PR #4167, opened 2019-02-12
Use goto_functiont::parameter_identifiers and remove goto_functiont::type

tautschnig:
Having two function type definitions (one in the symbol table and one in
goto_functiont) is just prone to become inconsistent. Instead, always look at
the symbol table when we do need type information. If all we care about are the
identifiers we can still use goto_functiont::parameter_identifiers.

~I'm also marking this RFC, @kroening in particular: it's quite painful to maintain consistency between `parameter_identifiers` and `code_typet::parameters()` - do we really need `parameter_identifiers`, or shouldn't we instead always use the information stored in the type?~

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4312, opened 2019-03-02
Store "is hidden" attribute of goto functions in the symbol table [blocks: #4167]

tautschnig:
There should only be a single place to hold type information, including
attributes, to ensure consistency. Future changes will remove the "type" member
of goto_functiont, making the type information stored in the symbol table the
single, authoritative source of information.

With this commit the information will remain available in both places, but all
read accesses only use the information in the symbol table.

This is a further, meant-to-be-non-breaking step towards removing `goto_functiont::type`, which #4167 tries to achieve.

### DIFF
--- a/src/goto-instrument/cover_filter.cpp
+++ b/src/goto-instrument/cover_filter.cpp
@@ -29,7 +29,7 @@ bool internal_functions_filtert::operator()(
   if(function.name == INITIALIZE_FUNCTION)
     return false;
 
-  if(goto_function.is_hidden())
+  if(function.is_hidden())
     return false;
 
   // ignore Java built-ins (synthetic functions)
--- a/src/goto-programs/goto_convert_functions.cpp
+++ b/src/goto-programs/goto_convert_functions.cpp
@@ -215,7 +215,10 @@ void goto_convert_functionst::convert_function(
   f.body.update();
 
   if(hide(f.body))
+  {
     f.make_hidden();
+    symbol_table.get_writeable_ref(identifier).set_hidden();
+  }
 
   lifetime = parent_lifetime;
 }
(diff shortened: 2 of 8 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a9cd4235cd6189b4a7ff4ad727b2445bf5dcc267).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/102911510

src/util/std_types.h
```diff
@@ -896,6 +896,16 @@ class code_typet:public typet
     set(ID_C_inlined, value);
   }
 
+  bool is_hidden() const
+  {
+    return get_bool(ID_C_hide);
+  }
+
+  void make_hidden()
```
smowton: `set_hidden` would be more usual

smowton (approved):
Though I note this is yet another candidate for `symbolt` having an `attributes` member for function attributes that aren't really part of its type signature.

tautschnig:
> Though I note this is yet another candidate for symbolt having an attributes member for function attributes that aren't really part of its type signature.

I guess we already have a ton of attributes in `symbolt`? I do know you suggested that rather than adding more binary flags we should have some data structure. I'm wondering whether, on this occasion, we should just use the existing `is_auxiliary` flag. We really have a very limited number of users of all this information here, and it's just advisory.

src/util/std_types.h
tautschnig: Done.

tautschnig:
> I'm wondering whether, on this occasion, we should just use the existing is_auxiliary flag. We really have a very limited number of users of all this information here, and it's just advisory.

I just did this. The `is_hidden`/`set_hidden` are now API functions of `symbolt` rather than `code_typet`.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: ebfab8d2c1b5b064828f0f3498cecb6f19a24cac).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103025037
Common spurious failures:

PR #3502, opened 2018-12-02
Enable perfect forwarding with irept/exprt/typet constructors

tautschnig:
This should enable use (with performance benefit) of rvalue references in
higher-level APIs.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4317, opened 2019-03-03
new constructor for function_application_exprt

kroening:
The new constructor
1) moves the arguments
2) enforces the type of the function to be mathematical_function
3) automatically sets the type of the application to be the codomain of the
function.
4) checks that the right number of arguments is given.

Note that the function symbol can't be moved during construction as its type
is needed.

### DIFF
--- a/src/util/mathematical_expr.cpp
+++ b/src/util/mathematical_expr.cpp
@@ -7,3 +7,17 @@ Author: Daniel Kroening, kroening@kroening.com
 \*******************************************************************/
 
 #include "mathematical_expr.h"
+#include "mathematical_types.h"
+
+function_application_exprt::function_application_exprt(
+  const symbol_exprt &_function,
+  argumentst _arguments)
+  : binary_exprt(
+      _function,
+      ID_function_application,
+      multi_ary_exprt(irep_idt(), std::move(_arguments), typet()),
+      to_mathematical_function_type(_function.type()).codomain())
+{
+  const auto &domain = to_mathematical_function_type(_function.type()).domain();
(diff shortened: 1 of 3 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 493324d26816c23e862b745a102ca668036f4b98).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/102948379

src/util/mathematical_expr.cpp
```diff
@@ -7,3 +7,17 @@ Author: Daniel Kroening, kroening@kroening.com
...
 #include "mathematical_expr.h"
+#include "mathematical_types.h"
+
+function_application_exprt::function_application_exprt(
+  const symbol_exprt &_function,
+  argumentst _arguments)
+  : binary_exprt(
+      _function,
+      ID_function_application,
+      multi_ary_exprt(irep_idt(), std::move(_arguments), typet()),
```
tautschnig, line 18: This looks very odd. Can we use `ID_arguments`? `exprt{ID_arguments, std::move(_arguments)}` would IMHO look cleaner. (But that's probably not possible until #3502 is merged.)

tautschnig (approved):

src/util/mathematical_expr.cpp
kroening: Sure, could use ID_arguments, but could also contemplate something like ID_tuple (which is what is happening here).

kroening: (Either should be a separate PR, and would need to change all constructors).

PR #4321, opened 2019-03-04
use a tuple for function arguments

kroening:
This is follow-up from PR #4317.

This introduces `tuple_exprt`.

`function_application_exprt` now uses `tuple_exprt` for the argument operand,
instead of an expression with empty ID.

### DIFF
--- a/src/util/mathematical_expr.h
+++ b/src/util/mathematical_expr.h
@@ -186,6 +186,15 @@ inline void validate_expr(const factorial_power_exprt &value)
   validate_operands(value, 2, "Factorial power must have two operands");
 }
 
+class tuple_exprt : public multi_ary_exprt
+{
+public:
+  explicit tuple_exprt(exprt::operandst operands)
+    : multi_ary_exprt(ID_tuple, std::move(operands), typet())
+  {
+  }
+};
+
 /// \brief Application of (mathematical) function
 class function_application_exprt : public binary_exprt
 {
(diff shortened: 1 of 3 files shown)


src/util/mathematical_expr.h
```diff
@@ -186,6 +186,15 @@ inline void validate_expr(const factorial_power_exprt &value)
   validate_operands(value, 2, "Factorial power must have two operands");
 }
 
+class tuple_exprt : public multi_ary_exprt
```
tautschnig, line 189: Would it be worth moving this to `std_expr.h` and use it in those places where we currently do use `ID_arguments`?

tautschnig (approved):

src/util/mathematical_expr.h
romainbrenguier: Alternatively I would suggest putting all the `.*_exprt` classes in a new module (separate from `util`) and have one class per header file, but that's a much bigger change.

tautschnig: I like this idea, but this is a very intrusive change -- a lot of code in `util` already depends on `std_expr.h`:
```
$ git grep 'include.*std_expr.h' src/util/ | wc -l
43
```

kroening: I think that the arguments/parameters of a C/C++ function are a very different thing from the arguments/parameters of a mathematical function. In particular, there is no tuple building going on in C/C++ or Java.

tautschnig: Ok.

kroening: @romainbrenguier getting slightly off-topic, we could separate the 'irep/expression' related things from the 'non irep utilities', say making a directory named "irep". There could be dummy headers for a transition period.

romainbrenguier: that sounds good

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: a8e67cbc118b52d9776b3e7d58a2bc2fdc13a9b8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103063401
Common spurious failures:

PR #4330, opened 2019-03-05
Add GDB pretty-printers

JohnDumbell:
A few of the more useful GDB pretty-printers and an installation python script to make their activation a little less painful for those unfamiliar.

Instructiont is there because when CLion attempts to pretty-print it, it goes around in an infinite loop because of the shared tree node. The actual print of the value is secondary to stopping the pretty-printers dying outright. 

If anyone knows how to do this without manually adding configuration to the .gdbinit let me know, because every option needed _something_ to go in there and it's not as seamless as I wanted.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,27 @@
+A collection of utility scripts and script applications.
+
+pretty-printers 
+------
+
+GDB:
+
+Pretty-printers for CBMC that enable easier debugging in IDEs and mitigate
+certain crashes due to the way some objects' memory is shared.
+
+Currently it deals with:
+* irep_idt
+* dstring
+* instructiont
+
+To install:
+
+1. Navigate to /pretty-printers/gdb.
+2. Run install.py with python 3+.
+3. If an exception occurs, create an empty '.gdbinit' file in your home
+    folder, and copy/paste the blob of code at the top of the install.py file.
+    
+The .gdbinit file is used by GDB during start-up to run any initial commands or 
+scripts, and the code injects the pretty-printers during that.
+
+Nothing else is required to get the pretty-printers to work, beside using 
+GDB to debug the code.
--- a/scripts/pretty-printers/gdb/auto_load.py
+++ b/scripts/pretty-printers/gdb/auto_load.py
@@ -1,0 +1,24 @@
+import gdb
+import pretty_printers
+
+
+def build_pretty_printer_collection():
+    
+    printers = gdb.printing.RegexpCollectionPrettyPrinter("CBMC")
+    
+    # First argument is the name of the pretty-printer, second is a regex match for which type
+    # it should be applied too, third is the class that should be called to pretty-print that type.
+    printers.add_printer(
+        'irep_idt', '^irep_idt', pretty_printers.DStringPrettyPrinter)
+    printers.add_printer(
+        'dstringt', '^dstringt', pretty_printers.DStringPrettyPrinter)
+    printers.add_printer(
+        'instructiont', '^goto_programt::instructiont', pretty_printers.InstructionPrettyPrinter)
+    return printers
+
+
+# If you change the name of this make sure to change install.py too.
+def load_pretty_printers():
+
+    # We aren't associating with a particular object file, so pass in None instead of gdb.current_objfile()
+    gdb.printing.register_pretty_printer(None, build_pretty_printer_collection(), replace=True)
--- a/scripts/pretty-printers/gdb/install.py
+++ b/scripts/pretty-printers/gdb/install.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+
+import os
+
+# This is the code that should be copied if you're applying the changes by hand.
+# Replace {0} with the path to this folder.
+file_contents = """
+python
+import sys
+import os
+
+pretty_printer_folder = '{0}'
+if os.path.exists(pretty_printer_folder):
+    sys.path.insert(1, pretty_printer_folder)
+    import auto_load
+    auto_load.load_pretty_printers()
+end
+"""
+
+
+def create_gdbinit_file():
+    """
+    Add or append to a .gdbinit file the python code to set-up cbmc pretty-printers.
+    """
+
+    print("Attempting to enable cbmc-specific pretty-printers.")
+
+    home_folder = os.path.expanduser("~")
+    if not home_folder:
+        print(home_folder + " is an invalid home folder, please manually create a .gdbinit file and apply the code.")
+        return
+
+    gdbinit_file = os.path.join(home_folder, ".gdbinit")
+    file_write_mode = 'w'
+    if os.path.exists(gdbinit_file):
+        print(".gdbinit file exists at " + gdbinit_file + "."
+              " Please type 'y' if you want to append the pretty-printer commands to it. Press any other key to exit.")
+        while True:
+            choice = input().lower()
+            if choice == 'y':
+                file_write_mode = 'a'
+                break
+            else:
+                print("Not appending to file. Exiting.")
+                return
+
+    if file_write_mode == 'w':
+        print("Creating .gdbinit file.")
+
+    print("Adding pretty-print commands to {0}.".format(gdbinit_file))
+    parent_directory = os.path.dirname(os.path.abspath(__file__))
+    try:
+        file = open(gdbinit_file, file_write_mode)
+        file.write(file_contents.format(parent_directory))
+        file.close()
+        print("Commands added.")
+    except:
+        print("Exception occured writing to file. Please apply changes manually.")
+
+
+if __name__ == "__main__":
+    create_gdbinit_file()
+
--- a/scripts/pretty-printers/gdb/pretty_printers.py
+++ b/scripts/pretty-printers/gdb/pretty_printers.py
@@ -1,0 +1,48 @@
+import gdb
+
+
+class DStringPrettyPrinter:
+    def __init__(self, val):
+        self.val = val
+
+    def to_string(self):
+        try:
+            raw_address = str(self.val.address)
+
+            # If it's ::empty, we know it's empty without going further.
+            if "::empty" in raw_address:
+                return ""
+
+            # Split the address on the first space, return that value
+            # Addresses are usually {address} {optional type_name}
+            typed_pointer = '({}*){}'.format(self.val.type, raw_address.split(None, 1)[0])
+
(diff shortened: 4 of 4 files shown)


src/util/dstring.h
```diff
@@ -73,6 +73,16 @@ class dstringt final
 
   // access
 
+  // In some situations we need to check for if this string has been
+  // correctly shared before doing an operation on it.
+
+  // Don't remove, used in python pretty-printers. (Also clang is wrong
+  // about it always being false)
+  bool uninitialized() const
```
tautschnig: Can we actually make this `private`? I don't think gdb will enforce access rules?

JohnDumbell: Can't due to the way python/gdb interacts (or at least the way it's being used here), but I can put DEPRECATED on it to make sure it's not called in the C++.

tautschnig: Ah, don't worry too much, there seems to be no risk in actually calling it, it would just be useless :-)

JohnDumbell:
Was able to remove the extra method in the C++ so this is now all in Python which is a bit nicer. Should also avoid the build errors/warnings too.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: fe4d51b52097f9cef36eb9e178ec6b116d090bdd).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103258014
Common spurious failures:

scripts/README.md
```diff
@@ -0,0 +1,21 @@
+A collection of utility scripts and script applications.
+
+pretty-printers
+------
+
+GDB:
+
+Pretty-printers for CBMC that enable easier debugging in IDE's and mitigate
```
owen-mc-diffblue: :pick: "IDE's" -> "IDEs" (this is a matter of style, but nowadays it's much more common to not use the apostrophe to pluralise an acronym)

scripts/pretty-printers/gdb/install.py
```diff
@@ -0,0 +1,59 @@
...
+
+    print("Attempting to enable cbmc-specific pretty-printers.")
+
+    home_folder = os.path.expanduser("~")
+    if not home_folder:
+        print(home_folder + " is an invalid home folder, please manually create a .gdbinit file and apply the code.")
+        return
+
+    gdbinit_file = os.path.join(home_folder, ".gdbinit")
+    file_write = 'w'
```
owen-mc-diffblue: :pick: `file_write` isn't a good variable name, especially as you call `file.write()` later. How about `file_write_mode`?

owen-mc-diffblue (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 46068e426756903bb935a273f350b6dde67b824b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103384883
Common spurious failures:

scripts/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+Pretty-printers for CBMC that enable easier debugging in IDEs and mitigate
+certain crashes due to the way some objects' memory is shared.
+
+Currently it deals with:
+* irep_idt
+* dstring
+* instructiont
+
+To install just run install.py and follow the instructions (if any are shown).
```
thk123: :question: from any particular directory? Answer: no but probably the docs should say this.

scripts/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+If installation fails it means simplistic introduction of the gdb commands was
+unable to be performed, so you'll have to open up the install script, copy the
```
thk123: :pick: It'd be great to elaborate on this a little.

scripts/pretty-printers/gdb/auto_load.py
```diff
@@ -0,0 +1,24 @@
+import gdb
```
thk123, line 1: :question: Can't comment on the actual file, but is the zero byte `__init__.py` needed?

scripts/README.md
thk123: No need - is explained in more detail in the python script

scripts/README.md
thk123: This should be
```suggestion
To install just run `pretty-printers/gdb/install.py` and follow the instructions (if any are shown).
```

Since this README is in scripts.

scripts/pretty-printers/gdb/install.py
```diff
@@ -0,0 +1,59 @@
+import os
```
thk123, line 3: this does not seem to have the correct permissions. And ideally it should have a flash bang at the top:
```bash
#!/usr/bin/env python
```

thk123: :question: What version of python should this be run with?

thk123: Seems like it only works with python 3, you should put some kind of check in I suppose

scripts/pretty-printers/gdb/install.py
```diff
@@ -0,0 +1,59 @@
...
+        print(home_folder + " is an invalid home folder, please manually create a .gdbinit file and apply the code.")
+        return
+
+    gdbinit_file = os.path.join(home_folder, ".gdbinit")
+    file_write_mode = 'w'
+    if os.path.exists(gdbinit_file):
+        print(".gdbinit file exists at " + gdbinit_file + "."
+              " Please type 'y' if you want to append the pretty-printer commands to it. Press any other key to exit.")
+        while True:
+            choice = input().lower()
```
thk123, line 39: I get:
```
Traceback (most recent call last):
  File "./pretty-printers/gdb/install.py", line 59, in <module>
    create_gdbinit_file()
  File "./pretty-printers/gdb/install.py", line 37, in create_gdbinit_file
    choice = input().lower()
  File "<string>", line 1, in <module>
NameError: name 'y' is not defined
```
When running python 2

scripts/pretty-printers/gdb/install.py
```diff
@@ -0,0 +1,59 @@
...
+    try:
+        file = open(gdbinit_file, file_write_mode)
+        file.write(file_contents.format(parent_directory))
+        file.close()
+        print("Commands added.")
+    except:
+        print("Exception occured writing to file. Please apply changes manually.")
+
+
+create_gdbinit_file()
```
thk123: :pick: good practise to wrap in a main method

thk123 (commented):
Sadly this still crashes for me on that example I gave you. For me, this isn't actually blocking, having it checked into source control and automatically update is better than the old approach of passing it around like a bootleg movie. 

Some wish list items:
 - no crashes :stuck_out_tongue: 
 - some tests?
 - perhaps a little detail on how to set it up in clion (just noting that CLion by default doesn't use the home directory `__init__` (I can provide this in a follow up PR if you like)

Couple of minor questions that I'd like to see answers/improvements

scripts/pretty-printers/gdb/install.py
thk123: If this isn't done, the docs should probably say 

> run `python3 install.py`

To indicate it can't be ran.

scripts/pretty-printers/gdb/install.py
```diff
@@ -0,0 +1,59 @@
...
+python
+import sys
+import os
+
+pretty_printer_folder = '{0}'
+if os.path.exists(pretty_printer_folder):
+    sys.path.insert(1, pretty_printer_folder)
+    import auto_load
+    auto_load.load_pretty_printers()
+end
```
thk123, line 17: :question: What is with this `end`? If pasted into an `__init__.py` then this end breaks things, is it ending the python section?

JohnDumbell: Yeah, specific syntax for the .gdbinit file to state the end of a python block (where 'python' states the beginning).

scripts/pretty-printers/gdb/auto_load.py
JohnDumbell: Just states the folder is a python module, though that's not overly important for this particular module, mostly just for standardisation / future.

JohnDumbell:
@thk123 You won't need to set anything up in CLion, just running the install script will set it up for gdb in general. I also have a feeling you might have old instances of this lurking around because this works for me on the case you showed me.

Because this has to be run inside gdb itself tests aren't really on the cards for it either, worst-case if something breaks you just won't get pretty-printed names. The segfault is something that should be considered an exceptional situation and absolutely fixed ASAP.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 28cc20168174d276dd8fd1a9cc7ec15af2e260fc).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103545691
Common spurious failures:

thk123:
When I tried to show this to John, the problem had evaporated, so let's assume I messed up something in configuration :tada:

scripts/pretty-printers/gdb/install.py
JohnDumbell: Won't put in a check, but I've amended the installation instructions to be more precise which should cover all your other comments as well.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 655e93689eb4e46326b7919ee5bdacbdf8062a3c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103908716

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 8be1bc1b47184c0f57cd2a3211e77e5627f1a6b2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103910923
Common spurious failures:

scripts/pretty-printers/gdb/install.py
thk123: Sad to see no flash bang, but the docs change is sufficient.

thk123 (approved):
Hoorary for more reliable printers that are actually checked in

scripts/pretty-printers/gdb/install.py
JohnDumbell: Oh, that's in. Should it just be `#!/usr/bin/env python3`for explicit python 3? (Not super familiar with the syntax, but it's relatively harmless, I was just talking about any other manual check for Python version)

thk123: Ah my bad - I looked at the wrong file!

PR #4334, opened 2019-03-06
Add an is_global() predicate to symbolt.

NlightNFotis:
This is a convenience predicate. Before that we had to remember to check a couple of properties. This makes it easier to handle by raising the abstraction level, and centralising the property checking. The properties checked have been reverse engineered from the properties that global symbols in the symbol table hold.

### DIFF
--- a/src/util/symbol.h
+++ b/src/util/symbol.h
@@ -123,6 +123,12 @@ public:
     return is_auxiliary;
   }
 
+  /// Returns true iff the symbol belongs to the global scope.
+  bool is_global() const
+  {
+    return is_static_lifetime && is_lvalue;
+  }
+
   /// Mark a symbol for internal use. This is advisory and may be utilized,
   /// e.g., to filter output.
   void set_hidden()


src/util/symbol.h
```diff
@@ -123,6 +123,12 @@ class symbolt
     return is_auxiliary;
   }
 
+  /// Returns true iff the symbol belongs to the global scope.
+  bool is_global() const
+  {
+    return is_static_lifetime && is_lvalue;
```
tautschnig, line 129: Well ... except this isn't actually right. This only works for non-type symbols. Maybe call it `is_global_object` and also check `is_type`?

tautschnig: And also it doesn't work for code-typed symbols (which aren't lvalues).

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 0ed99d8529758bb631519abf7010fc5822d902df).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103360850
Common spurious failures:

PR #4199, opened 2019-02-15
Move instead of merge symex L2 renaming map when possible

JohnDumbell:
This PR splits the level 2 renaming map into two: a global map whose state is kept across all symex runs, and a local one that only holds the generation counts of something along a particular branch (or specifically, a statet).

It works like this: on reading the current generation, the local map will always be used. When a generation is increased, we query the global map for the currently-taken generation number, increment that by one, then assign the local maps generation number as that. With this, we don't have to merge maps at each branch if for some reason a particular branch can't be taken, as generation numbers will always be unique. Downside is that generation numbers might look strange if symex has many paths to take.

Example of this: path symex, previously, would have taken the next free number on that particular path. With the global name map this is actually now the next free number across _all_ paths. So you could see:

name#1
name#2
name#54

Which might throw people off. You fix this by making the global name map only 'global' to an individual run of symex, but for now there shouldn't be any problems, and the change is minuscule anyway.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4343, opened 2019-03-07
Fix symex of functions that may throw

smowton:
Java functions that may throw do not always define `#return_value` (cf. the C front-end ensures it always defines it, even if the source code does not). Previously that situation was handled inconsistently: the first call on any such path would propagate the value-set on the normal-return path but refuse to constant-propagate, and the second call on a path to the same function would refuse to do either (using `return_value@object` and `return_value#0` to represent the possibility that the return-value is undefined in each case).

This partially addresses that by (a) using the same merge logic (keep the defined version) for `#return_value` variables as for locals, (b) erasing from the value-set when a variable dies rather than writing a failed object (consistent with the case where the variable has never been written at all), and (c) explicitly initialising nondet variables in the C front-end to permit (a).

With this change we can propagate the value-set to the normal-return path every time, and constant propagate on the first call to a given function.

One problem remains: on the second call constant propagation still fails, because we're merging an undefined generation of the `#return_value` variable (the one created by the previous `DEAD` instruction) but that generation is not zero. That case can be addressed once #4199 lands by having `DEAD` set the generation to zero (which is never defined) rather than incrementing it.

This is now based on #4386 -- please only review the last 4 commits.


smowton:
@tautschnig @peterschrammel this should now be ready for review, but is based on #4386 -- please only review final 4 commits.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: f0d7ae5a1558a78c4e0c6e129bbec23d230aa2f4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104456360

src/linking/static_lifetime_init.cpp
```diff
@@ -77,14 +69,18 @@ static_lifetime_init(const irep_idt &identifier, symbol_tablet &symbol_table)
...
 
-  if(symbol.value.id() == ID_nondet)
-  {
-    return {}; // do not initialize
-  }
-
   exprt rhs;
 
-  if(symbol.value.is_nil())
+  if((symbol.value.is_nil() && symbol.is_extern) ||
```
tautschnig, line 74: Does this also cover `argc` and `argv`?

tautschnig (approved):

src/linking/static_lifetime_init.cpp
smowton: Checked: no, they are initialised by dangling globals `argc'` and `argv'` created in `ansi_c_entry_point`, *after* `static_lifetime_init` has been called. I tried initialising them explicitly: this worked fine except for test `goto-instrument/harness1`, which tries to dump a C harness to replicate a particular situation. That led to a line like `argv' = nondet_0();`, which is not legal C.

So it's not trivial to fix this, but I think we're safe: there's no way for `argc'` and similar to be assigned, so they should never be merged. I agree it would be better to explicitly init them, but given the rabbit hole we appear to be diving down here I'm inclined to leave it to a subsequent PR.

src/goto-symex/symex_dead.cpp
```diff
@@ -22,26 +22,13 @@ void goto_symext::symex_dead(statet &state)
   const code_deadt &code = instruction.get_dead();
 
   ssa_exprt ssa = state.rename_ssa<L1>(ssa_exprt{code.symbol()}, ns);
-
-  // in case of pointers, put something into the value set
```
peterschrammel, line 26: (commit message) convergeance -> convergence

jbmc/regression/jbmc/throwing-function-return-value/test.desc
```diff
@@ -0,0 +1,18 @@
...
+--function Test.main --show-vcc
+java::Test\.main:\(Z\)V::14::t1!0@1#\d+ = address_of\(symex_dynamic::dynamic_object\d+\)
+java::Test\.main:\(Z\)V::9::x!0@1#\d+ = 5 \+ java::Test\.main:\(Z\)V::9::x!0@1#\d+
+^EXIT=0$
+^SIGNAL=0$
+--
+return_value!0#0
+java::Sub\.g:\(\)
+--
+This checks that when a function may throw, we can nontheless constant-propagate
```
peterschrammel: nonetheless

peterschrammel (approved):
LGTM

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d341910754d9a522805b2ae536ee569e123ad086).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104994782

PR #3962, opened 2019-01-27
add an implementation of std::variant<...>

kroening:
This appears to require update 3 of MSVC 2015, which may be an issue.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [X] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4344, opened 2019-03-07
Variant submodule

thk123:

This is a proposed replacement for  #3962 that I hope might be less controversial. It provides the following advantages:
 -  tests provided by the original repo are run in our CI
 - easy to find the source of variation

I didn't spend much time on this, so if people don't like it, no harm no fowl, but would quite like to be able to use variant.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "jbmc/lib/java-models-library"]
 	path = jbmc/lib/java-models-library
 	url = https://github.com/diffblue/java-models-library.git
+[submodule "lib/variant"]
+	path = lib/variant
+	url = https://github.com/mpark/variant.git
--- a/scripts/run_variant_tests.sh
+++ b/scripts/run_variant_tests.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd lib/variant
+mkdir -p build
+cd build
+cmake -DMPARK_VARIANT_INCLUDE_TESTS="mpark" -DCMAKE_CXX_FLAGS="-std=c++11" ..
+cmake --build . -- $@
+ctest --output-on-failure
--- a/src/util/module_dependencies.txt
+++ b/src/util/module_dependencies.txt
@@ -1,6 +1,8 @@
 big-int
 mach # system
 malloc # system
+mpark
 nonstd
 sys # system
 util
+
--- a/src/util/variant.h
+++ b/src/util/variant.h
@@ -1,0 +1,18 @@
+/*******************************************************************\
+
+Module: typedef for variant class template. To be replaced with
+        std::variant once C++17 support is enabled
+
+Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+#ifndef CPROVER_UTIL_VARIANT_H
+#define CPROVER_UTIL_VARIANT_H
+
+#include <mpark/variant.hpp>
+
+template <typename... Ts>
+using variantt = mpark::variant<Ts...>;
+
+#endif // CPROVER_UTIL_VARIANT_H
--- a/unit/util/variant.cpp
+++ b/unit/util/variant.cpp
@@ -1,0 +1,12 @@
+// Copyright 2016-2019 Diffblue Limited. All Rights Reserved.
+
+#include <testing-utils/use_catch.h>
+#include <util/variant.h>
+
+TEST_CASE("Ensure variant has been included", "[core][util][variant]")
+{
+  variantt<int, float> some_type = 1;
+  REQUIRE(some_type.index() == 0);
+  some_type = 1.0f;
+  REQUIRE(some_type.index() == 1);
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -326,6 +326,7 @@ install:
   - make -C src/cpp library_check
   - make -C src "CXX=${COMPILER} ${EXTRA_CXXFLAGS}" -j3
   - make -C jbmc/src "CXX=${COMPILER} ${EXTRA_CXXFLAGS}" -j3
+  - ./scripts/run_variant_tests.sh -j3
 
 script:
   - if [ -e bin/gcc ] ; then export PATH=$PWD/bin:$PATH ; fi ;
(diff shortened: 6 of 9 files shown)


src/util/variant.h
```diff
@@ -0,0 +1,18 @@
+/*******************************************************************\
+
+Module: typedef for optional class template. To be replaced with
+        std::optional once C++17 support is enabled
```
martin-cs: Should this be in 'nonstd/' with the existing `optional` code?

thk123: Since it is a submodule, I don't think I can change the namespace it is in? Or do you mean, making the `using nostd::variantt = ...` - the `optionalt` is declared in the global scope so I figured this was more consistent. 

 But your comment highlighted a copy-pasta error in the docs - [fixed](https://github.com/diffblue/cbmc/compare/3355d7708c0782fe4ba75e7233791db7532c1b1b..f0db1cda21b1b7af01c2028533a13b7e83d0bf73)

martin-cs: Thinking about it; just a comment saying ```// Swap for std::variant when switching to C++17``` or something.

thk123: Not sure I follow, comment currently reads (typo not withstanding):

>  To be replaced with std::variant once C++17 support is enabled

Is your suggestion your wording is clearer? Happy to change if that is indeed what you mean, but for me they are equivalent

thk123:
This is failing due to Makefile not linking the variant in util.

Before I spend some time figuring out how `Makefile`s work again, I would appreciate some indication from @tautschnig / @kroening / @peterschrammel as to whether this approach would be preferable or acceptable way of introducing a `variant`.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "jbmc/lib/java-models-library"]
 	path = jbmc/lib/java-models-library
 	url = https://github.com/diffblue/java-models-library.git
+[submodule "lib/variant"]
```
tautschnig, line 4: Instead of introducing a new top-level folder, can we place this in `nonstd/variant` instead of `lib/variant`?

scripts/run_variant_tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd lib/variant
+mkdir -p build
+cd build
+cmake -DMPARK_VARIANT_INCLUDE_TESTS="mpark" -DCMAKE_CXX_FLAGS="-std=c++11" ..
```
tautschnig, line 8: Are we sure that `cmake` is installed prior to running this (as this is being run in all Travis jobs)?

src/util/module_dependencies.txt
```diff
@@ -1,6 +1,8 @@
 big-int
 mach # system
 malloc # system
+mpark
 nonstd
 sys # system
 util
+
```
tautschnig, line 8: Not needed

src/util/variant.h
```diff
@@ -0,0 +1,18 @@
...
+        std::variant once C++17 support is enabled
+
+Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+#ifndef CPROVER_UTIL_VARIANT_H
+#define CPROVER_UTIL_VARIANT_H
+
+#include <mpark/variant.hpp>
```
tautschnig, line 13: I suppose this works with CMake, but shouldn't this be `#include <lib/variant/mpark/variant.hpp>`? If that, however, does _not_ work with `CMake` then we'd need to update `INCLUDES` in any `Makefile` that wants to use this.

kroening:
I am concerned about the following: so far, we have been able to build cbmc from a tarball (I am aware that this doesn't work with jbmc). We would now introduce a compile-time dependency on git, plus the ability to clone that repo.

unit/util/variant.cpp
```diff
@@ -0,0 +1,12 @@
+// Copyright 2016-2019 Diffblue Limited. All Rights Reserved.
```
peterschrammel, line 1: Wrong copyright header

smowton:
@kroening subtree rather than submodule would fix that

LAJW (approved):

scripts/run_variant_tests.sh
thk123: Hmm good point - what would you suggest - the project is in CMake not Make? (one option of course being: abandon this approach!)

src/util/variant.h
thk123: So only the util includes need to change, since all other users should use the `variantt` wrapper. But happy to do that if that's the preference.

thk123:
@kroening lmk if @smowton suggestion would work for you and I can swap this over to use a subtree.

scripts/run_variant_tests.sh
tautschnig: I don't think there's a problem asking for `cmake` to be installed. Just make sure it's documented/actually done in CI configurations.

src/util/variant.h
tautschnig: Well, but the `variant.h` file uses this include, so if I'm in directory `X` and have a file `x/y.cpp` that includes `variant.h` then the preprocessor still needs to know how to resolve that include.

peterschrammel:
If dependency on a repo that is not under our control is considered a problem we could also fork it into the org.

> we have been able to build cbmc from a tarball 

Note that CBMC already now requires an extra step (minisat) to build it from a tarball - tracking https://github.com/niklasso/minisat (or a fork of it) through a submodule would be more convenient.

tautschnig:
As a rather different suggestion: have people considered to just build using `-std=c++17` and just use `std::variant`? That's not something we can realistically do for CBMC at the moment, but if the main interest in using `variant` is in downstream projects then those projects could just change their compiler requirements.

LAJW:
> just build using -std=c++17

I'd love that, and nothing technical is stopping us from doing so. But someone doesn't want that.

Nevertheless, availability of variant in CBMC wouldn't be a bad thing.

smowton:
If you want a self-sufficient repo but also the ability to easily track what other-repo commit we're using, just subtree it. That would satisfy @kroening's concern

NlightNFotis:
Closing this PR on the basis of:

1) Inactivity for about 3 years now.
2) Significant divergence from mainline (CI is very outdated and merge conflicts exist).

If you believe this has been in err, feel free to reopen the issue.

PR #4347, opened 2019-03-07
Add ranget drop and skip methods

romainbrenguier:
This adds two utility methods to ranget which I would find useful. 
For instance we could iterate on a ranget with a while loop instead of a range-for: `while(!range.empty()) { ....; range.drop(1); ...}`

### DIFF
--- a/src/util/range.h
+++ b/src/util/range.h
@@ -341,6 +341,24 @@ public:
     return begin_value == end_value;
   }
 
+  /// Return an new range containing the same elements except for the first
+  /// \p count elements.
+  /// If the range has fewer elements, returns an empty range.
+  ranget<iteratort> drop(std::size_t count) &&
+  {
+    for(; count > 0 && begin_value != end_value; --count)
+      ++begin_value;
+    return ranget<iteratort>{std::move(begin_value), std::move(end_value)};
+  }
+
+  /// Return an new range containing the same elements except for the first
+  /// \p count elements.
+  /// If the range has fewer elements, returns an empty range.
+  ranget<iteratort> drop(std::size_t count) const &
+  {
+    return ranget<iteratort>{begin(), end()}.drop(count);
+  }
+
   iteratort begin() const
   {
     return begin_value;
--- a/src/util/range.h
+++ b/src/util/range.h
@@ -289,7 +289,8 @@ struct ranget final
 public:
   using value_type = typename iteratort::value_type;
 
-  ranget(iteratort begin, iteratort end) : begin_value(begin), end_value(end)
+  ranget(iteratort begin, iteratort end)
+    : begin_value(std::move(begin)), end_value(std::move(end))
   {
   }
 
--- a/unit/util/range.cpp
+++ b/unit/util/range.cpp
@@ -57,6 +57,45 @@ SCENARIO("range tests", "[core][util][range]")
       ++it;
       REQUIRE(it == filtered_range.end());
     }
+
+    THEN("Drop first 2 elements")
+    {
+      auto range = make_range(list);
+      auto drop_range = range.drop(2);
+      auto it = drop_range.begin();
+      REQUIRE(*it == "acdef");
+      drop_range = std::move(drop_range).drop(1);
+      REQUIRE(drop_range.empty());
+      // Check the original is unmodified
+      REQUIRE(!range.empty());
+      REQUIRE(*range.begin() == "abc");
+    }
+    THEN("Drop first 5 elements")
+    {
+      auto range = make_range(list);
+      auto skip_range = range.drop(5);
+      REQUIRE(skip_range.empty());
+      // Check the original is unmodified
+      REQUIRE(!range.empty());
+      REQUIRE(*range.begin() == "abc");
+    }
+    THEN("Drop first 2 elements, move version")
+    {
+      auto range = make_range(list);
+      range = std::move(range).drop(2);
+      REQUIRE(!range.empty());
+      auto it = range.begin();
+      REQUIRE(*it == "acdef");
+      range = std::move(range).drop(1);
+      REQUIRE(range.empty());
+    }
+    THEN("Drop first 5 elements, move version")
+    {
+      auto range = make_range(list);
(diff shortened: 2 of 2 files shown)


smowton (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 73cbe054fcab5fba6d8e15c3bfabd4cfd705ed9c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103580996
Common spurious failures:

src/util/range.h
```diff
@@ -341,6 +341,18 @@ struct ranget final
     return begin_value == end_value;
   }
 
+  /// Drop the first \p count elements.
+  /// If the range has fewer elements, drops everything.
+  void drop(std::size_t count)
+  {
+    std::size_t i = 0;
+    while(i < count && begin_value != end_value)
```
tautschnig: Nit pick, but wouldn't this make for a very nice `for` loop? As is, it looks a bit like assembler programming.

src/util/range.h
```diff
@@ -353,6 +353,16 @@ struct ranget final
     }
   }
 
+  /// Return an new range containing the same elements except for the first
+  /// \p count elements.
+  /// If the range has fewer elements, returns an empty range.
+  ranget<iteratort> skip(std::size_t count) const
```
tautschnig: I don't find the names `drop` and `skip` entirely intuitive. I am wondering whether using longer, more unambiguous names would be better?

unit/util/range.cpp
```diff
@@ -57,6 +57,34 @@ SCENARIO("range tests", "[core][util][range]")
...
+    }
+    THEN("Drop first 5 elements")
+    {
+      auto range = make_range(list);
+      range.drop(5);
+      REQUIRE(range.empty());
+    }
+    THEN("Skip first 2 elements")
+    {
+      auto range = make_range(list).skip(2);
```
tautschnig: I think it would be better to have a test where you also test the original range to make that indeed `skip` does not modify anything, just generates a copy.

romainbrenguier:
@smowton I made a slight change to the drop function: by returning a Boolean we can avoid some redundant iterator comparison and for instance write `while(range.drop(1)) { ... }`

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 39146d29b14055587994e30856c2bb9827224ee3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103603097
Common spurious failures:

src/util/range.h
romainbrenguier: Maybe. I'm open to suggestions. `drop` is also used by range-v3, so we should maybe keep this one.

tautschnig: Whatever is used elsewhere I'm happy with, just stick with that terminology then.

src/util/range.h
```diff
@@ -341,6 +341,32 @@ struct ranget final
     return begin_value == end_value;
   }
 
+  /// Drop the first \p count elements.
+  /// If the range has fewer elements, drops everything.
+  /// Similar to \ref skip but modifies the range in-place.
+  /// \return true if the range still contains element after the drop
+  bool drop(std::size_t count)
+  {
+    for(std::size_t i = 0; i < count; ++i)
```
tautschnig: You don't actually need `i`, do you? `for( ; count > 0; --count) ...` would do the trick.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: ac0932646704256dcb73f4d8c9f6a9caaf6a7390).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103676521
Common spurious failures:

src/util/range.h
thomasspriggs: `drop` and `skip` are both fine names, but they are kinda synonyms. I think we should choose one or the other, but not both. The name used in `.net`'s linq is `skip` - https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.skip?view=netframework-4.7.2
The name in haskell would be drop - http://zvon.org/other/haskell/Outputprelude/drop_f.html
Because C++ supports overloading we could actually have 3 different overloads, all with the same name if we wanted. A mutation overload, a `const &` overload and an r-value overload. I would suggest keeping just a `const &` and the r-value overload, in order to reduce the need to reason about mutation.

src/util/range.h
```diff
@@ -341,6 +341,26 @@ struct ranget final
     return begin_value == end_value;
   }
 
+  /// Drop the first \p count elements.
+  /// If the range has fewer elements, drops everything.
+  /// Similar to \ref skip but modifies the range in-place.
+  /// \return true if the range still contains element after the drop
+  ranget<iteratort> drop(std::size_t count) &&
+  {
+    for(; count > 0 && begin_value != end_value; --count)
```
thomasspriggs, line 350: :bulb: Could this be defined in terms of `std::advance` and `std::distance`? I think you could advance the begin iterator by the `std::min` of `count` and the distance between the begin and the end iterators.

romainbrenguier: Maybe, but I don't know if it's worth it. Some distance functions work by iterating from the current position to the end, so using that plus advance would end up being less efficient.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 2f9f9e21f6d82f96554e9da9c3c29cab16b3ab40).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103685752
Common spurious failures:

romainbrenguier:
Following @thomasspriggs suggestions, I made the two versions be named `drop` and return a new range, the difference between the two is that one works on ranges that are rvalues

tautschnig (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 962c9f4bff2006e7810b7444a088586db77bbfbe).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103695677

PR #4352, opened 2019-03-08
Refactor code related to bitvector type min/max values

antlechner:
This PR includes no functional changes, only refactoring.

### DIFF
--- a/jbmc/src/java_bytecode/java_object_factory.cpp
+++ b/jbmc/src/java_bytecode/java_object_factory.cpp
@@ -351,19 +351,6 @@ public:
   }
 };
 
-/// Get max value for an integer type
-/// \param type:
-///   Type to find maximum value for
-/// \return Maximum integer value
-static mp_integer max_value(const typet &type)
-{
-  if(type.id() == ID_signedbv)
-    return to_signedbv_type(type).largest();
-  else if(type.id() == ID_unsignedbv)
-    return to_unsignedbv_type(type).largest();
-  UNREACHABLE;
-}
-
 /// Initialise length and data fields for a nondeterministic String structure.
 ///
 /// If the structure is not a nondeterministic structure, the call results in
@@ -458,7 +445,9 @@ void initialize_nondet_string_fields(
   code.add(code_assumet(binary_relation_exprt(length_expr, ID_ge, min_length)));
 
   // assume (length_expr <= max_input_length)
-  if(max_nondet_string_length <= max_value(length_expr.type()))
+  if(
+    max_nondet_string_length <=
+    to_integer_bitvector_type(length_expr.type()).largest())
   {
     exprt max_length =
       from_integer(max_nondet_string_length, length_expr.type());
--- a/src/ansi-c/ansi_c_entry_point.cpp
+++ b/src/ansi-c/ansi_c_entry_point.cpp
@@ -377,18 +377,8 @@ bool generate_ansi_c_start_function(
         }
 
         // assume envp_size is INTMAX-1
-        mp_integer max;
-
-        if(envp_size_symbol.type.id()==ID_signedbv)
-        {
-          max=to_signedbv_type(envp_size_symbol.type).largest();
-        }
-        else if(envp_size_symbol.type.id()==ID_unsignedbv)
-        {
-          max=to_unsignedbv_type(envp_size_symbol.type).largest();
-        }
-        else
-          UNREACHABLE;
+        const mp_integer max =
+          to_integer_bitvector_type(envp_size_symbol.type).largest();
 
         exprt max_minus_one=from_integer(max-1, envp_size_symbol.type);
 
--- a/src/util/std_types.h
+++ b/src/util/std_types.h
@@ -1176,10 +1176,21 @@ public:
   {
   }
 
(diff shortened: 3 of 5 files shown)


martin-cs (approved):
Seems reasonable and code reduction patches are always a good thing!

src/util/std_types.h
```diff
@@ -1071,6 +1071,12 @@ class bitvector_typet : public typet
     set(ID_width, narrow_cast<long long>(width));
   }
 
+  mp_integer smallest() const;
+  mp_integer largest() const;
+  constant_exprt smallest_expr() const;
+  constant_exprt zero_expr() const;
+  constant_exprt largest_expr() const;
```
tautschnig: This is brilliant refactoring, thank you! Yet can you please hide these in `pointer_typet`? A `pointer_typet` happens to be a `bitvector_typet`, but really they must not be used (not even `zero_expr`) on a `pointer_typet`.

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -458,7 +446,7 @@ void initialize_nondet_string_fields(
   code.add(code_assumet(binary_relation_exprt(length_expr, ID_ge, min_length)));
 
   // assume (length_expr <= max_input_length)
-  if(max_nondet_string_length <= max_value(length_expr.type()))
+  if(max_nondet_string_length <= bv_spect(length_expr.type()).max_value())
```
tautschnig: How about `to_bitvector_type(length_expr.type()).largest()`?

src/ansi-c/ansi_c_entry_point.cpp
```diff
@@ -377,18 +378,7 @@ bool generate_ansi_c_start_function(
...
-        {
-          max=to_signedbv_type(envp_size_symbol.type).largest();
-        }
-        else if(envp_size_symbol.type.id()==ID_unsignedbv)
-        {
-          max=to_unsignedbv_type(envp_size_symbol.type).largest();
-        }
-        else
-          UNREACHABLE;
+        const mp_integer max = bv_spect(envp_size_symbol.type).max_value();
```
tautschnig: As above `to_bitvector_type(envp_size_symbol.type).largest()`?

tautschnig (approved):

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 23c17f1f6a97547f5cf840661bf77e59d8b2e30b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103691044

src/util/std_types.h
antlechner: Ah, I overlooked that there are actually quite a few classes that inherit from `bitvector_typet`. Besides the three that were already mentioned, there are `bv_typet`, `fixedbv_typet`, `floatbv_typet`, `c_bit_field_typet` and `c_bool_typet`. However these functions will hit an `UNREACHABLE` if the type id is anything other than `ID_unsignedbv` or `ID_signedbv`. Would it make sense to introduce a new class that inherits from `bitvector_typet` and that both `signedbv_typet` and `unsignedbv_typet` inherit from? Any suggestions for what such a class should be called?

jbmc/src/java_bytecode/java_object_factory.cpp
antlechner: Good point, why does the class `bv_spect` even (publicly) exist if this functionality is already made available by the typet?
I changed it to your suggestion for now. Depending on what you think about my previous question I might change `to_bitvector_type` to something more specific.
I also found another place where `bv_spect` was used to get the max value, and replaced this with `bitvector_type::largest` as well.

src/util/std_types.h
kroening: smallest()/largest() only make sense for the integer ones -- so that could be integer_bitvector_typet. smallest_expr, zero_expr, largest_expr() also work on fixedbv and floatbv; these could be a numeric_bitvector_typet.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c8f84588a66faca5c07ecf36423b15264e551e59).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103716955

src/util/std_types.h
antlechner: That makes sense, thank you! I only created the `integer_bitvector_typet` class as all five methods were previously only defined for the signedbv/unsignedbv types, and I added some documentation and a comment mentioning that the methods returning an exprt would make sense for fixedbv/floatbv as well in case we need them there in the future.

jbmc/src/java_bytecode/java_object_factory.cpp
antlechner: Following @kroening 's comment above, this is now `to_integer_bitvector_type(...)` (and similarly for your other comment).

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: f313fdf5ea16757a64bbd3fa3761b17aeb0f596b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103736558

PR #4353, opened 2019-03-08
Numeric cast from integral type to integral types

romainbrenguier:
This defines numeric_cast and numeric_castv for converting from one
integral type to another integral type.
Only handle the case where they are both signed or both unsigned for
now.

The numeric conversion is already used in CBMC code but until now it
would have implicitely converted the value to an mp_integer and the
mp_integer back to an integral type.

### DIFF
--- a/src/util/arith_tools.h
+++ b/src/util/arith_tools.h
@@ -121,6 +121,54 @@ public:
   }
 };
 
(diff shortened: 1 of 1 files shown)


romainbrenguier:
Closing this as using `narrow` seems like a better approach for doing this kind of conversions

PR #4355, opened 2019-03-08
Move get_null_assignment out of object factory

antlechner:
My goal is to tidy up the Java object factory a bit as I'm hoping to be able to re-use some of its functionality and it currently has a few helper functions that should probably live somewhere else.
The target of this PR is the simple `get_null_assignment`. My suggestion with this PR is to move it to a new util file `util/code_util.h`, mirroring the already existing `util/expr_util.h`. I couldn't find a better place to put it. However, if reviewers can think of a better place for it, or if they think I should just inline it because it's so simple, let me know and I'm happy to change it. All I want is for it to go out of the object factory. :slightly_smiling_face: 

### DIFF
--- a/src/util/Makefile
+++ b/src/util/Makefile
@@ -6,6 +6,7 @@ SRC = allocate_objects.cpp \
       byte_operators.cpp \
       c_types.cpp \
       cmdline.cpp \
+      code_util.cpp \
       config.cpp \
       cout_message.cpp \
       decision_procedure.cpp \
--- a/src/util/code_util.cpp
+++ b/src/util/code_util.cpp
@@ -13,8 +13,7 @@ code_assignt get_null_assignment(
   const pointer_typet &ptr_type,
   const source_locationt &loc)
 {
-  null_pointer_exprt null_pointer_expr(ptr_type);
-  code_assignt code(expr, null_pointer_expr);
+  code_assignt code{expr, null_pointer_exprt{ptr_type}};
   code.add_source_location() = loc;
   return code;
 }
--- a/src/util/code_util.cpp
+++ b/src/util/code_util.cpp
@@ -1,0 +1,20 @@
+/*******************************************************************\
+
+Module: codet utilities
+
+Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+#include "code_util.h"
+
+code_assignt get_null_assignment(
+  const exprt &expr,
+  const pointer_typet &ptr_type,
+  const source_locationt &loc)
+{
+  null_pointer_exprt null_pointer_expr(ptr_type);
+  code_assignt code(expr, null_pointer_expr);
+  code.add_source_location() = loc;
+  return code;
+}
--- a/jbmc/src/java_bytecode/java_object_factory.cpp
+++ b/jbmc/src/java_bytecode/java_object_factory.cpp
@@ -8,6 +8,7 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "java_object_factory.h"
 
+#include <util/code_util.h>
 #include <util/expr_initializer.h>
 #include <util/nondet.h>
 #include <util/nondet_bool.h>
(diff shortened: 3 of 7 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4562e6a2fa854786b67dbbd89025ffef06c0a4a3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103722584

src/util/code_util.cpp
```diff
@@ -0,0 +1,20 @@
+/*******************************************************************\
+
+Module: codet helper utilities
```
romainbrenguier: :pick: "helper utilities" sounds a redundant to me

src/util/code_util.cpp
```diff
@@ -0,0 +1,20 @@
...
+\*******************************************************************/
+
+#include "code_util.h"
+
+code_assignt get_null_assignment(
+  const exprt &expr,
+  const pointer_typet &ptr_type,
+  const source_locationt &loc)
+{
+  null_pointer_exprt null_pointer_expr(ptr_type);
```
romainbrenguier: The preferred style for constructor call is with braces `{ }`
This can also be inlined in the next line

smowton (approved):
Seems fine to me

src/util/Makefile
```diff
@@ -5,6 +5,7 @@ SRC = allocate_objects.cpp \
       bv_arithmetic.cpp \
       byte_operators.cpp \
       c_types.cpp \
+      code_util.cpp \
```
tautschnig: Nit pick: lexicographically sorts one line later (cmdline < code_util).

src/util/code_util.cpp
```diff
@@ -0,0 +1,20 @@
...
+  code_assignt code(expr, null_pointer_expr);
```
tautschnig: While you're at it ...
1) I'd use `code_assignt code{expr, null_pointer_exprt{ptr_type}};`
2) It would be nice to have a `PRECONDITION(expr.type() == ptr_type);` - or even get rid of `ptr_type`, because it must be the same as `expr.type()`.

tautschnig (approved):

romainbrenguier (commented):
I'm not convinced we need a helper function for that. If you add a constructor with a location argument we could write `code_assignt{expr, null_pointer_exprt{ptr_type}, loc}` which I think is more explicit than `get_null_assignment(expr, ptr_type, loc)`. I'm affraid people wouldn't use the helper function because they wouldn't think of looking in `code_util.h` for it, the util folder already contains to much. The situation would be a bit different if we had a `code_assign.h` where the class is defined and where you can put helper functions.  I also think the constructors could be used in many more places.

antlechner:
@romainbrenguier I was wondering if defining a new constructor for this might be better. I'll do that in a separate PR and see what reviewers think of it, and I'll mark this one as "don't merge" until a decision has been made.

src/util/code_util.cpp
antlechner: Changed to "codet utilities"

src/util/code_util.cpp
antlechner: I did 1) in a separate commit. As for 2), I'm guessing `pointer_type` could generally be more specific than `expr_type()` and I'm not sure if that could cause any problems with a null assignment. In any case, I'll raise another PR with Romain's suggestion of making a new constructor to replace the helper function, and not waste too much time on this PR until I have reviews on the alternative solution.

tautschnig: A `code_assignt` where `.lhs().type() != .rhs().type()` violates invariants. You could still have that more-specific type, but should then add `typecast_exprt::conditional_cast` to make sure a typecast is added as if and when it is needed.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 500a586e6b06792094b6d00b335039f688727482).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104086225

owen-mc-diffblue:
I agree that a new constructor would be better

antlechner:
@owen-jones-diffblue A PR for that is here: https://github.com/diffblue/cbmc/pull/4373

antlechner:
It looks like https://github.com/diffblue/cbmc/pull/4373 will replace this PR so I'm closing it.

PR #4357, opened 2019-03-08
Move total_ and remaining_vccs to goto_symex_statet

smowton:
These never belonged in goto_statet: we don't do anything with them in merge_goto.

### DIFF
--- a/src/goto-symex/goto_state.h
+++ b/src/goto-symex/goto_state.h
@@ -57,9 +57,6 @@ public:
   /// Threads
   unsigned atomic_section_id = 0;
 
-  unsigned total_vccs = 0;
-  unsigned remaining_vccs = 0;
-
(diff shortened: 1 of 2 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 2f8d1b3404ab3abd1481a2b74075bf91ab780955).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103730232

tautschnig (approved):
I believe this is ok, but it may be worth confirming that the test introduced in 7a897a583 still has exactly the same behaviour.

romainbrenguier (approved):

smowton:
Confirmed that test produces the same VCC counts (3 paths, 1 VCC per path)

PR #4364, opened 2019-03-11
Do not unnecessary follow tag types when storing the boolbv mapping

tautschnig:
Types in equations must match exactly, there is no need to follow tags
when checking for type consistency. Also, do not use a followed type
when storing the mapping. When retrieving a mapped value, assert type
consistency.

This is a bugfix, but apparently doesn't show up with current develop. Upcoming work such as field-sensitivity causes simplification to fail, because of the type inconsistency that `boolbvt::get` currently might introduce.

### DIFF
--- a/src/solvers/flattening/boolbv.cpp
+++ b/src/solvers/flattening/boolbv.cpp
@@ -573,11 +573,11 @@ bool boolbvt::boolbv_set_equality_to_true(const equal_exprt &expr)
   if(!equality_propagation)
     return true;
 
-  const typet &type=ns.follow(expr.lhs().type());
+  const typet &type = expr.lhs().type();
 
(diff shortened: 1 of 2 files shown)


kroening (approved):

kroening:
Those untyped expressions trouble me...

tautschnig:
> Those untyped expressions trouble me...

Ack. We should benchmark whether there really is much value in those clock types being as-small-as-possible. Maybe it's just fine to make them `size_type()` or so.

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 32a5f76cdee476add08c0adc338f752afc462443).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/103892078

PR #4369, opened 2019-03-12
Split out goto_statet::source

smowton:
This is the one member of goto_statet which is not trivially moved when it is combined
with an existing goto_symex_statet -- rather it labels the incoming state and is discarded
after perhaps being inspected by the merge_goto routine. By splitting these we prepare
goto_statet to use a trivial move-constructor when merging into a goto_symex_statet with
a false state guard.

There should be no behavioural change caused by this commit.

### DIFF
--- a/src/goto-symex/frame.h
+++ b/src/goto-symex/frame.h
@@ -13,12 +13,14 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 #define CPROVER_GOTO_SYMEX_FRAME_H
 
 #include "goto_state.h"
+#include "symex_target.h"
 
 /// Stack frames -- these are used for function calls and for exceptions
 struct framet
 {
   // gotos
-  using goto_state_listt = std::list<goto_statet>;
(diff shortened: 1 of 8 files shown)


JohnDumbell (approved):

src/goto-symex/frame.h
```diff
@@ -13,12 +13,14 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 #define CPROVER_GOTO_SYMEX_FRAME_H
 
 #include "goto_state.h"
+#include "symex_target_equation.h"
```
tautschnig: Should be `symex_target.h`

tautschnig (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 17a039bd36f11bc36de41fbc105341c76f89ab38).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104104408

PR #4379, opened 2019-03-13
SMT2 backend: remove byte operators before calling find_symbols

smowton:
`convert_expr` assumes that `find_symbols` has already made an entry for every `array_exprt`, but byte operator lowering can introduce new ones. Therefore run it up front.

The particular construction (depth-iterator plus recursion) is a bit ugly -- this is required because (a) we must keep the post-order strategy or we expose other bugs (see https://github.com/diffblue/cbmc/issues/4380) and (b) `expr_iterator.h` is broken when expressions are mutated and then the iterator is permitted to visit those mutated parts, despite the design intending this feature. I suggest we get this fix in and then tackle those two individually, particularly as this blocks other PRs.

### DIFF
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -14,6 +14,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/arith_tools.h>
 #include <util/c_types.h>
 #include <util/config.h>
+#include <util/expr_iterator.h>
 #include <util/expr_util.h>
 #include <util/fixedbv.h>
 #include <util/format_expr.h>
@@ -612,24 +613,6 @@ void smt2_convt::convert_address_of_rec(
         expr.id_string());
 }
 
-void smt2_convt::convert_byte_extract(const byte_extract_exprt &expr)
-{
-  // we just run the flattener
-  exprt flattened_expr = lower_byte_extract(expr, ns);
-  unflatten(wheret::BEGIN, expr.type());
-  convert_expr(flattened_expr);
-  unflatten(wheret::END, expr.type());
-}
-
-void smt2_convt::convert_byte_update(const byte_update_exprt &expr)
-{
-  // we just run the flattener
-  exprt flattened_expr = lower_byte_update(expr, ns);
-  unflatten(wheret::BEGIN, expr.type());
-  convert_expr(flattened_expr);
-  unflatten(wheret::END, expr.type());
-}
-
 literalt smt2_convt::convert(const exprt &expr)
 {
   PRECONDITION(expr.type().id() == ID_bool);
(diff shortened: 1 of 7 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7a3f22e7f51cfa8a04fc72a042a282a6ba828725).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104380042

smowton:
If you want rid of the expr-iterator workarounds, merge https://github.com/diffblue/cbmc/pull/4385 first

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -4068,38 +4051,98 @@ void smt2_convt::set_to(const exprt &expr, bool value)
...
+    else
+    {
+      ++it;
+    }
+  }
+
+  return lowered_expr;
+}
+
+exprt smt2_convt::prepare_expr(const exprt &expr)
```
romainbrenguier: Please document this function. Also a better name than prepare_expr would be helpful. 
I am a bit surprised this and lower_byte_operators_rec are not const, and if they could be made static this would be better, in particular you could avoid exposing auxiliary functions in the header.

smowton: Documented and renamed. The whole class uses member functions to pass the namespace around -- I won't get into changing that in this PR.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a1cfbb56270bfae7c7508ee8a1121cf777902f80).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104411418

tautschnig:
Did you by any chance check whether this also fixes other tests currently marked "broken-smt-backend"? I think there's a fair chance that it does.

tautschnig (approved):
+1 for getting #4385 merged first, and then please also take a look whether these fixes resolve even more problems with existing tests as just said.

smowton:
Checked; found 4 tests now working.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: e623944571f7f8a2871c395c3aebcafcf5cc216d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104457441
Common spurious failures:

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 01877d1a633b1a7645fabd2dfe42a7bd3ea854e2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104458674

tautschnig:
It seems that `union6` and `union7` are still failing on Windows.

smowton:
Returned the still-broken tests to their prior state; rebased to use the now-working depth-iterator.

PR #4382, opened 2019-03-13
Make JBMC build optional with cmake

karkhaz:
This commit adds an option to CMake, `WITH_JBMC`, that controls whether
the jbmc directory will be built. It is switched on by default,
preserving the current behaviour.

The motivation is to make it as easy as possible for users to get
started with CBMC, especially users who are only interested in the C
front-end and are on Windows, where installing the JDK and Maven is a
painful exercise.

Also updated `COMPILING.md` to note that you can pass `-DWITH_JBMC=OFF`
to cmake to avoid building jbmc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,10 @@ add_subdirectory(src)
 add_subdirectory(regression)
 add_subdirectory(unit)
 
-add_subdirectory(jbmc)
+set(CBMC_CXX_STANDARD 11)
+set(CBMC_CXX_STANDARD_REQUIRED true)
+set(CBMC_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY
+    "Developer ID Application: Daniel Kroening")
 
 set_target_properties(
     analyses
@@ -152,20 +155,13 @@ set_target_properties(
     util
     xml
 
-    java_bytecode
-    java-models-library
-    jbmc
-    jbmc-lib
-    janalyzer
-    janalyzer-lib
-    jdiff
-    jdiff-lib
-    java-testing-utils
-    java-unit
-    miniz
-
     PROPERTIES
-    CXX_STANDARD 11
-    CXX_STANDARD_REQUIRED true
-    XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "Developer ID Application: Daniel Kroening"
+    CXX_STANDARD ${CBMC_CXX_STANDARD}
+    CXX_STANDARD_REQUIRED ${CBMC_CXX_STANDARD_REQUIRED}
+    XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY ${CBMC_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY}
 )
+
+option(WITH_JBMC "Build the JBMC Java front-end" ON)
+if(WITH_JBMC)
+    add_subdirectory(jbmc)
+endif()
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -219,6 +219,9 @@ require manual modification of build files.
      ```
      You shoud also install [Homebrew](https://brew.sh), after which you can
      run `brew install cmake` to install CMake.
+   - On platforms where installing the Java Development Kit and Maven is
+     difficult, you can avoid needing these dependencies by not building
+     JBMC. Just pass `-DWITH_JBMC=OFF` to cmake in step (4) below.
    - On Windows, ensure you have Visual Studio 2015 or later installed.
      Then, download CMake from the [official download
      page](https://cmake.org/download).
--- a/jbmc/CMakeLists.txt
+++ b/jbmc/CMakeLists.txt
@@ -7,3 +7,22 @@ add_custom_target(java-models-library ALL
     COMMAND cp target/core-models.jar ${CMAKE_CURRENT_SOURCE_DIR}/src/java_bytecode/library/
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/lib/java-models-library
 )
+
+set_target_properties(
+    java_bytecode
+    java-models-library
+    jbmc
+    jbmc-lib
+    janalyzer
+    janalyzer-lib
+    jdiff
+    jdiff-lib
+    java-testing-utils
+    java-unit
+    miniz
+
+    PROPERTIES
+    CXX_STANDARD ${CBMC_CXX_STANDARD}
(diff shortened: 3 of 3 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 5282abd41db5acc021fbce00b9cfe329d573eed8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104342277
Common spurious failures:

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1e1b51ee1a4c2411c8c7ba7a3fa6da3a597cc604).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104342503

romainbrenguier (commented):
I don't think that's the right way of doing it. You can specify targets on cmake command line, so by specifying cbmc as a target, cmake shouldn't have to build jbmc. If the dependencies are too different between cbmc and jbmc then maybe jbmc should be on a separate repository.

chrisr-diffblue:
I'm a bit torn on this PR - Romain is right that you can already build just a single target (e.g. `cmake --target cbmc` ) I haven't yet found a way to say "build everything except the Java stuff" - so I do quite like what this PR does. It would be good to get the view of @hannes-steffenhagen-diffblue as well, as I know he's been doing a lot of thinking about the CMake builds recently.

karkhaz:
Thanks for these comments. Here are some notes on why I went for this approach:
* My first thought was similar to @chrisr-diffblue's suggestion of "build everything except java stuff". The way to do this is to create a new target (e.g. `c-stuff`) from the [`BUILDSYSTEM_TARGETS`](https://cmake.org/cmake/help/latest/prop_dir/BUILDSYSTEM_TARGETS.html#prop_dir:BUILDSYSTEM_TARGETS) list, which contains all the top-level targets, and then manually delete the Java components on that. Users can then do `cmake --target c-stuff`. This could be a good approach except that we would have to bump the CMake minimum version to 3.7 from 3.2.
* I don't really like the idea of providing individual targets. The whole point of this PR is to allow users to build with a minimum of fuss, and I'd like them to get (at least) CBMC, goto-cc, and goto-instrument built. CMake doesn't even allow you to specify multiple targets or multiple values to `--target`.
* I think that my approach is philosophically the right one from a dependency-checking POV. Reason is that we should check that the dependencies for JBMC are installed iff we configure to build JBMC; there is no need to check whether Maven is installed if we're only building C stuff. I acknowledge that we don't currently check that JBMC dependencies are installed during the configure stage anyway, but this is a bug IMO---the configure stage should not successfully complete if we're trying to build JBMC but `JAVA_HOME` is not correctly set or the JDK or Maven arent't installed.
* So my proposal is: if the user does want to build JBMC (the default), then check that all Java deps are installed during the _configure_ stage. If the user does _not_ want to build JBMC, then those dependency checks are skipped. My PR allows this to be done cleanly because JBMC dependency checks would be added in `jbmc/CMakeLists.txt`, which gets totally ignored if the user sets `WITH_JBMC=OFF`.

karkhaz:
In case you would like to go with the `BUILDSYSTEM_TARGET` approach, I searched for cmake 3.7 adoption: cmake 3.7 is [available](https://packages.debian.org/search?keywords=cmake&searchon=names&suite=stable&section=all) in the current release of Debian, which was released in 2017. The release before that included cmake 3.0, so Debian users would have had to install a later version anyway to build cbmc. I'm not really sure how macOS and Windows work, but I assume that users download the latest version of cmake from the web site, so again a version 3.7 requirement might not be too onerous.

chrisr-diffblue:
Standard way to install CMake on OSX is via the `brew` package manager, which currently installs CMake 3.13.4. Also worth checking what version of CMake is available in CI - but suspect 3.7 would be ok. A lot of folks around here use CLion (not me though), which also includes its own copy of CMake - not sure which version though.

karkhaz:
Thanks. Let's see what @hannes-steffenhagen-diffblue has to say about this, but I wouldn't mind implementing the `BUILDSYSTEM_TARGET` approach if you would prefer that.

CMakeLists.txt
```diff
@@ -152,20 +155,29 @@ set_target_properties(
...
-    java-unit
-    miniz
-
     PROPERTIES
     CXX_STANDARD 11
     CXX_STANDARD_REQUIRED true
     XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "Developer ID Application: Daniel Kroening"
 )
+
+if(WITH_JBMC)
```
hannes-steffenhagen-diffblue, line 165: Instead of doing this, just make the `add_subdirectory` controlled by the option and always unconditionally set these properties in `jbmc/CMakeLists.txt`?

EDIT: FWIW I also think that `jbmc` is still too tightly coupled with `cbmc`, it's arguably a separate project that depends on `cbmc` but we're using it the other way round...

hannes-steffenhagen-diffblue (approved):
Seems reasonable enough to me. I find the argument that if we're not going to be building a submodule we shouldn't be running its configure stage either quite compelling, especially if it's as involved as in this case.

hannes-steffenhagen-diffblue:
@romainbrenguier I think there's generally agreement that jbmc should be a separate thing from cbmc, but it isn't for a variety of reasons (tight coupling with cbmc meaning that upstream changes to cbmc might easily break jbmc, discoverability of jbmc and others if I recall the discussion correctly)

tautschnig (approved):
1) I approve of having some way to disable building the Java-related parts in CMake builds, much like we already do in Makefile builds.
2) I am _not_ a CMake expert, and would defer to others on what might be the best approach to implement this. The current one looks sensible to me, @chrisr-diffblue or @hannes-steffenhagen-diffblue are clearly much more authoritative on this.
3) We have a very poor track record of keeping projects outside the single repository alive, because we still have way too tight coupling of the various directories/modules. This might change one day, but I don't think this PR should be deferred until that day.

chrisr-diffblue (approved):
Agree with most of the comments, and as such, I'm also happy for this to go in.

karkhaz:
Thanks all. I'll implement proper dependency checking in jbmc's configure stage in a separate PR at some point.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 562fad22ee18d3d152ea23e73ebde4ff11361fdd).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104433588

PR #4385, opened 2019-03-14
Expr-depth-iterator: support iterating over mutated expressions

smowton:
Expr-iterator has always had a mutate() function to enable editing part of an expression tree, and a "next_sibling_or_parent" method to skip walking over the changed expression. However as it turns out using the latter with the former was effectively mandatory because when an expression was replaced wholesale (e.g. `it.mutate() = symbol_exprt("something_different", ...)`), then a stale iterator referring to the replaced expr's operands vector led to a segfault on the next `operator++`. 

This takes the simple approach to fixing that: use integer indices instead of iterators. I also add some basic unit tests, including one that exhibits the crash in question.

### DIFF
--- a/src/util/expr_iterator.h
+++ b/src/util/expr_iterator.h
@@ -42,22 +42,18 @@ class const_unique_depth_iteratort;
 struct depth_iterator_expr_statet final
 {
   typedef exprt::operandst::const_iterator operands_iteratort;
-  inline depth_iterator_expr_statet(
-    const exprt &expr,
-    operands_iteratort it,
-    operands_iteratort end):
-    expr(expr), it(it), end(end) { }
+  explicit depth_iterator_expr_statet(const exprt &expr) : expr(expr), op_idx(0)
+  {
+  }
   std::reference_wrapper<const exprt> expr;
-  operands_iteratort it;
-  operands_iteratort end;
+  std::size_t op_idx;
 };
 
 inline bool operator==(
   const depth_iterator_expr_statet &left,
   const depth_iterator_expr_statet &right)
 {
-  return distance(left.it, left.end) == distance(right.it, right.end) &&
-         left.expr.get() == right.expr.get();
+  return left.op_idx == right.op_idx && left.expr.get() == right.expr.get();
 }
 
 /// Depth first search iterator base - iterates over supplied expression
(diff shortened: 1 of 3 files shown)


src/util/expr_iterator.h
```diff
@@ -42,22 +42,18 @@ class const_unique_depth_iteratort;
 struct depth_iterator_expr_statet final
 {
   typedef exprt::operandst::const_iterator operands_iteratort;
-  inline depth_iterator_expr_statet(
-    const exprt &expr,
-    operands_iteratort it,
-    operands_iteratort end):
-    expr(expr), it(it), end(end) { }
+  inline depth_iterator_expr_statet(const exprt &expr) : expr(expr), op_idx(0)
```
tautschnig: Missing `explicit` (and you can get rid of the `inline`, which is actually implicit...).

src/util/expr_iterator.h
```diff
@@ -185,18 +184,11 @@ class depth_iterator_baset
     for(auto &state : m_stack)
     {
       // This deliberately breaks sharing as expr is now non-const
-      auto &operands = expr->operands();
-      // Get iterators into the operands of the new expr corresponding to the
-      // ones into the operands of the old expr
-      const auto it = operands.end() - (state.end - state.it);
+      (void)expr->operands();
```
tautschnig: Maybe that should be `(void)expr->write();` instead of a somewhat random function (`operands`)?

tautschnig (approved):
I'd suggest that @LAJW also takes a look, please?

src/util/expr_iterator.h
```diff
@@ -185,18 +184,11 @@ class depth_iterator_baset
...
       state.expr = *expr;
-      state.it=it;
-      state.end=operands.end();
       // Get the expr for the next level down to use in the next iteration
-      if(!(state==m_stack.back()))
-      {
-        expr = &*it;
-      }
+      if(!(state == m_stack.back()))
+        expr = &const_cast<exprt &>(state.expr.get().operands()[state.op_idx]);
```
LAJW: ❓ Why is another `const_cast` necessary here?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fc35a1800a34034a14807963566ec4f93bdf6f0b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104448724

LAJW (approved):

src/util/expr_iterator.h
smowton: Fair enough, replaced with the synonym `*expr`

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 17bee52504f98c922e577ddec65d23a08d153f2e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104465585

PR #4386, opened 2019-03-14
Symex: propagate assumptions and conditions

smowton:
This propagates simple tests and assumptions (e.g. `ASSUME(x == 5)`, or `IF y == null GOTO 1`) similar to how assignments are already propagated.

My particular use case is to note that non-exceptional return from a Java function (dominated by `IF @inflight_exception != NULL`) and a catch block (terminated by `@inflight_exception = NULL;`) mean that we can be certain that there is no exception still in flight when executing normal code. However as the C tests accompanying this PR indicate, the concept is more general and is likely to have other benefits by eliminating e.g. redundant return-value or error checks.

~This depends on #4199 (for easier L2 generation number management) and #4379 (to fix byte-update / SMT2 handling exposed by this PR), so please review only the last 5 commits.~


src/goto-symex/goto_state.cpp
```diff
@@ -34,3 +36,47 @@ std::size_t goto_statet::increase_generation(
...
+    {
+      const ssa_exprt &ssa_lhs = to_ssa_expr(lhs);
+      INVARIANT(
+        !ssa_lhs.get_level_2().empty(),
+        "apply_condition operand should be L2 renamed");
+
+      if(
+        previous_state.threads.size() > 1 &&
+        previous_state.write_is_shared(ssa_lhs, ns) ==
+        goto_symex_statet::write_is_shared_resultt::SHARED)
```
tautschnig: Not a proper review just yet, only poking at what I guessed would be key bits:
1) Why can't the goal of this PR be addressed completely by putting the RHS value in the propagation map? (And really just this, nothing else.)
2) I would have assumed the one case where we _cannot_ do this is when we're dealing with shared variables. But now the above condition suggests that it's exactly that case that we _do_ put in the propagation map?

smowton: 1. According to our current merge logic, equal L2 generation numbers --> merge is a no-op. We must either introduce the possibility that despite equal L2 generation numbers the constant prop might disagree (moderately costly), or allocate an "administrative" L2 generation, as we do for constant assignment.
2. I accidentally flipped the condition while refactoring, fixed now.

tautschnig: But doesn't this then introduce spurious merge assignments? And is it actually a problem if constant propagation disagrees?

smowton: (the value-set assignment... just seems like we might as well; this way @owen-jones-diffblue's value-set-filtering will have less work to do as we have established a straightforward constant.

smowton: Extra merge assignments: yes, harmless but some noise, indeed. We must do *something* about the constant propagator, or else the knowledge will outlive the code dominated by the branch or test.

Basically would you rather handle this through the L2 mechanism, or pay cost prop. to the size of the propagator map at each branch to merge them?

smowton:
The win32 CI failure is inherited from #4199

src/goto-symex/goto_state.cpp
tautschnig: Maybe it's worth adding an extra set of `irep_idt` that lists the constants inferred that way, and is then consulted at join points?

smowton: Happy to do that -- will appeal to @kroening for an opinion before doing it though

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 2fca1b8da079c696a78ae14a59dbfeec9cd493e5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104441777

src/goto-symex/goto_state.h
```diff
@@ -82,6 +82,11 @@ class goto_statet
     value_set = std::move(other_state.value_set);
   }
 
+  void apply_condition(
```
owen-mc-diffblue, line 78: Please briefly document this function

src/goto-symex/symex_goto.cpp
```diff
@@ -22,6 +22,28 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <analyses/dirty.h>
 #include <util/simplify_expr.h>
 
+static void apply_goto_condition(
```
owen-mc-diffblue, line 33: Please briefly document this function

src/goto-symex/symex_goto.cpp
```diff
@@ -228,6 +250,16 @@ void goto_symext::symex_goto(statet &state)
 
     symex_transition(state, state_pc, backward);
 
+    if(!symex_config.doing_path_exploration)
+    {
+      // In --paths mode we would require the GOTO condition as a constraint
+      // as well, effectively treating the branch like an assumption.
```
owen-mc-diffblue: I don't understand this comment. Possibly the condition is inverted by mistake.

smowton:
@owen-jones-diffblue all comments applied

src/goto-symex/goto_state.cpp
```diff
@@ -34,3 +36,59 @@ std::size_t goto_statet::increase_generation(
 
   return current_emplace_res.first->second.second;
 }
+
+/// Given a condition that must hold on this path, propagate as much knowledge
+/// as possible. For example, if the condition is (x == 5), whether that's an
+/// assumption or a GOTO condition that we just passed through, we can propagate
+/// the constant '5' for future 'x' references. If the condition is (y == &o1)
+/// then we can use that to populate the value set.
+/// \param condition: condition that must hold on this path
```
owen-mc-diffblue: Should you mention that this should be L2? There's a short end-of-line comment to that effect in the declaration.

smowton: Done

smowton:
This is now rebased and ready to go.

src/goto-symex/goto_state.h
smowton: Documented at definition site

peterschrammel (approved):
LGTM

jbmc/regression/jbmc/exception-cleanup/Test.java
```diff
@@ -0,0 +1,44 @@
...
+    // unreachable; if not it will pursue both paths.
+    try {
+      mayThrow(false);
+    }
+    catch(Exception e) {
+      unreachable();
+    }
+
+    // Try it once more, to check we also know after an unreachable catch that
+    // there is still no inflight exception
```
allredj, line 27: :bulb: A test with nested try-catch would be useful to prove that point.

jbmc/regression/jbmc/exception-cleanup/test.desc
```diff
@@ -0,0 +1,14 @@
...
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+1 remaining after simplification$
+--
+^warning: ignoring
+--
+The unreachable function should be successfully marked unreachable by symex,
+but the final uncaught exception is not yet, as it requires symex to note that
+within a catch block @inflight_exception is definitely *not* null, which it
+can't just yet.
```
allredj: Not sure I understand what that comment is here for. The test is already marked as `CORE`, so what difference should we see when the stated limitation is resolved?

allredj (commented):
Change looks sensible, but if the purpose is performance, it would be nice to post some performance results to at least check that there is no slowdown.

jbmc/regression/jbmc/exception-cleanup/test.desc
smowton: Then we'll resolve *all* conditions in symex. I'll amend to make that clearer.

jbmc/regression/jbmc/exception-cleanup/test.desc
```diff
@@ -0,0 +1,14 @@
...
+--function Test.main
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+1 remaining after simplification$
+--
+^warning: ignoring
+--
+The unreachable function should be successfully marked unreachable by symex,
+but the final uncaught exception is not yet, as it requires symex to note that
```
owen-mc-diffblue: I don't understand what "the final uncaught exception" means. All the exceptions appear to be caught to me.

owen-mc-diffblue (approved):
The commit message for "Symex: propagate constants implied by assumptions and conditions" says the constant propagator might be populated with new information. Is it also possible that the value-set gets updated? If so, please correct the commit message.

jbmc/regression/jbmc/exception-cleanup/Test.java
smowton: Such a case actually does not work yet, because symex can't tell that all exceptions were caught (i.e. the catch dispatch table is exhaustive). The reason is it isn't yet smart enough to notice that upon entering the catch block the incoming pointer cannot be null, and therefore e instanceof Exception certainly succeeds (null instanceof A returns false for any A).

Fixing that is my next task!

jbmc/regression/jbmc/exception-cleanup/test.desc
smowton: It refers to the "did any exception escape at top level?" test that's always generated in `__CPROVER_start`, and yes they are all caught, but symex isn't *quite* smart enough to determine that yet. This PR enables it to determine there's no exception in flight when we return to normal code (preventing exploration of catch blocks after future no-throw function calls), but doesn't yet enable it to determine catch block does not rethrow.

smowton: (I'll improve the wording here)

smowton:
@allredj I ran this on some Tika functions and observed that a small batch of 6 functions that took a total of 248s (symex-only) took 231s with this patch. Of course that's a small test, but it gives me some confidence this will be overall beneficial. I'll run a longer benchmark tomorrow to be sure and then merge if the results are also positive.

allredj (approved):
Great!

jbmc/regression/jbmc/exception-cleanup/Test.java
owen-mc-diffblue: You could add a KNOWNBUG test for it now and turn it on as part of your next task.

smowton:
perf_out

Extended benchmark results show a slight improvement, so merging this as-is and we can implement @tautschnig's suggestion re: using different data structures to record a path-local constant if and when we want to.

PR #4387, opened 2019-03-14
Use temp variable for `this` in virtual method dispatch tables

owen-mc-diffblue:
This is so that value-set filtering in symex will
filter elements which don't have the correct class
identifier out of the value-set for `this`.


src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -144,6 +146,63 @@ static void create_static_function_call(
...
+static void process_this_argument(
+  exprt &argument_for_this,
+  symbol_table_baset &symbol_table,
+  const source_locationt &vcall_source_loc,
+  goto_programt &new_code_for_this_argument)
+{
+  const auto dereference_it = std::find_if(
+    argument_for_this.depth_cbegin(),
+    argument_for_this.depth_cend(),
+    [](const exprt &expr) { return can_cast_expr<dereference_exprt>(expr); });
```
tautschnig: You could use `const bool argument_for_this_contains_a_dereference = has_subexpr(argument_for_this, ID_dereference);` If you don't like the way `has_subexpr` is implemented then you might still use `has_subexpr` here, but also replace the implementation of `has_subexpr` by your code.

src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -144,6 +146,63 @@ static void create_static_function_call(
...
+    dereference_it != argument_for_this.depth_cend();
+
+  if(argument_for_this_contains_a_dereference)
+  {
+    // Create a temporary for the `this` argument. This is so that value-set
+    // filtering in symex will reduce the value-set for `this` to those with the
+    // correct class identifier.
+    symbolt &temp_symbol = get_fresh_aux_symbol(
+      argument_for_this.type(),
+      "remove_virtual_functions", // TODO: pass function name in to use here?
```
smowton: Yes, see for example `remove_instanceof.cpp` is doing this (and its callers conveniently (a) live alongside remove-virtuals callers and (b) already have some way to lay their hands on the function ID. The mode can then be retrieved by looking it up in the symbol table.

smowton (commented):
LGTM apart from the noted TODO

src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -144,6 +146,63 @@ static void create_static_function_call(
...
+      "this_argument",
+      vcall_source_loc,
+      ID_java, //TODO: get this from somewhere
```
tautschnig: Require that name of the `goto_program` that's being processed is passed in as well (just like `remove_function_pointers` does this, search for `function_id`). You can then look up the symbol corresponding to `function_id` and use the `mode` of that symbol.

src/goto-programs/remove_virtual_functions.cpp
tautschnig: Yes, please.

src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -144,6 +146,63 @@ static void create_static_function_call(
...
+    // really use java_bytecode_instrumentt::check_null_dereference(), but it is
+    // in jbmc rather than cbmc. Or we could use create_fatal_assertion(), but it
+    // returns a code_blockt, which is a bit inconvenient.
+    const notequal_exprt null_check(
+      temp_var_for_this,
+      null_pointer_exprt(to_pointer_type(temp_var_for_this.type())));
+    new_code_for_this_argument.add(
+      goto_programt::make_assertion(null_check, vcall_source_loc));
+    new_code_for_this_argument.add(
+      goto_programt::make_assumption(null_check, vcall_source_loc));
```
tautschnig: That's only valid for Java, is it?

tautschnig (commented):
Just some very general comments, I'll defer to others for approval. My understanding of this bit of code isn't good enough.

src/goto-programs/remove_virtual_functions.cpp
owen-mc-diffblue: You're right, this is far too java-specific. It's needed because without it the local safe pointer analysis wasn't realising that `this` definitely wasn't null after the null check. Other options include:
1) making the local safe pointer analysis a teensy bit smarter;
2) checking for a null check in the two instructions above the virtual function call, and only inserting a null check here if we find one.

owen-mc-diffblue: I've done option 2.

owen-mc-diffblue:
Okay, I have addressed all review comments and fixed the CI issues that that caused. This is ready for re-review.

jbmc/regression/jbmc/virtual_temp_var_for_this_argument/test_one_candidate.desc
```diff
@@ -0,0 +1,12 @@
...
+--show-goto-functions --function Main.oneCandidate
+^EXIT=0$
+^SIGNAL=0$
+B\.f:\(\)V\(\);$
+--
+this_argument
+--
+The temporary variable `this_argument` should not be created, because it
+is only created when there is a dispatch table and the `this` argument has a
+dereference in it. Both conditions are false.
```
smowton: `. Both conditions are false.` -> `, neither of which is the case here`

src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -144,11 +155,126 @@ static void create_static_function_call(
...
+        ++expr_it)
+    {
+      if(*expr_it == argument_for_this)
+      {
+        expr_it.mutate() = temp_var_for_this;
+        changed = true;
+      }
+    }
+
+    if(!changed)
```
smowton: Probably don't do this -- for example, with
```
ASSUME(x->y->z != null);
ASSUME(some_param != null);
tmp_this = x->y->z;
```
we should keep the assumption concerning `x->y->z`

We might never currently generate any intervening assumptions, but given it's easy to do so we should probably tolerate them if present.

src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -144,11 +155,126 @@ static void create_static_function_call(
   }
 }
 
+/// If there are ASSERT or ASSUME instructions involving \p argument_for_this
+/// in the instructions directly preceding the virtual function call then we
+/// duplicate them for \p temp_var_for_this. This is mainly aimed at null
+/// checks, to make the local safe pointer analysis work. We stop as soon as we
```
smowton: `to make the local safe pointer analysis work` -> `because \ref local_safe_pointerst would otherwise lose track of known-not-null pointers due to the newly introduced assignment`

src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -144,11 +155,126 @@ static void create_static_function_call(
...
+static void process_this_argument(
+  const irep_idt &function_identifier,
+  const goto_programt &goto_program,
+  const goto_programt::targett target,
+  exprt &argument_for_this,
+  symbol_table_baset &symbol_table,
+  const source_locationt &vcall_source_loc,
+  goto_programt &new_code_for_this_argument)
+{
+  const bool argument_for_this_contains_a_dereference =
```
smowton: Given the variable is longer than the expression, just inline this :)

src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -144,11 +155,126 @@ static void create_static_function_call(
...
+    has_subexpr(argument_for_this, ID_dereference);
+
+  if(argument_for_this_contains_a_dereference)
+  {
+    // Create a temporary for the `this` argument. This is so that value-set
+    // filtering in symex will reduce the value-set for `this` to those elements
+    // with the correct class identifier.
```
smowton: Provide a reference to a function on `goto_symext` for further information

src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -144,11 +155,126 @@ static void create_static_function_call(
...
+  const exprt &argument_for_this,
+  const symbol_exprt &temp_var_for_this)
+{
+  goto_programt checks_directly_preceding_function_call;
+
+  while(instr_it != goto_program.instructions.cbegin())
+  {
+    instr_it = std::prev(instr_it);
+
+    if(!(instr_it->type == ASSERT || instr_it->type == ASSUME))
```
smowton: Agree to skipping over assertions, but suggest only duplicating assumptions (duplicating assertions would be confusing to a user, as any violation would show up twice)

smowton (approved):
Looks good, some mainly-cosmetic improvements

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 32b0e624efe13c13f3544db22d9de2481239d864).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104940146

src/goto-programs/remove_virtual_functions.cpp
owen-mc-diffblue: Let me just check I understand you correctly: the thing you're saying we shouldn't do is stopping when we encounter an assert or assume which doesn't contain `argument_for_this`; you want us to keep on going until we find an instruction which isn't an assert or an assume (or we hit the beginning of the function).

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9dbe4fdb9f9f0da1fedbf77da1c16d5ff44e640f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104983138

src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -23,14 +26,17 @@ class remove_virtual_functionst
...
   remove_virtual_functionst(
-    const symbol_table_baset &_symbol_table,
+    symbol_table_baset &_symbol_table,
     const class_hierarchyt &_class_hierarchy);
 
   void operator()(goto_functionst &goto_functions);
 
-  bool remove_virtual_functions(goto_programt &goto_program);
+  bool remove_virtual_functions(
+    irep_idt function_identifier,
```
peterschrammel:
```suggestion
    const irep_idt &function_id,
```

const & (several occurrences)

Also, I think we settled on calling this `function_id` throughout the codebase.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 77075e680bbdee84c3f2528426700c7b8717e4ef).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105077354

PR #4388, opened 2019-03-14
There is no need for argc and argv to be new

martin-cs:
With thanks to @natasha-jeppu who found this the hard way.  Consider the following unlikely sequence of events:

```
$ cat ~/tmp/can-delete/main.c
  return 0;
}
$ goto-cc main.c -o main.goto
$ cbmc main.goto --function main
CBMC version 5.11 (cbmc-5.11-881-gd52cd0d1e-dirty) 64-bit x86_64 linux
Reading: /home/martin/tmp/can-delete/main.goto
failed to insert argc symbol
SUPPORT FUNCTION GENERATION ERROR
```

`argc` and `argv` both exist and it is fine to re-use them.  If the
addition fails it will throw an exception, the `.second` bool is only
false in the case they already exist which is harmless.

### DIFF
--- a/src/ansi-c/ansi_c_entry_point.cpp
+++ b/src/ansi-c/ansi_c_entry_point.cpp
@@ -268,10 +268,12 @@ bool generate_ansi_c_start_function(
         argc_symbol.is_lvalue = true;
         argc_symbol.mode = ID_C;
 
-        if(!symbol_table.insert(std::move(argc_symbol)).second)
+        auto r = symbol_table.insert(argc_symbol);
+        if(!r.second && r.first != argc_symbol)
         {
           messaget message(message_handler);
-          message.error() << "failed to insert argc symbol" << messaget::eom;
+          message.error() << "argc already exists but is not usable"
+                          << messaget::eom;
           return true;
         }
       }
@@ -295,10 +297,12 @@ bool generate_ansi_c_start_function(
         argv_symbol.is_lvalue = true;
         argv_symbol.mode = ID_C;
 
-        if(!symbol_table.insert(std::move(argv_symbol)).second)
+        auto r = symbol_table.insert(argv_symbol);
+        if(!r.second && r.first != argv_symbol)
         {
           messaget message(message_handler);
-          message.error() << "failed to insert argv symbol" << messaget::eom;
+          message.error() << "argv already exists but is not usable"
+                          << messaget::eom;
           return true;
         }
       }


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 371fa5f7aa3b78fcf215be1947eb8da0596d6c3b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104461375

src/ansi-c/ansi_c_entry_point.cpp
```diff
@@ -268,12 +268,7 @@ bool generate_ansi_c_start_function(
         argc_symbol.is_lvalue = true;
         argc_symbol.mode = ID_C;
 
-        if(!symbol_table.insert(std::move(argc_symbol)).second)
-        {
-          messaget message(message_handler);
-          message.error() << "failed to insert argc symbol" << messaget::eom;
-          return true;
-        }
+        symbol_table.insert(std::move(argc_symbol));
```
tautschnig: Shouldn't there at least be some type compatibility check?

martin-cs: Where should it be done?

tautschnig: I'd say here is a good place? It will require some more code, but it would be nice to check whether argc/argv are present already and, if so, their types match expectations. Else we should continue to fail.

martin-cs: I have added checks that `argc` and `argv` are identical to the required symbols.

src/ansi-c/ansi_c_entry_point.cpp
```diff
@@ -268,10 +268,11 @@ bool generate_ansi_c_start_function(
         argc_symbol.is_lvalue = true;
         argc_symbol.mode = ID_C;
 
-        if(!symbol_table.insert(std::move(argc_symbol)).second)
+        auto r = symbol_table.insert(argc_symbol);
+        if (!r.second && r.first != argc_symbol)
         {
           messaget message(message_handler);
-          message.error() << "failed to insert argc symbol" << messaget::eom;
+          message.error() << "argc already exists but is not usable" << messaget::eom;
```
peterschrammel: What does `is not usable` mean?

martin-cs: It is not equal to the expected symbol used for `argc` or `argv`.  It might mean that it has an unexpected type, differing life-time, etc.  TBH I struggle to think how this could happen but @tautschnig wanted a test and a test without error handling seems ... stupid.

Hmmm... maybe if you had a version of `main` that did something unusual and very non-standard with it's arguments and then tried to set the entry point to `main`, maybe you could trigger this.

martin-cs:
CI should be clean now.

martin-cs:
Travis appears to be timing out on one of the clang builds.  This has happened several times now; please could someone with the power of Travis help?

martin-cs:
CI doesn't time out now; weird.  @tautschnig opinons ?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 2bee6b195f95dc425b7c72378d3bb2d2688fd5e5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105187801

tautschnig (approved):

PR #4203, opened 2019-02-15
Split static field test into two parts

smowton:
I realised validate-goto-model and show-goto-functions can't be used together, as validation happens after the show operation.

PR #4389, opened 2019-03-14
Move validate-goto-model before show-goto-functions

smowton:
As requested in https://github.com/diffblue/cbmc/pull/4203 -- if merged will render that redundant.


tautschnig:
@smowton `lazyloading_no_candidates/test.desc` now reports a type inconsistency.

tautschnig (approved):
Although there still is some debugging to do: this certainly seems like the right thing to do, and if it uncovered another bug then that's even better :-)

smowton:
Cause was remove-returns treating actually bodyless stubs wrongly, but it no longer matters now remove-returns doesn't change types, so I just rebased this.

peterschrammel (approved):

PR #4374, opened 2019-03-12
Additional unbounded-array test that should continue to pass

tautschnig:
This test passes without nondet-propagation, and should continue to pass
if/when it is merged. Test provided by @romainbrenguier.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- n/a Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4390, opened 2019-03-14
Fix segmentation fault when failing to lookup parameter names [depends-on: #4434]

tautschnig:
Users should not be able to trigger segmentation faults.

Only the second commit is new, the first one is #4374. It is likely possible to also trigger the same problem with other tests and we might thus be able to get rid of the dependency on that PR. It was just that test where I bumped into this problem.

### DIFF
--- a/jbmc/regression/jbmc/nondet_propagation1/lazy-loading-bug.desc
+++ b/jbmc/regression/jbmc/nondet_propagation1/lazy-loading-bug.desc
@@ -1,0 +1,13 @@
+CORE
+andNot_Bug.class
+--depth 1600 --java-unwind-enum-static --java-max-vla-length 48 --unwind 4 --max-nondet-string-length 100 andNot_Pass.class --function andNot_Pass.main
+^\*\* 1 of \d+ failed
+VERIFICATION FAILED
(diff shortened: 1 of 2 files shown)


smowton:
I'd say we generally mishandle the case where multiple class files are given on the command-line. I'd been meaning to allow language frontends to specify whether they are singletons (as Java effectively is), and if so notify the existing singleton of the extra arg rather than make a whole new instance of `java_bytecode_languaget`. Is this crash only achievable with multiple instances of the frontend?

tautschnig:
> Is this crash only achievable with multiple instances of the frontend?

Yes, at least in my (limited) testing I was only able to trigger this by passing multiple class files on the command line.

I am not very certain that the "fix" is right, but seg faulting certainly is wrong.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 59b277ff2bcbf40bba3036e9064c4deef1248b5d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105815861

tautschnig:
Superseded by #4434 and #4437.

PR #4394, opened 2019-03-15
Make rename return a renamedt<exprt, level> 

romainbrenguier:
This is to reflect in the type that an expression as been renamed and we
can then make some functions only accept expressions of that type.

### DIFF
--- a/src/goto-instrument/accelerate/scratch_program.cpp
+++ b/src/goto-instrument/accelerate/scratch_program.cpp
@@ -74,7 +74,7 @@ bool scratch_programt::check_sat(bool do_slice, guard_managert &guard_manager)
 
 exprt scratch_programt::eval(const exprt &e)
 {
-  return checker->get(symex_state->rename<L2>(e, ns));
+  return checker->get(symex_state->rename<L2>(e, ns).get());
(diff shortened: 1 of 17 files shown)


smowton (approved):
You could avoid a lot of `.get()` by permitting an implicit downcast to `const exprt &`, which is safe (the upcast to `renamedt` of course should stay explicit). Otherwise this looks good.

romainbrenguier:
@smowton 
> You could avoid a lot of .get() by permitting an implicit downcast to const exprt &, which is safe (the upcast to renamedt of course should stay explicit). Otherwise this looks good.

Yes that could be nice, but for now I'd prefer to keep it explicit so that it's easy to see where the type gets "ignored" (and the code could be improved), if that's ok.

smowton:
OK

peterschrammel (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fa227a0e51795183abb3a2111486c99492a26ff2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105088575

PR #4398, opened 2019-03-17
exprt::visit now accepts a lambda

kroening:
This is more modern than the current visitor base class.
Also adds two ranged for statements.

### DIFF
--- a/src/goto-programs/validate_goto_model.cpp
+++ b/src/goto-programs/validate_goto_model.cpp
@@ -141,18 +141,8 @@ void validate_goto_modelt::check_returns_removed()
 
 void validate_goto_modelt::check_called_functions()
 {
-  class test_for_function_addresst : public const_expr_visitort
-  {
-  public:
-    test_for_function_addresst(
-      const function_mapt &function_map,
-      const validation_modet &vm)
-      : function_map{function_map}, vm{vm}
-    {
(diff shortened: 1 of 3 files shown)


src/solvers/strings/string_refinement.cpp
```diff
@@ -1628,34 +1628,20 @@ compute_inverse_function(const exprt &qvar, const exprt &val, const exprt &f)
...
 static bool find_qvar(const exprt &index, const symbol_exprt &qvar)
 {
-  find_qvar_visitort v2(qvar);
-  index.visit(v2);
-  return v2.found;
+  bool found = false;
+
+  index.visit([&found, &qvar](const exprt &e) {
+    if(e == qvar)
+      found = true;
```
romainbrenguier: I would rather write `std::find(e.depth_begin(), e.depth_end(), qvar)`, but this version is still better than the previous one.

romainbrenguier (approved):
Looks good

smowton (approved):
Looks good. While you're there you might want to note that `exprt::depth_begin`/`depth_end` are probably a better alternative to `visit(exprt &)` because the former preserves sharing where it can while the latter breaks it everywhere.

kroening:
@smowton I'll try with letify!

PR #4402, opened 2019-03-17
exprt::visit post-order traversal variant

kroening:

### DIFF
--- a/src/util/expr.cpp
+++ b/src/util/expr.cpp
@@ -244,6 +244,50 @@ const source_locationt &exprt::find_source_location() const
 }
 
 template <typename T>
+void visit_post_template(std::function<void(T &)> visitor, T *_expr)
+{
+  struct stack_entryt
+  {
+    T *e;
+    bool operands_pushed;
+    explicit stack_entryt(T *_e) : e(_e), operands_pushed(false)
+    {
+    }
+  };
+
+  std::stack<stack_entryt> stack;
+
+  stack.emplace(_expr);
+
+  while(!stack.empty())
+  {
+    auto &top = stack.top();
+    if(top.operands_pushed)
+    {
+      visitor(*top.e);
+      stack.pop();
+    }
+    else
+    {
+      // do modification of 'top' before pushing in case 'top' isn't stable
+      top.operands_pushed = true;
+      for(auto &op : top.e->operands())
+        stack.emplace(&op);
+    }
+  }
+}
+
+void exprt::visit_post(std::function<void(exprt &)> visitor)
+{
+  visit_post_template(visitor, this);
+}
+
+void exprt::visit_post(std::function<void(const exprt &)> visitor) const
+{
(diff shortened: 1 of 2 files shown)


src/util/expr.cpp
```diff
@@ -279,14 +279,80 @@ void exprt::visit(std::function<void(const exprt &)> visitor) const
...
+    {
+      // do modification before pushing in case 'top' isn't stable
+      top.operands_pushed = true;
+      for(auto &op : top.e->operands())
+        stack.emplace(&op);
+    }
+  }
+}
+
+void exprt::visit_post(std::function<void(const exprt &)> visitor) const
```
romainbrenguier, line 285: There is a lot of duplication with the non-const version, which is not ideal. Maybe this could be avoided using templates?

src/goto-programs/mm_io.cpp
```diff
@@ -24,7 +24,7 @@ void collect_deref_expr(
   std::set<dereference_exprt> &dest)
 {
   src.visit_pre([&dest](const exprt &e) {
-    if(e.id()==ID_dereference)
+    if(e.id() == ID_dereference)
```
romainbrenguier: This change does not correspond to the commit message

romainbrenguier (approved):
I would have preferred an implementation as iterators, so that it could be used with standard algorithms, but otherwise this looks fine.

smowton (approved):
No complaints additional to @romainbrenguier

tautschnig (changes requested):
> Each commit message has a non-empty body, explaining why the change was made.

I'm not sure why this has been ticked. I am now requesting changes, because the justification is in particular missing with the first commit, which uses an iterator instead of the visitor. That seems like a good idea, but other commits instead introduce visitors, so I'm quite confused. Also, I am wondering whether the post-order visitor isn't duplicating a lot of the work that the depth iterator already does?

kroening:
The first commit is triggered by the comment by @romainbrenguier. Now moved to #4542.
The depth iterator does pre-order traversal, which should be documented.
This PR adds post-order traversal.

src/util/expr.cpp
kroening: done

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: cb77c8a30927900e1ae243b3537cb95f75d714b4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108583976

tautschnig (approved):

tautschnig:
@kroening This now just needs a rebase.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: f30d9db20e02614018cb673340486037d5c7d6c4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108661983

Issue #1521, opened 2017-10-25
Discussion : best practice for handling (command line) options


martin-cs:
This has come up in reviewing a number of PRs recently and I thought it might be good to have a documented consensus on this.  To achieve one, we possibly need a discussion.  So this is my understanding of the current state-of-affairs / how things are (but not necessarily how I, or anyone else) thinks they should be.

1. There has been a long-term goal of minimising the number of command-line options and making sure that the default options Do The Right Thing.  Verification systems seems to grow vast numbers of options, for comparison, another system I have worked with has over 300 flags, another requires a sophisticated config file (it has it's own include system!) and comes with around 75 sample configuration.  In these cases it is easy for configuration to become a skill in itself, creating a steep barrier to entry and meaning the majority of users do not use the majority of the functionality of the system.  There is also the obvious QA and testing nightmare.

2. There have always been undocumented options.  These are experimental features, features that only work / help in restricted / limited use-cases, features for specific customers, etc.  This is not really an ideal state of affairs but it allows some functionality to be maintained centrally but only really used by some users.

3. The command line gets transferred to a cmdlinet object owned by the appropriate child of parse_options_baset (which contains the fantastic "doit" method, the entry point to our binaries).  This is the unfiltered, lightly parsed command-line and shouldn't (there are exceptions but there probably shouldn't be new exceptions) really be used apart from parsing to options.

4. The parse_options methods should include code that checks cmdlinet, validates that the (combination of) flags make sense and sets values (including defaults, if applicable) in an optionst.  This tends to be a local variable of a top-level method and is passed (as a const reference) to parts of the system that need config information.  This should, preferably, but not necessarily exclusively be in the directory associated with the relevant binary.

5. Then there is configt.  This holds details of the language and platform for which the code is being verified.  It is auto-detected or loaded from the goto-program where possible but can be over-ridden from the command-line (and example of where a cmdline is used for something other than building options).  This structure is global because many of these options are used all over the place in weird corners.  It should only really contain options which are about the language / platform being analyzed.

So to add an option, one should:

A. Work out if it is needed and if so, which tool(s) needs it.  Is there any sane way of automatically configuring this information?

[...]

PR #4404, opened 2019-03-18
[RFC] Refactor command line options 

peterschrammel:
Before I go ahead with this rather big refactoring, I'd like to collect some feedback and additional requirements that you might think of.

a1496e0 is to showcase some of the capabilities.
CI is expected to fail.

Currently, I have the following requirements:
- export machine-readable commandline options output (JSON/XML) for downstream consumption
- export information on deprecated options for downstream consumption
- a step further towards making plain/json/xml output clean (there's no help/usage output in JSON/XML at the moment)
- make sure plain help text is always properly formatted and wrapped
- prepare for unifying cmdlinet->optionst parsing code

Checklist:

### DIFF
--- a/src/cbmc/cbmc_parse_options.h
+++ b/src/cbmc/cbmc_parse_options.h
@@ -15,6 +15,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <ansi-c/ansi_c_language.h>
 #include <ansi-c/c_object_factory_parameters.h>
 
+#include <util/cmdline_definition.h>
 #include <util/parse_options.h>
 #include <util/timestamper.h>
 #include <util/ui_message.h>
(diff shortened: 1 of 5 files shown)


LAJW:
Looks decent, though I had something like this in mind: https://www.npmjs.com/package/commander where each command could have its own parser. But it requires `std::any` (or a polyfill).

peterschrammel:
> where each command could have its own parser.

I don't think that we have an issue with the raw parsing. Most of the mess comes from duplicating the translation into a configuration object (optionst) across all driver programs. My vague plans for the second step there are to associate these translations with a cmdline_definitiont/cmdline_optiont. That will factor out all the verbose cmdline->options code. In third step we could then transition to more structured configuration objects.

martin-cs:
These sound like reasonable goals.  Bringing some sanity to the option handling of `goto-instrument` and `goto-analyzer` and not having to manually duplicate and parallel maintain their command options would make me very happy.  If we could also do a rationalisation / orthogonalisation on the options that would also be great.  One of the selling points of CBMC used to be that it didn't have an insane number of options.

I'm not sure that I understand from the patch how the new system is supposed to work; perhaps you could write a little bit describing it?  The main value(?) of separating `cmdlinet` and `optionst` is in giving sane defaults and flagging combinations that don't work.  As long as we can do that I think it could be a big improvement.

One point of pedantry; can we call the combined thing something more like options or parameters rather than command line because it would be nice to believe in a world where that config might come from somewhere else.

[ Inflammatory, unhelpful suggestion -- maybe we could also include `configt` in this refactor ... ]

martin-cs:
Oh yes; for keeping things in sync, there is also the question of the Java tools.

src/cbmc/cbmc_parse_options.h
```diff
@@ -38,49 +39,45 @@ class goto_functionst;
...
-  "(arrays-uf-always)(arrays-uf-never)" \
-  "(string-abstraction)(no-arch)(arch):" \
-  "(round-to-nearest)(round-to-plus-inf)(round-to-minus-inf)(round-to-zero)" \
-  OPT_FLUSH \
-  "(localize-faults)" \
-  OPT_GOTO_TRACE \
-  OPT_VALIDATE \
-  OPT_ANSI_C_LANGUAGE \
-  "(claim):(show-claims)(floatbv)(all-claims)(all-properties)" // legacy, and will eventually disappear // NOLINT(whitespace/line_length)
+const cmdline_definitiont ui_options
```
tautschnig, line 42: A rather unrelated comment: does anyone know why this ever needed to go in the header file? I.e., not just the changes in this PR, but also the current `CBMC_OPTIONS`?

src/util/cmdline_definition.h
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  typedef std::string deprecatedt;
+
+  cmdline_optiont(
+    std::string name,
+    optionalt<argumentt> argument,
+    std::string help_text,
+    std::string option_group,
+    optionalt<deprecatedt> deprecated);
```
tautschnig, line 40: Wouldn't it make for a very easy (and IMHO) nice extension to have a second constructor that takes a `std::function` for argument parsing (possibly achieving what @LAJW suggested)?

src/util/cmdline_definition.h
```diff
@@ -0,0 +1,70 @@
...
+
+    std::string name;
+    std::string type;
+  };
+
+  typedef std::string deprecatedt;
+
+  cmdline_optiont(
+    std::string name,
+    optionalt<argumentt> argument,
```
tautschnig, line 37: I don't think we should use `optionalt` here (or below, for `deprecated`), just have multiple constructors with varying numbers of parameters. The `optionalt` class members of course still remain in place.

src/util/cmdline_definition.h
```diff
@@ -0,0 +1,70 @@
...
+    std::string help_text,
+    std::string option_group,
+    optionalt<deprecatedt> deprecated);
+
+  std::string name;
+  optionalt<argumentt> argument;
+  std::string help_text;
+  std::string option_group;
+  optionalt<deprecatedt> deprecated;
```
tautschnig, line 46: I would like to propose one more member: `hidden`. At the moment, we have a number of command-line options that are intentionally undocumented, because either we consider them harmful (such as `--no-propagation`) or because they used to be supported and we just don't want to break scripts or other tooling still using them (such as `--floatbv`).

tautschnig (commented):
Thank you for starting to work on this! Yet I am surprised that #1521 isn't referenced here. Maybe this PR can even go all the way and eventually close that issue?

tautschnig:
Maybe we should actually have a class hierarchy under `cmdline_optiont`, including `cmdline_option_with_argumentt`, `deprecated_cmdline_optiont`, `hidden_cmdline_optiont`? `cmdline_optiont` would just provide some virtual functions so that printing and reading any set values can be achieved. But I guess for a start multiple constructors with a single class make things easier.

src/cbmc/cbmc_parse_options.h
peterschrammel: No idea :D

src/util/cmdline_definition.h
peterschrammel: That requires a bit more thinking as many options are not independent... I have a few ideas...

src/util/cmdline_definition.h
peterschrammel: Yes, although `{}` flows quite intuitively in the initialiser list...

src/util/cmdline_definition.h
tautschnig: Speaking of dependent options: it would actually be nice if one had a way to specify implied options (e.g., `--z3` implies `--smt2`). But I suppose the wish-list is unbounded (what about conflicting options...) 😇

PR #4395, opened 2019-03-16
SMT2 floating point encoding fix

kroening:
- [X] Each commit message has a non-empty body, explaining why the change was made.
- n/a Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4406, opened 2019-03-19
replace recursion in letify_rec by loop [blocks: #4395]

kroening:
This avoids stack overflow in the case of expressions with many unique subexpressions. PR #4395 is an exemplar.

### DIFF
--- a/src/solvers/smt2/letify.cpp
+++ b/src/solvers/smt2/letify.cpp
@@ -19,10 +19,12 @@ void letifyt::collect_bindings(
   seen_expressionst &map,
   std::vector<exprt> &let_order)
 {
+  // did we already see the expression?
   seen_expressionst::iterator entry = map.find(expr);
 
   if(entry != map.end())
   {
+    // yes, seen before, increase counter
     let_count_idt &count_id = entry->second;
     ++(count_id.count);
     return;
@@ -32,6 +34,7 @@ void letifyt::collect_bindings(
   if(expr.operands().empty())
     return;
 
+  // not seen before
   for(auto &op : expr.operands())
     collect_bindings(op, map, let_order);
 
@@ -46,6 +49,8 @@ void letifyt::collect_bindings(
   let_order.push_back(expr);
 }
 
+/// Construct a nested let expression for expressions
+/// in let_order that are used more than once
 exprt letifyt::letify(
   const exprt &expr,
   const std::vector<exprt> &let_order,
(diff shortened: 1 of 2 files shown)


peterschrammel (approved):

src/solvers/smt2/letify.cpp
```diff
@@ -46,26 +46,29 @@ void letifyt::collect_bindings(
...
-    map.find(current) != map.end(), "expression should have been seen already");
+  // go backwards in let order
+  for(auto r_it = let_order.rbegin(); r_it != let_order.rend(); r_it++)
+  {
+    const exprt &current = *r_it;
 
-  if(map.find(current)->second.count < LET_COUNT)
-    return letify_rec(expr, let_order, map, i + 1);
+    auto m_it = map.find(current);
+    INVARIANT(m_it != map.end(), "expression should have been seen already");
```
romainbrenguier: This invariant should be documented. It looks like a precondition of `letify`.

src/solvers/smt2/letify.h
```diff
@@ -46,11 +46,10 @@ class letifyt
     seen_expressionst &map,
     std::vector<exprt> &let_order);
 
-  static exprt letify_rec(
+  static exprt letify(
```
romainbrenguier, line 47: It's maybe a good occasion to document the function

src/solvers/smt2/letify.cpp
```diff
@@ -46,26 +46,29 @@ void letifyt::collect_bindings(
...
+
+    if(m_it->second.count >= LET_COUNT)
```
tautschnig: It would be nice to document this magic number. At least I have absolutely no idea why "2" is the right value, or indeed what it even is good for.

tautschnig (changes requested):
A request for documentation/aiding understanding below, but my bigger concern is the absence of newly viable tests. I believe we have a number of tests that are marked `broken-smt-backend` because of this stack overflow. If this fix does what it promises, then presumably it should be possible to enable those tests? I could be wrong, because there may be further problems down the line, but I'd at least like to see this discussed.

kroening:
@tautschnig The tests are in #4406!

src/solvers/smt2/letify.cpp
kroening: Two is really the only viable choice, and the symbolic name could well be replaced by 2.
"2" means that the expression is used more than once.
"3" (or higher) implies that the procedure becomes worst-case exponential.

tautschnig:
> @tautschnig The tests are in #4406!

(I suppose #4395, which solves that part of my request-for-changes. Will comment on the other one above.)

src/solvers/smt2/letify.cpp
tautschnig: Maybe using "1" instead of "2" would make things a lot clearer? That would make clear we are talking about something we have seen _more than once._ The named constant here really obfuscates this code, because 1) the name does not convey useful information and 2) one has to look elsewhere to find the value.

tautschnig: Adding to my comment moments ago: I think this one should be `if(m_it->second.count != 1)`.

tautschnig: ... and the other one in `substitute_let` is likely also best written as `!= 1`, mirroring this one.

martin-cs (approved):
Approve subject to other's comments.  This is something we do need though.

kroening:
Now with improved documentation

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7ed433fea28317e092c8d5189d923c995acdde23).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105593807

tautschnig (approved):

PR #4407, opened 2019-03-19
Symex code_deadt: always use L2 generation zero

smowton:
Now that we have #4199 and so decreasing the L2 generation number during symex is no longer a problem, we can simply use L2 generation 0 to represent all dead variables. Direct reads while the variable is certainly dead will still yield a nondet value, as generation 0 is never defined, while `merge_goto`'s special-casing of generation 0 (0 MERGE N == N for any nonzero generation N) can optimise the case where paths featuring a dead variable converge with those where it is still live, similar to how it already handles paths where a variable was not declared.

This is particularly useful for Java functions that throw, as symex will assume that code using `#return_value` flows from paths where it was defined (i.e. the normal return path), rather than considering the case where it is not defined (the exception-throwing path).

### DIFF
--- a/src/goto-symex/goto_symex_state.h
+++ b/src/goto-symex/goto_symex_state.h
@@ -223,14 +223,16 @@ public:
       l1_identifier, lhs, fresh_l2_name_provider);
   }
 
-  /// Increases the generation of the L1 identifier. Does nothing if there
-  /// isn't an expression keyed by it.
-  void increase_generation_if_exists(const irep_idt identifier);
-
   /// Drops an L1 name from the local L2 map
   void drop_l1_name(symex_renaming_levelt::current_namest::const_iterator it)
   {
     level2.current_names.erase(it);
+  }
+
+  /// Drops an L1 name from the local L2 map
+  void drop_l1_name(const irep_idt &l1_identifier)
+  {
+    level2.current_names.erase(l1_identifier);
   }
 
   std::function<std::size_t(const irep_idt &)> get_l2_name_provider() const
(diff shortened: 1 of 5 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c5ef60b170f3f7f33732cc1718806384a1b5cb62).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/104955360

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d201bad22dbd47e01e6f04ca77ddfcc6028f31c0).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105014174

peterschrammel (approved):

owen-mc-diffblue (approved):

src/goto-symex/goto_symex_state.h
```diff
@@ -223,16 +223,18 @@ class goto_symex_statet final : public goto_statet
...
   /// Drops an L1 name from the local L2 map
   void drop_l1_name(symex_renaming_levelt::current_namest::const_iterator it)
   {
     level2.current_names.erase(it);
   }
 
+  /// Drops an L1 name from the local L2 map
+  void drop_l1_name(const irep_idt &l1_identifier)
+  {
+    level2.current_names.erase(l1_identifier);
```
romainbrenguier, line 235: I would think these methods should belong to the `symex_level2t` class

smowton: I'll keep it alongside its close cousin `increase_generation` for now; if we want to refactor we should do them together (I don't mind either way)

PR #4408, opened 2019-03-19
multianewarray instruction: don't require a source line number

smowton:
Removing this doesn't appear to have any bad consequence, and line numbers cannot generally be relied upon, since compiling with -g:none or stripping all debug info means they won't be available.

Add a variant of the multianewarray test to check this works, and expand it slightly with comment-out parts that presumably at some point did not work, but do now.


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 35f0b2e9524d1639c98faa8419021f991e6a9f19).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105006212

kroening (approved):

peterschrammel (approved):

PR #4409, opened 2019-03-19
Allow `static` attribute to be stripped from symbols

karkhaz:
This PR is intended to help users that want to write CBMC proofs for individual functions. The typical way to do this is to write a 'harness' in a different file, which calls the function under test with unconstrained values. However, CBMC cannot see the definition of that function if it is marked `static` (in the C language. `static` in Java means something else). We would like to enable writing harnesses even for `static` functions, or functions that call `static` functions.

There are two issues solved in this PR.
1. `goto-cc` removes symbols from the symbol table if it determines that those symbols are not externally visible. This PR adds the `--keep-implementations` flag to `goto-cc`, which prevents it from removing those symbols from the symbol table.
1. Symbols naming functions defined as `static` have the `file_local` flag set. This PR adds the `--strip-static` flag to `goto-instrument`, which unsets the `file_local` flag on all symbols whose name is given as an argument.

So the intended workflow is: compile the file under test with goto-cc, passing in a list of functions that should be exposed. Then pass that same list of functions to goto-instrument so that it can unset the flag. This goto binary can then be successfully linked with a harness that calls the static function.


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 82a83e6d44e0c457b7486957fd604aa82be5033c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105034698

src/ansi-c/ansi_c_language.cpp
```diff
@@ -118,7 +118,8 @@ bool ansi_c_languaget::typecheck(
     return true;
   }
 
-  remove_internal_symbols(new_symbol_table);
```
smowton, line 121: The norm in the codebase is to pass `message_handlert` around rather than `messaget`. I suppose the idea is perhaps in the future an instance of `messaget` will denote a source of log messages, and a handler the means of printing those messages.

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -640,9 +640,10 @@ static void create_stub_global_symbols(
   }
 }
 
-bool java_bytecode_languaget::typecheck(
+bool java_bytecode_languaget::prv_typecheck(
```
smowton: Name it after what it does -- in this case, `typecheck_keeping_symbols`

regression/goto-cc-static/chain.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+goto_cc=$1
+cbmc=$2
```
smowton: The other commit changes this -- some aspect of them probably should be squashed (should this always have been goto-instrument?)

src/linking/remove_internal_symbols.cpp
```diff
@@ -147,6 +168,25 @@ bool remove_internal_symbols(symbol_tablet &symbol_table, messaget &log)
...
+  std::vector<irep_idt> diff;
+  std::set_difference(
+      user_kept.begin(),
+      user_kept.end(),
+      user_found.begin(),
+      user_found.end(),
+      std::inserter(diff, diff.begin()));
+  if(!diff.empty())
+  {
+    log.error() << "Several symbols requested to be kept were not found"
```
smowton: We could reasonably always warn on this (I agree it shouldn't usually be fatal)

smowton (commented):
Naming: `--strip-static` suggests a pass that removes static functions. How about `externalize-symbols`, since this is turning things with static linkage into ones with public linkage?

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 59d9c4f7848c07d4d0a4e56bbd4ba9d42075389b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105336054
Common spurious failures:

karkhaz:
@smowton I addressed all your suggestions, thanks.

My previous patchlist modified the interface of every subtype of `languaget`, which was a bit ugly. In addition to using your suggestion of changing `prv_typecheck` to something more descriptive, I changed the code so that only `ansi_c_languaget` needs to override the `languaget` API, other subtypes are no longer changed.

Could somebody please tell me why this is broken on Windows? It works on my local Windows box, but maybe I've done something wrong with goto-cl.

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1668,6 +1684,7 @@ void goto_instrument_parse_optionst::help()
     " --remove-function-pointers   replace function pointers by case statement over function calls\n" // NOLINT(*)
     HELP_REMOVE_CALLS_NO_BODY
     HELP_REMOVE_CONST_FUNCTION_POINTERS
+    HELP_STRIP_STATIC
```
martin-cs: I realise this is a pain but could this also be part of the option name update.

src/goto-instrument/goto_instrument_parse_options.h
```diff
@@ -83,6 +84,7 @@ Author: Daniel Kroening, kroening@kroening.com
   "(show-symbol-table)(show-points-to)(show-rw-set)" \
   "(cav11)" \
   OPT_TIMESTAMP \
+  OPT_STRIP_STATIC \
```
martin-cs: Likewise.

src/goto-instrument/externalize_symbols.h
```diff
@@ -0,0 +1,37 @@
...
+  bool externalize_symbols(const std::list<std::string> &);
+
+protected:
+  goto_modelt &model;
+  mutable messaget log;
+};
+
+#define OPT_STRIP_STATIC "(externalize-symbols):"
+#define HELP_STRIP_STATIC                                                      \
+  " --externalize-symbols fun[,fun...]  remove static from functions\n"
```
martin-cs: Given the many meanings of `static` could this say something like "removes file-local flag from functions".

martin-cs (approved):
The goto_instrument part seems reasonable.  I can see the wider problem and we do need a way of solving this.  I'm not 100% sure about the name "keep-implementations" as it seems a bit ambigious.  I also have the feeling that this could be simpler but I can't quite put my finger on how...

karkhaz:
@martin-cs thank you for your suggestions, I've just resolved them.

Making it simpler: one way of avoiding the goto-instrument pass altogether is to make the goto-cc flag do what I'm currently doing with goto-instrument. i.e., if you pass `--keep-implementations` to goto-cc, it will not only not remove the specified functions from the symbol table, but it would also unset the `file_local` flag. This was the approach I originally took---I'm happy to re-write it that way if you think it's better.

Naming of `--keep-implementations`: yeah, it's not great. Do you have suggestions? How about any of these:
* `--keep-functions`
* `--keep-file-local-functions`
* If we do both passes in goto-cc as I mentioned above, then the goto-cc option could be `--externalize-symbols` (which is currently what the goto-instrument option is)

karkhaz:
Two other things:
1. Could somebody please tell me why the windows build is failing, or let me know if I can get an AWS account that has access to that CI? I can't reproduce the failure on my windows box.
1. The `externalize_symbols.cpp` file is currently quite small, but I anticipate that I might add to it later. One possibility is that when you make `static` functions visible, there will be a multiple-definition error because of some other function with the same name in a different file. So one option to deal with this could be to name-mangle everything that we externalize, e.g. rename static function `foo` in file `bar.c` to `__CPROVER_file_local__bar__foo`. I don't plan to do this until it becomes a problem, though.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1d91fe0296858341dce76490061c962ebcbe9931).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105772029

martin-cs:
@karkhaz : Thanks for sorting this; various thoughts.

1. Is there a use case for just using either flag on their own?  I can't immediately see one but you are far more focused on this than me.  If not then maybe a flag for the C front-end which disregards the `static` keyword all together might be simpler, as you suggested.

2. I prefer `--keep-function-local-symbols` over `--keep-implementations`, would `--export-function-local-symbols` be even better?

3. Does this / should this / can this align with any GCC options : `-fkeep-static-functions` and `-fkeep-static-consts`.

4. I think it could be good to do the renaming ( `"__CPROVER_file_local_" + filename + "_" + function_name` maybe ) now because it will change the user interface of this feature.  I guess this might also affect which was this can be best done.

regression/goto-cc-static/chain.sh
```diff
@@ -0,0 +1,31 @@
...
+  fi
+fi
+
+name=${*:$#}
+name=${name%.c}
+
+if [[ "${is_windows}" == "true" ]]; then
+  "${goto_cc}" ${to_keep} "${name}.c"
+  mv "${name}.exe" "${name}.gb"
+else
```
smowton: To provide a double-check against some driver script wrongly supplying TRUE or 1 or something, suggest `elif [[ "${is_windows}" == "false" ]]; then` and error on anything else

src/linking/remove_internal_symbols.cpp
```diff
@@ -148,6 +169,22 @@ void remove_internal_symbols(
...
+  std::vector<irep_idt> diff;
+  std::set_difference(
+    user_kept.begin(),
+    user_kept.end(),
+    user_found.begin(),
+    user_found.end(),
+    std::inserter(diff, diff.begin()));
+  if(!diff.empty())
+  {
+    log.warning() << "Several symbols requested to be kept were not found"
```
smowton: Several symbols -> Symbol(s)

src/langapi/language_file.cpp
```diff
@@ -240,23 +253,29 @@ bool language_filest::typecheck_module(
...
   {
-    module.in_progress=false;
-    return true;
+    module.in_progress = !module.file->language->typecheck_keeping_symbols(
+      symbol_table, module.name, symbols_to_keep);
+  }
+  else
+  {
+    module.in_progress =
+      !module.file->language->typecheck(symbol_table, module.name);
```
smowton, line 272: If `symbols_to_keep` is non-empty we should surely at least warn here

smowton (approved):
A few more small fixes

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -642,7 +642,7 @@ static void create_stub_global_symbols(
 
 bool java_bytecode_languaget::typecheck(
   symbol_tablet &symbol_table,
-  const std::string &)
+  const std::string &module)
```
tautschnig: Unnecessary change?

regression/goto-cc-static/chain.sh
```diff
@@ -0,0 +1,31 @@
...
+    externalize="--externalize-symbols $4"
+  fi
+fi
+
+name=${*:$#}
+name=${name%.c}
+
+if [[ "${is_windows}" == "true" ]]; then
+  "${goto_cc}" ${to_keep} "${name}.c"
+  mv "${name}.exe" "${name}.gb"
```
tautschnig: You could use `"${goto_cc}" ${to_keep} "${name}.c" "/Fe${name}.exe"`

src/cpp/cpp_language.h
```diff
@@ -42,9 +42,8 @@ class cpp_languaget:public languaget
   bool generate_support_functions(
     symbol_tablet &symbol_table) override;
 
-  bool typecheck(
-    symbol_tablet &symbol_table,
-    const std::string &module) override;
+  bool
+  typecheck(symbol_tablet &symbol_table, const std::string &module) override;
```
tautschnig, line 46: Unnecessary change.

src/ansi-c/ansi_c_language.h
```diff
@@ -51,9 +51,21 @@ class ansi_c_languaget:public languaget
   bool generate_support_functions(
     symbol_tablet &symbol_table) override;
 
-  bool typecheck(
+  bool typecheck_keeping_symbols(
     symbol_tablet &symbol_table,
-    const std::string &module) override;
+    const std::string &module,
+    const std::set<std::string> &symbols_to_keep) override;
```
tautschnig: I'd use `irep_idt` instead of `std::string` at this point.

src/jsil/jsil_language.h
```diff
@@ -35,10 +35,10 @@ class jsil_languaget:public languaget
   virtual bool generate_support_functions(
     symbol_tablet &symbol_table) override;
 
-  virtual bool typecheck(
-    symbol_tablet &context,
-    const std::string &module) override;
+private:
+  bool typecheck(symbol_tablet &context, const std::string &module) override;
```
tautschnig: Unnecessary change?

src/linking/remove_internal_symbols.cpp
```diff
@@ -148,6 +169,22 @@ void remove_internal_symbols(
     }
   }
 
+  std::vector<irep_idt> diff;
+  std::set_difference(
```
tautschnig: I'd use `log.conditional_output(...)` for this entire block here as there is no need to compute a `set_difference` if warnings wouldn't even get printed.

src/linking/remove_internal_symbols.cpp
```diff
@@ -148,6 +169,22 @@ void remove_internal_symbols(
...
+    user_kept.begin(),
+    user_kept.end(),
+    user_found.begin(),
+    user_found.end(),
+    std::inserter(diff, diff.begin()));
```
tautschnig: You need to include `iterator` for this to work (on Windows).

tautschnig (approved):
Approving of the overall approach, but several comments (as posted earlier) need to be addressed.

karkhaz:
Thanks for all these comments. I'll need a few days to get this done.

I am in two minds about using an existing GCC option. Those options are obviously very appropriate. However, people may already have that option set in their existing build system, so adding behaviour to goto-cc that uses of that flag might break something...however, I am still inclined to use those flags, thanks for the suggestion.

Just to be clear, I wasn't proposing to make the C front-end disregard `static` entirely, I think it's better to export particular functions that the user requests only.

I'll implement name-mangling (and everything else) in goto-cc. This is because I will not only have to fix up the symbol table; I'll need to walk over the entire goto-program, mangling every call to the exported function too. It seems like a good idea to do this in goto-cc, which only ever sees one translation unit at a time, so I can safely rename every function I find without worrying what file it came from.

Question: it seems like I will need to recursively walk over irepts to do the name-mangling in the program. Is there an existing way of doing this, i.e. some sort of visitor-pattern thing in the CBMC codebase, or do I just do it manually? Asking because I did a similar thing for the linker script work (where I had to match multi-level nested irepts to change their type)...I remember having a conversation with @tautschnig long ago about having a tool called goto-grep to find particular shapes in irepts, and the natural extension of that would be to have goto-sed ;-)

smowton:
If they are `exprt`s the you can use `expr_iterator.h` which provides definitions for `exprt::depth_begin` and `_end`. This is a tree-walking iterator that breaks sharing only when its `mutate` method is called (`operator*` is read-only)

martin-cs:
On Tue, 2019-03-26 at 07:14 -0700, Kareem Khazem wrote:
> Thanks for all these comments. I'll need a few days to get this done.
> 
> I am in two minds about using an existing GCC option. Those options
> are obviously very appropriate. However, people may already have that
> option set in their existing build system, so adding behaviour to
> goto-cc that uses of that flag might break something...however, I am
> still inclined to use those flags, thanks for the suggestion.

I think you are right in that we should only use the gcc flags if we do
exactly the same thing.


> Just to be clear, I wasn't proposing to make the C front-end
> disregard `static` entirely, I think it's better to export particular
> functions that the user requests only.

OK.  I'm just not immediately seeing the need for only exporting some. 
Exporting all; you've convinced me of the use-case.  Some ... I remain
open minded and it seems like extra hassle if there isn't a use-case.

> I'll implement name-mangling (and everything else) in goto-cc. This
> is because I will not only have to fix up the symbol table; I'll need
> to walk over the entire goto-program, mangling every call to the
> exported function too. It seems like a good idea to do this in goto-
> cc, which only ever sees one translation unit at a time, so I can
> safely rename every function I find without worrying what file it
> came from.

Yes.  Sorry to ask for large changes but it will need to be done at
some point and it is better that it is now, with you giving it your
attention, than as a bug report.

> Question: it seems like I will need to recursively walk over irepts
> to do the name-mangling in the program. Is there an existing way of
> doing this, i.e. some sort of visitor-pattern thing in the CBMC
> codebase, or do I just do it manually?

There is `expr_visitort` in `util/expr.h` but `util/replace_symbol.h`
might suit your needs better.


>  Asking because I did a similar thing for the linker script work
> (where I had to match multi-level nested irepts to change their
> type)...I remember having a conversation with @tautschnig long ago
> about having a tool called goto-grep to find particular shapes in
> irepts, and the natural extension of that would be to have goto-sed
> ;-)

I tried to write something like goto-grep as one of my first CPROVER
projects.  It turns out to be much harder than you might think.  Maybe
I should try again at some point...

tautschnig:
> Question: it seems like I will need to recursively walk over irepts to do the name-mangling in the program. Is there an existing way of doing this, i.e. some sort of visitor-pattern thing in the CBMC codebase, or do I just do it manually? Asking because I did a similar thing for the linker script work (where I had to match multi-level nested irepts to change their type)...I remember having a conversation with @tautschnig long ago about having a tool called goto-grep to find particular shapes in irepts, and the natural extension of that would be to have goto-sed ;-)

Maybe `rename_symbolt` is all that you need?

karkhaz:
I'm going to leave the tests to run overnight, but I think I'm ready for review.

### Summary
- Everything is done in goto-cc now. There is no more goto-instrument pass.
- The flag is called `--export-function-local-symbols`, because there isn't actually a GCC flag called `-fkeep-static-functions` (you probably meant `-fkeep-inline-functions` or `-fkeep-static-consts`).
- If the user passes that flag, all static functions in the translation unit are kept and name-mangled.

### Name mangling
I've included two implementations. The active one mangles function `foo()` in `path/to/bar.c` into
```
__CPROVER_file_local_bar_c_foo()
```
which should be good enough almost all of the time. To deal with having the same-named function in a same-named file in a different part of the codebase, you can pass an argument to `--mangle-suffix`:
```
__CPROVER_file_local_bar_c_SUFFIX_foo()
```
this way, users decide whether they need a suffix. They can then pass a different suffix to each invocation of goto-cc. That way, all functions in the same file will be mangled with the same suffix, but different files will have different suffixes.

there is another implementation that hashes the pathname using Daniel J. Bernstein's hash, but it's not currently used. The implementation is chosen using templates, so there's no runtime cost to keeping both. The idea with that implementation is that it takes the entire file path into account, so there's no need for a suffix (but you can still provide one if you like). I turned it off because it's not really readable, I'd like harnesses to be readable.
```
__CPROVER_file_local_4a82bea1_foo()
```
### Misc
- `rename_symbolt` worked nicely, thank you!
- `regression/goto-cc-file-local/chain.sh` has become interesting. Usually we use the third line of `test.desc` to pass command-line args to whatever tool `chain.sh` is invoking, but I'm instead passing arguments to `chain.sh` itself to control what it does. Depending on the test, it sometimes links the goto-binary and sometimes doesn't, and then sometimes runs goto-instrument and other times runs cbmc.
- @martin-cs I just remembered why I wanted to only selectively export functions. The codebase I'm working on is full of function pointers. The more function implementations we have lying around, the more exponential blowup CBMC will encounter while exploring every possible function implementation that fits the signature of a function pointer. Nevertheless, I reasoned that this is such a serious performance problem that I should spend time actually fixing it, rather than merely avoiding making it worse by only exporting some functions. So I'll get round to fixing the function pointer problem one day.

karkhaz:
(assigning folks for re-review because I've changed the approach significantly. Thank you!)

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 067a6652e734ac24fbe40516f65023bf1fb19f66).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106670075

src/goto-programs/name_mangler.h
```diff
@@ -0,0 +1,155 @@
...
+  /// \brief Mangle all file-local function symbols in the program
+  ///
+  /// The way in which the symbols will be mangled is decided by which mangler
+  /// type this object is instantiated with, e.g. DJB_manglert mangles the path
+  /// name by hashing it.
+  void mangle()
+  {
+    rename_symbolt rename;
+    std::map<irep_idt, irep_idt> renamed_funs;
+    std::forward_list<symbolt> new_syms;
```
smowton: You're not using any list ops (e.g. splice), so suggest `std::vector`

src/goto-programs/name_mangler.h
```diff
@@ -0,0 +1,155 @@
...
+/// constructor and an `operator()` override with the following signature:
+///
+///     irep_idt operator()(const symbolt &, const std::string &);
+///
+/// The return type doesn't actually have to be an irep_idt, just something
+/// that can be assigned to one. The function is expected to return the
+/// mangled name of its \ref symbolt argument, incorporating the second
+/// argument into the mangled name if possible.
+template <class MangleFun>
+class name_manglert
```
smowton: Suggest `function_name_manglert`, as I supposed it really mangled all symbols at first

src/goto-programs/name_mangler.h
```diff
@@ -0,0 +1,155 @@
...
+        continue;
+      for(auto &ins : fun.second.body.instructions)
+      {
+        rename(ins.code);
+        rename(ins.guard);
+      }
+    }
+
+    // Add goto-programs with new function names
+    for(auto &pair : renamed_funs)
```
smowton: const

src/goto-programs/name_mangler.h
```diff
@@ -0,0 +1,155 @@
...
+protected:
+  const std::regex forbidden;
+  const std::regex multi_under;
+};
+
+/// \brief Mangle identifiers by hashing their working directory with djb2 hash
+///
+/// Hashes emitted by objects of this class include the leading 8 digits of the
+/// djb2 hash of the file path.
+class DJB_manglert
```
smowton: conventionally use lowercase class names even for acronyms

src/langapi/language.h
```diff
@@ -123,6 +124,32 @@ class languaget:public messaget
...
+  /// This function should only be called on objects for which a call to
+  /// can_keep_symbols() returns `true`.
+  virtual bool typecheck(
+    symbol_tablet &symbol_table,
+    const std::string &module,
+    const bool keep_file_local)
+  {
+    INVARIANT(
+      false,
+      "typecheck_keeping_symbols should only be called for files written"
```
smowton: `typecheck_keeping_symbols` no longer exists

src/langapi/language.h
```diff
@@ -123,6 +124,32 @@ class languaget:public messaget
     symbol_tablet &symbol_table,
     const std::string &module)=0;
 
+  /// \brief Is it possible to call typecheck_keeping_symbols() on this object?
+  virtual bool can_keep_file_local()
```
smowton, line 128: `typecheck_keeping_symbols` no longer exists

src/goto-programs/name_mangler.cpp
```diff
@@ -0,0 +1,48 @@
...
+operator()(const symbolt &src, const std::string &extra_info)
+{
+  char const *str = src.location.get_working_directory().c_str();
+  unsigned long hash = 5381;
+  int c;
+  while((c = *str++))
+    hash = ((hash << 5) + hash) + c;
+
+  // Make no more than 8 nibbles wide
+  while(hash > 0xffffffffl)
```
smowton: On a 32-bit UL arch this will surely generate a warning at least; suggest guarding by checking `sizeof(unsigned long)`

smowton: (Or just hard cast to `uint32_t`? Is there any particular value in keeping the high rather than the low bits of the computed hash?)

src/linking/remove_internal_symbols.h
```diff
@@ -12,7 +12,14 @@ Author: Daniel Kroening
 #ifndef CPROVER_LINKING_REMOVE_INTERNAL_SYMBOLS_H
 #define CPROVER_LINKING_REMOVE_INTERNAL_SYMBOLS_H
 
+#include <util/message.h>
+
+#include <set>
```
smowton: New includes but no apparent new users in this header file?

karkhaz:
Thanks @smowton, I've resolved all your suggestions.

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4438e777be0d9d3d4802a441d0ce0850d1ab7d55).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106771499

PR #4411, opened 2019-03-20
Memory snapshot harness lines

xbauch:
Before the user had to run the CBMC first to find out the location where they wanted to start their snaphot. This PR adds to option to specify the code line.

### DIFF
--- a/src/goto-harness/memory_snapshot_harness_generator.cpp
+++ b/src/goto-harness/memory_snapshot_harness_generator.cpp
@@ -36,32 +36,17 @@ void memory_snapshot_harness_generatort::handle_option(
   {
     memory_snapshot_file = require_exactly_one_value(option, values);
   }
-  else if(option == "initial-location")
+  else if(option == "initial-goto-location")
   {
-    const std::string initial_location =
-      require_exactly_one_value(option, values);
-
-    std::vector<std::string> start;
-    split_string(initial_location, ':', start, true);
-
-    if(
-      start.empty() || start.front().empty() ||
-      (start.size() == 2 && start.back().empty()) || start.size() > 2)
-    {
-      throw invalid_command_line_argument_exceptiont(
-        "invalid initial location specification", "--initial-location");
-    }
-
-    entry_function_name = start.front();
-
-    if(start.size() == 2)
-    {
-      location_number = optionalt<unsigned>(safe_string2unsigned(start.back()));
-    }
+    initial_goto_location_line = require_exactly_one_value(option, values);
   }
   else if(option == "havoc-variables")
   {
     variables_to_havoc.insert(values.begin(), values.end());
+  }
+  else if(option == "initial-source-location")
+  {
+    initial_source_location_line = require_exactly_one_value(option, values);
   }
   else
   {
--- a/src/goto-harness/memory_snapshot_harness_generator.cpp
+++ b/src/goto-harness/memory_snapshot_harness_generator.cpp
@@ -66,105 +66,50 @@ void memory_snapshot_harness_generatort::validate_options(
       "--harness-type initialise-from-memory-snapshot");
   }
 
-  if(entry_function_name.empty())
-  {
-    INVARIANT(
-      !location_number.has_value(),
-      "when `function` is empty then the option --initial-location was not "
-      "given and thus `location_number` was not set");
-
-    throw invalid_command_line_argument_exceptiont(
-      "option --initial-location is required",
-      "--harness-type initialise-from-memory-snapshot");
-  }
-
-  const auto &goto_functions = goto_model.goto_functions;
-  const auto &goto_function =
-    goto_functions.function_map.find(entry_function_name);
-  if(goto_function == goto_functions.function_map.end())
+  if(initial_source_location_line.empty() == initial_goto_location_line.empty())
   {
     throw invalid_command_line_argument_exceptiont(
-      "unknown initial location specification", "--initial-location");
+      "choose either source or goto location to specify the entry point",
+      "--initial-source/goto-location");
   }
 
-  if(!goto_function->second.body_available())
+  if(!initial_source_location_line.empty())
   {
-    throw invalid_command_line_argument_exceptiont(
-      "given function `" + id2string(entry_function_name) +
-        "` does not have a body",
-      "--initial-location");
+    entry_location = initialize_entry_via_source(
+      parse_source_location(initial_source_location_line),
+      goto_model.goto_functions);
   }
-
-  if(location_number.has_value())
+  else
   {
-    const auto &goto_program = goto_function->second.body;
-    const auto opt_it = goto_program.get_target(*location_number);
-
-    if(!opt_it.has_value())
-    {
-      throw invalid_command_line_argument_exceptiont(
-        "no instruction with location number " +
-          std::to_string(*location_number) + " in function " +
-          id2string(entry_function_name),
-        "--initial-location");
-    }
-  }
-
-  if(goto_functions.function_map.count(INITIALIZE_FUNCTION) == 0)
-  {
-    throw invalid_command_line_argument_exceptiont(
-      "invalid input program: " + std::string(INITIALIZE_FUNCTION) +
-        " not found",
-      "<in>");
+    entry_location = initialize_entry_via_goto(
+      parse_goto_location(initial_goto_location_line),
+      goto_model.goto_functions);
   }
 
   const symbol_tablet &symbol_table = goto_model.symbol_table;
+
   const symbolt *called_function_symbol =
-    symbol_table.lookup(entry_function_name);
+    symbol_table.lookup(entry_location.function_name);
 
   if(called_function_symbol == nullptr)
   {
     throw invalid_command_line_argument_exceptiont(
-      "function `" + id2string(entry_function_name) +
+      "function `" + id2string(entry_location.function_name) +
         "` not found in the symbol table",
       "--initial-location");
   }
 }
 
 void memory_snapshot_harness_generatort::add_init_section(
+  const symbol_exprt &func_init_done_var,
   goto_modelt &goto_model) const
 {
   goto_functionst &goto_functions = goto_model.goto_functions;
-  symbol_tablet &symbol_table = goto_model.symbol_table;
 
   goto_functiont &goto_function =
-    goto_functions.function_map[entry_function_name];
-  const symbolt &function_symbol = symbol_table.lookup_ref(entry_function_name);
+    goto_functions.function_map[entry_location.function_name];
 
   goto_programt &goto_program = goto_function.body;
-
-  // introduce a symbol for a Boolean variable to indicate the point at which
-  // the function initialisation is completed
-  symbolt &func_init_done_symbol = get_fresh_aux_symbol(
-    bool_typet(),
-    id2string(entry_function_name),
-    "func_init_done",
-    function_symbol.location,
-    function_symbol.mode,
-    symbol_table);
-  func_init_done_symbol.is_static_lifetime = true;
-  func_init_done_symbol.value = false_exprt();
-
-  const symbol_exprt func_init_done_var = func_init_done_symbol.symbol_expr();
-
-  // initialise func_init_done_var in __CPROVER_initialize if it is present
-  // so that it's FALSE value is visible before the harnessed function is called
-  goto_programt &cprover_initialize =
-    goto_functions.function_map.find(INITIALIZE_FUNCTION)->second.body;
-  cprover_initialize.insert_before(
-    std::prev(cprover_initialize.instructions.end()),
-    goto_programt::make_assignment(
-      code_assignt(func_init_done_var, false_exprt())));
 
   const goto_programt::const_targett start_it =
     goto_program.instructions.begin();
@@ -182,9 +127,8 @@ void memory_snapshot_harness_generatort::add_init_section(
   goto_program.compute_location_numbers();
   goto_program.insert_after(
     ins_it2,
-    goto_programt::make_goto(goto_program.const_cast_target(
-      location_number.has_value() ? *goto_program.get_target(*location_number)
-                                  : start_it)));
+    goto_programt::make_goto(
+      goto_program.const_cast_target(entry_location.start_instruction)));
 }
 
 code_blockt memory_snapshot_harness_generatort::add_assignments_to_globals(
@@ -310,12 +254,27 @@ void memory_snapshot_harness_generatort::generate(
   goto_functionst &goto_functions = goto_model.goto_functions;
 
   const symbolt *called_function_symbol =
-    symbol_table.lookup(entry_function_name);
+    symbol_table.lookup(entry_location.function_name);
 
-  add_init_section(goto_model);
+  // introduce a symbol for a Boolean variable to indicate the point at which
+  // the function initialisation is completed
+  auto &func_init_done_symbol = get_fresh_aux_symbol(
+    bool_typet(),
+    id2string(entry_location.function_name),
+    "func_init_done",
+    source_locationt::nil(),
+    called_function_symbol->mode,
+    symbol_table);
+  func_init_done_symbol.is_static_lifetime = true;
+  func_init_done_symbol.value = false_exprt();
+  symbol_exprt func_init_done_var = func_init_done_symbol.symbol_expr();
+
+  add_init_section(func_init_done_var, goto_model);
 
   code_blockt harness_function_body =
     add_assignments_to_globals(snapshot, goto_model);
+
+  harness_function_body.add(code_assignt{func_init_done_var, false_exprt{}});
 
   add_call_with_nondet_arguments(
     *called_function_symbol, harness_function_body);
--- a/src/goto-harness/memory_snapshot_harness_generator.cpp
+++ b/src/goto-harness/memory_snapshot_harness_generator.cpp
@@ -6,6 +6,8 @@ Author: Daniel Poetzl
 
 \******************************************************************/
 
+#include <algorithm>
+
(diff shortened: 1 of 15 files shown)


src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -69,6 +76,66 @@ void memory_snapshot_harness_generatort::handle_option(
...
+  const auto &goto_functions = goto_model.goto_functions;
+  if(line_number.has_value())
+  {
+    // by line: iterate over all instructions until source location match
+    for(const auto &entry : goto_functions.function_map)
+    {
+      const auto &goto_function = entry.second;
+
+      if(!goto_function.body_available())
+        continue;
```
tautschnig: That's actually unnecessary, the `std::find_if` will do the right thing either way.

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -69,6 +76,66 @@ void memory_snapshot_harness_generatort::handle_option(
...
+        goto_program.instructions.begin(),
+        goto_program.instructions.end(),
+        [this](const goto_programt::instructiont &instruction) {
+          return safe_string2unsigned(id2string(
+                   instruction.source_location.get_line())) >= *line_number;
+        });
+
+      if(it != goto_program.instructions.end())
+      {
+        entry_function_name = entry.first; //TODO - move elsewhere
```
tautschnig: I don't think I understand this `TODO`.

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -69,6 +76,66 @@ void memory_snapshot_harness_generatort::handle_option(
...
+      if(!goto_function.body_available())
+        continue;
+      const auto &goto_program = goto_function.body;
+
+      const auto it = std::find_if(
+        goto_program.instructions.begin(),
+        goto_program.instructions.end(),
+        [this](const goto_programt::instructiont &instruction) {
+          return safe_string2unsigned(id2string(
+                   instruction.source_location.get_line())) >= *line_number;
```
tautschnig: But then the same line number can occur in multiple files?

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -69,6 +76,66 @@ void memory_snapshot_harness_generatort::handle_option(
...
+    const auto &goto_function =
+      goto_functions.function_map.find(entry_function_name);
+    if(
+      goto_function != goto_functions.function_map.end() &&
+      goto_function->second.body_available())
+    {
+      const auto &goto_program = goto_function->second.body;
+      if(!location_number.has_value())
+        return goto_program.instructions.begin();
+      if(goto_program.instructions.size() > *location_number)
```
tautschnig: Why is this condition necessary?

src/goto-harness/memory_snapshot_harness_generator.h
```diff
@@ -34,6 +35,7 @@ Author: Daniel Poetzl
   "--initial-location <func[:<n>]>\n"                                          \
   "                              use given function and location number as "   \
   "entry\n                              point\n"                               \
+  "--initial-line <n>            use given line as entry point\n"              \
```
tautschnig: Can we actually use `file:line`? I think the line number alone is too ambiguous. You might actually end up picking an instruction from some system header file!

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Sorry, should have been removed before.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: I don't understand. I call `find_if` on the instructions inside `goto_function.body`. If the body is not `body_available` is it still safe to do?

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Because I'm advancing the iterator by `*location_number` position and I don't want to get past the `end()`.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Well `find_if` should return the first location with "high enough" line number. Is this still relevant if we insist on having the user specify the file as well?

src/goto-harness/memory_snapshot_harness_generator.h
xbauch: That's a very good idea. Added in the last commit. Which will also be squashed.

src/goto-harness/memory_snapshot_harness_generator.cpp
tautschnig: It is safe (`body_available()` just tests emptiness), but arguably that could change at some point in the future.

src/goto-harness/memory_snapshot_harness_generator.cpp
tautschnig: Ah, sorry, I missed that. But also this isn't how the `location_number` member of instructions works: these location numbers are guaranteed to be unique across _all_ goto functions, and monotonically increasing in a single function. Therefore a location number could well be larger than the size of a single function. At the same time, there could be gaps, so locations in a function could legitimately be numbered "7, 10, 15" instead of "1, 2, 3."

src/goto-harness/memory_snapshot_harness_generator.cpp
tautschnig: If the user can/has to specify the file then that's sorted, thanks!

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Ok, I removed it but left a comment.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: I give up. What about `find_if` to get the right instruction? That should be safe right?

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 1f0d7e9c2e3c8a1333545d729edb0626ebce28d8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105415411
Common spurious failures:

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: e2598d2d242dcba66c2226226cc787c82e232c4d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105422508
Common spurious failures:

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 948b5752df40dcd8bca1af57a740dc503689faf0).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105424064
Common spurious failures:

src/goto-programs/goto_program.h
```diff
@@ -588,28 +588,6 @@ class goto_programt
     return t;
   }
 
-  optionalt<const_targett> get_target(const unsigned location_number) const
```
martin-cs, line 591: I take it this was only used by goto-harness.

martin-cs (commented):
Seems like a reasonable feature.  Will approve once @tautschnig 's comments have been resolved.

src/goto-programs/goto_program.h
xbauch: It was used by the option `--initial-location` to find the instruction with this `location_number`. Now that we have two ways for the user to specify the entry point I generalised the search for the right instruction and thus `get_target` was no longer needed.

martin-cs: OK; sounds good.

src/goto-harness/memory_snapshot_harness_generator.cpp
tautschnig: Yes, I'm afraid you will have to do some form of search (technically binary search would be faster, but quite likely no one will notice the cost of linear search here).

tautschnig (approved):

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: But would binary search on a `std::list` really be faster? I thought that random access there is linear in the `distance` of the jump.

tautschnig: That's a question of how expensive the comparison operation is. Most likely you are right in that binary search would not add any value (or may indeed be more expensive).

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -61,6 +63,24 @@ void memory_snapshot_harness_generatort::handle_option(
   {
     variables_to_havoc.insert(values.begin(), values.end());
   }
+  else if(option == "initial-file-line")
```
hannes-steffenhagen-diffblue: :bike:  :derelict_house:  maybe `--initial-source-location`?

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -61,6 +63,24 @@ void memory_snapshot_harness_generatort::handle_option(
...
+  {
+    const std::string initial_file_line =
+      require_exactly_one_value(option, values);
+
+    std::vector<std::string> start;
+    split_string(initial_file_line, ':', start, true);
```
hannes-steffenhagen-diffblue: `split_string` has a special cased version for working with 2 values, maybe worth considering?

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -61,6 +63,24 @@ void memory_snapshot_harness_generatort::handle_option(
...
+
+    if(start.size() != 2 || start.front().empty() || start.back().empty())
+    {
+      throw invalid_command_line_argument_exceptiont(
+        "invalid initial location specification", "--initial-file-line");
+    }
+
+    initial_file_name = start.front();
```
hannes-steffenhagen-diffblue: :bulb:  might be worth wrapping these two in a struct to keep (and make obvious) the invariant that either both or neither of these are set

hannes-steffenhagen-diffblue (approved):
Looks good to me overall, not even really nitpicks just suggestion/questions

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Done.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Done.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: @hannes-steffenhagen-diffblue This required a little bigger change: the last 7 commits. Please have a look if its still acceptable.

src/goto-harness/memory_snapshot_harness_generator.h
```diff
@@ -69,6 +69,46 @@ class memory_snapshot_harness_generatort : public goto_harness_generatort
...
+  };
+
+  /// Wraps the information needed to identify the entry point. Initializes via
+  /// either \ref entry_goto_locationt or \ref entry_source_locationt
+  struct entry_locationt
+  {
+    irep_idt function_name;
+    optionalt<goto_programt::const_targett> start_instruction;
+
+    void initialize_via_goto(
```
hannes-steffenhagen-diffblue: I'd be more comfortable if these were builder/constructor functions, i.e. `entry_locationt make_from_goto_location(...);`, `entry_locationt make_from_source_location`

src/goto-harness/memory_snapshot_harness_generator.h
```diff
@@ -69,6 +69,46 @@ class memory_snapshot_harness_generatort : public goto_harness_generatort
...
 
 protected:
+  /// User provided goto location: function name and (maybe) location number;
+  /// the structure wraps this option with a parser
+  struct entry_goto_locationt
+  {
+    irep_idt function_name;
+    optionalt<unsigned> location_number;
+
+    entry_goto_locationt(const std::string &cmdl_option);
```
hannes-steffenhagen-diffblue: `explicit` - also, I'm personally not a huge fan of "do work" constructors, i.e. this one parses from string, I'd prefer if it was a function that said it did that in its name

src/goto-harness/memory_snapshot_harness_generator.h
```diff
@@ -69,6 +69,46 @@ class memory_snapshot_harness_generatort : public goto_harness_generatort
...
+  };
+
+  /// User provided source location: file name and line number; the structure
+  /// wraps this option with a parser
+  struct entry_source_locationt
+  {
+    irep_idt file_name;
+    unsigned line_number;
+
+    entry_source_locationt(const std::string &cmdl_option);
```
hannes-steffenhagen-diffblue: also `explicit` and the above comment

src/goto-harness/memory_snapshot_harness_generator.h
```diff
@@ -69,6 +69,46 @@ class memory_snapshot_harness_generatort : public goto_harness_generatort
...
+    optionalt<goto_programt::const_targett> start_instruction;
+
+    void initialize_via_goto(
+      const entry_goto_locationt &entry_goto_location,
+      const goto_functionst &goto_functions);
+    void initialize_via_source(
+      const entry_source_locationt &entry_source_location,
+      const goto_functionst &goto_functions);
+
+    bool valid() const
```
hannes-steffenhagen-diffblue: Ideally it shouldn't be possible to construct this in an invalid state, if it can be avoided

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -378,3 +378,113 @@ void memory_snapshot_harness_generatort::generate(
...
+  {
+    throw invalid_command_line_argument_exceptiont(
+      "invalid initial location specification", "--initial-location");
+  }
+
+  function_name = start.front();
+
+  if(start.size() == 2)
+  {
+    location_number = optionalt<unsigned>(safe_string2unsigned(start.back()));
```
hannes-steffenhagen-diffblue: this should be `string2optional_unsigned` instead, and you should check the result... if this function returns `nullopt` it means its argument wasn't a valid integer, which is something you don't want to silently ignore

(with `safe_string2unsigned` you're basically asserting that this conversion will always succeed which is not correct for user supplied strings)

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -378,3 +378,113 @@ void memory_snapshot_harness_generatort::generate(
...
+    cmdl_option, ':', initial_file_string, initial_line_string, true);
+
+  if(initial_file_string.empty() || initial_line_string.empty())
+  {
+    throw invalid_command_line_argument_exceptiont(
+      "invalid initial location specification", "--initial-file-line");
+  }
+
+  file_name = initial_file_string;
+  line_number = safe_string2unsigned(initial_line_string);
```
hannes-steffenhagen-diffblue: same as above

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -66,21 +66,43 @@ void memory_snapshot_harness_generatort::validate_options(
...
-  start_instruction = get_start_instruction(goto_model);
-  if(!start_instruction.has_value())
+  if(initial_source_location_line.empty() == initial_goto_location_line.empty())
   {
     throw invalid_command_line_argument_exceptiont(
-      "could not find the specified entry point", "--initial-location/line");
+      "choose either source or goto location to specify the entry point",
+      "--initial-source/goto-location");
+  }
+
```
hannes-steffenhagen-diffblue, line 75: I think this needs a check that only one of goto location and source location is set?

src/goto-harness/memory_snapshot_harness_generator.h
xbauch: Good idea: done.

src/goto-harness/memory_snapshot_harness_generator.h
xbauch: Done + constructors are now trivial.

src/goto-harness/memory_snapshot_harness_generator.h
xbauch: Done.

src/goto-harness/memory_snapshot_harness_generator.h
xbauch: Done.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Done.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Done.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: It's there somewhere (line 69).

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 0249ba059cebb09eba226f99f444c04de3779505).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106287968

PR #4288, opened 2019-02-27
Filter value sets in symex

owen-mc-diffblue:
When we process a goto or an assume instruction, we will sometimes be able to remove some entries from value-sets. This should make symex more efficient.

`value_sett::erase_value_from_entry` is inelegant - alternatives suggestions are appreciated.

I have not tested whether this improves performance, but I strongly suspect that it will in some cases.

I'll make separate PRs for the first five commits, as they are just clean ups.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4414, opened 2019-03-21
Fix MustNotThrow annotations and add invariant to avoid exception handling in __CPROVER_initialize

owen-mc-diffblue:


jbmc/src/java_bytecode/remove_exceptions.cpp
```diff
@@ -587,6 +589,14 @@ void remove_exceptionst::instrument_exceptions(
     }
   }
 
+  // INITIALIZE_FUNCTION should not contain any exception handling branches for
+  // two reasons: (1) with symex-driven lazy loading it means that code that
+  // references @inflight_exception might be generated before
+  // @inflight_exception is initialized; (2) symex can analyze
+  // INITIALIZE_FUNCTION much faster if it doesn't contain any branching.
+  INVARIANT(function_identifier != INITIALIZE_FUNCTION || !did_something,
+            INITIALIZE_FUNCTION "should not contain any exception handling branches");
```
smowton: `" should`

smowton (approved):
Once CI is passing, of course

jbmc/regression/jbmc/class-fields/java/lang/Class.java
```diff
@@ -4,6 +4,7 @@
 
   public Integer field;
 
+  @org.cprover.MustNotThrow
```
tautschnig, line 7: Wouldn't you also need to include updated .class files in the commit?

jbmc/src/java_bytecode/remove_exceptions.cpp
```diff
@@ -587,6 +589,14 @@ void remove_exceptionst::instrument_exceptions(
     }
   }
 
+  // INITIALIZE_FUNCTION should not contain any exception handling branches for
+  // two reasons: (1) with symex-driven lazy loading it means that code that
+  // references @inflight_exception might be generated before
+  // @inflight_exception is initialized; (2) symex can analyze
+  // INITIALIZE_FUNCTION much faster if it doesn't contain any branching.
+  INVARIANT(function_identifier != INITIALIZE_FUNCTION || !did_something,
```
tautschnig: Given that seemingly you had to modify the source code of a test to make this `INVARIANT` pass I am wondering whether this really is an invariant?

owen-mc-diffblue: What macro would capture my intent better? Quite a lot of work went into making sure that branches for exception-handling weren't created in `__CPROVER_initialize`. We want to make sure it doesn't creep back in by having an assertion which fails when it does.

owen-mc-diffblue:
I should have marked this `do not review` - I will fix it up when I have finished #4288 .

jbmc/src/java_bytecode/remove_exceptions.cpp
tautschnig: If user input can violate this constraint (which I thought was possible, given you had to change a test), then it should be an exception (for example ` unsupported_operation_exceptiont`).

owen-mc-diffblue: The "user input" in those tests is a cut-down model for java.lang.Class, implementing special CPROVER functions like `cproverNondetInitialize` or `cproverInitializeClassLiteral`. I don't think we expect users to be providing anything like that.

tautschnig: Oh, I see, I should probably just remain quiet about any Java-related code :-)  Thanks for clarifying this, the invariant does seem to make a lot of sense in that case.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 5aaaf25a58a6bc9e15d9a56d35365a7d323fee59).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105834354

owen-mc-diffblue:
This is now ready for review

smowton (approved):
I'd switch `_GOTOS` for `_MAY_THROW` or similar wording, since it's throwing rather than gotos that are actually the problem. Otherwise lgtm.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 58ded6248a60cc865b6ffc56f7e4546cbd92cd5c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105890480

PR #4415, opened 2019-03-21
Improve compilation with cmake Ninja

romainbrenguier:
This make cudd be compiled at the configuration step, so that it is not missing when we build with Ninja.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,29 +46,20 @@ if(DEFINED CMAKE_USE_CUDD)
     download_project(PROJ cudd
             URL https://sourceforge.net/projects/cudd-mirror/files/cudd-3.0.0.tar.gz/download
             URL_MD5 4fdafe4924b81648b908881c81fe6c30
-            CONFIGURE_COMMAND ./configure
-            BUILD_COMMAND make
             )
 
-    if(EXISTS ${cudd_SOURCE_DIR}/Makefile)
-        add_custom_target(cudd-make
-                COMMAND make
-                WORKING_DIRECTORY ${cudd_SOURCE_DIR}
-                )
-    else()
-        add_custom_target(cudd-make
-                COMMAND ./configure && make
-                WORKING_DIRECTORY ${cudd_SOURCE_DIR}
-                )
+    if(NOT EXISTS ${cudd_SOURCE_DIR}/Makefile)
+        message(STATUS "Configuring Cudd-3.0.0")
+        execute_process(COMMAND ./configure WORKING_DIRECTORY ${cudd_SOURCE_DIR})
     endif()
-
-    add_library(cudd-cplusplus STATIC IMPORTED)
-    add_dependencies(cudd-cplusplus cudd-make)
-    set_property(TARGET cudd-cplusplus PROPERTY IMPORTED_LOCATION ${cudd_SOURCE_DIR}/cplusplus/.libs/libobj.a)
+    message(STATUS "Building Cudd-3.0.0")
+    execute_process(COMMAND make WORKING_DIRECTORY ${cudd_SOURCE_DIR})
 
     add_library(cudd STATIC IMPORTED)
-    add_dependencies(cudd cudd-make)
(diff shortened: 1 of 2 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: a70d673c3428e212df7cf01d0f2879eb9d345265).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105334919
Common spurious failures:

src/solvers/CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ if("${sat_impl}" STREQUAL "minisat2")
         URL http://ftp.debian.org/debian/pool/main/m/minisat2/minisat2_2.2.1.orig.tar.gz
         PATCH_COMMAND patch -p1 -i ${CBMC_SOURCE_DIR}/../scripts/minisat-2.2.1-patch
         COMMAND cmake -E copy ${CBMC_SOURCE_DIR}/../scripts/minisat2_CMakeLists.txt CMakeLists.txt
+        URL_MD5 27faa19ee0508660bd6fb7f894646d42
```
tautschnig, line 70: That's a very nice addition! But then: is this really the only place where this is to be done? At least Glucose a few lines further down should see the same.

CMakeLists.txt
```diff
@@ -50,26 +50,19 @@ if(DEFINED CMAKE_USE_CUDD)
...
+    if(NOT EXISTS ${cudd_SOURCE_DIR}/Makefile)
+        message(STATUS "Configuring Cudd-3.0.0")
+        execute_process(COMMAND ./configure WORKING_DIRECTORY ${cudd_SOURCE_DIR})
     endif()
-
-    add_library(cudd-cplusplus STATIC IMPORTED)
-    add_dependencies(cudd-cplusplus cudd-make)
-    set_property(TARGET cudd-cplusplus PROPERTY IMPORTED_LOCATION ${cudd_SOURCE_DIR}/cplusplus/.libs/libobj.a)
+    message(STATUS "Building Cudd-3.0.0")
+    execute_process(COMMAND make WORKING_DIRECTORY ${cudd_SOURCE_DIR})
```
tautschnig, line 56: Hmm, how is it actually possible that CUDD is _not_ built at this time, given `download_project` above already has a `BUILD_COMMAND`? Or is the above `download_project` in fact invalid, if I read https://github.com/Crascit/DownloadProject/blob/master/DownloadProject.cmake correctly? That documentation says that `BUILD_COMMAND` and `CONFIGURE_COMMAND` are not permitted in `download_project`?!

tautschnig (approved):
Approving with some further comments:
1) If `download_project` in fact isn't triggering a build of CUDD, then the `download_project` invocation should be fixed in this PR, see my comments below.
2) Building CUDD at CMake configuration time does seem right to me, and also matches the spirit of our Makefile-based builds.

tautschnig:
I think it would be good if @chrisr-diffblue or @hannes-steffenhagen-diffblue could also take a look before this one is merged.

CMakeLists.txt
romainbrenguier: That explains why it wasn't working then, thanks! In that case I guess we could use ExternalProject directly https://cmake.org/cmake/help/v3.10/module/ExternalProject.html instead of DownloadProject

smowton (approved):

src/solvers/CMakeLists.txt
romainbrenguier: done

CMakeLists.txt
romainbrenguier: I haven't managed to make ExternalProject work as I want (it seems `./configure` can't get executed) so I've just cleaned up the BUILD_COMMAND and CONFIGURE_COMMAND which have no effect.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: b79624836f16b11caa43180d90dc0dcdc9c60775).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105446216

peterschrammel (approved):

PR #4423, opened 2019-03-23
Add the deprecation date to each deprecated entity [blocks: #3768, #4248, #4368, #4371]

tautschnig:
This makes the date of deprecation visible in warning messages and also
makes tracking much easier: git grep DEPRECATED or any code review will
yield immediate insight on how long an entity has been deprecated
already.

The date is taken from the commit that first marked the entity as
deprecated.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- n/a Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- n/a Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4424, opened 2019-03-23
Coding standard: Deprecation rules

tautschnig:
Thus far, we operated an informal deprecation policy. This codifies our
informal six-months rule.

This has a soft dependency on #4423 as the `SINCE` macro will only be available once that PR is merged.

### DIFF
--- a/CODING_STANDARD.md
+++ b/CODING_STANDARD.md
@@ -1,5 +1,15 @@
 Here a few minimalistic coding rules for the CPROVER source tree.
 
+# Interfaces
+- With any changes, consider the impact on other users of the code base. Users
+  frequently carry their own set of patches or build tools on top of the code
+  base. Large-scale changes negatively impact both scenarios.
+- Tools that link against the code base can reasonably expect a stable
+  interface. We consider as public interface any objects, procedures, or classes
+  (and their methods) that are meant to be used outside a single directory.
+- See below for how to document interfaces and how to mark parts of interfaces
+  as deprecated.
+
 # Whitespaces
 
 Formatting is enforced using clang-format. For more information about this, see
--- a/CODING_STANDARD.md
+++ b/CODING_STANDARD.md
@@ -101,6 +101,16 @@ Formatting is enforced using clang-format. For more information about this, see
 - Use comments to explain the non-obvious
 - Use #if 0 for commenting out source code
 - Use #ifdef DEBUG to guard debug code
+- When deprecating interfaces, use the `DEPRECATED` macro, preferably together
+  with `SINCE`. For example,
+  ```
+  DEPRECATED(SINCE(2019, 1, 31, "use other_method() instead"))
+  void deprecated_method();
+  ```
+  marks `deprecated_method` as deprecated as of 2019-01-31. Any deprecated
+  functionality should remain in place for at least six months from the date of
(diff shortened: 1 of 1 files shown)


peterschrammel (commented):
What's the scope of this rule? `code` seems too general to be a useful guidance.

tautschnig:
> `code` seems too general to be a useful guidance.

Good point - would "API" work? Or "public API?"

peterschrammel:
What that be anything declared in a header file?

tautschnig:
> What that be anything declared in a header file?

In C++ it's really hard to pin down what is/isn't (part of) an API. I'd say anything that is meant to be used outside a module, where I would consider as "module" a directory.

CODING_STANDARD.md
```diff
@@ -101,6 +101,15 @@ Formatting is enforced using clang-format. For more information about this, see
...
 - Use #ifdef DEBUG to guard debug code
+- When deprecating code, use the `DEPRECATED` macro, preferably together with
+  `SINCE`. For example,
+  ```
+  DEPRECATED(SINCE(2019, 1, 31, "use other_method() instead"))
+  void deprecated_method();
+  ```
+  marks `deprecated_method` as deprecated as of 2019-01-31. Any deprecated
+  functionality should remain in place for at least six months from the date of
+  deprecation. Before deprecating code, all in-tree uses should be replaced.
```
kroening, line 124: I would perhaps add "or marked as deprecated".

kroening (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: b39726bcbff8db8afcad22267c60fefcd1f9062b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105538719

peterschrammel:
> In C++ it's really hard to pin down what is/isn't (part of) an API.

That's not a property of C++, but of how we structure the code base.

> meant to be used outside a module

Please add this definition to make it a bit clearer.

peterschrammel (approved):

tautschnig:
I have made some changes to the wording and added a second commit that codifies the idea of an interface.

tautschnig:
> That's not a property of C++, but of how we structure the code base.

We might discuss this a bit more, but that's likely more efficient over drinks than via GitHub...

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: f3a837166adc4ac89cc2b5924151aa07696120ff).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105539964

martin-cs:
How does this fit with the release cycle?  If we are deprecating user-facing functionality should we give them at least one release where it has a warning?

tautschnig:
> How does this fit with the release cycle?

I think we were aiming for one release per quarter, so we should have one or two releases in between. But I'm not sure we are currently on track for quarterly releases...

martin-cs:
On Sat, 2019-03-23 at 15:20 -0700, Michael Tautschnig wrote:
> > How does this fit with the release cycle?
> 
> I think we were aiming for one release per quarter, so we should have
> one or two releases in between. But I'm not sure we are currently on
> track for quarterly releases...

This seems reasonable.  I guess there is a bit of a difference between
internal APIs used for tools based on CPROVER and user-facing features.

PR #4426, opened 2019-03-23
Sharing map refactoring to simplify usage [TG-6478]

danpoe:
This changes the interface of the sharing map to make it easier to use, by making it harder to accidentally break sharing. Now the only method where sharing might be broken unnecessarily is `replace()`, which replaces the mapped value for an item already existing in the map. In addition, a debug mode is provided (enabled via the boolean template parameter `check_equal`) to detect attempts to replace an old value with a new equal value. For this to work, the value type used must have a defined equality operator. As a trade-off, the new interface now deviates further from the interface of `std::map`.

### DIFF
--- a/src/util/sharing_map.h
+++ b/src/util/sharing_map.h
@@ -221,18 +221,61 @@ protected:
 public:
   // interface
 
+  /// Erase element
+  ///
+  /// Complexity:
+  /// - Worst case: O(H * S + M)
+  /// - Best case: O(H)
+  ///
+  /// \param k: The key of the element to erase
+  /// \param key_exists: Hint to indicate whether the element is known to exist
+  ///   (possible values `unknown` or` true`)
   size_type erase(const key_type &k, const tvt &key_exists = tvt::unknown());
 
+  /// Erase all elements
+  ///
+  /// Complexity:
+  /// - Worst case: O(K * (H * S + M))
+  /// - Best case: O(K * H)
+  ///
+  /// \param ks: The keys of the element to erase
+  /// \param key_exists: Hint to indicate whether the elements are known to
+  ///   exist (possible values `unknown` or `true`). Applies to all elements
+  ///   (i.e., have to use `unknown` if for at least one element it is not known
+  ///   whether it exists)
   size_type erase_all(
     const keyst &ks,
     const tvt &key_exists = tvt::unknown()); // applies to all keys
 
+  /// Insert element, element must not exist in map
+  ///
+  /// Complexity:
+  /// - Worst case: O(H * S + M)
+  /// - Best case: O(H)
+  ///
+  /// \param k: The key of the element to insert
+  /// \param m: The mapped value to insert
   template <class valueU>
   void insert(const key_type &k, valueU &&m);
 
+  /// Replace element, element must exist in map
+  ///
+  /// Complexity:
+  /// - Worst case: O(H * S + M)
+  /// - Best case: O(H)
+  ///
+  /// \param k: The key of the element to insert
+  /// \param m: The mapped value to replace the old value with
   template <class valueU>
   void replace(const key_type &k, valueU &&m);
-
+  /// Find element
+  ///
+  /// Complexity:
+  /// - Worst case: O(H * log(S) + M)
+  /// - Best case: O(H)
+  ///
+  /// \param k: The key of the element to search
+  /// \return optionalt containing a const reference to the value if found
   optionalt<std::reference_wrapper<const mapped_type>>
   find(const key_type &k) const;
 
@@ -315,8 +358,51 @@ public:
   /// that are not shared between them (also see get_delta_view()).
   typedef std::vector<delta_view_itemt> delta_viewt;
 
+  /// Get a view of the elements in the map
+  /// A view is a list of pairs with the components being const references to
+  /// the keys and values in the map.
+  ///
+  /// Complexity:
+  /// - Worst case: O(N * H * log(S))
+  /// - Best case: O(N + H)
+  ///
+  /// \param [out] view: Empty view
   void get_view(viewt &view) const;
 
+  /// Get a delta view of the elements in the map
+  ///
+  /// Informally, a delta view of two maps is a view of the key-value pairs in
+  /// the maps that are contained in subtrees that are not shared between them.
+  ///
+  /// A delta view is represented as a list of structs, with each struct having
+  /// four members (`in_both`, `key`, `value1`, `value2`). The elements `key`,
+  /// `value1`, and `value2` are const references to the corresponding elements
+  /// in the map. The first element indicates whether the key exists in both
+  /// maps, the second element is the key, the third element is the mapped value
+  /// of the first map, and the fourth element is the mapped value of the second
+  /// map, or a dummy element if the key exists only in the first map (in which
+  /// case `in_both` is false).
+  ///
+  /// Calling `A.delta_view(B, ...)` yields a view such that for each element in
+  /// the view one of two things holds:
+  /// - the key is contained in both A and B, and in the maps the corresponding
+  ///   key-value pairs are not contained in a subtree that is shared between
+  ///   them
+  /// - the key is only contained in A
+  ///
+  /// When `only_common=true`, the first case above holds for every element in
+  /// the view.
+  ///
+  /// Complexity:
+  /// - Worst case: O(max(N1, N2) * H * log(S) * M1 * M2) (no sharing)
+  /// - Best case: O(1) (maximum sharing)
+  ///
+  /// The symbols N1, M1 refer to map A, and symbols N2, M2 refer to map B.
+  ///
+  /// \param other: other map
+  /// \param [out] delta_view: Empty delta view
+  /// \param only_common: Indicates if the returned delta view should only
+  ///   contain key-value pairs for keys that exist in both maps
   void get_delta_view(
     const sharing_mapt &other,
     delta_viewt &delta_view,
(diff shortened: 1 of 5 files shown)


src/util/sharing_node.h
```diff
@@ -356,8 +356,20 @@ SN_TYPE_PAR_DECL class d_leaft : public small_shared_pointeet<unsigned>
...
+  }
+  d_leaft(const std::shared_ptr<keyT> &k, valueT &&v) : k(k), v(std::move(v))
+  {
+  }
   std::shared_ptr<keyT> k;
 #else
+  d_leaft(const keyT &k, const valueT &v) : k(k), v(v)
+  {
+  }
+  d_leaft(const keyT &k, valueT &&v) : k(k), v(std::move(v))
```
tautschnig: Instead of duplicating all the declarations, couldn't we just a single `typedef` (or `using`) that holds the effective key type? After all `k` is a public member as well, and has that effective key type, so it would be beneficial to just consistently use that typedef name.

src/util/sharing_map.h
```diff
@@ -92,21 +98,25 @@ Author: Daniel Poetzl
...
+/// pair is replaced) and its position is in a shared subtree, already existing
+/// nodes from the root of the subtree to the position of the key-value pair are
+/// copied and integrated with the map, and new nodes are created as needed.
 /// - When a key-value pair is erased from the map that is in a shared subtree,
 /// nodes from the root of the subtree to the last node that will still exist on
 /// the path to the erased element after the element has been removed are
 /// copied and integrated with the map, and the remaining nodes are removed.
 ///
+/// The replace() operation is the only method where sharing could unnecessarily
+/// be broken. This would happen when replacing and old value with a new equal
```
tautschnig: s/and/an/

src/util/sharing_map.h
```diff
@@ -92,21 +98,25 @@ Author: Daniel Poetzl
...
+/// value. The sharing map provides a debug mode to detect such cases. When the
+/// template parameter `check_equal` is set to true, then the replace() method
```
tautschnig: I'd find `fail_if_equal` a bit more descriptive.

src/util/sharing_map.h
```diff
@@ -887,105 +877,46 @@ ::erase_all(const keyst &ks, const tvt &key_exists)
...
+  PRECONDITION(lp != nullptr); // key must exist in map
 
-  if(lp != nullptr)
-    return find_type(lp->get_value(), false);
-
-  lp = cp->place_leaf(k, m);
-  num++;
+  INVARIANT(
+    value_equalt()(as_const(lp)->get_value(), m),
+    "values should not be replaced with equal values to maximize sharing");
```
tautschnig, line 969: Wouldn't a slightly less fancy `if(check_equal) ...` also do the trick instead of the `struct True ...` above?

src/util/sharing_map.h
```diff
@@ -995,70 +926,16 @@ SHARING_MAPT2(const, find_type)::find(const key_type &k, const tvt &key_exists)
...
-/// \return The mapped value
-SHARING_MAPT2(, mapped_type &)::at(
-  const key_type &k,
-  const tvt &key_exists)
-{
-  find_type r=find(k, key_exists);
-
-  if(!r.second)
-    throw std::out_of_range(not_found_msg);
+    return nullopt;
```
tautschnig: I think we had a discussion elsewhere: `return {};` appears to be more common than the use of `nullopt`.

unit/util/sharing_node.cpp
```diff
@@ -24,10 +24,10 @@ void sharing_node_internals_test()
     {
       leaft leaf(1, 2);
 
-      auto p = leaf.data.get();
+      /*auto p = leaf.data.get();
       leaf.write();
 
-      REQUIRE(leaf.data.get() == p);
+      REQUIRE(leaf.data.get() == p);*/
```
tautschnig: What to make of this? In fact the coding guidelines forbid the use of `/* ... */`. But also if you were to use `#if 0` instead I'd have no idea what is going on here. Needs comments.

tautschnig (approved):

unit/util/sharing_node.cpp
danpoe: Ah yes this code is obsolete now. I've removed it.

src/util/sharing_map.h
danpoe: This wouldn't compile when `valueT` doesn't have a defined equality operator, as the code in the `if(check_equal) ...` branch is still parsed when `check_equal` is false.

tautschnig: Well, note that it does always parse, it just wouldn't type check. And I don't think code that isn't instantiated would be subject to type checking (as in general that wouldn't work).

danpoe: A branch dependent on a boolean template parameter is still typechecked even when statically false. I've had this issue several times before.

tautschnig: Ok, thanks!

src/util/sharing_node.h
danpoe: Done

src/util/sharing_map.h
```diff
@@ -38,26 +40,46 @@ Author: Daniel Poetzl
...
+  template < \
+    typename keyT, \
+    typename valueT, \
+    bool fail_if_equal, \
+    typename hashT, \
+    typename equalT> \
+  CV typename sharing_mapt<keyT, valueT, fail_if_equal, hashT, equalT>::ST \
+    sharing_mapt<keyT, valueT, fail_if_equal, hashT, equalT>
 
 #define SHARING_MAPT3(T, CV, ST) \
```
smowton: This macros could use better names, Which one do I want, and why?

src/util/sharing_map.h
```diff
@@ -887,105 +893,46 @@ ::erase_all(const keyst &ks, const tvt &key_exists)
...
 ///
 /// \param k: The key of the element to insert
 /// \param m: The mapped value to insert
-/// \param key_exists: Hint to indicate whether the element is known to exist
-///   (possible values `false` or `unknown`)
-/// \return Pair of const reference to existing or newly inserted element, and
-///   boolean indicating if new element was inserted
-SHARING_MAPT2(const, const_find_type)
-::insert(const key_type &k, const mapped_type &m, const tvt &key_exists)
+SHARING_MAPT4(U, void)
```
smowton: Except for a single `<T>`, I think you should avoid single-character template parameter names. This one looks like ValueT perhaps?

unit/util/sharing_map.cpp
```diff
@@ -16,15 +16,15 @@ Author: Daniel Poetzl
 #include <testing-utils/use_catch.h>
 #include <util/sharing_map.h>
 
-class smt : public sharing_mapt<irep_idt, std::string, irep_id_hash>
+typedef sharing_mapt<irep_idt, std::string, false, irep_id_hash> smt;
+
+class sm2t : public smt
```
smowton: Give this a longer name, looks too much like `smt2` at first sight

smowton (approved):
Some abbreviation / naming I suggest improving, otherwise lgtm

src/util/sharing_map.h
danpoe: Yes this type is for use in the forwarding reference for the value. I'll rename it to `valueU` (`valueT` is already taken here).

src/util/sharing_map.h
danpoe: Hm I tried that at some point but it turned out to be difficult to distill the differences between the macros into meaningful names. I would suggest to keep the current names and I'll add some documentation instead.

tautschnig: Maybe then we shouldn't have that macro in the first place if it isn't possible to name its parameters in a meaningful way? Then really the information exclusively lies with the points where the macro is used. In that case I'd suggest to just expand the macro at all uses, at the expense of some code duplication.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 97965017c59d803fc2642d3ab668959e74c69e6f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105691473

src/util/sharing_map.h
danpoe: I can give the macro parameters better names and document the parameters and the macros. Just giving each macro a name that describes what it does and at the same time makes the distinctions between them clear seems hard. I would prefer though having the macros just to abbreviate those template expressions as otherwise the definitions of the methods and static variables become kind of cluttered.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1325f9854844ccd28e390445f8e8c8fc1f83041b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105822265

src/util/sharing_map.h
tautschnig: I'm happy with whatever makes this code easier to use. The sharing maps are really useful and efficient, but they are very hard to use, because their API is _not_ very easy to read.

danpoe: I agree, the sharing map is somewhat hard to use. My goal is that with this PR it will get easier. The macros above only appear in the definitions btw so don't show up in the interface declarations. Also I'll move the documentation of the public methods from the definitions to the declarations.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: e9943357d2db70e91f62ecbfb0b51d29729e9cd1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105854268

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 72457bdfaa91517973f5fd50ba2057aa8f0e93d5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105875796

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 19d8f0927502427c051732f58114f20a57422a76).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105896987

PR #4437, opened 2019-03-26
Use symbol_tablet::lookup_ref and ::get_writeable_ref

smowton:
These all replace uses of lookup and get_writeable that were immediately dereferenced without
being checked, thereby replacing any possible failure due to reading through a null pointer with
an invariant failure.

### DIFF
--- a/src/goto-instrument/accelerate/acceleration_utils.cpp
+++ b/src/goto-instrument/accelerate/acceleration_utils.cpp
@@ -227,7 +227,7 @@ void acceleration_utilst::stash_variables(
       it!=vars.end();
       ++it)
   {
-    symbolt orig=*symbol_table.lookup(*it);
+    symbolt orig = symbol_table.lookup_ref(*it);
     symbolt stashed_sym=fresh_symbol("polynomial::stash", orig.type);
     substitution[orig.symbol_expr()]=stashed_sym.symbol_expr();
     program.assign(stashed_sym.symbol_expr(), orig.symbol_expr());
--- a/src/goto-instrument/accelerate/acceleration_utils.cpp
+++ b/src/goto-instrument/accelerate/acceleration_utils.cpp
@@ -227,7 +227,7 @@ void acceleration_utilst::stash_variables(
       it!=vars.end();
       ++it)
   {
-    symbolt orig = symbol_table.lookup_ref(*it);
+    const symbolt &orig = symbol_table.lookup_ref(*it);
     symbolt stashed_sym=fresh_symbol("polynomial::stash", orig.type);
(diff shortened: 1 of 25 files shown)


martin-cs (commented):
This seems like a worth-while thing to do and the changes to my area are fine BUT could we make this simpler and more stringent?  Are there any legitimate uses of `get_writeable` where it is null?  Couldn't this become an invariant in general?  Likewise what exactly are the invariants on `lookup`?

smowton:
Users of `get_writeable`:
`src/jsil/jsil_typecheck.cpp`: to throw rather than raise an invariant
`src/goto-cc/linker_script_merge.cpp`: to skip undefined symbols
`src/cpp/cpp_typecheck_template.cpp`: to replace / consistency-check repeated template definitions
`src/cpp/cpp_declarator_converter.cpp`: to retry with an adjusted name on failure

So there aren't many users but they do exist. Unsure what you're asking about `lookup`, which is a plain old search function.

src/goto-instrument/accelerate/acceleration_utils.cpp
```diff
@@ -227,7 +227,7 @@ void acceleration_utilst::stash_variables(
       it!=vars.end();
       ++it)
   {
-    symbolt orig=*symbol_table.lookup(*it);
+    symbolt orig = symbol_table.lookup_ref(*it);
```
tautschnig: `const symbolt &orig`

src/goto-instrument/accelerate/polynomial_accelerator.cpp
```diff
@@ -768,7 +768,7 @@ void polynomial_acceleratort::stash_variables(
       it!=vars.end();
       ++it)
   {
-    symbolt orig=*symbol_table.lookup(*it);
+    symbolt orig = symbol_table.lookup_ref(*it);
```
tautschnig: `const symbolt &orig`

tautschnig (approved):
Thank you!!!

peterschrammel (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: b79a10069bddd467220888aecedb2f86e5d1cf16).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105870484

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 78c1c317ab167ab1526b34fb84b6540ae6d202c5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105877346

PR #4441, opened 2019-03-26
Remove unnecessary helper function that adds a function call to a code block

antlechner:
No functional change, only refactoring.
~The helper function is not directly related to nondet object creation and can be used in another function outside of the object factory.~ The function has now been removed entirely as it can be simplified and inlined.

### DIFF
--- a/jbmc/src/java_bytecode/java_object_factory.cpp
+++ b/jbmc/src/java_bytecode/java_object_factory.cpp
@@ -156,11 +156,6 @@ private:
     size_t depth,
     const source_locationt &location);
 
-  void gen_method_call_if_present(
-    code_blockt &assignments,
-    const exprt &instance_expr,
-    const irep_idt &method_name);
-
   void array_primitive_init_code(
     code_blockt &assignments,
     const exprt &init_array_expr,
@@ -589,7 +584,8 @@ void java_object_factoryt::gen_nondet_pointer_init(
     {
       const irep_idt &class_name = class_type->get_name();
       const irep_idt class_clinit = clinit_wrapper_name(class_name);
-      gen_method_call_if_present(assignments, expr, class_clinit);
+      if(const auto clinit_func = symbol_table.lookup(class_clinit))
+        assignments.add(code_function_callt{clinit_func->symbol_expr()});
     }
   }
 
@@ -953,7 +949,9 @@ void java_object_factoryt::gen_nondet_struct_init(
 
   const irep_idt init_method_name =
     "java::" + id2string(struct_tag) + ".cproverNondetInitialize:()V";
-  gen_method_call_if_present(assignments, expr, init_method_name);
+  if(const auto init_func = symbol_table.lookup(init_method_name))
+    assignments.add(
+      code_function_callt{init_func->symbol_expr(), {address_of_exprt{expr}}});
 }
 
 /// Initializes a primitive-typed or reference-typed object tree rooted at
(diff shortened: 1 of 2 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 54d5bf448bdf0223a9e268e9c629f6f1386cd646).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105908403

jbmc/src/java_bytecode/java_utils.cpp
```diff
@@ -262,6 +262,15 @@ void add_method_call_if_present(
   }
 }
 
+void add_static_method_call_if_present(
+  code_blockt &assignments,
+  const irep_idt &method_name,
+  const symbol_table_baset &symbol_table)
+{
+  add_method_call_if_present(
+    assignments, nil_exprt{}, method_name, symbol_table);
```
owen-mc-diffblue: :pick: I think it would be slightly nicer to use an optional `exprt` for the second argument, so we don't have to pass a `nil_exprt` like this. There could be a precondition that if the second argument is non-empty then the type has a `this` parameter.

owen-mc-diffblue (approved):

jbmc/src/java_bytecode/java_utils.cpp
```diff
@@ -246,6 +246,31 @@ void java_add_components_to_class(
     components.end(), components_to_add.begin(), components_to_add.end());
 }
 
+void add_method_call_if_present(
```
smowton: It's a bit weird to have a general-purpose utility function that only works on no-arg functions. I think you should take a vector-of-exprt and append those if given.

jbmc/src/java_bytecode/java_utils.cpp
antlechner: Passing in `null` as an instance to call a method on if the method is static is quite common in Java (e.g. `java.lang.reflect`) but I agree that an optionalt is probably clearer in the context of JBMC. So I changed it to that. I added an invariant that's the converse of your suggestion: if the type has a `this` parameter (i.e. it represents an instance method), the second argument has to be non-empty. It still seems safe to just ignore the second argument if the represented Java method is static. But if you think it's clearer I could change it to something like
```
    if(instance_expr)
    {
      INVARIANT(
        type.has_this(),
        "A call to a code_function_callt representing a static Java method "
        "should not be given an instance argument.");
      fun_call.arguments().push_back(address_of_exprt(*instance_expr));
    }
    else
    {
      INVARIANT(
        !type.has_this(),
        "A call to a code_function_callt representing a Java instance method "
        "should be given an instance argument.");
    }
```
Let me know what you think.

antlechner: Or I could just get rid of `instance_expr` altogether and make it part of the vector that @smowton suggested below? I'll try that.

antlechner:
@owen-jones-diffblue @smowton As I tried to implement @smowton 's suggestion to add a vector for the arguments, and to make the `instance_exprt` the first element of that vector, the helper function actually became so simple that it made more sense to just inline it. So I did that and the PR is completely different now - please re-review. :slightly_smiling_face:

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 25abc15ee3530b836726b5df9a129298943fa718).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/105996838

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 1c638651c1b2e4752c16e71ee4dfdcc4a5bbb9ed).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106006577

owen-mc-diffblue (approved):

smowton (approved):

PR #3619, opened 2018-12-21
Constant propagation of nondet-symbol [blocks: #35]

tautschnig:
- [x] Each commit message has a non-empty body, explaining why the change was made.
- n/a Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4444, opened 2019-03-27
Symex: resolve pointer comparisons using the value-set

owen-mc-diffblue:
This allows us to evaluate boolean expressions (including GOTO conditions) that test pointer equality or inequality (particularly "ptr == null" or "!= null") where the value-set contains sufficient information. We can already handle simple cases where `ptr` has a single known alias via constant propagation and aliasing, but not cases where `ptr` is non-constant, for example because it is drawn from an array.

### DIFF
--- a/regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.c
+++ b/regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.c
@@ -1,0 +1,181 @@
+#include <assert.h>
+
+static void noop()
+{
+}
+
+int test(int nondet_int_array[])
+{
+  int a = 1, b = 2, c = 3;
+  int *ptr_to_a = &a, *ptr_to_b = &b, *ptr_to_c = &c, *ptr_to_null = 0;
+
+  // Symex knows the value of ptr_to_a, ptr_to_b, ptr_to_c and ptr_to_null, so
+  // it should be able to evaluate simple conditions involving them.
+
+  // Equality "=="
+
+  // A non-null pointer and a matching value
+  int unconditionally_reachable_1;
+  if(ptr_to_a == &a)
+    unconditionally_reachable_1 = 7;
+
+  // A non-null pointer and a non-matching value (not a null pointer)
+  int unreachable_1;
+  if(ptr_to_a == &c)
+    unreachable_1 = 7;
+
+  // A non-null pointer and a non-matching value (a null pointer)
+  int unreachable_2;
+  if(ptr_to_a == 0)
+    unreachable_2 = 7;
+
+  // A null pointer and a matching value
+  int unconditionally_reachable_2;
+  if(ptr_to_null == 0)
+    unconditionally_reachable_2 = 7;
+
+  // A null pointer and a non-matching value
+  int unreachable_3;
+  if(ptr_to_null == &a)
+    unreachable_3 = 7;
+
+  // Disequality "!="
+
+  // A non-null pointer and a matching value
+  int unreachable_4;
+  if(ptr_to_a != &a)
+    unreachable_4 = 7;
+
+  // A non-null pointer and a non-matching value (not a null pointer)
+  int unconditionally_reachable_3;
+  if(ptr_to_a != &c)
+    unconditionally_reachable_3 = 7;
+
+  // A non-null pointer and a non-matching value (a null pointer)
+  int unconditionally_reachable_4;
+  if(ptr_to_a != 0)
+    unconditionally_reachable_4 = 7;
+
+  // A null pointer and a matching value
+  int unreachable_5;
+  if(ptr_to_null != 0)
+    unreachable_5 = 7;
+
+  // A null pointer and a non-matching value
+  int unconditionally_reachable_5;
+  if(ptr_to_null != &a)
+    unconditionally_reachable_5 = 7;
+
+  // Symex can't tell what ptr_to_a_or_b points to, but we can tell that it
+  // doesn't point to some things
+  int *ptr_to_a_or_b = nondet_int_array[0] == 0 ? &a : &b;
+  int *ptr_to_a_or_b_or_null =
+    nondet_int_array[1] == 0 ? &a : nondet_int_array[1] == 1 ? &b : 0;
+
+  // Equality "=="
+
+  // A non-null pointer and a matching value
+  int possibly_reachable_1;
+  if(ptr_to_a_or_b == &a)
+    possibly_reachable_1 = 7;
+
+  // A non-null pointer and a non-matching value (not a null pointer)
+  int unreachable_6;
+  if(ptr_to_a_or_b == &c)
+    unreachable_6 = 7;
+
+  // A non-null pointer and a non-matching value (a null pointer)
+  int unreachable_7;
+  if(ptr_to_a_or_b == 0)
+    unreachable_7 = 7;
+
+  // A possibly-null pointer and a matching value (not a null pointer)
+  int possibly_reachable_2;
+  if(ptr_to_a_or_b_or_null == &a)
+    possibly_reachable_2 = 7;
+
+  // A possibly-null pointer and a matching value (a null pointer)
+  int possibly_reachable_3;
+  if(ptr_to_a_or_b_or_null == 0)
+    possibly_reachable_3 = 7;
+
+  // A possibly-null pointer and a non-matching value
+  int unreachable_8;
+  if(ptr_to_a_or_b_or_null == &c)
+    unreachable_8 = 7;
+
+  // Disequality "!="
+
+  // A non-null pointer and a matching value
+  int possibly_reachable_4;
+  if(ptr_to_a_or_b != &a)
+    possibly_reachable_4 = 7;
+
+  // A non-null pointer and a non-matching value (not a null pointer)
+  int unconditionally_reachable_6;
+  if(ptr_to_a_or_b != &c)
+    unconditionally_reachable_6 = 7;
+
+  // A non-null pointer and a non-matching value (a null pointer)
+  int unconditionally_reachable_7;
+  if(ptr_to_a_or_b != 0)
+    unconditionally_reachable_7 = 7;
+
+  // A possibly-null pointer and a matching value (not a null pointer)
+  int possibly_reachable_5;
+  if(ptr_to_a_or_b_or_null != &a)
+    possibly_reachable_5 = 7;
+
+  // A possibly-null pointer and a matching value (a null pointer)
+  int possibly_reachable_6;
+  if(ptr_to_a_or_b_or_null != 0)
+    possibly_reachable_6 = 7;
+
+  // A possibly-null pointer and a non-matching value
+  int unconditionally_reachable_8;
+  if(ptr_to_a_or_b_or_null != &c)
+    unconditionally_reachable_8 = 7;
+
+  // We should also be able to do all of the above in compound expressions which
+  // use logical operators like AND, OR and NOT, or even ternary expressions.
+
+  int unconditionally_reachable_9;
+  if(!(ptr_to_a == &c) && ptr_to_a_or_b != 0)
+    unconditionally_reachable_9 = 7;
+
+  int unreachable_9;
+  if(!(ptr_to_null == 0) || ptr_to_a_or_b == 0)
+    unreachable_9 = 7;
+
+  int unconditionally_reachable_10;
+  if((ptr_to_a == &a && !(ptr_to_a_or_b == 0)) || ptr_to_a_or_b_or_null == &c)
+    unconditionally_reachable_10 = 7;
+
+  int unreachable_10;
+  if(ptr_to_a_or_b_or_null != 0 ? ptr_to_null != 0 : ptr_to_a_or_b == &c)
+    unreachable_10 = 7;
+
+  // And everything should work with the symbol on the left or the right
+
+  int unconditionally_reachable_11;
+  if(!(&c == ptr_to_a) && 0 != ptr_to_a_or_b)
+    unconditionally_reachable_11 = 7;
+
+  int unreachable_11;
+  if(!(0 == ptr_to_null) || 0 == ptr_to_a_or_b)
+    unreachable_11 = 7;
+
+  int unconditionally_reachable_12;
+  if((&a == ptr_to_a && !(0 == ptr_to_a_or_b)) || &c == ptr_to_a_or_b_or_null)
+    unconditionally_reachable_12 = 7;
+
+  int unreachable_12;
+  if(0 != ptr_to_a_or_b_or_null ? 0 != ptr_to_null : &c == ptr_to_a_or_b)
+    unreachable_12 = 7;
+
+  int possibly_reachable_7;
+  if(0 != ptr_to_a_or_b_or_null)
+    possibly_reachable_7 = 7;
+
+  assert(0);
+}
--- a/regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.desc
+++ b/regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.desc
@@ -1,0 +1,60 @@
+CORE paths-lifo-expected-failure
+test.c
+--function test --show-vcc
+^EXIT=0$
+^SIGNAL=0$
+test::1::unconditionally_reachable_1[^\s]+ = 7$
+test::1::unconditionally_reachable_2[^\s]+ = 7$
+test::1::unconditionally_reachable_3[^\s]+ = 7$
+test::1::unconditionally_reachable_4[^\s]+ = 7$
+test::1::unconditionally_reachable_5[^\s]+ = 7$
+test::1::unconditionally_reachable_6[^\s]+ = 7$
+test::1::unconditionally_reachable_7[^\s]+ = 7$
+test::1::unconditionally_reachable_8[^\s]+ = 7$
+test::1::unconditionally_reachable_9[^\s]+ = 7$
+test::1::unconditionally_reachable_10[^\s]+ = 7$
+test::1::possibly_reachable_1[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::possibly_reachable_1[^\s]+\)$
+test::1::possibly_reachable_2[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::possibly_reachable_2[^\s]+\)$
+test::1::possibly_reachable_3[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::possibly_reachable_3[^\s]+\)$
+test::1::possibly_reachable_4[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::possibly_reachable_4[^\s]+\)$
+test::1::possibly_reachable_5[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::possibly_reachable_5[^\s]+\)$
+test::1::possibly_reachable_6[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::possibly_reachable_6[^\s]+\)$
+test::1::possibly_reachable_7[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::possibly_reachable_7[^\s]+\)$
+--
+test::1::unconditionally_reachable_1[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_1[^\s]+\)$
+test::1::unconditionally_reachable_2[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_2[^\s]+\)$
+test::1::unconditionally_reachable_3[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_3[^\s]+\)$
+test::1::unconditionally_reachable_4[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_4[^\s]+\)$
+test::1::unconditionally_reachable_5[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_5[^\s]+\)$
+test::1::unconditionally_reachable_6[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_6[^\s]+\)$
+test::1::unconditionally_reachable_7[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_7[^\s]+\)$
+test::1::unconditionally_reachable_8[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_8[^\s]+\)$
+test::1::unconditionally_reachable_9[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_9[^\s]+\)$
+test::1::unconditionally_reachable_10[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_10[^\s]+\)$
+test::1::unconditionally_reachable_11[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_11[^\s]+\)$
+test::1::unconditionally_reachable_12[^\s]+ = \(goto_symex::\\guard[^\s]+ \? 7 : test::1::unconditionally_reachable_12[^\s]+\)$
+test::1::unreachable_1[^\s]+ = 7$
+test::1::unreachable_2[^\s]+ = 7$
+test::1::unreachable_3[^\s]+ = 7$
+test::1::unreachable_4[^\s]+ = 7$
+test::1::unreachable_5[^\s]+ = 7$
+test::1::unreachable_6[^\s]+ = 7$
+test::1::unreachable_7[^\s]+ = 7$
+test::1::unreachable_8[^\s]+ = 7$
+test::1::unreachable_9[^\s]+ = 7$
+test::1::unreachable_10[^\s]+ = 7$
+test::1::unreachable_11[^\s]+ = 7$
+test::1::unreachable_12[^\s]+ = 7$
+^warning: ignoring
+unreachable_1[3-9]
+unreachable[2-9][0-9]
+unconditionally_reachable_1[3-9]
+unconditionally_reachable[2-9][0-9]
+possibly_reachable_[8-9]
+possibly_reachable_[1-9][0-9]
+--
+Pointer comparisons will be resolved in symex by a mixture of constant
+propagation and value-set filtering in try_evaluate_pointer_comparisons.
+
+The expected failure for path-symex is because the lines we check for
+possibly_reachable_* would only appear when there is convergence behaviour.
--- a/src/goto-symex/renaming_level.h
+++ b/src/goto-symex/renaming_level.h
@@ -15,7 +15,9 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 #include <map>
 #include <unordered_set>
 
+#include <util/expr_iterator.h>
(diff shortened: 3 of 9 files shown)


src/goto-symex/goto_symex.h
```diff
@@ -287,6 +287,10 @@ class goto_symext
     statet &,
     bool keep_array);
 
+  exprt try_evaluate_pointer_comparisons(
```
owen-mc-diffblue: Any suggestions for a better name?

src/goto-symex/goto_symex.h
```diff
@@ -333,6 +337,11 @@ class goto_symext
     const exprt &new_guard,
     const namespacet &ns);
 
+  optionalt<exprt> build_reference_to_value_set_element(
```
owen-mc-diffblue: This function might fit better somewhere else - perhaps in `value_set_dereferencet`, since it's mainly a wrapper around `value_set_dereferencet::build_reference_to`?

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,103 @@ void goto_symext::apply_goto_condition(
...
+  const value_sett &value_set)
+{
+  if(condition.id() == ID_and || condition.id() == ID_or)
+  {
+    for(auto &op : condition.operands())
+    {
+      op = try_evaluate_pointer_comparisons(std::move(op), value_set);
+    }
+  }
+  else if(condition.id() == ID_not)
```
owen-mc-diffblue: anything else I should be recursing over, other than AND, OR and NOT?

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,103 @@ void goto_symext::apply_goto_condition(
...
+      op = try_evaluate_pointer_comparisons(std::move(op), value_set);
+    }
+  }
+  else if(condition.id() == ID_equal || condition.id() == ID_notequal)
+  {
+    if(can_cast_type<pointer_typet>(condition.op0().type()))
+    {
+      symbol_exprt *symbol_expr =
+        expr_try_dynamic_cast<symbol_exprt>(condition.op0());
+      exprt &constant = condition.op1();
```
owen-mc-diffblue: Currently it assumes the symbol is on the lhs and the constant is on the rhs. I will fix that up.

src/goto-symex/symex_main.cpp
```diff
@@ -583,6 +583,38 @@ find_unique_pointer_typed_symbol(const exprt &expr)
   return return_value;
 }
 
+optionalt<exprt> goto_symext::build_reference_to_value_set_element(
```
owen-mc-diffblue: This code was in `goto_symext::try_filter_value_sets` and I've pulled it out into a function to reuse it. Perhaps I'll make that a separate commit.

regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.c
```diff
@@ -0,0 +1,76 @@
...
+  int unconditionally_reachable_11;
+  if(!(ptr_to_a_or_b == &c))
+    unconditionally_reachable_11 = 7;
+
+  int possibly_reachable_12;
+  if(!(ptr_to_a_or_b == &a))
+    possibly_reachable_12 = 7;
+
+  int unconditionally_reachable_13;
+  if((ptr_to_a == &a && c == 3) || c == 0)
```
smowton: Bear in mind these constant comparisons with `c` may well be resolved very early as `c` is only assigned once and never address-taken.

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,103 @@ void goto_symext::apply_goto_condition(
...
+
+      if(symbol_expr && goto_symex_is_constantt()(constant))
```
smowton: Handle the reverse case too (constant OP symbol)

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,103 @@ void goto_symext::apply_goto_condition(
...
+      {
+        value_setst::valuest value_set_elements;
+        value_set.get_value_set(
```
smowton: Assert that `symbol_expr` is already L1 at least

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,103 @@ void goto_symext::apply_goto_condition(
...
+          to_ssa_expr(*symbol_expr).get_l1_object(), value_set_elements, ns);
+
+        INVARIANT(
+          constant.id() == ID_address_of || constant.id() == ID_NULL,
```
smowton: Surprising! Null is usually represented as an `ID_constant` with value = NULL -- suspect you need to add tests for this

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,103 @@ void goto_symext::apply_goto_condition(
...
+              constant_found = true;
+              // We can't break because we have to make sure we find any
+              // instances of ID_unknown or ID_invalid
+            }
+          }
+        }
+
+        if(!unknown_or_invalid_found)
+        {
+          if(!constant_found)
```
smowton: Give some brief comments setting out the logic here, as `!constant_found` and `value_set_elements.size() == 1` don't immediately chime with "x in { a, b, c }" and "x in { x }" respectively.

smowton (commented):
Looks good except I believe the null case is broken, and at least needs tests.

src/goto-symex/symex_goto.cpp
smowton: (and test this)

src/goto-symex/goto_symex.h
smowton: I like this one!

src/goto-symex/goto_symex.h
smowton: Proooobably. No strong opinion.

src/goto-symex/symex_goto.cpp
smowton: No I think that's enough

smowton: Oh wait, how about the arguments of a boolean-typed `if_exprt`

smowton: (e.g. `x == y ? z == w : a == b`)

smowton: Or indeed the first arg of an `if_exprt` in any context. Total recursion looking for `==` and `!=` nodes might be the simplest thing.

smowton: I 100% think these comments through before beginning to write FYI

regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.c
owen-mc-diffblue: I've changed them to other pointer comparisons. And I now take the address of all the integers at the beginning, if that's a thing that symex cares about.

owen-mc-diffblue:
I've addressed all of @smowton's review comments on the draft PR and got CI working (touch wood).

regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.c
```diff
@@ -0,0 +1,188 @@
...
+int test(int nondet_int) {
+  int a = 1, b = 2, c = 3;
+  int *ptr_to_a = &a, *ptr_to_b = &b, *ptr_to_c = &c, *ptr_to_null = 0;
+
+  // Symex knows the value of ptr_to_a, ptr_to_b, ptr_to_c and ptr_to_null, so
+  // it should be able to evaluate simple conditions involving them.
+
+  // Equality "=="
+
+  // A non-null pointer and the correct value
```
smowton: correct -> matching and incorrect -> non-matching everywhere

regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.c
```diff
@@ -0,0 +1,188 @@
...
+
+  // A possibly-null pointer and an incorrect value
+  int unreachable_8;
+  if(ptr_to_a_or_b_or_null == &c)
+    unreachable_8 = 7;
+
+
+  // Disequality "!="
+
+  // A non-null pointer and the correct value
```
smowton: the -> a

regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.desc
```diff
@@ -0,0 +1,51 @@
...
+test::1::unreachable_6[^\s]+ = 7$
+test::1::unreachable_7[^\s]+ = 7$
+test::1::unreachable_8[^\s]+ = 7$
+test::1::unreachable_9[^\s]+ = 7$
+test::1::unreachable_10[^\s]+ = 7$
+test::1::unreachable_11[^\s]+ = 7$
+test::1::unreachable_12[^\s]+ = 7$
+^warning: ignoring
+--
+Pointer comparisons will be resolved in symex by a mixture of constant
```
smowton, line 56: Suggest adding conditions that check you check for everything you should, by making sure `unreachable_1[3-9])` and `unreachable[2-9][0-9]` for example don't occur anywhere (i.e., if someone added a test but not a cross-check this regex would complain)

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,182 @@ void goto_symext::apply_goto_condition(
...
+  const namespacet &ns)
+{
+  const typecast_exprt *typecast_expr =
+    expr_try_dynamic_cast<typecast_exprt>(other_operand);
+  const constant_exprt *constant_expr =
+    expr_try_dynamic_cast<constant_exprt>(other_operand);
+
+  INVARIANT(
+    other_operand.id() == ID_address_of ||
+      (typecast_expr && typecast_expr->op().id() == ID_address_of) ||
```
smowton: If you're not sure this is simplified you might get redundant stacked typecast ops

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,182 @@ void goto_symext::apply_goto_condition(
...
+  const exprt &other_operand,
+  const value_sett &value_set,
+  const irep_idt language_mode,
+  const namespacet &ns)
+{
+  const symbol_exprt *symbol_expr_lhs =
+    expr_try_dynamic_cast<symbol_exprt>(symbol_expr);
+
+  if(!symbol_expr_lhs)
+  {
```
smowton: No need for braces with one-liner `if`s

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,182 @@ void goto_symext::apply_goto_condition(
...
+///   the lhs and rhs swapped)
+/// \param symbol_expr: The symbol expression in the condition
+/// \param other_operand: The other expression in the condition - must pass
+///   goto_symex_is_constant, and since it is pointer-typed it must therefore
+///   be an address of expression, a typecast of an address of expression or a
+///   null pointer
+/// \param value_set: The value-set for looking up what the symbol can point to
+/// \param language_mode: The language mode
+/// \param ns: A namespace
+static bool try_evaluate_pointer_comparisons_base_case_with_check(
```
smowton: Instead of calling this twice, suggest something like https://github.com/diffblue/cbmc/blob/develop/src/goto-symex/goto_state.cpp#L67

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,182 @@ void goto_symext::apply_goto_condition(
...
+///   the lhs and rhs swapped)
+/// \param symbol_expr: The symbol expression in the condition
+/// \param other_operand: The other expression in the condition - must pass
+///   goto_symex_is_constant, and since it is pointer-typed it must therefore
+///   be an address of expression, a typecast of an address of expression or a
+///   null pointer
+/// \param value_set: The value-set for looking up what the symbol can point to
+/// \param language_mode: The language mode
+/// \param ns: A namespace
+static void try_evaluate_pointer_comparisons_base_case(
```
smowton: How about `try_evaluate_pointer_comparison`, singular?

src/goto-symex/symex_goto.cpp
smowton: I don't think there's much gained by the two functions (with_check vs. without) since they're both only called in one place

owen-mc-diffblue:
Note also that the last two commits break sharing everywhere, so we'll have to rethink how to do that.

smowton:
Suggested change: https://github.com/smowton/cbmc/commit/1daa2c60056213d2eddce6f9e3a2945a139d81b2

That removes `renamed_vectort` and instead uses `depth_begin/depth_end`, which avoid sharing breaks when no changes are made, to selectively alter the renamed expression, only permitting changes where the renaming level is consistent. @romainbrenguier are you happy with this?

@owen-jones-diffblue note that also applies a couple of my suggestions above since it made it easier to code :)

src/goto-symex/symex_goto.cpp
owen-mc-diffblue: I prefer to use them in all cases

romainbrenguier:
@smowton 
> Suggested change: smowton@1daa2c6
> That removes renamed_vectort and instead uses depth_begin/depth_end, which avoid sharing breaks when no changes are made, to selectively alter the renamed expression, only permitting changes where the renaming level is consistent. @romainbrenguier are you happy with this?

Yes that looks better. What I'm not sure about is the fact that the mutator function takes an `exprt` as argument rather that a `renamedt`. That way you avoid calling constructors on all intermediary expressions, but I don't think it's nice for the interface. Maybe we could make `renamedt` privately inherit from exprt so that we can use `static_cast` instead of a constructor, so that conversion would be cheap and the interface nicer?
Here is the change I propose (if you want to use it): https://github.com/diffblue/cbmc/pull/4454

owen-mc-diffblue:
Thanks @smowton, that's much better. I've cherry-picked it and squashed it in, along with fixes for all the things you mentioned. I did add an extra commit which makes `selectively_mutate` a free function (which is a friend of `renamedt`), since it doesn't make sense to be using depth iterators for anything when `underlyingt` is derived from `exprt`. I'm not sure how this interacts with @romainbrenguier's suggestion.

There are three failing tests. The first is `exception_cleanup`, which is failing because symex can now tell that `@inflight_exception` is not null, so the test needs to be changed. I'm not totally sure what it should now be. The failing lines are:
```
1 remaining after simplification$
```
in test.desc and
```
file Test\.java line 6 function java::Test.main:\(I\)V
no uncaught exception
```
in vccs.desc.

The other two are nondet_elements_longer_lists and nondet_elements_longer_lists_global, which are failing in the same way. I'm looking into that at the moment.

owen-mc-diffblue:
For nondet_elements_longer_lists and nondet_elements_longer_lists_global, the test is failing to find these two lines: ```
\[test_function.unwind.\d+\] line \d+ unwinding assertion loop 0: SUCCESS
\[test_function.unwind.\d+\] line \d+ unwinding assertion loop 1: SUCCESS
```
Is there a good reason that we would not generate these unwinding assertions? If symex can determine that they are true then does it just not output them in the formula?

**Comment by smowton** [comment]
@owen-jones-diffblue unsure about the unwinding assertions (yet) but the exception one is entirely desirable -- that 1 VCC remaining was a shortcoming which we have now fixed (symex can tell no exception escaped from the whole program). Just amend the test to check *all* conditions are symex'd away.

**Comment by smowton** [comment]
@romainbrenguier I'm not sure I follow, what would your ideal interface look like?

**Comment by tautschnig** [comment]
@owen-jones-diffblue Presumably your changes just result in more constant propagation, and thus those unwinding assertions are resolved by the simplifier already. #2574 does the same, i.e., removes those lines from the test spec.

**Comment by romainbrenguier** [comment]
@smowton 
> @romainbrenguier I'm not sure I follow, what would your ideal interface look like?

I would think having `mutator_functiont` take `renamedt` arguments would be more useful (though in your case it doesn't really matter). For instance imagine we want to selectively apply `simplify` to operands of a `renamedt`, you could do it with a `renamedt -> optinalt<renamedt>` function. I'm not insisting on it, since it is not necessary here, but could be nice in the future.

**Comment by smowton** [comment]
Ah sure, ok. On the one hand the tag is useful if you want to call a function that demands a `renamedt`, but I certainly like not having to make a dedicated object for every subexpression, so your privately-inherit trick looks like a good solution.

**Comment by owen-mc-diffblue** [comment]
I've marked this as `do not merge` until we have some benchmarking data for it

**Comment by owen-mc-diffblue** [comment]
I've cherry-picked ##4454 and added a commit to use it. Before merging this should be cleaned up.

**Review by allredj — COMMENTED** [review]
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: eb9b8cecade7195d17aa7de83276962168a60daa).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106349820

File: src/util/ssa_expr.h [hunk 18]
```diff
@@ -171,4 +171,14 @@ inline bool is_ssa_expr(const exprt &expr)
...
+template <>
+inline bool can_cast_expr<ssa_exprt>(const exprt &base)
+{
+  return base.id() == ID_symbol && base.get_bool(ID_C_SSA_symbol);
+}
+
+inline void validate_expr(const ssa_exprt &value)
+{
+  validate_operands(value, 0, "SSA expressions must not have operands");
+}
```
tautschnig: If adding those, could we please unify the various checking-functions in the same way this has been implemented in `std_expr.h` for all the classes? That is, use `validate_expr` in the above.

src/pointer-analysis/value_set_dereference.h
```diff
@@ -74,6 +74,13 @@ class value_set_dereferencet final
     irep_idt language_mode,
     const namespacet &ns);
 
+  static optionalt<exprt> build_reference_to_value_set_element(
```
tautschnig: Please document.

regression/cbmc/symex_should_evaluate_simple_pointer_conditions/test.desc
```diff
@@ -0,0 +1,60 @@
...
+unconditionally_reachable_1[3-9]
+unconditionally_reachable[2-9][0-9]
+possibly_reachable_[8-9]
+possibly_reachable_[1-9][0-9]
+--
+Pointer comparisons will be resolved in symex by a mixture of constant
+propagation and value-set filtering in try_evaluate_pointer_comparisons.
+
+The expected failure for path-symex is because the lines we check for
+possibly_reachable_* would only appear when there is convergence behaviour.
```
tautschnig: GitHub suggests adding a final newline.

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,167 @@ void goto_symext::apply_goto_condition(
...
+  const exprt *other_operand_with_typecasts_removed = &other_operand;
+  const typecast_exprt *typecast_expr = expr_try_dynamic_cast<typecast_exprt>(
+    *other_operand_with_typecasts_removed);
+
+  while(typecast_expr)
+  {
+    other_operand_with_typecasts_removed = &typecast_expr->op();
+    typecast_expr = expr_try_dynamic_cast<typecast_exprt>(
+      *other_operand_with_typecasts_removed);
+  }
```
tautschnig: Use `skip_typecast`?

src/pointer-analysis/value_set_dereference.h
owen-mc-diffblue: What I want to call it is `build_reference_to`, but that's already taken by the function it's wrapping (which doesn't actually build a reference).

The reason I have a wrapper function is that `build_reference_to` makes it easy to get objects, but not pointers, e.g. it gives you `global`, not `&global`, and for null pointers it uses the special value `nil_exprt`, and doesn't give me the `null_pointer_exprt` that I want. That makes sense because I think it's only used for `dereference_rec`, which only cares about dereferenced pointers.  Now that I've read through that function a few more times, I think a better solution would be to add a field to `value_set_dereferencet::valuet` to put `&global` and `null_pointer_exprt` into. What would you say to that?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 926531eae1c1eeb6c03c9deab5570c422639484e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106591286

src/goto-symex/symex_goto.cpp
owen-mc-diffblue: I tried to search for such a utility but didn't hit the right combination of words. Next time I'll try looking in `expr_utils.h`.

owen-mc-diffblue:
@tautschnig I've addressed all but one comment. Let me know what you think about my suggested approach to addressing the last comment.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 44d45d645e96fc5787b21c97e20ca4e5e4f400a9).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106623135

src/pointer-analysis/value_set_dereference.h
owen-mc-diffblue: I have done this now

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c0e7ab8c6151a9943d901e145acffc7a581c10a8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106938073

smowton:
@tautschnig initial findings from studying this: the test `jbmc/regression/jbmc/exception-cleanup` gives a good example of where this can help, as even with #3619 and #4430 we get an expression where propagation works, but the simplifier isn't clever enough to spot that a unique value can be determined -- in this case we get `guard ? null : (guard ? &x : null)`, which is clearly always null but requires the simplifier to propagate that `guard => false` into the RHS of the conditional, which it doesn't attempt at present.

Here's another example where propagation alone doesn't work:
```c
#include <assert.h>

  int x, y;
  int *p = argc == 1 ? 0 : argc == 2 ? &x : &y;
  if(p != 0) {
    if(p == 0) {
      assert(0);
    }
  }
}
```

The redundant null-tests on `p` can't be resolved because by simplification because the knowledge that the null case has already been filtered out is encoded in the value-set but not in the expression itself. This could be done, for example by transforming `p == (argc == 1 ? 0 : argc == 2 ? &x : &y) && p != 0` into `argc == 2 ? &x : &y`, but we currently don't attempt this, and if we did I expect we would either (a) need to switch on local-definitions in the simplifier, which could get quite expensive, or (b) the augmented simplifier would probably only catch particular cases rather than generally propagating conditions into expressions.

In brief I continue to think that at least for the time being the value-set is a more robust way of tracking aliases than trying to propagate all if-expressions encoding aliases and improve the simplifier sufficiently to filter those nested conditionals and solve tests against them. I think this is particularly a problem as in the limiting case this results in spiralling expression sizes and so the cost of `simplify` and `merge_ireps` rises, as I'm currently trying to fight *against* in #4576.

src/goto-symex/renaming_level.h
```diff
@@ -98,12 +103,51 @@ class renamedt : private underlyingt
...
 
+  template <levelt make_renamed_level>
+  friend renamedt<exprt, make_renamed_level>
+  make_renamed(constant_exprt constant);
+
+  template <levelt selectively_mutate_level>
+  friend void selectively_mutate(
+    renamedt<exprt, selectively_mutate_level> &renamed,
+    typename renamedt<exprt, selectively_mutate_level>::mutator_functiont
+      get_mutated_expr);
```
tautschnig, line 114: I'm a bit surprised by the amount of friendship in here. Couldn't some of these be made static members instead?

tautschnig (approved):
Looks ok me, except the remaining clang-format problem and, more importantly, the absence of performance data - could this please be provided? I certainly see that this gives symex additional power in resolving some comparisons, but at what cost does this come?

smowton:
cbmc CORE unit tests times: unchanged

smowton:
Large Webgoat example: 55 seconds to reach a particular checkpoint (down 5 seconds)

smowton:
Romain's Sakai benchmark: One benchmark reduced around 15 seconds (from 145 seconds), the remainder tested showed negligible difference.

smowton:
Therefore I'm pretty confident this has no noticeable cost in context of the other fairly-expensive ops we perform in symex.

src/goto-symex/renaming_level.h
smowton: The gymnastics here are a little beyond me, so I think I'll leave this as-is and welcome any cleanup PRs by people who're better at this game than me :)

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,161 @@ void goto_symext::apply_goto_condition(
...
+/// \param language_mode: The language mode
+/// \param ns: A namespace
+/// \return If we were able to evaluate the condition as true or false then we
+///   return that, otherwise we return an empty optionalt
+static optionalt<renamedt<exprt, L2>> try_evaluate_pointer_comparison(
+  irep_idt operation,
+  const symbol_exprt &symbol_expr,
+  const exprt &other_operand,
+  const value_sett &value_set,
+  const irep_idt language_mode,
```
peterschrammel, line 87: `const &`

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,161 @@ void goto_symext::apply_goto_condition(
...
+///   goto_symex_is_constant, and since it is pointer-typed it must therefore
+///   be an address of expression, a typecast of an address of expression or a
+///   null pointer
+/// \param value_set: The value-set for looking up what the symbol can point to
+/// \param language_mode: The language mode
+/// \param ns: A namespace
+/// \return If we were able to evaluate the condition as true or false then we
+///   return that, otherwise we return an empty optionalt
+static optionalt<renamedt<exprt, L2>> try_evaluate_pointer_comparison(
+  irep_idt operation,
```
peterschrammel: `const &`

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,161 @@ void goto_symext::apply_goto_condition(
...
+/// \param renamed_expr: The L2-renamed expression to check
+/// \param value_set: The value-set for looking up what the symbol can point to
+/// \param language_mode: The language mode
+/// \param ns: A namespace
+/// \return If we were able to evaluate the condition as true or false then we
+///   return that, otherwise we return an empty optionalt
+static optionalt<renamedt<exprt, L2>> try_evaluate_pointer_comparison(
+  const renamedt<exprt, L2> &renamed_expr,
+  const value_sett &value_set,
+  const irep_idt language_mode,
```
peterschrammel: `const &`

src/goto-symex/symex_goto.cpp
```diff
@@ -65,6 +67,161 @@ void goto_symext::apply_goto_condition(
...
+/// \param [in,out] condition: An L2-renamed expression with boolean type
+/// \param value_set: The value-set for determining what pointer-typed symbols
+///   might possibly point to
+/// \param language_mode: The language mode
+/// \param ns: A namespace
+/// \return The possibly modified condition
+static renamedt<exprt, L2> try_evaluate_pointer_comparisons(
+  renamedt<exprt, L2> condition,
+  const value_sett &value_set,
+  const irep_idt language_mode,
```
peterschrammel: `const &`

peterschrammel (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: ff6002bb92522460931c3ffc6c3d517e3ac8b0b1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111618837

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 213a1b25c0eeb10405005619b45ce71eb82490a1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111750931

tautschnig:
I'm afraid this actually causes spurious counterexamples being reported. I'll try to build a small regression test.

tautschnig:
Actually it's pretty simple:
```
int main()
{
  void *ptr;
  if(ptr == 0)
  {
    return 0;
  }
  __CPROVER_assert(ptr != 0, "null");
}
```
I believe the problem is that `ptr` has never been assigned, and thus the value set has no information about it.

smowton:
Discussion on Slack summary: will revert and re-submit with a fix

PR #4449, opened 2019-03-28
Factor out solver resource limits capability

peterschrammel:
That need not be a part of prop_convt.

### DIFF
--- a/src/goto-checker/solver_factory.cpp
+++ b/src/goto-checker/solver_factory.cpp
@@ -27,6 +27,7 @@ Author: Daniel Kroening, Peter Schrammel
 #include <solvers/flattening/bv_dimacs.h>
 #include <solvers/prop/prop.h>
 #include <solvers/prop/prop_conv.h>
+#include <solvers/prop/solver_resource_limits.h>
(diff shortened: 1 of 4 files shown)


tautschnig (approved):

martin-cs (approved):
I believe that this works as described.  However, while we are on the subject, *please* can we add a reproducible resource limit?  We should only be using time limits as an absolute last resort option, to the point I am almost tempted to say this interface should be deprecated!  (For people who haven't come across this problem before -- it is really important that CPROVER is deterministic, a whole bunch of things become really nasty if it is not.)  The SMT-LIB standard has had reproducible resource limits for many years and it would not be hard to implement in our back-end.  While we are working on this bit of code, please can we do something about this?

peterschrammel:
@martin-cs, happy to discuss this separately from this PR. I'll ping you.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 4896d6101573245a7fde703127c4dd00df07e5f4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106156861

PR #4054, opened 2019-02-03
Context-based solver interface for incremental solvers [depends: 4451, blocks: 4361]

peterschrammel:
Based on #4451 - only review last 4 commits.

Upstreams part of 2LS's push/pop incremental solver interface 

- [X] Each commit message has a non-empty body, explaining why the change was made.
- [X] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [X] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4450, opened 2019-03-28
Use decision_proceduret instead of prop_convt [depends: 4533, blocks: 4451]

peterschrammel:
~Based on #4449, only review last 2 commits.~

* `decision_proceduret` can now be used instead instead of `prop_convt` in most of the codebase.
* `decision_procedure_incrementalt` is only used where incremental solving is actually necessary.

In the next step `decision_procedure_incrementalt` will then be further split into support for assumptions (see #4451) and contexts (see #4054).

### DIFF
--- a/jbmc/src/jbmc/all_properties.cpp
+++ b/jbmc/src/jbmc/all_properties.cpp
@@ -20,7 +20,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/json.h>
 #include <util/xml.h>
 
-#include <solvers/prop/prop_conv.h>
+#include <solvers/prop/decision_procedure.h>
 #include <solvers/sat/satcheck.h>
 
 #include <goto-programs/json_goto_trace.h>
@@ -61,7 +61,7 @@ safety_checkert::resultt bmc_all_propertiest::operator()()
   auto solver_start = std::chrono::steady_clock::now();
 
   convert_symex_target_equation(
-    bmc.equation, bmc.prop_conv, get_message_handler());
+    bmc.equation, bmc.decision_procedure, get_message_handler());
   bmc.freeze_program_variables();
 
   // Collect _all_ goals in `goal_map'.
@@ -301,7 +301,8 @@ void bmc_all_propertiest::report(const cover_goalst &cover_goals)
 safety_checkert::resultt
 bmct::all_properties(const goto_functionst &goto_functions)
 {
-  bmc_all_propertiest bmc_all_properties(goto_functions, prop_conv, *this);
+  bmc_all_propertiest bmc_all_properties(
+    goto_functions, decision_procedure, *this);
   bmc_all_properties.set_message_handler(get_message_handler());
   return bmc_all_properties();
 }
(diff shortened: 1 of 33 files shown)


tautschnig:
> decision_proceduret can now be used instead instead of prop_convt in most of the codebase.
> decision_procedure_incrementalt is only used where incremental solving is actually necessary.

I haven't looked at the actual commits, but would certainly require @kroening to comment on this. To me, it seems a bit surprising. My assumption was that `prop_convt` is the part that takes care of translating some higher-level representation into a solver-level representation.

peterschrammel:
> takes care of translating some higher-level representation into a solver-level representation.

That might have been the initial intuition, but actually `smt2_convt` implements the same interface without any `prop` involved.

kroening:
This does not make sense to me -- the decision procedure interface was already designed to be incremental, and has already been used in this fashion.

kroening:
@tautschnig has the right intuition here. The key idea isn't the conversion, but the fact that prop_conv offers a 'handle' for predicates. This is also what smt2_conv offers. However, the 'handle' has now been pushed up to `decision_proceduret`, which turns `prop_convt` into an implementation as opposed to an interface.

peterschrammel:
> This does not make sense to me -- the decision procedure interface was already designed to be incremental, and has already been used in this fashion.

No, `decision_proceduret` could not / cannot be used incrementally with MiniSAT under the hood, for example, because the freezing functionality is missing.

kroening:
The way forward is to consider what's an interface, who consumes that, and what implementations of that interface should exist.

peterschrammel:
Also, look at #4451 to see how I'm going to move this forward.

kroening:
The problem that Minisat with preprocessed needs freezing is a special case; this shouldn't necessarily be imposed on all users of incremental solving. It's certainly by no means intuitive.

peterschrammel:
> The problem that Minisat with preprocessed needs freezing is a special case; this shouldn't necessarily be imposed on all users of incremental solving. It's certainly by no means intuitive.

Unfortunately, you cannot hide that and just ignore it.

tautschnig:
@kroening I do think that class names need to be changed - `prop_conv` are two, ambiguous abbreviations (property conversion? propositional conversion? something else?).

peterschrammel:
If we can get rid of `decision_procedure_incrementalt` at some point because no special treatment for MiniSAT&Co is required anymore, then even better - it will just drop out of the hierarchy.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 8927b7dc032e9015ccb1889530d3f7022a91aa9b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106227811

peterschrammel:
@kroening, please have a look at the proposed hierarchy of interfaces:
* [decision_proceduret](https://github.com/peterschrammel/cbmc/blob/incremental-solver-interface/src/solvers/prop/decision_procedure.h)
* [decision_procedure_incrementalt](https://github.com/peterschrammel/cbmc/blob/incremental-solver-interface/src/solvers/prop/decision_procedure_incremental.h)
* [decision_procedure_assumptionst](https://github.com/peterschrammel/cbmc/blob/incremental-solver-interface/src/solvers/prop/decision_procedure_assumptions.h)
* [decision_procedure_contextst](https://github.com/peterschrammel/cbmc/blob/incremental-solver-interface/src/solvers/prop/decision_procedure_contexts.h)

peterschrammel:
Obsolete.

PR #4451, opened 2019-03-28
Separate interface for solving under assumptions [depends: 4450, blocks: 4054]

peterschrammel:
Based on #4450, only review last commit.

### DIFF
--- a/src/solvers/prop/decision_procedure_assumptions.h
+++ b/src/solvers/prop/decision_procedure_assumptions.h
@@ -20,9 +20,6 @@ public:
   /// Set assumptions for the next call to operator() to solve the problem
   virtual void set_assumptions(const bvt &) = 0;
 
-  /// Returns true if an assumption is in the final conflict
-  virtual bool is_in_conflict(literalt l) const = 0;
-
   virtual ~decision_procedure_assumptionst() = default;
 };
 
--- a/src/solvers/prop/decision_procedure_assumptions.h
+++ b/src/solvers/prop/decision_procedure_assumptions.h
@@ -1,0 +1,29 @@
+/*******************************************************************\
+
+Module: Decision procedure with incremental solving under assumptions
+
+Author: Peter Schrammel
+
+\*******************************************************************/
+
+/// \file
+/// Decision procedure with incremental solving under assumptions
+
+#ifndef CPROVER_SOLVERS_PROP_DECISION_PROCEDURE_ASSUMPTIONS_H
+#define CPROVER_SOLVERS_PROP_DECISION_PROCEDURE_ASSUMPTIONS_H
+
+#include "decision_procedure_incremental.h"
+
+class decision_procedure_assumptionst : public decision_procedure_incrementalt
+{
+public:
+  /// Set assumptions for the next call to operator() to solve the problem
+  virtual void set_assumptions(const bvt &) = 0;
+
+  /// Returns true if an assumption is in the final conflict
+  virtual bool is_in_conflict(literalt l) const = 0;
+
+  virtual ~decision_procedure_assumptionst() = default;
+};
+
+#endif // CPROVER_SOLVERS_PROP_DECISION_PROCEDURE_ASSUMPTIONS_H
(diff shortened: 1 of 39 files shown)


src/solvers/prop/decision_procedure_assumptions.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+Author: Peter Schrammel
+
+\*******************************************************************/
+
+#include "decision_procedure_assumptions.h"
+
+void decision_procedure_assumptionst::set_assumptions(const bvt &)
+{
+  UNREACHABLE;
```
martin-cs: Could this be pure virtual or something else?  Base implementations of methods that are just "nope!" make me kinda nervous.

src/solvers/prop/decision_procedure_assumptions.h
```diff
@@ -0,0 +1,29 @@
...
+#ifndef CPROVER_SOLVERS_PROP_DECISION_PROCEDURE_ASSUMPTIONS_H
+#define CPROVER_SOLVERS_PROP_DECISION_PROCEDURE_ASSUMPTIONS_H
+
+#include "decision_procedure_incremental.h"
+
+class decision_procedure_assumptionst : public decision_procedure_incrementalt
+{
+public:
+  /// Set assumptions for the next call to operator() to solve the problem
+  virtual void set_assumptions(const bvt &);
```
martin-cs: Why is this / should this be a separate call?  If the solver is deterministic then it's not like you are going to call solve multiple times?  One reason it might be separate is that you want to be able to call this multiple times, but, if that is the case, do you need interfaces to manage and remove these?

martin-cs (commented):
Am unsure but I guess it makes sense to resolve the discussion around #4450 first.

src/solvers/prop/decision_procedure_assumptions.cpp
peterschrammel: True, I've actually thought I had patched that. Will change.

src/solvers/prop/decision_procedure_assumptions.h
peterschrammel: There is no particular reason except legacy. You can remove assumptions by passing an empty `bvt`.

peterschrammel:
@martin-cs, another question is whether `is_in_conflict` should be placed on top of `assumptions`. This would be justified as `smt2_convt` can emulate assumptions, but cannot do `is_in_conflict`.

src/solvers/prop/decision_procedure_assumptions.cpp
peterschrammel: Done

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: a17fb3696880178346e020a6dfa7104daff4e034).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106239885

src/solvers/prop/decision_procedure_assumptions.h
martin-cs: If we are going to refactor then really the assumptions should be an argument to the solve call and not stateful.

martin-cs: Better yet; it can be an argument and a result of the solve call so we can get back a reduced set of assumptions that are sufficient to give UNSAT.

martin-cs:
To my mind `is_in_conflict` is something for an incremental solver interface (i.e. the main interface) rather than solve under assumption.  If you are monotonically adding to a set of constraints then tracking (maybe triggering solve?) when a call to solve has returned UNSAT is a meaningful thing to do.  If you are working with assumptions then it is just making the interface more stateful than it needs to be.

peterschrammel:
@martin-cs, no. Minisat Solver.cc:401: `...express the final conflict in terms of assumptions. Calculates the (possibly empty) set of assumptions that led to the assignment of 'p', and stores the result in 'out_conflict'.`

martin-cs:
@peterschrammel : my point is that this could be wrapped by this API and made into a single call to `solve : set of assumptions -> SAT | set of contradictory assumptions`.  The actual solver is going to be very stateful but our interface to it can be a lot less stateful.

peterschrammel:
obsolete

PR #4455, opened 2019-03-29
Standardise interface for accessing the owner of a symbol

thomasspriggs:
Without the functions added by this PR, the only way to access the
owner of a symbol for static fields and java methods is through
`symbol.type.get(ID_C_class)`. When this technique is used by downstream
repositories, they are vulnerable to being broken by upstream changes to
how this information is stored in JBMC. The intention of this PR is
that the two new functions added will be the standard interface to this
symbol -> owner link. This means that future changes to how this
information is stored can be made without breaking downstream
repositories by changing these two functions.

This PR includes changes to use the two new functions. This is preparatory 
work ready for moving the link from being stored in a comment
on a symbols type, to the symbol itself, as discussed with @smowton 


peterschrammel (approved):

tautschnig (dismissed):
What is meant by the "owner of a symbol?" For a given symbol `x`, Is it the class that `x` is a member of? But then those aren't separate symbols, they would just be members of a struct type. So I'm confused.

Edit: the only owner of a symbol that we have is the symbol table.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -419,6 +419,8 @@ void java_bytecode_convert_methodt::convert(
     get_method_identifier(class_symbol.name, m);
 
   method_id=method_identifier;
+  set_owning_class(
```
thk123: :pick: Lower down (line 537) you have a writeable ref for this symbol already which is used to update the type. It might be nicer to keep the mutation in one place and do this call down there

jbmc/src/java_bytecode/java_utils.cpp
```diff
@@ -464,3 +464,14 @@ symbolt &fresh_java_symbol(
   return get_fresh_aux_symbol(
     type, name_prefix, basename_prefix, source_location, ID_java, symbol_table);
 }
+
+optionalt<irep_idt> owning_class(const symbolt &symbol)
+{
+  const irep_idt &class_id = symbol.type.get(ID_C_class);
+  return class_id.empty() ? optionalt<irep_idt>{} : class_id;
```
thk123, line 471: :pick: can `optionalt<irep_idt>{}` just be `{}`?

thk123 (commented):
Code changes look good, but I think to make this robust, it needs a unit test that actually goes through the java parser and checks the properties of a real static symbol. At a minimum java class with a static field and a static method.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
thomasspriggs: I tried putting it lower down and things didn't work properly. I guess it must be read back somewhere between the two locations.

jbmc/src/java_bytecode/java_utils.cpp
thomasspriggs: Nope, because of the usage of the ternary. I'd have to refactor this into an `if` instead in order to be able to use just `{}`.

smowton (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: bd2bacd2e13387adb0658ea0c9ae7d83b4311540).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106300980

thomasspriggs:
I have
- Renamed `owning_class` to `declaring_class`, as discussed with @tautschnig 
- Added unit tests covering the java parsing setting `declaring_class` as requested by @thk123

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 65e3d31a79aa9c439b7bad473509a93d48092783).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106341108

PR #3486, opened 2018-12-01
Patch set for SV-COMP'19 [depends-on: #2000, #3462]

tautschnig:
This is just to provide a to-do list of PRs to be created or existing PRs to be merged to put in place the fixes applied for CBMC's SV-COMP'19 submission. No review needed or expected.

- [x] #2001 (bugfix: graph-ml counterexample)
- [x] #1997 (performance: don't sort operands)
- [x] #1996 (bugfix: local dead objects)
- [x] #1991 (performance: no SSA sanity checks)
- [ ] #2000 (bugfix: invalid object)
- [x] #1992 (performance: hash code, size of irept)
- [x] #2220 (bugfix: phi nodes)
- [ ] #3462 (feature: memory leak checks)
- [x] #4269 (feature: alloca)
- [x] #4270 (feature: alloca-allocated object being free'd)
- [x] #4271 (bugfix: pthread_cond_wait semantics)
- [x] #4272 (feature: pthread_key support)
- [x] #4458 (performance: size of irept)

PR #4458, opened 2019-03-29
Use std::forward_list instead of std::map in irept by default [blocks: #3486]

tautschnig:
This reduces the size of an irept by 5 pointers (i.e., 40 bytes on
64-bit systems).

Also make typet constructor compatible with the STL shipped with GCC 5.

### DIFF
--- a/src/memory-analyzer/gdb_api.h
+++ b/src/memory-analyzer/gdb_api.h
@@ -22,6 +22,7 @@ Author: Malte Mues <mail.mues@gmail.com>
 #include <algorithm>
 #include <exception>
 #include <forward_list>
+#include <map>
 
 #include <util/exception_utils.h>
 
--- a/src/util/irep.h
+++ b/src/util/irep.h
@@ -20,9 +20,11 @@ Author: Daniel Kroening, kroening@kroening.com
 #ifndef HASH_CODE
 #  define HASH_CODE 1
 #endif
-// #define NAMED_SUB_IS_FORWARD_LIST
+#ifndef NAMED_SUB_IS_FORWARD_LIST
+#  define NAMED_SUB_IS_FORWARD_LIST 1
+#endif
 
-#ifdef NAMED_SUB_IS_FORWARD_LIST
+#if NAMED_SUB_IS_FORWARD_LIST
 #  include "forward_list_as_map.h"
 #else
 #include <map>
@@ -386,7 +388,7 @@ class irept
   : public non_sharing_treet<
       irept,
 #endif
-#ifdef NAMED_SUB_IS_FORWARD_LIST
+#if NAMED_SUB_IS_FORWARD_LIST
       forward_list_as_mapt<irep_namet, irept>>
 #else
       std::map<irep_namet, irept>>
--- a/.travis.yml
+++ b/.travis.yml
@@ -182,7 +182,7 @@ jobs:
         - WITH_MEMORY_ANALYZER=1
       script: echo "Not running any tests for a debug build."
 
-    # cmake build using g++-7, enable NAMED_SUB_IS_FORWARD_LIST
+    # cmake build using g++-7, disable NAMED_SUB_IS_FORWARD_LIST
     - stage: Test different OS/CXX/Flags
       os: linux
       sudo: false
@@ -205,7 +205,7 @@ jobs:
       install:
         - ccache -z
         - ccache --max-size=1G
-        - cmake -H. -Bbuild '-DCMAKE_BUILD_TYPE=Release' '-DCMAKE_CXX_COMPILER=/usr/bin/g++-7' '-DCMAKE_CXX_FLAGS=-DNAMED_SUB_IS_FORWARD_LIST' '-DWITH_MEMORY_ANALYZER=On'
+        - cmake -H. -Bbuild '-DCMAKE_BUILD_TYPE=Release' '-DCMAKE_CXX_COMPILER=/usr/bin/g++-7' '-DCMAKE_CXX_FLAGS=-DNAMED_SUB_IS_FORWARD_LIST=0' '-DWITH_MEMORY_ANALYZER=On'
         - git submodule update --init --recursive
         - cmake --build build -- -j4
       script: (cd build; bin/unit "[core][irept]")
--- a/src/util/irep.cpp
+++ b/src/util/irep.cpp
@@ -92,7 +92,7 @@ void irept::set(const irep_namet &name, const long long value)
 
 void irept::remove(const irep_namet &name)
 {
-#ifdef NAMED_SUB_IS_FORWARD_LIST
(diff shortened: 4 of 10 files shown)


kroening:
Can you split? The constructor seems obvious and trivial, whereas the change to list will require us to do a lot of consulting & benchmarking.

tautschnig:
@kroening The constructor change is now in #4460.

tautschnig:
Just for the record: on ReachSafety-ECA with this change we can perform 3819.81 `symex_step` calls per second, compared to 2752.28 calls per second with develop. I'll now try to figure out what exactly makes so much of a difference.

tautschnig:
It seems that just about any `irept`-related operation is cheaper (creating fresh `irept`s of course, but also `irept::operator==`).

smowton (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: d1a788b53e53f832dd16126cd4ad02bdd07973e2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106442057

src/util/irep.h
```diff
@@ -18,7 +18,9 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #define SHARING
 // #define HASH_CODE
-// #define NAMED_SUB_IS_FORWARD_LIST
+#ifndef NAMED_SUB_IS_FORWARD_LIST
+#  define NAMED_SUB_IS_FORWARD_LIST
```
martin-cs: I'm probably being thick here but why not just `#define` it?

src/memory-analyzer/gdb_api.h
```diff
@@ -24,14 +27,13 @@ Author: Malte Mues <mail.mues@gmail.com>
...
+#  include <unistd.h>
 
-#include <exception>
-#include <forward_list>
+#  include <exception>
+#  include <forward_list>
+#  include <map>
 
-#include <util/exception_utils.h>
+#  include <util/exception_utils.h>
```
martin-cs: The indenting of includes seems inconsistent with the rest of the code.

martin-cs (commented):
1. I think we're going to need some large scale benchmarking before we can make this change.

2. Might part of the performance benefit be fitting an `irept` into a single cache line?

3. My guess is that there is a clear bias on which of the entries of the map are most used.  Does it make sense to sort them / use a list that pivots them to the front?

src/util/irep.h
tautschnig: This ensures compatibility with any builds (such as our CI!) using `-DNAMED_SUB_IS_FORWARD_LIST` on the compiler command line.

src/memory-analyzer/gdb_api.h
tautschnig: This is the preprocessor indentation used as of #4461 (which only applies to preprocessor directives).

kroening: That one seems highly unusual -- can clang-format be made not to do that?

tautschnig: In my opinion, this one is entirely appropriate, because of the `#if` that precedes it. But maybe I'm the only one thinking so and if that's the case we can of course avoid this by adding a `clang-format off`.

martin-cs: How much of a diff would it produce if we indent vs. if we don't.  I think that changes to the coding style that are a departure from the existing style (regardless of whether they are "right" or "wrong") need some thought and justification.  To me 'consistent' is a key priority.

tautschnig:
Although I responded with a 👍 I didn't actually provide meaningful comments to what @martin-cs said:

> I think we're going to need some large scale benchmarking before we can make this change.

Yes, all proposals or efforts most welcome.

> Might part of the performance benefit be fitting an `irept` into a single cache line?

Given that a typical cache line size today is 64B and that on 64-bit Linux systems we move from `sizeof(irept::dt) == 80` to `sizeof(irept::dt) == 40` you may very well be right!

> My guess is that there is a clear bias on which of the entries of the map are most used. Does it make sense to sort them / use a list that pivots them to the front?

We might as well want to measure the median size of the map: I'd suspect it would be 1 or 2 (expressions carry a type and maybe a source location).

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 0b0ee0591435dba607ff5fc6c11a887cd63c2540).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108706493

tautschnig:
@danpoe @smowton @romainbrenguier Would you be able to do some benchmarking of this change? I've also asked elsewhere, but it would be nice to have as much data as possible!

danpoe:
On a set of methods from Apache Tika, symex was overall 21% faster with the changes in this PR (20.8s vs 26.4s).

smowton:
On my large Webgoat run I find running time is practically identical, but the memory usage is down around 30%.

martin-cs:
> Although I responded with a 👍 I didn't actually provide meaningfu
> l comments to what @martin-cs said:

No worries -- I ignored your responses for a month! :-\


> > I think we're going to need some large scale benchmarking before we
> > can make this change.
> 
> Yes, all proposals or efforts most welcome.

Am a bit short on time but will try when I have some.


> > My guess is that there is a clear bias on which of the entries of
> > the map are most used. Does it make sense to sort them / use a list
> > that pivots them to the front?
> 
> We might as well want to measure the median size of the map: I'd
> suspect it would be 1 or 2 (expressions carry a type and maybe a
> source location). 

Yes; that would be the thing to do before thinking of any
optimisations.

romainbrenguier:
Here are the results on my benchmark:
chart (3)
overall forward lists were about 10% faster

martin-cs:
That's starting to look quite convincing isn't it?

tautschnig:
No known regressions (including external feedback), just needs final approval from at least one of @chrisr-diffblue @forejtv @peterschrammel @thk123.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 754dbfc39aee2fc15978f43abf08b630be7d678a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112051965

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: cfcc7e8e4070e9fde997676c742b14edd8c92777).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112097093

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 2b9849af73f150bab0ffd592b807aef7eea1feea).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112365181

peterschrammel (approved):

PR #4460, opened 2019-03-30
Ensure typet constructor is always compatible with GCC 5's STL [blocks: #4458]

tautschnig:
When enabling NAMED_SUB_IS_FORWARD_LIST, compilation failed as the
std::forward_list constructors shipped with GCC 5 do not support an
empty initializer list. GCC 6 and later do not have any such problem.

### DIFF
--- a/src/util/type.h
+++ b/src/util/type.h
@@ -32,10 +32,17 @@ public:
 
   explicit typet(const irep_idt &_id):irept(_id) { }
 
+#if defined(__clang__) || !defined(__GNUC__) || __GNUC__ >= 6
   typet(irep_idt _id, typet _subtype)
     : irept(std::move(_id), {}, {std::move(_subtype)})
   {
   }
+#else
+  typet(irep_idt _id, typet _subtype) : irept(std::move(_id))
+  {
+    subtype() = std::move(_subtype);
+  }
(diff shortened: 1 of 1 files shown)


src/util/type.h
```diff
@@ -32,10 +32,18 @@ class typet:public irept
 
   explicit typet(const irep_idt &_id):irept(_id) { }
 
+#if !defined(__GNUC__) || __GNUC__ >= 6
   typet(irep_idt _id, typet _subtype)
     : irept(std::move(_id), {}, {std::move(_subtype)})
   {
   }
+#else
```
smowton, line 40: Note this will also be used under Clang, which pretends to be gcc 4.2.1

smowton (approved):

src/util/type.h
tautschnig: Ah, thanks a lot, I've added `#defined(__clang__)` to the above `#if`.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: ca155d3db2b8777eec518900649642dc3ecda03a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106425825

PR #4461, opened 2019-03-30
Make clang-format indent nested preprocessor directives

tautschnig:
Since clang-format 6 there is support to indent preprocessor directives,
enabling formatting like

```
 #if foo
 #  define bar
 #endif
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -60,6 +60,7 @@ ForEachMacros: [
   'forall_subtypes',
   'Forall_subtypes']
 IndentCaseLabels: 'false'
+IndentPPDirectives: AfterHash
 IndentWidth: '2'
 IndentWrappedFunctionNames: 'false'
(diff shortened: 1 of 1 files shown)


smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 98f619f5b1c39289211beb7e7c7dd322f3f3ef04).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106430068

martin-cs:
Does this mean we are going to have to reformat all of the includes inside header guards?

tautschnig:
> Does this mean we are going to have to reformat all of the includes inside header guards?

No, clang-format is able to tell apart header guards from other `#if`. What you've seen in #4458 is just because that entire file is, in addition to the header guard, also `#if` guarded for Linux-only.

PR #4463, opened 2019-03-30
Use sharing_mapt as container in value_sett

tautschnig:
goto-symex copies a state, which contains a value set, at each branching
point. The paths originating at that branch point typically share most
of the points-to information. Thus the copy is unnecessarily expensive,
and so is the merge at the next join point, which will also result in
one of the value sets to be destroyed. With sharing_mapt, the copy has
constant cost, merging can make use of the delta view, and the cost of
destruction only depends on the size of the delta.

Includes the cleanup-commit from #4462.

Performance evaluation to be done.

### DIFF
--- a/src/goto-symex/symex_main.cpp
+++ b/src/goto-symex/symex_main.cpp
@@ -676,16 +676,16 @@ void goto_symext::try_filter_value_sets(
   }
   if(jump_taken_value_set && !erase_from_jump_taken_value_set.empty())
   {
-    value_sett::entryt *entry = jump_taken_value_set->get_entry_for_symbol(
+    auto entry_index = jump_taken_value_set->get_index_of_symbol(
       symbol_expr->get_identifier(), symbol_type, "", ns);
     jump_taken_value_set->erase_values_from_entry(
-      *entry, erase_from_jump_taken_value_set);
+      *entry_index, erase_from_jump_taken_value_set);
   }
   if(jump_not_taken_value_set && !erase_from_jump_not_taken_value_set.empty())
   {
-    value_sett::entryt *entry = jump_not_taken_value_set->get_entry_for_symbol(
+    auto entry_index = jump_not_taken_value_set->get_index_of_symbol(
       symbol_expr->get_identifier(), symbol_type, "", ns);
     jump_not_taken_value_set->erase_values_from_entry(
-      *entry, erase_from_jump_not_taken_value_set);
+      *entry_index, erase_from_jump_not_taken_value_set);
   }
 }
--- a/src/pointer-analysis/value_set.cpp
+++ b/src/pointer-analysis/value_set.cpp
@@ -48,19 +48,17 @@ bool value_sett::field_sensitive(const irep_idt &id, const typet &type)
   return type.id() == ID_struct || type.id() == ID_struct_tag;
 }
 
-value_sett::entryt *value_sett::find_entry(const irep_idt &id)
-{
-  auto found = values.find(id);
-  return found == values.end() ? nullptr : &found->second;
-}
-
 const value_sett::entryt *value_sett::find_entry(const irep_idt &id) const
 {
   auto found = values.find(id);
-  return found == values.end() ? nullptr : &found->second;
+  return !found.has_value() ? nullptr : &(found->get());
 }
 
-value_sett::entryt &value_sett::get_entry(const entryt &e, const typet &type)
+void value_sett::update_entry(
+  const entryt &e,
+  const typet &type,
+  const object_mapt &new_values,
+  bool add_to_sets)
 {
   irep_idt index;
 
@@ -69,10 +67,27 @@ value_sett::entryt &value_sett::get_entry(const entryt &e, const typet &type)
   else
     index=e.identifier;
 
-  std::pair<valuest::iterator, bool> r =
-    values.insert(std::pair<irep_idt, entryt>(index, e));
-
-  return r.first->second;
+  auto existing_entry = values.find(index);
+  if(existing_entry.has_value())
+  {
+    entryt replacement = *existing_entry;
+    if(add_to_sets)
+    {
+      if(make_union(replacement.object_map, new_values))
+        values.replace(index, std::move(replacement));
+    }
+    else
+    {
+      replacement.object_map = new_values;
+      values.replace(index, std::move(replacement));
+    }
+  }
+  else
+  {
+    entryt new_entry = e;
+    new_entry.object_map = new_values;
+    values.insert(index, std::move(new_entry));
+  }
 }
 
 bool value_sett::insert(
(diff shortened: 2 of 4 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 442b500f9cb2f4e10548b220cf7f7b7ad306cbe4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106434546

tautschnig:
Some stats for SV-COMP's ReachSafety-ECA (which barely uses pointers, i.e., the value set would mostly be near-empty):
* `goto_statet::goto_statet(goto_symex_statet const&)`: develop: 1423.62 seconds; this branch: 459.92 seconds
* Removing entries from the goto_state list: develop: 587.84 seconds; this branch: 246.81 seconds
* `value_sett::make_union`: develop: 645.59 seconds; this branch: 223.34 seconds
* `value_sett::assign`: develop: 306.79 seconds; this branch: 415.09 seconds

So adding new values to the value set has become more expensive (which is not unexpected), but merging and cloning is a lot cheaper (as intended).  @romainbrenguier @smowton would you be able to contribute some benchmarking on Java code, which makes a lot more use of value sets?

src/pointer-analysis/value_set.h
```diff
@@ -297,11 +295,7 @@ class value_sett
   ///
   /// The components of the ID are thus duplicated in the `valuest` key and in
   /// `entryt` fields.
-  #ifdef USE_DSTRING
```
martin-cs: Are we reaching the point of saying that `USE_DSTRING` should always be on?

martin-cs (approved):

src/pointer-analysis/value_set.h
tautschnig: Well, we do still _build_ with `USE_DSTRING` disabled, but I don't think the regression tests even pass. Though this change might be a step towards making that more likely, as it makes the code more uniform.

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 441eb6c568db89f7e9e02bc593cd7cd16614aae7).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106528011

src/pointer-analysis/value_set.cpp
```diff
@@ -69,10 +67,22 @@ value_sett::entryt &value_sett::get_entry(const entryt &e, const typet &type)
...
-  std::pair<valuest::iterator, bool> r =
-    values.insert(std::pair<irep_idt, entryt>(index, e));
-
-  return r.first->second;
+  auto existing_entry = values.find(index);
+  if(existing_entry.has_value())
+  {
+    entryt replacement = *existing_entry;
+    if(add_to_sets)
+      make_union(replacement.object_map, new_values);
```
danpoe: We could check the return value of `make_union()` and only do the replacement when the object map has changed.

src/pointer-analysis/value_set.cpp
```diff
@@ -69,10 +67,22 @@ value_sett::entryt &value_sett::get_entry(const entryt &e, const typet &type)
...
+    else
+      replacement.object_map = new_values;
```
danpoe, line 81: We could check whether `new_values.read() != existing_entry->object_map.read()`  before doing the replacement. Not sure about the impact on performance though. We might avoid breaking sharing in some cases but there'd be additional cost for the comparisons.

danpoe (approved):
LGTM

romainbrenguier:
Here are the results from my benchmarks, globaly using sharing_mapt was 13% faster and seems to scale better:
chart

src/goto-symex/symex_main.cpp
```diff
@@ -676,14 +676,14 @@ void goto_symext::try_filter_value_sets(
   }
   if(jump_taken_value_set && !erase_from_jump_taken_value_set.empty())
   {
-    value_sett::entryt *entry = jump_taken_value_set->get_entry_for_symbol(
+    auto entry = jump_taken_value_set->get_index_of_symbol(
```
owen-mc-diffblue: Rename `entry` to `entry_name` or `entry_index` :banana:

src/goto-symex/symex_main.cpp
```diff
@@ -676,14 +676,14 @@ void goto_symext::try_filter_value_sets(
...
       symbol_expr->get_identifier(), symbol_type, "", ns);
     jump_taken_value_set->erase_values_from_entry(
       *entry, erase_from_jump_taken_value_set);
   }
   if(jump_not_taken_value_set && !erase_from_jump_not_taken_value_set.empty())
   {
-    value_sett::entryt *entry = jump_not_taken_value_set->get_entry_for_symbol(
+    auto entry = jump_not_taken_value_set->get_index_of_symbol(
```
owen-mc-diffblue: :banana:

src/pointer-analysis/value_set.cpp
```diff
@@ -491,11 +472,11 @@ void value_sett::get_value_set_rec(
   }
   else if(expr.id()==ID_symbol)
   {
-    const entryt *entry = get_entry_for_symbol(
+    auto entry = get_index_of_symbol(
```
owen-mc-diffblue: :banana:

owen-mc-diffblue (approved):

smowton:
Evaluated this on one of our medium-sized Java applications (a particular WebGoat lesson) -- JBMC execution time down from ~30s to ~15s.

danpoe:
Here are some results for the runtime of symex only, with `--unwind 1`, on several methods from Apache Tika on which symex takes between 5 and 15 sec:

Rplot

Overall with sharing it was 7% faster, including the changes to `update_entry()` it was 9% faster.

src/pointer-analysis/value_set.cpp
tautschnig: I think there's a risk this is way too expensive and thus I'd rather not add this change.

src/pointer-analysis/value_set.cpp
tautschnig: Done.

src/goto-symex/symex_main.cpp
tautschnig: Done.

src/goto-symex/symex_main.cpp
tautschnig: Done.

src/pointer-analysis/value_set.cpp
tautschnig: Done.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d4b811b95ecebca3f5582962b2580eb8ad671350).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106864446

src/pointer-analysis/value_set.cpp
danpoe: Yeah that's ok, we'd probably have to specifically benchmark this. My hunch is that the map comparisons could be cheap on average if it's likely that they differ in size or that we find a mismatch quickly.

PR #4464, opened 2019-03-30
Use sharing_mapt to store propagation map

tautschnig:
goto-symex copies a state, which contains a propagation map, at each
branching point. The paths originating at that branch point typically
share most of the constants being propagated. Thus the copy is
unnecessarily expensive, and so is the merge at the next join point,
which will also result in one of the propagation maps to be destroyed.
With sharing_mapt, the copy has constant cost, and the cost of
destruction only depends on the size of the delta.

### DIFF
--- a/src/goto-symex/goto_state.cpp
+++ b/src/goto-symex/goto_state.cpp
@@ -17,7 +17,10 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 /// because there aren't any current callers.
 void goto_statet::output_propagation_map(std::ostream &out)
 {
-  for(const auto &name_value : propagation)
+  sharing_mapt<irep_idt, exprt>::viewt view;
+  propagation.get_view(view);
+
+  for(const auto &name_value : view)
   {
     out << name_value.first << " <- " << format(name_value.second) << "\n";
   }
@@ -86,7 +89,11 @@ void goto_statet::apply_condition(
         increase_generation(
           l1_identifier, l1_lhs, previous_state.get_l2_name_provider());
 
-        propagation[l1_identifier] = rhs;
+        const auto propagation_entry = propagation.find(l1_identifier);
+        if(!propagation_entry.has_value())
+          propagation.insert(l1_identifier, rhs);
+        else if(propagation_entry->get() != rhs)
+          propagation.replace(l1_identifier, rhs);
 
         value_set.assign(l1_lhs, rhs, ns, true, false);
       }
--- a/src/goto-symex/goto_symex_state.cpp
+++ b/src/goto-symex/goto_symex_state.cpp
@@ -201,8 +201,14 @@ void goto_symex_statet::assignment(
   // for value propagation -- the RHS is L2
 
   if(!is_shared && record_value && goto_symex_is_constantt()(rhs))
-    propagation[l1_identifier] = rhs;
-  else
+  {
+    const auto propagation_entry = propagation.find(l1_identifier);
+    if(!propagation_entry.has_value())
+      propagation.insert(l1_identifier, rhs);
+    else if(propagation_entry->get() != rhs)
+      propagation.replace(l1_identifier, rhs);
+  }
+  else if(propagation.has_key(l1_identifier))
     propagation.erase(l1_identifier);
 
   {
@@ -289,8 +295,8 @@ goto_symex_statet::rename(exprt expr, const namespacet &ns)
         // L1 identifiers are used for propagation!
         auto p_it = propagation.find(ssa.get_identifier());
 
-        if(p_it != propagation.end())
-          expr=p_it->second; // already L2
+        if(p_it.has_value())
+          expr = *p_it; // already L2
         else
           ssa = set_indices<L2>(std::move(ssa), ns).get();
       }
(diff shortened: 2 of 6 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 496eb15e1202d55078552ff43e4cfec620ca9fff).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106435134

tautschnig:
Some stats for SV-COMP's ReachSafety-ECA:
* `goto_statet::goto_statet(goto_symex_statet const&)`: develop: 1423.62 seconds; this branch: 827.72 seconds (cf. #4463, which reduces the cost in a different way, and their savings would actually add up)
* Removing entries from the goto_state list: develop: 587.84 seconds; this branch: 372.36 seconds (again, the savings of #4463 and this one would actually add up)
* L2-renaming: develop: 457.98 seconds; this branch: 494.23 seconds

Lookups in the propagation map are more expensive (as expected), but merging and cloning is a lot cheaper (as intended).

src/goto-symex/goto_state.cpp
```diff
@@ -86,7 +89,10 @@ void goto_statet::apply_condition(
         increase_generation(
           l1_identifier, l1_lhs, previous_state.get_l2_name_provider());
 
-        propagation[l1_identifier] = rhs;
+        if(propagation.has_key(l1_identifier))
```
smowton: Might want to just give `sharing_mapt` a primitive `set` operation to save the double lookup

smowton (approved):

src/goto-symex/goto_state.cpp
tautschnig: @danpoe maybe?

danpoe: I'd recommend to use `find(l1_identifier)` here and then check that `rhs` is different from the current value in the map, and only then do the replacement. When using this idiom a separate `set()` method should be unnecessary.

src/goto-symex/goto_symex_state.cpp
```diff
@@ -200,7 +200,12 @@ void goto_symex_statet::assignment(
   // for value propagation -- the RHS is L2
 
   if(!is_shared && record_value && goto_symex_is_constantt()(rhs))
-    propagation[l1_identifier] = rhs;
+  {
+    if(propagation.has_key(l1_identifier))
+      propagation.replace(l1_identifier, rhs);
```
danpoe: Check if `rhs` is different from the value already in the map (see above).

danpoe (approved):

smowton:
This one made almost no difference to our Webgoat lesson, I expect because generally the value-set is much bigger, but if we merge #3619 then that will no longer hold and this will be more important.

danpoe:
Here are some results for the runtime of symex only, with `--unwind 1`, on several methods from Apache Tika on which symex takes between 5 and 15 sec:
Rplot

Overall with sharing it was 9% slower, including the changes to `apply_condition()` and `assignment()` it was 7% slower. Somewhat unexpectedly, when using sharing in both the value set and the propagation map (i.e., this PR and PR #4463, both including the fixes) overall symex was 8% faster than develop. I would have expected the effects of both to cancel each other out. Does anyone have an idea of why it could behave like this?

tautschnig:
I'll try to implement the suggestions I have received across the two PRs, and would then tend to merge #4463, which seems to be uniformly better. We can then re-evaluate this one to see whether the somewhat surprising observation made by @danpoe carries on.

src/goto-symex/goto_state.cpp
tautschnig: Done.

src/goto-symex/goto_symex_state.cpp
tautschnig: Done.

tautschnig:
@danpoe As #4463 is merged I have rebased this one. Would you (and possibly also @romainbrenguier and @smowton) mind doing another benchmarking run to see where we are?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: be7d917525c7eeda5ef26bb68f2c192d6e6242b3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106873800

danpoe:
Here are some results for the runtime of symex only, with `--unwind 1`, on several methods from Apache Tika on which symex takes between 5 and 15 sec:

Rplot

The runtime is essentially the same as on the latest develop (overall with sharing it's 0.3% faster).

tautschnig:
It's probably best to put this on hold and look at other performance improvements first. It may also become a lot more relevant once we have field sensitivity in place as propagation maps will then likely be larger.

allredj:
Yep, that's also the conclusion we reached with @danpoe.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 3bb8488fbc0b662850889d6b5fc4900580457d0b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107164126

smowton:
Needs `propagation.erase` -> `propagation.erase_if_exists` in `symex_assign.cpp`

smowton:
Benchmark results for a long-running Webgoat analysis:

Develop: 77s
#4464: 57s
#4464 + #4505: 26s

tautschnig:
> Needs `propagation.erase` -> `propagation.erase_if_exists` in `symex_assign.cpp`

Thank you!! Done. Based on @smowton's data I'm inclined to merge this once CI passes.

PR #4465, opened 2019-03-30
[RFC] Make doxygen CI task fail on missing documentation

tautschnig:
Only review and comment on the first commit, the second one just records all resulting (>13k) Doxygen warnings thus generated. 


martin-cs (approved):
CI > manual view so, yes.

owen-mc-diffblue:
The DOXYGEN-CHECK Travis job is already doing something - how does this affect it? Is it that previously it would fail if you had partial documentation for an edited function (e.g. a missing parameter) but now it will also fail if there isn't any documentation?

tautschnig:
> Is it that previously it would fail if you had partial documentation for an edited function (e.g. a missing parameter) but now it will also fail if there isn't any documentation?

Indeed it would now require documentation to be added for any non-private member. But our list of undocumented members or fields is still very long as doxygen produces >13k warnings...

owen-mc-diffblue:
It would be good to force people to document the public methods they add. Possible downsides: (a) currently I think it's considered acceptable to not provide any documentation if the method is very short and well-described by its identifier, or to not document parameters if there isn't anything interesting to say. Presumably this will force us to do that for public methods; (b) am I right in thinking that if you edit a public method that isn't documented you will be forced to write the documentation for it to get CI to pass? In some cases that seems like an unnecessary burden, e.g. if you change the name of a function then you'll have to document all the undocumented public functions containing a callsite. (Please correct me if I'm wrong about this.)

peterschrammel:
>  seems like an unnecessary burden

I think it would be generally good to get all public declarations documented.

> if you change the name of a function then you'll have to document all the undocumented public functions containing a callsite

I think it only complains if the function declaration / definition (without the body) is in the diff.

TGWDB:
@tautschnig This does not seem like a bad idea, but also has been untouched for a long time. Do you wish to rebase/rerun this and continue the discussion, or should we close the PR as not progressing?

TGWDB:
Closing due to age and lack of progress/response. Please reopen if you have updates/progress or believe this closure was erroneous.

PR #4466, opened 2019-03-31
Constify collect_virtual_function_callees

smowton:
This was always a const operation in spirit, as it retrieves a set of potential virtual function
overrides without making any changes to the GOTO program, but was recently changed to take a non-
const symbol table because it shares code with remove_virtual_function(s), which indeed add new
symbols to the symbol table.

To once again allow calling collect_virtual_function_callees in a const context, this splits the
implementation of remove_virtual_functions into those parts that require a non-const reference and
those for which a const symbol table will suffice.

### DIFF
--- a/src/goto-programs/remove_virtual_functions.cpp
+++ b/src/goto-programs/remove_virtual_functions.cpp
@@ -22,38 +22,26 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "remove_skip.h"
 #include "resolve_inherited_component.h"
 
-class remove_virtual_functionst
+class get_virtual_calleest
 {
 public:
-  remove_virtual_functionst(
-    symbol_table_baset &_symbol_table,
-    const class_hierarchyt &_class_hierarchy);
+  get_virtual_calleest(
+    const symbol_table_baset &_symbol_table,
+    const class_hierarchyt &_class_hierarchy)
+    : ns(_symbol_table),
+      symbol_table(_symbol_table),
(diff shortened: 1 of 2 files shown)


src/goto-programs/remove_virtual_functions.cpp
```diff
@@ -70,14 +58,38 @@ class remove_virtual_functionst
   get_method(const irep_idt &class_id, const irep_idt &component_name) const;
 };
 
-remove_virtual_functionst::remove_virtual_functionst(
-  symbol_table_baset &_symbol_table,
-  const class_hierarchyt &_class_hierarchy)
-  : ns(_symbol_table),
-    symbol_table(_symbol_table),
-    class_hierarchy(_class_hierarchy)
+class remove_virtual_functionst : public get_virtual_calleest
```
tautschnig: Is "is-a" really the correct relationship? My guess would have been that `remove_virtual_functionst` uses-a/has-a `get_virtual_calleest`?!

smowton: Sure, was really just minimising churn. Changed to use-a.

tautschnig (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9bc196befd95489e789d49429a5dc70d29335bff).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/106467703

PR #4480, opened 2019-04-03
Get lhs_object from byte_extract for goto trace

peterschrammel:
Otherwise fields are missing from the XML output, which breaks downstream integration of a commercial CBMC user.

### DIFF
--- a/regression/cbmc/xml-trace/test.desc
+++ b/regression/cbmc/xml-trace/test.desc
@@ -1,0 +1,25 @@
+CORE broken-smt-backend
+main.c
+--xml-ui --function test --little-endian
+^EXIT=10$
+^SIGNAL=0$
+VERIFICATION FAILED
+<assignment assignment_type="state" base_name="u" display_name="test::u" hidden="false" identifier="test::u" mode="C" step_nr="\d+" thread="0">
+<location file=".*" function="test" line="\d+" working-directory=".*"/>
+<type>union myunion</type>
+<full_lhs>byte_extract_little_endian\(u, 0ll?, .*int.*\)</full_lhs>
+<full_lhs_value>\d+ll?</full_lhs_value>
+<value>\{ \.i=\d+ll? \}</value>
+<value_expression>
+<union>
+<member member_name="i">
+<integer binary="\d+" c_type=".*int.*" width="\d+">\d+</integer>
+</member>
+</union>
+</value_expression>
+</assignment>
+--
+^warning: ignoring
+--
+Tests whether the assignment XML output contains all the required elements
+such as identifier, full_lhs, full_lhs_value and value.
--- a/src/goto-programs/goto_trace.cpp
+++ b/src/goto-programs/goto_trace.cpp
@@ -16,6 +16,7 @@ Author: Daniel Kroening
 #include <ostream>
 
 #include <util/arith_tools.h>
+#include <util/byte_operators.h>
 #include <util/format_expr.h>
 #include <util/range.h>
 #include <util/string_utils.h>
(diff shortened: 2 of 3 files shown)


src/goto-programs/goto_trace.cpp
```diff
@@ -35,6 +36,12 @@ static optionalt<symbol_exprt> get_object_rec(const exprt &src)
     return get_object_rec(to_index_expr(src).array());
   else if(src.id()==ID_typecast)
     return get_object_rec(to_typecast_expr(src).op());
+  else if(
+    src.id() == ID_byte_extract_little_endian ||
+    src.id() == ID_byte_extract_big_endian)
+  {
+    return get_object_rec(to_byte_extract_expr(src).op());
```
tautschnig, line 43: Remind me, do we actually consider the offset? What you propose is consistent with what's done for index and member expressions, but I'm still wondering whether the overall result is meaningful.

regression/cbmc/xml-trace/test.desc
```diff
@@ -0,0 +1,25 @@
+CORE
+main.c
+--xml-ui --function test
+^EXIT=10$
+^SIGNAL=0$
+VERIFICATION FAILED
+<assignment assignment_type="state" base_name="u" display_name="test::u" hidden="false" identifier="test::u" mode="C" step_nr="\d+" thread="0">
+<location file=".*" function="test" line="\d+" working-directory=".*"/>
+<type>union myunion</type>
+<full_lhs>byte_extract_little_endian\(u, 0l, .*int.*\)</full_lhs>
```
tautschnig: Please also add `--little-endian` to the command line to make sure this test is portable.

tautschnig (approved):
Great to see this improving - I don't think you'd even need to add a new test, existing tests like `regression/cbmc/array_constraints1` also have this problem.

src/goto-programs/goto_trace.cpp
peterschrammel: No, it was not considered. The meaningfulness is questionable, but that probably goes beyond this fix.
It might be good to actually specify what these fields are supposed to mean precisely.

tautschnig: Here's an example to validate:
````c
#include <assert.h>

int main()
{
  int x;
  char *p = (char *)&x + 2;
  *p = 1;
  assert(x == 0);
}
````
The trace currently includes `byte_extract_little_endian(x, 2l, char)=1 (00000001)` - which might not be downstream-consumable, but at least is accurate. If after this PR the result is `x=1` then there's a problem I'm afraid.

kroening: goto_trace_steps offer two bits of information:
1) The thing returned by `get_lhs_object()` -- that is the object as defined in the C standard, and `x` is the right answer.
2) The `full_lhs`, which is documented to be "after dereferencing". It is supposed to be exactly what gets assigned. The byte_extract expression above is the best answer for that, in my view.

The trace given to the user should show the latter. We may wish to print byte_extract expressions there as a pointer typecast plus dereferencing.

tautschnig: Thank you for clarifying! +1 for printing `byte_extract` as some C expression, which will be important for SV-COMP as well. But I accept that this might be outside the scope of this PR.

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 60b532870fbbb34f30b6693a6221280c6c5e6c04).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107058275

PR #4481, opened 2019-04-04
i2x cast back to int for byte, short and char

JohnDumbell:
It's mentioned in the JVM spec (https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html#jvms-6.5.i2s) that char, byte, and short types should be truncated and then returned to an int when you have an int-to-x operation.

We weren't doing that for certain times, such as the example shown in the test. 


jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -1582,6 +1582,14 @@ code_blockt java_bytecode_convert_methodt::convert_instructions(
...
       typet type=java_type_from_char(statement[2]);
       results[0] = typecast_exprt::conditional_cast(op[0], type);
+
+      // These types get converted/truncated then immediately turned back into
+      // ints again, so we just double-cast here.
+      if(
+        type == java_char_type() || type == java_byte_type() ||
+        type == java_short_type())
+        results[0] =
+          typecast_exprt::conditional_cast(results[0], java_int_type());
```
tautschnig: Braces after multi-line conditions, please.

jbmc/regression/jbmc/short_arg_to_int_param/ShortToInt.java
```diff
@@ -0,0 +1,10 @@
+public class ShortToInt {
+    public void main(int arg) {
+        intMethod((short)arg);
+        assert true;
+    }
+
+    public int intMethod(int i) {
+        return i;
+    }
+}
```
tautschnig: So did this test actually fail before? It's not obvious to me what was causing trouble here before (but also I'm a Java noob). I think I would expect an assertion over the value of `i`.

JohnDumbell: The issue was to do with the `intMethod((short)arg);` call, I just saw this shown on a method that also returned an int so kept that little detail. The invariant here: https://github.com/diffblue/cbmc/blob/9bc196befd95489e789d49429a5dc70d29335bff/src/goto-programs/remove_virtual_functions.cpp#L160-L169 
fails, because the `intMethod((short)arg);` argument is a short (due to the cast) but the parameter is expecting an int.

smowton: (and the JVM spec actually says the types must be correct on the stack upon execution of a call instruction, we just implemented the cast opcodes wrongly before so that wasn't always true)

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -1582,6 +1582,16 @@ code_blockt java_bytecode_convert_methodt::convert_instructions(
...
       results[0] = typecast_exprt::conditional_cast(op[0], type);
+
+      // These types get converted/truncated then immediately turned back into
+      // ints again, so we just double-cast here.
+      if(
+        type == java_char_type() || type == java_byte_type() ||
+        type == java_short_type())
+      {
+        results[0] =
+          typecast_exprt::conditional_cast(results[0], java_int_type());
```
smowton: No need for conditional-cast, as the types *must* differ in these cases

jbmc/regression/jbmc/short_arg_to_int_param/ShortToInt.java
tautschnig: Ok, thanks both for the elaboration. I think it would be good to squash the two commits as otherwise there isn't sufficient context to understand what this test is actually testing.

smowton (commented):
Please also add tests for all Java primitive types, to verify we're doing the right thing elsewhere (so the cases should be `byte`, `short`, `long`, `float`, `double`, `char`. Double-check with `javap -c` that we indeed have an `i2x` opcode leading directly into a function call in each case.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 7b89fe8869565bc81d3f847080a54fac64602d05).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107047330

JohnDumbell:
Fixed up. @smowton I can't do it with this exact pattern for long, float or double, as they need explicit casts back to int or playing around with things that won't immediately cause a i2x -> function call.

Can you think of a situation that would and/or another way to test this?

smowton:
```java
public class Test {

  public static void f(int x) {
    g((float)x);
  }

  public static void g(float y) { }

}
```

```
  public static void f(int);
    Code:
       0: iload_0
       1: i2f
       2: invokestatic  #2                  // Method g:(F)V
       5: return
```

So that achieves the desired test (that cast instructions don't lead to type inconsistency)

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: e95e2c9337819fd2d50fe3426fbf5cb0ad30e566).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107050393

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 93327fc178d325e367470f9657816facf0972ea9).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107052689

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 2974fcde81dc7a4f006f37578fd05b3324b3ef9a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107054274

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: dcec44cc2cb9e54d7fb032da47b159b227d6ff93).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107056206

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -1582,6 +1582,13 @@ code_blockt java_bytecode_convert_methodt::convert_instructions(
       PRECONDITION(op.size() == 1 && results.size() == 1);
       typet type=java_type_from_char(statement[2]);
       results[0] = typecast_exprt::conditional_cast(op[0], type);
+
+      // These types get converted/truncated then immediately turned back into
+      // ints again, so we just double-cast here.
+      if(
+        type == java_char_type() || type == java_byte_type() ||
+        type == java_short_type())
+        results[0] = typecast_exprt(results[0], java_int_type());
```
owen-mc-diffblue, line 1592: You still need braces here because of the multi-line conditional

owen-mc-diffblue (approved):

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
JohnDumbell: Ah, thanks, sorry.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d9f8beb3804cf61ddcdf80ff267dae31740468a1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107091253

smowton (approved):
Possible problem: do you require 2+ virtual overloads to trigger the invariant you were hitting in the first place? If not then LGTM.

JohnDumbell:
Doesn't seem linked particularly to that, the tests fail when the fix is removed, and pass when it's in (as does running my problem code).

Only the ones that needed the double-cast though. The other ones were fine.

PR #4483, opened 2019-04-04
Add precondition that pointers point to structs in the object factory

antlechner:
I've been trying to find out what kinds of exprts can be passed to the object factory, and in the case of pointers, the only two cases that appeared in the code were pointers to structs and pointers to `void*`. The latter had a comment saying that it applies to `#exception_value`s, which according to @smowton have been replaced.

I tried adding a precondition that pointers always point to structs, ran most of the tests locally and haven't found any failures yet. So I am assuming that it is now safe to make this change.

I would like to do some refactoring for the struct case in follow-up PRs, so if reviewers think that there is any way that a pointer could still end up in the object factory and point to something other than a struct, please let me know and I will close this PR and instead add a test for this case.

- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message). [Various existing tests require constructing nondet values in the object factory.]

### DIFF
--- a/jbmc/src/java_bytecode/java_object_factory.cpp
+++ b/jbmc/src/java_bytecode/java_object_factory.cpp
@@ -221,21 +221,19 @@ void java_object_factoryt::gen_pointer_target_init(
 
   const namespacet ns(symbol_table);
   const typet &followed_target_type = ns.follow(target_type);
+  PRECONDITION(followed_target_type.id() == ID_struct);
 
-  if(followed_target_type.id() == ID_struct)
+  const auto &target_class_type = to_java_class_type(followed_target_type);
+  if(has_prefix(id2string(target_class_type.get_tag()), "java::array["))
   {
-    const auto &target_class_type = to_java_class_type(followed_target_type);
-    if(has_prefix(id2string(target_class_type.get_tag()), "java::array["))
-    {
-      gen_nondet_array_init(
-        assignments, expr, depth + 1, update_in_place, location);
+    gen_nondet_array_init(
+      assignments, expr, depth + 1, update_in_place, location);
+    return;
+  }
+  if(target_class_type.get_base("java::java.lang.Enum"))
+  {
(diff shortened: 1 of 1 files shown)


jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -518,44 +519,35 @@ void java_object_factoryt::gen_nondet_pointer_init(
...
+  const irep_idt &struct_tag = struct_type.get_tag();
 
-    // If this is a recursive type of some kind AND the depth is exceeded, set
-    // the pointer to null.
-    if(!recursion_set_entry.insert_entry(struct_tag) &&
-      depth>=object_factory_parameters.max_nondet_tree_depth)
+  // If this is a recursive type of some kind AND the depth is exceeded, set
+  // the pointer to null.
+  if(
+    !recursion_set_entry.insert_entry(struct_tag) &&
```
owen-jones-diffblue: This seems like an unrelated formatting change. Put it in a separate commit or remove it from the PR.

owen-mc-diffblue (approved):
Assuming CI passes

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7cada1b19e06629e5b0460c1fecd4dd23fd51ddf).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107086687

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: d195cc2187c0650ce8751ad0b8b689c8206d6969).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107095507

smowton (approved):

antlechner:
This causes a few test failures in TG, all of which are related to Strings. I will try to reproduce these failures using JBMC and either fix the problem or revert this PR and add a regression test.

PR #4471, opened 2019-03-31
C++ destructors: Ensure "this" is a fully qualified name

tautschnig:
Typechecking does not magically expand `ID_this`, and there isn't sufficient context to use "cpp-this."

Fixes: #661

Only the second commit is new, the first commit is #4470.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- n/a Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4485, opened 2019-04-04
Extend and cleanup usability of __CPROVER_{r,w}_ok

tautschnig:
`__CPROVER_r_ok` and `__CPROVER_w_ok` had surprising behaviour that wasn't documented. Make the behaviour less surprising and document it.

### DIFF
--- a/regression/cbmc/r_w_ok1/main.c
+++ b/regression/cbmc/r_w_ok1/main.c
@@ -1,0 +1,26 @@
+#include <assert.h>
+#include <stdlib.h>
+
+int main()
+{
+  int *p = NULL;
+
+  assert(!__CPROVER_r_ok(p, sizeof(int)));
+  assert(!__CPROVER_w_ok(p, sizeof(int)));
+
+  p = malloc(sizeof(int));
+
+  assert(__CPROVER_r_ok(p, 1));
+  assert(__CPROVER_w_ok(p, 1));
+  assert(__CPROVER_r_ok(p, sizeof(int)));
+  assert(__CPROVER_w_ok(p, sizeof(int)));
+
+  size_t n;
+  char *arbitrary_size = malloc(n);
+
+  assert(__CPROVER_r_ok(arbitrary_size, n));
+  assert(__CPROVER_w_ok(arbitrary_size, n));
+
+  assert(__CPROVER_r_ok(arbitrary_size, n + 1));
+  assert(__CPROVER_w_ok(arbitrary_size, n + 1));
+}
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -1076,9 +1076,7 @@ void goto_checkt::pointer_validity_check(
 goto_checkt::conditionst
 goto_checkt::address_check(const exprt &address, const exprt &size)
 {
-  if(!enable_pointer_check)
-    return {};
-
+  PRECONDITION(local_bitvector_analysis);
   PRECONDITION(address.type().id() == ID_pointer);
   const auto &pointer_type = to_pointer_type(address.type());
 
@@ -1700,9 +1698,8 @@ void goto_checkt::goto_check(
 
   bool did_something = false;
 
-  if(enable_pointer_check)
-    local_bitvector_analysis =
-      util_make_unique<local_bitvector_analysist>(goto_function, ns);
+  local_bitvector_analysis =
+    util_make_unique<local_bitvector_analysist>(goto_function, ns);
 
   goto_programt &goto_program=goto_function.body;
 
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -1718,7 +1718,18 @@ void goto_checkt::goto_check(
       assertions.clear();
 
     if(i.has_condition())
+    {
       check(i.get_condition());
+
+      if(has_subexpr(i.get_condition(), [](const exprt &expr) {
+           return expr.id() == ID_r_ok || expr.id() == ID_w_ok;
+         }))
+      {
+        auto rw_ok_cond = rw_ok_check(i.get_condition());
+        if(rw_ok_cond.has_value())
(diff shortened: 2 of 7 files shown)


danielsn:
You might want to add a test where "r_ok" should pass, but "w_ok" should fail. Something like
```
void test() {
  const char* str = "foobar";
  assert(!__CPROVER_w_ok(str, 6));
  assert(__CPROVER_r_ok(str, 6));
}
```

tautschnig:
> You might want to add a test where "r_ok" should pass, but "w_ok" should fail.

Thank you very much for the suggestion - I have added such a test, but marked it as `KNOWNBUG` as we do not currently have a way do distinguish lvalues at this point.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 7f8f20e6d13accf56faaa6badac9ccbccec8fe00).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107129865

regression/cbmc/r_w_ok1/main.c
```diff
@@ -0,0 +1,26 @@
...
+  assert(__CPROVER_r_ok(p, sizeof(int)));
+  assert(__CPROVER_w_ok(p, sizeof(int)));
+
+  size_t n;
+  char *arbitrary_size = malloc(n);
+
+  assert(__CPROVER_r_ok(arbitrary_size, n));
+  assert(__CPROVER_w_ok(arbitrary_size, n));
+
+  assert(__CPROVER_r_ok(arbitrary_size, n + 1));
```
danielsn, line 24: Why does this succeed?

tautschnig: Without `--pointer-check` any use of `__CPROVER_{r,w}_ok` is now a no-op (and is now also documented as such), but with `--pointer-check` this fails as expected (there are two regression tests using the same `main.c` file included).

kroening: Is this what we want? I would find it more intuitive if these were independent.

tautschnig: It does make the implementation a lot easier, and also removes a need for documentation, so that seems much better. The only cost is that we will unconditionally need to run the local alias analysis.

tautschnig: Ok, this will require #4471 to be merged first. And another bug left to be debugged.

tautschnig: Hmm, actually there is more to it than just #4471, we also seem to be messing up `ID_this` elsewhere.

kroening: If the cost of the local alias analysis is a concern, there could be a quick scan for the predicate in the function; but so far, that analysis has not been cause for concern.

tautschnig: Agreed, I just need to fix bugs in the C++ and Java front-ends first as the local alias analysis needs to succeed in doing symbol-table lookups...

src/analyses/goto_check.cpp
```diff
@@ -1765,6 +1773,15 @@ void goto_checkt::goto_check(
 
       // the LHS might invalidate any assertion
       invalidate(code_assign.lhs());
+
+      if(has_subexpr(i.code.op1(), [](const exprt &expr) {
```
smowton: `i.code.op1` -> `code_assign.rhs`

src/analyses/goto_check.cpp
```diff
@@ -1765,6 +1773,15 @@ void goto_checkt::goto_check(
...
+           return expr.id() == ID_r_ok || expr.id() == ID_w_ok;
+         }))
+      {
+        auto rw_ok_cond = rw_ok_check(i.code.op1());
+        if(rw_ok_cond.has_value())
+          i.code.op1() = *rw_ok_cond;
```
smowton: Hang on, so if any subexpr was an r_ok or w_ok then you overwrite the whole RHS? That seems surprising, what's the rationale for blowing away arbitrary AST nodes here?

tautschnig: No, `rw_ok_check` will take care of only replacing parts of the RHS.

src/analyses/goto_check.cpp
tautschnig: Done.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 1a8b8340d1c5f7a31af0710ea59c5d28cba2454f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107785716

src/analyses/goto_check.cpp
```diff
@@ -1813,6 +1831,15 @@ void goto_checkt::goto_check(
         check(i.code.op0());
         // the return value invalidate any assertion
         invalidate(i.code.op0());
+
+        if(has_subexpr(i.code.op0(), [](const exprt &expr) {
+             return expr.id() == ID_r_ok || expr.id() == ID_w_ok;
+           }))
+        {
+          auto rw_ok_cond = rw_ok_check(i.code.op0());
```
smowton: `to_code_return(code).return_value()`

smowton (approved):

src/analyses/goto_check.cpp
tautschnig: Oops, thank you for keeping me from being lazy.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: d7dc902cc535ac0115b1bd765182f865cc15fa13).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107793523

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7ecf1f69e859f76b5aa5ae86c2ff2ba9ccae62ef).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107802062

PR #4495, opened 2019-04-08
Fix stubbed String constructor special case

antlechner:
This PR fixes a bug that was caused/uncovered by https://github.com/diffblue/cbmc/pull/4483.
Generally JBMC should only be run on methods that include Strings if a model for String is loaded. However, `--show-properties` should work even without the model. After the changes in https://github.com/diffblue/cbmc/pull/4483, in the special case where the Java method includes a call to a String constructor and no model for it was loaded, we didn't check if the object had a string type, and a pointer to an unsignedbv ended up in the pointer case of the object factory, causing an invariant violation. This PR fixes the conditions in the struct case of the object factory so it always checks if the type is a String type, not only when `skip_classid` is false.

- [x] Each commit message has a non-empty body, explaining why the change was made. [except for the new test, which has a description in its .desc file]


jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -794,77 +794,69 @@ void java_object_factoryt::gen_nondet_struct_init(
   // * Not if the object has already been initialised by our caller, in which
   //   case they will set `skip_classid`
   // * Not if we're re-initializing an existing object (i.e. update_in_place)
-
-  bool skip_special_string_fields = false;
-
-  if(!is_sub &&
-     !skip_classid &&
-     update_in_place != update_in_placet::MUST_UPDATE_IN_PLACE)
+  // * Always if it has a string type (special case).
```
allredj: :pick: I don't think `(special case)` is necessary here. Rather an explanation of *why* strings are a special case would be useful.

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -794,77 +794,69 @@ void java_object_factoryt::gen_nondet_struct_init(
...
+
+  const bool is_char_sequence =
+    java_string_library_preprocesst ::implements_java_char_sequence(
```
allredj: :pick: :question:  Is it customary to have a space before the double colon?

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -801,12 +801,11 @@ void java_object_factoryt::gen_nondet_struct_init(
       struct_type);
   const bool has_length_and_data =
     struct_type.has_component("length") && struct_type.has_component("data");
-  const bool skip_special_string_fields =
-    is_char_sequence && has_length_and_data;
+  const bool is_string_type = is_char_sequence && has_length_and_data;
```
allredj, line 806: :+1:  Thanks for changing that. The name was really bad.

allredj (approved):

jbmc/src/java_bytecode/java_object_factory.cpp
smowton: No :p

smowton (approved):

thomasspriggs (approved):

jbmc/src/java_bytecode/java_object_factory.cpp
antlechner: Huh, clang-format did this, not sure why! I'll fix it as CI is failing with a linting error anyway.

jbmc/src/java_bytecode/java_object_factory.cpp
antlechner: I added a few more lines, with a reference to the function that initialises string fields. Any further documentation should probably go to that function.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 9bc1a6ed47b57c5a8211d16e8d76367a2c89e16b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107467585

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: c6e55522b3e7b057716a74f59f8956d2b91d7323).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107467525

PR #4440, opened 2019-03-26
Clean up decision_proceduret

peterschrammel:
`decision_proceduret` is only used through `prop_convt`, because it is lacking essential conversion functions. Moving these functions into `decision_proceduret` will enable to replace `prop_convt` by `decision_proceduret` throughout the code base (which is much more intuitive and makes the code more readable).
`prop_convt` will then evolve into interfaces to support advanced features for incremental solving, etc.

- [X] Each commit message has a non-empty body, explaining why the change was made.
- [X] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [X] My PR is restricted to a single feature or bugfix.
- [X] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4361, opened 2019-03-10
Incremental unwinding of one specified loop

peterschrammel:
Incremental unwinding and assertion checking of a specified loop.

~Based on https://github.com/diffblue/cbmc/pull/4054~

- [X] Each commit message has a non-empty body, explaining why the change was made.
- [X] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [X] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [X] My PR is restricted to a single feature or bugfix.
- [X] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4497, opened 2019-04-08
decision_proceduret interface

kroening:
This re-establishes a simpler interface for solvers. In particular, the concept of 'literal' is not exposed, which is specific to DPLL-style SAT solvers.

The new interface is in particular a much better fit for SMT2 solvers, which right now need to emulate a CNF-style interface.

### DIFF
--- a/src/solvers/prop/decision_procedure.h
+++ b/src/solvers/prop/decision_procedure.h
@@ -34,7 +34,13 @@ public:
   void set_to_false(const exprt &expr);
 
   /// Convert a Boolean expression and return the corresponding literal
+  /// This will go away, use handle(expr) instead
   virtual literalt convert(const exprt &expr) = 0;
+
+  /// Generate a handle for an expression; this offers an efficient way
+  /// to refer to the expression in subsequent calls to \ref get or
+  /// \ref set_to
+  virtual exprt handle(const exprt &expr) = 0;
 
   /// Result of running the decision procedure
   enum class resultt
@@ -54,6 +60,7 @@ public:
 
   /// Return value of literal \p l from satisfying assignment.
   /// Return tvt::UNKNOWN if not available
+  /// This will go away, use get instead.
   virtual tvt l_get(literalt l) const = 0;
 
   /// Print satisfying assignment to \p out
--- a/src/solvers/prop/prop_conv.h
+++ b/src/solvers/prop/prop_conv.h
@@ -22,6 +22,13 @@ public:
 
   using decision_proceduret::operator();
 
+  /// returns a 'handle', which is an expression that
+  /// has the same value as the argument in any model
+  /// that is generated.
+  /// This may be the expression itself or a more compact
+  /// but solver-specific representation.
+  exprt handle(const exprt &expr) override;
+
   // incremental solving
   virtual void set_frozen(literalt a);
   virtual void set_frozen(const bvt &);
--- a/src/solvers/prop/prop_conv.cpp
+++ b/src/solvers/prop/prop_conv.cpp
@@ -7,7 +7,15 @@ Author: Daniel Kroening, kroening@kroening.com
 \*******************************************************************/
 
 #include "prop_conv.h"
+
+#include "literal_expr.h"
+
 #include <algorithm>
+
+exprt prop_convt::handle(const exprt &expr)
+{
+  return literal_exprt(convert(expr));
+}
 
 /// determine whether a variable is in the final conflict
 bool prop_convt::is_in_conflict(literalt) const


src/solvers/flattening/functions.h
```diff
@@ -37,7 +39,7 @@ class functionst
   }
 
 protected:
-  prop_convt &prop_conv;
+  decision_proceduret &decision_procedure;
```
tautschnig: It's strange to see a file live under `flattening` but then not actually use the propositional variant of a decision procedure.

kroening: It really is a 'lowering', and should move there.

kroening: Now moved.

peterschrammel (approved):
Lgtm as an intermediate step

peterschrammel:
> In particular, the concept of 'literal' is not exposed, which is specific to DPLL-style SAT solvers.

Literal is not the correct term in this context. It's a boolean variable that is equal to a predicate. I don't see anything specific to DPLL SAT here. The problem is that the interface exposes the implementation by using `literalt` instead of something more general.

src/solvers/prop/prop_conv.h
```diff
@@ -10,18 +10,16 @@ Author: Daniel Kroening, kroening@kroening.com
...
 #define CPROVER_SOLVERS_PROP_PROP_CONV_H
 
-#include "decision_procedure.h"
+#include "prop_decision_procedure.h"
 
 // API that provides a "handle" in the form of a literalt
 // for expressions.
 
-class prop_convt:public decision_proceduret
+class prop_convt : public prop_decision_proceduret
```
peterschrammel: That means `smt2_convt` still implements `prop_decision_proceduret`.

The more I think about this PR the more doubts I have.

The predicate handle interface is something useful to have also in `smt2`. It is required for some of our algorithms and hence it doesn't make sense not to have a common interface for SAT-based implementations and SMT-based ones. IMO `prop_decision_proceduret` should be generalised, renamed and moved up into `solvers/`.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 4e1b1433e38733e94c64bcaf125e2d3c415deea6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107502331

kroening:
The problem here is the confusion between an interface and a particular realisation. The `prop_decision_proceduret` interface anticipates a particular realisation, which means it's tedious to implement for the SMT solvers.

It's fully agreed that the 'handle' for a propositional expression is useful. However, that handle shouldn't be fixed to be an integer with a sign flag. None of the users of the current interface in fact care about that.

In particular, observe that `decision_proceduret` already lets you get a handle for some predicate P: simply add a constraint "bX <=> P". We could add a convenience function `handle(exprt)` for that; the SAT-basedd solvers could implement it in the same way as they do now.

kroening:
Furthermore, note that `handle(exprt)` would not need to be restricted to Boolean at all.

tautschnig (commented):
Stopped short of blocking this PR, but admittedly I'm not very happy with the current state.

1) Moving functions.{h,cpp} to lowering seems like the right thing to do and should just happen, maybe in a separate PR if there is further discussion on this one.
2) Let's please stop using (possibly ambiguous) abbreviations. If "prop" stands for "propositional" then let's use "propositional." If it means something else then it is even more urgent that it be spelled out.
3) The main concern: it seems ok that an `smt_convt` (also) is-a `prop_convt`. But the class hierarchy appears messed up to me. Why _is_ the conversion a decision procedure? And is there actually value in  the `prop_convt` interface at all?

kroening:
Ok, let's split:
1) #4511 now moves functionst.
2) The renaming from prop -> propositional is agreed, but really very orthogonal to this PR. Will make separate one.
3) The key point of this PR is that it enables smt2_conv to become a decision_proceduret! That will need a bit more work, since the users need to use the new 'handle' interface first.

tautschnig:
> The key point of this PR is that it enables smt2_conv to become a decision_proceduret! That will need a bit more work, since the users need to use the new 'handle' interface first.

Would you mind elaborating as to why `smt2_conv` becoming a `decision_proceduret` is the right thing to do? We already have `smt2_dect`, which is documented as "Decision procedure interface for various SMT 2.x solvers."

kroening:
Ah, ok, let me be more precise: right now, `smt2_dect` needs to implement various `literalt` related methods that are now required by `decision_proceduret`. These methods were really meant for CNF-style solvers, and thus, do not fit SMT2.

tautschnig:
> Ah, ok, let me be more precise: right now, `smt2_dect` needs to implement various `literalt` related methods that are now required by `decision_proceduret`. These methods were really meant for CNF-style solvers, and thus, do not fit SMT2.

Ok. But can we at this point achieve a clean separation between a decision procedure (which ought to implement the `decision_proceduret` interface) and the code converting between `exprt` and the decision procedure's representation (variants of `*_convt`)? It may make sense that a decision procedure has an `X_convt`, but I don't see how the is-a relationship of `class X_convt : public decision_proceduret` can be right.

peterschrammel:
1.-3. are an improvement, in particular, generalising the handle interface (which would have made the motivation for https://github.com/diffblue/cbmc/pull/4440 clearer).

The changes here are orthogonal to #4450, #4451 and #4054, though, which tackle a different problem: distinguishing the various (incremental) solver features so that we don't have a god-interface `prop_convt`anymore, but we can clearly carve out the set of features required by each `prop_convt`user.

@kroening, can you sketch a plan how you'd like to move this forward beyond 1.-3.?

kroening:
@tautschnig on _convt being a decision procedure: The problem here is that an algorithm, i.e., an implementation got used as an interface. The variants of *_convt do implement decision procedures, so should inherit from `decision_proceduret`. The key problem I have with them is that their purpose is to combine theory solvers, and that is very much obscured. They also mix one-shot encodings and incremental encodings.

tautschnig:
> @tautschnig on _convt being a decision procedure: The problem here is that an algorithm, i.e., an implementation got used as an interface.

Can we just fix this?

> The variants of *_convt do implement decision procedures, so should inherit from `decision_proceduret`.

Maybe the fix is just to rename `*_convt` (assuming "conv" stands for "conversion"?!)?

> The key problem I have with them is that their purpose is to combine theory solvers, and that is very much obscured. They also mix one-shot encodings and incremental encodings.

That last part I again struggle with: is "encoding" and "solver" the same thing? Maybe "encoding" is such an integral part of a solver that it is. In my opinion, we care about the process going from an `exprt` that includes symbols via X to a model providing concrete values for (some of) those symbols. So now we can consider X a decision procedure or a solver an be done with it. Or we further decompose X into a conversion step that translates the `exprt` into some other representation, then runs a decision procedure on that other representation to yield a model (in yet another representation) and then allows us to convert back from some representation to values that fit the `exprt`.

kroening:
Roadmap proposal:
1) We need to fix the interface that a user of a decision procedure is offered. I would propose that to offer
`exprt handle(const exprt &)`
No restriction on the type of the expression. Solvers that use CNF may then return a `literal_exprt` for Booleans. SMT-LIB solvers could return a symbol, or perhaps use a named term. BDDs could return a node number. A fallback is always to simply return the parameter, and say rely on hashing.
This could simply be added to the current `decision_proceduret`, along with corresponding variants of `set_assumptions` and `get_in_conflict`.
2) The users need to be moved to that new interface. That is mostly replacing 'literalt` by `exprt`.
3) Then we can deprecate the `literalt` variants.
4) smt2_conv and smt2_dec can then simply stop implementing these.
5) In paralell, we can clean up the solver implementation, with a view towards offering an interface for theory solvers. This would replace the current 'stack' of theory solvers.

kroening:
> > @tautschnig on _convt being a decision procedure: The problem here is that an algorithm, i.e., an implementation got used as an interface.
> 
> Can we just fix this?

That is what steps 1-3 in the proposed roadmap do.

> > The variants of *_convt do implement decision procedures, so should inherit from `decision_proceduret`.
> 
> Maybe the fix is just to rename `*_convt` (assuming "conv" stands for "conversion"?!)?

There is more work to do than just rename; the final name might be something well understood in the academic community, say "theory_combinationt". We could go for "solver_combinationt" to make it less scary.

> That last part I again struggle with: is "encoding" and "solver" the same thing? Maybe "encoding" is such an integral part of a solver that it is. In my opinion, we care about the process going from an `exprt` that includes symbols via X to a model providing concrete values for (some of) those symbols. So now we can consider X a decision procedure or a solver an be done with it. Or we further decompose X into a conversion step that translates the `exprt` into some other representation, then runs a decision procedure on that other representation to yield a model (in yet another representation) and then allows us to convert back from some representation to values that fit the `exprt`.

An encoding (or "reduction") is one way of realising a solver, but there are certainly others. The process to go from a formula to a model is commonly a "decision procedure". Implementing that via reduction to something simpler is very common these days, but not essence of it. The current `decision_proceduret` essentially anticipates that there will be CNF at the end, and that is overly specific.

peterschrammel:
> along with corresponding variants of set_assumptions and get_in_conflict.

I don't think `set_assumptions` and `get_in_conflict` should be part of `decision_proceduret`. Not every solver can provide that functionality and also not every decision procedure user needs it. A decision procedure user should select the minimal interface that it requires.

peterschrammel:
Also, https://github.com/diffblue/cbmc/pull/4450#issuecomment-477555518 must be accounted for.

peterschrammel:
The lower bar for these refactorings is to make https://github.com/diffblue/cbmc/pull/4361 pass CI with MiniSAT. I'd call them a success if it also passes with Z3.

tautschnig:
In addition to what @peterschrammel said:

> [...] the final name [...]

I would just ask for a design involving that "final" name rather than intermediate steps that will come and go.

kroening:
@peterschrammel The key problem with #4361 is freezing.
I think it's a dead end to impose this on on the user of the solver, and it needs to go away. SMT-LIB agrees.
I think I have an idea how to make this happen, but this should really not be a stoper for us to have the right interfaces.

kroening:
@tautschnig I am not to concerned about the final name of the _implementation_ of the combination solver. Users won't see it; they see the interface, and they see an instance that comes with a particular set of theories.
The problem right now is that we are exposing users to implementation, which means that it's not stable, and it's expensive to improve the implementation.

peterschrammel:
> I think I have an idea how to make this happen, but this should really not be a stoper for us to have the right interfaces.

Then this needs to be solved with highest priority - otherwise it is a stopper in doing this refactoring if we end up with the "right interfaces" that are not usable.

kroening:
Item 1 now there!
Given that there are a number, I'll do a series of PRs for item 2.

src/solvers/prop/prop_conv.h
```diff
@@ -22,6 +22,9 @@ class prop_convt:public decision_proceduret
 
   using decision_proceduret::operator();
 
+  // handle
```
tautschnig: Maybe the notes from the roadmap (explaining that this could be a `literal_exprt` or a symbol or a BDD node or ...) can be copied in here instead of this comment?

src/solvers/prop/decision_procedure.h
```diff
@@ -54,6 +60,7 @@ class decision_proceduret
 
   /// Return value of literal \p l from satisfying assignment.
   /// Return tvt::UNKNOWN if not available
+  /// This will go away, use get instead.
```
tautschnig, line 63: Mark as deprecated?

src/solvers/prop/decision_procedure.h
```diff
@@ -34,8 +34,14 @@ class decision_proceduret
   void set_to_false(const exprt &expr);
 
   /// Convert a Boolean expression and return the corresponding literal
+  /// This will go away, use handle(expr) instead
```
tautschnig, line 37: Mark as deprecated? The sooner we mark it as deprecated, the earlier we can remove it ;-)

tautschnig (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: b20067aef6ce22f09be479f993fee4198ca8b450).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108055107

src/solvers/prop/prop_conv.h
peterschrammel: Shouldn't that be abstract here and implemented in `prop_conv_solvert`? Only `prop_conv_solvert` would use `literal_exprt`, other implementations of `prop_convt` might not?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9c3774b3a0b91fc6ae031f42ceb222ff16d29040).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108163619

src/solvers/prop/decision_procedure.h
kroening: Will remove the usage first (not a lot).

PR #4499, opened 2019-04-08
Fix Java exception handlers

NathanJPhillips:
Fix Java exception handlers when multiple exception table rows begin at the same offset. Previously this could lead to throws within the catch block (e.g. a straightforward log-and-rethrow) being directed to the same catch block, introducing a spurious loop.

This is not a perfect solution, as documented in the comments, but is an improvement on the previous version, which was actually broken as demonstrated by the added tests. If it wasn't for the unwind limit the test case shown would cause Symex never to complete. An ideal solution would probably be to get rid of push-catch and pop-catch altogether and carry a GOTO exception table around with the GOTO program until exceptions are removed.

- N/A Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.


smowton:
@NathanJPhillips Added the first paragraph to the description

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,80 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
-
-      c = code_blockt({catch_push, c});
-    }
-    else
+    if (exception.start_pc == cur_pc)
     {
-      // advance
-      ++pos;
+      std::size_t end_pc = exception.end_pc;
+      exceptionst &exceptions = exceptions_by_end.count(end_pc) == 0
```
smowton: Replace this whole block by `exceptions[exception.end_pc].push_back(exception);`

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,80 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
+  // a) the next code in the bytecode and therefore its start address matches
+  // the exclusive end position of the try block (probably a valid assumption
+  // for non-obfuscated bytecode generated by most compilers)
+  // b) the successor of a previous instruction and therefore in the working
+  // set (not a valid assumption when there aren't any throwing statements in
+  // the try block)
+  // We are therefore going to find all exception handlers that end at our
+  // earliest successor
+  if(working_set.empty())
+    // There was no throw in the try block
```
smowton: Add braces round these now-multiline blocks

jbmc/regression/jbmc/exceptions28/test.desc
```diff
@@ -0,0 +1,11 @@
...
+test.class
+--unwind 10
+^\[java::test.main:\(\[Ljava/lang/String;\)V\.1\] line 7 no uncaught exception: FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+This test verifies that we can handle the exception table not being ordered
```
smowton, line 11: Paste the actual exception table here as an example

smowton (approved):
Some cleanups otherwise lgtm

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: b76857c411179cfd7f126821f55954d8f93ccac9).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107511332

NathanJPhillips:
> This PR failed Diffblue compatibility checks (cbmc commit: [b76857c](https://github.com/diffblue/cbmc/commit/b76857c411179cfd7f126821f55954d8f93ccac9)).

This was confirmed to be an erroneous warning.

jbmc/regression/jbmc/exceptions28/test.desc
```diff
@@ -0,0 +1,17 @@
...
```
thk123, line 11: :question: What does "not being ordered" mean?

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,78 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
-    // reset
-    end_pc = 0;
-    exception_ids.clear();
-    handler_labels.clear();
+    if(exception.start_pc == cur_pc)
+      exceptions_by_end[exception.end_pc].push_back(exception);
+  }
+  for(const auto &exceptions : exceptions_by_end)
+  {
+    code_push_catcht catch_push;
```
thk123, line 2360: :pick: Seems a bit back to front to make the code before making the list, would be nicer to make the execption list, and construct the `code_push_catcht` from it, but I see it has no constructor from exception list so out of scope of this PR to fix it

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,78 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
+    code_push_catcht::exception_listt &exception_list =
+      catch_push.exception_list();
+    for(const java_bytecode_convert_methodt::methodt::exceptiont &exception
+      : exceptions.second)
+    {
+      exception_list.emplace_back(
+        exception.catch_type.get_identifier(),
+        // Record the exception handler in the CATCH-PUSH instruction by
```
thk123, line 2369: :pick: Initially found the comment inside the brackets a little disorientating. You could introduce a local with the result of `label` with the comment above, if you agree.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,78 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
-
-  // for each try-catch add a CATCH-PUSH instruction
-  // each CATCH-PUSH records a list of all the handler labels
-  // together with a list of all the exception ids
-
-  // be aware of different try-catch blocks with the same starting pc
-  method_offsett pos = 0;
-  method_offsett end_pc = 0;
-  while(pos < method.exception_table.size())
+  // For each try-catch add a CATCH-PUSH instruction
```
thk123: :bulb: Big block of code with a  comment, perhaps pull out into a function?

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,78 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
+  // for non-obfuscated bytecode generated by most compilers)
+  // b) the successor of a previous instruction and therefore in the working
+  // set (not a valid assumption when there aren't any throwing statements in
+  // the try block)
+  // We are therefore going to find all exception handlers that end at our
+  // earliest successor
+  if(working_set.empty())
+  {
+    // There was no throw in the try block
+    // TODO: This will result in non-balanced PUSH and POPs
```
thk123: :question: This seems a little scary... is this really not a regression? A knownbug test showing this limitation would be greatly appreciated.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,78 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
+    return code;
+  }
+  const method_offsett &working_set_elt = *working_set.begin();
+  if(cur_pc >= working_set_elt)
+  {
+    // This just fails if we have a back-edge here
+    // TODO: This will result in non-balanced PUSH and POPs
+    return code;
```
thk123: :question: Ditto here. Also might be a good idea to print some warning in this eventuallity?

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
thk123: :question: can we print a warning here

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,78 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
-          first_handler = true;
-        // remember the beginning of the try-catch so that we don't add
-        // another CATCH-POP for the same try-catch
-        start_pc = exception_row.start_pc;
-        // add CATCH_POP instruction
-        code_pop_catcht catch_pop;
-        c = code_blockt({c, catch_pop});
-      }
-    }
+    // Check if working_set_elt is the end of a try block
```
thk123: :question: what is `elt`? Maybe expand

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,78 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
+  }
+  const method_offsett &working_set_elt = *working_set.begin();
+  if(cur_pc >= working_set_elt)
+  {
+    // This just fails if we have a back-edge here
+    // TODO: This will result in non-balanced PUSH and POPs
+    return code;
+  }
+  // Count the distinct handler start positions that end at this location
+  std::set<std::size_t> start_positions;
```
thk123: Since you just care about the size, why not `std::count_if`?

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,78 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
-      }
-    }
+    // Check if working_set_elt is the end of a try block
+    if(working_set_elt == exception_row.end_pc)
+      start_positions.insert(exception_row.start_pc);
   }
-  return c;
+  for(std::size_t handler = 0; handler < start_positions.size(); ++handler)
+    // Append a CATCH-POP instruction before the end of the try block
+    code = code_blockt({ code, code_pop_catcht() });
```
thk123: :bulb: That's a lot of nested blocks, would it make sense to produce `code; pop; pop; pop` rather than `{{{code; pop} pop} pop}`?

E.g. by:

```C++
auto pop_catchs = std::vector<codet>{start_positions.size(), code_pop_catcht()};
code = code_blockt{code, pop_catchs)
```

thk123 (commented):
I have some questions, I'm also gonna try it on TG just to be safe about the limitations

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
NathanJPhillips: This is the current behaviour already. I suggested putting an assertion in but @smowton thought that this actually happens quite a lot and is relatively harmless. A warning seems like a good idea.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
NathanJPhillips: Managed to do this for both pushes and pops, thank you.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
NathanJPhillips: Comment added.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
NathanJPhillips: Also it turns out my thinking was slightly wrong. We can't tell at this point whether balancing POPs will be introduced. I've updated the code and clarified the comments to make it clearer what's going on.

jbmc/regression/jbmc/exceptions28/test.desc
NathanJPhillips: Added extra comment below example.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,91 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
+      if(*next_opcode_it == exception_row.end_pc)
+        start_positions.insert(exception_row.start_pc);
     }
+    for(std::size_t handler = 0; handler < start_positions.size(); ++handler)
+      // Append a CATCH-POP instruction before the end of the block
+      result.push_back(code_pop_catcht());
   }
-  return c;
+
+  if (result.size() != 1)
```
thk123: Seems a little over the top just to avoid one redundant block, but as you like

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,91 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
-    }
-
-    // reset
-    end_pc = 0;
-    exception_ids.clear();
-    handler_labels.clear();
+    if(exception.start_pc == cur_pc)
+      exceptions_by_end[exception.end_pc].push_back(exception);
+  }
+  // Loop over each block with a distinct end position
```
thk123: personally think this comment doesn't add clarity.

thk123 (approved):
Thanks :+1: I ran the tests locally (and they passed) but messed up running on CI, feel free to merge without waiting if you want, but I'll kick of CI again if you don't mind waiting till tomorrow

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -1806,7 +1806,7 @@ code_blockt java_bytecode_convert_methodt::convert_instructions(
     // (e.g. due to exceptional control flow)
     if(!start_new_block)
       start_new_block=address_pair.second.predecessors.size()>1;
-    // Start a new lexical block if we've just entered a try block
+    // Start a new lexical block if we've just entered a handler block
```
smowton: `try` was accurate, I'd interpret `handler` to mean a catch block, and the exception table's `start_pc` entry refers to the `try`

jbmc/regression/jbmc/exceptions28/test.desc
```diff
@@ -0,0 +1,18 @@
...
+test.main gives the following exception table:
+ from    to  target type
+    0     6    15   Class MyException
+    7    15    15   Class MyException
+    0     6    18   Class java/lang/Exception
+    7    15    18   Class java/lang/Exception
+This is not ordered in the sense that the two handlers for the block 0-6 are not next to each other.
```
peterschrammel: Missing newline

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2335,105 +2335,90 @@ codet &java_bytecode_convert_methodt::do_exception_handling(
...
+    working_set.end(),
+    [cur_pc](method_offsett offset) { return offset > cur_pc; });
+  if(next_opcode_it != working_set.end())
   {
-    // add the CATCH-POP before the end of the try block
-    if(
-      cur_pc < exception_row.end_pc && !working_set.empty() &&
-      *working_set.begin() == exception_row.end_pc)
+    // Count the distinct start positions of handlers that end at this location
+    std::set<std::size_t> start_positions;  // Use a set to deduplicate
```
peterschrammel, line 2401: Is the ordered set needed here?

NathanJPhillips: `unordered_set` wouldn't perform well when storing a small set of integers without good spread like these.

jbmc/regression/jbmc/exceptions28/test.desc
NathanJPhillips: :+1:

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
NathanJPhillips: I thought using the word "try" would imply the whole of a Java `try` block to someone who didn't know how exception tables worked.

smowton: Ah, true enough -- in that case I think you should say it in full, "if we've just entered a block in which exceptions are handled. This is usually the start of a `try` block, but a single `try` can be represented as multiple non-contiguous blocks in the exception table."

thk123:
TG has revealed a potential issue with either synchronized blocks or nested exception catch blocks and this PR. Nathan is investigating at the moment.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
NathanJPhillips: Unfortunately this was actually wrong. Reverted and gone with something like your initial idea.

NathanJPhillips:
A change I made in response to a review comment introduced a bug that wasn't caught by current CBMC tests. I have added a test to catch this in a separate commit and revised the change to overcome the problem.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 21bfdba8e1f728899030c35f48c2b2f76a2f8da4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107800935

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 2cf31a9f6b096206732684c72a2697c3f7fe791b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107823835

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 79d5d612e66f85bbfef75ae5121de09e68bf1e11).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107834277

PR #4502, opened 2019-04-09
Ensure Java stubbing names parameters [blocks: #4485]

tautschnig:
For functions with a body we require that parameters are named so that
we can generate assignments to them (even when the body may not read the
value). Also guard against future regressions by sanity checking
parameter names for any converted function.

### DIFF
--- a/jbmc/src/java_bytecode/simple_method_stubbing.cpp
+++ b/jbmc/src/java_bytecode/simple_method_stubbing.cpp
@@ -148,11 +148,28 @@ void java_simple_method_stubst::create_method_stub_at(
 void java_simple_method_stubst::create_method_stub(symbolt &symbol)
 {
   code_blockt new_instructions;
-  const java_method_typet &required_type = to_java_method_type(symbol.type);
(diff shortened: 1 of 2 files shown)


peterschrammel (approved):

smowton:
Got to this late, but this should be done when `assign_parameter_names` is called when a stub function is created in `java_bytecode_convert_method.cpp`, do you have any idea why that wasn't happening in your case (or should that call be removed now?) Could you submit a followup with whatever test case motivated this?

tautschnig:
It's jbmc/provide_object_implementation/ that fails without the patch (but the additional sanity check kept in). `assign_parameter_names` is never called for the symbol at fault (`java::java.lang.Class.forName:(Ljava/lang/String;)Ljava/lang/Class;`), which gets added into the symbol table in `typecheck_expr_symbol`. Note that this test uses `--no-lazy-methods`, which may be important here (I don't think the method at fault is ever invoked, and `assign_parameter_names` is AFAIK only called for invoked methods).

So I accept that maybe some other place is the right one to fix this, but I don't really have a sufficiently good understanding of the Java front-end to make that change.

PR #4503, opened 2019-04-09
Add tests for --java-assume-inputs-non-null

allredj:
I noticed that `--java-assume-inputs-non-null` was untested in JBMC. Here are some tests that verify that the option actually does something for Strings, arrays of Strings, and user-defined objects.

I also discovered that class fields were not affected by this option (as they are presumably considered to be fields of `this`, and the option does not extend to fields of inputs). Whether or not this was intentional, that behaviour is now documented by tests.


jbmc/regression/jbmc/assume-inputs-non-null/My.java
```diff
@@ -0,0 +1,41 @@
+class My {
+
+  String field;
+
+  public static void stringArrayArg(String[] arg) {
+    if (arg == null) {
+      assert false;
```
thk123, line 7: :bulb: It might make the tests a little easier to comprehend if you _only_ put the `assert false` in the branches that shouldn't be covered. 
Indeed could even be written as non-coverage style:

```java
assert(arg != null)
```

Then you only need to check for verification successful.

thk123 (approved):

jbmc/regression/jbmc/assume-inputs-non-null/My.java
allredj: I have a slight preference for the branching solution (using `if`s) as I would like to keep both assertions (to be sure that a trace exists at all), and I don't want to make too many assumptions as to whether the `assert` acts as a guard or not.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d0f608a12b92057756a870d66b2d57c21e0ae9cb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107658751

jbmc/regression/jbmc/assume-inputs-non-null/class_assume.desc
```diff
@@ -0,0 +1,16 @@
+CORE
+My.class
+--function My.classArg --java-assume-inputs-non-null
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+[java::My.classArg:(LOther;)V.assertion.1].*SUCCESS
```
romainbrenguier: `[ ]` is for character sets

romainbrenguier (changes requested):

jbmc/regression/jbmc/assume-inputs-non-null/class_assume.desc
allredj: Crap. Now escaped `[` `]` `(` `)`.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 5ee583ff0c2d5f3b697e3e52e8565ece1bc45168).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107788971

romainbrenguier (approved):

PR #4505, opened 2019-04-09
Use the sharing map for the renaming level maps

danpoe:
Use the sharing map for the renaming level maps `current_names` in `symex_renaming_levelt`. The code in this PR has a few issues which I'll fix once we've decided whether to merge it.

### DIFF
--- a/src/goto-symex/auto_objects.cpp
+++ b/src/goto-symex/auto_objects.cpp
@@ -90,9 +90,8 @@ void goto_symext::trigger_auto_object(const exprt &expr, statet &state)
         if(has_prefix(id2string(symbol.base_name), "auto_object"))
         {
           // done already?
-          if(
-            state.get_level2().current_names.find(ssa_expr.get_identifier()) ==
-            state.get_level2().current_names.end())
+          if(!state.get_level2().current_names.has_key(
+               ssa_expr.get_identifier()))
           {
             initialize_auto_object(e, state);
           }
--- a/src/goto-symex/goto_state.cpp
+++ b/src/goto-symex/goto_state.cpp
@@ -31,13 +31,22 @@ std::size_t goto_statet::increase_generation(
   const ssa_exprt &lhs,
   std::function<std::size_t(const irep_idt &)> fresh_l2_name_provider)
 {
-  auto current_emplace_res =
-    level2.current_names.emplace(l1_identifier, std::make_pair(lhs, 0));
+  std::size_t n = fresh_l2_name_provider(l1_identifier);
 
-  current_emplace_res.first->second.second =
-    fresh_l2_name_provider(l1_identifier);
+  const auto r_opt = level2.current_names.find(l1_identifier);
 
-  return current_emplace_res.first->second.second;
+  if(!r_opt)
+  {
+    level2.current_names.insert(l1_identifier, std::make_pair(lhs, n));
+  }
+  else
+  {
+    std::pair<ssa_exprt, unsigned> copy = r_opt->get();
+    copy.second = n;
+    level2.current_names.replace(l1_identifier, std::move(copy));
+  }
+
+  return n;
 }
 
 /// Given a condition that must hold on this path, propagate as much knowledge
--- a/src/goto-symex/goto_symex_state.cpp
+++ b/src/goto-symex/goto_symex_state.cpp
@@ -795,16 +795,21 @@ ssa_exprt goto_symex_statet::add_object(
   const irep_idt l0_name = ssa.get_identifier();
   const std::size_t l1_index = index_generator(l0_name);
 
-  // save old L1 name, if any
-  auto existing_or_new_entry = level1.current_names.emplace(
-    std::piecewise_construct,
-    std::forward_as_tuple(l0_name),
-    std::forward_as_tuple(ssa, l1_index));
+  const auto r_opt = level1.current_names.find(l0_name);
 
-  if(!existing_or_new_entry.second)
+  if(!r_opt)
   {
-    frame.old_level1.emplace(l0_name, existing_or_new_entry.first->second);
-    existing_or_new_entry.first->second = std::make_pair(ssa, l1_index);
+    level1.current_names.insert(l0_name, std::make_pair(ssa, l1_index));
+  }
+  else
+  {
+    // save old L1 name
+    if(!frame.old_level1.has_key(l0_name))
+    {
+      frame.old_level1.insert(l0_name, r_opt->get());
+    }
+
+    level1.current_names.replace(l0_name, std::make_pair(ssa, l1_index));
   }
 
   ssa = rename_ssa<L1>(std::move(ssa), ns).get();
--- a/src/goto-symex/goto_symex_state.h
+++ b/src/goto-symex/goto_symex_state.h
@@ -226,15 +226,15 @@ public:
   }
 
   /// Drops an L1 name from the local L2 map
-  void drop_l1_name(symex_renaming_levelt::current_namest::const_iterator it)
+  void drop_existing_l1_name(const irep_idt &l1_identifier)
   {
-    level2.current_names.erase(it);
+    level2.current_names.erase(l1_identifier);
   }
 
   /// Drops an L1 name from the local L2 map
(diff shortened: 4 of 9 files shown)


danpoe:
Here are some results for the runtime of symex only, with `--unwind 1`, on several methods from Apache Tika on which symex takes between 5 and 15 sec

Rplot
Overall with sharing it's 3% faster.

danpoe:
@smowton, @romainbrenguier Would you be able to try this out on your benchmarks too?

src/goto-symex/auto_objects.cpp
```diff
@@ -91,9 +91,7 @@ void goto_symext::trigger_auto_object(const exprt &expr, statet &state)
       if(has_prefix(id2string(symbol.base_name), "auto_object"))
       {
         // done already?
-        if(
-          state.get_level2().current_names.find(ssa_expr.get_identifier()) ==
-          state.get_level2().current_names.end())
+        if(!state.get_level2().current_names.has_key(ssa_expr.get_identifier()))
```
tautschnig: ~Formatting looks messed up?~ (edit: just a GitHub artifact)

src/goto-symex/goto_state.cpp
```diff
@@ -28,13 +28,22 @@ std::size_t goto_statet::increase_generation(
...
-  return current_emplace_res.first->second.second;
+  if(!r_opt)
+  {
+    level2.current_names.insert(l1_identifier, std::make_pair(lhs, n));
+  }
+  else
+  {
+    std::pair<ssa_exprt, unsigned> copy = r_opt->get();
+    copy.second = n;
+    level2.current_names.replace(l1_identifier, copy);
```
tautschnig: `std::move(copy)`

src/goto-symex/goto_symex_state.cpp
```diff
@@ -776,16 +776,20 @@ ssa_exprt goto_symex_statet::add_object(
   const irep_idt l0_name = ssa.get_identifier();
   const std::size_t l1_index = index_generator(l0_name);
 
-  // save old L1 name, if any
```
tautschnig, line 798: I'm not sure I appreciate that this comment has been removed.

src/goto-symex/goto_symex_state.h
```diff
@@ -223,15 +223,16 @@ class goto_symex_statet final : public goto_statet
   }
 
   /// Drops an L1 name from the local L2 map
-  void drop_l1_name(symex_renaming_levelt::current_namest::const_iterator it)
+  void drop_existing_l1_name(const irep_idt &l1_identifier)
   {
-    level2.current_names.erase(it);
+    level2.current_names.erase(l1_identifier);
   }
```
tautschnig, line 232: I think this method should either be removed or else be used from `drop_l1_name`.

src/util/ssa_expr.h
```diff
@@ -16,6 +16,8 @@ Author: Daniel Kroening, kroening@kroening.com
 class ssa_exprt:public symbol_exprt
 {
 public:
+  ssa_exprt() : symbol_exprt("", typet()) {}
```
tautschnig: This is pretty much a show stopper: why is this needed? Presumably a side effect of how objects are created in a sharing map, but can the sharing map please be fixed instead?

tautschnig (commented):
Comments (including one blocking one) below; I've started performance experiments, will report on their results later on today.

tautschnig:
The results for my recent favourite ReachSafety-ECA are: 160 points scored with this patch, 156 points without this patch. Symex steps per second increase from 2978.24 to 2991.21. So that's overall an improvement, albeit not a groundbreaking one.

romainbrenguier:
On my benchmarks the total time went from 106.85 to 99.31 chart (2)

tautschnig:
@danpoe @romainbrenguier Could you rerun your benchmarking now that field-sensitivity is merged? It's possible that the maps now actually get larger, and thus such changes may be more impactful.

src/goto-symex/goto_symex_state.h
```diff
@@ -223,15 +223,16 @@ class goto_symex_statet final : public goto_statet
...
 
   /// Drops an L1 name from the local L2 map
   void drop_l1_name(const irep_idt &l1_identifier)
   {
-    level2.current_names.erase(l1_identifier);
+    if(level2.current_names.has_key(l1_identifier))
```
smowton: Now have `erase_if_exists`

src/goto-symex/renaming_level.cpp
```diff
@@ -75,18 +78,18 @@ operator()(renamedt<ssa_exprt, L1> l1_expr) const
 
 void symex_level1t::restore_from(const current_namest &other)
 {
-  auto it = current_names.begin();
-  for(const auto &pair : other)
+  current_namest::delta_viewt delta_view;
+  other.get_delta_view(current_names, delta_view, false);
```
smowton, line 82: `delta_view` overestimates the keys that may differ, right? If so couldn't the `replace` below replace for no good reason?

smowton (approved):
Tested this with a long-running Webgoat analysis -- found before the changes it took ~200s, afterwards ~160s.

danpoe:
Here are some results for the runtime of symex only, with `--unwind 1`, on several methods from Apache Tika, after field-sensitivity was merged:

Rplot

Overall with sharing it's 3% faster.

Btw, I've also compared old symex vs. field sensitive symex and overall field-sensitive symex was 6.3X faster on the benchmarks.

romainbrenguier:
For my benchmarks it didn't seem to have a positive effect:
execution time on develop and sharing renaming (1)
Total time went from 73.86s to 77.218s

smowton:
Now that field-sensitivity is in this makes even more of a difference, mostly in skipping the large number of new keys in the renaming map via a sparse `delta_viewt`. I suspect the benchmarks above aren't really seeing the benefit as they're micro-benchmarks, whereas these effects only really show up once the renaming map gets large.

src/util/ssa_expr.h
tautschnig: Maybe somewhere between @danpoe and @smowton: can you look into this and see whether we can get away without it?

smowton: Sure, I'll take a look

danpoe: I was meaning to fix this once we have decided whether to merge this PR. Currently the sharing map creates a dummy object (using the default constructor of the stored value type) when for a `delta_view_itemt` the key is only in one of the maps. We can probably use an `optionalt` instead.

danpoe:
@smowton Would you be able to re-run the benchmarks on the Webgoat analysis now that field-sensitivity has been merged?

src/util/ssa_expr.h
smowton: https://github.com/diffblue/cbmc/pull/4612 @danpoe

smowton:
Benchmark results for a long-running Webgoat analysis:

Develop: 77s
#4464: 57s
#4464 + #4505: 26s

src/util/ssa_expr.h
tautschnig: #4612 is merged.

danpoe:
Thanks everyone for the benchmarking and comments. So in summary on @romainbrenguier's benchmarks it's about 4% slower, on @smowton's benchmark it's about twice as fast, and on my benchmarks it's about 3% faster. So I'd be inclined to merge it once comments have been addressed and CI passes.

src/goto-symex/goto_symex_state.cpp
danpoe: Put back in

src/goto-symex/goto_symex_state.h
danpoe: The other version is now using `erase_if_exists()` as suggested by @smowton

src/goto-symex/renaming_level.cpp
danpoe: Fixed

smowton:
CI failure is caused by the rebase, as `other_m` is now a pointer

PR #4506, opened 2019-04-09
Remove unused class_identifier parameter from Java object factory

antlechner:

### DIFF
--- a/jbmc/src/java_bytecode/java_object_factory.cpp
+++ b/jbmc/src/java_bytecode/java_object_factory.cpp
@@ -826,9 +826,10 @@ void java_object_factoryt::gen_nondet_struct_init(
     // This code mirrors the `remove_java_new` pass:
     auto initial_object = zero_initializer(expr.type(), source_locationt(), ns);
     CHECK_RETURN(initial_object.has_value());
-    const irep_idt qualified_clsid = "java::" + id2string(struct_tag);
     set_class_identifier(
-      to_struct_expr(*initial_object), ns, struct_tag_typet(qualified_clsid));
+      to_struct_expr(*initial_object),
+      ns,
+      struct_tag_typet("java::" + id2string(struct_tag)));
 
     // If the initialised type is a special-cased String type (one with length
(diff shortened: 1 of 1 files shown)


jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -826,9 +826,10 @@ void java_object_factoryt::gen_nondet_struct_init(
     // This code mirrors the `remove_java_new` pass:
     auto initial_object = zero_initializer(expr.type(), source_locationt(), ns);
     CHECK_RETURN(initial_object.has_value());
-    const irep_idt qualified_clsid = "java::" + id2string(struct_tag);
     set_class_identifier(
-      to_struct_expr(*initial_object), ns, struct_tag_typet(qualified_clsid));
+      to_struct_expr(*initial_object),
+      ns,
```
thomasspriggs, line 831: :bulb: `set_class_identifier` doesn't actually use `ns`. So this parameter could probably be removed, however that is probably a change for another PR.

thomasspriggs (approved):
Neat.

jbmc/src/java_bytecode/java_object_factory.cpp
antlechner: It has a call to `ns.follow`. I still find it a bit strange that `follow` is defined only for namespaces and not for symbol tables, but that would definitely be a change for another (much bigger) PR. :slightly_smiling_face:

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: b97b80aab54d9b67e55e3a6e1f4f5b8b1466e5a9).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107679876

jbmc/src/java_bytecode/java_object_factory.cpp
thomasspriggs: Ah, my mistake.

smowton (approved):

peterschrammel (approved):

PR #4507, opened 2019-04-10
Remove message_handler field of ui_message_handler

romainbrenguier:
Having ui_message_handlert both inherit from message_handler and have a
message handler pointer field is confusing.
We remove the field and keep inheritance. Because one of the constructor
becomes too complicated we make it into a static make function.
This in particular solves the problem of verbosity not being correct
because it is inherited and not fetched from the pointer field.


TGWDB:
Closing this PR since it has been "do not merge" and "do not review" without progress for two years. That said, please reopen if this is erroneous and feel free to request new reviews from codeowners.

PR #4509, opened 2019-04-10
Add a version of get_or_create_string_literal_symbol that takes a string parameter

antlechner:
No functional change, only refactoring.

### DIFF
--- a/jbmc/src/java_bytecode/java_string_literals.h
+++ b/jbmc/src/java_bytecode/java_string_literals.h
@@ -12,6 +12,14 @@ Author: Chris Smowton, chris.smowton@diffblue.com
 #include <util/symbol_table.h>
 #include <util/std_expr.h>
 
+/// Creates or gets an existing constant global symbol for a given string
+/// literal.
+/// \param string_expr: string literal expression to convert
+/// \param symbol_table: global symbol table. If not already present, constant
+///   global symbols will be added.
+/// \param string_refinement_enabled: if true, string refinement's string data
+///   structure will also be initialised and added to the symbol table.
+/// \return a symbol_expr corresponding to the new or existing literal symbol.
 symbol_exprt get_or_create_string_literal_symbol(
   const exprt &string_expr,
   symbol_table_baset &symbol_table,
--- a/jbmc/src/java_bytecode/java_string_literals.h
+++ b/jbmc/src/java_bytecode/java_string_literals.h
@@ -25,4 +25,12 @@ symbol_exprt get_or_create_string_literal_symbol(
   symbol_table_baset &symbol_table,
   bool string_refinement_enabled);
 
+/// Same as
+/// get_or_create_string_literal_symbol(const exprt&, symbol_table_baset&, bool)
+/// except it takes an id/string parameter rather than a string literal exprt.
+symbol_exprt get_or_create_string_literal_symbol(
+  const irep_idt &string_value,
+  symbol_table_baset &symbol_table,
+  bool string_refinement_enabled);
+
 #endif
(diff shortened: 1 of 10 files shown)


jbmc/src/java_bytecode/java_string_literals.h
```diff
@@ -12,9 +12,25 @@ Author: Chris Smowton, chris.smowton@diffblue.com
...
+/// Creates or gets an existing constant global symbol for a given string
+/// literal.
+/// \param string_expr: string literal expression to convert
+/// \param symbol_table: global symbol table. If not already present, constant
+///   global symbols will be added.
+/// \param string_refinement_enabled: if true, string refinement's string data
+///   structure will also be initialised and added to the symbol table.
+/// \return a symbol_expr corresponding to the new or existing literal symbol.
 symbol_exprt get_or_create_string_literal_symbol(
   const exprt &string_expr,
```
smowton: Are there any users of this overload left?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 78d181a3c313124a4add32f4690ddd80a34f317b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/107812797

jbmc/src/java_bytecode/java_string_literals.h
antlechner: Only one:
https://github.com/diffblue/cbmc/blob/7dd720b5db62d05bf814b4428c90dcca816d425a/jbmc/src/java_bytecode/java_bytecode_language.cpp#L424

smowton: In that case it would be a very welcome addition to this PR to make `java_string_literal_exprt` rather than directly twiddling operands / named subexpressions -- fair enough though if you think that's out of scope.

antlechner: It looks like there aren't any `java[\w_]+exprt` types yet. Should such a type go into `jbmc/src/java_bytecode/java_types.h` or a new file, e.g. something like `jbmc/src/java_bytecode/java_expressions.h` or similar?
Or does it make more sense to create a new folder `jbmc/src/util`? In that case I would definitely make a new PR for that.

romainbrenguier: Does symex have to handle these expressions in a specific way? If it's the case then they shouldn't be defined in a `jbmc` subdirectory. Otherwise `java_expr.h` would be clearer for showing the class inherits from `exprt` and I'd prefer one class per file, so `java_string_literal_expr.h` would be better.

tautschnig: Hearing about language-specific expressions makes me panic a bit. You will need to make sure all of these are gone by the time the code leaves the Java front-end.

smowton: They are; these are just descriptions of the exprts produced by the bytecode parser. The only other example I'm aware of now, and which I suggest we emulate, is `fieldref_exprt`, which is defined in `java_bytecode_parse_tree.h`.

smowton: (that expression is also local to the parser and becomes `member_exprt` later once we have more information about what it's referring to)

antlechner: I added the new class as you suggested, in a new commit.

smowton (approved):
Great, thanks a lot for this!

romainbrenguier (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9e499d53dbe32583b5c19f4acb90172e1fbe6e81).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108126079

PR #4510, opened 2019-04-10
prop_conv_solvert: introduce scopes

kroening:
This is the biggest change to the solver interface since it got added to the codebase.

Currently, bound identifiers (exists, forall, let) must be unique; the solver looses soundness if that assumption is violated.

This change introduces scopes, i.e., the uniqueness requirement is dropped. Nested scoping works as expected. This matches what SMT-LIB does, i.e., the renaming done in the SMT-LIB front-end can be dropped.


peterschrammel (commented):
Looks OK. CI is failing, though.

peterschrammel:
Tests for prop_conv and smt2 would be required.

src/solvers/flattening/boolbv.h
```diff
@@ -119,6 +119,10 @@ class boolbvt:public arrayst
   typedef std::unordered_map<const exprt, bvt, irep_hash> bv_cachet;
   bv_cachet bv_cache;
 
+  // record the subexpressions that use the scoped variable,
+  // in order to clear them from the cache after we are done
+  std::vector<bv_cachet::iterator> bv_cache_iterators;
```
tautschnig: Iterators into an unordered map aren't stable. Is there a guarantee that we will only ever put iterators in here after the source map isn't modified anymore?

src/solvers/flattening/boolbv_let.cpp
```diff
@@ -15,16 +15,30 @@ bvt boolbvt::convert_let(const let_exprt &expr)
...
+  bool previous_record_cache_iterators = record_cache_iterators;
+  std::size_t previous_prop_cache_iterators_size = prop_cache_iterators.size();
+  std::size_t previous_bv_cache_iterators_size = bv_cache_iterators.size();
+  record_cache_iterators = true;
+
+  // recursive call
   bvt result_bv=convert_bv(expr.where());
 
+  // clear cache_iterators that use the bound symbol from the cache
+  bv_cache_iterators.resize(previous_prop_cache_iterators_size);
```
tautschnig: `prop_cache_iterators.resize`?

tautschnig (changes requested):

src/solvers/flattening/boolbv.h
kroening: Uh, indeed.
This now makes me worried about `boolbvt::convert_bv`. It assumes that `cache_result.first` remains valid during `convert_bitvector`, which I don't believe is the case.

tautschnig: Making a few wild guesses: what makes iterators unstable is a possible `rehash` happening. Quite possibly the iterator object itself is ok, as is any reference/pointer contained in there to a value. What will have changed, however, is the relative position of objects, and thus going to the next or previous entry would not work. If this is true, maybe my concerns for this code are unfounded?

kroening:
This will need #4931 to work.

kroening:
This needs #4992.

src/solvers/flattening/boolbv.h
```diff
@@ -271,6 +271,9 @@ class boolbvt:public arrayst
 
   // strings
   numberingt<irep_idt> string_numbering;
+
+  // scopes
+  std::size_t scope_counter = 0;
```
tautschnig, line 276: Random place for comment:
- In the commit message fix s/looses/loses/
- The commit message promises "(which is the next  commit)" -- but there is no such change to be found.

tautschnig (commented):
1) See admin comments about what's being promised vs what is delivered below.
2) Could `array_comprehension_exprt` please also make use of scopes?

src/solvers/flattening/boolbv.h
kroening: Commit message fixed.

kroening:
> 2. Could `array_comprehension_exprt` please also make use of scopes?

There are many more expressions with scope (forall and exists in particular). Will deliver!

tautschnig (approved):

PR #4522, opened 2019-04-12
Avoid crash in linking conflict report by providing correct type for …

andreast271:
…dereference expression

CBMC currently crashes with an invariant check failure if there is a conflict discovered during linking.
Test case:
file one.c:
```
int f(char *i) { return  *i; }
```
file two.c:
```
int f(int *i) { return  *i; }
```
cbmc command:
```
cbmc one.c two.c
```
cbmc message without this commit:
```
Invariant check failed
File: ../util/std_expr.h:3041 function: dereference_exprt
Condition: op.type().id()==ID_pointer
Reason: Precondition
```
cbmc message with this commit:
```
file two.c line 1: 
reason for conflict at *#this: conflict on POD

char
signed int
error: conflicting function declarations f
old definition in module one file one.c line 1
signed int (char *i)
new definition in module two file two.c line 1
signed int (signed int *i$link1)
CONVERSION ERROR
```

### DIFF
--- a/src/linking/linking.cpp
+++ b/src/linking/linking.cpp
@@ -143,7 +143,8 @@ void linkingt::detailed_conflict_report_rec(
     if(depth>0 &&
        !base_type_eq(t1.subtype(), t2.subtype(), ns))
     {
-      conflict_path=dereference_exprt(conflict_path);
+      if(conflict_path.type().id() == ID_pointer)
+        conflict_path = dereference_exprt(conflict_path);
 
       detailed_conflict_report_rec(
         old_symbol,
--- a/regression/ansi-c/linking_conflicts2/main.c
+++ b/regression/ansi-c/linking_conflicts2/main.c
@@ -1,0 +1,4 @@
+int foo(int *i)
+{
+  return *i;
+}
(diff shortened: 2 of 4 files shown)


src/linking/linking.cpp
```diff
@@ -143,7 +143,7 @@ void linkingt::detailed_conflict_report_rec(
     if(depth>0 &&
        !base_type_eq(t1.subtype(), t2.subtype(), ns))
     {
-      conflict_path=dereference_exprt(conflict_path);
+      conflict_path=dereference_exprt(conflict_path, t1.subtype());
```
xbauch: clang-format would prefer spaces around `=`.

xbauch: Also assigning it the subtype of the first expression seems a bit random. Conditioning it works as well:
```
      if(conflict_path.type().id() == ID_pointer)
        conflict_path = dereference_exprt(conflict_path);
```

andreast271: I agree with your comments / suggestions and have updated the commit accordingly. I also added a regression test case for this defect.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 578ba08b4d78d4e57d9349f4173cc2a4eb2f8fa8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108103779

tautschnig (approved):

PR #4523, opened 2019-04-12
Use temporary variable to initialise array cells

smowton:
Previously when dealing with multidimensional arrays we could end up generating expressions
such as array[idx] = ALLOCATE Something; array[idx]->field1 = x; array[idx]->field2 = y;
This was a struggle for symex, as every reference to array[idx] could be a bad dereference,
and so symex::invalid_object (or perhaps some_symbol$object) references were created to account
for this possiblity. By using a temporary (generating code like
`t = ALLOCATE Something; t->field1 = x; t->field2 = y; array[idx] = t;`) it is much clearer
that the pointer dereferences must refer to the newly allocated object, and so the problematic
case is avoided.


owen-mc-diffblue (approved):

romainbrenguier (commented):
Could we have a test?

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -1207,6 +1207,29 @@ void java_object_factoryt::array_loop_init_code(
...
+  // `some_expr[idx]->x = y; some_expr[idx]->z = w;`
+  exprt init_expr;
+  if(new_item_is_primitive)
+  {
+    init_expr = arraycellref;
+  }
+  else
+  {
+    init_expr = allocate_objects.allocate_automatic_local_object(
+      arraycellref.type(), "new_array_item");
```
thk123, line 1225: Please use the same name as in the comment above (either this to `new_item`, or the comment to `new_array_item`

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -1207,6 +1207,29 @@ void java_object_factoryt::array_loop_init_code(
...
 
+  bool new_item_is_primitive = arraycellref.type().id() != ID_pointer;
+
+  // Use a temporary to initialise a new, or update an existing, non-primitive.
+  // This makes it clearer that in a sequence like
+  // `new_item->x = y; new_item->z = w;` that all the `new_item` references must
+  // alias, cf. the harder-to-analyse
+  // `some_expr[idx]->x = y; some_expr[idx]->z = w;`
+  exprt init_expr;
+  if(new_item_is_primitive)
```
thk123, line 1218: :bulb: Any reason to not to do the primitives in the same way, just in the interest of keeping this code simpler?

thk123 (commented):
Can we test this somehow?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 88a05ebac8001d8c0971b3632e6cb67849a09a2b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108119473

jbmc/src/java_bytecode/java_object_factory.cpp
smowton: Just to keep it simple indeed. `array[idx] = nondet(int)` -> `new_array_item = nondet(int); array[idx] = new_array_item;` seems like a disservice to the reader.

smowton:
Oops, I wrote a test then forgot to commit it, added now.

romainbrenguier (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 703e7c1f681d820ca05ebdf0812d650d495b5663).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108129000

thk123 (approved):

PR #4526, opened 2019-04-12
NULL compares equal to 0

tautschnig:
When comparing constants, we just compared their string representations.
This fails when one side of the (in)equality was NULL while the other
side was 0, which is wrong when config.ansi_c.NULL_is_zero is set. Thus
actually test both sides for being zero in this configuration.

### DIFF
--- a/regression/cbmc/union13/main.c
+++ b/regression/cbmc/union13/main.c
@@ -1,0 +1,12 @@
+#include <assert.h>
+#include <stdint.h>
+
+int main()
+{
+  union {
+    intptr_t i;
+    int *p;
+  } u;
+  u.i = 0;
+  assert(u.p == 0);
+}
--- a/regression/cbmc/union13/no-arch.desc
+++ b/regression/cbmc/union13/no-arch.desc
@@ -1,0 +1,9 @@
+CORE broken-smt-backend
+main.c
+--arch none --little-endian
+(Starting CEGAR Loop|^Generated 1 VCC\(s\), 1 remaining after simplification$)
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^warning: ignoring
--- a/regression/cbmc/union13/test.desc
+++ b/regression/cbmc/union13/test.desc
@@ -1,0 +1,8 @@
+CORE
+main.c
+
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^warning: ignoring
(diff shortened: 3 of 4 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 15bf786b295a494df2371bdcd22d83fda5788f4f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108192432

kroening:
If `config.ansi_c.NULL_is_zero` is false, can we really conclude that a non-zero pointer is NOT equal to NULL?

kroening:
(Very academic question; I am yet to spot a system in which NULL isn't a zero bit pattern).

tautschnig:
See http://c-faq.com/null/machexamp.html for the discussion of NULL-is-not-zero. But, yes, this remains academic.

tautschnig:
> If `config.ansi_c.NULL_is_zero` is false, can we really conclude that a non-zero pointer is NOT equal to NULL?

Hmm, indeed, we should probably just refrain from simplifying in that case. Let me fix this.

tautschnig:
> > If `config.ansi_c.NULL_is_zero` is false, can we really conclude that a non-zero pointer is NOT equal to NULL?

> Hmm, indeed, we should probably just refrain from simplifying in that case. Let me fix this.

Done.

tautschnig:
As a side note: a still-current scenario is that the address `0` is a valid place to store data, which is true in about any freestanding setting. There are probably several places where we don't currently account for this in CBMC...

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7562a998f7836d7ed5a1eed8e0f520c35fc7e8fe).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108214328

regression/cbmc/union13/main.c
```diff
@@ -0,0 +1,11 @@
+#include <assert.h>
+
+int main()
+{
+  union {
+    unsigned long long i;
```
smowton: AFAIK `unsigned long` is the type guaranteed to have same size as a pointer

smowton (commented):
Add a test exercising the `!null_is_zero` path?

regression/cbmc/union13/main.c
tautschnig: Not on Windows/LLP64... (and also it would be ok for the integer type to be larger than the pointer type in this union, but not the other way around)

smowton: Gah, so it isn't, Though perhaps `intptr_t` is better?

tautschnig: Done.

tautschnig:
> Add a test exercising the `!null_is_zero` path?

Test added!

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8f68057f665c5e6c89120598dd9773b39ae941e6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108223887

PR #4539, opened 2019-04-15
Minor sharing-map improvements, used to avoid copying in value_sett

smowton:
This adds `sharing_mapt::erase_if_exists`, `::iterate` and `::update`, and uses them to avoid copying entries in `value_sett`. The `value_sett` changes are much easier to review if you diff without whitespace sensitivity.

Note `erase_if_exists` isn't used in this PR (it's used out-of-tree), but is also the most simple of the additions.

### DIFF
--- a/src/util/sharing_map.h
+++ b/src/util/sharing_map.h
@@ -222,6 +222,19 @@ public:
   /// \param k: The key of the element to erase
   void erase(const key_type &k);
 
+  /// Erase element if it exists
+  ///
+  /// Complexity:
+  /// - Worst case: O(H * S + M)
+  /// - Best case: O(H)
+  ///
+  /// \param k: The key of the element to erase
+  void erase_if_exists(const key_type &k)
+  {
+    if(has_key(k))
+      erase(k);
+  }
+
   /// Insert element, element must not exist in map
   ///
   /// Complexity:
--- a/src/util/sharing_map.h
+++ b/src/util/sharing_map.h
@@ -138,22 +138,25 @@ Author: Daniel Poetzl
 /// to one of the maps, nodes are copied and sharing is lessened as described in
 /// the following.
 ///
-/// The replace(), insert(), and erase() operations interact with sharing as
-/// follows:
+/// The replace(), insert(), update() and erase() operations interact with
+/// sharing as follows:
 /// - When a key-value pair is inserted into the map (or a value of an existing
-/// pair is replaced) and its position is in a shared subtree, already existing
-/// nodes from the root of the subtree to the position of the key-value pair are
-/// copied and integrated with the map, and new nodes are created as needed.
+/// pair is replaced or updated) and its position is in a shared subtree,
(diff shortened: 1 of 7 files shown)


tautschnig (approved):
Looks ok to me, except:
1) It would be good for @danpoe to take a look if he's available.
2) The last commit includes a lot of whitespace changes.
3) Any data on how much of a performance difference this makes?

smowton:
I verbally ran these by @danpoe as I went but yes please do take a look. The whitespace is unavoidable per clang-format I'm afraid, due to a 2-char indent difference between iterate-with-lambda and a for-loop.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8118903982c7ab3e4706fad03110240df12e7e5d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108399530

owen-mc-diffblue (approved):

src/util/sharing_map.h
```diff
@@ -915,6 +954,18 @@ ::replace(const key_type &k, valueU &&m)
...
+SHARING_MAPT(void)
+::update(const key_type &k, std::function<void(mapped_type &)> mutator)
+{
+  innert *cp = get_container_node(k);
+  SM_ASSERT(cp != nullptr);
+
+  leaft *lp = cp->find_leaf(k);
+  PRECONDITION(lp != nullptr); // key must exist in map
+
+  lp->mutate_value(mutator);
```
danpoe, line 1002: Can we add a debugging check here that tests if the mutated element is different from the old element? (if `fail_if_equal` is set, similar to the invariant in `replace()`)

danpoe (approved):
Looks good to me. I'd also be curious if there's a performance difference.

danpoe:
We should also update the documentation at the top of `sharing_map.h` to mention `update()`.

smowton:
@tautschnig @danpoe conducted a quick micro-benchmark:

Procedure: populate a value-set with 100K keys, each with 10 values. Modify half the keys by adding 1 extra value, then make-union the two value sets.

Union operation took  ~300ms with these changes, or ~750ms without them. The direction of the union made little difference (in one direction it would be a no-op; in one direction it would add a single value to 50k entries), suggesting the cost of editing 50k keys in place is small compared to the cost of copying 100k values out in order to edit them.

smowton:
@tautschnig @danpoe you might want to re-review; I made @danpoe's requested changes re: checking the update makes a change; this required me to add a flag so invariants can throw exceptions instead of aborting (so I could unit test them) and I also fixed `replace`'s debug mode at the same time.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: ff16508074997b6cb3562e0fc6fa15f571efe1cc).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108601221

PR #4544, opened 2019-04-17
Clean up in string solver

romainbrenguier:
This deduplicate the use of `array_pool.find` and `char_array_of_pointer` and move the code for instantiation. This is preparation to make coming changes to the solver easier.

### DIFF
--- a/src/solvers/strings/array_pool.h
+++ b/src/solvers/strings/array_pool.h
@@ -97,13 +97,6 @@ private:
 /// are given as a struct containing a length and pointer to an array.
 array_string_exprt of_argument(array_poolt &array_pool, const exprt &arg);
 
-/// Return the array associated to the given pointer or creates a new one
-DEPRECATED("use pool.find(pointer, length) instead")
-const array_string_exprt &char_array_of_pointer(
-  array_poolt &pool,
-  const exprt &pointer,
-  const exprt &length);
-
 /// Fetch the string_exprt corresponding to the given refined_string_exprt
 /// \param pool: pool of arrays representing strings
 /// \param expr: an expression of refined string type
--- a/src/solvers/strings/string_constraint_instantiation.cpp
+++ b/src/solvers/strings/string_constraint_instantiation.cpp
@@ -14,10 +14,8 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
 #include <unordered_set>
 #include <util/expr_iterator.h>
 
-/// look for the symbol and return true if it is found
-/// \param index: an index expression
-/// \param qvar: a symbol expression
-/// \return a Boolean
+/// Look for symbol \p qvar in the expression \p index and return true if found
+/// \return True, iff \p qvar appears in \p index.
 static bool contains(const exprt &index, const symbol_exprt &qvar)
 {
   return std::find(index.depth_begin(), index.depth_end(), qvar) !=
--- a/src/solvers/strings/string_constraint_instantiation.cpp
+++ b/src/solvers/strings/string_constraint_instantiation.cpp
@@ -88,69 +88,66 @@ std::map<exprt, int> map_representation_of_sum(const exprt &f)
   return elems;
 }
 
-exprt
-sum_over_map(std::map<exprt, int> &m, const typet &type, bool negated)
+exprt sum_over_map(std::map<exprt, int> &m, const typet &type, bool negated)
 {
-  exprt sum = nil_exprt();
-  mp_integer constants = 0;
-  typet index_type;
   if(m.empty())
     return from_integer(0, type);
-  else
-    index_type = m.begin()->first.type();
+
+  exprt sum = nil_exprt();
+  mp_integer constants = 0;
+  typet index_type = m.begin()->first.type();
 
   for(const auto &term : m)
   {
-    // We should group constants together...
     const exprt &t = term.first;
-    int second = negated ? (-term.second) : term.second;
+    int factor = negated ? (-term.second) : term.second;
     if(t.id() == ID_constant)
     {
+      // Constants are accumulated in the variable \c constants
       const auto int_value = numeric_cast_v<mp_integer>(to_constant_expr(t));
-      constants += int_value * second;
+      constants += int_value * factor;
     }
     else
     {
-      switch(second)
+      switch(factor)
       {
-        case -1:
+      case -1:
+        if(sum.is_nil())
+          sum = unary_minus_exprt(t);
+        else
+          sum = minus_exprt(sum, t);
+        break;
+
+      case 1:
+        if(sum.is_nil())
+          sum = t;
+        else
+          sum = plus_exprt(sum, t);
+        break;
+
+      default:
+        if(factor > 1)
+        {
+          if(sum.is_nil())
+            sum = t;
+          else
+            plus_exprt(sum, t);
+          for(int i = 1; i < factor; i++)             sum = plus_exprt(sum, t);
+        }
+        else if(factor < -1)
+        {
           if(sum.is_nil())
             sum = unary_minus_exprt(t);
           else
             sum = minus_exprt(sum, t);
-          break;
-
-        case 1:
-          if(sum.is_nil())
-            sum = t;
-          else
-            sum = plus_exprt(sum, t);
-          break;
-
-        default:
-          if(second > 1)
-          {
-            if(sum.is_nil())
-              sum = t;
-            else
-              plus_exprt(sum, t);
-            for(int i = 1; i < second; i++)
-              sum = plus_exprt(sum, t);
-          }
-          else if(second < -1)
-          {
-            if(sum.is_nil())
-              sum = unary_minus_exprt(t);
-            else
-              sum = minus_exprt(sum, t);
-            for(int i = -1; i > second; i--)
-              sum = minus_exprt(sum, t);
-          }
+          for(int i = -1; i > factor; i--)
+            sum = minus_exprt(sum, t);
+        }
       }
     }
   }
 
-  exprt index_const = from_integer(constants, index_type);
+  const exprt index_const = from_integer(constants, index_type);
   if(sum.is_not_nil())
     return plus_exprt(sum, index_const);
   else
--- a/src/solvers/strings/string_constraint_instantiation.cpp
+++ b/src/solvers/strings/string_constraint_instantiation.cpp
@@ -131,8 +131,9 @@ exprt sum_over_map(std::map<exprt, int> &m, const typet &type, bool negated)
           if(sum.is_nil())
             sum = t;
           else
-            plus_exprt(sum, t);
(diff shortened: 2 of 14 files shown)


src/solvers/strings/array_pool.h
```diff
@@ -99,11 +99,6 @@ array_string_exprt of_argument(array_poolt &array_pool, const exprt &arg);
 
 /// Return the array associated to the given pointer or creates a new one
 DEPRECATED("use pool.find(pointer, length) instead")
```
tautschnig: Shouldn't the above also be removed?

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
 /// \file
 /// Defines related function for string constraints.
 
+#include <unordered_set>
+#include <algorithm>
+#include <util/expr_iterator.h>
 #include "string_constraint_instantiation.h"
 
+/// look for the symbol and return true if it is found
```
tautschnig: How about: "Look for symbol \p qvar in the expression \p index and return true if it was found."

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+/// \param index: an index expression
+/// \param qvar: a symbol expression
+/// \return a Boolean
```
tautschnig: "True, iff \p qvar appears in \p index."

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
 /// \file
 /// Defines related function for string constraints.
 
+#include <unordered_set>
+#include <algorithm>
+#include <util/expr_iterator.h>
 #include "string_constraint_instantiation.h"
 
+/// look for the symbol and return true if it is found
+/// \param index: an index expression
```
tautschnig: Does it actually matter whether this is an "index" expression?!

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+      to_process.emplace_back(cur.op0(), positive);
+    }
+    else if(cur.id() == ID_unary_minus)
+    {
+      to_process.emplace_back(cur.op0(), !positive);
+    }
+    else
+    {
+      if(positive)
+        elems[cur] = elems[cur] + 1;
```
tautschnig: `++elems[cur];`

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+      else
+        elems[cur] = elems[cur] - 1;
```
tautschnig: `--elems[cur];` (this actually makes a substantial difference here, because it avoids a duplicate map lookup)

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+  std::for_each(expr.depth_begin(), expr.depth_end(), [&](const exprt &e) {
+    if(index_str_containing_qvar(e))
+      result.insert(to_index_expr(e).index());
+  });
+  return result;
+}
+
+std::map<exprt, int> map_representation_of_sum(const exprt &f)
+{
+  // number of time the leaf should be added (can be negative)
```
tautschnig: s/time/times/

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+    }
+  }
+  return elems;
+}
+
+exprt
+sum_over_map(std::map<exprt, int> &m, const typet &type, bool negated)
+{
+  exprt sum = nil_exprt();
+  mp_integer constants = 0;
```
tautschnig: Don't declare or create them earlier than necessary. Notably they won't be needed at all if `m.empty()`.

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+  typet index_type;
+  if(m.empty())
+    return from_integer(0, type);
+  else
+    index_type = m.begin()->first.type();
+
+  for(const auto &term : m)
+  {
+    // We should group constants together...
```
tautschnig: What is this comment trying to suggest?

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+    const exprt &t = term.first;
+    int second = negated ? (-term.second) : term.second;
+    if(t.id() == ID_constant)
```
tautschnig: I suppose this is what the comment refers to - why isn't this already done in `map_representation_of_sum`?

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+
+        default:
+          if(second > 1)
+          {
+            if(sum.is_nil())
+              sum = t;
+            else
+              plus_exprt(sum, t);
+            for(int i = 1; i < second; i++)
+              sum = plus_exprt(sum, t);
```
tautschnig: Why doesn't this use multiplication?

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+/// \return an expression corresponding of $f^{-1}(val)$ where $f$ is seen as
+///   a function of $qvar$, i.e. the value that is necessary for `qvar` for `f`
+///   to be equal to `val`. For instance, if `f` corresponds to the expression
+///   $q + x$, `compute_inverse_function(q,v,f)` returns an expression
+///   for $v - x$.
+static exprt
+compute_inverse_function(const exprt &qvar, const exprt &val, const exprt &f)
+{
+  exprt positive, negative;
+  // number of time the element should be added (can be negative)
```
tautschnig: s/time/times/

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -9,8 +9,234 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
...
+  {
+    neg = (it->second == 1);
+  }
+  else
+  {
+    INVARIANT(
+      it->second == 0,
+      string_refinement_invariantt(
+        "a proper function must have exactly one "
+        "occurrences after reduction, or it cancelled out, and it does not"
```
tautschnig: s/occurrences/occurrence/

src/solvers/strings/string_constraint_instantiation.cpp
romainbrenguier: `map_representation_of_sum` just maps expressions to the number of time they appear in the sum, without doing anything special for constants, I clarified that in the example for `map_representation_of_sum`

src/solvers/strings/string_constraint_instantiation.cpp
romainbrenguier: I think the original intent was to have only plus_exprt and minus_exprt but this shouldn't matter for our purpose, I will replace it by multiplication to make the code simpler

romainbrenguier:
@tautschnig I implemented your suggestions in separate commits to keep the moving commit clean.

src/solvers/strings/string_constraint_instantiation.cpp
tautschnig: I understand that - but is that really the most appropriate design?

romainbrenguier: Another possibility would be to use an exprt equal to 1 in the map and add all constants to its factor. The current implementation of `sum_over_map` handles both cases so I wouldn't change it. We could make `map_representation_of_sum` do more work but I don't think there is not a huge need for it.

tautschnig: My interpretation of this code is that you are computing the coefficients of a polynomial. (I do not have a sufficient understanding of the code to tell why you need to do this, but at least that's my reading of this code.) If this interpretation is correct, then I'd consider renaming the function to make this clear. And if you're doing this, then certainly constants should be folded to become the coefficient of the zero-degree summand.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 9e50b3f93fe9ebc0d5a4955bb2cab0c1a64897b2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108807806

src/solvers/strings/string_constraint_instantiation.cpp
romainbrenguier: It's actually a linear function rather than a polynomial function, and computing this representation is useful to inverse the function. I've put this in a class `linear_functiont` to make that clearer (constants are now stored in a single field).

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 3b194e0adfe7ad32f0827f1335239b7382bbe7ed).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/108870055

src/solvers/strings/string_constraint_instantiation.cpp
```diff
@@ -10,6 +10,232 @@ Author: Jesse Sigal, jesse.sigal@diffblue.com
 /// Defines related function for string constraints.
 
 #include "string_constraint_instantiation.h"
+#include <algorithm>
```
peterschrammel, line 13: typo in commit message: instantiation

src/solvers/strings/string_constraint_instantiation.h
```diff
@@ -35,7 +35,7 @@ std::vector<exprt> instantiate_not_contains(
   const std::unordered_map<string_not_contains_constraintt, symbol_exprt>
     &witnesses);
 
-/// Connonical representation of linear function, for instance, expression
+/// Cannonical representation of linear function, for instance, expression
```
peterschrammel: typo canonical

peterschrammel (approved):

romainbrenguier:
@tautschnig do you want to have a second look?

PR #4555, opened 2019-04-18
Value-set dereference: use cond_exprt to avoid quadratic guards

smowton:
### The problem

When we have a moderately complicated pointer expression, `p`, with a large-ish alias set (size `n`) named `o1`, `o2` ... `on`, then `value_set_dereferencet` would create an expression of the form:
`p == &o1 ? o1 : p == &o2 ? o2 : p == &o3 : ...`

So far so good -- the expression `p` has been duplicated `n` times, but irep sharing means that's not actually very costly.

However, if the deref occurs on the LHS, then `symex_assign_if` expands this into a series of assignments:
```
o1#2 = p == &o1 ? new_value : o1#1
o2#2 = p != &o1 && p == &o2 ? new_value : o2#1
o3#2 = p != &o1 && p != &o2 && p == &o3 ? new_value : o3#1
...
```

There are now n^2 copies of `p`, and thanks to renaming they are no longer shared. The simplifier then proceeds to spend quite a long time on this assignment sequence if n is large-ish and so n^2 is very large.

### The solution

The conditions `p != &o1 && p != &o2 && p == &o3` are actually redundant, but our nested `if_exprt` chain doesn't actually make that clear. Therefore I add a flag `is_exclusive` to the already-existing `cond_exprt`, which expresses an if-elseif-elseif-... chain in one instruction. This means that we now get
```
o1#2 = p == &o1 ? new_value : o1#1
o2#2 = p == &o2 ? new_value : o2#1
o3#2 = p == &o3 ? new_value : o3#1
```
Back to linear complexity on the number of aliases. Much better. 90% of the diff in the PR is teaching various passes how to deal with `cond_exprt`, which presumably we'd want to do sooner or later; review commit-by-commit.

Note this doesn't implement cond-expr *everywhere*; particularly the string solver doesn't understand it yet. For now the expression is turned back into nested-ifs before it gets to that module.

### Rejected solutions

**Use a let-expression**: Rather than try to avoid quadratic conditionals we could just surround the whole thing in `let p2 = p in ...`. I think this would work: `symex_assign` would need to learn to evaluate let-around-if, presumably leading to `let p2 = p in p2 != &o1 && p2 != &o2 && p2 == &o3 ...`, and the simplifier would have a linear number of `p` instances to attack and a quadratic number of shallow `==` nodes. I thought it was better to avoid the quadratic syntax tree node problem at its root.

**Use caching in the simplifier**: While a quadratic number of comparisons would still occur, the simplifier can check before simplifying if it has seen an identical node before and avoid the repeated work. However this doesn't really deal with the case where `p` is complicated and doesn't simplify much or at all -- in this case renaming and subsequent passes still have a huge expression to deal with.

**Avoid quadratic guards by analysis**: when symex comes to add `p == &o3` to its stack of guards, it could also remove any that are incompatible with it. This feels brittle to me -- the tests generated by `value_set_dereferencet` are not always that simple, so we'd have to at least cover the various expressions it creates, and would rely on the guard elimination remaining clever enough if/when value-set-deref produces different expressions. We could try doing this as well, but I'd prefer if value-set-deref was explicit that it intends to produce a disjoint switch.

### DIFF
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -155,6 +155,9 @@ void goto_symext::symex_assign_rec(
   else if(lhs.id()==ID_if)
     symex_assign_if(
       state, to_if_expr(lhs), full_lhs, rhs, guard, assignment_type);
+  else if(lhs.id() == ID_cond)
+    symex_assign_cond(
+      state, to_cond_expr(lhs), full_lhs, rhs, guard, assignment_type);
   else if(lhs.id()==ID_typecast)
     symex_assign_typecast(
       state, to_typecast_expr(lhs), full_lhs, rhs, guard, assignment_type);
@@ -611,6 +614,44 @@ void goto_symext::symex_assign_if(
       state, lhs.false_case(), full_lhs, rhs, guard, assignment_type);
     guard.pop_back();
   }
+}
+
+void goto_symext::symex_assign_cond(
+  statet &state,
+  const cond_exprt &lhs,
+  const exprt &full_lhs,
+  const exprt &rhs,
+  exprt::operandst &guard,
+  assignment_typet assignment_type)
+{
+  std::size_t old_guard_size = guard.size();
+
+  for(std::size_t i = 0; i < lhs.get_n_cases(); ++i)
+  {
+    exprt renamed_guard = state.rename(lhs.condition(i), ns).get();
+    do_simplify(renamed_guard);
+    if(!renamed_guard.is_false())
+    {
+      guard.push_back(renamed_guard);
+      symex_assign_rec(
+        state, lhs.value(i), full_lhs, rhs, guard, assignment_type);
+      guard.pop_back();
+    }
+
+    // If this one is a certainty the remaining cases are irrelevant:
+    if(renamed_guard.is_true())
+      break;
+
+    // If the conditions are non-exclusive, further cases can only happen if
+    // this one did not. If they are exclusive then they can be tested
+    // independently.
+    if(!lhs.is_exclusive())
+      guard.push_back(not_exprt(renamed_guard));
+  }
+
+  // Restore the guard to its state before entering this function:
+  INVARIANT(guard.size() >= old_guard_size, "must not shrink the guard!");
+  guard.resize(old_guard_size);
 }
 
 void goto_symext::symex_assign_byte_extract(
--- a/src/goto-symex/symex_clean_expr.cpp
+++ b/src/goto-symex/symex_clean_expr.cpp
@@ -14,6 +14,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/arith_tools.h>
 #include <util/byte_operators.h>
 #include <util/c_types.h>
+#include <util/expr_iterator.h>
 #include <util/pointer_offset_size.h>
 #include <util/simplify_expr.h>
 
@@ -169,6 +170,33 @@ replace_nondet(exprt &expr, symex_nondet_generatort &build_symex_nondet)
   }
 }
 
+static void lower_cond_expr(exprt &expr)
+{
+  for(auto it = expr.depth_begin(), itend = expr.depth_end(); it != itend; ++it)
+  {
+    if(it->id() == ID_cond)
+    {
+      exprt new_expr;
+      const auto &cond_expr = to_cond_expr(*it);
+      INVARIANT(cond_expr.get_n_cases() >= 1, "cond_expr should not be empty");
+      for(std::size_t i = cond_expr.get_n_cases() - 1;; --i)
+      {
+        if(i == cond_expr.get_n_cases() - 1)
+          new_expr = cond_expr.value(i);
+        else
+        {
+          new_expr =
+            if_exprt(cond_expr.condition(i), cond_expr.value(i), new_expr);
+        }
+        if(i == 0)
+          break;
+      }
+
+      it.mutate() = std::move(new_expr);
+    }
+  }
+}
+
 void goto_symext::clean_expr(
   exprt &expr,
   statet &state,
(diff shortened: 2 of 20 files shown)


smowton:
Of course I'll add tests and benchmarking figures if people are happy with the overall approach. Anecdata so far: improves run time for an alias set of size 30 from around a minute to less than a second.

src/goto-symex/symex_assign.cpp
```diff
@@ -613,6 +616,44 @@ void goto_symext::symex_assign_if(
...
+  const exprt &full_lhs,
+  const exprt &rhs,
+  exprt::operandst &guard,
+  assignment_typet assignment_type)
+{
+  std::size_t old_guard_size = guard.size();
+
+  for(std::size_t i = 0; i < lhs.get_n_cases(); ++i)
+  {
+    exprt renamed_guard = state.rename(lhs.condition(i), ns).get();
```
owen-mc-diffblue, line 631: :bulb: It's confusing that the variable names don't distinguish between the incoming guard and the extra bit coming from the current case we're considering. Maybe `renamed_case_guard`?

src/util/expr_util.h
```diff
@@ -74,6 +75,9 @@ bool has_subtype(const typet &, const irep_idt &id, const namespacet &);
 /// lift up an if_exprt one level
 if_exprt lift_if(const exprt &, std::size_t operand_number);
 
+/// lift up an cond_exprt one level
```
owen-mc-diffblue, line 78: "an cond_exprt" -> "a cond_exprt"

smowton:
@tautschnig @kroening this now passes tests, and is ready for review

owen-mc-diffblue (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 2e48edd1ef796b8d4c986f8082a6b775e3b00311).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/109308178

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -107,12 +107,12 @@ exprt value_set_dereferencet::dereference(const exprt &pointer)
         may_fail=true;
   }
 
+  exprt failure_value;
```
tautschnig, line 117: I find it too non-trivial to check that `failure_value` is never used uninitialised. See comment below.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -138,36 +138,61 @@ exprt value_set_dereferencet::dereference(const exprt &pointer)
...
-  #if 0
-  std::cout << "R: " << format(value) << "\n\n";
-  #endif
+  // I'd like to put an invariant here that values without a pointer guard, such
+  // as integer_address, cannot co-occur with failed objects, but this isn't the
+  // case. There's no way to write a GOTO condition to discriminate between the
+  // two however, so purely by historical accident, the failed object takes
+  // precedence:
+
+  if(may_fail || value_without_condition.has_value())
```
tautschnig, line 181: Can't we just combine `value_without_condition` and `failure_value` into a single expression?

src/goto-symex/symex_assign.cpp
```diff
@@ -613,6 +616,44 @@ void goto_symext::symex_assign_if(
...
+
+    // If the conditions are non-exclusive, further cases can only happen if
+    // this one did not. If they are exclusive then they can be tested
+    // independently.
+    if(!lhs.is_exclusive())
+      guard.push_back(not_exprt(renamed_guard));
+  }
+
+  // Restore the guard to its state before entering this function:
+  INVARIANT(guard.size() >= old_guard_size, "must not shrink the guard!");
```
tautschnig, line 653: s/shrink/grow/?

src/goto-symex/symex_clean_expr.cpp
```diff
@@ -169,6 +170,33 @@ replace_nondet(exprt &expr, symex_nondet_generatort &build_symex_nondet)
...
+      {
+        if(i == cond_expr.get_n_cases() - 1)
+          new_expr = cond_expr.value(i);
+        else
+        {
+          new_expr =
+            if_exprt(cond_expr.condition(i), cond_expr.value(i), new_expr);
+        }
+        if(i == 0)
+          break;
```
tautschnig, line 213: I'd use `for(std::size_t i = cond_expr.get_n_cases(); i > 0; --i)` and then `if(i == cond_expr.get_n_cases())` as well as `cond_expr.value(i - 1)` and cond_expr.condition(i - 1)`.

tautschnig (approved):
Seems like a good idea, but can we also have some tests that, e.g., look at the output of `--program-only` or `--show-vcc`?

kroening:
I am a bit worried that the `is_exclusive` extension of `cond_exprt` may be a bit surprising/difficult to spot. How about a `switch_exprt`, inspired by the switch expression in Java 13?

smowton:
I re-tested this on security now that it passes tests and found that all the previously observed performance benefits have evaporated, so presumably I was benefiting from bugs (probably false pointer guards leading to whole assignments getting simplified away), not the actual improvement I thought I was. I observe some performance benefit when `cond_exprt` is used on the RHS as well as the LHS (but much less than the benefit I originally thought I was seeing), but that crashes the string solver later and would presumably require much more cases to handle `cond_exprt`, including the string solver and all analyses it relies on.

Notes going forward:
1. Trying to debug this I am *still* occasionally seeing guards like `p != &o1 && p != &o2 && p == &o3` on the right-hand side, despite `goto_symext::assign_if` not being used, which I thought was the culprit for introducing these.
2. Inspecting the case I was looking at more closely, the culprit is an array-invert function, of the form
```
for(...) { 
  result[i][j] = input[j][i];
}
```
The aliasing problem on the RHS is much worse than the LHS (192 aliases vs. 20), but significant savings could still probably be made by simplifying and renaming the RHS once prior to `symex_assign_if` / `symex_assign_cond`, rather than duplicating it and then renaming/simplifying every time we get down to `symex_assign_symbol`.

smowton:
Further investigation of point (1) above: the strings of negations result from simplifying an expression of the form `x ? false : z` into `!x && z`. Ultimately the simplifier is reflecting that the nested-if tower *does* mean dependent (non-exclusive) conditions. A moderate performance win can indeed be achieved by using an exclusive `switch_exprt`, but it must be used on the RHS too. I'll incrementally add support for this over the coming days/weeks.

tautschnig:
@smowton With #4576 merged, is this one still relevant?

PR #4556, opened 2019-04-19
RFC Upstream cnf dumper

nmanthey:
In this first iteration, I just want to share this code so that the performance of SAT solvers can be investigated outside of CBMC. This modification might be used to create coverage based CNFs that can be used for this years SAT competition.

Note, that actually only the first produced file is equivalent to what your internal SAT solver is doing, as follow up calls already build on top of internal state.

I might extend this class to produce incremental CNFs instead of plain CNFs, to close this gap.

### DIFF
--- a/src/solvers/sat/cnf.h
+++ b/src/solvers/sat/cnf.h
@@ -14,6 +14,13 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <solvers/prop/prop.h>
 
+#include <cstdlib>
+#include <vector>
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <unistd.h>
+
 class cnft:public propt
 {
 public:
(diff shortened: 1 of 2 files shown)


kroening:
Use `dimacs_cnft`?

nmanthey:
If only I would be familiar with all the classes you provide.

Lets see this as a very early prototype that allows us to dump (i)CNF today
(soonish) for SAT Race submissions.

I am not sure how much more i can make that CBMC style code in the near
future.

Best,
Norbert

kroening:
I would like to find out more about the use-case: is it essential that this is triggered via an environment variable? Or would command-line be ok?

kroening:
Also, the PR doesn't do incremental, but I understand that you'd like to get there?

nmanthey:
Command line would be okay, for rapid prototyping I went the getenv route.

Incremental is the next step yes, so that you can basically re-run all calls to a SAT engine offline based on the icnf files (http://www.siert.nl/icnf/). I have drafted a commit that extends the current commits into being able to also dump an icnf file once when destructing the object. The other pending change is to also log the assumptions that are used for the SAT call, so that you can actually do proper offline SAT benchmarking.

The other piece that is missing is an icnf front-end for IPASIR, so that solvers can be tested easily. I also drafted that, but it's not ready to be consumed or merged into the ipasir package. But I guess this is where this should actually go to.

TGWDB:
Closing this due to lack of update/progress. Please reopen if you believe this is erroneous and plan to update/rebase (also it is not clear to me if this relates to the `--dimacs` output option?).

PR #4564, opened 2019-04-23
Fix L2 renaming after shift-to-rhs and enable validation

tautschnig:
When rewriting the RHS to enable field-sensitive LHS-names, we may
introduce the LHS into the RHS, and thus need to re-rename the RHS
expression to ensure all of it is L2. We could have caught this earlier
if we had always enabled SSA validation in regression tests, which is
now done (except for C++ tests, which require a different set of fixes).


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 790e455af0b05596ee56ebbfb5b4622b9da9c1a9).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/109350154

karkhaz (approved):

smowton (approved):

PR #4565, opened 2019-04-23
Mangle static functions using basename, not path

karkhaz:
Previously, if we compiled a file called

    foo/bar/baz/main.c

containing a static function called `qux`, that function's name would be
mangled to

    __CPROVER_file_local_foo_bar_baz_main_c_qux

This commit changes the behaviour so that only the basename is used to
mangle the function name:

    __CPROVER_file_local_main_c_qux

This is so that users writing CBMC proofs for file-local functions don't
have to worry about where the goto-binary was compiled.

### DIFF
--- a/src/goto-programs/name_mangler.cpp
+++ b/src/goto-programs/name_mangler.cpp
@@ -8,18 +8,20 @@ Author: Kareem Khazem <karkhaz@karkhaz.com>, 2019
 
 #include "name_mangler.h"
 
+#include <util/get_base_name.h>
+
 #include <iomanip>
 #include <sstream>
 
 irep_idt file_name_manglert::
 operator()(const symbolt &src, const std::string &extra_info)
 {
+  std::string basename = get_base_name(src.location.get_file().c_str(), false);
+
   std::stringstream ss;
   ss << CPROVER_PREFIX << "file_local_";
   ss << std::regex_replace(
-          std::regex_replace(src.location.get_file().c_str(), forbidden, "_"),
-          multi_under,
-          "_")
+          std::regex_replace(basename, forbidden, "_"), multi_under, "_")
      << "_";
 
   if(extra_info != "")
--- a/regression/goto-cc-file-local/chain.sh
+++ b/regression/goto-cc-file-local/chain.sh
@@ -23,28 +23,59 @@ goto_instrument=$2
 cbmc=$3
 is_windows=$4
 
-for src in *.c; do
+if [[ "${is_windows}" != "true" && "${is_windows}" != "false" ]]; then
+  (>&2 echo "\$is_windows should be true or false (got '" "${is_windows}" "')")
+  exit 1
+fi
+
+if is_in use_find "$ALL_ARGS"; then
+  SRC=$(find . -name "*.c")
+else
+  SRC=*.c
+fi
+
+echo "Source files are ${SRC}"
+
+wall=""
+if is_in wall "$ALL_ARGS"; then
+  if [[ "${is_windows}" == "true" ]]; then
+    wall="/Wall"
+  else
+    wall="-Wall"
+  fi
+fi
+
+cnt=0
+for src in ${SRC}; do
+  cnt=$((cnt + 1))
+  out_suffix=""
+  if is_in out-file-counter "$ALL_ARGS"; then
+    out_suffix="_$cnt"
+    if is_in suffix "$ALL_ARGS"; then
+      suffix="--mangle-suffix custom_$cnt"
+    fi
+  fi
+
   base=${src%.c}
-  OUT_FILE="${base}.gb"
+  OUT_FILE=$(basename "${base}${out_suffix}")".gb"
 
   if [[ "${is_windows}" == "true" ]]; then
     "${goto_cc}"                        \
         --export-function-local-symbols \
         --verbosity 10                  \
+        ${wall}                         \
         ${suffix}                       \
         /c "${base}.c"                  \
         /Fo"${OUT_FILE}"
 
-  elif [[ "${is_windows}" == "false" ]]; then
+  else
     "${goto_cc}"                        \
         --export-function-local-symbols \
         --verbosity 10                  \
+        ${wall}                         \
         ${suffix}                       \
         -c "${base}.c"                  \
         -o "${OUT_FILE}"
-  else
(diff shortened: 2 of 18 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 45f834333de35d869eb1c4b38823794c226bae5b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/109366495

peterschrammel (commented):
How do you disambiguate static functions from different compilation units with the same basename?

src/goto-programs/name_mangler.cpp
```diff
@@ -14,12 +14,23 @@ Author: Kareem Khazem <karkhaz@karkhaz.com>, 2019
...
+  std::vector<std::string> path_components(
+    std::sregex_token_iterator(file.begin(), file.end(), path_separator, -1),
+    std::sregex_token_iterator());
+
+  INVARIANT(
+    path_components.size() > 0,
+    "A path should contain at least one component, the path was '" + file +
+      "'");
+
+  const std::string &basename = path_components[path_components.size() - 1];
```
tautschnig: Please use `get_base_name` from `util/get_base_name.{h,cpp}`.

tautschnig (commented):
@peterschrammel's question seems key, and it should be possible to simplify the implementation using `get_base_name`.

karkhaz:
@peterschrammel by default, they get the same name mangling. This was always my intention, this patch actually makes it true.

The idea is: this name mangling scheme will work almost all the time. If we have two files with the same basename, then the user should manually decide how the name should be mangled using the [--mangle-suffix](https://github.com/diffblue/cbmc/blob/45f834333de35d869eb1c4b38823794c226bae5b/src/goto-cc/gcc_cmdline.cpp#L30) switch to `goto-cc`. That suffix is appended to the rest of the name [on this line](https://github.com/diffblue/cbmc/blob/45f834333de35d869eb1c4b38823794c226bae5b/src/goto-programs/name_mangler.cpp#L36).

Here is the use case: I want to run CBMC on a static function in some file, and have that part of CI. In that case, I have the following:
```
// harness.c
void harness {
  int a, b, c;
  __CPROVER_file_local_main_c_foo(a, b, c);
}
// Makefile
default:
  goto-cc -c -o harness harness.c
  goto-cc -c -o main --export-function-local-symbols main.c
  goto-cc -o fully_linked.gb harness main
  cbmc fully_linked.gb
```

Now if I need to compile another `main.c` in a different directory, then I manually specify a suffix in the makefile and harness. Because I have explicitly written them down, it will still work in CI, as it doesn't rely on flaky filenames:

```
// harness.c
void harness {
  int a, b, c;
  __CPROVER_file_local_main_c_foo_qux(a, b, c);
}
// Makefile
default:
  goto-cc -c -o harness harness.c
  goto-cc -c -o main --export-function-local-symbols --mangle-suffix qux some/long/path/qux/main.c
  goto-cc -o fully_linked.gb harness main
  cbmc fully_linked.gb

# other main.c can be built with a different suffix
```

peterschrammel:
> the user should manually decide how the name should be mangled

That seems to be ok as an option to deal with your special use case, but why should I need to care about that in the normal case? Maybe I'm still misunderstanding the purpose of that switch.

tautschnig (approved):
I believe the link-time conflict will yield a warning or error so that the user can act upon that (do the tests actually exercise this case?) - but I don't think the single-file case currently has such warnings? I think the current use of `model.symbol_table.insert` is prone to hiding such errors. Could you please fix this either in this PR or in a follow-up work to make sure that any conflict is raised with the user?

karkhaz:
> Maybe I'm still misunderstanding the purpose of that switch

the `--mangle-suffix` switch is literally just for the user to be able to customize the name mangling, in precisely the situation you describe (where two files have the same basename and contain a static function with the same name, so the mangled name has to be different in each case).

Obviously it would be better to guarantee that the mangled name is unique all the time, but I haven't come up with a better way that also fulfills the following requirements:
* Not fragile (i.e. works even if you re-build on a different machine, under a different path)
* Readable. I did also include an implementation that uses a hash rather than a filename, but I want people to be able to look at the harness and know exactly what function is being called, the hash doesn't allow for that.

Do you have any suggestions? I agree that what I posted isn't totally ideal, but it does work for the majority of cases, and there's a manual override for the few cases where it doesn't work.

karkhaz:
@tautschnig bizarrely, even in the multi-file case, I get no errors when linking. One of the implementations gets clobbered, and it's the same implementation regardless of linking order:
```c
// foo/main.c
int bar();

static int static_fun()
{
  return bar();
}

int main()
{
  int x = static_fun();
}
```
```c
// bar/main.c
static int static_fun()
{
  return 3;
}

int bar()
{
  return static_fun();
}
```
```sh
> goto-cc -c -o foo.o --export-function-local-symbols foo/main.c
> goto-cc -c -o bar.o --export-function-local-symbols bar/main.c
> goto-cc foo.o bar.o
> #
> # a.out contains the foo.c implementation of static_fun
> #
> goto-cc bar.o foo.o
> #
> # a.out STILL contains the foo.c implementation of static_fun!
> #
```
Certainly it seems like a good idea to emit a warning or error from the linking code. I might need some pointers, will look into this at some point next week.

tautschnig:
@karkhaz You may want to retry the test that you posted with `-Wall` (and maybe also `-Werror`). I believe that should at least yield some warnings.

karkhaz:
@tautschnig correct, thanks:
```sh
> goto-cc -Wall bar.o foo.o
foo/main.c:3:1: warning: function `__CPROVER_file_local_main_c_foo' in module `main' 
is shadowed by a definition in module `main'
 static int foo()
```
So in that case I'm not clear what you're asking for in the "single file case". If there's only one file, then nothing would be shadowed, right? What warnings would you like to see?

tautschnig:
> So in that case I'm not clear what you're asking for in the "single file case". If there's only one file, then nothing would be shadowed, right? What warnings would you like to see?

What about the following two cases: 1) a user chose to name their non-static function in such a way that the newly added prefix would cause a collision; 2) we are linking an additional source file to an object file where static functions have already been externalised. I suppose the second case would be caught by the linker, but the first one likely remains an issue (although a very unlikely one - it's probably fair to consider that an invariant).

PR #4567, opened 2019-04-24
Symex: shortcut the common case of whole-struct initialisation

smowton:
Rather than assign a temporary symbol and then expand it, directly extract the fields
and assign them. This cuts the time taken on one initialisation-heavy benchmark to symex and print VCCs from 3.6s to 3.2s.

### DIFF
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -364,6 +364,51 @@ static void shift_indexed_access_to_lhs(
   }
 }
 
+/// Assign a struct expression to a symbol. If \ref symex_assign_symbol was used
+/// then we would assign the whole symbol, before extracting its components,
+/// with results like `x = {1, 2}; x..field1 = x.field1; x..field2 = x.field2;`
+/// This abbreviates the process, directly producing
+/// `x..field1 = 1; x..field2 = 2;`
+/// \param state: goto-symex state
+/// \param lhs: symbol to assign (already renamed to L1)
+/// \param full_lhs: expression skeleton corresponding to \p lhs, to be included
+///   in the result trace
+/// \param rhs: struct expression to assign to \p lhs
+/// \param guard: guard conjuncts that must hold for this assignment to be made
+/// \param assignment_type: assignment type (see
+///   \ref symex_targett::assignment_typet)
+void goto_symext::symex_assign_from_struct(
+  statet &state,
+  const ssa_exprt &lhs, // L1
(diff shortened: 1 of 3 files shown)


src/goto-symex/symex_assign.cpp
```diff
@@ -364,6 +364,50 @@ static void shift_indexed_access_to_lhs(
...
+  for(std::size_t i = 0; i < components.size(); ++i)
+  {
+    const auto &comp = components[i];
+    exprt lhs_field = member_exprt(lhs, comp.get_name(), comp.type());
+    state.field_sensitivity.apply(ns, state, lhs_field, true);
+    INVARIANT(
+      lhs_field.id() == ID_symbol,
+      "member of symbol should be susceptible to field-sensitivity");
+
+    const exprt &rhs_field = rhs.operands()[i];
```
tautschnig, line 401: I think you'd want a `PRECONDITION(rhs.operands().size() == components.size());`

src/goto-symex/symex_assign.cpp
```diff
@@ -364,6 +364,50 @@ static void shift_indexed_access_to_lhs(
...
+  const ssa_exprt &lhs, // L1
+  const exprt &full_lhs,
+  const struct_exprt &rhs,
+  exprt::operandst &guard,
+  assignment_typet assignment_type)
+{
+  const struct_typet &type = to_struct_type(ns.follow(lhs.type()));
+  const struct_union_typet::componentst &components = type.components();
+
+  for(std::size_t i = 0; i < components.size(); ++i)
```
tautschnig, line 392: How about using a ranged for and an iterator over `rhs.operands()`?

tautschnig (approved):

src/goto-symex/symex_assign.cpp
```diff
@@ -364,6 +364,50 @@ static void shift_indexed_access_to_lhs(
   }
 }
 
+/// Assign a struct expression to a symbol. If \ref symex_assign_symbol was used
+/// then we would assign the whole symbol, before extracting its components,
+/// with results like `x = {1, 2}; x..field1 = x.field1; x..field2 = x.field2;`
+/// This abbreviates the process, directly producing
+/// `x..field1 = 1; x..field2 = 2;`
+/// \param state: goto-symex state
+/// \param lhs: symbol to assign
```
owen-mc-diffblue: You should mention that it should be L1-renamed already here

owen-mc-diffblue (approved):

PR #4569, opened 2019-04-25
WIP Disable RTTI [DO NOT REVIEW]

chrisr-diffblue:
Building on #2826 at Daniels request. Currently a WIP so do not review at the moment.
On the Mac, this reduces the size of the binary by 200k.

### DIFF
--- a/jbmc/src/java_bytecode/java_qualifiers.cpp
+++ b/jbmc/src/java_bytecode/java_qualifiers.cpp
@@ -56,7 +56,7 @@ void java_qualifierst::write(typet &src) const
 qualifierst &java_qualifierst::operator+=(const qualifierst &other)
 {
   c_qualifierst::operator+=(other);
-  auto jq = dynamic_cast<const java_qualifierst *>(&other);
+  auto jq = static_cast<const java_qualifierst *>(&other);
   if(jq != nullptr)
   {
     std::copy(
@@ -69,7 +69,7 @@ qualifierst &java_qualifierst::operator+=(const qualifierst &other)
 
 bool java_qualifierst::operator==(const qualifierst &other) const
 {
-  auto jq = dynamic_cast<const java_qualifierst *>(&other);
+  auto jq = static_cast<const java_qualifierst *>(&other);
   if(jq == nullptr)
     return false;
   return c_qualifierst::operator==(other) && annotations == jq->annotations;
@@ -79,7 +79,7 @@ bool java_qualifierst::is_subset_of(const qualifierst &other) const
 {
   if(!c_qualifierst::is_subset_of(other))
     return false;
-  auto jq = dynamic_cast<const java_qualifierst *>(&other);
+  auto jq = static_cast<const java_qualifierst *>(&other);
   if(jq == nullptr)
     return annotations.empty();
   auto &other_a = jq->annotations;
--- a/jbmc/src/jbmc/jbmc_parse_options.cpp
+++ b/jbmc/src/jbmc/jbmc_parse_options.cpp
@@ -691,7 +691,7 @@ int jbmc_parse_optionst::get_goto_program(
     if(goto_model_ptr == nullptr)
       return CPROVER_EXIT_INTERNAL_ERROR;
 
-    goto_modelt &goto_model = dynamic_cast<goto_modelt &>(*goto_model_ptr);
+    goto_modelt &goto_model = static_cast<goto_modelt &>(*goto_model_ptr);
 
     if(cmdline.isset("validate-goto-model"))
     {
--- a/src/analyses/constant_propagator.cpp
+++ b/src/analyses/constant_propagator.cpp
@@ -143,8 +143,8 @@ void constant_propagator_domaint::transform(
   // information about dirty variables and config flags are
   // available. Otherwise, the below will be null and we use default
   // values
-  const constant_propagator_ait *cp=
-    dynamic_cast<constant_propagator_ait *>(&ai);
+  const constant_propagator_ait *cp =
+    static_cast<constant_propagator_ait *>(&ai);
   bool have_dirty=(cp!=nullptr);
 
   // Transform on a domain that is bottom is possible
--- a/src/analyses/goto_rw.h
+++ b/src/analyses/goto_rw.h
@@ -138,7 +138,9 @@ public:
 
   const range_domaint &get_ranges(objectst::const_iterator it) const
   {
+#if 0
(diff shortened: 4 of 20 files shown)


jbmc/src/java_bytecode/java_qualifiers.cpp
```diff
@@ -56,7 +56,7 @@ void java_qualifierst::write(typet &src) const
 qualifierst &java_qualifierst::operator+=(const qualifierst &other)
 {
   c_qualifierst::operator+=(other);
-  auto jq = dynamic_cast<const java_qualifierst *>(&other);
+  auto jq = static_cast<const java_qualifierst *>(&other);
```
hannes-steffenhagen-diffblue, line 59: :no_entry_sign: 

This doesn't work with the following logic, because it checks for the failure of the dynamic cast. `static_cast` cannot fail in this way.

jbmc/src/java_bytecode/java_qualifiers.cpp
```diff
@@ -69,7 +69,7 @@ qualifierst &java_qualifierst::operator+=(const qualifierst &other)
 
 bool java_qualifierst::operator==(const qualifierst &other) const
 {
-  auto jq = dynamic_cast<const java_qualifierst *>(&other);
+  auto jq = static_cast<const java_qualifierst *>(&other);
```
hannes-steffenhagen-diffblue, line 72: :no_entry_sign: ditto

jbmc/src/java_bytecode/java_qualifiers.cpp
```diff
@@ -79,7 +79,7 @@ bool java_qualifierst::is_subset_of(const qualifierst &other) const
 {
   if(!c_qualifierst::is_subset_of(other))
     return false;
-  auto jq = dynamic_cast<const java_qualifierst *>(&other);
+  auto jq = static_cast<const java_qualifierst *>(&other);
```
hannes-steffenhagen-diffblue, line 82: :no_entry_sign: ditto

jbmc/src/jbmc/jbmc_parse_options.cpp
```diff
@@ -691,7 +691,7 @@ int jbmc_parse_optionst::get_goto_program(
     if(goto_model_ptr == nullptr)
       return CPROVER_EXIT_INTERNAL_ERROR;
 
-    goto_modelt &goto_model = dynamic_cast<goto_modelt &>(*goto_model_ptr);
+    goto_modelt &goto_model = static_cast<goto_modelt &>(*goto_model_ptr);
```
hannes-steffenhagen-diffblue, line 694: Moving from checked to unchecked :(

src/analyses/constant_propagator.cpp
```diff
@@ -143,8 +143,8 @@ void constant_propagator_domaint::transform(
   // information about dirty variables and config flags are
   // available. Otherwise, the below will be null and we use default
   // values
-  const constant_propagator_ait *cp=
-    dynamic_cast<constant_propagator_ait *>(&ai);
+  const constant_propagator_ait *cp =
+    static_cast<constant_propagator_ait *>(&ai);
```
hannes-steffenhagen-diffblue, line 147: :no_entry_sign: Another check that doesn't work now

src/analyses/goto_rw.h
```diff
@@ -138,7 +138,9 @@ class rw_range_sett
 
   const range_domaint &get_ranges(objectst::const_iterator it) const
   {
+#if 0
```
hannes-steffenhagen-diffblue, line 141: I'd prefer if this came with documentation as to why it was commented out

src/analyses/goto_rw.h
```diff
@@ -362,8 +364,10 @@ class rw_guarded_range_set_value_sett:public rw_range_set_value_sett
 
   const guarded_range_domaint &get_ranges(objectst::const_iterator it) const
   {
+#if 0
```
hannes-steffenhagen-diffblue, line 367: same

src/analyses/reaching_definitions.cpp
```diff
@@ -71,12 +71,8 @@ void rd_range_domaint::transform(
...
   const namespacet &ns)
 {
-  reaching_definitions_analysist *rd=
-    dynamic_cast<reaching_definitions_analysist*>(&ai);
-  INVARIANT_STRUCTURED(
-    rd!=nullptr,
-    bad_cast_exceptiont,
-    "ai has type reaching_definitions_analysist");
+  reaching_definitions_analysist *rd =
+    static_cast<reaching_definitions_analysist *>(&ai);
```
hannes-steffenhagen-diffblue, line 75: another unchecked cast where we had a check before :(

src/goto-checker/solver_factory.cpp
```diff
@@ -83,17 +83,7 @@ void solver_factoryt::set_prop_conv_time_limit(prop_convt &prop_conv)
...
-    if(solver == nullptr)
-    {
-      messaget log(message_handler);
-      log.warning() << "cannot set solver time limit on "
-                    << prop_conv.decision_procedure_text() << messaget::eom;
-      return;
-    }
-
-    solver->set_time_limit_seconds(timeout_seconds);
+    prop_conv.set_time_limit_seconds(timeout_seconds);
```
hannes-steffenhagen-diffblue, line 86: This returns a `bool` now but the return value isn't used. Presumably it should emit the previous warning if this fails?

src/solvers/prop/solver_resource_limits.h
```diff
@@ -16,7 +16,7 @@ class solver_resource_limitst
 {
 public:
   /// Set the limit for the solver to time out in seconds
-  virtual void set_time_limit_seconds(uint32_t) = 0;
+  virtual bool set_time_limit_seconds(uint32_t) = 0;
```
hannes-steffenhagen-diffblue, line 19: return value should be documented at least, or better yet we could use a status enum to indicate whether this succeeded or failed (because we use `bool` for success/failure inconsistently throughout the code base, but most of the time `true` means _failure_ which is the opposite to how it works here).

Also, should the `return false;` implementation here be the default?

hannes-steffenhagen-diffblue (changes requested):
I'm not a particularly big fan of this move. Basically, a lot of places here would require us to implement "homebrew" versions of RTTI (via the normal virtual dispatch I guess). Either way if we're disabling the standard RTTI we need replacements for the cases where we do rely on it at the moment.

PR #4570, opened 2019-04-25
Use value_comparatort in replace() in the sharing map

danpoe:

### DIFF
--- a/src/util/sharing_map.h
+++ b/src/util/sharing_map.h
@@ -201,18 +201,6 @@ protected:
   typedef typename innert::to_mapt to_mapt;
   typedef typename innert::leaf_listt leaf_listt;
 
-  struct falset
-  {
-    bool operator()(const mapped_type &lhs, const mapped_type &rhs)
-    {
-      return false;
-    }
-  };
-
-  typedef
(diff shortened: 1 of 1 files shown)


tautschnig (approved):

smowton (changes requested):
I didn't do this for `update` in the first place because it needlessly takes a copy of the thing you're updating, which will be pretty costly for `value_sett`, whose `mapped_type` is pretty heavy

tautschnig:
@smowton Which operation is it that takes a copy? It seems all the operations involved in this change are constant, so really all of them should be taking const references?

smowton:
The comparator stores a copy of its first argument. This was so we can provide an (expensive) debug mode for in-place updates, where we copy the value before the update and then check it has changed afterwards. Compare `replace`, where we have old and new values available simultaneously.

danpoe:
@smowton I thought it didn't matter that `value_comparatort` takes an extra copy in `replace()`, since it would only happen in debug mode.

smowton:
If you like, personally I'd prefer if even debug mode didn't crawl if it can be avoided

PR #4572, opened 2019-04-26
Add getter/setter for ID_enumeration in java_class_typet

jeannielynnmoulton:
Adding getter/setter to avoid direct irept access

- [ x] Each commit message has a non-empty body, explaining why the change was made.
- [x ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] ~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~
- [ ] ~Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).~
- [ ] ~My commit message includes data points confirming performance improvements (if claimed).~
- [ x] My PR is restricted to a single feature or bugfix.

### DIFF
--- a/jbmc/src/java_bytecode/java_bytecode_convert_class.cpp
+++ b/jbmc/src/java_bytecode/java_bytecode_convert_class.cpp
@@ -332,7 +332,7 @@ void java_bytecode_convert_classt::convert(
     class_type.set(
       ID_java_enum_static_unwind,
       std::to_string(c.enum_elements+1));
-    class_type.set(ID_enumeration, true);
+    class_type.set_is_enumeration(true);
   }
 
   if(c.is_public)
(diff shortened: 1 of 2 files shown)


kroening:
Assuming that this is to determine that a class is an enum type, why not simply check that the base is `java.lang.enum`?

jeannielynnmoulton:
This is very much possible. There are a few places where we check `get_bool(ID_enumeration)` from a typet, so I thought that it might be more elegant to have the solution here than to check the base each time (or create a function outside of JBMC that does it)

antlechner:
@jeannielynnmoulton Some of those checks might have been added by me. I actually started with checks for "is the base `java.lang.Enum`" and at some point it was requested in a review to check the `ID_enumeration` instead, but I don't know if one way would really be better than the other.

thomasspriggs (approved):
:+1: An improvement. It would be nice if we had a catch test that checked the flag is set properly in the parsing and class conversion.

smowton (approved):

PR #4573, opened 2019-04-26
Split out is_in_conflict solver capability [blocks: 4577]

peterschrammel:
This is only provided by the prop_conv_solvert-based hierarchy
at the moment and is quite specific to MiniSAT-based solvers.
The functionality itself is used out-of-tree only (2LS).

### DIFF
--- a/src/solvers/conflict_provider.h
+++ b/src/solvers/conflict_provider.h
@@ -1,0 +1,30 @@
+/*******************************************************************\
+
+Module: Capability to check whether an expression is a reason for
+        the solver returning UNSAT
+
+Author: Peter Schrammel
+
+\*******************************************************************/
+
+/// \file
+/// Capability to check whether an expression is a reason for
+/// the solver returning UNSAT
+
+#ifndef CPROVER_SOLVERS_CONFLICT_PROVIDER_H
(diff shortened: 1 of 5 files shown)


src/solvers/conflict_provider.h
```diff
@@ -0,0 +1,30 @@
...
+
+class exprt;
+
+class conflict_providert
+{
+public:
+  /// Returns true if an expression is in the final conflict leading to UNSAT.
+  /// The argument can be a Boolean expression or something more
+  /// solver-specific, e.g. a `literal_exprt`.
+  virtual bool is_in_conflict(const exprt &) const = 0;
```
tautschnig, line 25: Is it really worth making it this general when really we just have support for this in `prop_conv_solvert`? I'd vote for taking a `literal_exprt` here instead.

tautschnig (approved):

src/solvers/conflict_provider.h
peterschrammel: I'd agree with that, but then I could as well leave it as `literalt`. Since we go for the more general `exprt` interfaces now, I think there is no harm keeping it more general.

PR #4574, opened 2019-04-26
goto-symex-is-constant: treat `x * sizeof(t)` and `sizeof(t) * x` alike

smowton:
Previously it would refuse and allow constant propagation respectively. (Needs a test)

### DIFF
--- a/src/goto-symex/goto_symex_is_constant.h
+++ b/src/goto-symex/goto_symex_is_constant.h
@@ -22,16 +22,18 @@ protected:
   {
     if(expr.id() == ID_mult)
     {
+      bool found_non_constant = false;
+
       // propagate stuff with sizeof in it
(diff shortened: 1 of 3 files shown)


tautschnig (approved):
A test should probably just look at the output of `--program-only` or `--show-vcc`.

owen-mc-diffblue (approved):

smowton:
@tautschnig turned out I can only unit-test this in mainline cbmc as its simplifier always wants to swap the operand order so the constant (sizeof) comes first. Still, that seems like a brittle requirement, so I've made a unit test to check that ordering isn't required.

tautschnig:
@smowton I feared as much, but this PR helps us stay out of trouble in light of the existing PRs that disable sorting in the simplifier. Thank you for adding a unit test!

PR #4575, opened 2019-04-26
value-set-dereference: factor out should_ignore_value

smowton:
Rather than build_reference_to returning a tuple which may indicate a value is ignored,
seperately test whether it ought to be ignored beforehand and then build references to those
values which are not ignored.

### DIFF
--- a/src/pointer-analysis/value_set_dereference.cpp
+++ b/src/pointer-analysis/value_set_dereference.cpp
@@ -67,26 +67,23 @@ exprt value_set_dereferencet::dereference(const exprt &pointer)
 
   // get the values of these
 
+  std::vector<exprt> retained_values;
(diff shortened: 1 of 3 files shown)


src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -77,16 +84,14 @@ exprt value_set_dereferencet::dereference(const exprt &pointer)
     valuet value =
       build_reference_to(*it, pointer, exclude_null_derefs, language_mode, ns);
```
tautschnig: This call needs to be amended as it is one of the reasons of the CI failures (Doxygen and clang-format are also unhappy).

owen-mc-diffblue (approved):

peterschrammel (approved):

PR #4576, opened 2019-04-26
Use let-expression for dereferenced pointers

smowton:
~Depends on #4574 and #4575.~ Since #4555 didn't really work out, here's my second stab at avoiding producing comically large expressions during nested dereferencing. 

In brief, if we have `**p` where `p` is the root of a binary tree with nodes L, R, LL, LR, RL, RR, then instead of `*p` -> `p == &L ? L : R` and then `**p` -> `(p == &L ? L : R) == LL ? LL : (p == &L ? L : R) == LR ? LR : ...` instead we get `let derefd = (p == &L ? L : R) in derefd == LL ? LL : derefd = LR ? LR : ...`. When we have a large alias set this avoids creating many duplicates of the inner dereference result.

Anecdotally this speeds up a nested deref due to a two-dimensional array from "practically forever" to "around a second".

~Putting up now for early review, but known TODOs:~
* ~Data~
* ~Check whether the assignments associated with a let expression's temporary variable ought to be `STATE` or `HIDDEN` assignments.~ (gone for HIDDEN, given that `#return_value` assignments use the same scheme)
* ~Forgo a `let` in more cases when the existing pointer expression is small / simple (e.g. if `pointer` is `(T*)p` there's not really any point in creating a temporary for it)~ (allowed expressions only involving one `symbol_exprt` to be used directly, such as `(type *)p` and `p[some_constant]`)
* ~Kill the temporaries created for `let` bound variables when they go out of scope, to avoid pointlessly carrying them around in `value_sett`, the constant propagator, etc for the rest of time~ ~(added an owning object returned by `clean_expr` that kills the binders after the instruction is done executing)~ (added a tracking vector of instruction-local bound variables, which are killed at the end of `symex_step`)

### DIFF
--- a/src/goto-symex/symex_clean_expr.cpp
+++ b/src/goto-symex/symex_clean_expr.cpp
@@ -186,6 +186,9 @@ void goto_symext::lift_let(statet &state, const let_exprt &let_expr)
     let_value,
     value_assignment_guard,
     symex_targett::assignment_typet::HIDDEN);
+
+  // Schedule the bound variable to be cleaned up at the end of symex_step:
+  instruction_local_symbols.push_back(let_expr.symbol());
 }
 
 void goto_symext::lift_lets(statet &state, exprt &rhs)
--- a/src/goto-symex/symex_clean_expr.cpp
+++ b/src/goto-symex/symex_clean_expr.cpp
@@ -128,6 +128,7 @@ void goto_symext::process_array_expr(statet &state, exprt &expr)
     ns, state.symbol_table, symex_dereference_state, language_mode, false);
 
   expr = dereference.dereference(expr);
+  lift_lets(state, expr, symex_targett::assignment_typet::STATE);
 
   ::process_array_expr(expr, symex_config.simplify_opt, ns);
 }
@@ -222,6 +223,7 @@ void goto_symext::clean_expr(
 {
   replace_nondet(expr, path_storage.build_symex_nondet);
   dereference(expr, state, write);
+  lift_lets(state, expr, symex_targett::assignment_typet::STATE);
 
   // make sure all remaining byte extract operations use the root
   // object to avoid nesting of with/update and byte_update when on
--- a/src/goto-symex/symex_clean_expr.cpp
+++ b/src/goto-symex/symex_clean_expr.cpp
@@ -14,6 +14,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/arith_tools.h>
 #include <util/byte_operators.h>
 #include <util/c_types.h>
+#include <util/expr_iterator.h>
 #include <util/pointer_offset_size.h>
 #include <util/simplify_expr.h>
 
@@ -166,6 +167,51 @@ replace_nondet(exprt &expr, symex_nondet_generatort &build_symex_nondet)
   {
     Forall_operands(it, expr)
       replace_nondet(*it, build_symex_nondet);
+  }
+}
+
+void goto_symext::lift_let(
+  statet &state,
+  const let_exprt &let_expr,
+  assignment_typet assignment_type)
+{
+  exprt let_value = let_expr.value();
+  clean_expr(let_value, state, false);
+  let_value = state.rename(std::move(let_value), ns).get();
+  do_simplify(let_value);
+
+  exprt::operandst value_assignment_guard;
+  symex_assign_symbol(
+    state,
+    to_ssa_expr(state.rename<L1>(let_expr.symbol(), ns).get()),
+    nil_exprt(),
+    let_value,
+    value_assignment_guard,
+    assignment_type);
+}
+
+void goto_symext::lift_lets(
+  statet &state,
+  exprt &rhs,
+  assignment_typet assignment_type)
+{
+  for(auto it = rhs.depth_begin(), itend = rhs.depth_end(); it != itend;)
+  {
+    if(it->id() == ID_let)
+    {
+      // Visit post-order, so more-local definitions are made before usage:
+      exprt &replaced_expr = it.mutate();
+      let_exprt &replaced_let = to_let_expr(replaced_expr);
+      lift_lets(state, replaced_let.value(), assignment_type);
+      lift_lets(state, replaced_let.where(), assignment_type);
+
+      lift_let(state, replaced_let, assignment_type);
+      replaced_expr = replaced_let.where();
+
+      it.next_sibling_or_parent();
+    }
+    else
+      ++it;
   }
 }
 
--- a/src/goto-symex/symex_clean_expr.cpp
+++ b/src/goto-symex/symex_clean_expr.cpp
@@ -128,7 +128,7 @@ void goto_symext::process_array_expr(statet &state, exprt &expr)
     ns, state.symbol_table, symex_dereference_state, language_mode, false);
 
   expr = dereference.dereference(expr);
-  lift_lets(state, expr, symex_targett::assignment_typet::STATE);
+  lift_lets(state, expr);
 
   ::process_array_expr(expr, symex_config.simplify_opt, ns);
 }
@@ -171,10 +171,7 @@ replace_nondet(exprt &expr, symex_nondet_generatort &build_symex_nondet)
   }
 }
 
-void goto_symext::lift_let(
-  statet &state,
-  const let_exprt &let_expr,
-  assignment_typet assignment_type)
+void goto_symext::lift_let(statet &state, const let_exprt &let_expr)
 {
   exprt let_value = let_expr.value();
   clean_expr(let_value, state, false);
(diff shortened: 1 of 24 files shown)


smowton:
@tautschnig @kroening @peterschrammel TODOs done -- only data-gathering remains.

smowton:
Here's a poster-child example:

```c
#include <assert.h>
#include <stdlib.h>

  int **pptrs[50];

  // Set up a straightforward vector of double-derefs that lead to
  // seperate integer allocations, but do so in a complicated-enough
  // way to avoid constant propagation solving the whole thing (in a
  // real example some user-controlled control-flow writing to one or
  // other of the arrays would provide the corresponding breakage):
  for(int i = 0; i < 50; ++i) {
    int **new_ptrs = malloc(2 * sizeof(int*));
    new_ptrs[argc % 2] = (int*)malloc(sizeof(int) * 2);
    new_ptrs[argc % 2][0] = 1;
    new_ptrs[argc % 2][1] = 2;    
    new_ptrs[1 - (argc % 2)] = (int*)malloc(sizeof(int) * 2);
    new_ptrs[1 - (argc % 2)][0] = 3;
    new_ptrs[1 - (argc % 2)][1] = 4;    

    pptrs[i % argc] = new_ptrs;
  }

  // Now just try to read from the pointers -- this will generate a
  // pretty big formula as the nested deref could point to any of 50
  // objects, and then the outer deref must condition on its result to
  // determine where the inner pointer leads.

  assert(pptrs[argc % 50][argc % 2][argc % 2] == 5);
}
```

Interestingly the multiple levels of arrays are necessary -- if straightforward derefs are used instead of array indices then this code (https://github.com/diffblue/cbmc/blob/develop/src/pointer-analysis/value_set_dereference.cpp#L40) recurses into the nested `if_exprt` and thus avoids the quadratic case. However it's pretty brittle -- an intervening pointer addition, as we get from an index operator, is enough to break it. It could be sufficient to improve that code, though I think this would boil down to rewriting `value_sett::get_value_set_rec` to special-case `if_exprt` wherever it occurs in the syntax tree.

Result: 
With develop, 4.7MB of VCCs (printed with --show-vcc), 8 seconds to run symex
With this patch, 337KB of VCCs, 0.3 seconds to run symex

Interestingly for this case the solver takes longer with the patch than without (26 seconds vs. 10), despite having very similar clause / variable counts after the simplifier / merge-ireps has got at them. This might be luck or might be systematic; if it is then I'd expect the main win here to come when the assertion violation we end up diagnosing doesn't directly depend on the big deref, or when there are a lot of pointer references meaning symex would be unlikely to finish without this patch.

smowton:
To elaborate on the remark about nested `if_exprt`s: the special case in `value_set_dereference` can handle directly nested dereferences, so if the first deref makes `*(pp == &o1 ? o1 : o2)` then we effectively push the deref in to make `(pp == &o1 ? *o1 : *o2)`, producing much smaller value-sets for each sub-deref. However if the `if_exprt` is not the topmost node in the syntax tree (e.g. from an array access, represented as `*(p + offset)`) then we use `get_value_set_rec`'s understanding of `if_exprt`, which simply sums the cases.

So, possibilities:
1. Use this patch
2. Push deref inside `if_exprt` more aggressively, especially inside the patterns generated by array accesses

Advantage of 1: flexible -- the kind of expression is irrelevant, all pointer expressions use let binders
Advantage of 2: less churn, and lower risk of unintentional performance effects

src/goto-symex/symex_dead.cpp
```diff
@@ -21,8 +21,17 @@ void goto_symext::symex_dead(statet &state)
   const goto_programt::instructiont &instruction=*state.source.pc;
 
   const code_deadt &code = instruction.get_dead();
+  symex_dead(state, code.symbol());
+}
 
-  ssa_exprt ssa = state.rename_ssa<L1>(ssa_exprt{code.symbol()}, ns).get();
+void goto_symext::symex_dead(statet &state, const symbol_exprt &symbol_expr)
+{
+  ssa_exprt ssa = state
```
peterschrammel: ugly clang-format

smowton:
Ran benchmarks on Romain's suite -- overall symex runtimes got worse by about 3-4%, though there were a couple of cases that got significantly faster. Will benchmark including the solver on Friday. I suspect the specific deref-of-deref hack is performing better than a let-bound temporary when it works correctly, so I should probably revise this to continue using that workaround when possible, and only resort to let-binding when the situation is more complicated.

smowton:
In fact it seems all the variation observed on that particular benchmark was due to the vagaries of Google's cloud, not any actual difference, as none of them reproduced on further inspection. Given this sort of thing only manifests in Java for multi-dimensional arrays it wouldn't be too surprising if the selection of Tika methods never uses the changed code.

src/goto-symex/symex_clean_expr.cpp
```diff
@@ -169,6 +170,51 @@ replace_nondet(exprt &expr, symex_nondet_generatort &build_symex_nondet)
...
 }
 
+void goto_symext::lift_let(
+  statet &state,
+  const let_exprt &let_expr,
+  assignment_typet assignment_type)
+{
+  exprt let_value = let_expr.value();
+  clean_expr(let_value, state, false);
+  let_value = state.rename(std::move(let_value), ns).get();
```
tautschnig, line 178: Just as a note: some time ago @kroening proposed to make `let` expressions be a scope of their own. Thus care would need to be taken as the `let` expression could have symbols of local scope even when the name overlaps with something occurring elsewhere. Not the case right now, but needs care moving forward.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -30,6 +31,30 @@ Author: Daniel Kroening, kroening@kroening.com
...
 #include <util/ssa_expr.h>
 
+/// Returns true if \p expr is complicated enough that a local definition (using
+/// a let expression) is preferable to repeating it, potentially many times.
+/// Of course this is just a heuristic -- currently we allow any expression that
+/// only involves one symbol, such as "x", "(type*)x", "x[0]" (but not "x[y]").
+/// Particularly we want to make sure to insist on a local definition of \p expr
+/// is a large if-expression, such as `p == &o1 ? o1 : p == &o2 ? o2 : ...`, as
+/// can result from dereferencing a subexpression.
+static bool should_use_local_definition_for(const exprt &expr)
```
tautschnig, line 44: At least the implementation could just be replaced by using `has_subexpr(ID_symbol)`, but maybe you want to remove the function altogether.

src/util/nodiscard.h
```diff
@@ -0,0 +1,26 @@
+/*******************************************************************\
+
+Module: No-discard-result macro
+
+Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+/// \file
+/// No-discard-result macro
```
tautschnig: Seems independent and might warrant a PR of its own?

src/goto-symex/symex_function_call.cpp
```diff
@@ -200,13 +200,14 @@ void goto_symext::symex_function_call_symbol(
 {
   code_function_callt code = original_code;
 
+  symex_live_let_variablest live_let_variables(*this, state);
```
tautschnig: Hmm, when exactly do we (not) need a new such context? Can we possibly reuse frames for this purpose? It's another bit of state to track and the interplay with branching I'm not entirely sure of.

smowton:
@tautschnig please re-review, have replaced the owning object with a `goto_symext`-owned vector `instruction_local_symbols`

src/pointer-analysis/value_set_dereference.cpp
smowton: No, this looks for *multiple* symbol expressions, looking to permit `x`, `(int*)x` etc but not `x[y]` or `*(x + (y ? z : w))` or similar

smowton:
@tautschnig I added tests that check let-exprs are / are not introduced in several particular circumstances. I then noticed that one of the particular cases that's currently handled badly, `*(int*)*p`, is easy to fix by analogy with the existing special case for `**p`, so I added an extra small commit fixing that.

I also have a fix for the `x->y->z` case that executes that neatly in one expression without a let-binder or expression size blowup, but it's more controversial as it will require moving `simplify` calls around, so I'll post that as a followup.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -76,9 +124,24 @@ exprt value_set_dereferencet::dereference(const exprt &pointer)
 
   std::list<valuet> values;
 
+  exprt compare_against_pointer = pointer;
+
+  if(retained_values.size() >= 2 && should_use_local_definition_for(pointer))
+  {
+    symbolt fresh_binder = get_fresh_aux_symbol(
+      pointer.type(),
+      "derefd_pointer",
```
peterschrammel, line 133: Should use the function_id here

peterschrammel (approved):

martin-cs:
Seems to be the right way of going about solving this; will review in detail when I have time.

src/pointer-analysis/value_set_dereference.cpp
smowton: I think not, the temporary should not be in the same namespace as the function's locals. Could just be "symex" though, like the namespace occupied by invalid-object symbols

tautschnig (approved):
Looks good to me. @smowton Could you please post some performance evaluation results? You did post some earlier, but there have been significant changes to this PR since. Feel free to merge once you've provided that additional data.

smowton:
Total process-seconds to run the cbmc CORE tests: 700s (-20)
Time to run my large Webgoat example: neither example terminates, but with this patch we get much further in the same time

smowton:
Heh, I actually have another benchmark in progess... will post it here when it finishes anyhow

smowton:
Good news! It was also a neutral finding.

PR #4578, opened 2019-04-27
Dynamic memory snapshot

xbauch:
Enable tracking information about dynamically allocated memory from `memory-analyzer` across to `goto-harness` via memory snapshots. The knowledge about what addresses have been allocated dynamically (and the scope of these allocation in byte size) is collected by breaking the execution at `malloc` call-sites and inspecting the registers of these function calls.

### DIFF
--- a/regression/memory-analyzer/pointer_to_struct_01/test.desc
+++ b/regression/memory-analyzer/pointer_to_struct_01/test.desc
@@ -1,8 +1,7 @@
 CORE
 main.gb
 --breakpoint checkpoint --symbols p
-struct S tmp;
-tmp = \{ \.next=\(\(struct S \*\)0\) \};
-p = \&tmp;
+st = \{ .next=\&st \};
+p = \&st;
 ^EXIT=0$
 ^SIGNAL=0$
--- a/regression/snapshot-harness/static-array-float/test.desc
+++ b/regression/snapshot-harness/static-array-float/test.desc
@@ -1,0 +1,15 @@
+FUTURE
+main.c
+array,iterator1,iterator2 --harness-type initialise-with-memory-snapshot --initial-goto-location main:4
+^EXIT=10$
+^SIGNAL=0$
+\[main.assertion.1\] line [0-9]+ assertion \*iterator1 \>= 1.10 \&\& \*iterator1 \<= 1.12: SUCCESS
+\[main.assertion.2\] line [0-9]+ assertion iterator1 != iterator2: SUCCESS
+\[main.assertion.3\] line [0-9]+ assertion iterator2 == \&array\[9\]: SUCCESS
+\[main.pointer_dereference.13\] line [0-9]+ dereference failure: pointer outside object bounds in \*iterator2: FAILURE
+\[main.assertion.4\] line [0-9]+ assertion \*iterator2 == 0: FAILURE
+VERIFICATION FAILED
+--
+unwinding assertion loop \d+: FAILURE
+--
+memory analyzer does not yet allow extract floating point values
--- a/src/goto-cc/Makefile
+++ b/src/goto-cc/Makefile
@@ -40,15 +40,16 @@ INCLUDES= -I ..
 
 LIBS =
 
-CLEANFILES = goto-cc$(EXEEXT) goto-cl$(EXEEXT)
+CLEANFILES = goto-cc$(EXEEXT) goto-gcc$(EXEEXT) goto-cl$(EXEEXT)
 
 include ../config.inc
 include ../common
 
 ifeq ($(BUILD_ENV_),MSVC)
 all: goto-cl$(EXEEXT)
+else
+all: goto-gcc$(EXEEXT)
 endif
-all: goto-cc$(EXEEXT)
 
 ifneq ($(wildcard ../jsil/Makefile),)
   OBJ += ../jsil/jsil$(LIBEXT)
@@ -56,6 +57,9 @@ ifneq ($(wildcard ../jsil/Makefile),)
 endif
 
 ###############################################################################
+
+goto-gcc$(EXEEXT): goto-cc$(EXEEXT)
+	ln -sf goto-cc$(EXEEXT) goto-gcc$(EXEEXT)
 
 goto-cc$(EXEEXT): $(OBJ)
 	$(LINKBIN)
--- a/src/goto-harness/memory_snapshot_harness_generator.cpp
+++ b/src/goto-harness/memory_snapshot_harness_generator.cpp
@@ -207,6 +207,18 @@ size_t memory_snapshot_harness_generatort::pointer_depth(const typet &t) const
     return pointer_depth(t.subtype()) + 1;
 }
 
+bool memory_snapshot_harness_generatort::refers_to(
+  const exprt &expr,
+  const irep_idt &name) const
+{
+  if(expr.id() == ID_symbol)
+    return to_symbol_expr(expr).get_identifier() == name;
+  return std::any_of(
+    expr.operands().begin(),
+    expr.operands().end(),
+    [this, name](const exprt &subexpr) { return refers_to(subexpr, name); });
+}
+
 code_blockt memory_snapshot_harness_generatort::add_assignments_to_globals(
   const symbol_tablet &snapshot,
   goto_modelt &goto_model) const
@@ -230,8 +242,11 @@ code_blockt memory_snapshot_harness_generatort::add_assignments_to_globals(
     ordered_snapshot_symbols.begin(),
     ordered_snapshot_symbols.end(),
     [this](const snapshot_pairt &left, const snapshot_pairt &right) {
-      return pointer_depth(left.second.symbol_expr().type()) <
-             pointer_depth(right.second.symbol_expr().type());
+      if(refers_to(right.second.value, left.first))
+        return true;
+      else
+        return pointer_depth(left.second.symbol_expr().type()) <
+               pointer_depth(right.second.symbol_expr().type());
     });
 
   code_blockt code;
--- a/src/memory-analyzer/analyze_symbol.cpp
+++ b/src/memory-analyzer/analyze_symbol.cpp
@@ -7,6 +7,8 @@ Author: Malte Mues <mail.mues@gmail.com>
 
 \*******************************************************************/
 
+#include <cstdlib>
+
 #include "analyze_symbol.h"
 
 #include <util/c_types.h>
@@ -27,6 +29,73 @@ gdb_value_extractort::gdb_value_extractort(
     c_converter(ns, expr2c_configurationt::clean_configuration),
     allocate_objects(ID_C, source_locationt(), irep_idt{}, this->symbol_table)
 {
+}
+
+gdb_value_extractort::memory_scopet::memory_scopet(
+  const memory_addresst &begin,
+  const mp_integer &byte_size,
+  const irep_idt &name)
+  : begin_int(safe_string2size_t(begin.address_string, 0)),
+    byte_size(byte_size),
+    name(name)
+{
+}
+
+size_t gdb_value_extractort::memory_scopet::address2size_t(
+  const memory_addresst &point) const
+{
+  return safe_string2size_t(point.address_string, 0);
+}
+
+mp_integer gdb_value_extractort::memory_scopet::distance(
+  const memory_addresst &point,
+  mp_integer member_size) const
+{
+  auto point_int = address2size_t(point);
+  CHECK_RETURN(check_containment(point_int));
+  return (point_int - begin_int) / member_size;
+}
+
+std::vector<gdb_value_extractort::memory_scopet>::iterator
+gdb_value_extractort::find_dynamic_allocation(irep_idt name)
+{
+  return std::find_if(
+    dynamically_allocated.begin(),
+    dynamically_allocated.end(),
+    [&name](const memory_scopet &scope) { return scope.id() == name; });
+}
+
+std::vector<gdb_value_extractort::memory_scopet>::iterator
+gdb_value_extractort::find_dynamic_allocation(const memory_addresst &point)
+{
+  return std::find_if(
+    dynamically_allocated.begin(),
+    dynamically_allocated.end(),
+    [&point](const memory_scopet &memory_scope) {
+      return memory_scope.contains(point);
+    });
+}
+
+mp_integer gdb_value_extractort::get_malloc_size(irep_idt name)
+{
+  const auto scope_it = find_dynamic_allocation(name);
+  if(scope_it == dynamically_allocated.end())
+    return 1;
+  else
+    return scope_it->size();
+}
+
+optionalt<std::string> gdb_value_extractort::get_malloc_pointee(
+  const memory_addresst &point,
+  mp_integer member_size)
+{
+  const auto scope_it = find_dynamic_allocation(point);
+  if(scope_it == dynamically_allocated.end())
+    return {};
+
+  const auto pointer_distance = scope_it->distance(point, member_size);
+  return id2string(scope_it->id()) +
+         (pointer_distance > 0 ? "+" + integer2string(pointer_distance) : "");
 }
 
 void gdb_value_extractort::analyze_symbols(const std::vector<irep_idt> &symbols)
--- a/src/memory-analyzer/analyze_symbol.cpp
+++ b/src/memory-analyzer/analyze_symbol.cpp
@@ -290,73 +290,31 @@ exprt gdb_value_extractort::get_pointer_to_member_value(
 
   const symbolt *struct_symbol = symbol_table.lookup(struct_name);
   DATA_INVARIANT(struct_symbol != nullptr, "unknown struct");
-  const auto maybe_struct_size =
-    pointer_offset_size(struct_symbol->symbol_expr().type(), ns);
-  bool found = false;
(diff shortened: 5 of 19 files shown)


src/memory-analyzer/analyze_symbol.cpp
```diff
@@ -29,6 +31,66 @@ gdb_value_extractort::gdb_value_extractort(
 {
 }
 
+bool gdb_value_extractort::memory_scopet::contains(
+  const memory_addresst &point) const
+{
+  size_t begin_int = std::strtoul(begin.address_string.c_str(), NULL, 0);
```
hannes-steffenhagen-diffblue: This is a unchecked conversion effectively (`size_t` is not necessarily the same size as `unsigned long`, even though it's often the case). I'd recommend using `safe_string2size_t` from `string2int.h` (same with all the other string to int conversions here).

src/memory-analyzer/analyze_symbol.h
```diff
@@ -91,6 +91,70 @@ class gdb_value_extractort
   ///   value of `symbol`.
   std::map<memory_addresst, exprt> values;
 
+  struct memory_scopet
+  {
+    memory_addresst begin;
```
hannes-steffenhagen-diffblue: Why not `private`, if these are not arbitrary (as the deleted default constructor implies)?

src/memory-analyzer/analyze_symbol.h
```diff
@@ -91,6 +91,70 @@ class gdb_value_extractort
...
+  std::vector<memory_scopet>::iterator
+  find_dynamic_allocation(const memory_addresst &point);
+
+  /// Search for the size of the allocated memory for \p name
+  /// \param name: name of the pointer used during allocation
+  /// \return the size if have a record of \p name's allocation
+  optionalt<mp_integer> get_malloc_size(irep_idt name);
+
+  /// Build the pointee string for address \p point assuming it points to a
+  ///   dynamic allocation of `n' elements each of size \p member_size. E.g.:
```
hannes-steffenhagen-diffblue, line 178: what's `n` here?

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -230,8 +242,11 @@ code_blockt memory_snapshot_harness_generatort::add_assignments_to_globals(
     ordered_snapshot_symbols.begin(),
     ordered_snapshot_symbols.end(),
     [this](const snapshot_pairt &left, const snapshot_pairt &right) {
-      return pointer_depth(left.second.symbol_expr().type()) <
-             pointer_depth(right.second.symbol_expr().type());
+      if(refers_to(right.second.value, left.first))
```
hannes-steffenhagen-diffblue, line 245: Can you explain what this is for?

src/memory-analyzer/gdb_api.h
```diff
@@ -81,10 +81,10 @@ class gdb_apit
     {
     }
 
-    const memory_addresst address;
-    const std::string pointee;
-    const std::string character;
-    const optionalt<std::string> string;
+    memory_addresst address;
```
hannes-steffenhagen-diffblue, line 91: These fields are all related, so I don't think they should be both be constant and public at the same time.

src/memory-analyzer/analyze_symbol.cpp
```diff
@@ -433,11 +440,22 @@ exprt gdb_value_extractort::get_non_char_pointer_value(
...
 
 bool gdb_value_extractort::points_to_member(
-  const pointer_valuet &pointer_value) const
+  pointer_valuet &pointer_value,
+  const typet &expected_type)
 {
   if(pointer_value.pointee.find("+") != std::string::npos)
     return true;
 
+  if(pointer_value.pointee.empty())
```
hannes-steffenhagen-diffblue, line 438: why is this only done if the `pointee` is empty? What does that mean?

hannes-steffenhagen-diffblue (approved):
Looks ok to me (with the disclaimer that I don't really understand this)

src/memory-analyzer/analyze_symbol.cpp
xbauch: I wasn't sure if the `safe`-conversion would work on hex-numbers.

src/memory-analyzer/analyze_symbol.cpp
xbauch: Well `pointer_value` is the result of the gdb query `get_memory`. When the `pointee` is empty it means that gdb did not know what _structure_ the pointer points to. This can happen if the pointer points to inside a dynamically allocated memory.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: For example, say `int* p = malloc(8); int *q = p+1;` is the source code for which we want to take the memory snapshot. Then the snapshot will contain: `int tmp[2]; int* p = tmp[0]; int* q = tmp[1];` And both `p` and `q` `refer_to` `tmp`. I sort the symbols based on this notion of _reference_ so that when building the harness entry function, `tmp` is assigned before the two pointers.

src/memory-analyzer/analyze_symbol.h
xbauch: `T* point = (T*) malloc(sizeof(T) * n);` .. that `n`.

src/memory-analyzer/analyze_symbol.cpp
hannes-steffenhagen-diffblue: They should, they use `stoll` which basically works the same way as the `str*` functions except it's easier to detect failures.

tautschnig: +1

src/memory-analyzer/analyze_symbol.h
```diff
@@ -91,6 +91,70 @@ class gdb_value_extractort
...
+  struct memory_scopet
+  {
+    memory_addresst begin;
+    mp_integer byte_size;
+    irep_idt name;
+
+    memory_scopet() = delete;
+    memory_scopet(
+      const memory_addresst &begin,
+      mp_integer byte_size,
```
tautschnig: `const mp_integer &byte_size`

src/memory-analyzer/analyze_symbol.h
```diff
@@ -91,6 +91,70 @@ class gdb_value_extractort
...
+      irep_idt name)
```
tautschnig: `const irep_idt &name`

src/memory-analyzer/analyze_symbol.cpp
```diff
@@ -29,6 +31,66 @@ gdb_value_extractort::gdb_value_extractort(
...
+}
+
+mp_integer gdb_value_extractort::memory_scopet::distance(
+  const memory_addresst &point,
+  mp_integer member_size) const
+{
+  CHECK_RETURN(contains(point));
+  size_t begin_int = std::strtoul(begin.address_string.c_str(), NULL, 0);
+  size_t point_int = std::strtoul(point.address_string.c_str(), NULL, 0);
+  return (point_int - begin_int) / member_size;
```
tautschnig, line 56: Shouldn't you first do `CHECK_RETURN(point_int >= begin_int);`?

src/memory-analyzer/analyze_symbol.h
```diff
@@ -91,6 +91,70 @@ class gdb_value_extractort
   ///   value of `symbol`.
   std::map<memory_addresst, exprt> values;
 
+  struct memory_scopet
+  {
+    memory_addresst begin;
+    mp_integer byte_size;
+    irep_idt name;
+
+    memory_scopet() = delete;
```
tautschnig: I'm not sure that's necessary if you explicitly define some other constructor?

src/memory-analyzer/analyze_symbol.cpp
```diff
@@ -112,6 +112,22 @@ void gdb_value_extractort::analyze_symbols(
     CHECK_RETURN(value.pointee.empty() || (value.pointee == id));
 
     values.insert({value.address, symbol_expr});
+
+    const symbolt &symbol = ns.lookup(id);
+    const symbol_exprt actual_expr = symbol.symbol_expr();
```
tautschnig: You can delay this until after the below `if`, because you can just use `symbol.type` instead of `actual_expr.type()`.

src/memory-analyzer/analyze_symbol.cpp
```diff
@@ -112,6 +112,22 @@ void gdb_value_extractort::analyze_symbols(
...
+    if(actual_expr.type().id() != ID_pointer)
+    {
+      memory_map[id2string(id)] = value;
```
tautschnig: I'm not entirely sure this should be indexed by `std::string`, wouldn't `irep_idt` be preferable?

src/memory-analyzer/analyze_symbol.cpp
```diff
@@ -330,13 +331,18 @@ exprt gdb_value_extractort::get_pointer_to_member_value(
...
+      from_integer(
+        member_offset / get_type_size(expr.type().subtype()), index_type())};
   }
 
   const auto it = values.find(memory_location);
   // if the structure we are pointing to does not exists we need to build a
   // temporary object for it: get the type from symbol table, query gdb for
   // value, allocate new object for it and then store into assignments
   if(it == values.end())
+  if(struct_symbol->symbol_expr().type().id() == ID_pointer)
```
tautschnig: This looks very weird?!

src/memory-analyzer/analyze_symbol.cpp
```diff
@@ -303,6 +303,7 @@ exprt gdb_value_extractort::get_pointer_to_member_value(
   }
 
   if(!found)
+  if(struct_symbol->symbol_expr().type().id() == ID_array)
```
tautschnig: Looks weird!?

src/memory-analyzer/analyze_symbol.cpp
```diff
@@ -291,6 +291,8 @@ exprt gdb_value_extractort::get_pointer_to_member_value(
   bool found = false;
   CHECK_RETURN(maybe_struct_size.has_value());
   for(const auto &value_pair : values)
+
+  if(memory_map.count(struct_name) == 0)
```
tautschnig: This looks weird.

regression/snapshot-harness/static-array-float/test.desc
```diff
@@ -0,0 +1,13 @@
+FUTURE
```
tautschnig, line 1: Please explain (after another `--` at the end of the file) what isn't supported here so far.

regression/memory-analyzer/pointer_to_struct_01/test.desc
```diff
@@ -1,8 +1,7 @@
 CORE
 main.exe
 --breakpoint checkpoint --symbols p
-struct S tmp;
-tmp = \{ \.next=\(\(struct S \*\)0\) \};
-p = \&tmp;
+st = \{ .next=\&st \};
+p = \&st;
```
tautschnig, line 5: Could you please squash this into the commit that made this possible?

src/memory-analyzer/analyze_symbol.cpp
xbauch: Done and refactored so that the conversion is computed once in case of `distance`. See: https://github.com/diffblue/cbmc/pull/4578/commits/3812c28519a207a0e5f247f4b7fe2cd4db19af2b.

src/memory-analyzer/analyze_symbol.h
xbauch: Done, also added getters when necessary.

src/memory-analyzer/gdb_api.h
xbauch: Well they are not `const` in the final version.

src/memory-analyzer/analyze_symbol.h
xbauch: Done.

src/memory-analyzer/analyze_symbol.h
xbauch: Done.

src/memory-analyzer/analyze_symbol.cpp
xbauch: That was implied by `contains`. But I refactored this section a bit to avoid duplicate computation of the address-to-string conversion. See: https://github.com/diffblue/cbmc/pull/4578/commits/3812c28519a207a0e5f247f4b7fe2cd4db19af2b.

src/memory-analyzer/analyze_symbol.h
xbauch: You are right. Removed.

src/memory-analyzer/analyze_symbol.cpp
xbauch: Delayed.

src/memory-analyzer/analyze_symbol.cpp
xbauch: Works with `irep_idt` as well. Changed.

src/memory-analyzer/analyze_symbol.cpp
xbauch: What about `struct_symbol->type.id()` ? Changed.

src/memory-analyzer/analyze_symbol.cpp
xbauch: As above.

src/memory-analyzer/analyze_symbol.cpp
xbauch: Wrapped in a helper function `has_known_memory_location`.

regression/snapshot-harness/static-array-float/test.desc
xbauch: Added.

regression/memory-analyzer/pointer_to_struct_01/test.desc
xbauch: It was this one: https://github.com/diffblue/cbmc/pull/4578/commits/33d0a79b6843c2c6c1b60ee8c8540dfe9df9da86. I will squash them together.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 41d398d0955cd267bc7fd331184a007cc0a1202c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113173659

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 356276ae695396ce3c7849e4828d7dc1b840a7ce).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113173817

xbauch:
@tautschnig Our initial approach at querying the GDB for malloc info turned out rather un-portable so we re-implemented the collection this info via breaking at malloc call-sites and parsing the registers (starting at this commit: https://github.com/diffblue/cbmc/pull/4578/commits/e6684572765628814daf1a67397b5d085f31c4b2). The current implementation is still lacking in robustness (we do not track calls of `free`, the name `malloc` is hard-coded, etc.) but I wasn't sure how much more we should include in this PR.

Previously, you reviewed it up to https://github.com/diffblue/cbmc/pull/4578/commits/28ad6d67921b6447ea9d87a8721ce4261b4cfb33. I can do more squashing but I first wanted to ask if our present approach is acceptable and if we should make it more robust in this PR.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 6bb1d2827c55e6e961e6d94f38a3574c53c7c5e8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113173970

src/goto-cc/Makefile
```diff
@@ -57,6 +57,9 @@ endif
 
 ###############################################################################
 
+goto-gcc$(EXEEXT): goto-cc$(EXEEXT)
+	@ln -sf goto-cc$(EXEEXT) goto-gcc$(EXEEXT)
```
tautschnig: I think make has `$(LN)` to improve portability, and I'd suggest not to hide this output, i.e., please remove the `@` - simply because we don't do quiet output elsewhere either.

src/goto-cc/Makefile
```diff
@@ -48,7 +48,7 @@ include ../common
 ifeq ($(BUILD_ENV_),MSVC)
 all: goto-cl$(EXEEXT)
 endif
-all: goto-cc$(EXEEXT)
+all: goto-gcc$(EXEEXT)
```
tautschnig, line 51: Could you please change the lines above so that this is within an `else`/`endif` instead of being done for MSVC as well? Furthermore, please add `goto-gcc$(EXEEXT)` to `CLEANFILES` above.

tautschnig (approved):

tautschnig:
@xbauch My apologies for taking this long! You might want to squash some commits and then just merge. Thank you for all the work!!!

tautschnig:
@xbauch Please rebase now that #4729 is in to make CI happy.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: ba8beb533cbf239a1eb65fc3e2a8be2389f24a7d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113721650

PR #4581, opened 2019-04-28
stack_decision_proceduret for solving under assumptions/contexts

peterschrammel:
Replaces #4451 and #4054 

### DIFF
--- a/src/solvers/prop/prop_conv_solver.cpp
+++ b/src/solvers/prop/prop_conv_solver.cpp
@@ -365,7 +365,7 @@ bool prop_conv_solvert::set_equality_to_true(const equal_exprt &expr)
   return true;
 }
 
-void prop_conv_solvert::set_to(const exprt &expr, bool value)
+void prop_conv_solvert::add_constraints_to_prop(const exprt &expr, bool value)
 {
   PRECONDITION(expr.type().id() == ID_bool);
 
@@ -380,7 +380,7 @@ void prop_conv_solvert::set_to(const exprt &expr, bool value)
     {
       if(expr.operands().size() == 1)
       {
-        set_to(expr.op0(), !value);
+        add_constraints_to_prop(expr.op0(), !value);
         return;
       }
     }
@@ -393,7 +393,7 @@ void prop_conv_solvert::set_to(const exprt &expr, bool value)
         if(expr.id() == ID_and)
         {
           forall_operands(it, expr)
-            set_to_true(*it);
+            add_constraints_to_prop(*it, true);
 
           return;
         }
@@ -437,14 +437,14 @@ void prop_conv_solvert::set_to(const exprt &expr, bool value)
         {
           const implies_exprt &implies_expr = to_implies_expr(expr);
 
-          set_to_true(implies_expr.op0());
-          set_to_false(implies_expr.op1());
+          add_constraints_to_prop(implies_expr.op0(), true);
+          add_constraints_to_prop(implies_expr.op1(), false);
           return;
         }
         else if(expr.id() == ID_or) // !(a || b)  ==  (!a && !b)
         {
           forall_operands(it, expr)
-            set_to_false(*it);
+            add_constraints_to_prop(*it, false);
           return;
         }
       }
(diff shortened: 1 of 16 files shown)


src/solvers/stack_decision_procedure.h
```diff
@@ -0,0 +1,43 @@
...
+
+Module: Decision procedure with incremental solving with contexts
+        and assumptions
+
+Author: Peter Schrammel
+
+\*******************************************************************/
+
+/// \file
+/// Decision procedure with incremental solving with contexts
```
smowton, line 11: Perhaps give an example of intended usage?

src/solvers/prop/prop_conv_solver.cpp
```diff
@@ -365,7 +365,7 @@ bool prop_conv_solvert::set_equality_to_true(const equal_exprt &expr)
   return true;
 }
 
-void prop_conv_solvert::set_to(const exprt &expr, bool value)
+void prop_conv_solvert::do_set_to(const exprt &expr, bool value)
```
smowton: Give this a better name, or provide some documentation to advise a future editor on whether they want `set_to` or `do_set_to`

src/solvers/prop/prop_conv_solver.h
```diff
@@ -124,10 +127,24 @@ class prop_conv_solvert : public conflict_providert,
 
   virtual void ignoring(const exprt &expr);
 
-  // deliberately protected now to protect lower-level API
+  /// Actually adds the constraints to `prop`
```
smowton: Explicitly mention that `set_to` is a wrapper around this, and which one ought to be used

src/solvers/prop/prop_conv_solver.cpp
```diff
@@ -525,3 +525,52 @@ std::size_t prop_conv_solvert::get_number_of_solver_calls() const
...
+  context_size_stack.push_back(1);
+
+  prop.set_assumptions(assumption_stack);
+}
+
+void prop_conv_solvert::pop()
+{
+  // We remove the context from the stack.
+  for(size_t i = 0; i < context_size_stack.back(); ++i)
+    assumption_stack.pop_back();
```
smowton: `assumption_stack.resize(assumption_stack.size() - context_size_stack.back())`?

smowton (approved):

src/solvers/prop/prop_conv_solver.cpp
peterschrammel: Yes, should work. Shrinking resize doesn't reallocate.

src/solvers/prop/prop_conv_solver.cpp
peterschrammel: documentation extended

src/solvers/prop/prop_conv_solver.h
peterschrammel: made private

src/solvers/stack_decision_procedure.h
```diff
@@ -0,0 +1,43 @@
...
+  /// This context becomes a child context nested in the current context.
+  /// An assumption is usually obtained by calling
+  /// `decision_proceduret::handle`. Thus it can be a Boolean expression or
+  /// something more solver-specific such as `literal_exprt`.
+  /// An empty vector of assumptions counts as an element on the stack.
+  virtual void push(const std::vector<exprt> &assumptions) = 0;
+
+  /// Push a new context on the stack
+  /// This context becomes a child context nested in the current context.
+  virtual void push() = 0;
```
tautschnig, line 72: Is this equivalent to a push of an empty vector?

src/solvers/stack_decision_procedure.h
```diff
@@ -0,0 +1,43 @@
...
+
+  /// Pop whatever has been pushed in the last call to `push`
+  virtual void pop() = 0;
```
tautschnig, line 76: I don't think the documentation is accurate: two calls to `pop` wouldn't remove the same element? I think the documentation should refer to the top of the stack. And what's the behaviour when stack underflow would occur? Is an implementation expected to handle this gracefully (by ignoring it), or should it fail hard?

src/solvers/prop/prop_conv_solver.h
```diff
@@ -124,10 +127,24 @@ class prop_conv_solvert : public conflict_providert,
...
+  void do_set_to(const exprt &expr, bool value);
+
   propt &prop;
 
   messaget log;
+
+  static const char *context_prefix;
```
tautschnig, line 134: Can we just initialise it right here rather than in the implementation file?

src/solvers/prop/prop_minimize.cpp
```diff
@@ -120,9 +117,9 @@ void prop_minimizet::operator()()
       {
         _iterations++;
 
-        bvt assumptions;
-        assumptions.push_back(c);
-        prop_conv.set_assumptions(assumptions);
+        exprt::operandst assumptions;
+        assumptions.push_back(literal_exprt(c));
+        prop_conv.push(assumptions);
```
tautschnig: How about simplifying all this to `prop_conv.push({literal_exprt{c}});`?

tautschnig (approved):

src/solvers/stack_decision_procedure.h
peterschrammel: No  - clarified in docs

src/solvers/prop/prop_conv_solver.h
romainbrenguier: `do_set_to` is not a great name, why not naming it `add_constraint_to_prop`?

romainbrenguier (approved):

src/solvers/prop/prop_conv_solver.h
peterschrammel: What would be the advantage of that?

src/solvers/stack_decision_procedure.h
peterschrammel: I've added two examples

src/solvers/prop/prop_conv_solver.h
tautschnig: Fewer lines of code and easier to locate. But clearly a nit-pick and all up to you.

PR #4582, opened 2019-04-28
Make sharing map iteration independent of tree depth

danpoe:
This makes the various methods that iterate over a sharing map independent of the depth of the nodes in the tree. This is in preparation to support variable height trees.

### DIFF
--- a/src/util/sharing_map.h
+++ b/src/util/sharing_map.h
@@ -530,6 +530,11 @@ protected:
 
   void gather_all(const innert &n, delta_viewt &delta_view) const;
 
+  bool is_singular(const leaf_listt &ll) const
+  {
+    return !ll.empty() && std::next(ll.begin()) == ll.end();
+  }
+
   std::size_t count_unmarked_nodes(
     bool leafs_only,
     std::set<const void *> &marked,
@@ -787,7 +792,7 @@ SHARING_MAPT(void)
     return;
   }
 
-  typedef std::tuple<unsigned, const baset *, const baset *> stack_itemt;
+  typedef std::pair<const innert *, const innert *> stack_itemt;
   std::stack<stack_itemt> stack;
 
   // We do a DFS "in lockstep" simultaneously on both maps. For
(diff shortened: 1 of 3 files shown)


src/util/sharing_map.h
```diff
@@ -894,6 +894,8 @@ SHARING_MAPT2(, innert *)::get_container_node(const key_type &k)
     key >>= chunk;
   }
 
+  SM_ASSERT(ip->is_container());
```
tautschnig, line 924: Why is it not `is_defined_container` here?

unit/util/sharing_map.cpp
```diff
@@ -378,6 +378,9 @@ TEST_CASE("Sharing map views and iteration", "[core][util]")
   SECTION("Iterate")
   {
     sharing_map_standardt sm;
+
+    sm.iterate([](const irep_idt &key, const std::string &value) {});
```
tautschnig, line 382: So ... what does this test other than that it compiles?

src/util/sharing_map.h
```diff
@@ -525,11 +525,10 @@ class sharing_mapt
   }
 
   void iterate(
-    const baset &n,
-    const unsigned start_depth,
+    const innert &n,
     std::function<void(const key_type &k, const mapped_type &m)> f) const;
 
-  void gather_all(const baset &n, const unsigned depth, delta_viewt &delta_view)
+  void gather_all(const innert &n, delta_viewt &delta_view)
```
tautschnig: `clang-format` would like the `const;` on this line as well.

src/util/sharing_map.h
```diff
@@ -798,71 +803,100 @@ ::get_delta_view(
...
-          {
-            delta_view.push_back({true, k1, l1.get_value(), p->get_value()});
-          }
-        }
-        else if(!only_common)
+    SM_ASSERT(ip2->is_container());
+
+    for(const auto &l1 : ip1->get_container())
+    {
+      const key_type &k1=l1.get_key();
```
tautschnig: Space around `=`

tautschnig (commented):
Looks ok to me, but given my judgement wasn't exactly spot-on on an earlier PR it would be great if @smowton could please review and approve.

src/util/sharing_map.h
```diff
@@ -798,71 +803,100 @@ ::get_delta_view(
...
+    const innert *ip1 = si.first;
+    const innert *ip2 = si.second;
 
     stack.pop();
 
-    if(depth < steps) // internal
+    SM_ASSERT(!ip1->empty());
+    SM_ASSERT(!ip2->empty());
+
+    if(ip1->is_internal() && ip2->is_container())
```
smowton, line 822: Why only this way around? Also, if a node is internal in tree 1 and a leaf in tree 2, why is it guaranteed to have a single child?

danpoe: The other cases are handled further down in the method.

A container node can only contain more than one element if it is at the very bottom of the tree (which happens when the hash codes of two different keys are the same). Since the corresponding node of the first map is an internal node, the node can't be at the bottom, since the bottom nodes always have to be container nodes.

smowton: Aha, got you. Suggest putting that in a comment!

src/util/sharing_map.h
danpoe: The non-const version of `get_container_node()` adds empty nodes if the key doesn't exist yet in the map. In this case, `!empty()` will not hold.

unit/util/sharing_map.cpp
danpoe: This just checks that calling `iterate()` on an empty map doesn't violate an internal assertion of the map.

src/util/sharing_map.h
danpoe: Ok, added a comment to explain this.

smowton (approved):

PR #4584, opened 2019-04-28
Fix, refactor, move xml_interface

peterschrammel:
- XML_UI was not enabled
- Turn the class into a free function
- Move to xmllang. Enables reuse in other driver programs. It's not CBMC-specific.


tautschnig (commented):
Facilitating reuse seems like a good idea, but:
1) All CI is failing.
2) The class lacks any documentation, thus I'm not sure "reuse" is actually possible? If you have figured out how to (re-)use it, could you please contribute documentation?
3) The implementation uses `null_message_handlert`, which raises alarm bells as there is a risk of hiding errors. I think it should just take a `message_handlert` as an argument.
4) There is a class, but really its only public member is a constructor. Using a free function may be much better and easier, with the two implementation methods just becoming static functions.

smowton (approved):

tautschnig (approved):

PR #4585, opened 2019-04-28
JSON cmdline interface

peterschrammel:
Allows reading the commandline in JSON format from stdin

### DIFF
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -79,6 +79,7 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <langapi/mode.h>
 
+#include <json/json_interface.h>
 #include <xmllang/xml_interface.h>
 
 #include "c_test_input_generator.h"
@@ -90,6 +91,7 @@ cbmc_parse_optionst::cbmc_parse_optionst(int argc, const char **argv)
       argv,
       std::string("CBMC ") + CBMC_VERSION)
 {
+  json_interface(cmdline, ui_message_handler);
   xml_interface(cmdline, ui_message_handler);
 }
 
@@ -103,6 +105,7 @@ cbmc_parse_optionst::cbmc_parse_optionst(int argc, const char **argv)
       argv,
       std::string("CBMC ") + CBMC_VERSION)
 {
+  json_interface(cmdline, ui_message_handler);
   xml_interface(cmdline, ui_message_handler);
 }
 
@@ -1021,6 +1024,7 @@ void cbmc_parse_optionst::help()
     " --xml-ui                     use XML-formatted output\n"
     " --xml-interface              bi-directional XML interface\n"
     " --json-ui                    use JSON-formatted output\n"
+    " --json-interface             bi-directional JSON interface\n"
     // NOLINTNEXTLINE(whitespace/line_length)
     HELP_VALIDATE
     HELP_GOTO_TRACE
--- a/src/json/json_interface.cpp
+++ b/src/json/json_interface.cpp
@@ -1,0 +1,120 @@
+/*******************************************************************\
+
+Module: JSON Commandline Interface
+
+Author: Peter Schrammel
+
+\*******************************************************************/
+
+/// \file
+/// JSON Commandline Interface
+
+#include "json_interface.h"
+
+#include <util/cmdline.h>
+#include <util/exception_utils.h>
+#include <util/json.h>
+#include <util/message.h>
+
+#include "json_parser.h"
+
+#include <iostream>
+
+/// Parse commandline options from \p json into \p cmdline
+static void get_json_options(const json_objectt &json, cmdlinet &cmdline)
+{
+  const jsont &arguments = json["arguments"];
+  if(!arguments.is_array())
+  {
+    throw invalid_command_line_argument_exceptiont(
+      "array expected", "'arguments'");
+  }
+
+  for(const auto &argument : to_json_array(arguments))
+  {
+    if(!argument.is_string())
+    {
+      throw invalid_command_line_argument_exceptiont(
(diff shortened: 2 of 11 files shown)


src/json/json_interface.cpp
```diff
@@ -0,0 +1,101 @@
...
+Author: Peter Schrammel
+
+\*******************************************************************/
+
+/// \file
+/// JSON Commandline Interface
+
+#include "json_interface.h"
+
+#include <iostream>
```
tautschnig: Nit pick: I'd prefer for standard-library includes to come last, see also some recent exchange that @kroening and I had; the idea is that any included header file should parse even without additional includes being present. So do the most minimal ones (the ones from `util`) first, and then by increasing complexity.

src/json/json_interface.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+
+  const jsont &options = json["options"];
+  if(!options.is_object())
+  {
+    throw invalid_command_line_argument_exceptiont(
+      "array expected", "'options'");
+  }
+  for(const auto &option_pair : to_json_object(options))
```
tautschnig, line 51: Nit pick: add a blank line before this one to be consistent with the code a few lines above.

src/json/json_interface.cpp
```diff
@@ -0,0 +1,101 @@
...
+  {
+    throw invalid_command_line_argument_exceptiont(
+      "array expected", "'arguments'");
+  }
+
+  for(const auto &argument : to_json_array(arguments))
+  {
+    if(argument.is_string())
+      cmdline.args.push_back(argument.value);
+    else
```
tautschnig: I believe you could increase readability by doing
```
if(!argument.is_string())
{
  throw ...
}

cmdline.args.push_back(argument.value);
```
because it reduces the number of closing braces that immediately follow each other. This applies even more so down below, where you currently have three closing braces.

src/json/json_interface.cpp
```diff
@@ -0,0 +1,101 @@
...
+    {
+      throw invalid_command_line_argument_exceptiont(
+        "unrecognized commandline option format",
+        option_pair.first,
+        "Boolean, string, number, or string array expected");
+    }
+  }
+}
+
+/// Parse JSON-formatted commandline options from stdin
```
tautschnig: I think this should just be removed - or at least I have no idea what Doxygen makes of documentation at both declaration and definition when they aren't the same.

src/json/json_interface.h
```diff
@@ -0,0 +1,20 @@
...
+
+/// \file
+/// JSON Commandline Interface
+
+#ifndef CPROVER_JSON_JSON_INTERFACE_H
+#define CPROVER_JSON_JSON_INTERFACE_H
+
+class cmdlinet;
+
+/// Parses the JSON-formatted command line from stdin
```
tautschnig, line 18: I think it would be rather important to document the expected JSON syntax! Otherwise you're bound to trial and error...

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1030,6 +1033,7 @@ void cbmc_parse_optionst::help()
     " --xml-ui                     use XML-formatted output\n"
     " --xml-interface              bi-directional XML interface\n"
     " --json-ui                    use JSON-formatted output\n"
+    " --json-interface             bi-directional JSON interface\n"
```
tautschnig, line 1027: Is this actually bi-directional?! Doesn't it only become "bi"directional when using both `--json-interface` and `--json-ui`? (The same is of course true for XML.)

tautschnig (approved):

src/cbmc/cbmc_parse_options.cpp
peterschrammel: I was wondering about that and have now more investigated in the XML case (see comments there). The original intention was that --xml-interface subsumes --xml-ui. 
Actually, the current behaviour was broken because the UI was set before the option could have been parsed from stdin, which seems difficult to fix without splitting the `parse_optionst` constructor into two phases and adding a `set_ui` to `ui_message_handlert` (rather no).
Also, the question is what happens with errors during parsing the options - they should already be output in a proper format. Thus, I'd rather stick with the original intention of 'bi-directional',

tautschnig: I'm not sure I fully understand - I agree that `--{json,xml}-interface` should subsume `--{json,xml}-ui` - but I don't think that was actually work as such? If things can be made to work as such then we'd have the desired bi-directional interface.

peterschrammel: `--json-interface` now implies `--json-ui`, and `--xml-interface` now implies `--xml-ui`.

>but I don't think that was actually work as such?

Not sure, I understand...

tautschnig: Let me retry, my apologies for completely broken English:

1) A "bi"-directional interface is about input _and_ output.
2) `--{json,xml}-ui` set up the _output_ side of an interface.
3) The comment claims that `--{json,xml}-interface` configure a bi-directional interface.
4) The actual implementation, however, only reads _input_ when using `--{json,xml}-interface`.
5) One possible fix would be that upon seeing `--{json,xml}-interface` we also set the option for `--{json,xml}-ui`.
6) This might not actually work as it could be too late, in which case a more extensive change would be required. If this is the case, a simple fix would be to change the comment such as not to promise a bi-directional interface. Users wanting to use a bi-directional interface would then need to provide both `--{json,xml}-interface` _and_ `--{json,xml}-ui`.

peterschrammel: 1. yes
2. yes
3. yes
4. yes
5. is being fixed here (was broken)
6. not required because 5 works.

PR #4590, opened 2019-04-29
Implement parsing of hexadecimal Unicode characters from JSON

antlechner:
Previously, if a JSON file contained a string in hexadecimal Unicode representation, e.g. "\u0001", the JSON parser would discard the "\u" part and store the string as (the four-character string) "0001". This PR fixes this so the resulting string is equal to (the one-character string) "\u0001".

### DIFF
--- a/src/json/parser.y
+++ b/src/json/parser.y
@@ -46,8 +46,15 @@ static std::string convert_TOK_STRING()
       case 'r':  result+='\r'; break;
       case 't':  result+='\t'; break;
       case 'u':
+      {
+        // Character in hexadecimal Unicode representation, in the format
+        // \uABCD, i.e. the following four digits are part of this character.
+        assert(p + 4 < yyjsontext + len - 1);
+        std::string hex(++p, 4);
+        result += std::stoi(hex, nullptr, 16);
+        p += 3;
         break;
-        
+      }
       default:; /* an error */
       }
     }
--- a/unit/json/json_parser.cpp
+++ b/unit/json/json_parser.cpp
@@ -84,4 +84,31 @@ SCENARIO("Loading JSON files")
       }
     }
   }
+  GIVEN("A JSON file containing a hexadecimal Unicode character")
+  {
+    temporary_filet unicode_json_file("cbmc_unit_json_parser_unicode", ".json");
+    const std::string unicode_json_path = unicode_json_file();
+    {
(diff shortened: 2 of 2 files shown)


tautschnig (approved):
cpplint and clang-format both have formatting complaints, otherwise looks good.

romainbrenguier (approved):

src/json/parser.y
```diff
@@ -46,8 +46,15 @@ static std::string convert_TOK_STRING()
       case 'r':  result+='\r'; break;
       case 't':  result+='\t'; break;
       case 'u':
+      {
+        // Character in hexadecimal Unicode representation, in the format
+        // \uABCD, i.e. the following four digits are part of this character.
+        assert(p + 4 < yyjsontext + len - 1);
```
smowton, line 52: Suggest `p + 5 < yyjsontext + len` would be better -- we're checking for 5-byte string uXXXX being in range, after all

src/json/parser.y
```diff
@@ -46,8 +46,15 @@ static std::string convert_TOK_STRING()
...
+        std::string hex(++p, 4);
+        result += std::stoi(hex, nullptr, 16);
+        p += 3;
```
smowton, line 55: Comment that you're intentionally leaving `p` pointing to the last char of the escape

src/json/parser.y
```diff
@@ -46,8 +46,15 @@ static std::string convert_TOK_STRING()
       case 'r':  result+='\r'; break;
       case 't':  result+='\t'; break;
       case 'u':
+      {
+        // Character in hexadecimal Unicode representation, in the format
+        // \uABCD, i.e. the following four digits are part of this character.
+        assert(p + 4 < yyjsontext + len - 1);
+        std::string hex(++p, 4);
+        result += std::stoi(hex, nullptr, 16);
```
smowton, line 54: this casts to a char (8-bit uint), so will only work for unicode characters <= 255

antlechner: This is addressed in a follow-up PR: https://github.com/diffblue/cbmc/pull/4594

src/json/parser.y
antlechner: I found the current version to be clearer: when pointing to `u`, if you advance by four steps (pointing to the last hex character), then you have to still be to the left of the last `"` of the string. But I'm happy to change it in a follow-up PR if you think your version is more intuitive.

PR #4176, opened 2019-02-13
Cleanup expression/code cast validation

tautschnig:
Please review commit-by-commit; several commits are text-motion only and others are automated text replacement - both indicated in the respective commit messages. ~This requires/includes #4175 ("Return statements always have one operand"), so please skip that while reviewing.~

The original aim was to get rid of all the sharing-breaking side effects in `to_*exprt` procedures, but when starting on this I noticed that really a lot more cleanup was indicated.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4593, opened 2019-04-30
Add can_cast_expr specialisation for ssa_exprt

smowton:


tautschnig (changes requested):
This looks fine, but would it be possible to go one step further and apply the same cleanup that happened to std_expr.h and friends in #4176?

smowton:
@tautschnig done, I think

src/util/ssa_expr.h
```diff
@@ -152,33 +152,34 @@ class ssa_exprt:public symbol_exprt
...
+
+/// Cast a generic exprt to an \ref ssa_exprt.
 /// \param expr: Source expression
 /// \return Object of type \ref ssa_exprt
 /// \ingroup gr_std_expr
 inline const ssa_exprt &to_ssa_expr(const exprt &expr)
 {
-  PRECONDITION(
-    expr.id() == ID_symbol && expr.get_bool(ID_C_SSA_symbol) &&
-    !expr.has_operands());
```
tautschnig, line 164: I think we've lost the condition on the absent operands?

src/util/ssa_expr.h
```diff
@@ -152,33 +152,34 @@ class ssa_exprt:public symbol_exprt
...
+{
+  return expr.id()==ID_symbol &&
+         expr.get_bool(ID_C_SSA_symbol);
+}
+
+template <>
+inline bool can_cast_expr<ssa_exprt>(const exprt &base)
+{
+  return is_ssa_expr(base);
+}
```
tautschnig, line 164: It would be great for this to use `ssa_exprt::check` - see `binary_exprt` in std_expr.h for how I think this could best achieve as much code re-use as possible.

smowton: They don't appear to use one another? I think this is small enough that it would be simpler for you to do what you want than to explain it to me.

tautschnig: Sure, my apologies for taking this long: https://github.com/smowton/cbmc/pull/3

tautschnig (approved):
Approving under the assumption that my PR will be merged :-)

PR #4594, opened 2019-04-30
Add support for all Unicode codepoints of the BMP in the JSON parser [TG-7634]

antlechner:
This is a follow-up PR to https://github.com/diffblue/cbmc/pull/4590, which implemented support in the JSON parser for hexadecimal Unicode representations (`\uABCD`) but was limited to codepoints <= `0x7f` (with values from `0x80` to `0xff` also sort of working, representing them as integers rather than characters).
This improved implementation supports all values from `\u0000` to `\uFFFF`, i.e. all codepoints in the BMP.

### DIFF
--- a/src/json/parser.y
+++ b/src/json/parser.y
@@ -51,10 +51,11 @@ static std::string convert_TOK_STRING()
       {
         // Character in hexadecimal Unicode representation, in the format
         // \uABCD, i.e. the following four digits are part of this character.
-        assert(p + 4 < yyjsontext + len - 1);
+        char *last_hex_digit = p + 4;
+        assert(last_hex_digit < yyjsontext + len - 1);
         std::string hex(++p, 4);
         result += codepoint_hex_to_utf8(hex);
-        p += 3;
+        p = last_hex_digit;
         break;
       }
       default:; /* an error */
--- a/src/json/parser.y
+++ b/src/json/parser.y
@@ -18,6 +18,8 @@
 %{
 #include "json_parser.h"
 
+#include <util/unicode.h>
+
 int yyjsonlex();
 extern char *yyjsontext;
 extern int yyjsonleng; // really an int, not a size_t
@@ -51,7 +53,7 @@ static std::string convert_TOK_STRING()
         // \uABCD, i.e. the following four digits are part of this character.
         assert(p + 4 < yyjsontext + len - 1);
         std::string hex(++p, 4);
-        result += std::stoi(hex, nullptr, 16);
+        result += codepoint_hex_to_utf8(hex);
         p += 3;
         break;
       }
--- a/src/util/unicode.cpp
+++ b/src/util/unicode.cpp
@@ -8,11 +8,14 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "unicode.h"
 
-#include <cstring>
-#include <locale>
-#include <iomanip>
-#include <sstream>
+#include <codecvt>
 #include <cstdint>
+#include <cstring>
+#include <iomanip>
+#include <locale>
+#include <sstream>
+
+#include "invariant.h"
 
 #ifdef _WIN32
 #include <util/pragma_push.def>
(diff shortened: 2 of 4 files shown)


src/util/unicode.h
```diff
@@ -31,6 +31,11 @@ std::string utf16_native_endian_to_java(const std::wstring &in);
 
 std::vector<std::string> narrow_argv(int argc, const wchar_t **argv_wide);
 
+/// \param hex: a representation of a UTF-16 character as a four-digit string
+///   (e.g.\ "0041" for \u0041)
```
tautschnig: It seems you will need to escape this for Doxygen to be happy.

src/util/unicode.cpp
```diff
@@ -315,3 +318,12 @@ std::string utf16_native_endian_to_java(const std::wstring &in)
     utf16_native_endian_to_java(ch, result, loc);
   return result.str();
 }
+
+std::string utf16_hex_to_utf8(const std::string &hex)
+{
+  PRECONDITION(hex.length() == 4);
+  const char16_t utf16_char = std::strtol(hex.c_str(), nullptr, 16);
+  const std::u16string u16(1, utf16_char);
+  return std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>{}
```
tautschnig, line 336: This is failing with a linker error with Visual Studio. It's a known problem and https://stackoverflow.com/questions/32055357/visual-studio-c-2015-stdcodecvt-with-char16-t-or-char32-t discusses some fixes. It seems that using `wchar_t` instead of `char16_t` might be the easiest of those, if that does actually work. But it would need to be wrapped in an `#ifdef _MSC_VER`.

unit/json/json_parser.cpp
```diff
@@ -84,17 +84,21 @@ SCENARIO("Loading JSON files")
       }
     }
   }
-  GIVEN("A JSON file containing a hexadecimal Unicode character")
+  GIVEN("A JSON file containing hexadecimal Unicode characters")
```
romainbrenguier: :pick: The use of "characters" here is a bit confusing. I would talk about symbols (or code points) for unicode and only use characters for the c++ `char` type :hamburger:

src/util/unicode.h
```diff
@@ -31,6 +31,11 @@ std::string utf16_native_endian_to_java(const std::wstring &in);
 
 std::vector<std::string> narrow_argv(int argc, const wchar_t **argv_wide);
 
+/// \param hex: a representation of a UTF-16 character as a four-digit string
```
romainbrenguier: :hamburger:

src/util/unicode.h
```diff
@@ -31,6 +31,11 @@ std::string utf16_native_endian_to_java(const std::wstring &in);
...
+///   (e.g.\ "0041" for \u0041)
+/// \return the UTF-8 encoding of the character.
```
romainbrenguier: :hamburger:

romainbrenguier (approved):

smowton:
Nitpick: chars between 00 and ff are not UTF-8 (which can represent all unicode characters, perhaps with many bytes) -- better to just say codepoints <= 255 or 0xff. Similarly codepoints <= 0xffff are the basic multilingual plane (BMP), but not UTF-16, which again can represent all codepoints, perhaps using multiple wchars. Therefore I suggest twiddling PR description / commit messages to clarify that this is adding support for all BMP unicode codepoints

src/json/parser.y
```diff
@@ -51,7 +53,7 @@ static std::string convert_TOK_STRING()
         // \uABCD, i.e. the following four digits are part of this character.
         assert(p + 4 < yyjsontext + len - 1);
         std::string hex(++p, 4);
-        result += std::stoi(hex, nullptr, 16);
+        result += utf16_hex_to_utf8(hex);
```
smowton: Rename codepoint_hex_to_utf8 -- the difference is that utf-16 implies the possibility of two 16-bit numbers representing a single codepoint, such as U+10437 == 0xD801 0xDC37

src/util/unicode.cpp
```diff
@@ -315,3 +318,25 @@ std::string utf16_native_endian_to_java(const std::wstring &in)
     utf16_native_endian_to_java(ch, result, loc);
   return result.str();
 }
+
+char16_t utf16_hex_to_utf16_native_endian(const std::string &hex)
```
smowton: Again, codepoint rather than utf16

src/util/unicode.h
```diff
@@ -31,6 +31,17 @@ std::string utf16_native_endian_to_java(const std::wstring &in);
 
 std::vector<std::string> narrow_argv(int argc, const wchar_t **argv_wide);
 
+char16_t utf16_hex_to_utf16_native_endian(const std::string &hex);
+
+std::string utf16_native_endian_to_utf8(const std::u16string &utf16_str);
+
+std::string utf16_native_endian_to_utf8(char16_t utf16_char);
+
+/// \param hex: a representation of a UTF-16 symbol as a four-digit string
```
smowton: of a BMP codepoint

src/util/unicode.cpp
antlechner: I tried the solution that uses `wchar_t`, the AWS Windows build is still failing and I can't tell from the logs if it's because of this or for some other reason... Can you tell from the logs what the problem is?

antlechner: Turns out the problem was a compilation error that, for some reason, did not show up anywhere in the AWS build logs.
Now everything builds fine but test are failing - again only on Windows...

antlechner: The problem was the encoding of string literals, which can vary between compilers. Should all be fixed now!

antlechner:
All review comments have been addressed. I split the original "hex to UTF8" function into four tiny functions, as this makes it easier to reuse them (and I'm planning to use one of them in some new code soon). All four functions are documented (@smowton could you check that the terminology in the Doxygen makes sense?). Some of the string literals in the unit test have been specified to use UTF-8 encoding as the Windows compiler used a different encoding by default.
I also addressed @smowton 's review comment from https://github.com/diffblue/cbmc/pull/4590, I used a variable name instead of comments to (hopefully) clarify how the parser processes Unicode representations.

smowton (approved):
I'd probably clarify that while anything taking a `char16_t` *is* converting UTF-16 to UTF-8, it's not tackling the whole of UTF-16, which can use low / high surrogate pairs (two successive char16_ts) to represent characters outside the BMP (i.e. those from U+0x10000 upwards). So really they're converting a UTF-16 BMP char... which is always just equal to the codepoint number. Otherwise lgtm.

smowton:
(In case that was a bit confused... what I'm saying is we're handling that subset of Unicode where the UTF-16 <-> codepoint number transformation is trivial (`int codepoint = 0x1234; char16_t utf16_codepoint = (char16_t)codepoint;`). By contrast those chars that don't fit into a char16_t need a more involved transformation.

tautschnig (approved):

PR #4050, opened 2019-02-03
languaget is not a messaget [blocks: #3800]

tautschnig:
Pass a message handler as an argument to several of its functions, but do not
construct a messaget object without a configured message handler as that is
deprecated.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- n/a Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4596, opened 2019-05-01
languaget now needs message_handler during construction

kroening:
This avoids usage of the deprecated `messaget` constructor without message
handler; it is follow-up for #4050.

### DIFF
--- a/src/goto-instrument/dump_c.cpp
+++ b/src/goto-instrument/dump_c.cpp
@@ -1405,13 +1405,17 @@ void dump_c(
   const namespacet &ns,
   std::ostream &out)
 {
+  null_message_handlert mh;
+
   dump_ct goto2c(
     src,
     use_system_headers,
     use_all_headers,
     include_harness,
     ns,
-    new_ansi_c_language);
+    new_ansi_c_language,
+    mh);
+
   out << goto2c;
 }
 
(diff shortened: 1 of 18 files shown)


src/goto-instrument/dump_c.cpp
```diff
@@ -1405,13 +1405,17 @@ void dump_c(
   const namespacet &ns,
   std::ostream &out)
 {
+  null_message_handlert mh;
```
tautschnig, line 1408: Maybe we could instead pass in a message handler?

src/goto-instrument/dump_c.cpp
```diff
@@ -1423,12 +1427,16 @@ void dump_cpp(
...
```
tautschnig, line 1430: As above: I'd favour a non-null message handler.

src/langapi/mode.cpp
```diff
@@ -21,6 +21,8 @@ Author: Daniel Kroening, kroening@cs.cmu.edu
 #include <util/invariant.h>
 #include <util/namespace.h>
 
+null_message_handlert null_message_handler;
```
tautschnig, line 24: To be honest, this one bit makes me like my proposal in #4050 much better.

tautschnig (commented):
Given the introduction of a global null message handler, and acknowledging that I may be biased: I do prefer #4050. But I'm open to arguments convincing me otherwise.

Apart from that, all CI is failing, so more work is necessary.

src/langapi/mode.cpp
```diff
@@ -38,7 +40,7 @@ languagest languages;
 void register_language(language_factoryt factory)
```
peterschrammel, line 40: I'd pass the message handler here instead of having a null_message_handler.

tautschnig:
Closing in favour of #4050.

PR #4598, opened 2019-05-01
use apply in cover_util

kroening:
This avoids the case split over the instruction type.


smowton (approved):

peterschrammel (approved):

tautschnig:
@kroening This is failing some of the MC/DC regression tests simply because the property numbers changed. I think the fix here is to just use `\d+` instead of fixed property numbers in the test.desc files as really the order of those does not matter.

PR #4599, opened 2019-05-01
Remove-virtual-functions: avoid computing class_hierarchyt when possible

smowton:


owen-mc-diffblue (approved):

jbmc/src/jbmc/jbmc_parse_options.cpp
```diff
@@ -753,7 +753,7 @@ void jbmc_parse_optionst::process_goto_function(
     *class_hierarchy,
     ui_message_handler);
   // Java virtual functions -> explicit dispatch tables:
-  remove_virtual_functions(function);
```
peterschrammel, line 756: I suggest to deprecate the variant without class_hierarchy.

peterschrammel (approved):

PR #4608, opened 2019-05-02
Clean-up string solver: refactor binary relations [TG-7439]

allredj:
Rename and refactor binary constraining functions to take two integers rather than a string and an integer. I.e. let the caller dereference the length of the string.

N.B. These relations are no longer string-specific. They should be moved to a more general location in a later PR.

Only review 4 last commits.

Based on:
* https://github.com/diffblue/cbmc/pull/4601
* https://github.com/diffblue/cbmc/pull/4607

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4609, opened 2019-05-02
Refactor: Introduce array_string_exprt.length_type() [TG-7439]

allredj:
Directly accessing the size of the string backing arrays will soon no longer be allowed, as the length of the string needs to be stored separately and the size of the backing array will no longer be relied upon.

We introduce a functionality to keep accessing the type of the array size, as this is widely used and we don't need to query the array_pool for that.

Based on https://github.com/diffblue/cbmc/pull/4608. Only review last 2 commits.

### DIFF
--- a/src/solvers/strings/string_builtin_function.cpp
+++ b/src/solvers/strings/string_builtin_function.cpp
@@ -182,7 +182,7 @@ optionalt<exprt> string_concat_char_builtin_functiont::eval(
   }();
   input_opt.value().push_back(char_val);
   const auto length =
-    from_integer(input_opt.value().size(), result.length().type());
+    from_integer(input_opt.value().size(), result.length_type());
   const array_typet type(result.type().subtype(), length);
   return make_string(input_opt.value(), type);
 }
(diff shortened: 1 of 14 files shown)


smowton (approved):

src/solvers/strings/string_builtin_function.cpp
```diff
@@ -182,7 +182,7 @@ optionalt<exprt> string_concat_char_builtin_functiont::eval(
   }();
   input_opt.value().push_back(char_val);
   const auto length =
-    from_integer(input_opt.value().size(), result.length().type());
+    from_integer(input_opt.value().size(), result.length_type());
```
romainbrenguier, line 185: :pick: In the commit message: "everywhere" is too imprecise

romainbrenguier (approved):
Needs to be rebased once the other pull requests are merged. Otherwise the two last commits look good.

src/solvers/strings/string_builtin_function.cpp
allredj: Added a commit description

PR #4611, opened 2019-05-03
Document and refactor goto check

xbauch:
If you go commit-by-commit it's not so horrible. There is not functional change only a bit of rewording.

### DIFF
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -136,11 +136,19 @@ protected:
 
   std::string array_name(const exprt &);
 
-  void add_guarded_claim(
-    const exprt &expr,
+  /// Include the \p asserted_expr in the code conditioned by the \p guard.
+  /// \param asserted_expr: expression to be asserted
+  /// \param comment: human readable comment
+  /// \param property_class: classification of the property
+  /// \param source_location: the source location of the original expression
+  /// \param src_expr: the original expression to be included in the comment
+  /// \param guard: the condition under which the asserted expression should be
+  ///   taken into account
+  void add_guarded_property(
+    const exprt &asserted_expr,
     const std::string &comment,
     const std::string &property_class,
-    const source_locationt &,
+    const source_locationt &source_location,
     const exprt &src_expr,
     const guardt &guard);
 
@@ -257,7 +265,7 @@ void goto_checkt::div_by_zero_check(
   exprt zero=from_integer(0, expr.op1().type());
   const notequal_exprt inequality(expr.op1(), std::move(zero));
 
-  add_guarded_claim(
+  add_guarded_property(
     inequality,
     "division by zero",
     "division-by-zero",
@@ -283,7 +291,7 @@ void goto_checkt::undefined_shift_check(
     binary_relation_exprt inequality(
       expr.distance(), ID_ge, from_integer(0, distance_type));
 
-    add_guarded_claim(
+    add_guarded_property(
       inequality,
       "shift distance is negative",
       "undefined-shift",
@@ -299,7 +307,7 @@ void goto_checkt::undefined_shift_check(
     exprt width_expr=
       from_integer(to_bitvector_type(op_type).get_width(), distance_type);
 
-    add_guarded_claim(
+    add_guarded_property(
       binary_relation_exprt(expr.distance(), ID_lt, std::move(width_expr)),
       "shift distance too large",
       "undefined-shift",
@@ -312,7 +320,7 @@ void goto_checkt::undefined_shift_check(
       binary_relation_exprt inequality(
         expr.op(), ID_ge, from_integer(0, op_type));
 
-      add_guarded_claim(
+      add_guarded_property(
         inequality,
         "shift operand is negative",
         "undefined-shift",
@@ -323,7 +331,7 @@ void goto_checkt::undefined_shift_check(
   }
   else
   {
-    add_guarded_claim(
+    add_guarded_property(
       false_exprt(),
       "shift of non-integer type",
       "undefined-shift",
@@ -345,7 +353,7 @@ void goto_checkt::mod_by_zero_check(
   exprt zero=from_integer(0, expr.op1().type());
   const notequal_exprt inequality(expr.op1(), std::move(zero));
 
-  add_guarded_claim(
+  add_guarded_property(
     inequality,
     "division by zero",
     "division-by-zero",
@@ -375,7 +383,7 @@ void goto_checkt::mod_overflow_check(const mod_exprt &expr, const guardt &guard)
     notequal_exprt minus_one_neq(
       expr.op1(), from_integer(-1, expr.op1().type()));
 
-    add_guarded_claim(
+    add_guarded_property(
       or_exprt(int_min_neq, minus_one_neq),
       "result of signed mod is not representable",
       "overflow",
@@ -426,7 +434,7 @@ void goto_checkt::conversion_check(
         const binary_relation_exprt no_overflow_lower(
           expr.op0(), ID_ge, from_integer(-power(2, new_width - 1), old_type));
 
-        add_guarded_claim(
+        add_guarded_property(
           and_exprt(no_overflow_lower, no_overflow_upper),
           "arithmetic overflow on signed type conversion",
           "overflow",
@@ -445,7 +453,7 @@ void goto_checkt::conversion_check(
           ID_le,
           from_integer(power(2, new_width - 1) - 1, old_type));
 
-        add_guarded_claim(
+        add_guarded_property(
           no_overflow_upper,
           "arithmetic overflow on unsigned to signed type conversion",
           "overflow",
@@ -466,7 +474,7 @@ void goto_checkt::conversion_check(
         const binary_relation_exprt no_overflow_lower(
           expr.op0(), ID_gt, lower.to_expr());
 
-        add_guarded_claim(
+        add_guarded_property(
           and_exprt(no_overflow_lower, no_overflow_upper),
           "arithmetic overflow on float to signed integer type conversion",
           "overflow",
@@ -489,7 +497,7 @@ void goto_checkt::conversion_check(
           const binary_relation_exprt no_overflow_lower(
             expr.op0(), ID_ge, from_integer(0, old_type));
 
-          add_guarded_claim(
+          add_guarded_property(
             no_overflow_lower,
             "arithmetic overflow on signed to unsigned type conversion",
             "overflow",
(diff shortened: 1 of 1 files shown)


NlightNFotis (approved):

src/analyses/goto_check.cpp
```diff
@@ -1515,30 +1521,25 @@ void goto_checkt::check_rec_logical_op(const exprt &expr, guardt &guard)
...
 
-  if(!expr.op0().is_boolean())
-  {
-    std::string msg=
-      "first argument of if must be boolean, but got "
-      +expr.op0().pretty();
-    throw msg;
-  }
+  INVARIANT(
+    !if_expr.cond().is_boolean(),
```
xbauch: Should have been negated: sorry. Will be fixed.

src/analyses/goto_check.cpp
```diff
@@ -102,6 +102,8 @@ class goto_checkt
   ///   operands (or their negations) for recursively calls)
   void check_rec_logical_op(const exprt &expr, guardt &guard);
 
+  void check_rec_if(const exprt &expr, guardt &guard);
```
tautschnig: I think this should just take an `if_exprt` instead of doing the check within the method.

src/analyses/goto_check.cpp
```diff
@@ -110,6 +110,8 @@ class goto_checkt
   ///   the) if-condition for recursively calls)
   void check_rec_if(const exprt &expr, guardt &guard);
 
+  void check_rec_member(const exprt &expr, guardt &guard);
```
tautschnig: Should just take a `member_exprt`

src/analyses/goto_check.cpp
```diff
@@ -1484,24 +1491,21 @@ void goto_checkt::check_rec_address(const exprt &expr, guardt &guard)
...
+      "`" + expr.id_string() + "' takes Boolean operands only, but got " +
+        op.pretty());
 
     check_rec(op, guard);
-
-    if(expr.id()==ID_or)
-      guard.add(not_exprt(op));
-    else
-      guard.add(op);
+    guard.add(expr.id() == ID_or ? not_exprt(op) : op);
```
tautschnig: If you are touching this, I'd suggest to use `boolean_negate(op)` instead of `not_exprt(op)`

src/analyses/goto_check.cpp
```diff
@@ -1569,46 +1618,8 @@ void goto_checkt::check_rec(const exprt &expr, guardt &guard)
...
-        typecast_exprt::conditional_cast(new_address, new_pointer_type);
-
-      dereference_exprt new_deref{new_address_casted};
-      new_deref.add_source_location() = deref.source_location();
-      pointer_validity_check(new_deref, guard);
-
-      return;
-    }
+    check_rec_member(expr, guard);
+    return;
```
tautschnig: This `return` doesn't seem correct to me. I think you should remove it, and also remove the final `forall_operands` from `check_rec_member`.

src/analyses/goto_check.cpp
```diff
@@ -120,6 +120,18 @@ class goto_checkt
   /// \param guard: the condition for the check (unmodified here)
   void check_rec_member(const exprt &expr, guardt &guard);
 
+  /// Check that a division is valid: check for division by zero, overflow and
+  ///   NaN (for floating point numbers).
+  /// \param expr: the expression to be checked
+  /// \param guard: the condition for the check (unmodified here)
+  void check_rec_div(const exprt &expr, guardt &guard);
```
tautschnig: Should take a `div_exprt`

src/analyses/goto_check.cpp
```diff
@@ -1473,24 +1473,19 @@ void goto_checkt::add_guarded_claim(
...
-
-  exprt guarded_expr;
-
-  if(guard.is_true())
-    guarded_expr.swap(simplified_expr);
-  else
-    guarded_expr = implies_exprt(guard.as_expr(), std::move(simplified_expr));
+  exprt guarded_expr =
+    guard.is_true()
+      ? simplified_expr
```
tautschnig: `std::move(simplified_expr)`

src/analyses/goto_check.cpp
```diff
@@ -1489,20 +1489,16 @@ void goto_checkt::add_guarded_claim(
...
-      static_cast<const codet &>(get_nil_irep()),
-      source_location,
-      type,
-      std::move(guarded_expr),
-      {}));
+    auto t = new_code.add(
+      enable_assert_to_assume ? goto_programt::make_assumption(
+                                  std::move(guarded_expr), source_location)
+                              : goto_programt::make_assertion(
+                                  std::move(guarded_expr), source_location));
```
tautschnig, line 1501: There is another instance of this pattern (around line 1800) - just search for "ASSUME". Could you please fix this as well?

src/analyses/goto_check.cpp
```diff
@@ -275,21 +275,14 @@ void goto_checkt::invalidate(const exprt &lhs)
...
-      next++;
-
       if(has_symbol(*it, find_symbols_set) || has_subexpr(*it, ID_dereference))
-        assertions.erase(it);
-
-      it=next;
+        it = assertions.erase(it);
+      else
+        ++it;
     }
```
tautschnig, line 289: What about using `std::remove_if` (until we can use `erase_if` with C++20)?

tautschnig (approved):
This does need cleanup as indicated in the comments, but is overall a good refactoring.

src/analyses/goto_check.cpp
xbauch: The return is there only to prevent the `forall_operands` loop in case the `check_rec_member` was existed via an explicit `return`. But it seems to be only an optimisation so I did what you asked for.

src/analyses/goto_check.cpp
xbauch: Done.

src/analyses/goto_check.cpp
xbauch: Done.

src/analyses/goto_check.cpp
xbauch: Done.

src/analyses/goto_check.cpp
xbauch: Done.

src/analyses/goto_check.cpp
xbauch: Done.

src/analyses/goto_check.cpp
xbauch: Added.

src/analyses/goto_check.cpp
xbauch: Unfortunately, `std::remove_if` requires the iterator to be `MoveAssignable`. Will have to wait for this one.

tautschnig: Ah, ok, too bad.

src/analyses/goto_check.cpp
xbauch: I take it back. Avoiding the recursive call on the member expression operands may be just an optimisation but one of our regression tests (`Malloc23`) requires exactly 4 failing assertions. If we change this section so that the recursion is always called, this test fails.

tautschnig: Which ... might be a bugfix? Does it generate more assertions or is it missing some?

xbauch: Doesn't seem so:
```
struct A
{
  int x;
  int y;
};

int main()
{
  struct A *p=malloc(sizeof(int));
  p->x=0; // OK
  p->y=0; // out of bounds
..
```
Unless we skip the recursive call, we would generate a pointer check for `p->ok=0` which, on it's own, would fail.

tautschnig: But still the behaviour before and after isn't the same: previously, we only returned in _some_ cases, now we always return early.

xbauch: Yes, but since I included the recursive call on sub-expressions in `rec_check_member` the only thing we skip by returning early is the remaining case analyses none of which can succeed (since we know that `expr.id()==ID_member`).

tautschnig: Fair enough, I was just a bit worried that one day someone would add a case. But I suppose testing (or rather: failing tests) should catch that. So it seems fine as is.

xbauch: I thought about it a bit and tried having the `check_rec_member` return a Boolean indicating if the subsequent analysis should be attempted or not: https://github.com/diffblue/cbmc/pull/4611/commits/d6f7f05f2f6c0b86a96dd7a36f132dd1e1b70264.

src/analyses/goto_check.cpp
```diff
@@ -174,11 +174,19 @@ class goto_checkt
...
 
+  /// Include the \p asserted_expr in the code conditioned by the \p guard.
+  /// \param asserted_expr: expression to be asserted
+  /// \param comment: human readable comment
+  /// \param property_class: classification of the property
+  /// \param source_location: the source location of the original expression
+  /// \param src_expr: the original expression to be included in the comment
+  /// \param guard: the condition under which the asserted expression should be
+  ///   taken into account
   void add_guarded_claim(
```
peterschrammel: While we are here, can we rename `claim` to `property` since we have replaced that terminology in other places in the codebase?

src/analyses/goto_check.cpp
```diff
@@ -194,6 +199,10 @@ class goto_checkt
   typedef std::set<exprt> assertionst;
   assertionst assertions;
 
+  /// Remove all assertions containing they symbol in \p lhs as well as all
```
peterschrammel: they -> the

peterschrammel (approved):
lgtm modulo squashing promised

src/analyses/goto_check.cpp
xbauch: Done.

src/analyses/goto_check.cpp
xbauch: Done.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: d62fe0ae386de0e75515b7208b55e6c47974e5de).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112345526

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: b5060ee0fca3fba80d6c977bbf052e2c3a2ada66).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112346598

PR #4614, opened 2019-05-03
String solver: add non-negative length constraint before solving [TG-7439]

allredj:
Add non-negativeness constraint on string lengths before doing the first iteration over the index set.
    
These constraints used to be added after that step, and some of the length variables were missing these constraints.
    
I don't have a failing example at hand, but this issue is made very visible by the upcoming changes to the way we deal with string lengths.

Based on https://github.com/diffblue/cbmc/pull/4609
**Only review last commit.**

### DIFF
--- a/src/solvers/strings/string_refinement.cpp
+++ b/src/solvers/strings/string_refinement.cpp
@@ -740,8 +740,8 @@ decision_proceduret::resultt string_refinementt::dec_solve()
   // All generated strings should have non-negative length
   for(const auto &string : generator.array_pool.created_strings())
   {
-    add_lemma(binary_relation_exprt{
-      string.length(), ID_ge, from_integer(0, string.length_type())});
+    add_lemma(greater_or_equal_to(
+      string.length(), from_integer(0, string.length_type())));
   }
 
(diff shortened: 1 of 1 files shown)


src/solvers/strings/string_refinement.cpp
```diff
@@ -737,6 +737,14 @@ decision_proceduret::resultt string_refinementt::dec_solve()
   for(const exprt &lemma : generator.constraints.existential)
     add_lemma(lemma);
 
+  // All generated strings should have non-negative length
+  for(const auto &string : generator.array_pool.created_strings())
+  {
+    add_lemma(binary_relation_exprt{
+      string.length(), ID_ge, from_integer(0, string.length_type())});
```
romainbrenguier: :pick: `greater_or_equal_to`

romainbrenguier (approved):

smowton (approved):

src/solvers/strings/string_refinement.cpp
allredj: done

PR #4617, opened 2019-05-03
C/JBMC: support distinguishing recursion from loop unwinding

smowton:
Specifying --unwind-loops and --unwind-recursion allows us to distinguish the two cases.


src/goto-instrument/unwindset.cpp
```diff
@@ -60,26 +79,28 @@ void unwindsett::parse_unwindset(const std::string &unwindset)
   }
 }
 
-optionalt<unsigned>
-unwindsett::get_limit(const irep_idt &loop_id, unsigned thread_nr) const
+optionalt<unsigned> unwindsett::get_limit(
+  const irep_idt &loop_id,
+  unsigned thread_nr,
+  bool is_iteration) const
```
hannes-steffenhagen-diffblue, line 85: Please don't use a boolean flag for this, they have too many overloaded meanings already. Either pass in the applicable global limit directly (+ possibly refactor this into `get_loop_limit` and `get_recursion_limit`?) or use an enum like

```
enum class unwind_kindt { loop, recursion };
```

src/goto-checker/bmc_util.h
```diff
@@ -186,6 +186,8 @@ void run_property_decider(
   "(show-symex-strategies)" \
   "(depth):" \
   "(unwind):" \
+  "(unwind-loops):" \
+  "(unwind-recursion):" \
```
hannes-steffenhagen-diffblue, line 190: These should be documented in the user facing documentation (i.e. help message and cprover-manual).

tautschnig (changes requested):
From a purely technical point of view this change looks fine. But still there are two blockers for me here which are all about the user interface (at different levels):
1) All user-facing documentation needs to be updated. Any place where we talk about unwinding needs to have this change reflected. Loop (and recursion) unwinding is probably the single most used command-line option. 
2) I'm missing the "why do we want this" explanation. This change is nicely done so that it doesn't break existing set-ups, but it still makes the command-line interface a bit bigger. It may actually make things easier (because it's not as implicit as `--unwind` is), but raises questions about the interaction between those options (which needs to be document) and also whether we should have the same refinement for `--unwindset`, `--partial-loops`, `--unwinding-asseritions`.

smowton:
Docs: will do. Why: I found it useful because in both C and Java recursion and iteration are stereotypically used quite differently (and recursion can have exponential cost if there are multiple recursive calls in a particular function or SCC-of-functions), so I found myself wanting to construct the particular `unwindset` that sets all recursion bounds shorter than the bound for loops.

peterschrammel (commented):
`partial-loops` and `unwinding-assertions` could be refined as @tautschnig suggests. I don't see why `unwindset` would need any change.

TGWDB:
Closing due to age and lack of update. Please reopen if you believe this is erroneous and plan to update as discussed in the conversation (and rebase).

PR #4622, opened 2019-05-06
C front-end: Do not lose comments in type checking

tautschnig:
In case of declarations and function calls we need to carry over
comments from the statement's source location into individual
subexpressions as the source location of those subexpressions will be
used when generating the goto program instruction.

### DIFF
--- a/src/ansi-c/c_typecheck_code.cpp
+++ b/src/ansi-c/c_typecheck_code.cpp
@@ -243,7 +243,16 @@ void c_typecheck_baset::typecheck_decl(codet &code)
   }
 
   ansi_c_declarationt declaration;
+  irep_idt comment = code.source_location().get_comment();
   declaration.swap(code.op0());
+  if(!comment.empty())
+  {
+    for(auto &d : declaration.declarators())
+    {
+      if(d.source_location().get_comment().empty())
+        d.add_source_location().set_comment(comment);
+    }
+  }
 
   if(declaration.get_is_static_assert())
   {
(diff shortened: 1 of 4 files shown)


peterschrammel (approved):

kroening (approved):

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2019,6 +2019,10 @@ void c_typecheck_baset::typecheck_side_effect_function_call(
     throw 0;
   }
 
+  irep_idt comment = expr.source_location().get_comment();
+  if(!comment.empty() && f_op.source_location().get_comment().empty())
```
danielsn, line 2023: If the comment is non empty, should we merge instead of ignoring?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -243,7 +243,16 @@ void c_typecheck_baset::typecheck_decl(codet &code)
...
 
   ansi_c_declarationt declaration;
+  irep_idt comment = code.source_location().get_comment();
   declaration.swap(code.op0());
+  if(!comment.empty())
+  {
+    for(auto &d : declaration.declarators())
+    {
+      if(d.source_location().get_comment().empty())
+        d.add_source_location().set_comment(comment);
```
danielsn, line 253: If the comment is non empty, should we merge instead of ignoring?

tautschnig:
I will merge this now, but will also explore a cleaner solution that doesn't abuse comments and instead has dedicated annotations. Then "merging" will actually be possible as the semantics of those annotations can be clearly defined.

src/ansi-c/c_typecheck_code.cpp
tautschnig: The new implementation in #4666 also resolves this problem (in parts by just removing the above code :-) ).

PR #4623, opened 2019-05-07
bugfix: pointer check failure in fgets

### LINKED ISSUE #4621: Pointer check failure in fgets
CBMC version: 5.11
Operating system: Linux
Exact command line resulting in the issue: cbmc --pointer-check fgets.c
What behaviour did you expect: VERIFICATION SUCCESSFUL
What happened instead: VERIFICATION FAILED

File fgets.c:
```
#include <stdio.h>

int main() {
  char buffer[10];
  FILE *fd = fopen("test.txt", "r");

  if (fd) {
    fgets(buffer, 10, fd);
    fclose(fd);
  }
}
```

This program should be successfully verified, however, it produces a dereference failure inside CBMC's fgets implementation. More precisely, a _pointer outside object bounds in *(((const char *)str + (signed long int)str_length) - (signed long int)1)_.

Checking the current implementation in **src/ansi-c/library/stdio.c**, the error is produced because str_length is a nondeterministically chosen int in the range [0, size). Since fgets documentation clearly states that if the End-of-File is encountered and no characters have been read, the contents of str remain unchanged, the dereference verification should only take place if str_length is greater than 0.

mauguignard:
Now we assume str_length greater than 0, since no dereference should take place when str_length equals 0.
Updated cbmc-library/fgets-01 regression test.

Fixes: #4621

### DIFF
--- a/regression/cbmc-library/fgets-01/test.desc
+++ b/regression/cbmc-library/fgets-01/test.desc
@@ -5,6 +5,6 @@ main.c
 ^SIGNAL=0$
 ^VERIFICATION FAILED$
 \[main.assertion.3\] line 16 assertion p\[1\] == 'b': FAILURE
-\*\* 2 of \d+ failed
+\*\* 1 of \d+ failed
 --
 ^warning: ignoring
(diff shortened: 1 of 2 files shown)


kroening:
May I suggest to switch this to using `__CPROVER_w_ok(s, n)`?
Say something like
`__CPROVER_precondition(__CPROVER_w_ok(str, str_length+1),
                         "fgets buffer writeable");`

tautschnig (approved):

PR #4626, opened 2019-05-08
String solver: stop relying on array size

allredj:
The string's backing array size (from its type) initially informs about the length of the string. However during refinement, the length can vary and if the length is tied to the size of the array, we need to rewrite the type as a conditional, which should be avoided (both the rewriting and the conditional).

This PR refactors string refinement so that during constraint generation, we refer to the size of the string solely through the array pool.
The size of a string thus always refers to a symbol, which may or may not be constrained.

The only time we get the array size is when inserting an array_string_exprt in the array pool (i.e. at the creation of the array_string_exprt).

There shouldn't be any visible change in JBMC, but that should fix some errors that have been happening with test-generation. I will keep looking to see if I can reproduce on JBMC.

I will also check that performance is not affected.


src/solvers/strings/array_pool.cpp
```diff
@@ -32,6 +32,18 @@ exprt array_poolt::get_length(const array_string_exprt &s) const
   return s.length();
 }
 
+/// As opposed to get_length(), do not create a new symbol if the length
```
romainbrenguier: :pick: `\ref`

romainbrenguier: this is part of the public interface so should be documented in the header file :grapes:

src/solvers/strings/array_pool.cpp
```diff
@@ -21,15 +21,14 @@ operator()(const irep_idt &prefix, const typet &type)
   return result;
 }
 
-exprt array_poolt::get_length(const array_string_exprt &s) const
+/// Get the length of an array_string_exprt from the array_pool. If the length
+/// does not yet exist, create a new symbol and add it to the pool.
```
romainbrenguier: :grapes:

src/solvers/strings/array_pool.cpp
```diff
@@ -21,15 +21,14 @@ operator()(const irep_idt &prefix, const typet &type)
   return result;
 }
 
-exprt array_poolt::get_length(const array_string_exprt &s) const
+/// Get the length of an array_string_exprt from the array_pool. If the length
```
romainbrenguier: :pick: typo in commit message: "accesssing"

src/solvers/strings/array_pool.cpp
```diff
@@ -46,8 +46,7 @@ array_poolt::get_length_if_exists(const array_string_exprt &s) const
 array_string_exprt
 array_poolt::fresh_string(const typet &index_type, const typet &char_type)
 {
-  symbol_exprt length = fresh_symbol("string_length", index_type);
-  array_typet array_type(char_type, length);
+  array_typet array_type(char_type, infinity_exprt(index_type));
```
romainbrenguier: :pick: brace initialization

src/solvers/strings/array_pool.cpp
```diff
@@ -23,8 +23,21 @@ operator()(const irep_idt &prefix, const typet &type)
...
+///
+/// If the array_string_exprt is an `if` expression, the true and false parts
+/// are handled independently (and recursively). That is, no new length symbol
+/// is added to the pool for `if` expressions, but symbols may be added for each
+/// of the parts.
 exprt array_poolt::get_length(const array_string_exprt &s)
 {
+  if(const auto &if_expr = expr_try_dynamic_cast<if_exprt>((exprt)s))
+  {
+    return if_exprt(
```
romainbrenguier: :pick: brace initialization

src/solvers/strings/array_pool.cpp
```diff
@@ -115,6 +115,54 @@ array_string_exprt array_poolt::make_char_array_for_char_pointer(
...
+  else
+  {
+    const exprt &size_from_type = to_array_type(array_expr.type()).size();
+    if(size_from_type != infinity_exprt(size_from_type.type()))
+    {
+      const auto emplace_result =
+        length_of_array.emplace(array_expr, size_from_type);
+      INVARIANT(
+        emplace_result.second,
+        "attempt_assign_length should only be called when no entry for the"
```
romainbrenguier: `attempt_assign_length_from_type`

src/solvers/strings/array_pool.cpp
```diff
@@ -65,6 +65,10 @@ array_poolt::fresh_string(const typet &index_type, const typet &char_type)
   created_strings_value.insert(str);
   arrays_of_pointers.emplace(
     address_of_exprt(index_exprt(str, from_integer(0, index_type))), str);
+
+  // add length to length_of_array map
+  get_length(str);
```
romainbrenguier: :pick: maybe worth defining a void function to avoid having to write that comment twice

src/solvers/strings/array_pool.cpp
```diff
@@ -184,9 +184,12 @@ void array_poolt::insert(
   attempt_assign_length_from_type(array_expr, length_of_array, fresh_symbol);
 }
 
-const std::set<array_string_exprt> &array_poolt::created_strings() const
+/// Return a map mapping all array_string_exprt of the array_pool to their
+/// length.
+const std::unordered_map<array_string_exprt, exprt, irep_hash> &
```
romainbrenguier, line 168: :bulb: exposing the whole map here seems unnecessary since we are only using the first component. An alternative would be to define a `for_each_string` function with takes a function as argument and apply it to each `array_string_exprt`.

romainbrenguier (commented):
Reviewed up to "Get rid of created_strings_value". No blocking request so far.

src/solvers/strings/string_refinement.cpp
```diff
@@ -1834,9 +1834,12 @@ exprt string_refinementt::get(const exprt &expr) const
     if(const auto arr_model_opt = get_array(super_get, ns, log.debug(), arr))
       return *arr_model_opt;
 
-    if(generator.array_pool.created_strings().count(arr))
+    if(
+      const auto &length_from_pool =
+        generator.array_pool.get_length_if_exists(arr))
     {
-      const exprt length = super_get(arr.length());
+      exprt length = super_get(length_from_pool.value());
```
romainbrenguier: no need to remove the `const`

src/solvers/strings/string_refinement.cpp
```diff
@@ -105,7 +105,8 @@ static optionalt<exprt> get_array(
   const std::function<exprt(const exprt &)> &super_get,
   const namespacet &ns,
   messaget::mstreamt &stream,
-  const array_string_exprt &arr);
+  const array_string_exprt &arr,
```
romainbrenguier, line 110: :pick: commit message doesn't say why

src/solvers/strings/string_refinement.cpp
```diff
@@ -965,7 +965,11 @@ static optionalt<exprt> get_array(
   const array_string_exprt &arr,
   const array_poolt &array_pool)
 {
-  const exprt &size = arr.length();
+  const auto &size_from_pool = array_pool.get_length_if_exists(arr);
+  exprt size = size_from_pool.has_value()
```
romainbrenguier: no need to remove the `const`

src/solvers/strings/string_constraint_generator_format.cpp
```diff
@@ -493,8 +498,9 @@ std::pair<exprt, string_constraintst> add_axioms_for_format(
           is_refined_string_type(arg.type()),
           "arguments of format should be strings");
         const array_string_exprt string_arg = get_string_expr(array_pool, arg);
-        get_arg = [string_arg](const irep_idt &id) {
-          return format_arg_from_string(string_arg, id);
+        /// REMOVE get_arg below
```
romainbrenguier: forgotten comment?

romainbrenguier (approved):

smowton:
Will look at this tomorrow, but do you intend it to remove the illegal `x ? y : z` expressions where x has type `char[x ? sizeof(y) : sizeof(z)]`?

src/solvers/strings/array_pool.cpp
allredj: `\ref` is unnecessary. Doxygen understands where to link.

src/solvers/strings/array_pool.cpp
```diff
@@ -32,6 +32,18 @@ exprt array_poolt::get_length(const array_string_exprt &s) const
   return s.length();
 }
 
+/// As opposed to get_length(), do not create a new symbol if the length
+/// of the array_string_exprt does not exist, but instead return an empty
```
smowton: Perhaps clarify what you mean by a length not existing? Is that any non-constant length, or the length of a string we haven't fully initialised yet, or...?

src/solvers/strings/array_pool.cpp
```diff
@@ -21,15 +21,14 @@ operator()(const irep_idt &prefix, const typet &type)
...
+exprt array_poolt::get_length(const array_string_exprt &s)
 {
-  if(s.length() == infinity_exprt(s.length().type()))
-  {
-    auto it = length_of_array.find(s);
-    if(it != length_of_array.end())
-      return it->second;
-  }
-  return s.length();
+  auto emplace_result =
```
smowton, line 34: Given this allocates a symbol even if `s` already has a length, I'd suggest inserting a dummy and then replacing it with a `fresh_symbol` if appropriate

src/solvers/strings/array_pool.cpp
```diff
@@ -115,6 +115,54 @@ array_string_exprt array_poolt::make_char_array_for_char_pointer(
...
+///
+/// This function should only be used at the creation of the
+/// `array_string_exprt`s, as it is the only place where we can reliably refer
+/// to the size in the type of the array.
+static void attempt_assign_length_from_type(
+  const array_string_exprt &array_expr,
+  std::unordered_map<array_string_exprt, exprt, irep_hash> &length_of_array,
+  symbol_generatort &symbol_generator)
+{
+  if(const auto &if_expr = expr_try_dynamic_cast<if_exprt>((exprt)array_expr))
```
smowton: Why the explicit cast (and to a non-reference-type?)

src/solvers/strings/array_pool.cpp
```diff
@@ -115,6 +115,54 @@ array_string_exprt array_poolt::make_char_array_for_char_pointer(
...
+      const auto emplace_result =
+        length_of_array.emplace(array_expr, size_from_type);
+      INVARIANT(
+        emplace_result.second,
+        "attempt_assign_length should only be called when no entry for the"
+        "array_string_exprt exists in the length_of_array map");
+    }
+    else
+    {
+      length_of_array.emplace(
```
smowton: Should check this one is new, too? Therefore, deduplicate the `emplace` calls and the invariant below this if block?

src/solvers/strings/array_pool.cpp
smowton: I wouldn't normally expect a `get_` function to have a side-effect like this, suggest renaming

src/solvers/strings/array_pool.h
```diff
@@ -76,6 +76,8 @@ class array_poolt final
 
 private:
   /// Associate arrays to char pointers
+  /// Invariant: Each array_string_exprt in this map has a corresponding entry
```
smowton, line 91: Where is this enforced?

src/solvers/strings/array_pool.cpp
allredj: I'll leave it like this for now so that the PR doesn't blow up.

src/solvers/strings/array_pool.cpp
allredj: Doc moved to header file

src/solvers/strings/array_pool.cpp
allredj: Doc moved to header file

src/solvers/strings/array_pool.cpp
allredj: Fixed

src/solvers/strings/array_pool.cpp
allredj: Done

src/solvers/strings/array_pool.cpp
allredj: Done

src/solvers/strings/array_pool.cpp
allredj: Fixed

src/solvers/strings/array_pool.cpp
allredj: Maybe in a later PR.

src/solvers/strings/string_refinement.cpp
allredj: Re-introduced

src/solvers/strings/string_refinement.cpp
allredj: Added description

src/solvers/strings/string_refinement.cpp
allredj: const re-introduced

src/solvers/strings/string_constraint_generator_format.cpp
allredj: Removed

src/solvers/strings/array_pool.cpp
allredj: You mean the `(exprt)`? Curiously `exprt_try_dynamic_cast` doesn't work directly on `array_string_exprt`.

smowton: Oh wait, this is obvious -- an `array_string_exprt` is certainly not an `if_exprt`, this can't ever be run.

smowton: There's one other case with the same cast and the same unreachable code. This suggests writing tests to exercise this code!

src/solvers/strings/array_pool.cpp
allredj: Would you be happy with `get_or_create_length`?

smowton: Sounds good to me

src/solvers/strings/array_pool.cpp
allredj: Improved the documentation. This now states: 
```
  /// As opposed to get_length(), do not create a new symbol if the length
  /// of the array_string_exprt does not have one in the array_pool, but instead
  /// return an empty optional.
```

src/solvers/strings/array_pool.cpp
allredj: Done in fixup commit

src/solvers/strings/array_pool.cpp
allredj: Done in fixup commit

src/solvers/strings/array_pool.h
allredj: Improved documentation. Now states:
```
  /// Associate arrays to char pointers
  /// Invariant: Each array_string_exprt in this map has a corresponding entry
  ///            in length_of_array.
  ///            This is enforced whenever we add an element to
  ///            `arrays_of_pointers`, i.e. fresh_string()
  ///            and make_char_array_for_char_pointer().
```

allredj:
@smowton About your general comment. Yes, removing the conditional types is the idea of that PR.

tautschnig (approved):
Great to see those conditionals in types go away! I'd just ask for the commit history to be cleaned up with all those fixup commits. Doing so will then also re-trigger the codebuild task.

src/solvers/strings/array_pool.cpp
allredj: I don't think it's unreachable. `array_string_exprt`s can have `ID_if` AFAIK. However you make a good point with the test coverage, as we don't seem to have anything to cover these lines so for now I might just put an assertion.

allredj:
@tautschnig Thanks. Of course I'll merge the fixup commits before merge. I have some failing tests on TG, so I need to fix that first anyway.

src/solvers/strings/array_pool.cpp
allredj: So in this place i replaced the if with an assertion because I can't find any test to cover this part. However in the other place where I use the condition, I left it in because it's often the case that the `array_string_exprt` is an if. Therefore that code is well covered by regression tests.

src/solvers/strings/array_pool.cpp
allredj: Renamed

allredj:
@romainbrenguier can you please review the 7 last commits which refactor `get_array` and solve the issue with the dangling string literals in the trace (in return values). I 'm still looking for a regression test for that.

src/solvers/strings/string_refinement.cpp
```diff
@@ -1003,7 +1003,7 @@ static optionalt<exprt> get_array(
            << messaget::eom;
     stream << "(sr::get_array) this can also happen on invalid object access"
            << messaget::eom;
-    return nil_exprt();
+    return {};
```
romainbrenguier: This can potentially change the behaviour in the test `jbmc/regression/jbmc-strings/long_string/test_abort.desc` because here MAX_CONCRETE_STRING_SIZE was exceeded (on purpose) and the value of the array replaced by nil in the trace. Can you check what happens now in the trace?

src/solvers/strings/string_refinement.cpp
```diff
@@ -949,16 +949,18 @@ void string_refinementt::add_lemma(
...
 /// \param super_get: function returning the valuation of an expression
 ///   in a model
 /// \param ns: namespace
 /// \param stream: output stream for warning messages
 /// \param arr: expression of type array representing a string
 /// \param array_pool: pool of arrays representing strings
-/// \return an optional array expression or array_of_exprt
-static optionalt<exprt> get_array(
+/// \return an optional expression representing the size of the array that can
+///         be cast to type_t
```
romainbrenguier: what's `type_t`?

src/solvers/strings/string_refinement.cpp
```diff
@@ -970,42 +972,72 @@ static optionalt<exprt> get_array(
...
+{
+  auto size = get_valid_array_size(super_get, ns, stream, arr, array_pool);
+  if(!size.has_value())
+  {
+    return {};
+  }
+
+  size_t n = numeric_cast<std::size_t>(size.value()).value();
+
+  exprt arr_val = simplify_expr(adjust_if_recursive(super_get(arr), ns), ns);
```
romainbrenguier: if the type problems are fixed by the previous commits, the calls to `adjust_if_recursive` should no longer be required

src/solvers/strings/string_refinement.cpp
```diff
@@ -970,42 +972,72 @@ static optionalt<exprt> get_array(
...
+  const array_string_exprt &arr,
+  const array_poolt &array_pool)
+{
+  auto size = get_valid_array_size(super_get, ns, stream, arr, array_pool);
+  if(!size.has_value())
+  {
+    return {};
+  }
+
+  size_t n = numeric_cast<std::size_t>(size.value()).value();
```
romainbrenguier: const?

src/solvers/strings/string_refinement.cpp
```diff
@@ -970,42 +972,72 @@ static optionalt<exprt> get_array(
...
+/// \param array_pool: pool of arrays representing strings
+/// \return an optional array expression or array_of_exprt
+static optionalt<exprt> get_array(
+  const std::function<exprt(const exprt &)> &super_get,
+  const namespacet &ns,
+  messaget::mstreamt &stream,
+  const array_string_exprt &arr,
+  const array_poolt &array_pool)
+{
+  auto size = get_valid_array_size(super_get, ns, stream, arr, array_pool);
```
romainbrenguier: const

src/solvers/strings/string_refinement.cpp
romainbrenguier: const

src/solvers/strings/string_refinement.cpp
```diff
@@ -982,13 +982,13 @@ static optionalt<exprt> get_valid_array_size(
     return {};
   }
 
-  auto n_opt = numeric_cast<std::size_t>(size_val);
-  if(!n_opt)
+  auto num_size_opt = numeric_cast<std::size_t>(size_val);
```
romainbrenguier: `num_size_opt` is also not a great name (what does "num" stand for?). `size_opt` could be good enough?

src/solvers/strings/string_refinement.cpp
```diff
@@ -988,12 +988,12 @@ static optionalt<exprt> get_valid_array_size(
     stream << "(sr::get_valid_array_size) size is not valid" << messaget::eom;
     return {};
   }
-  std::size_t n = *num_size_opt;
+  std::size_t num_size = *num_size_opt;
```
romainbrenguier: :pick: I would rather name this one `size` and the previous size `size_expr`

src/solvers/strings/string_refinement.cpp
allredj: Indeed that broke that test. I reverted that change and consequently changed the refactoring a bit. I moved the max length check back into `get_array`.

src/solvers/strings/string_refinement.cpp
allredj: I meant `size_t`. Fixed

src/solvers/strings/string_refinement.cpp
allredj: I don't know. I'll leave it for now.

src/solvers/strings/string_refinement.cpp
allredj: done

src/solvers/strings/string_refinement.cpp
allredj: done

src/solvers/strings/string_refinement.cpp
allredj: done

src/solvers/strings/string_refinement.cpp
allredj: Got rid of that change

src/solvers/strings/string_refinement.cpp
allredj: Got rid of that change

allredj:
Fixed outstanding issues and addressed comments. @romainbrenguier could you re-review the 5 last commits please?

src/solvers/strings/string_refinement.cpp
tautschnig: We should then have a follow-up PR that removes `adjust_if_recursive` I'd say.

src/solvers/strings/array_pool.cpp
smowton: Aha, my mistake, I thought `array_string_exprt` constrained the expression to have a particular id, and therefore not be an `if_exprt`, but actually it just constrains the type of a general expression. The implementation of the expr-casting stuff, though, assumes this sort of trait class doesn't exist, and so casts are only possible to children of the current type. That ought to be fixed, but is obviously out of scope for this PR, so approving.

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 6da6e01b7d8f039f789e1411b1eb064e8cf3aed5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111727112

src/solvers/strings/string_refinement.cpp
allredj: @tautschnig https://github.com/diffblue/cbmc/pull/4653

src/solvers/strings/string_refinement.cpp
```diff
@@ -993,8 +993,8 @@ static optionalt<exprt> get_array(
 
   if(n > MAX_CONCRETE_STRING_SIZE)
   {
-    stream << "(sr::get_array) long string (size " << format(size) << " = " << n
-           << ") " << format(arr) << messaget::eom;
+    stream << "(sr::get_array) long string (size = " << n << ") " << format(arr)
```
romainbrenguier: :pick: typo in commit message: "Renmove"

romainbrenguier (approved):
Looks good

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 09588ad748d4528da901958b49fdbf7618f298d5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112432177

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 27e27c0ee3860c446afcff7b63843ce99b827b1c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113065638

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: cd61e81953ec41d7235f767ac9e1906fac8a137c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113725722

PR #4627, opened 2019-05-08
Fix simplification of pointer-object comparison [blocks: #4628]

tautschnig:
1) pointer_object((T1 *)NULL) equals pointer_object((T2 *)NULL) for any
types T1, T2. Previously, this would return false, unless T1==T2.
2) Do not restrict the above to NULL, but instead let the existing logic
in simplify_inequality fully simplify this.
3) Add a unit test of this code, which highlighted further bugs and
limitations: the unit test previously did not set the instance of the
desired dynamic object, and address-of inequalities over dynamic objects
can also be simplified.

### DIFF
--- a/src/util/simplify_expr_pointer.cpp
+++ b/src/util/simplify_expr_pointer.cpp
@@ -447,6 +447,25 @@ bool simplify_exprt::simplify_inequality_address_of(exprt &expr)
 
     return false;
   }
+  else if(
+    tmp0.op0().id() == ID_dynamic_object &&
+    tmp1.op0().id() == ID_dynamic_object)
+  {
+    bool equal = to_dynamic_object_expr(tmp0.op0()).get_instance() ==
+                 to_dynamic_object_expr(tmp1.op0()).get_instance();
+
+    expr.make_bool(expr.id() == ID_equal ? equal : !equal);
+
+    return false;
+  }
+  else if(
+    (tmp0.op0().id() == ID_symbol && tmp1.op0().id() == ID_dynamic_object) ||
+    (tmp0.op0().id() == ID_dynamic_object && tmp1.op0().id() == ID_symbol))
+  {
+    expr.make_bool(expr.id() != ID_equal);
+
+    return false;
+  }
 
   return true;
 }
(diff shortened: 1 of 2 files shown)


src/util/simplify_expr_pointer.cpp
```diff
@@ -447,6 +447,25 @@ bool simplify_exprt::simplify_inequality_address_of(exprt &expr)
...
+  {
+    bool equal = to_dynamic_object_expr(tmp0.op0()).get_instance() ==
+                 to_dynamic_object_expr(tmp1.op0()).get_instance();
+
+    expr.make_bool(expr.id() == ID_equal ? equal : !equal);
+
+    return false;
+  }
+  else if(
+    (tmp0.op0().id() == ID_symbol && tmp1.op0().id() == ID_dynamic_object) &&
```
owen-mc-diffblue:
```suggestion
    (tmp0.op0().id() == ID_symbol && tmp1.op0().id() == ID_dynamic_object) ||
```

tautschnig: Thank you for catching this!!

smowton:
The unit test doesn't currently cover all the added behaviour (e.g. the line @owen-jones-diffblue suggested you change)

tautschnig:
Unit test extended (and locally tested that it does fail without @owen-jones-diffblue's fix).

src/util/simplify_expr_pointer.cpp
```diff
@@ -474,16 +494,23 @@ bool simplify_exprt::simplify_inequality_pointer_object(exprt &expr)
...
-  bool equal=expr.op0().op0()==expr.op1().op0();
-
-  expr.make_bool(expr.id()==ID_equal?equal:!equal);
+    if(new_inequality_ops.empty())
+      new_inequality_ops.push_back(op);
+    else
+    {
+      new_inequality_ops.push_back(typecast_exprt::conditional_cast(
+        op, new_inequality_ops.front().type()));
+      simplify_node(new_inequality_ops.back());
```
smowton, line 508: Thought for the future: in this case I think we have a very particular simplification in mind -- presumably that nulls are typed so the cast might be folded, and that `(T*)(S*)x` can become `(T*)x`. However bluntly using the simplifier always risks that a much more expensive transformation might be attempted. This suggests factoring out particular simplifications as utility methods, used both during simplification and as the routine way of creating a pointer cast.

smowton (approved):

src/util/simplify_expr_pointer.cpp
tautschnig: That certainly is something to review all the simplifier code for. I actually have some patches in a branch already, but it's not quite a systematic review yet. Will work on this in a separate PR.

PR #4629, opened 2019-05-08
fix up __float128 and _Float128 support

kroening:
`_Float128` is a keyword, defined in ISO/IEC TS 18661-3:2015, and exists since
gcc 7.0. clang does not offer it.

`__float128` is a typedef in gcc since 4.3/4.5, depending on architecture.

`__float128` is a keyword in clang.

This PR attempts to clean this up.

### DIFF
--- a/src/ansi-c/ansi_c_internal_additions.cpp
+++ b/src/ansi-c/ansi_c_internal_additions.cpp
@@ -208,8 +208,12 @@ void ansi_c_internal_additions(std::string &code)
       // https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html
       // For clang, __float128 is a keyword.
       // For gcc, this is a typedef and not a keyword.
-      if(config.ansi_c.mode != configt::ansi_ct::flavourt::CLANG)
+      if(
+        config.ansi_c.mode != configt::ansi_ct::flavourt::CLANG &&
+        config.ansi_c.gcc__float128_type)
+      {
         code += "typedef " CPROVER_PREFIX "Float128 __float128;\n";
+      }
     }
     else if(config.ansi_c.arch == "ppc64le")
     {
@@ -222,8 +226,12 @@ void ansi_c_internal_additions(std::string &code)
       // https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html
       // For clang, __float128 is a keyword.
       // For gcc, this is a typedef and not a keyword.
-      if(config.ansi_c.mode != configt::ansi_ct::flavourt::CLANG)
+      if(
+        config.ansi_c.mode != configt::ansi_ct::flavourt::CLANG &&
+        config.ansi_c.gcc__float128_type)
+      {
         code+="typedef long double __float128;\n";
+      }
     }
 
     if(
--- a/src/ansi-c/gcc_version.cpp
+++ b/src/ansi-c/gcc_version.cpp
@@ -142,3 +142,23 @@ std::ostream &operator<<(std::ostream &out, const gcc_versiont &v)
 {
   return out << v.v_major << '.' << v.v_minor << '.' << v.v_patchlevel;
 }
+
+void configure_gcc(const gcc_versiont &gcc_version)
+{
+  // ISO/IEC TS 18661-3:2015 support was introduced with gcc 7.0
+  if(
+    gcc_version.flavor == gcc_versiont::flavort::GCC &&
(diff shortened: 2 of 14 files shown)


tautschnig:
I thought we were explicitly testing different Clang and GCC versions in some build configurations. Did I get this wrong?

kroening:
@tautschnig This fix here is independent of that, meaning we did the wrong thing in response to a particular version.

Independently, the current behavior is that goto-cc does check the gcc version, whereas cbmc does not. CBMC uses a default, which works for gcc >=4.3 and <7.0, i.e., this will fail with gcc >= 7. We could/should consider checking in cbmc as well.

src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -222,7 +223,8 @@ void ansi_c_internal_additions(std::string &code)
       // https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html
       // For clang, __float128 is a keyword.
       // For gcc, this is a typedef and not a keyword.
-      if(config.ansi_c.mode != configt::ansi_ct::flavourt::CLANG)
+      if(config.ansi_c.mode != configt::ansi_ct::flavourt::CLANG &&
+         config.ansi_c.gcc__float128_type)
         code+="typedef long double __float128;\n";
```
tautschnig, line 233: Nit pick: please add braces

src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -208,7 +208,8 @@ void ansi_c_internal_additions(std::string &code)
       // https://gcc.gnu.org/onlinedocs/gcc/Floating-Types.html
       // For clang, __float128 is a keyword.
       // For gcc, this is a typedef and not a keyword.
-      if(config.ansi_c.mode != configt::ansi_ct::flavourt::CLANG)
+      if(config.ansi_c.mode != configt::ansi_ct::flavourt::CLANG &&
+         config.ansi_c.gcc__float128_type)
         code += "typedef " CPROVER_PREFIX "Float128 __float128;\n";
```
tautschnig, line 215: Nit pick: please add braces

tautschnig (approved):
I'm assuming that our tests now actually test something.

smowton (approved):

tautschnig:
@kroening This is failing CI on CodeBuild (Linux).

kroening:
@tautschnig Yes, this is to be expected. I am investigating whether enabling _Float128 by default will break users who have gcc <7.0.

kroening:
@tautschnig Unfortunately, math.h in gcc-6 does the following:
`
typedef __float128 _Float128;
`
So we can either
a) special-case that typedef; or
b) check the gcc version in CBMC.

kroening:
#4636 is meant to prepare for b).

tautschnig:
@kroening #4636 is merged, so this one should become feasible.

kroening:
Now there!

kroening:
BTW, checking the gcc version will cost some performance. A viable alternative is to simply assume a version of gcc >=7, which is the case now for all major distributions.

src/ansi-c/gcc_version.cpp
```diff
@@ -142,3 +142,23 @@ std::ostream &operator<<(std::ostream &out, const gcc_versiont &v)
 {
   return out << v.v_major << '.' << v.v_minor << '.' << v.v_patchlevel;
 }
+
+void configure_gcc(const gcc_versiont &gcc_version)
+{
+  // ISO/IEC TS 18661-3:2015 support was introduced with gcc 7.0
+  if(
+    gcc_version.flavor == gcc_versiont::flavort::GCC &&
+    gcc_version.is_at_least(7))
```
tautschnig: Nit pick: below there's always an explicit `u` suffix, so should this be used here as well? I'd just like to see consistency.

tautschnig (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: c5b0bd1af93a8827960cff3719ec4f44b8cde143).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111867719

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c937b3f6061132f1c86fc6faea632e967b75eb1c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111868662

PR #4631, opened 2019-05-08
[TG-7706] Add new JSON value assignment feature for static fields

antlechner:
This new feature allows us to read values from a JSON file and use them in assignments in JBMC. This is currently used to replace the static initialiser, `clinit`, with a synthetic function `json_clinit` that is equivalent but only contains assignments to static fields. This only happens when the option `--static-values` is given. Without this option, there is no change in behaviour.

I'm not sure if it is necessary to include the core models in all the regression tests; it looks like they are required for enums, Strings and inheritance from Object. I can investigate this further but it should not delay reviews.

@allredj and @jeannielynnmoulton offered to review this PR. A code owner will also have to approve of course, but maybe the code owner review can be less in-depth after two full reviews.


jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -291,6 +297,65 @@ static bool needs_clinit_wrapper(
   return false;
 }
 
+static void create_function_symbol(
```
allredj, line 306: :pick: It would have been great to separate the factoring out into its own commit.

antlechner:
The AWS "cbmc" build is failing (but showing up as not starting in Github for some reason) because this is not yet based on https://github.com/diffblue/cbmc/pull/4625. I will rebase this PR when the test-gen -> cbmc pointer has been advanced, as this PR requires some changes in TG too.

jbmc/src/java_bytecode/synthetic_methods_map.h
```diff
@@ -27,6 +27,8 @@ enum class synthetic_method_typet
   /// These are generated for both user and stub types, to ensure the actual
   /// static initializer is only run once on any given path.
   STATIC_INITIALIZER_WRAPPER,
+  /// Only exists if the `--static-values` option was used.
```
allredj, line 30: A bit more description would be good.

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -207,6 +209,7 @@ static void clinit_wrapper_do_recursive_calls(
   const irep_idt &class_name,
   code_blockt &init_body,
   const bool nondet_static,
+  const std::string &static_values_file,
```
allredj: :no_entry_sign: Only pass a boolean here, the string itself isn't read. Also, it's not clear whether the string represents the whole file or just the path. :postal_horn:

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -493,6 +500,7 @@ code_blockt get_thread_safe_clinit_wrapper_body(
   const irep_idt &function_id,
   symbol_table_baset &symbol_table,
   const bool nondet_static,
+  const std::string &static_values_file,
```
allredj: :postal_horn:

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -685,6 +696,7 @@ code_ifthenelset get_clinit_wrapper_body(
...
```
allredj: :postal_horn:

jbmc/src/java_bytecode/java_static_initializers.h
```diff
@@ -48,6 +48,31 @@ code_ifthenelset get_clinit_wrapper_body(
   const select_pointer_typet &pointer_type_selector,
   message_handlert &message_handler);
 
+/// Creates the body of a json_clinit function, which includes assignments for
```
allredj: :pick: `Creates` -> `Create`

jbmc/src/java_bytecode/java_static_initializers.h
```diff
@@ -48,6 +48,31 @@ code_ifthenelset get_clinit_wrapper_body(
...
+/// all static fields of a class to values read from a JSON file. If the JSON
```
allredj: :bulb: Maybe clarify that the `class` we're talking about here is the one where `function_id` is declared (if I understood the code correctly). By the way, it's not clear why we pass a function. We should rather pass the class directly.

jbmc/src/java_bytecode/java_static_initializers.h
```diff
@@ -48,6 +48,31 @@ code_ifthenelset get_clinit_wrapper_body(
...
+/// the JSON file, the function will only include a call to the "real" clinit
+/// function, and not include any assignments itself. If an entry for this class
+/// is found but some of its static fields are not mentioned in the JSON file,
+/// those fields will be assigned default values.
+/// \param function_id: the id of this json_clinit function
+/// \param static_values_file: JSON file containing values of static fields. The
+///   format is expected to be a map whose keys are class names, and whose
+///   values are maps from static field names to values.
+/// \param symbol_table: used to look up and create new symbols
+/// \param message_handler: used log any errors with parsing the JSON
```
allredj: :pick: missing word

jbmc/src/java_bytecode/java_static_initializers.h
```diff
@@ -48,6 +48,31 @@ code_ifthenelset get_clinit_wrapper_body(
...
+/// \param needed_lazy_methods: used to mark any runtime types given in the JSON
+///   file as needed
+/// \param max_user_array_length: maximum value for constant or variable length
+///   arrays. Any arrays that were specified to be of nondeterministic length in
+///   the JSON file will be limited by this value.
+/// \param references:
```
allredj: remove this

jbmc/src/java_bytecode/java_static_initializers.h
```diff
@@ -48,6 +48,31 @@ code_ifthenelset get_clinit_wrapper_body(
...
+/// \param static_values_file: JSON file containing values of static fields. The
+///   format is expected to be a map whose keys are class names, and whose
+///   values are maps from static field names to values.
+/// \param symbol_table: used to look up and create new symbols
+/// \param message_handler: used log any errors with parsing the JSON
+/// \param needed_lazy_methods: used to mark any runtime types given in the JSON
+///   file as needed
+/// \param max_user_array_length: maximum value for constant or variable length
+///   arrays. Any arrays that were specified to be of nondeterministic length in
+///   the JSON file will be limited by this value.
```
allredj: remove this

jbmc/src/java_bytecode/java_static_initializers.h
```diff
@@ -48,6 +48,31 @@ code_ifthenelset get_clinit_wrapper_body(
...
+/// is found but some of its static fields are not mentioned in the JSON file,
+/// those fields will be assigned default values.
+/// \param function_id: the id of this json_clinit function
+/// \param static_values_file: JSON file containing values of static fields. The
+///   format is expected to be a map whose keys are class names, and whose
+///   values are maps from static field names to values.
+/// \param symbol_table: used to look up and create new symbols
+/// \param message_handler: used log any errors with parsing the JSON
+/// \param needed_lazy_methods: used to mark any runtime types given in the JSON
+///   file as needed
```
allredj: remove this

jbmc/src/java_bytecode/java_static_initializers.h
```diff
@@ -48,6 +48,31 @@ code_ifthenelset get_clinit_wrapper_body(
...
+/// \param max_user_array_length: maximum value for constant or variable length
+///   arrays. Any arrays that were specified to be of nondeterministic length in
+///   the JSON file will be limited by this value.
+/// \param references:
+/// \return the body of the json_clinit function as a code block.
+code_blockt get_json_clinit_body(
+  const irep_idt &function_id,
+  const std::string &static_values_file,
+  symbol_table_baset &symbol_table,
+  message_handlert &message_handler);
```
allredj: :pick:  An extra space between the declaration and the class would make it more readable.

jbmc/src/java_bytecode/synthetic_methods_map.h
```diff
@@ -28,6 +28,10 @@ enum class synthetic_method_typet
   /// static initializer is only run once on any given path.
   STATIC_INITIALIZER_WRAPPER,
   /// Only exists if the `--static-values` option was used.
+  /// If a given type has an entry in the file given by this option, the JSON
+  /// JSON static initializer contains a sequence of assignments from static
```
allredj: :pick: `JSON` is repeated

jbmc/src/java_bytecode/synthetic_methods_map.h
allredj: I see it comes later.

src/util/allocate_objects.h
```diff
@@ -71,16 +71,20 @@ class allocate_objectst
...
-  /// \return A dereference_exprt that dereferences the pointer to the allocated
-  ///   memory, or an empty expression when `allocate_type` is void
+  /// \return The pointer to the allocated memory, or an empty expression
+  ///   when `allocate_type` is void
+  exprt allocate_dynamic_object_symbol(
+    code_blockt &output_code,
+    const exprt &target_expr,
+    const typet &allocate_type);
+
+  /// Generates the same code as \ref allocate_dynamic_object_symbol, but
```
allredj: :pick: `Generate`

src/util/allocate_objects.h
```diff
@@ -71,16 +71,20 @@ class allocate_objectst
...
+  /// returns a dereference_exprt that dereferences the newly created pointer
```
allredj: :pick: `return`

allredj (commented):
Reviewed up  to "Refactor allocate_dynamic_object into two functions "

jbmc/src/java_bytecode/synthetic_methods_map.h
antlechner: Yeah, sorry for the confusion, the full description wouldn't have made much sense in this commit, maybe I should have put a `TODO` and deleted it later like I did elsewhere.

jbmc/src/java_bytecode/java_static_initializers.cpp
antlechner: You're right, hope it wasn't too confusing. I've split it into two commits now.

jbmc/src/java_bytecode/synthetic_methods_map.h
antlechner: I've done this now (though it's probably not helpful for you any more).

jbmc/src/java_bytecode/java_static_initializers.cpp
antlechner: 1) Done.
2) I added more documentation for this in the first commit, where the `--static-values` option is introduced: the field of `java_bytecode_languaget` now has some documentation, and the new option is also mentioned in the `--help` output.

jbmc/src/java_bytecode/java_static_initializers.h
antlechner: Thanks, the `function_id` was legacy stuff - replaced it with `class_id`.

jbmc/src/java_bytecode/java_static_initializers.h
antlechner: Thanks, forgot about these while splitting existing work into commits. They will re-appear in a later commit.

antlechner:
@allredj I addressed all of your comments so far. :+1:

jbmc/src/java_bytecode/Makefile
```diff
@@ -1,4 +1,5 @@
-SRC = bytecode_info.cpp \
+SRC = assignments_from_json.cpp \
```
allredj, line 1: :pick: Commit message: remove `whose`.

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -87,6 +87,10 @@ void parse_java_language_options(const cmdlinet &cmd, optionst &options)
     options.set_option(
       "java-cp-include-files", cmd.get_value("java-cp-include-files"));
   }
+  if(cmd.isset("static-values"))
```
jeannielynnmoulton, line 90: :bulb: is something like `--external-static-value` more descriptive?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+followed_class_type(const exprt &expr, const symbol_table_baset &symbol_table)
+{
+  const pointer_typet &pointer_type = to_pointer_type(expr.type());
+  const java_class_typet &java_class_type =
+    to_java_class_type(namespacet{symbol_table}.follow(pointer_type.subtype()));
+  return java_class_type;
+}
+
+static bool
+has_array_type(const exprt &expr, const symbol_table_baset &symbol_table)
```
jeannielynnmoulton: :hammer_and_pick: Remove this and use `is_java_array_tag` to reduce duplication

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+{
+  return has_prefix(
+    id2string(followed_class_type(expr, symbol_table).get_tag()),
+    "java::array[");
+}
+
+static bool is_enum_type(const java_class_typet &class_type)
```
jeannielynnmoulton: :hammer_and_pick: remove this and use `get_is_enumeration` from `java_class_type`

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+static bool is_reference(const jsont &json)
+{
+  if(!json.is_object())
+    return false;
+  const auto &json_object = to_json_object(json);
+  return json_object.find("@ref") != json_object.end();
+}
+
+/// Returns the unique ID of all objects that are reference-equal to this one.
+/// See \ref has_id and \ref is_reference.
```
jeannielynnmoulton, line 155: Why if it doesn't have an id do we return the reference? This isn't clear to me (though maybe it will become so later)

Maybe two functions that return optionals would be better than this and the two others? e.g.
get_id return optionalt<std::string> and get_reference returns optional<std::string> :cactus:

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+    return false;
+  const auto &json_object = to_json_object(json);
+  return json_object.find("@id") != json_object.end();
+}
+
+/// Returns true iff the argument has a "@ref" key.
+/// The corresponding value is the unique ID of all objects that are reference-
+/// equal to this one.
+/// Any other key-value pairs of `json` will be ignored.
+static bool is_reference(const jsont &json)
```
jeannielynnmoulton, line 145: :cactus:

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+  return json_object.find("@type") != json_object.end();
+}
+
+/// Returns true iff the argument has a "@id" key.
+/// The presence of such a key means that there exist objects that are
+/// reference-equal to this object.
+/// The corresponding value is the unique ID of all objects that are reference-
+/// equal to this one.
+/// All other key-value pairs of `json` should be as usual.
+static bool has_id(const jsont &json)
```
jeannielynnmoulton, line 133: :cactus:

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+    info.block.add(code_assumet{binary_predicate_exprt{
+      length_expr, ID_ge, from_integer(0, java_int_type())}});
+    allocate_array(expr, length_expr, info);
+  }
+  const json_arrayt json_array = get_untyped_array(json, element_type);
+  const auto number_of_elements =
+    from_integer(json_array.size(), java_int_type());
+  info.block.add(code_assumet{
+    binary_predicate_exprt{length_expr, ID_ge, number_of_elements}});
+  if(has_nondet_length(json))
```
jeannielynnmoulton, line 481: optional get_nondet length instead of has and the relevant code in the block below? :cactus:

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+  return followed_class_type(expr, symbol_table) == declaring_class_type &&
+         is_enum_type(declaring_class_type);
+}
+
+/// Returns true iff the argument has a "@type" key.
+/// A runtime type that is different from the objects compile-time type should
+/// be specified in `json` in this way.
+/// Type values are of the format "my.package.name.ClassName".
+static bool has_type(const jsont &json)
+{
```
jeannielynnmoulton, line 118: :cactus: less impact than other below but optional could return json["@type"].value?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+/// \param type_from_array: may contain an element type name given by a
+///   containing array. Such types are stored in the form
+///   "Lmy.package.name.ClassName;".
+/// \param symbol_table: used to look up the type given its name.
+/// \return runtime type of the object, if specified by at least one of the
+///   parameters.
+static optionalt<java_class_typet> given_runtime_type(
+  const jsont &json,
+  const optionalt<std::string> &type_from_array,
+  symbol_table_baset &symbol_table)
```
jeannielynnmoulton: const!

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+{
+  std::string runtime_type;
+  if(!has_type(json) && !type_from_array)
+    return {};
+  if(has_type(json))
+    runtime_type = "java::" + json["@type"].value;
+  else
+  {
+    PRECONDITION(
```
jeannielynnmoulton: :pick: INVARIANT with helpful message?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+/// \param type_from_array: may contain a type name from a containing array.
+/// \return if the type of an array was given, the type of its elements.
+static optionalt<std::string> element_type_from_array_type(
+  const jsont &json,
+  const optionalt<std::string> &type_from_array)
+{
+  if(has_type(json))
+  {
+    const std::string &json_array_type = json["@type"].value;
+    PRECONDITION(json_array_type.find('[') == 0);
```
jeannielynnmoulton: :pick: INVARIANT with message? :ice_skate:

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+    return json_array_type.substr(1);
+  }
+  else if(type_from_array)
+  {
+    PRECONDITION(type_from_array->find('[') == 0);
```
jeannielynnmoulton: :ice_skate:

jbmc/src/java_bytecode/java_utils.cpp
```diff
@@ -183,6 +183,21 @@ irep_idt resolve_friendly_method_name(
   }
 }
 
+/// Given a pointer type to a Java class and a type representing a more specific
```
jeannielynnmoulton: docs belong in .h file

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+/// For characters, the encoding in \p json is assumed to be UTF-8.
+/// See \ref assign_from_json_rec.
+static void assign_primitive_from_json(
+  const exprt &expr,
+  const jsont &json,
+  code_blockt &block)
+{
+  if(json.is_null()) // field is not mentioned in json, leave as default value
+    return;
+  if(expr.type() == java_boolean_type())
```
jeannielynnmoulton, line 341: In the factory, we sometimes can see a `bool_typet`. My memory is failing me, but the commit says `Allow state_factory to interpreter non-java bool`. Since this is java, I guess it is probably OK, I just thought I'd mention it in case you even see this problem :)

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+      block.add(code_assignt{expr, true_exprt{}});
+    else
+      block.add(code_assignt{expr, false_exprt{}});
+  }
+  else if(
+    expr.type() == java_int_type() || expr.type() == java_byte_type() ||
+    expr.type() == java_short_type() || expr.type() == java_long_type())
+  {
+    block.add(
+      code_assignt{expr, from_integer(std::stoi(json.value), expr.type())});
```
jeannielynnmoulton: :question: / :no_entry_sign: Will std::stoi work for very large values of a long?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+{
+  if(!json.is_object())
+    return false;
+  const auto &json_object = to_json_object(json);
+  if(json_object.find("@nondetLength") != json_object.end())
+    return (json["@nondetLength"].is_true());
+  return false;
+}
+
+/// For typed versions primitive, string or array types, looks up their untyped
```
jeannielynnmoulton: :question: I don't understand these docs.

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
+/*******************************************************************\
```
jeannielynnmoulton, line 1: I now understand the typed vs untyped this as full json object vs value, so I think a better name for `get_untyped*` might be `get_value*`

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+  const exprt &expr,
+  const jsont &json,
+  det_creation_infot &info)
+{
+  const java_class_typet &java_class_type =
+    to_java_class_type(namespacet{info.symbol_table}.follow(expr.type()));
+  for(const auto &component : java_class_type.components())
+  {
+    const irep_idt &component_name = component.get_name();
+    if(
```
jeannielynnmoulton, line 524: :bulb: in the java_trace_interpreter there is a function that does exactly this. It would be cool maybe if this existed in jbmc instead and both instances replaced?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+      component_name == "@class_identifier" ||
+      component_name == "cproverMonitorCount")
+    {
+      continue;
+    }
+    member_exprt member_expr{expr, component_name, component.type()};
+    if(component_name[0] == '@') // component is parent struct type
```
jeannielynnmoulton, line 531: :hammer_and_pick: : These two separate branches are probably unnecessary, as simply calling assign_from_json_rec should handle both cases just fine. The this could be really simple:

```
if(!internal_field())
  assign_from_json_rec()
```

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+      assign_struct_components_from_json(member_expr, json, info);
```
jeannielynnmoulton, line 532: :question: if ignoring above, why does this accept json and not the json of the member?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+  const exprt &expr,
+  const symbol_table_baset &symbol_table,
+  const java_class_typet &declaring_class_type)
+{
+  PRECONDITION(can_cast_type<pointer_typet>(expr.type()));
+  return followed_class_type(expr, symbol_table) == declaring_class_type &&
+         is_enum_type(declaring_class_type);
+}
+
+/// Returns true iff the argument has a "@type" key.
```
allredj: :pick: `Return`

jbmc/src/java_bytecode/assignments_from_json.cpp
allredj: `versions primitive`?

jbmc/src/java_bytecode/assignments_from_json.h
```diff
@@ -0,0 +1,92 @@
...
+///   things more complicated.
+/// For examples of JSON representations of objects, see the regression tests
+/// for this feature in
+/// jbmc/regression/jbmc/deterministic_assignments_json.
+///
+/// Known limitations:
+/// - If two reference-equal objects are defined in two different functions, and
+///   following the specification of json-io, one of them is stored as
+///   `{"@ref":1}` or similar, then that object is assigned a default value and
+///   the reference is ignored. If both objects are defined in the same function
```
allredj: I don't understand that sentence. What is that "default value"?

jbmc/src/java_bytecode/assignments_from_json.cpp
allredj: What is an "untyped primitive"? Why do we need them?

jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
```diff
@@ -0,0 +1,424 @@
...
+        "\u0000\u0001AB?;⿃"
+      ]
+    },
+    "fibCharField":{
+      "@type":"java.lang.Character",
+      "value":"ɢ"
+    },
+    "longField":{
+      "@type":"java.lang.Long",
+      "value":"202630"
```
jeannielynnmoulton: :no_entry_sign: Please make this long field bigger than the max int value to test how you are using std::stoi on previous commit

jeannielynnmoulton (changes requested):
Would like to see the std::stoi comment tested and addressed. There is maybe some room for simplification in some places that I suggested, but these are not blocking.

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+///   `@type` field, this takes priority over \p type_from_array. Types for non-
+///   array objects are stored in the JSON in the format
+///   "my.package.name.ClassName".
+/// \param type_from_array: may contain an element type name given by a
+///   containing array. Such types are stored in the form
+///   "Lmy.package.name.ClassName;".
+/// \param symbol_table: used to look up the type given its name.
+/// \return runtime type of the object, if specified by at least one of the
+///   parameters.
+static optionalt<java_class_typet> given_runtime_type(
```
allredj: Why not just `runtime_type`?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+  else
+  {
+    PRECONDITION(
+      type_from_array->find('L') == 0 &&
+      type_from_array->rfind(';') == type_from_array->length() - 1);
+    runtime_type =
+      "java::" + type_from_array->substr(1, type_from_array->length() - 2);
+  }
+  if(!symbol_table.has_symbol(runtime_type))
+    return {}; // Fall back to compile-time type if runtime type was not found
```
allredj, line 282: How frequent is that? Should we output a warning message?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+}
+
+/// Entry point of the recursive deterministic assignment algorithm.
+/// \param expr: expression to assign a deterministic value to. In the case of
+///   the entry point, this is either a pointer to a struct, or an expression
+///   corresponding to a Java primitive.
+/// \param json: a JSON representation of the deterministic value to assign.
+/// \param type_from_array: if `expr` was found as an element of an array,
+///   the element type of this array.
+/// \param info: references used throughout the recursive algorithm.
```
allredj: :pick: I think that documentation should be at definition, as this function is not exported.

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+  code_blockt &block)
+{
+  if(json.is_null()) // field is not mentioned in json, leave as default value
+    return;
+  if(expr.type() == java_boolean_type())
+  {
+    if(json.is_true())
+      block.add(code_assignt{expr, true_exprt{}});
+    else
+      block.add(code_assignt{expr, false_exprt{}});
```
allredj, line 346: :pick: Could be written down as a single line.
```
block.add(code_assignt{expr, json.is_true()?true_exprt{}:false_exprt{}})
```
Not sure it's more readable though....

jbmc/src/java_bytecode/java_bytecode_language.cpp
antlechner: @forejtv suggested the `--static-values` name because it's short (I think).

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+/// For the overall algorithm, see \ref assign_from_json_rec.
+static void assign_array_from_json(
+  const exprt &expr,
+  const jsont &json,
+  const optionalt<symbol_exprt> &given_length_expr,
+  const optionalt<std::string> &type_from_array,
+  det_creation_infot &info)
+{
+  PRECONDITION(has_array_type(expr, info.symbol_table));
+  PRECONDITION(can_cast_type<pointer_typet>(expr.type()));
```
allredj: This precondition is not explicitly needed, as you do the actual cast in the following line.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Sounds like a good idea but probably best left for a follow-up PR.

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+/// function assigns the data component of the array, which contains the array
+/// elements. \p expr is a pointer to the array containing the component.
+static void assign_array_data_component_from_json(
+  const exprt &expr,
+  const jsont &json,
+  const optionalt<std::string> &type_from_array,
+  det_creation_infot &info)
+{
+  const auto &java_class_type = followed_class_type(expr, info.symbol_table);
+  const auto &components = java_class_type.components();
```
allredj: For clarity, you could directly get `java_class_type.components()[2]` here and name the variable appropriately. I admit I don't know what element `2` contains.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Ah, I should have made this clearer than just the short comment. If `component_name` starts with `@` then it has a struct type representing the parent class. I can't send this back to `assign_from_json_rec` because the latter expects pointer types or primitives, but not structs. (And I don't want to give it a struct case because I think it's confusing and requires passing around a boolean or defining more functions to keep this "struct storing components of the parent" as a special case - the object factory does it like this and it's not pretty.) All other components are fields that are declared in `java_class_type`, and those have pointer types and go back to the recursive case.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Again, the "member" here is not actually a class member, but a struct storing all components of the parent type. components = (components of parent) + class members. I agree that this is confusing.
So in this case, there is no member to look up in the JSON file. We're recursively going through parent types here to find inherited fields. Only in the `else` case have we actually found a field.

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+    member_exprt{dereference_exprt{expr}, "data", components[2].type()},
+    pointer_type(element_type));
+
+  const symbol_exprt &array_init_data =
+    info.allocate_objects.allocate_automatic_local_object(
+      data_member_expr.type(), "det_array_data_init");
+  const code_assignt data_assign{array_init_data, data_member_expr, info.loc};
+  info.block.add(data_assign);
+
+  size_t index = 0;
```
allredj, line 407: :pick: I think this can be defined inside the for statement.

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+  const optionalt<std::string> inferred_element_type =
+    element_type_from_array_type(json, type_from_array);
+  const json_arrayt json_array = get_untyped_array(json, element_type);
+  for(auto it = json_array.begin(); it < json_array.end(); it++, index++)
+  {
+    const dereference_exprt element_at_index = array_element_from_pointer(
+      array_init_data, from_integer(index, java_int_type()));
```
allredj, line 414: :question: Why do we use `dereference_exprt` instead of `index_exprt` here?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+  {
+    info.block.add(code_assumet{binary_predicate_exprt{
+      length_expr,
+      ID_le,
+      from_integer(info.max_user_array_length, java_int_type())}});
+  }
+  else
+  {
+    info.block.add(code_assumet{
+      binary_predicate_exprt{length_expr, ID_le, number_of_elements}});
```
allredj, line 491: If I understand correctly, in the deterministic case, we add these two statements:
`length_expr <= number_of_elements`
`length_expr >= number_of_elements`
Why add an equality assumption in that case instead?

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: To give an example, `MyClassWithInheritedFields` from the regression tests has the following components:
```
  * components: 
    0: 
      * type: struct_tag
          * identifier: java::com.diffblue.example.MyParent
      * name: @com.diffblue.example.MyParent
      * base_name: @com.diffblue.example.MyParent
      * pretty_name: @com.diffblue.example.MyParent
    1: 
      * type: signedbv
          * width: 32
      * name: childField
      * base_name: childField
      * access: default
      * pretty_name: childField
      * final: 0
```
Component 0 is the parent type, and component 1 is a pointer to `childField`.
We then call `assign_struct_components_from_json` on `MyParent`, which gives us the components
```
  * components: 
    0: 
      * type: struct_tag
          * identifier: java::java.lang.Object
      * name: @java.lang.Object
      * base_name: @java.lang.Object
      * pretty_name: @java.lang.Object
    1: 
      * type: signedbv
          * width: 32
      * name: parentField
      * base_name: parentField
      * access: default
      * pretty_name: parentField
      * final: 0
    2: 
      * type: pointer
          * width: 64
          * #reference: 1
          0: struct_tag
              * identifier: java::java.lang.String
              * #base_name: java.lang.String
      * name: parentString
      * base_name: parentString
      * access: default
      * pretty_name: parentString
      * final: 0
```
So we find all the fields declared in `MyParent` and also see that it inherits directly from Object. Then we call this function one more time on the `Object` type.

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,828 @@
...
+  pointer_typet replacement_pointer_type =
+    pointer_to_replacement_type(pointer_type, runtime_type);
+  if(!equal_java_types(pointer_type, replacement_pointer_type))
+  {
+    const auto &new_symbol =
+      info.allocate_objects.allocate_automatic_local_object(
+        replacement_pointer_type, "det_subtype_symbol");
+    if(info.needed_lazy_methods)
+      info.needed_lazy_methods->add_all_needed_classes(
+        replacement_pointer_type);
```
allredj, line 661: Braces for multi-line `if`s

jbmc/src/java_bytecode/assignments_from_json.h
```diff
@@ -0,0 +1,92 @@
...
+///   precision when representing longs.
+/// This rule has the following exceptions:
+/// - It seems that strings are always printed in "primitive" representation by
+///   json-io, i.e.\ they are always JSON strings, and never JSON objects with
+///   a `@type` key. For cases where we don't know that an expression has a
+///   string type (e.g.\ if its type is generic and specialized to
+///   java.lang.String), we need to sometimes represent strings as JSON objects
+///   with a `@type` key. In this case, the content of the string will be the
+///   value associated with a `value` key (similarly to StringBuilder in
+///   json-io).
```
allredj: That processing of strings is not very apparent in the code. It seems like strings are processed in the "reference" case (in `assign_struct_from_json`). So how does this special case translate into your implementation?

jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
```diff
@@ -0,0 +1,424 @@
...
+      "name":"WHITE","ordinal":3,"grayscale":{
+        "@type":"java.lang.Boolean",
+        "value":true
+      }
+    }
+  },
+  "OtherClass":{
+    "otherField":{"@ref":1}
+  }
+}
```
allredj, line 424: It would be great to include the json-io command line that generated that file. And also the json-io version. Maybe in a README in this folder.

allredj (approved):
Stunning!
A few comments and questions, but I can't find much to complain about.
I can't wait to see this in action on our benchmarks!

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: I thought about that, both here and in the enum case, which has a fallback if we can't find `$VALUES` in the symbol table. I thought I'd add warnings in a follow-up PR, is that ok?

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: I tried this before, sadly it doesn't work because C++ wants the expressions in a ternary to have _exactly_ the same type. The types here are different (`true_exprt` and `false_exprt`). The way to get around this to write the whole thing in one statement would be to use an immediately invoked lambda and specify its return type, but in this case this really wouldn't make it more readable:
```
    block.add(code_assignt{expr, [&]() -> exprt {
                             if(json.is_true())
                               return true_exprt{};
                             return false_exprt{};
                           }()});
 ```

Comment (antlechner) [reply jbmc/src/java_bytecode/assignments_from_json.cpp:346]: Maybe slightly nicer:
```
    const exprt boolean_expr = [&]() -> exprt {
      if(json.is_true())
        return true_exprt{};
      return false_exprt{};
    }();
    block.add(code_assignt{expr, boolean_expr});
 ```
Any preference?

allredj: I see. Don't change it then.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Took me a while to figure this one out myself. It's because we're dealing with a pointer array here, i.e. the array is represented as a pointer to its first element, and other elements can be accessed with pointer offsets (the `plus_exprt` in the definition of `array_element_from_pointer`). `index_exprt`s represent indexing of "actual" arrays. Java arrays are always represented as pointer arrays.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: That was something I considered - doing it this way seemed easier to do since one of the constraints is always necessary, and I'm not sure if an equality constraint would really simplify things for the solver?

jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
antlechner: The version is mentioned in `assignments_from_json.h` but I can mention it again in the README. I did use a script to generate this file but it's messy and not really merge-able, and I did some manual changes after running it... I thought we decided that the script wasn't necessary for this PR? But if you think it would be useful, I'll tidy it up and add it to this commit.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Interesting, and quite confusing (`java_boolean_type()` returns a `c_bool_typet`, which is unrelated to `bool_typet` in the `typet` type hierarchy, even though a Java type probably isn't a type of C type...). Let's hope that this only happens for variables that aren't directly translated from Java code, e.g. `clinit_wrapper_already_run` or similar things that we made up.

tautschnig:
@antlechner Could you please sync with @hannes-steffenhagen-diffblue and @chrisr-diffblue (and possibly others) about the following: isn't this repeating some effort that is going into `goto-harness`? Specifically, in `goto-harness` there already is support for reading a JSON memory snapshot. I acknowledge that `goto-harness` wouldn't work with Java, because we assume goto binaries to exist. But using JSON to represent initial values seems like a commonality?

jbmc/src/java_bytecode/assignments_from_json.cpp
peterschrammel: `c_bool_typet` is the C99 `_Bool` type. The representation of `_Bool` is not defined in the standard. It only needs to have at least one bit. Typically, it's encoded by 8 bits. 
Also, the representation of a Java `boolean` is not defined in the JVM spec (but e.g. in OpenJDK it's 8 bits). Hence, for all practical purposes it seemed fine to use `c_bool_typet` for Java `boolean`.
`bool_typet` is a 1-bit boolean, which is not mapped to any C nor Java type.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 3dc728f332a7d0d75fec1f8190b6bd54be51e5a7).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111638108

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: `is_java_array_type` is exactly what I needed - thanks, I removed the new function.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Done. `has_enum_type` and `followed_class_type` are still there; I am planning to move those to `util` in follow-up PRs.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Done here and for `get_enum_id`.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: I had a look but apparently this isn't possible (or at least not without making it much less readable) as `index` and `it` have different types.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: You're right, it will not. Changed `stoi` to `stol` and all seems fine now.

jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
antlechner: Now equal to the max long value (`9_223_372_036_854_775_807L`).

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: `has_nondet_length` returns `true` if there is a `"@nondetLength": true` entry for this object. We only care if the value is `true` or not, so all we need is a boolean.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: ... or at least it does on Linux. Windows does not seem happy with this so I'll have to use a more precisely defined type.

tautschnig: How about using one of the wrappers from `util/string2int.h`?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 26ee2324786058dd88ee08d671cfbd75c069c183).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111899133

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Done: renamed it to `get_type` and it returns an `optionalt<std::string>`. I also used this to simplify `given_runtime_type` a bit and make sure all local variables in it are `const`.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: According to https://en.cppreference.com/w/cpp/language/types, `long long` has to have at least 64 bits. So I tried `std::stoll`, at that seems to do the trick.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 867ab4353a8962b346e5b9261be5fa1a778824ec).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111936093

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: I tried changing this to use optionals but whatever I do, the result seems to be more confusing than what I have now, so I'll leave it for now. We have to distinguish between objects that have `@id` fields and `@ref` fields (we want to allocate a new symbol only if we see an `@id` field), and we need the actual value of the ID (value associated with an `@id` or `@ref` field) only in one place.
I did change the name of `get_id` to `get_id_or_reference_value` and added a line of Doxygen to it and a short description of how the ref/id stuff works in the header file so hopefully this makes things clearer.
Generally there is no difference between "the id" and "the reference" - there is just one ID for a group of reference-equal objects, and it is stored as the value of either `@id` or `@reference`.
I hope this makes things clearer.

martin-cs:
@antlechner : I seem to have got a review request for this, although I am not quite sure why.  It looks like it has picked up some unrelated commits on the way.  Is there any chance you could rebase to loose these?  If, after that, there are bits that I should be reviewing, I am happy to do so.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: I added documentation about this to `assign_from_json` in the header file (under "Some examples of json-io...").
On this function I linked back to this new documentation. Hope it makes more sense now.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: I don't like `get_value` because the key is called `value` for most types but `@items` for arrays (that's a json-io thing so I didn't want to change it), and the name also doesn't specify _which_ value we're getting (the object without an `@type`). So I'll leave the terminology as it is, but with the definitions in the header file it should hopefully be clearer now.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: 1) Fixed: `versions of primitive...`
2) @jeannielynnmoulton had the same question. This is part of the json-io documentation but it clearly helps to have the basics documented here as well. I added documentation for this in the header (under "Some examples of json-io...").

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Because of how I'm calling it:
```
      const auto runtime_type =
        given_runtime_type(json, type_from_array, info.symbol_table))
```
I can't call both the variable and the function `runtime_type`.

jbmc/src/java_bytecode/assignments_from_json.h
antlechner: `assign_struct_from_json` has a call to `assign_string_from_json`, which has a call to `get_untyped_string`, which looks up the `value` key if it exists.
I added a `\ref get_untyped_string` to the docs here.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: I think the components of ~a java.lang.String~ an array type are:
0: java.lang.Object struct type (parent type)
1: length
2: data
I followed your suggestion and renamed the variable to `data_component`.

jbmc/src/java_bytecode/assignments_from_json.h
antlechner: Good question - I rewrote this part and I _think_ it's correct now...

antlechner:
@martin-cs I'm not quite sure how that happened: yesterday only my commits showed up, then today there were various other commits mixed in. A rebase fixed it, and I'll remove you from the list of reviewers. :slightly_smiling_face:

antlechner:
@allredj @jeannielynnmoulton All review comments addressed. This is ready for re-review.
@tautschnig I talked to them before I made this PR, as the two features both read from a JSON file. However, the context and format of the JSON file are quite different in the two cases so it's not obvious how to unify the approach. I'm not very familiar with `goto-harness` but if I find anything in the future that can be refactored, I will do so.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: f9a8be6466a8fe4b72d37cdfd4907fbb6d4ee86f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112121100

tautschnig:
> @tautschnig I talked to them before I made this PR, as the two features both read from a JSON file. However, the context and format of the JSON file are quite different in the two cases so it's not obvious how to unify the approach.

Sounds good. The only bit not clear to me is: are the formats of the JSON files _necessarily_ different?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a39b3aaeca00c8b3739b93f41771e1bd97c55419).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112210229

jeannielynnmoulton (approved):

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -69,65 +69,55 @@ followed_class_type(const exprt &expr, const symbol_table_baset &symbol_table)
...
   const java_class_typet &declaring_class_type)
 {
   PRECONDITION(can_cast_type<pointer_typet>(expr.type()));
   return followed_class_type(expr, symbol_table) == declaring_class_type &&
-         is_enum_type(declaring_class_type);
+         declaring_class_type.get_is_enumeration();
 }
 
-/// Returns true iff the argument has a "@type" key.
+/// Return true iff the argument has a "@type" key.
```
allredj: :no_entry_sign: Documentation needs updating. It's now an optional.

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -257,25 +255,24 @@ static jsont get_untyped_string(const jsont &json)
...
-  else
-  {
-    PRECONDITION(
+  const symbol_table_baset &symbol_table)
+{
+  const auto runtime_type = [&]() -> optionalt<std::string> {
+    if(const auto type_from_json = get_type(json))
+      return "java::" + *type_from_json;
+    if(!type_from_array)
+      return {};
```
allredj: :pick: Why not pull this test out of the lambda? Obviously `get_untyped` should return `{}` whenever `type_from_array` is empty.

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -257,25 +255,24 @@ static jsont get_untyped_string(const jsont &json)
...
-    runtime_type =
-      "java::" + type_from_array->substr(1, type_from_array->length() - 2);
-  }
-  if(!symbol_table.has_symbol(runtime_type))
+        type_from_array->rfind(';') == type_from_array->length() - 1,
+      "Types inferred from the type of a containing array should be of the "
+      "form Lmy.package.name.ClassName;");
+    return "java::" + type_from_array->substr(1, type_from_array->length() - 2);
+  }();
+  if(!runtime_type || !symbol_table.has_symbol(*runtime_type))
```
allredj: :pick: For code clarity I would put the `!runtime_type` test before the invariant check.

allredj (changes requested):
One documentation issue. Otherwise looks good.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Changed to
```
/// If the argument has a "@type" key, return the corresponding value, else
/// return an empty optional.
```

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Done. The `if(!type_from_json && !type_from_array)` check is done before the lambda now, and the lambda returns a std::string.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Not sure if I understand this comment but `runtime_type` is no longer an optional so this part of the test is gone.

antlechner:
@allredj I updated the documentation. I also squashed the fixup commits.

jbmc/src/java_bytecode/java_bytecode_language.h
```diff
@@ -71,7 +72,11 @@ Author: Daniel Kroening, kroening@kroening.com
   "                              METHODNAME can be a regex that will be matched\n" /* NOLINT(*) */ \
   "                              against all symbols. If missing a java:: prefix\n"    /* NOLINT(*) */ \
   "                              will be added. If no descriptor is found, all\n"/* NOLINT(*) */ \
-  "                              overloads of a method will also be added.\n"
+  "                              overloads of a method will also be added.\n" \
+  " --static-values f            Load initial static values from given JSON file.\n"/* NOLINT(*) */ \
```
smowton: static field values

jbmc/src/java_bytecode/java_bytecode_language.h
```diff
@@ -71,7 +72,11 @@ Author: Daniel Kroening, kroening@kroening.com
...
+  "                              Assignments to these values replace calls to clinit.\n"/* NOLINT(*) */ \
```
smowton: Static fields are assigned these values instead of calling the normal static initializer (clinit) method

allredj (approved):

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,837 @@
...
+#include <util/array_element_from_pointer.h>
+#include <util/expr_initializer.h>
+#include <util/prefix.h>
+#include <util/unicode.h>
+
+/// Values passed around between most functions of the recursive deterministic
+/// assignment algorithm entered from \ref assign_from_json.
+/// The values in a given `det_creation_infot` are never reassigned, but the
+/// ones that are not marked `const` may be mutated.
+struct det_creation_infot
```
smowton: Suggest naming this `object_creation_infot`, `det`, presumably short for "deterministic", isn't really used in cbmc apart from its antonym "nondet".

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,837 @@
...
+{
+  return followed_class_type(expr, symbol_table).get_is_enumeration();
+}
+
+/// This function is used as a workaround until reference-equal objects defined
+/// across several classes are tracked correctly. Once reference-equality works
+/// in all cases, this function can be removed.
+/// Until then, in the case of an enum expression that needs to be assigned a
+/// value, we distinguish between two cases:
+/// 1) the enum expression is declared in its own type - in this case,
```
smowton: Not sure what "declared in its own type" means, could you expand the comment with an example?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,837 @@
...
+    set_class_identifier(
+      to_struct_expr(*initial_object),
+      ns,
+      struct_tag_typet("java::" + id2string(java_class_type.get_tag())));
+    info.block.add(code_assignt{expr, *initial_object});
+    assign_struct_components_from_json(expr, json, info);
+  }
+}
+
+/// Same as \ref assign_pointer_from_json without special cases (enums).
```
smowton, line 568: `assign_non_enum_pointer_from_json` then?

smowton (approved):
Some naming and doc suggestions; nothing blocking. Most prominently:

1. There are various places where this thing is called `json_` when actually the fact that it was specified as JSON is irrelevant. For example, `create_object_from_json` is fine, but the thing that results is just an `object`, not an `object_from_json`. `json_clinit` itself I'd suggest could just be `synthetic_clinit` or `replacement_clinit` or `user_specified_clinit` or...

2. Along similar lines, suggest not using the `det_` prefix, which isn't widely used in cbmc to refer to not-nondet things (nondet is the exceptional case, not-nondet things are just... things :)) The det/nondet nomenclature is a peculiarity of test-gen / the models library and likely wouldn't be readily understood by a cbmc developer.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: b9153f0de76f3e74858f50feb69de88e06da86ac).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112231161

jbmc/src/java_bytecode/java_bytecode_language.h
antlechner: I changed it to something very similar to your suggestion.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Changed it to `object_creation_infot`. Similarly, I also renamed `det_creation_referencet` (in `assignments_from_json.h`) to `object_creation_referencet`.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Rephrased it, added examples, and renamed it to `is_enum_with_type_equal_to_declaring_type` (long name but more descriptive and it's only used in one place anyway; let me know if you have an idea for a better name).

jbmc/src/java_bytecode/assignments_from_json.cpp
allredj: It's not ideal because it feels that `runtime_type` and `given_runtime_type` are two different things. What about `runtime_type_from_json`?

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -207,6 +209,7 @@ static void clinit_wrapper_do_recursive_calls(
   const irep_idt &class_name,
   code_blockt &init_body,
   const bool nondet_static,
+  const bool use_json_clinit,
```
peterschrammel: I don't like `json` in the name as the source format of the data is irrelevant; it could as well be read from an XML file  or whatever.

jbmc/src/java_bytecode/synthetic_methods_map.h
```diff
@@ -27,6 +27,9 @@ enum class synthetic_method_typet
   /// These are generated for both user and stub types, to ensure the actual
   /// static initializer is only run once on any given path.
   STATIC_INITIALIZER_WRAPPER,
+  /// Only exists if the `--static-values` option was used.
+  /// TODO Currently has no content, future commits will add it.
+  JSON_STATIC_INITIALIZER,
```
peterschrammel: `STATIC_INITIALIZER_VALUES` ?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -0,0 +1,843 @@
...
+/// return an empty optional.
+/// A runtime type that is different from the objects compile-time type should
+/// be specified in `json` in this way.
+/// Type values are of the format "my.package.name.ClassName".
+static optionalt<std::string> get_type(const jsont &json)
+{
+  if(!json.is_object())
+    return {};
+  const auto &json_object = to_json_object(json);
+  if(json_object.find("@type") == json_object.end())
```
peterschrammel, line 122: Does this comply with https://google.github.io/styleguide/jsoncstyleguide.xml ?

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -10,11 +10,12 @@ Author: Chris Smowton, chris.smowton@diffblue.com
 #include "java_object_factory.h"
 #include "java_utils.h"
 #include <goto-programs/class_hierarchy.h>
-#include <util/std_types.h>
-#include <util/std_expr.h>
+#include <json/json_parser.h>
```
peterschrammel, line 13: I'd prefer if the use of JSON was limited to the assignments_from_json class.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: Do you mean the `@`? It's directly taken from the `json-io` library, and in the Google style guide it says "JSON maps can use any Unicode character in key names."

jbmc/src/java_bytecode/assignments_from_json.cpp
allredj: What shall we do to not forget about that follow-up PR?

jbmc/src/java_bytecode/assignments_from_json.cpp
allredj: From what I understand about CDCLs, an equality expression should be better as it allows boolean constraint propagation. An equality is also easier to read when debugging the VCCs, and it wouldn't make the present code any more complicated. I suggest a follow up PR for that.

jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
allredj: I think having the version here would be good. It's all about understanding where this file comes from, and how we can reproduce it. It's unfortunate that you had to do some manual tweaks. What is the extend of those?

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: I renamed the function to `runtime_type` and I'm now calling it like
```
      const auto runtime_type =
        ::runtime_type(json, type_from_array, info.symbol_table))
```
looking up `runtime_type` in the global namespace. Looks like this will do the trick (thanks @jeannielynnmoulton ).

antlechner:
> The only bit not clear to me is: are the formats of the JSON files _necessarily_ different?

@tautschnig I would have to go through the details of `goto-harness` to answer this question. The format I used in this PR is almost the same as what you get if you run `JsonWriter.objectToJson` from the `json-io` library on a map that stores field names and values. The output from other Java to JSON serialising libraries (e.g. Gson) is similar. So I think the format of JSON files for the new feature is close to what someone with experience in Java-to-JSON serialisation would expect. I will tidy up the script I used to generate the JSON, as requested in a review comment by @allredj , and add it to the regression test folder, to make it clearer why I chose this format (following a standard) and easier to try out the feature on other examples.

jbmc/src/java_bytecode/assignments_from_json.cpp
peterschrammel: Depends on whether it is considered a `key` in a map or a `property name`.

antlechner:
@smowton 
1) I renamed `json_clinit` to `user_specified_clinit`. Related variables are renamed as well, e.g. `use_json_clinit` -> `replace_clinit`, `is_json_clinit_needed` -> `is_user_clinit_needed`, `json_clinit_name` -> `user_specified_clinit_name`.
2) I renamed everything starting with `det_`. For structs I replaced the prefix with `object_`, and for symbol names I replaced it with `user_specified`.

jbmc/src/java_bytecode/java_static_initializers.cpp
antlechner: I renamed this to `replace_clinit`. Similarly for other variable names, see my reply to @smowton 's comment below.

jbmc/src/java_bytecode/synthetic_methods_map.h
antlechner: This is now called `USER_SPECIFIED_STATIC_INITIALIZER`, for consistency with the (new) name of the new function symbol.

jbmc/src/java_bytecode/java_static_initializers.cpp
antlechner: JSON is now used/referenced only in `assignments_from_json` and the `get_user_specified_clinit_body` function in this file. The latter can be changed to allow other formats if/when we support them.

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: I will try to have small follow-up PRs ready quickly; for everything else I'll create a ticket/issue. I currently have all plans for follow-up PRs as hand-written notes...

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 42b70d37d9374ee49854e83022b3b2c67f6735d8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112477579

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: d0e503158c00b9e6a2e3fe45c5ac3ce95ea1a810).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112480809

jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
antlechner: I added a script to generate something very close to `clinit-state.json` - minus some newlines, as the enum/String custom writers don't support pretty printing. I documented this in the README. The version of json-io is in the pom file of the newly added Maven project (for the script).

jbmc/src/java_bytecode/assignments_from_json.cpp
antlechner: > property name - the name (or key) portion of the property.

Looks like keys and property names are the same thing?

peterschrammel: Not quite. A key is itself a piece of data,  eg a parameter name that is mapped to its type, whereas a property name is a name for a piece of data, eg 'zipCode' in an address object.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 6a012bb6a3e5c85f7410accc192412c37fb9f391).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112988694

jbmc/regression/jbmc/deterministic_assignments_json/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+            </manifest>
+          </archive>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <properties>
+   <maven.compiler.source>1.8</maven.compiler.source>
+   <maven.compiler.target>1.8</maven.compiler.target>
```
allredj: :pick: weird indent

jbmc/regression/jbmc/deterministic_assignments_json/src/main/java/org/cprover/EnumWriter.java
```diff
@@ -0,0 +1,41 @@
...
+  public void write(Object obj, boolean showType, Writer output, Map<String, Object> args)
+    throws IOException {
+    JsonWriter outerWriter = getWriter(args);
+    Enum<?> enumm = (Enum<?>) obj;
+    output.write("\"name\":\"" + enumm.name() + "\"");
+    output.write(',');
+    output.write("\"ordinal\":" + enumm.ordinal());
+    for (Field field : enumm.getDeclaringClass().getDeclaredFields()) {
+      if ((field.getModifiers() & Modifier.STATIC) != Modifier.STATIC) {
+          output.write(",\"" + field.getName() + "\":");
```
allredj: :pick: indent

allredj (approved):

jeannielynnmoulton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c9546e57ffac8cd90d389f39824333191464fa06).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113095894

PR #4634, opened 2019-05-09
Add return_value_identifier function

majakusber:
This function was removed in https://github.com/diffblue/cbmc/pull/4616, I would like to bring it back - it makes `return_value_symbol` more abstract (no need to work with the `RETURN_VALUE_SUFFIX` string directly), there is at least one place where creating the whole symbol is not necessary and last but not least, the function was quite widely used in test-gen (so in order to update the cbmc pointer, this needs to be resolved).

### DIFF
--- a/src/goto-programs/remove_returns.cpp
+++ b/src/goto-programs/remove_returns.cpp
@@ -295,13 +295,9 @@ bool remove_returnst::restore_returns(
   const irep_idt function_id=f_it->first;
 
   // do we have X#return_value?
-  const namespacet ns(symbol_table);
-  auto rv_symbol = return_value_symbol(function_id, ns);
-  auto rv_name = rv_symbol.get_identifier();
-
+  auto rv_name = return_value_identifier(function_id);
   symbol_tablet::symbolst::const_iterator rv_it=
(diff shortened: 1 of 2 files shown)


danpoe (approved):

src/goto-programs/remove_returns.h
```diff
@@ -95,6 +95,10 @@ void restore_returns(symbol_table_baset &, goto_functionst &);
 
 void restore_returns(goto_modelt &);
 
+/// produces the identifier that is used to store the return
+/// value of the function with the given identifier
+irep_idt return_value_identifier(const irep_idt &);
```
romainbrenguier: it doesn't look like it needs to be exposed in the header file

src/goto-programs/remove_returns.cpp
```diff
@@ -296,8 +296,7 @@ bool remove_returnst::restore_returns(
 
   // do we have X#return_value?
   const namespacet ns(symbol_table);
```
antlechner: :bulb: Looks like we no longer need this namespace.

antlechner (approved):

smowton (approved):

PR #4637, opened 2019-05-09
Add is_return_value_* methods

smowton:
Now that RETURN_VALUE_SUFFIX is hidden, these provide a way to easily differentiate an ordinary global variable from one introduced by remove_returns.


owen-mc-diffblue:
Note #4634

peterschrammel (approved):

PR #4638, opened 2019-05-09
Question as a PR: What is string_abstractiont for?

allredj:
Just a question for now (asked in a controversial way), not to be merged as is.

The `--string-abstraction` option doesn't seem to be documented. Is it used in any way?
I'm aware that there are some regression tests, so the feature doesn't seem at all broken.

This code contains quite a few strange calls to the `index_exprt` constructor with what seems like inconsistent types. Before I dig any further, is there any known usage of that option?


martin-cs (changes requested):
My understanding is that string-abstraction is a very old (pre 2010?) feature to avoid the complexity of handling the `str*` functions.  I believe it is supposed to replace the strings with just their length, writes to them are ignored and reads are nondet'd; but it is a *long* time since I looked at the details.  I am surprised there are tests (but not surprised they are from CVS) and I would not be surprised if it turned out that it was broken in some subtle (or maybe not so subtle) way.

Given that there is quite a bit of embedded code which uses strings but doesn't reason over them and that this kind of abstraction is quite effective, I would be in favour of this feature staying.

tautschnig:
The "string abstraction" code replaces all uses of `char *` by a struct that carries that string, but also the size of the underlying allocation and the C string length. Doing these type replacements is somewhat flakey, because pointers in C may of course be subject to type casts, possibly even through integers, at which point we'd lose track. If we had support for the string solver with the C/C++ front-ends we could likely get rid of the string abstraction code, because the string solver probably does a much better job while achieving the same benefit. But until that's the case I'd rather not see this removed.

peterschrammel:
>  If we had support for the string solver with the C/C++ front-ends we could likely get rid of the string abstraction code

For C++ yes. However, for C you would have to do a similar preprocessing to bundle array and length together to a struct in order to become string-solver-ready.

tautschnig:
> For C++ yes. However, for C you would have to do a similar preprocessing to bundle array and length together to a struct in order to become string-solver-ready.

Ok, so that seems to be even more of a case to keep this code, and properly fix it if/when we get to work on integration with the string solver.

allredj:
@martin-cs @tautschnig @peterschrammel  Thanks for your input. I'll drop this PR and will create another one with some header documentation.

allredj:
Very lightweight documentation PR: https://github.com/diffblue/cbmc/pull/4837

PR #4641, opened 2019-05-09
Sharing map variable-height trees

danpoe:
This changes the sharing map such that not all leafs are stored at the same depth (which was depth 6 previously), but that the tree grows dynamically as elements are inserted.

### DIFF
--- a/src/util/sharing_map.h
+++ b/src/util/sharing_map.h
@@ -997,9 +997,9 @@ SHARING_MAPT2(, innert *)::get_container_node(const key_type &k)
 
   std::size_t key = hash()(k);
   innert *ip = &map;
-  SM_ASSERT(ip->is_internal());
+  SM_ASSERT(ip->is_defined_internal());
 
-  for(std::size_t i = 0; i < steps; i++)
+  while(true)
   {
     std::size_t bit = key & mask;
 
@@ -1024,10 +1024,9 @@ SHARING_MAPT2(const, innert *)::get_container_node(const key_type &k) const
 
   std::size_t key = hash()(k);
   const innert *ip = &map;
-
   SM_ASSERT(ip->is_defined_internal());
 
-  for(std::size_t i = 0; i < steps; i++)
+  while(true)
   {
     std::size_t bit = key & mask;
 
@@ -1058,7 +1057,7 @@ SHARING_MAPT(void)::erase(const key_type &k)
   std::size_t key = hash()(k);
   innert *ip = &map;
 
-  for(std::size_t i = 0; i < steps; i++)
+  while(true)
   {
     std::size_t bit = key & mask;
 
@@ -1072,7 +1071,9 @@ SHARING_MAPT(void)::erase(const key_type &k)
 
     ip = ip->add_child(bit);
 
-    if(ip->is_defined_container())
+    SM_ASSERT(!ip->empty());
+
+    if(ip->is_container())
       break;
 
     key >>= chunk;
@@ -1180,7 +1181,9 @@ SHARING_MAPT4(valueU, void)
   // The root cannot be a container node
   SM_ASSERT(ip->is_internal());
 
-  for(std::size_t i = 0; i < steps; i++)
+  std::size_t i = 0;
+
+  while(true)
   {
     std::size_t bit = key & mask;
 
@@ -1203,12 +1206,20 @@ SHARING_MAPT4(valueU, void)
       return;
     }
 
-    if(child->is_container() && i < steps - 1)
+    if(child->is_container())
     {
-      // Migrate the elements downwards
-      innert *cp = migrate(i, key, bit, *ip);
+      if(i < steps - 1)
+      {
+        // Migrate the elements downwards
+        innert *cp = migrate(i, key, bit, *ip);
 
-      cp->place_leaf(k, std::forward<valueU>(m));
+        cp->place_leaf(k, std::forward<valueU>(m));
+      }
+      else
+      {
+        // Add to the bottom container
+        child->place_leaf(k, std::forward<valueU>(m));
+      }
 
       num++;
 
@@ -1219,14 +1230,8 @@ SHARING_MAPT4(valueU, void)
 
     ip = child;
     key >>= chunk;
+    i++;
   }
-
-  SM_ASSERT(ip->is_defined_container());
-
-  // Add to the bottom container
-  ip->place_leaf(k, std::forward<valueU>(m));
-
-  num++;
 }
 
 SHARING_MAPT4(valueU, void)
--- a/src/util/sharing_map.h
+++ b/src/util/sharing_map.h
@@ -564,7 +564,6 @@ protected:
     std::set<const void *> &marked,
     bool mark = true) const;
 
-  static const std::string not_found_msg;
   static const std::size_t dummy_level;
 
   // config
@@ -1291,7 +1290,6 @@ SHARING_MAPT2(optionalt<std::reference_wrapper<const, mapped_type>>)::find(
 
 // static constants
 
-SHARING_MAPT(const std::string)::not_found_msg="key not found";
 SHARING_MAPT(const std::size_t)::dummy_level = 0xff;
 
 SHARING_MAPT(const std::size_t)::bits = 30;
--- a/src/util/sharing_map.h
+++ b/src/util/sharing_map.h
@@ -1284,7 +1284,7 @@ SHARING_MAPT2(optionalt<std::reference_wrapper<const, mapped_type>>)::find(
 
 SHARING_MAPT(const std::string)::not_found_msg="key not found";
 
-SHARING_MAPT(const std::size_t)::bits = 18;
+SHARING_MAPT(const std::size_t)::bits = 30;
 SHARING_MAPT(const std::size_t)::chunk = 3;
 
 SHARING_MAPT(const std::size_t)::mask = 0xffff >> (16 - chunk);
--- a/src/util/sharing_map.h
+++ b/src/util/sharing_map.h
@@ -548,6 +548,7 @@ protected:
   void add_item_if_not_shared(
     const innert &container,
     const innert &inner,
+    const std::size_t level,
     delta_viewt &delta_view,
     const bool only_common) const;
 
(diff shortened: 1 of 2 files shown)


danpoe:
On a set of methods from Apache Tika, the sum of the runtimes of the sharing map methods (timing just those) improved by 22%. Overall symex was 4% faster.

src/util/sharing_map.h
```diff
@@ -779,6 +791,73 @@ ::gather_all(const innert &n, delta_viewt &delta_view) const
   iterate(n, f);
 }
 
+SHARING_MAPT(void)::add_item_if_not_shared(
```
smowton, line 793: Document this method

src/util/sharing_map.h
```diff
@@ -779,6 +791,73 @@ ::gather_all(const innert &n, delta_viewt &delta_view) const
...
+      }
+    }
+    else
+    {
+      SM_ASSERT(ip->is_container());
+
+      if(ip->shares_with(container))
+        return;
+
+      const leaft &l1 = container.get_container().front();
```
smowton: Suggest making `l`, `ll`, `l1` more descriptive, they're pretty easy to muddle as it stands

src/util/sharing_map.h
```diff
@@ -779,6 +791,73 @@ ::gather_all(const innert &n, delta_viewt &delta_view) const
...
+
+      const leaf_listt &ll = ip->get_container();
+      SM_ASSERT(!ll.empty());
+
+      for(const auto &l : ll)
```
smowton: Comment explaining this list? It looks like you're sure `container` doesn't have multiple entries but `ip->get_container()` might do, and if it does and an entry is shared that means all further entries are too?

src/util/sharing_map.h
```diff
@@ -779,6 +791,73 @@ ::gather_all(const innert &n, delta_viewt &delta_view) const
...
+      {
+        if(l1.shares_with(l))
+          return;
+
+        if(l1.get_key() == l.get_key())
```
smowton: Why don't we do anything when differing keys occur in `container` and `ip->get_container()`? Is that impossible for some reason?

src/util/sharing_map.h
```diff
@@ -917,21 +995,28 @@ ::get_delta_view(
...
+      return ip;
 
     key >>= chunk;
   }
 
-  SM_ASSERT(ip->is_container());
-
-  return ip;
+  UNREACHABLE;
+  return nullptr;
```
smowton, line 1025: I think some compilers will complain about this unreachable statement

src/util/sharing_map.h
```diff
@@ -1001,13 +1098,138 @@ SHARING_MAPT(void)::erase(const key_type &k)
   num--;
 }
 
+SHARING_MAPT2(, innert *)::migrate(
```
smowton, line 1108: Document this methid

unit/util/sharing_map.cpp
```diff
@@ -126,6 +131,163 @@ void sharing_map_internals_test()
     sm.count_unmarked_nodes(false, marked, true);
     REQUIRE(marked.size() >= 2);
   }
+
+  SECTION("single map shape")
```
smowton, line 135: Describe this in some more detail, what is shape?

unit/util/sharing_map.cpp
```diff
@@ -214,15 +214,14 @@ void sharing_map_internals_test()
 
       REQUIRE(delta_view.size() == 2);
 
-      REQUIRE(!delta_view[0].in_both);
+      REQUIRE(!delta_view[0].is_in_both_maps());
```
smowton: Squash this commit

src/util/sharing_map.h
```diff
@@ -1283,7 +1288,7 @@ SHARING_MAPT2(optionalt<std::reference_wrapper<const, mapped_type>>)::find(
 
 SHARING_MAPT(const std::string)::not_found_msg="key not found";
 
-SHARING_MAPT(const std::size_t)::bits = 18;
+SHARING_MAPT(const std::size_t)::bits = 30;
```
smowton, line 1295: Surely belongs in its own commit

src/util/sharing_map.h
```diff
@@ -779,6 +791,73 @@ ::gather_all(const innert &n, delta_viewt &delta_view) const
...
+
+    if(ip->is_internal())
+    {
+      const to_mapt &m = ip->get_to_map();
+      SM_ASSERT(!m.empty());
+
+      for(const auto &item : m)
+      {
+        const innert *i = &item.second;
+        stack.push(i);
```
tautschnig: I don't think `i` adds any value here, I'd just do `stack.push(&item.second);`

src/util/sharing_map.h
```diff
@@ -779,6 +791,73 @@ ::gather_all(const innert &n, delta_viewt &delta_view) const
...
+
+      const leaft &l1 = container.get_container().front();
+
+      const leaf_listt &ll = ip->get_container();
+      SM_ASSERT(!ll.empty());
+
+      for(const auto &l : ll)
+      {
+        if(l1.shares_with(l))
+          return;
```
tautschnig, line 838: I would have expected `continue` - why is it ok to return?

src/util/sharing_map.h
```diff
@@ -779,6 +791,73 @@ ::gather_all(const innert &n, delta_viewt &delta_view) const
...
+
+        if(l1.get_key() == l.get_key())
+        {
+          // element is in both maps and not shared
+          delta_view.push_back(
+            {true, l1.get_key(), l1.get_value(), l.get_value()});
+          return;
```
tautschnig, line 843: Again, I would have expected `continue` instead of returning.

src/util/sharing_map.h
```diff
@@ -1001,13 +1098,138 @@ SHARING_MAPT(void)::erase(const key_type &k)
...
 }
 
+SHARING_MAPT2(, innert *)::migrate(
+  const std::size_t step,
+  const std::size_t key_suffix,
+  const std::size_t bit_last,
+  innert *ip)
+{
+  SM_ASSERT(step < steps - 1);
+  SM_ASSERT(ip != nullptr);
```
tautschnig: If it must not be a null pointer, why not pass in a reference?

src/util/sharing_map.h
```diff
@@ -1001,13 +1098,138 @@ SHARING_MAPT(void)::erase(const key_type &k)
...
+
+  // Copy the container
+  innert container_copy(child);
+
+  // Delete existing container
+  ip->remove_child(bit_last);
+
+  // Add internal node
+  ip = ip->add_child(bit_last);
+  SM_ASSERT(ip->empty());
```
tautschnig, line 1138: I would just declare a local pointer variable instead of re-using the `ip` parameter.

src/util/sharing_map.h
```diff
@@ -1001,13 +1098,138 @@ SHARING_MAPT(void)::erase(const key_type &k)
...
+
+  std::size_t i = step + 1;
+  std::size_t key = key_suffix;
+
+  key_existing >>= chunk;
+  key >>= chunk;
+
+  SM_ASSERT(i < steps);
+
+  do
```
tautschnig, line 1150: How about using a `for` loop instead: `for(std::size_t i = step + 1; i < steps; ++i)`?

src/util/sharing_map.h
```diff
@@ -1001,13 +1098,138 @@ SHARING_MAPT(void)::erase(const key_type &k)
...
+
+    key >>= chunk;
+    key_existing >>= chunk;
+
+    i++;
+  } while(i < steps);
+
+  leaft leaf_copy(as_const(&container_copy)->get_container().front());
+  leaf_listt &c = ip->get_container();
+  c.push_front(leaf_copy);
```
tautschnig: `c` doesn't seem to add much value: `ip->get_container().push_front(leaf_copy);`

src/util/sharing_map.h
```diff
@@ -922,6 +923,7 @@ ::get_delta_view(
         if(!child.shares_with(*ip2))
         {
           stack.push(stack_itemt(&child, ip2));
+          level_stack.push(0xff);
```
tautschnig: Maybe use `optionalt`, or else name that constant.

src/util/sharing_map.h
danpoe: Yes, `add_item_if_not_shared()` is called by `get_delta_view()` only for containers which only have a single element. If we know that the leaf in the first container has the same key as a leaf in the second container (either because they share or the keys compare equal), then since the maps cannot contain duplicate keys, we can ignore all further keys in the second container since they will be different from the key of the leaf in the first container. We can ignore the further leafs in the second container since `map1.get_delta_view(map2, ...)` does not include items that are only in `map2`.

src/util/sharing_map.h
danpoe: That's because if `l1` shares with `l`, then since the maps don't contain duplicate keys, all further keys in the container must be different from `l1.get_key()`. Since `map1.get_delta_view(map2, ...)` does not include items that are only in `map2`, we can return immediately.

src/util/sharing_map.h
danpoe: Similar to above, since we have found a matching key, all other keys will be different and don't have to be included in the `delta_view`.

src/util/sharing_map.h
danpoe: Ah yes there's a case missing for when we don't find a matching key. The bug would only appear when there are hash collisions. I'll fix it and add a test.

src/util/sharing_map.h
danpoe: For the first iteration we don't need to check the condition so a do-while is probably fine.

src/util/sharing_map.h
danpoe: Changed

src/util/sharing_map.h
danpoe: Done

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 38c8388639e406c7e79109f6731584872a870fec).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111936752

tautschnig (approved):

PR #4642, opened 2019-05-10
Refactoring of symex_assign_symbol

romainbrenguier:
Having to debug some changes related to `symex_assign` I found it difficult to understand what variable represents what in the execution of `symex_assign_symbol` because of several functions which have parameters that are also output. This pull request is an attempt to get rid of these.

### DIFF
--- a/src/goto-symex/goto_symex_state.cpp
+++ b/src/goto-symex/goto_symex_state.cpp
@@ -151,9 +151,9 @@ goto_symex_statet::set_indices<L2>(ssa_exprt ssa_expr, const namespacet &ns)
   return level2(level1(level0(std::move(ssa_expr), ns, source.thread_nr)));
 }
 
-void goto_symex_statet::assignment(
-  ssa_exprt &lhs, // L0/L1
-  const exprt &rhs,  // L2
+renamedt<ssa_exprt, L2> goto_symex_statet::assignment(
+  ssa_exprt lhs,    // L0/L1
+  const exprt &rhs, // L2
   const namespacet &ns,
   bool rhs_is_simplified,
   bool record_value,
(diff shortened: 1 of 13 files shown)


smowton (approved):

src/goto-symex/goto_symex_state.cpp
```diff
@@ -438,19 +441,20 @@ bool goto_symex_statet::l2_thread_read_encoding(
...
+      ssa_l2,
+      ssa_l2,
+      ssa_l2.get_original_expr(),
       tmp,
       source,
       symex_targett::assignment_typet::PHI);
 
-    expr = set_indices<L2>(std::move(ssa_l1), ns).get();
+    // TODO: are we setting l2 indices of something that is already l2?
+    expr = set_indices<L2>(std::move(ssa_l2), ns).get();
```
tautschnig: Could you try adding an assertion that this effectively returns `ssa_l2` - I agree that this seems unnecessary, but maybe we're missing something.

src/goto-symex/goto_symex_state.cpp
```diff
@@ -452,8 +452,7 @@ bool goto_symex_statet::l2_thread_read_encoding(
       source,
       symex_targett::assignment_typet::PHI);
 
-    // TODO: are we setting l2 indices of something that is already l2?
-    expr = set_indices<L2>(std::move(ssa_l2), ns).get();
+    expr = std::move(ssa_l2);
```
tautschnig, line 463: See my earlier comment: did you add an assertion to confirm this is doing what you expected across our regression tests?

src/goto-symex/goto_symex_state.cpp
romainbrenguier: I've now added an invariant before the call to check that ssa_l2 was correctly renamed.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fdc03e0ac87d45be70296b902aafc091b0e8f319).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111540186

romainbrenguier:
@tautschnig I've added the check you suggested and everything works fine

PR #4643, opened 2019-05-10
Set max-nondet-string-length to 10000 by default

majakusber:
Unless no-refine-strings is defined. If no max-nondet-string-length is set and a long string is generated, it gets discarded in the trace which causes problems in test-gen trace interpreter.

### DIFF
--- a/jbmc/regression/jbmc-strings/StringContains03/test.desc
+++ b/jbmc/regression/jbmc-strings/StringContains03/test.desc
@@ -1,11 +1,11 @@
 CORE
 Test.class
---function Test.check
+--function Test.check --max-nondet-string-length 1000000000
 ^EXIT=10$
 ^SIGNAL=0$
 ^VERIFICATION FAILED$
 file Test.java line 20 .*: FAILURE$
 --
 --
---max-nondet-string-length is not used on purpose, because this tests the behaviour
-of string refinement on very large strings.
+This tests the behaviour of string refinement on very large strings.
+max-nondet-string-length is set big enough to cause an overflow.
--- a/jbmc/regression/jbmc-strings/string-non-empty-option/test_non_empty.desc
+++ b/jbmc/regression/jbmc-strings/string-non-empty-option/test_non_empty.desc
@@ -1,6 +1,6 @@
 CORE
 Test.class
---function Test.checkMinLength --string-non-empty
+--function Test.checkMinLength --string-non-empty --max-nondet-string-length 2000000000
 ^EXIT=10$
 ^SIGNAL=0$
 assertion.* line 11 function java::Test.checkMinLength.*: SUCCESS
@@ -9,3 +9,5 @@ assertion.* line 17 function java::Test.checkMinLength.*: FAILURE
 assertion.* line 19 function java::Test.checkMinLength.*: FAILURE
 --
 ^Building error trace
+--
+max-nondet-string-length is set big enough to cause an overflow.
--- a/src/solvers/strings/string_refinement.h
+++ b/src/solvers/strings/string_refinement.h
@@ -21,6 +21,7 @@ Author: Alberto Griggio, alberto.griggio@gmail.com
 #define CPROVER_SOLVERS_REFINEMENT_STRING_REFINEMENT_H
 
 #include <limits>
+#include <util/magic.h>
 #include <util/replace_expr.h>
 #include <util/string_expr.h>
 #include <util/union_find_replace.h>
@@ -46,7 +47,7 @@ Author: Alberto Griggio, alberto.griggio@gmail.com
   "                              the switch can be used multiple times to give\n" /* NOLINT(*) */ \
   "                              several strings\n" /* NOLINT(*) */ \
   " --max-nondet-string-length n bound the length of nondet (e.g. input) strings;\n" /* NOLINT(*) */ \
-  "                              set to 10000 by default\n" /* NOLINT(*) */
+  "                              set to " + std::to_string(MAX_NONDET_STRING_LENGTH_DEFAULT) + " by default\n" /* NOLINT(*) */
 
 // The integration of the string solver into CBMC is incomplete. Therefore,
 // it is not turned on by default and not all options are available.
--- a/src/solvers/strings/string_refinement.h
+++ b/src/solvers/strings/string_refinement.h
@@ -45,7 +45,8 @@ Author: Alberto Griggio, alberto.griggio@gmail.com
   " --string-input-value st      restrict non-null strings to a fixed value st;\n" /* NOLINT(*) */ \
   "                              the switch can be used multiple times to give\n" /* NOLINT(*) */ \
   "                              several strings\n" /* NOLINT(*) */ \
-  " --max-nondet-string-length n bound the length of nondet (e.g. input) strings\n" /* NOLINT(*) */
+  " --max-nondet-string-length n bound the length of nondet (e.g. input) strings;\n" /* NOLINT(*) */ \
+  "                              set to 10000 by default\n" /* NOLINT(*) */
 
 // The integration of the string solver into CBMC is incomplete. Therefore,
 // it is not turned on by default and not all options are available.
--- a/src/util/object_factory_parameters.cpp
+++ b/src/util/object_factory_parameters.cpp
@@ -74,6 +74,10 @@ void parse_object_factory_options(const cmdlinet &cmdline, optionst &options)
       "max-nondet-string-length",
       cmdline.get_value("max-nondet-string-length"));
   }
+  else if(!cmdline.isset("no-refine-strings"))
+  {
+    options.set_option("max-nondet-string-length", 10000);
+  }
   if(cmdline.isset("string-printable"))
   {
     options.set_option("string-printable", true);
--- a/src/util/object_factory_parameters.cpp
+++ b/src/util/object_factory_parameters.cpp
@@ -8,8 +8,9 @@ Author: Diffblue Ltd
 
 #include "object_factory_parameters.h"
 
-#include <util/cmdline.h>
-#include <util/options.h>
+#include "cmdline.h"
+#include "magic.h"
+#include "options.h"
 
 void object_factory_parameterst::set(const optionst &options)
 {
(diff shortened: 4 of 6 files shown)


owen-mc-diffblue (approved):

jbmc/regression/jbmc-strings/string-non-empty-option/test_non_empty.desc
```diff
@@ -1,6 +1,6 @@
 CORE
 Test.class
---function Test.checkMinLength --string-non-empty
+--function Test.checkMinLength --string-non-empty --max-nondet-string-length 1000000000
```
romainbrenguier: This is not big enough to cause an overflow (which is what one of the assertion is checking here), the maximum java integer is 2147483647 so you should use at least half that, 2000000000 should work

martin-cs: I agree with @romainbrenguier about the overflow.  However, if this is about finding overflows / checking if they occur, please could it say that somewhere in the test or the commit message.

majakusber: Thank you, I adjusted the value

src/solvers/strings/string_refinement.h
```diff
@@ -45,7 +45,8 @@ Author: Alberto Griggio, alberto.griggio@gmail.com
   " --string-input-value st      restrict non-null strings to a fixed value st;\n" /* NOLINT(*) */ \
   "                              the switch can be used multiple times to give\n" /* NOLINT(*) */ \
   "                              several strings\n" /* NOLINT(*) */ \
-  " --max-nondet-string-length n bound the length of nondet (e.g. input) strings\n" /* NOLINT(*) */
+  " --max-nondet-string-length n bound the length of nondet (e.g. input) strings;\n" /* NOLINT(*) */ \
+  "                              set to 10000 by default\n" /* NOLINT(*) */
```
martin-cs: Could we possibly have a constant or a macro for 10000 so that we don't have a problem with the comment and the setting of the value getting out of sync?

src/util/object_factory_parameters.cpp
```diff
@@ -74,6 +74,10 @@ void parse_object_factory_options(const cmdlinet &cmdline, optionst &options)
       "max-nondet-string-length",
       cmdline.get_value("max-nondet-string-length"));
   }
+  else if(!cmdline.isset("no-refine-strings"))
+  {
+    options.set_option("max-nondet-string-length", 10000);
```
martin-cs: Why 10,000?  Does this affect performance?  If so is this a good choice?  Creating an array of 10,000 16-bit characters is a non-trivial amount for the SAT solver to handle.

martin-cs (commented):
Seems reasonable in principle but if there are overflow issues, I would rather they were actually documented / discussed.

jbmc/regression/jbmc-strings/StringContains03/test.desc
```diff
@@ -1,11 +1,11 @@
...
+--function Test.check --max-nondet-string-length 1000000000
 ^EXIT=10$
 ^SIGNAL=0$
 ^VERIFICATION FAILED$
 file Test.java line 20 .*: FAILURE$
 --
 --
---max-nondet-string-length is not used on purpose, because this tests the behaviour
-of string refinement on very large strings.
+>his tests the behaviour of string refinement on very large strings.
```
peterschrammel: typo: This

jbmc/regression/jbmc-strings/string-non-empty-option/test_non_empty.desc
romainbrenguier: The java file already say:
```
 // For this assertion to be valid, also need to limit the length of
 // strings because overflows could cause the sum to be negative.
 assert arg1.length() + arg2.length() > 1;
```
maybe the desc file should mention that we explicitly set the length limit high to be able to detect overflow behaviors.

src/util/object_factory_parameters.cpp
majakusber: As documented here: https://github.com/diffblue/test-gen/blob/develop/doc/cl-options.md 
The performance should not be affected in any significant way. The value is anyway recommended to be set, so setting the default value is only to make sure that we do not trip over it if we forget.

jbmc/regression/jbmc-strings/string-non-empty-option/test_non_empty.desc
majakusber: I added explanation to the `.desc` files about the overflow

src/util/object_factory_parameters.cpp
peterschrammel: Has the option any effect if no-refine-strings is specified?

majakusber: No, the two options cannot be specified together: https://github.com/diffblue/cbmc/blob/88f2a96f6e105803760a85d6c725ef5165cbd48e/jbmc/src/jbmc/jbmc_parse_options.cpp#L281

martin-cs: @svorenova : I don't have access to that documentation.

jbmc/regression/jbmc-strings/StringContains03/test.desc
majakusber: Corrected

src/solvers/strings/string_refinement.h
majakusber: Added a constant

majakusber:
@peterschrammel @martin-cs @romainbrenguier Ready for re-review. Please note that this is high priority as it's needed for a long-overdue test-gen bump.

src/util/object_factory_parameters.cpp
```diff
@@ -8,6 +8,7 @@ Author: Diffblue Ltd
 
 #include "object_factory_parameters.h"
 
+#include <solvers/strings/max_nondet_string_length_default.h>
```
romainbrenguier: `solvers` is not the right module for this file (`util` shouldn't depend on `solvers`).
Consider moving the definition to `util/magic.h`

src/util/object_factory_parameters.cpp
majakusber: @martin-cs Here is screenshot:
image

src/util/object_factory_parameters.cpp
majakusber: Makes sense, moved

src/util/object_factory_parameters.cpp
```diff
@@ -74,6 +74,11 @@ void parse_object_factory_options(const cmdlinet &cmdline, optionst &options)
       "max-nondet-string-length",
       cmdline.get_value("max-nondet-string-length"));
   }
+  else if(!cmdline.isset("no-refine-strings"))
```
smowton, line 79: Suggest add blank lines to make clear we've hidden an `if/else` amongst this sequence of `if`s

smowton (approved):
LGTM, one formatting suggestion

src/util/object_factory_parameters.cpp
majakusber: I get your point, although I would really like to merge this PR and the test-gen bump today (to avoid even more accumulation for the bump over the weekend) so I will not update this.

peterschrammel:
```suggestion
  else
```
I don't think the `if` part is needed.

peterschrammel (approved):

src/util/object_factory_parameters.cpp
majakusber: That would cause the max length to be bounded if `--no-refine-strings` is given which I think is not desired. And it would be bounded by a reasonably small value. If anyone would like to increase it, while keeping `--no-refine-strings`, they would get an error as one cannot define both options at the same time.

majakusber: @smowton I had to restart the CI anyway so I applied your suggestion

src/solvers/strings/string_refinement.h
martin-cs: Thank you.

src/util/object_factory_parameters.cpp
martin-cs: That sounds like an argument for it to be 100 rather than 10000.

martin-cs (approved):
The docs seem to indicate that a lower value would be better but ... that's between you and the major users of this, which is not me so approve.

src/util/object_factory_parameters.cpp
romainbrenguier: I think that's because the doc is more concerned with test-generation than verification. If you are interested in the trace generated by jbmc it's better to use a reasonable value like 100, otherwise you can set it as high as you want.

tautschnig:
I'm obviously late to the party, but I'm not particularly pleased with this fixed bound. My concern, in line with what @martin-cs and @romainbrenguier said is that there does not seem to be any rationale provided as to why it is the most appropriate number. Now we seem to be in a situation where the string length cannot be set to be unbounded anymore?

majakusber:
@tautschnig To use unbounded length, you could use either `--no-refine-strings` (which however means that no other string refinement options can be used either, namely `--string-printable` and `--string-input-value`) or you can set the bound high enough to allow overflow, see @romainbrenguier's comment here: https://github.com/diffblue/cbmc/pull/4643#discussion_r282863550

In any case, I'm happy to change the value to what makes the most sense to both CBMC and test-gen users. I do like the idea of having a default value though - I have wasted quite some time trying to debug a failing test before realizing (with the help of Romain) that the problem was the overflow.

The reason why I chose the bound 10000 rather than 100 is that I wanted to restrict the length as little as possible comparing to the unbound length (previous default). And according to the doc, this is the largest number that should still have little effect on performance.

romainbrenguier:
> In any case, I'm happy to change the value to what makes the most sense to both CBMC and test-gen users.

You could have different ones for test-gen and CBMC I guess, in particular not having one for CBMC is maybe better.

majakusber:
@romainbrenguier Good idea, I will shortly open a PR for revert and a test-gen PR introducing the bound.

PR #3546, opened 2018-12-07
Add brace initialisers to coding standard

hannes-steffenhagen-diffblue:
- [ ] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

Major benefit:
* C++ standard requires a diagnostic on narrow conversions in brace initialisers, e.g.
```
#include <iostream>

struct X {
  char x; 
  X(char x) : x(x) {}
};

struct Y {
  float x;
  Y(float x) : x(x) {}
};

int main()
{
  int val = 10;
  // may or may not warn (does not by default in g++)
  auto const a = X(val);
  // diagnostic required by C++ standard!
  auto const b = X{val};

  double dval = 3.0;
  // again, no warning with g++
  auto const c = Y(dval);
  // diagnostic required
  auto const d = Y{dval};
}
```

Minor benefits:
* Can visually distinguish constructor from function calls
* Can use slightly more consistent syntax `type_name variable_name{};` instead of `type_name variable_name;` (parentheses version doesn't work because it's indistinguishable from a function declaration)

PR #4645, opened 2019-05-10
Add symtab2gb to enable linking of .json_symtab files

hannes-steffenhagen-diffblue:
Right now we have the `goto-cc` tool to create goto binaries from C
files and link goto binaries together. This adds a similar type of
"linker" tool for the symtab language.

I had considered extending 'goto-cc' to handle symtab files as well,
however goto-cc (quite reasonably) makes some C-specific assumptions
about its input files, and I'd figured rather than working around that
it'd be easier to just have a simple command line tool to invoke
`goto_convert` and `link_goto_model` (which, in the end, is all we want
to do for json_symtabs).


src/Makefile
```diff
@@ -21,6 +21,7 @@ DIRS = analyses \
        solvers \
        util \
        xmllang \
+			 symtab2gb \
```
xbauch: funny whitespaces.

hannes-steffenhagen-diffblue:
There's no real need to look at the `.json_symtab` files in the tests, they're just big generated files containing symbol tables.

xbauch (approved):
Looks good.

martin-cs:
@hannes-steffenhagen-diffblue : can you elaborate a bit on why this can't be part of goto-cc?  It seems that a general for / against is something like:

- Where possible we want to stop the proliferation of different tools and we already have a tool which is front-end producing goto-binaries.
+ We may want to be able to do Ada specific linking things, particularly around the questions of elaboration order and set-up.
+ Other languages that use .json_symtab may also need to do custom linking things.

I would particularly like to hear from @tautschnig as the author / warden of the linker.

hannes-steffenhagen-diffblue:
@martin-cs regarding Ada, I think we can get away without any special elaboration support during linking, but that's not what I mean; The problem with `goto-cc` is that it assumes `main` is the entry point to the program (which isn't true in Ada, `main` is just a normal identifier there), and it also replaces any `__CPROVER_start` which is already present with the one from `ansi_c_entry_point`.

I thought it'd be useful to have something that just does a straight forward combining of symbol tables, and I'm not sure if that's really compatible with `goto-cc` which is supposed to act like a C compiler.

martin-cs:
@hannes-steffenhagen-diffblue I am not sure if it is compatible either but I think it is a discussion worth having.  If I understand the points you raise correctly there are:

A. where should we do language-specific linking things?
A.1. how and by whom is the entry point specified?
A.2. what do we do about set-up code and how does that get through linking?

B. where should linking be a stand-alone tool? In goto-cc? In goto-instrument?

I think in the next few weeks and months there are likely to be Ada-specific linking things to be done (we may need to support control over elaboration orders for example), so I am not exactly clear on what the best route forward is.

I would very much appreciate thoughts from others (including, ofcourse, @hannes-steffenhagen-diffblue )

regression/symtab2gb/multiple_symtabs/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+entry_point.json_symtab
+user.json_symtab library.json_symtab
```
tautschnig, line 3: Could you please add a readme file or some other explanation describing how the symtab files were generated?

src/CMakeLists.txt
```diff
@@ -104,3 +104,5 @@ add_subdirectory(goto-instrument)
 add_subdirectory(goto-analyzer)
 add_subdirectory(goto-diff)
 add_subdirectory(goto-harness)
+
```
tautschnig: Unnecessary blank line?

src/Makefile
```diff
@@ -21,6 +21,7 @@ DIRS = analyses \
        solvers \
        util \
        xmllang \
+       symtab2gb \
```
tautschnig: Sort lexicographically.

src/symtab2gb/symtab2gb_main.cpp
```diff
@@ -0,0 +1,113 @@
+#include <fstream>
```
tautschnig: Missing copyright header

src/symtab2gb/symtab2gb_main.cpp
```diff
@@ -0,0 +1,113 @@
...
+      throw invalid_source_file_exceptiont{
+        "failed to parse symbol table from file `" + symtab_filename + "'"};
+    }
+    auto symtab = symbol_tablet{};
+    if(failed(symtab_language->typecheck(symtab, "<unused>")))
+    {
+      throw invalid_source_file_exceptiont{
+        "failed to typecheck symbol table from file `" + symtab_filename + "'"};
+    }
+    auto goto_model = goto_modelt{};
```
tautschnig: I'd prefer `goto_modelt goto_model;`

src/symtab2gb/symtab2gb_main.cpp
```diff
@@ -0,0 +1,113 @@
...
+  for(std::size_t ix = 0; ix < symtab_files.size(); ++ix)
+  {
+    auto const &symtab_filename = symtab_filenames[ix];
+    auto &symtab_file = symtab_files[ix];
+    if(failed(symtab_language->parse(symtab_file, symtab_filename)))
+    {
+      throw invalid_source_file_exceptiont{
+        "failed to parse symbol table from file `" + symtab_filename + "'"};
+    }
+    auto symtab = symbol_tablet{};
```
tautschnig: `symbol_tablet symtab;`

src/symtab2gb/symtab2gb_main.cpp
```diff
@@ -0,0 +1,113 @@
...
+    std::vector<std::string> symtab_filenames;
+    optionalt<std::string> gb_file;
+    for(int i = 1; i < argc; ++i)
+    {
+      if(argv[i] == std::string("--out"))
+      {
+        if(i + 1 >= argc)
+        {
+          throw invalid_command_line_argument_exceptiont{
+            "expected output file name", "--out"};
```
tautschnig: To be honest, this is a bit too simplistic to my taste: I think this should have support for `--help` and `--version` at bare minimum. Therefore I'd recommend inheriting from `parse_options_baset`.

tautschnig (commented):
I'd lean towards keeping such a simple linking tool as a separate binary, because `goto-cc` primarily deals with the various command-line options of existing compilers, none of which is of interest here. But then I'm wondering whether the binary could just go in the existing `json-symtab-language` folder?

src/symtab2gb/symtab2gb_main.cpp
hannes-steffenhagen-diffblue: I prefer the `auto` form because with that I can make all declarations look the same, but I don't mind changing it if it bothers you.

hannes-steffenhagen-diffblue:
@tautschnig I believe I've addressed your review comments

src/symtab2gb/symtab2gb_main.cpp
```diff
@@ -0,0 +1,18 @@
...
+\******************************************************************/
+
+/// \file
+/// symtab2gb Main Module
+
+#include "symtab2gb_parse_options.h"
+
+int main(int argc, const char *argv[])
+{
+  symtab2gb_parse_optionst parse_options{argc, argv};
```
martin-cs, line 16: Is there a reason for `{` `}` rather than `(` `)` ?

src/symtab2gb/symtab2gb_parse_options.cpp
```diff
@@ -0,0 +1,138 @@
...
+static void run_symtab2gb(
+  const std::vector<std::string> &symtab_filenames,
+  const std::string &gb_filename)
+{
+  // try opening all the files first to make sure we can
+  // even read/write what we want
+  std::ofstream out_file{gb_filename};
+  if(!out_file.is_open())
+  {
+    throw system_exceptiont{"couldn't open output file `" + gb_filename + "'"};
```
martin-cs, line 48: Ditto

src/symtab2gb/symtab2gb_parse_options.cpp
```diff
@@ -0,0 +1,138 @@
...
+  }
+  std::vector<std::ifstream> symtab_files;
+  for(auto const &symtab_filename : symtab_filenames)
+  {
+    std::ifstream symtab_file{symtab_filename};
+    if(!symtab_file.is_open())
+    {
+      throw system_exceptiont{"couldn't open input file `" + symtab_filename +
```
martin-cs, line 56: Ditto.

martin-cs (approved):
After a discussion with @chrisr-diffblue I think this is probably the best architecture for now.

src/symtab2gb/symtab2gb_main.cpp
hannes-steffenhagen-diffblue: Brace initialisation avoids parsing ambiguities and prevents implicit conversions, and is recommended "by default" by the [CPP Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es23-prefer-the--initializer-syntax). Unless we have a good reason my preference would be to stick to those as the 'default'.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: dc8b83cd8ca90fc144d116d72e67dcf847b5f37c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111644348

hannes-steffenhagen-diffblue:
Windows tests seem to fail the goto binaries that were written aren't readable? Honestly no idea what's going on there

src/symtab2gb/symtab2gb_main.cpp
tautschnig: @martin-cs Preferred since #3546

tautschnig:
> Windows tests seem to fail the goto binaries that were written aren't readable? Honestly no idea what's going on there

I looked at the code again and nothing obvious springs to my mind. I can try to play with this in a Windows environment later this week.

src/symtab2gb/symtab2gb_main.cpp
tautschnig: It does bother _me_, but I'm just one voice here. If others like this better then there's no need to fight about it. (I actually don't think that all declarations looking the same is desirable - quite often I don't even care much about the name, just about the type.)

regression/symtab2gb/Makefile
```diff
@@ -0,0 +1,27 @@
...
+test:
+	@../test.pl -e -p -c '../chain.sh $(exe) ../../../src/symtab2gb/symtab2gb ../../../src/cbmc/cbmc'
+
+tests.log:
+	@../test.pl -e -p -c '../chain.sh $(exe) ../../../src/symtab2gb/symtab2gb ../../../src/cbmc/cbmc'
+
+show:
+	@for dir in *; do \
+		if [ -d "$$dir" ]; then \
+			vim -o "$$dir/*.c" "$$dir/*.out"; \
```
chrisr-diffblue, line 15: $VISUAL or $EDITOR, rather than "vim" ?

src/symtab2gb/symtab2gb_parse_options.cpp
```diff
@@ -0,0 +1,138 @@
...
+  return error_indicator;
+}
+
+static void run_symtab2gb(
+  const std::vector<std::string> &symtab_filenames,
+  const std::string &gb_filename)
+{
+  // try opening all the files first to make sure we can
+  // even read/write what we want
+  std::ofstream out_file{gb_filename};
```
chrisr-diffblue: Regarding your windows CI issue, does windows still care about the difference between binary and ascii file modes, and if so, what does std::ofstream do by default? I'm wondering if you need to put the stream into a binary mode?

chrisr-diffblue: src/goto-cc/compiler.cpp does this: `std::ofstream outfile(file_name, std::ios::binary);` so I suspect you might need the binary mode.

chrisr-diffblue: Also, do you need to explicitly close() the output file as well?

regression/symtab2gb/Makefile
tautschnig: The `-o` option is specific to `vim`, and it's the same bit of code we have everywhere else. But I don't know whether anyone ever users this target?!

chrisr-diffblue: Ah, this was the first time I've actually noticed this target - never used it myself!

hannes-steffenhagen-diffblue: Ditto on never using it. It's honestly a bit of cargo culting, I was kind of assuming _someone_ around here liked to work that way. But I certainly don't mind removing it, I have no attachment to it.

src/symtab2gb/symtab2gb_parse_options.cpp
hannes-steffenhagen-diffblue: `binary` yes, I had not considered that at all but now that you mention it that does seem to be the kind of thing that'd cause this problem.

`close` I don't think so, the `std` streams are all RAII controlled, so they should close even on exception.

I can put the input files in a smaller scope so they get closed before the actual writing starts, although I'm not sure how much of a benefit that's going to be (I'm opening all files at the start to be as-sure-as-reasonably-possible-in-advance that our IO operations are going to succeed, so we don't go half way through the linking bit only to find one of the files is misspelled or we don't have r/w access or something like that).

tautschnig: Could you actually add a precondition to `write_goto_binary` that tests that the stream is open in binary mode?

regression/symtab2gb/Makefile
tautschnig: I'd vote for having a separate PR that removes all of those. If someone is still using them, we can then have that discussion on the PR.

src/symtab2gb/symtab2gb_parse_options.cpp
hannes-steffenhagen-diffblue: @tautschnig AFAIK iostreams don't expose a way to query or change the openmode

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 10d7a51c62e13092474e4dcb8fbd554070ae216f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111720442

chrisr-diffblue (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: da5fd4508c48c1636e4df4fb7e5c4004faf5a610).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/111772089

PR #4646, opened 2019-05-10
Default fall-through for function pointer removal

xbauch:
When the pointer points to neither of the candidates we shouldn't just call the
first one (which does happen if the user does not request pointer-check). This
PR inserts an ASSUME (false) which makes all subsequent assertions
unreachable (and thus reported as not being violated).

A small regression test is included.

### DIFF
--- a/src/goto-programs/remove_function_pointers.cpp
+++ b/src/goto-programs/remove_function_pointers.cpp
@@ -416,6 +416,7 @@ void remove_function_pointerst::remove_function_pointer(
     t->source_location.set_property_class("pointer dereference");
     t->source_location.set_comment("invalid function pointer");
   }
+  new_code_gotos.add(goto_programt::make_assumption(false_exprt()));
 
   goto_programt new_code;
 
--- a/regression/cbmc/Function_Pointer18/main.c
+++ b/regression/cbmc/Function_Pointer18/main.c
@@ -1,0 +1,34 @@
+// f is set to 0 -> does not point to either f1 or f2
+#include <assert.h>
+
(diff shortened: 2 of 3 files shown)


smowton (approved):

src/goto-programs/remove_function_pointers.cpp
```diff
@@ -408,14 +408,18 @@ void remove_function_pointerst::remove_function_pointer(
       goto_programt::make_goto(call, equal_exprt(pointer, casted_address)));
   }
 
+  goto_programt::targett t;
   // fall-through
   if(add_safety_assertion)
   {
-    goto_programt::targett t =
-      new_code_gotos.add(goto_programt::make_assertion(false_exprt()));
+    t = new_code_gotos.add(goto_programt::make_assertion(false_exprt()));
```
hannes-steffenhagen-diffblue: FWIW there should probably be an `assume(false)` after this as well.

hannes-steffenhagen-diffblue (approved):

src/goto-programs/remove_function_pointers.cpp
tautschnig: I do agree with this as it  would be consistent with what happens with loop-unwinding assertions (where we add the negated loop condition to the path condition). This would effectively simplify the change here as you'd just unconditionally add the assumption below.

tautschnig (approved):
This is a very good catch, but I'd agree with what @hannes-steffenhagen-diffblue said, so please fix this before merging the change.

martin-cs (approved):
No changes needed other than the ones already requested.

PR #4647, opened 2019-05-10
Suppress color output on aws

hannes-steffenhagen-diffblue:
Color codes don't show up correctly on AWS CodeBuild console, so we're disabling them to make the output more legible (they're still on by default!). Having the ability to turn color codes off can be useful in other situations (like scripts) as well.

### DIFF
--- a/buildspec-linux-clang.yml
+++ b/buildspec-linux-clang.yml
@@ -1,5 +1,10 @@
 version: 0.2
 
+env:
+  variables:
+    # CodeBuild console doesn't display color codes correctly
+    TESTPL_COLOR_OUTPUT: 0
+
 phases:
   install:
     runtime-versions:
--- a/regression/test.pl
+++ b/regression/test.pl
@@ -21,6 +21,16 @@ my $has_thread_pool = eval
 #
 # runs a test and check its output
 
+my $color_output_enabled;
+sub with_color {
+  my ($text, $color) = @_;
+  if ($color_output_enabled) {
+    colored($text, $color);
+  } else {
+    $text;
+  }
+}
+
 sub run($$$$$) {
   my ($name, $input, $cmd, $options, $output) = @_;
   my $cmdline = "$cmd $options '$input' >'$output' 2>&1";
@@ -290,6 +300,10 @@ Usage: test.pl -c CMD [OPTIONS] [DIRECTORIES ...]
              independent logs.
   -f         forward the test name to CMD
 
+  --[no]color enable/disable color output; enabled by default unless
+              TESTPL_COLOR_OUTPUT is set to 0, in which case it is
+              disabled by default.
+
 test.pl expects a test.desc file in each subdirectory. The file test.desc
 follows the format specified below. Any line starting with // will be ignored.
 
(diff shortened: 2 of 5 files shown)


buildspec-linux-clang.yml
```diff
@@ -1,5 +1,10 @@
 version: 0.2
 
+env:
+  variables:
+    # CodeBuild console doesn't display color codes correctly
+    TESTPL_SUPPRESS_COLOR_OUTPUT: 1
```
smowton: I'd recommend `TESTPL_COLOR_OUTPUT` rather than `SUPPRESS_` with an implicit negation, otherwise lgtm

smowton (approved):

buildspec-linux-clang.yml
chrisr-diffblue: +1

regression/test.pl
```diff
@@ -289,6 +301,8 @@ ($$$$)
              as runs with different suffixes will operate independently and keep
              independent logs.
   -f         forward the test name to CMD
+  -S <0|1>   (un-)suppress color output, default 0 if TESTPL_SUPPRESS_COLOR_OUTPUT
```
chrisr-diffblue: Rather than -S 0/1 to turn this on/off, could you have say "-P" for "plain output" - and just default to colour unless "-P" or TESTPL_COLOR_OUTPUT=0 ?

hannes-steffenhagen-diffblue: @chrisr-diffblue I wanted a way to force it back on through command line arguments if `TESTPL_SUPPRESS_COLOR_OUTPUT=1`

hannes-steffenhagen-diffblue:
@chrisr-diffblue @smowton I changed the option name to `--[no]color` and the env var name to `TESTPL_COLOR_OUTPUT`, and flipped the logic to match the new names

regression/test.pl
```diff
@@ -290,6 +300,11 @@ ($$$$)
              independent logs.
   -f         forward the test name to CMD
 
+  --[no]color enable/disable color output; enabled by default unless
+              TESTPL_COLOR_OUTPUT is set, in which case it is enabled by
+              default if its value is nonempty and nonzero, and enabled
+              by default otherwise.
+
```
chrisr-diffblue, line 306: I may just be being particularly slow today, but I find that message a little difficult to parse :-) Maybe a clearer description might be something like:

enabled by default, unless the TESTPL_COLOR_OUTPUT environment variable is set to a nonempty, nonzero value. This option overrides any value of TESTPL_COLOR_OUTPUT

what do you think?

regression/test.pl
```diff
@@ -324,9 +339,13 @@ ($$$$)
...
 $Getopt::Std::STANDARD_HELP_VERSION = 1;
-our ($opt_c, $opt_e, $opt_f, $opt_i, $opt_j, $opt_n, $opt_p, $opt_h, $opt_C, $opt_T, $opt_F, $opt_K, $opt_s, %defines, @include_tags, @exclude_tags); # the variables for getopt
-GetOptions("D=s" => \%defines, "X=s" => \@exclude_tags, "I=s" => \@include_tags);
-getopts('c:efi:j:nphCTFKs:') or &main::HELP_MESSAGE(\*STDOUT, "", $main::VERSION, "");
+our ($opt_c, $opt_e, $opt_f, $opt_i, $opt_j, $opt_n, $opt_p, $opt_h, $opt_C, $opt_T, $opt_F, $opt_K, $opt_s, $opt_S, %defines, @include_tags, @exclude_tags); # the variables for getopt
+$color_output_enabled = 1;
+if (exists $ENV{'TESTPL_COLOR_OUTPUT'}) {
+  $color_output_enabled = $ENV{'TESTPL_COLOR_OUTPUT'};
+}
+GetOptions("D=s" => \%defines, "X=s" => \@exclude_tags, "I=s" => \@include_tags, 'color!' => \$color_output_enabled);
```
chrisr-diffblue, line 350: Is it worth adding a comment to explain that the test/setting of value from the ENV *must* come before the call to GetOptions to preserve the behaviour that the command line option overrides the ENV variable?

regression/test.pl
hannes-steffenhagen-diffblue: I think it might be hard to parse because I mistyped the latter half of the sentence, it was meant to be

> and disabled by default otherwise.

instead of

> and enabled by default otherwise.

regression/test.pl
hannes-steffenhagen-diffblue: done

regression/test.pl
hannes-steffenhagen-diffblue: I've simplified the text, should be easier to parse now

chrisr-diffblue (approved):

PR #4536, opened 2019-04-15
Interface for collecting function pointer targets

xbauch:
Motivation: we want `remove_returns` to work even if not preceded by `remove_function_pointers`. For example:

```
int (* const fptbl2[])(void) = {g1, g2};

int func(int id1)
{
  t = fptbl2[id1];  // for this instruction the set is {"g1", "g2"}
  return t;
}
```

This PR exposes the function-pointer-target-collecting part of `remove_function_pointers` to be used in other analyses and the does so in `remove_returns`.

In `remove_returns` use the `possible_fp_targets_map` to be able to remove returns for calls to function pointers. `CALL x=*fp()` should become `CALL *fp(); ASSIGN x = (fp == &f00) ? return_value_f00 : (fp == &f01) ?` ....

The first 3 commits are refactoring/preparation. The fourth introduces the interface. The fifth implements the new `do_function_call` for function pointers.

Then we refactor the `remove_function_pointers` to extract the collect-candidate-functions part and separate the modify-function-call part.

Finally -- now that the file is open -- we also refactor the internal `remove_function_pointers` without any functional change.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4650, opened 2019-05-13
Separate collecting function pointer targets [alternative-to: #4536]

xbauch:
The goal is similar to that of #4536 but here we physically separate the collection from remove-function-pointers into a stand-alone module.

Similarly to #4536: remove returns analysis is no longer depended on remove-function-pointers but the later can produce more efficient GOTO, e.g. in case the remove-const-function-pointer analysis was precise (result was exactly one target).

### DIFF
--- a/jbmc/src/janalyzer/janalyzer_parse_options.cpp
+++ b/jbmc/src/janalyzer/janalyzer_parse_options.cpp
@@ -634,7 +634,7 @@ bool janalyzer_parse_optionst::process_goto_program(const optionst &options)
     goto_partial_inline(goto_model, ui_message_handler);
 
     // remove returns, gcc vectors, complex
-    remove_returns(goto_model);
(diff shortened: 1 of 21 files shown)


tautschnig:
Is there any chance to get this in smaller doses? The overarching idea of this PR seems great to me, but 35 commits and/or +1,379 −398 changed lines is quite scary I have to admit. Or maybe an explanation as to why it has to be this large? And then just the initial look at the overall diff made me wonder why adding message handlers was a necessary part of this change. I'll certainly bite the bullet if it needs to be this big, but if there is a good way to break it up then I'd appreciate this.

xbauch:
@tautschnig Splitting in would not be a problem but I think @kroening should first decide if the end-result is  something we should aim for in the first place.

TGWDB:
Closing due to age and draft status. If you believe this is erroneous please reopen and update the PR.

PR #4660, opened 2019-05-16
Revert "Set max-nondet-string-length to 10000 by default"

majakusber:
Reverts diffblue/cbmc#4643 which introduced a default value for `max-nondet-string-length`. Instead, the value will remain unbounded in CBMC by default and a default value will be defined in test-gen only.

### DIFF
--- a/jbmc/regression/jbmc-strings/StringContains03/test.desc
+++ b/jbmc/regression/jbmc-strings/StringContains03/test.desc
@@ -1,11 +1,11 @@
 CORE
 Test.class
---function Test.check --max-nondet-string-length 1000000000
+--function Test.check
 ^EXIT=10$
 ^SIGNAL=0$
 ^VERIFICATION FAILED$
 file Test.java line 20 .*: FAILURE$
 --
 --
-This tests the behaviour of string refinement on very large strings.
-max-nondet-string-length is set big enough to cause an overflow.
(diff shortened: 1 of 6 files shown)


peterschrammel (approved):

tautschnig (approved):

tautschnig:
+100 for distinguishing a verification tool (CBMC) from a test-input generator. Thank you @svorenova!

majakusber:
@tautschnig Thank you for pointing it out :+1: 
Test-gen PR is ready here: https://github.com/diffblue/test-gen/pull/3425

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c35c16a70dbbf2cf5f08fd134426c1a88d8ba17d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112055707

martin-cs:
I realise this patch caused a disproportionate amount of discussion on it's way in and it is not my intent to replicate that but, @svorenova -- why did you decide to revert this?

majakusber:
Because from the discussion it seemed like this is not desired for CBMC (or if it is, what the default should be), but it would be useful for test-gen. So basically to not limit CBMC users unnecessarily.

martin-cs:
Fair.  Thanks.

PR #4674, opened 2019-05-20
Refactor clean_expr so that it returns an exprt and takes argument by copy

romainbrenguier:
This makes the information flow clearer, avoid errors where a subtype of
exprt could be passed as argument and its id mutated to something
inconsistent with this subtype, and avoids having to make explicit
copies in the places where the function is called.

### DIFF
--- a/src/goto-symex/symex_clean_expr.cpp
+++ b/src/goto-symex/symex_clean_expr.cpp
@@ -173,8 +173,7 @@ replace_nondet(exprt &expr, symex_nondet_generatort &build_symex_nondet)
 
 void goto_symext::lift_let(statet &state, const let_exprt &let_expr)
 {
-  exprt let_value = let_expr.value();
-  clean_expr(let_value, state, false);
+  exprt let_value = clean_expr(let_expr.value(), state, false);
   let_value = state.rename(std::move(let_value), ns).get();
   do_simplify(let_value);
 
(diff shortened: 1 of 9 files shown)


src/goto-symex/symex_clean_expr.cpp
```diff
@@ -213,8 +212,8 @@ void goto_symext::lift_lets(statet &state, exprt &rhs)
   }
 }
 
-void goto_symext::clean_expr(
-  exprt &expr,
+exprt goto_symext::clean_expr(
```
tautschnig: Could you please introduce a macro wrapping `__attribute__((warn_unused_result))` - I thought that @smowton had already done this, but now I can't find it. Otherwise changes of this kind result in bugs that are terrible to debug.

tautschnig (changes requested):
Looks good except we need to make sure we warn about unused return values.

src/goto-symex/symex_clean_expr.cpp
smowton: It was on an abandoned branch of the let-expr work, but it's just a port of test-gen's `NODISCARD` macro so should be easily recreated

smowton (approved):

tautschnig (approved):

tautschnig:
@romainbrenguier clang-format and cpp-lint aren't quite happy yet.

PR #4675, opened 2019-05-20
Document byte_update_exprt and rename non-const functions

romainbrenguier:
In some cases the non-const version was unnecessarily called because it
has the same name as the const version.
This pull requests avoids that from happening in 9 cases for op, 2 for value and 3 for
offset.

### DIFF
--- a/src/util/simplify_expr.cpp
+++ b/src/util/simplify_expr.cpp
@@ -2064,7 +2064,7 @@ bool simplify_exprt::simplify_byte_update(byte_update_exprt &expr)
     expr.offset() == to_byte_update_expr(expr.op()).offset() &&
     expr.value().type() == to_byte_update_expr(expr.op()).value().type())
   {
-    expr.op()=expr.op().op0();
+    expr.set_op()=expr.op().op0();
     return false;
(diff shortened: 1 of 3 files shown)


tautschnig (changes requested):
Renaming functions without following an appropriate deprecation path is a no-go.

romainbrenguier:
@tautschnig 
> Renaming functions without following an appropriate deprecation path is a no-go.

I've now kept both versions and deprecated the old one.

src/util/simplify_expr.cpp
```diff
@@ -2057,20 +2057,22 @@ bool simplify_exprt::simplify_byte_extract(byte_extract_exprt &expr)
 
 bool simplify_exprt::simplify_byte_update(byte_update_exprt &expr)
 {
+  const byte_update_exprt &expr_const = as_const(expr);
```
tautschnig: Is this use of `as_const` actually necessary?

tautschnig (approved):
Looks good to me - except I'd prefer if the `set_*` methods actually took an argument as discussed.

PR #4676, opened 2019-05-20
Add --nondet-static-except flag to CBMC

mmuesly:
--nondet-static-except behaves as --nondet-static but allows to filter
the symbols which are effected by --nondet-static. Every symbol
that is effected by nondet-static and is not in the comma separated
symbol list passed to --nondet-static-except will be nondet when
using this flag. All symbols in the list keep their concret initalization.

### DIFF
--- a/regression/goto-instrument/chain.sh
+++ b/regression/goto-instrument/chain.sh
@@ -8,13 +8,24 @@ cbmc=$3
 is_windows=$4
 
 name=${*:$#}
-name=${name%.c}
+
+if [[ x$name == x ]];  then
+  name=${name%.c}
+fi
 
 args=${*:5:$#-5}
 
-if [[ "${is_windows}" == "true" ]]; then
+if [[ "${is_windows}" == "true" && x$name != x ]]; then
+  $goto_cc "main.gb" ${name}
+  name="main"
+  mv "${name}.exe" "${name}.gb"
+elif [[ "${is_windows}" == "true" ]]; then
   $goto_cc "${name}.c"
   mv "${name}.exe" "${name}.gb"
+elif [[ x$name != x ]]; then
+  $goto_cc -o "main.gb" ${name}
+  echo "name: ${name}"
+  name="main"
 else
   $goto_cc -o "${name}.gb" "${name}.c"
 fi
--- a/regression/goto-instrument/nondet-static-exclude_subfolder/package1/file.c
+++ b/regression/goto-instrument/nondet-static-exclude_subfolder/package1/file.c
@@ -1,0 +1,8 @@
+static int myVal = 3;
+static int myVal2 = 4;
+
+void method2()
+{
+  __CPROVER_assert(myVal == 3, "method2 myVal");
+  __CPROVER_assert(myVal2 == 4, "method2 myVal2");
+}
--- a/regression/goto-instrument/nondet-static-exclude_subfolder/package2/file.c
+++ b/regression/goto-instrument/nondet-static-exclude_subfolder/package2/file.c
@@ -1,0 +1,8 @@
+static int myVal = 3;
+static int myVal2 = 4;
+
+void method1()
+{
+  __CPROVER_assert(myVal == 3, "method1 myVal");
+  __CPROVER_assert(myVal2 == 4, "method1 myVal2");
+}
--- a/regression/goto-instrument/nondet_static_exclude/test.desc
+++ b/regression/goto-instrument/nondet_static_exclude/test.desc
@@ -1,0 +1,18 @@
+CORE
+main_1.c main.c 
+--nondet-static --nondet-static-exclude test1 --nondet-static-exclude test3 --nondet-static-exclude test4 --nondet-static-exclude main_1.c:value
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+assertion test1 == (NULL|0): SUCCESS
+assertion test2.a == 5: FAILURE
+assertion test3.a == 1: SUCCESS
+assertion test3.b == 4: SUCCESS
+assertion test4->a == 1: SUCCESS
+assertion test4->b == 4: SUCCESS
+assertion test5->a == 5: FAILURE
+assertion test5->b == 2: FAILURE
+assertion value == 7: FAILURE
+assertion value == 5: SUCCESS
+--
+--
--- a/src/goto-instrument/goto_instrument_parse_options.cpp
+++ b/src/goto-instrument/goto_instrument_parse_options.cpp
@@ -1205,10 +1205,19 @@ void goto_instrument_parse_optionst::instrument_goto_program()
 
   // ignore default/user-specified initialization of variables with static
   // lifetime
-  if(cmdline.isset("nondet-static"))
+  if(cmdline.isset("nondet-static-exclude"))
   {
-    log.status() << "Adding nondeterministic initialization of static/global "
-                    "variables"
+    log.status() << "Adding nondeterministic initialization "
+                    "of static/global variables except for "
+                    "the specified ones."
+                 << messaget::eom;
+
+    nondet_static(goto_model, cmdline.get_values("nondet-static-exclude"));
+  }
+  else if(cmdline.isset("nondet-static"))
+  {
+    log.status() << "Adding nondeterministic initialization "
+                    "of static/global variables"
                  << messaget::eom;
     nondet_static(goto_model);
   }
@@ -1607,6 +1616,8 @@ void goto_instrument_parse_optionst::help()
     " --isr <function>             instruments an interrupt service routine\n"
     " --mmio                       instruments memory-mapped I/O\n"
     " --nondet-static              add nondeterministic initialization of variables with static lifetime\n" // NOLINT(*)
+    " --nondet-static-exclude e    same as nondet-static except for the variable e\n" //NOLINT(*)
+    "                              (use multiple times if required)\n"
     " --check-invariant function   instruments invariant checking function\n"
     " --remove-pointers            converts pointer arithmetic to base+offset expressions\n" // NOLINT(*)
     " --splice-call caller,callee  prepends a call to callee in the body of caller\n"  // NOLINT(*)
--- a/src/goto-instrument/nondet_static.cpp
+++ b/src/goto-instrument/nondet_static.cpp
@@ -46,6 +46,12 @@ bool is_nondet_initializable_static(
 
   const symbolt &symbol = ns.lookup(id);
 
+  // is the symbol explicitly marked as not to be nondet initialized?
+  if(symbol.value.get_bool(ID_C_no_nondet_initialization))
+  {
+    return false;
+  }
+
   // is the type explicitly marked as not to be nondet initialized?
   if(symbol.type.get_bool(ID_C_no_nondet_initialization))
     return false;
@@ -110,24 +116,22 @@ void nondet_static(
         nondet_static(ns, goto_functions, fsym.get_identifier());
     }
   }
+
+  // update counters etc.
+  goto_functions.update();
 }
 
 /// Nondeterministically initializes global scope variables in
 /// CPROVER_initialize function.
 /// \param ns: Namespace for resolving type information.
 /// \param [out] goto_functions: Existing goto-functions to be updated.
-void nondet_static(
-  const namespacet &ns,
-  goto_functionst &goto_functions)
+void nondet_static(const namespacet &ns, goto_functionst &goto_functions)
 {
   nondet_static(ns, goto_functions, INITIALIZE_FUNCTION);
-
-  // update counters etc.
-  goto_functions.update();
 }
 
-/// Main entry point of the module. Nondeterministically initializes global
-/// scope variables, except for constants (such as string literals, final
+/// First main entry point of the module. Nondeterministically initializes
+/// global scope variables, except for constants (such as string literals, final
 /// fields) and internal variables (such as CPROVER and symex variables,
 /// language specific internal variables).
 /// \param [out] goto_model: Existing goto-model to be updated.
@@ -135,4 +139,60 @@ void nondet_static(goto_modelt &goto_model)
 {
   const namespacet ns(goto_model.symbol_table);
   nondet_static(ns, goto_model.goto_functions);
+}
+
+/// Second main entry point of the module. Nondeterministically initializes
+/// global scope variables, except for constants (such as string literals, final
+/// fields), internal variables (such as CPROVER and symex variables,
+/// language specific internal variables) and variables passed to except_value.
+/// \param [out] goto_model: Existing goto-model to be updated.
+/// \param except_values: list of symbol names that should not be updated.
+void nondet_static(
+  goto_modelt &goto_model,
+  const optionst::value_listt &except_values)
+{
+  const namespacet ns(goto_model.symbol_table);
+  std::set<std::string> to_exclude;
+
+  for(auto const &except : except_values)
+  {
+    const bool file_prefix_found = except.find(":") != std::string::npos;
+
+    if(file_prefix_found)
+    {
+      to_exclude.insert(except);
+      if(has_prefix(except, "./"))
+      {
+        to_exclude.insert(except.substr(2, except.length() - 2));
+      }
+      else
+      {
+        to_exclude.insert("./" + except);
+      }
+    }
+    else
+    {
+      irep_idt symbol_name(except);
+      symbolt lookup_results = ns.lookup(symbol_name);
+      to_exclude.insert(
+        id2string(lookup_results.location.get_file()) + ":" + except);
+    }
+  }
+
+  symbol_tablet &symbol_table = goto_model.symbol_table;
+
+  for(symbol_tablet::iteratort symbol_it = symbol_table.begin();
+      symbol_it != symbol_table.end();
+      symbol_it++)
+  {
+    symbolt &symbol = symbol_it.get_writeable_symbol();
+    std::string qualified_name = id2string(symbol.location.get_file()) + ":" +
+                                 id2string(symbol.display_name());
+    if(to_exclude.find(qualified_name) != to_exclude.end())
+    {
+      symbol.value.set(ID_C_no_nondet_initialization, 1);
+    }
+  }
+
+  nondet_static(ns, goto_model.goto_functions, INITIALIZE_FUNCTION);
 }
--- a/src/goto-instrument/nondet_static.h
+++ b/src/goto-instrument/nondet_static.h
@@ -20,6 +20,8 @@ Date: November 2011
 #ifndef CPROVER_GOTO_INSTRUMENT_NONDET_STATIC_H
 #define CPROVER_GOTO_INSTRUMENT_NONDET_STATIC_H
 
+#include <util/options.h>
+
 class goto_modelt;
 class namespacet;
 class goto_functionst;
@@ -34,5 +36,7 @@ void nondet_static(
   goto_functionst &goto_functions);
 
 void nondet_static(goto_modelt &);
+
+void nondet_static(goto_modelt &, const optionst::value_listt &);
 
 #endif // CPROVER_GOTO_INSTRUMENT_NONDET_STATIC_H
--- a/regression/goto-instrument/nondet-static-exclude_subfolder/main.c
+++ b/regression/goto-instrument/nondet-static-exclude_subfolder/main.c
@@ -1,0 +1,9 @@
+#include "package1/file.c"
+#include "package2/file.c"
+
+void main()
+{
+  method1();
+  method2();
+  return 0;
+}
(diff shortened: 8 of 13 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: f3502ff69956b84bc8dbf09c20e46b4871080ddb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/112499997

mmuesly:
@allredj could you give me an advise regarding what I should do to pass the compatibility check?
I can not access the linked build.

allredj:
@mmuesly Hi! You can disregard the compatibility check for now. Actually I've just disabled it because TG and CBMC are already incompatible at this time.

smowton:
Could I suggest for symmetry we should have `--nondet-static-variables` and `--nondet-static-exclude-variables`, with the rule "nondet if blanket nondet-static or nondet-static-variables AND not nondet-static-exclude-variables"?

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1006,7 +1021,8 @@ void cbmc_parse_optionst::help()
     "\n"
     "Semantic transformations:\n"
     // NOLINTNEXTLINE(whitespace/line_length)
-    " --nondet-static              add nondeterministic initialization of variables with static lifetime\n"
+    " --nondet-static              add nondeterministic initialization of variables with static lifetime\n" // NOLINT(*)
+    " --nondet-static-except e     same as nondet-static except variables listed in e\n" // NOLINT(*)
```
peterschrammel: As @smowton, I have a preference for the suffix `-exclude`. Also, comma-separation should not be used, but multiple instances of the option should be supplied instead.

regression/cbmc/nondet-static-except_subfolder/main.c
```diff
@@ -0,0 +1,9 @@
+#include "package1/file.c"
+#include "package2/file.c"
+
+void main()
+{
+  method1();
+  method2();
+  return 0;
+}
```
peterschrammel: missing newline

kroening:
This appears to be a feature that should go into the harness generator?

regression/cbmc/nondet-static-except_subfolder/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+package2/file.c
+--nondet-static-except package1/file.c:myVal,package2/file.c:myVal2 main.c package1/file.c
```
peterschrammel:
```suggestion
--nondet-static-exclude package1/file.c:myVal --nondet-static-exclude package2/file.c:myVal2 main.c package1/file.c 
```

karkhaz:
@kroening to clarify, are you asking for @mmuesly to do additional work to integrate this into the harness generator, or was your comment just a note to Diffblue people?

karkhaz:
@smowton not sure if I totally understand the semantics that you're proposing, but I see a potential for confusion here: what happens if a user passes both `--nondet-static` and `--nondet-static-variables-exclude`, but no `--nondet-static-variables`?

Your wording seems to imply that the exclude list would be entirely ignored, because `--nondet-static` makes everything nondet. On the other hand, if `--nondet-static` does in fact take into account the exclude list, then I don't see why it's any different to `--nondet-static-variables`.

The concrete reason why this is a concern is: we use a "common" Makefile that sets up most of the options to CBMC (like `--nondet-static`), using a variable like `CBMCFLAGS`. Then, for individual CBMC tests, we have a local makefile that appends to `CBMCFLAGS`---this is where we would set `--nondet-static-variables-except`. The local Makefile then runs `cbmc $(CBMCFLAGS) foo.c`. With your proposal, it seems very easy for a user to set `--nondet-static` (rather than `--nondet-static-variables`) at the global level, which would then silently override the exclude list.

If you really prefer the symmetry, I propose that we make `--nondet-static-variables` an alias for `--nondet-static`. Maybe even remove `--nondet-static` from all documentation, but keep it in the codebase so that we don't break existing users of that flag.

mmuesly:
I changed the PR as requested by @peterschrammel.
Further, I made sure that `--nondet-static` does not overwrite `--nondet-static-exclude`. I missed this side effect in the first draft. As described by @karkhaz, this is important for our use case.

kroening:
@karkhaz The PR should change to make this feature an addition to the harness generator (unless I am misunderstanding the purpose).
All features that produce a harness for a given piece of code should go there.

karkhaz:
@kroening this feature doesn't generate new harnesses. The new flag gives users finer control over what global variables are initialized to nondet. This flag can be used on arbitrary goto-binaries, so it's not related to harness generation.

mmuesly:
@kroening This PR is intended to modify the behavior of an existing CBMC flag and is a flag CBMC borrows from goto-instrument. It is adding only a couple of lines to the existing function implementation. Do you plan to move the original code from goto-instrument to the goto-harness package?

kroening:
@karkhaz if it's about transforming an existing program, it should go into goto-instrument!
cbmc should not get any options that change semantics of the goto program.

kroening:
@mmuesly The --nondet-static flag is indeed an issue; ideally, it should be removed from cbmc, which is at least hinted by the fact that its implementation lives in goto-instrument. The reason why it's still there is that there are existing users.

karkhaz:
@kroening OK. So I suggest that @mmuesly does the following:
1. Add the new flag to goto-instrument and remove it from cbmc
1. Move the tests to the goto-instrument regression directory. You won't
   have to change the tests very much, because (looking at
   `regressin/goto-instrument/chain.sh`) the tests in that directory have
   both `goto-instrument` and `cbmc` run over them. So you would use
   `goto-instrument` to instrument the binaries with
   `--nondet-static-exclude`, and then use CBMC to check that the
   verification result is as expected.

As a *separate* PR, we might consider removing `--nondet-static` from
the help output of CBMC (but keeping it in the actual codebase, to
prevent breaking legacy users).

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: dbda7c01c9f767974c21d337cbb863b5628759c6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114465285

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 585fdff016c43eef6e9d8a3ea8f94c528f098509).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114480311

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 5cd8b252b3cc68c3274363814135fd42cad58940).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114523926

mmuesly:
@peterschrammel I implemented @karkhaz suggestions.
What should I do next?
Should I open a second PR for removing nondet-static from the cbmc help output?
That is from my point of view the last open point no yet addressed raised in this discussion.

regression/goto-instrument/chain.sh
```diff
@@ -8,13 +8,23 @@ cbmc=$3
 is_windows=$4
 
 name=${*:$#}
-name=${name%.c}
+
+if [[ ! ($name =~ " ") ]];  then
```
tautschnig: Would it be easier to use `if [[ x$name != x ]] ; then` ?

regression/goto-instrument/chain.sh
```diff
@@ -8,13 +8,23 @@ cbmc=$3
...
+  name=${name%.c}
+fi
 
 args=${*:5:$#-5}
 
-if [[ "${is_windows}" == "true" ]]; then
+if [[ "${is_windows}" == "true" && ("${name}" =~ " ") ]]; then
```
tautschnig: As above - matching on whitespace is just a bit odd.

regression/goto-instrument/chain.sh
```diff
@@ -8,13 +8,23 @@ cbmc=$3
...
+  $goto_cc "main.gb" ${name}
+  name="main"
+  mv "${name}.exe" "${name}.gb"
+elif [[ "${is_windows}" == "true" ]]; then
   $goto_cc "${name}.c"
   mv "${name}.exe" "${name}.gb"
+elif [[ ("${name}" =~ " ") ]]; then
```
tautschnig: As above

regression/goto-instrument/nondet-static-exclude_subfolder/package1/file.c
```diff
@@ -0,0 +1,10 @@
+#include "assert.h"
+#include "stdint.h"
```
tautschnig: `#include <assert.h>`, `#include <stdint.h>` unless you really want to have the option to use a local copy of those files?

regression/goto-instrument/nondet-static-exclude_subfolder/package1/file.c
```diff
@@ -0,0 +1,10 @@
...
+static uint32_t myVal = 3;
+static uint32_t myVal2 = 4;
```
tautschnig: Is the use of `uint32_t` relevant for this test? Wouldn't `unsigned` or even `int` do?

regression/goto-instrument/nondet-static-exclude_subfolder/package2/file.c
```diff
@@ -0,0 +1,10 @@
+#include "assert.h"
+#include "stdint.h"
+static uint32_t myVal = 3;
+static uint32_t myVal2 = 4;
```
tautschnig: As above

regression/goto-instrument/nondet_static_exclude/test.desc
```diff
@@ -0,0 +1,18 @@
+CORE
+main_1.c main.c 
+--nondet-static --nondet-static-exclude test1 --nondet-static-exclude test3 --nondet-static-exclude test4 --nondet-static-exclude main_1.c:value
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+assertion test1 == (?:NULL||0): SUCCESS
```
tautschnig: Why such elaborate pattern matching?

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1607,6 +1616,7 @@ void goto_instrument_parse_optionst::help()
     " --isr <function>             instruments an interrupt service routine\n"
     " --mmio                       instruments memory-mapped I/O\n"
     " --nondet-static              add nondeterministic initialization of variables with static lifetime\n" // NOLINT(*)
+    " --nondet-static-exclude e    same as nondet-static except for the variable e (use multiple times if required)\n" // NOLINT(*)
```
tautschnig: This is a very long line (also in the output!), please line-wrap

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,68 @@ void nondet_static(goto_modelt &goto_model)
...
+  goto_modelt &goto_model,
+  const optionst::value_listt &except_values)
+{
+  const namespacet ns(goto_model.symbol_table);
+  std::set<std::string> to_exclude;
+
+  for(optionst::value_listt::const_iterator i = except_values.begin();
+      i != except_values.end();
+      ++i)
+  {
```
tautschnig, line 158: I'd suggest to use a ranged for: `for(const auto &except : except_values)`

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,68 @@ void nondet_static(goto_modelt &goto_model)
...
+    bool prefix_found = false;
+    for(size_t j = 0; j < i->length(); ++j)
+    {
+      if((*i)[j] == ':')
```
tautschnig: Use `i->find(':')`?!

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,68 @@ void nondet_static(goto_modelt &goto_model)
...
+    if(!prefix_found)
+    {
+      irep_idt symbol_name(*i);
+      symbolt lookup_results = ns.lookup(symbol_name);
+      to_exclude.insert(
+        id2string(lookup_results.location.get_file()) + ":" + *i);
+    }
+  }
+
+  symbol_tablet symbol_table = ns.get_symbol_table();
```
tautschnig: Why is a copy taken?

regression/goto-instrument/nondet_static_exclude/test.desc
mmuesly: Because in some test in CI this was NULL and in other it was 0.

What would you expect to be the default value in CI for all compilers?

src/goto-instrument/nondet_static.cpp
mmuesly: Because namespacet has a const symbol_table.
If I take a const reference the iterators are also const and I cannot call get_writable_symbol on the iterator. Which makes sense from my point of view that a const symbol_table must not be modified. Anyhow, I want to modify it. Therefore, I copy it and then modify the copy.

Eventually, I have overseen a possibility to this without the copy. But I don’t know how to get a symbol_tablet::iteratort instead of the symbolst::const_iterator if I have a const reference to a symbol table.

Any advice?

regression/goto-instrument/nondet_static_exclude/test.desc
tautschnig: That seems fine, but wouldn't then `(NULL|0)` do instead of `(?:NULL||0)`?

src/goto-instrument/nondet_static.cpp
tautschnig: You don't really need to go through `ns`, because the namespace itself you only constructed earlier from `goto_model.symbol_table` - I think you could just use that symbol table directly?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: b1450414a60e144b319d4cd92e987a40fb6e32ce).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115658522

regression/goto-instrument/nondet_static_exclude/test.desc
mmuesly: Yes

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 362ab4de235c4560e5567717827cf1eb5eef1e4d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115660321

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: afd0b0abbc6f272b27f78a00b76beeaef6aab97d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115662790

mmuesly:
@tautschnig Thank you for the review and comments. I fixed all build issues by now and addressed all your comments.

peterschrammel (approved):

src/goto-instrument/nondet_static.h
```diff
@@ -20,6 +20,8 @@ Date: November 2011
 #ifndef CPROVER_GOTO_INSTRUMENT_NONDET_STATIC_H
 #define CPROVER_GOTO_INSTRUMENT_NONDET_STATIC_H
 
+#include <util/options.h>
```
romainbrenguier, line 23: you can use a forward declaration instead if you use `std::list<std::string>` instead of `optionst::value_listt` below

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,64 @@ void nondet_static(goto_modelt &goto_model)
...
+      if(except.length() > 2 && except[0] == '.' && except[1] == '/')
+      {
+        to_exclude.insert(except.substr(2, except.length() - 2));
+      }
+      else
+      {
+        to_exclude.insert("./" + except);
+      }
+    }
+    if(!prefix_found)
```
romainbrenguier: replace by `else`

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,64 @@ void nondet_static(goto_modelt &goto_model)
...
+  goto_modelt &goto_model,
+  const optionst::value_listt &except_values)
+{
+  const namespacet ns(goto_model.symbol_table);
+  std::set<std::string> to_exclude;
+
+  for(auto const &except : except_values)
+  {
+    bool prefix_found = false;
+    auto it = except.find(":");
```
romainbrenguier: `const bool prefix_found = except.find(":") != std::string::npos;`

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,64 @@ void nondet_static(goto_modelt &goto_model)
...
+
+    if(it != std::string::npos)
+    {
+      prefix_found = true;
+      to_exclude.insert(except);
+      if(except.length() > 2 && except[0] == '.' && except[1] == '/')
```
romainbrenguier: use `has_prefix`

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,64 @@ void nondet_static(goto_modelt &goto_model)
...
+      {
+        to_exclude.insert(except.substr(2, except.length() - 2));
+      }
+      else
+      {
+        to_exclude.insert("./" + except);
```
romainbrenguier, line 170: I don't really understand this piece of code, if `except` is "./path" we insert "path" and if it is "path" we insert "./path"?

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,64 @@ void nondet_static(goto_modelt &goto_model)
...
+    std::string qualified_name = id2string(symbol.location.get_file()) + ":" +
+                                 id2string(symbol.display_name());
+    if(to_exclude.find(qualified_name) != to_exclude.end())
+    {
+      symbol.value.set(ID_C_no_nondet_initialization, 1);
+    }
+  }
+
+  // Update the symbol_table in the model
+  goto_model.symbol_table = symbol_table;
```
romainbrenguier: not needed, symbol table is already a reference to `goto_model.symbol_table`

romainbrenguier (changes requested):

src/goto-instrument/nondet_static.cpp
mmuesly: Yes. The location path returned by the namespace symbol after lookup sometimes starts with `./` and sometimes not. To make the lookup more deterministic, I add the prefix eventually to the passed parameter or remove it. So the exclude list always contains `path` and `./path` as an excluded location.

src/goto-instrument/nondet_static.h
mmuesly: Unless there is a plan to drop the `optionst::value_listt` in the long term, I personally think using explicitly this type makes it clearer what is expected as a parameter.

mmuesly:
@peterschrammel and @romainbrenguier, thank you for the reviews. I addressed romainbrenguier's comments.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8e158b5a57af04100f84a4eb3e1b80b1020fd6a6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116671062

src/goto-instrument/nondet_static.cpp
romainbrenguier: Ok I see. Maybe add a comment about that. Ideally I would expect the path to be stored in some canonical form (but I guess it's alright for now)

romainbrenguier: Ok I see. Maybe add a comment about that. Ideally I would expect the path to be stored in some canonical form (but I guess it's alright for now)

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,59 @@ void nondet_static(goto_modelt &goto_model)
...
+      to_exclude.insert(
+        id2string(lookup_results.location.get_file()) + ":" + except);
+    }
+  }
+
+  symbol_tablet &symbol_table = goto_model.symbol_table;
+
+  for(symbol_tablet::iteratort symbol_it = symbol_table.begin();
+      symbol_it != symbol_table.end();
+      symbol_it++)
```
romainbrenguier, line 186: Range for if possible (otherwise ++symbol_it and you can use auto for the iterator type)

src/goto-instrument/nondet_static.cpp
```diff
@@ -136,3 +140,59 @@ void nondet_static(goto_modelt &goto_model)
...
+  {
+    symbolt &symbol = symbol_it.get_writeable_symbol();
+    std::string qualified_name = id2string(symbol.location.get_file()) + ":" +
```
romainbrenguier, line 189: Const

src/goto-instrument/nondet_static.h
romainbrenguier: I personally disagree that it is clearer (the name value_listt carries strictly less information than list<string>), but I know people can have a different opinion on that so I won't insist.

romainbrenguier (approved):

PR #4300, opened 2019-02-28
[WIP] Use Visual Studio and CMake in Windows CI builds

tautschnig:
With CMake we can actually build natively on Windows/Visual Studio, there should
not be any need for Cygwin.

This is early testing, the PR will remain a draft until all issues are ironed out.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- n/a Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4678, opened 2019-05-21
Update generate_vcxproj to work with current source tree

tautschnig:
1) We have multiple source files with the same basename, and thus need
to instruct Visual Studio not to put build artefacts into a single
directory.
2) Directory dependencies have changed.
3) Specifically, cbmc now also depends on (parts of) goto-instrument;
yet we must not include goto_instrument_main.cpp in the CBMC build as it
would cause duplicate main functions.
4) Cleanup: While reviewing 2) I noticed that we had a duplicate
json$(LIBEXT) entry in CBMC's Makefile.

### DIFF
--- a/scripts/generate_vcxproj
+++ b/scripts/generate_vcxproj
@@ -16,6 +16,9 @@ function doit {
   for dir in $dirs ; do
     sources="`(cd $dest/src/$dir; make sources)`"
     for s in $sources ; do
+      if [ "$s" = "goto_instrument_main.cpp" ] && [ "$1" != "goto-instrument" ] ; then
+        continue
+      fi
       echo "    <ClCompile Include=\"src\\${dir}\\${s}\"/>" >> $dest/$1.vcxproj
     done
   done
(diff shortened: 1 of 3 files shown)


smowton:
Would it be easy to run and briefly verify the function of this script in our Win32 CI?

tautschnig:
> Would it be easy to run and briefly verify the function of this script in our Win32 CI?

Doable, if we are ok with increasing the Windows build time (Travis usually takes the most time, so it would likely be ok). I'll add a commit taking care of this.

scripts/generate_vcxproj
```diff
@@ -16,6 +16,9 @@ function doit {
   for dir in $dirs ; do
     sources="`(cd $dest/src/$dir; make sources)`"
     for s in $sources ; do
+      if [ "$s" = "goto_instrument_main.cpp" ] && [ "$s" != "${1}_main.cpp" ] ; then
+        continue
```
kroening, line 20: The first conjunct isn't subsumed by the second?

kroening:
I am quite worried about the ever-slower CI...

scripts/generate_vcxproj
tautschnig: We want to `continue` if seeing `goto_instrument_main.cpp` while not actually building `goto-instrument.vcxproj` - which actually tells me I still need to fix this and change the second conjunct to `[ "$1" != "goto-instrument" ]`... will fix.

tautschnig:
> I am quite worried about the ever-slower CI...

We could run it in parallel, but I don't know how/whether we can add more CodeBuild jobs.

tautschnig:
The Windows build is now failing in an expected way, PR to fix this will follow the next hours.

forejtv (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 3d0d151a320af965de456e09d1e19bf9ac1f0601).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113716934

scripts/generate_vcxproj
chrisr-diffblue: Very minor nit pick on inconsistent use of `${blah}` versus `$blah`?

chrisr-diffblue (approved):
Looks reasonable to me, so I'm happy for this to go in as is. The one nit pick is very minor, so I don't mind if you ignore it.

A more general question: Rather than having our own script to generate the project file, and effectively having yet another parallel build system (in the sense that its another script that has to be told about directories, various dependencies, etc) - can we just use CMake to geneerate the Visual Studio project file?

tautschnig:
> [...] can we just use CMake to geneerate the Visual Studio project file?

I think that's a route worth exploring, but does require more work. I'd prefer to make it part of my yet-to-be-completed work in #4300, which requires me experimenting with Windows a lot more.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 6c2d601ee3e440eb8c39feb2c871fd61b224f4f4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113745591

PR #4680, opened 2019-05-21
Array cell sensitivity

romainbrenguier:
This pull request activates the array field sensitivity code, which should allow constant propagation in arrays of constant size.
 


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4b5023f9716980216f570827977e7047fa944261).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122071988

romainbrenguier:
I ran some experiments on 250 methods from tika-parsers (with unwind 15). It seems it doesn't help on simple cases (even a bit slower) but on some more complicated examples it is much faster.
I tested this with two version of models for the standard library, one uses non-determinstic size arrays and the other one uses constant size arrays so that we can make more use of array cell sensitivity.
   
Develop, Models and Array cell sensitivity
(time in sec.)

https://docs.google.com/spreadsheets/d/e/2PACX-1vRJNu9_cmWBc2ptmMLNWkU2895p-Kk40XGPnGU6yuP4eI4okC5muU0S9dm-u7Wxyc5mqQlrraW3qogD/pubchart?oid=830404797&format=interactive

src/goto-symex/field_sensitivity.cpp
```diff
@@ -97,10 +97,21 @@ exprt field_sensitivityt::apply(
       // SSA expression
       ssa_exprt tmp = to_ssa_expr(index.array());
       bool was_l2 = !tmp.get_level_2().empty();
+
+      exprt array_size_expr = to_array_type(l2_array.get().type()).size();
+      if(array_size_expr.is_nil() && index.array().id() == ID_symbol)
+      {
+        // In case the array type was incomplete, attempt to retrieve it from
+        // the symbol table.
+        const symbolt *array_from_symbol_table = ns.get_symbol_table().lookup(
```
tautschnig, line 108: What is the justification for this to be correct? I don't see why we could use values from the symbol table during symbolic execution.

tautschnig (commented):
Just trying to make this manageable:
1) One comment on something that surprised me below.
2) Could any commits be factored out and be merged independently?
3) What is the timeline along which we are hoping to merge this?

regression/cbmc/address_space_size_limit3/main.i
```diff
@@ -92,6 +92,11 @@ void f__L_0x3c6_0()
 
 int main()
 {
+  int unknown;
+  // Avoid constant propagation solving for pointer-offsets of esp,
+  // which prevent demonstrating the object-size limitation this test
+  // intends to exhibit:
+  esp -= (unknown & 1);
```
tautschnig, line 99: This actually changes the semantics of the program? Maybe do `esp += (unknown & 1);` to counter that?

regression/cbmc/array-cell-sensitivity9/test.desc
```diff
@@ -0,0 +1,32 @@
+CORE
+test.c
+--show-vcc
```
tautschnig, line 3: It seems that all the tests only exercise `--show-vcc` - I think we should also be running the full flow (in addition to `--show-vcc`) to make sure we actually get the right results!

src/goto-symex/field_sensitivity.h
```diff
@@ -9,6 +9,10 @@ Author: Michael Tautschnig
 #ifndef CPROVER_GOTO_SYMEX_FIELD_SENSITIVITY_H
 #define CPROVER_GOTO_SYMEX_FIELD_SENSITIVITY_H
 
+/// Limit the size of arrays for which field_sensitivity gets applied.
+/// Necessary because large constant arrays slow-down the process.
+static constexpr unsigned max_field_sensitive_array_size = 64;
```
tautschnig: This should go in `util/magic.h`

tautschnig:
There are some `KNOWNBUG` tests in `regression/cbmc-concurrency` - I believe at least `struct_and_array1` should be fixed by this PR. The others are also worth exploring.

tautschnig:
@romainbrenguier @smowton Thanks you for building this! I've done a high-level review with a bunch of questions.

src/goto-symex/field_sensitivity.cpp
romainbrenguier: I've only seen that happen in one case which is this regression test: https://github.com/diffblue/cbmc/blob/develop/regression/cbmc/Global_Initialization2/main.c
In this example since `A1` is declared but not defined the length we get at this point in symex is not defined, although it appears correctly in the symbol table. @peterschrammel seemed to think that could be caused by a bug in the C frontend but we couldn't identify the source of this issue.

romainbrenguier:
@tautschnig 
> There are some KNOWNBUG tests in regression/cbmc-concurrency - I believe at least struct_and_array1 should be fixed by this PR. The others are also worth exploring.

The test doesn't work on windows:
`main.c(1): fatal error C1083: Cannot open include file: 'pthread.h': No such file or directory `
Do you have an idea of how to fix that?

smowton:
Tag the test `gcc-only`, and check the Makefile for that directory is generating an exclusion (`-X gcc-only`) when running under MSVC. See regression/cbmc/Makefile for an example.

smowton:
Benchmarking with SVCOMP shows no particular performance difference, but does uncover a probable bug: test `linux-4.0-rc1---drivers--net--ethernet--ethoc.ko.cil.i` exits with `l2_rename_rvalues case 'constant' not handled`

smowton:
Minimised bug to

```

  int x;
  const char *c = "Hello world";

  int *p = (argc ? &x : (int *)c);

  *p = 1;

  return 0;

}
```

smowton:
That example faults on current develop; fix at https://github.com/diffblue/cbmc/pull/4998

src/goto-symex/field_sensitivity.cpp
peterschrammel: As far as I remember the problem is that the array type is created with undefined length and then attached to various expressions from which the goto program is constructed. When The implementation of rule 6.9.2 (https://github.com/diffblue/cbmc/blob/develop/regression/cbmc/Global_Initialization2/main.c#L15) then updates the array length of the symbol in the symbol table only, but does not end up in the goto program. Symex then takes the length from the goto program, which doesn't contain the correct length - the correct length is in the symbol table only.
IMO there are two solutions for that: 1. get rid of the length in the array type. 2. In the C frontend, process declarations first to see whether all array types are complete, then update; only afterwards convert function bodies.

src/util/magic.h
```diff
@@ -16,4 +16,8 @@ const std::size_t MAX_CONCRETE_STRING_SIZE = 1 << 26;
 // The top end of the range of integers for which dstrings are precomputed
 constexpr std::size_t DSTRING_NUMBERS_MAX = 64;
 
+/// Limit the size of arrays for which field_sensitivity gets applied.
+/// Necessary because large constant arrays slow-down the process.
+constexpr unsigned max_field_sensitive_array_size = 64;
```
peterschrammel: Any reason why this is `unsigned` and not `size_t`?

peterschrammel: Any reason why this is lower case and not all-caps as all the other constants defined in this file?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a8a3e941df9a891ebbda25786446b2c8d30305d7).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123201104

src/goto-symex/field_sensitivity.cpp
peterschrammel: @romainbrenguier, please create a ticket in the issue tracker for this so that we can go ahead.

peterschrammel (approved):
modulo fixing clang and doxygen issues

src/goto-symex/field_sensitivity.cpp
romainbrenguier: issue created here https://github.com/diffblue/cbmc/issues/5022, I will add it to the commit message

romainbrenguier:
> Just trying to make this manageable:
> 1) One comment on something that surprised me below.

I've now raised an issue for it 

> 2) Could any commits be factored out and be merged independently?

I've already done that for as many things as I could, I don't think there is much we can extract now that would make sense on its own

> 3) What is the timeline along which we are hoping to merge this?

We think it's ready now, so as soon as possible.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 53e8c93c78ed97cdac3223bff787596273e360bc).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123473784

romainbrenguier:
@peterschrammel @smowton I've added options to choose the maximum length for which field sensitivity applies (two last commits)

src/goto-symex/symex_main.cpp
```diff
@@ -46,7 +46,11 @@ symex_configt::symex_configt(const optionst &options)
     run_validation_checks(options.get_bool_option("validate-ssa-equation")),
     show_symex_steps(options.get_bool_option("show-goto-symex-steps")),
     max_field_sensitivity_array_size(
-      options.get_unsigned_int_option("max-field-sensitivity-array-size"))
+      options.is_set("no-array-field-sensitivity")
```
smowton, line 49: We should produce an error if both are set

smowton:
Also genuine CI problems to fix; after that let's merge!

src/goto-symex/symex_main.cpp
romainbrenguier: I've added a check for that in jbmc/cbmc_parse_options

smowton:
Merged as #5057

PR #4681, opened 2019-05-21
Add support for --xml-ui to goto-analyzer

hannes-steffenhagen-diffblue:

### DIFF
--- a/src/goto-analyzer/static_verifier.cpp
+++ b/src/goto-analyzer/static_verifier.cpp
@@ -17,6 +17,7 @@ Author: Martin Brain, martin.brain@cs.ox.ac.uk
 #include <goto-programs/goto_model.h>
 
 #include <analyses/ai.h>
+#include <util/ui_message.h>
 
 struct static_verifier_resultt
 {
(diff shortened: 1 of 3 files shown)


xbauch (approved):
Looks good.

kroening:
I was really hoping to kill the XML stuff one day...

src/goto-analyzer/static_verifier.cpp
```diff
@@ -17,6 +17,7 @@ Author: Martin Brain, martin.brain@cs.ox.ac.uk
 #include <goto-programs/goto_model.h>
 
 #include <analyses/ai.h>
+#include <util/ui_message.h>
```
tautschnig, line 20: Nit pick: please add to the other `util` includes above.

src/goto-analyzer/static_verifier.cpp
```diff
@@ -156,67 +157,104 @@ static void static_verifier_console(
...
+      return "SUCCESS";
+    case static_verifier_resultt::FALSE:
+      return "FAILURE";
+    case static_verifier_resultt::BOTTOM:
+      return "SUCCESS";
+    case static_verifier_resultt::UNKNOWN:
+      return "UNKNOWN";
+    default:
+      INVARIANT(false, "result is one of TRUE, FALSE, BOTTOM or UNKNOWN");
+      return "";
```
tautschnig, line 176: The `default` case should be unnecessary. We have compiler warnings enabled for switch/case that don't cover all values of an enum.

src/goto-analyzer/static_verifier.cpp
```diff
@@ -156,67 +157,104 @@ static void static_verifier_console(
 {
   irep_idt last_function_id;
   irep_idt function_file;
-
-  for(const auto &result : results)
+  ui_message_handlert *ui_message_handler =
+    dynamic_cast<ui_message_handlert *>(&m.get_message_handler());
```
tautschnig, line 161: Hmm, couldn't we just have a `ui_message_handlert` as a class member here?

src/goto-analyzer/static_verifier.cpp
hannes-steffenhagen-diffblue: I always feel a bit weird about that because my opinion is that while _in theory_ we should be able to rely on enums having only the values of the constants they're declared with, but in practice the standard doesn't actually guarantee that :(

But I suppose if we have a code convention that we'll never convert from integers to enum values this is fine as well.

src/goto-analyzer/static_verifier.cpp
```diff
@@ -156,67 +157,104 @@ static void static_verifier_console(
...
+  auto const status_to_string =
```
peterschrammel, line 162: OMG... Time to continue my work on porting all analysers to goto-checker...

src/goto-analyzer/static_verifier.cpp
tautschnig: If we do those kinds of conversions we're probably in pretty big trouble. Ideally we'd actually have compiler warnings in place to really forbid those conversions.

PR #4682, opened 2019-05-21
Make check in XML output function a precondition

hannes-steffenhagen-diffblue:
Previously this check would just lead to a silent return when
trying to output an empty XML document. As per the comment this is
actually an error.

Also fixes one instance where this function was used incorrectly in
goto-analyzer.

### DIFF
--- a/src/util/xml.cpp
+++ b/src/util/xml.cpp
@@ -34,8 +34,7 @@ void xmlt::output(std::ostream &out, unsigned indent) const
   // 'name' needs to be set, or we produce mal-formed
   // XML.
 
-  if(name.empty())
-    return;
+  PRECONDITION(!name.empty());
 
   do_indent(out, indent);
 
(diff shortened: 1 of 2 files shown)


src/util/xml.cpp
```diff
@@ -34,8 +34,7 @@ void xmlt::output(std::ostream &out, unsigned indent) const
   // 'name' needs to be set, or we produce mal-formed
   // XML.
 
-  if(name.empty())
-    return;
+  PRECONDITION(!name.empty());
```
chrisr-diffblue, line 37: Good catch! - but should this also be an invariant/precondition in the constructor as well?

hannes-steffenhagen-diffblue: That'd make `xmlt` not default constructible anymore which'd be a bigger change.

I'm also noticing that the `xmlt` class seems to be limited to a single piece of text content, rather than allowing text and markup to intersperse like normal xml, but that's also a different issue.

tautschnig (approved):

chrisr-diffblue (approved):
Discussed off-line with Hannes. Approving as is.

PR #4686, opened 2019-05-21
Fix value-set-based pointer test resolution for non-deterministic pointer values

tautschnig:
When we do not have any entry in the value set we shouldn't rely on it
being a trustworthy source of information.


owen-mc-diffblue (approved):

src/goto-symex/symex_goto.cpp
```diff
@@ -142,7 +142,11 @@ static optionalt<renamedt<exprt, L2>> try_evaluate_pointer_comparison(
     }
   }
 
-  if(!constant_found)
+  if(value_set_elements.empty())
+  {
+    return {};
+  }
```
romainbrenguier: The fact that the function is not correct when the value set is empty is suspicious. 
Maybe we should check that the value set has size at least one just after the call to `get_value_set`, and can you say in which case it can be empty?

tautschnig: That is a very good point. It seems value sets are missing an `ID_nondet_symbol` case. I'll fix this. Marking do-not-merge in the meantime.

smowton (approved):
Adjusted title to more-accurately reflect what's being fixed

PR #4688, opened 2019-05-22
fix for handling of wchars on windows in getting the directory name [blocks: #4678]

BrettSchiff:
directory names needed to be fetched using wchar instead of char

### DIFF
--- a/src/util/file_util.cpp
+++ b/src/util/file_util.cpp
@@ -50,7 +50,7 @@ Date: January 2012
 /// \return current working directory
 std::string get_current_working_directory()
 {
-  #ifndef _WIN32
+#ifndef _WIN32
   errno=0;
   char *wd=realpath(".", nullptr);
 
@@ -60,13 +60,19 @@ std::string get_current_working_directory()
 
   std::string working_directory=wd;
   free(wd);
-  #else
-  char buffer[4096];
+#else
+  TCHAR buffer[4096];
   DWORD retval=GetCurrentDirectory(4096, buffer);
   if(retval == 0)
     throw system_exceptiont("failed to get current directory of process");
+
+#  ifdef UNICODE
+  std::string working_directory(narrow(buffer));
+#  else
   std::string working_directory(buffer);
-  #endif
+#  endif
+
+#endif
 
   return working_directory;
 }


src/util/file_util.cpp
```diff
@@ -61,11 +61,12 @@ std::string get_current_working_directory()
...
   free(wd);
   #else
-  char buffer[4096];
+  wchar_t buffer[4096];
   DWORD retval=GetCurrentDirectory(4096, buffer);
   if(retval == 0)
     throw system_exceptiont("failed to get current directory of process");
-  std::string working_directory(buffer);
+  std::wstring unicode_buffer(buffer);
+  std::string working_directory(unicode_buffer.begin(), unicode_buffer.end());
```
tautschnig: I'd suggest to use `std::string working_directory(narrow(buffer));` with an additional `#include "unicode.h"`.

src/util/file_util.cpp
```diff
@@ -61,11 +61,12 @@ std::string get_current_working_directory()
   std::string working_directory=wd;
   free(wd);
   #else
-  char buffer[4096];
+  wchar_t buffer[4096];
   DWORD retval=GetCurrentDirectory(4096, buffer);
```
tautschnig, line 65: Just checking: was Visual Studio now happy with `4096` or did it still complain about a buffer overrun?

tautschnig (commented):
I'd hope that my suggestions below also address the failing CI steps.

src/util/file_util.cpp
```diff
@@ -61,11 +61,12 @@ std::string get_current_working_directory()
   std::string working_directory=wd;
   free(wd);
   #else
-  char buffer[4096];
+  wchar_t buffer[4096];
```
smowton: This depends on whether you compile with `-DUNICODE` or not. Since this ever got committed / builds in CI in this state, it looks like some builds do not define `UNICODE`. `TCHAR buffer[4096]` will give a Unicode-agnostic buffer of either 8- or 16-bit chars as per configuration, and similarly some sort of `#ifdef UNICODE` should be used to guard the use of `wstring` below.

tautschnig: Oh, thank you, that explains why building via Visual Studio/msbuild yields different results than via `make`.

smowton: Alternatively of course we could assert that no right-minded person would ever build without `-DUNICODE` any longer and do this everywhere.

smowton: see https://docs.microsoft.com/en-us/windows/desktop/learnwin32/working-with-strings

src/util/file_util.cpp
BrettSchiff: It is happy with 4096, I imagine it's number of character rather than size in bytes

src/util/file_util.cpp
BrettSchiff: Ah, that is very useful, thank you!

tautschnig (approved):

PR #919, opened 2017-05-13
Stop inheriting from std containers

reuk:
The std containers aren't designed to be inherited-from. This patch aims to replace inheritance from std containers with composition. It also removes util/error.h because it seems that this file isn't included anywhere. Let me know if that should be done in a different PR.

PR #4690, opened 2019-05-22
Clean up in the value set API

romainbrenguier:
Several methods/members of value set are actually not used and should be deprecated to simplify the API. 
The `get_value_set` interface in which the result has to be given as argument is also not clear and it is better to return the result.

### DIFF
--- a/src/goto-symex/symex_dereference_state.h
+++ b/src/goto-symex/symex_dereference_state.h
@@ -34,11 +34,11 @@ protected:
   goto_symext::statet &state;
   const namespacet &ns;
 
-  DEPRECATED(SINCE(2019, 05, 22, "use list returning version instead"))
+  DEPRECATED(SINCE(2019, 05, 22, "use vector returning version instead"))
   void get_value_set(const exprt &expr, value_setst::valuest &value_set)
     const override;
 
-  value_setst::valuest get_value_set(const exprt &expr) const override;
+  std::vector<exprt> get_value_set(const exprt &expr) const override;
 
   const symbolt *get_or_create_failed_symbol(const exprt &expr) override;
 };
--- a/src/goto-symex/symex_dereference_state.h
+++ b/src/goto-symex/symex_dereference_state.h
@@ -34,8 +34,11 @@ protected:
   goto_symext::statet &state;
   const namespacet &ns;
 
+  DEPRECATED(SINCE(2019, 05, 22, "use list returning version instead"))
   void get_value_set(const exprt &expr, value_setst::valuest &value_set)
     const override;
+
+  value_setst::valuest get_value_set(const exprt &expr) const override;
 
   const symbolt *get_or_create_failed_symbol(const exprt &expr) override;
 };
--- a/src/pointer-analysis/dereference_callback.h
+++ b/src/pointer-analysis/dereference_callback.h
@@ -29,8 +29,11 @@ class dereference_callbackt
 public:
   virtual ~dereference_callbackt() = default;
 
+  DEPRECATED(SINCE(2019, 05, 22, "use list returning version instead"))
   virtual void
   get_value_set(const exprt &expr, value_setst::valuest &value_set) const = 0;
+
+  virtual value_setst::valuest get_value_set(const exprt &expr) const = 0;
 
   virtual const symbolt *get_or_create_failed_symbol(const exprt &expr) = 0;
 };
--- a/src/pointer-analysis/dereference_callback.h
+++ b/src/pointer-analysis/dereference_callback.h
@@ -29,11 +29,11 @@ class dereference_callbackt
 public:
   virtual ~dereference_callbackt() = default;
 
-  DEPRECATED(SINCE(2019, 05, 22, "use list returning version instead"))
+  DEPRECATED(SINCE(2019, 05, 22, "use vector returning version instead"))
   virtual void
   get_value_set(const exprt &expr, value_setst::valuest &value_set) const = 0;
 
-  virtual value_setst::valuest get_value_set(const exprt &expr) const = 0;
+  virtual std::vector<exprt> get_value_set(const exprt &expr) const = 0;
 
   virtual const symbolt *get_or_create_failed_symbol(const exprt &expr) = 0;
 };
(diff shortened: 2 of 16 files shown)


owen-mc-diffblue (approved):

tautschnig:
@romainbrenguier Various CI steps in Travis are failing.

src/pointer-analysis/value_set.h
```diff
@@ -90,70 +90,15 @@ class value_sett
   /// offsets (`offsett` instances). This is the RHS set of a single row of
   /// the enclosing `value_sett`, such as `{ null, dynamic_object1 }`.
   /// The set is represented as a map from numbered `exprt`s to `offsett`
-  /// instead of a set of pairs to make lookup by `exprt` easier. All
-  /// methods matching the interface of `std::map` forward those methods
-  /// to the internal map.
+  /// instead of a set of pairs to make lookup by `exprt` easier.
   class object_map_dt
+    : public std::map<object_numberingt::number_type, offsett>
```
tautschnig: This basically is a revert of 3b5263ff. We should have really good reasons if that's what should happen.

src/pointer-analysis/value_set.h
```diff
@@ -276,6 +277,7 @@ class value_sett
     const namespacet &ns) const;
 
   /// Appears to be unimplemented.
+  DEPRECATED(SINCE(2019, 05, 22, "Unimplemented"))
   expr_sett &get(const irep_idt &identifier, const std::string &suffix);
```
tautschnig, line 282: I think we can just remove what isn't even implemented.

src/pointer-analysis/goto_program_dereference.h
```diff
@@ -65,9 +65,12 @@ class goto_program_dereferencet:protected dereference_callbackt
 
   const symbolt *get_or_create_failed_symbol(const exprt &expr) override;
 
+  DEPRECATED(SINCE(2019, 05, 22, "use list returning version instead"))
   void
   get_value_set(const exprt &expr, value_setst::valuest &dest) const override;
 
+  std::list<exprt> get_value_set(const exprt &expr) const override;
```
smowton: If we're adding a new entry-point we probably shouldn't use `std::list`, there's no call for a doubly-linked list here AFAICT

smowton:
@owen-jones-diffblue could you check if `local_value_sett` uses `location_number`?

smowton (changes requested):
Suggest dropping the `std::map`-inheriting commit, which is a straight revert of https://github.com/diffblue/cbmc/pull/919, for the reasons given in that PR

src/pointer-analysis/value_set.h
romainbrenguier: Ok I see. I will drop that commit for now and try to find a better solution

romainbrenguier:
@tautschnig @smowton here is a PR to go instead of the std::map inheritance https://github.com/diffblue/cbmc/pull/4694

owen-mc-diffblue:
@smowton It does, but we can just add that field ourselves

romainbrenguier:
actually it seems location_number is used by value_set_domain_templatet so I will have to drop that commit too...

src/pointer-analysis/goto_program_dereference.h
romainbrenguier: but it is a list, and the fact that `value_setst::valuest` is a list is public, by writing `value_setst::valuest` we are just making it more complicated for the reader

src/pointer-analysis/value_set.h
romainbrenguier: I don't know. In test-gen there were cases of cpp files used instead of the cbmc ones which could add new implementation not existing in cbmc

src/pointer-analysis/goto_program_dereference.h
smowton: I'm not suggesting `valuest`, I'm suggesting `std::vector` because that's the most performant C++ sequence container when you don't have a good reason for something else (e.g. deque or list)

romainbrenguier: oh right

romainbrenguier:
@smowton I have now replace the use of list by vector

smowton (approved):

src/goto-symex/symex_dereference_state.h
```diff
@@ -34,9 +34,12 @@ class symex_dereference_statet:
   goto_symext::statet &state;
   const namespacet &ns;
 
+  DEPRECATED(SINCE(2019, 05, 22, "use list returning version instead"))
```
smowton: vector returning

src/pointer-analysis/dereference_callback.h
```diff
@@ -29,9 +29,12 @@ class dereference_callbackt
 public:
   virtual ~dereference_callbackt() = default;
 
+  DEPRECATED(SINCE(2019, 05, 22, "use list returning version instead"))
```
smowton: vector returning

src/pointer-analysis/value_sets.h
```diff
@@ -28,12 +28,19 @@ class value_setst
   typedef std::list<exprt> valuest;
 
   // this is not const to allow a lazy evaluation
+  DEPRECATED(SINCE(2019, 05, 22, "use list returning version instead"))
```
smowton: vector returning

romainbrenguier:
@owen-jones-diffblue @smowton I've added a commit which gets rid of doxygen warning that were generated for  this pull request, in case you want to have a look

tautschnig:
@romainbrenguier Please rebase now that #4729 is in to make CI happy.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 27f87e83a40685b0073ef88e1c34dcb65f9c2e8e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113842424

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fa1412c758edd7d4f50c92ddc73fbf5ba34791ae).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114073541

PR #4694, opened 2019-05-22
Replace definition of value_sett::object_map_dt 

romainbrenguier:
Instead of defining a new class with the same interface as std::map and
which use std::map internally, we use using and define outside the
static field `blank` which is the only thing added to std::map.


smowton (approved):
s/blank/empty, otherwise lgtm

tautschnig (approved):
clang-format isn't happy and +1 on using `empty_object_map` instead of `blank_object_map`, otherwise looks good, another nice piece of cleanup!

PR #4703, opened 2019-05-24
Preparation to make goto-analyzer work with the goto_verifiert framework

hannes-steffenhagen-diffblue:
* Make ai_baset work with `abstract_goto_modelt` rather than `goto_modelt`
* Add a version of `static_verifier` that works with `abstract_goto_modelt` and `propertiest`

This will (eventually) allow us to use the goto-checker stuff in `goto-analyzer`, which will be useful to make reporting of results and things like that more consistent with the rest of CPROVER.

### DIFF
--- a/src/goto-analyzer/static_verifier.cpp
+++ b/src/goto-analyzer/static_verifier.cpp
@@ -27,6 +27,49 @@ struct static_verifier_resultt
   irep_idt function_id;
 };
 
+void static_verifier(
+  const abstract_goto_modelt &abstract_goto_model,
+  const ai_baset &ai,
+  propertiest &properties)
+{
+  const namespacet ns{abstract_goto_model.get_symbol_table()};
+  // this is mutable because we want to change the property status
+  // in this loop
+  for(auto &property : properties)
+  {
+    auto &property_status = property.second.status;
+    const goto_programt::const_targett &property_location = property.second.pc;
+    exprt condition = property_location->get_condition();
+    const std::unique_ptr<ai_baset::statet> predecessor_state_copy =
+      ai.abstract_state_before(property_location);
+    // simplify the condition given the domain information we have
+    // about the state right before the assertion is evaluated
+    predecessor_state_copy->ai_simplify(condition, ns);
+    // if the condition simplifies to true the assertion always succeeds
+    if(condition.is_true())
+    {
+      property_status = property_statust::PASS;
+    }
+    // if the condition simplifies to false the assertion always fails
+    else if(condition.is_false())
+    {
+      property_status = property_statust::FAIL;
+    }
+    // if the domain state is bottom then the assertion is definitely
+    // unreachable
+    else if(predecessor_state_copy->is_bottom())
+    {
+      property_status = property_statust::NOT_REACHABLE;
+    }
+    // if the condition isn't definitely true, false or unreachable
+    // we don't know whether or not it may fail
+    else
+    {
+      property_status = property_statust::UNKNOWN;
+    }
+  }
+}
+
 /// Makes a status message string from a status.
 static const char *message(const static_verifier_resultt::statust &status)
 {
--- a/src/goto-analyzer/static_verifier.h
+++ b/src/goto-analyzer/static_verifier.h
@@ -9,8 +9,10 @@ Author: Martin Brain, martin.brain@cs.ox.ac.uk
 #ifndef CPROVER_GOTO_ANALYZER_STATIC_VERIFIER_H
 #define CPROVER_GOTO_ANALYZER_STATIC_VERIFIER_H
 
+#include <goto-checker/properties.h>
 #include <iosfwd>
 
+class abstract_goto_modelt;
 class ai_baset;
 class goto_modelt;
 class message_handlert;
@@ -22,5 +24,16 @@ bool static_verifier(
   const optionst &,
   message_handlert &,
   std::ostream &);
+
+/// Use the information from the abstract interpreter to fill out the statuses
+/// of the passed properties
+/// \param abstract_goto_model The goto program to verify
+/// \param ai The abstract interpreter (should be run to fixpoint
+///           before calling this function)
+/// \param properties The properties to fill out
+void static_verifier(
+  const abstract_goto_modelt &abstract_goto_model,
+  const ai_baset &ai,
+  propertiest &properties);
 
 #endif // CPROVER_GOTO_ANALYZER_STATIC_VERIFIER_H
--- a/src/analyses/ai.h
+++ b/src/analyses/ai.h
@@ -68,12 +68,12 @@ public:
   }
 
   /// Run abstract interpretation on a whole program
-  void operator()(const goto_modelt &goto_model)
+  void operator()(const abstract_goto_modelt &goto_model)
   {
-    const namespacet ns(goto_model.symbol_table);
-    initialize(goto_model.goto_functions);
-    entry_state(goto_model.goto_functions);
-    fixedpoint(goto_model.goto_functions, ns);
+    const namespacet ns(goto_model.get_symbol_table());
+    initialize(goto_model.get_goto_functions());
+    entry_state(goto_model.get_goto_functions());
+    fixedpoint(goto_model.get_goto_functions(), ns);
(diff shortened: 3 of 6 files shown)


src/goto-analyzer/static_verifier.cpp
```diff
@@ -27,6 +27,39 @@ struct static_verifier_resultt
...
+  const ai_baset &ai,
+  propertiest &properties)
+{
+  auto const ns = namespacet{abstract_goto_model.get_symbol_table()};
+  for(auto &property : properties)
+  {
+    auto &property_status = property.second.status;
+    auto const &property_location = property.second.pc;
+    auto condition = property_location->get_condition();
+    auto const predecessor = ai.abstract_state_before(property_location);
```
xbauch: It is not obvious to me why we want to evaluate the condition in the preceding state, rather than the current one? Is it possible that evaluating the property will change the value of the condition?

xbauch (approved):
It's not really used here, but it seems to do what it claims to be doing.

owen-mc-diffblue (approved):
The first commit is fine. I didn't review the second one as I know nothing about goto-analyzer.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4b0eb7f0a0cc9812e080edcc00a7508fc3d5a3d0).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113100556

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fadb25a30814da299d5252e1ad277fd881c4c129).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113111770

src/goto-analyzer/static_verifier.cpp
```diff
@@ -27,6 +27,39 @@ struct static_verifier_resultt
...
+    auto const &domain = *predecessor;
+    domain.ai_simplify(condition, ns);
+    if(condition.is_true())
+    {
+      property_status = property_statust::PASS;
+    }
+    else if(condition.is_false())
+    {
+      property_status = property_statust::FAIL;
```
peterschrammel, line 56: Is this a definite fail? If not, then it should be `UNKNOWN`.

src/goto-analyzer/static_verifier.cpp
```diff
@@ -27,6 +27,39 @@ struct static_verifier_resultt
...
+    }
+    else if(domain.is_bottom())
+    {
+      property_status = property_statust::NOT_REACHABLE;
+    }
+    else
+    {
+      property_status = property_statust::UNKNOWN;
```
peterschrammel, line 68: How can this happen? If the location is never assigned a value then it should be `NOT_REACHED`.

peterschrammel (approved):

src/goto-analyzer/static_verifier.cpp
hannes-steffenhagen-diffblue: Yes, if a condition is false that means that using domain information it could be simplified to false, i.e. definite fail.

src/goto-analyzer/static_verifier.cpp
hannes-steffenhagen-diffblue: This means the condition could not be simplified to true or false using domain information (so it's neither definite fail or success) and the domain state at the location is _not_ bottom so the location is reachable (edit: or rather, potentially reachable), therefore it is unknown.

src/goto-analyzer/static_verifier.h
```diff
@@ -9,6 +9,8 @@ Author: Martin Brain, martin.brain@cs.ox.ac.uk
 #ifndef CPROVER_GOTO_ANALYZER_STATIC_VERIFIER_H
 #define CPROVER_GOTO_ANALYZER_STATIC_VERIFIER_H
 
+#include <goto-checker/properties.h>
+#include <goto-programs/abstract_goto_model.h>
```
tautschnig: Wouldn't a forward declaration be sufficient?

src/goto-analyzer/static_verifier.cpp
```diff
@@ -27,6 +27,39 @@ struct static_verifier_resultt
   irep_idt function_id;
 };
 
+void static_verifier(
+  const abstract_goto_modelt &abstract_goto_model,
+  const ai_baset &ai,
+  propertiest &properties)
+{
+  auto const ns = namespacet{abstract_goto_model.get_symbol_table()};
```
tautschnig: I have complained about this before and I'll keep doing so: why is the above better than `const namespacet ns{abstract_goto_model.get_symbol_table()};`? To me it seems that would be shorter and avoid any doubt about 1) the type of the object and 2) whether some `operator=` is being invoked or not. And if "consistency" is part of the rationale: in fact the above is the only case of `auto` that's not in violation of our current coding guidelines...

I keep going after this, because I'm worried that we will at some point arrive at write-only code. Explicitly spelling out types may seem like a nuisance to the person writing the code, but they are really helpful to any reader. In the code below, I actually have no idea what the types of `property_status` or `predecessor` are. Alas I'm actually unable to judge whether (not) using references is the right thing.

src/goto-analyzer/static_verifier.cpp
chrisr-diffblue: It would be helpful to have this commentary on the various states in the code as well, and not just in the PR :-)

src/goto-analyzer/static_verifier.cpp
chrisr-diffblue: I have to say I share the concerns of @tautschnig too, though he's put them more eloquently than I probably would have done. I'm not a huge fan of `auto` in anything other than very trivial contexts.

hannes-steffenhagen-diffblue: I don't omit the types here because I don't care about readability, but rather because in my opinion they're not very useful for readability; I did another scan over the code and tried to clarify as much a possible, adding types where I thought they're useful. I kept the `auto` in the for loop and for `property_status` - the former because it'd otherwise just be `propertiest::value_type` which IMHO doesn't really add much useful information, and the latter because the type would be `property_statust` which again doesn't really communicate anything over the variable name.

(Side note: `operator=` is never involved in initialisation)

src/goto-analyzer/static_verifier.cpp
hannes-steffenhagen-diffblue: I added some comments explaining what was going on

src/goto-analyzer/static_verifier.h
hannes-steffenhagen-diffblue: Do you mean just for `abstract_goto_modelt`?

tautschnig: Yes

src/goto-analyzer/static_verifier.cpp
tautschnig: Thank you!

src/goto-analyzer/static_verifier.h
```diff
@@ -23,4 +25,9 @@ bool static_verifier(
   message_handlert &,
   std::ostream &);
 
+void static_verifier(
```
romainbrenguier, line 34: This should be documented. And it's a bit confusing to use a noun for a function which is returning void. (for me "verifier"    here should imply you return an object which has a `verify` method)

hannes-steffenhagen-diffblue: This is doing more or less the same thing as the other `static_verifier` function that already is in this file, the only difference is that it writes the result to the `propertiest` passed in rather than writing them to stdout or to a file.

The "return" value here is the `properties` parameter (it's an in/out parameter, we read some information from each property and write its verification status back to it)

tautschnig:
@hannes-steffenhagen-diffblue Please rebase now that #4729 is in to make CI happy.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 0f4a362aad04ca3d9a201b75b45f47ca59fdd727).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114085606

PR #4710, opened 2019-05-27
build_reference_to must consider the offset when returning a null object

tautschnig:
This has been buggy for a long time, but is now used more extensively
with value-set based filtering. The included regression test shows an
example where filtering would wrongly conclude that a condition is
always false.

### DIFF
--- a/regression/cbmc/null6/main.c
+++ b/regression/cbmc/null6/main.c
@@ -1,0 +1,17 @@
+void *guard_malloc_counter = 0;
+
+void *my_malloc(int size)
+{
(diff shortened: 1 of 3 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fc3a6ef2c77e844e3c2c2c10df8d668df8cf84ef).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113344747

owen-mc-diffblue (approved):

PR #4712, opened 2019-05-27
Fix unbounded recursion in byte-update lowering of structs

tautschnig:
The missing update of the type of the byte-update expression caused the
very same procedure to be invoked recursively. Instead we should use
byte-update-of-array.

### DIFF
--- a/src/solvers/lowering/byte_operators.cpp
+++ b/src/solvers/lowering/byte_operators.cpp
@@ -1787,6 +1787,10 @@ static exprt lower_byte_update_struct(
     }
     else if(!offset_bytes.has_value())
     {
+      // The offset to update is not constant; abort the attempt to update
+      // indiviual struct members and instead turn the operand-to-be-updated
+      // into a byte array, which we know how to update even if the offset is
+      // non-constant.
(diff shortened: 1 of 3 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a51a4afd8616b6fa796f0b18b75a4e6dad300b8e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113349760

src/solvers/lowering/byte_operators.cpp
```diff
@@ -1802,6 +1806,7 @@ static exprt lower_byte_update_struct(
 
       byte_update_exprt bu = src;
       bu.set_op(lower_byte_extract(byte_extract_expr, ns));
+      bu.type() = bu.op().type();
```
smowton, line 1809: Perhaps this should always be a side-effect of `set_op`, much as the constructor copies its type?

smowton (approved):

src/solvers/lowering/byte_operators.cpp
tautschnig: Fair point, I'll create a separate PR for that.

PR #4717, opened 2019-05-28
Byte-operator lowering: Add support for byte-extracting unions

tautschnig:
We already had support for structs, arrays, vectors in place; add unions
following the same approach. This is required for some SV-COMP device
driver benchmarks, including
ldv-linux-3.16-rc1/43_2a_consumption_linux-3.16-rc1.tar.xz-43_2a-drivers--scsi--qla2xxx--tcm_qla2xxx.ko-entry_point.cil.out.i

### DIFF
--- a/src/solvers/lowering/byte_operators.cpp
+++ b/src/solvers/lowering/byte_operators.cpp
@@ -940,6 +940,18 @@ static exprt unpack_rec(
         ns,
         true);
     }
+    else if(!union_type.components().empty())
+    {
+      member_exprt member{src, union_type.components().front()};
+      return unpack_rec(
+        member,
+        little_endian,
+        offset_bytes,
+        max_bytes,
+        bits_per_byte,
+        ns,
+        true);
+    }
   }
   else if(src.type().id() == ID_pointer)
   {
--- a/regression/cbmc/union17/main.c
+++ b/regression/cbmc/union17/main.c
@@ -1,0 +1,19 @@
+int main()
+{
+  // create a union type of non-constant, non-zero size
+  unsigned x;
+  __CPROVER_assume(x > 0);
+  union U
+  {
+    unsigned A[x];
+  };
+  // create an integer of arbitrary value
+  int i, i_before;
+  i_before = i;
+  // initialize a union of non-zero size from the integer
+  unsigned u = ((union U *)&i)->A[0];
+  // reading back an integer out of the union should yield the same value for
+  // the integer as it had before
+  i = u;
+  __CPROVER_assert(i == i_before, "going through union works");
+}
--- a/regression/cbmc/union17/test.desc
+++ b/regression/cbmc/union17/test.desc
@@ -1,0 +1,13 @@
+CORE broken-smt-backend
+main.c
+--no-simplify
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^warning: ignoring
+--
+This test passes when simplification is enabled (which gets rid of
+byte-extracting a union of non-constant size), but yielded a wrong verification
+outcome with both the SAT back-end before. The SMT back-end fails for it would
+like to flatten an array of non-constant size.


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9384d50b3551815d7f7ac296375a38e16149b163).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/113362951

src/solvers/lowering/byte_operators.cpp
```diff
@@ -108,6 +108,40 @@ static struct_exprt bv_to_struct_expr(
...
+      continue;
+
+    max_width = *element_width;
+    max_comp_type = comp.type();
+    max_comp_name = comp.get_name();
+  }
+
+  exprt operand = nil_exprt{};
+
+  if(max_width > 0)
```
smowton: Can a zero-width element occur, or does that always indicate the loop above never found a known-size component?

src/solvers/lowering/byte_operators.cpp
```diff
@@ -108,6 +108,40 @@ static struct_exprt bv_to_struct_expr(
...
+  irep_idt max_comp_name;
+
+  for(const auto &comp : components)
+  {
+    auto element_width = pointer_offset_bits(comp.type(), ns);
+
+    if(!element_width.has_value() || *element_width <= max_width)
+      continue;
+
+    max_width = *element_width;
```
smowton: Explicitly  note that if two components have equal size we favour the later one

src/solvers/lowering/byte_operators.cpp
```diff
@@ -108,6 +108,40 @@ static struct_exprt bv_to_struct_expr(
...
+    max_comp_type = comp.type();
+    max_comp_name = comp.get_name();
+  }
+
+  exprt operand = nil_exprt{};
+
+  if(max_width > 0)
+    operand = bv_to_expr(bitvector_expr, max_comp_type, endianness_map, ns);
+
+  return union_exprt{std::move(max_comp_name), std::move(operand), union_type};
```
smowton: `max_comp_name` may be uninitialised here, in which case I suspect we shouldn't be trying to raise to a `union_exprt` at all?

smowton:
Also tests please, including one with multiple like-sized components (e.g. `union { double d; uint64_t ul; }`)

src/solvers/lowering/byte_operators.cpp
tautschnig: Cleanup done.

src/solvers/lowering/byte_operators.cpp
tautschnig: (Actually the former one, but also I added some comments.)

src/solvers/lowering/byte_operators.cpp
tautschnig: There's the case of an empty union - comments added (and there isn't any risk of uninitialised values anymore).

tautschnig:
> Also tests please, including one with multiple like-sized components (e.g. `union { double d; uint64_t ul; }`)

Test added, but note that this code will only be reached when none of the members as a known non-zero size.

smowton (commented):
Hmm, are you really sure we want to do lifting to the first component if it's the wrong size, rather than just abandoning conversion and doing whatever we did before this code existed?

tautschnig:
> Hmm, are you really sure we want to do lifting to the first component if it's the wrong size, rather than just abandoning conversion and doing whatever we did before this code existed?

No, I'm not sure :-) - but what we did before this code existed is running into a failing invariant that said that this case is unsupported.

smowton:
Perhaps it would be better to have an invariant that if we're trying to turn a `bv[N]` into a `union { T1 e1; T2 e2; ... }` then at least one of `T1` ... `Tn` has the correct size, then we'll know when this code is doing something nonsensical and can address that particular case?

regression/cbmc/union14/main.c
```diff
@@ -0,0 +1,13 @@
...
+  unsigned x;
+  __CPROVER_assume(x > 0);
+  union U {
+    unsigned A[x];
+  };
+  int i, i_before;
+  i_before = i;
+  union U u = *((union U *)&i);
+  i = u.A[0];
+  __CPROVER_assert(i == i_before, "going through union works");
```
martin-cs: There is quite a lot going on in this test, esp. in making it endian independent. If you have to modify the PR please can you add a few comments on what is going on.

src/solvers/flattening/boolbv_byte_extract.cpp
```diff
@@ -38,15 +38,15 @@ bvt map_bv(const bv_endianness_mapt &map, const bvt &src)
 
 bvt boolbvt::convert_byte_extract(const byte_extract_exprt &expr)
 {
+  const std::size_t width = boolbv_width(expr.type());
+
   // array logic does not handle byte operators, thus lower when operating on
   // unbounded arrays
-  if(is_unbounded_array(expr.op().type()))
+  if(is_unbounded_array(expr.op().type()) || is_unbounded_array(expr.type()) || width == 0)
```
martin-cs: The test of `expr.type()` and `expr.op.type()` makes me a little nervous.  What about arrays at the start of structs or unions containing unions?

martin-cs (approved):
Approve but please do have a look at the comments.

src/solvers/flattening/boolbv_byte_extract.cpp
tautschnig: I have expanded `is_unbounded_array` to cover these cases.

regression/cbmc/union14/main.c
tautschnig: Done.

peterschrammel (approved):

PR #2946, opened 2018-09-13
[Draft] Refactor more simplify_plus to take plus_exprt

hannes-steffenhagen-diffblue:
Based on https://github.com/diffblue/cbmc/pull/2939

Does the same sort of refactoring for plus_exprt. Because this depends on multiple PRs that each are still subject to change this should be considered a draft that's not ready to be merged.

Only 97cd89870b44f3b0ff4cfbcffd07b20d70dd94e9 is really new in this PR, the others are from work-in-progress PRs. Please disregard them.

PR #2939, opened 2018-09-12
Refactor simplify floatbv typecast

hannes-steffenhagen-diffblue:
Note: Only f9165bae (3fa4e19 before rebase, still has some comments on it) is new, the prior commits are from https://github.com/diffblue/cbmc/pull/2898

PR #4732, opened 2019-05-30
introduce simplify_exprt::resultt

kroening:
This is follow-up on #2939 and #2946, and the biggest change to the simplifier since its existence.

It attempts to make the implementation of individual simplification steps more type safe, as opposed to using `exprt`.

The alternative considered was to return `optionalt<exprt>`, which has the disadvantage that it doesn't "chain", i.e., given two transformations `f` and `g` one cannot write `f(g(...))`.

### DIFF
--- a/src/util/simplify_expr.cpp
+++ b/src/util/simplify_expr.cpp
@@ -2405,131 +2405,131 @@ bool simplify_exprt::simplify_node_preorder(exprt &expr)
 bool simplify_exprt::simplify_node(exprt &expr)
 {
   if(!expr.has_operands())
-    return true;
+    return true; // no change
 
-  // #define DEBUGX
+    // #define DEBUGX
 
-  #ifdef DEBUGX
+#ifdef DEBUGX
   exprt old(expr);
-  #endif
+#endif
 
-  bool result=true;
+  bool no_change = true;
 
-  result=sort_and_join(expr) && result;
+  no_change = sort_and_join(expr) && no_change;
 
   if(expr.id()==ID_typecast)
-    result=simplify_typecast(expr) && result;
+    no_change = simplify_typecast(expr) && no_change;
   else if(expr.id()==ID_equal || expr.id()==ID_notequal ||
           expr.id()==ID_gt    || expr.id()==ID_lt ||
           expr.id()==ID_ge    || expr.id()==ID_le)
-    result=simplify_inequality(expr) && result;
+    no_change = simplify_inequality(expr) && no_change;
   else if(expr.id()==ID_if)
-    result=simplify_if(to_if_expr(expr)) && result;
+    no_change = simplify_if(to_if_expr(expr)) && no_change;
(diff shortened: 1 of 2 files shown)


tautschnig:
Could the commit message please include a body - which might actually just what's being said in the PR description?

tautschnig (commented):
The commit in itself looks promising, but also it's not easy to judge without seeing it in action.

kroening:
@tautschnig Now with non-trivial example.

src/util/simplify_expr.cpp
```diff
@@ -2405,139 +2405,139 @@ bool simplify_exprt::simplify_node_preorder(exprt &expr)
 bool simplify_exprt::simplify_node(exprt &expr)
 {
   if(!expr.has_operands())
-    return true;
+    return true; // no change
 
-  // #define DEBUGX
+    // #define DEBUGX
```
tautschnig, line 2371: Except for this seemingly wrong/unnecessary change: could we please just have this commit merged as soon as possible? It seems like a big improvement towards readability without any immediate dependence on the other changes.

src/util/simplify_expr.cpp
```diff
@@ -272,91 +272,79 @@ bool simplify_exprt::simplify_function_application(exprt &expr)
...
+  if(expr.op().id() == ID_infinity)
   {
     typet new_type=expr.type();
-    exprt tmp;
-    tmp.swap(expr.op0());
+    exprt tmp = expr.op();
     tmp.type()=new_type;
-    expr.swap(tmp);
-    return false;
+    return tmp;
```
tautschnig: Do we need `std::move(tmp)` here?

tautschnig: Actually, don't we need `return changed(std::move(tmp));` here?

src/util/simplify_expr.cpp
```diff
@@ -272,91 +272,79 @@ bool simplify_exprt::simplify_function_application(exprt &expr)
...
     binary_relation_exprt inequality(
-      expr.op0(),
+      expr.op(),
       op_type.id() == ID_floatbv ? ID_ieee_float_notequal : ID_notequal,
       from_integer(0, op_type));
     inequality.add_source_location()=expr.source_location();
     simplify_node(inequality);
-    expr.swap(inequality);
-    return false;
+    return inequality;
```
tautschnig: Unless I'm mistaken this needs a lot more use of `std::move` at about any return statement?

kroening:
The std::move are added!

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 23bfef15f033d2dea9009451e9ead3e2dc923a72).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114390981

tautschnig (approved):

PR #4735, opened 2019-05-30
Clean-up symex assignments

romainbrenguier:
This introduce an SSA_assignment_stept class to clarify what describes an assignment step (this could be a start to progressively split the SSA_stept class into several more specialized ones) and does some simple clean-up. 

### DIFF
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -79,9 +79,13 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
   }
 }
 
-exprt goto_symext::add_to_lhs(
-  const exprt &lhs,
-  const exprt &what)
+/// Store the \p what expression by recursively descending into the operands
+/// of \p lhs until the first operand \c op0 is _nil_: this _nil_ operand
+/// is then replaced with \p what.
+/// \param lhs: Non-symbol pointed-to expression
+/// \param what: The expression to be added to the \p lhs
+/// \return The resulting expression
+static exprt add_to_lhs(const exprt &lhs, const exprt &what)
 {
   PRECONDITION(lhs.id() != ID_symbol);
   exprt tmp_what=what;
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -569,8 +569,6 @@ void goto_symext::symex_assign_array(
   //   a'==a WITH [i:=e]
 
   with_exprt new_rhs(lhs_array, lhs_index, rhs);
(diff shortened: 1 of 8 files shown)


src/goto-symex/symex_assign.cpp
```diff
@@ -17,6 +17,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/exception_utils.h>
 #include <util/pointer_offset_size.h>
 #include <util/simplify_expr.h>
+#include <util/expr_util.h>
```
tautschnig: Please sort lexicographically (I'm pretty sure clang-format will block on this)

tautschnig (approved):

src/goto-symex/symex_assign.cpp
```diff
@@ -485,12 +486,12 @@ void goto_symext::symex_assign_symbol(
               << messaget::eom;
     });
 
-  // Temporarily add the state guard
-  guard.emplace_back(state.guard.as_expr());
+  const exprt assignment_guard =
+    make_and(state.guard.as_expr(), conjunction(guard));
```
owen-mc-diffblue, line 490: Presumably this was done this way in the name of performance. Have you checked that your change doesn't noticeably slow symex down?

owen-mc-diffblue (approved):

tautschnig:
@romainbrenguier This now needs a rebase, which I guess would also sort out the CI failure (it's a spurious clang-format report).

src/goto-symex/symex_assign.cpp
romainbrenguier: Before this was adding an exprt to a vector then using this vector to make the `and_exprt` of the conjunction, now we first make the conjunction and then add the exprt to the vector of the `and_exprt` operands. So the operation are the same, just in a different order, this shouldn't impact performances.

romainbrenguier: @owen-jones-diffblue I just noticed that my change could lead to `assignment_guard` being `TRUE && TRUE` so I've added a commit to have `make_and` handle the constant cases, can you check that?

owen-mc-diffblue: It seems good, once I saw from `exprt::is_true()` that boolean constant `exprt`s must be either `true_exprt` or `false_exprt`.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 63e01d9bf47bc49a3f4231927db3c6f5a0b51662).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114089912

PR #4733, opened 2019-05-30
Unit test ssa_exprt

romainbrenguier:
This adds unit test for the ssa_exprt class ~and clean-up some things I noticed while writing the tests, in particular the constructor cannot take any kind of `exprt` but a very specific class of these, which I documented.~

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [na] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [na] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [na] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4736, opened 2019-05-30
Clean-up ssa-exprt

romainbrenguier:
These are clean-up commits extracted https://github.com/diffblue/cbmc/pull/4733
Performance needs to be evaluated

### DIFF
--- a/src/util/ssa_expr.cpp
+++ b/src/util/ssa_expr.cpp
@@ -50,71 +50,6 @@ ssa_exprt::ssa_exprt(const exprt &expr) : symbol_exprt(expr.type())
   set(ID_L1_object_identifier, id);
 }
 
-void ssa_exprt::set_expression(const exprt &expr)
-{
-  type() = expr.type();
-  add(ID_expression, expr);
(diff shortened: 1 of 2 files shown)


romainbrenguier:
@tautschnig Here are the results from my benchmarks:

chart (4)

on average this branch was 3% faster (85.4s instead of 88.2) so there is no big slow-down because of putting functions in the cpp file (I wouldn't expect to have one anyway since they probably wouldn't be inlined), and I think we get a slight speed up because of the initialization which a bit more clever.

tautschnig (approved):
Thank you for the cleanup and the performance confirmation. clang-format is still unhappy, so that's the only bit that needs cleaning up.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 926c7c084d1cf465a5190d04e6f37c6e8cef364a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114599007

owen-mc-diffblue (approved):

PR #4755, opened 2019-06-06
Clean-up and unit test the renaming level classes

romainbrenguier:
This removes unnecessary inheritance, and makes the public interface of symex_level1t smaller so that it can be more significantly unit-tested. In particular we ensure that it is only used to map identifiers which come from L0 renamed ssa_exprt.

### DIFF
--- a/src/goto-symex/goto_symex_state.cpp
+++ b/src/goto-symex/goto_symex_state.cpp
@@ -798,23 +798,13 @@ ssa_exprt goto_symex_statet::add_object(
 
   ssa_exprt ssa = rename_ssa<L0>(ssa_exprt{expr}, ns).get();
   const irep_idt l0_name = ssa.get_identifier();
-  const std::size_t l1_index = index_generator(l0_name);
+  const auto l1_index = narrow_cast<unsigned>(index_generator(l0_name));
 
-  const auto r_opt = level1.current_names.find(l0_name);
-
-  if(!r_opt)
-  {
-    level1.current_names.insert(l0_name, std::make_pair(ssa, l1_index));
-  }
-  else
+  if(const auto old_value = level1.insert_or_replace(ssa, l1_index))
   {
     // save old L1 name
     if(!frame.old_level1.has_key(l0_name))
-    {
-      frame.old_level1.insert(l0_name, r_opt->get());
-    }
-
-    level1.current_names.replace(l0_name, std::make_pair(ssa, l1_index));
+      frame.old_level1.insert(l0_name, *old_value);
   }
 
   ssa = rename_ssa<L1>(std::move(ssa), ns).get();
--- a/src/goto-symex/goto_symex_state.cpp
+++ b/src/goto-symex/goto_symex_state.cpp
@@ -135,21 +135,22 @@ template <>
 renamedt<ssa_exprt, L0>
 goto_symex_statet::set_indices<L0>(ssa_exprt ssa_expr, const namespacet &ns)
 {
-  return level0(std::move(ssa_expr), ns, source.thread_nr);
+  return symex_level0(std::move(ssa_expr), ns, source.thread_nr);
 }
 
 template <>
 renamedt<ssa_exprt, L1>
 goto_symex_statet::set_indices<L1>(ssa_exprt ssa_expr, const namespacet &ns)
 {
-  return level1(level0(std::move(ssa_expr), ns, source.thread_nr));
+  return level1(symex_level0(std::move(ssa_expr), ns, source.thread_nr));
 }
 
 template <>
 renamedt<ssa_exprt, L2>
 goto_symex_statet::set_indices<L2>(ssa_exprt ssa_expr, const namespacet &ns)
 {
-  return level2(level1(level0(std::move(ssa_expr), ns, source.thread_nr)));
+  return level2(
+    level1(symex_level0(std::move(ssa_expr), ns, source.thread_nr)));
 }
 
 renamedt<ssa_exprt, L2> goto_symex_statet::assignment(
--- a/src/goto-symex/goto_symex_state.cpp
+++ b/src/goto-symex/goto_symex_state.cpp
@@ -803,8 +803,8 @@ ssa_exprt goto_symex_statet::add_object(
   if(const auto old_value = level1.insert_or_replace(renamed, l1_index))
   {
     // save old L1 name
-    if(!frame.old_level1.has_key(l0_name))
-      frame.old_level1.insert(l0_name, *old_value);
+    if(!frame.old_level1.has(renamed))
+      frame.old_level1.insert(renamed, old_value->second);
   }
 
   const ssa_exprt ssa = rename_ssa<L1>(renamed.get(), ns).get();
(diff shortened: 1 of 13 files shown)


src/goto-symex/goto_symex_state.cpp
```diff
@@ -796,28 +796,18 @@ ssa_exprt goto_symex_statet::add_object(
...
+  if(const auto old_value = level1.insert_or_replace(renamed, l1_index))
   {
     // save old L1 name
-    if(!frame.old_level1.has_key(l0_name))
-    {
-      frame.old_level1.insert(l0_name, r_opt->get());
-    }
-
-    level1.current_names.replace(l0_name, std::make_pair(ssa, l1_index));
+    if(!frame.old_level1.has(renamed))
```
smowton, line 807: Normally I'd expect an `insert` in C++ to be a no-op if the key is not present

src/goto-symex/renaming_level.cpp
```diff
@@ -97,6 +136,12 @@ void symex_level1t::restore_from(const current_namest &other)
   }
 }
 
+unsigned symex_level2t::current_count(const irep_idt &identifier) const
```
smowton: This name sounds too much like `std::map::count`, how about `latest_renaming_index`?

src/goto-symex/symex_start_thread.cpp
```diff
@@ -67,16 +67,16 @@ void goto_symext::symex_start_thread(statet &state)
     ssa_exprt lhs(pair.second.first.get_original_expr());
 
     // get L0 name for current thread
-    lhs.set_level_0(t);
-    const irep_idt &l0_name = lhs.get_identifier();
+    const renamedt<ssa_exprt, L0> l0_lhs =
+      symex_level0t{}(std::move(lhs), ns , t);
```
smowton: So it's default constructed then `operator()`'d? Could that be improved? Perhaps comment what's going on?

smowton (approved):
Mostly lgtm just a few minor comments

src/goto-symex/goto_symex_state.cpp
romainbrenguier: You mean if it's present?
Instead `sharing_mapt` has a precondition what ensures we never insert twice the same key, it seems more secure to abort the program there rather than silently doing nothing. The alternative solution would be to give another name to insert.

src/goto-symex/renaming_level.cpp
romainbrenguier: Maybe `latest_index`? "renaming" is kind of implied when talking about `symex_level2t`

src/goto-symex/goto_symex_state.cpp
smowton: Ah crap forgot this was `sharing_map`. Indeed it surely ought to have the same semantics as `std::map::insert` and its current `insert` should be called something like `insert_fresh`

src/goto-symex/symex_start_thread.cpp
romainbrenguier: I've made this functor a function instead so it should be better

src/goto-symex/goto_symex_state.cpp
romainbrenguier: It makes sense but I would leave that for another pull request since this is related more to sharing map than renamings

src/goto-symex/renaming_level.h
```diff
@@ -56,7 +56,14 @@ struct symex_level0t
 /// This is to preserve locality in case of recursion
 struct symex_level1t
 {
-  symex_renaming_levelt current_names;
```
owen-mc-diffblue: This deletion should be in the next commit (" Make current_names private in symex_level1t"), I think

unit/goto-symex/symex_level1.cpp
```diff
@@ -0,0 +1,68 @@
...
+        auto renamed = symex_level1(l0_symbol);
+        REQUIRE(renamed.get().get_identifier() == "foo!50");
+      }
+    }
+
+    WHEN("The symbol is inserted in symex_level1 several times")
+    {
+      symex_level1.insert(l0_symbol, 12134);
+      auto renamed = symex_level1(l0_symbol);
+      REQUIRE(renamed.get().get_identifier() == "foo!50@12134");
```
owen-mc-diffblue: I'd expect some `THEN` sections

unit/goto-symex/symex_level1.cpp
```diff
@@ -0,0 +1,68 @@
...
+      REQUIRE(old.has_value());
+      REQUIRE(old->second == 12134);
+      auto renamed2 = symex_level1(l0_symbol);
+      REQUIRE(renamed2.get().get_identifier() == "foo!50@43950");
+    }
+
+    WHEN("insert_or_replace is called with a symbol not already inserted")
+    {
+      auto old = symex_level1.insert_or_replace(l0_symbol, 20051);
+      ;
```
owen-mc-diffblue: Stray semicolon

owen-mc-diffblue (approved):

src/goto-symex/renaming_level.h
```diff
@@ -42,6 +42,9 @@ void get_variables(
 /// The renaming is built for one particular interleaving.
 struct symex_level0t
 {
+  /// Rename \p ssa_expr using \p thread_nr as L0 tag, unless \p ssa_expr is
+  /// a guard, a shared variable or a function. \p ns is queried to decide
+  /// whether we are in one of this cases.
```
peterschrammel: typo `these`

src/goto-symex/goto_symex_state.cpp
```diff
@@ -798,23 +798,13 @@ ssa_exprt goto_symex_statet::add_object(
 
   ssa_exprt ssa = rename_ssa<L0>(ssa_exprt{expr}, ns).get();
   const irep_idt l0_name = ssa.get_identifier();
-  const std::size_t l1_index = index_generator(l0_name);
+  const auto l1_index = narrow_cast<unsigned>(index_generator(l0_name));
```
peterschrammel, line 802: Why `unsigned` and not `size_t`?

src/goto-symex/renaming_level.cpp
```diff
@@ -56,6 +56,28 @@ operator()(ssa_exprt ssa_expr, const namespacet &ns, unsigned thread_nr) const
   return renamedt<ssa_exprt, L0>{ssa_expr};
 }
 
+void symex_level1t::insert(ssa_exprt ssa, unsigned index)
+{
+  const irep_idt identifier = ssa.get_identifier();
```
peterschrammel: `const &` ? (several occurrences)

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9992388747163a707bfd6a5a3afd501eed8ef871).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114705446

src/goto-symex/goto_symex_state.cpp
romainbrenguier: `symex_renaming_levelt` uses `unsigned` for indexes

src/goto-symex/renaming_level.cpp
romainbrenguier: the problem is `ssa` is moved below, and the move will happen before we use the identifier, so we need a copy

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 530c35b8e45c72217225c3fc7b78747355c8698f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114711776

PR #4758, opened 2019-06-06
Add CaDiCaL support to CMake

tautschnig:
Add CMake configuration to build using CaDiCaL, which was already
supported with Makefiles. Also bump the version of CaDiCaL to sr19 and
fix the use of the cnf_solvert interface.


tautschnig:
FWIW: running the regression suite with CaDiCaL appears to be quite a bit slower, but an evaluation on more realistic benchmarks is yet to be done.

src/solvers/CMakeLists.txt
```diff
@@ -98,6 +98,36 @@ elseif("${sat_impl}" STREQUAL "glucose")
     target_sources(solvers PRIVATE ${glucose_source})
 
     target_link_libraries(solvers glucose-condensed)
+elseif("${sat_impl}" STREQUAL "cadical")
+    message(STATUS "Building solvers with cadical")
+
+    download_project(PROJ cadical
+        URL https://github.com/arminbiere/cadical/archive/sr19.tar.gz
+        PATCH_COMMAND true
+        COMMAND CXX=${CXX_COMPILER} CXXFLAGS=-O3 ./configure --debug
```
smowton: Double-check: is the `--debug` asking `configure` for verbose output, or setting `-g` for the build, or could it even be overriding the `-O3` given in `CXXFLAGS` to get `-g -O0`?

smowton (approved):
Build script stuff looks fine to me, no opinion on the solvers stuff that I'm not that familiar with

src/solvers/CMakeLists.txt
tautschnig: I did double-check when building locally: it just adds `-g`, the `-O3` from `CXXFLAGS` is passed on as desired.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7931e35fe6f23fd32dd6227231d3df26c7089f28).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/114606145

kroening (approved):

src/solvers/sat/satcheck_cadical.cpp
```diff
@@ -79,6 +79,21 @@ propt::resultt satcheck_cadicalt::do_prop_solve()
   }
   else
   {
+    // if assumptions contains false, we need this to be UNSAT
```
thk123: :pick: I don't understand this change or why the assume falses need special casing, but it should probably go in its own commit explaining in more detail why this is required

thk123: Preferably with a test that only passes with this

thk123 (commented):
I don't have any strong opinions on this, but it should probably build `cadical` on CI and run the tests?

src/solvers/sat/satcheck_cadical.cpp
```diff
@@ -120,13 +135,21 @@ satcheck_cadicalt::~satcheck_cadicalt()
...
 void satcheck_cadicalt::set_assumptions(const bvt &bv)
 {
-  INVARIANT(false, "method not supported");
+  assumptions = bv;
+
+  for(const auto &a : assumptions)
+  {
+    if(a.is_true())
+    {
+      assumptions.clear();
```
peterschrammel: :no_entry_sign: I don't understand this. Assumptions are connected conjunctively.

tautschnig: That was me copy&pasting from the `satcheck_minisat2.cpp` ... probably worth reviewing.

peterschrammel: See https://github.com/diffblue/cbmc/pull/4810

src/solvers/sat/satcheck_cadical.cpp
tautschnig: Just for the record: this is done in keeping with other solvers, and thus about the way in which we use the solver interface. I agree that a test would be useful.

kroening: BTW, the reason why 'assume false' needs special casing is that SAT solvers don't have a 'false' literal.

martin-cs (approved):
Seems good to me.

peterschrammel (approved):
I'd suggest to merge also https://github.com/diffblue/cbmc/pull/5305 so that we can use the other solvers too without having to implement hardness_collectort everywhere.

src/solvers/CMakeLists.txt
```diff
@@ -98,6 +98,36 @@ elseif("${sat_impl}" STREQUAL "glucose")
...
+
+    target_compile_definitions(solvers PUBLIC
+        SATCHECK_CADICAL HAVE_CADICAL
+    )
+
+    add_library(cadical STATIC IMPORTED)
+
+    set_target_properties(
+        cadical
+        PROPERTIES IMPORTED_LOCATION ${cadical_SOURCE_DIR}/build/libcadical.a
```
hannes-steffenhagen-diffblue, line 122: This will only work on Linux, is that OK? If cadical also works on Windows/OSX it'd be nice if just used something like `find_library` here to get a more portable library path.

src/solvers/sat/satcheck_cadical.cpp
```diff
@@ -23,7 +23,7 @@ tvt satcheck_cadicalt::l_get(literalt a) const
 
   tvt result;
 
-  if(a.var_no() > static_cast<unsigned>(solver->max()))
+  if(a.var_no() > static_cast<unsigned>(solver->vars()))
```
hannes-steffenhagen-diffblue: Preferably use `narrow` from `util/narrow.h` here

src/solvers/sat/satcheck_cadical.cpp
```diff
@@ -72,31 +89,37 @@ propt::resultt satcheck_cadicalt::do_prop_solve()
...
-          "solving inside CaDiCaL SAT solver has been interrupted");
+      log.status() << "got FALSE as assumption: instance is UNSATISFIABLE"
+                   << messaget::eom;
+      status = statust::UNSAT;
+      return resultt::P_UNSATISFIABLE;
     }
   }
 
+  for(const auto &a : assumptions)
+    solver->assume(a.dimacs());
```
hannes-steffenhagen-diffblue, line 106: Why do we need to do this now? This wasn't here previously?

hannes-steffenhagen-diffblue (approved):
@tautschnig I'm noticing that the second commit has quite a few related changes at once. Is there any chance these could be split up into indivial commits that still make sense on their own? No real issue with the changes themselves.

src/solvers/sat/satcheck_cadical.cpp
tautschnig: Support for assumptions simply was missing before.

src/solvers/sat/satcheck_cadical.cpp
tautschnig: Will do.

tautschnig:
> @tautschnig I'm noticing that the second commit has quite a few related changes at once. Is there any chance these could be split up into indivial commits that still make sense on their own? No real issue with the changes themselves.

I'm not entirely convinced this adds value: the previous implementation was pretty broken, and not really usable. All these changes are necessary to make things work.

src/solvers/CMakeLists.txt
tautschnig: Cadical's makefile hard-codes this, so I don't think it'll work out of the box on other platforms.

PR #4765, opened 2019-06-07
Avoid guard to expr conversions

romainbrenguier:
For complicated programs, guards can get big and when represented as BDDs the conversion to exprt can be expensive. The goal of this pull request is to avoid these conversions, in particular in the solver we can convert directly from BDDs to literals without having to go through exprt first.


src/goto-symex/symex_target_equation.cpp
```diff
@@ -590,8 +590,6 @@ void symex_target_equationt::convert_io(
 /// \param SSA_step The step you want to have shared values.
 void symex_target_equationt::merge_ireps(SSA_stept &SSA_step)
 {
-  merge_irep(SSA_step.guard);
```
owen-mc-diffblue, line 590: This should be in the next commit, I think

src/solvers/prop/bdd_expr.h
```diff
@@ -36,6 +37,110 @@ class bdd_exprt
...
   exprt as_expr(const bddt &root) const;
 
+  template <typename returnt>
+  struct strategyt
+  {
+    std::function<returnt(const exprt &)> leaf;
+    std::function<returnt(const returnt &, const returnt &)> make_or;
+    std::function<returnt(const returnt &, const returnt &)> make_and;
+    std::function<returnt(const returnt &)> make_not;
+    std::function<returnt(const returnt &, const returnt &, const returnt)> make_ite;
```
owen-mc-diffblue: I don't understand `ite`. After a minute I guessed "if then else". Perhaps make it clearer somehow.

src/solvers/prop/prop_conv_solver.cpp
```diff
@@ -584,3 +585,21 @@ std::string prop_conv_solvert::decision_procedure_text() const
...
+  };
+  strategy.make_or = [&](const literalt &l1, const literalt &l2) {
+    return prop.lor(l1, l2);
+  };
+  strategy.make_ite =
+    [&](const literalt &i, const literalt &t, const literalt &e) {
+      return prop.lor(prop.land(i, t), prop.land(!i, e));
+  };
+  return guard.apply(strategy);
+}
```
owen-mc-diffblue: new line missing

owen-mc-diffblue (commented):
I got up to " Use version of assignment with guardt argument "

kroening:
This PR may set a new record in terms of the number of labels.

src/goto-symex/symex_assign.cpp
```diff
@@ -486,8 +486,9 @@ void goto_symext::symex_assign_symbol(
               << messaget::eom;
     });
 
-  const exprt assignment_guard =
-    make_and(state.guard.as_expr(), conjunction(guard));
+  // Temporarily add the state guard
+  guardt assignment_guard = state.guard;
+  assign_guard.add(conjunction(guard));
```
owen-mc-diffblue: `assign_guard` -> `assignment_guard`?

src/goto-symex/symex_atomic_section.cpp
```diff
@@ -57,11 +57,9 @@ void goto_symext::symex_atomic_end(statet &state)
         it != pair.second.second.end();
         ++it)
       read_guard|=*it;
-    exprt read_guard_expr=read_guard.as_expr();
-    do_simplify(read_guard_expr);
```
owen-mc-diffblue, line 61: Is it a functional change that we no longer simplify `read_guard` here?

src/goto-symex/symex_atomic_section.cpp
```diff
@@ -77,11 +77,9 @@ void goto_symext::symex_atomic_end(statet &state)
         it != pair.second.end();
         ++it)
       write_guard|=*it;
-    exprt write_guard_expr=write_guard.as_expr();
-    do_simplify(write_guard_expr);
```
owen-mc-diffblue, line 83: Is it a functional change that we no longer simplify `write_guard` here?

owen-mc-diffblue (dismissed):

src/goto-symex/symex_target_equation.cpp
```diff
@@ -29,7 +29,7 @@ void symex_target_equationt::shared_read(
   SSA_steps.emplace_back(source, goto_trace_stept::typet::SHARED_READ, guard_manager);
   SSA_stept &SSA_step=SSA_steps.back();
 
-  SSA_step.guard=guard;
+  SSA_step.guard.make_true().add(guard);
```
smowton: Given all these uses of `make_true().add(...)`, perhaps add a utility method `set` or `replace`?

src/solvers/prop/bdd_expr.h
```diff
@@ -36,6 +37,110 @@ class bdd_exprt
...
+      INVARIANT(index < bdd_expr.node_map.size(), "Index should be in node_map");
+      const exprt &n_expr = bdd_expr.node_map[index];
+      returnt node_result = leaf(n_expr);
+
+      // Look-up cache for already computed value
+      // (false_exprt used as temporary value)
+      auto insert_result = cache.emplace(r.id(), leaf(false_exprt()));
+      if(insert_result.second)
+      {
+        insert_result.first->second = [&]() -> returnt {
```
smowton: Suggest this big function deserves a name, rather than being anonymous

src/solvers/prop/bdd_expr.h
```diff
@@ -36,6 +37,110 @@ class bdd_exprt
   bddt from_expr(const exprt &expr);
   exprt as_expr(const bddt &root) const;
 
+  template <typename returnt>
+  struct strategyt
+  {
+    std::function<returnt(const exprt &)> leaf;
```
smowton: Why `std::function` everywhere, instead of virtual functions?

src/goto-symex/symex_target_equation.cpp
```diff
@@ -381,7 +381,7 @@ void symex_target_equationt::convert_guards(
         mstream << messaget::eom;
```
smowton, line 428: Commit message: convertion -> conversion

src/goto-symex/symex_target.h
```diff
@@ -12,6 +12,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #ifndef CPROVER_GOTO_SYMEX_SYMEX_TARGET_H
 #define CPROVER_GOTO_SYMEX_SYMEX_TARGET_H
```
smowton, line 13: Commit message: conveting -> converting

src/goto-symex/symex_atomic_section.cpp
smowton: I suspect we ought to have a simplify(guardt &) method, which simplifies when `guardt` == `exprt` but is a no-op on a BDD, which is always inherently in simplified form

smowton:
Oh damn, one of the labels was "do not review". Oh well.

Re: `strategyt`, I suspect actually given how it's used we should simply use normal non-virtual functions, i.e. something like

```
template<resultt> struct x {
  resultt f();
  resultt g();
}

struct y : public x<int> {
}

int y::f() { }
int y::g() { }
```

Then any function that uses it (just `apply` I think) is templated?

smowton:
As it stands we get a weird mix of templates and dynamic dispatch, which I don't know to be wrong but feels... novel.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: db7ae45732de31ff559a9c57b4535a606959625c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/124790348

PR #4770, opened 2019-06-10
[TG-5572] Add option to restrict numerical input values

majakusber:
Adds command line option `--java-assume-input-limits [int:int]`. Does not work for arrays yet.


jbmc/regression/jbmc/assume-input-limits/My.java
```diff
@@ -0,0 +1,131 @@
...
+  byte byteField;
+  short shortField;
+  int intField;
+  long longField;
+  float floatField;
+  double doubleField;
+
+  int[] intArrField;
+
+  public static void numericalArg(byte byteArg, short shortArg, int intArg, long longArg, float floatArg, double doubleArg) {
```
romainbrenguier: this line is too long

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -1011,6 +1011,28 @@ void java_object_factoryt::gen_nondet_init(
     assign.add_source_location() = location;
 
     assignments.add(assign);
+    // add assumes to obey numerical limits (given via java-assume-input-limits)
+    if(type.id() != ID_c_bool && type.id() != ID_char)
```
romainbrenguier: :pick: I would prefer `type != java_bool_type() && type != java_char_type()`

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -1011,6 +1011,28 @@ void java_object_factoryt::gen_nondet_init(
...
+    {
+      if(object_factory_parameters.assume_input_limits.lower.has_value())
```
romainbrenguier: :pick: maybe do `if(const auto &lower = ...)` so that you can use `lower.value()` below instead of the long expression

src/util/object_factory_parameters.cpp
```diff
@@ -46,6 +47,24 @@ void object_factory_parameterst::set(const optionst &options)
...
       options.get_unsigned_int_option("min-nondet-string-length");
   }
+  if(options.is_set("java-assume-input-limits"))
+  {
+    const auto &limits = options.get_option("java-assume-input-limits");
+    const auto &colon = limits.find(':');
+    DATA_INVARIANT(
+      colon != std::string::npos && limits.front() == '[' &&
+        limits.back() == ']',
+      "java-assume-input-limits value should be pf the form [int:int]");
```
romainbrenguier: typo "pf" -> of

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 01f3e63a3673746ccb08b09c0cec50d3ecba47db).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115054010

src/util/object_factory_parameters.cpp
```diff
@@ -46,6 +47,24 @@ void object_factory_parameterst::set(const optionst &options)
...
+    assume_input_limits = [&]() -> numeric_input_limitst {
+      numeric_input_limitst temp;
+      if(colon > 1)
+        temp.lower = unsafe_string2int(limits.substr(1, colon - 1));
+      if(colon < limits.size() - 2)
+        temp.upper = unsafe_string2int(
+          limits.substr(colon + 1, limits.size() - colon - 2));
```
romainbrenguier: I think using a regex `"\\[(\\d*):(\\d*)\\]"` would be a bit clearer and avoid having to think about indexes

src/util/object_factory_parameters.h
```diff
@@ -70,6 +71,14 @@ struct object_factory_parameterst
   /// Force string content to be ASCII printable characters when set to true.
   bool string_printable = false;
 
+  /// Force numerical primitive inputs to fall within the bounds
+  struct numeric_input_limitst
+  {
+    optionalt<int> lower;
+    optionalt<int> upper;
+  };
+  numeric_input_limitst assume_input_limits;
```
romainbrenguier: I'm not sure these two need to be grouped in a struct type, since the type itself doesn't seem to be used outside of this field.

jbmc/src/java_bytecode/java_object_factory.cpp
thk123: This behaviour should be documented in the command line options help

src/util/object_factory_parameters.cpp
```diff
@@ -46,6 +47,24 @@ void object_factory_parameterst::set(const optionst &options)
     min_nondet_string_length =
       options.get_unsigned_int_option("min-nondet-string-length");
   }
+  if(options.is_set("java-assume-input-limits"))
+  {
+    const auto &limits = options.get_option("java-assume-input-limits");
+    const auto &colon = limits.find(':');
+    DATA_INVARIANT(
```
thk123: :no_entry_sign: User errors should throw an exception rather than an invariant. Specifically `user_input_error_exceptiont`

src/util/object_factory_parameters.h
thk123: I disagree - they are logically one thing (a range). However, does it make sense to have a lower but not an upper or vice versa? I'd probably make the whole thing optional if you only support having both or neither set

jbmc/regression/jbmc/assume-input-limits/My.java
```diff
@@ -0,0 +1,131 @@
...
+    if (intArg > 3) {
```
thk123: :pick: put the ifs in the same order as the parameters

jbmc/regression/jbmc/assume-input-limits/My.java
```diff
@@ -0,0 +1,131 @@
...
+    } else if (doubleField > 3) {
+      assert false;
+    } else if (doubleField < -1) {
+      assert false;
+    } else {
+      assert false;
+    }
+  }
+
+  public void classArrArg(Other other, int [] intArr) {
```
thk123: :pick: Ideally this would go in the other (Array) commit to ease reviewing

jbmc/regression/jbmc/assume-input-limits/arg_assume_no_zero.desc
```diff
@@ -0,0 +1,23 @@
+CORE
+My.class
+--function My.numericalArg --java-assume-input-limits [1:3]
```
thk123: :question: Not clear on what this test checks? There is no assert that can be reached by only using 0?

thk123 (commented):
:no_entry_sign: Please add some tests for invalid format of parameter and check the exit code is `1` :white_check_mark:

src/util/object_factory_parameters.h
majakusber: I also find it more understandable if they are grouped. 

Originally, I had the whole thing as optional but I think supporting only one limit could be quite useful (allows to for example force using only positive values). Alternatively, at one point I had default values for the limits - min and max long. However, asking whether the optional has a value is I think nicer than comparing the value to max long. I also thought of making the whole thing as well as the individual limits optional but that seems like an overkill :slightly_smiling_face:

src/util/object_factory_parameters.h
```diff
@@ -70,6 +71,14 @@ struct object_factory_parameterst
   /// Force string content to be ASCII printable characters when set to true.
   bool string_printable = false;
 
+  /// Force numerical primitive inputs to fall within the bounds
+  struct numeric_input_limitst
+  {
+    optionalt<int> lower;
+    optionalt<int> upper;
```
tautschnig: Is `int` really the right type? I'd use `mp_integer` here as the type of the system JBMC is being compiled on is unrelated to the type that you want to talk about.

jbmc/src/java_bytecode/java_object_factory.cpp
majakusber: Done

src/util/object_factory_parameters.cpp
majakusber: Corrected

src/util/object_factory_parameters.cpp
majakusber: Done, I used `invalid_command_line_argument_exceptiont`

jbmc/src/java_bytecode/java_object_factory.cpp
majakusber: Done and done

src/util/object_factory_parameters.cpp
majakusber: Done

jbmc/regression/jbmc/assume-input-limits/My.java
majakusber: Done

jbmc/regression/jbmc/assume-input-limits/My.java
majakusber: Shortened

jbmc/regression/jbmc/assume-input-limits/My.java
majakusber: Done

jbmc/regression/jbmc/assume-input-limits/arg_assume_no_zero.desc
majakusber: I changed the tests a bit - all desc files check for positive limits only `[1:3]` and there is a specific test fro checking that negative limit works too `arg_assume_neg`

src/util/object_factory_parameters.h
majakusber: Good point, done!

majakusber:
@romainbrenguier @thk123 @tautschnig All comments addressed, please have another look. I left most of the changes in the `Fixups`/`Fixup` commits but will squash those before merging.

General question regarding the option name: would you prefer `--java-assume-input-limits` or `--java-assume-input-bounds`?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 56a84cfd4a5cd859723af9167bf1e36eb0ddc40c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115267886

src/util/object_factory_parameters.cpp
```diff
@@ -46,6 +50,37 @@ void object_factory_parameterst::set(const optionst &options)
     min_nondet_string_length =
       options.get_unsigned_int_option("min-nondet-string-length");
   }
+  if(options.is_set("java-assume-input-limits"))
+  {
+    const auto &limits = options.get_option("java-assume-input-limits");
+    const std::regex limits_regex("\\[(-\\d+|\\d*):(-\\d+|\\d*)\\]");
```
romainbrenguier, line 56: :+1:

src/util/object_factory_parameters.h
```diff
@@ -70,6 +72,14 @@ struct object_factory_parameterst
   /// Force string content to be ASCII printable characters when set to true.
   bool string_printable = false;
 
+  /// Force numerical primitive inputs to fall within the bounds
```
romainbrenguier: The documentation should also tell what the type represents not how it will be used.

src/util/object_factory_parameters.h
romainbrenguier: In that case "range" or "interval" should probably appear in the name. 
Rather that defining a new type you should reuse existing ones, here `integer_intervalt` would be the right one. It will avoid having to define a new type and you can make your code simpler by using `.empty()`.

romainbrenguier (approved):

src/util/object_factory_parameters.h
majakusber: Oh yes! That looks like exactly what I need, thank you :+1:

majakusber:
@tautschnig @thk123 Ready for re-review, as @romainbrenguier suggested I used `integer_intervalt` instead of my own struct, I renamed the option to match this and also to match the related `java-assume-inputs-non-null` option, and I added tests for static fields.

src/util/object_factory_parameters.h
```diff
@@ -14,6 +14,9 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <list>
 
 #include <util/irep.h>
+#include <util/optional.h>
+#include <analyses/interval_domain.h>
```
romainbrenguier: I think we shouldn't have dependencies of util on analyses, so the best solution is probably to move `interval_template.h` to  `util` (and we can drop "template" from the name)

jbmc/src/java_bytecode/java_bytecode_language.h
```diff
@@ -56,6 +57,11 @@ Author: Daniel Kroening, kroening@kroening.com
   " --java-assume-inputs-non-null\n" \
   "                              never initialize reference-typed parameter to the\n" /* NOLINT(*) */ \
   "                              entry point with null\n" /* NOLINT(*) */ \
+  " --java-assume-inputs-interval [L:U] or [L:] or [:U]\n" \
+  "                              never initialize numerical primitive-typed parameter\n" /* NOLINT(*) */ \
+  "                              (byte, short, int, long, float, double) to an entry\n" /* NOLINT(*) */ \
```
thk123, line 62: :bulb: I'd phrase this as the positive "only initialize primitive typed parameters to values within the specified range"

thk123 (approved):
lgtm

src/util/object_factory_parameters.h
majakusber: Yep, moved... with a bunch of other files it needed...

tautschnig (changes requested):
I'm afraid the move of files is a bit too far-reaching. It seems that all that you need is `integer_intervalt`, so I guess that should just be placed in a file of its own. The `ai*.h` files really shouldn't be moved.

unit/analyses/ai/ai.cpp
```diff
@@ -12,7 +12,7 @@ Author: Diffblue Ltd.
 #include <testing-utils/message.h>
 #include <testing-utils/use_catch.h>
 
-#include <analyses/ai.h>
+#include <util/ai.h>
```
romainbrenguier: the header file and cpp file shouldn't be in separate modules and I think ai should stay in analyses

jbmc/src/java_bytecode/java_bytecode_language.h
majakusber: Rephrased

majakusber:
I minimized the move :crossed_fingers:

src/util/integer_interval.h
```diff
@@ -0,0 +1,18 @@
...
+\*******************************************************************/
+
+
+#ifndef CPROVER_UTIL_INTEGER_INTERVAL_H
+#define CPROVER_UTIL_INTEGER_INTERVAL_H
+
+#include "interval_template.h"
+#include "mp_arith.h"
+
+typedef interval_templatet<mp_integer> integer_intervalt;
```
romainbrenguier: :pick: `using` is preferred

romainbrenguier (approved):

src/util/integer_interval.h
majakusber: Done

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: f303e99ca7afca324b27498bc6548902db99d4fe).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115425794

majakusber:
@tautschnig Would you have time to give a full review? I need a code-owner :)

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -1011,6 +1011,27 @@ void java_object_factoryt::gen_nondet_init(
     assign.add_source_location() = location;
 
     assignments.add(assign);
+    // add assumes to obey numerical limits (given via
+    // java-assume-inputs-interval)
+    if(type != java_boolean_type() && type != java_char_type())
```
tautschnig, line 1016: Why is `java_char_type()` also special cased? It would seem that intervals still make sense for that type?

src/util/object_factory_parameters.cpp
```diff
@@ -46,6 +50,34 @@ void object_factory_parameterst::set(const optionst &options)
...
+    std::smatch base_match;
+    if(!std::regex_match(interval, base_match, limits_regex))
+    {
+      throw invalid_command_line_argument_exceptiont(
+        "interval must be of the form [int:int], [int:] or [:int]",
+        "--java-assume-inputs-interval");
+    }
+    assume_inputs_interval = [&]() -> integer_intervalt {
+      integer_intervalt temp;
+      if(base_match[1] != "")
```
tautschnig, line 66: Why not `!base_match[1].empty()`

src/util/object_factory_parameters.cpp
```diff
@@ -46,6 +50,34 @@ void object_factory_parameterst::set(const optionst &options)
...
+        temp.make_ge_than(string2integer(base_match[1]));
+      if(base_match[2] != "")
```
tautschnig, line 68: Why not `!base_match[2].empty()`

tautschnig (approved):

jbmc/src/java_bytecode/java_object_factory.cpp
romainbrenguier: it makes sense but we would probably want a different one, for instance if we use [0:10] for integers it would only give non-printable characters

majakusber: I think that wouldn't be very intuitive for the user (definitely not for me :slightly_smiling_face: ). We have `--string-input-value` option which accepts a list of strings to use during test generation (it's not exactly the same though as what you meant).

src/util/object_factory_parameters.cpp
majakusber: The type doesn't have `empty()` function. Note that because the regex allows empty string here, this would always be a match. I just need to check that what matched was an empty string (no bound).

jbmc/src/java_bytecode/java_object_factory.cpp
thk123: A little forward visibility - a forthcoming feature will bound `java_char_type` when `--string-printable` is used (using the same range of characters).

PR #4774, opened 2019-06-11
Unit-test try_evaluate_pointer_comparison and bugfix

romainbrenguier:
This adds some unit test for `try_evaluate_pointer_comparison`, then fix a small bug, and add an unit-test for the case that was buggy, then add some checks for an implicit invariant that I noticed while unit testing, and make value_sett::output more usable in particular for debugging unit tests. 

### DIFF
--- a/src/goto-symex/goto_symex.h
+++ b/src/goto-symex/goto_symex.h
@@ -726,4 +726,20 @@ void symex_transition(
   goto_programt::const_targett to,
   bool is_backwards_goto);
 
+/// Try to evaluate pointer comparisons where they can be trivially determined
+/// using the value-set. This is optional as all it does is allow symex to
+/// resolve some comparisons itself and therefore create a simpler formula for
+/// the SAT solver.
+/// \param [in,out] condition: An L2-renamed expression with boolean type
+/// \param value_set: The value-set for determining what pointer-typed symbols
+///   might possibly point to
+/// \param language_mode: The language mode
+/// \param ns: A namespace
+/// \return The possibly modified condition
+renamedt<exprt, L2> try_evaluate_pointer_comparisons(
+  renamedt<exprt, L2> condition,
+  const value_sett &value_set,
+  const irep_idt &language_mode,
+  const namespacet &ns);
+
 #endif // CPROVER_GOTO_SYMEX_GOTO_SYMEX_H
--- a/src/goto-symex/symex_goto.cpp
+++ b/src/goto-symex/symex_goto.cpp
@@ -196,17 +196,7 @@ static optionalt<renamedt<exprt, L2>> try_evaluate_pointer_comparison(
     expr.id(), *symbol_expr_lhs, rhs, value_set, language_mode, ns);
 }
 
-/// Try to evaluate pointer comparisons where they can be trivially determined
-/// using the value-set. This is optional as all it does is allow symex to
-/// resolve some comparisons itself and therefore create a simpler formula for
-/// the SAT solver.
-/// \param [in,out] condition: An L2-renamed expression with boolean type
-/// \param value_set: The value-set for determining what pointer-typed symbols
-///   might possibly point to
-/// \param language_mode: The language mode
-/// \param ns: A namespace
-/// \return The possibly modified condition
-static renamedt<exprt, L2> try_evaluate_pointer_comparisons(
+renamedt<exprt, L2> try_evaluate_pointer_comparisons(
   renamedt<exprt, L2> condition,
   const value_sett &value_set,
   const irep_idt &language_mode,
--- a/src/goto-symex/symex_goto.cpp
+++ b/src/goto-symex/symex_goto.cpp
@@ -100,8 +100,10 @@ static optionalt<renamedt<exprt, L2>> try_evaluate_pointer_comparison(
   const ssa_exprt *ssa_symbol_expr =
     expr_try_dynamic_cast<ssa_exprt>(symbol_expr);
 
+  ssa_exprt l1_expr{*ssa_symbol_expr};
+  l1_expr.remove_level_2();
   const std::vector<exprt> value_set_elements =
-    value_set.get_value_set(ssa_symbol_expr->get_l1_object(), ns);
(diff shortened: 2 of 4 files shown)


src/pointer-analysis/value_set.cpp
```diff
@@ -1220,7 +1221,8 @@ void value_sett::assign(
   std::cout << "--------------------------------------------\n";
   output(ns, std::cout);
 #endif
-
+  PRECONDITION(is_l1_renamed(lhs));
```
smowton: This is only true when used in symex-- otherwise we expect to operate on plain un-renamed expressions. Therefore these invariants should move to `symex_dereference` or to some wrapper function in `goto_statet`.

owen-mc-diffblue (approved):
Great unit tests. Some other edge cases you might want to consider testing: if the value-set contains `ID_unknown`, `ID_invalid` or a failed symbol then `try_evaluate_pointer_comparison` should not be able to determine anything.

src/pointer-analysis/value_set.cpp
romainbrenguier: I defined a wrapper class for `value_sett` (a wrapper function wouldn't ensure we didn't continue to use the unchecked one)

romainbrenguier:
@owen-jones-diffblue I added some cases in the unit test, in particular some with unknown in the value set.

owen-mc-diffblue:
Btw failed symbols are covered by the regression test that was merged in https://github.com/diffblue/cbmc/pull/4659, I believe.

src/goto-symex/symex_value_set.h
```diff
@@ -0,0 +1,74 @@
...
+
+#ifndef CPROVER_GOTO_SYMEX_SYMEX_VALUE_SET_H
+#define CPROVER_GOTO_SYMEX_SYMEX_VALUE_SET_H
+
+#include <pointer-analysis/value_set.h>
+#include "renamed.h"
+
+/// Wrapper for value_sett which ensures we access it in a consistent way.
+/// In particular level 1 names should be used.
+class symex_value_sett : value_sett
```
smowton: Generally prefer composition to inheritence

src/goto-symex/goto_symex_state.cpp
```diff
@@ -222,13 +138,13 @@ renamedt<ssa_exprt, L2> goto_symex_statet::assignment(
     ssa_exprt l1_lhs(lhs);
     l1_lhs.remove_level_2();
 
-    if(run_validation_checks)
```
smowton: These should still be guarded by `if(run_validation_checks)`

src/goto-symex/renamed.cpp
```diff
@@ -14,9 +14,6 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 #include <util/std_expr.h>
 bool is_l1_renamed(const exprt &expr)
 {
-  if(!is_l2_renamed(expr.type()))
-    return false;
```
tautschnig: The claim made in the commit message seems wrong: the type may well be renamed to L2 even when the object itself is only L1. On the contrary, it actually makes no sense for the type to be renamed to L1 ever.

src/pointer-analysis/value_set.cpp
```diff
@@ -22,6 +22,8 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <langapi/language_util.h>
 #include <util/range.h>
+#include <util/format_expr.h>
+#include <util/format_type.h>
```
tautschnig: But then `#include <langapi/language_util.h>` better be removed as well (and hopefully also the corresponding entry in `module_dependencies.txt`).

src/pointer-analysis/value_set.cpp
```diff
@@ -20,10 +20,11 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/prefix.h>
 #include <util/simplify_expr.h>
 
+#include <goto-symex/renamed.h>
```
tautschnig: I don't think a dependency from pointer-analysis towards goto-symex is right.

tautschnig (changes requested):
This PR appears to be doing at least 3 different things (and some of them I believe are wrong, see detailed comments).

src/goto-symex/renamed.cpp
romainbrenguier: Ok I didn't know that the type could be L2 while the object L1. But is it true that L1 object always have L2 types? It sounds wrong to me. It seems that when an `ssa_exprt` is renamed to L1, its type is also renamed to L1.

tautschnig: Renaming to L1 effectively is instantiation/determines which instance of a named object we are using. At that point the type must be complete, i.e., must be L2.

romainbrenguier:
@tautschnig 
> This PR appears to be doing at least 3 different things (and some of them I believe are wrong, see detailed comments).

Yes you are right, I extracted what comes after the unit tests and bugfix into a new pull request
https://github.com/diffblue/cbmc/pull/4778 so that we can merge the fix more quickly and leave time for discussion on the other part.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: e26f4cd523f751cb6c3f5f6614343a28f0a8d983).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115206278

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 767d8f09ff7d519fc64c2f11b5e912050d5d3e40).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115226966

src/goto-symex/renamed.cpp
romainbrenguier: It seems to be the case in symex_decl but not in symex_dereference and the other places where L1 renaming is used. Should we make `rename<L1>` apply `L2` renaming on the type to ensure this is consistent?

tautschnig (approved):

PR #4787, opened 2019-06-13
Synthesise lambda classes

smowton:
This scans invokedynamic instructions during class-loading, creating a synthetic type for each one that we can understand, and registers its constructor (which captures) and method (which passes captured and actual parameters to the real lambda function) as synthetic methods. These are then created on demand using the same mechanism as e.g. synthetic static initializer methods.

Needs tests, but the code is in reviewable state.


tautschnig:
@smowton For Java-clueless people like myself: can you explain _why_ this change is desirable to have?

smowton:
At the moment `invokedynamic` instructions are always stubbed to return null. With this change they actually work, so you can verify programs that use lambdas, such as `assert (MyInterface)(x -> x + 1)(1) == 2`

tautschnig:
@smowton Thanks! Would you mind adding that info to the commit message?

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -94,6 +94,35 @@ static void assign_parameter_names(
   }
 }
 
+void create_method_stub_symbol(
```
thk123, line 98: :pick: I think would be more natural if this returned the symbol, and elt the caller add to the symbol table

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2185,23 +2214,15 @@ void java_bytecode_convert_methodt::convert_invoke(
...
-    assign_parameter_names(
-      to_java_method_type(symbol.type), symbol.name, symbol_table);
-    set_declaring_class(symbol, arg0.get(ID_C_class));
-
-    debug() << "Generating codet:  new opaque symbol: method '" << symbol.name
-            << "'" << eom;
-    symbol_table.add(symbol);
+    create_method_stub_symbol(
+      invoked_method_id,
+      arg0.get(ID_C_base_name),
```
thk123, line 2260: :bulb: Lots of the parameters are derived from `ID_C_base_name`, perhaps just pass that in and let the method do this substring nonsense

thk123: Or possibly even better - pass in arg0

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -324,14 +324,15 @@ optionalt<symbolt> java_bytecode_convert_methodt::get_lambda_method_symbol(
...
+/// \param class_symbol: The class this method belongs to. The method, if not
+///   static, will be added to the class' list of methods.
 /// \param method_identifier: The fully qualified method name, including type
 ///   descriptor (e.g. "x.y.z.f:(I)")
 /// \param m: The parsed method object to convert.
 /// \param symbol_table: The global symbol table (will be modified).
 /// \param message_handler: A message handler to collect warnings.
 void java_bytecode_convert_method_lazy(
-  const symbolt &class_symbol,
+  symbolt &class_symbol,
```
thk123, line 308: :bulb: This should be unit tested

smowton:
@tautschnig done

smowton:
Added an expansion of the `jbmc/lambda1` test to actually check behaviour

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a94e7a9909ea856ff8d6cda127253685d9587c75).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115779638

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 87a435fc9fee0561c6b59e4e1e00ae5a49f4e144).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115786915

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 447de944b2eddaa85dc97b910118a18d49872b01).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115796756

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4b8837eb8435c5bd3b7f2f6fc5baa08d99c4ab75).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115930962

smowton:
Pinging @thk123 @peterschrammel for review

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2957,40 +2937,71 @@ code_blockt java_bytecode_convert_methodt::convert_astore(
...
+  code_function_callt::argumentst arguments = pop(parameters.size());
 
-  const java_method_typet::parameterst &parameters(method_type.parameters());
+  irep_idt synthetic_class_name =
+    lambda_synthetic_class_name(method_id, instruction_address);
 
-  pop(parameters.size());
+  if(!symbol_table.has_symbol(synthetic_class_name))
+  {
+    // We failed to parse the invokedynamic handle as a Java 8+ lambda;
```
peterschrammel, line 3001: Should we output a warning in this case?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -134,6 +134,18 @@ void create_invokedynamic_synthetic_classes(
                     << messaget::eom;
         continue;
       }
+      else if(implemented_interface_type.methods().size() != 1)
```
peterschrammel: (commit message) distiniguish -> distinguish

peterschrammel (approved):
LGTM

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
smowton: A debug message is already produced in `java_bytecode_parser.cpp` in this case. That's probably enough for now, as otherwise we would print a warning for every non-Java use of invokedynamic, which would be pretty user-unfriendly

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -719,6 +720,33 @@ bool java_bytecode_languaget::typecheck(
     }
   }
 
+  // Now add synthetic classes for every invokedynamic instruction found (it
+  // makes this easier that all interface types and their methods have been
+  // created above):
+  {
```
thk123, line 747: :bulb: Maybe make this a function rather than unlabelled scope?

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -719,6 +720,33 @@ bool java_bytecode_languaget::typecheck(
...
+    std::vector<irep_idt> function_symbols;
+
+    for(const auto &id_and_symbol : symbol_table)
```
thk123: :bulb: This looks like a natural fit for a filter

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -719,6 +720,33 @@ bool java_bytecode_languaget::typecheck(
...
+    {
+      if(id_and_symbol.second.type.id() != ID_code)
+        continue;
+      auto cmb = method_bytecode.get(id_and_symbol.first);
```
thk123: :no_entry_sign: I don't know what `cmb` stands - please de-acro-fy
:question: I don't know what `id_and_symbol.first` is likely to be, and I don't think direct irep access using a dynamic string is a good idea - can we do something a bit better here

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -719,6 +720,33 @@ bool java_bytecode_languaget::typecheck(
...
+      if(!cmb)
+        continue;
+      function_symbols.push_back(id_and_symbol.first);
+    }
+
+    for(const auto &id : function_symbols)
```
thk123: :bulb: `id` isn't an informative name, perhaps `method_id`

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -719,6 +720,33 @@ bool java_bytecode_languaget::typecheck(
...
 
+  // Now add synthetic classes for every invokedynamic instruction found (it
+  // makes this easier that all interface types and their methods have been
+  // created above):
+  {
+    std::vector<irep_idt> function_symbols;
+
+    for(const auto &id_and_symbol : symbol_table)
+    {
+      if(id_and_symbol.second.type.id() != ID_code)
```
thk123: :pick: `if(can_cast_type<code_typet>(id_and_symbol.second.type))`

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -1031,6 +1059,15 @@ static void notify_static_method_calls(
           expr_dynamic_cast<symbol_exprt>(fn_call->function());
         needed_lazy_methods->add_needed_method(fn_sym.get_identifier());
       }
+      else if(
```
thk123, line 1084: :question: What is this block about and how come it is appearing as part of this commit?

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -1031,6 +1059,15 @@ static void notify_static_method_calls(
...
+        it->id() == ID_side_effect &&
```
thk123, line 1085: :pick: `expr_try_dynamic_cast`

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -1136,6 +1173,14 @@ bool java_bytecode_languaget::convert_single_method(
           get_pointer_type_selector(),
           get_message_handler());
       break;
+    case synthetic_method_typet::INVOKEDYNAMIC_CAPTURE_CONSTRUCTOR:
```
thk123, line 1204: :question: What are these? can they do in a seperate commit with justification?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+  message_handlert &message_handler)
+{
+  messaget log(message_handler);
+  namespacet ns(symbol_table);
+
+  for(const auto &instruction : instructions)
+  {
+    if(!strcmp(bytecode_info[instruction.bytecode].mnemonic, "invokedynamic"))
+    {
+      // invokedynamic will be called with operands that should be stored in a
```
thk123: Seems like this big picture documentation should live in the doxygen for the function?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+      // }
+      // This method just creates the outline; the methods will be populated on
+      // demand via java_bytecode_languaget::convert_lazy_method.
+
+      // Check that we understand the lambda method handle; if we don't then
+      // we will not create a synthetic class at all, and the corresponding
+      // invoke instruction will return null when eventually converted by
+      // java_bytecode_convert_method.
+
+      const auto &method_symbol = ns.lookup(method_identifier);
```
thk123: You might find the immediately invoked lambda a useful pattern here. Or pull out a little static function to get the `class_type` from the `method_identifier` since all other variables are used only here

thk123: Actually even `class_type` isn't used outside of the next block

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+      const auto &implemented_interface_tag = to_struct_tag_type(
+        to_java_reference_type(dynamic_method_type.return_type()).subtype());
+      const symbolt &implemented_interface_symbol =
+        ns.lookup(implemented_interface_tag.get_identifier());
+      const java_class_typet &implemented_interface_type =
+        to_java_class_type(implemented_interface_symbol.type);
+
+      if(implemented_interface_type.get_is_stub())
+      {
+        log.debug() << "ignoring invokedynamic at " << method_identifier
```
thk123: :question: Seems like this should be a `log.warning()`?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+                  << " address " << instruction.address
+                  << " for lambda: " << lambda_method_symbol.value().name
+                  << messaget::eom;
+
+      // Create the class symbol:
+
+      irep_idt synthetic_class_name =
+        lambda_synthetic_class_name(method_identifier, instruction.address);
+
+      java_class_typet synthetic_class_type;
```
thk123: :bulb: Seems like this could be pulled out into a separate function too `create_synthetic_lambda_class` or something?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+      synthetic_class_type.set_synthetic(true);
+      synthetic_class_type.set(
+        ID_java_lambda_method_identifier, lambda_method_symbol.value().name);
+      struct_tag_typet base_tag("java::java.lang.Object");
+      synthetic_class_type.add_base(base_tag);
+      synthetic_class_type.add_base(implemented_interface_tag);
+
+      // Add the class fields:
+
+      java_class_typet::componentt base_field;
```
thk123: :question: There must be code that already creates this component somewhere else - can it be reused?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+        java_class_typet::componentt new_field;
+        new_field.set_name(field_basename);
+        new_field.set_base_name(field_basename);
+        new_field.set_pretty_name(field_basename);
+        new_field.set_access(ID_private);
+        new_field.type() = param.type();
+        synthetic_class_type.components().emplace_back(std::move(new_field));
+      }
+
+      // Create constructor symbol:
```
thk123: :bulb: It seems like you've already named this function!

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+        java_reference_type(struct_tag_typet(synthetic_class_name));
+
+      constructor_type.parameters().insert(
+        constructor_type.parameters().begin(), constructor_this_param);
+
+      constructor_symbol.type = constructor_type;
+      set_declaring_class(constructor_symbol, synthetic_class_name);
+      symbol_table.add(constructor_symbol);
+
+      // Create implemented method symbol:
```
thk123: :bulb: And this one!

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+      type_symbolt synthetic_class_symbol(synthetic_class_type);
+      synthetic_class_symbol.name = synthetic_class_name;
+      synthetic_class_symbol.mode = ID_java;
+      symbol_table.add(synthetic_class_symbol);
+    }
+  }
+}
+
+static const symbolt &get_or_create_method_symbol(
```
thk123, line 349: :question: How come this function is needed here specifically? Seems like we already have method that do this?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+    base_name,
+    pretty_name,
+    type,
+    declaring_class,
+    symbol_table,
+    log);
+  return symbol_table.lookup_ref(identifier);
+}
+
+codet get_invokedynamic_synthetic_constructor(
```
thk123: :no_entry_sign: Personally find `get` a little misleading in the name here, would `synthesize` be more accurate? Perhaps adding a doxygen summary would help

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+  const auto &parameters = to_code_type(function_symbol.type).parameters();
+
+  const symbolt &class_symbol = ns.lookup(*declaring_class(function_symbol));
+  const class_typet &class_type = to_class_type(class_symbol.type);
+
+  const symbol_exprt this_param(
+    parameters.at(0).get_identifier(), parameters.at(0).type());
+  const dereference_exprt deref_this(this_param);
+
+  // Call super-constructor (always java.lang.Object):
```
thk123, line 391: :bulb: seems like another function

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+    parameter_symbolt param_symbol;
+    param_symbol.name = parameter.get_identifier();
+    param_symbol.base_name = parameter.get_base_name();
+    param_symbol.mode = ID_java;
+    param_symbol.type = parameter.type();
+    symbol_table.add(param_symbol);
+
+    if(parameter.get_this())
+      continue;
+
```
thk123, line 427: :bulb: Probably worth throwing some invariants here that the members and the parameters all exist, their types are correct etc?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+    code_assignt assign_field(
+      member_exprt(deref_this, field_iterator->get_name(), parameter.type()),
+      symbol_exprt(parameter.get_identifier(), parameter.type()));
+    result.add(assign_field);
+
+    ++field_iterator;
+  }
+
+  return std::move(result);
```
thk123, line 436: :question: Why move? In this case is a local variable, won't RVO kick in anyway?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+}
+
+codet get_invokedynamic_synthetic_method(
```
thk123: Same comment - I think `get` is not a good term for this

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+  for(const auto &parameter : parameters)
+  {
+    // Give the parameter its symbol:
+    parameter_symbolt param_symbol;
+    param_symbol.name = parameter.get_identifier();
+    param_symbol.base_name = parameter.get_base_name();
+    param_symbol.mode = ID_java;
+    param_symbol.type = parameter.type();
+    symbol_table.add(param_symbol);
```
thk123, line 481: :question: Why is this happening so late, is that normal?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+    lambda_method_args.push_back(param_symbol.symbol_expr());
+  }
+
+  const auto &lambda_method_symbol =
+    ns.lookup(class_type.get(ID_java_lambda_method_identifier));
+
+  if(return_type != empty_typet())
+  {
+    result.add(code_returnt(side_effect_expr_function_callt(
```
thk123, line 494: :question: why is this a `side_effect_expr_function_callt` and not just a regular `code_function_callt`? Why do both methods even exist?

thk123: A: to avoid declaring, storing and then returning a value... meh

jbmc/src/java_bytecode/java_bytecode_language.cpp
thk123: These are added because these are new kinds of methods to convert.

jbmc/regression/jbmc/lambda1/BiFunction.java
```diff
@@ -0,0 +1,6 @@
+
+interface BiFunction<From1, From2, To> {
```
thk123, line 2: :question: Why? I guess these interfaces don't exist in core models, but might be easier to just add them?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -134,6 +134,18 @@ void create_invokedynamic_synthetic_classes(
                     << messaget::eom;
         continue;
       }
+      else if(implemented_interface_type.methods().size() != 1)
+      {
+        log.debug() << "ignoring invokedynamic at " << method_identifier
```
thk123: :pick: Seems like should definitely be a warning?

jbmc/regression/jbmc/lambda1/Lambdatest.java
```diff
@@ -76,17 +74,54 @@ public boolean custom(Integer i) {
...
-        lt.f(0.0f, 0, 0);
-        lt.i(0);
-        lt.j(0);
-        lt.k(0);
-        lt.l(0);
-        lt.m(0);
+  public static void main(String[] args, int unknown) {
+      Lambdatest lt = new Lambdatest();
+      if(unknown == 0)
+        assert lt.g(1.0f, 2, (x, y) -> x.intValue() + y) == 3; // should succeed
```
thk123: :pick: I know you didn't name these functions, but if we could maybe rename some of these so less impossible to know what is being tested here `g` -> `executeLambda`

jbmc/regression/jbmc/lambda1/Lambdatest.java
```diff
@@ -76,17 +74,54 @@ public boolean custom(Integer i) {
...
+      else if(unknown == 1)
+        assert lt.g(1.0f, 2, (x, y) -> x.intValue() + y) == 4; // should fail
+      else if(unknown == 2)
+        assert lt.f(1.0f, 2, 3) == 9; // should succeed
```
thk123: :pick: `f` -> `capture_local`

jbmc/regression/jbmc/lambda1/Lambdatest.java
```diff
@@ -76,17 +74,54 @@ public boolean custom(Integer i) {
...
+      else if(unknown == 3)
+        assert lt.f(1.0f, 2, 3) == 10; // should fail
+      else if(unknown == 4)
+        assert lt.i(2) == 10; // should succeed
```
thk123: `i` -> `capture_local_primitive`

jbmc/regression/jbmc/lambda1/Lambdatest.java
```diff
@@ -76,17 +74,54 @@ public boolean custom(Integer i) {
...
+      else if(unknown == 5)
+        assert lt.i(2) == 11; // should fail
+      else if(unknown == 6)
+        assert lt.j(3) == 30; // should succeed
```
thk123: `j` -> `capture_member`

jbmc/regression/jbmc/lambda1/Lambdatest.java
```diff
@@ -76,17 +74,54 @@ public boolean custom(Integer i) {
...
+      else if(unknown == 7)
+        assert lt.j(3) == 31; // should fail
+      else if(unknown == 8)
+        assert lt.k(4) == 40; // should succeed
```
thk123: `k` -> `capture_reference_member`

jbmc/regression/jbmc/lambda1/Lambdatest.java
```diff
@@ -76,17 +74,54 @@ public boolean custom(Integer i) {
...
+      else if(unknown == 9)
+        assert lt.k(4) == 41; // should fail
+      else if(unknown == 10)
+      {
+        assert lt.l(5) == 70; // should succeed
```
thk123: `l` -> `side_effect_lambda`

thk123 (dismissed):
:no_entry_sign: Missing tests for:
 - capturing a primitive array :heavy_check_mark: 
 - capturing a reference array :heavy_check_mark: 
 - capturing an object factory produced array/reference type (reference done, but can't see array?)
 - returning something captured (sort of tested - but might be good to verify the reference equality holds?)
 - assigning a lambda from an interface to an interface:
```java
Func<A, B> x = (a) -> return a;
Func<A, B> y = x;
y.Apply(1);
```
 - lambda that capture members of outer classes :heavy_check_mark: 
 - lamdbas that are initalized in `<clinit>` (missing but in fairness GH had rendered this bullet point incomprehensible)

The tests are also quite difficult to figure out what is tested due to the cryptic names (not yours but still)

There are also unit tests that should, with a bit of tweaking, now work inside `jbmc/unit/java_bytecode/java_bytecode_convert_method` - could these be turned on?

My biggest complaint on the code is there are huge chunks of straight line code that would be easier to digest if they were split up into smaller functions. You're half way there - you've broken them up with whitespace and stuck a comment that is essentially a function name above them.

jbmc/src/java_bytecode/java_bytecode_language.cpp
smowton: Separate commit sounds good, they're added because we now make side-effect-function-call expressions to avoid a pointless temporary (the expression is equivalent to a `code_function_callt` with a symbol on the LHS followed by immediately returning that symbol)

jbmc/src/java_bytecode/lambda_synthesis.cpp
smowton: Disagree, this would be a very noisy high-priority warning, suggest this should only be upgraded as/when stub classes stop being the norm

jbmc/src/java_bytecode/lambda_synthesis.cpp
smowton: `create_` then?

jbmc/src/java_bytecode/lambda_synthesis.cpp
smowton: Needed when you return `code_blockt` from a function whose type says it produces `codet`, as there will be an implicit copy- or move-constructor call to perform the conversion. Some Clang versions actually warn (or with `-Werror` give a fatal error) when you don't do this (or if the types are the same, if you *do*)

jbmc/src/java_bytecode/lambda_synthesis.cpp
smowton: Matches what we do with normal (non-synthetic) methods

jbmc/src/java_bytecode/lambda_synthesis.cpp
smowton: Yeah, one's an expression, the other a statement

jbmc/regression/jbmc/lambda1/BiFunction.java
smowton: That's up to them, I just wanted a free-standing test

jbmc/src/java_bytecode/lambda_synthesis.cpp
smowton: Default methods, Object methods, and stub types all produce false positives at the moment, so suggest this is still debug level until those issues are addressed.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
LAJW: @thk123 This directly manipulates the symbol table, so returning/constness is not very relevant. I don't see a way to make this substantially better 😞

smowton:
@lajw reviewing my own PR, realised the lambda1 test is needlessly changed from tab size 4 to 2. I'd fix it myself but I'm on the road until Sunday with only a phone.

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -66,8 +66,8 @@ void create_invokedynamic_synthetic_classes(
   synthetic_methods_mapt &synthetic_methods,
   message_handlert &message_handler)
 {
-  messaget log(message_handler);
-  namespacet ns(symbol_table);
+  messaget log{message_handler};
```
thk123: :pick: Putting this sort of changes in the same commit increases the burden for the reviewer, if possible put them in a separate commit in future

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -120,10 +123,11 @@ void create_invokedynamic_synthetic_classes(
 
       const auto &implemented_interface_tag = to_struct_tag_type(
         to_java_reference_type(dynamic_method_type.return_type()).subtype());
-      const symbolt &implemented_interface_symbol =
-        ns.lookup(implemented_interface_tag.get_identifier());
-      const java_class_typet &implemented_interface_type =
-        to_java_class_type(implemented_interface_symbol.type);
+      const java_class_typet &implemented_interface_type = [&] {
```
thk123: :question: I'll trust your answer - but is storing the result of function in a reference (albeit const ref) a good idea?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -118,6 +118,96 @@ static optionalt<irep_idt> interface_method_id(
...
+  constructor_this_param.type() =
+    java_reference_type(struct_tag_typet(synthetic_class_name));
+
+  constructor_type.parameters().insert(
+    constructor_type.parameters().begin(), constructor_this_param);
+
+  constructor_symbol.type = constructor_type;
+  set_declaring_class(constructor_symbol, synthetic_class_name);
+  return constructor_symbol;
+      }
```
thk123: :pick: This commit `Convert lambdas into functions` seems to need clang formatting?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -206,7 +206,56 @@ static symbolt create_constructor_symbol(
   constructor_symbol.type = constructor_type;
   set_declaring_class(constructor_symbol, synthetic_class_name);
   return constructor_symbol;
-      }
+}
+
+symbolt create_implemented_method_symbol(
```
thk123: :no_entry_sign: Please make the commit message more informative:
`Conver the last function ` -> `Pull out function create_implemented_method_symbol for invokedynamic conversion`

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -118,7 +118,7 @@ static optionalt<irep_idt> interface_method_id(
   return implemented_interface_type.methods().at(0).get_name();
 }
 
-const java_class_typet synthetic_class_type(
+symbolt synthetic_class_symbol(
```
thk123, line 130: :green_heart:

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -723,20 +723,11 @@ bool java_bytecode_languaget::typecheck(
...
-      if(id_and_symbol.second.type.id() != ID_code)
-        continue;
-      auto cmb = method_bytecode.get(id_and_symbol.first);
-      if(!cmb)
-        continue;
-      function_symbols.push_back(id_and_symbol.first);
-    }
-
-    for(const auto &id : function_symbols)
+    const auto &id = id_and_symbol.first;
```
thk123: :pick: `id` is better than `cmb`, but might be easier to just use `symbol.name` whether `id` is used, since then obvious is just the id of the symbol we're talking abotu

jbmc/src/java_bytecode/lambda_synthesis.cpp
thk123: :question: I still don't understand what this method is for or why is lambda specific

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -0,0 +1,499 @@
...
+    const auto lambda_method_name = ::lambda_method_name(
+      symbol_table, method_identifier, dynamic_method_type);
+    if(!lambda_method_name)
+    {
+      log.debug() << "ignoring invokedynamic at " << method_identifier
+                  << " address " << instruction.address
+                  << " with unknown handle type" << messaget::eom;
+      continue;
+    }
+    const auto &implemented_interface_tag = to_struct_tag_type(
```
thk123: :bulb: Name suggestion: `functional_interface_tag`

thk123 (commented):
Thanks for all the clean up Lukasz - this is looking much better. There are a couple of tests still missing:

- capturing an object factory produced array type (reference done, but not array?):

```java
int testFunction(Foo[] inputArr) {
  lambda = () -> return inputArr.length;
  lambda.execute();
}
```

- returning something captured (sort of tested - but might be good to verify the reference equality holds?):

```java
int testFunction() {
  Object reference = new Object();
  lambda = () -> return reference;
  Object o2 = lambda.execute();
  assert o2 == reference;
}
```

- assigning a lambda from an interface to an interface:
```java
Func<A, B> x = (a) -> return a;
Func<A, B> y = x;
y.Apply(1);
```

- lamdbas that are initalized in <clinit> (missing but in fairness GH had rendered this bullet point incomprehensible)

```java
static Func<A, A> x = (a) -> return a;
int testFunction() {
  A a = new A();
  A result = x.Apply(a);
  assert(a == result);
}
```

It would also be really good to tidy up the variable names in the existing test as I commented in the original review, but this can be done in a follow up PR. 

Finally could we have some rudimentary tests in TG with a bump pointing at this branch.

jbmc/src/java_bytecode/lambda_synthesis.cpp
LAJW: `const ref` extends the lifespan of the value on the right. The `&` is superfluous. This is the result of me trying to not replace all types with `const auto` 😁

LAJW:
Test-gen update with extra tests: https://github.com/diffblue/test-gen/pull/3697
Edit: Rebased

LAJW:
Rebased
Reformatted
Added tests:
 - nondet capture
 - statically initialized lambda
 - test for checking reference equality of captured object

Test not added:
 - Assignment of lambda from one variable to the next, because it's implicitly there.

LAJW:
Nondet array tests (reference and primitive) added.

jeannielynnmoulton (approved):
Verifying that the tests that @thk123 requested have been added since he is on holiday.

LAJW:
Rebasing on top of latest develop

tautschnig:
@smowton @LAJW Two of the Travis/cmake configurations are still failing.

smowton:
@LAJW @tautschnig just rebased on current develop which should work around the dodgy goto-harness test.

tautschnig:
@smowton CI still seems unhappy, though now in different ways.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: a316fd5be7b0baaa63da8857823041e313554984).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122445309

PR #4791, opened 2019-06-14
Add ranget::zip method

romainbrenguier:
This adds a zip method to ranges and add an example of usage.

### DIFF
--- a/src/util/range.h
+++ b/src/util/range.h
@@ -433,6 +433,14 @@ public:
       zip_begin, zip_end);
   }
 
+  template <typename containert>
+  auto zip(containert &container)
+    -> ranget<zip_iteratort<iteratort, decltype(container.begin())>>
+  {
+    return zip(
+      ranget<decltype(container.begin())>{container.begin(), container.end()});
+  }
+
   bool empty() const
   {
     return begin_value == end_value;
--- a/src/util/range.h
+++ b/src/util/range.h
@@ -265,6 +265,90 @@ private:
   second_iteratort second_begin;
 };
 
+/// Zip two ranges to make a range of pairs.
+/// On increment, both iterators are incremented.
+/// Ends when the two ranges reach their ends.
+/// Invariants are checking that one does not end before the other.
+template <typename first_iteratort, typename second_iteratort>
+struct zip_iteratort
+{
+public:
+  using difference_type = typename first_iteratort::difference_type;
+  using value_type = std::pair<
+    typename first_iteratort::value_type,
+    typename second_iteratort::value_type>;
+  using pointer = value_type *;
+  using reference = value_type &;
+  using iterator_category = std::forward_iterator_tag;
+
+  bool operator==(const zip_iteratort &other) const
+  {
+    return first_begin == other.first_begin && first_end == other.first_end &&
+           second_begin == other.second_begin && second_end == other.second_end;
+  }
+
+  bool operator!=(const zip_iteratort &other) const
+  {
+    return !(*this == other);
+  }
+
+  /// Preincrement operator
+  zip_iteratort &operator++()
+  {
+    PRECONDITION(first_begin != first_end && second_begin != second_end);
+    ++first_begin;
+    ++second_begin;
+    INVARIANT(
+      (first_begin == first_end) == (second_begin == second_end),
+      "Zipped ranges should have the same size");
+    current = first_begin != first_end
+                ? std::make_shared<value_type>(*first_begin, *second_begin)
+                : nullptr;
+    return *this;
+  }
+
+  /// Postincrement operator
+  const zip_iteratort operator++(int)
+  {
+    zip_iteratort tmp(first_begin, first_end, second_begin, second_end);
+    this->operator++();
+    return tmp;
+  }
+
+  reference operator*() const
+  {
+    PRECONDITION(current != nullptr);
+    return *current;
+  }
+
+  pointer operator->() const
+  {
+    return current.get();
+  }
+
+  zip_iteratort(
+    first_iteratort _first_begin,
+    first_iteratort _first_end,
+    second_iteratort _second_begin,
+    second_iteratort _second_end)
+    : first_begin(std::move(_first_begin)),
+      first_end(std::move(_first_end)),
+      second_begin(std::move(_second_begin)),
+      second_end(std::move(_second_end))
+  {
+    PRECONDITION((first_begin == first_end) == (second_begin == second_end));
+    if(first_begin != first_end)
+      current = util_make_unique<value_type>(*first_begin, *second_begin);
+  }
+
+private:
+  first_iteratort first_begin;
+  first_iteratort first_end;
+  second_iteratort second_begin;
+  second_iteratort second_end;
+  std::shared_ptr<value_type> current = nullptr;
+};
+
 /// A range is a pair of a begin and an end iterators.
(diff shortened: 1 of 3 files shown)


src/util/range.h
```diff
@@ -265,6 +265,88 @@ struct concat_iteratort
...
+    first_iteratort first_end,
+    second_iteratort second_begin,
+    second_iteratort second_end)
+    : first_begin(std::move(first_begin)),
+      first_end(std::move(first_end)),
+      second_begin(std::move(second_begin)),
+      second_end(std::move(second_end))
+  {
+    if(first_begin != first_end && second_begin != second_end)
+      current = util_make_unique<value_type>(*first_begin, *second_begin);
```
thomasspriggs, line 341: Because we know `current` is a `shared_ptr`, if my memory serves me correctly, then there is a marginal performance gain to be had by using std::make_shared instead of `util_make_unique`. Something to do with allocating a single contiguous block of memory instead of performing multiple separate allocations.

src/util/range.h
```diff
@@ -265,6 +265,88 @@ struct concat_iteratort
...
+  zip_iteratort &operator++()
+  {
+    ++first_begin;
+    ++second_begin;
+    INVARIANT(
+      (first_begin == first_end) == (second_begin == second_end),
+      "Zipped ranges should have the same size");
+    current = first_begin != first_end && second_begin != second_end
+                ? util_make_unique<value_type>(
+                    std::make_pair(*first_begin, *second_begin))
```
thomasspriggs: I think I would have been trying to define a `zip_with`, for which I then supplied `make_pair` or something similar as the function parameter in order to define `zip`. But I am happy for that to be a later refinement if we find a use for `zip_with`.

src/util/range.h
```diff
@@ -337,6 +419,18 @@ struct ranget final
       concat_begin, concat_end);
   }
 
+  template <typename other_iteratort>
+  ranget<zip_iteratort<iteratort, other_iteratort>>
+  zip(ranget<other_iteratort> other)
```
thomasspriggs, line 426: It might be nice to have an overload which applies `make_range` and calls the existing overload in the case where `other` is an STL container instead of a range. That way your example could be -
`make_range(components).zip(rhs.operands())` instead of `make_range(components).zip(make_range(rhs.operands()))`.

src/util/range.h
thomasspriggs: `std::make_pair` is handy, where we don't have the full type of the `std::pair` to hand, because it infers the type of the pair from the types of the arguments. I think that we have the full type of the `pair` specified here in `value_type`, which may mean that this call to `std::make_pair` is not needed.

thomasspriggs (commented):
How handy. I haven't spotted any problems. Therefore my comments can be considered to be refinements for consideration rather than anything which needs to be done before merging.

thomasspriggs (approved):

unit/util/range.cpp
```diff
@@ -188,6 +189,34 @@ SCENARIO("range tests", "[core][util][range]")
       REQUIRE(input2 == expected_result2);
     }
   }
+  GIVEN("A vectors of int and a list of strings.")
+  {
+    std::vector<int> int_vector{1, 2};
+    std::list<std::string> string_list{"foo", "bar"};
```
thomasspriggs, line 195: It would be good to have an equivalent test where `int_vector` and `string_list` are defined const.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7730ac3f5be05584338bf475be31d619c0951ef0).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115569228

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1b954ce5c24b4cfa61782128a0ea0bcbcfb4b110).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115575644

src/util/range.h
```diff
@@ -265,6 +265,87 @@ struct concat_iteratort
...
+  {
+    return !(*this == other);
+  }
+
+  /// Preincrement operator
+  zip_iteratort &operator++()
+  {
+    ++first_begin;
+    ++second_begin;
+    INVARIANT(
```
smowton, line 301: Your doc text claims an exception is raised in this circumstance

src/util/range.h
```diff
@@ -265,6 +265,87 @@ struct concat_iteratort
...
+    first_iteratort first_begin,
+    first_iteratort first_end,
+    second_iteratort second_begin,
+    second_iteratort second_end)
+    : first_begin(std::move(first_begin)),
+      first_end(std::move(first_end)),
+      second_begin(std::move(second_begin)),
+      second_end(std::move(second_end))
+  {
+    if(first_begin != first_end && second_begin != second_end)
```
smowton: This will silently swallow the case where one range is empty but not the other, which per docs should lead to an exception

src/util/range.h
```diff
@@ -265,6 +265,87 @@ struct concat_iteratort
   second_iteratort second_begin;
 };
 
+/// Zip two ranges to make a range of pairs.
+/// On increment, both iterators are incremented.
+/// Ends when the two ranges reach their ends, raise an exception if one ends
+/// before the other.
```
tautschnig: See @smowton's comment below, but I think this should never be an exception in the first place, it should just be an invariant.

src/util/range.h
```diff
@@ -265,6 +265,87 @@ struct concat_iteratort
...
+  bool operator!=(const zip_iteratort &other) const
+  {
+    return !(*this == other);
+  }
+
+  /// Preincrement operator
+  zip_iteratort &operator++()
+  {
+    ++first_begin;
+    ++second_begin;
```
tautschnig, line 300: I'd add a `PRECONDITION` that neither of them have reached the end before being incremented.

src/util/range.h
```diff
@@ -265,6 +265,87 @@ struct concat_iteratort
...
+    INVARIANT(
+      (first_begin == first_end) == (second_begin == second_end),
+      "Zipped ranges should have the same size");
+    current = first_begin != first_end && second_begin != second_end
```
tautschnig: Given the `INVARIANT` one of those is redundant.

src/util/range.h
```diff
@@ -265,6 +265,87 @@ struct concat_iteratort
...
+  const zip_iteratort operator++(int)
+  {
+    zip_iteratort tmp(first_begin, first_end, second_begin, second_end);
+    this->operator++();
+    return tmp;
+  }
+
+  reference operator*() const
+  {
+    return *current;
```
tautschnig, line 321: Add a `PRECONDITION` that `current != nullptr`

tautschnig (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 2530c2ca9dbcf6db55b66859e44a4402bc316b46).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115765736

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1011350503f94cd3a0b45f9f95203dfaf05fa449).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115793437

PR #4796, opened 2019-06-14
Specify linux distribution xenial in .travis.yml

owen-mc-diffblue:
Currently no dist is specified and it defaults to trusty. Soon the
default will change to xenial. Try xenial out now to see what breaks,
and if it can be easily fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ jobs:
         apt:
           sources:
             - llvm-toolchain-trusty-7
-            - ubuntu-toolchain-r-test
           packages:
             - clang-format-7
       install:
@@ -94,8 +93,6 @@ jobs:
       cache: ccache
       addons:
         apt:
-          sources:
-            - ubuntu-toolchain-r-test
           packages:
             - libwww-perl
             - g++-5
@@ -134,8 +131,6 @@ jobs:
       cache: ccache
       addons:
         apt:
-          sources:
-            - ubuntu-toolchain-r-test
           packages:
             - libwww-perl
             - g++-5
@@ -163,7 +158,6 @@ jobs:
       addons:
         apt:
           sources:
-            - ubuntu-toolchain-r-test
             - llvm-toolchain-trusty-7
           packages:
             - libwww-perl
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -88,6 +88,7 @@ jobs:
     # Ubuntu Linux with glibc using g++-5
     - stage: Linter + Doxygen + non-debug Ubuntu/gcc-5 test
       os: linux
+      dist: xenial
```
tautschnig, line 90: Do we still need `sources: ubuntu-toolchain-r-test` with Xenial?

owen-mc-diffblue:
Here's the announcement by the way: https://changelog.travis-ci.com/xenial-as-the-default-build-environment-99476

.travis.yml
owen-mc-diffblue: I can experiment with removing that, and other sources, but only once CI passes. I expect there will be a few things to sort out to get it to pass.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: b49065613e0cbfc02b1e372efef454c9a42d0f0b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115604735

owen-mc-diffblue:
It seems I was wrong and everything passed. I'll see what sources we can remove now.

owen-mc-diffblue:
@tautschnig It seems that those sources can't be removed. I've left the CI in a failing state so you can see. Anything else you want me to try? Otherwise I'll remove the second commit and get reviews.

tautschnig:
@owen-jones-diffblue I think we may need to dig deeper. Note that only the util/unicode.cpp unit test is failing. It might be https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79511, which would be consistent with 16.04 shipping GCC 5.4, while our other Travis runs use GCC 5.5. As such we probably should raise our minimum GCC version to 5.5, and possibly even fail the build if that version constraint is not satisfied (the unicode bug could lead to subtle errors).

tautschnig:
Closing as Travis is no longer being used.

PR #4799, opened 2019-06-14
Move utf16_constant_array_to_java to string_refinement_util.h

allredj:
Function is used by both `string_refinement.cpp` and `string_format_builtin_function.cpp`.

### DIFF
--- a/src/solvers/strings/string_format_builtin_function.cpp
+++ b/src/solvers/strings/string_format_builtin_function.cpp
@@ -17,10 +17,10 @@ Date:   June 2019
 
 #include "format_specifier.h"
 #include "string_format_builtin_function.h"
+#include "string_refinement_util.h"
 #include <util/range.h>
 #include <util/simplify_expr.h>
 #include <util/symbol_table.h>
-#include <util/unicode.h>
 
 static exprt format_arg_from_string(
   const array_string_exprt &string,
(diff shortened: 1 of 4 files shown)


src/solvers/strings/string_format_builtin_function.cpp
```diff
@@ -17,10 +17,10 @@ Date:   June 2019
 
 #include "format_specifier.h"
 #include "string_format_builtin_function.h"
+#include "string_refinement_util.h"
```
allredj, line 20: @romainbrenguier Do you see any problem with this added dependency?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 39af3f001bffef0817864266a4844c25564f54c3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115615820

src/solvers/strings/string_format_builtin_function.cpp
romainbrenguier: I think it's ok, but ideally `string_refinement_util.h` should be cleaned-up to only have utility functions, and the data types inside should be moved to other files.

romainbrenguier (approved):

src/solvers/strings/string_format_builtin_function.cpp
allredj: That's the plan. Hopefully we can free some time to do that.

majakusber (approved):

PR #4798, opened 2019-06-14
Replace array if-exprts before giving to solver

allredj:
This should allow us to remove all the `adjust_if_recursive` calls (https://github.com/diffblue/cbmc/pull/4653).

~Now based on #4805. Only review 2 last commits.~

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4805, opened 2019-06-17
Use global symbol generator in substitute_array_access

allredj:
Using a const symbol generator doesn't make sense and results in several symbols having the same name, thus messing up the bool_bv map.

There are no regression tests as the problem is not visible at the moment, but this fix prevents issues with more extensive usage of `substitute_array_access`, as experienced in https://github.com/diffblue/cbmc/pull/4798.

### DIFF
--- a/src/solvers/strings/string_refinement.cpp
+++ b/src/solvers/strings/string_refinement.cpp
@@ -113,8 +113,7 @@ static optionalt<exprt> get_array(
 
 static exprt substitute_array_access(
   const index_exprt &index_expr,
-  const std::function<symbol_exprt(const irep_idt &, const typet &)>
-    &symbol_generator,
+  symbol_generatort &symbol_generator,
   const bool left_propagate);
 
 /// Convert index-value map to a vector of values. If a value for an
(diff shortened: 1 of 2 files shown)


romainbrenguier (approved):

src/solvers/strings/string_refinement.cpp
```diff
@@ -1410,7 +1396,7 @@ static std::pair<bool, std::vector<exprt>> check_axioms(
 
     stream << std::string(2, ' ') << i << ".\n";
     const exprt with_concretized_arrays =
-      substitute_array_access(negaxiom, gen_symbol, true);
+      substitute_array_access(negaxiom, generator.fresh_symbol, true);
```
majakusber, line 1399: I was looking in our code base what this actually is and just to make sure I understand it right - we shouldn't make it `const` because it is an instance of `symbol_generatort` and by calling the `()` operator we actually change the underlying instance (by increasing the `symbol_count`). Right? Because we had it `const` it was behaving unexpectedly and probably not updating the count, thus resulting in multiple symbols of the same name.

majakusber (approved):

src/solvers/strings/string_refinement.cpp
allredj: @svorenova exactly.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: aeb03cf09ea5a55df52b2aecfed50843e13e9ce1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/115782889

PR #4815, opened 2019-06-19
Make nondet_symbol_exprt constructor take location

romainbrenguier:
This is how it is used in practice so this makes the constructor calls
clearer.

### DIFF
--- a/src/goto-symex/path_storage.cpp
+++ b/src/goto-symex/path_storage.cpp
@@ -13,10 +13,12 @@ Author: Kareem Khazem <karkhaz@karkhaz.com>
 #include <util/exit_codes.h>
 #include <util/make_unique.h>
 
-nondet_symbol_exprt symex_nondet_generatort::operator()(const typet &type)
+nondet_symbol_exprt symex_nondet_generatort::
+operator()(typet type, source_locationt location)
 {
-  irep_idt identifier = "symex::nondet" + std::to_string(nondet_count++);
-  return nondet_symbol_exprt(identifier, type);
+  return nondet_symbol_exprt{"symex::nondet" + std::to_string(nondet_count++),
+                             std::move(type),
+                             std::move(location)};
 }
 
 // _____________________________________________________________________________
--- a/src/goto-symex/path_storage.h
+++ b/src/goto-symex/path_storage.h
@@ -22,7 +22,7 @@
 class symex_nondet_generatort
 {
 public:
-  nondet_symbol_exprt operator()(const typet &type);
+  nondet_symbol_exprt operator()(typet type, source_locationt location);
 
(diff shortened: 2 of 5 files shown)


src/goto-symex/path_storage.cpp
```diff
@@ -19,6 +19,14 @@ nondet_symbol_exprt symex_nondet_generatort::operator()(const typet &type)
   return nondet_symbol_exprt(identifier, type);
 }
 
+nondet_symbol_exprt symex_nondet_generatort::operator()(
+  typet type, source_locationt location)
+{
+  irep_idt identifier = "symex::nondet" + std::to_string(nondet_count++);
```
smowton: Factor this with the function above into `fresh_symex_nondet_identifier` or something, so we don't accidentally change one but not the other in future

smowton (approved):

src/goto-symex/path_storage.h
```diff
@@ -22,8 +22,12 @@
 class symex_nondet_generatort
 {
 public:
+  DEPRECATED(SINCE(
+    28, 05, 2019, "Use operator()(typet, source_locationt) instead"))
   nondet_symbol_exprt operator()(const typet &type);
```
tautschnig: I think it's ok to just remove this without a deprecation path. It's not even that long ago that this code has been introduced in the first place, and no one bothered to actually do this in a way that would maintained API compatibility.

src/goto-symex/path_storage.cpp
tautschnig: See below: I think the above should just be removed.

src/util/std_expr.h
```diff
@@ -203,14 +203,26 @@ inline symbol_exprt &to_symbol_expr(exprt &expr)
 class nondet_symbol_exprt : public nullary_exprt
 {
 public:
-  /// \param identifier: Name of symbol
-  /// \param type: Type of symbol
+  DEPRECATED(SINCE(
+    28, 05, 2019,
+    "Use nondet_symbol_exprt(irep_idt, typet, source_locationt) instead"))
```
tautschnig: I disagree. There is really nothing wrong with this constructor. No need to deprecate it.

tautschnig (changes requested):

romainbrenguier:
@tautschnig I implemented the changes

tautschnig (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 112afb8018ce5129c8e6636eb447f115bb1756fc).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116268596

PR #4825, opened 2019-06-20
Add unit-test for goto symex state assignments

romainbrenguier:
This adds unit test for state.assignment, and a couple of clean-up for things I noticed while looking at the function.

### DIFF
--- a/src/goto-symex/goto_symex_state.cpp
+++ b/src/goto-symex/goto_symex_state.cpp
@@ -173,7 +173,6 @@ renamedt<ssa_exprt, L2> goto_symex_statet::assignment(
   {
     DATA_INVARIANT(!check_renaming_l1(lhs), "lhs renaming failed on l1");
   }
-  const ssa_exprt l1_lhs = lhs;
 
 #if 0
   PRECONDITION(l1_identifier != get_original_name(l1_identifier)
--- a/src/goto-symex/goto_symex_state.cpp
+++ b/src/goto-symex/goto_symex_state.cpp
@@ -174,14 +174,6 @@ renamedt<ssa_exprt, L2> goto_symex_statet::assignment(
     DATA_INVARIANT(!check_renaming_l1(lhs), "lhs renaming failed on l1");
   }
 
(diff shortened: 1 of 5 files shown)


src/goto-symex/goto_symex_state.cpp
```diff
@@ -193,8 +193,7 @@ renamedt<ssa_exprt, L2> goto_symex_statet::assignment(
     throw unsupported_operation_exceptiont(
       "pointer handling for concurrency is unsound");
 
-  // for value propagation -- the RHS is L2
-
+  // Update constant propagation map -- the RHS is L2
```
smowton, line 196: Belongs in different commit

unit/goto-symex/goto_symex_state.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+SCENARIO(
+  "Goto symex state assignment",
+  "[core][goto-symex][goto-symex-state][assignment]")
+{
+  // Initialize goto state
+  std::list<goto_programt::instructiont> target;
+  symex_targett::sourcet source{"fun", target.begin()};
+  guard_managert manager;
+  std::size_t fresh_name_count = 1;
```
smowton, line 39: Static variable inside the lambda is usual

smowton (approved):

unit/goto-symex/goto_symex_state.cpp
romainbrenguier: Yes but when ran in catch that makes its result a bit unpredictable because the same counter will be used in each test but we don't have a guarantee on the order, so I wouldn't be able to use assertions like `result2.get().get_identifier() == "foo!0#2"`

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 81aec47172121ae142e1a124415e30cdbb98076e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116391013

forejtv (approved):

PR #4525, opened 2019-04-12
Refactor element access of pointer arrays into a separate function

antlechner:
~This PR is based on https://github.com/diffblue/cbmc/pull/4524. Please only review the last commit.~

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- n/a Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- n/a My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4826, opened 2019-06-20
Refactor array object factory to allow customising the element construction

thk123:

In TG need a way to initalise an array but to not use the object factory for its elements. This refactors the array loop init code to allow customizing on this point. I also intend to upstream a refactor to make it use a `code_fort` which makes intent clearer and code simpler. 

I would also like to refactor it to take a function that deals with adding symbols - would this be a reasonable change? 

Suggest reviewing commit by commit

The interface for the free function can be simplified a bit:
 - No need to take the array_element expression, instead let the object factory add the assignment inside the loop
 - No need to return the symbol for the array, since it is passed in.

### DIFF
--- a/jbmc/src/java_bytecode/java_object_factory.cpp
+++ b/jbmc/src/java_bytecode/java_object_factory.cpp
@@ -162,6 +162,13 @@ private:
     size_t depth,
     update_in_placet update_in_place,
     const source_locationt &location);
+
+  code_blockt assign_element(
+    const exprt &element_at_counter,
+    update_in_placet update_in_place,
+    const typet &element_type,
+    size_t depth,
+    const source_locationt &location);
 };
 
 /// Initializes the pointer-typed lvalue expression `expr` to point to an object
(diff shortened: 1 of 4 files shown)


jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -153,6 +154,7 @@ class java_object_factoryt
     const exprt &max_length_expr,
     const source_locationt &location);
 
+
```
smowton: Accidental change?

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -161,6 +163,14 @@ class java_object_factoryt
     const exprt &max_length_expr,
     size_t depth,
     update_in_placet update_in_place,
+    const source_locationt &location,
+    const array_element_generatort &element_generator);
+
+  code_blockt assign_element(
+    const exprt &element_at_counter,
+    const update_in_placet &update_in_place,
```
smowton: enum type usually taken by value

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -161,6 +163,14 @@ class java_object_factoryt
     const exprt &max_length_expr,
     size_t depth,
     update_in_placet update_in_place,
+    const source_locationt &location,
+    const array_element_generatort &element_generator);
+
+  code_blockt assign_element(
+    const exprt &element_at_counter,
```
smowton: I don't immediately understand "element_at_counter", will see if I have a useful suggestion when I find the implementation...

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -1591,3 +1626,29 @@ void gen_nondet_init(
     update_in_place,
     log);
 }
+std::pair<code_blockt, symbol_exprt> nondet_array(
```
smowton: Unused here, so needs a test

smowton (commented):
`element_at_counter` was introduced in https://github.com/diffblue/cbmc/pull/4525 it turns out. It just about makes sense there when you've got `counter_expr`, the current array index, and `element_at_counter` is `array->data[counter]`, but it doesn't really make sense once factored out. I think just `element_expr` or `element` is better, perhaps with a comment that this is necessarily an lvalue expression (typically `array->data[idx]`) which we're going to assign to.

jbmc/src/java_bytecode/java_object_factory.cpp
thk123: Blast you're right - might move it into TG for now.

thk123: This ends up going away - the whole free function is now used by the Java Object Factory.

jbmc/src/java_bytecode/java_object_factory.cpp
thk123: What's the accidental change here?

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -1669,11 +1669,3 @@ std::pair<code_blockt, symbol_exprt> nondet_array(
     max_array_size);
   return std::make_pair(assignments, array_symbol);
 }
-//array_element_generatort
```
thk123: squash into  `Make creating an array a free function`

jbmc/src/java_bytecode/java_object_factory.cpp
smowton: Adding an extra blank line apparently unrelated to the rest of the commit?

src/util/nondet.cpp
```diff
@@ -21,13 +21,36 @@ symbol_exprt generate_nondet_int(
   const source_locationt &source_location,
   allocate_objectst &allocate_objects,
   code_blockt &instructions)
+{
+  const allocate_localt cls =
```
LAJW: ⛏ What's `cls`?.

LAJW (approved):
Alright.

thk123:
@smowton Could I get a re-review, hopefully it is now clearer. Also note the function is used in object factory so no untested code

thk123:
CI failure looks like a Maven 403, restarted and hope Maven smiles on me todaqy

tautschnig:
> CI failure looks like a Maven 403, restarted and hope Maven smiles on me todaqy

Good luck: https://travis-ci.community/t/continuous-maven-repo-403/3908 (from https://github.com/travis-ci/travis-ci/issues/10053).

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -225,8 +200,27 @@ void java_object_factoryt::gen_pointer_target_init(
...
   {
-    gen_nondet_array_init(
-      assignments, expr, depth + 1, update_in_place, location);
+    assignments.append(gen_nondet_array_init(
+      expr,
+      depth + 1,
+      update_in_place,
+      location,
+      [this, update_in_place, depth, location](
+        const exprt &element_at_counter,
```
smowton: Please rename

smowton (approved):
One missed rename otherwise lgtm

jbmc/src/java_bytecode/java_object_factory.cpp
thk123: [Done](https://github.com/diffblue/cbmc/compare/651ea4eed9164a7c853fdce31ff0a5208e9e1d53..8e69bd0c177c5b2672cae7823a89530a70505ac0)

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 20e4cfc5a4785e5c26ca1a1f23489697234c205a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117019855

PR #4827, opened 2019-06-20
Refactor and unit test symex_assign_symbol

romainbrenguier:
~This is based on https://github.com/diffblue/cbmc/pull/4825 (first 10 commits).~ (merged and rebased)
The goal was to add unit test for symex_assign_symbol, to make that easier we make `symex_assign_rec` and the functions it calls static, and add a header `symex_assign.h` which corresponds to `symex_assign.cpp`.

### DIFF
--- a/src/goto-symex/goto_symex.cpp
+++ b/src/goto-symex/goto_symex.cpp
@@ -10,6 +10,7 @@ Author: Daniel Kroening, kroening@kroening.com
 /// Symbolic Execution
 
 #include "goto_symex.h"
+#include "symex_assign.h"
 
 #include <util/format_expr.h>
 #include <util/pointer_offset_size.h>
--- a/src/goto-symex/goto_symex.cpp
+++ b/src/goto-symex/goto_symex.cpp
@@ -79,15 +79,7 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
       assignment_type = symex_targett::assignment_typet::HIDDEN;
 
     exprt::operandst lhs_if_then_else_conditions;
-    symex_assign_rec(
-      state,
-      lhs,
-      nil_exprt(),
-      rhs,
-      lhs_if_then_else_conditions,
-      assignment_type,
-      ns,
-      symex_config,
-      target);
+    symex_assignt{state, assignment_type, ns, symex_config, target}
+      .symex_assign_rec(lhs, nil_exprt(), rhs, lhs_if_then_else_conditions);
   }
 }
--- a/src/goto-symex/goto_symex.cpp
+++ b/src/goto-symex/goto_symex.cpp
@@ -79,7 +79,7 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
       assignment_type = symex_targett::assignment_typet::HIDDEN;
 
     exprt::operandst lhs_if_then_else_conditions;
-    symex_assignt{state, assignment_type, ns, symex_config, target}
-      .symex_assign_rec(lhs, nil_exprt(), rhs, lhs_if_then_else_conditions);
+    symex_assignt{state, assignment_type, ns, symex_config, target}.assign_rec(
+      lhs, nil_exprt(), rhs, lhs_if_then_else_conditions);
   }
 }
--- a/src/goto-symex/goto_symex.cpp
+++ b/src/goto-symex/goto_symex.cpp
@@ -11,6 +11,8 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "goto_symex.h"
 
+#include <util/format_expr.h>
+#include <util/pointer_offset_size.h>
 #include <util/simplify_expr.h>
 
 unsigned goto_symext::dynamic_counter=0;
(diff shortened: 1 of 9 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1afc94b2fc10fb23e4d9b0744f2126c4ccec3cf1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116617890

src/goto-symex/symex_assign.cpp
```diff
@@ -407,13 +421,26 @@ void goto_symext::symex_assign_from_struct(
...
       "member of symbol should be susceptible to field-sensitivity");
 
-    symex_assign_symbol(
-      state,
-      to_ssa_expr(lhs_field),
-      full_lhs,
-      comp_rhs.second,
-      guard,
-      assignment_type);
+    if(comp_rhs.second.id() == ID_struct)
```
smowton: This is duplicated from `symex_assign_symbol`, extract that too?

unit/goto-symex/symex_assign.cpp
```diff
@@ -0,0 +1,124 @@
...
+            expr_try_dynamic_cast<if_exprt>(step.ssa_rhs);
+          REQUIRE(rhs_if != nullptr);
+          REQUIRE(to_symbol_expr(rhs_if->cond()).get_identifier() == "g!0#0");
+          const auto then_value =
+            numeric_cast_v<mp_integer>(to_constant_expr(rhs_if->true_case()));
+          REQUIRE(then_value == 475);
+          const symbol_exprt rhs_symbol = to_symbol_expr(rhs_if->false_case());
+          REQUIRE(rhs_symbol.get_identifier() == "foo!0#0");
+        }
+
```
smowton: Extra whitespace

unit/goto-symex/symex_assign.cpp
```diff
@@ -0,0 +1,124 @@
...
+        ns,
+        symex_config,
+        target);
+      THEN("An equation is added to the target")
+      {
+        REQUIRE(target.SSA_steps.size() == 1);
+        SSA_stept step = target.SSA_steps.back();
+        REQUIRE(step.type == goto_trace_stept::typet::ASSIGNMENT);
+        REQUIRE(step.assignment_type == symex_targett::assignment_typet::STATE);
+        REQUIRE(step.cond_expr.id() == ID_equal);
```
smowton: Seems redundant with the check below

unit/goto-symex/symex_assign.cpp
```diff
@@ -129,4 +126,73 @@ SCENARIO(
       }
     }
   }
+  GIVEN(
+    "A lhs and rhs of type int, an empty guard, and constant propagation "
+    "activated")
+  {
+    exprt::operandst guard;
+    null_message_handlert log;
```
smowton: There's already a null message handler in https://github.com/diffblue/cbmc/blob/develop/unit/testing-utils/message.h

unit/goto-symex/symex_assign.cpp
```diff
@@ -129,4 +126,73 @@ SCENARIO(
...
+      exprt full_lhs = nil_exprt{};
+      full_lhs.type() = int_type;
+      symex_assign_symbol(info, ssa_foo, full_lhs, rhs1, guard);
+      THEN("An equation with an empty guard is added to the target")
+      {
+        REQUIRE(target_equation.SSA_steps.size() == 1);
+        SSA_stept step = target_equation.SSA_steps.back();
+        REQUIRE(step.guard == true_exprt{});
+      }
+      THEN("The propagation map contains an entry for foo")
```
smowton: The propagation map maps "foo" -> 5721

src/goto-symex/goto_symex.cpp
```diff
@@ -13,8 +13,8 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "symex_assign.h"
 
 #include <util/format_expr.h>
-#include <util/simplify_expr.h>
 #include <util/pointer_offset_size.h>
+#include <util/simplify_expr.h>
```
smowton, line 17: Don't use en-masse clang-format commits, instead format-and-squash at each commit

smowton (commented):
I agree that pulling the assign-related mutually recursive functions into their own class, which only carries around the information necessary for executing an assignment, makes them easier to test, but turning that class into a parameter pack named `info` and making all the functions static (effectively just making all that class' functions public and substituting implicit `this` for `info` everywhere) seems like a lot of churn for marginal benefit to me. What would this look like if we just created `goto_symex_assignt` and made the symex-assign-* functions members of that class instead of `goto_symext`?

Also since assign-symbol is a very hot function I'm a little worried about anonymous function used to initialise the info struct, though perhaps the compiler is smart enough for that.

romainbrenguier:
@smowton 
> I agree that pulling the assign-related mutually recursive functions into their own class, which only carries around the information necessary for executing an assignment, makes them easier to test, but turning that class into a parameter pack named info and making all the functions static (effectively just making all that class' functions public and substituting implicit this for info everywhere) seems like a lot of churn for marginal benefit to me. What would this look like if we just created goto_symex_assignt and made the symex-assign-* functions members of that class instead of goto_symext?

That would be basically the same thing, but if the `goto_symex_assignt` class would be defined in the header then all its method need to be declared there, which I find distracting when navigating header file as the public methods get lost among the private ones. On the other hand, a class would make the syntax in the body of the functions a bit nicer. So I don't mind it being done the other way but my preference goes to having the nicer interface rather than nicer field access in the code.

smowton:
Guess we need a tie-breaker from @tautschnig @peterschrammel @kroening, but my preference goes to a class, both to reduce churn and to deviate less from the codebase norms

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 667c1457ab6cdb402104cf95f7abec878b262a18).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116642140

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: f0e5c845ed20f8ad13599052742f82fc59338c06).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116645015

peterschrammel:
I vote for 'class'.

romainbrenguier:
@peterschrammel @smowton I replaced the info struct by a class

smowton:
Keep the `assign_` prefixes on the function names (function names should contain a verb), otherwise looks good

romainbrenguier:
@smowton 
> Keep the assign_ prefixes on the function names (function names should contain a verb), otherwise looks good

Done, but I disagree with the general statement that function names should be verbs, there are plenty of examples of the opposite in the standard library (for instance std::vector has size, max_size, capacity, empty, at, front, back, data), and for a functor like here, the verb is already in the class/object name.

smowton:
Sure ok, but those are properties or accessors; in this case we have an imperative action. Without the actual action specifier you get code like `symbol(s);`, which suggests to me "make a symbol out of s" or "get the symbol for s", rather than "assign to symbol s", which `assign_symbol` makes clear.

smowton (approved):
There's likely some commits to squash, otherwise lgtm

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 37321a7762715a6017450ffd8a94bf88748d6afb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116808254

src/goto-symex/goto_symex.cpp
```diff
@@ -20,3 +23,59 @@ void goto_symext::do_simplify(exprt &expr)
...
+
+  if(rhs.id() == ID_side_effect)
+  {
+    const side_effect_exprt &side_effect_expr = to_side_effect_expr(rhs);
+    const irep_idt &statement = side_effect_expr.get_statement();
+
+    if(
+      statement == ID_cpp_new || statement == ID_cpp_new_array ||
+      statement == ID_java_new_array_data)
+      symex_cpp_new(state, lhs, side_effect_expr);
```
peterschrammel, line 51: ⛏  braces after multi-line condition

src/goto-symex/goto_symex.cpp
```diff
@@ -20,3 +23,59 @@ void goto_symext::do_simplify(exprt &expr)
...
+  else
+  {
+    assignment_typet assignment_type = symex_targett::assignment_typet::STATE;
+
+    // Let's hide return value assignments.
+    if(
+      lhs.id() == ID_symbol &&
+      id2string(to_symbol_expr(lhs).get_identifier()).find("#return_value!") !=
+        std::string::npos)
+      assignment_type = symex_targett::assignment_typet::HIDDEN;
```
peterschrammel, line 70: ⛏ braces after multi-line condition

peterschrammel (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d929d3443f3a8a636e930a4c57dde22be7dddcab).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116823785

PR #4828, opened 2019-06-20
Introduce --context-include/exclude options

peterschrammel:
Allows restricting for which packages/classes/methods we will not elaborate method bodies and thus they will be stubbed.

It does not restrict parsing the bytecode itself, which can already be restricted by java-cp-include-files.
This distinction is made because we still want to parse the bytecode to obtain meta-information from the class files even though we want to stub parts of them.


jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -1049,6 +1070,15 @@ bool java_bytecode_languaget::convert_single_method(
   symbol_table_baset &symbol_table,
   optionalt<ci_lazy_methods_neededt> needed_lazy_methods)
 {
+  // Method is not in context, hence we don't convert it,
+  // unless it is a clinit_wrapper.
+  if(
+    !is_clinit_wrapper_function(function_id) && method_in_context &&
```
peterschrammel: I don't want this to merged as is. Currently, clinit_wrappers append a `::clinit_wrapper` suffix directly to the class name, which makes the structure of the name different from other methods. 
We have to fix a naming convention for synthetic methods that doesn't produce unintended side-effects. What about 'class`.<clinit_wrapper>`'?

smowton: Fine with me

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 0244c2b09bd6c8c7abd5fe4c72c2973d9cfbaf5a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116491107

kroening:
What's the meaning in jbmc when a class is loaded but the methods are stubbed?
I would say that the command-line interface needs rethinking.

peterschrammel:
@kroening, a method might be stubbed, but we still need information about class hierarchy, access qualifiers, etc.
This is less useful for JBMC, but essential for TG.

peterschrammel:
> I would say that the command-line interface needs rethinking.

I agree that the current CLI is a huge mess, but that's what I'm tackling here.
The proposed options offer a cleaner approach to what we need and will allow us to remove some of the messy options in future.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 52d62c7b3e2d6dd252f4b842c80a141681949aba).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116659701

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -1049,6 +1070,12 @@ bool java_bytecode_languaget::convert_single_method(
   symbol_table_baset &symbol_table,
   optionalt<ci_lazy_methods_neededt> needed_lazy_methods)
 {
+  // Method is not in context, hence we don't convert it
```
allredj: Slightly misleading comment. I would rephrase: `Do not convert if method is not in context`.

jbmc/src/java_bytecode/java_bytecode_language.h
```diff
@@ -240,6 +253,9 @@ class java_bytecode_languaget:public languaget
   /// It tracks objects that should be reference-equal to each other by mapping
   /// IDs of such objects to symbols that store their values.
   std::unordered_map<std::string, object_creation_referencet> references;
+
+  /// Enables checking whether we should elaborate a method body
```
allredj: The documentation should state what the variable should contain, alongside what its purpose is.

src/util/prefix_filter.h
```diff
@@ -0,0 +1,36 @@
...
+/// \file
+/// Prefix Filtering
+
+#ifndef CPROVER_UTIL_PREFIX_FILTER_H
+#define CPROVER_UTIL_PREFIX_FILTER_H
+
+#include <string>
+#include <vector>
+
+class prefix_filtert
```
allredj, line 19: Needs header documentation for the class and the file.

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -551,7 +551,7 @@ code_blockt get_thread_safe_clinit_wrapper_body(
   // the static initializer wrapper. Enable this for debugging the symex going
   // through the clinit_wrapper.
   //
-  // java::C::clinit_wrapper()
+  // java::C.<clinit_wrapper>()
```
allredj, line 554: I suppose this will make some test-gen regression tests fail, so we should synchronize with a test-gen PR and mark this one as "Requires TG approval".

allredj (commented):
Looks very good. Only significant request is for some header documentation on `prefix_filtert`.

src/util/prefix_filter.cpp
```diff
@@ -0,0 +1,44 @@
...
+/// Prefix Filtering
+
+#include "prefix_filter.h"
+
+#include <algorithm>
+
+#include "prefix.h"
+
+prefix_filtert::prefix_filtert(
+  std::vector<std::string> &&included_prefixes,
```
thk123: :question: Requiring rvalue here seems a little odd?

src/util/prefix_filter.h
```diff
@@ -0,0 +1,36 @@
...
+public:
+  prefix_filtert(
+    std::vector<std::string> &&included_prefixes,
+    std::vector<std::string> &&excluded_prefixes);
+
+  /// Return true iff \p value matches a prefix in `included_prefixes`,
+  /// but doesn't match a prefix in `excluded_prefixes`.
+  /// An empty vector of `included_prefixes` is treated as 'match all'.
+  /// An empty vector of `excluded_prefixes` is treated as 'match nothing'.
+  bool operator()(const std::string &value);
```
thk123, line 30: :bulb: To avoid the clunky calling syntax, you might just want to make this a free function that operators on a prefix_filter so you can give it a name.

unit/util/prefix_filter.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+Author: Peter Schrammel
+
+\*******************************************************************/
+
+#include <testing-utils/use_catch.h>
+
+#include <util/prefix_filter.h>
+
+TEST_CASE("include all", "[core][util][prefix_filter]")
```
thk123, line 13: :green_heart:

jbmc/src/java_bytecode/java_bytecode_language.h
```diff
@@ -240,6 +253,9 @@ class java_bytecode_languaget:public languaget
...
+  std::unique_ptr<prefix_filtert> method_in_context;
```
thk123: :bulb: It is a pair of list of user provided strings, storing as an `optional<prefix_filtert>` seems OK and simplifies the calling code

thk123: As a unique pointer it makes me the reader assume there is going to be a virtual function or something.

thk123 (approved):
The tests are perfect. I think a few tweaks to the code would greatly simplify it

tautschnig:
I accept (even when I don't like it) that we need to special-case this for Java, but if we could still aim to be as close as possible to the `remove-function-body`/`generate-function-body` interface then this would be a nice improvement.

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -52,8 +52,8 @@ static typet clinit_states_type()
 
 // Disable linter here to allow a std::string constant, since that holds
 // a length, whereas a cstr would require strlen every time.
-const std::string clinit_wrapper_suffix = "::clinit_wrapper"; // NOLINT(*)
-const std::string user_specified_clinit_suffix = "::user_specified_clinit"; // NOLINT(*)
+const std::string clinit_wrapper_suffix = ".<clinit_wrapper>"; // NOLINT(*)
```
thk123, line 55: :question: Why are you fixing in this PR?

peterschrammel: Because I would have to special-case it in my tests. I could put this change into a separate PR that goes before this one here.

jbmc/src/java_bytecode/java_static_initializers.cpp
peterschrammel: There's a corresponding TG PR.

jbmc/src/java_bytecode/java_static_initializers.cpp
thk123: No need on my behalf :slightly_smiling_face:

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a5b5ae27ca67488eecfdf7e59d003066d6c90b9d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117955655

src/util/prefix_filter.h
romainbrenguier: I agree with @thk123, the fields of this class are just inputs of the function and there is no invariant preserved by this class. I would prefer a struct with two public field, and a free function. That would be easier to use.

src/util/prefix_filter.cpp
```diff
@@ -0,0 +1,45 @@
+/*******************************************************************\
+
+Module: Prefix Filtering
+
+Author: Peter Schrammel
```
smowton, line 5: I think we usually attribute to Diffblue?

src/util/prefix_filter.cpp
```diff
@@ -0,0 +1,45 @@
...
+  std::vector<std::string> included_prefixes,
+  std::vector<std::string> excluded_prefixes)
+  : included_prefixes(std::move(included_prefixes)),
+    excluded_prefixes(std::move(excluded_prefixes))
+{
+}
+
+bool prefix_filtert::operator()(const std::string &value)
+{
+  if(!included_prefixes.empty())
```
smowton, line 28: Make `is_included` and `is_excluded` their own functions, then you can write like
```
return included_prefixes.empty() || std::any_of(...)
```
and the main function can be clearer: `is_included(value) && !is_excluded(value)`

smowton (approved):

smowton:
Oops, reviewed one commit by itself. Review of the main bit coming up.

jbmc/regression/jbmc/context-include-exclude/test_exclude_from_include.desc
```diff
@@ -0,0 +1,17 @@
...
+^EXIT=10$
+^SIGNAL=0$
+.* line 10 assertion at file Main.java line 10 .*: FAILURE
+.* line 11 assertion at file Main.java line 11 .*: SUCCESS
+.* line 12 assertion at file Main.java line 12 .*: FAILURE
+WARNING: no body for function java::Main\.myMethod:\(I\)I
+WARNING: no body for function java::org\.cprover\.MyClass\$Inner\.doIt:\(I\)I
+--
+WARNING: no body for function .*clinit_wrapper
+WARNING: no body for function java::org\.cprover\.MyClass\.<init>:\(I\)V
```
smowton, line 13: Might as well be more general -- we shouldn't see `no body for function java::org\.cprover\.MyClass\.` at all, as any exclusion should include `MyClass$Inner`

jbmc/regression/jbmc/context-include-exclude/test_exclude_from_all.desc
```diff
@@ -0,0 +1,16 @@
...
+--context-exclude 'org.cprover.MyClass$Inner.'
+^EXIT=10$
+^SIGNAL=0$
+.* line 10 assertion at file Main.java line 10 .*: SUCCESS
+.* line 11 assertion at file Main.java line 11 .*: SUCCESS
+.* line 12 assertion at file Main.java line 12 .*: FAILURE
+WARNING: no body for function java::org\.cprover\.MyClass\$Inner\.doIt:\(I\)I
+--
+WARNING: no body for function .*clinit_wrapper
+WARNING: no body for function java::org\.cprover\.MyClass\.<init>:\(I\)V
```
smowton, line 12: Similarly can generalise

jbmc/regression/jbmc/context-include-exclude/test_include.desc
```diff
@@ -0,0 +1,16 @@
...
+^SIGNAL=0$
+.* line 10 assertion at file Main.java line 10 .*: SUCCESS
+.* line 11 assertion at file Main.java line 11 .*: FAILURE
+.* line 12 assertion at file Main.java line 12 .*: FAILURE
+WARNING: no body for function java::org\.cprover\.MyClass\.<init>:\(I\)V
+WARNING: no body for function java::org\.cprover\.MyClass\.get:\(\)I
+WARNING: no body for function java::org\.cprover\.MyClass\$Inner\.doIt:\(I\)I
+--
+WARNING: no body for function .*clinit_wrapper
+WARNING: no body for function java::Main\.myMethod:\(I\)I
```
smowton, line 14: Generalise to `Main\.`?

smowton (approved):

jbmc/regression/jbmc/context-include-exclude/test_exclude_from_include.desc
peterschrammel: I've listed the methods explicitly so that the difference between the expected outputs of the different test cases is clearly visible.

PR #4829, opened 2019-06-22
added create_directory(path)

kroening:
This replaces `mkdir(...)`, which differs on Unix and Windows.

### DIFF
--- a/src/util/file_util.cpp
+++ b/src/util/file_util.cpp
@@ -193,3 +193,13 @@ bool is_directory(const std::string &path)
 
 #endif
 }
+
+bool create_directory(const std::string &path)
+{
+#ifdef _WIN32
+  return _mkdir(path.c_str()) == 0;
+#else
+  // the umask matches what std::filesystem::create_directory does
+  return mkdir(path.c_str(), 0777) == 0;
+#endif
+}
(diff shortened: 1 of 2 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 177b9688157a3c8177f38ee10d7ae506f533e4fc).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116505255

src/util/file_util.cpp
```diff
@@ -193,3 +193,12 @@ bool is_directory(const std::string &path)
 
 #endif
 }
+
+void create_directory(const std::string &path)
+{
+#ifdef _WIN32
+  _mkdir(path.c_str());
+#else
+  mkdir(path.c_str(), 0777);
```
tautschnig: Is that really the umask we should be using? Also, in either case we should at bare minimum return the value that `mkdir` returned. The current approach necessarily hides errors.

tautschnig (changes requested):

src/util/file_util.cpp
kroening: Now with comment about the umask, and with return value that matches what `std::filesystem::create_directory` does.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4a1b7cbc8426ffc3abf1747927a152f68a386ed3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/116546419

tautschnig (approved):

PR #4841, opened 2019-06-26
Define and use an expr_skeletont class

romainbrenguier:
This clarifies the meaning of skeletons, which we use in assignmentt.
This makes sure the skeleton field is always of the correct form for a
skeleton.

### DIFF
--- a/src/goto-symex/symex_assign.h
+++ b/src/goto-symex/symex_assign.h
@@ -20,6 +20,49 @@ class byte_extract_exprt;
 class ssa_exprt;
 struct symex_configt;
 
+/// Expression in which some part is missing and can be substituted for another
+/// expression.
+///
+/// For instance, a skeleton `☐[index]` where `☐` is the missing part, can be
+/// applied to an expression `x` to get `x[index]` (see
+/// \ref expr_skeletont::apply). It can also be composed with another skeleton,
+/// let say `☐.some_field` which would give the skeleton `☐.some_field[index]`
+/// (see \ref expr_skeletont::compose).
+class expr_skeletont final
+{
+public:
+  /// Empty skeleton. Applying it to an expression would give the same
+  /// expression unchanged
+  expr_skeletont() : skeleton(nil_exprt{})
+  {
+  }
+
+  /// Replace the missing part of the current skeleton by another skeleton,
+  /// ending in a bigger skeleton corresponding to the two combined.
+  expr_skeletont compose(expr_skeletont other) const;
+
+  /// Replace the missing part by the given expression, to end-up with a
+  /// complete expression
+  exprt apply(exprt expr) const;
+
+  /// Create a skeleton by removing the first operand of \p e. For instance,
+  /// remove_op0 on `array[index]` would give a skeleton in which `array` is
+  /// missing, and applying that skeleton to `array2` would give
+  /// `array2[index]`.
+  static expr_skeletont remove_op0(exprt e);
+
+private:
+  /// In \c skeleton, nil_exprt is used to mark the sub expression to be
+  /// substituted. The nil_exprt always appears recursively following the first
+  /// operands because the only way to get a skeleton is by removing the first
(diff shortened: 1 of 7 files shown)


smowton (commented):
I note all uses of `compose` and `remove_op0` go together, combine them into one method `compose(exprt)`?

romainbrenguier:
@smowton 
> I note all uses of compose and remove_op0 go together, combine them into one method compose(exprt)?

I'm not sure I would like that because it's hiding the fact that the first operand of the expression is discarded. To be clear it would have to be named something like `compose_by_removing_op0`.

smowton (approved):
OK, this is certainly better than the current code anyhow

src/goto-symex/symex_assign.h
```diff
@@ -20,6 +20,49 @@ class byte_extract_exprt;
...
 
+/// Expression in which some part is missing and can be substituted for another
+/// expression.
+///
+/// For instance, a skeleton `☐[index]` where `☐` is the missing part, can be
+/// applied to an expression `x` to get `x[index]` (see
+/// \ref expr_skeletont::apply). It can also be composed with another skeleton,
+/// let say `☐.some_field` which would give the skeleton `☐.some_field[index]`
+/// (see \ref expr_skeletont::compose).
+class expr_skeletont final
```
thk123, line 31: :bulb: I wouldn't have guessed the behaviour of this expression from the name. Perhaps `algebraic_expr`, `partial_expr` `expr_template`?

src/goto-symex/symex_assign.h
```diff
@@ -20,6 +20,49 @@ class byte_extract_exprt;
...
+public:
+  /// Empty skeleton. Applying it to an expression would give the same
+  /// expression unchanged
+  expr_skeletont() : skeleton(nil_exprt{})
+  {
+  }
+
+  /// Replace the missing part of the current skeleton by another skeleton,
+  /// ending in a bigger skeleton corresponding to the two combined.
+  expr_skeletont compose(expr_skeletont other) const;
```
thk123, line 42: Seems like this would be very easy and worthwhile to unit test

thk123 (approved):
Personally think the name isn't very clear, but still more clear than using the nil_expr directly

src/goto-symex/symex_assign.h
romainbrenguier: Yes I couldn't find a very good name, but `algebraic_expr` could be mistaken for a special type of `exprt` (like expr containing symbols), `expr_template` would be expected to be a template, `partial_expr` is as good as `expr_skeleton` I think.

src/goto-symex/symex_assign.h
romainbrenguier: I will add them in another pull request https://github.com/diffblue/cbmc/pull/4843 because it takes a lot of tries to get CI to pass these days.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c852e7241fcc5c69a947e91ef112e2f0d875c98d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117093045

unit/goto-symex/symex_assign.cpp
```diff
@@ -156,9 +152,7 @@ SCENARIO(
         WHEN("foo is assigned a second time")
         {
           const exprt rhs2 = from_integer(1841, int_type);
-          exprt full_lhs2 = nil_exprt{};
-          full_lhs.type() = int_type;
-          symex_assign.assign_symbol(ssa_foo, full_lhs2, rhs2, guard);
+          symex_assign.assign_symbol(ssa_foo, expr_skeletont{}, rhs2, guard);
```
tautschnig, line 155: I know I'm late to the party, but how is this equivalent to the previous code? The previous code had a type assigned. @romainbrenguier ?

PR #4845, opened 2019-06-27
Fix full_lhs not being updated in symex_assignt::assign_symbol after rewrite_with_to_field_symbols

romainbrenguier:
The function rewrite_with_two_field_symbols can transform the assignment and in particular lhs without the inverse transformation being reported on full_lhs leading the two to be desynchronized. The problem is illustrated in a unit test, in which the previous version of assign_symbol would have caused `original_full_lhs` to be `struct1.field1.field1` instead of `struct1.field1`.

### DIFF
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -195,6 +195,8 @@ void symex_assignt::assign_rec(
 struct assignmentt final
 {
   ssa_exprt lhs;
+  /// Skeleton to reconstruct the original lhs in the assignment
+  expr_skeletont original_lhs_skeleton;
   exprt rhs;
 };
 
@@ -326,7 +328,9 @@ static assignmentt shift_indexed_access_to_lhs(
 
     if(byte_extract.id() == ID_symbol)
     {
-      return assignmentt{to_ssa_expr(byte_extract), byte_update.value()};
+      return assignmentt{to_ssa_expr(byte_extract),
+                         std::move(assignment.original_lhs_skeleton),
+                         byte_update.value()};
     }
     else if(byte_extract.id() == ID_index || byte_extract.id() == ID_member)
     {
@@ -376,6 +380,7 @@ static assignmentt shift_indexed_access_to_lhs(
       // We may have shifted the previous lhs into the rhs; as the lhs is only
       // L1-renamed, we need to rename again.
       return assignmentt{to_ssa_expr(byte_extract),
+                         std::move(assignment.original_lhs_skeleton),
                          state.rename(std::move(ssa_rhs), ns).get()};
     }
   }
@@ -436,10 +441,11 @@ void symex_assignt::assign_non_struct_symbol(
   // introduced by assign_struct_member, are transformed into member
   // expressions on the LHS. If we add an option to disable field-sensitivity
   // in the future these should be omitted.
-  auto assignment = shift_indexed_access_to_lhs<use_update()>(
-    state, assignmentt{lhs, std::move(l2_rhs)}, ns, symex_config.simplify_opt);
-  assignment = rewrite_with_to_field_symbols<use_update()>(
-    state, std::move(assignment), ns);
+  assignmentt assignment = rewrite_with_to_field_symbols<use_update()>(
+    state,
+    shift_indexed_access_to_lhs<use_update()>(
+      state, {lhs, full_lhs, std::move(l2_rhs)}, ns, symex_config.simplify_opt),
+    ns);
 
   if(symex_config.simplify_opt)
     assignment.rhs = simplify_expr(std::move(assignment.rhs), ns);
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -264,11 +264,20 @@ static assignmentt rewrite_with_to_field_symbols(
        lhs_mod.type().id() == ID_struct_tag))
     {
       exprt field_sensitive_lhs;
+      expr_skeletont lhs_skeleton;
       const with_exprt &with_expr = to_with_expr(ssa_rhs);
 
       if(lhs_mod.type().id() == ID_array)
       {
         field_sensitive_lhs = index_exprt(lhs_mod, with_expr.where());
+        // Access in an array can appear as an index_exprt or a byte_extract
+        auto index_reverted = expr_skeletont::clear_innermost_index_expr(
+          assignment.original_lhs_skeleton);
+        lhs_skeleton = index_reverted
+                         ? *index_reverted
+                         : get_value_or_abort(
+                             expr_skeletont::clear_innermost_byte_extract_expr(
+                               assignment.original_lhs_skeleton));
       }
       else
       {
@@ -276,6 +285,9 @@ static assignmentt rewrite_with_to_field_symbols(
           lhs_mod,
           with_expr.where().get(ID_component_name),
           with_expr.new_value().type());
+        lhs_skeleton =
+          get_value_or_abort(expr_skeletont::clear_innermost_member_expr(
+            assignment.original_lhs_skeleton));
       }
 
       field_sensitive_lhs = state.field_sensitivity.apply(
@@ -286,6 +298,7 @@ static assignmentt rewrite_with_to_field_symbols(
 
       ssa_rhs = with_expr.new_value();
       lhs_mod = to_ssa_expr(field_sensitive_lhs);
+      assignment.original_lhs_skeleton = lhs_skeleton;
     }
   }
   return assignment;
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -67,6 +67,54 @@ expr_skeletont expr_skeletont::compose(expr_skeletont other) const
   return expr_skeletont(apply(other.skeleton));
 }
 
+/// In the expression corresponding to a skeleton returns a pointer to the
+/// deepest subexpression before we encounter nil.
+static exprt *deepest_not_nil(exprt &e)
+{
+  exprt *ptr = &e;
+  while(!ptr->op0().is_nil())
+    ptr = &ptr->op0();
+  return ptr;
+}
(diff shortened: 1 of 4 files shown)


src/goto-symex/symex_assign.cpp
```diff
@@ -67,6 +67,31 @@ expr_skeletont expr_skeletont::compose(expr_skeletont other) const
...
 }
 
+optionalt<expr_skeletont> expr_skeletont::array() const
+{
+  if(skeleton.id() != ID_index)
+    return {};
+  return expr_skeletont{to_index_expr(skeleton).array()};
+}
+
+optionalt<expr_skeletont> expr_skeletont::compound() const
```
smowton: Name this `member_compound`, as I would expect a method named `compound` to also return the array of an `index_exprt`

src/goto-symex/symex_assign.cpp
```diff
@@ -67,6 +67,31 @@ expr_skeletont expr_skeletont::compose(expr_skeletont other) const
...
+{
+  if(skeleton.id() != ID_member)
+    return {};
+  return expr_skeletont{to_member_expr(skeleton).compound()};
+}
+
+optionalt<expr_skeletont> expr_skeletont::object() const
```
smowton: `byte_extract_or_member_object`? Again I'd expect `object` to address `index_exprt` too.

src/goto-symex/symex_assign.cpp
```diff
@@ -214,18 +241,29 @@ static assignmentt rewrite_with_to_field_symbols(
...
       exprt field_sensitive_lhs;
+      expr_skeletont lhs_skeleton;
       const with_exprt &with_expr = to_with_expr(ssa_rhs);
 
       if(lhs_mod.type().id() == ID_array)
       {
         field_sensitive_lhs = index_exprt(lhs_mod, with_expr.where());
+        lhs_skeleton = [&] {
+          auto tmp = assignment.original_lhs_skeleton.array();
+          CHECK_RETURN(tmp.has_value());
```
smowton: Make a utility in `optional.h` for this rather than using lambdas everywhere.

src/goto-symex/symex_assign.h
```diff
@@ -51,6 +51,18 @@ class expr_skeletont final
...
 
+  /// If the skeleton corresponds to an index expression, return the array part,
+  /// which itself is a skeleton
+  optionalt<expr_skeletont> array() const;
+
+  /// If the skeleton corresponds to a member expression, return the compound
+  /// part, which itself is a skeleton
+  optionalt<expr_skeletont> compound() const;
+
+  /// If the skeleton corresponds to a byte-extract expression, return the
```
smowton: Also handles `member_exprt`

smowton (commented):
Please give the PR a descriptive name

src/goto-symex/symex_assign.cpp
romainbrenguier: Actually the `member` part was a mistake, I will make it `byte_extract_object`

romainbrenguier:
@smowton I've updated the pull request, it turned out that the original version of the expr_skeletont methods were not what was needed, they were removing from the root instead of the last branch.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 210d0b1995c68b10eebe2349a2014a1f8c15afb8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117212091

src/goto-symex/symex_assign.cpp
```diff
@@ -67,6 +67,52 @@ expr_skeletont expr_skeletont::compose(expr_skeletont other) const
...
+/// deepest subexpression before we encounter nil.
+static exprt *deepest_not_nil(exprt &e)
+{
+  exprt *ptr = &e;
+  while(!ptr->op0().is_nil())
+    ptr = &ptr->op0();
+  return ptr;
+}
+
+optionalt<expr_skeletont> expr_skeletont::revert_index(expr_skeletont skeleton)
```
smowton: `clear_innermost_index_expr`? (and similar below)

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: aa052234479a1cffaec490d7e7b9a38d848a298b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117324278

src/goto-symex/symex_assign.h
```diff
@@ -51,6 +51,24 @@ class expr_skeletont final
   /// `array2[index]`.
   static expr_skeletont remove_op0(exprt e);
 
+  /// If the deepest subexpression in the skeleton is a member
+  /// expression, remove it and return the obtained skeleton.
```
smowton: "remove it" -> "replace it with a nil expression" throughout

src/util/optional.h
```diff
@@ -38,4 +38,21 @@ typedef nonstd::bad_optional_access bad_optional_accesst;
 
 using nonstd::nullopt;
 
+/// Similar to optionalt::value but in case of empty optional, generates an
+/// invariant failure instead of throwing an exception.
+template <typename T>
+T &value(optionalt<T> &opt)
```
smowton: That's a very generic word to introduce into the top-level namespace. How about `get_optional_value`?

romainbrenguier: I don't know, optional is already in the signature so it is redundant and "get" is unnecessary. I don't see a problem with having it in the global namespace.

smowton: My intent is it should be possible to read the code without navigating to the declaration to find out what that is. `value(e)` leaves ambiguous whether I'm retrieving a value from `e` or creating one that wraps `e`. If I remove that ambiguity with `get_value(e)` then it's clear in the context `optionalt<int> e = f(); auto i = get_value(e);`, but not `get_value(f())` where the `optionalt` type is not visible.

romainbrenguier: Don't we have the same problem with `f().value()`? 
With `get_optional_value` I would expect the result to be an optional.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: fc5911be5c5f164046b8a1464292d702586f957e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117475127

src/util/optional.h
smowton: OK, let's drop the `optional`, I still think `get_` is necessary to distinguish creation from retrieval though. @NathanJPhillips notes the name really ought to make clear this might throw, so `get_value_or_abort`?

romainbrenguier: @smowton done

smowton (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 674758a1aa68adf9dd7c37e0d2d6b9d3b7a7a5fa).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117486014

unit/goto-symex/symex_assign.cpp
```diff
@@ -192,4 +192,61 @@ SCENARIO(
       }
     }
   }
+  GIVEN(
+    "An symbol `struct1` a with_exprt `struct1 with [field1 <- 234]` and "
```
peterschrammel: A symbol

peterschrammel (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: fcc0f815d174260b3edd3f97d41c309092c55c2f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117493294

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 6e93a64cbddb385af15c7b8c55fa412b1d49bcca).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117531497

unit/goto-symex/symex_assign.cpp
```diff
@@ -192,4 +192,61 @@ SCENARIO(
       }
     }
   }
+  GIVEN(
+    "A symbol `struct1` a with_exprt `struct1 with [field1 <- 234]` and "
+    "a skeleton `☐.field_1`")
```
tautschnig, line 197: This should NOT have been unit tested, but rather be covered in a regression test. The unit test merely makes sure that the code does whatever the author thinks it does, instead of testing what the user should expect. The trace output is all about providing value to humans and other actual users. But obviously I'm too late to the party.

romainbrenguier: Sometimes the function has an obvious bug but it is difficult to hit the problematic behavior with a regression test, and even if we can hit that behavior now, it may no longer be the case when other parts of the code change. In those cases unit-test are the right thing to add because it is easier to hit the piece of code you want, and it is tested independently of the rest.

tautschnig: Unit tests can be useful, and the point about making sure it is hit is a good one. But I don't accept "difficult" as a reason not to do other testing. (I typically spend most of my time designing tests rather than coding the fix.) Via unit tests we may as well be testing code that is entirely irrelevant and should possibly be removed. If future changes result in this code no longer being covered by regression tests then surely something is going wrong: either the result is no longer as expected, or the code has become redundant (and should be removed).

romainbrenguier: In this particular case I'm adding a new class and this unit-test can act as documentation by example of what the class is doing, and in addition there is value in testing that the code does what the author thinks it does.
Unit tests for a user of this function which is `assign_symbol` are also included in this PR, so we are also testing that the user is using it right.

tautschnig: I'm sorry, I was ambiguous in my use of the term "user." I care about the end user, i.e., human running CBMC. These are the only "users" that matter to me.

romainbrenguier: :+1: but I also care about developers and they read these unit tests (sometimes)

PR #4862, opened 2019-07-01
Clean up/enable StringBuilder.append tests

allredj:
Some StringBuilder tests were marked as FUTURE and caused by https://github.com/diffblue/test-gen/issues/434.
This seems now fixed, so I moved the tests to the main StringBuilder.append tests and enabled them,  but I added a `--max-nondet-string-length 1000` option. Not sure whether these tests should always pass without the option.
@romainbrenguier do you think we can remove it (the tests don't fail locally without the option, but I'm worried they might at a later time)?


jbmc/regression/jbmc-strings/StringBuilderAppend03/test.java
```diff
@@ -1,10 +1,16 @@
 public class test
 {
-   public static void main(String[] args)
-   {
-      StringBuilder sb = new StringBuilder(args[0]);
-      sb.append("Z");
-      String s = sb.toString();
-      assert(s.equals("fg"));
-   }
+    public static void nondet(String stringArg, int intArg)
```
romainbrenguier: I don't see where intArg is used

romainbrenguier: :pick: the name suggest there should also be a `det` test

romainbrenguier (approved):

jbmc/regression/jbmc-strings/StringBuilderAppend03/test.java
allredj: It's not. Removed.

allredj: The tests for deterministic inputs are in the StringBuilderAppend[01] folders.

jbmc/regression/jbmc-strings/StringBuilderAppend03/test1.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test.class
+--function test.nondet --max-nondet-string-length 1000 --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar` --property "java::test.nondet:(Ljava/lang/String;)V.assertion.1"
+^EXIT=10$
+^SIGNAL=0$
+\[.*assertion\.1\] .* line 9 .* FAILURE
```
peterschrammel, line 6: What about the other two assertions?

peterschrammel (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 62ceb1bd82bf1d9ce6f340ca566e06d1020f1ea1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117650646

jbmc/regression/jbmc-strings/StringBuilderAppend03/test1.desc
allredj: I'm testing per-property, so the other assertions are checked in the two other desc files. Is that what you mean?

PR #4864, opened 2019-07-02
[TG-8445] Dynamically allocate arguments to java entry point

owen-mc-diffblue:
This was already the case for arguments that had more than
one object type. This PR makes it the case for arguments
that only have one object type too.

The motivation for this is that when nondet-initialising the
argument with automatic local allocation we were creating the
object to be pointed to and only initialising it when the
argument is non-null, and then symex couldn't resolve the
class identifier, which stopped it from doing various
optimisations and producing a simpler formula.


smowton:
The ideal test for this should exhibit its expected benefit -- that a classid test against something so allocated can be resolved by symex, and so a subsequent virtual call only explores the really-reachable version of the function.

jbmc/src/java_bytecode/java_entry_point.cpp
```diff
@@ -351,7 +351,7 @@ std::pair<code_blockt, std::vector<exprt>> java_build_arguments(
         init_code,
         symbol_table,
         factory_parameters,
-        lifetimet::AUTOMATIC_LOCAL,
```
thk123, line 354: :bulb: I wonder if now all of the objects are dynamic if we can simplify the `allocate_objectst` to just be a function

thk123 (commented):
As Chris says, ideally we need some kind of test. Also, this would benefit from a run of test-gens tests before merging, I'll can do this tomorrow unless you beat me to it :slightly_smiling_face:

thk123:
TG looks OK - but this needs a rebase and to pass CI here first...

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d3bc7bd2f78e0fea18a8f88929718b13a93a129c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/118108484

jbmc/src/java_bytecode/java_entry_point.cpp
owen-mc-diffblue: I don't know anything about it. If so, that could be a follow-up PR.

owen-mc-diffblue:
@thk123 @smowton I've fixed CI and added a test, so this is ready to review.

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -104,20 +104,35 @@ require_goto_statements::find_struct_component_assignments(
             object_descriptor_exprt ode;
             const namespacet ns(symbol_table);
             ode.build(superclass_expr, ns);
-            if(
-              superclass_expr.get_component_name() == supercomponent_name &&
-              to_symbol_expr(ode.root_object()).get_identifier() ==
-                structure_name)
+            if(superclass_expr.get_component_name() == supercomponent_name)
             {
+              irep_idt identifier = irep_idt{};
```
smowton: `irep_idt identifier;` or `irep_idt identifier{};` achieves the same result

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -132,21 +147,35 @@ require_goto_statements::find_struct_component_assignments(
...
           const namespacet ns(symbol_table);
           ode.build(member_expr, ns);
-          if(
-            ode.root_object().id() == ID_symbol &&
-            to_symbol_expr(ode.root_object()).get_identifier() ==
-              structure_name &&
-            member_expr.get_component_name() == component_name)
+          if(member_expr.get_component_name() == component_name)
           {
+            irep_idt identifier = irep_idt{};
```
smowton: Again no need for init

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -104,20 +104,35 @@ require_goto_statements::find_struct_component_assignments(
...
             {
+              irep_idt identifier = irep_idt{};
               if(
-                code_assign.rhs() ==
-                null_pointer_exprt(to_pointer_type(code_assign.lhs().type())))
+                auto symbol_expr =
+                  expr_try_dynamic_cast<symbol_exprt>(ode.root_object()))
+                identifier = symbol_expr->get_identifier();
+              else if(
+                auto dereference_expr =
```
smowton: This makes the function a bit weird-- we're happy with `x.a = b;` and `(*x).a = b;` aka `x->a = b;`? The first one assigns (a component of) `x`, the second one looks through it to assign something else... that suggests the code *calling* this function has the wrong idea about what it's looking for.

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +324,71 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
+  const code_assignt &assignment)
+{
+  // We are expecting that the resulting statement can be of the form:
+  // 1. structure_name.(@superclass_name if given).component =
+  //     (optional type cast *) tmp_object_factory$1;
+  //   followed by a direct assignment like this:
+  //     tmp_object_factory$1 = &tmp_object_factory$2;
+  // 2. structure_name.component = (optional cast *)&tmp_object_factory$1
+  // 3. malloc_site->@UnsupportedWrapper1.field = malloc_site$0;
+  //    followed by a direct assignment like this:
```
smowton: like what?

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +324,71 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
+
+  // We must be in case 1
+  if(expr_assigned_to_rhs_symbol.id() == ID_address_of)
+  {
+    return {to_symbol_expr(to_address_of_expr(expr_assigned_to_rhs_symbol).op())
+              .get_identifier(),
+            false};
+  }
+
+  // Case 4! symbol
```
smowton: There are only 3 cases listed in the comment above

smowton (commented):
It's a mess, what with the client unit tests clearly looking for something much more specific than is warranted... I'd suggest something testing all of `java_entry_point` should be checking for its functional properties -- that it produces initial assignments capable of satisfying constraints X Y Z but not A B C -- rather than that it produces precise output code (that would be appropriate for code testing `java_object_factory` specifically in `AUTOMATIC_LOCAL` mode). Can we just turn these tests into regression tests?

jbmc/unit/java-testing-utils/require_goto_statements.cpp
allredj: Don't some compilers require that? (cf https://github.com/diffblue/cbmc/pull/4876)

tautschnig: I am still puzzled about #4876. Maybe clang++-3.7 is completely broken?

jbmc/unit/java-testing-utils/require_goto_statements.cpp
owen-mc-diffblue: Yes, it is weird. Maybe I should make a separate function for the `x->a` case.

owen-mc-diffblue: It turns out that we no longer need the `x.a` case, so I've fully switched it over to `x->a`.

owen-mc-diffblue:
@smowton This is ready to re-review

jbmc/unit/java_bytecode/goto_program_generics/generic_bases_test.cpp
```diff
@@ -469,17 +471,18 @@ SCENARIO(
...
         require_goto_statements::require_entry_point_argument_assignment(
           ID_this, entry_point_code);
 
       THEN("Object 'this' created has unspecialized inherited field")
       {
-        //   tmp_object_factory$1.@UnsupportedWrapper1.field =
-        // &tmp_object_factory$2;
-        // struct java.lang.Object { __CPROVER_string @class_identifier; }
-        // tmp_object_factory$2;
+        // malloc_site->@UnsupportedWrapper1.field = malloc_site$0;
```
smowton: Surprising to see the comments changing but not the code?

smowton: Also surely we always expect both a declaration and an allocation, not either/or?

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +294,54 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
+  const std::vector<codet> &entry_point_instructions,
+  const code_assignt &assignment)
+{
+  // We are expecting that the resulting statement is of the form:
+  //   malloc_site->(@superclass_name if given.)field = malloc_site$0;
+  // where the symbol on the right hand side is a symbol that either
+  // 1. has previously been dynamically allocated like this:
+  //      malloc_site$0 = ALLOCATE(struct java.lang.Object, 4ul, false);
+  // 2. has not
+
```
smowton: In the has not case are we expecting it is nulled?

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +294,54 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
+  auto side_effect_expr = expr_try_dynamic_cast<side_effect_exprt>(expr);
+  return side_effect_expr && side_effect_expr->get_statement() == ID_allocate;
+}
+
+static const irep_idt &
+get_name_of_symbol_with_struct_component_ptr_type(
+  const std::vector<codet> &entry_point_instructions,
+  const code_assignt &assignment)
+{
+  // We are expecting that the resulting statement is of the form:
```
smowton: Resulting statement = `assignment`?

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +294,54 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
 }
 
+static bool is_allocate(const exprt &expr)
+{
+  auto side_effect_expr = expr_try_dynamic_cast<side_effect_exprt>(expr);
+  return side_effect_expr && side_effect_expr->get_statement() == ID_allocate;
+}
+
+static const irep_idt &
+get_name_of_symbol_with_struct_component_ptr_type(
```
smowton: I'm not sure either the name or comments match what this does? It appears to expect an assignment a = b, then go looking for either b = ALLOCATE(...) or b = c, where c is again a symbol? Either way I can't see how struct_component_ptr_types are involved?

smowton (commented):
Certainly much tidier than before, but I'd appreciate some explanation regarding what we're looking for in the modified unit test

jbmc/unit/java_bytecode/goto_program_generics/generic_bases_test.cpp
owen-mc-diffblue: The comment is giving what code we expect to find. That has changed because `this_tmp_name` comes `require_entry_point_argument_assignment()`, and entry point arguments are now assigned differently.

owen-mc-diffblue: Yes. I should make the comment clearer about what's going on. Also, I think I'm trying to explain too much in the wrong place.

jbmc/unit/java-testing-utils/require_goto_statements.cpp
owen-mc-diffblue: I'll expand that comment to make things clearer. It should say that we're expecting a symbol has been assigned to it, and it's that symbol which we return. In the short term it doesn't matter which symbol we use as they both have the same type, but we return the symbol and some tests continue to use it for other things, and in case 2 they need the second symbol.

jbmc/unit/java-testing-utils/require_goto_statements.cpp
owen-mc-diffblue: Yes. I'll fix that. I think I moved this comment from somewhere else, where that phrasing presumably made more sense.

jbmc/unit/java-testing-utils/require_goto_statements.cpp
owen-mc-diffblue: Your summary is correct, except it's `a->x = b`. The fact that the returned symbol (`b` or `c`) has the correct type comes after the call to `get_name_of_symbol_with_struct_component_ptr_type`. So it's more like `get_name_of_symbol_which_we_believe_will_have_struct_component_ptr_type`. The function doesn't check that it has that type in any way.

owen-mc-diffblue:
I've addressed @smowton's comments, which helped me think more about what's going on and make it much clearer.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 11de12c3394a5f7664e8325cb55183ed1336ed37).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/118588948

allredj:
This looks like it takes code coverage down a bit. Do you see why?

owen-mc-diffblue:
I had a quick look (without really knowing how to use codecov.io) and didn't find anything that seemed bad. Here's my summary of the biggest changes:

Lost 5 lines of coverage in `exprt boolbvt::bv_get_unbounded_array(const exprt &expr) const` because it is no longer called with values.size() finite and known in advance

In `src/statement-list/statement_list_typecheck.cpp` various typecheck functions are no longer called, presumably because the things they check for are no longer created.
```
typecheck_statement_list_load(op_code, tia_element);
typecheck_statement_list_transfer(op_code, tia_element);
typecheck_statement_list_accu_int_add(op_code);
typecheck_statement_list_accu_int_sub(op_code);
typecheck_statement_list_accu_int_mul(op_code);
typecheck_statement_list_accu_int_div(op_code);
typecheck_statement_list_accu_real_mul(op_code);
typecheck_statement_list_accu_real_div(op_code);
typecheck_statement_list_set_rlo(op_code);
```

`static void get_sub_arrays(const exprt &array_expr, std::vector<exprt> &accu)` seems to now only be called with array_exprts now, not if_exprts or anything expression which contains an array_exprt as an operand.

owen-mc-diffblue:
To some extent changes like this are to be expected: this PR means that we allocate some objects in different ways, producing different code, which then exercises different code paths. I'm not sure we should remove any of the lines which are not covered any more, because they might well be useful in future.

(This is all assuming that the results of  codecov.io are fully reliable.)

allredj:
@owen-jones-diffblue OK, thanks for looking into this. It seems fine then! :+1:

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +294,49 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
+      input_symbol_identifier, entry_point_instructions)
+      .non_null_assignments;
+
+  REQUIRE(non_null_assignments_to_input_symbol.size() == 1);
+
+  const symbol_exprt *symbol_assigned_to_input_symbol =
+    expr_try_dynamic_cast<symbol_exprt>(
+      skip_typecast(non_null_assignments_to_input_symbol[0].rhs()));
+
+  // Check for case 2
```
smowton: I think these case labels are the wrong way around? Case 2 is when there *is* a source symbol

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +294,49 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
+/// Follow the chain of non-null assignments until we find a symbol that
+/// hasn't ever had another symbol assigned to it.
+/// \param entry_point_instructions: A vector of instructions
+/// \param input_symbol_identifier: The identifier of the symbol we are
+///   currently considering
+/// \return The identifier of the symbol which is (possibly indirectly) assigned
+///   to \p input_symbol_identifier and which does not have any symbol assigned
+///   to it
+static const irep_idt &
+get_symbol_at_end_of_chain_of_non_null_assignments(
```
smowton: How about get_ultimate_source_symbol?

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +294,49 @@ const code_declt &require_goto_statements::require_declaration_of_name(
   throw no_decl_found_exceptiont(variable_name.c_str());
 }
 
+/// Follow the chain of non-null assignments until we find a symbol that
```
smowton, line 331: This would be easier to explain if split into three functions: one that finds a unique non-null *expression* feeding a symbol, one that finds a unique *symbol*, and one which does the iteration. Then the first can be explained tersely that the input program may contain many "input = null" assignments and must contain exactly one "input = e"; returns e. The others are then easier to define in terms of that operation.

smowton (commented):
Apologies for being a pain in the arse on this one, but I'd really like to come up with a clearer formulation for this function if possible, and I suspect the way to do that is to split it into smaller functions with simpler individual responsibilities.

jbmc/unit/java_bytecode/goto_program_generics/mutually_recursive_generics.cpp
```diff
@@ -10,11 +10,14 @@ Author: Diffblue Ltd.
...
 #include <java-testing-utils/require_type.h>
 #include <testing-utils/use_catch.h>
+#include <util/config.h>
 
 SCENARIO(
   "Generics class with mutually recursive_generic parameters",
   "[core][java_bytecode][goto_programs_generics]")
 {
+  config.ansi_c.set_LP64();
+
```
majakusber, line 20: :pick: Why are these needed? Especially, why is the one here required?

majakusber (approved):
Looks good to me :+1:

jbmc/unit/java_bytecode/goto_program_generics/mutually_recursive_generics.cpp
owen-mc-diffblue: Without it there is failure when simplifying 0+0, because certain indexes have width 0 and the simplify_plus method checks the widths are non-zero. I assume it is needed now because some things became pointers that weren't before.

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +294,66 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
+  const auto &assignments = require_goto_statements::find_pointer_assignments(
+                              symbol_identifier, entry_point_instructions)
+                              .non_null_assignments;
+  REQUIRE(assignments.size() == 1);
+  return assignments[0].rhs();
+}
+
+/// Get the unique symbol assigned to a symbol, if one exists. There is always
+/// a unique non-null assignment to the symbol, and it is either another symbol
+/// or a dynamic allocation.
```
smowton: This doesn't enforce that it's a dynamic allocation, so just say it's either copied from another symbol (in which case we return that symbol expression) or from something else (when we return null)

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +294,66 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
+  const irep_idt &symbol_identifier)
+{
+  const auto &assignments = require_goto_statements::find_pointer_assignments(
+                              symbol_identifier, entry_point_instructions)
+                              .non_null_assignments;
+  REQUIRE(assignments.size() == 1);
+  return assignments[0].rhs();
+}
+
+/// Get the unique symbol assigned to a symbol, if one exists. There is always
```
smowton: "is always" -> "must be"

jbmc/unit/java-testing-utils/require_goto_statements.cpp
```diff
@@ -295,6 +294,66 @@ const code_declt &require_goto_statements::require_declaration_of_name(
...
+  const irep_idt &symbol_identifier)
+{
+  const auto &expr = get_unique_non_null_expression_assigned_to_symbol(
+    entry_point_instructions, symbol_identifier);
+
+  return expr_try_dynamic_cast<symbol_exprt>(skip_typecast(expr));
+}
+
+/// Follow the chain of non-null assignments until we find a symbol that
+/// hasn't ever had another symbol assigned to it.
```
smowton: Give a short example, like 
```
a = 5 + g(7)
b = a
c = b
// Given input c, returns a
```

smowton (approved):
Just nitpicks now, thanks for improving this!

owen-mc-diffblue:
TG bump: https://github.com/diffblue/test-gen/pull/3681

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 81b7e7c091c0d7f7560b1cdaf57cb69fa321d5a4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119131808

owen-mc-diffblue:
I will merge this when I've got the TG bump passing

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 2c36f47d7d7c173b7e40b30e9bdbc790187a21a6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119297350

PR #4866, opened 2019-07-02
Do not store duplicates of inherited static fields in tests for --static-values

antlechner:
In the tests for the --static-values option, we previously stored the value for an inherited static field twice, when it fact it is only needed in its declaring class.

### DIFF
--- a/jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
+++ b/jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
@@ -313,10 +313,6 @@
       "@type":"java.lang.Long",
       "value":"9223372036854775807"
     },
-    "inheritedStatic":{
-      "@type":"java.lang.Integer",
-      "value":10
-    },
     "interfaceContainerFirst":{
       "@type":"MyInterfaceContainerA",
       "field":{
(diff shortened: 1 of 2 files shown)


allredj (approved):

jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
```diff
@@ -313,10 +313,6 @@
       "@type":"java.lang.Long",
       "value":"9223372036854775807"
     },
-    "inheritedStatic":{
```
majakusber, line 316: :question: Is this how it works in the actual script too?

majakusber (approved):
I didn't quite understand the structure of these tests, only after some offline discussion. Consider adding a note to the readme file about the script in the `cprover` folder being just a simplified version of the actual, internal script.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 057e4da613ccafa5b5f7da5f5038e1dfa7eddbbb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117691838

jbmc/regression/jbmc/deterministic_assignments_json/clinit-state.json
antlechner: It will be. :slightly_smiling_face: 
The "actual script" comments might be confusing for some people. I'll think about how to resolve this and will try to add some more info in a follow-up PR.

PR #4868, opened 2019-07-02
--string-printable option also forces chars to be printable

jeannielynnmoulton:


src/solvers/strings/string_refinement.h
```diff
@@ -41,7 +41,7 @@ Author: Alberto Griggio, alberto.griggio@gmail.com
 
 #define HELP_STRING_REFINEMENT \
   " --no-refine-strings          turn off string refinement\n" \
-  " --string-printable           restrict to printable strings (experimental)\n" /* NOLINT(*) */ \
+  " --string-printable           restrict to printable non-static strings and characters (experimental)\n" /* NOLINT(*) */ \
```
allredj: `(experimental)` can go

src/solvers/strings/string_constraint_generator_main.cpp
```diff
@@ -114,6 +114,19 @@ void merge(string_constraintst &result, string_constraintst other)
     std::back_inserter(result.not_contains));
 }
 
+and_exprt char_set_restrictions(const exprt &expr, const std::string &char_set)
+{
+  // Parse char_set
+  PRECONDITION(char_set.length() == 3);
+  PRECONDITION(char_set[1] == '-');
```
LAJW: 💡  `std::regex_match` with capture groups could be a bit easier to get my head around.
⛏ You could separate concerns - get character bounds and parse character bounds
❓ Why restrictions and not bounds?
❓ Should we check lower bound is actually lower than the upper bound?

src/solvers/strings/string_constraint_generator_main.cpp
```diff
@@ -114,6 +114,19 @@ void merge(string_constraintst &result, string_constraintst other)
...
+  const char &low_char = char_set[0];
+  const char &high_char = char_set[2];
+
+  return and_exprt(
+    binary_relation_exprt(expr, ID_ge, from_integer(low_char, expr.type())),
+    binary_relation_exprt(expr, ID_le, from_integer(high_char, expr.type())));
```
LAJW: 💡 What helped me parse these things in the past - a formula in a comment above the block, like so:
```
// expr >= low_char && expr <= high_char
```

LAJW (approved):

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -303,6 +303,8 @@ class recursion_set_entryt
   }
 };
 
+const std::string printable_char_set = " -~";
```
thk123: :no_entry_sign: Please document this and maybe give it a better name, I would expect a printable_char_set to include, for example, `'a'`! 

:pick: I also would expect a set to be a `std::unordered_set<char>`

src/solvers/strings/string_constraint_generator.h
```diff
@@ -111,6 +111,11 @@ std::pair<exprt, string_constraintst> add_axioms_for_string_of_int_with_radix(
   const namespacet &ns,
   array_poolt &array_pool);
 
+/// Given an exprt and a string representing a character set of the form
+/// [low_char]-[high_char], return an and_exprt that represents restricting
+/// the selection of the character to the range in the character set.
+and_exprt char_set_restrictions(const exprt &expr, const std::string &char_set);
```
thk123: :bulb: @svorenova added a range class for expressing the `low_char-high_car` that might be good to reuse

thk123 (commented):
Realised Joel and Lukasz had already started reviewing this before I got to it, so I've only reviewed commits 1 & 2

src/solvers/strings/string_constraint_generator_main.cpp
allredj: "constraint" is the common term for that (rather than "restriction").

src/solvers/strings/string_constraint_generator_main.cpp
```diff
@@ -114,6 +114,19 @@ void merge(string_constraintst &result, string_constraintst other)
     std::back_inserter(result.not_contains));
 }
 
+and_exprt char_set_restrictions(const exprt &expr, const std::string &char_set)
```
allredj: This `char_set` format is a bit weird. It's not a regexp, and its not a string either.
That's not really your fault, but if we want to make `char_set_restrictions` work for other things than just to constraint to printable chars, it would be better for that function to take a more flexible format. I suggest giving a lower char and an upper char as parameter.

unit/solvers/strings/string_constraint_generator_main/char_set_restrictions.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+  REQUIRE(result.op0().id() == ID_ge);
+  REQUIRE(result.op0().op0().type() == unsignedbv_typet(16));
+  REQUIRE(can_cast_expr<constant_exprt>(result.op0().op1()));
+  REQUIRE(to_constant_expr(result.op0().op1()).get_value() == "61");
+
+  REQUIRE(result.op1().id() == ID_le);
+  REQUIRE(result.op1().op0().type() == unsignedbv_typet(16));
+  REQUIRE(can_cast_expr<constant_exprt>(result.op1().op1()));
+  REQUIRE(to_constant_expr(result.op1().op1()).get_value() == "7A");
```
allredj: It's hard to see what you're trying to assert here. Suggest documenting with (`GIVEN` `WHEN` `THEN`).

allredj (approved):

jbmc/regression/strings-smoke-tests/java_string_printable/Test.java
```diff
@@ -18,4 +18,10 @@ public static boolean check_char(String s, char c) {
         assert(b);
         return b;
     }
+
+    public static boolean printable_char(char c) {
+      boolean b = c >= ' ' && c <= '~';
+      assert(b);
+      return b;
+    }
```
allredj, line 26: Tests for char fields, and array of chars would be very good.

jbmc/src/java_bytecode/java_object_factory.cpp
romainbrenguier: it's actually a range, equivalent to what you would get with `[ -~]` with grep for instance. so it includes 'a'. It could be renamed to `printable_char_range`. Alternatively the interval class @thk123 mentioned below can be used here, and can be made into a string where `add_character_set_contraint` is called.

jeannielynnmoulton: The range of printable characters is from '  ' to '~'

src/solvers/strings/string_constraint_generator_main.cpp
jeannielynnmoulton: Are you feeling OK?

jbmc/regression/strings-smoke-tests/java_string_printable/Test.java
jeannielynnmoulton: Arrays weren't working! I've added a commit that addresses and tests arrays of chars.

jeannielynnmoulton: Also added test for field.

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -1420,7 +1432,9 @@ code_blockt gen_nondet_array_init(
     init_array_expr =
       typecast_exprt(init_array_expr, pointer_type(element_type));
 
-  if(element_type.id() == ID_pointer || element_type.id() == ID_c_bool)
+  if(
+    element_type.id() == ID_pointer || element_type.id() == ID_c_bool ||
+    (element_type == java_char_type() && string_printable))
```
allredj: I think we shouldn't do that. Sorry, I should have raised that earlier...
But this re-introduces an initialisation loop on all character arrays (as I imagine `--string-printable` will be universally used).
This will affect general performance, and also constrains nondeterministic char arrays to be smaller than the chosen unwind value. After a brief scan over Tika, I'm not worried that we'll lose much coverage because of the latter (because methods that have a char array as input usually contain a loop to deal with the array anyway), but the former is more of a concern I think.

Do we really need this feature to work for arrays? We decided otherwise in https://github.com/diffblue/test-gen/blob/develop/doc/cl-options.md#--java-assume-inputs-interval-lu-or-l-or-u so I guess the answer here will be the same?

allredj (changes requested):
I have some concerns about re-introducing the initialization loop. I think we should seek @peterschrammel's direction on that.

jbmc/src/java_bytecode/java_object_factory.cpp
jeannielynnmoulton: If it had been harder to implement, I would have pushed back and said that maybe we should ask Peter if this really needs to work on arrays :) so in my opinion, if it will be much worse, then no, we don't need to do it.

allredj: It's not clear what the performance handicap really is, it's more of a moral issue...
Can we accept to have loops in our society?
If you don't mind reverting it (and documenting the limitation), then it will be more consistent with `--java-assume-inputs-interval`.

peterschrammel: For `--java-assume-inputs-interval` we decided not to do it for arrays for now until we have a solution that works for arrays without re-introducing an initialisation loop,

jeannielynnmoulton: OK, reverted for now, but left the tests marked as FUTURE.

@allredj Can you suggest where to document it?

allredj: Yes, I am the doc man.
I would exactly as it was done for `--java-assume-inputs-interval`: https://github.com/diffblue/test-gen/blob/develop/doc/cl-options.md#--java-assume-inputs-interval-lu-or-l-or-u
with a `Limitation` subsection (I personally think the Jira ticket is optional, but others might feel more strongly about it).

jeannielynnmoulton: OK, I've done it in test-gen, but I'll look at the limitation subsection thing. If you don't think it should be in CBMC, can you reapprove the PR given that I've removed the array stuff?

allredj (approved):

jbmc/src/java_bytecode/java_object_factory.cpp
allredj: No, there's no good place in cbmc to document that kind of thing.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8b2afa3498f9944da490a5f83d6a9bc4b240c458).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/117946425

src/solvers/strings/string_constraint_generator.h
```diff
@@ -111,6 +111,12 @@ std::pair<exprt, string_constraintst> add_axioms_for_string_of_int_with_radix(
   const namespacet &ns,
   array_poolt &array_pool);
 
+/// Given an exprt and a string representing a character range of the form
+/// [low_char]-[high_char], return an and_exprt that represents restricting
+/// the selection of the character to the range.
+and_exprt
```
romainbrenguier: :pick: there is no particular reason this should return an `and_exprt`, I suggest making it return an `exprt`

src/solvers/strings/string_constraint_generator_main.cpp
```diff
@@ -136,18 +154,10 @@ string_constraintst add_constraint_on_characters(
...
-  const char &low_char = char_set[0];
-  const char &high_char = char_set[2];
-
   // Add constraint
   const symbol_exprt qvar = fresh_symbol("char_constr", s.length_type());
   const exprt chr = s[qvar];
-  const and_exprt char_in_set(
-    binary_relation_exprt(chr, ID_ge, from_integer(low_char, chr.type())),
-    binary_relation_exprt(chr, ID_le, from_integer(high_char, chr.type())));
+  const and_exprt char_in_set = char_range_constraints(chr, char_set);
```
romainbrenguier: can be inlined in the next constructor call

jbmc/src/java_bytecode/java_object_factory.cpp
```diff
@@ -17,6 +17,7 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <goto-programs/class_identifier.h>
 #include <goto-programs/goto_functions.h>
+#include <solvers/strings/string_constraint_generator.h>
```
romainbrenguier: I'm not sure it's a good thing to have a dependency of java bytecode on solvers (@peterschrammel?). A solution for that would be to move the function you need to `util`, but make it take a lower bound and an upper bound instead of a string, rename it `interval_constraint` or something similar, and make it work for more types (it should take mp_integers as inputs).

src/util/interval_constraint.h
```diff
@@ -0,0 +1,22 @@
...
+Author: Jeannie Moulton
+
+\*******************************************************************/
+
+#ifndef CPROVER_UTIL_INTERVAL_CONSTRAINT_H
+#define CPROVER_UTIL_INTERVAL_CONSTRAINT_H
+
+#include "expr.h"
+#include "integer_interval.h"
+#include "mp_arith.h"
```
romainbrenguier: unnecessary

src/util/interval_constraint.h
```diff
@@ -0,0 +1,22 @@
...
+
+Author: Jeannie Moulton
+
+\*******************************************************************/
+
+#ifndef CPROVER_UTIL_INTERVAL_CONSTRAINT_H
+#define CPROVER_UTIL_INTERVAL_CONSTRAINT_H
+
+#include "expr.h"
+#include "integer_interval.h"
```
romainbrenguier, line 12: exprt and integer_intervalt can be forward declared

unit/util/interval_constraint.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <util/std_types.h>
+
+SCENARIO(
+  "interval_constraint with characters",
+  "[core][util][interval_constraint]")
+{
+  GIVEN("A string representing a character range and an character expression")
+  {
+    const integer_intervalt char_range('a', 'z');
+    const exprt expr = exprt("dummy", unsignedbv_typet(16));
```
romainbrenguier: :pick: the style `const exprt expr{"dummy", unsignedbv_typet(16)};` is preferred :newspaper:

unit/util/interval_constraint.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+SCENARIO(
+  "interval_constraint with integers",
+  "[core][util][interval_constraint]")
+{
+  GIVEN("A string representing a int range and an int expression")
+  {
+    const integer_intervalt int_range(6, 9);
+    const exprt expr = exprt("dummy", unsignedbv_typet(32));
```
romainbrenguier: :newspaper:

src/util/interval_constraint.h
jeannielynnmoulton: done for exprt. cannot be done for integer_interval it seems.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 27f2894fb2bc3285e6aa2d359c9429932bdb2f2b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/118291489

romainbrenguier (approved):

jbmc/src/java_bytecode/java_object_factory.cpp
peterschrammel: Yes, that's preferable.

jbmc/regression/strings-smoke-tests/java_string_printable/test_nonprintable_char_array.desc
```diff
@@ -0,0 +1,11 @@
...
+Test.class
+--function Test.printable_char_array --max-nondet-string-length 100
+assertion.* file Test.java line 42 .* FAILURE
+^EXIT=10$
+^SIGNAL=0$
+--
+Shows that the assertion can be violated when string-printable is not enabled
+for character arrays
+
+The constraint has not been implemented for char arrays.
```
peterschrammel, line 11: Add the ticket number TG-x

peterschrammel (approved):

PR #4875, opened 2019-07-04
Replace `<stuff>' with '<stuff>'

owen-mc-diffblue:
This changes the error message from
failed to load class `java.nio.file.Path'
to
failed to load class 'java.nio.file.Path'

### DIFF
--- a/jbmc/regression/jbmc/generics_type_param/test.desc
+++ b/jbmc/regression/jbmc/generics_type_param/test.desc
@@ -7,6 +7,6 @@ Reading class AWrapper
 Reading class FWrapper
 Reading class IWrapper
 --
-failed to load class \'AWrapper\'
-failed to load class \'FWrapper\'
-failed to load class \'IWrapper\'
+failed to load class AWrapper
+failed to load class FWrapper
+failed to load class IWrapper
--- a/jbmc/regression/jbmc/generics_type_param/test.desc
+++ b/jbmc/regression/jbmc/generics_type_param/test.desc
@@ -7,6 +7,6 @@ Reading class AWrapper
 Reading class FWrapper
 Reading class IWrapper
 --
-failed to load class \`AWrapper\'
-failed to load class \`FWrapper\'
-failed to load class \`IWrapper\'
+failed to load class \'AWrapper\'
+failed to load class \'FWrapper\'
+failed to load class \'IWrapper\'
--- a/jbmc/regression/jbmc/class-loading-annotations/test-array-value-annotation-on-class.desc
+++ b/jbmc/regression/jbmc/class-loading-annotations/test-array-value-annotation-on-class.desc
@@ -3,8 +3,8 @@ ArrayValueAnnotationOnClass.class
 --verbosity 10
 ^EXIT=6$
 ^SIGNAL=0$
-Getting class `MyClassA' from file \.[\\/]MyClassA\.class
-Getting class `MyClassB' from file \.[\\/]MyClassB\.class
+Getting class 'MyClassA' from file \.[\\/]MyClassA\.class
+Getting class 'MyClassB' from file \.[\\/]MyClassB\.class
 --
 --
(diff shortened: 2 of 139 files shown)


danpoe:
That's probably intentional as the pattern with left hand ``` ` ``` and right hand `'` occurs more frequently in the code than two single quotes.

owen-mc-diffblue:
Thanks for pointing that out @danpoe , on closer inspection I see that you are right. In that case, I'm expanding the scope of this PR to replace ` with ' everywhere where the closing delimiter is '. I think this is a much more common usage of punctuation.

tautschnig:
@owen-jones-diffblue Why is this a desirable change?

kroening:
This is certainly a very unimportant point one could get very excited about.
https://www.cl.cam.ac.uk/~mgk25/ucs/quotes.html
We could add `quotebegin` and `quoteend` stream modifiers. The result could then be quotes, a change in font, etc., depending on target.

owen-mc-diffblue:
@tautschnig As the link that @kroening posted says, the convention of quoting `text like this' looks odd in most modern fonts.

I appreciate this is not a majorly important change, but it's something that annoys me. I think it makes the output look unprofessional. It was a treat I gave myself after doing some very tedious work I didn't really want to do. The prompt was someone asking about an error message, and they obviously thought it was a mistake that there was a ` instead of a '.

tautschnig:
@owen-jones-diffblue Thank you for elaborating - it would be really helpful to have the "Why?" answered in commit messages (always, not just here). That settled, we are probably very inconsistent about our quoting and thus this is a really good step forward! My only reservation is that there may be users parsing warnings and thus we'd (negatively) surprise them, but arguably no one should rely on formatting of plain-text messages so I'm ok to ignore this one.

jbmc/regression/jbmc/generics_type_param/test.desc
```diff
@@ -7,6 +7,6 @@ Reading class AWrapper
 Reading class FWrapper
 Reading class IWrapper
 --
-failed to load class \`AWrapper\'
-failed to load class \`FWrapper\'
-failed to load class \`IWrapper\'
+failed to load class \'AWrapper\'
+failed to load class \'FWrapper\'
+failed to load class \'IWrapper\'
```
tautschnig: Nit picking: why are we using escapes here?

tautschnig (approved):

peterschrammel (commented):
I actually like @kroening's suggestion of stream modifiers... For the next rainy day...

jbmc/regression/jbmc/generics_type_param/test.desc
owen-mc-diffblue: I agree, and it's probably closely enough related to this PR to include it

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8e86fe201a0277ffcff2837248b2fe0b68f09838).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/118429422

smowton (approved):

chrisr-diffblue (approved):
Looks good to me. I'd also +1 on the idea of having stream modifiers. That would be very helpful for the use-case of wanting to be able to 'redact' the output of tools, such as in cases where we would like to have the end-user be able to send us output from the tools, but where they do not want to leak any information about their own input source code. With stream modifiers it would be simple to add a redaction mode or other tokens that could be used for post-processing redaction.

romainbrenguier (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4c900d1bf8dda708f9b63812114faecf366490f5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121070770

owen-mc-diffblue:
TG bump passed CI

PR #4885, opened 2019-07-08
[TG-8293] Constant propagation for empty strings and string concatenation [blocks: #4941]

danpoe:


src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,141 @@ void symex_assignt::assign_from_struct(
...
+{
+  // assign length of string
+  assign_symbol(
+    length,
+    expr_skeletont{},
+    new_length,
+    {});
+
+  symbol_table_baset &st_mutable =
+    const_cast<symbol_table_baset &>(ns.get_symbol_table());
```
danpoe: :skull:

src/goto-symex/symex_assign.cpp
```diff
@@ -14,9 +14,17 @@ Author: Daniel Kroening, kroening@kroening.com
...
+#include <util/arith_tools.h>
 #include <util/byte_operators.h>
+#include <util/c_types.h>
 #include <util/expr_util.h>
 #include <util/format_expr.h>
+#include <util/fresh_symbol.h>
+#include <util/mathematical_expr.h>
+#include <util/string_expr.h>
+
+#include <iostream>
```
romainbrenguier: probably not necessary

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
   }
 }
 
+void symex_assignt::assign_constants(
+  const ssa_exprt &length,
+  const constant_exprt &new_length,
+  const ssa_exprt &char_array,
+  const array_exprt &new_char_array)
+{
+  // assign length of string
```
romainbrenguier: :pick: I find a comment like // length = new_length would be clearer

src/goto-symex/symex_assign.cpp
romainbrenguier: :no_entry_sign:

src/goto-symex/symex_assign.h
```diff
@@ -109,6 +115,15 @@ class symex_assignt
     const expr_skeletont &full_lhs,
     const exprt &rhs,
     exprt::operandst &guard);
+
+  bool constant_propagate_empty_string(
+    const function_application_exprt &f_l1);
+
+  void assign_constants(
```
romainbrenguier: :no_entry_sign: documentation is missing for these two functions

src/goto-symex/symex_assign.h
```diff
@@ -35,7 +36,9 @@ class symex_assignt
       assignment_type(assignment_type),
       ns(ns),
       symex_config(symex_config),
-      target(target)
+      target(target),
+      char_type(16),
+      length_type(32)
```
romainbrenguier: these fields shouldn't be necessary and this is not language independent like symex should be.

src/goto-symex/symex_assign.cpp
```diff
@@ -378,7 +472,41 @@ void symex_assignt::assign_non_struct_symbol(
     ns);
 
   if(symex_config.simplify_opt)
+  {
+    if(rhs.id() == ID_function_application)
+    {
+      INVARIANT(guard.empty(), "assignment does not have a guard");
+      INVARIANT(
+        rhs.id() == assignment.rhs.id(),
+        "rhs has same kind after renaming and field sensitivity processing");
```
romainbrenguier: I'm not sure of the reason for these invariant, maybe they are just needed for `constant_propagate_empty_string`, in which case they should come later.

src/goto-symex/symex_assign.cpp
```diff
@@ -378,7 +472,41 @@ void symex_assignt::assign_non_struct_symbol(
...
+        const irep_idt &func_id =
+          to_symbol_expr(f_l1.function()).get_identifier();
+
+        int ret = false;
+
+        if(func_id == ID_cprover_string_empty_string_func)
+        {
+          ret = constant_propagate_empty_string(f_l1);
+        }
+
```
romainbrenguier: too many blank lines

src/goto-symex/symex_assign.cpp
```diff
@@ -378,7 +472,41 @@ void symex_assignt::assign_non_struct_symbol(
...
+      const function_application_exprt &f_l2 =
+        to_function_application_expr(assignment.rhs);
+
+      if(f_l1.function().id() == ID_symbol)
+      {
+        const irep_idt &func_id =
+          to_symbol_expr(f_l1.function()).get_identifier();
+
+        int ret = false;
```
romainbrenguier: :no_entry_sign: types do not match

src/goto-symex/symex_assign.cpp
```diff
@@ -378,7 +472,41 @@ void symex_assignt::assign_non_struct_symbol(
...
+      INVARIANT(guard.empty(), "assignment does not have a guard");
+      INVARIANT(
+        rhs.id() == assignment.rhs.id(),
+        "rhs has same kind after renaming and field sensitivity processing");
+
+      const function_application_exprt &f_l1 =
+        to_function_application_expr(rhs);
+
+      const function_application_exprt &f_l2 =
+        to_function_application_expr(assignment.rhs);
```
romainbrenguier: I don't see where this is used

src/goto-symex/symex_assign.cpp
```diff
@@ -378,7 +472,41 @@ void symex_assignt::assign_non_struct_symbol(
...
+
+      if(f_l1.function().id() == ID_symbol)
+      {
+        const irep_idt &func_id =
+          to_symbol_expr(f_l1.function()).get_identifier();
```
romainbrenguier: can be inlined in the if

src/goto-symex/symex_assign.cpp
```diff
@@ -378,7 +472,41 @@ void symex_assignt::assign_non_struct_symbol(
...
+
+        int ret = false;
+
+        if(func_id == ID_cprover_string_empty_string_func)
```
romainbrenguier: the expression for ret could be simplified into a conjunction

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+    expr_skeletont{},
+    new_length,
+    {});
+
+  symbol_table_baset &st_mutable =
+    const_cast<symbol_table_baset &>(ns.get_symbol_table());
+
+  array_typet new_char_array_type(
+    char_type,
+    new_length);
```
romainbrenguier: :pick: brace `{ }` initialization is preferred. :baby_chick:

romainbrenguier: const

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+    ID_java,
+    st_mutable);
+
+  CHECK_RETURN(aux_symbol.is_state_var);
+  CHECK_RETURN(aux_symbol.is_lvalue);
+
+  aux_symbol.is_static_lifetime = true;
+  aux_symbol.is_file_local = false;
+  aux_symbol.is_thread_local = false;
+
```
romainbrenguier: :pick: unnecessary blank line

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+  aux_symbol.value = new_char_array;
+
+  exprt arr =
```
romainbrenguier: const ?

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+        ns)
+      .get();
+
+  assign_symbol(
+    to_ssa_expr(arr).get_l1_object(),
+    expr_skeletont{},
+    new_char_array,
+    {});
+
+  address_of_exprt aoe(
```
romainbrenguier: const

romainbrenguier: give a better name than `aoe`

romainbrenguier: alternatively inline this expression in the next call

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+    index_exprt(
+      aux_symbol.symbol_expr(),
+      from_integer(0, length_type)));
+
+  assign_symbol(
```
romainbrenguier: :pick: add a comment like // char_array = &aux_symbol[0]

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+    {});
+
+  symbol_table_baset &st_mutable =
+    const_cast<symbol_table_baset &>(ns.get_symbol_table());
+
+  array_typet new_char_array_type(
+    char_type,
+    new_length);
+
+  symbolt &aux_symbol = get_fresh_aux_symbol(
```
romainbrenguier: :pick: give a better name than `aux_symbol`, one that would make clear that it is a char array

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
 }
 
+void symex_assignt::assign_constants(
+  const ssa_exprt &length,
+  const constant_exprt &new_length,
+  const ssa_exprt &char_array,
+  const array_exprt &new_char_array)
+{
+  // assign length of string
+  assign_symbol(
```
romainbrenguier: since you know the rhs is not a `struct` you can call directly `assign_non_struct_symbol` :strawberry:

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+  aux_symbol.value = new_char_array;
+
+  exprt arr =
+    state
+      .rename(
+        aux_symbol.symbol_expr(),
+        ns)
+      .get();
+
+  assign_symbol(
```
romainbrenguier: :strawberry:

src/goto-symex/symex_assign.cpp
romainbrenguier: :strawberry:

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+    aoe,
+    {});
+}
+
+bool symex_assignt::constant_propagate_empty_string(
+  const function_application_exprt &f_l1)
+{
+  const constant_exprt length = from_integer(0, length_type);
+
+  array_typet char_array_type(
```
romainbrenguier: length_type and char_type should be deduced from the type of the arguments of f_l1

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+
+  assign_symbol(
+    char_array,
+    expr_skeletont{},
+    aoe,
+    {});
+}
+
+bool symex_assignt::constant_propagate_empty_string(
+  const function_application_exprt &f_l1)
```
romainbrenguier: :pick: could we use a clearer name, like `function_application_l1`?

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+{
+  const constant_exprt length = from_integer(0, length_type);
+
+  array_typet char_array_type(
+    char_type,
+    length);
+
+  INVARIANT(
+    f_l1.arguments().size() == 2,
+    "empty string primitive does not take arguments");
```
romainbrenguier: not very clear because it asserts there are 2 arguments, but the comment says there are 0

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
   }
 }
 
+void symex_assignt::assign_constants(
```
romainbrenguier: should be made clear in the name that it is for string constants

src/goto-symex/symex_assign.cpp
romainbrenguier: :baby_chick:

src/goto-symex/symex_assign.cpp
```diff
@@ -350,6 +358,92 @@ void symex_assignt::assign_from_struct(
...
+
+  array_typet char_array_type(
+    char_type,
+    length);
+
+  INVARIANT(
+    f_l1.arguments().size() == 2,
+    "empty string primitive does not take arguments");
+
+  array_exprt char_array({}, char_array_type);
```
romainbrenguier: :baby_chick:

romainbrenguier (changes requested):
Review for the first commit

src/goto-symex/symex_assign.cpp
```diff
@@ -444,6 +444,55 @@ bool symex_assignt::constant_propagate_empty_string(
   return true;
 }
 
+bool symex_assignt::constant_propagate_concat(
+  const function_application_exprt &f_l1,
+  const function_application_exprt &f_l2)
+{
+  const refined_string_exprt &s1 = to_string_expr(f_l2.arguments().at(2));
```
romainbrenguier: used only once, so can be inlined below. There are a lot of variables representing basically the same thing in this function (s1, s1_data_opt, s1_data) so it's bette to try to minimize that :lemon:

src/goto-symex/symex_assign.cpp
```diff
@@ -444,6 +444,55 @@ bool symex_assignt::constant_propagate_empty_string(
...
+  const auto s1_data_opt = try_get_string_data_array(s1, ns);
+
+  if(!s1_data_opt)
+    return false;
+
+  const array_exprt &s1_data = s1_data_opt->get();
+
+  const refined_string_exprt &s2 = to_string_expr(f_l2.arguments().at(3));
```
romainbrenguier: :lemon:

src/goto-symex/symex_assign.cpp
```diff
@@ -444,6 +444,55 @@ bool symex_assignt::constant_propagate_empty_string(
...
+
+  const array_exprt &s2_data = s2_data_opt->get();
+
+  const std::size_t new_size =
+    s1_data.operands().size() + s2_data.operands().size();
+
+  const constant_exprt new_char_array_length =
+    from_integer(new_size, length_type);
+
+  array_typet new_char_array_type(
```
romainbrenguier: :strawberry:

src/goto-symex/symex_assign.cpp
```diff
@@ -444,6 +444,55 @@ bool symex_assignt::constant_propagate_empty_string(
...
+    char_type,
+    new_char_array_length);
+
+  exprt::operandst operands(s1_data.operands());
+  operands.insert(
+    operands.end(),
+    s2_data.operands().begin(),
+    s2_data.operands().end());
```
romainbrenguier: Ideally we should avoid duplication with the code from `solver/strings/string_concatenation_builtin_function.h`, concatenation is pretty simple and unlikely to change, but for functions like format we definitely need to share the code

src/goto-symex/symex_assign.cpp
```diff
@@ -444,6 +444,55 @@ bool symex_assignt::constant_propagate_empty_string(
...
+
+  array_exprt new_char_array(
+    std::move(operands),
+    new_char_array_type);
```
romainbrenguier: :strawberry:

src/util/simplify_expr.h
```diff
@@ -27,4 +31,9 @@ bool simplify(
 // this is the preferred interface
 exprt simplify_expr(exprt src, const namespacet &ns);
 
+optionalt<std::reference_wrapper<const array_exprt>>
+try_get_string_data_array(
```
romainbrenguier: :no_entry_sign:  needs documentation

src/util/simplify_expr.h
```diff
@@ -27,4 +31,9 @@ bool simplify(
 // this is the preferred interface
 exprt simplify_expr(exprt src, const namespacet &ns);
 
+optionalt<std::reference_wrapper<const array_exprt>>
```
romainbrenguier, line 46: :pick: optional references can be replaced by pointers

romainbrenguier (changes requested):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 19f8245930aa1cf3240fdee9344dc58878559ca3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119510875

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 686dd8ae485b112206f693a5365d0086dd02524f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120002129

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d0b5245dc260be952c0880e18ec547d8a2e11afc).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120004102

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 0db7b6ecac31904256dd030c0345be226b9c875d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120051213

jbmc/regression/jbmc-strings/StringBuilderAppendConstantEvaluation1/Main.java
```diff
@@ -0,0 +1,9 @@
+public class Main {
+  public void test() {
+    StringBuilder sb = new StringBuilder("abc");
+    sb.append("xyz");
+    String s = sb.toString();
+    assert s.length() == 6;
+    assert s.startsWith("abcxyz");
```
smowton, line 7: Surely the tests should check `equals`; whether that decomposes into `startsWith` and `length` is the model's business

jbmc/regression/jbmc-strings/StringConcatenationConstantEvaluation3/Main.java
```diff
@@ -0,0 +1,8 @@
+public class Main {
+  public void test(String s1) {
+    String s2 = "abc";
+    assert (s1 + s2).startsWith("abc");
```
smowton, line 4: Comment either here or in the `.desc` files that we're testing that constant propagation does not happen in these cases

jbmc/regression/jbmc-strings/StringConcatenationConstantEvaluation3/Main.java
```diff
@@ -0,0 +1,8 @@
...
+    assert ("" + s1).startsWith("abc");
```
smowton: The way these are sequenced means there's no consistent value of `s1` that will get through the first two assertions, so the last one is actually dead code -- suggest putting these in separate functions and using `--function` instead of `--property`.

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +86,204 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+    if(f_l1.function().id() == ID_symbol)
+    {
+      const irep_idt &func_id = f_l1.function().get_identifier();
+
+      if(func_id == ID_cprover_string_concat_func)
+      {
+        return constant_propagate_string_concat(
+          state, symex_assign, f_l1, f_l2);
+      }
+      else if(func_id == ID_cprover_string_empty_string_func)
```
smowton, line 177: Surprising this is handled in `_with_side_effects`

jbmc/regression/jbmc-strings/StringBuilderAppendConstantEvaluation1/Main.java
allredj, line 7: :no_entry_sign: I think it's not sufficient to have tests for the successful case. We should have tests that check some assertion is violated.
Otherwise we're only checking for under-approximation, and assertions are true when unreachable.
See https://github.com/danpoe/cbmc/tree/feature/simplify-expression-strings/jbmc/regression/jbmc-strings/CompareToConstantEvaluation

src/goto-symex/goto_symex.h
```diff
@@ -528,6 +530,60 @@ class goto_symext
   /// \param code: The assignment to execute
   void symex_assign(statet &state, const code_assignt &code);
 
+  /// Attempt to constant propagate side effects of the assignment (if any)
+  ///
+  /// \param state: goto symex state
+  /// \param symex_assign: object handling symbol assignments
+  /// \param lhs: lhs of the assignment
+  /// \param rhs: rhs of the assignment
```
allredj, line 538: Please document return value :hamster:

src/goto-symex/goto_symex.h
```diff
@@ -528,6 +530,60 @@ class goto_symext
...
+    const function_application_exprt &f_l1);
+
+  /// Attempt to constant propagate string concatenation
+  ///
+  /// \param state: goto symex state
+  /// \param symex_assign: object handling symbol assignments
+  /// \param f_l1: application of function ID_cprover_string_concat_func with
+  ///   l1 renaming applied
+  /// \param f_l2: same as above but with l2 renaming applied
+  bool constant_propagate_string_concat(
```
allredj, line 566: Please document return value :hamster:

src/solvers/strings/array_pool.cpp
```diff
@@ -162,7 +162,10 @@ void array_poolt::insert(
   INVARIANT(
     it_bool.second, "should not associate two arrays to the same pointer");
 
-  attempt_assign_length_from_type(array_expr, length_of_array, fresh_symbol);
+  if(length_of_array.find(array_expr) == length_of_array.end())
+  {
+    attempt_assign_length_from_type(array_expr, length_of_array, fresh_symbol);
+  }
```
allredj, line 168: Some of these changes look like they should be in separate commits. :lock:

src/util/simplify_expr.cpp
```diff
@@ -1673,10 +1668,8 @@ optionalt<std::string> simplify_exprt::expr2bits(
 /// \return array expression representing the char sequence which forms the
 ///   content of the refined string expression, empty optional if the content
 ///   cannot be determined
-static optionalt<std::reference_wrapper<const array_exprt>>
-  try_get_string_data_array(
-    const refined_string_exprt &s,
-    const namespacet &ns)
+optionalt<std::reference_wrapper<const array_exprt>>
+try_get_string_data_array(const refined_string_exprt &s, const namespacet &ns)
```
allredj: :lock:

src/util/simplify_expr.cpp
```diff
@@ -64,11 +64,6 @@ struct simplify_expr_cachet
 simplify_expr_cachet simplify_expr_cache;
 #endif
 
-static optionalt<std::reference_wrapper<const array_exprt>>
-try_get_string_data_array(
-  const refined_string_exprt &s,
-  const namespacet &ns);
-
```
allredj, line 71: :lock:

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +86,204 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  CHECK_RETURN(aux_symbol.is_state_var);
+  CHECK_RETURN(aux_symbol.is_lvalue);
+
+  aux_symbol.is_static_lifetime = true;
+  aux_symbol.is_file_local = false;
+  aux_symbol.is_thread_local = false;
+
+  aux_symbol.value = new_char_array;
+
+  exprt arr = state.rename(aux_symbol.symbol_expr(), ns).get();
```
smowton: `aux_symbol_expr`?

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +86,204 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+        constant_propagate_empty_string(state, symex_assign, f_l1);
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+void goto_symext::assign_string_constant(
```
smowton, line 190: Comment on what this actually does -- generate assignments `length = new_length`, `char aux[new_length]; aux = new_char_array; char_array = &aux[0]; dummy = associate_array_to_pointer(new_char_array, &aux[0])`

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +86,204 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+      return;
+
     exprt::operandst lhs_if_then_else_conditions;
-    symex_assignt{state, assignment_type, ns, symex_config, target}.assign_rec(
+    symex_assign.assign_rec(
       lhs, expr_skeletont{}, rhs, lhs_if_then_else_conditions);
   }
 }
+
+bool goto_symext::constant_propagate_assignment_with_side_effects(
```
smowton, line 158: Suggest a better term than  `assignment_with_side_effects` might be `constraint`? It looks like this is constraining its two operands to have a particular relationship, rather than acting as the pure function the `ID_function_application` suggests. @allredj thoughts?

src/goto-symex/goto_symex.h
```diff
@@ -528,6 +530,60 @@ class goto_symext
...
+  /// \param state: goto symex state
+  /// \param symex_assign: object handling symbol assignments
+  /// \param f_l1: application of function ID_cprover_string_empty_string_func
+  ///   with l1 renaming applied
+  void constant_propagate_empty_string(
+    statet &state,
+    symex_assignt &symex_assign,
+    const function_application_exprt &f_l1);
+
+  /// Attempt to constant propagate string concatenation
```
smowton, line 558: return value? (here and elsewhere)

src/goto-symex/symex_assign.h
```diff
@@ -17,6 +17,7 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 
 class byte_extract_exprt;
 class expr_skeletont;
+class function_application_exprt;
```
smowton: This looks like it belongs in the cpp file, since we haven't changed anything here

smowton (commented):
I note you're generating a fresh constant char array every time this happens, which particularly for the empty string function is likely to generate many duplicates. How about naming the constant arrays after their constants, as in the Java frontend (https://github.com/diffblue/cbmc/blob/develop/jbmc/src/java_bytecode/java_string_literals.cpp#L59), so you can avoid recreating the same array again and again?

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +86,204 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  statet &state,
+  symex_assignt &symex_assign,
+  const exprt &lhs,
+  const exprt &rhs)
+{
+  if(!symex_config.simplify_opt)
+  {
+    return false;
+  }
```
allredj: :bulb: Somehow I would prefer this check to be external to the function. If you want to keep it in the behaviour should be documented in the function description.

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +86,204 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+
+  if(state.threads.size() > 1)
```
allredj: Needs some documentation (preferably in the function documentation).

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +86,204 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  {
+    return false;
+  }
+
+  if(rhs.id() == ID_function_application)
+  {
+    exprt rhs_l2 = state.rename(rhs, ns).get();
+
+    const function_application_exprt &f_l1 = to_function_application_expr(rhs);
```
allredj, line 166: Move this assignment below the `f_l2` one (to have it closer to its usage).

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +86,204 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  }
+
+  if(state.threads.size() > 1)
+  {
+    return false;
+  }
+
+  if(rhs.id() == ID_function_application)
+  {
+    exprt rhs_l2 = state.rename(rhs, ns).get();
```
allredj: @danpoe I don't really understand why we do an L2 renaming here. Can we look at this together please?

allredj (commented):
Looks generally good to me.
One thing to addess is that we're loosing test coverage for `add_axioms_for_empty_string()` in src/solvers/strings/string_constraint_generator_constants.cpp and a couple of lines in src/solvers/strings/string_concatenation_builtin_function.cpp.
I suppose some of these tests now get caught into constant propagation (which is good sign) but the tests need to be adapted so they don't.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 085828faedd37018456a91997f8fce45d8c2fffc).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120752999

jbmc/regression/jbmc-strings/StringBuilderAppendConstantEvaluation1/Main.java
danpoe: This test doesn't use the models. I've not included them for tests where they are not needed.

src/goto-symex/goto_symex.cpp
danpoe: Expanded the documentation in `goto_symex.h`.

src/goto-symex/goto_symex.cpp
danpoe: The `lhs` argument is currently not used (it's not needed for the string operations currently handled but might be needed in the future). The `rhs` is a `function_application_exprt` with the first two arguments being output parameters (so it's not a pure a function, a bit unfortunate I know). So overall here we handle the potential side effects of `lhs = rhs` since the evaluation of the `rhs` has side effects.

src/goto-symex/goto_symex.cpp
danpoe: The assignment of the empty string is a side effect of the whole assignment `lhs = f(length, data)` with `f` being a function application expression with ID `ID_cprover_string_empty_string_func` and `length` and `data` being the output parameters to which the empty string length and data array pointer are assigned to.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a9af62f634e3be3c700a8718281aeccaca2138fa).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120981015

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: bab613afca0f46cae9e5f8f37e9c8a21042e6d46).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120981224

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +89,257 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  const constant_exprt &new_length,
+  const ssa_exprt &char_array,
+  const array_exprt &new_char_array)
+{
+  // assign length of string
+  symex_assign.assign_symbol(length, expr_skeletont{}, new_length, {});
+
+  array_typet new_char_array_type = new_char_array.type();
+
+  new_char_array_type.set(ID_C_constant, true);
```
romainbrenguier: The blank line is distracting

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +89,257 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+    CHECK_RETURN(new_aux_symbol.is_state_var);
+    CHECK_RETURN(new_aux_symbol.is_lvalue);
+
+    new_aux_symbol.is_static_lifetime = true;
+    new_aux_symbol.is_file_local = false;
+    new_aux_symbol.is_thread_local = false;
+
+    new_aux_symbol.value = new_char_array;
+
+    exprt arr = state.rename(new_aux_symbol.symbol_expr(), ns).get();
```
romainbrenguier: const?

src/goto-symex/goto_symex.cpp
```diff
@@ -80,8 +89,257 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  else
+  {
+    aux_symbol = state.symbol_table.lookup(aux_symbol_name);
+
+    INVARIANT(
+      aux_symbol->value == new_char_array,
+      "symbol shall have value derived from char array content");
+  }
+
+  address_of_exprt aoe(
```
romainbrenguier: avoid abbreviations like `aoe`

src/goto-symex/goto_symex.cpp
danpoe: Ah, I initially did this since `l2` renaming also replaces `l1` symbols with `l2` expressions stored in the propagation map apparently. But I think I can retrieve the expressions from the propagation map directly instead of doing the renaming.

jbmc/regression/jbmc-strings/StringBuilderAppendConstantEvaluation1/Main.java
danpoe: Test added by @yumibagge

src/goto-symex/goto_symex.cpp
smowton: It's a bit weird that these only apply at the root of an assignment though. How about instead recursing over the LHS and RHS with something like `propagate_expression_side_effects`, and if it wanted to also propagate a return value it would just replace the expression with its result and let that be propagated as an ordinary assignment?

jbmc/regression/jbmc-strings/ConstantEvaluationStringConcatenation05-VerificationFailure/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+Main.class
+--function Main.test --property "java::Main.test:()V.assertion.1" --property "java::Main.test:()V.assertion.2"
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
```
allredj: Does it make sense to check for two properties at the same time?
I think this only checks that at least one of them fails.

src/goto-symex/goto_symex.cpp
smowton: and again `lhs` is not used -- this is purely a constraint on `length` and `data`, right? In which case this should be operating over expressions, not assignments

src/goto-symex/goto_symex.cpp
allredj: @danpoe, so are you doing this?

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,277 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+}
+
+void goto_symext::assign_string_constant(
+  statet &state,
+  symex_assignt &symex_assign,
+  const ssa_exprt &length,
+  const constant_exprt &new_length,
+  const ssa_exprt &char_array,
+  const array_exprt &new_char_array)
+{
```
allredj, line 197: :pick: It looks like this function could do with some refactoring. It's a bit too long to be readable. Extracting parts into named functions would make it more understandable.

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,277 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+
+  if(!s_pointer_opt)
+  {
+    return {};
+  }
+
+  return try_get_string_data_array(s_pointer_opt->get(), ns);
+}
+
+std::string goto_symext::get_alnum_string(const array_exprt &char_array)
```
allredj: Documentation would be useful.

jbmc/regression/jbmc-strings/ConstantEvaluationStringConcatenation07-NonPrintableChar/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+Main.class
+--function Main.test --property "java::Main.test:()V.assertion.1" --property "java::Main.test:()V.assertion.2"
+^Generated [0-9]+ VCC\(s\), 0 remaining after simplification$
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
```
allredj, line 7: Sorry for being a bore, but regression tests that check for success only verify that cbmc doesn't do anything. They should always be paired with a test checking for failure.

allredj (approved):

src/goto-symex/goto_symex.cpp
danpoe: Documented in the header file.

allredj: :+1:  Indeed. I don't know how I missed it...

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: cd8a0a8ca7a254abea6ef812e4ec938d8234511e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121103986

src/goto-symex/goto_symex.cpp
danpoe: Ah yes this code can now be removed.

src/goto-symex/goto_symex.cpp
danpoe: We could of course recurse into the operands. Since we currently don't need that though, can we leave this for a future PR?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: cdf4102b9257f1edb78b7e3ad1a66420f9b82a5c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121115810

jbmc/regression/jbmc-strings/ConstantEvaluationStringConcatenation07-NonPrintableChar/test.desc
allredj: After discussion with @danpoe, it seems to be unnecessary to "assertion failure" tests for each case, as not many bad things can happen.
We do want to check that we're not completely discarding the trace, so a few tests for assertion failure makes sense. The tests currently included now look sufficient to me.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: bbbe08c3b88a55755f8471bd1079ca9fac3e148c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121121929

src/goto-symex/goto_symex.cpp
danpoe: Ok, added to helper functions being called by this function.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 72624884ae1cce9e66831abaed86b4f175df22d2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121134021

src/util/simplify_expr.h
```diff
@@ -31,19 +31,19 @@ bool simplify(
 // this is the preferred interface
 exprt simplify_expr(exprt src, const namespacet &ns);
 
-/// Get char sequence from refined string expression
+/// Get char sequence from content() field of a refined string expression
```
romainbrenguier: :pick: `content()` -> `content`

src/solvers/strings/array_pool.cpp
```diff
@@ -162,7 +162,10 @@ void array_poolt::insert(
   INVARIANT(
     it_bool.second, "should not associate two arrays to the same pointer");
 
-  attempt_assign_length_from_type(array_expr, length_of_array, fresh_symbol);
+  if(length_of_array.find(array_expr) == length_of_array.end())
```
romainbrenguier, line 165: :bulb: in test-gen repository we have a `has` function which could be nice to use in cbmc too

romainbrenguier, line 165: The commit message seems to say the opposite of the change since the function is called only when there is no mapping.

jbmc/regression/jbmc-strings/StringConcatenationConstantEvaluation3/test1.desc
```diff
@@ -0,0 +1,10 @@
+CORE symex-driven-lazy-loading-expected-failure
+Main.class
+--function Main.test1 --property "java::Main.test1:(Ljava/lang/String;)V.assertion.1" --show-vcc
+^Generated [0-9]+ VCC\(s\), 1 remaining after simplification$
+^EXIT=0$
+^SIGNAL=0$
```
romainbrenguier, line 6: Check also for VERIFICATION FAILED? :bird:

jbmc/regression/jbmc-strings/StringConcatenationConstantEvaluation3/test2.desc
```diff
@@ -0,0 +1,10 @@
+CORE symex-driven-lazy-loading-expected-failure
+Main.class
+--function Main.test2 --property "java::Main.test2:(Ljava/lang/String;)V.assertion.1" --show-vcc
+^Generated [0-9]+ VCC\(s\), 1 remaining after simplification$
+^EXIT=0$
+^SIGNAL=0$
```
romainbrenguier, line 6: :bird:

jbmc/regression/jbmc-strings/StringConcatenationConstantEvaluation3/test3.desc
```diff
@@ -0,0 +1,10 @@
+CORE symex-driven-lazy-loading-expected-failure
+Main.class
+--function Main.test3 --property "java::Main.test3:(Ljava/lang/String;)V.assertion.1" --show-vcc
+^Generated [0-9]+ VCC\(s\), 1 remaining after simplification$
+^EXIT=0$
+^SIGNAL=0$
```
romainbrenguier, line 6: :bird:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  const auto &length_type = f_type.domain().at(0);
+  const auto &char_type = to_pointer_type(f_type.domain().at(1)).subtype();
+
+  const constant_exprt length = from_integer(0, length_type);
+
+  array_typet char_array_type(char_type, length);
+
+  DATA_INVARIANT(
+    f_l1.arguments().size() == 2,
+    "empty string primitive takes two output arguments");
```
romainbrenguier, line 328: this should be a precondition of the function

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+
+  const symbolt &aux_symbol =
+    string_constant_exists
+      ? state.symbol_table.lookup_ref(aux_symbol_name)
+      : get_new_string_data_symbol(
+          state, symex_assign, aux_symbol_name, char_array, new_char_array);
+
+  INVARIANT(
+    aux_symbol.value == new_char_array,
+    "symbol shall have value derived from char array content");
```
romainbrenguier, line 215: :warning: This is relying on symbols ending with `_constant_char_array` only being used to represent string constant by this function.

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+    state.symbol_table);
+
+  CHECK_RETURN(new_aux_symbol.is_state_var);
+  CHECK_RETURN(new_aux_symbol.is_lvalue);
+
+  new_aux_symbol.is_static_lifetime = true;
+  new_aux_symbol.is_file_local = false;
+  new_aux_symbol.is_thread_local = false;
+
+  new_aux_symbol.value = new_char_array;
```
romainbrenguier, line 255: :pick: this blank line can be removed

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+void goto_symext::associate_array_to_pointer(
+  statet &state,
+  symex_assignt &symex_assign,
+  const array_exprt &new_char_array,
+  const address_of_exprt &string_data)
+{
+  const symbolt &function_symbol =
+    ns.lookup(ID_cprover_associate_array_to_pointer_func);
+
+  function_application_exprt fae(
```
romainbrenguier: :pick: avoid abbreviations like fae :fox_face:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  {
+    return {};
+  }
+
+  return try_get_string_data_array(s_pointer_opt->get(), ns);
+}
+
+std::string goto_symext::get_alnum_string(const array_exprt &char_array)
+{
+  const auto &ibv_type =
```
romainbrenguier: :fox_face:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+    to_integer_bitvector_type(to_array_type(char_array.type()).subtype());
+
+  const std::size_t n_bits = ibv_type.get_width();
+  CHECK_RETURN(n_bits % 8 == 0);
+
+  static_assert(CHAR_BIT == 8, "bitwidth of char assumed to be 8");
+
+  const std::size_t n = n_bits / 8;
```
romainbrenguier: Better name than `n`?

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  symex_assignt &symex_assign,
+  const function_application_exprt &f_l1)
+{
+  const auto &f_type = to_mathematical_function_type(f_l1.function().type());
+  const auto &length_type = f_type.domain().at(0);
+  const auto &char_type = to_pointer_type(f_type.domain().at(1)).subtype();
+
+  const constant_exprt length = from_integer(0, length_type);
+
+  array_typet char_array_type(char_type, length);
```
romainbrenguier: `const` or inline in the `char_array` declaration

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+
+  DATA_INVARIANT(
+    f_l1.arguments().size() == 2,
+    "empty string primitive takes two output arguments");
+
+  array_exprt char_array({}, char_array_type);
```
romainbrenguier: const or inline below :bee:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+bool goto_symext::constant_propagate_string_concat(
+  statet &state,
+  symex_assignt &symex_assign,
+  const function_application_exprt &f_l1)
+{
+  const auto &f_type = to_mathematical_function_type(f_l1.function().type());
+  const auto &length_type = f_type.domain().at(0);
+  const auto &char_type = to_pointer_type(f_type.domain().at(1)).subtype();
+
+  const refined_string_exprt &s1 = to_string_expr(f_l1.arguments().at(2));
```
romainbrenguier, line 350: blank line above can be removed to group abbreviations together

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  const auto s1_data_opt = try_evaluate_constant_string(state, s1.content());
+
+  if(!s1_data_opt)
```
romainbrenguier, line 353: :pick: I find it more logical to but the blank line before `s1_data_opt` definition rather than after :lollipop:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+    return false;
+
+  const array_exprt &s1_data = s1_data_opt->get();
+
+  const refined_string_exprt &s2 = to_string_expr(f_l1.arguments().at(3));
+  const auto s2_data_opt = try_evaluate_constant_string(state, s2.content());
+
+  if(!s2_data_opt)
```
romainbrenguier, line 361: :lollipop:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+
+  const array_exprt &s2_data = s2_data_opt->get();
+
+  const std::size_t new_size =
+    s1_data.operands().size() + s2_data.operands().size();
+
+  const constant_exprt new_char_array_length =
+    from_integer(new_size, length_type);
+
+  array_typet new_char_array_type(char_type, new_char_array_length);
```
romainbrenguier: :bee:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+
+  exprt::operandst operands(s1_data.operands());
+  operands.insert(
+    operands.end(), s2_data.operands().begin(), s2_data.operands().end());
+
+  array_exprt new_char_array(std::move(operands), new_char_array_type);
```
romainbrenguier: :bee:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+
+  if(!s_pointer_opt)
+  {
+    return {};
+  }
+
+  return try_get_string_data_array(s_pointer_opt->get(), ns);
+}
+
+std::string goto_symext::get_alnum_string(const array_exprt &char_array)
```
romainbrenguier: :no_entry_sign: This shouldn't be a member method of `goto_symext`, it's not using any of its fields

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,285 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+
+  const exprt arr = state.rename(new_aux_symbol.symbol_expr(), ns).get();
+
+  symex_assign.assign_symbol(
+    to_ssa_expr(arr).get_l1_object(), expr_skeletont{}, new_char_array, {});
+
+  return new_aux_symbol;
+}
+
+void goto_symext::associate_array_to_pointer(
```
romainbrenguier, line 265: It doesn't look like this needs to be a member function of `goto_symext` (it only uses `ns` which can be deduced from the symbol table)

romainbrenguier (dismissed):
The `goto_symext` class is already too big, I'd prefer we try to not add to much to it. In particular functions which do not touch its private fields should not be members of `goto_symext`.

src/goto-symex/goto_symex.cpp
danpoe: Intentional for better readability.

src/goto-symex/goto_symex.cpp
danpoe: Prefer with blank line for readability.

src/goto-symex/goto_symex.cpp
danpoe: I prefer the current version, and to always have a blank line before an `if`.

src/goto-symex/goto_symex.cpp
danpoe: Prefer current version.

src/goto-symex/goto_symex.cpp
danpoe: We need the namespace here as it contains two symbol tables (those from the goto program and those minted during symex).

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 16becf6a196d4abe68db101d456e06918e6bd95d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121290733

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 71d11d56f4ee3e659aea31740e6f1380b51deed3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121745220

src/goto-symex/goto_symex.cpp
danpoe: Abbreviation is fine imo

src/goto-symex/goto_symex.cpp
danpoe: A precondition of the method is that `f_l1` is a `function_application_exprt` with ID `ID_cprover_string_empty_string_func`. So at this point it feels more appropriate to phrase the condition on the number of arguments as a `DATA_INVARIANT`. Plus it documents the fact that we have two output parameters.

src/goto-symex/goto_symex.cpp
danpoe: True. This uses `state.symbol_table` which contains only the symbols minted during symbolic execution so the naming is within our control.

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,294 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+/// \param char_array: array_exprt containing characters represented by
+///   expressions of kind constant_exprt and type unsignedbv_typet or
+///   signedbv_typet
+/// \return a string containing only alphanumeric characters
+static std::string get_alnum_string(const array_exprt &char_array)
+{
+  const auto &ibv_type =
+    to_integer_bitvector_type(to_array_type(char_array.type()).subtype());
+
+  const std::size_t n_bits = ibv_type.get_width();
```
romainbrenguier, line 131: :pick: `ibv_type` is used only once, it can be inlined here since the variable name doesn't give any information.

src/goto-symex/goto_symex.cpp
romainbrenguier: `ns` can be given as an argument

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,294 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  symbolt &return_symbol = get_fresh_aux_symbol(
+    to_mathematical_function_type(function_symbol.type).codomain(),
+    "",
+    "return_value",
+    source_locationt(),
+    function_symbol.mode,
+    ns,
+    state.symbol_table);
+
+  ssa_exprt ssa_expr(return_symbol.symbol_expr());
```
romainbrenguier: const

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,294 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  const array_exprt &new_char_array,
+  const address_of_exprt &string_data)
+{
+  const symbolt &function_symbol =
+    ns.lookup(ID_cprover_associate_array_to_pointer_func);
+
+  function_application_exprt array_to_pointer_app(
+    function_symbol.symbol_expr(), {new_char_array, string_data});
+
+  symbolt &return_symbol = get_fresh_aux_symbol(
```
romainbrenguier: const

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,294 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+void goto_symext::associate_array_to_pointer(
+  statet &state,
+  symex_assignt &symex_assign,
+  const array_exprt &new_char_array,
+  const address_of_exprt &string_data)
+{
+  const symbolt &function_symbol =
+    ns.lookup(ID_cprover_associate_array_to_pointer_func);
+
+  function_application_exprt array_to_pointer_app(
```
romainbrenguier: const and `{` `}` initialization is prefered

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,294 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+    string_constant_exists
+      ? state.symbol_table.lookup_ref(aux_symbol_name)
+      : get_new_string_data_symbol(
+          state, symex_assign, aux_symbol_name, char_array, new_char_array);
+
+  INVARIANT(
+    aux_symbol.value == new_char_array,
+    "symbol shall have value derived from char array content");
+
+  address_of_exprt string_data(
```
romainbrenguier: const

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,294 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+        constant_propagate_empty_string(state, symex_assign, f_l1);
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+void goto_symext::assign_string_constant(
```
romainbrenguier, line 190: I don't think this function belongs in the `goto_symext` class, it doesn't seem to use any of its field (apart from ns which could be given as argument). :octopus:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,294 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+  symex_assign.assign_symbol(char_array, expr_skeletont{}, string_data, {});
+
+  if(!string_constant_exists)
+  {
+    associate_array_to_pointer(
+      state, symex_assign, new_char_array, string_data);
+  }
+}
+
+const symbolt &goto_symext::get_new_string_data_symbol(
```
romainbrenguier, line 229: :octopus:

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,294 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+
+  if(!s_pointer_opt)
+  {
+    return {};
+  }
+
+  return try_get_string_data_array(s_pointer_opt->get(), ns);
+}
+
+void goto_symext::constant_propagate_empty_string(
```
romainbrenguier, line 313: :octopus:

src/goto-symex/goto_symex.cpp
danpoe: Does provide information, and is more readable as is imo

romainbrenguier: I wouldn't know what `ibv` meant if it wasn't followed by `to_integer_bitvector_type`, so at least to me it gives no information

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: e78d73ad4b454aa09daf107717962b4d1c69aa4c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122123110

src/goto-symex/goto_symex.h
```diff
@@ -528,6 +530,103 @@ class goto_symext
...
+    const exprt &lhs,
+    const exprt &rhs);
+
+  /// Create an empty string constant
+  ///
+  /// \param state: goto symex state
+  /// \param symex_assign: object handling symbol assignments
+  /// \param f_l1: application of function ID_cprover_string_empty_string_func
+  ///   with l1 renaming applied
+  void constant_propagate_empty_string(
```
romainbrenguier, line 553: Sorry, I forgot to ask for unit tests for this function and the one below (ideally also the others that are added to `goto_symext`)

danpoe:
Regarding @romainbrenguier's comment above, we have agreed that adding unit tests for the added methods in `goto_symext` is not strictly necessary as the feature is well-tested by regression tests already. The "don't like the changes" refers to some of the added methods being members of `goto_symext` as opposed to being static functions (in the same file).

jbmc/regression/jbmc-strings/ConstantEvaluationStringConcatenation07-NonPrintableChar/Main.java
```diff
@@ -0,0 +1,9 @@
+public class Main {
+  public void test() {
+    String s1 = "\t";
+    String s2 = "\\";
+    String s3 = s1 + s2;
+    assert s3.length() == 2;
+    assert s3.startsWith("\t");
```
peterschrammel: Maybe add `assert s3.endsWith("\\");`

jbmc/regression/jbmc-strings/StringBuilderDefaultConstructorConstantEvaluation/Main.java
```diff
@@ -0,0 +1,8 @@
+public class Main {
+  public void test() {
+    StringBuilder sb = new StringBuilder();
+    String s = sb.toString();
+    assert s.isEmpty();
+    assert s.startsWith("");
```
peterschrammel: Isn't that always `true`?

jbmc/regression/jbmc-strings/StringBuilderAppendConstantEvaluation1/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+Main.class
+--function Main.test --property "java::Main.test:()V.assertion.1" --property "java::Main.test:()V.assertion.2"
+^Generated [0-9]+ VCC\(s\), 0 remaining after simplification$
```
peterschrammel, line 4: Shouldn't this be exactly `1 VCC` if you use `--property`?

src/goto-symex/goto_symex.cpp
peterschrammel: >  but might be needed in the future

@danpoe, are there any operations where this is actually used?

src/goto-symex/goto_symex.cpp
```diff
@@ -87,8 +96,294 @@ void goto_symext::symex_assign(statet &state, const code_assignt &code)
...
+    state.symbol_table);
+
+  const ssa_exprt ssa_expr(return_symbol.symbol_expr());
+
+  symex_assign.assign_symbol(
+    ssa_expr, expr_skeletont{}, array_to_pointer_app, {});
+}
+
+optionalt<std::reference_wrapper<const array_exprt>>
+goto_symext::try_evaluate_constant_string(
```
peterschrammel, line 293: 🐙

danpoe: I prefer those methods to live in `goto_symext` for now as opposed to being static. The plan is to eventually create a `goto_symex_constant_propagationt` member of `goto_symext` which handles the constant propagation.

jbmc/regression/jbmc-strings/StringBuilderAppendConstantEvaluation1/test.desc
danpoe: `--property` is passed in twice here, so yes I think the value should be 2. I'll change it to use the precise value.

jbmc/regression/jbmc-strings/StringBuilderDefaultConstructorConstantEvaluation/Main.java
danpoe: Yes it should be.

jbmc/regression/jbmc-strings/ConstantEvaluationStringConcatenation07-NonPrintableChar/Main.java
danpoe: Ok

danpoe: `s1.equals(s2)` is implemented in the models as `s1.length() == s2.length() && s1.startsWith(s2)` so here we're checking that the strings are equal. To keep the tests as simple as possible we're not using the models library. I don't think it's necessary to use `endsWith()` here as this is a test for concatenation, `endsWith()` is tested elsewhere.

peterschrammel: Then it should better be `assert s3.startsWith("\t\\");` if you want to emulate equals?

src/goto-symex/goto_symex.cpp
danpoe: No, and I think that's ok.

jbmc/regression/jbmc-strings/StringBuilderAppendConstantEvaluation1/test.desc
danpoe: So it turns out that `jbmc` translates those assertions to something like:
```
if(length_return_value != 6)
{
  assert(false);
}
```
So if `length_return_value` is 6, we don't symex the assertion and thus get 0 VCCs before simplification. This is probably not stable behaviour though so I'll leave it as is. The important thing is that after simplification we have 0 remaining VCCs.

peterschrammel (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4fecf4ec42b96ee26398ef99e5bf002f8d5e25eb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122424687

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: d9f75eb459feee99ee53921418b7ed94f2ad5b2f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122439299

PR #4664, opened 2019-05-16
 Symex: resolve pointer comparisons using the value-set (second attempt)

owen-mc-diffblue:
This is a fixed version of #4444, which was reverted in #4656 because a bug was found (which has since been added as a test). The last commit is the only change from the original PR.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4891, opened 2019-07-11
pre/post traversal expression transformers

kroening:
Transformers for expressions that offer an interface similar to that offered
by the goto-program API. Both pre- and post-traversal options are available.

The key benefit over the non-const `visit` method is that sharing is only
broken up when required.

### DIFF
--- a/src/util/expr.cpp
+++ b/src/util/expr.cpp
@@ -287,6 +287,63 @@ void exprt::visit_post(std::function<void(const exprt &)> visitor) const
   visit_post_template(visitor, this);
 }
 
+optionalt<exprt>
+exprt::transform_pre(std::function<optionalt<exprt>(exprt)> visitor) const
+{
+  auto visitor_result = visitor(*this);
+
+  // make a copy
+  exprt tmp = visitor_result.value_or(*this);
+
+  bool op_changed = false;
+
+  for(auto &op : tmp.operands()) // this breaks sharing of the copy
+  {
+    auto op_result = op.transform_pre(visitor);
+    if(op_result.has_value())
+    {
+      op = std::move(op_result.value());
+      op_changed = true;
+    }
+  }
+
+  if(op_changed)
+    return std::move(tmp);
+  else
+    return visitor_result;
+}
+
+optionalt<exprt>
+exprt::transform_post(std::function<optionalt<exprt>(exprt)> visitor) const
+{
+  // make a copy
+  exprt tmp = *this;
+
+  bool op_changed = false;
+
+  for(auto &op : tmp.operands()) // this breaks sharing of the copy
+  {
+    auto op_result = op.transform_post(visitor);
+    if(op_result.has_value())
+    {
+      op = std::move(op_result.value());
+      op_changed = true;
+    }
+  }
+
+  if(op_changed)
+  {
+    auto visitor_result = visitor(tmp);
+
+    if(visitor_result.has_value())
+      return std::move(visitor_result.value());
+    else
+      return std::move(tmp);
+  }
+  else
+    return visitor(*this);
+}
+
 template <typename T>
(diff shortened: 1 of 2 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: eda58d05c391943378aec378063033d441e5775a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/118808656

src/util/expr.cpp
```diff
@@ -287,6 +287,65 @@ void exprt::visit_post(std::function<void(const exprt &)> visitor) const
...
+    return visitor_result;
+}
+
+optionalt<exprt>
+exprt::transform_post(std::function<optionalt<exprt>(exprt)> visitor) const
+{
+  exprt tmp = *this;
+  bool op_changed = false;
+
+  for(auto &op : tmp.operands()) // this breaks sharing
```
smowton: Surely breaking sharing before you know if that's necessary will negate the benefit of this

smowton (commented):
Can't find it right now as on a phone, but I implemented a similar visitor for renamed expressions recently in symex and tried to retain sharing whenever possible, perhaps use that as a starting point?

src/util/expr.cpp
kroening: Yes, but note it's done on a copy.

kroening:
You mean `renamedt<exprt, level>
goto_symex_statet::rename(exprt expr, const namespacet &ns)`?

src/util/expr.cpp
```diff
@@ -287,6 +287,65 @@ void exprt::visit_post(std::function<void(const exprt &)> visitor) const
   visit_post_template(visitor, this);
```
tautschnig, line 287: Commit message: "Transformers for expressions that offer an interface similar to that offered
by the goto-program API." There isn't enough context available to understand why what looks like a comparison of apples and oranges is useful to the reader.

src/util/expr.cpp
```diff
@@ -287,6 +287,65 @@ void exprt::visit_post(std::function<void(const exprt &)> visitor) const
...
+exprt::transform_pre(std::function<optionalt<exprt>(exprt)> visitor) const
+{
+  auto visitor_result = visitor(*this);
+
+  exprt tmp;
+
+  if(visitor_result.has_value())
+    tmp = visitor_result.value();
+  else
+    tmp = *this;
```
tautschnig: How about `exprt tmp = visitor_result.value_or(*this);`

tautschnig (changes requested):
This needs unit tests.

owen-mc-diffblue:
I believe @smowton is referring to `void selectively_mutate(renamedt<exprt, level> &renamed, typename renamedt<exprt, level>::mutator_functiont get_mutated_expr)`, from #4664.

src/util/expr.cpp
kroening: Fixed

src/util/expr.cpp
kroening: sure

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 884d002bc86101c40c4f3260dc1ae6f6a840ef55).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119023646

src/util/expr.cpp
smowton: So what? Make a copy then immediately break sharing and you've copied the input expression

smowton:
Indeed -- https://github.com/diffblue/cbmc/blob/1ea4a764caba0744e8ccec9a49da6ec51d0b78e4/src/goto-symex/renamed.h#L88 preserves sharing when possible using the depth iterator. A variant of that with pre-order visiting would permit sharing to be broken only when actually required.

PR #4895, opened 2019-07-12
Fix skeleton in shift_indexed_access_to_lhs

romainbrenguier:
If a byte_extract operation is added to the lhs, then the invert
operation must be performed on the skeleton so that
new_skeleton[new_lhs] is equivalent to skeleton[lhs].

This requires adding a method `revert_byte_extract` on skeletont.
Unit tests are added for this method and for convert assign symbol.

### DIFF
--- a/src/goto-symex/symex_assign.cpp
+++ b/src/goto-symex/symex_assign.cpp
@@ -238,80 +238,76 @@ static assignmentt shift_indexed_access_to_lhs(
   assignmentt assignment,
   const namespacet &ns)
 {
+  if(!can_cast_expr<byte_update_exprt>(assignment.rhs))
+    return assignment;
   exprt &ssa_rhs = assignment.rhs;
   ssa_exprt &lhs_mod = assignment.lhs;
-  if(
-    ssa_rhs.id() == ID_byte_update_little_endian ||
-    ssa_rhs.id() == ID_byte_update_big_endian)
+  const byte_update_exprt &byte_update = to_byte_update_expr(ssa_rhs);
+  exprt byte_extract = simplify_exprt{ns}
+                         .simplify_byte_extract(byte_extract_exprt{
+                           byte_update.id() == ID_byte_update_big_endian
+                             ? ID_byte_extract_big_endian
+                             : ID_byte_extract_little_endian,
+                           lhs_mod,
+                           byte_update.offset(),
+                           byte_update.value().type()})
+                         .expr;
+
+  if(byte_extract.id() == ID_symbol)
   {
-    const byte_update_exprt &byte_update = to_byte_update_expr(ssa_rhs);
-    exprt byte_extract = simplify_exprt{ns}
-                           .simplify_byte_extract(byte_extract_exprt{
-                             byte_update.id() == ID_byte_update_big_endian
-                               ? ID_byte_extract_big_endian
-                               : ID_byte_extract_little_endian,
-                             lhs_mod,
-                             byte_update.offset(),
-                             byte_update.value().type()})
-                           .expr;
+    return assignmentt{to_ssa_expr(byte_extract),
+                       std::move(assignment.original_lhs_skeleton),
+                       byte_update.value()};
+  }
+  else if(byte_extract.id() == ID_index || byte_extract.id() == ID_member)
+  {
+    ssa_rhs = byte_update.value();
 
-    if(byte_extract.id() == ID_symbol)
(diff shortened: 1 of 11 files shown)


tautschnig (changes requested):
Could I please ask for a an explanation why we shouldn't instead just revert #4841? If a bugfix requires +584/-94 changes then we've got a serious problem in our architecture. This doesn't scale.

romainbrenguier:
@tautschnig 
> Could I please ask for a an explanation why we shouldn't instead just revert #4841? If a bugfix requires +584/-94 changes then we've got a serious problem in our architecture. This doesn't scale.

The problem is not with #4841, this was a refactoring to makes things clearer. The bug was already there before. It was introduced when `shift_indexed_access_to_lhs` was added, but there are other places with similar problems.

tautschnig:
> The problem is not with #4841, this was a refactoring to makes things clearer. The bug was already there before. It was introduced when `shift_indexed_access_to_lhs` was added, but there are other places with similar problems.

So which bug does it fix? I cannot seem to see a single regression test. I think this is the third PR on the `expr_skeletont` stuff, which tells me that we are not testing this properly.

src/goto-symex/symex_assign.cpp
```diff
@@ -244,14 +245,15 @@ static assignmentt shift_indexed_access_to_lhs(
...
     const byte_update_exprt &byte_update = to_byte_update_expr(ssa_rhs);
-    exprt byte_extract = simplify_expr(
-      byte_extract_exprt{byte_update.id() == ID_byte_update_big_endian
-                           ? ID_byte_extract_big_endian
-                           : ID_byte_extract_little_endian,
-                         lhs_mod,
-                         byte_update.offset(),
-                         byte_update.value().type()},
-      ns);
+    exprt byte_extract = simplify_exprt{ns}
```
smowton: Let's not -- `simplify` can decide how to do whatever simplification seems appropriate; we shouldn't lift the lid in one place.

smowton:
After staring at this for a while, thoughts:

1. `shift_indexed_access_to_lhs`'s documentation is not quite right: it seems designed to usually turn `x = byte_update y offset z <- w` into `x = x with .field = (x.field with x.field[1] = w)` or similar. That's assuming the `while(byte_extract.id() == ID_index || byte_extract.id() == ID_member)` loop is expected to terminate when it hits the original `x` symbol again, and it can't generate something like `(byte_extract struct A from x offset (z-2)).field`. I'm quite surprised there are situations where the `byte_extract` operation transferred to the LHS will simplify into array-cell or field accesses but the `byte_update` operation on the RHS won't simplify into `with` expressions! @tautschnig ?
2. Regardless of what `shift_indexed_access_to_lhs` is doing, it doesn't affect the *real* lhs (the one recorded in the GOTO program), which is I think what `full_lhs` is supposed to be tracking. Therefore to my mind it should *not* change the skeleton.
3. The same line of reasoning applies to `rewrite_with_to_field_symbols` -- yes, that should turn `x = x with .field = value` into `x..field = value`, but again the actual operation happening at GOTO level is unchanged, only the way symex is encoding it has changed. Therefore I think this function should also not touch the skeleton.
4. On the other hand, line `const exprt l2_full_lhs = state.rename(assignment.original_lhs_skeleton.apply(l2_lhs), ns).get();` in `symex_assign.cpp` seems highly suspicious -- during the `symex_assign_rec` loop we might have recorded various index and member operations which were undone by `rewrite_with_to_field_symbols`, then we proceed the blithely stick the ssa expressions together regardless. I think all the dancing around with the skeleton in `rewrite_with_to_field_symbols` and `shift_indexed_access_to_lhs` is just trying to make that step go as we hope it might, and instead we should take `l2_lhs`'s root symbol, which should be the same as that of the `lhs` passed into `assign_non_struct_symbol`, then rename as at present.

To sum up: `l2_full_lhs` should be tracking the GOTO-level operation, not how symex chooses to represent it, therefore it should take no part in the field-sensitivity dance in `symex_assignt::assign_non_struct_symbol`. I'll put together an alternative PR tomorrow morning to test this solution.

smowton:
Here's my rival PR: https://github.com/diffblue/cbmc/pull/4917

Long story short: do all the LHS rvalue renaming and simplifying right at the start of `symex_assign`, that way we can delete `shift_indexed_access_to_lhs` and `rewrite_with_to_field_symbols` entirely and thus the skeleton manipulations get much simpler.

romainbrenguier:
https://github.com/diffblue/cbmc/pull/4917 makes this irrelevant

PR #4912, opened 2019-07-15
JSON equality operator

LAJW:

### DIFF
--- a/scripts/cpplint.py
+++ b/scripts/cpplint.py
@@ -3928,60 +3928,6 @@ def CheckBracesSpacing(filename, clean_lines, linenum, nesting_state, error):
   """
   line = clean_lines.elided[linenum]
 
-  # Except after an opening paren, or after another opening brace (in case of
-  # an initializer list, for instance), you should have spaces before your
-  # braces when they are delimiting blocks, classes, namespaces etc.
-  # And since you should never have braces at the beginning of a line,
-  # this is an easy test.  Except that braces used for initialization don't
-  # follow the same rule; we often don't want spaces before those.
-  match = Match(r'^(.*[^ ({>]){', line)
-
-  if match:
-    # Try a bit harder to check for brace initialization.  This
-    # happens in one of the following forms:
-    #   Constructor() : initializer_list_{} { ... }
-    #   Constructor{}.MemberFunction()
-    #   Type variable{};
-    #   FunctionCall(type{}, ...);
-    #   LastArgument(..., type{});
-    #   LOG(INFO) << type{} << " ...";
-    #   map_of_type[{...}] = ...;
-    #   ternary = expr ? new type{} : nullptr;
-    #   OuterTemplate<InnerTemplateConstructor<Type>{}>
-    #
-    # We check for the character following the closing brace, and
-    # silence the warning if it's one of those listed above, i.e.
-    # "{.;,)<>]:".
-    #
-    # To account for nested initializer list, we allow any number of
-    # closing braces up to "{;,)<".  We can't simply silence the
-    # warning on first sight of closing brace, because that would
-    # cause false negatives for things that are not initializer lists.
-    #   Silence this:         But not this:
-    #     Outer{                if (...) {
-    #       Inner{...}            if (...){  // Missing space before {
-    #     };                    }
-    #
-    # There is a false negative with this approach if people inserted
-    # spurious semicolons, e.g. "if (cond){};", but we will catch the
-    # spurious semicolon with a separate check.
-    leading_text = match.group(1)
-    (endline, endlinenum, endpos) = CloseExpression(
-        clean_lines, linenum, len(match.group(1)))
-    trailing_text = ''
-    if endpos > -1:
-      trailing_text = endline[endpos:]
-    for offset in xrange(endlinenum + 1,
-                         min(endlinenum + 3, clean_lines.NumLines() - 1)):
-      trailing_text += clean_lines.elided[offset]
-    # We also suppress warnings for `uint64_t{expression}` etc., as the style
-    # guide recommends brace initialization for integral types to avoid
-    # overflow/truncation.
-    if (not Match(r'^[\s}]*[{.;,)<>\]:]', trailing_text)
-        and not _IsType(clean_lines, nesting_state, leading_text)):
-      error(filename, linenum, 'whitespace/braces', 5,
-            'Missing space before {')
-
   # You shouldn't have a space before a semicolon at the end of the line.
   # There's a special case for "for" since the style guide allows space before
   # the semicolon there.
--- a/src/util/json.cpp
+++ b/src/util/json.cpp
@@ -9,6 +9,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "json.h"
 
 #include <ostream>
+#include <algorithm>
 
 const jsont jsont::null_json_object(jsont::kindt::J_NULL);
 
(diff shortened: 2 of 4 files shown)


src/util/json.cpp
```diff
@@ -162,3 +164,49 @@ void jsont::swap(jsont &other)
...
+           std::equal(
+             left_array.begin(), left_array.end(), right_array.begin());
+  }
+  case jsont::kindt::J_OBJECT:
+  {
+    const auto &left_object = static_cast<const json_objectt &>(left);
+    const auto &right_object = static_cast<const json_objectt &>(right);
+    const auto yes = [](const std::pair<std::string, jsont> &) { return true; };
+    const size_t left_size = narrow<size_t>(
+      std::count_if(left_object.begin(), left_object.end(), yes));
```
thomasspriggs: :grey_question: Can this be done using `std::count`, so that you don't need to define `yes`?

thomasspriggs (approved):
Looks useful for writing unit tests.

src/util/json.cpp
smowton: `std::distance(left_object.begin(), left_object.end())`

smowton: Or since `object` is `std::map`, just use `.size()`. At present since it is `map` you could also use `zip` and safely assume the keys occur in the same order, if guarded by a static-assert checking that `typeof(object)` really is an (ordered) map.

LAJW: `json_objectt` is not a `map`, it just has some `map`-like methods exposed. `size` sadly is not one of them.
`zip` lives in test-gen AFAIK.
I don't want to rely on the order of elements in the map. We've been shifting towards using `unordered_map` in many places, and JSON looks like it could be one of them.

LAJW: `std::count` counts occurrences of a specified element, so no.

smowton: `json_objectt` is not, but its member `object` is. `zip` lives in https://github.com/diffblue/cbmc/blob/develop/src/util/range.h

smowton: Sure, but if you guard with an appropriate `static_assert` (ideally there would be some type trait `is_ordered`) then you could be sure this code will be changed as/when we do that

LAJW: `object` is a protected property.

There isn't an `is_ordered` trait. Being "ordered" is a runtime property.

If I didn't have to worry about future proofing - `zip` is unnecessary - `std::equal(b1, e1, b2)` is the tool for this job.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: cfe09dccb93f4a913a02c5bc7baabc277c58cf6e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119292309

src/util/json.cpp
```diff
@@ -7,8 +7,10 @@ Author: Daniel Kroening, kroening@kroening.com
 \*******************************************************************/
 
 #include "json.h"
+#include <util/narrow.h>
```
tautschnig: For consistency: `#include "narrow.h"`

src/util/json.cpp
```diff
@@ -162,3 +164,48 @@ void jsont::swap(jsont &other)
...
+  case jsont::kindt::J_NULL:
+    return true;
+  case jsont::kindt::J_TRUE:
+    return true;
+  case jsont::kindt::J_FALSE:
+    return true;
+  case jsont::kindt::J_NUMBER:
+    return left.value == right.value;
+  case jsont::kindt::J_STRING:
+    return left.value == right.value;
```
tautschnig, line 182: Overall design question: shouldn't we do this more in a visitor-style where each call deriving from `jsont` defines each own `operator==` and the switch-case here would only do the downcasts and then invoke the more specific equality operators?

src/util/json.cpp
```diff
@@ -162,3 +164,48 @@ void jsont::swap(jsont &other)
...
+             left_array.begin(), left_array.end(), right_array.begin());
+  }
+  case jsont::kindt::J_OBJECT:
+  {
+    const auto &left_object = static_cast<const json_objectt &>(left);
+    const auto &right_object = static_cast<const json_objectt &>(right);
+    const size_t left_size =
+      narrow<size_t>(std::distance(left_object.begin(), left_object.end()));
+    const size_t right_size =
+      narrow<size_t>(std::distance(right_object.begin(), right_object.end()));
```
tautschnig: Why do those conversions when afterwards you're only going to use the values for comparing them to each other? I'd go with `const auto left_size = std::distance ...`

src/util/json.cpp
```diff
@@ -162,3 +164,48 @@ void jsont::swap(jsont &other)
...
+    if(left_size != right_size)
+      return false;
+    return std::all_of(
+      left_object.begin(),
+      left_object.end(),
+      [&](const std::pair<std::string, jsont> &pair) {
+        return right_object[pair.first] == pair.second;
```
tautschnig, line 201: That's an unnecessary lookup when instead you could just use two iterators over the two containers.

unit/json/json_parser.cpp
```diff
@@ -133,3 +133,110 @@ SCENARIO("Loading JSON files")
     }
   }
 }
+
+TEST_CASE("json equality", "[core][test-gen-util][state][type]")
```
tautschnig: I actually don't know what those [...][...] parts are good for at all, but this particular entry ("test-gen-util") strikes me as information of no value in this repository.

scripts/cpplint.py
```diff
@@ -3974,13 +3974,6 @@ def CheckBracesSpacing(filename, clean_lines, linenum, nesting_state, error):
     for offset in xrange(endlinenum + 1,
                          min(endlinenum + 3, clean_lines.NumLines() - 1)):
       trailing_text += clean_lines.elided[offset]
-    # We also suppress warnings for `uint64_t{expression}` etc., as the style
-    # guide recommends brace initialization for integral types to avoid
-    # overflow/truncation.
-    if (not Match(r'^[\s}]*[{.;,)<>\]:]', trailing_text)
-        and not _IsType(clean_lines, nesting_state, leading_text)):
-      error(filename, linenum, 'whitespace/braces', 5,
-            'Missing space before {')
```
tautschnig, line 3983: Maybe removing this is the right thing to do, but then we should remove all the code computing `leading_text`, `trailing_text` above. Or maybe you should just use `NOLINTNEXTLINE(whitespace/braces)` in the test.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1a7ca85632241d0c7a41d6c4afd412b5c36dd05a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119297176

src/util/json.cpp
smowton: At least expose a `size` property on `json_objectt` rather than hacking around it with the existing interface. The ordered vs. unordered thing I'll pass on.

src/util/json.cpp
LAJW: Yes, except `jsont` much like `irept` doesn't allow any virtual functions. Therefore the visitor pattern is not applicable.

src/util/json.cpp
LAJW: Yeah, but if someone decides to swap `std::map` for `std::unordered_map`, it could be difficult to debug, so this is more stable/doesn't rely on any invariants.

scripts/cpplint.py
LAJW: `clang-format` already checks for positions and spacing of curly braces and parentheses. The problem stems from the fact this isn't aware of initializer lists/uniform initialization.
```c++
std::vector<std::pair<int, int>>{{1, 2}, {3, 4}, {5, 6}};
```
In the days of C++98, the above wasn't a valid syntax. Curly braces would appear only after `if` /`for` statements, as block scopes or surround function bodies. The linter expects this to still be the case.

src/util/json.cpp
tautschnig: Of course, hence me saying "visitor-style." But also my comment was intended as a thought, not any kind of hard requirement.

src/util/json.cpp
tautschnig: I do wonder which containers would exhibit differences in strict equality testing? But I suppose it is conceivable that two `std::unordered_map` had their elements constructed in a different order, yet should compare equal. So, yes, keep it as is.

scripts/cpplint.py
tautschnig: Ok, in which case the linter should be cleaned properly so as not to leave in place unused code.

src/util/json.cpp
smowton: Personally I still favour 
```
// This relies on ordered maps:
static_assert(std::is_same<typeof(jsont::object), std::map<std::string, jsont>>::value)
```

LAJW:
Done!
 - Exposed `jsont::object::size` as `json_objectt::size`
 - Used that instead of `std::distance`
 - Removed dead linter code

smowton (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 2f63cb81329a35b4b75a7de7eec980adff4f8bfd).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119350287

src/util/json.cpp
```diff
@@ -162,3 +163,44 @@ void jsont::swap(jsont &other)
...
+    return true;
+  case jsont::kindt::J_FALSE:
+    return true;
+  case jsont::kindt::J_NUMBER:
+    return left.value == right.value;
+  case jsont::kindt::J_STRING:
+    return left.value == right.value;
+  case jsont::kindt::J_ARRAY:
+  {
+    const auto &left_array = static_cast<const json_arrayt &>(left);
```
thomasspriggs, line 185: :information_source: There is a function called `to_json_array` at the bottom of this file, which you could re-use instead of adding your own static casts.

LAJW: `static_cast` is a feature of C++, and not "my" cast. `to_json_array` checks for kind, which in this case is already checked. We'd be doing extra work, so no.

thk123 (approved):
Code owner review of the cpplint change only (though the other changes lgtm)

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c70cbabf2ed0daeb86d385306d6fb38f9a0bb8eb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119476532

PR #4913, opened 2019-07-15
[TG-7815] Fix bug in get_recursively_instantiated_type

NathanJPhillips:
This reduces the amount of code and in so doing fixes the implementation of specialization of generics.

Previously the following invariant was being triggered in non-symmetric mutual generic specialization:
```
Invariant check failed
File: .../lib/cbmc/jbmc/src/java_bytecode/select_pointer_type.cpp:187 function: get_recursively_instantiated_type
Condition: depth < replacements.size()
Reason: cannot access elements outside stack
```

- [x] Regression or unit tests are included in a downstream PR.

### DIFF
--- a/jbmc/src/java_bytecode/java_types.h
+++ b/jbmc/src/java_bytecode/java_types.h
@@ -735,13 +735,23 @@ public:
   }
 };
 
+/// Check whether a reference to a typet is a java generic parameter/variable,
+/// e.g., `T` in `List<T>`.
+/// \param type: Source type.
+/// \return True if \p type is a generic Java parameter type
+template <>
+inline bool can_cast_type<java_generic_parametert>(const typet &base)
+{
+  return is_reference(base) && is_java_generic_parameter_tag(base.subtype());
+}
+
 /// Checks whether the type is a java generic parameter/variable, e.g., `T` in
 /// `List<T>`.
 /// \param type: the type to check
 /// \return true if type is a generic Java parameter type
 inline bool is_java_generic_parameter(const typet &type)
 {
-  return is_reference(type) && is_java_generic_parameter_tag(type.subtype());
+  return can_cast_type<java_generic_parametert>(type);
 }
 
 /// \param type: source type
--- a/jbmc/src/java_bytecode/select_pointer_type.cpp
+++ b/jbmc/src/java_bytecode/select_pointer_type.cpp
@@ -126,86 +126,24 @@ pointer_typet select_pointer_typet::specialize_generics(
 ///   such instantiation exists.
 optionalt<pointer_typet>
 select_pointer_typet::get_recursively_instantiated_type(
-  const irep_idt &parameter_name,
-  const generic_parameter_specialization_mapt
-    &generic_parameter_specialization_map) const
+  irep_idt parameter_name,
+  generic_parameter_specialization_mapt
+    generic_parameter_specialization_map) const
 {
-  generic_parameter_recursion_trackingt visited;
-  const size_t max_depth =
-    generic_parameter_specialization_map.find(parameter_name)->second.size();
-
-  irep_idt current_parameter = parameter_name;
-  for(size_t depth = 0; depth < max_depth; depth++)
+  while(true)
   {
-    const auto retval = get_recursively_instantiated_type(
-      current_parameter, generic_parameter_specialization_map, visited, depth);
-    if(retval.has_value())
-    {
-      CHECK_RETURN(!is_java_generic_parameter(*retval));
-      return retval;
-    }
-    CHECK_RETURN(visited.empty());
-
-    const auto &entry =
-      generic_parameter_specialization_map.find(current_parameter)
-        ->second.back();
-    current_parameter = to_java_generic_parameter(entry).get_name();
+    auto types_it = generic_parameter_specialization_map.find(parameter_name);
+    INVARIANT(
+      types_it != generic_parameter_specialization_map.end(),
+      "Type parameter must have a type argument");
+    if(types_it->second.empty())
+      return {};
+    reference_typet type = types_it->second.back();
+    types_it->second.pop_back();
+    if(!is_java_generic_parameter(type))
+      return std::move(type);
+    parameter_name = to_java_generic_parameter(type).get_name();
   }
-  return {};
-}
-
-/// See get_recursively instantiated_type, the additional parameters just track
-/// the recursion to prevent visiting the same depth again and specify which
-/// stack depth is analyzed.
-/// \param parameter_name: The name of the generic parameter type we want to
-///   have instantiated
-/// \param generic_parameter_specialization_map: Map of type names to
-///   specialization stack
-/// \param visited: Tracks the visited parameter names
-/// \param depth: Stack depth to analyze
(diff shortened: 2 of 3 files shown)


jbmc/src/java_bytecode/select_pointer_type.cpp
```diff
@@ -126,86 +126,24 @@ pointer_typet select_pointer_typet::specialize_generics(
...
-    return {};
-  }
-
-  const size_t index = (replacements.size() - 1) - depth;
-  const auto &type = replacements[index];
-
-  if(!is_java_generic_parameter(type))
-  {
-    return type;
+    auto types_it = generic_parameter_specialization_map.find(parameter_name);
```
smowton: Commit message: note what the bug was and how this fixes it

jbmc/src/java_bytecode/select_pointer_type.cpp
```diff
@@ -126,86 +126,24 @@ pointer_typet select_pointer_typet::specialize_generics(
...
+  generic_parameter_specialization_mapt
+    generic_parameter_specialization_map) const
 {
-  generic_parameter_recursion_trackingt visited;
-  const size_t max_depth =
-    generic_parameter_specialization_map.find(parameter_name)->second.size();
-
-  irep_idt current_parameter = parameter_name;
-  for(size_t depth = 0; depth < max_depth; depth++)
+  while(true)
```
smowton: Looks good, but since taking a map by value like this is quite unusual please add comments noting the strategy you're using here and why

jbmc/src/java_bytecode/select_pointer_type.cpp
```diff
@@ -82,26 +81,27 @@ pointer_typet select_pointer_typet::specialize_generics(
...
-          to_pointer_type(array_element_type),
-          generic_parameter_specialization_map,
-          visited_nodes);
-
-        pointer_typet replacement_array_type = java_array_type('a');
-        replacement_array_type.subtype().set(ID_element_type, new_array_type);
-        return replacement_array_type;
-      }
-    }
+    auto array_element_type =
```
smowton, line 80: `const auto *` for clarity?

jbmc/src/java_bytecode/select_pointer_type.cpp
```diff
@@ -82,26 +81,27 @@ pointer_typet select_pointer_typet::specialize_generics(
...
+    const pointer_typet &new_array_type = specialize_generics(
+      *array_element_type,
+      generic_parameter_specialization_map,
+      visited_nodes);
+
+    pointer_typet replacement_array_type = java_array_type('a');
+    replacement_array_type.subtype().set(ID_element_type, new_array_type);
+    return replacement_array_type;
   }
+
```
smowton, line 93: Don't re-space in the same commit as a substantial change

smowton (commented):
Also please add at least one regression test exposing the bug (or better yet, KNOWNBUG -> CORE an existing test), and there seem to be real unit test failures in CI.

jbmc/src/java_bytecode/select_pointer_type.cpp
allredj: The map itself is not taken by value. I don't think we need to document why passing a struct by value is OK.

jbmc/src/java_bytecode/java_types.h
```diff
@@ -735,13 +735,23 @@ class java_generic_parametert : public reference_typet
...
+}
+
 /// Checks whether the type is a java generic parameter/variable, e.g., `T` in
 /// `List<T>`.
 /// \param type: the type to check
 /// \return true if type is a generic Java parameter type
 inline bool is_java_generic_parameter(const typet &type)
 {
-  return is_reference(type) && is_java_generic_parameter_tag(type.subtype());
+  return can_cast_type<java_generic_parametert>(type);
```
allredj, line 754: Should probably be in a separate commit. It does not correspond to the commit description. There are several things being changed here, and it makes it difficult to check that this commit does not introduce a behavioural change.

jbmc/src/java_bytecode/select_pointer_type.h
```diff
@@ -23,6 +24,32 @@ typedef std::unordered_map<irep_idt, std::vector<reference_typet>>
...
 
+struct print_generic_parameter_specialization_map
+{
+  const generic_parameter_specialization_mapt &map;
+  const namespacet &ns;
+};
+
+template <typename ostream>
+ostream &
+operator<<(ostream &stm, const print_generic_parameter_specialization_map &map)
```
majakusber, line 34: :ok_hand: This is useful

allredj (commented):
Should be fine once the tests are in.

jbmc/src/java_bytecode/select_pointer_type.cpp
majakusber: I'm not sure what you mean by "not taken by value". We do make a copy here, so we can pop from the stacks of particular parameters in the map just for the sake of finding the parameter specialization, but at the call site of this function we need to keep the original full information. I agree that adding a note that points this out is useful (not arguing why it's ok, just to make the reader notice), I missed it myself at first.

jbmc/src/java_bytecode/select_pointer_type.cpp
```diff
@@ -25,61 +25,60 @@ pointer_typet select_pointer_typet::convert_pointer_type(
...
+/// - T -> (Integer, U)
+/// - U -> (String, T)
+///
+/// Note that the top of the stacks for T and U create a recursion T -> U,
+/// U-> T. We want to break it and specialize `T my.gen.t` to `String my.gen.t`.
+/// \param pointer_type: The pointer type to specialize
+/// \param generic_parameter_specialization_map: Map of type names to
+///   specialization stack
+/// \return The first instantiated type for the generic type or nothing if no
+///   such instantiation exists.
```
majakusber, line 43: This should be updated slightly - it sounds l like it's only handling the recursion case (because it's just moved from the function that was breaking recursion). Also, move the documentation to the header.

jbmc/src/java_bytecode/select_pointer_type.cpp
```diff
@@ -25,61 +25,60 @@ pointer_typet select_pointer_typet::convert_pointer_type(
...
   if(parameter != nullptr)
   {
-    const irep_idt &parameter_name = parameter->get_name();
+    irep_idt parameter_name = parameter->get_name();
 
-    // avoid infinite recursion by looking at each generic argument from
-    // previous assignments
-    if(visited_nodes.find(parameter_name) != visited_nodes.end())
+    generic_parameter_specialization_mapt spec_map_copy =
+      generic_parameter_specialization_map;
```
majakusber, line 55: :bulb: You're making a copy of the map over and over again, changing a single element in it in each call. I don't have any good idea how to make this less costly though.. not so nice ideas being keeping two functions, make the code uglier at the call site, or add extra arguments to the function (depth of each stack to look at). Can you think of a better way?

majakusber (commented):
Needs some cleanup but I love it :slightly_smiling_face:

NathanJPhillips:
This solution was inadequate. Replacing with a new solution.

PR #4914, opened 2019-07-17
Include SSA lhs and rhs in extended JSON trace

tautschnig:
The full SSA identifiers will permit building tools that trace
assigned values back to inputs.

### DIFF
--- a/src/goto-programs/goto_trace.h
+++ b/src/goto-programs/goto_trace.h
@@ -118,6 +118,9 @@ public:
   bool cond_value;
   exprt cond_expr;
 
+  // for assignments and declarations (ssa_lhs only)
+  exprt ssa_lhs, ssa_rhs;
+
   // for assert
   irep_idt property_id;
   std::string comment;
@@ -161,6 +164,8 @@ public:
     full_lhs.make_nil();
     full_lhs_value.make_nil();
     cond_expr.make_nil();
+    ssa_lhs.make_nil();
+    ssa_rhs.make_nil();
   }
 };
 
--- a/regression/cbmc/trace_options_json_extended/extended.desc
+++ b/regression/cbmc/trace_options_json_extended/extended.desc
@@ -4,4 +4,6 @@ test.c
 ^EXIT=10$
 ^SIGNAL=0$
 rawLhs
+"ssaLhs": "main::argc!0@1#1",
+"ssaRhs": "argc'#0",
 --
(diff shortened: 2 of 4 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7a4135f10708e0ef5761436b301a2555c0070af5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119404690

src/goto-programs/goto_trace.h
```diff
@@ -118,6 +118,9 @@ class goto_trace_stept
   bool cond_value;
   exprt cond_expr;
 
+  // for assignments
```
smowton: ...and declarations (lhs only)

smowton: Suggest naming these `ssa_lhs/rhs` to make clear what distinguishes them from `full_lhs` and `full_lhs_value` below.

tautschnig: Good points, all done as suggested!

smowton (commented):
Let's have a couple of regression tests exercising this, then lgtm

tautschnig:
> Let's have a couple of regression tests exercising this, then lgtm

Which aspects specifically would you like to see tested? Right now, there is one test checking "yes, something is printed." Since the consumer is not yet very well-defined (tentatively cbmc-viewer) I'm not entirely sure what the expectation is.

smowton:
Given the support you've added, I'd expect a program with only a declaration in addition to the assignment already tested; other than that as you say the requirements are pretty weak so far

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 754b2455e778cdeced022294861c14488e6c3243).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119481318

kroening:
Note that we used to have these fields.

Part of the rationale for removing them was the idea that a "goto trace" should be an artefact that is just that -- and independent of _how_ it got computed. In particular, a goal was that it should be possible to produce one using a simulator or plain execution.

I am wondering whether the higher-level goal (dependency analysis) can perhaps be achieved in other ways?

tautschnig:
> I am wondering whether the higher-level goal (dependency analysis) can perhaps be achieved in other ways?

@kroening  All ideas welcome - it's about ways to understand how we arrived at a certain value being assigned. With CBMC's scalability improving, counterexamples get longer and thus become less useful as a diagnostic tool. This is just what seemed like the simplest step.

tautschnig:
Follow-up note: maybe `--localize-fault` is the way to go, but then it needs to be made to work. At present, its regression tests are the only cases where I am not getting "unable to localize fault."

martin-cs:
All ideas... well... we could implement an `ai_history_baset` child that uses the `goto_trace` to control the abstract interpreter and use it's dependency and information flow tracking.

PR #4917, opened 2019-07-17
Rename LHS rvalues up front

smowton:
The rough idea: instead of recursing over the structure of an assignment LHS, then performing post-hoc adjustments to permit field-sensitivity to have maximum effect, perform all expression simplification up to LHS field-sensitivity application right at the start of `symex_assign`. This both simplifies the `symex_assign` workflow, as `shift_indexed_access_to_lhs` and `rewrite_with_to_field_symbols` are no longer needed, and makes maintaining `l2_full_lhs` simpler, as the expression skeleton no longer needs to undergo any complicated transformations.

There is one down-side: `with` expressions *not* introduced by `goto-symex` are no longer handled. If they are only ever really produced within goto-symex then this is no loss -- the simplification and field-sensitivity application ahead of `symex_assign` achieves the same thing. If we expect to handle GOTO programs with `with_exprt` already present then we should do something a little cleverer.

This of course needs tests (or existing tests should be reviewed) to ensure we still have field-sensitivity in all the cases we intend.

Test PR for this + array-cell sensitivity: https://github.com/diffblue/cbmc/pull/4918

---

Long version: Rename LHS rvalues up front

This was already happening in most cases (e.g. symex_dereference would apply field-sensitivity
before symex_assign_rec was entered), but the case of a statically non-constant but dynamically
constant array index or byte-extract offset would only be handled *after* symex_assign_rec, leading
to an asymmetry in which operations were combined into the ssa_exprt and which ones accumulated in
the expr_skeletont.

With this change the LHS expression is maximally renamed and simplified before symex_assign_rec is
entered, which means that any field-sensitive symbols on the LHS are fully constructed as early as
possible, and expr_skeletont only accumulates those member, index and byte-extract operations which
*cannot* be associated with an ssa_exprt. This means there is no need to undo with-operations or
try to simplify byte-extract operations in goto_symext::assign_from_non_struct_symbol, as this has
been taken care of already, and the l2_full_lhs can be constructed from the expression skeleton
trivially.


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 167f5623e195a39789d5cea977127d4aca876551).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119511318

kroening:
I agree with the approach -- should this be rename_rhs, since 'rvalue' has a slightly different meaning since C++ has rvalue references? For symmetry, have rename_lhs?

@tautschnig ?

smowton:
Added tests as https://github.com/diffblue/cbmc/pull/4944, which this PR passes. I regard this as ready to go once those tests are in.

peterschrammel (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c4bc80029d83c4e7c572162345065c74cdcce5bf).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120656593

smowton:
@kroening re: naming, it's definitely not `rename_rhs`, since this acts specifically on an lvalue to rename components of it which are read (e.g. array and byte-extract indices, and ternary-conditional condition operands). Not wedded to `rvalues` as a term for these though.

smowton:
Perhaps just `l2_rename_lvalue`, which can call regular old `rename` on things which are read and `l2_rename_lvalue` on sub-lvalues?

smowton:
@andreast271 I hear from email relayed via @kroening that the out-of-bounds issue was not introduced by that PR, is that accurate? If so are you happy for this to go in?

andreast271:
@smowton  That is accurate and I posted about it on [4933](https://github.com/diffblue/cbmc/issues/4933#issuecomment-515004238). I would be very happy for this PR to go in. It makes a huge positive difference for my use case.

smowton:
So you did, I'm not sure why I didn't get email about that. Polling @kroening for a final approval here.

Issue #4882, opened 2019-07-08
Field sensitivity appears to have broken display of counterexamples


hannes-steffenhagen-diffblue:
CBMC version: develop (acb3abaf5cd863c4e37df0a79db7680c21037a27)
Operating system: Linux (Ubuntu 18.04)
Exact command line resulting in the issue: cbmc --trace test.c
What behaviour did you expect: It should show a counterexample with the field set to a non-0 value
What happened instead: The counterexample had its field set to 0

Here we have a test with a failing assertion:

```
//test.c
struct SomeStruct
{
  int x;
};

int main(void)
{
  struct SomeStruct x;
  assert(x.x == 0);
}
```

`cbmc` finds the problem (`x.x` can be non-zero) just fine, but in the produced trace we have

```
State 21 file test.c function main line 8 thread 0
----------------------------------------------------
  x={ .x=0 } ({ 00000000 00000000 00000000 00000000 })

Violated property:
  file test.c function main line 9 thread 0
  assertion x.x != 0
  x.x != 0
```

A git bisect revealed that this issue was introduced in 6e514deeebdbd47493d218c37ef5692371bd7591.

**EDIT**: issue had it wrong before, the problem appears with `assert(x.x == 0)`, not `!=`

smowton:
Fixed by #4964

PR #4883, opened 2019-07-08
Add test demonstrating bug in field sensitivity

hannes-steffenhagen-diffblue:
This adds a KNOWNBUG test demonstrating an issue where struct fields in counterexamples seem to have nonsensical values (apparently always 0). See also #4882

- [ ] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4919, opened 2019-07-17
Symex trace: show implicit value associated with composite declarations

smowton:
Previously when a struct- or other composite-typed variable was declared, symex_decl
would initialize its individual fields, then return an L2-renamed symbol relating to
the whole object. With field-sensitivity enabled, the returned symbol is e.g. my_struct!0@1#0,
while anybody reading from it will read my_struct!0@1#1..field, a different symbol which
to the solver is unrelated. The result was that in the trace, the declaration would show
up with the value of the free symbol my_struct!0@1#0 (most likely zero, since it can't be
referred to), while the rest of the trace may make it obvious that it actually had some other
non-zero value.

With this change we expand the struct into its constituent field symbols before recording an
SSA_stept for the trace. This restores the correspondence between what the trace reads and what
code reading from the struct reads.

Argument for the reviewers to have: should the `goto_symex_statet::rename` in `symex_decl` already be expanding like this? That function has an invariant insisting that it should return a single symbol, not a struct-exprt of field symbols, but that may not really be necessary. Note that if we change `::rename` to always perform struct expansion then we will probably break any user that is applying this to an lvalue.

### DIFF
--- a/src/goto-symex/symex_decl.cpp
+++ b/src/goto-symex/symex_decl.cpp
@@ -54,10 +54,10 @@ void goto_symext::symex_decl(statet &state, const symbol_exprt &expr)
   target.decl(
     state.guard.as_expr(),
     ssa,
+    state.field_sensitivity.apply(ns, state, ssa, false),
     state.source,
-    hidden?
-      symex_targett::assignment_typet::HIDDEN:
-      symex_targett::assignment_typet::STATE);
+    hidden ? symex_targett::assignment_typet::HIDDEN
+           : symex_targett::assignment_typet::STATE);
 
   if(path_storage.dirty(ssa.get_object_name()) && state.atomic_section_id == 0)
     target.shared_write(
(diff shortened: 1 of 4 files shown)


hannes-steffenhagen-diffblue:
This should fix #4882

hannes-steffenhagen-diffblue (approved):
This seems like a reasonable thing to do to me, but I'd prefer if people more familiar with this part of the code could chime in.

smowton:
@hannes-steffenhagen-diffblue since this is your fix, would you mind taking charge of finding reviewers and addressing their comments?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7c6370adbe249db1bb4537ac1933223e75b3ef5e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120255611

peterschrammel (approved):
@hannes-steffenhagen-diffblue, what was the failing test? Please ping me to check how to upstream it.

smowton:
@hannes-steffenhagen-diffblue you should probably also push a couple of tests representative of the problems reported to you by customers

hannes-steffenhagen-diffblue:
Sorry for the late reply, this PR over here has a KNOWNBUG test for what's being fixed here (I did check, and the problem does indeed go away):

https://github.com/diffblue/cbmc/pull/4883

I'm not sure if there are other situations in which it can happen, but I don't think so.

smowton:
OK, suggest in that case you get that merged and I'll flip this to CORE

smowton:
Continued in #4964

PR #4920, opened 2019-07-17
Make goto_rw less conservative

karkhaz:
When handling the case of pointers, goto_rw would mark both the object pointed
to and the pointer itself as written. For example, if x = &i, the line `*x = 1'
would yield a write set containing both x and i, instead of just i.

This resolves that issue by ensuring that value_set_dereference always gives at
least one object that the dereference can refer to.

This also resolves a bug wherein &a is marked as reading from a by removing the
ID_symbol special case from get_objects_address_of.

### DIFF
--- a/regression/goto-instrument/goto_rw_pointer_handling-2/main.c
+++ b/regression/goto-instrument/goto_rw_pointer_handling-2/main.c
@@ -1,0 +1,6 @@
+int main()
+{
+  char a = 'a';
+  char *b = &a;
+  *b = 'c';
+}
--- a/regression/goto-instrument/goto_rw_pointer_handling-2/test.desc
+++ b/regression/goto-instrument/goto_rw_pointer_handling-2/test.desc
@@ -1,0 +1,16 @@
+CORE
+main.c
+--show-dependence-graph
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+^Data dependencies: \d+$
+--
+^warning: ignoring
+^Data dependencies: \d+,\s*\d+$
+--
+This tests that the write-set of a pointer contains only the pointed-to
+object when the pointer is dereferenced and written to.
+
+There should therefore be no data dependence between the pointer itself
+and the write to the pointed-to object.
--- a/src/analyses/goto_rw.cpp
+++ b/src/analyses/goto_rw.cpp
@@ -429,17 +429,13 @@ void rw_range_sett::get_objects_typecast(
 
 void rw_range_sett::get_objects_address_of(const exprt &object)
 {
-  if(object.id() == ID_string_constant ||
-     object.id() == ID_label ||
-     object.id() == ID_array ||
-     object.id() == ID_null_object)
+  if(
+    object.id() == ID_string_constant || object.id() == ID_label ||
+    object.id() == ID_array || object.id() == ID_null_object ||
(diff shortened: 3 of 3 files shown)


karkhaz:
(CI passing is currently blocked on  #4922)

A couple of questions, not sure who the right person to answer this is:

this looks like a performance improvement, not something that would actually affect the verification result---right? In which case I'm not sure if I should write a test, or how I would do so.

There is a TODO in the code, not sure whether to leave it in or what I might do about it.

thank you!

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 872cc8c486243fff82fd724c32b6a6026fb3c086).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119522812

smowton:
Let's have some regression tests illustrating the change? This is no worry for symex, as it doesn't use dynamic objects this way (it makes up symbols for them instead)

src/analyses/reaching_definitions.cpp
```diff
@@ -321,20 +321,17 @@ void rd_range_domaint::transform_assign(
...
 
     const range_domaint &ranges=rw_set.get_ranges(it);
 
-    if(is_must_alias &&
-       (!rd.get_is_threaded()(from) ||
-        (!symbol_ptr->is_shared() &&
-         !rd.get_is_dirty()(identifier))))
+    if(
+      is_must_alias && (!rd.get_is_threaded()(from) ||
+                        (symbol_ptr != nullptr && symbol_ptr->is_shared() &&
```
tautschnig: Maybe we should do `(!not_found && symbol_ptr->is_shared() && ...`

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -398,8 +398,12 @@ value_set_dereferencet::valuet value_set_dereferencet::build_reference_to(
...
     result.pointer_guard = dynamic_object(pointer_expr);
 
-    // can't remove here, turn into *p
-    result.value = dereference_exprt{pointer_expr};
+    // TODO should this be object or root_object?
+    // TODO It's unclear whether this is a good approach to take --- it
+    // successfully ensures that every (non-null) dereference points to
+    // something, making the write set computation work correctly, but dynamic
+    // objects do not seem to be intended to be used in this way.
+    result.value = object;
```
tautschnig: The problem is the offset, which might not be known. I believe this should be correct when `o.offset().is_constant() == true`. I think in all other cases you largely need to fall back to what the `byte_extract` base does below.

tautschnig (changes requested):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 29f8f6986041d0b61e670dafb74dfc3be2c86308).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122637868

karkhaz:
@smowton thanks. Some of the patch was causing a regression, so I've included the most relevant part and added a test case. Thanks to @tautschnig for help with the test.

smowton (approved):

karkhaz:
@peterschrammel may I have a review of this please? Although I won't be able to merge until @tautschnig gets back and approves...

karkhaz:
@tautschnig are you happy with the changes that I've made?

tautschnig (approved):

PR #4921, opened 2019-07-17
Add a __CPROVER_postcondition builtin

angelhof:
Add a builtin __CPROVER_postcondition builtin function that internally becomes an assert (like __CPROVER_precondition).

This will be used (by a later PR) to turn function preconditions, and postconditions to code contracts.


smowton:
Please add a regression test exercising the new code. There actually don't appear to be any exercising `__CPROVER_precondition` either, so if you're feeling generous add that too!

angelhof:
> Please add a regression test exercising the new code. There actually don't appear to be any exercising `__CPROVER_precondition` either, so if you're feeling generous add that too!

@smowton  Thanks, I will add a regression test for both. On a different note, why is CPP-LINT in travis complaining only for the newly added line and not for the ones above that use `__CPROVER_`?

owen-mc-diffblue:
I think CPP-LINT in travis only checks newly added lines. This is so that things can be slowly fixed as we go, rather than having to fix them all at once so that we can see if we add any new ones.

angelhof:
> I think CPP-LINT in travis only checks newly added lines. This is so that things can be slowly fixed as we go, rather than having to fix them all at once so that we can see if we add any new ones.

Thanks. Shouldn't this check in [cpplint.py](https://github.com/diffblue/cbmc/blob/develop/scripts/cpplint.py#L4617) only check for uses of `__CPROVER_` in strings? I am not aware of any way to define a function using a string literal (CPROVER_PREFIX).

owen-mc-diffblue:
@tautschnig, you wrote that check in cpplint.py (last October), can you help @angelhof?

smowton:
Here's the check: https://github.com/diffblue/cbmc/blob/develop/scripts/cpplint.py#L4607

I would suggest simply whitelisting paths beginning with `regession/ or `src/ansi-c/library/` for that particular check. Hard-coding that should be fine given the check itself is our own custom invention.

angelhof:
> Here's the check: https://github.com/diffblue/cbmc/blob/develop/scripts/cpplint.py#L4607
> 
> I would suggest simply whitelisting paths beginning with `regession/ or `src/ansi-c/library/` for that particular check. Hard-coding that should be fine given the check itself is our own custom invention.

Wouldn't something like `r'\".*__CPROVER_.*\"'` make more sense? Since (from what I understand) this check suggests using CPROVER_PREFIX in string literals and not in general (e.g. function names).

smowton:
That precise regex would misfire for `f("Hello", __CPROVER_blah(y, z), "world")`, but looking for precisely `"__CPROVER_` would probably do the trick 90% of the time. I'd be happy with either solution.

angelhof:
@smowton I don't have any preference between the two. I could implement any of them in a separate PR. Is this PR blocked until the lint check succeeds?

smowton:
Yes without administrative intervention. I'd recommend making the change as a second commit in this PR.

angelhof:
Could someone review this?

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 44197bbbb2d25bf8df72dba7f0a968891c7b417e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120852254

smowton (approved):

owen-mc-diffblue (approved):

angelhof:
@peterschrammel could you review this?

kroening (approved):

angelhof:
@peterschrammel @thk123 could you review this PR?

thk123 (approved):
Approving the changes to cpplint only - seems like there has been more than enough eyes on the rest!

src/ansi-c/cprover_builtin_headers.h
```diff
@@ -2,6 +2,7 @@ void __CPROVER_assume(__CPROVER_bool assumption);
 void __VERIFIER_assume(__CPROVER_bool assumption);
 void __CPROVER_assert(__CPROVER_bool assertion, const char *description);
 void __CPROVER_precondition(__CPROVER_bool precondition, const char *description);
+void __CPROVER_postcondition(__CPROVER_bool assertion, const char *description);
```
tautschnig, line 5: Random place for a comment: I think this strictly requires documentation. When should users make use of this? I note that the commit message doesn't have a body either, so it's really hard to see when that should be used.

scripts/cpplint.py
```diff
@@ -4614,7 +4614,7 @@ def Check__CPROVER_(filename, clean_lines, linenum, error):
     error: The function to call with any errors found.
   """
   line = clean_lines.lines[linenum]
-  match = Match(r'.*__CPROVER_.*', line)
+  match = Match(r'.*\"__CPROVER_.*\"', line)
```
tautschnig: I would appreciate if the commit message could have a body explaining why this change is right.

tautschnig (changes requested):
The empty commit bodies plus the absence of other documentation make this hard for anyone to use effectively. Otherwise looks good to me.

src/ansi-c/cprover_builtin_headers.h
angelhof: Thanks. Where is a good place to add documentation for this?

tautschnig: As discussed elsewhere: `doc/cprover-manual/cbmc-assertions.md` is likely the best place for this. @kroening Would you be able to contribute some documentation of `__CPROVER_precondition`, which AFAIK never got documented?

kroening: I would structure this differently.
`__CPROVER_precondition` was introduced as a means to add preconditions to library functions, i.e., it was never meant to be used by anyone but CBMC developers.

The new feature, which is contract checking, is a different use-case, and externalises this to users of CBMC. My experience is that users like manuals that are structured by use-case. Thus, I would suggest to add a new chapter, say titled "Modular Verification using Code Contracts", or the like, to discuss this use-case separately.

tautschnig: > [...] never meant to be used by anyone but CBMC developers.

While this is a good intention, I don't think it will remain as constrained if people find it useful in other places as well. Anything that _can_ be used _will_ be used at some point...

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 59ab9e94b677208fa824e08572e9c1d32ef3222c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122362598

scripts/cpplint.py
angelhof: I made the lint change a separate PR, together with the removal of some NOLINT comments. #5000

angelhof:
The lint error is fixed in #5000.

doc/cprover-manual/cbmc-assertions.md
```diff
@@ -59,6 +59,34 @@ for(i=0; i<100; i++)
   assert(a[i]==0);
 ```
 
+CPROVER also supports writing function pre and postconditions, using
+the built-in functions `__CPROVER_precondition` and
+`__CPROVER_postcondition`. They can be used to express intent, and at
+the moment they are just transformed to assertions in the goto
+program. As such, they can be used as simple assertions in
+code. However, it is advised to use `__CPROVER_precondition` in the
```
tautschnig: s/in the/at the/

doc/cprover-manual/cbmc-assertions.md
```diff
@@ -59,6 +59,34 @@ for(i=0; i<100; i++)
...
+  int rval = a / b;
+
+  __CPROVER_postcondition(rval >= 0);
+  return rval;
+}
+```
+
+A future release of CPROVER will support using these pre and
+postconditions to create a function contract, that can be used for
+modular verification.
```
tautschnig, line 87: s/that can/which can/

tautschnig (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 119ce6020dad05ba2f0f5f8ebac28a866b7920a2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122640419

PR #4923, opened 2019-07-17
get_l1_object: only set L0 and L1 if they are non-empty

smowton:
This is important because an expression "L0: '', L1: ''" is semantically equivalent to
one with either index not specified at all, but they do not compare equal (irept::compare
distinguishes an empty-string from an absent key), leading to confusing behaviour when
ssa_exprts are used as keys in a map, for example.


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 56bb4c4fa0fa9d436c8f09301016213cd73ad15f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120260619

peterschrammel (approved):

allredj (approved):

PR #4931, opened 2019-07-20
introduce binding_exprt [blocks: #4510]

kroening:
This is a base class that factors out some common functionality of various
expressions that are variable bindings; examples are let, lambda, forall,
exists; future additions might include choose.

### DIFF
--- a/src/solvers/flattening/boolbv_quantifier.cpp
+++ b/src/solvers/flattening/boolbv_quantifier.cpp
@@ -140,8 +140,8 @@ instantiate_quantifier(const quantifier_exprt &expr, const namespacet &ns)
     return re;
   }
 
-  const auto min_i = get_quantifier_var_min(var_expr, re);
-  const auto max_i = get_quantifier_var_max(var_expr, re);
+  const optionalt<constant_exprt> min_i = get_quantifier_var_min(var_expr, re);
+  const optionalt<constant_exprt> max_i = get_quantifier_var_max(var_expr, re);
 
   if(!min_i.has_value() || !max_i.has_value())
     return nullopt;
--- a/src/util/mathematical_expr.h
+++ b/src/util/mathematical_expr.h
@@ -313,35 +313,35 @@ public:
 };
 
 /// \brief A base class for quantifier expressions
-class quantifier_exprt : public binary_predicate_exprt
+class quantifier_exprt : public binding_exprt
 {
 public:
-  quantifier_exprt(
-    const irep_idt &_id,
-    const symbol_exprt &_symbol,
-    const exprt &_where)
-    : binary_predicate_exprt(_symbol, _id, _where)
+  /// constructor for single variable
+  quantifier_exprt(irep_idt _id, symbol_exprt _symbol, exprt _where)
+    : binding_exprt(_id, {std::move(_symbol)}, std::move(_where), bool_typet())
   {
   }
 
+  /// constructor for multiple variables
+  quantifier_exprt(irep_idt _id, const variablest &_variables, exprt _where)
+    : binding_exprt(_id, _variables, std::move(_where), bool_typet())
+  {
+  }
+
+  // for the special case of one variable
   symbol_exprt &symbol()
   {
-    return static_cast<symbol_exprt &>(op0());
+    auto &variables = this->variables();
+    PRECONDITION(variables.size() == 1);
+    return variables.front();
   }
 
+  // for the special case of one variable
   const symbol_exprt &symbol() const
   {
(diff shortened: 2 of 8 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: ce8b26cd60bcedc22a900e899b36675a05caeb46).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/119891946

src/solvers/flattening/boolbv_quantifier.cpp
```diff
@@ -145,14 +140,14 @@ instantiate_quantifier(const quantifier_exprt &expr, const namespacet &ns)
     return re;
   }
 
-  const exprt min_i = get_quantifier_var_min(var_expr, re);
-  const exprt max_i = get_quantifier_var_max(var_expr, re);
+  const auto min_i = get_quantifier_var_min(var_expr, re);
```
smowton: use `constant_exprt`, nothing else here makes it obvious we're talking about a constant

smowton (commented):
Why don't the users (e.g. `typecheck_expr_main`) use the newly-introduced interface instead of the pretty-opaque `expr.op0().op0().op0()`?

src/solvers/flattening/boolbv_quantifier.cpp
kroening: changed

tautschnig (changes requested):
These `.op0()` chains are very opaque and make it rather impossible to understand what's going on. And then wouldn't we really also need to make some changes in the back-end now that the structure of those expressions is different? And the commit message talks about lambdas, lets, but only quantifiers received this new treatment. I don't think we should do half-a-job here.

kroening:
I've introduce further typing to reduce the `op0()` chains.
Note that we don't have lamba expressions yet, so this PR can't change them.

tautschnig:
> Note that we don't have lamba expressions yet, so this PR can't change them.

Well, we have `array_comprehension_exprt` (with `id() == ID_lambda`).

kroening:
I'll clean up `array_comprehension_exprt` vs. Church's lamba_exprt in a follow-up PR; this should not be done here.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 190f1c8950762802f3335cd0fadc83a5b4cabcb8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121976290

src/util/mathematical_expr.h
```diff
@@ -271,6 +271,47 @@ inline function_application_exprt &to_function_application_expr(exprt &expr)
...
+        tuple_exprt((const operandst &)_variables),
+        _id,
+        std::move(_where),
+        std::move(_type))
+  {
+  }
+
+  variablest &variables()
+  {
+    return (variablest &)static_cast<tuple_exprt &>(op0()).operands();
```
tautschnig, line 296: Quite likely this isn't the only such place, but this clearly makes `USE_LIST` no longer possible -- but in this case without causing compilation errors. So should we really get rid of `USE_LIST` right now?

tautschnig (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 3243b986fb94d6d05a40c540dbef29c99d6c0e2c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122132439

src/util/mathematical_expr.h
kroening: It would be possible to make this work with `USE_LIST`. Is `USE_LIST` sill useful?

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 0983a0cd93ff0a2cf533a81afa338cd054fc833a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122168140

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 0983a0cd93ff0a2cf533a81afa338cd054fc833a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122168140

src/util/mathematical_expr.h
tautschnig: The code wouldn't actually compile with `USE_LIST`. If `USE_LIST` were to work, we could change `irept` to have only a single container and some cursor(s) into that. Whether that's more efficient or not I can't tell, because it wouldn't compile...

Issue #4930, opened 2019-07-19
Void pointer argument cannot be pointer-checked


xbauch:
```main.c
#include <assert.h>

void foo(void *arg)
{
  assert(*(unsigned *)(arg) == 5);
}
```

CBMC version: 5.12 (cbmc-5.12-d8598f8-552-g03cc69c04-dirty)
Operating system: Ubuntu 18.04
Exact command line resulting in the issue: `$ cbmc main.c --function foo --pointer-check`
What behaviour did you expect: VERIFICATION FAILED
What happened instead:
```
--- begin invariant violation report ---
Invariant check failed
/home/diffblue/petr.bauch/code/cbmc/src/analyses/goto_check.cpp:1116 function: pointer_validity_check
Condition: size_of_expr_opt.has_value()
Reason: Check return value
Backtrace:
cbmc(print_backtrace(std::ostream&)+0x4d) [0x55c5af45da7a]
cbmc(get_backtrace[abi:cxx11]()+0x45) [0x55c5af45dc81]
cbmc(std::enable_if<std::is_base_of<invariant_failedt, invariant_failedt>::value, void>::type invariant_violated_structured<invariant_failedt, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)+0x4e) [0x55c5aed83491]
cbmc(+0x711c3d) [0x55c5aed82c3d]
cbmc(goto_checkt::pointer_validity_check(dereference_exprt const&, exprt const&, guard_exprt const&)+0x17b) [0x55c5aeff43eb]
cbmc(goto_checkt::check_rec(exprt const&, guard_exprt&)+0x654) [0x55c5aeffa79a]
cbmc(goto_checkt::check(exprt const&)+0x6a) [0x55c5aeffa820]
cbmc(goto_checkt::goto_check(dstringt const&, goto_functiont&)+0xb0f) [0x55c5aeffb9e9]
cbmc(goto_check(namespacet const&, optionst const&, goto_functionst&)+0xec) [0x55c5aeffd7a4]
[...]

PR #4936, opened 2019-07-22
Stop assertion size-of-expr for pointer-checks

xbauch:
Since these can be void-pointers. In this case we simply skip introducing the
address-check altogether. See https://github.com/diffblue/cbmc/issues/4930.

### DIFF
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -1113,7 +1113,8 @@ void goto_checkt::pointer_validity_check(
   const exprt &pointer=expr.pointer();
 
   auto size_of_expr_opt = size_of_expr(expr.type(), ns);
-  CHECK_RETURN(size_of_expr_opt.has_value());
+  if(!size_of_expr_opt.has_value())
+    return; // in the case of `void*`
 
   auto conditions = address_check(pointer, size_of_expr_opt.value());
 
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -2214,6 +2214,10 @@ void smt2_convt::convert_typecast(const typecast_exprt &expr)
         convert_typecast(tmp);
       }
     }
+    else if(src_type.id() == ID_empty)
+    {
+      convert_expr(src);
+    }
     else
     {
       std::ostringstream e_str;
@@ -4625,6 +4629,11 @@ void smt2_convt::convert_type(const typet &type)
   else if(type.id()==ID_c_bit_field)
   {
     convert_type(c_bit_field_replacement_type(to_c_bit_field_type(type), ns));
+  }
+  else if(type.id() == ID_empty)
+  {
+    // the NONDET(void) is in fact of type char
+    out << "(_ BitVec " << boolbv_width(char_type()) << " )";
   }
   else
   {
--- a/regression/cbmc/Void_Pointer_Init/main.c
+++ b/regression/cbmc/Void_Pointer_Init/main.c
@@ -1,0 +1,6 @@
+#include <assert.h>
+
+void foo(void *arg)
+{
+  assert(*(unsigned *)(arg) == 5);
+}
--- a/regression/cbmc/Void_Pointer_Init/test.desc
+++ b/regression/cbmc/Void_Pointer_Init/test.desc
@@ -1,0 +1,9 @@
+CORE
+main.c
+--function foo --pointer-check
+^\[foo.assertion.1\] line \d+ assertion \*\(unsigned \*\)\(arg\) == 5: FAILURE$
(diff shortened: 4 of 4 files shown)


src/analyses/goto_check.cpp
```diff
@@ -1113,7 +1113,8 @@ void goto_checkt::pointer_validity_check(
   const exprt &pointer=expr.pointer();
 
   auto size_of_expr_opt = size_of_expr(expr.type(), ns);
-  CHECK_RETURN(size_of_expr_opt.has_value());
+  if(!size_of_expr_opt.has_value())
+    return; // in the case of `void*`
```
danpoe, line 1117: Shouldn't `size_of_expr()` also work for void pointers? In that case we should fix `size_of_expr()` if possible.

xbauch: But even if we implement size-of for `void` the result will be `0`, right? I'm not sure we should build the `address-check` in that case.

hannes-steffenhagen-diffblue: `sizeof(void)` is undefined in C, some compilers set it to 1 just because there's too much code out there that assumes that `void* == char*`. For example, `gcc` does this but warns against it with `-Wpointer-arith` (in the `-Wpedantic` group):

```
test.c:1:16: warning: invalid application of ‘sizeof’ to a void type [-Wpointer-arith]
 int x = sizeof(void);
```

danpoe: Ah, I see. It's surprising that we get a dereference expression of a void pointer here.

kroening:
What is the semantics of a valid `void *` pointer? Would that be true for a zero-sized object?

xbauch:
> What is the semantics of a valid `void *` pointer? Would that be true for a zero-sized object?

Well if anything it should be legal to allocate memory, cast it to `void*`, pass to a function, which can then cast back, and access all the elements (within the allocation). That is probably not what we should allow for zero-sized objects.

src/analyses/goto_check.cpp
xbauch:
```
__CPROVER__start {
  char tmp;
  void *arg;
  if(NONDET(__CPROVER_bool))
    arg = NULL;

  else
  {
    arg = (void *)&tmp;
    *((void *)&tmp) = NONDET(void);
  }
  foo(arg);
}
```
This is what the entry-point looks like.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: faf21d043517234fefeb58e9c61357e226d7af51).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120033995

hannes-steffenhagen-diffblue:
The problem is that there's no "valid" semantics for what a `void*` points to per se - if it's "valid" it should point to a memory location we're allowed to access, but since there's no such thing as a "`void` object" in C we can't really tell what's supposed to be there... Now we could do some sort of abstract interpretation thingy or some such to figure out what the pointer is supposed to be dereferenced as by looking at where it's being dereferenced, the problem of course being that `void*` are often used for closures like this

```
typedef struct event event;
void register_event_handler(void *closure, void (*handler)(const event*));
```

so in that case, what the void* is supposed to be pointing to depends on what we set the handler to... There are lots of similar cases where we have instances of several parameters have nontrivial dependencies between each other, that was (for me) a big part of the motivation for `goto-harness`.

kroening:
> > What is the semantics of a valid `void *` pointer? Would that be true for a zero-sized object?
> 
> Well if anything it should be legal to allocate memory, cast it to `void*`, pass to a function, which can then cast back, and access all the elements (within the allocation). That is probably not what we should allow for zero-sized objects.

Note that it _is_ valid to allocate zero bytes; so to make the first part of your statement work, the would need to make this pass for the case of a zero-sized object.

kroening:
I.e., my proposal is to treat the case of `void *` as "zero size", but otherwise do the same checks as they are done now. In particular, something like `valid_pointer((void *)null)` should fail.

src/analyses/goto_check.cpp
tautschnig: Wouldn't `address_check` below with a zero-sized object still do some good?

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -2214,6 +2214,10 @@ void smt2_convt::convert_typecast(const typecast_exprt &expr)
         convert_typecast(tmp);
       }
     }
+    else if(src_type.id() == ID_empty)
+    {
+      convert_expr(src);
+    }
```
tautschnig, line 2220: This might actually fix some of the `broken-smt-backend` tests - would you mind testing that?

romainbrenguier (approved):

tautschnig:
Closing as superseded by #5427.

PR #4940, opened 2019-07-23
Stop generating nondet pointees for function pointers

xbauch:
when building the entry point. Because that leads to declaring function-type
objects and then assigning values to them. Rather we should simply initialise
the function pointer with NONDET(fptr_type) and let the function-pointer removal
find the candidates.

### DIFF
--- a/src/ansi-c/c_nondet_symbol_factory.cpp
+++ b/src/ansi-c/c_nondet_symbol_factory.cpp
@@ -54,6 +54,13 @@ void symbol_factoryt::gen_nondet_init(
     const pointer_typet &pointer_type=to_pointer_type(type);
     const typet &subtype = pointer_type.subtype();
 
+    if(subtype.id() == ID_code)
+    {
+      assignments.add(
+        code_assignt{expr, side_effect_expr_nondett{pointer_type, loc}});
+      return;
+    }
+
     if(subtype.id() == ID_struct_tag)
     {
       const irep_idt struct_tag = to_struct_tag_type(subtype).get_identifier();
--- a/src/ansi-c/c_nondet_symbol_factory.cpp
+++ b/src/ansi-c/c_nondet_symbol_factory.cpp
@@ -56,6 +56,9 @@ void symbol_factoryt::gen_nondet_init(
 
     if(subtype.id() == ID_code)
     {
+      // Handle the pointer-to-code case separately:
+      // leave as nondet_ptr to allow `remove_function_pointers`
+      // to replace the pointer.
       assignments.add(
         code_assignt{expr, side_effect_expr_nondett{pointer_type, loc}});
       return;
--- a/regression/cbmc/Function_Pointer_Init_No_Candidate/main.c
+++ b/regression/cbmc/Function_Pointer_Init_No_Candidate/main.c
@@ -1,0 +1,13 @@
+#include <assert.h>
+
+typedef int (*other_function_type)(int n);
+
+void foo(other_function_type other_function)
+{
+  // returning from the function call is unreachable -> the following assertion
+  //   should succeed
+  // requesting `pointer-check` will then catch the fact that there is no valid
+  //   candidate function to call resulting in an invalid function pointer
+  //   failure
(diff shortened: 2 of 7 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: e5d5c9faf2d577e3163a114d451aef206c1ee5d2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120202139

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 68c9554a3d1bd658a20969cb09775fa568a0fa75).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/144957668

hannes-steffenhagen-diffblue (approved):
This comes up a lot, maybe we should have a utility function like `is_pointer_to_object` or something of the sort?

regression/cbmc/Function_Pointer_Init/main.c
```diff
@@ -0,0 +1,10 @@
+#include <assert.h>
+
+typedef int (*other_function_type)(int n);
+
+void foo(other_function_type other_function)
+{
+  // returning from the function call is unreachable -> the following assertion
```
thk123: :question: This comment doesn't make sense wrt to the test checking for verification failed?

src/ansi-c/c_nondet_symbol_factory.cpp
```diff
@@ -48,7 +48,7 @@ void symbol_factoryt::gen_nondet_init(
     return;
   }
 
-  if(type.id()==ID_pointer)
+  if(type.id() == ID_pointer && to_pointer_type(type).subtype().id() != ID_code)
```
thk123: :question: What happens with pointers to code then? Just random number? 
Read the PR description - that is worth a comment here explaining why we don't need to initialize it here

thk123 (commented):
:no_entry_sign: Please add a test for when there _is_ a function of the right type, to check it is selected. And two functions of the right type, to check it picks between them (as I believe the new behaviour should / old behaviour did not).

regression/cbmc/Function_Pointer_Init/main.c
xbauch: Extended.

src/ansi-c/c_nondet_symbol_factory.cpp
xbauch: Now the symbol is explicitly assigned a nondet (of the function-ptr type).

xbauch:
>  Please add a test for when there _is_ a function of the right type, to check it is selected. And two functions of the right type, to check it picks between them (as I believe the new behaviour should / old behaviour did not).

I've added the tests. Note that existing functions will only be considered as candidates if dereferenced somewhere in the code.

src/ansi-c/c_nondet_symbol_factory.cpp
```diff
@@ -54,6 +54,13 @@ void symbol_factoryt::gen_nondet_init(
     const pointer_typet &pointer_type=to_pointer_type(type);
     const typet &subtype = pointer_type.subtype();
 
+    if(subtype.id() == ID_code)
+    {
+      assignments.add(
```
thk123, line 62: :pick: I'd still like a comment to the effect 
```
// leave as nondet_ptr to allow `remove_function_pointers`
// to replace the pointer. 
```

thk123 (approved):
LGTM, thanks for the extra tests. 
~This just needs code owner approval @tautschnig @kroening~

thk123:
currently failing CI, but looks transient, have restarted the job

src/ansi-c/c_nondet_symbol_factory.cpp
xbauch: Added.

thk123:
This is now passing CI and just needs code owner approval @tautschnig @kroening

tautschnig (approved):

PR #4941, opened 2019-07-23
[TG-7648] Constant propagation of substring operations

danpoe:


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: eaeadb62f60b7dce81bd2c69fcc194f742d690ab).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120411659

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 415535ecd286be4ffac801726a11676251f51ea9).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122582457

jbmc/regression/jbmc-strings/StringBuilderSubstringConstantEvaluation1/Main.java
```diff
@@ -0,0 +1,25 @@
...
+    assert s3.startsWith("abc");
+
+    String s4 = sb.substring(0, 0);
+    assert s4.length() == 0;
+    assert s4.startsWith("");
+
+    String s5 = sb.substring(0, 1);
+    assert s5.length() == 1;
+    assert s5.startsWith("a");
+  }
```
smowton, line 24: Additionally check that out-of-range begin- or end-indices still throw as expected?

jbmc/regression/jbmc-strings/StringSubstringConstantEvaluation2/test3.desc
```diff
@@ -0,0 +1,9 @@
+CORE symex-driven-lazy-loading-expected-failure
```
smowton, line 1: Any idea why the xfail?

src/goto-symex/goto_symex.cpp
```diff
@@ -310,6 +314,26 @@ goto_symext::try_evaluate_constant_string(
   return try_get_string_data_array(s_pointer_opt->get(), ns);
 }
 
+optionalt<std::reference_wrapper<const constant_exprt>>
```
smowton, line 317: `optionalt<reference_wrapper<T>>` -> `T*`

smowton (approved):

jbmc/regression/jbmc-strings/StringBuilderSubstringConstantEvaluation1/Main.java
allredj: @smowton Exceptions are a thing of the model library. There are already a couple of tests there:
https://github.com/diffblue/models-library/blob/develop/modelTests/maven/java/lang/String/src/main/java/L1Substring_int.java
https://github.com/diffblue/models-library/blob/develop/modelTests/maven/java/lang/String/src/main/java/L1Substring_int_int.java
which seem sufficient to me.

Of course these would only fail when the test-gen bump is updated, but I don't see how they coiuld.

allredj (approved):

src/goto-symex/goto_symex.cpp
danpoe: It was suggested in a previous PR to use optional references as they provide some additional semantic information. I don't really mind either way but we should be consistent about it.

romainbrenguier: I don't think there is a semantic difference, a reference wrapper is basically a pointer that can't be null, by making it optional you can get the empty optional which is basically equivalent to a null pointer, so it's just a verbose way of saying the same thing.

danpoe: Yes, I agree that they behave the same. What I meant was that when using an `optionalt` it's clear from the signature of the method that it returns an optional value, whereas when using a plain pointer that's not necessarily clear.

smowton: I suppose I would always assume a pointer might be null, otherwise you'd use a reference

jbmc/regression/jbmc-strings/StringSubstringConstantEvaluation2/test3.desc
danpoe: When using symex driven lazy loading it seems that jbmc does not fully honor the `--property` option as several additional VCCs are generated (e.g. null pointer checks). Those regression tests though check that a specific VCC is not simplified. This is currently done by selecting that VCC via `--property` and then checking that jbmc prints `Generated 1 VCC(s), 1` remaining after simplification. If there would be more VCCs we could not check that this one VCC is the one that is not simplified.

jbmc/regression/jbmc-strings/StringBuilderSubstringConstantEvaluation1/Main.java
danpoe: I think @smowton is referring to the case where all the arguments to `substring()` are constant, but the indices are out of bounds. In this case the constant propagator should not simplify anything and the statement should just be symex'd normally. Whether that then later leads to an exception is a thing of the models library. I'll add a test to check that no simplification occurs for out of bounds indices.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: e0ef78e57eafa62c9f0709f02ba31de29401412d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123016840

src/goto-symex/goto_symex.cpp
danpoe: Sure, though a nullpointer return value could suggest that some kind of error has occured, whereas `optionalt` would suggest that this is part of normal and expected behavior.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 7da4f78b6e5edb66afb27ec9cb9f17ba6ad179aa).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123397270

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: e56041f4cf250f77a010b7231c74a3fa6051b1a9).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123399979

PR #4943, opened 2019-07-23
Error message when entry function body cannot be produced [TG-8299]

peterschrammel:
Previously crashed on a failed map.at

### DIFF
--- a/jbmc/src/java_bytecode/java_bytecode_language.cpp
+++ b/jbmc/src/java_bytecode/java_bytecode_language.cpp
@@ -899,6 +899,17 @@ bool java_bytecode_languaget::generate_support_functions(
   // parameter names
   convert_lazy_method(res.main_function.name, symbol_table_builder);
 
+  const symbolt &symbol =
+    symbol_table_builder.lookup_ref(res.main_function.name);
+  if(symbol.value.is_nil())
+  {
+    throw invalid_command_line_argument_exceptiont(
+      "the program has no entry point",
+      "function",
+      "Check that the specified entry point is included by your "
+      "--context-include or --context-exclude options");
+  }
+
   // generate the test harness in __CPROVER__start and a call the entry point
   return java_entry_point(
     symbol_table_builder,
--- a/jbmc/regression/jbmc/context-include-exclude/test_excluded_entry_point.desc
+++ b/jbmc/regression/jbmc/context-include-exclude/test_excluded_entry_point.desc
@@ -1,0 +1,11 @@
+CORE
+Main.class
+--context-exclude 'org.cprover.MyClass$Inner.' --function 'org.cprover.MyClass$Inner.doIt:(I)I'
+^EXIT=1$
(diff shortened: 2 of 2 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: b590a06d18761ccf6e2444c1cec0fd07fb79146e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120286881

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -899,6 +899,16 @@ bool java_bytecode_languaget::generate_support_functions(
...
   convert_lazy_method(res.main_function.name, symbol_table_builder);
 
+  const symbolt &symbol =
+    symbol_table_builder.lookup_ref(res.main_function.name);
+  if(symbol.value.is_nil())
+  {
+    throw invalid_command_line_argument_exceptiont(
+      "the program has no entry point",
+      "function",
+      "Specify an entry point that is in the analysis context");
```
smowton: Not sure most readers would comprehend "in the analysis context". How about "Specify an entry-point function defined on the classpath"? If they're using "--context-exclude" and so on they presumably understand the hole they're making in the realm of loaded classes?

peterschrammel: That would be misleading. The case that it is not on the classpath is already covered elsewhere. This only happens with context restrictions.

smowton: Hmm. Perhaps name-check the offending `--context-exclude` argument?

peterschrammel: I don't understand

smowton: For example, "The specified entry-point was not in the analysis context; check that it is included by your --context-include or --context-exclude options"

smowton: Or even just "was not available"; I'm trying to avoid saying "the analysis context" and hoping they know what that means

smowton:
Side-note, and not for this PR (I should have talked about this back when `--context-in/exclude` went in): why are we calling this "context" anyway? It's a pretty broad term that refers to a lot of different things, but I don't see any relationship between "the set of methods we don't stub" and "context"?

jbmc/src/java_bytecode/java_bytecode_language.cpp
peterschrammel: Ok, changed.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 18fa5da9b9e93f8d9fccdfa7a8a063800429da45).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120460442

smowton (approved):

PR #4944, opened 2019-07-24
Add field-sensitivity tests

smowton:
These look at various cases of struct-field assignment, directly and via pointers; only one
case is identified as not using the field-sensitivity machinery so far (that of a pointer that
may point to several different offsets within the same target allocation).

These are added now so that forthcoming changes to field-sensitivity have a reasonably comprehensive
regression test suite to check against.

### DIFF
--- a/regression/cbmc/field-sensitivity1/test.c
+++ b/regression/cbmc/field-sensitivity1/test.c
@@ -1,0 +1,15 @@
+#include <assert.h>
+
+struct A
+{
+  int x;
+  int y;
+};
+
+int main(int argc, char **argv)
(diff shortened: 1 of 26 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: ab6ca36e95f5a84d826f29788bb9de60b978ad5c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120464194

peterschrammel (approved):

allredj (approved):

PR #4945, opened 2019-07-24
Generation of function contracts from preconditions

angelhof:
The goal of this PR is to extend the function contracts using preconditions.

An overview of the code changes follows:

I have added a flag `turn-preconditions-to-contracts` in goto-cc that collects preconditions from all function bodies and extends the function contract with their conjunction. The code for this functionality is in  `src/ansi-c/ansi_c_language.cpp`.  This analysis is done on the AST and not in the goto-binary because it is much cleaner to extract the condition in the preconditions from the AST (as in the goto-binary function calls are flattened, and some transformations happen so it becomes difficult to reconstruct the conditions). It is assumed that calls to `__CPROVER_precondition` are in the beginning of the function body and consecutive.

### DIFF
--- a/src/ansi-c/ansi_c_language.cpp
+++ b/src/ansi-c/ansi_c_language.cpp
@@ -9,9 +9,9 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "ansi_c_language.h"
 
 #include <cstring>
-#include <sstream>
 #include <fstream>
 #include <iostream>
+#include <sstream>
 
 #include <util/config.h>
 #include <util/expr_iterator.h>
@@ -138,13 +138,14 @@ bool ansi_c_languaget::generate_support_functions(
     symbol_table, get_message_handler(), object_factory_params);
 }
 
-// Question: Are those [is_*] functions correct? Are they sound/complete?
-bool is_call_to_function(irep_idt function_name, exprt expr) {
+bool is_call_to_function(irep_idt function_name, exprt expr)
+{
   if(can_cast_expr<side_effect_expr_function_callt>(expr))
   {
-    const side_effect_expr_function_callt function_call = to_side_effect_expr_function_call(expr);
+    const side_effect_expr_function_callt function_call =
+      to_side_effect_expr_function_call(expr);
     const exprt function = function_call.function();
-    
+
     if(can_cast_expr<symbol_exprt>(function))
     {
       const symbol_exprt function_symbol = to_symbol_expr(function);
@@ -154,49 +155,35 @@ bool is_call_to_function(irep_idt function_name, exprt expr) {
   return false;
 }
 
-
-exprt::operandst filter_preconditions(irep_idt target_function_name, code_blockt function_body) {
-  // TODO: I probably have to add some check that inside the
-  // code in the precondition there is nothing funky going on?
-  exprt::operandst conditions;
-  for (depth_iteratort it = function_body.depth_begin(); it != function_body.depth_end(); ++it) {
-    // std::cout << it->id() << " -- " << it->find(ID_statement).id() << "\n";
-    if (is_call_to_function(target_function_name, *it)) {
-      const side_effect_expr_function_callt function_call = to_side_effect_expr_function_call(*it);
-      exprt condition = function_call.arguments().front();
-      // std::cout << "Precondition:\n" << precondition.pretty() << "\n";
-      // std::cout << "Precondition type:\n" << precondition.type().id() << "\n";
-      
-      conditions.push_back(condition);
-
-      // TODO: I should probably go to the next sibling or parent
-      // after finding a function call to the target name and not just
-      // iterate it once
-    }
-  }
-  return conditions;
-}
-
 // Question: Is there a way to optimize this (Returning a balanced
 // tree instead of a chain tree)? Is there a standard function that I
 // can call for this?
 //
-// If this function returns nil expression, it means that no (pre/post)condition was aggregated.
+// If this function returns nil expression, it means that no
+// (pre/post)condition was aggregated.
 exprt condition_conjunction(const exprt::operandst &nil_op)
 {
   exprt::operandst op;
-  
+
   // filter non nil expressions
-  std::copy_if (nil_op.begin(), nil_op.end(), std::back_inserter(op), [](exprt e){return e.is_not_nil();} );
-  
-  if(op.empty()) {
+  std::copy_if(
+    nil_op.begin(), nil_op.end(), std::back_inserter(op), [](exprt e) {
+      return e.is_not_nil();
+    });
+
+  if(op.empty())
+  {
     // Question: What is a better way to do this?
     exprt expr = exprt(ID_nil);
-    INVARIANT(expr.is_nil(), "Postcondition conjunction should be nil if no postcondition group was found.");
+    INVARIANT(
+      expr.is_nil(),
+      "Postcondition conjunction should be nil if no postcondition group was "
+      "found.");
     return expr;
     // return make_boolean_expr(true);
   }
-  else if(op.size()==1) {
+  else if(op.size() == 1)
+  {
     return op.front();
   }
   else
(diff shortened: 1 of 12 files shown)


karkhaz:
Here is a list of reasons that @angelhof is doing this pass in goto-cc rather than goto-instrument:
* The goto-cc AST contains the argument to `__CPROVER_precondition` as a single expression. Whereas it will have been changed into multiple instructions by the time it has been transformed into a goto-program.
* The multiple instructions in the goto-program will contain temporary variables (for function return values) etc. So these would need to be put into the generated stub, but in general these temporary variables don't make any sense outside the original function.
* Some functions have multiple return points, each with a different postcondition. To be sound, the stub's postcondition should be a disjunction of all these postconditions. It's much easier to get the postcondition from the AST, since again, the expression may have been transformed into several instructions by the time we get to goto-instrument.

Asking for advice: should we proceed with this PR, or is it better to implement the stub-generating functionality in goto-instrument? If the latter, how would we solve the above problems? My thoughts so far are that we either need to write an analysis that recovers the original {pre,post}condition from several goto-instructions; or we could make goto-cc write the original pre/postcondition as a string in an irep comment, so that goto-instrument can later make use of it. I don't have any better ideas right now

karkhaz:
@chrisr-diffblue @martin-cs is there any chance we could get some implementation advice about this? There's no need for a lengthy review at this stage, but I'd just like to point @angelhof in the right direction and we can iterate from there. Thank you!

chrisr-diffblue:
@martin-cs I suggested it might be worth pinging you too because I wondered if this might overlap/compliment some of the stuff we wanted for more flexible handling of Ada language checks in gnat2goto...

src/ansi-c/ansi_c_language.cpp
```diff
@@ -135,6 +138,492 @@ bool ansi_c_languaget::generate_support_functions(
     symbol_table, get_message_handler(), object_factory_params);
 }
 
+// Question: Are those [is_*] functions correct? Are they sound/complete?
+bool is_call_to_function(std::string function_name, exprt expr) {
+  if(can_cast_expr<side_effect_expr_function_callt>(expr))
```
hannes-steffenhagen-diffblue, line 146: Note that aside from `side_effect_expr_function_callt` there is also `code_function_callt` (function call statement as opposed to expression). Also, I think this function would be somewhat cleaner if it just took a `side_effect_expr_function_callt` in the first place.

src/ansi-c/ansi_c_language.cpp
```diff
@@ -135,6 +138,492 @@ bool ansi_c_languaget::generate_support_functions(
...
+  {
+    const side_effect_expr_function_callt function_call = to_side_effect_expr_function_call(expr);
+    const exprt function = function_call.function();
+    
+    if(can_cast_expr<symbol_exprt>(function))
```
hannes-steffenhagen-diffblue, line 152: The function to be called is not necessarily a `symbol_exprt`, it might just as well be an expression returning a function pointer. If that's not a concern for you that's fine, but I think this is then a somewhat more limited scope than the name of the function implies.

src/ansi-c/ansi_c_language.cpp
```diff
@@ -135,6 +138,492 @@ bool ansi_c_languaget::generate_support_functions(
     symbol_table, get_message_handler(), object_factory_params);
 }
 
+// Question: Are those [is_*] functions correct? Are they sound/complete?
+bool is_call_to_function(std::string function_name, exprt expr) {
```
hannes-steffenhagen-diffblue: `function_name` should be `irep_idt`

src/ansi-c/ansi_c_language.cpp
```diff
@@ -135,6 +138,492 @@ bool ansi_c_languaget::generate_support_functions(
...
+    if(can_cast_expr<symbol_exprt>(function))
+    {
+      const symbol_exprt function_symbol = to_symbol_expr(function);
+      return function_symbol.get_identifier() == function_name;
+    }
+  }
+  return false;
+}
+
+bool is_code_postcondition(exprt expr) {
```
hannes-steffenhagen-diffblue: This should be at minium a `codet`, possibly something more specific.

src/ansi-c/ansi_c_language.cpp
```diff
@@ -135,6 +138,492 @@ bool ansi_c_languaget::generate_support_functions(
...
+  if(can_cast_expr<code_expressiont>(expr))
+  {
+    exprt expr0 = to_code_expression(to_code(expr)).expression();
+    return is_call_to_function("__CPROVER_postcondition", expr0);
+  }
+  return false;
+}
+
+
+exprt::operandst filter_pre_post_conditions(std::string target_function_name, exprt function_body) {
```
hannes-steffenhagen-diffblue: `target_function_name` should be `irep_idt`, `function_body` should probably be `code_blockt`.

src/ansi-c/ansi_c_language.cpp
```diff
@@ -135,6 +138,492 @@ bool ansi_c_languaget::generate_support_functions(
...
+    // Question: What is the correct way of finding a function call?
+    // std::cout << it->id() << " -- " << it->find(ID_statement).id() << "\n";
+    // Filters the function calls
+    if (is_call_to_function(target_function_name, *it)) {
+      const side_effect_expr_function_callt function_call = to_side_effect_expr_function_call(*it);
+      exprt condition = function_call.arguments().front();
+      // std::cout << "Precondition:\n" << precondition.pretty() << "\n";
+      // std::cout << "Precondition type:\n" << precondition.type().id() << "\n";
+      
+      // Note: it is not very efficient to push back in a vector.
```
hannes-steffenhagen-diffblue: ? vector `push_back` is amortised O(1). You can't really get any faster than that unless you already know in advance how many elements there are.

src/langapi/language.h
```diff
@@ -158,6 +158,15 @@ class languaget:public messaget
 
   // show parse tree
 
+  virtual bool preconditions_to_contracts()
+    {
+      INVARIANT(
```
hannes-steffenhagen-diffblue: Why is it a virtual function at all in that case?

src/goto-cc/compile.cpp
```diff
@@ -504,8 +504,23 @@ bool compilet::parse(
       error() << "PARSING ERROR" << eom;
       return true;
     }
+    // Question: Should this move after a typecheck. I don't think so,
+    // because typecheck should also typecheck the contracts in
+    // theory, so it is actually even better to let typecheck happen
+    // after this transformation so that we make sure that the
+    // transformed contracts are correct
```
hannes-steffenhagen-diffblue: I think what we call "typecheck" actually isn't a type check per se, but as far as that's concerned if we only did a pure typecheck we should really do it _both_ before and after the transformation because if something is wrong after the transformation we'd like to know if was wrong to begin with or if the transformation messed it up.

hannes-steffenhagen-diffblue (commented):
Some codestyle nitpicks mostly. As to the overall direction, I don't completely understand the use case here but if the issue is that tracking pre/postconditions in later stages is hard (but apparently still useful) perhaps function pre and post conditions should just be added as a separate thing to the goto model itself?

angelhof:
@hannes-steffenhagen-diffblue Thanks a lot for the comments :) There is a separate PR that I will submit later that generates a stub for a function given its contract. The use case of these 2 PRs is to generate these stubs from the pre and postconditions, which are heavily used in our codebase as they can also be asserted at runtime (in contrast to the contracts).

My main concern (and the reason why I submitted this as a draft PR) is whether it is fine to do this analysis in `goto-cc`, instead of in `goto-instrument`. Do you have any opinion on that? As I mentioned above, there are several issues when trying to do this analysis in `goto-instrument` and that is why I implemented it on the AST in `goto-cc`.

martin-cs:
@angelhof Sorry for the lag in getting back to you.  I will summarise my understanding of the situation; please let me know if I have misunderstood or left something out.

*. The general context of this work is in doing modular deductive verification things.  So we are in a world where SPARK and Boogie are state-of-the-art.

*. Your source code has annotated pre and post conditions (and loop invariants?) using magic CPROVER functions.  What, exactly, are the semantics of these?  How do they relate to things like ACSL and Misra-C's contracts?

*. The placement of the pre and post condition statements in the program are assumed to be following some kind of rules.

*. The goal of this patch set is to assemble all of the pre conditions (implicitly taking the conjunction) and the post conditions (implicitly taking the disjunction) in a function.  The resulting object is then what you are referring to as a code contract.

*. The goal of the next patch is to take a code contract and generate a stub function which asserts that the precondition must hold, sets some variables nondeterministically and then assumes that the postcondition must hold.

*. For the reasons that @karkhaz observes, you feel that goto-cc is a better place to do this than goto-instrument.

Have I got the context right?

karkhaz:
@martin-cs this context is accurate. We have agreed to discuss this offline, and I'll update the thread with what we agree on how to proceed. Thank you!

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 2943f2234f335e731eb6356ac3a4204ea29591b5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121659378

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: db2aff1dab374e0de062c5765708e1f9f366d082).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121659962

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 70f0cb3975eb7495e092b367ccf25d0b9ec877c6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121662172

PR #4949, opened 2019-07-25
Simplify and apply field sensitivity before value-set-deref

smowton:
This means that any member-of-symbol constructs introduced by a nested dereference,
such as `x->y ==> (x == &o1 ? o1 : o2).y`, can be simplified to produce e.g.
`(x == &o1 ? o1.y : o2.y) ==> (x == &o1 ? o1..y : o2..y)`. value_set_dereferencet will
then special-case the if-expression, dereferencing the inner `o1..y` and `o2..y` individually.
In the best case where each has a single possible alias, `&o3` and `&o4` respectively, we
end up with `(x == &o1 ? &o3 : &o4)`, rather than the current result:
`let p = (x == &o1 ? o1 : o2).y in (p == &o3 ? o3 : o4)`.

The net benefit: smaller expressions, and thanks to simpler expressions a higher chance that future simplification will succeed.

### DIFF
--- a/regression/cbmc/double_deref/double_deref_with_member.desc
+++ b/regression/cbmc/double_deref/double_deref_with_member.desc
@@ -1,10 +1,10 @@
 CORE
 double_deref_with_member.c
 --show-vcc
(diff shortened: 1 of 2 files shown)


peterschrammel (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8ff892778dc640a88486e7b5820bfc8467ea721a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120726270

owen-mc-diffblue (approved):

PR #4950, opened 2019-07-26
Check for presence of SMT solver

xbauch:
and report early if missing.

### DIFF
--- a/regression/cbmc/CVC4_Not_Present/test.desc
+++ b/regression/cbmc/CVC4_Not_Present/test.desc
@@ -1,0 +1,8 @@
+CORE
+main.c
+--cvc4
+^SMT2 solver \'cvc4\' is not on the PATH$
+^EXIT=1$
+^SIGNAL=0$
+--
+^\[main.assertion.1\] line \d+ assertion i \> 5\: ERROR$
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -21,6 +21,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/exit_codes.h>
 #include <util/invariant.h>
 #include <util/make_unique.h>
+#include <util/run.h>
 #include <util/unicode.h>
 #include <util/version.h>
 
@@ -360,6 +361,7 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
   if(cmdline.isset("boolector"))
   {
     options.set_option("boolector", true), solver_set=true;
+    check_smt2("boolector");
     options.set_option("smt2", true);
   }
 
@@ -372,24 +374,28 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
   if(cmdline.isset("mathsat"))
   {
     options.set_option("mathsat", true), solver_set=true;
+    check_smt2("mathsat");
     options.set_option("smt2", true);
   }
 
   if(cmdline.isset("cvc4"))
   {
     options.set_option("cvc4", true), solver_set=true;
+    check_smt2("cvc4");
     options.set_option("smt2", true);
   }
 
   if(cmdline.isset("yices"))
   {
     options.set_option("yices", true), solver_set=true;
+    check_smt2("yices");
     options.set_option("smt2", true);
   }
 
   if(cmdline.isset("z3"))
   {
     options.set_option("z3", true), solver_set=true;
+    check_smt2("z3");
     options.set_option("smt2", true);
   }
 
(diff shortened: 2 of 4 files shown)


src/cbmc/cbmc_parse_options.cpp
```diff
@@ -450,6 +457,27 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
...
 }
 
+void cbmc_parse_optionst::check_smt2(
+  const std::string &name,
+  const std::string &pre_message)
+{
+  int exit_code;
+#ifdef _WIN32
+  {
+    std::string where_string = "\%windir\%\\system32\\where.exe";
```
smowton: % isn't a special character (if you were thinking of format-string escaping, that's "%%")

smowton (approved):

regression/cbmc/CVC4_Not_Present/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+--cvc4
+^SMT2 solver \'cvc4\' is not on the PATH$
+^EXIT=1$
+^SIGNAL=0$
+--
+^\[main.assertion.1\] line \d+ assertion i \> 5\: ERROR$
```
danpoe, line 8: We shouldn't include such a test as it would fail for someone who has `cvc4` on the `PATH`. Maybe `check_smt2` could be tested via a unit test instead.

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -450,6 +457,27 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
...
+    std::string where_string = "%%windir%%\\system32\\where.exe";
+    exit_code = run(where_string, {where_string, name}, "", "", "");
+  }
+#else
+  exit_code = run("which", {"which", name}, "", "", "");
+#endif
+  if(exit_code != 0)
+  {
+    log.error() << pre_message << "SMT2 solver \'" << name
+                << "\' is not on the PATH" << messaget::eom;
```
hannes-steffenhagen-diffblue, line 476: Throw an exception instead

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -450,6 +457,27 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
...
+    exit(CPROVER_EXIT_USAGE_ERROR);
```
danpoe, line 477: Instead of exiting here I think we should throw an exception instead and let the logging and exiting be handled in `parse_options_baset::main()`.

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -450,6 +457,27 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
...
+  const std::string &pre_message)
+{
+  int exit_code;
+#ifdef _WIN32
+  {
+    std::string where_string = "%%windir%%\\system32\\where.exe";
+    exit_code = run(where_string, {where_string, name}, "", "", "");
+  }
+#else
+  exit_code = run("which", {"which", name}, "", "", "");
```
hannes-steffenhagen-diffblue, line 471: This is a bit roundabout (maybe we should have a separate utility for querying the path ourselves?) but I suppose this is fine for now.

hannes-steffenhagen-diffblue (approved):

regression/cbmc/CVC4_Not_Present/test.desc
hannes-steffenhagen-diffblue: This test could be run with its own `chain.sh` that sets up an empty `PATH` before running cbmc.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 77311930ad55e88754f07ec985d20618f90f22a1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120840714

kroening:
This incurs non-trivial cost on Windows. We have recently added a version detection for gcc, so a full run might run four executables. Eventually the cbmc command-line tool may become unusable.

tautschnig:
As an alternative, could we just have a better error report upon the actual (attempted) invocation of the solver? This would avoid the concern that @kroening raised, but of course would delay the problem report.

NlightNFotis:
Closing this due to a number of reasons:

1) It focuses changes on the older SMT backend (which is to be superseded by the work done in `/src/solvers/smt2_incremental/`)
2) It's going to require a bit of work to tidy this up and bring it up to speed with the current version of `develop`.

If you would rather this gets merged, feel free to reopen the PR.

PR #4951, opened 2019-07-26
java_bytecode tweaks

LAJW:
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own. (where possible)

### DIFF
--- a/jbmc/src/java_bytecode/java_bytecode_parser.cpp
+++ b/jbmc/src/java_bytecode/java_bytecode_parser.cpp
@@ -40,6 +40,19 @@ public:
 
   bool parse() override;
 
+  struct pool_entryt
+  {
+    u1 tag = 0;
+    u2 ref1 = 0;
+    u2 ref2 = 0;
+    irep_idt s;
+    u8 number = 0;
+    exprt expr;
+  };
+
+  java_bytecode_parse_treet parse_tree;
+
+private:
   typedef java_bytecode_parse_treet::classt classt;
   typedef java_bytecode_parse_treet::classt::fieldst fieldst;
   typedef java_bytecode_parse_treet::classt::methodst methodst;
@@ -55,22 +68,9 @@ public:
     lambda_method_handlet;
   typedef java_bytecode_parse_treet::classt::u2_valuest u2_valuest;
 
-  java_bytecode_parse_treet parse_tree;
-
-  struct pool_entryt
-  {
-    u1 tag = 0;
-    u2 ref1 = 0;
-    u2 ref2 = 0;
-    irep_idt s;
-    u8 number = 0;
-    exprt expr;
-  };
-
   typedef std::vector<pool_entryt> constant_poolt;
   constant_poolt constant_pool;
 
-private:
   const bool skip_instructions = false;
 
   pool_entryt &pool_entry(u2 index)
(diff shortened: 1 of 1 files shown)


jbmc/src/java_bytecode/java_bytecode_parser.cpp
```diff
@@ -38,7 +38,7 @@ class java_bytecode_parsert:public parsert
   {
   }
 
-  virtual bool parse();
+  virtual bool parse() override;
```
owen-mc-diffblue: :pick: The coding standard says you should remove the `virtual` since you have `override`.

owen-mc-diffblue (approved):

jbmc/src/java_bytecode/java_bytecode_parser.cpp
owen-mc-diffblue: Indeed, the linter is failing for this reason

jbmc/src/java_bytecode/java_bytecode_parser.cpp
```diff
@@ -137,8 +137,12 @@ class java_bytecode_parsert:public parsert
     }
   }
 
-  u8 read_bytes(size_t bytes)
+  template <typename T>
```
smowton, line 131: commit message "Make read_bytes less likely to be called incorrectly" truncated (larger than... what?)

jbmc/src/java_bytecode/java_bytecode_parser.cpp
```diff
@@ -653,15 +653,9 @@ void java_bytecode_parsert::rconstant_pool()
 
   constant_pool.resize(constant_pool_count);
 
-  for(constant_poolt::iterator
-      it=constant_pool.begin();
-      it!=constant_pool.end();
+  for(auto it = std::next(constant_pool.begin()); it != constant_pool.end();
```
smowton, line 656: Could the constant pool be empty? If so this is wrong

smowton:
Looks ok except the `std::next` on a maybe-end iterator, and I don't see the value in inlining e.g. `annotationst`, where I guess the idea was if we want to turn that into `std::list` or whatever then we only need to make the change in one place?

jbmc/src/java_bytecode/java_bytecode_parser.cpp
LAJW: No, this is already checked at the start of the function.

LAJW:
@smowton I don't like obstructing collections like that, because each type has different properties and advantages. Code already relies on these assumptions, so swapping containers is not really an option. Having it spelled out in full makes it easier to see what you can/cannot use. I've seen cases where people were manually converting vectors, just because they were typedefed.

And in this case this code was just stealing typedefs from other unrelated class for convenience, not because this is what a particular class returns.

smowton:
Eh, if you want. It's widely done in this codebase so you'll have much more to do if you want to go that way, and hundreds of lines of churn for your trouble.

LAJW:
@smowton Commit message updated. @owen-jones-diffblue linter problems fixed.

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8d3559bba095fb725ad6fa7a7b45a5880e56b155).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120884008

PR #4953, opened 2019-07-26
Stop sharing removal on refine-strings expression walk

JohnDumbell:
When string refinement was walking a tree it broke all sharing as it went due to requesting a non-constant iterable, but then only infrequently needed to actually make changes. This causes memory problems depending upon size of expressions that are then created - more complicated the program and string expressions, the worse this got.

### DIFF
--- a/src/solvers/strings/string_refinement.cpp
+++ b/src/solvers/strings/string_refinement.cpp
@@ -111,7 +111,7 @@ static optionalt<exprt> get_array(
   const array_string_exprt &arr,
   const array_poolt &array_pool);
 
-static exprt substitute_array_access(
+static optionalt<exprt> substitute_array_access(
   const index_exprt &index_expr,
   symbol_generatort &symbol_generator,
   const bool left_propagate);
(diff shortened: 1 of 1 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4563469d2eb70d2f263cdfb8c34f558e3746bf3f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/120867895

src/solvers/strings/string_refinement.cpp
```diff
@@ -1200,12 +1206,24 @@ static void substitute_array_access_in_place(
...
+    if(result)
+    {
+      expr.write();
+      expr.swap(*result);
+    }
   }
 
-  for(auto &op : expr.operands())
-    substitute_array_access_in_place(op, symbol_generator, left_propagate);
+  // Trick it into giving us a const iterable (thus not copying anything) as we
```
smowton: Rather than using const-hacking, use the depth-iterator, which has a `mutate()` operation for exactly this purpose. It would look something like
```
for(auto it = expr.depth_begin(), itend = expr.depth_end(); it != itend; ++it)
{
  if(const auto index_expr = expr_try_dynamic_cast<index_exprt>(*it))
  {
    optionalt<exprt> result =
￼       substitute_array_access(*index_expr, symbol_generator, left_propagate);		￼       ￼     
      if(result)
        it.mutate() = *result;
  }
}
```
That performs CoW breaks from the root of the expression up to the modified node on demand.

src/solvers/strings/string_refinement.cpp
```diff
@@ -1155,16 +1155,22 @@ static exprt substitute_array_access(
   symbol_generatort &symbol_generator,
   const bool left_propagate)
 {
-  // Substitute recursively in branches of conditional expressions
-  const exprt true_case = substitute_array_access(
-    index_exprt(if_expr.true_case(), index), symbol_generator, left_propagate);
-  const exprt false_case = substitute_array_access(
-    index_exprt(if_expr.false_case(), index), symbol_generator, left_propagate);
+  auto true_index = index_exprt(if_expr.true_case(), index);
+  auto false_index = index_exprt(if_expr.false_case(), index);
```
romainbrenguier: :pick: `exprt` instead of `auto`

src/solvers/strings/string_refinement.cpp
```diff
@@ -1200,12 +1206,24 @@ static void substitute_array_access_in_place(
...
   {
-    expr =
+    optionalt<exprt> result =
       substitute_array_access(*index_expr, symbol_generator, left_propagate);
+
+    // Only perform a write when we have something changed.
+    if(result)
+    {
+      expr.write();
+      expr.swap(*result);
```
romainbrenguier: That's not very clear. Could we just do `expr = std::move(*result)` instead?

src/solvers/strings/string_refinement.cpp
romainbrenguier: :+1: but I'm unsure about `itend` being defined at the beginning of a loop which can mutate `expr`

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 6642d127e6fedf73fcf65b235e649fd989a27f7a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121067189

JohnDumbell:
Made the requested changes.

romainbrenguier (approved):

PR #4957, opened 2019-07-29
Simplify when combining string-refinement result constraints

JohnDumbell:
Due to the improvements in constant prop we increasingly have constraints that are trivially simplified, like:

`if 5 >= 2 do x else y`

Also because of the way these conditions are combined - the resulting merged if gets added to all subsequent conditions and values - we end up with expressions like this:

`if 5 >= 2 then (if 5 >= 2 do x ... else y ...) else (if 5 >= 2 do x ... else y ...)`

Where the size of the expression increases exponentially to the complexity of the original if statement even if the entire conditional is trivial to prove right/wrong. This could likely be worked out by the SAT solver happily enough but due to the size of the resulting expressions (I've seen ones 1.34 billion nodes deep) it never gets there within a reasonable time-frame.

### DIFF
--- a/src/solvers/strings/string_builtin_function.cpp
+++ b/src/solvers/strings/string_builtin_function.cpp
@@ -407,8 +407,8 @@ optionalt<exprt> string_of_int_builtin_functiont::eval(
 string_constraintst string_of_int_builtin_functiont::constraints(
   string_constraint_generatort &generator) const
 {
-  auto pair = add_axioms_for_string_of_int_with_radix(
-    result, arg, radix, 0, generator.ns, array_pool);
+  auto pair =
+    generator.add_axioms_for_string_of_int_with_radix(result, arg, radix, 0);
   pair.second.existential.push_back(equal_exprt(pair.first, return_code));
   return pair.second;
 }
--- a/src/solvers/strings/string_constraint_generator.h
+++ b/src/solvers/strings/string_constraint_generator.h
@@ -74,364 +74,235 @@ private:
   exprt associate_length_to_array(
     const exprt &return_code,
     const function_application_exprt &f);
+
+public:
+  std::pair<exprt, string_constraintst> add_axioms_for_concat(
+    const array_string_exprt &res,
+    const array_string_exprt &s1,
+    const array_string_exprt &s2);
+
+  std::pair<exprt, string_constraintst> add_axioms_for_concat_substr(
+    const array_string_exprt &res,
+    const array_string_exprt &s1,
+    const array_string_exprt &s2,
+    const exprt &start_index,
+    const exprt &end_index);
+  std::pair<exprt, string_constraintst> add_axioms_for_insert(
+    const array_string_exprt &res,
+    const array_string_exprt &s1,
+    const array_string_exprt &s2,
+    const exprt &offset);
+  std::pair<exprt, string_constraintst> add_axioms_for_string_of_int_with_radix(
+    const array_string_exprt &res,
+    const exprt &input_int,
+    const exprt &radix,
+    size_t max_size);
+
+  string_constraintst add_constraint_on_characters(
+    const array_string_exprt &s,
+    const exprt &start,
+    const exprt &end,
+    const std::string &char_set);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_constrain_characters(const function_application_exprt &f);
+
+  // The following functions add axioms for the returned value
+  // to be equal to the result of the function given as argument.
+  // They are not accessed directly from other classes: they call
+  // `add_axioms_for_function_application` which determines which of
+  // these methods should be called.
+
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_char_at(const function_application_exprt &f);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_code_point_at(const function_application_exprt &f);
+
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_code_point_before(const function_application_exprt &f);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_contains(const function_application_exprt &f);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_equals(const function_application_exprt &f);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_equals_ignore_case(const function_application_exprt &f);
+
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_is_empty(const function_application_exprt &f);
+  std::pair<exprt, string_constraintst> add_axioms_for_is_prefix(
+    const array_string_exprt &prefix,
+    const array_string_exprt &str,
+    const exprt &offset);
+  std::pair<exprt, string_constraintst> add_axioms_for_is_prefix(
+    const function_application_exprt &f,
+    bool swap_arguments);
+
+  std::pair<exprt, string_constraintst> add_axioms_for_is_suffix(
+    const function_application_exprt &f,
+    bool swap_arguments);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_length(const function_application_exprt &f);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_empty_string(const function_application_exprt &f);
+
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_copy(const function_application_exprt &f);
+
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_concat_code_point(const function_application_exprt &f);
+  std::pair<exprt, string_constraintst> add_axioms_for_constant(
+    const array_string_exprt &res,
+    irep_idt sval,
+    const exprt &guard = true_exprt());
+
+  std::pair<exprt, string_constraintst> add_axioms_for_delete(
+    const array_string_exprt &res,
+    const array_string_exprt &str,
+    const exprt &start,
+    const exprt &end);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_delete(const function_application_exprt &f);
+  std::pair<exprt, string_constraintst>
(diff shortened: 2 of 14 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c70f7ef5e52aea0091216d605a6a45e5b8cbad23).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121394358

src/solvers/strings/string_constraint_generator_main.cpp
```diff
@@ -441,9 +441,12 @@ exprt zero_if_negative(const exprt &expr)
...
-  std::pair<exprt, string_constraintst> result2)
+  std::pair<exprt, string_constraintst> result2,
+  const namespacet &ns)
 {
-  const exprt return_code = maximum(result1.first, result2.first);
+  exprt return_code = maximum(result1.first, result2.first);
   merge(result2.second, std::move(result1.second));
+  simplify(return_code, ns);
+
   return {return_code, std::move(result2.second)};
```
tautschnig: May I suggest to use `return {simplify_expr(return_code, ns), std::move(result2.second)};` at which point the `return_code` can be made `const exprt`.

tautschnig (approved):

src/solvers/strings/string_constraint_generator_main.cpp
romainbrenguier: :+1:

src/solvers/strings/string_constraint_generator_main.cpp
```diff
@@ -441,9 +441,12 @@ exprt zero_if_negative(const exprt &expr)
...
 std::pair<exprt, string_constraintst> combine_results(
   std::pair<exprt, string_constraintst> result1,
-  std::pair<exprt, string_constraintst> result2)
+  std::pair<exprt, string_constraintst> result2,
+  const namespacet &ns)
 {
-  const exprt return_code = maximum(result1.first, result2.first);
+  exprt return_code = maximum(result1.first, result2.first);
   merge(result2.second, std::move(result1.second));
+  simplify(return_code, ns);
```
romainbrenguier: the change should be documented

src/solvers/strings/string_constraint_generator_constants.cpp
```diff
@@ -86,17 +87,19 @@ std::pair<exprt, string_constraintst> add_axioms_for_cprover_string(
...
   {
     const and_exprt guard_true(guard, if_expr->cond());
     const and_exprt guard_false(guard, not_exprt(if_expr->cond()));
     return combine_results(
       add_axioms_for_cprover_string(
-        fresh_symbol, res, if_expr->true_case(), guard_true, array_pool),
+        fresh_symbol, res, if_expr->true_case(), guard_true, array_pool, ns),
       add_axioms_for_cprover_string(
-        fresh_symbol, res, if_expr->false_case(), guard_false, array_pool));
+        fresh_symbol, res, if_expr->false_case(), guard_false, array_pool, ns),
```
romainbrenguier: It looks like `add_axioms_for_cprover_string` is only called with `guard=true_exprt{}` so you could remove this argument and get rid of this `if`, that will avoid having to introduce the `ns` argument here and in the calling function (and maybe others).

src/solvers/strings/string_constraint_generator_main.cpp
JohnDumbell: Anywhere specific?

src/solvers/strings/string_constraint_generator_constants.cpp
JohnDumbell: That might reduce the arguments by one, but pretty sure everything else is still required? The guard may always be true, but the condition on the if is variable and having to do the branching / merge is still required as well right?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 46cf0721817f124a0b1ae0958549675957ea837c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121998505

src/solvers/strings/string_constraint_generator_constants.cpp
romainbrenguier: oh yes I was mistaken, the recursive call changes the guard

src/solvers/strings/string_constraint_generator_main.cpp
romainbrenguier: in the documentation of `combine_results`

romainbrenguier (approved):

src/solvers/strings/string_constraint_generator_constants.cpp
```diff
@@ -78,6 +78,7 @@ add_axioms_for_empty_string(const function_application_exprt &f)
 /// \param arg: expression of type string typet
 /// \param guard: condition under which `res` should be equal to arg
 /// \param array_pool: pool of arrays representing strings
+/// \param ns: the namespace
```
romainbrenguier: :bulb: Maybe it will be good to say that this argument is so that we can simplify some expressions. We could pass a function `simplify_expr` as argument instead to make that obvious (just an idea)

tautschnig: Or maybe we could have a class and have some const members (such as a `namespacet`)? The architecture of the string solver code makes it really painful to add new capabilities.

romainbrenguier: We could basically use the `string_constraint_generatort` class for that (either pass it as an argument of the function or make the function a member), but we should clean-up the members of the class first: the `string_constraintst` field should just be an output of the methods and not a field and I think the `messaget` field does not belong there.

tautschnig: I don't know what exactly the right solution looks like, but I get the impression that making everything free functions has been taken a little too far here. It severely hampers maintainability. The PR we are looking at could have been a one-line change, but instead it's currently +87/-48. We need to make sure we find a good balance between perfectly clean code and maintainability.

romainbrenguier: Here is the clean-up I suggest https://github.com/diffblue/cbmc/pull/4995, you could then replace the `fresh_symbol`, `array_pool`, and `ns` arguments in these functions by just a string_constraint_generatort object and then further changes like this one could be small.

JohnDumbell: I'll happily do that. I was tempted to do it originally while doing this, but the changes were just under the acceptable threshold for me.

smowton:
Is this good to go?

JohnDumbell:
Not yet, I'll make the required changes now to help clean it up then should be good.

JohnDumbell:
I've made the changes which results in the original change being one line. Right now the clean-up commits are all in separate blocks of work to make things a little easier to see what's been done, but I'll squash them together before merging.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: dad8579406f524ebec2646c2eb4fa1374da7a86e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123219879

src/solvers/strings/string_constraint_generator.h
```diff
@@ -74,392 +74,248 @@ class string_constraint_generatort final
...
 
-size_t max_printed_string_length(const typet &type, unsigned long ul_radix);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_replace(const function_application_exprt &f);
+  std::pair<exprt, string_constraintst>
+  add_axioms_for_set_length(const function_application_exprt &f);
+
+  /// \todo The specification may not be correct for the case where the
+  /// string is shorter than end. An actual java program should throw an
+  /// exception in that case.
```
romainbrenguier, line 247: :bulb: this looks outdated compared to the `add_axioms_for_substring(const function_application_exprt)` (but not related to this PR)

src/solvers/strings/string_constraint_generator.h
```diff
@@ -74,392 +74,248 @@ class string_constraint_generatort final
...
-  array_poolt &array_pool);
-std::pair<exprt, string_constraintst>
-add_axioms_for_empty_string(const function_application_exprt &f);
-
-std::pair<exprt, string_constraintst> add_axioms_for_copy(
-  symbol_generatort &fresh_symbol,
-  const function_application_exprt &f,
-  array_poolt &array_pool);
+public:
+  std::pair<exprt, string_constraintst> add_axioms_for_concat(
```
romainbrenguier, line 79: :bulb: at some point we should remove the `add_` from these method names, since it doesn't add them but return them (doesn't need to be in this PR)

src/solvers/strings/string_builtin_function.cpp
```diff
@@ -426,8 +426,8 @@ string_constraintst string_insertion_builtin_functiont::constraints(
 {
   if(args.size() == 1)
   {
-    auto pair = add_axioms_for_insert(
-      generator.fresh_symbol, result, input1, input2, args[0], array_pool);
+    auto pair = generator.add_axioms_for_insert(
+      result, input1, input2, args[0]);
```
romainbrenguier: this commit should be squashed in the previous one if it doesn't compile without it

romainbrenguier (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 88683783b67ab1152ac565c427b9077d46cc1e27).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123237780

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 303ec4d038700fca1f14adf03d9e95eab095d353).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/125072367

PR #4961, opened 2019-07-29
Symex rename functions: use NODISCARD

smowton:
These are particularly hazardous to people rebasing old branches, since they used to have void-type and side-effect one of their arguments, and now have a signature that's still compatible but which requires you to use the return value.


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: e46afa195242c3adc2aab2b971efe11dff1121ac).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121130757

kroening (approved):

PR #4964, opened 2019-07-30
Symex trace: show implicit value associated with composite declarations

hannes-steffenhagen-diffblue:
Previously when a struct- or other composite-typed variable was declared, symex_decl
would initialize its individual fields, then return an L2-renamed symbol relating to
the whole object. With field-sensitivity enabled, the returned symbol is e.g. my_struct!0@1#0,
while anybody reading from it will read my_struct!0@1#1..field, a different symbol which
to the solver is unrelated. The result was that in the trace, the declaration would show
up with the value of the free symbol my_struct!0@1#0 (most likely zero, since it can't be
referred to), while the rest of the trace may make it obvious that it actually had some other
non-zero value.

With this change we expand the struct into its constituent field symbols before recording an
SSA_stept for the trace. This restores the correspondence between what the trace reads and what
code reading from the struct reads.

Argument for the reviewers to have: should the goto_symex_statet::rename in symex_decl already be expanding like this? That function has an invariant insisting that it should return a single symbol, not a struct-exprt of field symbols, but that may not really be necessary. Note that if we change ::rename to always perform struct expansion then we will probably break any user that is applying this to an lvalue.

### DIFF
--- a/regression/cbmc/field-sensitivity-trace-wrong-counterexample-1/test.desc
+++ b/regression/cbmc/field-sensitivity-trace-wrong-counterexample-1/test.desc
@@ -1,0 +1,13 @@
+CORE
+test.c
+--trace
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+x=\{ \.x=0 \}
+^warning: ignoring
+--
+This tests for the counterexample shown in the trace having sensible values.
+In this case, we are asserting that x.x == 0, so x.x should have a value other
+than 0 if the assertion fails.
--- a/regression/cbmc/field-sensitivity-trace-wrong-counterexample-1/test.c
+++ b/regression/cbmc/field-sensitivity-trace-wrong-counterexample-1/test.c
@@ -1,0 +1,16 @@
+#include <assert.h>
+
+struct SomeStruct
+{
+  int x;
+};
+
+int main(void)
+{
+  struct SomeStruct x;
+  // this should cause
+  // visible initialization of
+  // x to value other than 0
+  // if assertion fails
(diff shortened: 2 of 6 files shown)


hannes-steffenhagen-diffblue:
This is copied from [this PR](https://github.com/diffblue/cbmc/pull/4919) because I will be maintaining this until it can be merged.

regression/cbmc/field-sensitivity-trace-wrong-counterexample-1/test.desc
```diff
@@ -0,0 +1,13 @@
+CORE
+test.c
+--trace
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+x=\{ \.x=0 \}
```
hannes-steffenhagen-diffblue, line 8: I originally had this as a different check (checking that there _was_ a non-zero assignment to `x.x`) but the `cprover-smt2` run didn't produce a valid trace at all which made this test fail.

smowton (approved):

peterschrammel (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: f458a87d8437c3d4bfae08c91dfa0238d97336e3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121252412

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 407c1497cee743c1838c9c10103cc0829de8932e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121252599

PR #4966, opened 2019-07-31
Stop unneccessary copy of lbva location info

JohnDumbell:
This now holds only the changes, so will look like this:

Changes array: [0,0,0,x,0,y]
Original array: [a,b,c,d,e,f]

Where previously it would've been a direct copy we now just have a zero in any place that has no changes, so when it merges instead of b | b, it's b | 0. In a changed scenario nothing it does the same thing.

Saw an instance where the location info was big enough that the local bit vector analysis was taking minutes to do due to the copies. 

### DIFF
--- a/src/analyses/local_bitvector_analysis.cpp
+++ b/src/analyses/local_bitvector_analysis.cpp
@@ -264,7 +264,7 @@ void local_bitvector_analysist::build()
     work_queue.pop();
 
     auto &loc_info_src=loc_infos[loc_nr];
-    auto loc_info_dest=loc_infos[loc_nr];
+    auto loc_info_dest=points_tot();
 
     switch(instruction.type)
     {


smowton:
Is the difference in any way observable? Can we test it?

JohnDumbell:
@smowton Easily observable in the wild, testable might be a bit more tricky. I'll have a look to see if what's causing this is easily reproducable.

smowton:
I mean is there a goto-instrument version of LBVA that we can print and test it's still producing sensible results?

JohnDumbell:
Not sure, I'll have a look.

JohnDumbell:
Looks like things weren't that simple as there are huge differences between the two outputs, I'll see what's wrong.

JohnDumbell:
I missed something when originally looking at the code. It doesn't deal well with large numbers, but I don't think there's a very quick solution either. 

Going to close this and will look at a proper solution if it becomes more problematic.

PR #4967, opened 2019-07-31
Don't use temporaries in remove_instanceof

smowton:
This replaces `"if(x == null) result = false; else { tmp = x->@clsid; result = (tmp == "A" || tmp == "B" || ...)`
with `(x != null && (x->@clsid == "A" || x->@clsid == "B" || ...))`

Advantages:
* The ASSERT/ASSUME pair is more obvious when expressed as a single expression rather than a small if-block, so when asserts are converted to assumes (e.g. for coverage testing) then the redundant assume can be spotted and removed.
* By operating directly on a pointer rather than using a temporary (`x->@clsid == "A"` rather than `id = x->@clsid; id == "A"`) we enable symex's value-set filtering to discard no-longer-viable candidates for "x" (in the case where 'x' is a symbol, not a general expression like `x->y->@clsid`). This means there are more clean dereference operations (ones where there is no possibility of reading a null, invalid or incorrectly-typed object), and less pessimistic aliasing assumptions possibly causing goto-symex to explore in-fact-unreachable paths.
* By not creating a small if-block for every instanceof operation, we spare symex from a fork/join sequence with attendant copying of data structures

Disadvantage:
* By not using a temporary for the class-identifier, we force symex to carry out the dereference multiple times (i.e. `x->@clsid == "A" || x->@clsid == "B"` will become `(x == &o1 ? o1 : o2).@clsid == "A" || (x == &o1 ? o1 : o2).@clsid == "B")`. This could be expensive if either the alias set or the list of allowable types is long.


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: d34916fc6e3eae582b9b133e6ca60e503bc82958).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121427927

smowton:
SVCOMP evaluation: one test made faster (goes from timing out to completing in 15s), otherwise little change. Mostly no change expected for this PR, as it will only affect accesses made after a cast but before the instance type becomes clear due to a virtual function call or similar.

JohnDumbell (approved):

jbmc/src/java_bytecode/remove_instanceof.cpp
```diff
@@ -110,70 +110,31 @@ bool remove_instanceoft::lower_instanceof(
...
-  else_block.add(code_assignt(clsid_tmp_sym.symbol_expr(), object_clsid));
+  // Replace the instanceof operation with (check_ptr != null &&
+  //  (check_ptr->@class_identifier == "A" ||
+  //   check_ptr->@class_identifier == "B" || ...
+
+  // Note we deliberately do not use a temporary for the *check_ptr redundant
+  // subexpression because by testing check_ptr->@class_identifier rather than
+  // (e.g.) tmp = check_ptr->@class_identifier followed by testing 'tmp', we
+  // increase the chances that goto-symex can reduce the alias set of check_ptr
+  // according to the condition imposed here.
```
allredj: That's all very well explained, but not as complete as the commit message. I don't think anyone would complain if you pasted it here completely, thus making the life of future maintainers ever so wonderful.

allredj (approved):
Looks good and amazingly well explained.
An improvement on one SV-COMP example is good, but it would be more reassuring to try this out with TG on some real-world benchmarks and see the effect there.

smowton:
@romainbrenguier @JohnDumbell would either of you mind trying your benchmarks on this?

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 581540c0483b35b836233dd5a190216df9ea3bf7).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122594097

JohnDumbell:
Ran my benchmark against it. Lot of methods on Tika it helped, one outlier that added about a second but reductions on others far outweigh that.

Ran one run with the change, another with the commit reverted, so otherwise it was exactly the same.

smowton:
Awesome! Suggestion: tweak your benchmark output so the summary data is easily paste-able? For example, `X methods run, Y timeouts, Z total cpu time, ...`

JohnDumbell:
That json was always intended to be consumed by something else to make a nice shiny report, but if I don't get around to it I'll just add a summary like that somewhere.

peterschrammel (approved):

PR #4968, opened 2019-07-31
Do not generate initializations for struct padding in goto-harness

hannes-steffenhagen-diffblue:
Previously the generated code for struct fields included assignments to
struct padding. This is not necessary, makes the generated code harder
to read and makes --dump-c generated code based on it not compile, so we
shouldn't do this.

### DIFF
--- a/src/goto-harness/recursive_initialization.cpp
+++ b/src/goto-harness/recursive_initialization.cpp
@@ -155,7 +155,10 @@ void recursive_initializationt::initialize_struct_tag(
   auto const &ns = namespacet{goto_model.symbol_table};
   for(auto const &component : ns.follow_tag(type).components())
   {
-    initialize(member_exprt{lhs, component}, depth, new_known_tags, body);
+    if(!component.get_is_padding())
+    {
+      initialize(member_exprt{lhs, component}, depth, new_known_tags, body);
+    }
   }
 }
 


chrisr-diffblue:
Is it worth adding a regression test (probably using a new chain.sh) that uses --dump-c and then compiles and tests the output?

hannes-steffenhagen-diffblue:
I had considered it. I don't think it makes a lot of sense to do it to test for this particular change, but it _may_ be worth having a couple tests checking that the `--dump-c` output for `goto-harness` compiles.

chrisr-diffblue:
Yeah, that's what I was thinking - the more general use of such a test, and since this is the first PR I've seen that specifically addresses what the output of --dump-c for harness looks like, I wondered if it was worth including in this PR.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 52dba357916b5eb4bb4e76361a7cb8643f70bcd6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121444685

src/goto-harness/recursive_initialization.cpp
```diff
@@ -155,7 +155,10 @@ void recursive_initializationt::initialize_struct_tag(
   auto const &ns = namespacet{goto_model.symbol_table};
   for(auto const &component : ns.follow_tag(type).components())
   {
-    initialize(member_exprt{lhs, component}, depth, new_known_tags, body);
+    if(!component.get_is_padding())
+    {
+      initialize(member_exprt{lhs, component}, depth, new_known_tags, body);
+    }
```
tautschnig, line 161: What about function-pointer members - are they correctly supported at the moment? Else those may need a special case as well.

PR #4970, opened 2019-07-31
Fix catch stringification of ireps in failure messages

thomasspriggs:
The existing debug printing of ireps in `unit_tests.cpp` was not being
called by catch for some reason. This pr adds a more thorough
implementation of the pretty printing configuration/specializtion in
`use_catch.h` instead. This implementation works for any type which can
be converted to `irept`.

### DIFF
--- a/unit/testing-utils/use_catch.h
+++ b/unit/testing-utils/use_catch.h
@@ -1,6 +1,7 @@
 /*******************************************************************\
 
 Module: Wrapper around CATCH to disable selected compiler warnings
+        and add pretty printing for failure messages involving ireps.
 
 Author: Michael Tautschnig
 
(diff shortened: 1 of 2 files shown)


thk123:
IRL Tom showed me a one line fix for this that he is working on to replace this (with a test and everything!)

thomasspriggs:
Rebased on develop just to see if it passes CI.

tautschnig:
Is this code also being called/used somewhere? If not, can we have a test that actually does use it?

thomasspriggs:
> Is this code also being called/used somewhere? If not, can we have a test that actually does use it?

The code is called by the catch framework in the case of errors for pretty printing purposes. The plan was to add a test before merging this, but it fell to the bottom of my priority list.

thomasspriggs:
> IRL Tom showed me a one line fix for this that he is working on to replace this (with a test and everything!)

I'm struggling to remember what the one-line fix was at this point and it could have been to do with the way I was using/including the framework downstream rather than missing features in the printing itself. Therefore I am going to close this out as it isn't a priority and there isn't a viable route to completing this work without something jogging my memory.

PR #4972, opened 2019-08-01
Enum range check

angelhof:
Adds checks before every appearance of an enum type expression. These checks ensure that the value of the enum expression is one of the defined enum values. It doesn't insert checks for occurrences  on the left hand side of assignments, as this introduces false positives (e.g when a variable is just declared but not initialized).

### DIFF
--- a/regression/goto-instrument/enum-range-check1/main.c
+++ b/regression/goto-instrument/enum-range-check1/main.c
@@ -1,0 +1,26 @@
+#include <stdio.h>
+
+enum day
+{
+  sunday = 1,
+  monday,
+  tuesday = 5,
+  wednesday,
+  thursday = 10,
+  friday,
+  saturday
+};
+
+int main()
+{
+  enum day temp = friday;
+  temp = sunday;
+  temp = monday;
+  temp = tuesday;
+  temp = wednesday;
+  temp = thursday;
+  temp = thursday + 1;
+  temp = saturday;
+  printf("%d\n", temp);
+  return 0;
+}
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -47,10 +47,12 @@ public:
     ns(_ns),
     local_bitvector_analysis(nullptr)
   {
+    no_enum_check = false;
     enable_bounds_check=_options.get_bool_option("bounds-check");
     enable_pointer_check=_options.get_bool_option("pointer-check");
     enable_memory_leak_check=_options.get_bool_option("memory-leak-check");
     enable_div_by_zero_check=_options.get_bool_option("div-by-zero-check");
+    enable_enum_range_check = _options.get_bool_option("enum-range-check");
     enable_signed_overflow_check=
       _options.get_bool_option("signed-overflow-check");
     enable_unsigned_overflow_check=
@@ -90,6 +92,7 @@ protected:
   std::unique_ptr<local_bitvector_analysist> local_bitvector_analysis;
   goto_programt::const_targett current_target;
   guard_managert guard_manager;
+  bool no_enum_check;
 
   /// Check an address-of expression:
   ///  if it is a dereference then check the pointer
@@ -167,6 +170,7 @@ protected:
   void div_by_zero_check(const div_exprt &, const guardt &);
   void mod_by_zero_check(const mod_exprt &, const guardt &);
   void mod_overflow_check(const mod_exprt &, const guardt &);
+  void enum_range_check(const exprt &, const guardt &);
   void undefined_shift_check(const shift_exprt &, const guardt &);
   void pointer_rel_check(const binary_relation_exprt &, const guardt &);
   void pointer_overflow_check(const exprt &, const guardt &);
@@ -220,6 +224,7 @@ protected:
   bool enable_pointer_check;
   bool enable_memory_leak_check;
   bool enable_div_by_zero_check;
+  bool enable_enum_range_check;
   bool enable_signed_overflow_check;
   bool enable_unsigned_overflow_check;
   bool enable_pointer_overflow_check;
@@ -320,6 +325,35 @@ void goto_checkt::div_by_zero_check(
     inequality,
     "division by zero",
     "division-by-zero",
+    expr.find_source_location(),
+    expr,
+    guard);
+}
+
+void goto_checkt::enum_range_check(const exprt &expr, const guardt &guard)
+{
+  if(!enable_enum_range_check || no_enum_check)
+    return;
+
+  const c_enum_tag_typet &c_enum_tag_type = to_c_enum_tag_type(expr.type());
+  symbolt enum_type = ns.lookup(c_enum_tag_type.get_identifier());
+  const c_enum_typet &c_enum_type = to_c_enum_type(enum_type.type);
+
+  const c_enum_typet::memberst enum_values = c_enum_type.members();
+
+  std::vector<exprt> disjuncts;
+  for(const auto &enum_value : enum_values)
+  {
+    const constant_exprt val{enum_value.get_value(), c_enum_tag_type};
+    disjuncts.push_back(equal_exprt(expr, val));
+  }
+
+  const exprt check = disjunction(disjuncts);
+
+  add_guarded_property(
+    check,
+    "enum range check",
+    "enum-range-check",
     expr.find_source_location(),
     expr,
     guard);
@@ -1668,6 +1702,9 @@ void goto_checkt::check_rec(const exprt &expr, guardt &guard)
   forall_operands(it, expr)
     check_rec(*it, guard);
 
+  if(expr.type().id() == ID_c_enum_tag)
+    enum_range_check(expr, guard);
+
   if(expr.id()==ID_index)
   {
     bounds_check(to_index_expr(expr), guard);
@@ -1880,7 +1917,13 @@ void goto_checkt::goto_check(
     {
       const code_assignt &code_assign=to_code_assign(i.code);
 
-      check(code_assign.lhs());
+      // Reset the no_enum_check with the flag reset for exception
+      // safety
+      {
+        flag_resett no_enum_check_flag_resetter;
+        no_enum_check_flag_resetter.set_flag(no_enum_check, true);
+        check(code_assign.lhs());
+      }
       check(code_assign.rhs());
 
       // the LHS might invalidate any assertion
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -1843,6 +1843,8 @@ void goto_checkt::goto_check(
         flag_resetter.set_flag(enable_memory_leak_check, false);
       else if(d.first == "disable:div-by-zero-check")
         flag_resetter.set_flag(enable_div_by_zero_check, false);
+      else if(d.first == "disable:enum-range-check")
+        flag_resetter.set_flag(enable_enum_range_check, false);
       else if(d.first == "disable:signed-overflow-check")
         flag_resetter.set_flag(enable_signed_overflow_check, false);
       else if(d.first == "disable:unsigned-overflow-check")
--- a/src/ansi-c/c_typecheck_type.cpp
+++ b/src/ansi-c/c_typecheck_type.cpp
@@ -1078,7 +1078,7 @@ typet c_typecheck_baset::enum_constant_type(
   }
 }
 
-typet c_typecheck_baset::enum_underlying_type(
+bitvector_typet c_typecheck_baset::enum_underlying_type(
   const mp_integer &min_value,
   const mp_integer &max_value,
   bool is_packed) const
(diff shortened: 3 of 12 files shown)


angelhof:
Why does only one build fail with error:
```
c_typecheck_type.cpp:1244:33: error: use of undeclared identifier 'as_string'
    auto value = string2integer(as_string(member.get_value()));
                                ^
1 error generated.
```
whereas all other builds succeed?

smowton:
Most likely that's a build using `typedef std::string irep_idt`. Use `id2string`, which is always defined, rather than `as_string`, which is a `dstring` function only included when `irep_idt == dstringt`

angelhof:
> Most likely that's a build using `typedef std::string irep_idt`. Use `id2string`, which is always defined, rather than `as_string`, which is a `dstring` function only included when `irep_idt == dstringt`

@smowton Thanks, this seems to work :)

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: ec750efd37d4e830abee5e7193891536d11a30e3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121635227

src/analyses/goto_check.cpp
```diff
@@ -325,6 +330,35 @@ void goto_checkt::div_by_zero_check(
...
+    return;
+
+  const c_enum_tag_typet &c_enum_tag_type = to_c_enum_tag_type(expr.type());
+  symbolt enum_type = ns.lookup(c_enum_tag_type.get_identifier());
+  const c_enum_typet &c_enum_type = to_c_enum_type(enum_type.type);
+
+  const c_enum_typet::memberst enum_values = c_enum_type.members();
+
+  std::vector<exprt> disjuncts;
+  for(auto it = enum_values.begin(); it != enum_values.end(); ++it)
```
karkhaz: `const auto`?

src/analyses/goto_check.cpp
```diff
@@ -325,6 +330,35 @@ void goto_checkt::div_by_zero_check(
...
+  {
+    const constant_exprt val = constant_exprt(it->get_value(), c_enum_tag_type);
```
smowton: `val{it->get_value(), c_enum_tag_type}`

src/analyses/goto_check.cpp
smowton: Use C++11-style for-each loop

src/analyses/goto_check.cpp
```diff
@@ -1880,7 +1917,13 @@ void goto_checkt::goto_check(
     {
       const code_assignt &code_assign=to_code_assign(i.code);
 
+      // Question: I am not sure whether adding it as a class field is
+      // better than adding it as a function argument. Adding it as an
+      // argument means that it needs to be passed as a third argument
+      // to many check_rec functions that don't really need it.
+      no_enum_check = true;
```
smowton: Not a fan of the class-field style here (it is likely to go wrong if exceptions are caught, or it may not nest well), but when can you get enum values being used on the LHS anyhow?

src/ansi-c/c_typecheck_type.cpp
```diff
@@ -1224,27 +1224,27 @@ void c_typecheck_baset::typecheck_c_enum_type(typet &type)
     // produce value for next constant
     ++value;
   }
-  
+
```
smowton: This file is only changed cosmetically in commit "Add the enum-in-range check"

src/ansi-c/c_typecheck_type.cpp
```diff
@@ -1216,19 +1216,35 @@ void c_typecheck_baset::typecheck_c_enum_type(typet &type)
...
+    // Note: The value will be correctly set to a bv type when we know
+    // the width of the bitvector
     member.set_value(integer2string(value));
     enum_members.push_back(member);
 
     // produce value for next constant
     ++value;
   }
-
+
```
smowton: Accidental change?

src/ansi-c/c_typecheck_type.cpp
```diff
@@ -1265,14 +1281,10 @@ void c_typecheck_baset::typecheck_c_enum_type(typet &type)
 
   // throw in the enum members as 'body'
   irept::subt &body=enum_tag_symbol.type.add(ID_body).get_sub();
-
+
```
smowton: Accidental change? (Consider setting your editor to truncate trailing blanks)

regression/goto-instrument/enum-range-check1/main.c
```diff
@@ -1,23 +1,26 @@
 #include <stdio.h>
 
-enum day {sunday = 1,
-          monday,
-          tuesday = 5, 
-          wednesday,
-          thursday = 10,
-          friday,
-          saturday}; 
+enum day
```
smowton, line 3: Apply clang-format as you go (e.g. do a rebase with `git-clang-format HEAD~1` after each commit)

src/ansi-c/c_typecheck_type.cpp
```diff
@@ -1241,7 +1241,7 @@ void c_typecheck_baset::typecheck_c_enum_type(typet &type)
   {
     // Note: This is inefficient as it first turns integers to strings
     // and then turns them back to bvrep
-    auto value = string2integer(as_string(member.get_value()));
+    auto value = string2integer(id2string(member.get_value()));
```
smowton, line 1244: Squash this change into the original broken commit

src/analyses/goto_check.cpp
angelhof: If I remember correctly, enum type variables are on the LHS, and this assertion always fails when they are not initialized. 

Regarding the class-field, do you think it would be better passed as an argument to check_rec?

regression/goto-instrument/enum-range-check1/main.c
angelhof: Thanks for the advice, should I squash the clang-format commit with the rest now?

src/ansi-c/c_typecheck_type.cpp
angelhof: thanks, will do :)

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 66d7172156f0929954a7b64bf9f90929193c0c87).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121771820

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 99387472f4d221f03c87b2b10d083fe399aa85c2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121772613

src/analyses/goto_check.cpp
smowton: Mmm, neither option is good is it... how about using an on-stack class instance that memorises the old state and stores it, then at least it's exception safe and nests properly, that's probably the least bad option

regression/goto-instrument/enum-range-check1/main.c
smowton: I guess once the others are incrementally formatted then it won't exist at all

src/analyses/goto_check.cpp
angelhof: I am not sure I understand your proposal. Could you clarify that? Thanks!

smowton: For example,

```c++
class disable_no_enum_checkt
{
  bool old_value;
  goto_checkt &goto_check;

public:
  disable_no_enum_checkt(goto_checkt &current_state) :
    goto_check(current_state),
    old_value(current_state.no_enum_check)
  { 
    goto_check.no_enum_check = true;
  }

  ~disable_no_enum_checkt()
  {
    goto_check.no_enum_check = old_value;
  }
};
```

Then
```c++
  {
    disable_no_enum_checkt disable_check(*this);
    process_lhs(...);
  }
```
The advantage being, (a) does the right thing in the presence of exceptions (i.e. restores the setting to its old value), and (b) notices if the check was already disabled (e.g. recursing into a function that disables it), rather than always re-enabling it on exit.

angelhof: @smowton Thanks for the suggestion! A minor comment is that this function cannot be recursively called, and so there shouldn't be any recursion issues. In that case, I guess that we can get away with:

```cpp
class disable_no_enum_checkt
{
  goto_checkt &goto_check;

public:
  disable_no_enum_checkt(goto_checkt &current_state) :
    goto_check(current_state)
  { 
    goto_check.no_enum_check = true;
  }

  ~disable_no_enum_checkt()
  {
    goto_check.no_enum_check = false;
  }
};
```

Would this work?

smowton: Personally I'd still remember the old value to defend against this code being cut-and-pasted inside the recursion at some later point

angelhof: That is a reasonable point, thanks

angelhof: I just noticed that a similar `flag_resett` class exists, and I used that instead. @smowton it seems that it does exactly what you proposed.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: a895b402415f62de4cdcc92b7115c97a4cd05596).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122197970

src/analyses/goto_check.cpp
```diff
@@ -1880,7 +1917,13 @@ void goto_checkt::goto_check(
     {
       const code_assignt &code_assign=to_code_assign(i.code);
 
+      // Question: I am not sure whether adding it as a class field is
```
smowton: This comment can go now

src/analyses/goto_check.cpp
```diff
@@ -342,9 +342,9 @@ void goto_checkt::enum_range_check(const exprt &expr, const guardt &guard)
   const c_enum_typet::memberst enum_values = c_enum_type.members();
 
   std::vector<exprt> disjuncts;
-  for(auto it = enum_values.begin(); it != enum_values.end(); ++it)
+  for(const auto it : enum_values)
```
smowton: `const auto &`, and `it` -> `enum_value` since it's not longer an iterator

smowton (approved):
Couple of minor nitpicks, and you should squash the last two fixup commits into the "main" two wherever is appropriate, then LGTM

src/analyses/goto_check.cpp
angelhof: Done, thanks

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 85b90f304ba90ca83d4f1f3c78e28293c51c057b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122303742

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1624,6 +1624,7 @@ void goto_instrument_parse_optionst::help()
     "Safety checks:\n"
     " --no-assertions              ignore user assertions\n"
     HELP_GOTO_CHECK
+    " --enum-range-check           checks that all enum type expressions have values in the enum range\n" // NOLINT(*)
```
tautschnig: Please add this to `HELP_GOTO_CHECK` rather than locally adding it here.

src/analyses/goto_check.cpp
```diff
@@ -1880,7 +1917,13 @@ void goto_checkt::goto_check(
...
       const code_assignt &code_assign=to_code_assign(i.code);
 
-      check(code_assign.lhs());
+      // Reset the no_enum_check with the flag reset for exception
+      // safety
+      {
+        flag_resett no_enum_check_flag_resetter;
+        no_enum_check_flag_resetter.set_flag(no_enum_check, true);
+        check(code_assign.lhs());
+      }
```
tautschnig, line 1928: Placing this comment at a random point: can you please also extend `ansi-c/scanner.l` to recognize enum-range as part of `named_checks` and add a `disable:` case around line 1800 in this file here (`analyses/goto_check.cpp`).

tautschnig (changes requested):
Looks good, just needs a few extensions to make sure pragmas to disable the new check work fine.

src/goto-instrument/goto_instrument_parse_options.cpp
angelhof: Thanks

src/analyses/goto_check.cpp
angelhof: That's a good point, thanks

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 933eb5e0f70637764b90bb7c8eaf77974124ab72).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122622217

tautschnig (approved):

PR #4926, opened 2019-07-18
Enable clang-format for Java

tautschnig:
This is to avoid clang-format silently aborting when encountering Java
files, as happened in #4912.

The configuration for Java likely needs tweaking to avoid
mass-reformatting, but we might do so in an incremental fashion.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- n/a Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- n/a Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4973, opened 2019-08-01
[TG-8623] Fix primitive wrapper types in JSON value retriever

antlechner:
Following the format of the json-io library, the JSON input to the value retriever does not distinguish between primitive types and their corresponding wrapper types. A class `Foo` with a field `fooField` of type Integer might appear as
```
{
  "@type":"org.cprover.Foo",
  "fooField":100
}
```
To support this "primitive representation", we need a special case in the struct assignment part.

The new function `is_primitive_wrapper_type_name` is moved here from TG. Any changes to it should be done in future refactoring PRs.


antlechner:
@tautschnig I had to reformat all of the test commits on this PR because clang-format complained. It seems that this is a consequence of https://github.com/diffblue/cbmc/pull/4926. What is the reason behind enforcing clang-format on Java source files for regression tests?

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 91e652e8eb6efa664b49692d312e36c05c583368).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121647072

tautschnig:
> @tautschnig I had to reformat all of the test commits on this PR because clang-format complained. It seems that this is a consequence of #4926.

I'm pretty sure that the clang-format configuration for Java actually should be amended in some way, which might in many cases help avoid the need for reformatting (assuming our Java tests are actually formatted in some consistent fashion...). I was in a way hoping for this to happen so that we could figure out what the correct formatting rules should be. Maybe we can do this here and now?

> What is the reason behind enforcing clang-format on Java source files for regression tests?

Previously, clang-format would abort when encountering changes to Java files, leaving the C++ files untouched even when they weren't correctly formatted.

smowton (approved):

allredj (approved):
tidy

antlechner:
@tautschnig 
> assuming our Java tests are actually formatted in some consistent fashion...

It doesn't look like they are. No format has ever been enforced for them... Do you think specifying a format for test files is generally useful? It certainly is for source code but tests are much more isolated and slightly different formats usually don't make them less readable.
> Previously, clang-format would abort when encountering changes to Java files, leaving the C++ files untouched even when they weren't correctly formatted.

That sounds like a bug? Did we report this somewhere? Are only Java files affected, or do we have other file types that could secretly silence clang-format when they are changed?

smowton:
If we want to there's an argument to git-clang-format giving wildcard excludes (i.e. *,java would do)

antlechner:
@smowton Let's look into that. I don't think the clang-format issue should block this PR so I'll rebase it and merge it if it (and its TG bump) passes.

PR #4976, opened 2019-08-01
Add an --assert-ensures check

angelhof:
Add an --assert-ensures check that asserts the ensures part of a function's contract. Many function names can be given as arguments to the check.

### DIFF
--- a/regression/contracts/assert_ensures_01/test.desc
+++ b/regression/contracts/assert_ensures_01/test.desc
@@ -1,0 +1,7 @@
+CORE
+main.c
+--assert-ensures min
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
--- a/regression/contracts/assert_ensures_02/test.desc
+++ b/regression/contracts/assert_ensures_02/test.desc
@@ -1,0 +1,7 @@
+CORE
+main.c
+--assert-ensures min
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -465,7 +465,9 @@ void code_contractst::operator()()
 
     goto_functionst::function_mapt::iterator i_it =
       goto_functions.function_map.find(INITIALIZE_FUNCTION);
-    assert(i_it != goto_functions.function_map.end());
+    INVARIANT(
+      i_it != goto_functions.function_map.end(),
+      "There must exist an INITIALIZE_FUNCTION");
 
     for(const auto &contract : summarized)
       add_contract_check(contract, i_it->second.body);
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -72,6 +72,9 @@ protected:
   void
   assert_ensures(goto_programt &goto_program, goto_programt::targett target);
 
+  goto_programt
+  make_ensures_assertion_for_call(const code_function_callt &call);
+
   void add_contract_check(
     const irep_idt &function,
     goto_programt &dest);
@@ -259,15 +262,14 @@ void code_contractst::code_contracts(
       apply_contract(goto_function.body, it);
 }
 
-void code_contractst::assert_ensures(
-  goto_programt &goto_program,
-  goto_programt::targett target)
+goto_programt code_contractst::make_ensures_assertion_for_call(
+  const code_function_callt &call)
 {
-  const code_function_callt &call = target->get_function_call();
+  goto_programt goto_assertion;
 
   // we don't handle function pointers
   if(call.function().id() != ID_symbol)
-    return;
+    return goto_assertion;
 
   const irep_idt &function = to_symbol_expr(call.function()).get_identifier();
   const symbolt &f_sym = ns.lookup(function);
@@ -277,7 +279,7 @@ void code_contractst::assert_ensures(
 
   // is there a contract?
   if(ensures.is_nil())
-    return;
+    return goto_assertion;
 
   // replace formal parameters by arguments, replace return
   replace_symbolt replace;
@@ -295,23 +297,36 @@ void code_contractst::assert_ensures(
   for(code_typet::parameterst::const_iterator p_it = type.parameters().begin();
       p_it != type.parameters().end() && a_it != call.arguments().end();
       ++p_it, ++a_it)
+  {
     if(!p_it->get_identifier().empty())
     {
       symbol_exprt p(p_it->get_identifier(), p_it->type());
       replace.insert(p, *a_it);
     }
+  }
 
   replace(ensures);
 
-  // assert the ensures part of the contract after the call
-  ++target;
-  code_assertt a(ensures);
-  goto_programt new_goto_program;
+  code_assertt assertion(ensures);
   null_message_handlert message_handler;
   // We first need to convert the expression to a goto_program
-  goto_convert(a, symbol_table, new_goto_program, message_handler, ID_C);
+  goto_convert(assertion, symbol_table, goto_assertion, message_handler, ID_C);
 
-  goto_program.insert_before_swap(target, new_goto_program);
+  return goto_assertion;
+}
+
+void code_contractst::assert_ensures(
+  goto_programt &goto_program,
+  goto_programt::targett target)
+{
+  const code_function_callt &call = target->get_function_call();
+
+  // Create the goto program for the assertion
+  goto_programt goto_assertion = make_ensures_assertion_for_call(call);
+
+  // assert the ensures part of the contract after the call
+  ++target;
+  goto_program.insert_before_swap(target, goto_assertion);
 }
 
 void code_contractst::assert_ensures_after_calls(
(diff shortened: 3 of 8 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 497b74003cfd06e1df69da6976179b1c5b4131c8).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121669324

angelhof:
@chrisr-diffblue @martin-cs could you review this PR?

kroening:
Seems sensible overall -- perhaps call it `--check-postconditions`?

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1643,6 +1659,7 @@ void goto_instrument_parse_optionst::help()
     " --nondet-static-exclude e    same as nondet-static except for the variable e\n" //NOLINT(*)
     "                              (use multiple times if required)\n"
     " --check-invariant function   instruments invariant checking function\n"
+    " --assert-ensures function    checks that function satisfies its contract ensures after each call\n" // NOLINT(*)
     " --remove-pointers            converts pointer arithmetic to base+offset expressions\n" // NOLINT(*)
```
kroening, line 1663: What does "after each call" mean?
So this is really checking the post-conditions in particular calling contexts?

angelhof: @kroening 

> Seems sensible overall -- perhaps call it `--check-postconditions`?

I agree that this is a much better name.


> What does "after each call" mean?
> So this is really checking the post-conditions in particular calling contexts?

It asserts the postcondition after all calls to the function. So it checks the postcondition if the function is reachable, and is called with its name (i.e. it wouldn't work for function pointers). If this is a problem, I could re-implement this as a check before each return point in the function body. I didn't do that initially for simplicity. What do you think?

chrisr-diffblue: Or maybe `--check-ensures` (because they are written in the source code as `__CPROVER_ensures`) or `--check-contracts` ?

src/goto-instrument/code_contracts.cpp
```diff
@@ -242,6 +259,70 @@ void code_contractst::code_contracts(
...
+    symbol_exprt ret_val(CPROVER_PREFIX "return_value", call.lhs().type());
+    replace.insert(ret_val, call.lhs());
+  }
+
+  // formal parameters
+  code_function_callt::argumentst::const_iterator a_it =
+    call.arguments().begin();
+  for(code_typet::parameterst::const_iterator p_it = type.parameters().begin();
+      p_it != type.parameters().end() && a_it != call.arguments().end();
+      ++p_it, ++a_it)
```
chrisr-diffblue, line 299: Minor style comment - I'd put explicit braces in for the body of the for loop here, even though it's technically only "one statement" in the loop.... but given its' multiple lines of text I think it would be more readable with explicit braces.

src/goto-instrument/code_contracts.cpp
```diff
@@ -242,6 +259,70 @@ void code_contractst::code_contracts(
       apply_contract(goto_function.body, it);
 }
 
+void code_contractst::assert_ensures(
+  goto_programt &goto_program,
+  goto_programt::targett target)
```
chrisr-diffblue: How about `code_function_callt` rather than `targett` and avoid the call to `get_function_call()` ?

src/goto-instrument/code_contracts.cpp
angelhof: Thanks, that makes sense :)

src/goto-instrument/code_contracts.cpp
angelhof: I addressed that by factoring out the assertion generation in a different function.

src/goto-instrument/goto_instrument_parse_options.cpp
angelhof: What about `--check-contract-postconditions` or `--check-contract-ensures`?

chrisr-diffblue: My personal opinion is that I like these `--check-blah` options to mirror whatever is in the source code - so I'd lean towards `--check-ensures` or `--check-contract-ensures` - others may have different opinions :-)

danielsn:
Would it also make sense to add an ```check-all-ensures``` flag to allow this to be turned on for all functions by default?  I wouldn't suggest holding this PR up for that, but could be a good second PR.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 167e36db0c6331bb436887fdd72f8f0975695de9).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122308767

regression/contracts/assert_ensures_01/test.desc
```diff
@@ -0,0 +1,7 @@
+CORE
+main.c
+--assert-ensures min
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
```
smowton, line 6: Could fail for other reasons -- suggest checking the particular failure line you're expecting, and/or comparing against a companion test that doesn't use `--assert-ensures` and for which verification succeeds

regression/contracts/assert_ensures_02/test.desc
```diff
@@ -0,0 +1,7 @@
+CORE
+main.c
+--assert-ensures min
```
smowton, line 3: Same goes for this test

src/goto-instrument/code_contracts.cpp
```diff
@@ -30,12 +32,18 @@ class code_contractst
 public:
   code_contractst(
     symbol_tablet &_symbol_table,
-    goto_functionst &_goto_functions):
-      ns(_symbol_table),
+    goto_functionst &_goto_functions,
+    std::list<std::string> _functions_to_assert_ensures)
```
smowton, line 36: Take by const ref, or std::move into the local variable

src/goto-instrument/code_contracts.cpp
```diff
@@ -55,6 +66,15 @@ class code_contractst
     goto_programt &goto_program,
     goto_programt::targett target);
 
+  void
+  assert_ensures_after_calls(goto_functionst::goto_functiont &goto_function);
+
+  void
+  assert_ensures(goto_programt &goto_program, goto_programt::targett target);
```
smowton, line 73: `_after_call` for consistency with function above?

src/goto-instrument/code_contracts.cpp
```diff
@@ -242,6 +262,82 @@ void code_contractst::code_contracts(
...
+
+  const irep_idt &function = to_symbol_expr(call.function()).get_identifier();
+  const symbolt &f_sym = ns.lookup(function);
+  const code_typet &type = to_code_type(f_sym.type);
+
+  exprt ensures = static_cast<const exprt &>(type.find(ID_C_spec_ensures));
+
+  // is there a contract?
+  if(ensures.is_nil())
+    return goto_assertion;
```
smowton, line 282: This intends `return None`, but is vulnerable to accidentally returning a partial program if someone moves some prep work to the top of the function, e.g. allocating a temporary, or labelled SKIP or whatever. Suggest instead this function should return `optionalt<goto_programt>` so you can use `return {}` when you intend to report failure / no-action.

src/goto-instrument/code_contracts.cpp
```diff
@@ -242,6 +262,82 @@ void code_contractst::code_contracts(
...
+
+  // replace formal parameters by arguments, replace return
+  replace_symbolt replace;
+
+  // TODO: return value could be nil
```
smowton, line 287: ? As it just not set at all? I don't think that's valid

src/goto-instrument/code_contracts.cpp
```diff
@@ -242,6 +262,82 @@ void code_contractst::code_contracts(
...
+  {
+    symbol_exprt ret_val(CPROVER_PREFIX "return_value", call.lhs().type());
+    replace.insert(ret_val, call.lhs());
+  }
+
+  // formal parameters
+  code_function_callt::argumentst::const_iterator a_it =
+    call.arguments().begin();
+  for(code_typet::parameterst::const_iterator p_it = type.parameters().begin();
+      p_it != type.parameters().end() && a_it != call.arguments().end();
```
smowton, line 298: `arguments` ending before `parameters` would be a static error AFAIK -- should probably just `INVARIANT` that you have at least enough arguments (and decide what you want to do for varargs?)

src/goto-instrument/code_contracts.cpp
```diff
@@ -372,23 +468,39 @@ void code_contractst::add_contract_check(
...
-  Forall_goto_functions(it, goto_functions)
-    code_contracts(it->second);
-
-  goto_functionst::function_mapt::iterator i_it=
-    goto_functions.function_map.find(INITIALIZE_FUNCTION);
-  assert(i_it!=goto_functions.function_map.end());
+  if(mode_assert_ensures)
+  {
+    Forall_goto_functions(it, goto_functions)
+      assert_ensures_after_calls(it->second);
```
smowton, line 474: Does this share any code with `code_contracts`? If so perhaps it should be a different pass rather than two modes of the same one

PR #4977, opened 2019-08-02
[TG-8996] Avoid redundant expressions when dereferencing

smowton:
Our dereferencing code currently special-cases nested dereferences, allowing us to turn
`**x` directly into something like `(x == &o1 ? 1 : 2)` in the best case where 'x' has two
possible aliases, rather than going via an intermediary like `*(x == &o1 ? o1 : o2)`
==> `((x == &o1 ? o1 : o2) == &o3 ? 1 : 2)`.

However, this can lead to redundnacy when o1 and o2 have overlapping alias sets, such as
`(x == &o1 ? (o1 == &o3 ? o3 : o4) : (o2 == &o3 ? o3 : o4))`

In this case it is better to use a temporary:
```
derefd_pointer = x == &o1 ? o1 : o2
result = derefd_pointer == &o3 ? o3 : o4
```

This restructuring of `value_set_dereferencet::dereference` attempts to keep the best of both
worlds by dividing the algorithm into two stages: one which determines the alias sets that
would be created, using placeholder symbols to mark where they occur in the overall expression,
followed by a commit stage which decides whether to commit the tentative result or to combine
the alias sets and use an intermediate temporary symbol instead.

The result should be that when alias sets overlap we use an intermediate variable, but when they
are disjoint we apply the double-dereference in-place.

### DIFF
--- a/regression/cbmc/double_deref/double_deref_with_pointer_arithmetic_single_alias.desc
+++ b/regression/cbmc/double_deref/double_deref_with_pointer_arithmetic_single_alias.desc
@@ -1,7 +1,7 @@
 CORE
 double_deref_with_pointer_arithmetic_single_alias.c
 --show-vcc
-\{1\} \(derefd_pointer::derefd_pointer!0#1 = address_of\(symex_dynamic::dynamic_object2\) \? main::argc!0@1#1 = 1 : symex::invalid_object!0#0 = main::argc!0@1#1\)
+\{1\} \(symex_dynamic::dynamic_object1#2\[cast\(mod #source_location=""\(main::argc!0@1#1, 2\), signedbv\[64\]\)\] = address_of\(symex_dynamic::dynamic_object2\) \? main::argc!0@1#1 = 1 : symex::invalid_object!0#0 = main::argc!0@1#1\)
 ^EXIT=0$
 ^SIGNAL=0$
 --
--- a/src/pointer-analysis/value_set_dereference.cpp
+++ b/src/pointer-analysis/value_set_dereference.cpp
@@ -30,6 +30,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/pointer_offset_size.h>
 #include <util/pointer_predicates.h>
 #include <util/range.h>
+#include <util/replace_symbol.h>
 #include <util/simplify_expr.h>
 #include <util/ssa_expr.h>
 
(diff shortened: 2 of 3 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 970421cfab60bbe7e1762926839abaf70c56ff71).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121755751

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 83575f5a598b7dd55bd5b0df2d4bb249561f5d17).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121761390

regression/cbmc/double_deref/double_deref_with_pointer_arithmetic_single_alias.desc
```diff
@@ -1,7 +1,7 @@
 CORE
 double_deref_with_pointer_arithmetic_single_alias.c
 --show-vcc
-\{1\} \(derefd_pointer::derefd_pointer!0#1 = address_of\(symex_dynamic::dynamic_object2\) \? main::argc!0@1#1 = 1 : symex::invalid_object!0#0 = main::argc!0@1#1\)
+\{1\} \(symex_dynamic::dynamic_object1#2\[cast\(mod #source_location=""\(main::argc!0@1#1, 2\), signedbv\[64\]\)\] = address_of\(symex_dynamic::dynamic_object2\) \? main::argc!0@1#1 = 1 : symex::invalid_object!0#0 = main::argc!0@1#1\)
```
tautschnig, line 4: This `64` will only apply to certain architectures. I'd suggest using `\d+` instead.

tautschnig (approved):

allredj:
Waiting for performance benchmark on this.

smowton:
SVCOMP/C benchmark results: on my 2.5-hour subsample, negligible difference

allredj:
@smowton Is it worth all that boilerplate then?

smowton:
@tautschnig can you confirm it improves things for your problematic example? Perhaps the case where it helps just isn't exercised in SVCOMP / takes longer than the timeout?

smowton:
So encouragingly that suggests this is generally neutral at least; I'll also have a go at recreating pathological cases where this is helpful

smowton:
If people end up unhappy with this btw then one possibility is to *never* use the recursive-dereference case, i.e. when an expression contains multiple nested deref operations always introduce a temporary. That would be more verbose, e.g. `derefd_pointer = x == &o1 ? o1 : o2; y = 5` in the optimal case where `*o1` and `*o2` both resolve to the same constant, or `y = derefd_pointer == &o3 ? 1 : 2` in the case where its referees are constant but distinct (cf. the cleaner expression `y = x == &o1 ? 1 : 2`).

This would mean simpler code, somewhat more verbose VCCs, and the same avoidance of redundant nested derefs as we get with this PR.

allredj:
I would advocate for cleaner code (which we care about) over smaller VCCs (which we don't really care about).

peterschrammel:
> one possibility is to never use the recursive-dereference case

If this doesn't degrade performance then this seems preferable to me - both, in terms of simpler implementation and more readable VCCs (I wouldn't say nested conditionals are easy to read).

peterschrammel (approved):

tautschnig:
@smowton Would you be able to find time to look into this one as well?

smowton:
@tautschnig looked at this one, but the rebase is difficult and people weren't happy with the implementation anyhow, so I'd encourage someone with more time to work on CBMC to take another pass at this starting with the basic idea:

In a double-deref situation where the points-to graph has at least one convergence (I recall you sent me a pathalogical case where the `next` pointer of a circular list is followed twice and we weren't certain where in the list we were starting, so each follow may access any list cell), then we might generate a redundant expression like `x == &o1 ? (y == &o3 ? result3 : result4) : (y == &o3 ? result3 : result4)`.

It would be better to generate:

```
tmp = x == &o1 ? &o3 : &o4
tmp == &o3 ? result3 : result4
```

This is particularly bad when the deref occurs on the LHS (e.g. `node->next->prev = node`), because this can lead to many more conditional-updates than are really necessary.

Using some heuristic or other, I recommend noticing when value_set_dereferencet is about to / has already generated either an expression containing such redundancy or just any expression above some size and introducing temporaries for the first deref instead of pushing the second deref in.

As a first step, you should dig up your circular-list pathological case that stimulated this work to begin with.

tautschnig:
Closing as #5494 provides a different implementation towards the same goal.

PR #4980, opened 2019-08-02
Made get_all_generic_parameters a non-private function

NathanJPhillips:
This moves a piece of code to a place where it can be made publicly visible.

- N/A Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).

### DIFF
--- a/jbmc/src/java_bytecode/java_types.cpp
+++ b/jbmc/src/java_bytecode/java_types.cpp
@@ -839,6 +839,31 @@ bool equal_java_types(const typet &type1, const typet &type2)
   return (type1 == type2 && arrays_with_same_element_type);
 }
 
+std::vector<java_generic_parametert>
+get_all_generic_parameters(const typet &type)
+{
+  std::vector<java_generic_parametert> generic_parameters;
+  if(is_java_implicitly_generic_class_type(type))
+  {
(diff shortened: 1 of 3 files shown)


jbmc/src/java_bytecode/java_types.cpp
```diff
@@ -839,6 +839,34 @@ bool equal_java_types(const typet &type1, const typet &type2)
   return (type1 == type2 && arrays_with_same_element_type);
 }
 
+std::vector<java_generic_parametert>get_all_generic_parameters(
+  const typet &type)
+{
+  PRECONDITION(
```
thk123: :bulb: Now this is a more general function, perhaps returning an empty vector for_non_ generic types makes more sense than crashing (it isn't obvious from the name or signature that this only works on specific kinds of types).

thomasspriggs (approved):

thk123 (approved):
Obviously I've seen this used in TG - but a unit test would be a good idea (perhaps in a follow up PR that also addresses the precondition comment?)

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d3f55b5af874934afd782304cf718de679aa8b91).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/121793624

PR #4985, opened 2019-08-05
Insert final assert false pass

nchong-at-aws:
A `goto-instrument` pass that adds `assert(false)` to the end of a function body---usually expected to be the test harness. If verification still passes then this might indicate inconsistent assumptions. Useful as an automatic check for vacuous proofs.

### DIFF
--- a/src/goto-instrument/goto_instrument_parse_options.cpp
+++ b/src/goto-instrument/goto_instrument_parse_options.cpp
@@ -86,6 +86,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "function.h"
 #include "havoc_loops.h"
 #include "horn_encoding.h"
+#include "insert_final_assert_false.h"
 #include "interrupt.h"
 #include "k_induction.h"
 #include "mmio.h"
@@ -255,6 +256,20 @@ int goto_instrument_parse_optionst::doit()
       }
 
       return CPROVER_EXIT_SUCCESS;
+    }
+
+    if(cmdline.isset("insert-final-assert-false"))
+    {
+      log.status() << "Inserting final assert false" << messaget::eom;
+      bool fail = insert_final_assert_false(
+        goto_model,
+        cmdline.get_value("insert-final-assert-false"),
+        ui_message_handler);
+      if(fail)
+      {
+        return CPROVER_EXIT_INTERNAL_ERROR;
+      }
+      // otherwise, fall-through to write new binary
     }
 
     if(cmdline.isset("show-value-sets"))
@@ -1648,6 +1663,7 @@ void goto_instrument_parse_optionst::help()
     " --undefined-function-is-assume-false\n"
     // NOLINTNEXTLINE(whitespace/line_length)
     "                              convert each call to an undefined function to assume(false)\n"
+    HELP_INSERT_FINAL_ASSERT_FALSE
     HELP_REPLACE_FUNCTION_BODY
     HELP_ANSI_C_LANGUAGE
     "\n"
--- a/src/goto-instrument/insert_final_assert_false.cpp
+++ b/src/goto-instrument/insert_final_assert_false.cpp
@@ -1,0 +1,60 @@
+/*******************************************************************\
+
+Module: Insert assert(false) at end of entry function
+
+Author: Nathan Chong, Kareem Khazem
+
+\*******************************************************************/
+
+/// \file
+/// Insert final assert false into a function
+
+#include "insert_final_assert_false.h"
+
+#include <goto-programs/goto_model.h>
+#include <util/irep.h>
+
+#include <iterator>
+#include <list>
+
+insert_final_assert_falset::insert_final_assert_falset(
+  message_handlert &_message_handler)
+  : log(_message_handler)
+{
+}
+
+bool insert_final_assert_falset::
+operator()(goto_modelt &goto_model, const std::string &function_to_instrument)
+{
+  const irep_idt entry(function_to_instrument);
+  auto it = goto_model.goto_functions.function_map.find(entry);
+  if(it == goto_model.goto_functions.function_map.end())
+  {
+    log.error() << "insert-final-assert-false: could not find function "
+                << "'" << function_to_instrument << "'" << messaget::eom;
+    return true;
+  }
+  goto_programt &entry_function = (it->second).body;
+  goto_programt::instructionst &instructions = entry_function.instructions;
+  auto instr_it = instructions.end();
+  auto last_instruction = std::prev(instr_it);
+  DATA_INVARIANT(
+    last_instruction->is_end_function(),
+    "last instruction in function should be END_FUNCTION");
+  source_locationt assert_location = last_instruction->source_location;
+  assert_location.set_property_class(ID_assertion);
+  assert_location.set_comment("insert-final-assert-false (should fail) ");
+  goto_programt::instructiont false_assert =
+    goto_programt::make_assertion(false_exprt(), assert_location);
+  entry_function.insert_before_swap(last_instruction, false_assert);
+  return false;
+}
+
+bool insert_final_assert_false(
+  goto_modelt &goto_model,
+  const std::string &function_to_instrument,
+  message_handlert &message_handler)
+{
+  insert_final_assert_falset ifaf(message_handler);
+  return ifaf(goto_model, function_to_instrument);
+}
--- a/src/goto-instrument/insert_final_assert_false.h
+++ b/src/goto-instrument/insert_final_assert_false.h
@@ -1,0 +1,59 @@
+/*******************************************************************\
+
+Module: Insert assert(false) at end of entry function
+
+Author: Nathan Chong, Kareem Khazem
+
+\*******************************************************************/
+
+/// \file
+/// Insert final assert false into a function
+
+/// This transform can be used to check for some cases of vacuous proofs
+/// Usually the given function is the test harness / entry point
+/// The instrumented assert is expected to fail
+/// If it does not then this may indicate inconsistent assumptions
+
+#ifndef CPROVER_GOTO_INSTRUMENT_INSERT_FINAL_ASSERT_FALSE_H
+#define CPROVER_GOTO_INSTRUMENT_INSERT_FINAL_ASSERT_FALSE_H
+
+#include <string>
+
+#include <util/message.h>
+
+class goto_modelt;
+
+class insert_final_assert_falset
+{
+public:
+  explicit insert_final_assert_falset(message_handlert &_message_handler);
+  bool operator()(goto_modelt &, const std::string &);
+
+private:
+  messaget log;
+};
+
+/// Transform a goto program by inserting assert(false) at the end of a given
+/// function \p function_to_instrument. The instrumented assert is expected
+/// to fail. If it does not then this may indicate inconsistent assumptions.
+///
+/// \param goto_model: The goto program to modify.
+/// \param function_to_instrument: Name of the function to instrument.
+/// \param message_handler: Handles logging.
+/// \returns false on success
+bool insert_final_assert_false(
+  goto_modelt &goto_model,
+  const std::string &function_to_instrument,
+  message_handlert &message_handler);
+
+// clang-format off
+#define OPT_INSERT_FINAL_ASSERT_FALSE \
+  "(insert-final-assert-false):"
+
+#define HELP_INSERT_FINAL_ASSERT_FALSE \
+  " --insert-final-assert-false <function>\n" \
+  /* NOLINTNEXTLINE(whitespace/line_length) */ \
+  "                              generate assert(false) at end of function\n"
+// clang-format on
+
+#endif // CPROVER_GOTO_INSTRUMENT_INSERT_FINAL_ASSERT_FALSE_H
--- a/regression/goto-instrument/insert-final-assert-false1/main.c
+++ b/regression/goto-instrument/insert-final-assert-false1/main.c
@@ -1,0 +1,13 @@
+#include <stdlib.h>
+
+int main()
+{
+  size_t n;
+  int *p = (int *)malloc(n);
+  int *q = (int *)malloc(n);
+  __CPROVER_assert(p, "not-null");
+  __CPROVER_assert(q, "not-null");
+  // the following is the same as assuming false
+  __CPROVER_assume(__CPROVER_POINTER_OBJECT(p) == __CPROVER_POINTER_OBJECT(q));
+  return 0;
+}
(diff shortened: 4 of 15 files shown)


src/goto-instrument/insert_final_assert_false.h
```diff
@@ -0,0 +1,48 @@
...
+#ifndef CPROVER_GOTO_INSTRUMENT_INSERT_FINAL_ASSERT_FALSE_H
+#define CPROVER_GOTO_INSTRUMENT_INSERT_FINAL_ASSERT_FALSE_H
+
+#include <string>
+
+#include <util/message.h>
+
+class goto_modelt;
+
+class insert_final_assert_falset : public messaget
```
karkhaz: I would prefer that this class has a `messaget` as a member, rather than inheriting from it. I would do
```c++
private:
  messaget &log;
```
and the constructor can be:
```c++
insert_final_assert_falset::insert_final_assert_falset(
  message_handlert &_message_handler)
  : log(_message_handler)
```
then you can call `log.error()` instead of just `error()` and so on.

src/goto-instrument/insert_final_assert_false.h
```diff
@@ -0,0 +1,48 @@
...
+{
+public:
+  explicit insert_final_assert_falset(message_handlert &_message_handler);
+  bool operator()(goto_modelt &, const std::string &);
+};
+
+bool do_insert_final_assert_false(
```
karkhaz: Doxygen comment here please. In particular, please note that the second parameter is the name of the function that you would like to instrument with `assert(false)`.

src/goto-instrument/insert_final_assert_false.cpp
```diff
@@ -0,0 +1,61 @@
...
+bool insert_final_assert_falset::
+operator()(goto_modelt &goto_model, const std::string &function_to_instrument)
+{
+  const irep_idt entry(function_to_instrument);
+  auto it = goto_model.goto_functions.function_map.find(entry);
+  if(it == goto_model.goto_functions.function_map.end())
+  {
+    error() << "insert-final-assert-false: could not find function "
+            << "'" << function_to_instrument << "'" << eom;
+    return false;
```
karkhaz: CBMC convention is to return `true` on failure, `false` on success.

src/goto-instrument/insert_final_assert_false.cpp
```diff
@@ -0,0 +1,61 @@
...
+    last_instruction->is_end_function(),
+    "last instruction in function should be END_FUNCTION");
+  goto_programt::targett assert_target =
+    entry_function.insert_before(last_instruction);
+  exprt false_expr;
+  false_expr.make_bool(false);
+  assert_target->make_assertion(false_expr);
+  assert_target->source_location.set_comment(
+    "added by insert-final-assert-false pass");
+  return true;
```
karkhaz: `return false;` please.

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -257,6 +258,19 @@ int goto_instrument_parse_optionst::doit()
       return CPROVER_EXIT_SUCCESS;
     }
 
+    if(cmdline.isset("insert-final-assert-false"))
+    {
+      log.status() << "Inserting final assert false" << messaget::eom;
+      bool success = do_insert_final_assert_false(
```
karkhaz: `bool fail = ...` here please, see below comments about `true` indicating failure

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -257,6 +258,19 @@ int goto_instrument_parse_optionst::doit()
...
+        goto_model,
+        cmdline.get_value("insert-final-assert-false"),
+        ui_message_handler);
+      if(!success)
```
karkhaz: We should also return & exit on success, right? Suggest
`return fail ? CPROVER_EXIT_INTERNAL_ERROR : CPROVER_EXIT_SUCCESS;`

karkhaz (changes requested):
Thank you for this instrumentation pass, it will be very useful!

Apart from the comments, please could you have a go at writing a regression test? You already wrote a few small programs while creating this patch, it should be reasonably easy to adapt them to a test somewhere under the `regression/` directory.

kroening:
@nchong-at-aws The way people usually check whether there are inconsistent assumptions is using --cover location. Is that good for your use-case?

karkhaz:
@kroening many of our proofs don't have 100% coverage, for various valid reasons. If using `--cover`, we would need to check for sudden *drops* in coverage, rather than checking for less-than-100%. That implies that we need to save the state of the previous coverage check.

@nchong-at-aws's patch is thus very useful because it provides a stateless way of alerting on inconsistent assumptions.

This check is also much simpler and much more black-or-white. There can be nuanced reasons for less-than-100% coverage. But it is certainly a bad sign if code instrumented with this pass successfully verifies, no doubt about it.

src/goto-instrument/goto_instrument_parse_options.cpp
smowton: Remove `do_` prefix from name (already has a verb)

smowton (approved):
No further comments besides @karkhaz's. From a UI perspective, would `--assert-function-does-not-terminate` make more sense, since that's what we're actually introducing?

src/goto-instrument/insert_final_assert_false.h
nchong-at-aws: Thanks Kareem. There's a note in `CODING_STANDARD.md` that suggests deriving from `messaget`. Which way is preferable?

src/goto-instrument/insert_final_assert_false.h
nchong-at-aws: Done.

src/goto-instrument/insert_final_assert_false.cpp
nchong-at-aws: Done.

src/goto-instrument/goto_instrument_parse_options.cpp
nchong-at-aws: Done.

src/goto-instrument/goto_instrument_parse_options.cpp
nchong-at-aws: We need to fall-through because we want to write out the modified goto program.

src/goto-instrument/insert_final_assert_false.cpp
nchong-at-aws: Done.

src/goto-instrument/goto_instrument_parse_options.cpp
nchong-at-aws: @smowton Thanks for the feedback. I fixed the name of the function as suggested. Regarding calling the switch `--assert-function-does-not-terminate`. So if I have this correct if the assert passes (resp. fails) then the function does not terminate (does terminate)? The difficulty here is the "does not terminate" reason could be for multiple reasons: unbounded loop, not enough loop unrollings, inconsistent assumptions.

nchong-at-aws:
@kroening Yes, cover is useful for seeing this problem too. This check is another way to sanity check a result.

src/goto-instrument/insert_final_assert_false.h
karkhaz: Ah, I remembered being told the opposite by a reviewer at some point. I suppose the CODING_STANDARD.md is the truth. @smowton do you have an opinion on this?

(FWIW I still believe having a messaget member is cleaner, but I'll defer to whatever the prevalent style is)

src/goto-instrument/goto_instrument_parse_options.cpp
karkhaz: @smowton I'm not sure I agree with the choice of name, this is only about checking for inconsistent assumptions---not asserting that the function doesn't terminate.

If you'd like the switch to reflect what it's for, perhaps something like `--check-for-inconsistent-assumptions` or `--ensure-consistent-assumptions`? However these also seem somewhat confusing to me---if the assumptions are inconsistent, the false assertion will actually *pass* rather than fail, which is different to all the other `--check-*` flags.

smowton: Hmm, even if that's what it's for, it'll still pass/fail according to whether the function contains an infinite loop or similar, right? It seems like the only concrete information you could ascertain is "assertion reachable (i.e., fails)" -> "function can somehow terminate, i.e. is not impeded by inconsistent assumptions, unwind limit, depth limit, etc etc etc". In other words is this actually a version of `--cover`, with the specific coverage goal "end of specified function"?

smowton: (actually not quite, since `--cover` would usually convert existing assertions to assumptions which this does not)

src/goto-instrument/insert_final_assert_false.cpp
```diff
@@ -0,0 +1,63 @@
...
+  auto last_instruction = std::prev(instr_it);
+  DATA_INVARIANT(
+    last_instruction->is_end_function(),
+    "last instruction in function should be END_FUNCTION");
+  last_instruction->turn_into_skip();
+  goto_programt::targett assert_target = entry_function.add_instruction();
+  exprt false_expr;
+  false_expr.make_bool(false);
+  assert_target->make_assertion(false_expr);
+  assert_target->source_location.set_comment(
```
kroening: Instead of the `.make_bool(...)` simply write
`false_exprt()`
(this is in util/std_expr.h)

src/goto-instrument/insert_final_assert_false.cpp
```diff
@@ -0,0 +1,63 @@
...
+  goto_programt &entry_function = (it->second).body;
+  goto_programt::instructionst &instructions = entry_function.instructions;
+  auto instr_it = instructions.end();
+  auto last_instruction = std::prev(instr_it);
+  DATA_INVARIANT(
+    last_instruction->is_end_function(),
+    "last instruction in function should be END_FUNCTION");
+  last_instruction->turn_into_skip();
+  goto_programt::targett assert_target = entry_function.add_instruction();
+  exprt false_expr;
```
kroening: The sequence of overwriting, adding, and then producing a new end_function instruction could be shortened by simply saying
`entry_function.insert_before_swap(last_instruction);
*last_instruction=goto_programt::make_assertion(false_exprt());
`

kroening (approved):

src/goto-instrument/insert_final_assert_false.h
chrisr-diffblue: I wouldn't trust the CODING_STANDARD.md as far as that, to be honest. `insert_final_assert_falset` isn't a type of `messaget`, it just wants to use it - so it shouldn't derive from `messaget`. If the coding standard still says that, then it should be updated. :-)

src/goto-instrument/insert_final_assert_false.cpp
nchong-at-aws: Done.

src/goto-instrument/insert_final_assert_false.cpp
nchong-at-aws: Done.

src/goto-instrument/insert_final_assert_false.h
nchong-at-aws: I went with messaget as a class member. I'll leave updating the CODING_STANDARD to someone else!

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 11c2f8ac915acff14c1e5d8ae650e07ec4f53fef).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123517019

nchong-at-aws:
@allredj I think this could be because I'm a first time contributor?

src/goto-instrument/goto_instrument_parse_options.cpp
karkhaz: Yes. The intention is that we would run CBMC on a binary instrumented with this flag, in parallel with an otherwise-identical binary that has not. Both CBMC runs would include exactly the same flags, including `--unwind` (which I believe one must always use on any nontrivial program anyway). So reachability and termination are for granted, unless the entire function that we're instrumenting is dead/unreachable.

This flag really is intended to show up contradictory assumptions, and nothing else. If you think there are other scenarios that might cause the assertion to be hit (or not hit), @nchong-at-aws should fix that. But if there are no such cases, then I think the flag should be precisely named to reflect either what it does (`--insert-final-assert-false`) or what it's for (`--ensure-consistent-assumptions`).

kroening:
This is now getting ready to merge -- may I suggest that you squash the commits into one.

src/goto-instrument/goto_instrument_parse_options.cpp
smowton: So if the input program was
```
void f(int x) {
  assert(x == 5);
  assume(x % 2 == 0);
  assume(x % 2 == 1);
  // assert false introduced here
}
```
Then your non-instrumented binary would report failing the assertion `x == 5` and the instrumented one would fail that assertion but not the `assert(false)` introduced at the end? That would indicate that maybe there are contradictory assumptions (as there are in the example), but this example would behave the same, assuming a reasonable unwind limit:
```
void f(int x) {
  assert(x == 5);
  for(int i = 0; i < 1000000; ++i) { } // Presumably we have --unwind < 1000000
  // assert false introduced here
}
```
So I don't see how this plan disambiguates ability to reach the end due to assumptions vs. ability to reach the end due to unwind or depth limitation.

allredj:
@nchong-at-aws Very probably. You can ignore the message.

src/goto-instrument/goto_instrument_parse_options.cpp
nchong-at-aws: I agree with you. There are multiple reasons why the instrumented assert could pass. This makes me lean towards the flag saying what it does and then leaving the "interpretation" of what to do with the result with the user. The way I imagine using this flag is after the harness has all-passes wrt its check-* and user-provided asserts.

karkhaz (approved):

src/goto-instrument/goto_instrument_parse_options.cpp
smowton: Fair enough, let's go with that.

smowton:
@nchong-at-aws looks like this is good to go, please squash commits

src/goto-instrument/goto_instrument_parse_options.cpp
nchong-at-aws: Thanks! I'll close this conversation.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 639ef8195ce865d800bffcca853440b61f7b8e04).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123735093

PR #4991, opened 2019-08-06
introduce ID_array_comprehension

kroening:
The ID is introduced to avoid an inconsistency between class name and the ID
that was introduced with 2fbe5e1e29.

This will allow the introduction of lambda expressions, to mean nameless
(mathematical) functions.


smowton:
I didn't do this when naming `array_comprehension_exprt` because I thought it would break GOTO binaries -- should this be reserved for a major release?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: b5ad3e432db7c944c9029efcecafb015fee07daa).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122237743

tautschnig (changes requested):
Looks good, but as @smowton said this requires a bump of the goto binary version.

kroening:
No, there is no way to get that into a goto binary.

tautschnig:
> No, there is no way to get that into a goto binary.

Ah, the only places that generated an `array_comprehension_exprt` were back-end components and thus it was never written to disk.

tautschnig (approved):

PR #4992, opened 2019-08-06
let_exprt can now do multiple bindings

kroening:
This matches what the SMT-LIB standard does.  It also matches what our own
`binding_exprt` can do, i.e., `let_exprt` can be re-written into a lambda
binding plus a function application in the standard way.

### DIFF
--- a/src/util/std_expr.h
+++ b/src/util/std_expr.h
@@ -4327,48 +4327,112 @@ public:
 };
 
 /// \brief A let expression
-class let_exprt : public ternary_exprt
+class let_exprt : public binary_exprt
 {
 public:
-  let_exprt(symbol_exprt symbol, exprt value, const exprt &where)
-    : ternary_exprt(
+  let_exprt(
+    binding_exprt::variablest variables,
+    operandst values,
+    const exprt &where)
+    : binary_exprt(
+        binding_exprt(
+          ID_let_binding,
+          std::move(variables),
+          where,
+          where.type()),
         ID_let,
-        std::move(symbol),
-        std::move(value),
-        where,
+        multi_ary_exprt(irep_idt(), std::move(values), typet(ID_tuple)),
         where.type())
   {
+    PRECONDITION(this->variables().size() == this->values().size());
   }
 
+  /// convenience constructor for the case of a single binding
+  let_exprt(symbol_exprt symbol, exprt value, const exprt &where)
+    : let_exprt(
+        binding_exprt::variablest{std::move(symbol)},
+        operandst{std::move(value)},
+        where)
+  {
+  }
+
+  binding_exprt &binding()
+  {
+    return static_cast<binding_exprt &>(op0());
+  }
+
+  const binding_exprt &binding() const
+  {
+    return static_cast<const binding_exprt &>(op0());
+  }
+
+  /// convenience accessor for the symbol of a single binding
   symbol_exprt &symbol()
   {
-    return static_cast<symbol_exprt &>(op0());
+    auto &variables = binding().variables();
+    PRECONDITION(variables.size() == 1);
+    return variables.front();
   }
 
+  /// convenience accessor for the symbol of a single binding
   const symbol_exprt &symbol() const
   {
-    return static_cast<const symbol_exprt &>(op0());
+    const auto &variables = binding().variables();
+    PRECONDITION(variables.size() == 1);
+    return variables.front();
   }
 
+  /// convenience accessor for the value of a single binding
   exprt &value()
   {
-    return op1();
+    auto &values = this->values();
+    PRECONDITION(values.size() == 1);
+    return values.front();
   }
 
+  /// convenience accessor for the value of a single binding
   const exprt &value() const
   {
-    return op1();
+    const auto &values = this->values();
+    PRECONDITION(values.size() == 1);
+    return values.front();
   }
 
+  operandst &values()
+  {
+    return static_cast<multi_ary_exprt &>(op1()).operands();
+  }
+
+  const operandst &values() const
+  {
+    return static_cast<const multi_ary_exprt &>(op1()).operands();
+  }
+
+  /// convenience accessor for binding().variables()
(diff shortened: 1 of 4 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 13c457b8c5ce26b10321749b3ee9e659d49b2811).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122229305

src/util/std_expr.h
```diff
@@ -4283,47 +4283,81 @@ class infinity_exprt : public nullary_exprt
...
+          typet(ID_tuple)),
         ID_let,
-        std::move(symbol),
-        std::move(value),
         where,
         where.type())
   {
   }
 
+  symbol_value_sequencet &symbol_value_sequence()
```
smowton: So someone retrieving this needs to know it's an alternating sequence of symbol, value, symbol, value? Seems better to either have a sequence of 2-tuples called bindings, or else provide accessors for nth_symbol, nth_value etc. Either of these can also enforce that the symbols are symbol_exprt.

kroening: I did consider both options.
Problems of the "accessor" option are that a) you can't easily iterate, and b) it kills the USE_LIST option for performance reasons (but we may want to kill that anyway).
The two-tuple option would be somewhat more memory-consuming, but maybe that's not a genuine issue.

src/util/std_expr.h
```diff
@@ -4283,47 +4283,81 @@ class infinity_exprt : public nullary_exprt
...
+          ID_tuple,
+          {std::move(symbol), std::move(value)},
+          typet(ID_tuple)),
+        ID_let,
+        where,
+        where.type())
+  {
+  }
+
+  using symbol_value_sequencet = std::vector<exprt>;
```
tautschnig: I know this isn't the first time/only place we are doing this, but we really shouldn't: this should be `exprt::operandst`, unless we are very intentionally (but still somewhat questionably) using a subtype that we know has the same memory layout (as was the case for `binding_exprt`).

src/util/std_expr.h
```diff
@@ -4283,47 +4283,81 @@ class infinity_exprt : public nullary_exprt
...
-        std::move(symbol),
-        std::move(value),
         where,
         where.type())
   {
   }
 
+  symbol_value_sequencet &symbol_value_sequence()
+  {
+    return (symbol_value_sequencet &)op0().operands();
```
tautschnig: I do find these type casts scary. Right now, this one actually is unnecessary. But if in future there were to be a difference then only runtime crashes would tell us.

tautschnig (commented):
It doesn't strike me as particularly great design to build something rather fragile just to satisfy a potential, yet unproven future use.

src/util/std_expr.h
kroening: I don't see much of an alternative to strengthen typing while still exposing something that can be iterated over.
An option could be to invent a custom range-like setup, which would essentially wrap the underlying iterator. That would eliminate the scary casts at the level of the container.

kroening: I'd be happy to look into this, but given that this is a very global problem, I would declare that to be outside of the scope of this PR.

tautschnig: Well, can we at least not make it any worse? I.e., not introduce yet another local definition of what we believe to be the type of `exprt::operands()`?

src/util/std_expr.h
kroening: Now with the "sequence of 2-tuples" option.

src/util/std_expr.h
kroening: Without that, it's not possible to type the symbols.

src/util/std_expr.h
```diff
@@ -4283,47 +4283,109 @@ class infinity_exprt : public nullary_exprt
 };
 
 /// \brief A let expression
-class let_exprt : public ternary_exprt
+class let_exprt : public binary_exprt
 {
 public:
-  let_exprt(symbol_exprt symbol, exprt value, const exprt &where)
-    : ternary_exprt(
+  class let_binding_exprt : public binary_exprt
```
tautschnig: Another question: why is a `let_binding_exprt` not at all related to a `binding_exprt`? That seems surprising to me.

kroening: This is not trivial: there's only one "where", and not one per binding, whereas the let bindings need an additional "value". Thus, the "is-a" relationship is not easy to establish.

tautschnig: Ack, but then the `binding_exprt` may need to be reconsidered if it isn't actually as generic as was hoped? Maybe it should just not be called `binding_exprt` and instead its API should be folded into that of a `quantifier_exprt`? I would just like to avoid confusion.

kroening: I did consider that, but it does fit `array_comprehension_exprt`. That would then need to be a quantifier (odd) or have it's own API.

kroening: Another idea. An option could be that `let_exprt` gets a `binding_exprt` as operand!

tautschnig: That would seem nice and consistent with the other uses of `binding_exprt`!

kroening: Ok, will go for it!

kroening: done

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: eba457866078e68596fba7636b0e63b8816f7f0d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122360496

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 71a20e96589a1862c6bf3b9bd2719161603c9e36).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122980484

peterschrammel (approved):

PR #4993, opened 2019-08-07
goto-symex: Reformat parameter conversion error message

tautschnig:
Language front-ends cannot always type check parameter conversions, for
example in case of incomplete forward declaration there isn't sufficient
information available. Thus some of that work is left to be done during
symbolic execution. In such cases, late type checking can fail and the
problem needs to be reported back to the users. We previously did not
include the source location of the problem in the output, and did not
make generous use of newlines, making diagnosing the problem difficult.

### DIFF
--- a/src/goto-symex/symex_function_call.cpp
+++ b/src/goto-symex/symex_function_call.cpp
@@ -116,15 +116,16 @@ void goto_symext::parameter_assignments(
               from_integer(0, index_type()),
               parameter_type);
         }
+        // clang-format on
         else
         {
           std::ostringstream error;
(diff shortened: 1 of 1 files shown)


smowton (approved):

owen-mc-diffblue (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 0b01b29fce3eb16ebd58a151b963477d5a7d4531).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122254808

PR #4998, opened 2019-08-08
[TG-8994] Tolerate a constant on the LHS of an assignment

smowton:
This can happen due to a parital write to a string constant, e.g. *(int*)"hello world".
In real programs these would fault, but we must tolerate them because symex's over-approximate
alias analysis means it may think they can happen when in fact they are forbidden by other
constraints.

### DIFF
--- a/regression/cbmc/lhs-pointer-aliases-constant/test.c
+++ b/regression/cbmc/lhs-pointer-aliases-constant/test.c
@@ -1,0 +1,15 @@
+#include <assert.h>
+
+int main(int argc, char **argv)
+{
+  int x;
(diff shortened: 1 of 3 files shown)


owen-mc-diffblue (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: e98540fb8faac8bee05af8f0ea72606695a8a3da).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122587749

peterschrammel (approved):

PR #5000, opened 2019-08-08
Refine CPPLINT script to not report __CPROVER_ usage outside of strings

angelhof:
This commit refines the CPPLINT script to only report `__CPROVER_` usages in strings (and not in function names for example).

I also removed several unnecessary NOLINT comments.

### DIFF
--- a/scripts/cpplint.py
+++ b/scripts/cpplint.py
@@ -4605,7 +4605,7 @@ def CheckAssert(filename, clean_lines, linenum, error):
 
 
 def Check__CPROVER_(filename, clean_lines, linenum, error):
-  """Check for uses of __CPROVER_.
+  """Check for uses of __CPROVER_ in strings.
 
   Args:
     filename: The name of the current file.
@@ -4614,7 +4614,7 @@ def Check__CPROVER_(filename, clean_lines, linenum, error):
     error: The function to call with any errors found.
   """
   line = clean_lines.lines[linenum]
-  match = Match(r'.*__CPROVER_.*', line)
+  match = Match(r'.*"([^"\\]|\\.)*__CPROVER_([^"\\]|\\.)*".*', line)
   if match:
     error(filename, linenum, 'build/deprecated', 4,
           'use CPROVER_PREFIX instead of __CPROVER_')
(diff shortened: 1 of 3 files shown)


tautschnig (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 60c49fdaf6fc260a846831e67a14866de5d01c67).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122513922

scripts/cpplint.py
```diff
@@ -4614,7 +4614,7 @@ def Check__CPROVER_(filename, clean_lines, linenum, error):
     error: The function to call with any errors found.
   """
   line = clean_lines.lines[linenum]
-  match = Match(r'.*__CPROVER_.*', line)
+  match = Match(r'.*"([^"\\]|\\.)*__CPROVER_([^"\\]|\\.)*".*', line)
```
thk123, line 4617: :question: I don't quite understand this regex, probably because I'm not clear on what ends up being escaped in the final regex - it'd be really nice to write a quick test that shows what this does and does not match. Specifically, does it handle:

```
std::string a_really_long_string = "some initial part \
  __CPROVER_gumpf";
```

I suppose it doesn't matter hugely since more likely to do a series of concatenating strings anyway

thk123 (approved):
I won't block this PR as cpplint.py doesn't even have any tests running in CI - but I think it'd be really great to have a few tests of that regex so it is clearer what it does and does not match

scripts/cpplint.py
angelhof: If I am not mistaken, the lint script checks each line separately, so indeed this wouldn't be caught by this regex. Note that this regex also (erroneously) complains about the following:
```
... "foo"; __CPROVER_assume(true); "bar";
``` 
though I don't think this ever appears in practice. Regarding tests, there exists a regression test directory `cpp-linter`. However most of the tests fail on my machine, and it doesn't seem to be included in the standard regression suite.

thk123: I'd forgotten about those tests, and yup they look quite broken, might try and tidy them up. I was more thinking along the lines of a python unit test - but no worries if not.

PR #5003, opened 2019-08-09
[TG-9002] Symex: revert guards after call return

smowton:
Executing a function may have a cumulative effect on the state guard. For example, if the callee contained ASSUME statements that rendered one or more control-flow options unviable then the guard might still embody that restriction (i.e. for if(x) ASSUME(false) the guard might still be `!x`). However, on function return we know that all control-flow paths have converged or been shown to be unviable, so we can restore the simpler guard as it was when we entered the callee function.

Exceptions:
(a) if the guard is false it would be correct but inefficient to restore it; keep it false until we find a convergeance with another viable path
(b) if we're doing path-sensitive symex then we do tail duplication, and there are no control-flow     convergeances. Keep the guard as-was.
(c) if we're executing a multi-threaded program then symex_assume_l2 uses the guard to accumulate assumptions, which we must not discard.

In truth this optimisation is applicable whenever a block postdominates another, but function structure gives us a cheap way to establish postdominance without analysis (in the absence of setjmp/longjmp at least)

### DIFF
--- a/regression/cbmc/residual-guards-1/test_execution.desc
+++ b/regression/cbmc/residual-guards-1/test_execution.desc
@@ -1,0 +1,11 @@
+CORE
+test.c
+--unwind 10
+^\[main\.assertion\.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+The main test is in test.desc; this just checks that the test is executable and
+the assertion fails as expected.
--- a/regression/cbmc/residual-guards-2/test_execution.desc
+++ b/regression/cbmc/residual-guards-2/test_execution.desc
@@ -1,0 +1,11 @@
+CORE
+test.c
+--unwind 10
+^\[main\.assertion\.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+The main test is in test.desc; this just checks that the test is executable and
+the assertion fails as expected.
--- a/regression/cbmc/residual-guards-3/test_execution.desc
+++ b/regression/cbmc/residual-guards-3/test_execution.desc
@@ -1,0 +1,11 @@
+CORE
+test.c
+--depth 1000
+^\[main\.assertion\.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+The main test is in test.desc; this just checks that the test is executable and
+the assertion fails as expected.
--- a/regression/cbmc/residual-guards-4/test_execution.desc
+++ b/regression/cbmc/residual-guards-4/test_execution.desc
@@ -1,0 +1,11 @@
+CORE
+test.c
+
+^\[main\.assertion\.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+The main test is in test.desc; this just checks that the test is executable and
+the assertion fails as expected.
--- a/regression/cbmc/residual-guards-1/test.c
+++ b/regression/cbmc/residual-guards-1/test.c
@@ -1,0 +1,19 @@
+void g()
+{
+  g();
+}
+
+void f(int y)
+{
+  if(y == 5)
+  {
+    g();
+    y = 10;
+  }
+}
+
+int main(int argc, char **argv)
+{
+  f(argc);
+  assert(argc == 1);
+}
--- a/regression/cbmc/residual-guards-1/test.desc
+++ b/regression/cbmc/residual-guards-1/test.desc
@@ -1,0 +1,14 @@
+CORE paths-lifo-expected-failure
+test.c
+--show-vcc --unwind 10
+^\{1\} main::argc!0@1#1 = 1$
+^EXIT=0$
+^SIGNAL=0$
+--
+^\{-[0-9]+\} f::y!0@1#[0-9]+ = 10$
+--
+This checks that the assertion, argc == 1, is unguarded (i.e. is not of the form 
+'guard => condition'). Such a guard is redundant, but could be added before goto-symex
+made sure to restore guards to their state at function entry.
+We also check that no VCC is created for the unreachable code 'y = 10;'
+--paths mode is excluded as it currently always accrues large guards as it proceeds through execution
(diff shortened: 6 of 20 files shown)


smowton:
Poster-child example courtesy of @hannes-steffenhagen-diffblue:

```c++
#include <assert.h>
#include <string.h>
#include <stdlib.h>


char *nondet_str(void) {
  const char empty[] = "                               ";
  char *result = strdup(empty);
  int x;
  __CPROVER_assume(x >= 0 && x <= 9);
  result[30] = (char)('0'+x);
  return result;
}

int main(void) {
  long x = 2;
  const char *sx = "                              2";
  for(int i = 0; i < 10; ++i) {
    assert(strtol(nondet_str(), NULL, 10) == x);
  }
  return 0;
}
```

This previously got bogged down in an ever-growing guard, I think because the guard `|=` operator didn't quite manage to squash the guard back down so that it left `strtol` in the same state as it entered, and so the run slowed ever down and down. With this patch that example takes around 5 seconds to complete on my machine.

kroening:
Guards getting fat & monstrous has bugged me for many years -- and I believe it's a major performance problem.
I agree it's probably not sensible to try to solve this exactly, but that cheap, syntactic stuff is the way to go.
Other ideas like "function body"?

kroening:
Another BTW: I think that path-pased symex suffers from this just as much.

smowton:
@kroening I don't follow "function body"?

smowton:
SVCOMP/Java: 3 tests stop timing out (timeout = 30 seconds), overall CPU time down 5%

smowton:
SVCOMP/C subsample: 3 tests stop timing out (timeout = 30 seconds), total CPU time down 2%

regression/cbmc/assume-false/test.desc
```diff
@@ -0,0 +1,11 @@
...
+test.c
+--show-vcc
+main::1::i!0@1#[0-9]+ = 0
+^EXIT=0$
+^SIGNAL=0$
+--
+main::1::i!0@1#[0-9]+ = 2
+--
+This checks that code rendered unreachable by a __CPROVER_assume(0) does not generate constraints.
+
```
peterschrammel: redundant blank line

regression/cbmc/residual-guards-1/test_execution.desc
```diff
@@ -0,0 +1,11 @@
+CORE paths-lifo-expected-failure
+test.c
+--unwind 10
+^\[main.assertion.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
```
peterschrammel:
```suggestion
^\[main\.assertion\.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
```

smowton:
CI failure: this harms performance for cases with many conditional `ASSUME false;` instructions in one function -- previously these would be ignored for guard purposes and the unviable instructions executed regardless, which was wasteful but lead to the guard being simplified more effectively. This is especially punitive with BDD guards.

Note this already happens now when paths are truncated by falsifying the guard -- for example exceeding `depth` bound, or the (specifically recursion) `unwind` bound.

Proposed course of action: merge this PR with the guard reversion but not the path truncation cleanup, then separately pursue guard shrinkage when paths are truncated, then make path truncation uniform so that we don't execute instructions after an `ASSUME false`.

smowton:
In view of this I've removed the cleanup of `ASSUME` vs. guard manipulation from this PR and only make the on-function-return guard simplification. I'm re-running my benchmarks on this modified version and expect CI to pass here.

smowton:
Current status: causes about 10 SVCOMP/Java tests to stop timing out and instead return incorrect answers. Looking into it...

smowton:
Conclusion: reintroduced those parts of the cleanup that ensure an ASSUME / real constraint is generated wherever we truncate execution due to the unwind or depth limits, as otherwise using the dominant guard would actually mean broadening the guard by re-including impossible paths. This PR now (a) ensures we generate a real constraint whenever we truncate a path, and (b) uses guard-as-of-entering-function whenever we exit that function, possibly broadening it by including paths that have been ruled out and/or simplifying it better than the expression-simplifier is capable of.

Forthcoming work will separate the concepts of guards and reachability, such that we can reduce the execution guard at control-flow convergances even when one of the predecessors is unreachable, while maintaining the benefit of not generating constraints for unreachable code.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9aec581b9656b1a2a7fac2a50cf9abc57cd75a65).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122999685

smowton:
Re-benchmarked with neutral results: all I claim now is that specific examples where the guard gets out of hand, like Hannes' example above, become analysable with this change but were previously impractically slow.

regression/cbmc/residual-guards-2/test_execution.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+test.c
+--unwind 10
+^\[main.assertion.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
```
peterschrammel:
```suggestion
^\[main\.assertion\.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
```

regression/cbmc/residual-guards-3/test_execution.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+test.c
+--depth 1000
+^\[main.assertion.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
```
peterschrammel:
```suggestion
^\[main\.assertion\.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
```

regression/cbmc/residual-guards-4/test_execution.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+test.c
+
+^\[main.assertion.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
```
peterschrammel:
```suggestion
^\[main\.assertion\.[0-9]+\] line [0-9]+ assertion argc == 1: FAILURE$
```

peterschrammel (approved):

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 5dbbf291974316a9fec32315e4850afbb57d4904).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123209903

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 66133a6b0f4844d14c123f64739c21aab31b0315).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123215484

PR #5008, opened 2019-08-12
WIP add fake_dynamic_cast

hannes-steffenhagen-diffblue:
This basically lets us do a dynamic_cast-like thing without RTTI

Comes with an example of how to do this for `ai_domaint`, which is one of the
places where we need dynamic cast right now.

### DIFF
--- a/src/analyses/dependence_graph.cpp
+++ b/src/analyses/dependence_graph.cpp
@@ -334,3 +334,13 @@ void dep_graph_domaint::populate_dep_graph(
   for(const auto &d_dep : data_deps)
     dep_graph.add_dep(dep_edget::kindt::DATA, d_dep, this_loc);
 }
+
+const void *dep_graph_domaint::match_type_identifier(
+  const type_identifiert &identifier) const
+{
+  if(identifier == type_identifiert::get<dep_graph_domaint>())
+  {
+    return this;
+  }
+  return nullptr;
+}
--- a/src/util/type_identifier.h
+++ b/src/util/type_identifier.h
@@ -1,0 +1,59 @@
+#pragma once
+
+#include <string>
+#include <type_traits>
+
+class type_identifiert final
+{
+private:
+  static int identifier_counter;
+  int id;
+  explicit type_identifiert(int id) : id(id)
+  {
+  }
+
+public:
+  type_identifiert(const type_identifiert &) = default;
+  type_identifiert(type_identifiert &&) = default;
+  type_identifiert &operator=(const type_identifiert &) = default;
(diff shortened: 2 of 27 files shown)


src/util/type_identifier.h
```diff
@@ -0,0 +1,59 @@
...
+};
+
+struct has_type_indext
+{
+  void *match_type_identifier(const type_identifiert &id)
+  {
+    const has_type_indext *cthis = this;
+    return const_cast<void *>(cthis->match_type_identifier(id));
+  }
+  virtual const void *match_type_identifier(const type_identifiert &) const = 0;
```
hannes-steffenhagen-diffblue, line 47: these methods should probably be private

src/analyses/dependence_graph.cpp
```diff
@@ -334,3 +334,13 @@ void dep_graph_domaint::populate_dep_graph(
   for(const auto &d_dep : data_deps)
     dep_graph.add_dep(dep_edget::kindt::DATA, d_dep, this_loc);
 }
+
+const void *dep_graph_domaint::match_type_identifier(
+  const type_identifiert &identifier) const
+{
+  if(identifier == type_identifiert::get<dep_graph_domaint>())
```
hannes-steffenhagen-diffblue, line 341: These "base" versions could easily be created as something like

```
template<typename T>
match_type_identifier_implt : public has_type_indext
{
   const void *match_type_identifier(const type_identifiert &identifier) const
   {
     if(identifier == type_identifiert::get<T>()) { return this; }
     return nullptr;
   }
};
```

unit/util/type_identifier.cpp
```diff
@@ -0,0 +1,161 @@
...
+  class Bottom : public Middle
+  {
+    const void *
+    match_type_identifier(const type_identifiert &identifier) const override
+    {
+      if(identifier == type_identifiert::get<Bottom>())
+      {
+        return this;
+      }
+      return Middle::match_type_identifier(identifier);
```
hannes-steffenhagen-diffblue, line 96: I'm not sure if there's a simple way to generate the implementations for cases like this; the logic is pretty easy (first check this, then check the parents) but this can't really be done via template (because base class impls aren't visible from there) or via macro (or at least I don't know how to do iteration in macros).

src/util/type_identifier.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <string>
+#include <type_traits>
+
+class type_identifiert final
```
hannes-steffenhagen-diffblue, line 6: Names in this file all subject to change

PR #5013, opened 2019-08-13
Improve error message when functional method is inherited

thk123:
This error can be erroneously triggered both when there are default methods on the interface (meaning >1 methods) or when the method is actually defined in an inherited interface (meaning 0 methods). Extend the warning message to explain both cases.

- [n.a] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).

### DIFF
--- a/jbmc/src/java_bytecode/lambda_synthesis.cpp
+++ b/jbmc/src/java_bytecode/lambda_synthesis.cpp
@@ -121,7 +121,9 @@ static optionalt<irep_idt> interface_method_id(
       << functional_interface_tag.get_identifier()
       << " which should have exactly one abstract method but actually has "
       << implemented_interface_type.methods().size()
-      << ". Note default methods are not supported yet." << messaget::eom;
+      << ". Note default methods are not supported yet."
+      << "Also note methods declared in an inherited interface are not "
+      << "supported either." << messaget::eom;
     return {};
   }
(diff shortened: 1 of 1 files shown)


NathanJPhillips (approved):

smowton (approved):
Oooh, hadn't thought of that

smowton:
Consider adding a `FUTURE` test for this?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 5466c0883d5799f6af65d5fb22e6c4f92de51f02).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/122986140

thk123:
@smowton will do, though I note there is no FUTURE test for the original problem of default methods :stuck_out_tongue:

smowton:
True. My shortcomings probably shouldn't serve as excuse though ;)

thk123:
Not only is there such a test, but there is also a test for exactly this - both in `lambda-unhandled-types`

thk123:
Tested here: https://github.com/diffblue/cbmc/blob/develop/jbmc/regression/jbmc/lambda-unhandled-types/test.desc#L8

(by you!! :stuck_out_tongue: )

smowton:
Note that test specifically looks for a *stub* superinterface, is that the case here?

PR #5016, opened 2019-08-14
Handle more formulas than just equations in string solver

romainbrenguier:
Previously the string-solver was restricting builtin-functions to be the left-hand-side of an equations, so formulas like `g1 => c = char_at(string, 10)` would be forbidden.
This removes this restriction by replacing these function applications by symbols.
The array_to_length and pointer_to_array associations should still only appear on lhs of an equations. Char array equalities should also only appear as an equation.

The changes are tested with unit tests as the current symex is not producing the kind of formulas which were not handled.
The motivation for this feature is that I have seen this kind of equations appear while experimenting with BDD guards https://github.com/diffblue/cbmc/pull/4765
 

### DIFF
--- a/src/solvers/strings/string_dependencies.h
+++ b/src/solvers/strings/string_dependencies.h
@@ -182,22 +182,27 @@ private:
     const std::function<void(const nodet &)> &f) const;
 };
 
-/// When right hand side of equation is a builtin_function add
+/// When a sub-expression of \p expr is a builtin_function, add
 /// a "string_builtin_function" node to the graph and connect it to the strings
 /// on which it depends and which depends on it.
 /// If the string builtin_function is not a supported one, mark all the string
 /// arguments as depending on an unknown builtin_function.
 /// \param dependencies: graph to which we add the node
-/// \param equation: an equation whose right hand side is possibly a call to a
+/// \param expr: an expression which may contain a call to a
 ///   string builtin_function.
 /// \param array_pool: array pool containing arrays corresponding to the string
 ///   exprt arguments of the builtin_function call
-/// \return true if a node was added, if false is returned it either means that
-///   the right hand side is not a function application
+/// \param fresh_symbol: used to create new symbols for the return values of
+///   builtin functions
+/// \return An expression in which function applications have been replaced
+///   by symbols corresponding to the `return_value` field of the associated
+///   builtin function. Or an empty optional when no function applications has
+///   been encountered
 /// \todo there should be a class with just the three functions we require here
-bool add_node(
+optionalt<exprt> add_node(
   string_dependenciest &dependencies,
-  const equal_exprt &equation,
-  array_poolt &array_pool);
+  const exprt &expr,
+  array_poolt &array_pool,
+  symbol_generatort &fresh_symbol);
 
 #endif // CPROVER_SOLVERS_STRINGS_STRING_DEPENDENCIES_H
--- a/src/solvers/strings/string_dependencies.h
+++ b/src/solvers/strings/string_dependencies.h
@@ -79,9 +79,8 @@ public:
   std::unique_ptr<const string_nodet>
   node_at(const array_string_exprt &e) const;
 
-  /// `builtin_function` is reset to an empty pointer after the node is created
   builtin_function_nodet &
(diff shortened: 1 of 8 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 10418f6d5102cc0952470c99c639b7aafdad2cca).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123485329

peterschrammel (approved):

src/solvers/strings/string_dependencies.h
```diff
@@ -188,16 +188,20 @@ class string_dependenciest
 /// If the string builtin_function is not a supported one, mark all the string
 /// arguments as depending on an unknown builtin_function.
 /// \param dependencies: graph to which we add the node
-/// \param equation: an equation whose right hand side is possibly a call to a
```
allredj, line 192: The function description needs updating. It still mentions the rhs.

src/solvers/strings/string_dependencies.h
```diff
@@ -188,16 +188,20 @@ class string_dependenciest
...
 ///   string builtin_function.
 /// \param array_pool: array pool containing arrays corresponding to the string
 ///   exprt arguments of the builtin_function call
-/// \return true if a node was added, if false is returned it either means that
-///   the right hand side is not a function application
+/// \param fresh_symbol: used to create new symbols for the return values of
+///   builtin functions
+/// \return an expression in which function applications have been replaced
+///   by symbols corresponding to the `return_value` field of the associated
+///   builtin function.
```
allredj: :no_entry_sign: Missing: Document that the function returns empty when there is no substitution.

src/solvers/strings/string_refinement.cpp
```diff
@@ -630,7 +632,18 @@ decision_proceduret::resultt string_refinementt::dec_solve()
...
 
   const union_find_replacet string_id_symbol_resolve =
-    string_identifiers_resolution_from_equations(equations, ns, log.debug());
+    string_identifiers_resolution_from_equations(
+      [&] {
+        std::vector<equal_exprt> equalities;
+        for(const auto &eq : equations)
+        {
+          if(auto equal_expr = expr_try_dynamic_cast<equal_exprt>(eq))
+            equalities.push_back(*equal_expr);
```
allredj, line 633: Not sure I understand. Why do equalities that don't appear as top-level `equal_exprt`s not need to be added to the vector of equalities? :racehorse:

src/solvers/strings/string_refinement.cpp
```diff
@@ -283,18 +283,10 @@ void string_refinementt::set_to(const exprt &expr, bool value)
...
   PRECONDITION(expr.type().id() == ID_bool);
   PRECONDITION(equality_propagation);
-
-  if(expr.id() == ID_equal && value)
-  {
-    const equal_exprt &eq_expr = to_equal_expr(expr);
-    equations.push_back(eq_expr);
-  }
+  if(!value)
+    equations.push_back(not_exprt{expr});
```
allredj, line 287: Is that to compensate for :racehorse: ?

allredj (changes requested):
A change requested in the documentation, otherwise looks good.

src/solvers/strings/string_refinement.cpp
romainbrenguier: If they don't appear at the top then we can't be sure they always hold, for instance if we have `g1 => (s1 == s2)`, we cannot be sure `s1` and `s2` will be equal in the end and we have to keep them as separate symbols (instead of replacing occurences of `s1` by `s2`).

src/solvers/strings/string_refinement.cpp
romainbrenguier: not really, it's more like the opposite: because we introduce things that may not be `equal_exprt` in `equations` they need to be filtered out in the previous commit

allredj (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 462d355468fa735c7600134f78b66eb7206d7351).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123848151

PR #5019, opened 2019-08-14
byte-operators unpack_rec: enforce max-bytes if supplied

smowton:
Otherwise it may return more bytes than were requested, leading to an invariant violation in lower_byte_update which passes a constant limit and enforces that it is obeyed.

Also remove the broken-smt-backend tag from tests which now pass.

### DIFF
--- a/src/solvers/lowering/byte_operators.cpp
+++ b/src/solvers/lowering/byte_operators.cpp
@@ -1072,7 +1072,7 @@ exprt lower_byte_extract(const byte_extract_exprt &src, const namespacet &ns)
     src.id() == ID_byte_extract_big_endian);
   const bool little_endian = src.id() == ID_byte_extract_little_endian;
 
-  // determine an upper bound of the number of bytes we might need
+  // determine an upper bound of the last byte we might need
   auto upper_bound_opt = size_of_expr(src.type(), ns);
   if(upper_bound_opt.has_value())
   {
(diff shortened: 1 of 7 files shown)


tautschnig:
@smowton (knowing that this is a somewhat dated PR:) Would you mind rebasing?

smowton:
@tautschnig done

tautschnig:
Thanks @smowton! It seems the unit test is failing an invariant, would you be able to debug this?

smowton:
Took a look today, but it's pretty awful; there's much confusion about counts vs. offsets. Will have another go tomorrow.

smowton:
@tautschnig fixed this -- the big-endian case needed to be done differently, plus there was another case (unpacking from an array) where the `max_bytes` parameter was handled incorrectly. I've also added a second commit improving terminology a bit and noting a case that is surely a latent bug but which I haven't the time to examine here.

smowton:
Had to re-mark tests `union6`, `union7` and `va_list3` as broken-smt because they appear to fail specifically with the SMT backend and specifically on Windows (or perhaps somehow only Windows CI runs the SMT-backend tests, but probably not)

smowton:
@tautschnig rebased, passing tests, ready for review

smowton:
@tautschnig nudge, since I see you're working in this area anyhow (unrelatedly, also nudge on #5418 which I worked on over the holidays)

src/solvers/lowering/byte_operators.cpp
```diff
@@ -1106,6 +1106,7 @@ exprt lower_byte_extract(const byte_extract_exprt &src, const namespacet &ns)
     if(element_bits.has_value() && *element_bits >= 1 && *element_bits % 8 == 0)
     {
       auto num_elements = numeric_cast<std::size_t>(array_type.size());
+      // XXX: This can't be right -- unpacked is a byte array, whereas array_type may have larger elements
```
tautschnig, line 1109: Agreed, this could result in inconsistent expressions being generated. We should really stick with the `else` branch here (which AFAIK was added much later). Can you amend your PR to remove this `if` statement here?

tautschnig (approved):

src/solvers/lowering/byte_operators.cpp
tautschnig: Nevermind, I have found a test exercising this path. I'll merge this one and will create a follow-up PR.

PR #5033, opened 2019-08-16
Support instanceof and checkcast for array types

smowton:
Previously java::array[reference] always carried the matching classid (i.e. "java::array[reference]"). Now it carries java::array[reference(Subtype)] for a single-dimensioned array of type Subtype, or java::array[reference(array[reference(Subtype])]) for a two-dimensional array, and so on. This means that instanceof tests for these types now work as expected, and ClassCastExceptions are thrown when casting between array types (or corresponding ASSERT/ASSUME statements created when not running with Java runtime exceptions enabled). remove_java_new and remove_instanceof passes are improved to deal with this new dimension to the type system, and the array clone method propagates the type of the array it is cloning as expected.

This should improve capacity to determine the real type of an array in a trace, and may help symex to filter value-sets according to instanceof / checkcast instructions.

The remaining missing piece of the array type system is ArrayStoreException, as this requires us to compare two unknown type IDs, cf. all existing instanceof operations which have a constant RHS operand.


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 18d596c88bbee8d74e703991e777b6b98228bf29).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123507222

jbmc/regression/jbmc-strings/max-length-generic-array/IntegerTests.java
```diff
@@ -28,9 +29,9 @@ public static Boolean testMySet(Integer key) {
 
 class MyGenSet<E> {
     E[] array;
-    @SuppressWarnings("unchecked")
     MyGenSet() {
-        array = (E[]) new Object[1];
+      // Note the caller must instantiate the array as generic array creation
```
thk123, line 33: :question: I don't understand why this is changing - was the java here previous illegal?

jbmc/regression/jbmc/instanceof9/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
```
thk123, line 1: :pick: Generally easier if tests are in a separate commit to code changes.

jbmc/src/java_bytecode/java_bytecode_convert_class.cpp
```diff
@@ -881,11 +882,36 @@ void add_java_array_types(symbol_tablet &symbol_table)
       ID_java_new_array, java_reference_type(struct_tag_type), location);
     dereference_exprt old_array{this_symbol.symbol_expr()};
     dereference_exprt new_array{local_symexpr};
+
```
thk123: :pick: misc white space change should be in own commit or not at all

jbmc/src/java_bytecode/java_bytecode_convert_class.cpp
```diff
@@ -881,11 +882,36 @@ void add_java_array_types(symbol_tablet &symbol_table)
...
+      // which should be copied:
+      // Note we can't use `get_class_identifier_field` as at the time this
+      // method is called java.lang.Object hasn't actually been defined yet.
+
+      member_exprt old_base_class(
+        old_array,
+        base_class_component.get_name(),
+        base_class_component.type());
+      member_exprt old_classid(
+        old_base_class, "@class_identifier", string_typet());
```
thk123: :no_entry_sign: We should abstract out this magic string that is used in multiple places to at least a string constant

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -2488,35 +2489,35 @@ code_blockt java_bytecode_convert_methodt::convert_newarray(
...
+    }
     else
-      element_type = '?';
-  }
-  else
-    element_type = 'a';
-
-  const reference_typet ref_type = java_array_type(element_type);
+      return java_reference_array_type(to_struct_tag_type(arg0.type()));
+  }();
```
thk123, line 2521: :green_heart:

thk123: :pick: Though AFAICT this is just refactoring -that would be clearer done in a stand alone commit (and if there is a change here, it is lost in the refactoring and should :no_entry_sign: definitely be pulled out into a stand alone commit)

jbmc/src/java_bytecode/java_types.cpp
```diff
@@ -140,6 +140,25 @@ java_reference_typet java_array_type(const char subtype)
   return java_reference_type(struct_tag_type);
 }
 
+irep_idt java_classid_for_type(const struct_tag_typet &tag)
+{
+  if(tag.get_identifier() == "java::array[reference]")
+  {
+    const auto &element_type =
```
thk123, line 147: :no_entry_sign: Please use `java_array_element_type`
:pick: No need `to_pointer_type` as `typet` declares `subtype`

jbmc/src/java_bytecode/java_types.cpp
```diff
@@ -140,6 +140,25 @@ java_reference_typet java_array_type(const char subtype)
...
+      to_pointer_type(static_cast<const typet &>(tag.find(ID_element_type)))
+        .subtype();
+    if(element_type.id() == ID_struct_tag)
```
thk123: :pick: `can_cast_type<struct_tag_typet>`

jbmc/src/java_bytecode/java_types.cpp
```diff
@@ -140,6 +140,25 @@ java_reference_typet java_array_type(const char subtype)
...
+    {
+      irep_idt subtype_classid =
+        java_classid_for_type(to_struct_tag_type(element_type));
+      return "java::array[reference(" +
+             id2string(strip_java_namespace_prefix(subtype_classid)) + ")]";
```
thk123, line 154: :question: Why `strip_java_namespace_prefix`? Just have to add it back on if we want to look up the symbol in the symbol table?

jbmc/src/java_bytecode/remove_instanceof.cpp
```diff
@@ -56,6 +56,24 @@ class remove_instanceoft
...
 };
 
+static irep_idt add_array_qualifiers(
+  const irep_idt &underlying_type_name,
+  std::size_t array_dimensions)
+{
+  if(array_dimensions == 0)
+    return underlying_type_name;
+
+  // Make "java::array[array[A]]" for example, for "A[][]"
```
thk123: :no_entry_sign: This comment seems wrong, looks like it does:

```
java::array[reference(array[reference(A)])]
```

:bulb: I'd also put this in the doxygen

jbmc/src/java_bytecode/remove_instanceof.cpp
```diff
@@ -98,11 +116,37 @@ bool remove_instanceoft::lower_instanceof(
...
+    // means we don't need to check all array dimensionalities, which also
+    // derive from Object, but finding the highest-dimensioned array in the
+    // program under test is not trivial.
+    expr = true_exprt();
+    return true;
+  }
+
+  // If checking for an array type, the target type must have the same
+  // dimension:
+  std::size_t array_dimensions = 0;
```
thk123: :bulb: Pull out a utiltity for getting dimension of an array type? Maybe there already is one?

jbmc/regression/jbmc/instanceof2/test.desc
```diff
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
```
thk123, line 1: :pick: While your in the area and presumably understand the test - could you add a description to the bottom of it.

thk123 (commented):
:no_entry_sign: ~Please make manual TG bump for this~ -> diffblue/test-gen#3822 (failing CI)
:no_entry_sign: Please add a test with a generic array :heavy_check_mark: 
:no_entry_sign: Please add a test of an inner class array (to check the naming) :heavy_check_mark:

jbmc/regression/jbmc-strings/max-length-generic-array/IntegerTests.java
smowton: Not illegal, but produced code which immediately failed with a ClassCastException, both on the real JVM and in CBMC with this patch applied, so it wasn 't really testing anything

jbmc/src/java_bytecode/java_types.cpp
jeannielynnmoulton: I noticed this as well :)

jbmc/src/java_bytecode/java_bytecode_convert_class.cpp
smowton: Added commit to do this

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
smowton: The change is because `java_reference_typet` doesn't have a nullary constructor, so we can't use the old structure of a default-init'd variable which is then assigned one way or the other

jbmc/src/java_bytecode/java_types.cpp
smowton: I wanted to avoid a classid with multiple `java::`s in it, thought that might confuse some downstream classid parsing code

jbmc/src/java_bytecode/remove_instanceof.cpp
smowton: Factored out into `java_array_dimension_and_element_type` in `java_types.h`. Recommend test-gen's `dim_and_element_type` (which is less general) should go away / use it.

smowton:
@thk123 @jeannielynnmoulton  comments addressed please re-review

smowton:
Ye olde style test-gen bump at https://github.com/diffblue/test-gen/pull/3822

thk123:
:no_entry_sign: Just noting that currently TG CI fails on this PR.

jbmc/src/java_bytecode/java_types.cpp
thk123: @smowton This seems to be outstanding?

thk123: Here: https://github.com/diffblue/cbmc/pull/5033/files#diff-ab3efb660b8cd14f23e5fcf8cb285c86R214

thk123: And here: https://github.com/diffblue/cbmc/pull/5033/files#diff-ab3efb660b8cd14f23e5fcf8cb285c86R148

smowton: Not so (see the Outdated tag)

jbmc/src/java_bytecode/remove_instanceof.cpp
thk123: :angel:  it'd be great if you fix this in your bump!

thk123:
Re-reviewed and resolved comments, outstanding:

 - test-gen bump passing
 - still some direct `ID_element_type` access rather than java_element_type

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
thk123: Sure - but you can do, refactor old code to use immmediately invoked lambda, then add the extra condition?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: cc40a4ad30f103aa3c0b3e49acde90a997986bd0).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/124402330

smowton:
@thk123 fixed bump, used `java_array_element_type` (now I've figured out that's a function not a local variable naming suggestion :)), please re-review

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d0b086310b215fee3b967e98754feaffa2a9a9c5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/124507656

smowton:
Realised that this is broken for casts to `Object[]`, `Object[][]` etc, which need to match all arrays of that dimensionality and higher. That either requires knowing every array type in use in the program, or splitting the classid field into two parts: the element class ID and the array dimensionality. I'll rework this with the latter approach to avoid needing another whole-program analysis.

smowton:
Replaced by #5066

PR #5036, opened 2019-08-19
Refactor the abstract interpretation code

martin-cs:
Hi all,
        This is quite a substantial patch set and has been a long time in the making.  It radically restructures the abstract interpreter with the goal of making it easier to implement various advanced features such as merging VSD, context dependent analysis, loop bounds analysis, handling function pointers during AI, etc. and also deprecating the other two abstract interpretation frameworks.

  I have tried my best to make all of the changes backwards compatible, so the results given by existing AI users should not change at all.  The only things this should break are out of tree classes that inherit from ait<domainT> and use the interfaces in an unusual way.  As such there aren't test cases because there should be no user / externally visible effects.  I will start a separate draft PR that adds new features and functionality using this patch to demonstrate some of the new things we can do.

  I realise this is a lot to review but each patch should be self contained and documented enough that they can be reviewed one at a time.  I'm also happy to discuss / talk through the changes with anyone interested.

### DIFF
--- a/src/analyses/Makefile
+++ b/src/analyses/Makefile
@@ -1,5 +1,6 @@
 SRC = ai.cpp \
       ai_domain.cpp \
+      ai_history.cpp \
       call_graph.cpp \
       call_graph_helpers.cpp \
       constant_propagator.cpp \
--- a/src/analyses/ai.cpp
+++ b/src/analyses/ai.cpp
@@ -157,7 +157,8 @@ xmlt ai_baset::output_xml(
   return function_body;
 }
 
-void ai_baset::entry_state(const goto_functionst &goto_functions)
+ai_baset::trace_ptrt
+ai_baset::entry_state(const goto_functionst &goto_functions)
 {
   // find the 'entry function'
 
@@ -165,14 +166,18 @@ void ai_baset::entry_state(const goto_functionst &goto_functions)
     f_it=goto_functions.function_map.find(goto_functions.entry_point());
 
   if(f_it!=goto_functions.function_map.end())
-    entry_state(f_it->second.body);
+    return entry_state(f_it->second.body);
+
+  // It is not clear if this is even a well-structured goto_functionst object
+  return nullptr;
 }
 
-void ai_baset::entry_state(const goto_programt &goto_program)
+ai_baset::trace_ptrt ai_baset::entry_state(const goto_programt &goto_program)
 {
   // The first instruction of 'goto_program' is the entry point
   trace_ptrt p = history_factory->epoch(goto_program.instructions.begin());
   get_state(p).make_entry();
+  return p;
 }
 
 void ai_baset::initialize(
@@ -218,19 +223,16 @@ ai_baset::trace_ptrt ai_baset::get_next(working_sett &working_set)
 }
 
 bool ai_baset::fixedpoint(
+  trace_ptrt start_trace,
   const irep_idt &function_id,
   const goto_programt &goto_program,
   const goto_functionst &goto_functions,
   const namespacet &ns)
 {
-  working_sett working_set;
+  PRECONDITION(start_trace != nullptr);
 
-  // Put the first location in the working set
-  if(!goto_program.empty())
-  {
-    put_in_working_set(
-      working_set, history_factory->bang(goto_program.instructions.begin()));
-  }
+  working_sett working_set;
+  put_in_working_set(working_set, start_trace);
 
   bool new_data=false;
 
@@ -247,14 +249,15 @@ bool ai_baset::fixedpoint(
 }
 
 void ai_baset::fixedpoint(
+  trace_ptrt start_trace,
   const goto_functionst &goto_functions,
   const namespacet &ns)
 {
-  goto_functionst::function_mapt::const_iterator
-    f_it=goto_functions.function_map.find(goto_functions.entry_point());
+  goto_functionst::function_mapt::const_iterator f_it =
+    goto_functions.function_map.find(goto_functions.entry_point());
 
-  if(f_it!=goto_functions.function_map.end())
-    fixedpoint(f_it->first, f_it->second.body, goto_functions, ns);
+  if(f_it != goto_functions.function_map.end())
+    fixedpoint(start_trace, f_it->first, f_it->second.body, goto_functions, ns);
 }
 
 bool ai_baset::visit(
@@ -469,7 +472,7 @@ bool ai_recursive_interproceduralt::visit_edge_function_call(
   {
     locationt l_begin = callee.instructions.begin();
 
-    working_sett ignore_working_set; // Redundant; fixpoint will add l_begin
+    working_sett catch_working_set; // The trace from the next fixpoint
 
     // Do the edge from the call site to the beginning of the function
     bool new_data = visit_edge(
@@ -478,11 +481,16 @@ bool ai_recursive_interproceduralt::visit_edge_function_call(
       callee_function_id,
       l_begin,
       ns,
-      ignore_working_set);
+      catch_working_set);
 
     // do we need to do/re-do the fixedpoint of the body?
     if(new_data)
-      fixedpoint(callee_function_id, callee, goto_functions, ns);
+      fixedpoint(
+        get_next(catch_working_set),
+        callee_function_id,
+        callee,
+        goto_functions,
+        ns);
   }
 
   // This is the edge from function end to return site.
@@ -493,7 +501,7 @@ bool ai_recursive_interproceduralt::visit_edge_function_call(
       l_end->is_end_function(),
       "The last instruction of a goto_program must be END_FUNCTION");
 
-    // Construct a history from a location
+    // Find the histories for a location
     auto traces = storage->abstract_traces_before(l_end);
 
     bool new_data = false;
--- a/src/analyses/ai.cpp
+++ b/src/analyses/ai.cpp
@@ -248,30 +248,18 @@ bool ai_baset::visit(
   bool new_data=false;
 
   // Function call and end are special cases
-  if(l->is_function_call() && !goto_functions.function_map.empty())
+  if(l->is_function_call())
   {
     DATA_INVARIANT(
       goto_program.get_successors(l).size() == 1,
       "function calls only have one successor");
+
     DATA_INVARIANT(
       *(goto_program.get_successors(l).begin()) == std::next(l),
       "function call successor / return location must be the next instruction");
 
-    const code_function_callt &code = to_code_function_call(l->code);
-
-    locationt to_l = std::next(l);
-    if(do_function_call_rec(
-         function_id,
-         l,
-         to_l,
-         code.function(),
-         code.arguments(),
-         goto_functions,
-         ns))
-    {
-      new_data = true;
-      put_in_working_set(working_set, to_l);
-    }
+    new_data = visit_function_call(
+      function_id, l, working_set, goto_program, goto_functions, ns);
   }
   else if(l->is_end_function())
   {
@@ -279,7 +267,8 @@ bool ai_baset::visit(
       goto_program.get_successors(l).empty(),
       "The end function instruction should have no successors.");
 
-    // Do nothing
+    new_data = visit_end_function(
+      function_id, l, working_set, goto_program, goto_functions, ns);
   }
   else
   {
@@ -327,108 +316,110 @@ bool ai_baset::visit_edge(
   return false;
 }
 
-bool ai_baset::do_function_call(
+bool ai_baset::visit_edge_function_call(
   const irep_idt &calling_function_id,
   locationt l_call,
   locationt l_return,
+  const irep_idt &,
+  working_sett &working_set,
+  const goto_programt &,
+  const goto_functionst &,
+  const namespacet &ns)
+{
+  // The default implementation is not interprocedural
+  // so the effects of the call are approximated but nothing else
+  return visit_edge(
+    calling_function_id,
+    l_call,
+    calling_function_id,
+    l_return,
+    ns,
+    working_set);
+}
+
+bool ai_baset::visit_function_call(
+  const irep_idt &calling_function_id,
+  locationt l_call,
+  working_sett &working_set,
+  const goto_programt &caller,
   const goto_functionst &goto_functions,
-  const goto_functionst::function_mapt::const_iterator f_it,
-  const exprt::operandst &,
   const namespacet &ns)
 {
   PRECONDITION(l_call->is_function_call());
 
-  const goto_functionst::goto_functiont &goto_function=
-    f_it->second;
+  locationt l_return = std::next(l_call);
 
-  // This branch is for legacy support only and is deprecated
-
-  if(!goto_function.body_available())
+  // operator() allows analysis of a single goto_program independently
+  // it generates a synthetic goto_functions object for this
+  if(!goto_functions.function_map.empty())
   {
-    working_sett working_set; // Redundant; visit will add l_return
+    const code_function_callt &code = to_code_function_call(l_call->code);
+    const exprt &callee_expression = code.function();
 
-    // If we don't have a body, we just do an edge call -> return
-    return visit_edge(
-      calling_function_id,
-      l_call,
-      calling_function_id,
-      l_return,
-      ns,
-      working_set);
+    if(callee_expression.id() == ID_symbol)
+    {
+      const irep_idt &callee_function_id =
+        to_symbol_expr(callee_expression).get_identifier();
+
+      goto_functionst::function_mapt::const_iterator it =
+        goto_functions.function_map.find(callee_function_id);
+
+      DATA_INVARIANT(
+        it != goto_functions.function_map.end(),
+        "Function " + id2string(callee_function_id) + "not in function map");
+
+      const goto_functionst::goto_functiont &callee_fun = it->second;
+
+      if(callee_fun.body_available())
+      {
+        return visit_edge_function_call(
+          calling_function_id,
+          l_call,
+          l_return,
+          callee_function_id,
+          working_set,
+          callee_fun.body,
+          goto_functions,
+          ns);
+      }
+      else
+      {
+        // Fall through to the default, body not available, case
+      }
+    }
+    else
+    {
+      // Function pointers are not currently supported and must be removed
+      DATA_INVARIANT(
+        callee_expression.id() == ID_symbol,
+        "Function pointers and indirect calls must be removed before "
+        "analysis.");
+    }
   }
 
-  DATA_INVARIANT(
-    !goto_function.body.instructions.empty(),
-    "By analysis time, all functions should have bodies");
-
-  // This is the edge from call site to function head.
-
-  {
-    locationt l_begin=goto_function.body.instructions.begin();
-
-    working_sett working_set; // Redundant; fixpoint will add l_begin
-
-    // Do the edge from the call site to the beginning of the function
-    bool new_data = visit_edge(
-      calling_function_id, l_call, f_it->first, l_begin, ns, working_set);
-
-    // do we need to do/re-do the fixedpoint of the body?
-    if(new_data)
-      fixedpoint(f_it->first, goto_function.body, goto_functions, ns);
-  }
-
-  // This is the edge from function end to return site.
-
-  {
-    // get location at end of the procedure we have called
-    locationt l_end=--goto_function.body.instructions.end();
-    assert(l_end->is_end_function());
-
-    // do edge from end of function to instruction after call
-    const statet &end_state=get_state(l_end);
-
-    if(end_state.is_bottom())
-      return false; // function exit point not reachable
-
-    working_sett working_set; // Redundant; visit will add l_return
-
-    return visit_edge(
-      f_it->first, l_end, calling_function_id, l_return, ns, working_set);
-  }
+  // If the body is not available then we just do the edge from call to return
+  // in the caller.  Domains should over-approximate what the function might do.
+  return visit_edge(
+    calling_function_id,
+    l_call,
+    calling_function_id,
+    l_return,
+    ns,
+    working_set);
 }
 
-bool ai_baset::do_function_call_rec(
-  const irep_idt &calling_function_id,
-  locationt l_call,
-  locationt l_return,
-  const exprt &function,
-  const exprt::operandst &arguments,
+bool ai_baset::visit_end_function(
+  const irep_idt &function_id,
+  locationt l,
+  working_sett &working_set,
+  const goto_programt &goto_program,
   const goto_functionst &goto_functions,
   const namespacet &ns)
 {
-  PRECONDITION(!goto_functions.function_map.empty());
+  PRECONDITION(l->is_end_function());
 
-  // This is quite a strong assumption on the well-formedness of the program.
-  // It means function pointers must be removed before use.
-  DATA_INVARIANT(
-    function.id() == ID_symbol,
-    "Function pointers and indirect calls must be removed before analysis.");
-
-  bool new_data=false;
-
-  const irep_idt &identifier = to_symbol_expr(function).get_identifier();
-
-  goto_functionst::function_mapt::const_iterator it =
-    goto_functions.function_map.find(identifier);
-
-  DATA_INVARIANT(
-    it != goto_functions.function_map.end(),
-    "Function " + id2string(identifier) + "not in function map");
-
-  new_data = do_function_call(
-    calling_function_id, l_call, l_return, goto_functions, it, arguments, ns);
-
-  return new_data;
+  // Do nothing
+  return false;
 }
 
 void ai_baset::fixedpoint(
@@ -442,3 +433,57 @@ void ai_baset::fixedpoint(
     fixedpoint(f_it->first, f_it->second.body, goto_functions, ns);
 }
 
+bool ai_recursive_interproceduralt::visit_edge_function_call(
+  const irep_idt &calling_function_id,
+  locationt l_call,
+  locationt l_return,
+  const irep_idt &callee_function_id,
+  working_sett &working_set,
+  const goto_programt &callee,
+  const goto_functionst &goto_functions,
+  const namespacet &ns)
+{
+  // This is the edge from call site to function head.
+  {
+    locationt l_begin = callee.instructions.begin();
+
+    working_sett ignore_working_set; // Redundant; fixpoint will add l_begin
+
+    // Do the edge from the call site to the beginning of the function
+    bool new_data = visit_edge(
+      calling_function_id,
+      l_call,
+      callee_function_id,
+      l_begin,
+      ns,
+      ignore_working_set);
+
+    // do we need to do/re-do the fixedpoint of the body?
+    if(new_data)
+      fixedpoint(callee_function_id, callee, goto_functions, ns);
+  }
+
+  // This is the edge from function end to return site.
+
+  {
+    // get location at end of the procedure we have called
+    locationt l_end = --callee.instructions.end();
+    DATA_INVARIANT(
+      l_end->is_end_function(),
+      "The last instruction of a goto_program must be END_FUNCTION");
+
+    // do edge from end of function to instruction after call
+    const statet &end_state = get_state(l_end);
+
+    if(end_state.is_bottom())
+      return false; // function exit point not reachable
+
+    return visit_edge(
+      callee_function_id,
+      l_end,
+      calling_function_id,
+      l_return,
+      ns,
+      working_set);
+  }
+}
--- a/src/analyses/ai.cpp
+++ b/src/analyses/ai.cpp
@@ -246,6 +246,17 @@ bool ai_baset::fixedpoint(
   return new_data;
 }
 
+void ai_baset::fixedpoint(
+  const goto_functionst &goto_functions,
+  const namespacet &ns)
+{
+  goto_functionst::function_mapt::const_iterator
+    f_it=goto_functions.function_map.find(goto_functions.entry_point());
+
+  if(f_it!=goto_functions.function_map.end())
+    fixedpoint(f_it->first, f_it->second.body, goto_functions, ns);
+}
+
 bool ai_baset::visit(
   const irep_idt &function_id,
   trace_ptrt p,
@@ -442,17 +453,6 @@ bool ai_baset::visit_end_function(
 
   // Do nothing
   return false;
-}
-
-void ai_baset::fixedpoint(
-  const goto_functionst &goto_functions,
-  const namespacet &ns)
-{
-  goto_functionst::function_mapt::const_iterator
-    f_it=goto_functions.function_map.find(goto_functions.entry_point());
-
-  if(f_it!=goto_functions.function_map.end())
-    fixedpoint(f_it->first, f_it->second.body, goto_functions, ns);
 }
 
 bool ai_recursive_interproceduralt::visit_edge_function_call(
--- a/src/analyses/ai.cpp
+++ b/src/analyses/ai.cpp
@@ -14,6 +14,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <cassert>
 #include <memory>
 #include <sstream>
+#include <type_traits>
 
 #include <util/invariant.h>
 #include <util/std_code.h>
@@ -170,7 +171,8 @@ void ai_baset::entry_state(const goto_functionst &goto_functions)
 void ai_baset::entry_state(const goto_programt &goto_program)
 {
   // The first instruction of 'goto_program' is the entry point
-  get_state(goto_program.instructions.begin()).make_entry();
+  trace_ptrt p = history_factory->epoch(goto_program.instructions.begin());
+  get_state(p).make_entry();
 }
 
 void ai_baset::initialize(
@@ -197,16 +199,22 @@ void ai_baset::finalize()
   // Nothing to do per default
 }
 
-ai_baset::locationt ai_baset::get_next(
-  working_sett &working_set)
+ai_baset::trace_ptrt ai_baset::get_next(working_sett &working_set)
 {
   PRECONDITION(!working_set.empty());
 
-  working_sett::iterator i=working_set.begin();
-  locationt l=i->second;
-  working_set.erase(i);
+  static_assert(
+    std::is_same<
+      working_sett,
+      std::set<trace_ptrt, ai_history_baset::compare_historyt>>::value,
+    "begin must return the minimal entry");
+  auto first = working_set.begin();
 
-  return l;
+  trace_ptrt t = *first;
+
+  working_set.erase(first);
+
+  return t;
 }
 
 bool ai_baset::fixedpoint(
@@ -219,18 +227,19 @@ bool ai_baset::fixedpoint(
 
   // Put the first location in the working set
   if(!goto_program.empty())
+  {
     put_in_working_set(
-      working_set,
-      goto_program.instructions.begin());
+      working_set, history_factory->bang(goto_program.instructions.begin()));
+  }
 
   bool new_data=false;
 
   while(!working_set.empty())
   {
-    locationt l=get_next(working_set);
+    trace_ptrt p = get_next(working_set);
 
     // goto_program is really only needed for iterator manipulation
-    if(visit(function_id, l, working_set, goto_program, goto_functions, ns))
+    if(visit(function_id, p, working_set, goto_program, goto_functions, ns))
       new_data=true;
   }
 
@@ -239,13 +248,14 @@ bool ai_baset::fixedpoint(
 
 bool ai_baset::visit(
   const irep_idt &function_id,
-  locationt l,
+  trace_ptrt p,
   working_sett &working_set,
   const goto_programt &goto_program,
   const goto_functionst &goto_functions,
   const namespacet &ns)
 {
   bool new_data=false;
+  locationt l = p->current_location();
 
   // Function call and end are special cases
   if(l->is_function_call())
@@ -259,7 +269,7 @@ bool ai_baset::visit(
       "function call successor / return location must be the next instruction");
 
     new_data = visit_function_call(
-      function_id, l, working_set, goto_program, goto_functions, ns);
+      function_id, p, working_set, goto_program, goto_functions, ns);
   }
   else if(l->is_end_function())
   {
@@ -268,7 +278,7 @@ bool ai_baset::visit(
       "The end function instruction should have no successors.");
 
     new_data = visit_end_function(
-      function_id, l, working_set, goto_program, goto_functions, ns);
+      function_id, p, working_set, goto_program, goto_functions, ns);
   }
   else
   {
@@ -280,7 +290,7 @@ bool ai_baset::visit(
         continue;
 
       new_data |=
-        visit_edge(function_id, l, function_id, to_l, ns, working_set);
+        visit_edge(function_id, p, function_id, to_l, ns, working_set);
     }
   }
 
@@ -289,27 +299,36 @@ bool ai_baset::visit(
 
 bool ai_baset::visit_edge(
   const irep_idt &function_id,
-  locationt l,
+  trace_ptrt p,
   const irep_idt &to_function_id,
-  const locationt &to_l,
+  locationt to_l,
   const namespacet &ns,
   working_sett &working_set)
 {
+  // Has history taught us not to step here...
+  auto next = p->step(to_l, *(storage->abstract_traces_before(to_l)));
+  if(next.first == ai_history_baset::step_statust::BLOCKED)
+    return false;
+  trace_ptrt to_p = next.second;
+
   // Abstract domains are mutable so we must copy before we transform
-  statet &current = get_state(l);
+  statet &current = get_state(p);
 
   std::unique_ptr<statet> tmp_state(make_temporary_state(current));
   statet &new_values = *tmp_state;
 
   // Apply transformer
-  new_values.transform(function_id, l, to_function_id, to_l, *this, ns);
+  new_values.transform(function_id, p, to_function_id, to_p, *this, ns);
 
-  // Initialize state(s), if necessary
-  get_state(to_l);
-
-  if(merge(new_values, l, to_l))
+  // Expanding a domain means that it has to be analysed again
+  // Likewise if the history insists that it is a new trace
+  // (assuming it is actually reachable).
+  if(
+    merge(new_values, p, to_p) ||
+    (next.first == ai_history_baset::step_statust::NEW &&
+     !new_values.is_bottom()))
   {
-    put_in_working_set(working_set, to_l);
+    put_in_working_set(working_set, to_p);
     return true;
   }
 
@@ -318,7 +337,7 @@ bool ai_baset::visit_edge(
 
 bool ai_baset::visit_edge_function_call(
   const irep_idt &calling_function_id,
-  locationt l_call,
+  trace_ptrt p_call,
   locationt l_return,
   const irep_idt &,
   working_sett &working_set,
@@ -330,7 +349,7 @@ bool ai_baset::visit_edge_function_call(
   // so the effects of the call are approximated but nothing else
   return visit_edge(
     calling_function_id,
-    l_call,
+    p_call,
     calling_function_id,
     l_return,
     ns,
@@ -339,12 +358,14 @@ bool ai_baset::visit_edge_function_call(
 
 bool ai_baset::visit_function_call(
   const irep_idt &calling_function_id,
-  locationt l_call,
+  trace_ptrt p_call,
   working_sett &working_set,
   const goto_programt &caller,
   const goto_functionst &goto_functions,
   const namespacet &ns)
 {
+  locationt l_call = p_call->current_location();
+
   PRECONDITION(l_call->is_function_call());
 
   locationt l_return = std::next(l_call);
@@ -374,7 +395,7 @@ bool ai_baset::visit_function_call(
       {
         return visit_edge_function_call(
           calling_function_id,
-          l_call,
+          p_call,
           l_return,
           callee_function_id,
           working_set,
@@ -401,7 +422,7 @@ bool ai_baset::visit_function_call(
   // in the caller.  Domains should over-approximate what the function might do.
   return visit_edge(
     calling_function_id,
-    l_call,
+    p_call,
     calling_function_id,
     l_return,
     ns,
@@ -410,12 +431,13 @@ bool ai_baset::visit_function_call(
 
 bool ai_baset::visit_end_function(
   const irep_idt &function_id,
-  locationt l,
+  trace_ptrt p,
   working_sett &working_set,
   const goto_programt &goto_program,
   const goto_functionst &goto_functions,
   const namespacet &ns)
 {
+  locationt l = p->current_location();
   PRECONDITION(l->is_end_function());
 
   // Do nothing
@@ -435,7 +457,7 @@ void ai_baset::fixedpoint(
 
 bool ai_recursive_interproceduralt::visit_edge_function_call(
   const irep_idt &calling_function_id,
-  locationt l_call,
+  trace_ptrt p_call,
   locationt l_return,
   const irep_idt &callee_function_id,
   working_sett &working_set,
@@ -452,7 +474,7 @@ bool ai_recursive_interproceduralt::visit_edge_function_call(
     // Do the edge from the call site to the beginning of the function
     bool new_data = visit_edge(
       calling_function_id,
-      l_call,
+      p_call,
       callee_function_id,
       l_begin,
       ns,
@@ -464,7 +486,6 @@ bool ai_recursive_interproceduralt::visit_edge_function_call(
   }
 
   // This is the edge from function end to return site.
-
   {
     // get location at end of the procedure we have called
     locationt l_end = --callee.instructions.end();
@@ -472,18 +493,35 @@ bool ai_recursive_interproceduralt::visit_edge_function_call(
       l_end->is_end_function(),
       "The last instruction of a goto_program must be END_FUNCTION");
 
-    // do edge from end of function to instruction after call
-    const statet &end_state = get_state(l_end);
+    // Construct a history from a location
+    auto traces = storage->abstract_traces_before(l_end);
 
-    if(end_state.is_bottom())
-      return false; // function exit point not reachable
+    bool new_data = false;
 
-    return visit_edge(
-      callee_function_id,
-      l_end,
-      calling_function_id,
-      l_return,
-      ns,
-      working_set);
+    // The history used may mean there are multiple histories at the end of the
+    // function.  Some of these can be progressed (for example, if the history
+    // tracks paths within functions), some can't be (for example, not all
+    // calling contexts will be appropriate).  We rely on the history's step,
+    // called from visit_edge to prune as applicable.
+    for(auto p_end : *traces)
+    {
+      // do edge from end of function to instruction after call
+      const statet &end_state = get_state(p_end);
+
+      if(!end_state.is_bottom())
+      {
+        // function exit point reachable in that history
+
+        new_data |= visit_edge(
+          callee_function_id,
+          p_end,
+          calling_function_id,
+          l_return,
+          ns,
+          working_set);
+      }
+    }
+
+    return new_data;
   }
 }
--- a/src/analyses/ai.cpp
+++ b/src/analyses/ai.cpp
@@ -19,8 +19,6 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/std_code.h>
 #include <util/std_expr.h>
 
-#include "is_threaded.h"
-
 void ai_baset::output(
   const namespacet &ns,
   const goto_functionst &goto_functions,
@@ -435,7 +433,7 @@ bool ai_baset::do_function_call_rec(
   return new_data;
 }
 
-void ai_baset::sequential_fixedpoint(
+void ai_baset::fixedpoint(
   const goto_functionst &goto_functions,
   const namespacet &ns)
 {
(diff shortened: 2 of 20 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8017845487d0410bb1c23b157ce560cf1f899769).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/123863950

src/analyses/ai.h
```diff
@@ -543,24 +513,107 @@ class concurrency_aware_ait:public ait<domainT>
   {
   }
 
-  bool merge_shared(
+  virtual bool merge_shared(
```
hannes-steffenhagen-diffblue, line 651: Why is this virtual? Is this class supposed to be inherited from?

src/analyses/ai.h
```diff
@@ -100,7 +100,8 @@ class ai_baset
   /// \return The abstract state before `l`. We return a pointer to a copy as
   ///   the method should be const and there are some non-trivial cases
   ///   including merging abstract states, etc.
-  virtual std::unique_ptr<statet> abstract_state_before(locationt l) const = 0;
+  virtual std::shared_ptr<const statet>
```
hannes-steffenhagen-diffblue: This one is a bit hard to follow; While the changes themselves are obvious enough it's not easy to understand what's meant by 'refactoring the domain storage' here (maybe that'll make more sense later on in the PR?)

src/analyses/ai.h
```diff
@@ -395,16 +395,6 @@ class ait:public ai_baset
 
   typedef goto_programt::const_targett locationt;
 
-  /// Find the analysis result for a given location.
-  domainT &operator[](locationt l)
```
hannes-steffenhagen-diffblue, line 420: Typo in 46c10cebf8d9589bfa75c787b01ea6faf9db274a: "Clarrifies -> Clarifies"

src/analyses/ai.h
```diff
@@ -431,16 +431,25 @@ class ait:public ai_baset
...
 
   /// Look up the analysis state for a given location, instantiating a new state
   /// if required. Used internally by the analysis.
   virtual statet &get_state(locationt l) override
   {
-    return state_map[l]; // calls default constructor
+    typename state_mapt::const_iterator it = state_map.find(l);
+    if(it == state_map.end())
+    {
+      auto p = state_map.insert(std::make_pair(l, std::make_shared<domainT>()));
```
hannes-steffenhagen-diffblue: You could also use `emplace` to skip the `make_pair`

src/analyses/dependence_graph.h
```diff
@@ -257,7 +257,8 @@ class dependence_grapht:
   {
     for(const auto &location_state : state_map)
     {
-      location_state.second.populate_dep_graph(*this, location_state.first);
+      static_cast<dep_graph_domaint &>(*(location_state.second))
```
hannes-steffenhagen-diffblue: Could also use `std::static_pointer_cast`, then you won't need the extra dereference

src/analyses/ai.h
```diff
@@ -431,16 +431,25 @@ class ait:public ai_baset
...
+      CHECK_RETURN(p.second);
+      it = p.first;
+    }
+
+    return *(it->second);
```
hannes-steffenhagen-diffblue: It's a bit unfortunate that we have to assume the pointer is non-null here; C++ GSL has a `not_null` wrapper around pointers that makes such things more explicit. Maybe we should add it to our `util` library as well?

src/analyses/ai_storage.h
```diff
@@ -0,0 +1,125 @@
...
+  friend invariant_propagationt;
+  friend dependence_grapht;
+  state_mapt &internal(void)
+  {
+    return state_map;
+  }
+
+public:
+  std::shared_ptr<const statet> abstract_state_before(
+    locationt t,
```
xbauch: Could the variable be named `l`, for consistency?

src/analyses/ai.cpp
```diff
@@ -248,38 +248,28 @@ bool ai_baset::visit(
...
+  if(l->is_function_call())
   {
     DATA_INVARIANT(
       goto_program.get_successors(l).size() == 1,
       "function calls only have one successor");
+
+    locationt to_l = std::next(l);
     DATA_INVARIANT(
-      *(goto_program.get_successors(l).begin()) == std::next(l),
+      *(goto_program.get_successors(l).begin()) == to_l,
```
xbauch: Inline maybe? Seems we never use `to_l` again.

src/analyses/ai.cpp
```diff
@@ -327,118 +317,174 @@ bool ai_baset::visit_edge(
...
+          ns);
+      }
+      else
+      {
+        // Fall through to the default, body not available, case
+      }
+    }
+    else
+    {
+      // Function pointers are not currently supported and must be removed
```
xbauch, line 427: This is a strange structure of code: do you plan on implementing this else-branch later?

martin-cs: Yes.  I wanted to keep this patch set functionality neutral but with the work you and your team have done, this is *very* much on the cards.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 2942e99b83024105b0fac33b507dca1d94967aef).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/124408878

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+
+#include <memory>
+
+#include <goto-programs/goto_model.h>
+
+#include <util/json.h>
+#include <util/xml.h>
+
+/// A history object is an abstraction / representation of the control
+/// control-flow part of a set of traces.  The simplest example is a single
```
xbauch: `control control-flow` .. just checking if this is intentional.

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+  /// The custom comparison allows particular histories to control
+  /// the order of exploration.
+  struct compare_historyt
+  {
+    bool operator()(const trace_ptrt &l, const trace_ptrt &r) const
+    {
+      return *l < *r;
+    }
+  };
+  typedef std::set<trace_ptrt, compare_historyt> trace_sett;
```
xbauch: I guess some people would prefer the `unordered` variant. With `shared_ptr` you'd get `std::hash` support for free.

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+  ///  2. Merge with an existing history
+  ///     This will mean the analyser will continue of the domain is modified
+  ///     POSTCONDITION(IMPLIES(result.first == MERGED,
+  ///                           result.second is an element of others));
+  ///  3. Block this flow of execution
+  ///     This indicates the transition is impossible (returning to a location
+  ///     other than the call location) or undesireable (omitting some traces)
+  ///     POSTCONDITION(IMPLIES(result.first == BLOCKED,
+  ///                           result.second == nullptr()));
+  typedef std::pair<step_status, trace_ptrt> step_returnt;
```
xbauch: Could this `typedef` precede the above comment?

src/analyses/ai_history.h
martin-cs: I would prefer to keep it ordered.  Histories have to have an ordering for working out which to propagate first in the abstract interpreter (see `working_sett`) and I would rather keep to just one kind of set.

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+  ///              current_location()->is_function_call() ||
+  ///              current_location()->is_end_function());
+  ///
+  /// Step may do one of three things :
+  ///  1. Create a new history object and return a pointer to it
+  ///     This will force the analyser to continue regardless of domain changes
+  ///     POSTCONDITION(IMPLIES(result.first == NEW,
+  ///                           result.second.use_count() == 1 ));
+  ///  2. Merge with an existing history
+  ///     This will mean the analyser will continue of the domain is modified
```
xbauch: Could you please rephrase this line?

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+  /// POSTCONDITION(return value is dereferenceable)
+  virtual const locationt &current_location(void) const = 0;
+
+  /// Domains with a substantial height may need to widen when merging
+  /// this method allows the history to provide a hint on when to do this
+  /// The suggested use of this is for domains merge operation to check this
+  /// and then widen in a domain specific way.  However it could be used in
+  /// other ways (such as the transformer).  The key idea is that the history
+  /// tracks / should know when to widen and when to continue.
+  virtual bool widen(const ai_history_baset &other) const
```
xbauch: If this is just a predicate, then maybe it could be called something like `should_widen`. Just `widen` sounds like an imperative.

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+  virtual void output(std::ostream &out) const
+  {
+  }
+  virtual jsont output_json(void) const;
+  virtual xmlt output_xml(void) const;
+};
+
+/// The common case of history is to only care about where you are now,
+/// not how you got there!
+/// Invariants are not checkable due to C++...
```
xbauch, line 143: But you could make it a precondition to constructors.

martin-cs:
@xbauch : I realise I suggested reviewing patch by patch but, I think one or two of the things you have flagged are fixed / replaced / obsoleted by later patches.  If this is the case, do you want to me to fix the original patches or just comment that it is fixed later?

xbauch:
@martin-cs Comment will do, thanks.

src/analyses/ai_storage.h
```diff
@@ -103,7 +178,18 @@ class location_sensitive_storaget : public ai_storage_baset
...
-  statet &get_state(locationt l, const ai_domain_factory_baset &fac) override
+  statet &get_state(trace_ptrt p, const ai_domain_factory_baset &fac) override
+  {
+    register_trace(p);
+    return get_state(p->current_location(), fac);
+  }
+
+  // For backwards compatability
+  // Care should be exercised in using this.  It is possible to create domains
+  // without the any corresponding history.  This can lead to somewhat
```
xbauch: -> "without the corresponding history."

src/analyses/ai.h
```diff
@@ -96,6 +104,29 @@ class ai_baset
...
+  /// the end of the instruction.
+  /// PRECONDITION(l is dereferenceable)
+  /// \param l: The location before which we want the set of histories
+  /// \return The set of histories before `l`.
+  virtual ctrace_set_ptrt abstract_traces_after(locationt l) const
+  {
+    /// PRECONDITION(l is dereferenceable && std::next(l) is dereferenceable)
+    /// Check relies on a DATA_INVARIANT of goto_programs
+    INVARIANT(!l->is_end_function(), "No state after the last instruction");
+    return storage->abstract_traces_before(l);
```
xbauch: The precondition suggests we should pass `std::next(l)` here. Just checking this is intentional.

src/analyses/ai.h
```diff
@@ -28,10 +61,60 @@ Author: Daniel Kroening, kroening@kroening.com
...
+///    \ref ai_baset#abstract_state_before(locationt)const
+/// 3. Outputting the results of the analysis; see
+///    \ref ai_baset#output(const namespacet&, const irep_idt&,
+///    const goto_programt&, std::ostream&)const et cetera.
+///
+/// Where possible, uses should be agnostic of the particular configuration of
+/// the abstract interpreter.  The "tasks" that goto-analyze uses are good
+/// examples of how to do this.
+///
+/// From a development point of view, there are several directions in which
```
xbauch, line 88: I wonder if parts of this documentation wouldn't be better suited for a README file.

xbauch (approved):
Only small nitpicks. Could be split into several PRs.

src/analyses/ai_storage.h
martin-cs: Changed (but at a later patch so it might not show up here).

src/analyses/ai.cpp
martin-cs: Inlined.

src/analyses/ai_history.h
martin-cs: Nope; editing error.  Fixed.

src/analyses/ai_history.h
martin-cs: Re-ordered.

src/analyses/ai_history.h
martin-cs: "This means the analyser will only continue if the domain is updated"

src/analyses/ai_history.h
martin-cs: OK; `should_widen` it is.

src/analyses/ai_history.h
martin-cs: Not easily.  Non-debugging implementations of iterators don't seem to have a way of checking "is this dereferenceable".  I could pass the `goto_programt` or at least the end marker but that is kind of ugly from an interface point of view and doesn't fully check the "is dereferenceable" part as it won't catch invalidated iterators, etc.  So I think I would rather leave it as a comment and let the CI's iterator debugging builds catch if anyone gets it wrong.

src/analyses/ai_storage.h
martin-cs: "It is possible to create domains without any corresponding history object(s)"

src/analyses/ai.h
martin-cs: Nope; you are correct and that is a bug.  Fixed.  Documentation and preconditions, even not enforced ones, are useful!

src/analyses/ai.h
martin-cs: Maybe.  At some point I want to write a paper describing how this works in detail, the interfaces, etc.  The goal of this commit was to keep the documentation up to date and keep the "if you want to do X then you should ..." comments.  Before that we need to have it merged and working!

martin-cs:
Thanks @xbauch, I have pushed an extra commit that fixes the issues you raised.  Let me know if there is more that you want.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d5ba77b00ae0933fce4ef630cf3adb9b9cadc2a6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/124808826

martin-cs:
Apologies @xbauch , I've had to rebase and alter things because #5064 was merged so I think a number of your comments have been marked as outdated.  I folded the changes you requested in wherever possible.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 304180ed4e527b4a88d5cd8f0d8a59fbb07a3230).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/124954646

src/analyses/ai_domain.h
```diff
@@ -122,4 +184,74 @@ class ai_domain_baset
...
+public:
+  typedef ai_domain_baset statet;
+  typedef ai_domain_baset::locationt locationt;
+  typedef ai_domain_baset::trace_ptrt trace_ptrt;
+
+  virtual ~ai_domain_factory_baset()
+  {
+  }
+
+  virtual std::unique_ptr<statet> make_domain(locationt l) const = 0;
```
hannes-steffenhagen-diffblue: Why are these `unique_ptr` if they're supposed to be used as `shared_ptr`?

src/analyses/dependence_graph.h
```diff
@@ -242,6 +242,10 @@ class dependence_grapht:
   {
     ait<dep_graph_domaint>::initialize(function, goto_program);
 
+    // The dependency graph requires that all nodes are explicitly created
+    forall_goto_program_instructions(i_it, goto_program)
+      get_state(i_it).make_bottom();
```
hannes-steffenhagen-diffblue, line 238: :pick: I know you're just moving this, but I think this is just equivalent to doing a regular for each loop over `goto_program.instructions`, might as well change that while we're at it.

src/analyses/ai.h
```diff
@@ -408,17 +409,18 @@ class ait:public ai_baset
...
     if(it == state_map.end())
     {
-      std::unique_ptr<statet> d = util_make_unique<domainT>();
+      auto d = std::make_shared<domainT>();
       CHECK_RETURN(d->is_bottom());
       return d;
     }
 
-    return util_make_unique<domainT>(it->second);
+    return std::make_shared<domainT>(domainT(it->second));
```
hannes-steffenhagen-diffblue: If everything is `shared_ptr` why not in the `state_map` itself?

src/analyses/dependence_graph.h
```diff
@@ -278,17 +279,21 @@ class dependence_grapht:
...
   virtual statet &get_state(goto_programt::const_targett l)
   {
-    std::pair<state_mapt::iterator, bool> entry=
-      state_map.insert(std::make_pair(l, dep_graph_domaint()));
-
-    if(entry.second)
+    typename state_mapt::const_iterator it = state_map.find(l);
+    if(it == state_map.end())
     {
+      auto d = std::make_shared<dep_graph_domaint>();
```
hannes-steffenhagen-diffblue: I'd have a preference for avoiding single letter variable names where reasonable names can be given.

src/analyses/ai.h
```diff
@@ -433,16 +433,25 @@ class ait:public ai_baset
 protected:
   /// Map from locations to domain elements, for the results of a static
   /// analysis.
+  typedef std::shared_ptr<statet> s_ptrt;
```
hannes-steffenhagen-diffblue: `state_ptrt`?

src/analyses/ai.h
```diff
@@ -439,14 +443,17 @@ class ait:public ai_baset
...
+
   /// Look up the analysis state for a given location, instantiating a new state
   /// if required. Used internally by the analysis.
   virtual statet &get_state(locationt l) override
   {
     typename state_mapt::const_iterator it = state_map.find(l);
     if(it == state_map.end())
     {
-      auto p = state_map.insert(std::make_pair(l, std::make_shared<domainT>()));
+      std::shared_ptr<statet> d(domain_factory->make_domain(l));
```
hannes-steffenhagen-diffblue: `domain`, `domain_factory->make_domain(location)`?

src/analyses/ai.h
```diff
@@ -439,14 +443,17 @@ class ait:public ai_baset
...
+      auto p = state_map.insert(std::make_pair(l, d));
```
hannes-steffenhagen-diffblue: `insertion_result`?

src/analyses/invariant_propagation.cpp
```diff
@@ -15,6 +15,39 @@ Author: Daniel Kroening, kroening@kroening.com
...
+  {
+    auto p = util_make_unique<invariant_set_domaint>(
+      ip.value_sets, ip.object_store, ip.ns);
+    CHECK_RETURN(p->is_bottom());
+
+    return p;
+  }
+
+private:
+  invariant_propagationt &ip;
```
hannes-steffenhagen-diffblue, line 37: Why is this a reference rather than a constant pointer like in the other factory?

src/analyses/ai.h
```diff
@@ -36,7 +36,8 @@ class ai_baset
   typedef ai_domain_baset statet;
   typedef goto_programt::const_targett locationt;
 
-  ai_baset()
+  explicit ai_baset(std::unique_ptr<ai_domain_factory_baset> &&df)
```
hannes-steffenhagen-diffblue: Hm, so this means now it's impossible to implement an `ait` without a factory? I think this was possible before.

src/analyses/ai.h
```diff
@@ -406,70 +416,36 @@ class ait:public ai_baset
...
-  }
-
-  /// Used internally by the analysis.
-  std::shared_ptr<const statet>
-  abstract_state_before(locationt t) const override
-  {
-    typename state_mapt::const_iterator it = state_map.find(t);
-    if(it == state_map.end())
-      return domain_factory->make_domain(t);
+    auto p(storage->abstract_state_before(l, *domain_factory));
```
hannes-steffenhagen-diffblue: Could you change this to `auto p = storage->....`? or `auto p{storage->....}`? With the combination of `auto` and the parentheses I had to do a double take here.

Also what's `p`?

src/analyses/ai_storage.h
hannes-steffenhagen-diffblue: Rather, could it just be called `location`?

src/analyses/ai_history.h
hannes-steffenhagen-diffblue: FWIW it's worth commenting if there was a deliberate choice to use an ordered map here, lest someone 'improve' it in the future (and it passing the regression tests by accident).

src/analyses/Makefile
```diff
@@ -1,5 +1,6 @@
 SRC = ai.cpp \
       ai_domain.cpp \
+      ai_history.cpp \
```
hannes-steffenhagen-diffblue, line 3: Commit message on 7ba3feb41979dd9b6763cc2825a559e75641e1fb: s/compatable/compatible

Also, this seems like it could be a whole separate PR on its own

hannes-steffenhagen-diffblue (commented):
Some comments. Mostly nitpicks about spelling and naming and things like that. It's a lot of changes to make in a single PR though

martin-cs:
@hannes-steffenhagen-diffblue I appreciate that opinions vary on variable naming schemes.  I tend to use single letters for short term variables.  Also for things that are "small parameters" I tend to try to match the letter with the type, so we have 'a location l', 'a number n', 'an expression e', etc.  I think I have been consistent about this throughout the patch and followed the existing conventions in this part of the code.

If you insist that variable names are changed before it can be merged, then I will do it.  If not then I would prefer not to and I will try to be more considerate in future patches.

src/analyses/ai_history.h
martin-cs: I could write a test case that will break if the order of exploration / working_sett is changed substantially?  Would that help?

src/analyses/ai.h
martin-cs: Not really.  I mean, ideally, `concurrency_aware_ait` would be deprecated and replaced with a solution that would work with other domains.  With the new storage it should be pretty easy to merge the relevant locations and VSD's three-way-merge would allow concurrent AI to be done properly.  But, right now, I don't have the time to refactor all of that.

I made it virtual so that it is consistent with the merge method in ai_baset.

src/analyses/ai.h
martin-cs: It's referring to the refactoring of how domains are stored that happens 11 commits later.  Although it looks like it from the history, this PR wasn't really written in commit order, or, where it was, it was written with a lot of fore-knowledge of what would happen.

src/analyses/ai.h
martin-cs: Fixed.

src/analyses/ai.h
martin-cs: Changed.

src/analyses/dependence_graph.h
martin-cs: Done.

src/analyses/ai.h
martin-cs: Maybe but I think that is beyond the scope of this PR.

src/analyses/ai_domain.h
martin-cs: The domain factory creates them and hands over the responsibility for their deallocation to the caller.  To me a `unique_ptr` is a clearer and more precise expression of this contract.  There are places (in `ait::visit`) where these are used to create temporary domains that do not have shared pointers to them.  Also, although `ai_storage_baset::abstract_state_before` returns a shared pointer, there is no need for it to be the same domain it stores; these can be stored using `unique_ptr`'s if you want.

src/analyses/dependence_graph.h
martin-cs: I need the iterator (i.e. the `locationt`) rather than the instruction themselves.  Is there a version of the ranged-for construct that supports something like `for(goto_programt::targett i_it : goto_program.instructions)`?

src/analyses/ai.h
martin-cs: ??? By the end of the PR it is...?  I have tried to break into smaller commits so that this is more reviewable and so that each commit has a single purpose.

src/analyses/ai.h
martin-cs: I think this gets fixes later commit.

src/analyses/invariant_propagation.cpp
martin-cs: Ideally they should all be references but I wanted to keep the changes as minimal and obviously correct as possible.

src/analyses/ai.h
martin-cs: It was possible before and remains possible.  `ait<domainT>` has a constructor that creates a suitable factory (using `domainT::domainT()`).  This PR should be source compatible with the previous code base.

src/analyses/ai.h
martin-cs: Now `auto p = storage->abstract_state_before(l, *domain_factory);`.  Better?

src/analyses/Makefile
martin-cs: It's really the key commit for this whole PR.  All of the other commits are just there to make it easier / simpler.  I didn't want to submit those separately as many of them are kind of pointless on their own.  There have already been maybe 4 PRs broken out of this, with more to come.

martin-cs:
@hannes-steffenhagen-diffblue I believe I have addressed all of your comments.  As some of them were mistakes in commit messages, I knew that it would break the tracking of comments to locations and require other rebasing so I rebased to the latest origin/develop and folded the code tweaks back into the original commits.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fe163e6a22ae33ecb3b9cfa47cfde79485a7ab5d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/125268131

src/analyses/ai_domain.h
```diff
@@ -122,4 +128,48 @@ class ai_domain_baset
...
+public:
+  typedef ai_domain_baset statet;
+  typedef ai_domain_baset::locationt locationt;
+
+  virtual std::unique_ptr<statet> make_domain(locationt l) const = 0;
+  virtual std::unique_ptr<statet> copy_domain(const statet &s) const = 0;
+};
+
+// It would be great to have a single (templated) default implementation.
+// However, a default constructor is not part of the ai_domain_baset inteface
```
peterschrammel: interface

src/analyses/invariant_propagation.h
```diff
@@ -48,6 +42,8 @@ class invariant_propagationt:public
   typedef ait<invariant_set_domaint> baset;
 
 protected:
+  friend invariant_set_domain_factoryt;
```
peterschrammel, line 51: Why that?

src/analyses/ai.h
```diff
@@ -333,16 +334,28 @@ class ai_baset
...
+  }
 
-  virtual bool merge(const statet &src, locationt from, locationt to)=0;
+  /// Make a copy of a state
+  virtual std::unique_ptr<statet> make_temporary_state(const statet &s)
+  {
+    return domain_factory->copy_domain(s);
+  }
+
+  // abstract methods
```
peterschrammel: This comment can be removed IMO.

src/analyses/ai.h
```diff
@@ -333,16 +334,28 @@ class ai_baset
     const exprt::operandst &arguments,
     const namespacet &ns);
 
-  // abstract methods
+  // Domain creation
```
peterschrammel: `/// For domain creation` ?

src/analyses/ai_storage.h
```diff
@@ -0,0 +1,125 @@
...
+#include <goto-programs/goto_program.h>
+
+#include "ai_domain.h"
+
+/// This is the basic interface for storing domains.
+/// The abstract interpreters should use this interface by default.
+class ai_storage_baset
+{
+public:
+  ai_storage_baset()
```
peterschrammel, line 36: protected?

src/analyses/ai_storage.h
```diff
@@ -0,0 +1,125 @@
...
+
+  /// Reset the abstract state
+  virtual void clear()
+  {
+    return;
+  }
+
+  /// Notifies the storage that the user of the domain will not need the value
+  /// for this location.  After this has been called abstract_state_before may
+  /// return an over-approximation of the domain and get_state may give an
```
peterschrammel: over-approximation of the value ?

src/analyses/ai.cpp
```diff
@@ -327,118 +316,174 @@ bool ai_baset::visit_edge(
...
+  const goto_functionst &goto_functions,
+  const namespacet &ns)
+{
+  // This is the edge from call site to function head.
   {
-    locationt l_begin=goto_function.body.instructions.begin();
+    locationt l_begin = callee.instructions.begin();
 
-    working_sett working_set; // Redundant; fixpoint will add l_begin
+    working_sett ignore_working_set; // Redundant; fixpoint will add l_begin
```
peterschrammel: Not sure I understand what the comment means

src/analyses/ai.h
```diff
@@ -312,6 +312,26 @@ class ai_baset
     const goto_functionst &goto_functions,
     const namespacet &ns);
 
+  // function calls and return are special cases
```
peterschrammel, line 446: Can these comments be made more doxygen-friendly (as generally more systematic class/function-level documentation would be appreciated)?

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+  typedef ai_history_baset tracet;
+
+  /// History objects are intended to be immutable so they can be shared
+  /// to reduce memory overhead
+  typedef std::shared_ptr<const tracet> trace_ptrt;
+
+  /// Create a new history starting from a given location
+  /// This is used to start the analysis from scratch
+  /// PRECONDITION(l.is_dereferenceable());
+  explicit ai_history_baset(locationt)
```
peterschrammel, line 49: Why is this empty definition needed here?

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+/// per history.  Depending on what storage is used, they may also control
+/// or influence the number of domains that are used, supporting delayed
+/// merging, loop unwinding, context dependent analysis, etc.
+
+/// This is the base interface; derive from this.
+class ai_history_baset
+{
+public:
+  typedef goto_programt::const_targett locationt;
+  typedef ai_history_baset tracet;
```
peterschrammel: What does this typedef add? I find it rather confusing (besides, the term trace already being used for something different).

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+
+/// As more detailed histories can get complex (for example, nested loops
+/// or deep, mutually recursive call stacks) they often need some user
+/// controls or limits.  The best way to do this is to give the limits to
+/// the first, start-of-history object and let it propagate.  Having a
+/// factory gives a way of wrapping this information up in a common interface
+class ai_history_factory_baset
+{
+public:
+  // Creates a new history from the given starting point
```
peterschrammel: ///

src/analyses/ai_history.h
```diff
@@ -0,0 +1,237 @@
...
+  virtual ai_history_baset::trace_ptrt bang(ai_history_baset::locationt) = 0;
+};
+
+// An easy factory implementation for histories that don't need parameters
```
peterschrammel: ///
etc

src/analyses/ai_storage.h
```diff
@@ -35,20 +36,36 @@ class ai_storage_baset
   }
 
   typedef ai_domain_baset statet;
-  typedef std::shared_ptr<const statet> state_ptrt;
+  typedef std::shared_ptr<statet> state_ptrt;
+  typedef std::shared_ptr<const statet> cstate_ptrt;
+  typedef ai_history_baset tracet;
```
peterschrammel, line 48: I don't like `tracet`...

src/analyses/ai.h
```diff
@@ -8,6 +8,39 @@ Author: Daniel Kroening, kroening@kroening.com
 
 /// \file
 /// Abstract Interpretation
+///
+/// This is the core of the abstract interpretation framework.
```
peterschrammel, line 12: Cool. Great!!!

src/analyses/ai.h
```diff
@@ -8,6 +8,39 @@ Author: Daniel Kroening, kroening@kroening.com
...
+///  2. A history factory, derived from ai_history_factory_baset.
+///     This generates history objects (derived from ai_history_baset) which
+///     control the number of steps that the analysis performs.  These can be
+///     simple, just tracking location, or they can be complex, tracking
+///     location, function calls, branches (forwards and backwards), even
+///     threading.
+///     See ai_history.h for more information.
+///
+///  3. A domain factory, derived from ai_domain_factory_baset.
+///     This generates domain objects (derived from ai_domain_baset) which
```
peterschrammel, line 30: We should decide on a terminology here and then use it consistently to avoid confusion. Domain, domain object, domain element, domain value, abstract value, ... all of these are used somewhere (all of which are legit - except 'domain').

src/analyses/ai_storage.h
```diff
@@ -168,12 +168,12 @@ class location_sensitive_storaget : public trace_map_storaget
   }
 
   cstate_ptrt abstract_state_before(
-    locationt t,
+    locationt l,
```
peterschrammel, line 178: Please squash with the commit that introduces this.

src/analyses/ai.cpp
```diff
@@ -156,22 +156,27 @@ xmlt ai_baset::output_xml(
...
 
   goto_functionst::function_mapt::const_iterator
     f_it=goto_functions.function_map.find(goto_functions.entry_point());
 
   if(f_it!=goto_functions.function_map.end())
-    entry_state(f_it->second.body);
+    return entry_state(f_it->second.body);
+
+  // It is not clear if this is even a well-structured goto_functionst object
+  return nullptr;
```
peterschrammel, line 172: That should probably go into the interface docs

src/analyses/ai_history.h
hannes-steffenhagen-diffblue: @martin-cs The problem with tests like this is that if an unordered collection is used you can always get the ordering you _meant_ to get by sheer accident. This has bitten us in other places in the past. I think it's fair enough to just leave a comment here and hopefully that should at the very least flag things up during review

martin-cs: I have expanded the comment to make it clearer why the ordering and per-history control of it is important.

src/analyses/ai_domain.h
martin-cs: Thanks.  Fixed.

src/analyses/invariant_propagation.h
martin-cs: Each `invariant_set_domain` needs access to a few of the fields of the invariant_propagation object.  This is a kinda wonky encapsulation violation but it would require a substantial refactor to remove this.  Instead what I have done is the `invariant_set_domain_factory` is created with a reference to the `invariant_propagationt` and creates the domain objects with references to the relevant parts of `invariant_propagationt`.  To do this it has to be a friend.

It's ugly but what it replaced was worse.

src/analyses/ai.h
martin-cs: I think it is in a later patch.  (I am assuming you are reviewing per patch.)

src/analyses/ai.h
martin-cs: `/// For creating domain objects`

src/analyses/ai_storage.h
martin-cs: Do you want the constructors of all abstract interfaces to be `protected`?

src/analyses/ai_storage.h
martin-cs: I've reworded this comment somewhat.

src/analyses/ai.cpp
martin-cs: Up until the last commit of the PR the fixpoint function has a design oddity; it starts the fixpoint calculation with the first element of the function.  What this means is that the entry in the working set that is generated by computing the call edge will be ignored and the same thing recreated when the recursive call to fixpoint is made.  The naming of the variable is intended to make that a little clearer.

The last patch adds a 'starting history' parameter to fixpoint so that history objects can be threaded through calls in the expected way rather than being destroyed and recreated.  So both the code and the comment should be gone by the time the PR is merged.

src/analyses/ai.h
martin-cs: Can I take "improve the documentation" along with the point about terminology below as actions for future PRs rather than things that need to be done in this PR?

src/analyses/ai_history.h
martin-cs: To make clear that it is part of the the interface and to document it.

src/analyses/ai_history.h
martin-cs: Removed.

src/analyses/ai_history.h
martin-cs: Done.

src/analyses/ai_history.h
martin-cs: Fixed.

src/analyses/ai_storage.h
martin-cs: Removed.

src/analyses/ai.h
martin-cs: :-)

src/analyses/ai.h
martin-cs: Can that be an action for another PR as I think that issue is wider than just the code touched by this PR?

src/analyses/ai_storage.h
martin-cs: I tried; it was a mess.  I will try again when doing a final rebase and squash of revision comments.

src/analyses/ai.cpp
martin-cs: I had hoped that the invariants / check whether a `goto_programt` is well structured would eventually give a testable documentation of this kind of thing.  I am still not clear about this.  The interface that calls the abstract interpreter on a single `goto_programt` by synthesising a `goto_functionst` for it to live in is kinda suspect but I didn't want to change the functionality in this patch so ... it is still there.

martin-cs:
@hannes-steffenhagen-diffblue and @peterschrammel : I have pushed a series of commits that should address your comments.  If they resolve your concerns I will:

1. Rebase on to the latest master.
2. Squash them into the relevant commits (along with the one Peter wanted squashed in).
3. Add "improve the doxygen quality of comments" to my to-do list.
4. Start an issue for sorting out the terminology.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 09f4e1a265d55b04a6a990120344b20ad28b0c63).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/127098399

martin-cs:
I'm guessing the @allredj bot failure is because I haven't rebased on to the latest develop.  I will see if it goes away after the rebase as this has passed before.

hannes-steffenhagen-diffblue (approved):
My preferences for variable names notwithstanding I believe this is probably OK now

martin-cs:
Thanks @hannes-steffenhagen-diffblue I will try to avoid single variable names in future PRs.

src/analyses/ai.cpp
```diff
@@ -304,231 +308,223 @@ bool ai_baset::visit(
...
+  trace_ptrt p,
   const irep_idt &to_function_id,
-  const locationt &to_l,
+  locationt to_l,
   const namespacet &ns,
   working_sett &working_set)
 {
+  // Has history taught us not to step here...
+  auto next = p->step(to_l, *(storage->abstract_traces_before(to_l)));
+  if(next.first == ai_history_baset::step_statust::BLOCKED)
```
chrisr-diffblue, line 324: This kind of thing would be easier to read if we had proper accessors rather than just "first"...

src/analyses/ai.h
chrisr-diffblue: Seems reasonable to me to leave that for a different PR - maybe raising a github issue to track that it's something we should do?

src/analyses/ai.h
```diff
@@ -51,8 +147,8 @@ class ai_baset
   {
     goto_functionst goto_functions;
     initialize(function_id, goto_program);
-    entry_state(goto_program);
-    fixedpoint(function_id, goto_program, goto_functions, ns);
+    trace_ptrt p = entry_state(goto_program);
```
chrisr-diffblue, line 150: In this instance, and the couple that follow this, I'm not sure that having a temporary `p` actually adds much - I'd be tempted to just do `fixpoint(entry_state(blah), blah...` instead.

src/analyses/ai_domain.h
```diff
@@ -122,4 +184,74 @@ class ai_domain_baset
...
+// methods.
+template <typename domainT>
+class ai_domain_factoryt : public ai_domain_factory_baset
+{
+public:
+  typedef ai_domain_factory_baset::statet statet;
+  typedef ai_domain_factory_baset::locationt locationt;
+  typedef ai_domain_factory_baset::trace_ptrt trace_ptrt;
+
+  std::unique_ptr<statet> copy_domain(const statet &s) const override
```
chrisr-diffblue: Is there any guidance on whether the 'copy' is/should be a deep copy, or whether the 'copy' might be a copy-on-write copy, or something in between? Also, why is this called `copy_domain` and not just `copy` ?

src/analyses/ai_history.h
```diff
@@ -0,0 +1,244 @@
...
+
+      return std::make_pair(step_statust::MERGED, *it);
+    }
+  }
+
+  bool operator<(const ai_history_baset &op) const override
+  {
+    PRECONDITION(dynamic_cast<const ahistoricalt *>(&op) != nullptr);
+
+    return this->current_location()->location_number <
```
chrisr-diffblue, line 186: Is this always a valid comparison? Is there any circumstance where `op` could give a location number in a different function, say?

src/analyses/ai_history.h
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  const locationt &current_location(void) const override
+  {
+    return current;
+  }
+
+  // Use default widening
+  // without history there is no reason to say any location is better than
+  // another to widen.
```
chrisr-diffblue, line 210: This comment seems a bit random to have here (though I understand why its here). Would it be more immediately clear if there was an explicit implementation here, even if that implementation just defers to the superclass implementation?

src/analyses/ai_history.h
```diff
@@ -0,0 +1,244 @@
...
+/// As more detailed histories can get complex (for example, nested loops
+/// or deep, mutually recursive call stacks) they often need some user
+/// controls or limits.  The best way to do this is to give the limits to
+/// the first, start-of-history object and let it propagate.  Having a
+/// factory gives a way of wrapping this information up in a common interface
+class ai_history_factory_baset
+{
+public:
+  /// Creates a new history from the given starting point
+  virtual ai_history_baset::trace_ptrt bang(ai_history_baset::locationt) = 0;
```
chrisr-diffblue: While I appreciate the humour :-) Would  'epoch' be a better/more instantly understandable name?

src/analyses/ai_storage.h
```diff
@@ -0,0 +1,298 @@
...
+    const ai_domain_factory_baset &fac) const = 0;
+
+  virtual cstate_ptrt abstract_state_before(
+    locationt l,
+    const ai_domain_factory_baset &fac) const = 0;
+
+  /// Look up the analysis state for a given history,
+  /// instantiating a new domain if required.
+  virtual statet &
+  get_state(trace_ptrt p, const ai_domain_factory_baset &fac) = 0;
```
chrisr-diffblue, line 75: Nit: Indentation seems a bit whacky here? Or is that clang_format being annoying?

src/analyses/ai_storage.h
```diff
@@ -0,0 +1,298 @@
...
+
+  /// Reset the abstract state
+  virtual void clear()
```
chrisr-diffblue, line 78: Is it worth making clear (hah!) what is meant by "resetting" or "clearing" the abstract state as a concept? When might this be called, for instance?

src/analyses/ai_storage.h
```diff
@@ -0,0 +1,298 @@
...
+    return;
+  }
+
+  /// Notifies the storage that the user will not need the domain object(s)
+  /// for this location.  After this has been called abstract_state_before may
+  /// return an over-approximation of the value and get_state may give an
+  /// under-approximation (forcing recomputation).
+  /// If there are multiple histories that reach this location all will be
+  /// affected
+  virtual void prune(locationt l)
```
chrisr-diffblue, line 89: So this is like `clear()` but for only a single location?

src/analyses/ai_storage.h
```diff
@@ -0,0 +1,298 @@
...
+  trace_mapt trace_map;
+
+  // This retains one part of a shared_ptr to the history object
+  void register_trace(trace_ptrt p)
+  {
+    // Save the trace_ptrt
+    trace_mapt::iterator it = trace_map.find(p->current_location());
+    if(it == trace_map.end())
+    {
+      trace_set_ptrt s(new trace_sett());
```
chrisr-diffblue, line 111: Is this construct the kind of thing that should just have a static method on trace_sett? e.g. `trace_sett::new_trace_set_ptr()` and then hide it's actual constructor?

src/analyses/ai_storage.h
```diff
@@ -0,0 +1,298 @@
...
+    // Strictly this should be "it->second points to a trace_set"
+    POSTCONDITION(it->second != nullptr);
+    return it->second;
+  }
+
+  void clear() override
+  {
+    ai_storage_baset::clear();
+    trace_map.clear();
+    return;
```
chrisr-diffblue, line 140: Why the return?

chrisr-diffblue (approved):
A bunch of minor comments/queries, and I'll vote with @hannes-steffenhagen-diffblue that I don't particularly like the single character identifiers in a lot of places - but I'm not going to block this PR on any of those.

src/analyses/ai.cpp
martin-cs: At the moment I was using `std::pair` because I figured it was better to use the standard types than to replicate stuff but if you want I can create a `step_return_valuet` object?

src/analyses/ai.h
martin-cs: I will as part of my "things to do once everyone is happy with this PR".

src/analyses/ai_history.h
martin-cs: It can and that is intentional.  As I understand it, location numbers are intended to be globally unique within a goto_model.  So comparing location numbers is the safe version of comparing `locationt`s.

src/analyses/ai_storage.h
martin-cs: I just used clang-format ...

src/analyses/ai_storage.h
martin-cs: At the moment I am not sure that it is called.  I left it in for backwards compatibility.  I am not sure that it has a use-case that wouldn't be better served by destroying the abstract interpreter object and starting again TBH.

src/analyses/ai_storage.h
martin-cs: Sort of.  At the moment the abstract interpreter stores all of its results in memory.  That's great if you want to be able to do random access into them after the analysis is done but if you only care about, say, proving assertions, then most of the domains could be thrown away without it being a problem.  I'm planning on using this to implement a much more memory compact analyser.

The key trick is that if you prune the domain, you can return bottom when the internal interface asks for it (forcing it to be recomputed) and top when the external interface asks for it.

src/analyses/ai_storage.h
martin-cs: Maybe... trace_sett is a typedef around standard library objects so it would require quite a bit of code for simplifying two calls.  If it were already a custom object then, hell yeah, but as it is I am not sure it is a net saving.

src/analyses/ai_storage.h
martin-cs: I tend to put explicit return statements into things with void return type so that it is clear that it is done, I haven't just got part way through and wandered off and forgotten what I was doing.  Not that that has ever happened.

src/analyses/ai.h
martin-cs: Fair point but if you inline it then you have to wrap the function calls over multiple lines so this is actually a smaller change.

src/analyses/ai_domain.h
martin-cs: How the factory and the domain handle copy is entirely up to them.  It is an implementation decision and this interface shouldn't constrain it.

Have renamed s/copy_domain/copy/ to s/make_domain/make/ .

src/analyses/ai_history.h
martin-cs: I could but it is an inherited virtual function so that kind of is duplication.  I'll do it if you want but I'd rather leave it as just a comment.

src/analyses/ai_history.h
martin-cs: Yeah... Changed.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fa1c03d1c2810fd85f3d60e0a6b749c399d5f362).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/127648219

martin-cs:
@chrisr-diffblue : I have responded to all of your comments and patched the ones I think need patching.  Unless you object I will continue with the PR.

@peterschrammel : If you can approve the changes I made I think we are ready to do the rebase / squash / merge.

src/analyses/ai_storage.h
peterschrammel: If their only purpose is to be called from a deriving class then yes.

src/analyses/invariant_propagation.h
peterschrammel: Ok, adding your insights above as comment in the code might help future refactoring efforts.

peterschrammel (approved):

src/analyses/invariant_propagation.h
martin-cs: Added as a comment.

src/analyses/ai_storage.h
martin-cs: Done.

martin-cs:
All comments resolved; next:

1. Rebase on to the latest master.
2. Squash them into the relevant commits (along with the one Peter wanted squashed in).
3. Add "improve the doxygen quality of comments" to my to-do list.
4. Start an issue for sorting out the terminology.

src/analyses/reaching_definitions.cpp
```diff
@@ -26,9 +26,34 @@ Date: February 2013
...
 #include "dirty.h"
 
+/// This ensures that all domains are constructed with the appropriate pointer
+/// back to the analysis engine itself.  Using a factory is a tad verbose
+/// but it works well with the ait infrastructure.
+class rd_range_domain_factoryt : public ai_domain_factoryt<rd_range_domaint>
+{
+public:
+  rd_range_domain_factoryt(
+    sparse_bitvector_analysist<reaching_definitiont> *_bv_container)
```
smowton, line 36: If this is never null, should it just be a reference?

src/analyses/ai_domain.h
```diff
@@ -137,6 +137,11 @@ class ai_domain_factory_baset
 
   virtual std::unique_ptr<statet> make_domain(locationt l) const = 0;
   virtual std::unique_ptr<statet> copy_domain(const statet &s) const = 0;
+
+  // Not domain construction but requires knowing the precise type of statet
+  virtual bool
+  merge(statet &dest, const statet &src, locationt from, locationt to)
```
smowton: Rather than apologise for "factory" containing a non-factory method, how about "domain-manager"?

src/analyses/ai.cpp
```diff
@@ -197,16 +198,18 @@ void ai_baset::finalize()
...
-ai_baset::locationt ai_baset::get_next(
-  working_sett &working_set)
+ai_baset::trace_ptrt ai_baset::get_next(working_sett &working_set)
 {
   PRECONDITION(!working_set.empty());
 
-  working_sett::iterator i=working_set.begin();
-  locationt l=i->second;
-  working_set.erase(i);
+  // STL guarantees this will be key minimal
```
smowton: If you're relying on this you should probably static-assert that `working_sett` is the type you expect

src/analyses/ai.cpp
```diff
@@ -289,27 +294,36 @@ bool ai_baset::visit(
 
 bool ai_baset::visit_edge(
   const irep_idt &function_id,
-  locationt l,
+  trace_ptrt p,
```
smowton, line 316: I'd appreciate `p` being named something more informative (just "trace_ptr"?) here and elsewhere. The various "l_" things at least referred to "location" rather than just "pointer" (to what?)

src/analyses/ai_storage.h
```diff
@@ -211,4 +211,82 @@ class location_sensitive_storaget : public trace_map_storaget
   }
 };
 
+// The most precise form of storage
+class history_sensitive_storaget : public trace_map_storaget
+{
+protected:
+  typedef std::map<trace_ptrt, state_ptrt, ai_history_baset::compare_historyt>
```
smowton, line 225: Do you need the ordered-ness of `map`?

smowton (approved):

src/analyses/reaching_definitions.cpp
martin-cs: I converted at least one case of these when I was doing this, the others I didn't convert for reasons.  IIRC these were wanting to keep the patches as small and focused as possible and it's not quite that simple; there are some cases where it isn't that straight forward.

If it bothers you I can write a separate PR for it?

src/analyses/ai_domain.h
martin-cs: Yes ... probably the right name for it but ... uhhh... that's going to be a massive pain to fix as it is in comments, commit messages and so on and for much of the patch set it is a sensible name.

I am going to open a discussion issue on renaming things as @peterschrammel requested, can we deal with this as part of that?

src/analyses/ai.cpp
martin-cs: Done.

src/analyses/ai_storage.h
martin-cs: Not directly but we already have a compare function and no hash function so I figured it was better.  Also I would prefer to avoid the reallocation that comes with growing hash maps.

src/analyses/ai.cpp
martin-cs: Fair point.  As there is already going to be a PR for source code text changes that don't alter the functionality, can I address this as part of that?

martin-cs:
If the latest rebase passes CI and no-one objects I will merge this and open a discussion issue for fixing the source text issues.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 27f27f85b2e59c7a93421b2346551799cb65be65).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/136626871

src/analyses/ai.cpp
chrisr-diffblue: I'm OK with that approach, but it might be helpful to open an issue to track all the name-related cleanup that needs to be done so that things like this can be logged and then checked in the PR for text-only changes later - otherwise I fear they might get forgotten...

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 529d7a52be4374390f0b14616501cdb2b17a2808).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/138137061

martin-cs:
( @chrisr-diffblue re-ran the test-gen build and said the one above is a Travis failure rather than an actual issue. )

PR #1787, opened 2018-01-31
Use more efficient algorithm to calculate dominators

hannes-steffenhagen-diffblue:
Use Lengauer&Tarjan's algorithm to calculate the tree of
immediate dominators instead of the previous algorithm,
which gives us a significant performance improvement.

Also, instead of storing the dominators in a std::set we
now just store the immediate dominators and make lookups
in there instead. This is a significant performance improvement
as long as the dominators set are only iterated over or only
queried infrequently, as is currently the case, and also
saves a lot of memory in the case of large functions (previous
memory usage was quadratic with function size, now it is linear).

In cases where a lot of queries are made against the same set of dominators,
they can still be copied to a local std::set prior to that.

PR #5042, opened 2019-08-20
Dominator analysis: analyse on basic block granularity

smowton:
Our dominator analysis could produce very large data structures because it computed dominator sets on an instruction granularity. In typical programs the basic-block graph is of course considerably smaller, which combined with the worst-case n-squared space complexity of dominator sets can save a great deal of memory.

Of course there are knock-on consequences for users: most are simple, `natural_loopst` needs a little restructuring, and `accelerate` et al need some awkward functions to accommodate their desire to edit the program while keeping the analysis up to date (if this is too awful we could insist that it fully recomputes the analysis after each change).

Accelerate has no tests, so it's hard to know if it's still working, but the Java front-end uses dominator analysis for local variable analysis, so the JBMC testsuite provides good testing of `cfg_dominatorst` (but not `natural_loopst`)

### DIFF
--- a/jbmc/src/java_bytecode/java_local_variable_table.cpp
+++ b/jbmc/src/java_bytecode/java_local_variable_table.cpp
@@ -41,6 +41,7 @@ struct procedure_local_cfg_baset<
                    java_bytecode_convert_methodt::method_offsett>
     entry_mapt;
   entry_mapt entry_map;
+  std::vector<java_bytecode_convert_methodt::method_offsett> possible_keys;
 
   procedure_local_cfg_baset() {}
 
@@ -52,6 +53,7 @@ struct procedure_local_cfg_baset<
     {
       // Map instruction PCs onto node indices:
       entry_map[inst.first]=this->add_node();
+      possible_keys.push_back(inst.first);
       // Map back:
       (*this)[entry_map[inst.first]].PC=inst.first;
     }
@@ -121,6 +123,11 @@ struct procedure_local_cfg_baset<
   static bool nodes_empty(const method_with_amapt &args)
   {
     return args.second.empty();
+  }
+
+  const std::vector<java_bytecode_convert_methodt::method_offsett> &keys()
+  {
+    return possible_keys;
   }
 };
 
@@ -463,13 +470,19 @@ static java_bytecode_convert_methodt::method_offsett get_common_dominator(
     candidate_dominators;
   for(auto v : merge_vars)
   {
-    const auto &dominator_nodeidx=
+    const auto &var_start_basic_block =
       dominator_analysis.cfg.entry_map.at(v->var.start_pc);
-    const auto &this_var_doms=
-      dominator_analysis.cfg[dominator_nodeidx].dominators;
-    for(const auto this_var_dom : this_var_doms)
-      if(this_var_dom<=first_pc)
-        candidate_dominators.push_back(this_var_dom);
+    const auto &this_var_dom_blocks =
+      dominator_analysis.cfg[var_start_basic_block].dominators;
+    for(const auto this_var_dom_block_index : this_var_dom_blocks)
+    {
+      const auto &this_var_dom_block =
+        dominator_analysis.cfg[this_var_dom_block_index].block;
+      // Only consider placing variable declarations at the head of
+      // a basic block (which conveniently is always a safe choice, even
+      // for a live range starting midway through a block)
+      candidate_dominators.push_back(this_var_dom_block.front());
+    }
   }
   std::sort(candidate_dominators.begin(), candidate_dominators.end());
 
--- a/src/analyses/cfg_dominators.h
+++ b/src/analyses/cfg_dominators.h
@@ -248,9 +248,9 @@ inline void dominators_pretty_print_node(
 template <class P, class T, bool post_dom>
 void cfg_dominators_templatet<P, T, post_dom>::output(std::ostream &out) const
 {
-  for(const auto &node : cfg.entries())
+  for(const auto &n : cfg.keys())
   {
-    auto n=node.first;
+    const auto &node = cfg.get_node(n);
 
     dominators_pretty_print_node(n, out);
     if(post_dom)
@@ -258,7 +258,7 @@ void cfg_dominators_templatet<P, T, post_dom>::output(std::ostream &out) const
     else
       out << " dominated by ";
     bool first=true;
-    for(const auto &d : cfg[node.second].dominators)
+    for(const auto &d : node.dominators)
     {
       if(!first)
         out << ", ";
--- a/src/analyses/cfg_dominators.h
+++ b/src/analyses/cfg_dominators.h
@@ -37,14 +37,14 @@ template <class P, class T, bool post_dom>
 class cfg_dominators_templatet
 {
 public:
-  typedef std::set<T> target_sett;
+  typedef std::set<std::size_t> target_sett;
 
   struct nodet
   {
     target_sett dominators;
   };
 
-  typedef procedure_local_cfg_baset<nodet, P, T> cfgt;
+  typedef cfg_basic_blockst<procedure_local_cfg_baset, nodet, P, T> cfgt;
   cfgt cfg;
 
   void operator()(P &program);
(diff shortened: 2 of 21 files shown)


src/goto-programs/goto_program.h
```diff
@@ -1068,34 +1072,24 @@ class goto_programt
...
-///   program
-/// \param target: pointer to the instruction of which to get the successors of
-/// \return List of control-flow successors of the pointed-to goto program
-///   instruction
 template <typename Target>
-std::list<Target> goto_programt::get_successors(
-  Target target) const
+std::list<Target> get_successors(
+  Target target,
+  std::function<bool(const Target &)> is_instruction_out_of_range)
```
hannes-steffenhagen-diffblue, line 1078: If this is already a template and you're not storing the function anywhere there's no real reason to use `std::function` here, you could just add the function type as another template parameter (I'm not sure if it's going to matter in this particular instance but since this avoids type erasure it can under some circumstance dramatically affect optimisation, for example in [this case](https://godbolt.org/z/9rsYy0) the difference is in one case generating a straight loop with virtual function calls and another that's partially unrolled/vectorised).

jbmc/src/java_bytecode/java_local_variable_table.cpp
```diff
@@ -442,13 +442,19 @@ static java_bytecode_convert_methodt::method_offsett get_common_dominator(
...
-      if(this_var_dom<=first_pc)
-        candidate_dominators.push_back(this_var_dom);
+    const auto &this_var_dom_blocks =
+      dominator_analysis.cfg[var_start_basic_block].dominators;
+    for(const auto this_var_dom_block_index : this_var_dom_blocks)
+    {
+      const auto &this_var_dom_block =
+        dominator_analysis.cfg[this_var_dom_block_index].block;
+      // Only consider placing variable declarations at the head of
+      // a basic block (which conveniently is always a safe choice, even
```
hannes-steffenhagen-diffblue, line 482: I don't know what this means (i.e. why is it a safe choice and why would it not be a safe choice?)

src/analyses/cfg_dominators.h
```diff
@@ -26,20 +26,44 @@ template <class P, class T, bool post_dom>
 class cfg_dominators_templatet
 {
 public:
-  typedef std::set<T> target_sett;
+  typedef std::set<std::size_t> target_sett;
```
hannes-steffenhagen-diffblue, line 40: I'm guessing this is referring to the node indices? In that case, shouldn't this use `node_indext`?

:pick: (As a side note, for cases like this I'd also prefer a "strong typedef" i.e. wrapping it in a struct just to avoid accidentally doing things with them they were not meant to do - like comparing them to unrelated indices or adding them together, things like that).

src/analyses/cfg_dominators.h
```diff
@@ -26,20 +26,44 @@ template <class P, class T, bool post_dom>
...
   cfgt cfg;
 
   void operator()(P &program);
 
   T entry_node;
 
+  /// Returns true if basic block \p lhs [post]dominates \p rhs
+  bool basic_block_dominates(std::size_t lhs, std::size_t rhs) const
+  {
+    return cfg[rhs].dominators.count(lhs);
```
hannes-steffenhagen-diffblue, line 75: :pick: ` > 0` for additional verbosity? Never a big fan of implicit to-bool conversion myself

src/analyses/cfg_dominators.h
```diff
@@ -26,20 +26,44 @@ template <class P, class T, bool post_dom>
...
+  bool dominates(T lhs, T rhs) const;
+
+  const target_sett &basic_block_dominators(std::size_t block) const
+  {
+    return cfg[block].dominators;
+  }
+
+  bool basic_block_reachable(std::size_t block) const
+  {
+    return !cfg[block].dominators.empty();
```
hannes-steffenhagen-diffblue: Wouldn't this be false for the first block?

src/analyses/cfg_dominators.h
```diff
@@ -158,12 +183,39 @@ void cfg_dominators_templatet<P, T, post_dom>::fixedpoint(P &program)
...
+
+  for(const auto &instruction : cfg[lhs_block].block)
+  {
+    if(instruction == lhs)
+      return !post_dom;
+    else if(instruction == rhs)
+      return post_dom;
+  }
+
+  UNREACHABLE; // Entry map is inconsistent with block members?
```
hannes-steffenhagen-diffblue, line 278: :pick: I think in this case it might be better to just use a regular invariant and put the comment on that instead

src/analyses/cfg_dominators.h
```diff
@@ -158,12 +183,39 @@ void cfg_dominators_templatet<P, T, post_dom>::fixedpoint(P &program)
...
   }
 }
 
+template <class P, class T, bool post_dom>
+bool cfg_dominators_templatet<P, T, post_dom>::dominates(T lhs, T rhs) const
+{
+  if(lhs == rhs)
+    return true;
+
+  const auto lhs_block = cfg.entry_map.at(lhs);
```
hannes-steffenhagen-diffblue: :pick: Maybe call these `block_id` or something of the sort

hannes-steffenhagen-diffblue (approved):
Looks fine to me, mostly nitpicks (and because the second commit has a bunch of changes all at once it's a bit hard to review but since a lot of these need to be dragged through pretty winded code paths I'm not sure if this even can be sensibly broken up further)

src/analyses/cfg_dominators.h
smowton: All blocks dominate themselves by definition

smowton:
@hannes-steffenhagen-diffblue take another look, I had to make the get-basic-block-from-instruction operation cheaper to get acceptable performance, for which I introduced `dense_integer_mapt`.

Preliminary performance figures, using the same benchmarking example as you did in https://github.com/diffblue/cbmc/pull/1787#issuecomment-362337130

Examples:
* `straightline` is like `v1` in the linked comment: a long straight-line function.
* `small`, `medium` and `largeish` are 100, 1000 and 3000 repeats of example `v2` in the linked comment (a function with lots of small basic blocks)
* `blocks` is a version of `medium` with basic blocks of size 3

As with the linked comment, I used `dependency-graph` to benchmark as a pass that both computes postdoms and then hammers it with queries (58% of runtime is spent in `control_dependencies` of which almost all of that is postdom queries).

Results:

* Straightline: 4.5s -> 0.5s (down 89%)
* Blocks: 7.5s -> 6.0s (down 20%)
* Small: 0.05s -> 0.04s (unmeasurable)
* Medium: 4.6s -> 3.3s (down 28%)
* Largeish: 107s -> 61s (down 43%)

So up front this looks to me like we should get a huge win for mostly-straight-line code (as you'd expect from using basic-block form), and moderate wins (20-30%) from code consisting of 2-3 line basic blocks. Unlike #1787 we continue to use Cooper's dominator algorithm, but we shrink the sizes of the sets it tracks using basic-block form, which particularly shows its worth when we deal with large GOTO programs.

src/analyses/cfg_dominators.h
smowton: (except ones not reachable from the entry point, which are never visited by `cfg_dominatorst`)

smowton:
re: breaking up the PR, a lot of this now introduces and uses a sensible API for dominators, cf. the original code which merrily rummaged in both `cfg_dominatorst` and `cfgt`'s fields. If this is looking like it might go in then I'll submit those first as cleanups.

smowton:
Best case against the Java testsuite: `regression/jbmc/coreModels`, down from 3.2s to 1.7s. Total `jbmc-CORE` test suite time down from 67 to 59 seconds.

smowton:
Have so far split out #5045 #5051 #5055, which should be pretty uncontroversial, and #5049, which floats the `dense_integer_mapt` concept for `cfgt` by itself, without the basic-blocks innovation introduced here. I will make one more PR in the morning, to factor out this PR's cleanup changes to `natural_loopst`, then this one will sit on top of the whole stack.

martin-cs:
@smowton I am listed as a reviewer for this and am happy to do so but I think it might be best to wait until things have been rebased / merged.  Correct me if I am wrong.

smowton:
Agreed.

PR #5047, opened 2019-08-22
Update Doxygen version

owen-mc-diffblue:
The old version is no longer available from this host

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ jobs:
     - stage: Linter + Doxygen + non-debug Ubuntu/gcc-5 test
       env:
         NAME: "DOXYGEN-CHECK"
-        DOXYGEN_VERSION: "1.8.14"
+        DOXYGEN_VERSION: "1.8.16"
       addons:
         apt:
           sources:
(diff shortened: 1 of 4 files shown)


thk123 (approved):
Can't imagine point releases mattering. 

Is wget-ing from this serve the best option?

chrisr-diffblue (approved):
Assuming that a minor point release update doesn't break CI doxygen, approved. Depending on how much we want to tie ourselves to a specific Doxygen version it might be better to host the chosen version ourselves - but I don't think the download links have broken very often, so might not be worth the extra management.

chrisr-diffblue (approved):
Assuming that a minor point release update doesn't break CI doxygen, approved. Depending on how much we want to tie ourselves to a specific Doxygen version it might be better to host the chosen version ourselves - but I don't think the download links have broken very often, so might not be worth the extra management.

smowton (approved):

owen-mc-diffblue:
I'm having problems with multi-line doxygen warnings. They seemed to work before.

thk123:
Replaced by #5058

PR #5048, opened 2019-08-22
Conversion functions from primitive types to strings

danpoe:
- n/s My commit message includes data points confirming performance improvements (if claimed).


jbmc/src/java_bytecode/java_string_library_preprocess.cpp
```diff
@@ -1548,6 +1548,21 @@ void java_string_library_preprocesst::initialize_conversion_table()
   cprover_equivalent_to_java_string_returning_function
     ["java::org.cprover.CProverString.substring:(Ljava/Lang/"
      "StringBuffer;II)Ljava/lang/String;"] = ID_cprover_string_substring_func;
+  cprover_equivalent_to_java_string_returning_function
+    ["java::org.cprover.CProverString.toString:(Z)Ljava/lang/String;"] =
+      ID_cprover_string_of_bool_func;
```
romainbrenguier: :pick: actually this could be implemented directly as a model, there is not complicated logic involved, so I don't know if this is needed :lollipop:

jbmc/src/java_bytecode/java_string_library_preprocess.cpp
```diff
@@ -1548,6 +1548,21 @@ void java_string_library_preprocesst::initialize_conversion_table()
...
+  cprover_equivalent_to_java_string_returning_function
+    ["java::org.cprover.CProverString.toString:(C)Ljava/lang/String;"] =
+      ID_cprover_string_of_char_func;
```
romainbrenguier: :lollipop:

jbmc/src/java_bytecode/java_string_library_preprocess.cpp
```diff
@@ -1548,6 +1548,21 @@ void java_string_library_preprocesst::initialize_conversion_table()
...
+  cprover_equivalent_to_java_string_returning_function
+    ["java::org.cprover.CProverString.toString:(I)Ljava/lang/String;"] =
+      ID_cprover_string_of_int_func;
+  cprover_equivalent_to_java_string_returning_function
+    ["java::org.cprover.CProverString.toString:(J)Ljava/lang/String;"] =
+      ID_cprover_string_of_long_func;
+  cprover_equivalent_to_java_string_returning_function
+    ["java::org.cprover.CProverString.toString:(F)Ljava/lang/String;"] =
+      ID_cprover_string_of_float_func;
```
romainbrenguier: why is there no string_of_double entry?

jbmc/regression/jbmc-strings/CProverStringToString/Main.java
```diff
@@ -0,0 +1,184 @@
...
+      r = "デ";
+      break;
+    case 4:
+      c = Character.MIN_VALUE;
+      r = "\u0000";
+      break;
+    /*case 5:
+      c = Character.MAX_VALUE;
+      r = "\uFFFF";
+      break;*/
```
romainbrenguier: why is this commented out? :coffee:

jbmc/regression/jbmc-strings/CProverStringToString/Main.java
```diff
@@ -0,0 +1,184 @@
...
+      r = "Infinity";
+      break;
+    case 7:
+      f = Float.NEGATIVE_INFINITY;
+      r = "-Infinity";
+      break;
+    case 8:
+      f = Float.NaN;
+      r = "NaN";
+      break;*/
```
romainbrenguier: :coffee:

romainbrenguier (approved):

jbmc/src/java_bytecode/java_string_library_preprocess.cpp
danpoe: Double to string is handled by a model via `CProverString.toString(CProver.doubleToFloat(d))` as double to string conversion is not implemented AFAIK.

jbmc/regression/jbmc-strings/CProverStringToString/Main.java
danpoe: This case does not work. I have to investigate why.

jbmc/regression/jbmc-strings/CProverStringToString/Main.java
danpoe: Does not work yet.

jbmc/src/java_bytecode/java_string_library_preprocess.cpp
danpoe: Yes, I think in that case we can remove both `ID_cprover_string_of_bool_func` as well as `CProverString.toString(char)` and use `b ? "true" : "false"` directly.

jbmc/regression/jbmc-strings/CProverStringToString/Main.java
```diff
@@ -0,0 +1,184 @@
...
+      f = 1.1F;
+      r = "1.1";
+      break;
+    /*case 4:
+      f = Float.MAX_VALUE;
+      r = "3.4028235E38";
+      break;
+    case 5:
+      f = Float.MIN_VALUE;
+      r = "1.4E-45";
```
martin-cs, line 106: Where did this number come from?

jbmc/regression/jbmc-strings/CProverStringToString/Main.java
```diff
@@ -0,0 +1,184 @@
...
+      f = -1F;
+      r = "-1.0";
+      break;
+    case 3:
+      f = 1.1F;
+      r = "1.1";
+      break;
+    /*case 4:
+      f = Float.MAX_VALUE;
+      r = "3.4028235E38";
```
martin-cs, line 102: This is kind of a weird way of formatting this.

martin-cs (commented):
A passing thought

jbmc/regression/jbmc-strings/CProverStringToString/Main.java
danpoe: That's the result of `String.valueOf(f)` (same as `Float.toString(f)`) :bread:

jbmc/regression/jbmc-strings/CProverStringToString/Main.java
danpoe: :bread:

jbmc/src/java_bytecode/java_string_library_preprocess.cpp
danpoe: We could use
```
char arr[] = {c};
return new String(arr);
```
which calls `ofCharArray()` which creates a `StringBuilder` and uses `append()` to insert the char and then returns `StringBuilder.toString()`. Any idea if this might have performance implications?

smowton (approved):
No further comments except what @romainbrenguier already raised

danpoe:
Ok, I'll merge this once CI passes. The behavior of the string operations does not change, but a new test has revealed some issues in the float to string conversion. I've marked the test as `KNOWNBUG` for now.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: ba122158a2b12df849dd09b1784cecfff08d8871).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/124413147

Issue #5002, opened 2019-08-09
Tracking pointer-members across bi-recursion is broken


xbauch:
```main.c
#include <assert.h>
#include <stdlib.h>

#ifndef NULL
#define NULL ((void *)0)
#endif

struct list;

typedef struct list list_nodet;

list_nodet fill_node(signed int depth_tag_list);

struct list {
  int datum; // comment out to make it work
  struct list *next;
};

int max_depth = 4;

list_nodet *build_node(int depth) {
  if (max_depth < depth)
    return ((list_nodet *)NULL);
  else {
    list_nodet *result = malloc(16ul);
    if (result) {
      // result->next = build_node(depth + 1);
      *result = fill_node(depth + 1); // inline the call to make it work
    }
    return result;
  }
}

list_nodet fill_node(int depth) {
  list_nodet result;
  result.next = build_node(depth);
  return result;
}

int main() {
  list_nodet *node = build_node(0);
  int i = 0;
  list_nodet *list_walker = node;
  for (; list_walker; list_walker = list_walker->next) {
    i = i + 1;
  }
  assert(i <= 5);
  return 0;
}

```

CBMC version: 5.12 (cbmc-5.12-d8598f8-758-g719e9a985)
Operating system: Ubuntu 18.04
Exact command line resulting in the issue: `cbmc main.c --unwind 11 --unwinding-assertions`
What behaviour did you expect: The unwinding assertion to succeed
What happened instead: The unwinding assertion fails

Note that:
1: removing the (unused) integer member `datum` fixes the issue
2: inlining the bi-recursive call fixes the issue

3: trace contains assignment `list_walker=((list_nodet *)NULL) + 4398046511104l` where the number is random.

PR #5050, opened 2019-08-22
Fix assigning to dynamically allocated structs

xbauch:
that is when
```
lhs :: [char]
rhs :: byte-extract-expr(struct, 0)
```
before we created a member expression at the offset and, in this case, assigned
to the first member of the struct. Now we assign the whole struct.

A test is included that was previously failing in the unwinding assertion (CBMC
did not see that the next pointer is eventually NULL). Which is issue #5002.

### DIFF
--- a/regression/cbmc/dynamic_struct/main.c
+++ b/regression/cbmc/dynamic_struct/main.c
@@ -1,0 +1,58 @@
+#include <assert.h>
+#include <stdlib.h>
+
+#ifndef NULL
+#  define NULL ((void *)0)
+#endif
+
+struct list;
+
+typedef struct list list_nodet;
+
+list_nodet fill_node(signed int depth_tag_list);
+
+struct list
+{
+  int datum;
+  struct list *next;
+};
+
+int max_depth = 4;
+
+list_nodet *build_node(int depth)
+{
+  if(max_depth < depth)
+    return ((list_nodet *)NULL);
+  else
+  {
+    // previous CBMC would work with this line:
+    // list_nodet *result = malloc(sizeof(struct list));
+    // but not this:
+    list_nodet *result = malloc(16);
+    if(result)
+    {
+      *result = fill_node(depth + 1);
+    }
+    return result;
+  }
+}
+
+list_nodet fill_node(int depth)
+{
+  list_nodet result;
+  result.next = build_node(depth);
+  return result;
+}
+
+int main()
+{
+  list_nodet *node = build_node(0);
+  int i = 0;
+  list_nodet *list_walker = node;
+  for(; list_walker; list_walker = list_walker->next)
+  {
+    i = i + 1;
+  }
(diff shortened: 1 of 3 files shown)


regression/cbmc/dynamic_struct/main.c
```diff
@@ -0,0 +1,55 @@
...
+
+int max_depth = 4;
+
+list_nodet *build_node(int depth)
+{
+  if(max_depth < depth)
+    return ((list_nodet *)NULL);
+  else
+  {
+    list_nodet *result = malloc(16);
```
hannes-steffenhagen-diffblue, line 31: this should be changed to `sizeof(struct list)` so it doesn't randomly break in the future

hannes-steffenhagen-diffblue (approved):
I don't quite get this, would prefer if someone more familiar with the code would sanity check this before merging

regression/cbmc/dynamic_struct/main.c
NlightNFotis: Agreed that this should be changed to something so that at least the value is self-documented.

NlightNFotis (approved):
LGTM

regression/cbmc/dynamic_struct/main.c
xbauch: The bug only appears if I don't use `sizeof`.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 0ed6adc34cff70937a074d73984ff45a0c855c8b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128901538

regression/cbmc/dynamic_struct/main.c
smowton: Because `malloc(sizeof(type))` is special-cased to produce a `type*` instead of a `char*`

src/pointer-analysis/value_set.cpp
```diff
@@ -936,6 +936,11 @@ void value_sett::get_value_set_rec(
     const typet &op_type = ns.follow(byte_extract_expr.op().type());
     if(op1_offset.has_value() && op_type.id() == ID_struct)
     {
+      if(*op1_offset == 0)
```
smowton, line 939: This is forcing the not-found case when the access offset is zero, regardless of the access size. I can see no reason why zero is special here.

smowton (commented):
The problem is clearly genuine, but the solution seems like a somewhat-arbitrary hack. Why is the access against the zeroth member incorrect? Where did the write end up recorded in the value-set, and why was that write not similarly directed into a struct field as the read code is trying to do?

tautschnig:
Closing in favour of #6107: I've re-used the regression test (modulo a slight simplification and cleanup), but provided an implementation that should be less arbitrary.

PR #5054, opened 2019-08-22
Statement List: Jump instructions

MatWise:
Includes support for the following STL instructions:
- JU (unconditional jump)
- JC (conditional jump)
- JCN (inverted conditional jump)

Since not all label destinations in STL are valid (this can depend on several conditions), it is needed to introduce data structures which track the properties of label and jump locations to the typecheck. See the documentation for details about the whole strategy.

Limitations: This implementation is only a rough concept and will likely be changed in the future. CFGs prove more useful for this task than the current concept, since they allow all parts of the PLC CPU to be modelled implicitly, without exceptions. Currently the state of the CPU needs to be saved as an intermediate result when encountering labels and jumps. So far this has **only** been realised for logic sequences. Jump instructions in blocks that depend on the accumulator will yield to false results.


MatWise:
Travis build fails because of known doxygen issue, works locally.

tautschnig (approved):

PR #5062, opened 2019-08-27
Allow user of run_doxygen script to specify path to doxygen

thk123:
Modify the `run_doxygen.sh` shell script to optionally take the path to doxygen as an argument. 

This change made it a little easier to test the broken doxygen problem so I'm suggesting it as an improvement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,11 +69,10 @@ jobs:
           then
             scripts/build_doxygen.sh ${DOXYGEN_VERSION}
           fi
-        - export PATH="$PATH:${TRAVIS_BUILD_DIR}/doxygen/build/bin"
       script:
         - echo $PATH
-        - doxygen --version
-        - scripts/run_doxygen.sh
+        - ${TRAVIS_BUILD_DIR}/doxygen/build/bin/doxygen --version
+        - scripts/run_doxygen.sh ${TRAVIS_BUILD_DIR}/doxygen/build/bin/doxygen
       before_cache:
       after_success:
         # Google Cloud Integration
--- a/scripts/run_doxygen.sh
+++ b/scripts/run_doxygen.sh
@@ -1,15 +1,19 @@
 #!/bin/bash
 
+doxygen_executable=${1:-doxygen}
+
 # Check doxygen version
 EXPECTED_VERSION="1.8.16"
-doxygen --version | grep ^$EXPECTED_VERSION > /dev/null
+$doxygen_executable --version | grep ^$EXPECTED_VERSION > /dev/null
 if [ $? -ne 0 ]
 then
   echo "WARNING: Using wrong version of doxygen.\
   The list of expected warnings is for version $EXPECTED_VERSION."
+  $doxygen_executable --version
+
 fi
 
 # Run doxygen and filter warnings
 SCRIPT_FOLDER=`dirname $0`
 cd $SCRIPT_FOLDER/../src
-doxygen 2>&1 | ../scripts/filter_expected_warnings.py ../scripts/expected_doxygen_warnings.txt
+$doxygen_executable 2>&1 | ../scripts/filter_expected_warnings.py ../scripts/expected_doxygen_warnings.txt


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: a17503376885afc456478e4b6e65c364d7b94fb6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/124779570

thk123:
Joel bot failure looks spurious.

scripts/run_doxygen.sh
```diff
@@ -1,15 +1,19 @@
...
+
 # Check doxygen version
 EXPECTED_VERSION="1.8.16"
-doxygen --version | grep ^$EXPECTED_VERSION > /dev/null
+$doxygen_executable --version | grep ^$EXPECTED_VERSION > /dev/null
 if [ $? -ne 0 ]
 then
   echo "WARNING: Using wrong version of doxygen.\
   The list of expected warnings is for version $EXPECTED_VERSION."
+  $doxygen_executable --version
```
owen-mc-diffblue, line 12: :pick: Ideally this would be in a separate commit

.travis.yml
```diff
@@ -69,11 +69,10 @@ jobs:
           then
             scripts/build_doxygen.sh ${DOXYGEN_VERSION}
           fi
-        - export PATH="$PATH:${TRAVIS_BUILD_DIR}/doxygen/build/bin"
       script:
         - echo $PATH
-        - doxygen --version
-        - scripts/run_doxygen.sh
+        - ${TRAVIS_BUILD_DIR}/doxygen/build/bin/doxygen --version
+        - scripts/run_doxygen.sh ${TRAVIS_BUILD_DIR}/doxygen/build/bin/doxygen
```
owen-mc-diffblue, line 75: As far as I understand (and my bash is pretty rusty), this commit isn't needed. Please explain why you want to make this change, preferably in the commit message.

chrisr-diffblue: I assume this commit is needed to make sure that the newly built version of doxygen is picked up, rather than a pre-installed version (which would appear sooner on the PATH). If that's the case, then an alternative solution would be to just re-write the `export PATH` as : `export PATH="${TRAVIS_BUILD_DIR}/doxygen/build/bin:${PATH}"`

chrisr-diffblue: I guess also that the added bonus here of explicitly specifying which doxygen binary is used by the run_doxygen.sh script, is that we make sure that script always works in that way for developers too.

scripts/run_doxygen.sh
```diff
@@ -1,15 +1,19 @@
 #!/bin/bash
 
+doxygen_executable=${1:-doxygen}
```
chrisr-diffblue, line 3: Pondering this a bit more - Should we drop this magic "fallback to whatever doxygen is on the path" stuff, and instead just *always* require the script to have the doxygen binary specified? I'm completely open either way, just wondering what the "least surprises" route is.

chrisr-diffblue (approved):

.travis.yml
thk123: Yes it isn't needed - this all works now. The intent was to:
 - make it clearer which version is being used
 - help developer know how they can run a specific version of it using the script

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a17503376885afc456478e4b6e65c364d7b94fb6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/124779570

scripts/run_doxygen.sh
owen-mc-diffblue: The fallback makes it easier to run locally. I don't think that it's surprising that the default is whatever `doxygen` resolves to.

peterschrammel (approved):

NlightNFotis:
This will need a bit of re-work to resolve conflicts and get it to pass the current CI configuration.

Closing this now - if you think this was a mistake, please reopen the PR - it will need a rebase a on develop and a drop of the last commit (as we are no longer using travis)

PR #5067, opened 2019-08-29
Natural loops: do not insist that loops only have one backedge

smowton:
I accidentally added this check when altering the natural loops data structures as it appeared
that each loop head was visited once -- in fact each *backedge* is visited once, and many such
edges may target the same loop header.

### DIFF
--- a/src/analyses/natural_loops.h
+++ b/src/analyses/natural_loops.h
@@ -241,7 +241,12 @@ void natural_loops_templatet<P, T>::compute_natural_loop(T m, T n)
   std::stack<T> stack;
 
   auto insert_result = loop_map.emplace(n, natural_loopt{*this});
-  INVARIANT(insert_result.second, "each loop head should only be visited once");
+  // Note the emplace *may* access a loop that already exists: this happens when
+  // a given header has more than one incoming edge, such as
+  // head: if(x) goto head; else goto head;
+  // In this case this compute routine is run twice, one for each backedge, with
+  // each adding whatever instructions can reach this 'm' (the program point
+  // that branches to the loop header, 'n').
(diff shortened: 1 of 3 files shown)


peterschrammel (approved):

src/analyses/natural_loops.h
```diff
@@ -241,7 +241,12 @@ void natural_loops_templatet<P, T>::compute_natural_loop(T m, T n)
   std::stack<T> stack;
 
   auto insert_result = loop_map.emplace(n, natural_loopt{*this});
-  INVARIANT(insert_result.second, "each loop head should only be visited once");
+  // Note the emplace *may* access a loop that already exists: this happens when
+  // a given header has more than one incoming edge, such as
+  // head: if(x) goto head; else goto head;
+  // In this case this compute routine is run twice, one for each backedge, with
+  // each adding whatever instructions can reach this 'm' (the program point
+  // that branches to the loop header, 'n').
```
danpoe, line 249: I suspect this also happens for loops with `continue` statements? That might be a more natural example than one involving gotos.

danpoe (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8ce372d1bf58a50230ff6a6f9791507107689878).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/125124986

src/analyses/natural_loops.h
smowton: Deliberately didn't do that because `goto_convert` is liable to turn `continue;` into a branch to the bottom (leading directly into the backedge), similarly `javac`

PR #5070, opened 2019-08-30
Set default max string length below existing max concrete string size

smowton:
Previously by default JBMC could guess at arbitrarily long strings but could not concretise
them (i.e. decision_proceduret::get() could not handle the ensuing very large arrays). This
meant JBMC would crash with --validate-trace (always enabled in regression testing), or
without it client code would encounter an unexpected nil_exprt.

After one too many difficult-to-reproduce regression test failures, I think it's finally time we set this by default...

### DIFF
--- a/src/solvers/strings/string_refinement.h
+++ b/src/solvers/strings/string_refinement.h
@@ -21,6 +21,7 @@ Author: Alberto Griggio, alberto.griggio@gmail.com
 #define CPROVER_SOLVERS_REFINEMENT_STRING_REFINEMENT_H
 
 #include <limits>
+#include <util/magic.h>
 #include <util/replace_expr.h>
 #include <util/string_expr.h>
 #include <util/union_find_replace.h>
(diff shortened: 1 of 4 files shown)


owen-mc-diffblue (approved):

src/solvers/strings/string_refinement.h
```diff
@@ -46,7 +47,9 @@ Author: Alberto Griggio, alberto.griggio@gmail.com
   " --string-input-value st      restrict non-null strings to a fixed value st;\n" /* NOLINT(*) */ \
   "                              the switch can be used multiple times to give\n" /* NOLINT(*) */ \
   "                              several strings\n" /* NOLINT(*) */ \
-  " --max-nondet-string-length n bound the length of nondet (e.g. input) strings\n" /* NOLINT(*) */
+  " --max-nondet-string-length n bound the length of nondet (e.g. input) strings\n" /* NOLINT(*) */ \
+  "   default is " + std::to_string(MAX_CONCRETE_STRING_SIZE - 1) + "; note that\n" /* NOLINT(*) */ \
```
romainbrenguier: indentation doesn't look correct

romainbrenguier (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: f06f5e5b7b3ce366cb74a806e70dd6ea486dedab).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/125260755

peterschrammel (approved):

PR #5072, opened 2019-08-30
(Calling) context sensitivity in the abstract interpreter

martin-cs:
This patch builds on #5036 and is intended as a demonstration of what the refactored infrastructure can do.  This adds optional context sensitivity to the abstract interpreter by just implementing a new history object.

### DIFF
--- a/regression/goto-analyzer/command_line_01/test.desc
+++ b/regression/goto-analyzer/command_line_01/test.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+--verify --recursive-interprocedural --ahistorical --constants --one-domain-per-history
+\[main.assertion.1\] line 13 assertion v != 0: FAILURE \(if reachable\)
+\[f00.assertion.1\] line 5 assertion x != 0: FAILURE \(if reachable\)
+Summary: 0 pass, 2 fail if reachable, 0 unknown
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+Tests the full AI configuration tuple AI/history/domain/storage
--- a/regression/goto-analyzer/command_line_02/main.c
+++ b/regression/goto-analyzer/command_line_02/main.c
@@ -1,0 +1,16 @@
+#include <assert.h>
+
+int f00(int x)
+{
+  assert(x != 0);
+  return 0;
+}
+
+int main(int argc, char **argv)
+{
+  int v = 0;
+  v = f00(v);
+  assert(v != 0);
+
+  return 0;
+}
--- a/regression/goto-analyzer/context_sensitivity_05/main.c
+++ b/regression/goto-analyzer/context_sensitivity_05/main.c
@@ -1,0 +1,54 @@
+#include <assert.h>
+
+void f00(int x, int y)
+{
+  if(x < 0)
+  {
+    // Unreachable in all traces if they are considered individually
+    assert(x < 0);
+    assert(1);
+    assert(0);
+  }
+
+  assert(x >= 0); // True in all traces
+  assert(x < 0);  // False in all traces
+  assert(x < 2);  // Split; true in some, false in others
+
+  assert((x <= 0) ? 1 : y);                  // True in some, unknown in others
+  assert((x <= 1) ? 0 : y);                  // False in some, unknown in others
+  assert((x <= 2) ? ((x <= 3) ? 1 : 0) : y); // A mix of all three
+
+  if(x < 5)
+  {
+    // Not reachable in all traces
+    assert((x <= 0) ? 1 : y); // True in some, unknown in others
+    assert((x <= 1) ? 0 : y); // False in some, unknown in others
+    assert((x <= 2) ? ((x <= 3) ? 1 : 0) : y); // A mix of all three
+  }
+
+  if(x < 3)
+  {
+    // Not reachable in all traces
+    assert((x <= 0) ? 1 : y); // True in some, unknown in others
+    assert((x <= 1) ? 0 : y); // False in some, unknown in others
+    assert((x <= 2) ? ((x <= 3) ? 1 : 0) : y); // A mix of all three
+  }
+}
+
+int nondet_int();
+
+int main(int argc, char **argv)
+{
+  int y = nondet_int();
+
+  // Because the history is context sensitive these should be analysed independently
+  // Just showing the domain will merge them giving top for everything interesting
+  f00(0, y);
+  f00(1, y);
+  f00(2, y);
+  f00(3, y);
+  f00(4, y);
+  f00(5, y);
+
+  return 0;
+}
--- a/src/analyses/ai_history.cpp
+++ b/src/analyses/ai_history.cpp
@@ -50,7 +50,40 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
     else
     {
       // An interprocedural (call) edge; add to the current stack
-      next_stack = cse_ptrt(new call_stack_entryt(to, current_stack));
+
+      // If the recursion limit has been reached
+      // shorten the stack / merge with the most recent invocation
+      // before we extend
+      cse_ptrt shorten = current_stack;
+
+      if(has_recursion_limit())
+      {
+        unsigned int number_of_recursive_calls = 0;
+        cse_ptrt first_found = nullptr; // The most recent recursive call
+
+        // Iterate back through the call stack
+        for(cse_ptrt i = current_stack->caller; i != nullptr; i = i->caller)
+        {
+          if(
+            i->current_location->location_number ==
+            current_stack->current_location->location_number)
+          {
+            // Found a recursive instance
+            if(first_found == nullptr)
+            {
+              first_found = i;
+            }
+            ++number_of_recursive_calls;
+            if(number_of_recursive_calls == recursion_limit)
+            {
+              shorten = first_found;
+              break;
+            }
+          }
+        }
+      }
+
+      next_stack = cse_ptrt(new call_stack_entryt(to, shorten));
     }
   }
   else if(current_stack->current_location->is_end_function())
@@ -88,7 +121,7 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
   INVARIANT(next_stack != nullptr, "All branches should initialise next_stack");
 
   // Create the potential next history
-  trace_ptrt next(new call_stack_historyt(next_stack));
+  trace_ptrt next(new call_stack_historyt(next_stack, recursion_limit));
 
   // If there is already an equivalent history, merge with that instead
   auto it = others.find(next);
--- a/src/analyses/ai_history.cpp
+++ b/src/analyses/ai_history.cpp
@@ -27,3 +27,169 @@ xmlt ai_history_baset::output_xml(void) const
   xml.data = out.str();
   return xml;
 }
+
+ai_history_baset::step_returnt
+call_stack_historyt::step(locationt to, const trace_sett &others) const
+{
+  DATA_INVARIANT(current_stack != nullptr, "current_stack must exist");
+
+  cse_ptrt next_stack = nullptr;
+
+  // First construct what the new history would be.
+  // This requires special handling if this edge is a function call or return
+  if(current_stack->current_location->is_function_call())
+  {
+    locationt l_return = std::next(current_stack->current_location);
+
+    if(l_return->location_number == to->location_number)
+    {
+      // Is skipping the function call, only update the location
+      next_stack =
+        cse_ptrt(new call_stack_entryt(l_return, current_stack->caller));
+    }
+    else
+    {
+      // An interprocedural (call) edge; add to the current stack
+      next_stack = cse_ptrt(new call_stack_entryt(to, current_stack));
+    }
+  }
+  else if(current_stack->current_location->is_end_function())
+  {
+    if(current_stack->caller == nullptr)
+    {
+      // Trying to return but there was no caller?
+      // Refuse to do the transition outright
+      return std::make_pair(ai_history_baset::step_statust::BLOCKED, nullptr);
+    }
+    else
+    {
+      // The expected call return site...
+      locationt l_caller_return =
+        std::next(current_stack->caller->current_location);
+
+      if(l_caller_return->location_number == to->location_number)
+      {
+        // ... which is where we are going
+        next_stack =
+          cse_ptrt(new call_stack_entryt(to, current_stack->caller->caller));
+      }
+      else
+      {
+        // Not possible to return to somewhere other than the call site
+        return std::make_pair(ai_history_baset::step_statust::BLOCKED, nullptr);
+      }
+    }
+  }
+  else
+  {
+    // Just update the location
+    next_stack = cse_ptrt(new call_stack_entryt(to, current_stack->caller));
+  }
+  INVARIANT(next_stack != nullptr, "All branches should initialise next_stack");
+
+  // Create the potential next history
+  trace_ptrt next(new call_stack_historyt(next_stack));
+
+  // If there is already an equivalent history, merge with that instead
+  auto it = others.find(next);
+
+  if(it == others.end())
+  {
+    return std::make_pair(ai_history_baset::step_statust::NEW, next);
+  }
+  else
+  {
+    return std::make_pair(ai_history_baset::step_statust::MERGED, *it);
+  }
+}
+
+bool call_stack_historyt::operator<(const ai_history_baset &op) const
+{
+  auto other = dynamic_cast<const call_stack_historyt *>(&op);
+  PRECONDITION(other != nullptr);
+
+  return *(this->current_stack) < *(other->current_stack);
+}
+
+bool call_stack_historyt::operator==(const ai_history_baset &op) const
+{
+  auto other = dynamic_cast<const call_stack_historyt *>(&op);
+  PRECONDITION(other != nullptr);
+
+  return *(this->current_stack) == *(other->current_stack);
+}
+
+void call_stack_historyt::output(std::ostream &out) const
+{
+  out << "call_stack_history : stack "
+      << current_stack->current_location->location_number;
+  cse_ptrt working = current_stack->caller;
+  while(working != nullptr)
+  {
+    out << " from " << working->current_location->location_number;
+    working = working->caller;
+  }
+  return;
+}
+
+bool call_stack_historyt::call_stack_entryt::
+operator<(const call_stack_entryt &op) const
+{
+  if(
+    this->current_location->location_number <
+    op.current_location->location_number)
+  {
+    return true;
+  }
+  else if(
+    this->current_location->location_number >
+    op.current_location->location_number)
+  {
+    return false;
+  }
+  else
+  {
+    INVARIANT(
+      this->current_location->location_number ==
+        op.current_location->location_number,
+      "Implied by if-then-else");
+
+    if(this->caller == op.caller)
+    {
+      return false; // Because they are equal
+    }
+    else if(this->caller == nullptr)
+    {
+      return true; // Shorter stacks are 'lower'
+    }
+    else if(op.caller == nullptr)
+    {
+      return false;
+    }
+    else
+    {
+      // Sort by the rest of the stack
+      return *(this->caller) < *(op.caller);
+    }
+  }
+  UNREACHABLE;
+}
+
+bool call_stack_historyt::call_stack_entryt::
+operator==(const call_stack_entryt &op) const
+{
+  if(
+    this->current_location->location_number ==
+    op.current_location->location_number)
+  {
+    if(this->caller == op.caller)
+    {
+      return true;
+    }
+    else if(this->caller != nullptr && op.caller != nullptr)
+    {
+      return *(this->caller) == *(op.caller);
+    }
+  }
+  return false;
+}
--- a/src/analyses/ai_history.cpp
+++ b/src/analyses/ai_history.cpp
@@ -44,8 +44,8 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
     if(l_return->location_number == to->location_number)
     {
       // Is skipping the function call, only update the location
-      next_stack =
-        cse_ptrt(new call_stack_entryt(l_return, current_stack->caller));
+      next_stack = cse_ptrt(
+        std::make_shared<call_stack_entryt>(l_return, current_stack->caller));
     }
     else
     {
@@ -83,7 +83,7 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
         }
       }
 
-      next_stack = cse_ptrt(new call_stack_entryt(to, shorten));
+      next_stack = cse_ptrt(std::make_shared<call_stack_entryt>(to, shorten));
     }
   }
   else if(current_stack->current_location->is_end_function())
@@ -103,8 +103,8 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
       if(l_caller_return->location_number == to->location_number)
       {
         // ... which is where we are going
-        next_stack =
-          cse_ptrt(new call_stack_entryt(to, current_stack->caller->caller));
+        next_stack = cse_ptrt(std::make_shared<call_stack_entryt>(
+          to, current_stack->caller->caller));
       }
       else
       {
@@ -116,12 +116,15 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
   else
   {
     // Just update the location
-    next_stack = cse_ptrt(new call_stack_entryt(to, current_stack->caller));
+    next_stack =
+      cse_ptrt(std::make_shared<call_stack_entryt>(to, current_stack->caller));
   }
   INVARIANT(next_stack != nullptr, "All branches should initialise next_stack");
 
   // Create the potential next history
   trace_ptrt next(new call_stack_historyt(next_stack, recursion_limit));
+  // It would be nice to use make_shared here but ... that doesn't work with
+  // protected constructors
 
   // If there is already an equivalent history, merge with that instead
   auto it = others.find(next);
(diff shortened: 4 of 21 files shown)


martin-cs:
I guess there probably isn't too much point reviewing this in detail until #5036 is merged.  I will rebase after that.  Certainly only look at the commits that are unique to this PR.

xbauch:
In case of reviewing: the first new commit is https://github.com/diffblue/cbmc/pull/5072/commits/76b7c1b3fc8391842ccf71acce797d6934036042.

src/goto-analyzer/goto_analyzer_parse_options.cpp
```diff
@@ -737,15 +829,27 @@ void goto_analyzer_parse_optionst::help()
     "\n"
     "Abstract interpreter options:\n"
     // NOLINTNEXTLINE(whitespace/line_length)
-    " --location-sensitive         use location-sensitive abstract interpreter\n"
-    " --concurrent                 use concurrency-aware abstract interpreter\n"
+    " --recursive-interprocedural  use recursion to handle interprocedural reasoning\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    " --legacy-ait                 recursion for function and one domain per location\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    " --legacy-concurrent          legacy-ait with a extended fixed-point for concurrency\n"
```
xbauch: _an_ extended fixed-point

regression/goto-analyzer/command_line_01/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--verify --recursive-interprocedural --ahistorical --constants --one-domain-per-history
+\[main.assertion.1\] line 13 assertion v != 0: FAILURE \(if reachable\)
+\[f00.assertion.1\] line 5 assertion x != 0: FAILURE \(if reachable\)
+Summary: 0 pass, 2 fail if reachable, 0 unknown
+^EXIT=0$
```
xbauch, line 7: Just a question: `goto-analyzer` returns `0` even when there are failing assertions?

src/analyses/ai_history.h
```diff
@@ -210,6 +210,67 @@ class ahistoricalt : public ai_history_baset
...
+
+  cse_ptrt current_stack;
+  // DATA_INVARIANT(current_stack != nullptr, "current_stack must exist");
+  // DATA_INVARIANT(current_stack->current.is_dereferenceable(),
+  //                "Must not be _::end()")
+
+  explicit call_stack_historyt(cse_ptrt p)
+    : ai_history_baset(p->current_location), current_stack(p)
+  {
+    PRECONDITION(p != nullptr); // A little late by now but worth documenting
```
xbauch: This could almost be a comment.

src/analyses/ai_history.cpp
```diff
@@ -27,3 +27,169 @@ xmlt ai_history_baset::output_xml(void) const
...
+      // An interprocedural (call) edge; add to the current stack
+      next_stack = cse_ptrt(new call_stack_entryt(to, current_stack));
+    }
+  }
+  else if(current_stack->current_location->is_end_function())
+  {
+    if(current_stack->caller == nullptr)
+    {
+      // Trying to return but there was no caller?
+      // Refuse to do the transition outright
```
xbauch, line 94: Why not assert it then?

regression/goto-analyzer/context_sensitivity_05/main.c
```diff
@@ -0,0 +1,54 @@
...
+    assert((x <= 1) ? 0 : y); // False in some, unknown in others
+    assert((x <= 2) ? ((x <= 3) ? 1 : 0) : y); // A mix of all three
+  }
+
+  if(x < 3)
+  {
+    // Not reachable in all traces
+    assert((x <= 0) ? 1 : y); // True in some, unknown in others
+    assert((x <= 1) ? 0 : y); // False in some, unknown in others
+    assert((x <= 2) ? ((x <= 3) ? 1 : 0) : y); // A mix of all three
```
xbauch, line 34: I think the comment here should say success.

xbauch (approved):
Looks good. Could we have a test showing the difference between `domain-per-location/history`?

src/goto-analyzer/goto_analyzer_parse_options.cpp
```diff
@@ -252,15 +251,50 @@ void goto_analyzer_parse_optionst::get_command_line_options(optionst &options)
...
-      options.set_option("concurrent", true);
+    if(cmdline.isset("recursive-interprocedural"))
+      options.set_option("recursive-interprocedural", true);
+    else if(cmdline.isset("legacy-ait") || cmdline.isset("location-sensitive"))
+    {
+      options.set_option("legacy-ait", true);
+      // Fixes a number of other options as well
+      options.set_option("ahistorical", true);
+      options.set_option("history set", true);
+      options.set_option("one-domain-per-location", true);
```
hannes-steffenhagen-diffblue, line 262: with Peter's comments on #5036 maybe this should be called `one-domain-object-per-location` or something of the sort?

src/goto-analyzer/goto_analyzer_parse_options.cpp
```diff
@@ -737,15 +829,27 @@ void goto_analyzer_parse_optionst::help()
...
-    " --concurrent                 use concurrency-aware abstract interpreter\n"
+    " --recursive-interprocedural  use recursion to handle interprocedural reasoning\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    " --legacy-ait                 recursion for function and one domain per location\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    " --legacy-concurrent          legacy-ait with a extended fixed-point for concurrency\n"
+    "\n"
+    "History options:\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    " --ahistorical                the most basic history, tracks locations only\n"
```
hannes-steffenhagen-diffblue, line 852: I'd expect "ahistorical" to mean nothing is tracked, but it does track something?

src/analyses/ai_history.h
hannes-steffenhagen-diffblue: This is just one of the issues with not having proper contracts support in C++ :(

src/analyses/ai_history.cpp
```diff
@@ -27,3 +27,169 @@ xmlt ai_history_baset::output_xml(void) const
...
+  // This requires special handling if this edge is a function call or return
+  if(current_stack->current_location->is_function_call())
+  {
+    locationt l_return = std::next(current_stack->current_location);
+
+    if(l_return->location_number == to->location_number)
+    {
+      // Is skipping the function call, only update the location
+      next_stack =
+        cse_ptrt(new call_stack_entryt(l_return, current_stack->caller));
```
hannes-steffenhagen-diffblue: `make_shared`

src/analyses/ai_history.cpp
```diff
@@ -27,3 +27,169 @@ xmlt ai_history_baset::output_xml(void) const
...
+    }
+    else
+    {
+      // An interprocedural (call) edge; add to the current stack
+      next_stack = cse_ptrt(new call_stack_entryt(to, current_stack));
```
hannes-steffenhagen-diffblue: `make_shared`

src/analyses/ai_history.cpp
```diff
@@ -27,3 +27,169 @@ xmlt ai_history_baset::output_xml(void) const
...
+
+      if(l_caller_return->location_number == to->location_number)
+      {
+        // ... which is where we are going
+        next_stack =
+          cse_ptrt(new call_stack_entryt(to, current_stack->caller->caller));
+      }
+      else
+      {
+        // Not possible to return to somewhere other than the call site
```
hannes-steffenhagen-diffblue, line 111: Is this not an error then?

src/analyses/ai_history.cpp
```diff
@@ -27,3 +27,169 @@ xmlt ai_history_baset::output_xml(void) const
...
+        return std::make_pair(ai_history_baset::step_statust::BLOCKED, nullptr);
+      }
+    }
+  }
+  else
+  {
+    // Just update the location
+    next_stack = cse_ptrt(new call_stack_entryt(to, current_stack->caller));
```
hannes-steffenhagen-diffblue: `make_shared`

src/analyses/ai_history.cpp
```diff
@@ -27,3 +27,169 @@ xmlt ai_history_baset::output_xml(void) const
...
+  }
+  INVARIANT(next_stack != nullptr, "All branches should initialise next_stack");
+
+  // Create the potential next history
+  trace_ptrt next(new call_stack_historyt(next_stack));
```
hannes-steffenhagen-diffblue: `make_shared`

src/analyses/ai_history.cpp
```diff
@@ -50,7 +50,40 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
...
     {
       // An interprocedural (call) edge; add to the current stack
-      next_stack = cse_ptrt(new call_stack_entryt(to, current_stack));
+
+      // If the recursion limit has been reached
+      // shorten the stack / merge with the most recent invocation
+      // before we extend
+      cse_ptrt shorten = current_stack;
+
+      if(recursion_limit > 0)
```
hannes-steffenhagen-diffblue: I'm guessing `0` here means "no limit"? I'd prefer if that was more obvious in the code, either by making `recursion_limit` optional or having a small helper like `has_recursion_limit()`

src/analyses/ai_history.cpp
```diff
@@ -50,7 +50,40 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
...
+            if(found == recursion_limit)
+            {
+              shorten = first_found;
+              break;
+            }
+          }
+        }
+      }
+
+      next_stack = cse_ptrt(new call_stack_entryt(to, shorten));
```
hannes-steffenhagen-diffblue: `make_shared`

src/analyses/ai_history.cpp
```diff
@@ -50,7 +50,40 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
...
-      next_stack = cse_ptrt(new call_stack_entryt(to, current_stack));
+
+      // If the recursion limit has been reached
+      // shorten the stack / merge with the most recent invocation
+      // before we extend
+      cse_ptrt shorten = current_stack;
+
+      if(recursion_limit > 0)
+      {
+        unsigned int found = 0;
```
hannes-steffenhagen-diffblue: Something like `number_of_recursive_calls`?

src/analyses/ai_history.h
```diff
@@ -237,20 +237,37 @@ class call_stack_historyt : public ai_history_baset
   // DATA_INVARIANT(current_stack->current.is_dereferenceable(),
   //                "Must not be _::end()")
 
-  explicit call_stack_historyt(cse_ptrt p)
-    : ai_history_baset(p->current_location), current_stack(p)
+  // At what point to merge with a previous call stack when handling recursion
+  // Setting to 0 disables completely
+  unsigned int recursion_limit;
+
+  call_stack_historyt(cse_ptrt p, unsigned int rl)
```
hannes-steffenhagen-diffblue: `current_stack`, `recursion_limit`

src/analyses/ai_history.cpp
```diff
@@ -88,7 +121,7 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
   INVARIANT(next_stack != nullptr, "All branches should initialise next_stack");
 
   // Create the potential next history
-  trace_ptrt next(new call_stack_historyt(next_stack));
+  trace_ptrt next(new call_stack_historyt(next_stack, recursion_limit));
```
hannes-steffenhagen-diffblue, line 125: `make_shared`

src/analyses/ai_history.h
```diff
@@ -299,4 +316,26 @@ class ai_history_factory_default_constructort : public ai_history_factory_baset
...
+  unsigned int recursion_limit;
+
+public:
+  explicit call_stack_history_factoryt(unsigned int rl) : recursion_limit(rl)
+  {
+  }
+
+  ai_history_baset::trace_ptrt bang(ai_history_baset::locationt l) override
+  {
+    ai_history_baset::trace_ptrt p(new call_stack_historyt(l, recursion_limit));
```
hannes-steffenhagen-diffblue: `make_shared`

src/analyses/ai_history.h
```diff
@@ -237,20 +237,37 @@ class call_stack_historyt : public ai_history_baset
...
+      recursion_limit(rl)
   {
     PRECONDITION(p != nullptr); // A little late by now but worth documenting
   }
 
 public:
   explicit call_stack_historyt(locationt l)
-    : ai_history_baset(l), current_stack(new call_stack_entryt(l, nullptr))
+    : ai_history_baset(l),
+      current_stack(new call_stack_entryt(l, nullptr)),
```
hannes-steffenhagen-diffblue: `make_shared`

src/goto-analyzer/static_verifier.cpp
```diff
@@ -307,7 +307,12 @@ bool static_verifier(
       results.push_back(static_verifier_resultt());
       auto &result = results.back();
 
-      if(e.is_true())
+      if(domain.is_bottom())
```
hannes-steffenhagen-diffblue: I agree this seems to be an improvement, but it appears unrelated to the rest of the PR?

src/goto-analyzer/static_verifier.cpp
```diff
@@ -262,6 +262,31 @@ static void static_verifier_console(
...
+  else if(e.is_false())
+  {
+    return static_verifier_resultt::FALSE;
+  }
+  else
+  {
+    return static_verifier_resultt::UNKNOWN;
+  }
+
+  UNREACHABLE;
```
hannes-steffenhagen-diffblue, line 287: I believe the `UNREACHABLE` is unnecessary as this place is just syntactically unreachable right now

src/goto-analyzer/static_verifier.cpp
```diff
@@ -300,32 +325,127 @@ bool static_verifier(
...
+              // All traces that (may) reach it are false
+              ++fail;
+              result.status = static_verifier_resultt::FALSE;
+            }
+            else
+            {
+              // The awkward case, there are traces that (may) reach it and
+              // some are true, some are false.  It is not entirely fair to say
+              // "FAILURE (if reachable)" because it's a bit more complex than
+              // that, "FAILURE (if reachable via a particular trace)" would be
```
hannes-steffenhagen-diffblue, line 441: So normally `FALSE` just meant "definitely fails and is definitely reachable" IIRC? Isn't this breaking convention?

regression/goto-analyzer/context_sensitivity_05/main.c
hannes-steffenhagen-diffblue: In context this would always be true, but I believe this is about the semantics of `--constants` where it can also be unknown - however, this definitely can't be `false`.

hannes-steffenhagen-diffblue (commented):
Looks good to me generally, just some questions/nitpicks

regression/goto-analyzer/command_line_01/test.desc
martin-cs: Yeah... At the moment it does because the analysis has run successfully / it hasn't shown the presence of bugs.  I can see an argument for it returning 10 but I think that is beyond the scope of the current PR.

src/analyses/ai_history.h
martin-cs: I have seen contract languages for C and C++ and they get kinda scary kinda quickly.

src/analyses/ai_history.cpp
martin-cs: I did but there is a legitimate case when this happens -- the end of the initial calling function.  This is the code that finally finishes the analysis.

src/goto-analyzer/goto_analyzer_parse_options.cpp
martin-cs: The object knows it's current location but nothing more; it knows nothing of it's history.  I mean, ahistorical is kinda a bad name but I couldn't come up with anything more exact.

src/analyses/ai_history.cpp
martin-cs: It's not an error.  The `ait` dispatch loop doesn't know which transitions are possible so it simply tries all of them (in this case, trying all transitions from *any* history that reaches the end of the callee function back to the call site) and relies on the history to block out any of these that the history can rule impossible.

This is kind of subtle and the first time this feature has been used; that's sort of why I wanted to write this patch to demonstrate this.

src/goto-analyzer/static_verifier.cpp
martin-cs: Sort of.  The next commit refactors the same bit of code so I wanted to separate the 'improvement that changes the regression tests a bit' from the 'refactoring that adds new capabilities'.

src/goto-analyzer/static_verifier.cpp
martin-cs: It is right now, I tend to use them defensively so I can't later break the code by removing an option / adding a value to the enum without finding out.

src/goto-analyzer/static_verifier.cpp
martin-cs: It is not really possible or meaningful to follow BMC style conventions in an abstract interpreter.  In BMC we are always dealing with the possibility that there are executions that we haven't explored, so "TRUE / SAFE" comes with an implicit "(in all the traces we have explored)" but on the plus side "FALSE / UNSAFE" is a definite thing.

In the AI case, "TRUE / SAFE" really is in *all* traces but "FALSE / UNSAFE (if reachable)" is the best we can manage.

I'm not sure I see an easy way of resolving this; it is certainly more fundamental than this PR.

martin-cs:
My thanks to @xbauch and @hannes-steffenhagen-diffblue for their comments.  I have responded to all of the 'discussion' ones.  I may leave the 'changes required' ones until the rebase once the main PR is merged.

hannes-steffenhagen-diffblue (approved):
Going over my comments and your responses, unless there are substantial changes after the rebase this'll probably not require a re-review from me. (I think the `make_shared`s are needed for consistency and _some_ correctness though)

regression/goto-analyzer/command_line_02/main.c
```diff
@@ -0,0 +1,16 @@
+#include <assert.h>
```
chrisr-diffblue, line 1: Since this test file appears to be identical to the one in command_line_01, why not just have a single file, and use two `.desc` files in the one directory?

src/analyses/ai_history.cpp
chrisr-diffblue: Probably out of scope of this PR, but how would all this work with exceptions or something like setjmp/longjmp ?

src/analyses/ai_history.h
```diff
@@ -210,6 +210,67 @@ class ahistoricalt : public ai_history_baset
...
+  bool operator==(const ai_history_baset &op) const override;
+
+  const locationt &current_location(void) const override
+  {
+    return current_stack->current_location;
+  }
+
+  // Use default widening
+  // Typically this would be used for loops, which are not tracked
+  // it would be possible to use this to improve the handling of recursion
```
chrisr-diffblue, line 301: Rather than just a comment saying that the default implementation is used, I still prefer to explicitly define the function and invoke the parent.

chrisr-diffblue (approved):
Some minor queries and nit picks, but nothing blocking.

src/analyses/ai_history.cpp
martin-cs: At the moment, it doesn't.  `ait` will through an exception if it finds try `THROW` or `CATCH` instructions and some domains will also fail in transform if they meet them.  So the following is hypothetical but...

(-2. Find a use case for analysing setjmp/longjmp. )
(-1. Get setjmp/longjmp working with the rest of the code. )
0. Actually get some agreement and documentation of the semantics of THROW/CATCH.
1. Inherit from `ait` and create an abstract interpreter that doesn't use recursion for interprocedural stuff.  This will need to track call-sites for functions.
2. Add tracking of catch locations and override `visit` to handle throw and catch.  Throw will then attempt to step edges to all possible catch locations.
3. Modify this, or another history so that it blocks steps to implausible exception handlers (i.e. not in the call stack) as given above.
4. Modify domains as needed so they can compute the edge.

Thinking about this ... actually the part from 1 to 4 is probably a single PR's worth.  I suspect the hard work is up to and including 0.

martin-cs:
TODO : Triggering the recursion limit should result in widening returning true.

martin-cs:
@chrisr-diffblue : exceptions and `setjmp`/`longjmp`, possible but beyond the scope of this PR, way beyond.

Somewhere I have a written up plan for how to do exceptions (assuming anyone can document the semantics of CPROVER exceptions) but the short version is to compute an "edge" from every throw to every (previously reached) catch in `ai_baset`.  The `step` function in the history allows these to be pruned down to only those that are historically feasible -- that would work really well for this calling context history.

`setjmp`/`longjmp` ... is an interesting challenge ... that may have to wait until someone has some real code using it that they want to analyse.  In principle you would think that you could associate the `jmp_buf` object with a set of histories and then on `longjmp` you could compute "edges" to each of these and again, let `step` prune them.  BUT that wouldn't be enough as you would also have to overwrite the globals and heap variables in the saved domains with those from the caller and there isn't currently a good interface for doing that and it might require changes to the way domains are handled... yeah... not a starter task...

src/goto-analyzer/goto_analyzer_parse_options.cpp
martin-cs: Fixed.

regression/goto-analyzer/context_sensitivity_05/main.c
martin-cs: I think they should say true and false because the comments are about the truth of the propositions.  To me success and failure are about the tool being able to / not being able to prove something.

src/goto-analyzer/goto_analyzer_parse_options.cpp
martin-cs: See https://github.com/diffblue/cbmc/issues/5197 for discussion of this kind of thing.

src/analyses/ai_history.cpp
martin-cs: Fixed.

src/analyses/ai_history.cpp
martin-cs: Fixed.

src/analyses/ai_history.cpp
martin-cs: Fixed.

src/analyses/ai_history.cpp
martin-cs: Fixed.

src/analyses/ai_history.cpp
martin-cs: Fixed.

src/analyses/ai_history.cpp
martin-cs: Fixed.

src/analyses/ai_history.h
martin-cs: Fixed.

src/analyses/ai_history.h
martin-cs: Fixed.

src/analyses/ai_history.cpp
martin-cs: Or... potentially not fixed.  Turns out that `std::make_shared` doesn't work with protected and private constructors.  All of the work-arounds are more horrible than leaving this one instance:

https://stackoverflow.com/questions/54454620/why-cant-i-access-the-private-constructor-in-my-static-function-as-below/54456533#54456533
https://stackoverflow.com/questions/3541632/using-make-shared-with-a-protected-constructor-abstract-interface

( Yes, `friend` should work, no, it doesn't. )

I've had a chat with @hannes-steffenhagen-diffblue about this and he says that leaving the new is the least worst option.

src/analyses/ai_history.cpp
martin-cs: Done.

src/analyses/ai_history.cpp
martin-cs: Done.

src/analyses/ai_history.h
martin-cs: Renamed to `cur_stack` and `rec_lim` so the naming doesn't clash with member variables.

regression/goto-analyzer/command_line_02/main.c
martin-cs: Fair point, will do in future.  Don't particularly want to change this right now as a hassle getting this back into the right commit.

src/analyses/ai_history.h
martin-cs: Done!

martin-cs:
@xbauch , @hannes-steffenhagen-diffblue and @chrisr-diffblue : I have made the requested changes and have pushed a rebased version ( post merging #5036 ).  If^WWhen this passes CI I will squash back as many of the fix commits as possible and then merge.  Shout if you want something else done!

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: ae0d7ade97d9437d0d88ae61738257751fba7809).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/138337121

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fddd0987118614033edb7c90ccacb8c4b5df1ebb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/138404904

PR #5073, opened 2019-09-02
[TG-9299] Fix crash in case of user provided clinit for a clinit which is not in the symbol table

romainbrenguier:
The look-up for clinit could fail when we had a user provided clinit but the real clinit was not in the symbol table.
This adds a test which illustrate the case that was previously causing the exception.

### DIFF
--- a/jbmc/src/java_bytecode/assignments_from_json.cpp
+++ b/jbmc/src/java_bytecode/assignments_from_json.cpp
@@ -838,7 +838,9 @@ void assign_from_json(
   location.set_function(function_id);
   allocate_objectst allocate(ID_java, location, function_id, symbol_table);
   code_blockt body_rec;
-  const auto class_name = declaring_class(symbol_table.lookup_ref(function_id));
+  const symbolt *function_symbol = symbol_table.lookup(function_id);
+  INVARIANT(function_symbol, "Function must appear in symbol table");
+  const auto class_name = declaring_class(*function_symbol);
   INVARIANT(
     class_name,
     "Function " + id2string(function_id) + " must be declared by a class.");
--- a/jbmc/src/java_bytecode/java_static_initializers.cpp
+++ b/jbmc/src/java_bytecode/java_static_initializers.cpp
@@ -792,6 +792,14 @@ code_blockt get_user_specified_clinit_body(
     &class_to_declared_symbols_map)
 {
   const irep_idt &real_clinit_name = clinit_function_name(class_id);
+  const auto clinit_func = symbol_table.lookup(real_clinit_name);
+  if(clinit_func == nullptr)
+  {
+    // Case where the real clinit doesn't appear in the symbol table, even
+    // though their is user specifed one. This may occur when some class
+    // substitution happened after compilation.
+    return code_blockt{};
+  }
   const auto class_entry =
     static_values_json.find(id2string(strip_java_namespace_prefix(class_id)));
   if(class_entry != static_values_json.end())
@@ -833,9 +841,7 @@ code_blockt get_user_specified_clinit_body(
       return body;
     }
   }
-  if(const auto clinit_func = symbol_table.lookup(real_clinit_name))
-    return code_blockt{{code_function_callt{clinit_func->symbol_expr()}}};
-  return code_blockt{};
+  return code_blockt{{code_function_callt{clinit_func->symbol_expr()}}};
 }
 
 /// Create static initializer wrappers and possibly user-specified functions for
--- a/jbmc/src/java_bytecode/java_static_initializers.cpp
+++ b/jbmc/src/java_bytecode/java_static_initializers.cpp
@@ -802,44 +802,40 @@ code_blockt get_user_specified_clinit_body(
   }
   const auto class_entry =
     static_values_json.find(id2string(strip_java_namespace_prefix(class_id)));
-  if(class_entry != static_values_json.end())
+  if(class_entry != static_values_json.end() && class_entry->second.is_object())
   {
-    const auto &class_json_value = class_entry->second;
-    if(class_json_value.is_object())
+    const auto &class_json_object = to_json_object(class_entry->second);
+    std::map<symbol_exprt, jsont> static_field_values;
+    for(const auto &symbol_pair :
+        equal_range(class_to_declared_symbols_map, class_id))
     {
-      const auto &class_json_object = to_json_object(class_json_value);
-      std::map<symbol_exprt, jsont> static_field_values;
-      for(const auto &class_symbol_pair :
-          equal_range(class_to_declared_symbols_map, class_id))
+      const symbolt &symbol = symbol_pair.second;
+      if(symbol.is_static_lifetime)
       {
-        const symbolt &symbol = class_symbol_pair.second;
-        if(symbol.is_static_lifetime)
+        const symbol_exprt &static_field_expr = symbol.symbol_expr();
+        const auto &static_field_entry =
+          class_json_object.find(id2string(symbol.base_name));
+        if(static_field_entry != class_json_object.end())
         {
-          const symbol_exprt &static_field_expr = symbol.symbol_expr();
-          const auto &static_field_entry =
-            class_json_object.find(id2string(symbol.base_name));
-          if(static_field_entry != class_json_object.end())
-          {
-            static_field_values.insert(
-              {static_field_expr, static_field_entry->second});
-          }
+          static_field_values.insert(
+            {static_field_expr, static_field_entry->second});
         }
       }
-      code_blockt body;
-      for(const auto &value_pair : static_field_values)
-      {
-        assign_from_json(
-          value_pair.first,
-          value_pair.second,
-          real_clinit_name,
-          body,
-          symbol_table,
-          needed_lazy_methods,
-          max_user_array_length,
-          references);
-      }
-      return body;
     }
+    code_blockt body;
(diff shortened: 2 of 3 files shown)


jbmc/regression/jbmc/deterministic_assignments_json/test_empty_non_empty.desc
```diff
@@ -0,0 +1,12 @@
...
+EmptyNonEmpty.class
+--unwind 1 --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar` --static-values clinit-state.json --function EmptyNonEmpty.test
+^VERIFICATION SUCCESSFUL$
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+This tests what happens when the non-empty class is substitued with an empty
+one after compilation: After compiling EmptyNonEmpty.class and Empty.class,
+Empty.class is moved to NonEmpty.class.
```
peterschrammel: What does 'moved' mean?

jbmc/regression/jbmc/deterministic_assignments_json/EmptyNonEmpty.java
```diff
@@ -0,0 +1,9 @@
+public class EmptyNonEmpty {
+
+  /// This test is meant to illustrate what happens when the non-empty
+  /// class is substitued with an empty one after compilation.
```
peterschrammel: substituted

jbmc/regression/jbmc/deterministic_assignments_json/test_empty_non_empty.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+EmptyNonEmpty.class
+--unwind 1 --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar` --static-values clinit-state.json --function EmptyNonEmpty.test
+^VERIFICATION SUCCESSFUL$
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+This tests what happens when the non-empty class is substitued with an empty
```
peterschrammel: substituted

jbmc/regression/jbmc/deterministic_assignments_json/Empty.java
```diff
@@ -0,0 +1,3 @@
+public class Empty {
```
smowton: Rather that emptiness, is having or not having static fields the critical attribute?

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -838,7 +838,9 @@ void assign_from_json(
   location.set_function(function_id);
   allocate_objectst allocate(ID_java, location, function_id, symbol_table);
   code_blockt body_rec;
-  const auto class_name = declaring_class(symbol_table.lookup_ref(function_id));
+  const symbolt *function_symbol = symbol_table.lookup(function_id);
```
smowton, line 841: is lookup-ref vs. lookup-then-invariant really that different?

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -808,13 +809,15 @@ code_blockt get_user_specified_clinit_body(
             }
           }
         }
+        if(symbol_table.lookup(real_clinit_name) == nullptr)
+          return code_blockt{};
```
smowton: Add a comment here about how/why this could happen

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -777,6 +777,7 @@ code_blockt get_user_specified_clinit_body(
   std::unordered_map<std::string, object_creation_referencet> &references)
 {
   jsont json;
+  const irep_idt real_clinit_name = clinit_function_name(class_id);
```
peterschrammel: (commit message) careful

jbmc/src/java_bytecode/assignments_from_json.cpp
peterschrammel: yes, you get a reasonable message on failure.

jbmc/regression/jbmc/deterministic_assignments_json/Empty.java
romainbrenguier: Yes, I've clarified that in the class names

jbmc/src/java_bytecode/java_static_initializers.cpp
romainbrenguier: done

jbmc/regression/jbmc/deterministic_assignments_json/test_empty_non_empty.desc
romainbrenguier: I've clarified that

jbmc/regression/jbmc/deterministic_assignments_json/EmptyNonEmpty.java
```diff
@@ -0,0 +1,9 @@
+public class EmptyNonEmpty {
```
smowton: Change this name too

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 81e1c9b1b6ca20f7d32da67410d135f3d72530cc).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/125753721

jbmc/src/java_bytecode/assignments_from_json.cpp
```diff
@@ -838,7 +838,9 @@ void assign_from_json(
...
+  INVARIANT(function_symbol, "Function must appear in symbol table");
```
antlechner, line 842: :bulb: This message is quite general, maybe include `function_id` in it and/or mention the context in which this was looked up.

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -817,6 +817,13 @@ code_blockt get_user_specified_clinit_body(
           }
         }
       }
+      if(symbol_table.lookup(real_clinit_name) == nullptr)
+      {
+        // Case where there is a user provided clinit, but the real clinit
+        // doesn't appear in the symbol table. This may occur when some class
```
antlechner: :question: Do you have an example of such a class substitution? Having these additional checks in is good of course, but I don't understand how this ever happened in practice. `get_user_specified_clinit_body` is only called to populate the body of a `synthetic_method_typet::USER_SPECIFIED_STATIC_INITIALIZER` if we see one. Such a synthetic method is only created in `create_user_specified_clinit_function_symbol`, which in turn is only called after several `if`s, one of which is `symbol_table.has_symbol(clinit_function_name(class_name))`.
So it looks like the only way we could end up with a `user_specified_clinit` but no `clinit` would be if `clinit` was first added to the symbol table, then `user_specified_clinit` was added, and finally `clinit` was removed - but I'm not aware of the latter ever happening?

jbmc/src/java_bytecode/java_static_initializers.cpp
```diff
@@ -817,6 +817,13 @@ code_blockt get_user_specified_clinit_body(
           }
         }
       }
+      if(symbol_table.lookup(real_clinit_name) == nullptr)
```
antlechner: :bulb: Looks like it would make more sense to move this check just below the definition of `real_clinit_name`.
Edit: Ah, this is done in the next commit. Those two commits should probably be squashed.

antlechner (approved):
Is there any way to add a regression test for this? See my question below: I can't see how the bug that this is fixing would occur in practice.

jbmc/src/java_bytecode/java_static_initializers.cpp
romainbrenguier: @antlechner 
`create_user_specified_clinit_function_symbol` is called if `needs_clinit_wrapper` is true, which is when `clinit` exists *or* some other condition. If you think this is wrong then `needs_clinit_wrapper` needs to be corrected.

jbmc/src/java_bytecode/assignments_from_json.cpp
romainbrenguier: I don't know, the backtrace of the invariant should already give us the context.

jbmc/src/java_bytecode/java_static_initializers.cpp
antlechner: Ah, that's the part I overlooked! `needs_clinit_wrapper` is true if `clinit` exists for this class _or_ a superclass.
So it looks like the mistake is that currently, we create a `user_specified_clinit` iff we create a `clinit_wrapper`. But really we should be creating a `user_specified_clinit` iff there is a `clinit`.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1b3c289e208651bdf5bcb74b8dd29299149e8b0e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126241524

jbmc/src/java_bytecode/java_static_initializers.cpp
romainbrenguier: ok, I'm still going to merge this PR since it prevents a crash but the condition for calling `create_user_specified_clinit_function_symbol` should be fixed. Do you want to create an issue for that?

PR #5082, opened 2019-09-03
Add pass to (try to) ensure a single backedge per loop header

smowton:
Actually to reduce costs this redirects *any* (lexical) backedge to enforce the rule that each loop header (or block otherwise branched back to, including acyclic code) has a single incoming edge from below.

This pass will be useful for regularising Java `while` and `do-while` loops, which can have `continue` edges directed to the top of the loop. At present goto-symex executes these loops more times than the `--unwind` directive would suggest because it regards the multiple distinct backedges as different loops.

### DIFF
--- a/src/goto-programs/ensure_one_backedge_per_target.cpp
+++ b/src/goto-programs/ensure_one_backedge_per_target.cpp
@@ -1,0 +1,122 @@
+/*******************************************************************\
+
+Module: Ensure one backedge per target
+
+Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+/// \file
+/// Ensure one backedge per target
+
+#include "ensure_one_backedge_per_target.h"
+
+static bool location_number_less_than(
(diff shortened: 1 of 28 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 82f239c7fd7123c77b227c99c65c993196a7e2d1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/125758859

src/goto-programs/ensure_one_backedge_per_target.cpp
```diff
@@ -0,0 +1,122 @@
+/*******************************************************************\
+
+Module: Ensure one backedge per target
```
peterschrammel, line 3: MIght as well live in goto-instrument

peterschrammel (approved):

PR #5077, opened 2019-09-02
Add more specific INVARIANT messages in order to make problem finding easier.

thomasspriggs:
Add more specific INVARIANT messages in order to make problem finding easier.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- ~~[ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~~ N/A - No user visible behaviour.
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- ~~[ ] My commit message includes data points confirming performance improvements (if claimed).~~ N/A - None claimed.
- [x] My PR is restricted to a single feature or bugfix.
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #5085, opened 2019-09-04
Improve catch support for checking of thrown instances of `invariant_failedt`

thomasspriggs:
This a follow up after a review comment on #5077

- [x] ~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~ N/A no user visible behaviour.
- [x] ~Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).~ N/A changes are to tests only.
- [x] ~My commit message includes data points confirming performance improvements (if claimed).~ N/A None claimed.

### DIFF
--- a/unit/testing-utils/invariant.cpp
+++ b/unit/testing-utils/invariant.cpp
@@ -1,0 +1,29 @@
+/// Author: Diffblue Ltd.
+
+#include "invariant.h"
+
+#include <utility>
+
+invariant_failure_containingt invariant_failure_containing(std::string expected)
+{
+  return invariant_failure_containingt{std::move(expected)};
+}
+
+invariant_failure_containingt::invariant_failure_containingt(
+  std::string expected)
+  : expected{std::move(expected)}
+{
+}
+
+bool invariant_failure_containingt::match(
+  const invariant_failedt &exception) const
+{
+  const std::string what = exception.what();
+  return what.find(expected) != std::string::npos;
+}
+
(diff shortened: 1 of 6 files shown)


unit/testing-utils/invariant.cpp
```diff
@@ -28,3 +28,11 @@ std::string invariant_failure_containingt::describe() const
          expected + "\"";
 }
 
+std::ostream &
+operator<<(std::ostream &out, const invariant_failedt &invariant_failed)
```
NathanJPhillips, line 32: Would this be useful generally with `invariant_failedt` rather than in testing utils?

NathanJPhillips (approved):

owen-mc-diffblue (approved):

unit/testing-utils/invariant.cpp
thomasspriggs: I am not sure. This printing seems fairly specific to the unit tests. If I put it with `invariant_failedt` then I think it would increase the risk that someone might unwittingly rename it to `pretty_print` or something rather than leaving it as an operator; which would revert the unit test printing back to using `{?}`, without giving any kind of error.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 91d162646a03feda6cf24fcf4db53b02406cfdb7).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128493068

PR #5086, opened 2019-09-04
Remove buggy model of Character.toCodePoint

smowton:
The models library provides a fine implementation of this function anyhow, without the type mismatches present in the model.


jbmc/regression/jbmc-strings/toCodePoint/Test.java
```diff
@@ -0,0 +1,10 @@
+public class Test {
+
+  public void test(char a, char b) {
+    assert Character.toCodePoint((char)0xD800, (char)0xDC00) == 0x10000;
+    assert Character.toCodePoint((char)0xDBFF, (char)0xDFFF) == 0x10ffff;
+    assert Character.toCodePoint((char)0xD800, (char)0xDFFF) == 0x103ff;
+    assert Character.toCodePoint((char)0xDBFF, (char)0xDC00) == 0x10fc00;
+    assert Character.toCodePoint(a, b) == 0x10000;
```
romainbrenguier, line 8: To make the test better maybe assert `!= 0x10000` and check in the trace produced by jbmc that it found the correct values for `a` and `b` to falsify the assertion.

romainbrenguier (approved):
I think ultimately character_refine_preprocess should disappear and all these methods should be modeled in the java-models-library (where not introducing bugs is easier), so if you want to remove more don't hesitate.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 0d9223e7fb98778f246d2e1c8b017441ec09e495).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/125916728

peterschrammel (approved):

PR #5087, opened 2019-09-04
Improve harness generated code readability

xbauch:
Previously all recursive initialisation took place in a single function. Now each type has it's own constructor function, which are called recursively. The results (variables to be initialised) are passed by pointer.

### DIFF
--- a/src/goto-harness/function_call_harness_generator.cpp
+++ b/src/goto-harness/function_call_harness_generator.cpp
@@ -18,7 +18,7 @@ Author: Diffblue Ltd.
 #include <util/string_utils.h>
 #include <util/ui_message.h>
 
-#include <goto-programs/goto_convert.h>
+#include <goto-programs/goto_convert_functions.h>
 #include <goto-programs/goto_model.h>
 
 #include <algorithm>
@@ -300,14 +300,7 @@ void function_call_harness_generatort::implt::
     symbol_table->lookup_ref(harness_function_name);
   goto_functions->function_map[harness_function_name].type =
     to_code_type(generated_harness.type);
-  auto &body = goto_functions->function_map[harness_function_name].body;
-  goto_convert(
-    static_cast<const codet &>(generated_harness.value),
-    *symbol_table,
-    body,
-    *message_handler,
-    function_to_call.mode);
-  body.add(goto_programt::make_end_function());
+  goto_convert(*symbol_table, *goto_functions, *message_handler);
 }
 
 code_function_callt::argumentst
--- a/src/goto-harness/function_call_harness_generator.cpp
+++ b/src/goto-harness/function_call_harness_generator.cpp
@@ -11,7 +11,6 @@ Author: Diffblue Ltd.
 #include <util/allocate_objects.h>
 #include <util/arith_tools.h>
 #include <util/exception_utils.h>
-#include <util/prefix.h>
 #include <util/std_code.h>
 #include <util/std_expr.h>
 #include <util/string2int.h>
@@ -213,10 +212,7 @@ void function_call_harness_generatort::implt::generate_nondet_globals(
     for(const auto &symbol_table_entry : *symbol_table)
     {
       const auto &symbol = symbol_table_entry.second;
-      if(
-        symbol.is_static_lifetime && symbol.is_lvalue &&
-        symbol.type.id() != ID_code &&
-        !has_prefix(id2string(symbol.name), CPROVER_PREFIX))
+      if(recursive_initialization->is_initialization_allowed(symbol))
       {
         globals.push_back(symbol.symbol_expr());
       }
--- a/src/goto-harness/function_call_harness_generator.cpp
+++ b/src/goto-harness/function_call_harness_generator.cpp
@@ -10,6 +10,7 @@ Author: Diffblue Ltd.
 
 #include <util/allocate_objects.h>
 #include <util/arith_tools.h>
+#include <util/c_types.h>
 #include <util/exception_utils.h>
 #include <util/std_code.h>
 #include <util/std_expr.h>
@@ -52,6 +53,8 @@ struct function_call_harness_generatort::implt
 
   std::map<irep_idt, irep_idt> function_argument_to_associated_array_size;
   std::map<irep_idt, irep_idt> function_parameter_to_associated_array_size;
+
+  std::set<symbol_exprt> global_pointers;
 
   /// \see goto_harness_generatort::generate
   void generate(goto_modelt &goto_model, const irep_idt &harness_function_name);
@@ -194,6 +197,11 @@ void function_call_harness_generatort::implt::generate(
 
   generate_nondet_globals(function_body);
   call_function(arguments, function_body);
+  for(const auto &global_pointer : global_pointers)
+  {
+    function_body.add(code_function_callt{
+      recursive_initialization->get_free_function(), {global_pointer}});
+  }
   add_harness_function_to_goto_model(std::move(function_body));
 }
 
@@ -228,7 +236,10 @@ void function_call_harness_generatort::implt::generate_initialisation_code_for(
   code_blockt &block,
   const exprt &lhs)
 {
-  recursive_initialization->initialize(lhs, 0, {}, block);
+  recursive_initialization->initialize(
+    lhs, from_integer(0, signed_int_type()), block);
+  if(lhs.type().id() == ID_pointer)
+    global_pointers.insert(to_symbol_expr(lhs));
 }
 
 void function_call_harness_generatort::validate_options(
@@ -366,6 +377,8 @@ void function_call_harness_generatort::implt::call_function(
   for(auto const &argument : arguments)
   {
     generate_initialisation_code_for(function_body, argument);
+    if(argument.type().id() == ID_pointer)
+      global_pointers.insert(to_symbol_expr(argument));
   }
   code_function_callt function_call{function_to_call.symbol_expr(),
                                     std::move(arguments)};
--- a/src/goto-harness/memory_snapshot_harness_generator.cpp
+++ b/src/goto-harness/memory_snapshot_harness_generator.cpp
@@ -231,7 +231,7 @@ code_blockt memory_snapshot_harness_generatort::add_assignments_to_globals(
   for(auto pair : snapshot)
   {
     const auto name = id2string(pair.first);
-    if(name.find(CPROVER_PREFIX) != 0)
+    if(!has_prefix(name, CPROVER_PREFIX))
       ordered_snapshot_symbols.push_back(pair);
   }
 
@@ -249,7 +249,22 @@ code_blockt memory_snapshot_harness_generatort::add_assignments_to_globals(
                pointer_depth(right.second.symbol_expr().type());
     });
 
-  code_blockt code;
+  code_blockt code{};
+
+  // add initialization for existing globals
+  for(const auto &pair : goto_model.symbol_table)
+  {
+    const auto &global_symbol = pair.second;
+    if(
+      global_symbol.is_static_lifetime && global_symbol.is_lvalue &&
+      global_symbol.type.id() != ID_code)
+    {
+      auto symeexr = global_symbol.symbol_expr();
+      if(symeexr.type() == global_symbol.value.type())
+        code.add(code_assignt{symeexr, global_symbol.value});
+    }
+  }
+
   for(const auto &pair : ordered_snapshot_symbols)
   {
     const symbolt &snapshot_symbol = pair.second;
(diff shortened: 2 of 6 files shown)


src/goto-harness/function_call_harness_generator.cpp
```diff
@@ -195,6 +197,9 @@ void function_call_harness_generatort::implt::generate(
 
   generate_nondet_globals(function_body);
   call_function(arguments, function_body);
+  for(const auto &global_pointer : global_pointers)
```
hannes-steffenhagen-diffblue, line 200: :pick: braces please

src/goto-harness/function_call_harness_generator.cpp
```diff
@@ -195,6 +197,9 @@ void function_call_harness_generatort::implt::generate(
...
+    function_body.add(code_function_callt{
```
hannes-steffenhagen-diffblue, line 202: Hm, I thought you'd have to pass more than one parameter here sometimes?

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -249,7 +249,21 @@ code_blockt memory_snapshot_harness_generatort::add_assignments_to_globals(
                pointer_depth(right.second.symbol_expr().type());
     });
 
-  code_blockt code;
+  code_blockt code{};
+
+  // add initialization for existing globals
+  for(auto pair : goto_model.symbol_table)
```
hannes-steffenhagen-diffblue: `const &`

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: c7c9708e41a864a73b8c1c4c2a2b93486cec270b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126018554

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -249,7 +249,21 @@ code_blockt memory_snapshot_harness_generatort::add_assignments_to_globals(
...
+  {
+    if(
+      id2string(pair.first).find(CPROVER_PREFIX) != 0 &&
+      pair.second.is_static_lifetime)
```
hannes-steffenhagen-diffblue: See the function call harness, you'll also want 
```
&& symbol.is_lvalue 
&& symbol.type.id() != ID_code 
&& !has_prefix(id2string(symbol.name), CPROVER_PREFIX)
```

because we don't want to initialise function symbols, non-lvalues or cprover internal variables (we should probably factor out this filtering stuff into a separate utility)

src/goto-harness/recursive_initialization.cpp
```diff
@@ -79,54 +81,172 @@ recursive_initializationt::recursive_initializationt(
...
+    max_depth_var_name(get_fresh_global_name(
+      "max_depth",
+      from_integer(
+        initialization_config.max_nondet_tree_depth,
+        signed_int_type()))),
+    min_depth_var_name(get_fresh_global_name(
+      "min_depth",
+      from_integer(
+        initialization_config.min_null_tree_depth,
+        signed_int_type())))
```
hannes-steffenhagen-diffblue, line 94: it's kind of irrelevant but I'd have expected these to be `size_t`

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8cb2f6ed82a64f24075739ad4010e1f9437dfc73).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126026333

src/goto-harness/recursive_initialization.cpp
```diff
@@ -79,54 +81,172 @@ recursive_initializationt::recursive_initializationt(
...
+  const exprt &depth,
   code_blockt &body)
 {
-  auto const &type = lhs.type();
-  if(type.id() == ID_struct_tag)
-  {
-    initialize_struct_tag(lhs, depth, known_tags, body);
-  }
-  else if(type.id() == ID_pointer)
+  const irep_idt &fun_name = build_constructor(lhs);
```
hannes-steffenhagen-diffblue, line 103: Surely `build_constructor` only needs `lhs.type()`?

src/goto-harness/recursive_initialization.cpp
```diff
@@ -79,54 +81,172 @@ recursive_initializationt::recursive_initializationt(
...
+  const symbolt &fun_symbol = goto_model.symbol_table.lookup_ref(fun_name);
+
+  if(lhs.id() == ID_symbol)
   {
-    if(lhs.id() == ID_symbol)
+    const irep_idt &lhs_name = to_symbol_expr(lhs).get_identifier();
+    // skip initialisation of max/min depth globals
+    if(lhs_name == max_depth_var_name || lhs_name == min_depth_var_name)
```
hannes-steffenhagen-diffblue: I think it'd be cleaner to remove these from the global set in the first place - IIRC at least the function-harness automatically strips all variables with a `__CPROVER_` prefix from the list of variables to nondet (all harnesses should do this, may be worth extracting into separate functionality). All our global (maybe even local, given that cbmc doesn't really care all that much about scope) variables should have the `__CPROVER_` prefix attached to them anyway (that way we can avoid accidentally overriding variables that we add in here without having to remember to block each of them individually here)

src/goto-harness/recursive_initialization.cpp
```diff
@@ -79,54 +81,172 @@ recursive_initializationt::recursive_initializationt(
...
+      auto size_var = get_associated_size_variable(lhs_name);
+      if(size_var.has_value())
       {
-        initialize_cstring(lhs, depth, known_tags, body);
-        return;
+        const symbol_exprt &size_symbol =
+          goto_model.symbol_table.lookup_ref(size_var.value()).symbol_expr();
+        body.add(code_function_callt{
+          fun_symbol.symbol_expr(),
+          {depth, address_of_exprt{lhs}, address_of_exprt{size_symbol}}});
```
hannes-steffenhagen-diffblue, line 118: That's one way to do it, but I don't like having the replicated logic between the constructor and the call; I'd prefer if the dynamic array constructor just always took a size pointer and just set it to null if it's not needed (kind of mirroring similar functions in the real world)

src/goto-harness/recursive_initialization.cpp
```diff
@@ -426,49 +332,421 @@ std::string recursive_initialization_configt::to_string() const
...
 
-recursive_initializationt::array_convertert
-recursive_initializationt::default_array_member_initialization()
+irep_idt recursive_initializationt::get_fresh_global_name(
+  const std::string &symbol_name,
+  const exprt &initial_value) const
+{
+  symbolt &fresh_symbol = get_fresh_aux_symbol(
+    signed_int_type(),
+    "__goto_harness",
```
hannes-steffenhagen-diffblue: Probably makes more sense to use `CPROVER_PREFIX`, same in following

src/goto-harness/recursive_initialization.cpp
```diff
@@ -426,49 +332,421 @@ std::string recursive_initialization_configt::to_string() const
...
+    return "int";
+  else if(type.id() == ID_unsignedbv)
+    return "uint";
+  else
+    return "";
+}
+
+symbol_exprt recursive_initializationt::get_free_function()
+{
+  auto free_sym = goto_model.symbol_table.lookup("free");
```
hannes-steffenhagen-diffblue, line 486: FYI we had some talk at some point about using CBMCs internal allocate/deallocate instructions for this, but I didn't have much luck for them (no need for changes, just a remark)

hannes-steffenhagen-diffblue (approved):
Mostly looks fine to me, but I think the decision of what things to initialise or not should be made outside of `recursive_initialisation`

src/goto-harness/recursive_initialization.cpp
xbauch: We also need the name of the symbol (if `lhs` is a symbol) to get the `associated_size_variable` etc.

src/goto-harness/recursive_initialization.cpp
xbauch: The only reason I went for `int` is that it looks better in the dumped code.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Extracted and moved to `recursive_initializationt`.

src/goto-harness/memory_snapshot_harness_generator.cpp
xbauch: Added.

src/goto-harness/function_call_harness_generator.cpp
xbauch: Added.

src/goto-harness/recursive_initialization.cpp
xbauch: Done.

src/goto-harness/recursive_initialization.cpp
xbauch: Done.

src/goto-harness/recursive_initialization.cpp
xbauch: Done.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 099cf1208958436f0353c9c485cbc965077d10d6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128903045

src/goto-harness/recursive_initialization.cpp
```diff
@@ -428,6 +428,9 @@ std::string recursive_initialization_configt::to_string() const
 
 recursive_initializationt::array_convertert
 recursive_initializationt::default_array_member_initialization()
+irep_idt recursive_initializationt::get_fresh_global_name(
+  const std::string &symbol_name,
```
NlightNFotis, line 347: `get_fresh_global_name` and `get_fresh_global_symexpr` look very similar. Are we sure the bulk of the work performed in one function, and then have the helpers call into it?

src/goto-harness/recursive_initialization.cpp
```diff
@@ -121,7 +121,77 @@ void recursive_initializationt::initialize(
...
+    size_symbol_expr = size_symbol.symbol_expr();
   }
+
+  const symbolt &function_symbol = get_fresh_fun_symbol(
+    "type_constructor_" + pretty_type, code_typet{fun_params, empty_typet{}});
+  type_constructor_names[type] = function_symbol.name;
+  symbolt *mutable_symbol = symbol_table.get_writeable(function_symbol.name);
+
+  // the body is specific for each type of expression
+  mutable_symbol->value = build_constructor_body(
```
NlightNFotis, line 246: slightly weird that `build_constructor_body` is being referenced here but added in the next commit and there's a prototype added for it as well in `recursive_initialization.h`

src/goto-harness/recursive_initialization.cpp
```diff
@@ -122,6 +122,40 @@ void recursive_initializationt::initialize(
...
+  const typet &type = result_symbol.type().subtype();
+  if(type.id() == ID_struct_tag)
+  {
+    return build_struct_constructor(depth_symbol, result_symbol);
+  }
+  else if(type.id() == ID_pointer)
+  {
+    if(lhs_name.has_value())
+    {
+      if(should_be_treated_as_cstring(*lhs_name) && type == char_type())
```
NlightNFotis, line 161: Just a question, since I'm not sure why it's happening: 

Why in other places we compare the `type.id()`, like for instance in line 137 `type.id() == ID_pointer` and here we compare the type itself with the constructor result? Are we sure equality is working as it should? If yes, should all of these be harmonised?

NlightNFotis (commented):
Looks good to me overall, just some minor comments.

src/goto-harness/recursive_initialization.cpp
xbauch: I wasn't sure how to break the PR into individual commits so I sometimes formed a commit around a body of a function but only declared the functions called inside that body.

src/goto-harness/recursive_initialization.cpp
xbauch: That's just my being lazy and not finding out what type-id `char` should have.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9ca0824bb380132818fce832ca4e06dc733b2d5c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130278481

PR #5089, opened 2019-09-05
GOTO-symex: propagate notequal conditions for boolean types [TG-9390]

smowton:
Conditional branches on "x != false", as well as "x" and "!x", can and should lead to constant
propagation just as "x == true" does now. This is particularly beneficial for the
"clinit_already_run" variables that are maintained to determine if a static initialiser should
be run or not -- the current "if(already_run != true) clinit(); already_run = true;" condition
now leaves symex certain that it has been run at the bottom of the function regardless of the
already_run flag's initial value, whereas before it could remain uncertain and so explore the
static initialiser more than is necessary.

Command line to see the benefit of this change on Tika-parsers:

`test-generator --java-max-vla-length 192 --unwind 2 --java-unwind-enum-static --trace --export-imports --java-assume-inputs-non-null --classpath ~/test-gen/lib/models-library/model/target/models.jar tika-all.jar --function 'org.apache.tika.parser.jdbc.JDBCTableReader.handleClob'`

where `tika-all.jar` is a jar containing all of the `org/apache/tika` package from all Tika modules (i.e. tika-core, tika-parsers etc).

### DIFF
--- a/src/goto-symex/goto_state.cpp
+++ b/src/goto-symex/goto_state.cpp
@@ -46,11 +46,72 @@ void goto_statet::apply_condition(
 {
   if(condition.id() == ID_and)
   {
+    // A == B && C == D && E == F ...
+    // -->
+    // Apply each condition individually
     for(const auto &op : condition.operands())
       apply_condition(op, previous_state, ns);
   }
+  else if(condition.id() == ID_notequal &&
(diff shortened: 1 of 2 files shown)


owen-mc-diffblue:
I'm taking this work over, so I have to close this PR and open a new: https://github.com/diffblue/cbmc/pull/5091

PR #5078, opened 2019-09-03
Avoid redundant computations in get_user_specified_clinit

romainbrenguier:
Some computation like json parsing and iterating through the map were done at each call to `get_user_specified_clinit` which is called a lot of times for projects with large json file for static values.
With this pull request on such an example the execution time went from 10 seconds to 2 seconds (mostly thanks to removing the redundant json parsing).

 

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [x] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [na] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #5090, opened 2019-09-05
Add a lazyt class for delaying computations

romainbrenguier:
This class is meant to be used for delaying the computation of a value to the first time we need it, but avoid recomputing it if it is needed several times.
In particular I would have found such a class useful when writing this pull request: https://github.com/diffblue/cbmc/pull/5078

### DIFF
--- a/src/util/lazy.h
+++ b/src/util/lazy.h
@@ -1,0 +1,46 @@
+/*******************************************************************\
+
+Module: Util
+
+Author: Romain Brenguier, romain.brenguier@diffblue.com
+
+\*******************************************************************/
+
+#ifndef CPROVER_UTIL_LAZY_H
+#define CPROVER_UTIL_LAZY_H
+
+#include <functional>
+#include <util/optional.h>
+
+template <typename valuet>
+class lazyt
+{
+public:
+  /// Delay the computation of \p fun to the next time the \c force method
+  /// is called.
+  static lazyt from_fun(std::function<valuet()> fun)
+  {
+    return lazyt{std::move(fun)};
+  }
+
+  /// Force the computation of the value. If it was already computed,
+  /// return the same result.
+  valuet force()
+  {
+    if(value)
+      return *value;
+    value = evaluation_function();
+    return *value;
+  }
+
+private:
+  optionalt<valuet> value;
+  std::function<valuet()> evaluation_function;
+
+  explicit lazyt(std::function<valuet()> fun)
+    : evaluation_function(std::move(fun))
+  {
+  }
+};
+
+#endif // CPROVER_UTIL_LAZY_H
--- a/unit/util/lazy.cpp
+++ b/unit/util/lazy.cpp
@@ -1,0 +1,29 @@
+/*******************************************************************\
+
+Module: Unit tests for lazy
+
+Author: Romain Brenguier, romain.brenguier@diffblue.com
+
+\*******************************************************************/
+
+#include <testing-utils/use_catch.h>
+#include <util/lazy.h>
+
+SCENARIO("lazy test", "[core][util][lazy]")
+{
+  std::size_t call_counter = 0;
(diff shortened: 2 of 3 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: e5bf82d4eb3acbd229330cf0e2d0091beb79b77e).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126077346

src/util/lazy.h
```diff
@@ -0,0 +1,46 @@
...
+  /// Delay the computation of \p fun to the next time the \c force method
+  /// is called.
+  static lazyt from_fun(std::function<valuet(infot)> fun)
+  {
+    return lazyt{std::move(fun)};
+  }
+
+  /// Force the computation of the value. If it was already computed (even if
+  /// \p info was different), return the same result.
+  valuet force(infot info)
```
owen-mc-diffblue: I find "force" an unintuitive name. How about "evaluate", "compute" or "force_evaluation"?

owen-mc-diffblue (approved):

src/util/lazy.h
romainbrenguier: `force` is used by .net, scheme and ocaml https://en.wikipedia.org/wiki/Lazy_evaluation so I think it's relatively standard, but I don't know if someone encountered similar concepts in other libraries or languages?

owen-mc-diffblue: If it's a standard term then that's fine.

src/util/lazy.h
```diff
@@ -0,0 +1,46 @@
...
+template <typename valuet, typename infot>
+class lazyt
+{
+public:
+  /// Delay the computation of \p fun to the next time the \c force method
+  /// is called.
+  static lazyt from_fun(std::function<valuet(infot)> fun)
+  {
+    return lazyt{std::move(fun)};
+  }
```
danpoe, line 24: Why not a public constructor?

danpoe (approved):

src/util/lazy.h
romainbrenguier: I think having a static function with a name is clearer in that case, when I read `lazyt::from_fun(f)` I know we use a function `f` to make a `lazyt` object, but `lazyt(f)` is more obscure.

romainbrenguier:
@owen-mc-diffblue @danpoe following the suggestion of @LAJW I've removed the `infot` for force, so that it cannot be called twice with different arguments.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 8dbbca1c18d2928d03dbc0010848c83262c14c40).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126233098

unit/util/lazy.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+SCENARIO("lazy test", "[core][util][lazy]")
+{
+  std::size_t call_counter = 0;
+  auto length_with_counter = [&call_counter](const std::string &s) {
+    ++call_counter;
+    return s.length();
+  };
+  auto lazy_length =
+    lazyt<int>::from_fun([&]() { return length_with_counter("foo"); });
```
LAJW, line 20: 😊 https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/lazy-expressions
And now you could even infer return type by doing:
```
template<T> auto lazy(T func) -> lazyt<decltype(func())> { ... }
```

Also, `()` in zero-argument lambdas is optional.

LAJW (approved):

peterschrammel (approved):

unit/util/lazy.cpp
romainbrenguier: I'm adding it in a follow-up PR: https://github.com/diffblue/cbmc/pull/5101

PR #5091, opened 2019-09-05
[TG-9390] GOTO-symex: propagate notequal conditions for boolean types

owen-mc-diffblue:
Conditional branches on "x != false", as well as "x" and "!x", can and should lead to constant
propagation just as "x == true" does now. This is particularly beneficial for the
"clinit_already_run" variables that are maintained to determine if a static initialiser should
be run or not -- the current "if(already_run != true) clinit(); already_run = true;" condition
now leaves symex certain that it has been run at the bottom of the function regardless of the
already_run flag's initial value, whereas before it could remain uncertain and so explore the
static initialiser more than is necessary.


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 357dfbf6488e2c3700297bd59f91f5a66306c611).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126092608

owen-mc-diffblue:
This is a cleaned-up version of #5089, with a few bugs fixed and tests added. I added tests in C and java take make sure it worked with the boolean types in both languages.

The last commit was a part of the code from #5089 that I found wasn't needed for any of my tests. It seems to be doing the right kind of thing, so one argument is to keep it in so that if the any of the language frontends start producing GOTO like that then we'll still be able to determine the value of the symbol. The counter-argument is that it isn't tested, so it might not be doing the right thing, and it's misleading to put in code for a case that doesn't currently happen.

owen-mc-diffblue:
One test is failing (regression/cbmc/Empty_struct1). The problem is that we declare `b` and `c` like this:
```
_Bool b, c=b;
```
We then condition on `b`. The new code sets `b` to be 0 on one branch and 1 on the other. The test ends with the assertion `assert(b==c);`. cbmc falsifies this assertion by initially setting `b` to some value other than 0 or 1, e.g. 9 in the case I looked at, and we end up asserting that `1 == 9`.

The only reasonable resolutions I can think of are (a) restricting `_Bool` types in C to have values 0 and 1, which would take us away from how C actually works, (b) restricting this feature to languages which have a proper boolean type, like java

smowton:
I'd suggest only doing this for real booleans (bool_typet, 1-bit) not cbools (8-bit, conventionally either 1 or 0). That loses some cases but ensures compliance with C, and the static init fields I set out to handle are "proper" booleans

smowton:
(presumably this is because of the unrestricted typecasting in the equals case)

owen-mc-diffblue:
@smowton Okay, that makes sense. `bool_typet` symbols aren't made by the C or java front-ends, are they? They're only synthetic, right? Is there any way of making them apart from these `A_clinit_already_run` symbols we make for java? I'm thinking about how I can test this feature.

src/goto-symex/goto_state.cpp
```diff
@@ -46,11 +47,46 @@ void goto_statet::apply_condition(
 {
   if(condition.id() == ID_and)
   {
+    // A == B && C == D && E == F ...
+    // -->
+    // Apply each condition individually
     for(const auto &op : condition.operands())
       apply_condition(op, previous_state, ns);
   }
+  else if(condition.id() == ID_not)
```
romainbrenguier: :pick: `expr_try_dynamic_cast<not_exprt>` :tractor:

src/goto-symex/goto_state.cpp
```diff
@@ -46,11 +47,46 @@ void goto_statet::apply_condition(
...
+  {
+    const exprt &operand = to_not_expr(condition).op();
+    if(operand.id() == ID_notequal)
```
romainbrenguier: :tractor:

src/goto-symex/goto_state.cpp
```diff
@@ -46,11 +47,46 @@ void goto_statet::apply_condition(
...
+      // !(A != B)
+      // -->
+      // A == B
+      const auto &notequal_expr = to_notequal_expr(operand);
+      apply_condition(
+        equal_exprt{notequal_expr.lhs(), notequal_expr.rhs()},
+        previous_state,
+        ns);
+    }
+    else if(operand.id() == ID_equal)
```
romainbrenguier: :tractor:

src/goto-symex/symex_goto.cpp
```diff
@@ -55,21 +55,12 @@ void goto_symext::apply_goto_condition(
...
-  {
-    auto &not_equal_expr = to_notequal_expr(new_guard);
-    jump_not_taken_state.apply_condition(
-      equal_exprt(not_equal_expr.lhs(), not_equal_expr.rhs()),
-      current_state,
-      ns);
-  }
+  exprt negated_new_guard = new_guard.id() == ID_not
+                              ? to_not_expr(new_guard).op()
+                              : not_exprt(new_guard);
```
romainbrenguier: use `boolean_negate`

src/goto-symex/symex_goto.cpp
```diff
@@ -55,21 +55,12 @@ void goto_symext::apply_goto_condition(
...
-      to_not_expr(new_guard).op(), current_state, ns);
-  else if(new_guard.id() == ID_notequal)
-  {
-    auto &not_equal_expr = to_notequal_expr(new_guard);
-    jump_not_taken_state.apply_condition(
-      equal_exprt(not_equal_expr.lhs(), not_equal_expr.rhs()),
-      current_state,
-      ns);
-  }
+  exprt negated_new_guard = new_guard.id() == ID_not
```
romainbrenguier: const

romainbrenguier: const

romainbrenguier (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 6bf2df69a12058545f94c48138ed1477795c7292).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126485597

romainbrenguier:
It would be good to have some benchmark results to see if get a performance improvement from this.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 96155fb6b5196a5e64de38b044c9f9167544b280).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126563612

src/goto-symex/goto_state.cpp
```diff
@@ -10,6 +10,7 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 #include "goto_symex_is_constant.h"
 #include "goto_symex_state.h"
 
+#include <util/arith_tools.h>
```
JohnDumbell: Is this used? (After a quick scan I couldn't see an obvious use)

JohnDumbell (approved):

src/goto-symex/goto_state.cpp
owen-mc-diffblue: You're right, I think I must have added it for `to_integer` in an earlier version, but I'm not using that any more. I've removed it.

owen-mc-diffblue:
I benchmarked this branch vs develop on 13 methods from tika-parsers that had been chosen because they don't complete quickly. I ran with a timeout of 10 minutes, and `--show-vcc` in an attempt to stop after symex, and not run the SAT solver, since I expect the change in this PR to speed up symex (it may also speed up SAT solving a little bit). 3 of the methods errored in both runs (2 because of a models library problem, one because it ran out of memory just after symex. 6 of the methods timed out in symex in both runs. That leaves 4 methods to compare. Here is a table, with develop on the left and this branch on the right

64.6 | 63.2
11.2 | 9.7
8.0 | 7.3 
103.4 | 29.9

The runtime decreased by 2%, 13%, 9% and 71% respectively. The first one is within normal variation. The middle 2 show a small but noticeable improvement. But the last one shows without a doubt that this PR can drastically reduce symex run times.

owen-mc-diffblue:
@kroening @peterschrammel @tautschnig This needs a code owner review - do any of you have time to look at it? (@smowton wrote most of it, so he probably shouldn't review it as well.)

segun3d:
> I benchmarked this branch vs develop on 13 methods from tika-parsers that had been chosen because they don't complete quickly. I ran with a timeout of 10 minutes, and `--show-vcc` in an attempt to stop after symex, and not run the SAT solver, since I expect the change in this PR to speed up symex (it may also speed up SAT solving a little bit). 3 of the methods errored in both runs (2 because of a models library problem, one because it ran out of memory just after symex. 6 of the methods timed out in symex in both runs. That leaves 4 methods to compare. Here is a table, with develop on the left and this branch on the right
> 
> 64.6 | 63.2
> 11.2 | 9.7indly 
> 8.0 | 7.3
> 103.4 | 29.9
> 
> The runtime decreased by 2%, 13%, 9% and 71% respectively. The first one is within normal variation. The middle 2 show a small but noticeable improvement. But the last one shows without a doubt that this PR can drastically reduce symex run times.

@owen-mc-diffblue Indeed the last bench mark is impressive. Could you kindly state which benchmark tool you used. Thanks.

owen-mc-diffblue:
I used the benchmarking script in test-gen, written by @JohnDumbell. I used the following list of methods, provided by @smowton
```
org.apache.tika.parser.DefaultParser.<init>:(Lorg/apache/tika/mime/MediaTypeRegistry;Ljava/lang/ClassLoader;)V
org.apache.tika.parser.dwg.DWGParser.get2004Props:(Ljava/io/InputStream;Lorg/apache/tika/metadata/Metadata;Lorg/apache/tika/sax/XHTMLContentHandler;)V
org.apache.tika.parser.gdal.GDALParser.processCommand:(Ljava/io/InputStream;)Ljava/lang/String;
org.apache.tika.parser.html.HtmlHandler.resolve:(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;
org.apache.tika.parser.isatab.ISATabUtils.parseAssay:(Ljava/io/InputStream;Lorg/apache/tika/sax/XHTMLContentHandler;Lorg/apache/tika/metadata/Metadata;Lorg/apache/tika/parser/ParseContext;)V
org.apache.tika.parser.iwork.PagesContentHandler.characters:([CII)V
org.apache.tika.parser.iwork.PagesContentHandler.resolveMetaDataKey:(Ljava/lang/String;)Ljava/lang/Object;
org.apache.tika.parser.jdbc.JDBCTableReader.getDetector:()Lorg/apache/tika/detect/Detector;
org.apache.tika.parser.jdbc.JDBCTableReader.handleClob:(Ljava/lang/String;Ljava/lang/String;ILjava/sql/ResultSet;ILorg/xml/sax/ContentHandler;Lorg/apache/tika/parser/ParseContext;)V
org.apache.tika.parser.microsoft.ooxml.XWPFWordExtractorDecorator.<init>:(Lorg/apache/tika/parser/ParseContext;Lorg/apache/poi/xwpf/extractor/XWPFWordExtractor;)V
org.apache.tika.parser.microsoft.ooxml.xwpf.XWPFEventBasedWordExtractor.getText:()Ljava/lang/String;
org.apache.tika.parser.ocr.TesseractOCRParser.extractOutput:(Ljava/io/InputStream;Lorg/apache/tika/sax/XHTMLContentHandler;)V
org.apache.tika.parser.xml.ElementMetadataHandler.isMatchingElement:(Ljava/lang/String;Ljava/lang/String;)Z
```
The method with the impressive speed-up was `org.apache.tika.parser.xml.ElementMetadataHandler.isMatchingElement`

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 5bc512acc39a231fd9be56d90ad1587211a018a7).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/127177250

jbmc/regression/jbmc/symex_can_track_when_clinits_have_definitely_been_run/test.desc
```diff
@@ -0,0 +1,16 @@
+CORE
+Test.class
+--function Test.test --show-vcc
+^EXIT=0$
+^SIGNAL=0$
+--
+java::ClassWithStaticInit::clinit_already_run#[1-9][0-9]
+^warning: ignoring
+--
+After the second call to ClassWithStaticInit.getStaticValue(), symex should have
```
peterschrammel: Maybe explain how many generations you expect to be per call without the optimisation. That might it easier to understand why you expect >=10 to be bad.

peterschrammel (approved):

jbmc/regression/jbmc/symex_can_track_when_clinits_have_definitely_been_run/test.desc
owen-mc-diffblue: Good point Peter. Actually, the number of generations per call could varies depending on whether `apply_condition` thinks it can set a value to be constant on one or both of the branches, and also future changes. I think it's safest to just say that we expect at least one new generation per call if symex doesn't know that `clinit_already_run` is true. I have increased the number of static function calls to 20, to make absolutely sure that we'd go over if the feature isn't working properly.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d312dad5cabab49a49544d940bc718530cfd81c2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/127355957

segun3d:
@antlechner has confirmed that regression tests are sufficient, this in addition to the benchmarks results (see above) means this is now QA approved.

romainbrenguier (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a2cbcc2d547af419021344f8eaa38527037b1a2b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/127405553

Issue #4978, opened 2019-08-02
Snapshot-harness may initialise variables in the wrong order


smowton:
When snapshot-harness tries to set up aliasing pointers, it may use the name of one of the aliases to set up the other. For example, when working correctly it might generate:
```
int memory[10];
x = memory;
y = &x[1];
```
But when generating those same assignments it might emit them in the wrong order, as it appears to expect that their RHS values are independent, leading to:
```
int memory[10];
y = &x[1];
x = memory;
```

Since `x` is a global variable this leads to y being `(int*)4` rather than pointing into the same array as `x` as intended.

Real-world example from regression test `snapshot-harness/dynamic-array-int`:
Good:
```
generated_entry_function /* generated_entry_function */
        // 177 no location
        tmp = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
        // 178 no location
        array = tmp;
        // 179 no location
        iterator1 = tmp;
        // 180 no location
        iterator2 = &array[1ul];
        // 181 no location
        iterator3 = &array[1ul];
        // 182 no location
        func_init_done = FALSE;
        // 183 no location
        main();
        // 184 no location
        END_FUNCTION
```
Bad (note `array` is read at line 45 before it is set at line 47)
```
generated_entry_function /* generated_entry_function */
        // 44 no location
        tmp = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
        // 45 no location
        iterator2 = &array[1ul];
        // 46 no location
        iterator1 = tmp;
        // 47 no location
        array = tmp;
        // 48 no location
        iterator3 = &array[1ul];
        // 49 no location
        func_init_done = FALSE;
        // 50 no location
        main();
        // 51 no location
        END_FUNCTION
```

Whether or not `snapshot-harness` malfunctions on any given occasion depends on the order it chooses to emit the assignments, which depends on compiler version, standard library version, phase of moon, etc.

GOTO binaries showing the problem: [examples.zip](https://github.com/diffblue/cbmc/files/3461931/examples.zip)

PR #5095, opened 2019-09-06
Fix the ordering of variable initialisation in goto-harness

xbauch:
When using the `memory-snapshot` harness.

See https://github.com/diffblue/cbmc/issues/4978 for the problem.

Previously we were using the `refers-to` relation (`a` refers to `b` if the value `a` should be initialised with contains -- as a subexpression -- the symbol `b`) as if it was a linear ordering between the snapshot symbols. That is, we called `std::stable_sort` with this relation which is incorrect.

Now we select one linearisation of the partial order (given by `refers-to`) to be the initilisation ordering. Note that we ignore cycles, or more precisely in the presence of a cycle we start from the first visited symbol. This is correct, because for

```
struct A { A*next; };
A a1;
A a2;
a1.next = &a2;
a2.next = &a1;
```
the snapshot is

```
a1 = {.next=&a2}
a2 = {.next=&a1}
```

And the order is irrelevant. There is also a test-case which was failing before.

### DIFF
--- a/src/goto-harness/memory_snapshot_harness_generator.cpp
+++ b/src/goto-harness/memory_snapshot_harness_generator.cpp
@@ -209,18 +209,6 @@ size_t memory_snapshot_harness_generatort::pointer_depth(const typet &t) const
     return pointer_depth(t.subtype()) + 1;
 }
 
-bool memory_snapshot_harness_generatort::refers_to(
-  const exprt &expr,
-  const irep_idt &name) const
-{
-  if(expr.id() == ID_symbol)
-    return to_symbol_expr(expr).get_identifier() == name;
-  return std::any_of(
-    expr.operands().begin(),
-    expr.operands().end(),
-    [this, name](const exprt &subexpr) { return refers_to(subexpr, name); });
-}
-
 code_blockt memory_snapshot_harness_generatort::add_assignments_to_globals(
   const symbol_tablet &snapshot,
   goto_modelt &goto_model) const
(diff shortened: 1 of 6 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: e034622be65d7b4a06a7f058bc846af1992abac9).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126261769

src/goto-harness/memory_snapshot_harness_generator.cpp
```diff
@@ -207,16 +207,16 @@ size_t memory_snapshot_harness_generatort::pointer_depth(const typet &t) const
     return pointer_depth(t.subtype()) + 1;
 }
 
-bool memory_snapshot_harness_generatort::refers_to(
+void memory_snapshot_harness_generatort::collect_references(
```
NlightNFotis: While I understand the need for the change for this, I feel slightly uneasy about the change to the return value (which provides some sort of feedback) to a void return value.

hannes-steffenhagen-diffblue:
I think the first commit here should probably just have the entire description of this PR in it

NlightNFotis (approved):
LGTM

src/goto-harness/memory_snapshot_harness_generator.h
```diff
@@ -275,11 +275,68 @@ class memory_snapshot_harness_generatort : public goto_harness_generatort
...
+  struct partial_ordert
+  {
+  public:
+    explicit partial_ordert(const std::map<Key, std::set<Key>> &por) : por(por)
+    {
+    }
+
+    void sort(
+      const std::vector<std::pair<Key, Value>> &input,
+      std::vector<std::pair<Key, Value>> &output)
```
hannes-steffenhagen-diffblue: Is there a reason this can't do an in-place sort?

src/goto-harness/memory_snapshot_harness_generator.h
```diff
@@ -275,11 +275,68 @@ class memory_snapshot_harness_generatort : public goto_harness_generatort
...
+    const std::map<Key, std::set<Key>> &por;
+
+    std::set<Key> inserted;
+    std::set<Key> seen;
+
+    void recurse_then_insert(
+      const Key &key,
+      const std::map<Key, Value> &input,
+      std::vector<std::pair<Key, Value>> &output)
+    {
```
hannes-steffenhagen-diffblue, line 339: This seems to be basically doing a DFS. Might be worth just renaming the function to reflect that.

It could just have an API like

```
template<typename Value, typename GetNodeFromValue, typename GetNeighbors, typename OnFirstVisit>
void depth_first_search(Value &&entry_node, GetKeyFromValue &&get_key_from_value, GetNeighbors &&getNeighbors, OnFirstVisit &&onFirstVisit);
```

and then you could use it above like
```
auto get_key_from_value = [](const std::pair<Key, Value>& kv_pair) { return kv_pair.first; };
auto get_neighbors = [&](const Key& key) { return por.at(key); }
auto on_first_visit = [&](const Value& value) { output.push_back(value); }
for(const auto &item : input) {
  dfs(item, get_key_from_value, get_neighbors, on_first_visit);
}
```

(Note that I've changed `Value` to what was previously `std::pair<Key,Value>`, and instead there's a separate function for getting the key from the value).

:exclamation: This is just a suggestion, I'm not saying you have to do it exactly this way obviously, but in any case I think it's worth emphasising that "recurse then insert" is basically doing DFS in the code, either by renaming the function or mentioning it in a comment

src/goto-harness/memory_snapshot_harness_generator.h
```diff
@@ -275,11 +275,68 @@ class memory_snapshot_harness_generatort : public goto_harness_generatort
...
+
+      for(const auto &item : input)
+        recurse_then_insert(item.first, searchable_input, output);
+    }
+
+  private:
+    const std::map<Key, std::set<Key>> &por;
+
+    std::set<Key> inserted;
+    std::set<Key> seen;
```
hannes-steffenhagen-diffblue: Note that these make `sort` a destructive method, i.e. an instance becomes unusable after `sort` is first called. Might be better to just pass these on as parameters, or loopify the recursion and just keep these as local variables

hannes-steffenhagen-diffblue (approved):
Looks mostly good to me, but I think the partial order sort class could use some cleanup.

src/goto-harness/memory_snapshot_harness_generator.h
xbauch: I went half-way. Given that the ordering is on `keys` only, the structure is only templated by `Key` and the `sort`, `dfs` functions and templated by some `T`.

src/goto-harness/memory_snapshot_harness_generator.h
xbauch: I wrapped the `dfs` in a `cleaner` function.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 373947eae08aaad5bb1da3fa81af594f85ed0fbb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/129287253

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: a4df02c57215a52841fad69cabfd1b4711495695).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/129292085

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: fbb40a044ee56870d6f4d8007a0ec62d173f9e2c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/134848870

PR #5097, opened 2019-09-07
Symex: distinguish reachability from the state guard

smowton:
Previously the state guard being false was taken to indicate unreachable code,
and different causes of unreachability were handled inconsistently:

* ASSUME false implied unreachability but had no effect on the guard
* --unwind or --depth being exceeded usually meant setting the guard false

This meant that expensive code downstream of an ASSUME false *was* executed, but the state
guard was maintained as usual including shrinking the guard when branches converge, while
--unwind or --depth limit breaks led to code *not* being executed (vital for these options
to function properly) but the state guard grew ever larger because the part that had gone
missing (been set to false) never got merged back in.

This commit distinguishes the two concepts: the "reachable" flag is set false whenever a
limit break OR and ASSUME false happens, but state guard maintenance continues as normal.
A new symex_unreachable_goto tries to find some way through the CFG back to reachable code;
when it can't due to loops *then* it sets the state guard to false, signifying our not
being able to figure out where the guard ought to be merged back in.

This means that ASSUME false instructions now correctly truncate execution and are
recognised as unreachable code for the purposes of state merging (leading to better
constant propagation downstream), but in some cases we may see more guard growth
because code downstream of an ASSUME false is executed less rigorously that before,
which could result in the guard not being merged or being merged later than it otherwise
would have been.

Depth and unwind limit breaks behave roughly as before, except that there is some chance
the state guard will be merged rather than discarded, reducing costs downstream.

### DIFF
--- a/src/goto-symex/goto_symex.h
+++ b/src/goto-symex/goto_symex.h
@@ -332,6 +332,9 @@ protected:
   /// Symbolically execute a GOTO instruction
   /// \param state: Symbolic execution state for current instruction
   virtual void symex_goto(statet &state);
+  /// Symbolically execute a GOTO instruction in the context of unreachable code
+  /// \param state: Symbolic execution state for current instruction
+  void symex_unreachable_goto(statet &state);
   /// Symbolically execute a START_THREAD instruction
   /// \param state: Symbolic execution state for current instruction
   virtual void symex_start_thread(statet &state);
--- a/src/goto-symex/symex_goto.cpp
+++ b/src/goto-symex/symex_goto.cpp
@@ -211,6 +211,8 @@ renamedt<exprt, L2> try_evaluate_pointer_comparisons(
 
 void goto_symext::symex_goto(statet &state)
 {
+  PRECONDITION(state.reachable);
+
   const goto_programt::instructiont &instruction=*state.source.pc;
 
   exprt new_guard = clean_expr(instruction.get_condition(), state, false);
@@ -413,6 +415,7 @@ void goto_symext::symex_goto(statet &state)
     symex_transition(state, state_pc, backward);
 
     state.guard = guardt(false_exprt(), guard_manager);
+    state.reachable = false;
   }
   else
   {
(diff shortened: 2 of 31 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: fe7b5069f877a8a77d6ff1030b37923e30dd5810).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128777244

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 14bf1717c6dd9ffb7e237a7738b51ce30c713d02).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128782532

peterschrammel (commented):
Do you have performance data w/wo this change?

smowton:
Coming up

src/goto-symex/goto_symex.h
```diff
@@ -332,6 +332,9 @@ class goto_symext
   /// Symbolically execute a GOTO instruction
   /// \param state: Symbolic execution state for current instruction
   virtual void symex_goto(statet &state);
+  /// Symbolically execute a GOTO instruction in the context of unreachable code
+  /// \param state: Symbolic execution state for current instruction
+  void symex_unreachable_goto(statet &state);
```
romainbrenguier, line 337: I suspect this method can be marked `const` or even `static`

src/goto-symex/symex_goto.cpp
```diff
@@ -513,6 +516,80 @@ void goto_symext::symex_goto(statet &state)
...
+      // Give up trying to salvage the guard:
+      state.guard.add(false_exprt());
+      symex_transition(state);
+    }
+    else
+    {
+      // Take the branch:
+      queue_unreachable_state_at_target();
+      state.guard.add(false_exprt());
+      symex_transition(state);
```
romainbrenguier: the two last instructions are common to the if and else cases so this can be simplified

src/goto-symex/symex_goto.cpp
```diff
@@ -513,6 +516,80 @@ void goto_symext::symex_goto(statet &state)
...
+    }
+  }
+  else
+  {
+    // Arbitrarily assume a conditional branch is not-taken, except for when
+    // there's an incoming backedge, when we guess that the taken case is less
```
romainbrenguier, line 557: back-edge?

src/goto-symex/symex_goto.cpp
```diff
@@ -513,6 +516,80 @@ void goto_symext::symex_goto(statet &state)
...
+    {
+      // Assume branch not taken:
+      symex_transition(state);
+    }
+    else
+    {
+      // Assume branch taken:
+      queue_unreachable_state_at_target();
+      state.guard.add(false_exprt());
+      symex_transition(state);
```
romainbrenguier: happens in both if and else branches so can be simplified

src/goto-symex/symex_goto.cpp
```diff
@@ -547,9 +624,15 @@ void goto_symext::merge_goto(
       "atomic sections differ across branches",
       state.source.pc->source_location);
 
-  if(!goto_state.guard.is_false())
+  // adjust guard, even using guards from unreachable states. This helps to
+  // shrink the state guard if the incoming edge is from a path that was
+  // truncated by config.unwind, config.depth or an assume-false instruction.
+  auto new_guard = std::move(state.guard);
+  new_guard |= goto_state.guard;
```
romainbrenguier: why do you define this new_guard instead of directly modifying state.guard?

romainbrenguier (approved):

src/goto-symex/goto_symex.h
smowton: Not done: calls `symex_transition` which is itself not const... and so on. Can do constification of the `goto_symext` class but not here.

src/goto-symex/symex_goto.cpp
smowton: Added a comment explaining that the `if(!state.reachable)` case overwrites `state.guard`, hence calculating it here but assigning it after the `if(goto_state.reachable)` block.

allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 4f8065f6b48b2894676feb4706ad000cbd7a4aa1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/129298761

smowton:
@peterschrammel results of SV-comp subset: two C benchmarks stop timing out and finish quickly. Regrettably both are false results, but because the benchmark is nuts (may free a NONDET pointer before allocating it) rather than due to an actual bug. One Java benchmark starts timing out, but the difference is entirely in the solver and the VCCs are identical, so I think it's SAT solver performance weirdness rather than pessimisation of the problem it gets passed.

C total times down about 2 minutes; Java total times roughly the same.

Conclusion: this doesn't hurt performance at all, but SV-comp doesn't use ASSUME false to truncate execution much / at all, so we don't see much benefit there. The real benefit comes in when `ASSUME false` truncating a path means that constant propagation can proceed in the wake of a control-flow convergence where it couldn't before -- this happens particularly when used with `--java-assume-no-throws`.

peterschrammel (approved):
Would be good to run it through TG on the tika performance benchmarks to evaluate the effect there.

smowton:
Awaiting performance data from @romainbrenguier

romainbrenguier:
Here is the data I got, running develop against this branch with `--assert-no-exceptions-thrown` the total time went from 3743s to 3430s (8% decrease) and coverage from 6429 goals to 6132 (4.6% decrease)
chart (7)
(axis are time in seconds, for develop x-axis and this branch y-axis)

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9dcc43f8435c7d3cb510d2364bbebea2301d8a61).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/129827614

PR #5098, opened 2019-09-07
Symex: use single-entry, single-exit region analysis to simplify state guards

smowton:
By saving the state guard on entry to an SESE region and the restoring it on exit,
we prevent the guard from growing out of control due to paths being truncated for
breaking unwind or depth limits.


danpoe (approved):

TGWDB:
Closing due to age and lack of progress. Please reopen if you believe this is erroneous and will bring this code up to date/rebase.

PR #5102, opened 2019-09-09
Fix build dependencies of symtab2gb

tautschnig:
Depending on the exact order of building directories, Makefile-based
builds of symtab2gb would sometimes fail as json.a had not been built
(and no rule for building it was present).

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -88,7 +88,8 @@ goto-cc.dir: languages goto-programs.dir linking.dir
 
 memory-analyzer.dir: util.dir goto-programs.dir ansi-c.dir
 
-symtab2gb.dir: util.dir goto-programs.dir json-symtab-language.dir
+symtab2gb.dir: util.dir goto-programs.dir langapi.dir linking.dir \
+               json.dir json-symtab-language.dir
 
 # building for a particular directory
 


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4ca7c9bb9f8a0fbf71f3a8d7341ca0badc36b1f1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/126418525

smowton (approved):

src/Makefile
```diff
@@ -88,7 +88,8 @@ goto-cc.dir: languages goto-programs.dir linking.dir
 
 memory-analyzer.dir: util.dir goto-programs.dir ansi-c.dir
 
-symtab2gb.dir: util.dir goto-programs.dir json-symtab-language.dir
+symtab2gb.dir: util.dir goto-programs.dir langapi.dir linking.dir \
+               json.dir json-symtab-language.dir
```
hannes-steffenhagen-diffblue, line 92: This seems a bit sketchy. `json-symtab-language.dir` should already depend on `langapi.dir` and `json.dir` right? And `linking.dir` should already be included in `goto-programs.dir`?

If not, then I think those should be fixed to include these.

tautschnig: No, there isn't really that kind of dependency: it is perfectly safe to build, e.g., `json-symtab-language`, just the linking will fail. And that's really the only dependencies that are expressed here.

PR #5104, opened 2019-09-11
Make cproverNondetInitialize be inherited [TG-8677]

owen-mc-diffblue:


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4097d449e6850238f01678804a3a14871ed88886).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/127175289

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 62b5999be9a0501c4c18263866571007e806bfcc).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/127332523

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 38a93c44582eae8ef1e590201c7e128ec2319a68).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/127404727

jbmc/src/java_bytecode/java_utils.cpp
```diff
@@ -510,3 +511,41 @@ void set_declaring_class(symbolt &symbol, const irep_idt &declaring_class)
...
+  const irep_idt method_name(class_name + ".cproverNondetInitialize:()V");
+  const auto ret = symbol_table.symbols.find(method_name);
+  if(ret != symbol_table.symbols.end())
+  {
+    return &ret->second;
+  }
+  return nullptr;
+}
+
+const symbolt *find_cprover_nondet_initialize(
```
JohnDumbell: Is `optionalt` usable here? I know it causes problems in certain situations (sadly), so wouldn't be surprised if the answer is no.

jbmc/src/java_bytecode/java_utils.cpp
```diff
@@ -510,3 +511,41 @@ void set_declaring_class(symbolt &symbol, const irep_idt &declaring_class)
...
+  std::string class_name)
+{
+  namespacet ns{symbol_table};
+  const symbolt *ret =
+    find_cprover_nondet_initialize_on_class(symbol_table, class_name);
+
+  while(!ret)
+  {
+    const irep_idt &class_id =
+      to_java_class_type(ns.lookup(class_name).type).get_super_class();
```
JohnDumbell: Worth being more defensive here? I can't remember if `ns.lookup` throws on not found or returns a default. Same with the call on `get_super_class` if the type is dodgy.

JohnDumbell (approved):

jbmc/src/java_bytecode/java_utils.cpp
owen-mc-diffblue: You can't have an `optionalt` of a reference. They say to use a pointer. I could use optionalt of a reference wrapper, I think.

jbmc/src/java_bytecode/java_utils.cpp
owen-mc-diffblue: `ns.lookup` has an INVARIANT that the thing was found. That's the only reason I'm using a namespace instead of the symbol table directly, to avoid having to handle that error myself.

`to_java_class_type` asserts the type has ID_struct. If it doesn't have a super class then `get_super_class()` will return an empty irep_idt, which is handled on the next line.

So actually I think the errors are handled okay, despite how little explicit error handling there is.

jbmc/regression/jbmc-inheritance/cprover_nondet_initialize/inheritanceFromAbstractClass.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+Test.class
+--function Test.inheritanceFromAbstractClass --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar`
+^EXIT=0$
+^SIGNAL=0$
```
thk123, line 5: :no_entry_sign: please add more asserts (e.g. a verification successful) :rescue_worker_helmet:

jbmc/regression/jbmc-inheritance/cprover_nondet_initialize/Test.java
```diff
@@ -0,0 +1,45 @@
...
+class F extends A {
+  public void cproverNondetInitialize() {}
+}
+
+abstract class J {
+  int x;
+
+  public void cproverNondetInitialize() { CProver.assume(x == 42); }
+
+  abstract void neverCalled();
```
thk123: :question: what is the purpose of this method?

jbmc/regression/jbmc-inheritance/cprover_nondet_initialize/inheritanceTwoLevels.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+Test.class
+--function Test.inheritanceTwoLevels --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar`
+^EXIT=0$
+^SIGNAL=0$
```
thk123, line 5: :rescue_worker_helmet:

scripts/pretty-printers/gdb/pretty_printers.py
```diff
@@ -1,4 +1,21 @@
 import gdb
+import json
```
thk123: :no_entry_sign: please don't put prettty printing changes into the same commit as the body of the work

thk123: Or even in this PR??

jbmc/src/java_bytecode/java_utils.cpp
```diff
@@ -510,3 +511,47 @@ void set_declaring_class(symbolt &symbol, const irep_idt &declaring_class)
...
+  const irep_idt method_name(class_name + ".cproverNondetInitialize:()V");
+  const auto ret = symbol_table.symbols.find(method_name);
+  if(ret != symbol_table.symbols.end())
+  {
+    return &ret->second;
+  }
+  return nullptr;
+}
+
+const symbolt *find_cprover_nondet_initialize(
```
thk123: :bulb: great candidate for a unit test

scripts/pretty-printers/gdb/pretty_printers.py
thk123: I've not reviewed these changes on the assumption they are in error!

thk123 (commented):
:no_entry_sign: please add a test for class that implements an interface :heavy_check_mark: 
:no_entry_sign: please add a test for calling a base cproverNondetInitalize from derived :heavy_check_mark: 
:question: does virtual function resolution make sense to test? answered IRL - no
:question: can an interface define `cproverNondetInitalize` - answer IRL - we don't yet support default methods so don't need to worry about this

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 3b87e83685a6f9059af3972548da929f57f1aecb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128040215

jbmc/regression/jbmc-inheritance/cprover_nondet_initialize/Test.java
owen-mc-diffblue: I thought an abstract class needed an abstract method. I'll remove it.

scripts/pretty-printers/gdb/pretty_printers.py
owen-mc-diffblue: Oops

owen-mc-diffblue:
@thk123 I've address most of the review comments. I haven't added tests for interfaces because I don't know how they would implement cproverNondetInitialize (it seems it is possible using default implementations, which came in in java 8, but I'm not sure we support that because I've never heard anyone mention it).

The one thing I think it remaining is unit testing `find_cprover_nondet_initialize`. @thomasspriggs suggested using `resolve_inherited_componentt` instead, which needs a class hierarchy. I have switched to using this in one of the two places, where it is easy to get one. The other place is in the java object factory. I don't think it is worth making every user of the java object factory supply a populated class hierarchy. I think it would be too slow building it each time. I think the best solution is moving my code into `resolve_inherited_component.cpp`, using the same interface, and making it an alternative implementation that doesn't need a class hierarchy.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 366d3eccec53342e799c57b1f0a4f159636cdddb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128291529

thk123:
I still think a test with an interface just to ensure you're going up the correct side of the inheritance tree has value (since not reusing the `resolve_inherited_compnoentt`) but other than that lgtm

owen-mc-diffblue:
With https://github.com/diffblue/cbmc/pull/5118 I should be able to use `resolve_inherited_componentt`, and hence I can get rid of my little utility functions, and even better I won't have to write unit tests for them :smile:

owen-mc-diffblue:
@thk123 I've now merged #5118 and rebased this on top of it, so I can use `resolve_inherited_componentt` in both places. That means I no longer have the utility function which you wanted me to unit test. (I didn't do this just to avoid unit testing them - it's also much nicer code now.) I've also added a test that we find cproverNondetInitialize on a super class even when the base class also implements an interface.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 13eac315352160e55519255e76697fbe1b5178fa).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128470565

thk123 (approved):
Blocking comments addressed

PR #5113, opened 2019-09-17
Add runtime-check replacing instrumentation pass

xbauch:
replacing calls to `__CPROVER_X_check-name` checks with a user-specified
combination of assume/assert (for a user-defined language X).

For example Ada checks can be generated with `gnat2goto`, and this pass allows
us to have more fine-grained and option-able workflow when verifying them. Also this allows us to customise how `check-name` is reported to the user via the assertion description.

### DIFF
--- a/regression/goto-instrument/ada-checks/main.c
+++ b/regression/goto-instrument/ada-checks/main.c
@@ -1,0 +1,10 @@
+void __CPROVER_Ada_Overflow_Check(int condition)
+{
+}
+
+int main()
+{
+  int i = 0;
+  __CPROVER_Ada_Overflow_Check(i > 5);
+  return 0;
+}
--- a/regression/goto-instrument/ada-checks/test.desc
+++ b/regression/goto-instrument/ada-checks/test.desc
@@ -1,0 +1,9 @@
+CORE
+main.c
+--replace-runtime-checks assert-then-assume-first-arg --runtime-checks-language Ada
+^EXIT=10$
+^SIGNAL=0$
+^\[main.\d+\] line \d+ Ada Overflow check i > 5: FAILURE$
+^VERIFICATION FAILED$
+--
+^warning: ignoring
--- a/src/goto-instrument/goto_instrument_parse_options.cpp
+++ b/src/goto-instrument/goto_instrument_parse_options.cpp
@@ -100,6 +100,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "race_check.h"
 #include "reachability_slicer.h"
 #include "remove_function.h"
+#include "replace_runtime_checks.h"
 #include "rw_set.h"
 #include "show_locations.h"
 #include "skip_loops.h"
@@ -1459,6 +1460,17 @@ void goto_instrument_parse_optionst::instrument_goto_program()
     havoc_loops(goto_model);
   }
 
+  if(cmdline.isset("replace-runtime-checks"))
+  {
+    std::string replace_option = cmdline.get_value("replace-runtime-checks");
+    log.status() << "Replacing runtime checks" << messaget::eom;
+    if(!cmdline.isset("runtime-checks-language"))
+      throw invalid_command_line_argument_exceptiont{
+        "unspecified language for runtime checks", "--runtime-checks-language"};
+    std::string language_id = cmdline.get_value("runtime-checks-language");
+    replace_runtime_checks(goto_model, replace_option, language_id);
+  }
+
   if(cmdline.isset("k-induction"))
   {
     bool base_case=cmdline.isset("base-case");
@@ -1782,6 +1794,16 @@ void goto_instrument_parse_optionst::help()
     " --xml-ui                     use XML-formatted output\n"
     " --json-ui                    use JSON-formatted output\n"
     HELP_TIMESTAMP
+    " --replace-runtime-checks <rplc-option>\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    "                              replace runtime checks with assume/assert/both\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    "                              <rplc-option> should be \'assert-first-arg\'\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    "                              \'assert-first-arg\' or \'assert-then-assume-first-arg\'\n"
+    " --runtime-checks-language <lang>\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    "                              language identification for the runtime checks\n"
     "\n";
   // clang-format on
 }
--- a/src/goto-instrument/replace_runtime_checks.cpp
+++ b/src/goto-instrument/replace_runtime_checks.cpp
@@ -1,0 +1,209 @@
+/*******************************************************************\
+
+Module: Replace Runtime Checks
+
+Author: diffblue
+
+\*******************************************************************/
+
+/// \file
+/// Replace Runtime Checks
+
+#include <util/cprover_prefix.h>
+#include <util/prefix.h>
+#include <util/string_constant.h>
+
+#include <langapi/language.h>
+#include <langapi/mode.h>
+
+#include "replace_runtime_checks.h"
+
+/// A simple pass that replace each call to a function with name:
+/// __CPROVER_X_`check-name` with a user-specified sequence of instructions. The
+/// language X can also by specified by the user via
+/// `--runtime-checks-language`. The function is expected to take one argument
+/// (the condition to be checked) and this condition is either asserted or
+/// assumed (or both). The `check-name` is then included in the assertion
+/// description (via source-location).
+struct runtime_checks_replacert
+{
+public:
+  explicit runtime_checks_replacert(
+    goto_modelt &goto_model,
+    const namespacet &ns,
+    const std::string &language_id)
+    : goto_model(goto_model),
+      ns(ns),
+      language_id(language_id),
+      common_prefix(std::string{CPROVER_PREFIX} + language_id)
+  {
+  }
+
+  void replace_with(const std::string &replace_option);
+
+private:
+  goto_modelt &goto_model;
+  const namespacet &ns;
+  const std::string language_id;
+  const std::string common_prefix;
+  irep_idt mode;
+
+  enum class replace_typet
+  {
+    assert_only,
+    assume_only,
+    assert_then_assume
+  };
+
+  enum class runtime_check_typet
+  {
+    overflow,
+    range,
+    division_by_zero
+  };
+
+  void replace_by_option(
+    const exprt &condition,
+    goto_programt &goto_program,
+    goto_programt::targett instruction,
+    const std::string &function_name,
+    const std::string &replace_option);
+
+  runtime_check_typet get_check_type(const std::string &function_name) const
+  {
+    if(function_name == common_prefix + "_Overflow_Check")
+      return runtime_check_typet::overflow;
+    if(function_name == common_prefix + "_Range_Check")
+      return runtime_check_typet::range;
+    if(function_name == common_prefix + "_Division_Check")
+      return runtime_check_typet::division_by_zero;
+    throw invalid_source_file_exceptiont{"unsupported " + language_id +
+                                         " check: " + function_name};
+  }
+
+  replace_typet get_replace_type(const std::string &replace_option) const
+  {
+    if(replace_option == "assert-first-arg")
+      return replace_typet::assert_only;
+    if(replace_option == "assume-first-arg")
+      return replace_typet::assume_only;
+    if(replace_option == "assert-then-assume-first-arg")
+      return replace_typet::assert_then_assume;
+    throw invalid_command_line_argument_exceptiont{
+      "unknown " + language_id + " check replacement option" + replace_option,
+      "--replace-runtime-checks"};
+  }
+
+  friend std::string to_string(runtime_check_typet type)
+  {
+    switch(type)
+    {
+    case runtime_check_typet::overflow:
+      return "Overflow";
+    case runtime_check_typet::range:
+      return "Range";
+    case runtime_check_typet::division_by_zero:
+      return "Division by zero";
+    }
+    UNREACHABLE;
+  }
+};
+
+void runtime_checks_replacert::replace_with(const std::string &replace_option)
+{
+  auto &goto_functions = goto_model.goto_functions;
+  for(auto &goto_function_pair : goto_functions.function_map)
+  {
+    goto_programt &function_body = goto_function_pair.second.body;
+    for(goto_programt::targett instruction = function_body.instructions.begin();
+        instruction != function_body.instructions.end();
+        ++instruction)
+    {
+      if(instruction->is_function_call())
+      {
+        const auto &function_call = instruction->get_function_call();
+        const auto &function = function_call.function();
+
+        const auto &function_name =
+          id2string(to_symbol_expr(function).get_identifier());
+        const auto &function_symbol = ns.lookup(function_name);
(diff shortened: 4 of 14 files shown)


xbauch:
@martin-cs  Hi, Martin. This PR is supposed to be the second part of https://github.com/martin-cs/ASVAT/issues/9. What do you think?

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+  ada_checkst(goto_modelt &goto_model) : goto_model(goto_model)
+  {
+  }
+
+  void replace_with(const std::string &replace_option);
+
+private:
+  goto_modelt &goto_model;
+
+  enum class Replace_Type
```
hannes-steffenhagen-diffblue: :pick: `replace_typet`

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+  {
+    AssertOnly,
```
hannes-steffenhagen-diffblue: :pick: `assert_only`, `assume_only`, `assert_then_assume`

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+    AssumeOnly,
+    AssertThenAssume
+  };
+
+  enum class Ada_Check
```
hannes-steffenhagen-diffblue: naming etc

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+    const exprt &condition,
+    goto_programt &goto_program,
+    goto_programt::targett instruction,
+    const std::string &function_name,
+    const std::string &replace_option);
+
+  Ada_Check get_check_type(const std::string &function_name) const
+  {
+    if(function_name.find("Overflow") != std::string::npos)
+      return Ada_Check::Overflow;
```
hannes-steffenhagen-diffblue: Hm, shouldn't these names be fixed? If that's the case, shouldn't these be exact matches?

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+      return Ada_Check::Range;
+    if(function_name.find("Division") != std::string::npos)
+      return Ada_Check::DivByZero;
+    throw invalid_source_file_exceptiont{"unsupported Ada check: " +
+                                         function_name};
+  }
+
+  Replace_Type get_replace_type(const std::string &replace_option) const
+  {
+    if(replace_option == "assert-first-arg")
```
hannes-steffenhagen-diffblue: What's with the `first-arg`?

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+      return Replace_Type::AssertOnly;
+    if(replace_option == "assume-first-arg")
+      return Replace_Type::AssumeOnly;
+    if(replace_option == "assert-then-assume-first-arg")
+      return Replace_Type::AssertThenAssume;
+    throw invalid_command_line_argument_exceptiont{
+      "unknown Ada check replacement option" + replace_option, "--ada-checks"};
+  }
+
+  std::string type2string(Ada_Check type) const
```
hannes-steffenhagen-diffblue: `Ada_Check` type2string if anything I guess? I think you could also just overload `std::to_string` for this

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+    switch(type)
+    {
+    case Ada_Check::Overflow:
+      return "Overflow";
+    case Ada_Check::Range:
+      return "Range";
+    case Ada_Check::DivByZero:
+      return "Division by zero";
+    }
+    UNREACHABLE;
```
hannes-steffenhagen-diffblue: Funnily enough, this is actually very easily reachable, you just have to call it with stupid arguments (no need to change this, just my attempt at a funny remark)

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1b87ab2c65bce47548d8489752a4f316f5409da6).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128026883

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+void ada_checkst::replace_by_option(
+  const exprt &condition,
+  goto_programt &goto_program,
+  goto_programt::targett instruction,
+  const std::string &function_name,
+  const std::string &replace_option)
+{
+  exprt boolean_condition = condition;
+  while(boolean_condition.id() == ID_typecast)
```
NlightNFotis: Why is this `while`? I thought the `to_typecast_expr` below always simplifies? Is it possible it returns something that has the same structure, and hasn't been structurally reduced?

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+  case Replace_Typet::AssumeOnly:
+    goto_program.insert_after(
+      instruction,
+      goto_programt::make_assumption(
+        boolean_condition, commented_source_location));
+    break;
+  }
+
+  auto skip_instruction = goto_programt::make_skip();
+  instruction->swap(skip_instruction);
```
NlightNFotis: Why is this happening here? 

(I'm not saying it's wrong, I just don't understand what purpose it serves)

NlightNFotis (commented):
It looks good to me, I just have two questions:

src/goto-instrument/ada_checks.cpp
NlightNFotis: Okay, ignore this comment. After I wrote it, I remembered that the `to_typecast_expr` is not even simplifying, it's just a cast. But I still wonder why the need for `while` here.

xbauch: In a C program this is probably unnecessary, but gnat2goto can produce nested typecasts.

src/goto-instrument/ada_checks.cpp
xbauch: Well the idea is that we _replace_ the original call to `__CPROVER_Ada_check` with assert/assume. On the lines above we have added the assertion, and here we change the original function call to a `SKIP`. The only interface for changing an existing instruction I could find is `swap`, so that's what I used.

src/goto-instrument/ada_checks.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+void replace_ada_checks(
+  goto_modelt &goto_model,
+  const std::string &replace_option)
+{
+  ada_checkst ada_checks{goto_model};
+  ada_checks.replace_with(replace_option);
```
hannes-steffenhagen-diffblue: FYI I think you need to call `update` somewhere to fix the location numbers after inserting or removing instructions, unless that was changed (or maybe that's already done by goto-instrument? No idea tbh)

hannes-steffenhagen-diffblue (approved):
Some nitpicks, I think the naming needs to be changed, otherwise LGTM

src/goto-instrument/ada_checks.cpp
xbauch: https://github.com/martin-cs/ASVAT/issues/9 .. That is the exact wording is part of the task specification.

src/goto-instrument/ada_checks.cpp
xbauch: Even if the argument is an `enum class`?

src/goto-instrument/ada_checks.cpp
xbauch: Sure, why not. Changed.

src/goto-instrument/ada_checks.cpp
xbauch: Renamed. Here and above.

src/goto-instrument/ada_checks.cpp
xbauch: I used the `to_string`.

src/goto-instrument/ada_checks.cpp
xbauch: I added the updating, once all check are replaced.

regression/goto-instrument/ada-checks/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+--ada-checks assert-then-assume-first-arg
+^EXIT=10$
+^SIGNAL=0$
+^\[1\] Ada Overflow check: FAILURE$
```
xbauch: The goto-functions line looks like this:
`ASSERT i > 5 // Ada Overflow check`
but somehow the asserted condition is not included in the CBMC output.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: d1ecf617715a82dc8dfed6f09c4b0e6d672b7963).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128144409

src/goto-instrument/ada_checks.cpp
hannes-steffenhagen-diffblue: Yes, hilariously enough. `std::byte` is defined without any named constants but is used to hold any byte value (however, I think treating an enum that isn't explicitly intended to be used as a bitfield or a numeric type like that shouldn't pass code review, hence why this is just a passing remark rather than a genuine warning)

regression/goto-instrument/ada-checks/test.desc
xbauch: Fixed.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7c23fd75c2953de1aa4ca13829c214b8dc2197bf).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128161589

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 94b39beb46e219f383d86a9dbe4f18c7ba683343).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128366180

regression/goto-instrument/ada-checks/main.c
```diff
@@ -0,0 +1,10 @@
+void __CPROVER_Ada_Overflow_Check(int condition)
+{
+}
```
martin-cs, line 3: I think it might be nice to have the typical assert-then-assume body here.

martin-cs (commented):
I had been thinking of using `--generate-function-body` and one of the inlining functions.  I think this would simplify things a bit.

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1759,6 +1771,16 @@ void goto_instrument_parse_optionst::help()
     " --xml-ui                     use XML-formatted output\n"
     " --json-ui                    use JSON-formatted output\n"
     HELP_TIMESTAMP
+    " --replace-runtime-checks <rplc-option>\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    "                              replace runtime checks with assume/assert/both\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    "                              <rplc-option> should be \'assert-first-arg\'\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    "                              \'assert-first-arg\' or \'assert-then-assume-first-arg\'\n"
```
xbauch, line 1803: `assume-first-arg`

xbauch:
@martin-cs The second commit tries to do the same using `generate-function-body` pass. There are several problems with that: 
1) the `generate-function-body` didn't like generating bodies for function that already had bodies
2) the `--function-inline foo` option inlines all calls to functions inside `foo` (and not the calls to `foo`), we therefore have to use `--function-inline main`
3) inlining creates temporary variables for arguments and thus the original asserted expressions are not visible in the CBMC report. The `main` looks like this:
```
main /* main */
        // 0 file main.c line 7 function main
        signed int i;
        // 1 file main.c line 7 function main
        i = 0;
        // 2 file main.c line 8 function main
        LOCATION
        // 3 file main.c line 8 function main
        signed int condition;
        // 4 file main.c line 8 function main
        condition = (signed int)(i > 5);
        // 5 file main.c line 8 function main
        ASSERT (__CPROVER_bool)condition // assertion cast(__CPROVER_Ada_Overflow_Check::condition, 𝔹)
        // 6 file main.c line 8 function main
        ASSUME (__CPROVER_bool)condition
        // 7 file main.c line 8 function main
```
Otherwise it works.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 1def9093cbbe90fecf3458cb25c5a5b282ffd670).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128520921

martin-cs:
> 1. the generate-function-body didn't like generating bodies for function that already had bodies

It should be easy enough to fix this by deleting the bodies first.


> 2. the --function-inline foo option inlines all calls to functions inside foo (and not the calls to foo), 
> we therefore have to use --function-inline main

Ah; sorry.  There are two ways of inlining things, either inline everything *in* the target function or inline every call *to* the target function.  Unfortunately CPROVER in both code and options is remarkably unclear which is which and it looks like I picked the wrong one.

My suggestion was to inline every call *to* the run-time checks.  Inlining everything into `main` won't scale and is kinda scary.

> 3. inlining creates temporary variables for arguments and thus the original asserted expressions 
> are not visible in the CBMC report.

That is annoying.  If the other set of inlining function have this as well then we may have to work around this.

src/goto-instrument/replace_runtime_checks.cpp
```diff
@@ -0,0 +1,209 @@
...
+    assume_only,
+    assert_then_assume
+  };
+
+  enum class runtime_check_typet
+  {
+    overflow,
+    range,
+    division_by_zero
+  };
```
martin-cs, line 63: Do we need to enumerate these here?  IIRC there are a number of others and it would be a little bit annoying to have to do parallel maintenance between this and gnat2goto.

src/goto-instrument/replace_runtime_checks.cpp
```diff
@@ -0,0 +1,209 @@
...
+    const std::string &replace_option);
+
+  runtime_check_typet get_check_type(const std::string &function_name) const
+  {
+    if(function_name == common_prefix + "_Overflow_Check")
+      return runtime_check_typet::overflow;
+    if(function_name == common_prefix + "_Range_Check")
+      return runtime_check_typet::range;
+    if(function_name == common_prefix + "_Division_Check")
+      return runtime_check_typet::division_by_zero;
```
martin-cs, line 79: Is there a way of doing "all checks"

martin-cs (approved):
I think this is a good implementation of this design.  Please have another look and see if the generate body / inline calls option works and produces something more compact.  If not then I am fine for this to be merged.

xbauch:
> My suggestion was to inline every call _to_ the run-time checks. Inlining everything into `main` won't scale and is kinda scary.

If there is such an option, I did not find it.

xbauch:
@martin-cs How do you want to handle this one? I did not find a version of `function-inline` that would inline all calls to a target function. I guess we could write our own.

xbauch:
@martin-cs So I've implemented an inliner that inlines calls to a function, but if we want it to preserve argument values (so that we can see the assertions in the cbmc report), that would be a much bigger change. For that, the version in the first commit would be easier.

TGWDB:
Closing due to age and "Draft" status with lack of progress/update in years. Please reopen if you will update this and progress upon it.

PR #5119, opened 2019-09-18
Java type parser [TG-9756]

NathanJPhillips:
This is a draft PR for the Java type parser.
In order to properly interpret generic parameters from outer classes it is necessary to load the outer class first to find the bounds of the generic parameter. The bytecode parser has thus been split into two phases, one that loads the bytecode just enough to find the outer class and a second phase that uses type information to parse this loaded bytecode after outer classes have been loaded and parsed.

### DIFF
--- a/src/util/parsable_string.h
+++ b/src/util/parsable_string.h
@@ -1,0 +1,100 @@
+// Copyright 2019 Diffblue Limited.
+
+/// \file
+/// A wrapper for strings that facilitates parsing
+
+#ifndef CPROVER_UTIL_PARSABLE_STRING_H
+#define CPROVER_UTIL_PARSABLE_STRING_H
+
+#include <stdexcept>
+#include <string>
+
+class parsable_stringt
+{
+private:
+  const std::string &underlying;
+  std::size_t start_pos;
+  std::size_t after_end_pos;
+
+  parsable_stringt(
+    const std::string &underlying,
+    std::size_t start_pos,
+    std::size_t after_end_pos)
+    : underlying(underlying), start_pos(start_pos), after_end_pos(after_end_pos)
+  {
+  }
+
+public:
+  /// Create an instance containing the whole of the given string
+  /// \param underlying The string to wrap in the input
+  // NOLINTNEXTLINE(runtime/explicit) - loses no information
+  parsable_stringt(const std::string &underlying)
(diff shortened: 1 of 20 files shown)


JohnDumbell:
@NathanJPhillips do you have TODO statements on everything right now that can be improved/needs additional work in the future? I think authors notes before you head out might be the best idea for future reviewers.

src/util/parsable_string.h
```diff
@@ -0,0 +1,100 @@
...
+#include <stdexcept>
+#include <string>
+
+class parsable_stringt
+{
+private:
+  const std::string &underlying;
+  std::size_t start_pos;
+  std::size_t after_end_pos;
+
```
kroening, line 18: Note that this mostly re-invents `std::string_view`.
I would suggest that we pull in an implementation of that, and use it.

NathanJPhillips: First Google result is this one: https://github.com/martinmoene/string-view-lite/blob/master/include/nonstd/string_view.hpp
Seem OK to pull that in?
With others we've renamed to string_viewt, but that will make forward-porting harder when we upgrade - should we leave the name alone?
Where are files like this (Boost licence) kept?

owen-mc-diffblue: I would use an alias, like we do with `optionalt`. I was going to suggest putting the file in the same place as `nonstd::optionalt` but the licence is different, so I'm not sure. Maybe @smowton or @tautschnig knows?

owen-mc-diffblue: That's funny, `nonstd::optional-lite` is also Boost licence. Hmm. It seems it used to be MIT-licenced, and it was re-licenced in June 2018: https://github.com/martinmoene/optional-lite/commit/03f75a98c48eb9e88d04e0995c69eb1baf5e9f2f. It seems that `nonstd::string_view-lite` was also re-licenced at the same time: https://github.com/martinmoene/string-view-lite/commit/3aed8744ee6df0266e79c4a4b568ba9800f2fbf8. If it's easier, perhaps it would be legal to copy it the version just before it was re-licenced?

TGWDB:
Closing due to age and major rebase required indicating lack of progress. Please reopen if you believe this closure is erroneous and you will rebase and update.

PR #5123, opened 2019-09-20
GDB pretty-printers v2

JohnDumbell:
This cleans up the final few bits of the irep pretty-printer that was still outstanding. This is what it now looks like:

pretty_printer_phase2

(That's part of a seal in the background, pay it no mind)

It's a little bit more verbose but far more consistent with how other things look and work in the IDE. Means you can just ctrl-c the value now and it'll work properly, as well as delegating the printing to the dstring pretty-printer, etc. 

Unless bugs crop up this is now considered finished.

### DIFF
--- a/scripts/pretty-printers/gdb/pretty_printers.py
+++ b/scripts/pretty-printers/gdb/pretty_printers.py
@@ -72,12 +72,10 @@ def get_node_value(data_ref):
     if id_value:
         if has_children:
             return "[{0}]".format(id_value)
-        else:
-            return "\"{0}\"".format(id_value)
     elif has_children:
         return "[...]"
 
-    return "\"\""
+    return None
 
 
 def get_id(data_ref):
@@ -148,6 +146,13 @@ class IrepPrettyPrinter:
         we return a single child with the value of the node.
         """
 
+        if has_children_nodes(self.val):
+            if self.clion_representation:
+                yield "ID", self.val["data"]
+            else:
+                yield "ID key", "ID"
+                yield "ID value", self.val["data"]
+
         sub = self.val["sub"]
         sub_count = 0
         item = sub["_M_impl"]["_M_start"]
@@ -158,9 +163,12 @@ class IrepPrettyPrinter:
             iter_item = item.dereference()
 
             if self.clion_representation:
+                item_deref = iter_item["data"].referenced_value()
                 nested_id = get_node_value(iter_item["data"].referenced_value())
                 if nested_id:
                     node_key = "{0}: {1}".format(node_key, nested_id)
+                else:
+                    iter_item = item_deref["data"]
 
                 yield node_key, iter_item
             else:
@@ -188,9 +196,12 @@ class IrepPrettyPrinter:
 
             iter_item = result["second"]
             if self.clion_representation:
-                nested_id = get_node_value(iter_item["data"].referenced_value())
+                item_deref = iter_item["data"].referenced_value()
+                nested_id = get_node_value(item_deref)
                 if nested_id:
                     node_key = "{0}: {1}".format(node_key, nested_id)
+                else:
+                    iter_item = item_deref["data"]
 
                 yield node_key, iter_item
             else:
--- a/scripts/pretty-printers/gdb/pretty_printers.py
+++ b/scripts/pretty-printers/gdb/pretty_printers.py
@@ -1,7 +1,8 @@
-import gdb
 import json
 import os
 from json import JSONDecodeError
+
+import gdb
 
 script_directory = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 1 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9685850d22653763efa8dc469ae0f7b428f60fc4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/128525144

JohnDumbell:
@thk123 Can I get an approval when you have a moment. Nothing too subtle going on here, either it works or breaks everything (and I've been making sure it does the former).

scripts/pretty-printers/gdb/pretty_printers.py
```diff
@@ -236,14 +237,57 @@ def display_hint(self):
         return "string"
 
 
+def untypedef(type_obj):
```
thk123: :pick: document this function

scripts/pretty-printers/gdb/pretty_printers.py
```diff
@@ -236,14 +237,57 @@ def display_hint(self):
...
+
+        if "irept" in hierarchy_types:
+            return IrepPrettyPrinter(val)
+
+    return None
+
+
 # If you change the name of this make sure to change install.py too.
 def load_cbmc_printers():
+    gdb.printing.register_pretty_printer(None, child_of_irept)
```
thk123: :question: why none?

scripts/pretty-printers/gdb/pretty_printers.py
```diff
@@ -13,7 +13,8 @@
         pretty_printer_options = json.load(json_file)
 except JSONDecodeError as e:
     print("Options file at {0} failed to load. Exception: {1}".format(options_path, e.msg))
-
+except Exception as e:
```
thk123, line 16: :question: err -- maybe print the error? is there a reason to expect errors that can be ignored?

scripts/pretty-printers/gdb/pretty_printers.py
```diff
@@ -249,16 +282,19 @@ def display_hint(self):
         return "string"
 
 
-def untypedef(type_obj):
-    if (type_obj.code == gdb.TYPE_CODE_REF or
-            type_obj.code == getattr(gdb, 'TYPE_CODE_RVALUE_REF', None)):
+# If you change the name of this make sure to change install.py too.
+def load_cbmc_printers():
+    gdb.printing.register_pretty_printer(None, child_of_irept)
```
thk123, line 286: :question: reman confused by the `None` here

scripts/pretty-printers/gdb/pretty_printers.py
```diff
@@ -6,6 +6,7 @@
 
 script_directory = os.path.dirname(os.path.abspath(__file__))
 
+# Load up the options in case there's something changed.
```
thk123: :pick: This commit message isn't very helpful:
> Clean-up 

Perhaps:

> Add comments to irep gdb pretty printer

thk123: :pick: I don't think this comment helps readability

scripts/pretty-printers/gdb/pretty_printers.py
```diff
@@ -79,13 +82,27 @@ def get_node_value(data_ref):
 
 
 def get_id(data_ref):
+    """ Return the ID of a node has a string. """
```
thk123: :pick: `as a string` I think

thk123 (commented):
Couple of questions, but this is awesome without a doubt

scripts/pretty-printers/gdb/pretty_printers.py
JohnDumbell: Yes - an options file isn't integral, so if anything goes wrong in the block, even on the import statements, we don't want it to stop the pretty-printer. Might print an error though, just so it isn't silently failing for a long time.

scripts/pretty-printers/gdb/pretty_printers.py
JohnDumbell: So it's a 'global' pretty-printer. GDB allows you to attach pretty-printers to specific file objects, binaries etc, but I never got it working reliably. Making it global means any time something needs to be printed, it'll check to see if this matches.

Not precisely optimal, but right now I don't want to spend the time coaxing it to work (and the benefits are a big question-mark).

thk123: Ohhh I see, for some reason I thought the second param was like the matcher.

thk123 (approved):
Thanks for addressing all the nit picks as well :green_heart:

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 5f741a127d090c1385bb5b81a118b088eb868aa5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/129855318

PR #5127, opened 2019-09-24
Optionally lift clinit calls after Java method conversion

smowton:
This lifts all clinit calls in Java function bodies to the top, thereby
reducing the chances that a clinit method is repeatedly maybe-called (e.g.
due to `if(x) { A.staticfield.set(1); }`, which will only run A.<clinit> if
the branch is taken). If code like this occurs in a loop, for example, the
static initializer could be run by symex on every iteration, creating large
alias sets to account for all the possible initializer calls and sometimes
producing much larger formulas than are necessary.

On the other hand, with this option set, if `x` is definitely false we will
nontheless needlessly run the initializer, so in some circumstances the
option can increase costs.


owen-mc-diffblue (approved):
The tests involving `--show-goto-functions` seem a bit brittle. Would it be possible for the clinits to have a side effect, like incrementing a static field, and assert on the value of that field?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: f46b0fb222714de3410e3c21cd6866b169ba7864).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/129488086

smowton:
Benchmarked this on Tika and found that of 318 methods, 11 were saved from (a 2 minute) timeout, 4 were significant harmed (slowed down by 10s or more), 5 more were significantly improved. Total analysis time down 20%, so I suggest we should merge this, keep it optional and use it as a phase.

owen-mc-diffblue:
I agree with the suggestion to merge this as an option. Whether we use it on some or all phases is something we can experiment with - we just have to make sure we don't forget to use it at all.

jbmc/src/java_bytecode/lift_clinit_calls.cpp
```diff
@@ -0,0 +1,69 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  if(clinit_wrappers_called.empty())
+    return input;
+
+  // 2. Unique:
```
cesaro: I would briefly explain here why this step is necessary.

cesaro (approved):
Looks good. I'm surprised that only 16 in 318 methods improve with this.

allredj:
@smowton Could you please link a ticket, and add to it a note to use it in the phase configuration?

kroening (approved):

jbmc/src/java_bytecode/lift_clinit_calls.cpp
```diff
@@ -0,0 +1,69 @@
...
+            // Replace call with skip:
+            it.mutate() = code_skipt();
+          }
+        }
+      }
+    }
+  }
+
+  if(clinit_wrappers_called.empty())
+    return input;
```
pkesseli, line 47: Since you're explicitly moving all other returns, perhaps:
```
  return std::move(input);
```

pkesseli (approved):

jbmc/src/java_bytecode/lift_clinit_calls.cpp
smowton: I actually mustn't -- when I return a `codet` I must not `move` so that RVO works (otherwise Clang warns, and we have `-Werror`), whereas if I return a type that would require a conversion (e.g. `code_blockt`) then I must use `std::move`.

jbmc/src/java_bytecode/java_bytecode_language.h
```diff
@@ -117,7 +118,12 @@ Author: Daniel Kroening, kroening@kroening.com
   "                              instead of calling the normal static initializer\n"/* NOLINT(*) */ \
   "                              (clinit) method.\n" \
   "                              The argument can be a relative or absolute path to\n"/* NOLINT(*) */ \
-  "                              the file.\n"
+  "                              the file.\n" \
+  " --java-lift-clinit-calls     Lifts clinit calls in function bodies to the top of the\n" /* NOLINT(*) */ \
+  "                              function. This may reduce the overall cost of static\n" /* NOLINT(*) */ \
+  "                              initialisation, but may be unsound if there are\n" /* NOLINT(*) */ \
+  "                              cyclic dependencies between static initializers due\n" /* NOLINT(*) */ \
```
pkesseli, line 125: There are scenarios where this is unsound even without cyclic dependencies:
```
class A {
  static int VALUE;
}

class B {
  static {
    A.VALUE = 10;
  }

  static void foo() {
  }
}
// ...
  public void testInit(int value) {
    if (value > 0) {
      B.foo();
    }
    System.out.println(A.VALUE);
  }
```

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 87017649f8d47a9b42d351b4b837ea9fbc4e01db).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130011481

PR #5129, opened 2019-09-25
Constant propagate string-solver function results

JohnDumbell:
Allows constant propagation to work on a range of general string operations.


src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+
+  array_exprt string_data = s_data_opt->get();
+
+  auto &operands = string_data.operands();
+  for(auto & operand : operands)
+  {
+    int character = std::stoi(id2string(operand.get(ID_value)), nullptr, 16);
+    if (isalpha(character))
+    {
+      char buffer[33];
```
smowton: Justify 33

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+
+  if(!s_data_opt)
+    return false;
+
+  array_exprt string_data = s_data_opt->get();
+
+  auto &operands = string_data.operands();
+  for(auto & operand : operands)
+  {
+    int character = std::stoi(id2string(operand.get(ID_value)), nullptr, 16);
```
smowton: `get(ID_value)` -> `to_constant_expr(...).get_value()`

smowton: Don't parse directly, use `numeric_cast_v` (https://github.com/diffblue/cbmc/blob/7bbef4ae34c408425d1b900a9c186911f7ea90ff/src/util/arith_tools.h#L149)

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+    if (isalpha(character))
+    {
+      char buffer[33];
+      if (to_upper)
+      {
+        if(islower(character))
+        {
+          sprintf(buffer, "%x", toupper(character));
+          operand.set(ID_value, buffer);
```
smowton: Don't set value, use `from_integer` (https://github.com/diffblue/cbmc/blob/7bbef4ae34c408425d1b900a9c186911f7ea90ff/src/util/arith_tools.h#L160)

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+      state,
+      symex_assign,
+      to_ssa_expr(f_l1.arguments().at(0)),
+      new_char_array_length,
+      to_ssa_expr(f_l1.arguments().at(1)),
+      new_char_array);
+  }
+
+  return true;
+}bool goto_symext::constant_propagate_trim(
```
smowton: Line break after `}`

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  const refined_string_exprt &s = to_string_expr(f_l1.arguments().at(2));
+  const auto s_data_opt = try_evaluate_constant_string(state, s.content());
+
+  if(!s_data_opt)
+    return false;
+
+  auto &new_data = f_l1.arguments().at(3);
+  auto &old_data = f_l1.arguments().at(4);
+
+  // If it's a symbol we're using the single character override.
```
smowton: What's the alternative? Write the signatures down here for simplicity

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  if (new_data.id() == ID_symbol)
+  {
+    const auto &oldChar = to_symbol_expr(new_data);
```
smowton: snake case, unless these are intentionally mirroring JDK identifiers

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+    }
+
+    auto new_character = new_char_pointer->get().get(ID_value);
+    auto old_character = old_char_pointer->get().get(ID_value);
+
+    array_exprt string_data = s_data_opt->get();
+    auto &operands = string_data.operands();
+    for(auto &operand : operands)
+    {
+      if(operand.get(ID_value) == old_character)
```
smowton: As before use `arith_tools.h` not direct `ID_value` hacking

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  if(!s_data_opt)
+    return false;
+
+  array_exprt string_data = s_data_opt->get();
+
+  auto &operands = string_data.operands();
+  for(auto it = operands.begin();
+      it != operands.end();)
+  {
+    if(it->get_int(ID_value) == 20)
```
smowton: arith_tools

src/util/simplify_expr.cpp
```diff
@@ -256,10 +298,12 @@ static simplify_exprt::resultt<> simplify_string_compare_to(
 ///
 /// \param expr: the expression to simplify
 /// \param ns: namespace
+/// \param last_index: return the last instead of the first index
 /// \return: the modified expression or an unchanged expression
 static simplify_exprt::resultt<> simplify_string_index_of(
   const function_application_exprt &expr,
-  const namespacet &ns)
+  const namespacet &ns,
+  const bool last_index)
```
smowton: Sounds like it's a bound (like `starting_index`), rename `search_from_end`

src/util/simplify_expr.cpp
```diff
@@ -319,11 +360,30 @@ static simplify_exprt::resultt<> simplify_string_index_of(
...
 
-    it = std::search(
-      std::next(s1_data.operands().begin(), starting_index),
-      s1_data.operands().end(),
-      s2_data.operands().begin(),
-      s2_data.operands().end());
+    if(last_index)
+    {
+      auto it = std::search(
+        std::next(s1_data.operands().rbegin(), starting_index),
```
smowton: This use of `starting_index` looks wrong, surely it should be applied to `rend` instead? Make sure the case where reverse-search uses a starting-index limit is tested.

src/util/simplify_expr.cpp
```diff
@@ -339,26 +399,33 @@ static simplify_exprt::resultt<> simplify_string_index_of(
...
     };
 
-    it = std::find_if(
-      std::next(s1_data.operands().begin(), starting_index),
-      s1_data.operands().end(),
-      pred);
+    if(last_index)
+    {
+      auto it = std::find_if(
+        std::next(s1_data.operands().rbegin(), starting_index),
```
smowton: Similarly this use of starting-index

src/util/simplify_expr.cpp
```diff
@@ -405,6 +472,88 @@ static simplify_exprt::resultt<> simplify_string_char_at(
   return c;
 }
 
+/// Take the passed-in constant string array and lower-case every character.
+static void lower_case_string_expression(array_exprt &string_data)
+{
+  auto &operands = string_data.operands();
+  for(auto &operand : operands)
+  {
+    int character = std::stoi(id2string(operand.get(ID_value)), nullptr, 16);
```
smowton: Use arith-tools

src/util/simplify_expr.cpp
```diff
@@ -405,6 +472,88 @@ static simplify_exprt::resultt<> simplify_string_char_at(
...
+    if(isalpha(character))
+    {
+      char buffer[33];
+      if(isupper(character))
+      {
+        sprintf(buffer, "%x", tolower(character));
```
smowton: Use arith-tools

src/util/simplify_expr.cpp
```diff
@@ -405,6 +472,88 @@ static simplify_exprt::resultt<> simplify_string_char_at(
...
+        operand.set(ID_value, buffer);
+      }
+    }
+  }
+}
+
+/// Simplify String.startsWith function when arguments are constant
```
smowton: Wrong comment

src/util/simplify_expr.cpp
```diff
@@ -405,6 +472,88 @@ static simplify_exprt::resultt<> simplify_string_char_at(
...
+    return simplify_exprt::unchanged(expr);
+  }
+
+  array_exprt second_value = second_value_opt->get();
+
+  // Just lower-case both expressions.
+  lower_case_string_expression(first_value);
+  lower_case_string_expression(second_value);
+
+  // test whether second_value is a prefix of first_value
```
smowton: Given you check the sizes are equal, so this is *not* a prefix check, surely this is just `first_value == second_value`

smowton:
Also, tests: these currently only look at the cases where you *do* expect a change. Make sure to exercise the cases where there should *not* be a simplification / constant propagation, and to exercise optional arguments like the starting-index for String.find / findLast

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+      operands.erase(it);
```
danpoe: Should be `it = operands.erase(it)`

src/goto-symex/goto_symex.cpp
danpoe: We should remove whitespace characters other than space as well. Also a space is `20` in hex not in decimal.

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+    else
+      break;
+  }
+
+  for(auto it = operands.rbegin();
+      it != operands.rend();)
+  {
+    if(it->get_int(ID_value) == 20)
+      operands.erase(it.base()),
+      it++;
```
danpoe: The `erase(it.base())` invalidates the iterator `it`.

src/goto-symex/goto_symex.h
```diff
@@ -610,6 +610,25 @@ class goto_symext
     symex_assignt &symex_assign,
     const function_application_exprt &f_l1);
 
+  /// Attempt to constant propogate trim operations.
```
danpoe: propogate -> propagate

src/goto-symex/goto_symex.h
```diff
@@ -610,6 +610,25 @@ class goto_symext
...
+    const function_application_exprt &f_l1);
+
+  /// Attempt to constant propagate case changes, both upper and lower.
+  bool constant_propagate_case_change(
+    statet &state,
+    symex_assignt &symex_assign,
+    const function_application_exprt &f_l1,
+    bool to_upper);
+
+  /// Attempt to constant proagate character replacement.
```
danpoe, line 673: proagate -> propagate

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,179 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  auto &old_data = f_l1.arguments().at(4);
+
+  // If it's a symbol we're using the single character override.
+  if (new_data.id() == ID_symbol)
+  {
+    const auto &oldChar = to_symbol_expr(new_data);
+    const auto &newChar = to_symbol_expr(old_data);
+
+    const auto new_char_pointer =
+      state.propagation.find(newChar.get_identifier());
```
danpoe: Can use `try_evaluate_constant()` instead which does the lookup in the propagation map and returns an optional `constant_exprt`.

jbmc/regression/jbmc-strings/ConstantEvaluationContains/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+Main.class
+--function Main.contains --show-vcc --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar`
+^Generated [0-9]+ VCC\(s\), 0 remaining after simplification$
```
danpoe: Since `--show-vcc` is used, the test only prints the VCCs and checks that we have 0 remaining VCCs after symex. We also need to check though that constant propagation does compute the correct result. We should leave off `--show-vcc` and check that verification succeeded. (also for the other tests)

Also we should check the negative cases, i.e. here if a string doesn't contain another string.

danpoe (commented):
The tests should also check that constant propagation computes the right result in addition to that there are 0 remaining VCCs after symex.

src/goto-symex/goto_symex.cpp
JohnDumbell: Don't think it makes a difference the way I'm using it here though?

JohnDumbell:
Applied all comments and added more tests.

src/goto-symex/goto_symex.cpp
danpoe: It might work though according to the documentation `it` becomes invalid when calling `erase(it)`. So I'd say it's safer to use the returned iterator.

jbmc/regression/jbmc-strings/ConstantEvaluationContains/Main.java
```diff
@@ -0,0 +1,15 @@
+public class Main {
+    public void contains() {
+        String str = "abcde";
+        assert str.contains("cd");
+    }
+
+    public void notContains() {
+        String str = "abcde";
+        assert !str.contains("cd");
```
smowton: This doesn't really test anything more than `contains`, except that `operator!` is working -- instead I think you should test some cases where `contains` should return false. `assert !str.contains("fgh")`, `assert !str.contains("abcdef")`, and so on -- things that should evaluate to `false` for different reasons.

jbmc/regression/jbmc-strings/ConstantEvaluationContains/notcontains.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+Main.class
+--function Main.notContains --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar`
+^Generated [1-9]\d* VCC\(s\), [1-9]\d* remaining after simplification$
```
smowton: As a non-noprop case I expect these *should* be decided by the simplifier

jbmc/regression/jbmc-strings/ConstantEvaluationEqualsIgnoreCase/Main.java
```diff
@@ -0,0 +1,16 @@
+public class Main {
+    public void equalsIgnoreCase() {
+        assert "abc".equalsIgnoreCase("ABC");
+    }
+
+    public void notEqualsIgnoreCase() {
+        assert !"abc".equalsIgnoreCase("ABC");
```
smowton: As above, test some real negatives, not just verification success  vs. failure

jbmc/regression/jbmc-strings/ConstantEvaluationEqualsIgnoreCase/notequals.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+Main.class
+--function Main.notEqualsIgnoreCase --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar`
+^Generated [1-9]\d* VCC\(s\), [1-9]\d* remaining after simplification$
```
smowton: Should be decided

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,181 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  if(!s_data_opt)
+    return false;
+
+  array_exprt string_data = s_data_opt->get();
+
+  auto &operands = string_data.operands();
+  for(auto &operand : operands)
+  {
+    auto &constant_value = to_constant_expr(operand);
+    char character = numeric_cast_v<int>(constant_value);
```
smowton: Careful, Java chars are 16-bit. It's fine if we only treat ASCII for now, but we should decline to const-prop when we're dealing with e.g. strings of Greek letters, which do have case-change semantics but aren't handled by 8-bit `islower` etc

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,181 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+    if(isalpha(character))
+    {
+      if(to_upper)
+      {
+        if(islower(character))
+        {
+          std::stringstream stream;
+          stream << std::hex << toupper(character);
+          constant_value.set_value(stream.str());
```
smowton: Honestly these are 16-bit ints as far as CBMC is concerned, I'd use `from_integer` rather than hope you get the expected formatting for `constant_exprt`'s value member

smowton: Like this for example: https://github.com/diffblue/cbmc/blob/73e410ab653edffe27e4a74db86169d0a627443f/src/solvers/strings/string_constraint_generator_code_points.cpp#L40

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +681,181 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  const auto s_data_opt = try_evaluate_constant_string(state, s.content());
+
+  if(!s_data_opt)
+    return false;
+
+  auto &new_data = f_l1.arguments().at(3);
+  auto &old_data = f_l1.arguments().at(4);
+
+  // If it's a symbol we're using the replace(Character, Character) overload.
+  // There's another that takes CharSequence which isn't dealt with.
```
smowton: I'm still surprised you're looking for the ID of the *expression* not the type or the signature of the called function, though. What's going on there, why couldn't a symbol_exprt get passed to the other overload?

src/goto-symex/goto_symex.cpp
smowton: +1, it's just luck that it works here based on how `std::vector`'s iterators happen to work (i.e. being pointers to the data array, and the array not usually being reallocated on an erase). How about

```
auto expr_is_space = [](const exprt &e)
{
  const auto &constant_value = to_constant_expr(*it);
  return numeric_cast_v<int>(constant_value) <= ' ';
}

while(!operands.empty() && expr_is_space(operands.front()))
  operands.erase(operands.begin());
while(!operands.empty() && expr_is_space(operands.back()))
  operands.erase(operands.rbegin().base());
```

src/util/simplify_expr.cpp
```diff
@@ -405,6 +474,69 @@ static simplify_exprt::resultt<> simplify_string_char_at(
...
 }
 
+/// Take the passed-in constant string array and lower-case every character.
+static void lower_case_string_expression(array_exprt &string_data)
+{
+  auto &operands = string_data.operands();
+  for(auto &operand : operands)
+  {
+    auto &constant_value = to_constant_expr(operand);
+    char character = numeric_cast_v<int>(constant_value);
```
smowton: Again, watch out that you're restricting to ASCII here and refuse to const-prop anything outside that range, since we don't have a C++ widget with the same case semantics as Character.toUpperCase

src/util/simplify_expr.cpp
```diff
@@ -294,21 +338,17 @@ static simplify_exprt::resultt<> simplify_string_index_of(
 
   const array_exprt &s1_data = s1_data_opt->get();
 
-  if(starting_index >= s1_data.operands().size())
+  const auto search_string_size = s1_data.operands().size();
+  if(starting_index >= search_string_size)
```
smowton: AFAIK still not tests for `starting_index` or perhaps `compareTo` at all?

smowton (commented):
Main comment: the tests for negative cases shouldn't just be positive cases with a `!` at the start, they should test cases where we give a proper negative response, such as `equalsIgnoreCase("abc", "ADC")`, for as many different reasons as our code has reasons to reject the comparison (searched-for string too short and contains a mismatch seem like the obvious ones). The goal is to make sure it doesn't accidentally fire *every* time. Where there are optional index restrictions as for `String.compareTo` make sure there are cases where the restriction is the reason for it not to work -- e.g. `String.lastIndexOf("abc", "abc") == 0` but `String.lastIndexOf("abc", "abc", 1)` == -1

src/goto-symex/goto_symex.cpp
danpoe: Thinking about it the current implementation has n^2 worst case runtime due to the repeated shifting of the vector when erasing elements at the beginning. It might be better to use `find_if()` to get iterators to the first and last non-space characters, and then copy that segment to a new vector (instead of making a copy of the original vector and then modifying it).

src/goto-symex/goto_symex.cpp
JohnDumbell: Fair point!

src/goto-symex/goto_symex.cpp
JohnDumbell: Sure, modified.

jbmc/regression/jbmc-strings/ConstantEvaluationContains/notcontains.desc
JohnDumbell: These cases usually resolve in a VCC but with a directly false guard somewhere. Might be to do with building a trace? Always been like this when I've tested and just assumed that's how things were.

src/goto-symex/goto_symex.cpp
JohnDumbell: Never did in any of my testing, but yes, it's fragile. I've changed it to check for the signature instead, which is also fragile, just in different ways.

I'll get around to adding the other overload if I have time.

src/util/simplify_expr.cpp
JohnDumbell: This was/is right (although yes, I've added a test).

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +669,66 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  }
+
+  for(auto it = operands.rbegin(); it != operands.rend(); it++)
+  {
+    auto &constant_value = to_constant_expr(*it);
+    char character = numeric_cast_v<int>(constant_value);
+    if(character > ' ')
+    {
+      end_iter = it.base();
+      break;
```
romainbrenguier: use `std::find_if`

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +669,66 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+    }
+  }
+
+  // Then copy in what the string would be with whitespace removed.
+  exprt::operandst new_operands(start_iter, end_iter);
+  const constant_exprt new_char_array_length =
+    from_integer(new_operands.size(), length_type);
+
+  const array_typet new_char_array_type(char_type, new_char_array_length);
```
romainbrenguier, line 1149: use curly braces `{}`

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +669,66 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  auto start_iter = operands.begin();
+  auto end_iter = operands.end();
+  for(auto it = operands.begin(); it != operands.end(); it++)
+  {
+    auto &constant_value = to_constant_expr(*it);
+    char character = numeric_cast_v<int>(constant_value);
+    if(character > ' ')
+    {
+      start_iter = it;
+      break;
```
romainbrenguier: use `std::find_if`

src/goto-symex/goto_symex.h
```diff
@@ -610,6 +610,12 @@ class goto_symext
     symex_assignt &symex_assign,
     const function_application_exprt &f_l1);
 
+  /// Attempt to constant propagate trim operations.
+  bool constant_propagate_trim(
+    statet &state,
+    symex_assignt &symex_assign,
+    const function_application_exprt &f_l1);
```
romainbrenguier, line 664: can this be marked `const` or even `static`? :cactus:

jbmc/regression/jbmc-strings/ConstantEvaluationTrim/Main.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public void trimLeft() {
+        String str = "   abc";
+        str = str.trim();
+        assert str.equals("abc");
+    }
+
+    public void trimRight() {
+        String str = "abc   ";
```
romainbrenguier, line 21: can we have tests with other characters than ` ` that would get trimmed away?

src/util/simplify_expr.cpp
```diff
@@ -180,6 +180,48 @@ static simplify_exprt::resultt<> simplify_string_endswith(
...
+/// Simplify String.contains function when arguments are constant
+static simplify_exprt::resultt<> simplify_string_contains(
+  const function_application_exprt &expr,
+  const namespacet &ns)
+{
+  // We want to get both arguments of any starts-with comparison, and
+  // trace them back to the actual string instance. All variables on the
+  // way must be constant for us to be sure this will work.
+  auto &first_argument = to_string_expr(expr.arguments().at(0));
+  auto &second_argument = to_string_expr(expr.arguments().at(1));
```
romainbrenguier, line 192: add a precondition on the size before calling `at`

src/util/simplify_expr.cpp
```diff
@@ -180,6 +180,48 @@ static simplify_exprt::resultt<> simplify_string_endswith(
...
+
+  if(!second_value_opt)
+  {
+    return simplify_exprt::unchanged(expr);
+  }
+
+  const array_exprt &second_value = second_value_opt->get();
+
+  // Is our 'contains' array directly contained in our target.
+  bool includes =
```
romainbrenguier: const

src/goto-symex/goto_symex.h
```diff
@@ -616,6 +616,13 @@ class goto_symext
     symex_assignt &symex_assign,
     const function_application_exprt &f_l1);
 
+  /// Attempt to constant propagate case changes, both upper and lower.
+  bool constant_propagate_case_change(
```
romainbrenguier, line 667: :cactus:

src/goto-symex/goto_symex.cpp
```diff
@@ -670,6 +678,73 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+
+    if(isalpha(character))
+    {
+      if(to_upper)
+      {
+        if(islower(character))
+        {
+          irept hex_value =
+            from_integer(toupper(character), constant_value.type());
+          constant_value.swap(hex_value);
```
romainbrenguier: why swap instead of direct assignment?

src/goto-symex/goto_symex.cpp
```diff
@@ -670,6 +678,73 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  if(!s_data_opt)
+    return false;
+
+  array_exprt string_data = s_data_opt->get();
+
+  auto &operands = string_data.operands();
+  for(auto &operand : operands)
+  {
+    auto &constant_value = to_constant_expr(operand);
+    unsigned char character = numeric_cast_v<int>(constant_value);
```
romainbrenguier: :no_entry_sign: should use `numeric_cast_v<unsigned char>` or if unavailable `narrow`

src/goto-symex/goto_symex.cpp
```diff
@@ -670,6 +678,73 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+    if(character >= 128)
+      return false;
+
+    if(isalpha(character))
+    {
+      if(to_upper)
+      {
+        if(islower(character))
+        {
+          irept hex_value =
```
romainbrenguier: why hex_value? this is actually an exprt

src/util/simplify_expr.cpp
```diff
@@ -361,11 +359,38 @@ static simplify_exprt::resultt<> simplify_string_index_of(
...
-    it = std::search(
-      std::next(s1_data.operands().begin(), starting_index),
-      s1_data.operands().end(),
-      s2_data.operands().begin(),
-      s2_data.operands().end());
+    if(search_from_end)
+    {
+      auto starting_offset = 0;
+      if(starting_index > 0)
+        starting_offset = search_string_size - starting_index;
```
romainbrenguier: Use ternary if to define starting_offset and make it const. Avoid using `auto` when the type is not explicit.

src/util/simplify_expr.cpp
```diff
@@ -361,11 +359,38 @@ static simplify_exprt::resultt<> simplify_string_index_of(
...
+        s2_data.operands().rbegin(),
+        s2_data.operands().rend());
+
+      if(it != s1_data.operands().rend())
+      {
+        // Zero-indexed arrays, so -1.
+        return from_integer(
+          std::distance(s1_data.operands().begin(), it.base()) - 1,
+          expr.type());
+      }
```
romainbrenguier: should return `-1` in the other case no? :tangerine:

src/util/simplify_expr.cpp
```diff
@@ -361,11 +359,38 @@ static simplify_exprt::resultt<> simplify_string_index_of(
...
+    {
+      auto it = std::search(
+        std::next(s1_data.operands().begin(), starting_index),
+        s1_data.operands().end(),
+        s2_data.operands().begin(),
+        s2_data.operands().end());
+
+      if(it != s1_data.operands().end())
+        return from_integer(
+          std::distance(s1_data.operands().begin(), it), expr.type());
```
romainbrenguier, line 397: :tangerine:

romainbrenguier (commented):
I only reviewed until `Add String.lastIndexOf(...)` but some of the comments could apply to the following commits as well

src/goto-symex/goto_symex.cpp
```diff
@@ -739,6 +743,73 @@ bool goto_symext::constant_propagate_case_change(
...
+  auto &old_data = f_l1.arguments().at(3);
+
+  if(f_l1.source_location().is_nil())
+    return false;
+
+  irep_idt target_overload = f_l1.source_location().get_function();
+
+  // If it's a symbol we're using the replace(Character, Character) overload.
+  // There's another that takes CharSequence which isn't dealt with.
+  if(target_overload == "java::java.lang.String.replace:(CC)Ljava/lang/String;")
```
romainbrenguier: :no_entry_sign: check the type of the arguments instead

src/goto-symex/goto_symex.cpp
```diff
@@ -739,6 +743,73 @@ bool goto_symext::constant_propagate_case_change(
...
+
+  auto &new_data = f_l1.arguments().at(4);
+  auto &old_data = f_l1.arguments().at(3);
+
+  if(f_l1.source_location().is_nil())
+    return false;
+
+  irep_idt target_overload = f_l1.source_location().get_function();
+
+  // If it's a symbol we're using the replace(Character, Character) overload.
```
romainbrenguier: `replace(char, char)`

romainbrenguier (changes requested):

jbmc/regression/jbmc-strings/ConstantEvaluationEqualsIgnoreCase/notequals.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+Main.class
+--function Main.mismatch --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar`
+^Generated [0-9]+ VCC\(s\), 0 remaining after simplification$
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
```
romainbrenguier: :pick: missing new line

src/util/simplify_expr.cpp
```diff
@@ -480,6 +480,74 @@ static simplify_exprt::resultt<> simplify_string_char_at(
...
+
+  array_exprt second_value = second_value_opt->get();
+
+  // Just lower-case both expressions.
+  if(
+    !lower_case_string_expression(first_value) ||
+    !lower_case_string_expression(second_value))
+    return simplify_exprt::unchanged(expr);
+
+  bool is_equal = first_value == second_value;
```
romainbrenguier, line 552: const or can be inlined in the next expression

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +669,57 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+
+  if(!s_data_opt)
+    return false;
+
+  const array_exprt &string_data = s_data_opt->get();
+  auto &operands = string_data.operands();
+
+  auto is_not_whitespace = [](const exprt &expr) {
+    auto &constant_value = to_constant_expr(expr);
+    char character = numeric_cast_v<int>(constant_value);
```
romainbrenguier: numeric_cast_v<char> if possible otherwise use narrow, but actually the next operation doesn't require a char and could directly be done on an mp_integer

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +669,57 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  const auto s_data_opt = try_evaluate_constant_string(state, s.content());
+
+  if(!s_data_opt)
+    return false;
+
+  const array_exprt &string_data = s_data_opt->get();
+  auto &operands = string_data.operands();
+
+  auto is_not_whitespace = [](const exprt &expr) {
+    auto &constant_value = to_constant_expr(expr);
```
romainbrenguier: :pick: inline below

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +669,57 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+  const auto &length_type = f_type.domain().at(0);
+  const auto &char_type = to_pointer_type(f_type.domain().at(1)).subtype();
+
+  const refined_string_exprt &s = to_string_expr(f_l1.arguments().at(2));
+  const auto s_data_opt = try_evaluate_constant_string(state, s.content());
+
+  if(!s_data_opt)
+    return false;
+
+  const array_exprt &string_data = s_data_opt->get();
```
romainbrenguier: :pick: used only once, can be inlined below

src/goto-symex/goto_symex.cpp
```diff
@@ -665,3 +669,57 @@ bool goto_symext::constant_propagate_delete_char_at(
...
+
+  // Note where the trim would start/end.
+  auto end_iter =
+    std::find_if(operands.rbegin(), operands.rend(), is_not_whitespace);
+  auto start_iter =
+    std::find_if(operands.begin(), operands.end(), is_not_whitespace);
+
+  // If we haven't found anything, just want to copy the whole array.
+  if(start_iter == operands.end())
+    start_iter = operands.begin();
```
romainbrenguier: :no_entry_sign: this seems wrong, if there is no character which is not a whitespace, then the result should be empty

jbmc/regression/jbmc-strings/ConstantEvaluationTrim/Main.java
romainbrenguier: can you also add a test with a string which is only white spaces

romainbrenguier (dismissed):

src/goto-symex/goto_symex.cpp
smowton: I don't think this should block, either one achieves what's needed

romainbrenguier: that's java specific code in symex

smowton: Doh, right, sure. Use the types, as Romain said.

src/goto-symex/goto_symex.cpp
JohnDumbell: Where would those go? (And curious about why)

src/util/simplify_expr.cpp
JohnDumbell: Would think so, but no, the other case works fine without it. Assume it's notched along +1 due to the way the reverse iterator works.

Edit: Yeah, something's off with the way the reverse iterator works in this instance, swapped it to use `find_end` instead.

JohnDumbell:
Added more tests, added in both character/string replace, did general clean-up pass. If you want any more tests please just list the specific ones.

src/goto-symex/goto_symex.cpp
romainbrenguier: `const array_typet new_char_array_type{char_type, new_char_array_length};`
it's the preferred way of constructing an object, here are some reasons why  https://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 2fe2434b914965934e68be7dce4f12e713c9a5d2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130009103

jbmc/regression/jbmc-strings/ConstantEvaluationEqualsIgnoreCase/pi.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+Main.class
+--function Main.pi --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar`
+^Generated [1-9]\d* VCC\(s\), [1-9]\d* remaining after simplification$
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
```
smowton, line 9: Might want to note *why* this one doesn't work, either here or in a comment in the java file

jbmc/regression/jbmc-strings/ConstantEvaluationReplace/Main.java
```diff
@@ -0,0 +1,90 @@
...
+        assert str.equals("abcde");
+    }
+
+    public void nopropReplace(String str) {
+        str = str.replace('d', 'c');
+        assert str.equals("abcec");
+    }
+
+    public void stringReplace() {
+        String str = "abded";
```
smowton, line 38: Duplicate of `replace`

jbmc/regression/jbmc-strings/ConstantEvaluationReplace/Main.java
```diff
@@ -0,0 +1,90 @@
...
+        assert str.equals("fghjded");
+    }
+
+    public void stringMultiReplace() {
+        String str = "abcdab";
+        str = str.replace("ab", "gh");
+        assert str.equals("ghcdgh");
+    }
+
+    public void stringMultiLargerReplace() {
```
smowton: Larger and Smaller names the wrong way around here

jbmc/regression/jbmc-strings/ConstantEvaluationToLowerCase/pi.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+Main.class
+--function Main.pi --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar`
+^Generated [1-9]\d* VCC\(s\), [1-9]\d* remaining after simplification$
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
```
smowton, line 9: Again mention the missing support, either here or inline in the java file

jbmc/regression/jbmc-strings/ConstantEvaluationToUpperCase/pi.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+Main.class
+--function Main.pi --cp `../../../../scripts/format_classpath.sh . ../../../lib/java-models-library/target/core-models.jar`
+^Generated [1-9]\d* VCC\(s\), [1-9]\d* remaining after simplification$
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
```
smowton, line 7: Surprising! Pi *should* be affected by the real JDK method; is our models lib wrong here?

src/goto-symex/goto_symex.cpp
```diff
@@ -922,3 +938,225 @@ bool goto_symext::constant_propagate_set_char_at(
...
+  if(!s_data_opt)
+    return false;
+
+  auto &new_data = f_l1.arguments().at(4);
+  auto &old_data = f_l1.arguments().at(3);
+
+  array_exprt characters_to_find(s_data_opt->get().type());
+  array_exprt characters_to_replace(s_data_opt->get().type());
+
+  // Two main ways to perform a replace: characters or strings.
```
smowton, line 1025: Main ways, or ways? Is there a third unusual one?

src/goto-symex/goto_symex.cpp
```diff
@@ -922,3 +938,225 @@ bool goto_symext::constant_propagate_set_char_at(
...
+    {
+      return {};
+    }
+
+    characters_to_find.operands().emplace_back(old_char_pointer->get());
+    characters_to_replace.operands().emplace_back(new_char_pointer->get());
+  }
+  else
+  {
+    auto &new_char_array = to_string_expr(new_data);
```
smowton, line 1043: INVARIANT that the types are as you expect (String, String by the sounds of it)

src/goto-symex/goto_symex.cpp
smowton: This actually is producing an empty string, but it's not obvious (`end_iter` must have traversed the whole string too, and so equals `operands.begin()`). Suggest instead do

```
exprt::operandst new_operands;
if(start_iter != operands.end())
  new_operands = exprt::operandst{start_iter, end_iter.base()};
```

src/goto-symex/goto_symex.cpp
```diff
@@ -922,3 +938,225 @@ bool goto_symext::constant_propagate_set_char_at(
...
+  };
+
+  // Note the points where a trim would trim too.
+  auto &operands = s_data_opt->get().operands();
+  auto end_iter =
+    std::find_if(operands.rbegin(), operands.rend(), is_not_whitespace);
+  auto start_iter =
+    std::find_if(operands.begin(), operands.end(), is_not_whitespace);
+
+  // If our entire string is empty, our end_iter will be at begin(), so just
```
smowton: "If our entire string is empty" -> "If the entire string is whitespace"

src/util/simplify_expr.cpp
```diff
@@ -319,11 +361,40 @@ static simplify_exprt::resultt<> simplify_string_index_of(
 
     const array_exprt &s2_data = s2_data_opt->get();
 
-    it = std::search(
-      std::next(s1_data.operands().begin(), starting_index),
-      s1_data.operands().end(),
-      s2_data.operands().begin(),
-      s2_data.operands().end());
+    // Searching for empty string is a special case and is simply the
+    // first match you find. This needs to take into account starting position
```
smowton: "first match you find" -> "always found at the first searched position"?

smowton (approved):
Non-blocking suggestions

jbmc/regression/jbmc-strings/ConstantEvaluationReplace/Main.java
JohnDumbell: One takes 2 characters, other takes 2 strings.

JohnDumbell:
@romainbrenguier Mind having another look and seeing if you're blocking issue is dealt with?

src/goto-symex/goto_symex.cpp
```diff
@@ -922,3 +926,55 @@ bool goto_symext::constant_propagate_set_char_at(
...
   return true;
 }
+
+bool goto_symext::constant_propagate_trim(
+  statet &state,
+  symex_assignt &symex_assign,
+  const function_application_exprt &f_l1)
+{
+  const auto &f_type = to_mathematical_function_type(f_l1.function().type());
+  const auto &length_type = f_type.domain().at(0);
```
romainbrenguier, line 1118: Add a precondition for `f_type.domain().size() == 2`

src/goto-symex/goto_symex.cpp
```diff
@@ -922,3 +926,55 @@ bool goto_symext::constant_propagate_set_char_at(
...
+
+  if(!s_data_opt)
+    return false;
+
+  auto is_not_whitespace = [](const exprt &expr) {
+    auto character = numeric_cast_v<unsigned int>(to_constant_expr(expr));
+    return character > ' ';
+  };
+
+  // Note the points where a trim would trim too.
```
romainbrenguier, line 1132: "to"?

jbmc/regression/jbmc-strings/ConstantEvaluationToUpperCase/pi.desc
romainbrenguier: the builtin upper_case and lower_case are only correct for a (small) subset of unicode (I think ascii and latin-1 supplement) so this is incorrect but expected. It sould be marked as knownbug.

src/goto-symex/goto_symex.cpp
```diff
@@ -927,6 +935,68 @@ bool goto_symext::constant_propagate_set_char_at(
...
 }
 
+bool goto_symext::constant_propagate_case_change(
+  statet &state,
+  symex_assignt &symex_assign,
+  const function_application_exprt &f_l1,
+  bool to_upper)
+{
+  const auto &f_type = to_mathematical_function_type(f_l1.function().type());
+  const auto &length_type = f_type.domain().at(0);
```
romainbrenguier, line 949: precondition

src/goto-symex/goto_symex.cpp
```diff
@@ -997,6 +1001,114 @@ bool goto_symext::constant_propagate_case_change(
   return true;
 }
 
+bool goto_symext::constant_propagate_replace(
+  statet &state,
+  symex_assignt &symex_assign,
+  const function_application_exprt &f_l1)
+{
+  const auto &f_type = to_mathematical_function_type(f_l1.function().type());
+  const auto &length_type = f_type.domain().at(0);
```
romainbrenguier, line 1010: precondition

src/goto-symex/goto_symex.h
```diff
@@ -667,6 +667,12 @@ class goto_symext
     const function_application_exprt &f_l1,
     bool to_upper);
 
+  /// Attempt to constant proagate character replacement.
+  bool constant_propagate_replace(
+    statet &state,
+    symex_assignt &symex_assign,
+    const function_application_exprt &f_l1);
```
romainbrenguier, line 677: :cactus:

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: b5d8053ce0ab57c683120538a124b5535ad4a0bf).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130056711

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: bdde6756a248cdc469916ab1bfe8f0698730416a).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130408178

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: db1f84488305fa3578f61cdc4ebdd405a44c28ff).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130783671

PR #5136, opened 2019-09-30
Fix end iterator dereference

LAJW:
Detected by VS2017 Debug mode
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own. (There are none)

The reason why it worked was that it's unlikely for the end iterator to point to a memory address with the exact required value, so it was skipped over in the while loop.

### DIFF
--- a/src/goto-instrument/cover_instrument_other.cpp
+++ b/src/goto-instrument/cover_instrument_other.cpp
@@ -75,10 +75,19 @@ void cover_instrument_end_of_function(
   const irep_idt &function_id,
   goto_programt &goto_program)
 {
-  auto if_it = goto_program.instructions.end();
-  while(!if_it->is_function_call())
-    if_it--;
-  if_it++;
+  const auto last_function_call = std::find_if(
+    goto_program.instructions.rbegin(),
+    goto_program.instructions.rend(),
+    [](const goto_programt::instructiont &instruction) {
+      return instruction.is_function_call();
+    });
+  INVARIANT(
+    last_function_call != goto_program.instructions.rend(),
+    "Goto program should have at least one function call");
+  INVARIANT(
+    last_function_call != goto_program.instructions.rbegin(),
+    "Goto program shouldn't end with a function call");
+  const auto if_it = last_function_call.base();
   const std::string &comment =
     "additional goal to ensure reachability of end of function";
   goto_program.insert_before_swap(if_it);
--- a/unit/goto-instrument/cover_instrument.cpp
+++ b/unit/goto-instrument/cover_instrument.cpp
@@ -1,0 +1,28 @@
+/*******************************************************************\
+
+Module: Tests for coverage instrumentation
+
+Author: Diffblue Ltd
+
+\*******************************************************************/
+
+#include <goto-instrument/cover_instrument.h>
+#include <testing-utils/use_catch.h>
+
+TEST_CASE("cover_intrument_end_of_function", "[core]")
+{
+  // Arrange
+  goto_programt goto_program{};
+  goto_program.add(
+    goto_programt::make_function_call(code_function_callt({}, {})));
+  goto_program.add(
+    goto_programt::make_function_call(code_function_callt({}, {})));
+  goto_program.add(goto_programt::make_return());
+  // Act
+  cover_instrument_end_of_function("foo", goto_program);
+  // Assert
+  REQUIRE(goto_program.instructions.size() == 4);
+  const auto newly_inserted = std::next(goto_program.instructions.begin(), 2);
+  REQUIRE(newly_inserted->is_assert());
+  REQUIRE(newly_inserted->source_location.get_function() == "foo");
+}
--- a/unit/Makefile
+++ b/unit/Makefile
@@ -16,6 +16,7 @@ SRC += analyses/ai/ai.cpp \
        analyses/does_remove_const/is_type_at_least_as_const_as.cpp \
        big-int/big-int.cpp \
        compound_block_locations.cpp \
+       goto-instrument/cover_instrument.cpp \
        goto-instrument/cover/cover_only.cpp \
        goto-programs/goto_model_function_type_consistency.cpp \
        goto-programs/goto_program_assume.cpp \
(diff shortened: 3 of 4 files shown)


smowton (approved):

src/goto-instrument/cover_instrument_other.cpp
```diff
@@ -75,10 +75,19 @@ void cover_instrument_end_of_function(
...
+    [](const goto_programt::instructiont &instruction) {
+      return instruction.is_function_call();
+    });
+  INVARIANT(
+    last_function_call != goto_program.instructions.rend(),
+    "Goto program should have at least one function call");
+  INVARIANT(
+    last_function_call != goto_program.instructions.rbegin(),
+    "Goto program shouldn't end with a function call");
+  const auto if_it = std::next(last_function_call.base());
```
danpoe: Should this be `std::prev`?

src/goto-instrument/cover_instrument_other.cpp
```diff
@@ -75,10 +75,19 @@ void cover_instrument_end_of_function(
...
-  if_it++;
+  const auto last_function_call = std::find_if(
+    goto_program.instructions.rbegin(),
+    goto_program.instructions.rend(),
+    [](const goto_programt::instructiont &instruction) {
+      return instruction.is_function_call();
+    });
+  INVARIANT(
+    last_function_call != goto_program.instructions.rend(),
+    "Goto program should have at least one function call");
```
danpoe, line 86: It looks like the reason that there is always a function call is that this function is only called for the entry point, which always calls the main function. We should probably say this in the invariant message.

Also nit: invariant messages should start with a lowercase letter.

src/goto-instrument/cover_instrument_other.cpp
LAJW: No, because I called `.base()` to convert from reverse iterator to forward iterator.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4bc5c00cb0a5b0beb275fb888226a90a164d67fb).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/129704556

src/goto-instrument/cover_instrument_other.cpp
danpoe: Yes, but `.base()` gives an iterator one to the right to the element that dereferencing the reverse iterator would yield. Thus to get a forward iterator to the same position we need to go back again via `std::prev()`.

thk123 (commented):
:no_entry_sign: Given that this didn't work and is uncovered by existing tests, please add a test.

Adding a test looks challenging - this code is only exercised using a test-gen option: `cover-traces-must-terminate`. Probably the code shouldn't live here, but in the mean time, perhaps a unit test can be written

It will also resolve whether it should be `std::prev` or `std::next`

LAJW:
@thk123 @danpoe Unit tests added. `std::next` removed. It is indeed unnecessary. `base()` "offsets" the iterator.

unit/goto-instrument/cover_instrument.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+TEST_CASE("cover_intrument_end_of_function")
+{
+  // Arrange
+  goto_programt goto_program{};
+  goto_program.add(
+    goto_programt::make_function_call(code_function_callt({}, {})));
+  goto_program.add(
+    goto_programt::make_function_call(code_function_callt({}, {})));
+  goto_program.add(goto_programt::make_return());
```
danpoe, line 20: Suggest to make the goto program end with an `END_FUNCTION` instruction to make it more well-formed.

danpoe (approved):
Looks good to me modulo fixes to the invariant messages. Indeed `std::prev` is unnecessary too since we want the instruction after the function call, not the function call itself.

unit/goto-instrument/cover_instrument.cpp
```diff
@@ -0,0 +1,20 @@
+#include <goto-instrument/cover_instrument.h>
+#include <testing-utils/use_catch.h>
+
+TEST_CASE("cover_intrument_end_of_function")
+{
+  // Arrange
```
thk123, line 14: :pick: I'd add one for no function call (check for the invariant)

unit/goto-instrument/cover_instrument.cpp
```diff
@@ -0,0 +1,20 @@
...
+  goto_program.add(
+    goto_programt::make_function_call(code_function_callt({}, {})));
+  goto_program.add(
+    goto_programt::make_function_call(code_function_callt({}, {})));
+  goto_program.add(goto_programt::make_return());
+  // Act
+  cover_instrument_end_of_function("foo", goto_program);
+  // Assert
+  REQUIRE(goto_program.instructions.size() == 4);
+  const auto newly_inserted = std::next(goto_program.instructions.begin(), 2);
```
thk123, line 25: :pick: Personally would find it easier to see that this is being inserted after the second function call phrased like this:

```
const auto newly_inserted = goto_program.instructions[2]
```

though I can see this is the same.

thk123 (approved):

unit/goto-instrument/cover_instrument.cpp
LAJW: But then it crashes the test? I'd have to start throwing exceptions to be able to test that, no? I really would want to avoid adding more of these "failed as expected" tests which pollute the output.

unit/goto-instrument/cover_instrument.cpp
LAJW: can't do, it's a `std::list` no index operator here :cry:

smowton:
This seems to be consuming quite a bit of time and effort. In cases like this where the change is quite easy to verify by eye, do we really need to write new tests just because the original author neglected to? That seems orthogonal to the change being made, and it strikes me we would prioritise "write code for this apparently working but untested code" quite low at the moment.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 320eeacb32a3df9df34524c2b5361162466f6306).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130261710

PR #5138, opened 2019-10-01
Add lexical loop analysis

smowton:
This gives a cheaper loop analysis than `natural_loopst` at the expense of missing some loops. It should identify loops that are generated by the C frontend, and apparently most that are generated by `javac` as well.


chrisr-diffblue (approved):
Looks reasonable.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 7787803c9be6019dbc8d8683657d186ade44849f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/129886909

regression/goto-instrument/lexical-loops8/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+test.c
+--show-lexical-loops
+^2 is head of \{ 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 \(backedge\) \}$
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+This checks that a loop with a break edge and a continue edge nontheless
```
danpoe, line 10: nontheless -> nonetheless

src/analyses/lexical_loops.h
```diff
@@ -0,0 +1,175 @@
...
+Module: Compute lexical loops in a goto_function
+
+Author: Diffblue Ltd
+
+\*******************************************************************/
+
+/// \file
+/// Compute lexical loops in a goto_function.
+///
+/// A lexical loop is a block of (lexically) contiguous GOTO program
```
danpoe, line 12: Can you define lexically contiguous here and say that loops can have "holes"?

danpoe (approved):

src/analyses/loop_analysis.h
```diff
@@ -0,0 +1,181 @@
...
+  bool insert_instruction(const T instruction)
+  {
+    return loop_instructions.insert(instruction).second;
+  }
+
+private:
+  parent_analysist &loop_analysis;
+};
+
+template <class P, class T>
```
hannes-steffenhagen-diffblue, line 99: The template parameters should be documented (I know they are in `lexical_loops.h` but that's not immediately visible from here).

src/analyses/loop_analysis.h
```diff
@@ -0,0 +1,181 @@
...
+  /// instruction does not alter loop structure; if it does they must discard
+  /// and recompute the related \ref parent_analysist instance.
+  /// \return true if the instruction is new
+  bool insert_instruction(const T instruction)
+  {
+    return loop_instructions.insert(instruction).second;
+  }
+
+private:
+  parent_analysist &loop_analysis;
```
hannes-steffenhagen-diffblue, line 96: I'm not entirely clear why the `parent_analysis` needs to be referenced in the loop? Seems like it's only used for `contains` which could just as well be implemented by this class

src/analyses/loop_analysis.h
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  /// Returns true if \p instruction is in this loop
+  bool contains(const T instruction) const
+  {
+    return loop_analysis.loop_contains(*this, instruction);
+  }
+
+  /// Get the \ref parent_analysist analysis this loop relates to
+  const parent_analysist &get_loop_analysis() const
```
hannes-steffenhagen-diffblue, line 49: This isn't referenced anywhere, and I'm not entirely clear when/how this would be useful

src/analyses/loop_analysis.h
```diff
@@ -0,0 +1,181 @@
...
+  // back-pointers on copy/move.
+  loop_analysist(const loop_analysist &) = delete;
+  loop_analysist(loop_analysist &&) = delete;
+  loop_analysist &operator=(const loop_analysist &) = delete;
+  loop_analysist &operator=(loop_analysist &&) = delete;
+};
+
+/// Print all natural loops that were found
+template <class P, class T>
+void loop_analysist<P, T>::output(std::ostream &out) const
```
hannes-steffenhagen-diffblue, line 137: This appears to be the only reason there even is a parent class, but this could just as well be implemented as a free function

```
template<typename P, typename T> void output_loop_map(std::ostream &out, const loop_mapt<P, T>& loop_map);
```

or

```
template<typename LoopAnalysis>
void output_loop_analysis(std::ostream& out, const LoopAnalysis&);
```

src/analyses/lexical_loops.h
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  lexical_loops_templatet() = default;
+
+  explicit lexical_loops_templatet(P &program)
+  {
+    compute(program);
+  }
+
+  bool all_cycles_in_lexical_loop_form() const
```
hannes-steffenhagen-diffblue, line 64: This also doesn't seem to be used anywhere in this PR. Do you have a usage in mind for this?

src/analyses/loop_analysis.h
smowton: This is for when we move to a basic block representation, in which case we'll need to refer back to the analysis to get the basic-block graph (or else carry many more auxiliary references around in each loop structure)

src/analyses/loop_analysis.h
smowton: Really only provided for convenience considering we need it to provide `contains` etc. Getting from an analysis data structure back to the analysis itself seems generally useful

src/analyses/loop_analysis.h
smowton: I wanted the common class to unify the read-side interface (contans, is-header, the loop map etc); the output function was just a useful coincidence

src/analyses/lexical_loops.h
smowton: Yes, when we use it in symex we'll want to check if the function contains cycles that aren't lexical loops

PR #5143, opened 2019-10-02
SMT2 parser: use hash tables instead of if() ... else...

kroening:
This refactors the SMT2 parser to use hash tables instead of chains of `if()...else if()...` for commands, sorts and expressions.

This enables extensions, and may have a performance benefit as the number of commands/expressions/sorts grows.

### DIFF
--- a/src/solvers/smt2/smt2_parser.h
+++ b/src/solvers/smt2/smt2_parser.h
@@ -10,6 +10,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #define CPROVER_SOLVERS_SMT2_SMT2_PARSER_H
 
 #include <map>
+#include <unordered_map>
 
 #include <util/mathematical_types.h>
 #include <util/std_expr.h>
@@ -22,6 +23,7 @@ public:
   explicit smt2_parsert(std::istream &_in)
     : exit(false), smt2_tokenizer(_in), parenthesis_level(0)
   {
+    setup_commands();
   }
 
   virtual ~smt2_parsert() = default;
@@ -82,10 +84,6 @@ protected:
   std::size_t parenthesis_level;
   smt2_tokenizert::tokent next_token();
 
-  void command_sequence();
-
-  virtual void command(const std::string &);
-
   // for let/quantifier bindings, function parameters
   using renaming_mapt=std::map<irep_idt, irep_idt>;
   renaming_mapt renaming_map;
@@ -116,7 +114,6 @@ protected:
     }
   };
 
-  void ignore_command();
   exprt expression();
   exprt function_application();
   exprt function_application_ieee_float_op(
@@ -144,6 +141,14 @@ protected:
 
   /// Apply typecast to unsignedbv to given expression
   exprt cast_bv_to_unsigned(const exprt &);
+
+  // hashtable for all commands
+  std::unordered_map<std::string, std::function<void()>> commands;
+
+  void command_sequence();
+  void command(const std::string &);
+  void ignore_command();
+  void setup_commands();
 };
 
 #endif // CPROVER_SOLVERS_SMT2_SMT2_PARSER_H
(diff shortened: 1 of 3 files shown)


allredj (commented):
:warning:
This PR failed Diffblue compatibility checks (cbmc commit: 0b4db941798ea19a5db6deb9de61569db029b4da).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130035567

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4ce9a7e6bd45c1c2ec0d611bddf17b5d07bb90d4).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130040813

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 2f7c9fbe60a2f9a260b7270457cd5458001425c0).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130524976

src/solvers/smt2/smt2_parser.h
```diff
@@ -142,6 +142,11 @@ class smt2_parsert
   /// Apply typecast to unsignedbv to given expression
   exprt cast_bv_to_unsigned(const exprt &);
 
+  // sorts
+  typet sort();
+  std::unordered_map<std::string, std::function<typet()>> sorts;
+  virtual void setup_sorts();
```
romainbrenguier: why virtual?

src/solvers/smt2/smt2_parser.h
```diff
@@ -115,6 +116,9 @@ class smt2_parsert
     }
   };
 
+  // expressions
+  std::unordered_map<std::string, std::function<exprt()>> expressions;
+  virtual void setup_expressions();
```
romainbrenguier: why virtual?

romainbrenguier (approved):

polgreen:
This doesn't allow declaration of variables that have the same names as variables in function signatures, which would be allowed by SMT. e.g., 
~~~
(set-logic LIA)
(define-fun myfun ((x Int) (y Int)) Int
  (+ x y))
(declare-const x Int)

(assert (= (myfun x x)(+ x x )))
(check-sat)
~~~

src/solvers/smt2/smt2_parser.h
kroening: Indeed no need, removed

kroening:
> This doesn't allow declaration of variables that have the same names as variables in function signatures, which would be allowed by SMT. e.g.,

Indeed; will make that a separate PR!

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: a8aa59493556937347b4bb11b65bcd011109d44c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130873673

polgreen:
> > This doesn't allow declaration of variables that have the same names as variables in function signatures, which would be allowed by SMT. e.g.,
> 
> Indeed; will make that a separate PR!

I think the lack of local variable tracking may also affect let expressions

polgreen:
Another failing case. Needs to expand function applications:
~~~
(set-logic LIA)
(define-fun myfun ((x Int) (y Int)) Int
  (+ x y))
(declare-const z Int)

(assert (= (myfun z z)(+ z z )))
(check-sat)
~~~
This invariant fails because the procedure isn't implemented for function_applications
~~~
Invariant check failed
flattening/boolbv_width.cpp:196 function: get_entry
Condition: false
Reason: Unimplemented
~~~

kroening:
> I think the lack of local variable tracking may also affect let expressions

Indeed, added a test for that.

kroening:
> This invariant fails because the procedure isn't implemented for function_applications

The problem here is that we don't have an integer solver. The example works on bit-vectors.

PR #5145, opened 2019-10-03
Refinements to `class_method_descriptor_exprt`

thomasspriggs:
This PR renames the components of `class_method_descriptor_exprt`, in order to make their purpose clearer. The constructor of this class and its getters are documented and validation is added.

- [ ] ~~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~~ N/A (No user visible behaviour.)
- [ ] ~~My commit message includes data points confirming performance improvements (if claimed).~~ N/A (None claimed)

### DIFF
--- a/jbmc/src/java_bytecode/ci_lazy_methods.cpp
+++ b/jbmc/src/java_bytecode/ci_lazy_methods.cpp
@@ -263,7 +263,7 @@ bool ci_lazy_methodst::handle_virtual_methods_with_no_callees(
     // didn't already exist. It can't be instantiated already, otherwise it
     // would give a concrete definition of the called method, and
     // candidate_target_methods would be non-empty.
-    const irep_idt &call_class = virtual_function.get_class_name();
+    const irep_idt &call_class = virtual_function.class_id();
     bool was_missing = instantiated_classes.count(call_class) == 0;
     CHECK_RETURN(was_missing);
     any_new_classes = true;
@@ -477,7 +477,7 @@ void ci_lazy_methodst::get_virtual_method_targets(
   std::unordered_set<irep_idt> &callable_methods,
   symbol_tablet &symbol_table)
 {
-  const auto &call_class = called_function.get_class_name();
+  const auto &call_class = called_function.class_id();
   INVARIANT(
     !call_class.empty(), "All virtual calls should be aimed at a class");
   const auto &call_basename = called_function.get_component_name();
--- a/jbmc/src/java_bytecode/ci_lazy_methods.cpp
+++ b/jbmc/src/java_bytecode/ci_lazy_methods.cpp
@@ -478,12 +478,7 @@ void ci_lazy_methodst::get_virtual_method_targets(
   symbol_tablet &symbol_table)
 {
   const auto &call_class = called_function.class_id();
-  INVARIANT(
-    !call_class.empty(), "All virtual calls should be aimed at a class");
   const auto &method_name = called_function.mangled_method_name();
-  INVARIANT(
-    !method_name.empty(),
-    "Virtual function must have a reasonable name after removing class");
 
   class_hierarchyt::idst self_and_child_classes =
     class_hierarchy.get_children_trans(call_class);
--- a/jbmc/src/java_bytecode/ci_lazy_methods.cpp
+++ b/jbmc/src/java_bytecode/ci_lazy_methods.cpp
@@ -292,13 +292,13 @@ bool ci_lazy_methodst::handle_virtual_methods_with_no_callees(
     }
 
     // Check that `get_virtual_method_target` returns a method now
-    const irep_idt &call_basename = virtual_function.get_component_name();
-    const irep_idt method_name = get_virtual_method_target(
-      instantiated_classes, call_basename, call_class, symbol_table);
-    CHECK_RETURN(!method_name.empty());
+    const irep_idt &method_name = virtual_function.mangled_method_name();
+    const irep_idt method_id = get_virtual_method_target(
+      instantiated_classes, method_name, call_class, symbol_table);
+    CHECK_RETURN(!method_id.empty());
 
     // Add what it returns to methods_to_convert_later
-    methods_to_convert_later.insert(method_name);
+    methods_to_convert_later.insert(method_id);
   }
   return any_new_classes;
 }
@@ -480,9 +480,9 @@ void ci_lazy_methodst::get_virtual_method_targets(
   const auto &call_class = called_function.class_id();
   INVARIANT(
     !call_class.empty(), "All virtual calls should be aimed at a class");
-  const auto &call_basename = called_function.get_component_name();
+  const auto &method_name = called_function.mangled_method_name();
   INVARIANT(
-    !call_basename.empty(),
+    !method_name.empty(),
     "Virtual function must have a reasonable name after removing class");
 
   class_hierarchyt::idst self_and_child_classes =
@@ -491,10 +491,10 @@ void ci_lazy_methodst::get_virtual_method_targets(
 
   for(const irep_idt &class_name : self_and_child_classes)
   {
-    const irep_idt method_name = get_virtual_method_target(
-      instantiated_classes, call_basename, class_name, symbol_table);
-    if(!method_name.empty())
-      callable_methods.insert(method_name);
+    const irep_idt method_id = get_virtual_method_target(
+      instantiated_classes, method_name, class_name, symbol_table);
+    if(!method_id.empty())
+      callable_methods.insert(method_id);
   }
 }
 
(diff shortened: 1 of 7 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 89d2d734868576be0ea6ccb6d94ffef6ef161c84).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130227916

src/util/std_expr.h
```diff
@@ -4611,7 +4614,10 @@ class class_method_descriptor_exprt : public nullary_exprt
     return get(ID_component_name);
   }
 
-  const irep_idt &get_class_name() const
+  /// Unique identifier in the symbol table, of the compile time type of the
+  /// class which this expression is applied to. For example this could be -
+  /// `java::java.lang.Object`.
+  const irep_idt &class_id() const
```
thk123, line 4640: :question: not sure I agree with rename - it is the classes symbol name, I'd have thought the identifier as as without the java:: prefix... actually not sure I agree with myself here, seems we use tag / id(entifier) / name throughout. Perahps you have a convincing argument for why this is a good rename?

src/util/std_expr.h
```diff
@@ -4622,7 +4625,9 @@ class class_method_descriptor_exprt : public nullary_exprt
     return get(ID_C_class);
   }
 
-  const irep_idt &get_base_name() const
+  /// The name of the method to which this expression is applied as would be
+  /// seen in the source code. For example this could be - `toString`.
+  const irep_idt &base_method_name() const
```
thk123, line 4647: :pick: for me, `method_base_name` reads a little more naturally

jbmc/src/java_bytecode/ci_lazy_methods.cpp
```diff
@@ -292,13 +292,13 @@ bool ci_lazy_methodst::handle_virtual_methods_with_no_callees(
     }
 
     // Check that `get_virtual_method_target` returns a method now
-    const irep_idt &call_basename = virtual_function.get_component_name();
-    const irep_idt method_name = get_virtual_method_target(
-      instantiated_classes, call_basename, call_class, symbol_table);
-    CHECK_RETURN(!method_name.empty());
+    const irep_idt &method_name = virtual_function.mangled_method_name();
```
thk123, line 295: :pick: keep one rename per commit in future

src/util/std_expr.h
```diff
@@ -4598,21 +4598,34 @@ class class_method_descriptor_exprt : public nullary_exprt
...
   }
 
-  const irep_idt &get_component_name() const
+  /// The method name after mangling it by combining it with the descriptor.
+  /// The mangled name is distinguished from other overloads of the method with
+  /// different counts of or types of parameters. It is not distinguished
+  /// between different implementations within a class hierarchy. For example if
+  /// the overall expression refers to the `java.lang.Object.toString` method,
+  /// then the mangled_method_name would be `toString:()Ljava/lang/String;`
+  const irep_idt &mangled_method_name() const
```
thk123, line 4632: :pick: Would `method_descriptor` be more precise?

jbmc/src/java_bytecode/ci_lazy_methods.cpp
thk123: :question: `method_name` and `method_id` is confusing as to what is the difference between them. Perhaps `method_name` and `resolved_method_symbol_id`?

src/util/std_expr.h
```diff
@@ -4654,6 +4657,23 @@ class class_method_descriptor_exprt : public nullary_exprt
...
+  validate_operands(value, 0, "class method descriptor must not have operands");
+  DATA_INVARIANT(
+    !value.mangled_method_name().empty(),
+    "class method descriptor must have a mangled method name.");
+  DATA_INVARIANT(
+    !value.class_id().empty(), "class method descriptor must have a class id.");
+  DATA_INVARIANT(
+    !value.base_method_name().empty(),
+    "class method descriptor must have a base method name.");
+  DATA_INVARIANT(
```
thk123, line 4672: :bulb: Rather than enforcing this data invariant, why not have `get_identifier` return `id2string(value.class_id()) + "." + id2string(value.mangled_method_name())`?

src/util/std_expr.h
```diff
@@ -4588,6 +4588,8 @@ inline array_comprehension_exprt &to_array_comprehension_expr(exprt &expr)
   return ret;
 }
 
+inline void validate_expr(const class class_method_descriptor_exprt &value);
```
thk123, line 4591: :question: is this the correct way to add a validation rule (someone from C&Ada will know, @hannes-steffenhagen-diffblue ?

thk123 (commented):
Couple of questions but lgtm. But the :bulb: seems to be like a really strong improvement

src/util/std_expr.h
thomasspriggs: No. It would be less precise. The method descriptor only includes the type information and not the name.

jbmc/src/java_bytecode/ci_lazy_methods.cpp
smowton, line 295: Might be easier if we used `method_base_name` and `mangled_method_base_name`, making it clear that neither includes the package/class qualifiers?

smowton (approved):
One suggestion

src/util/std_expr.h
thomasspriggs: I considered doing that. However I didn't wish to change the underlying data structure (yet), just in case there are any remaining direct accesses. I think ideally we would also keep the mangled name as a separate `method_base_name` and `method_descriptor`, so that the `mangled_method_name` would return `id2string(base_method_name()) + ":" + id2string(method_descriptor())`.

src/util/std_expr.h
thomasspriggs: To me an `_id` is a unique identifier, whereas a name might be a "friendlier" identifier and is not guaranteed to be unique. Because the return of this function is a unique identifier, I think the function should be named `_id`.

hannes-steffenhagen-diffblue: IMHO `name` and `id` should both mean the same thing, i.e. unique if said unqualified. But `id` is a bit more common for this use so I think it’s good to leave this as is. For human readable representations we have the `pretty_` prefix elsewhere

jbmc/src/java_bytecode/ci_lazy_methods.cpp
thomasspriggs: To me, `base_name` means "not mangled". So `mangled_method_base_name` would read as a contradiction.

src/util/std_expr.h
hannes-steffenhagen-diffblue: I think it’s a question of whether it’s the base of the method name or the name of the method base.

src/util/std_expr.h
```diff
@@ -4654,6 +4657,23 @@ class class_method_descriptor_exprt : public nullary_exprt
   }
 };
 
+inline void validate_expr(const class class_method_descriptor_exprt &value)
```
hannes-steffenhagen-diffblue, line 4661: I’m not sure what the `class` keyword is doing here. Seems redundant?

src/util/std_expr.h
hannes-steffenhagen-diffblue: I wasn’t involved with the validation stuff myself, but I’ve asked and this seems OK to me.

src/util/std_expr.h
thomasspriggs: "base/basis of the method's name". We aren't dealing with a method called "base" here. I wanted to insert the word "method" in order to make it clear that this particular name relates to the method, rather than somehow to the class or the expression as a whole. I don't really see that either ordering would be clearer than the other.

jbmc/src/java_bytecode/ci_lazy_methods.cpp
thomasspriggs: Would it be an improvement if I renamed the `method_name` variable to `mangled_method_name`?

src/util/std_expr.h
thomasspriggs: It isn't redundant. It is an inline forward declaration of the `class_method_descriptor_exprt` class. The forward declaration of this `validate_expr` function needs to be before the definition of the `class_method_descriptor_exprt` class because it is used from inside the classes constructor. The class also needs to be forward declared in order for the function forward declaration to be valid. So the two forward declarations break a cyclic dependency problem and can't be removed without either re-introducing the problem, or defining the constructor out of line instead of in-line. I could remove the keyword `class` from this line of code and put `class class_method_descriptor_exprt;` on the proceeding line, but this was a little more succinct.

thk123: FWIW coding guidelines say forward declarations should be put at the top of the file with the includes

thk123 (approved):

src/util/std_expr.h
thomasspriggs: Ah, I'll try and remember for next time.

PR #5146, opened 2019-10-03
Unit tests: remove pointless command-line argument [TG-2708]

smowton:
Setting java-cp-include-files requires registering the option properly,
but there's no point in doing so as its only purpose is to limit the classes
explored by java_class_loadert, and limiting it to the entire classpath is
equivalent to not limiting it at all.


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 094fa2a159ff1b606b8841569a06ab594c89ce03).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130284395

romainbrenguier (approved):

PR #5148, opened 2019-10-04
Remove symbol_module_map from `symbol_tablet`

thomasspriggs:
The `symbol_module_map` is not referenced from outside of the symbol
table. It is only read from inside the symbol table for consistency
checking purposes. Therefore the `symbol_module_map` and all code
referring to it can be removed resulting in less code to maintain.

- [ ] ~~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~~ No user visible feature or behaviour.
- [ ] ~~My commit message includes data points confirming performance improvements (if claimed).~~ Non claimed.

### DIFF
--- a/src/util/journalling_symbol_table.h
+++ b/src/util/journalling_symbol_table.h
@@ -51,8 +51,7 @@ private:
   explicit journalling_symbol_tablet(symbol_table_baset &base_symbol_table)
     : symbol_table_baset(
         base_symbol_table.symbols,
-        base_symbol_table.symbol_base_map,
-        base_symbol_table.symbol_module_map),
+        base_symbol_table.symbol_base_map),
       base_symbol_table(base_symbol_table)
   {
   }
@@ -61,10 +60,7 @@ public:
   journalling_symbol_tablet(const journalling_symbol_tablet &other) = delete;
 
   journalling_symbol_tablet(journalling_symbol_tablet &&other)
-    : symbol_table_baset(
-        other.symbols,
-        other.symbol_base_map,
-        other.symbol_module_map),
+    : symbol_table_baset(other.symbols, other.symbol_base_map),
       base_symbol_table(other.base_symbol_table),
       inserted(std::move(other.inserted)),
       updated(std::move(other.updated)),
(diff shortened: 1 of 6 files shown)


smowton (changes requested):
Believe this is used in derivative products such as the Verilog frontend (EBMC?), blocking until we know otherwise (@kroening ?)

chrisr-diffblue:
@xbauch - Any thoughts?

thomasspriggs:
My motivation for making this change is that in the case where all symbols have the same module, then `symbol_tablet::erase` has to effectively iterate over the entire symbol table in order to find the particular entry to erase. This then caused a performance problem in my use case. If there is however a motivation for keeping this otherwise unused data structure, then I can investigate alternative solutions.

chrisr-diffblue:
@martin-cs might also have some thoughts on the history/uses of this data structure.

xbauch:
@thomasspriggs I confirm that `ebmc` uses that map.

thomasspriggs:
@xbauch If I were to make an alternative change which prevented adding symbols to the module map in the case where the symbol's module was empty, would that be compatible with the existing ebmc code base?

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: bbb8f9bfd27c95a3b5d39763ef84f4fe9065fee5).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130421593

xbauch:
> @xbauch If I were to make an alternative change which prevented adding symbols to the module map in the case where the symbol's module was empty, would that be compatible with the existing ebmc code base?

It seems that `ebmc` only every iterates over the module-map using some identifier so symbols without a module wouldn't be found anyway.

thomasspriggs:
Given that this is used in EBMC, I am going to close out this PR, rather than making extra work for ebmc maintainers.

thk123:
@xbauch doesn't need to be done now, but longer term, ebmc derive from `symbol_table_baset` and add this field itself, or is there some technical reason this wouldn't' work?

PR #5149, opened 2019-10-04
String solver setLength routine: tolerate negative lengths

smowton:
These are handled by the wrapping models library function in practice, but the string solver function must not constrain the length to be >= 0, or the constraint propagates backwards into the calling 
 context and prevents parallel branches where the value *is* negative from being covered.


smowton:
Replacing branch because Travis is confused

smowton:
New branch doesn't help; trivial punctuation fiddling to change the commit SHA does.

allredj (approved):
Looks fine, but the commit message doesn't explain why we need that change in the first place, as the model should properly prevent negative lengths from being a thing.

smowton:
@allredj "the constraint propagates backwards into the calling context and prevents parallel branches where the value *is* negative from being covered.", which means that unlike a normal `__CPROVER_assume` the constraints applied don't only apply on this branch, they apply everywhere. So if I had `if(x < 0) { // HERE } else { someString.setLength(x); }` then if the string solver  constrains the length >= 0 then it will render `HERE` inaccessible.

src/solvers/strings/string_constraint_generator_transformation.cpp
```diff
@@ -47,12 +49,12 @@ string_constraint_generatort::add_axioms_for_set_length(
...
 
   // We add axioms:
-  // a1 : |res|=k
+  // a1 : |res|=max(k, 0)
   // a2 : forall i< min(|s1|, k) .res[i] = s1[i]
   // a3 : forall |s1| <= i < |res|. res[i] = 0
 
   constraints.existential.push_back(
-    equal_to(array_pool.get_or_create_length(res), k));
+    equal_to(array_pool.get_or_create_length(res), zero_if_negative(k)));
```
danpoe, line 57: Now that we have this constraint we could remove `zero_if_negative()` in line 70.

src/solvers/strings/string_constraint_generator_transformation.cpp
```diff
@@ -47,12 +49,12 @@ string_constraint_generatort::add_axioms_for_set_length(
   const typet &char_type = s1.content().type().subtype();
 
   // We add axioms:
-  // a1 : |res|=k
+  // a1 : |res|=max(k, 0)
   // a2 : forall i< min(|s1|, k) .res[i] = s1[i]
```
danpoe, line 53: Could update doc here to say that upper bound is `max(0, min(|s1|, k))`.

danpoe (approved):
Looks good, comments optional

allredj:
I still don't see where that `length >= 0` constraints exists.

jbmc/regression/jbmc-strings/StringBufferSetLength/SetLengthTest.java
```diff
@@ -0,0 +1,30 @@
+public class SetLengthTest {
+
+    public static void main(int unknown) {
+
+        StringBuffer s = new StringBuffer("ABCD");
+        // Avoid producing a huge string that consumes too much memory, or which
+        // string refinement can't handle:
```
romainbrenguier, line 7: :information_source: string refinement could handle a huge string there, but concretizing it in a trace would consume too much memory so it would get replaced by nil (which would make an invariant fail in trace checking)

romainbrenguier (approved):

smowton:
@allredj the constraint is because the string solver constrains any string it creates to have a valid length, and specifically for `setLength` constrains the input to equal the new string length

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 88d43e61fd0009ce16338915338325da3b89aee1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130444588

PR #5150, opened 2019-10-04
String solver setLength routine: tolerate negative lengths

smowton:
These are handled by the wrapping models library function in practice, but the string solver function must not constrain the length to be >= 0, or the constraint propagates backwards into the calling context and prevents parallel branches where the value *is* negative from being covered.


smowton:
Travis still confused, disregard

PR #5157, opened 2019-10-08
[TG-9830] [UFC] Exclude symbols with empty module from the symbol module map

antlechner:
This improves performance whenever we don't use the symbol module map (which, as far as I know, is currently only used in EBMC). See first commit message for further details.
Benchmarking results will be added here when I have them.

- [x] Each commit message has a non-empty body, explaining why the change was made. (except for documentation and unit test)

### DIFF
--- a/src/util/symbol_table.cpp
+++ b/src/util/symbol_table.cpp
@@ -93,10 +93,10 @@ void symbol_tablet::erase(const symbolst::const_iterator &entry)
 {
   const symbolt &symbol=entry->second;
 
-  symbol_base_mapt::const_iterator
-    base_it=symbol_base_map.lower_bound(entry->second.base_name);
-  symbol_base_mapt::const_iterator
-    base_it_end=symbol_base_map.upper_bound(entry->second.base_name);
+  symbol_base_mapt::const_iterator base_it =
+    symbol_base_map.lower_bound(symbol.base_name);
+  symbol_base_mapt::const_iterator base_it_end =
+    symbol_base_map.upper_bound(symbol.base_name);
   while(base_it!=base_it_end && base_it->second!=symbol.name)
     ++base_it;
   INVARIANT(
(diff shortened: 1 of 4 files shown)


src/util/symbol_table.cpp
```diff
@@ -93,10 +93,8 @@ void symbol_tablet::erase(const symbolst::const_iterator &entry)
 {
   const symbolt &symbol=entry->second;
 
-  symbol_base_mapt::const_iterator base_it =
-    symbol_base_map.lower_bound(symbol.base_name);
-  symbol_base_mapt::const_iterator base_it_end =
-    symbol_base_map.upper_bound(symbol.base_name);
+  auto base_it = symbol_base_map.lower_bound(symbol.base_name);
+  auto base_it_end = symbol_base_map.upper_bound(symbol.base_name);
```
thomasspriggs: :bulb: `base_it_end` isn't mutated therefore it could be declared `const auto`.

smowton (approved):
Should be fine; are we 100% convinced they don't use any empty-module symbols?

Also please document that symbols in the empty module are not recorded in the map.

xbauch (approved):
Looks good.

thomasspriggs (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 5843a4457c3f07091564cedc986888c027c7b2fd).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/130996175

antlechner:
I have benchmarked this using TG. Using our benchmarking script on the default subset of Tika, the methods that succeeded on both runs (before and after the fix from this PR) were overall 216 seconds faster after the fix (before: 1251, after: 1035). There are 231 methods in the subset so this is a significant improvement.

Another comparison I ran was a trivial method (`return true`, no args), with the TG version of the models (which are compatible with rt.jar) and rt.jar on the classpath. Before the fix, it took about 40 seconds, after the fix it takes 6 seconds.

antlechner:
@smowton 
> are we 100% convinced they don't use any empty-module symbols?

We asked @xbauch and he confirmed that this change is compatible with EBMC.
> Also please document that symbols in the empty module are not recorded in the map.

Done in the same commit as the other documentation.

segun3d:
Dev has benchmarked her fix against the statusquo on develop with proven gains in performance. I have attached a copy of the benchmark results to the Jira ticket.

QA approved.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4718929976178f8a5eb3790fc4a4e179ee2afa1c).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/131182785

PR #5159, opened 2019-10-10
Collect complexity statistics of solver queries

xbauch:
Introduces a functionality to track and collect the complexity of individual
solver queries. For SAT solvers we collect the number of clauses, literals, and
variables. The functionality matches solver queries to the SSA steps that
produced them. After running the solver, we produce a JSON report that matches
program line to GOTO instruction to SSA step to solver hardness.

At present the collection is command-line optional and only works with the
default SAT solver (minisat2). Extension to other SAT and SMT solvers should be
possible.

A simple regression test is included.

### DIFF
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -460,6 +460,12 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
     }
   }
 
+  if(cmdline.isset("write-solver-stats-to"))
+  {
+    options.set_option(
+      "write-solver-stats-to", cmdline.get_value("write-solver-stats-to"));
+  }
+
   if(cmdline.isset("beautify"))
     options.set_option("beautify", true);
 
@@ -1106,6 +1112,8 @@ void cbmc_parse_optionst::help()
     HELP_FLUSH
     " --verbosity #                verbosity level\n"
     HELP_TIMESTAMP
+    " --write-solver-stats-to json-file\n"
+    "                              collect the solver query complexity\n"
     "\n";
   // clang-format on
 }
--- a/src/cbmc/cbmc_parse_options.h
+++ b/src/cbmc/cbmc_parse_options.h
@@ -46,6 +46,7 @@ class optionst;
   OPT_REACHABILITY_SLICER \
   "(debug-level):(no-propagation)(no-simplify-if)" \
   "(document-subgoals)(outfile):(test-preprocessor)" \
+  "(write-solver-stats-to):"  \
   "D:I:(c89)(c99)(c11)(cpp98)(cpp03)(cpp11)" \
   "(object-bits):" \
   OPT_GOTO_CHECK \
--- a/src/goto-checker/bmc_util.cpp
+++ b/src/goto-checker/bmc_util.cpp
@@ -26,6 +26,7 @@ Author: Daniel Kroening, Peter Schrammel
 #include <linking/static_lifetime_init.h>
 
 #include <solvers/decision_procedure.h>
+#include <solvers/flattening/boolbv.h>
 
 #include <util/make_unique.h>
 #include <util/ui_message.h>
@@ -149,14 +150,22 @@ void output_graphml(
 
 void convert_symex_target_equation(
   symex_target_equationt &equation,
-  decision_proceduret &decision_procedure,
+  goto_symex_property_decidert &property_decider,
   message_handlert &message_handler)
 {
   messaget msg(message_handler);
   msg.status() << "converting SSA" << messaget::eom;
 
   // convert SSA
-  equation.convert(decision_procedure);
+  if(dynamic_cast<boolbvt *>(&property_decider.get_decision_procedure()))
+  {
+    equation.convert(
+      property_decider.get_decision_procedure(), property_decider.get_prop());
+  }
+  else
+  {
+    equation.convert(property_decider.get_decision_procedure());
+  }
 }
 
 std::unique_ptr<memory_model_baset>
@@ -361,8 +370,7 @@ std::chrono::duration<double> prepare_property_decider(
     << property_decider.get_decision_procedure().decision_procedure_text()
     << messaget::eom;
 
-  convert_symex_target_equation(
-    equation, property_decider.get_decision_procedure(), ui_message_handler);
+  convert_symex_target_equation(equation, property_decider, ui_message_handler);
   property_decider.update_properties_goals_from_symex_target_equation(
     properties);
   property_decider.convert_goals();
--- a/src/goto-checker/bmc_util.h
+++ b/src/goto-checker/bmc_util.h
@@ -36,7 +36,7 @@ class ui_message_handlert;
 
 void convert_symex_target_equation(
   symex_target_equationt &,
-  decision_proceduret &,
+  goto_symex_property_decidert &,
   message_handlert &);
 
 /// Returns a function that checks whether an SSA step is an assertion
--- a/src/goto-checker/goto_symex_property_decider.cpp
+++ b/src/goto-checker/goto_symex_property_decider.cpp
@@ -110,6 +110,11 @@ goto_symex_property_decidert::get_decision_procedure() const
   return solver->decision_procedure();
 }
 
+propt &goto_symex_property_decidert::get_prop() const
+{
+  return solver->prop();
+}
+
 stack_decision_proceduret &
 goto_symex_property_decidert::get_stack_decision_procedure() const
 {
--- a/src/goto-checker/incremental_goto_checker.h
+++ b/src/goto-checker/incremental_goto_checker.h
@@ -77,6 +77,12 @@ public:
   /// failing properties any more.
   virtual resultt operator()(propertiest &properties) = 0;
 
+  /// Additional reporting that may result from the underlying solver, no-op by
+  /// default.
+  virtual void report()
+  {
+  }
+
 protected:
   incremental_goto_checkert(const optionst &, ui_message_handlert &);
 
--- a/src/goto-checker/multi_path_symex_checker.cpp
+++ b/src/goto-checker/multi_path_symex_checker.cpp
@@ -25,6 +25,17 @@ multi_path_symex_checkert::multi_path_symex_checkert(
     equation_generated(false),
     property_decider(options, ui_message_handler, equation, ns)
 {
+  if(options.is_set("write-solver-stats-to"))
+  {
+    if(
+      auto hardness_collector =
+        dynamic_cast<hardness_collectort *>(&property_decider.get_prop()))
+    {
+      hardness_collector->enable_hardness_collection();
+      hardness_collector->get_solver_hardness().set_outfile(
+        options.get_option("write-solver-stats-to"));
+    }
+  }
 }
 
 incremental_goto_checkert::resultt multi_path_symex_checkert::
@@ -154,4 +165,20 @@ multi_path_symex_checkert::localize_fault(const irep_idt &property_id) const
     property_decider.get_stack_decision_procedure());
 
   return fault_localizer(property_id);
+}
+
+void multi_path_symex_checkert::report()
+{
+  if(options.is_set("write-solver-stats-to"))
+  {
+    if(
+      auto hardness_collector =
+        dynamic_cast<hardness_collectort *>(&property_decider.get_prop()))
+    {
+      if(hardness_collector->is_hardness_collection_enabled())
+      {
+        hardness_collector->get_solver_hardness().produce_report();
+      }
+    }
+  }
 }
--- a/src/goto-checker/single_loop_incremental_symex_checker.cpp
+++ b/src/goto-checker/single_loop_incremental_symex_checker.cpp
@@ -89,7 +89,7 @@ operator()(propertiest &properties)
 
         log.status() << "converting SSA" << messaget::eom;
         equation.convert_without_assertions(
-          property_decider.get_decision_procedure());
+          property_decider.get_decision_procedure(), nullptr);
 
         property_decider.update_properties_goals_from_symex_target_equation(
           properties);
@@ -97,7 +97,7 @@ operator()(propertiest &properties)
         // We convert the assertions in a new context.
         property_decider.get_stack_decision_procedure().push();
         equation.convert_assertions(
-          property_decider.get_decision_procedure(), false);
+          property_decider.get_decision_procedure(), nullptr, false);
         property_decider.convert_goals();
 
         current_equation_converted = true;
--- a/src/goto-symex/symex_target_equation.cpp
+++ b/src/goto-symex/symex_target_equation.cpp
@@ -16,6 +16,8 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/std_expr.h>
 
 #include <solvers/decision_procedure.h>
+#include <solvers/hardness_collector.h>
+#include <solvers/prop/prop.h>
 
 #include "goto_symex_state.h"
 #include "ssa_step.h"
@@ -319,27 +321,52 @@ void symex_target_equationt::constraint(
 }
 
 void symex_target_equationt::convert_without_assertions(
-  decision_proceduret &decision_procedure)
+  decision_proceduret &decision_procedure,
+  solver_hardnesst *solver_hardness)
 {
-  convert_guards(decision_procedure);
-  convert_assignments(decision_procedure);
-  convert_decls(decision_procedure);
-  convert_assumptions(decision_procedure);
-  convert_goto_instructions(decision_procedure);
-  convert_function_calls(decision_procedure);
-  convert_io(decision_procedure);
-  convert_constraints(decision_procedure);
+  if(solver_hardness)
+    solver_hardness->register_ssa_size(SSA_steps.size());
+
+  convert_guards(decision_procedure, solver_hardness);
+  convert_assignments(decision_procedure, solver_hardness);
+  convert_decls(decision_procedure, solver_hardness);
+  convert_assumptions(decision_procedure, solver_hardness);
+  convert_goto_instructions(decision_procedure, solver_hardness);
+  convert_function_calls(decision_procedure, solver_hardness);
+  convert_io(decision_procedure, solver_hardness);
+  convert_constraints(decision_procedure, solver_hardness);
+}
+
+void symex_target_equationt::convert(
+  decision_proceduret &decision_procedure,
+  propt &prop)
+{
+  auto solver_hardness = [&prop]() -> solver_hardnesst * {
+    if(auto hardness_collector = dynamic_cast<hardness_collectort *>(&prop))
+    {
+      if(hardness_collector->is_hardness_collection_enabled())
+      {
+        return &hardness_collector->get_solver_hardness();
+      }
+    }
+    return nullptr;
+  }();
+
+  convert_without_assertions(decision_procedure, solver_hardness);
+  convert_assertions(decision_procedure, solver_hardness); //, solver_hardness);
 }
 
 void symex_target_equationt::convert(decision_proceduret &decision_procedure)
 {
-  convert_without_assertions(decision_procedure);
-  convert_assertions(decision_procedure);
+  convert_without_assertions(decision_procedure, nullptr);
+  convert_assertions(decision_procedure, nullptr); //, solver_hardness);
 }
 
 void symex_target_equationt::convert_assignments(
-  decision_proceduret &decision_procedure)
+  decision_proceduret &decision_procedure,
+  solver_hardnesst *solver_hardness)
 {
+  std::size_t step_index = 0;
   for(auto &step : SSA_steps)
   {
     if(step.is_assignment() && !step.ignore && !step.converted)
@@ -351,13 +378,22 @@ void symex_target_equationt::convert_assignments(
 
       decision_procedure.set_to_true(step.cond_expr);
       step.converted = true;
+
+      if(solver_hardness != nullptr)
+      {
+        solver_hardness->register_ssa(
+          step_index, step.cond_expr, step.source.pc);
+      }
     }
+    ++step_index;
   }
 }
 
 void symex_target_equationt::convert_decls(
-  decision_proceduret &decision_procedure)
+  decision_proceduret &decision_procedure,
+  solver_hardnesst *solver_hardness)
 {
+  std::size_t step_index = 0;
   for(auto &step : SSA_steps)
   {
     if(step.is_decl() && !step.ignore && !step.converted)
@@ -366,13 +402,22 @@ void symex_target_equationt::convert_decls(
       // the satisfiability of the formula.
       decision_procedure.handle(step.cond_expr);
       step.converted = true;
+
+      if(solver_hardness != nullptr)
+      {
+        solver_hardness->register_ssa(
+          step_index, step.cond_expr, step.source.pc);
+      }
     }
+    ++step_index;
   }
 }
 
 void symex_target_equationt::convert_guards(
-  decision_proceduret &decision_procedure)
+  decision_proceduret &decision_procedure,
+  solver_hardnesst *solver_hardness)
 {
+  std::size_t step_index = 0;
   for(auto &step : SSA_steps)
   {
     if(step.ignore)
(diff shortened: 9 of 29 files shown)


src/goto-checker/solver_factory.cpp
```diff
@@ -11,6 +11,8 @@ Author: Daniel Kroening, Peter Schrammel
 
 #include "solver_factory.h"
 
+#include <sys/stat.h>
```
xbauch: Will be removed.

src/solvers/solver_hardness.cpp
```diff
@@ -0,0 +1,324 @@
...
+  current_hardness.literals += bv.size();
+  for(const auto &literal : bv)
+  {
+    current_hardness.variables.insert(literal.var_no());
+  }
+}
+
+void solver_hardnesst::produce_report(const namespacet &ns)
+{
+  if(outfile.empty())
```
hannes-steffenhagen-diffblue: This behaviour is undocumented. It should be documented.

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -426,6 +426,11 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
     }
   }
 
+  if(cmdline.isset("solver-stats") && !cmdline.isset("paths"))
+  {
+    options.set_option("solver-stats", cmdline.get_value("solver-stats"));
```
hannes-steffenhagen-diffblue: The name of this option should reflect that it contains the name of a file to write output to. Maybe something as simple as `write-solver-stats-to`?

regression/cbmc/solver-hardness-simple/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--solver-stats ""
+^EXIT=10$
+^SIGNAL=0$
+^\s*\"GOTO\": \"ASSERT a \+ b \< 10\",$
+^\s*\"GOTO_ID\": \d+,$
+^\s*\"Clauses\": \d+,$
+^\[main.assertion.\d+\] line \d+ assertion a \+ b \< 10: FAILURE$
+^VERIFICATION FAILED$
```
hannes-steffenhagen-diffblue: This is one of those cases where it might be worth having a separate test script that can actually parse the JSON instead; Not sure if it's worth doing, but we could use such a thing to make sure the output we get is actually valid JSON, and that it is of the structure we're expecting.

I'm also not quite convinced this covers all of the code of the stat collection

src/goto-checker/solver_factory.cpp
hannes-steffenhagen-diffblue: This header is for the posix stat syscalls. I don't think you need it here.

src/goto-symex/symex_target_equation.cpp
```diff
@@ -424,31 +465,48 @@ void symex_target_equationt::convert_goto_instructions(
           });
 
         step.cond_handle = decision_procedure.handle(step.cond_expr);
+
+        if(solver_hardness != nullptr)
```
hannes-steffenhagen-diffblue, line 496: Not the end of the world, but I'm usually not a huge fan of using `nullptr` to indicate optional things if we can avoid it

src/goto-symex/symex_target_equation.cpp
```diff
@@ -424,31 +465,48 @@ void symex_target_equationt::convert_goto_instructions(
...
+  auto solver_hardness = decision_procedure.solver_hardness;
   for(auto &step : SSA_steps)
   {
-    if(step.is_constraint() && !step.ignore && !step.converted)
-    {
-      log.conditional_output(log.debug(), [&step](messaget::mstreamt &mstream) {
-        step.output(mstream);
-        mstream << messaget::eom;
-      });
+    if(step.is_goto())
```
hannes-steffenhagen-diffblue: Please document why you need this additional check here

src/goto-symex/symex_target_equation.cpp
```diff
@@ -458,6 +516,7 @@ void symex_target_equationt::convert_assertions(
   if(number_of_assertions==0)
     return;
 
+  auto solver_hardness = decision_procedure.solver_hardness;
```
hannes-steffenhagen-diffblue: const

src/goto-symex/symex_target_equation.cpp
```diff
@@ -520,19 +601,34 @@ void symex_target_equationt::convert_assertions(
...
         assumption = and_exprt(assumption, step.cond_handle);
+
+      if(solver_hardness != nullptr)
+      {
+        involved_steps.push_back(step.source.pc);
+      }
     }
   }
 
+  auto assertion_disjunction = disjunction(disjuncts);
```
hannes-steffenhagen-diffblue: const

src/goto-symex/symex_target_equation.cpp
```diff
@@ -520,19 +601,34 @@ void symex_target_equationt::convert_assertions(
...
+      from_expr(ns, ID_C, assertion_disjunction), involved_steps);
+  }
 }
 
 void symex_target_equationt::convert_function_calls(
-  decision_proceduret &dec_proc)
+  decision_proceduret &dec_proc,
+  const namespacet &ns)
 {
+  auto solver_hardness = dec_proc.solver_hardness;
```
hannes-steffenhagen-diffblue: const

src/goto-symex/symex_target_equation.cpp
```diff
@@ -520,19 +601,34 @@ void symex_target_equationt::convert_assertions(
...
   for(auto &step : SSA_steps)
     if(!step.ignore)
     {
+      and_exprt::operandst conjuncts;
```
hannes-steffenhagen-diffblue, line 661: what's happening here?

src/solvers/solver_hardness.cpp
```diff
@@ -0,0 +1,324 @@
...
+#include <util/string2int.h>
+
+void solver_hardnesst::register_ssa(
+  const std::string &ssa_expression,
+  goto_programt::const_targett pc)
+{
+  current_stats.ssa_expression = ssa_expression;
+  current_stats.pc = pc;
+  hardness_stats.push_back(current_stats);
+  current_stats.clear();
```
hannes-steffenhagen-diffblue, line 44: What's with the clear?

src/solvers/solver_hardness.cpp
```diff
@@ -0,0 +1,324 @@
...
+}
+
+void solver_hardnesst::report_to_stream(std::ostream &out, const namespacet &ns)
+{
+  json_stream_arrayt json_stream_array{out};
+
+  std::sort(
+    hardness_stats.begin(),
+    hardness_stats.end(),
+    [](const hardness_statst &l_stats, const hardness_statst &r_stats) {
```
hannes-steffenhagen-diffblue: I think this should probably be moved out into its own function for clarity, something like 

```
static bool order_by_instruction_source_location(const hardness_statst& l_stats, const hardness_statst &r_stats);
``` 

```
std::sort(hardness_stats.begin(), hardness_stats.end(), order_by_instruction_source_location);
```

src/solvers/solver_hardness.cpp
```diff
@@ -0,0 +1,324 @@
...
+    hardness_stats_json["SSA_expr"] = json_stringt{stat.ssa_expression};
+    hardness_stats_json["GOTO"] =
+      json_stringt{goto_instruction2string(stat.pc, ns)};
+
+    // It might be desirable to collect all SAT hardness statistics pertaining
+    // to a particular GOTO instruction, since there may be a number of SSA
+    // steps per GOTO instruction. The following JSON contains a unique
+    // identifier for each one.
+    hardness_stats_json["GOTO_ID"] = json_numbert{
+      std::to_string(reinterpret_cast<std::uintptr_t>(&(*stat.pc)))};
```
hannes-steffenhagen-diffblue: This should be fine for machine processing, but we should still consider at some point if we shouldn't have more predictable "identifiers".

src/solvers/solver_hardness.cpp
```diff
@@ -0,0 +1,324 @@
...
+    auto hardness_stats_json = json_objectt{};
+    hardness_stats_json["SSA_expr"] = json_stringt{stat.ssa_expression};
+    hardness_stats_json["GOTO"] =
+      json_stringt{goto_instruction2string(stat.pc, ns)};
+
+    // It might be desirable to collect all SAT hardness statistics pertaining
+    // to a particular GOTO instruction, since there may be a number of SSA
+    // steps per GOTO instruction. The following JSON contains a unique
+    // identifier for each one.
+    hardness_stats_json["GOTO_ID"] = json_numbert{
```
hannes-steffenhagen-diffblue: I'm still baffled that a json number is a string

src/solvers/solver_hardness.cpp
```diff
@@ -0,0 +1,324 @@
...
+      std::to_string(reinterpret_cast<std::uintptr_t>(&(*stat.pc)))};
+    hardness_stats_json["Source"] = json(stat.pc->source_location);
+
+    auto sat_hardness_json = json_objectt{};
+    sat_hardness_json["Clauses"] =
+      json_numbert{std::to_string(stat.sat_hardness.clauses)};
+    sat_hardness_json["Literals"] =
+      json_numbert{std::to_string(stat.sat_hardness.literals)};
+    sat_hardness_json["Variables"] =
```
hannes-steffenhagen-diffblue: super minor point, but in my experience it's more common to use `camelCase` or `snake_case` for identifiers in JSON; This isn't at all required, but it might trip some people up a bit to see something else.

src/solvers/solver_hardness.h
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+private:
+  std::string outfile;
+  std::vector<hardness_statst> hardness_stats;
+  hardness_statst current_stats;
+  assertion_statst assertion_stats;
+
+  // A minor modification of \ref goto_programt::output_instruction
+  std::string goto_instruction2string(
```
hannes-steffenhagen-diffblue: This should be static

src/solvers/solver_hardness.cpp
```diff
@@ -0,0 +1,324 @@
...
+    assertion_stats_json["SAT_hardness"] = assertion_hardness_json;
+
+    json_stream_array.push_back(assertion_stats_json);
+  }
+}
+
+std::string solver_hardnesst::goto_instruction2string(
+  goto_programt::const_targett pc,
+  const namespacet &ns) const
+{
```
hannes-steffenhagen-diffblue, line 158: I'm not completely sold on the idea of repeating so much of the printing logic here. I get the existing printing functionality didn't quite meet our purposes here, but it still feels a bit "off" to repeat this ad-hoc over here hidden in a private function somewhere.

hannes-steffenhagen-diffblue (commented):
I think the overall direction here is OK, but there's some bits that might need cleanup or a rethink

src/goto-symex/symex_target_equation.cpp
xbauch: I don't, copy-pasted. Will be removed.

src/goto-checker/solver_factory.cpp
xbauch: Removed.

src/cbmc/cbmc_parse_options.cpp
xbauch: Renamed.

src/solvers/solver_hardness.cpp
xbauch: This should be part of the debate as to how to test this feature. Originally, I wanted only to print to a JSON file, but I did not know how to simply test the content of that file.

src/goto-symex/symex_target_equation.cpp
xbauch: Done.

src/goto-symex/symex_target_equation.cpp
xbauch: Done.

src/goto-symex/symex_target_equation.cpp
xbauch: Done.

src/goto-symex/symex_target_equation.cpp
xbauch: There is a number of solver calls for a single SSA step, each representing an equality between function argument and parameter. I wanted a single expression for all there equalities.

src/solvers/solver_hardness.cpp
xbauch: Well the solver always updates the statistics into `current_stats` and it's up to the symex to match them to an SSA step. After the match, we clean the `current_stats` before the solver starts updating again.

src/solvers/solver_hardness.cpp
xbauch: Done.

src/solvers/solver_hardness.h
xbauch: Done.

src/goto-symex/symex_target_equation.cpp
```diff
@@ -379,13 +404,21 @@ void symex_target_equationt::convert_guards(
       });
 
       step.guard_handle = decision_procedure.handle(step.guard);
+
+      if(solver_hardness != nullptr)
+      {
+        solver_hardness->register_ssa(
+          from_expr(ns, ID_C, step.guard), step.source.pc);
```
peterschrammel: 🚫 If you need a namespace then pass it through solver_hardness and don't thread it through symex_target_equation. Besides, from_expr is pretty much deprecated.

src/solvers/solver_hardness.cpp
```diff
@@ -0,0 +1,329 @@
...
+#include <cstdint>
+
+#include <langapi/language_util.h>
+
+#include <util/json_irep.h>
+#include <util/json_stream.h>
+#include <util/string2int.h>
+
+void solver_hardnesst::register_ssa(
+  const std::string &ssa_expression,
```
peterschrammel: Take an expr here

src/cbmc/cbmc_parse_options.h
```diff
@@ -45,7 +45,7 @@ class optionst;
   "(no-simplify)(full-slice)" \
   OPT_REACHABILITY_SLICER \
   "(debug-level):(no-propagation)(no-simplify-if)" \
-  "(document-subgoals)(outfile):(test-preprocessor)" \
+  "(document-subgoals)(outfile):(write-solver-stats-to):(test-preprocessor)"  \
```
peterschrammel: put on separate lines

src/goto-symex/symex_target_equation.cpp
xbauch: Done. Went for `format` instead. Prints `main::1::a` instead of `a`, but ok.

src/solvers/solver_hardness.cpp
xbauch: Done.

src/cbmc/cbmc_parse_options.h
xbauch: Done.

xbauch:
@hannes-steffenhagen-diffblue small rework in the second commit.

martin-cs:
Forgive the likely daft question but ... why?  The cost per SSA instruction is generally pretty straight-forward and there are only really a few costly operations -- you could just look for those.  Maybe I'm just not getting this...

xbauch:
@martin-cs I guess you could call it a feature request. At the beginning I had some filtering for non-trivial steps, but it was suggested that people could do it on their own, once they have the JSON. I think the big-picture motivation is to overlay the statistics in some IDE so that you can track how hard each line of code is. Code -> GOTO -> SSA -> clauses. That way you can optimise your code (for the solver) on whichever level you choose.

martin-cs:
OK... but... 1. clauses / variables are a measure of size, not difficulty, 2. the amount used depends on the context not just the assignment, 3. you can approximate expensive instructions much more simply, 4. the really big wins are from reorganisations of the code (see the N different ways of doing recursion) not individual instructions, 5. the inbuilt solver is *very* naive and you won't really be able to track similar stats for other solvers.

I'm not going to oppose this, esp. if this is something you are doing for a specific customer request but ... I think there may be a better way of achieving your high-level goal and I worry that this may not actually solve your problem.

src/goto-checker/incremental_goto_checker.h
```diff
@@ -83,6 +87,8 @@ class incremental_goto_checkert
   const optionst &options;
   ui_message_handlert &ui_message_handler;
   messaget log;
+
+  std::shared_ptr<solver_hardnesst> solver_hardness;
```
hannes-steffenhagen-diffblue: I think you explained this at some point but I still don't quite get why this is shared rather than unique. Do all the things that want a handle on this thing have different/non-hierarchical lifetimes?

hannes-steffenhagen-diffblue (approved):
Codewise, tis is probably fine modulo some smaller concerns, but this should probably not be merged until we're sure we're actually solving the right problem here.

martin-cs (commented):
Github seems to want a review from me for this so...

I think there are likely easier ways of achieving most of the information that you can get like this.  I don't think that information, regardless how it is computed, is really what you need to solve actual end-user problems.  I am happy to help with what I know about solver hardness / to discuss this if it would help.

But, I'm not going to block this, so the review is "comment" not "request changes".

karkhaz (approved):
Thanks, I believe this will be useful. Happy for this to go in!

@martin-cs thanks for your comments–they're very valid, and I do hope that people don't read more into these numbers than what they really mean. We have a very specific use-case for them, which I'm unable to share unfortunately...

src/cbmc/cbmc_parse_options.h
```diff
@@ -46,6 +46,7 @@ class optionst;
   OPT_REACHABILITY_SLICER \
   "(debug-level):(no-propagation)(no-simplify-if)" \
   "(document-subgoals)(outfile):(test-preprocessor)" \
+  "(write-solver-stats-to):"  \
```
peterschrammel, line 49: Define an `OPT_...`

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1106,6 +1112,8 @@ void cbmc_parse_optionst::help()
     HELP_FLUSH
     " --verbosity #                verbosity level\n"
     HELP_TIMESTAMP
+    " --write-solver-stats-to json-file\n"
```
peterschrammel, line 1115: Define a `HELP_...`

xbauch:
@peterschrammel I'm not sure this is exactly what you meant; I didn't know how to get the `prop` (since minisat is a `prop`) out of `decision-procedure` so I extended the conversion interface for symex-target-equation.

src/goto-symex/symex_target_equation.h
```diff
@@ -178,60 +181,93 @@ class symex_target_equationt:public symex_targett
   /// \param decision_procedure: A handle to a decision procedure interface
   void convert(decision_proceduret &decision_procedure);
 
+  /// \copydoc convert(decision_proceduret &)
+  /// \param prop: A handle to a prop interface to query the solver hardness
+  void convert(decision_proceduret &decision_procedure, propt &prop);
```
peterschrammel, line 186: 🚫 None of these signature changes are necessary.

src/goto-symex/symex_target_equation.cpp
```diff
@@ -583,10 +714,18 @@ void symex_target_equationt::convert_io(
           merge_irep(eq);
 
           dec_proc.set_to(eq, true);
+          conjuncts.push_back(eq);
           step.converted_io_args.push_back(symbol);
         }
       }
+      if(solver_hardness != nullptr)
```
peterschrammel, line 721: Check whether decision_procedure implements hardness_collectort. If yes, then do the necessary.
You may want to write a little helper method to wrap that logic to avoid repetition throughout the conversion functions.

src/goto-symex/symex_target_equation.cpp
```diff
@@ -319,27 +321,52 @@ void symex_target_equationt::constraint(
...
+      if(hardness_collector->is_hardness_collection_enabled())
+      {
+        return &hardness_collector->get_solver_hardness();
+      }
+    }
+    return nullptr;
+  }();
+
+  convert_without_assertions(decision_procedure, solver_hardness);
+  convert_assertions(decision_procedure, solver_hardness); //, solver_hardness);
```
peterschrammel, line 356: remove commented code

src/goto-symex/symex_target_equation.cpp
```diff
@@ -319,27 +321,52 @@ void symex_target_equationt::constraint(
...
 }
 
 void symex_target_equationt::convert(decision_proceduret &decision_procedure)
 {
-  convert_without_assertions(decision_procedure);
-  convert_assertions(decision_procedure);
+  convert_without_assertions(decision_procedure, nullptr);
```
peterschrammel, line 361: I don't want to pass in a `nullptr` when I don't need the feature.

src/goto-symex/symex_target_equation.cpp
```diff
@@ -319,27 +321,52 @@ void symex_target_equationt::constraint(
...
-  convert_guards(decision_procedure);
-  convert_assignments(decision_procedure);
-  convert_decls(decision_procedure);
-  convert_assumptions(decision_procedure);
-  convert_goto_instructions(decision_procedure);
-  convert_function_calls(decision_procedure);
-  convert_io(decision_procedure);
-  convert_constraints(decision_procedure);
+  auto solver_hardness = [&prop]() -> solver_hardnesst * {
+    if(auto hardness_collector = dynamic_cast<hardness_collectort *>(&prop))
```
peterschrammel, line 345: 🚫 Make `prop_conv_solvert` implement `hardness_collectort` and forward the calls. `propt` MUST not be exposed here.

src/goto-checker/single_loop_incremental_symex_checker.cpp
```diff
@@ -89,15 +89,15 @@ operator()(propertiest &properties)
 
         log.status() << "converting SSA" << messaget::eom;
         equation.convert_without_assertions(
-          property_decider.get_decision_procedure());
+          property_decider.get_decision_procedure(), nullptr);
```
peterschrammel, line 92: 🚫

src/goto-checker/multi_path_symex_checker.cpp
```diff
@@ -155,3 +166,19 @@ multi_path_symex_checkert::localize_fault(const irep_idt &property_id) const
...
   return fault_localizer(property_id);
 }
+
+void multi_path_symex_checkert::report()
+{
+  if(options.is_set("write-solver-stats-to"))
+  {
+    if(
+      auto hardness_collector =
+        dynamic_cast<hardness_collectort *>(&property_decider.get_prop()))
```
peterschrammel, line 176: See below. `propt` must not be exposed. Forward to `prop_conv_solvert` which is a `decision_proceduret`.

src/goto-checker/goto_symex_property_decider.cpp
```diff
@@ -110,6 +110,11 @@ goto_symex_property_decidert::get_decision_procedure() const
   return solver->decision_procedure();
 }
 
+propt &goto_symex_property_decidert::get_prop() const
```
peterschrammel, line 113: Unnecessary.

src/goto-checker/bmc_util.h
```diff
@@ -36,7 +36,7 @@ class ui_message_handlert;
 
 void convert_symex_target_equation(
   symex_target_equationt &,
-  decision_proceduret &,
+  goto_symex_property_decidert &,
```
peterschrammel, line 39: 🚫

src/goto-checker/bmc_util.cpp
```diff
@@ -149,14 +150,22 @@ void output_graphml(
...
-  decision_proceduret &decision_procedure,
+  goto_symex_property_decidert &property_decider,
   message_handlert &message_handler)
 {
   messaget msg(message_handler);
   msg.status() << "converting SSA" << messaget::eom;
 
   // convert SSA
-  equation.convert(decision_procedure);
+  if(dynamic_cast<boolbvt *>(&property_decider.get_decision_procedure()))
```
peterschrammel, line 160: 🚫  unnecessary

peterschrammel (changes requested):
From my conversion with @xbauch, the forwarding from `propt` to `prop_conv_solvert` was missed. Doing that removes most of the ugly plumbing from this PR.

hannes-steffenhagen-diffblue:
Closing in favour of https://github.com/diffblue/cbmc/pull/5266

PR #2620, opened 2018-07-26
More precise function pointer removal [depends-on: #2504]

polgreen:
Currently the function pointer removal replaces each function pointer with a case split between all functions that have the same signature and have had their address taken. 

This PR is the beginning of work to refine that.  Submitting as PR for early feedback. 

This PR:
- scans the goto program to find all assigns to function pointers
- builds map of function pointers -> sets of functions
- uses these maps when replacing the function pointers

Currently only works for:
- standalone function pointers
- function pointers 1 layer deep in a struct

If a function pointer is written to with anything other than a straight forward address of a function, default to original behaviour. 

Current unsoundness:
- if a function pointer is part of a struct, and the struct is written to with a struct, this needs to be detected.

PR #1385, opened 2017-09-14
test for function pointer removal

kroening:
This tests the interplay between the analysis in get_called_functions and remove_function_pointers.

PR #5168, opened 2019-10-28
Add function pointer restriction option

hannes-steffenhagen-diffblue:
This adds a new option, --restrict-function-pointer, to cbmc. This lets a user
specify a list of possible pointer targets for calls to a particular function
pointer variable, rather than have remove_function_pointers guess possible
values. The intended purpose behind this is to prevent excessive symex time
wasted on exploring paths the user knows the program can never actually take.

### DIFF
--- a/doc/cprover-manual/restrict-function-pointer.md
+++ b/doc/cprover-manual/restrict-function-pointer.md
@@ -1,0 +1,148 @@
+[CPROVER Manual TOC](../../)
+
+## Restricting function pointers
+
+### Motivation
+
+CBMC comes with a way to resolve calls to function pointers to direct function
+calls. This is needed because symbolic execution itself can't handle calls to
+function pointers. In practice, this looks something like this:
+
+Given that there are functions with these signatures available in the program:
+
+```
+int f(int x);
+int g(int x);
+int h(int x);
+```
+
+And we have a call site like this:
+
+```
+typedef int(*fptr_t)(int x);
+void call(fptr_t fptr) {
+  int r = fptr(10);
+  assert(r > 0);
+}
+```
+
+Function pointer removal will turn this into code similar to this:
+
+```
+void call(fptr_t fptr) {
+  int r;
+  if(fptr == &f) {
+    r = f(10);
+  } else if(fptr == &g) {
+    r = g(10);
+  } else if(fptr == &h) {
+    r = h(10);
+  } else {
+    // sanity check
+    assert(false);
+    assume(false);
+  }
+  return r;
+}
+```
+
+This works well enough for simple cases. However, this is a very simple
+replacement only based on the signature of the function (and whether or not they
+have their address taken somewhere in the program), so if there are many
+functions matching a particular signature, or if some of these functions are
+expensive in symex (e.g. functions with lots of loops or recursion), then this
+can be a bit cumbersome - especially if we, as a user, already know that a
+particular function pointer will only resolve to a single function or a small
+set of functions. This is what the `--restrict-function-pointer` option allows.
+
+### Example
+
+Take the motivating example. Let us assume that we know for a fact that `call`
+will always receive pointers to either `f` or `g` during actual executions of
+the program, and symbolic execution for `h` is too expensive to simply ignore
+the cost of its branch. For this, we will label the places in each function
+where function pointers are being called, to this pattern:
+
+```
+<function-name>.function_pointer_call.<N>
+```
+
+where `N` is referring to which function call it is - so the first call to a
+function pointer in a function will have `N=1`, the 5th `N=5` etc.
+
+We can call `cbmc` with `--restrict-function-pointer
+call.function_pointer_call.1/f,g`. This can be read as
+
+> For the first call to a function pointer in the function `call`, assume that
+> it can only be a call to `f` or `g`
+
+The resulting code looks similar to the original example, except now there will
+not be a call to `h`:
+
+```
+void call(fptr_t fptr) {
+  int r;
+  if(fptr == &f) {
+    r = f(10);
+  } else if(fptr == &g) {
+    r = g(10);
+  } else {
+    // sanity check
+    assert(false);
+    assume(false);
+  }
+  return r;
+}
+```
+
+Another example: Imagine we have a simple virtual filesystem API and implementation
+like this:
+
+```
+typedef struct filesystem_t filesystem_t;
+struct filesystem_t {
+  int (*open)(filesystem_t *filesystem, const char* file_name);
+};
+
+int fs_open(filesystem_t *filesystem, const char* file_name) {
+  filesystem->open(filesystem, file_name);
+}
+
+int nullfs_open(filesystem_t *filesystem, const char* file_name) {
+  return -1;
+}
+
+filesystem_t nullfs_val = {.open = nullfs_open};
+filesystem *const nullfs = &nullfs_val;
+
+filesystem_t *get_fs_impl() {
+  // some fancy logic to determine
+  // which filesystem we're getting -
+  // in-memory, backed by a database, OS file system
+  // - but in our case, we know that
+  // it always ends up being nullfs
+  // for the cases we care about
+  return nullfs;
+}
+int main(void) {
+  filesystem_t *fs = get_fs_impl();
+  assert(fs_open(fs, "hello.txt") != -1);
+}
+```
+
+In this case, the assumption is that *we* know that in our `main`, `fs` can be
+nothing other than `nullfs`; But perhaps due to the logic being too complicated
+symex ends up being unable to figure this out, so in the call to `fs_open()` we
+end up branching on all functions matching the signature of
+`filesystem_t::open`, which could be quite a few functions within the program.
+Worst of all, if it's address is ever taken in the program, as far as the "dumb"
+function pointer removal is concerned it could be `fs_open()` itself due to it
+having a matching signature, leading to symex being forced to follow a
+potentially infinite recursion until its unwind limit.
+
+In this case we can again restrict the function pointer to the value which we
+know it will have:
+
+```
+--restrict-function-pointer fs_open.function_pointer_call.1/nullfs_open
+```
--- a/doc/cprover-manual/restrict-function-pointer.md
+++ b/doc/cprover-manual/restrict-function-pointer.md
@@ -146,3 +146,28 @@ know it will have:
 ```
 --restrict-function-pointer fs_open.function_pointer_call.1/nullfs_open
 ```
+
+### Loading from file
+
+If you have many places where you want to restrict function pointers, it'd be a
+nuisance to have to specify them all on the command line. In these cases, you
+can specify a file to load the restrictions from instead, via the
+`--function-pointer-restrictions-file` option, which you can give the name of a
+JSON file with this format:
+
+```
+{
+  "function_call_site_name": ["function1", "function2", ...],
+   ...
+}
+```
+
+**Note:** If you pass in multiple files, or a mix of files and command line
+restrictions, the final restrictions will be a set union of all specified
+restrictions.
+
+**Note:** as of now, if something goes wrong during type checking (i.e. making
+sure that all function pointer replacements refer to functions in the symbol
+table that have the correct type), the error message will refer the command line
+option `--restrict-function-pointer` regardless of whether the restriction in
+question came from the command line or a file.
--- a/regression/cbmc/restrict-function-pointer-to-single-function/test.c
+++ b/regression/cbmc/restrict-function-pointer-to-single-function/test.c
@@ -1,0 +1,40 @@
+#include <assert.h>
+
+typedef int (*fptr_t)(int);
+
+fptr_t get_f(void);
+
+void use_f(fptr_t fptr)
+{
+  assert(fptr(10) == 11);
+}
+
+int main(void)
+{
+  use_f(get_f());
+}
+
+int f(int x)
+{
+  return x + 1;
+}
+int g(int x)
+{
+  return x;
+}
+
+// this is just here to distinguish the behavior from FP removal, which'd include
+// only f if we didn't reference g anywhere.
+int g_always_false_cond = 0;
+
+fptr_t get_f(void)
+{
+  if(!g_always_false_cond)
+  {
+    return f;
+  }
+  else
+  {
+    return g;
+  }
+}
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -60,6 +60,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <goto-programs/remove_skip.h>
 #include <goto-programs/remove_unused_functions.h>
 #include <goto-programs/remove_vector.h>
+#include <goto-programs/restrict_function_pointers.h>
 #include <goto-programs/rewrite_union.h>
 #include <goto-programs/set_properties.h>
 #include <goto-programs/show_goto_functions.h>
@@ -78,6 +79,7 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <pointer-analysis/add_failed_symbols.h>
 
+#include <goto-programs/label_function_pointer_call_sites.h>
 #include <langapi/mode.h>
 
 #include "c_test_input_generator.h"
@@ -135,6 +137,8 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
 
   cbmc_parse_optionst::set_default_options(options);
   parse_c_object_factory_options(cmdline, options);
+
+  parse_function_pointer_restriction_options_from_cmdline(cmdline, options);
 
   if(cmdline.isset("function"))
     options.set_option("function", cmdline.get_value("function"));
@@ -749,7 +753,6 @@ int cbmc_parse_optionst::get_goto_program(
     show_symbol_table(goto_model, ui_message_handler);
     return CPROVER_EXIT_SUCCESS;
   }
-
   if(cbmc_parse_optionst::process_goto_program(goto_model, options, log))
     return CPROVER_EXIT_INTERNAL_ERROR;
 
@@ -836,6 +839,15 @@ bool cbmc_parse_optionst::process_goto_program(
   // remove function pointers
   log.status() << "Removal of function pointers and virtual functions"
                << messaget::eom;
+
+  const auto function_pointer_restrictions =
+    function_pointer_restrictionst::from_options(
+      options, log.get_message_handler());
+  if(!function_pointer_restrictions.restrictions.empty())
+  {
+    label_function_pointer_call_sites(goto_model);
+    restrict_function_pointers(goto_model, function_pointer_restrictions);
+  }
   remove_function_pointers(
     log.get_message_handler(),
     goto_model,
@@ -1066,6 +1078,7 @@ void cbmc_parse_optionst::help()
     HELP_FLUSH
     " --verbosity #                verbosity level\n"
     HELP_TIMESTAMP
+    RESTRICT_FUNCTION_POINTER_HELP
     "\n";
   // clang-format on
 }
--- a/src/goto-programs/goto_program.h
+++ b/src/goto-programs/goto_program.h
@@ -1169,6 +1169,30 @@ struct pointee_address_equalt
   }
 };
 
+template <typename GotoFunctionT, typename PredicateT, typename HandlerT>
+void for_each_goto_location_if(
+  GotoFunctionT &&goto_function,
+  PredicateT predicate,
+  HandlerT handle)
+{
+  auto &&instructions = goto_function.body.instructions;
+  for(auto it = instructions.begin(); it != instructions.end(); ++it)
+  {
+    if(predicate(it))
+    {
+      handle(it);
+    }
+  }
+}
+
+template <typename GotoFunctionT, typename HandlerT>
+void for_each_goto_location(GotoFunctionT &&goto_function, HandlerT handle)
+{
+  using iterator_t = decltype(goto_function.body.instructions.begin());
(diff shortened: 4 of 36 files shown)


src/cbmc/cbmc_parse_options.cpp
```diff
@@ -78,6 +79,7 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <pointer-analysis/add_failed_symbols.h>
 
+#include <goto-programs/simplify_function_pointers.h>
```
xbauch: Move above (to the rest of `goto-program` includes).

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -750,7 +752,8 @@ int cbmc_parse_optionst::get_goto_program(
     return CPROVER_EXIT_SUCCESS;
   }
 
-  if(cbmc_parse_optionst::process_goto_program(goto_model, options, log))
+  if(cbmc_parse_optionst::process_goto_program(
+       goto_model, options, log, cmdline))
```
xbauch: Couldn't you just copy the `cmdline` part that you need into `options` instead of passing the whole thing?

src/goto-programs/goto_program.h
```diff
@@ -1169,6 +1169,30 @@ struct pointee_address_equalt
...
+  {
+    if(predicate(it))
+    {
+      handle(it);
+    }
+  }
+}
+
+template <typename GotoFunctionT, typename HandlerT>
+void for_each_goto_location(GotoFunctionT &&goto_function, HandlerT handle)
```
xbauch, line 1189: It's called `for_each_goto_location` but inside you iterate over `instructions`.

src/goto-programs/goto_program.h
```diff
@@ -1169,6 +1169,30 @@ struct pointee_address_equalt
   }
 };
 
+template <typename GotoFunctionT, typename Predicate, typename HandlerT>
+void for_each_goto_location_if(
+  GotoFunctionT &&goto_function,
+  Predicate predicate,
```
xbauch: `PredicateT` for consistency?

src/goto-programs/restrict_function_pointers.h
```diff
@@ -0,0 +1,60 @@
...
+/*******************************************************************\
+
+Module: GOTO Program Utilities
+
+Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+/// \file
+/// Given goto functions and a list function parameters or globals
```
xbauch: a list **of** function parameters

src/goto-programs/simplify_function_pointers.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#include "goto_model.h"
+
+void simplify_function_pointers(goto_modelt &goto_model);
```
xbauch: This probably deserves documenting as well.

src/goto-programs/simplify_function_pointers.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <util/fresh_symbol.h>
+#include <util/simplify_expr.h>
+
+void simplify_function_pointers(goto_modelt &goto_model)
+{
+  for(auto &goto_function : goto_model.goto_functions.function_map)
+  {
+    for_each_goto_location(
+      goto_function.second, [&](goto_programt::targett it) {
+        if(it->is_function_call())
```
xbauch: Not sure it would help much, but you could use the `for_each_goto_location_if` and split the `is_function_call /\ is_dereference` as `Predicate`.

src/goto-programs/simplify_function_pointers.h
xbauch: Also, is there a test for this?

src/goto-programs/simplify_function_pointers.cpp
```diff
@@ -0,0 +1,49 @@
...
+              auto const &goto_function_symbol_mode =
+                goto_model.symbol_table.lookup_ref(goto_function.first).mode;
+              auto new_function_pointer_symbol = get_fresh_aux_symbol(
+                function_pointer_dereference.pointer().type(),
+                CPROVER_PREFIX,
+                "function_ptr_tmp",
+                function_call.source_location(),
+                goto_function_symbol_mode,
+                goto_model.symbol_table);
+              auto assign_instruction = goto_programt::instructiont{ASSIGN};
```
xbauch: There is `goto_programt::make_assignment`.

src/goto-programs/simplify_function_pointers.cpp
```diff
@@ -0,0 +1,49 @@
...
+              assign_instruction.code =
+                code_assignt{new_function_pointer_symbol.symbol_expr(),
+                             function_pointer_dereference.pointer()};
+              goto_function.second.body.instructions.insert(
```
xbauch: Maybe `insert_before` would be more explicit.

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+exprt make_restricted_function_pointer_conditional(
+  const symbol_exprt &function_pointer,
+  const symbol_tablet &symbol_table,
+  const id_iteratort begin,
+  const id_iteratort end)
+{
+  PRECONDITION(begin != end);
+  auto const pointer_to_function =
+    address_of_exprt{symbol_table.lookup_ref(*begin).symbol_expr()};
+  id_iteratort const next = begin + 1;
```
xbauch: So is `const` first or is actual type first?

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+exprt make_restricted_function_pointer_conditional(
+  const symbol_exprt &function_pointer,
+  const symbol_tablet &symbol_table,
+  const id_iteratort begin,
+  const id_iteratort end)
+{
+  PRECONDITION(begin != end);
+  auto const pointer_to_function =
+    address_of_exprt{symbol_table.lookup_ref(*begin).symbol_expr()};
```
xbauch: At this point we know the names are in the symbol table?

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+      throw invalid_command_line_argument_exceptiont{
+        "not a function pointer: " + id2string(restriction.first),
+        "--restrict-function-pointer"};
+    }
+    auto const &function_type =
+      to_pointer_type(function_pointer_type).subtype();
+    if(function_type.id() != ID_code)
+    {
+      throw invalid_command_line_argument_exceptiont{
+        "not a function pointer: " + id2string(restriction.first),
```
xbauch, line 48: If it gives the same message (as the `throw` above), couldn't it be merged?

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+          "symbol not found: " + id2string(function_pointer_target),
+          "--restrict-function-pointer"};
+      }
+      auto const &function_pointer_target_type =
+        function_pointer_target_sym->type;
+      if(function_type != function_pointer_target_type)
+      {
+        throw invalid_command_line_argument_exceptiont{
+          "type mismatch: `" + id2string(restriction.first) + "' points to `" +
+            type2c(function_type, ns) + "', but restriction `" +
```
xbauch, line 68: There is also `util/format` but I'm not sure which should be prefered.

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+  return source_location;
+}
+
+template <typename Handler>
+void for_each_function_call(goto_functiont &goto_function, Handler handler)
+{
+  for_each_goto_location_if(
+    goto_function,
+    [](goto_programt::targett target) { return target->is_function_call(); },
+    [&](goto_programt::targett target) {
```
xbauch: What?

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+}
+} // namespace
+
+void restrict_function_pointers(
+  goto_modelt &goto_model,
+  const function_pointer_restrictionst &restrictions)
+{
+  typecheck_function_pointer_restrictions(goto_model, restrictions);
+
+  // we iterate over all restrictions we might have
```
xbauch: The loop below does not seem to iterate over restrictions.

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+  for(auto &goto_function : goto_model.goto_functions.function_map)
+  {
+    // for each function call, we l
```
xbauch: I don't get it.

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+          auto const &pointer_symbol = to_symbol_expr(called_function_pointer);
+          auto const restriction_iterator =
+            restrictions.restrictions.find(pointer_symbol.get_identifier());
+          if(restriction_iterator != restrictions.restrictions.end())
+          {
+            auto const &restriction = *restriction_iterator;
+            auto const is_calling_pointer_symbol =
+              to_dereference_expr(original_function_call.function())
+                .pointer() == pointer_symbol;
+            const symbol_exprt &pointer_symbol =
```
xbauch: First, can we rename this: there already is one `pointer_symbol` above? Second, can we move it below, closer to it's use?

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+      }
+      comment << target;
+    }
+    comment << ']';
+  }
+  source_location.set_comment(string2id(comment.str()));
+  source_location.set_property_class(ID_assertion);
+  // XXX I want to hide these assertions unless they fail - apparently this is
+  ///    not the correct way to do this?
+  source_location.set_hide();
```
NlightNFotis: We probably don't want this here at all.

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: Fixed thanks

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+    auto const target_names = ([&target_name_strings] {
+      auto target_names = std::vector<irep_idt>(target_name_strings.size());
+      std::transform(
+        target_name_strings.begin(),
+        target_name_strings.end(),
+        target_names.begin(),
+        string2id);
+      return target_names;
+    })();
+    for(auto const &target_name : target_names)
```
xbauch, line 291: Couldn't you run this check on `target_name_strings`?

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: I removed this function entirely, it was there for an earlier version of this that I since discarded.

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: Maybe? But I think of them as two different failure conditions that just happen to have the same text in them.

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,309 @@
...
+          "--" OPT_RESTRICT_FUNCTION_POINTER,
+          restriction_format_message);
+      }
+    }
+    if(!function_pointer_restrictions
+          .emplace(irep_idt{pointer_name}, target_names)
+          .second)
+    {
+      throw invalid_command_line_argument_exceptiont{
+        "function pointer restriction for `" + id2string(pointer_name) +
```
xbauch: `pointer_name` _is_ a `string`.

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: Vestigial comment, will fix

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: Brain stopped mid sentence

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: These two are actually the same thing, I'll remove this one

src/goto-programs/restrict_function_pointers.cpp
xbauch: Ok, (what about "not a pointer to a function").

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: I'll keep it in here for now. I'll remove this when this goes out of draft, I'm still hoping someone knows how to do this correctly.

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: removed thanks

doc/cprover-manual/restrict-function-pointer.md
```diff
@@ -0,0 +1,105 @@
...
+  } else {
+    // sanity check
+    assert(false);
+    assume(false);
+  }
+  return r;
+}
+```
+
+This works well enough for simple cases. However, this is a "dumb" replacement
```
xbauch: While this perhaps isn't customer-facing text, I think it's high enough for us to call existing methods _dumb_. What about "inefficient" or "non-optimal". (Hedging level _scientist_.)

doc/cprover-manual/restrict-function-pointer.md
```diff
@@ -0,0 +1,105 @@
...
+only based on the signature of the function, so if there are many functions
+matching a particular signature, or if some of these functions are expensive in
+symex (e.g. functions with lots of loops or recursion), then this can be a bit
+cumbersome - especially if we, as a user, already know that a particular
+function pointer will only resolve to a single function or a small set of
+functions. This is what the `--restrict-function-pointer` option allows.
+
+### Example
+
+Take the motivating example. Let us assume that we know for a fact that `call`
```
xbauch, line 60: Maybe: the motivating example **above**?

xbauch (approved):
Looks good. Maybe squash a bit.

doc/cprover-manual/restrict-function-pointer.md
martin-cs: "very over-approximate" or "too much of an over-approximation for some users".

doc/cprover-manual/restrict-function-pointer.md
```diff
@@ -0,0 +1,147 @@
...
+    // sanity check
+    assert(false);
+    assume(false);
+  }
+  return r;
+}
+```
+
+This works well enough for simple cases. However, this is a "dumb" replacement
+only based on the signature of the function, so if there are many functions
```
martin-cs: IIRC there also used to be a global analysis of "which functions have their addresses taken".

regression/cbmc/restrict-function-pointer-to-single-function/test.c
```diff
@@ -0,0 +1,39 @@
...
+int f(int x)
+{
+  return x + 1;
+}
+int g(int x)
+{
+  return x;
+}
+
+// this is just here to distinguish the behavior from FP removal, which'd include
```
martin-cs, line 26: ??? I'm not sure what this sentence is saying.

src/goto-programs/goto_program.h
martin-cs: I'm not against these but please note the amount of hatred that `forall_goto_program_instructions` et al. used to get and try to get some kind of agreement on whether we should be doing more or less of this.

martin-cs (commented):
I am not opposed to this from a user-space point of view, nor the code in general.  However there are lots of open issues to do with function pointer removal:

https://github.com/diffblue/cbmc/pull/4536
https://github.com/diffblue/cbmc/pull/4650
https://github.com/diffblue/cbmc/pull/2620
https://github.com/diffblue/cbmc/pull/1385
https://github.com/diffblue/cbmc/pull/915

(to mention just a few) and it would be good to try and get some consistency / order / plan here.

As far as I can see, the common theme is to separate the code into two parts:

A. things that get candidate sets of targets for a function pointer : from a user, from the signature, from the signature plus some kind of 'is dirty' analysis for when functions have their address taken, from the expression and the symbol table, from the abstract interpreter, etc.

B. things that use these candidate sets : for transforming the program, for doing AI with the function pointers still in, etc.

Adding this interface / separation should not be too hard.

doc/cprover-manual/restrict-function-pointer.md
hannes-steffenhagen-diffblue: @martin-cs I believe this is still in place actually, so maybe I should update this to be more accurate? On the other hand, I don't really want to replicate the documentation of remove-function-pointers (if it exists) here.

regression/cbmc/restrict-function-pointer-to-single-function/test.c
hannes-steffenhagen-diffblue: That's because I didn't finish the sentence - this comes back to how FP removal works now, which is to scan for

1. matching functions
2. which have their addresses taken

so I'm adding this so it looks to FP removal as if an `int(*)(int)` could also pint to `g` rather than just to `f`.

src/goto-programs/goto_program.h
hannes-steffenhagen-diffblue: I've added this here for now, but I don't feel too strongly about this - I'll remove this if people don't like it.

doc/cprover-manual/restrict-function-pointer.md
martin-cs: It should be accurate and it should be consolidated with any other documentation for this.  I don't think there is any but do have a grep through the manual at least.

regression/cbmc/restrict-function-pointer-to-single-function/test.c
martin-cs: OK, that makes a little more sense.

src/goto-programs/goto_program.h
martin-cs: I'm fine with it but I know it was a big source of argument at one point (I don't remember what was decided TBH) so be a little careful.  I think it might be in the coding standard...

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 860117dfcbfe08d6d397620211a6817b4d781cc2).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/135089826

kroening:
A use-case question: This feature appears to be something that requires a specialist user, and I therefore don't expect wide usage. How about adding this to goto-instrument, to avoid adding complexity to CBMC's interface?
(Independently, CBMC's symbolic execution ought to learn how to do function pointers, but that is outside of the scope here).

hannes-steffenhagen-diffblue:
@kroening I don't particularly mind; However the goal was for these mappings to be generated by something like e.g. goto-analyzer most of the time, so they hopefully shouldn't require too much expertise to use (hand writing these may be useful in some circumstances as well, but that wasn't the main point).

karkhaz:
We don't mind having this in goto-instrument at all. In general, we don't expect to be generating anything: often we have a good idea of the function that we want to restrict to.

Thank you very much for this work, it looks great, and especially thank you for the documentation and test cases. Here are a couple of cases I'm not sure about.

```c
#include <assert.h>

typedef int (*fptr_t)(int);

void call(fptr_t fptr)
{
  assert(fptr(10) == 10);
}

int f(int x)
{
  return x;
}
int g(int x)
{
  return x + 1;
}
int h(int x)
{
  return x + 2;
}

int main(void)
{
  call(nondet_int() ? f : g);
}
```
```sh
> cbmc test.c --restrict-function-pointer call.function_pointer_call.1/f
** Results:
test.c function call
[call.assertion.2] line 7 assertion fptr(10) == 10: SUCCESS
[call.assertion.1] line 7 dereferenced function pointer at call.function_pointer_call.1 must be f: FAILURE
```
☝️is that expected? I'm restricting to `f`, and `f` has the right type and potentially could be the right function, so why is it failing?

Also, if I change the command line to restrict to `h` rather than `f` (which really is an impossible candidate, since we're only choosing between `f` and `g` in `main`), I get an almost identical error message:
```
[call.assertion.1] line 7 dereferenced function pointer at call.function_pointer_call.1 must be h: FAILURE
```
Would it be possible for this to be distinct, to make it clear that this is the case of a user error passing in an impossible function?

hannes-steffenhagen-diffblue:
@karkhaz thanks for having a look at this. The assertion

```
dereferenced function pointer at call.function_pointer_call.1 must be f
```

is intended to mean “the function pointer we got passed in was, indeed, a pointer to `f`”. This is a safeguard to prevent accidental misuse, so for instance in the example you quoted the function _could_ also be `g`. The option generates code like this:

```
if(fptr == &f) { f(args...); }
else if...
else { 
  assert(false); // we assume the set of restrictions is a superset of the set of actually possible targets
  assume(false); // if this assumption is wrong we can’t guarantee anything about the following code
}
```

The `assume(false)` is necessary because if the restriction of function pointers was incorrect, then we can’t say anything meaningful about that case anymore. The assertion is just there for documentation.

In that regard it is really no different than how the already existing `remove_function_pointers` works, this is intended to mirror that behaviour except that there is user control over the restriction.

PR #5109, opened 2019-09-15
Protect exprt::opX()

kroening:
This is a big change.

The key point is to prevent direct access to `exprt::opX()` -- the expectation is that the relevant specializations are used, which continue to offer these methods.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- n/a Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #5170, opened 2019-10-30
Make access to exprt::opN bounds checked

hannes-steffenhagen-diffblue:
These not being bounds checked is a potential source for pretty annoying bugs
where a non-existent operand of a malformed irep is accessed, a reference to it
passed around, causing the program to crash in a potentially quite distant
location when it eventually needs to be actually dereferenced.

Of course ideally we don't want any malformed Ireps at all, but sometimes we do end up getting them anyway because we don't do extensive structure invariant checking everywhere all the time, so it helps to be defensive with these.

### DIFF
--- a/src/util/expr.h
+++ b/src/util/expr.h
@@ -99,28 +99,44 @@ public:
   { return (const operandst &)get_sub(); }
 
   exprt &op0()
-  { return operands().front(); }
+  {
+    return operands().at(0);
+  }
 
   exprt &op1()
-  { return operands()[1]; }
+  {
+    return operands().at(1);
+  }
 
   exprt &op2()
-  { return operands()[2]; }
+  {
+    return operands().at(2);
+  }
 
   exprt &op3()
-  { return operands()[3]; }
+  {
+    return operands().at(3);
+  }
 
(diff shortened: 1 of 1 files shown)


kroening:
There is no need for that once #5109 is merged.

hannes-steffenhagen-diffblue:
@kroening Theoretically there shouldn't be, but that's only if we never violate our structure invariants anywhere - e.g. "`code_function_callt` has exactly 3 operands" etc. We've had several cases just this year when this happened in generated code - and of course it shouldn't, but these things not being checked made it much harder to find the problem once it did go wrong.

kroening:
@hannes-steffenhagen-diffblue
If there is an exprt-derived type that you consider at risk for this, then simply derive those from `multi_ary_exprt` -- the accesses to `opX()` will then be range checked.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 10db3ccb7090b70008ee5035a50da512c112ace1).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/134254883

PR #5171, opened 2019-11-04
Allow function pointer arguments to be equal

xbauch:
for the function-call goto-harness.

We add a new command-line option `treat-pointers-equal` that takes a list of
formal parameter names and results in a harness that only generates constructor
for one of the pointers. The other pointers are simply assigned with the
constructed one.

### DIFF
--- a/src/goto-harness/function_call_harness_generator.cpp
+++ b/src/goto-harness/function_call_harness_generator.cpp
@@ -169,6 +169,10 @@ void function_call_harness_generatort::handle_option(
     p_impl->function_parameters_to_treat_as_cstrings.insert(
       values.begin(), values.end());
   }
+  else if(option == FUNCTION_HARNESS_GENERATOR_TREAT_POINTERS_EQUAL_MAYBE_OPT)
+  {
+    p_impl->recursive_initialization_config.arguments_may_be_equal = true;
+  }
   else
   {
     throw invalid_command_line_argument_exceptiont{
@@ -219,9 +223,9 @@ void function_call_harness_generatort::implt::generate(
   call_function(arguments, function_body);
   for(const auto &global_pointer : global_pointers)
   {
-    function_body.add(code_function_callt{
-      recursive_initialization->get_free_function(), {global_pointer}});
+    recursive_initialization->free_if_possible(global_pointer, function_body);
   }
+  recursive_initialization->free_cluster_origins(function_body);
   add_harness_function_to_goto_model(std::move(function_body));
 }
 
--- a/src/goto-harness/function_call_harness_generator.cpp
+++ b/src/goto-harness/function_call_harness_generator.cpp
@@ -48,8 +48,8 @@ struct function_call_harness_generatort::implt
   std::set<irep_idt> function_parameters_to_treat_as_arrays;
   std::set<irep_idt> function_arguments_to_treat_as_arrays;
 
-  std::set<irep_idt> function_parameters_to_treat_equal;
-  std::set<irep_idt> function_arguments_to_treat_equal;
+  std::vector<std::set<irep_idt>> function_parameters_to_treat_equal;
+  std::vector<std::set<irep_idt>> function_arguments_to_treat_equal;
 
   std::set<irep_idt> function_parameters_to_treat_as_cstrings;
   std::set<irep_idt> function_arguments_to_treat_as_cstrings;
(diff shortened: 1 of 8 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 9db8bd4aef5c103c755cedabff4cbf9746bac26f).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/134853929

NlightNFotis (approved):
LGTM

src/goto-harness/recursive_initialization.cpp
```diff
@@ -763,3 +783,17 @@ code_blockt recursive_initializationt::build_dynamic_array_constructor(
 
   return body;
 }
+
+bool recursive_initializationt::needs_freeing(const exprt &expr) const
+{
+  if(expr.type().id() != ID_pointer)
```
hannes-steffenhagen-diffblue: Function pointers also shouldn't be freed but we're already fixing this in a follow up PR

hannes-steffenhagen-diffblue (commented):
This is currently only capable of expressing that a single set of function params should be treated equal, I'd like this to be able to specify multiple (non-overlapping I'd hope?) sets of possibly equal pointers.

Also, these should be _maybe_ equal, but as currently specified they'll _definitely_ be equal, which is a bit unfortunate (as now we won't be covering the case where they're not equal, we'd have to generate a different harness for these again).

xbauch:
@hannes-steffenhagen-diffblue both sets of equal pointers and conditional equality should be implemented now.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: dcf0ea3a5789ebcf3b8214002f0f71ec621f2662).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/136572662

src/goto-harness/function_call_harness_generator.cpp
```diff
@@ -112,6 +115,21 @@ void function_call_harness_generatort::handle_option(
     p_impl->function_parameters_to_treat_as_arrays.insert(
       values.begin(), values.end());
   }
+  else if(option == FUNCTION_HARNESS_GENERATOR_TREAT_POINTERS_EQUAL_OPT)
+  {
+    for(auto const &value : values)
+    {
+      for(auto const &param_cluster : split_string(value, ';'))
```
hannes-steffenhagen-diffblue, line 122: Super minor nitpick that `;` might not necessarily be the best separator to use here because on most shells it’s also a terminator, but tbh I don’t think it’s big enough of a problem to worry about here.

src/goto-harness/recursive_initialization.cpp
```diff
@@ -763,3 +817,52 @@ code_blockt recursive_initializationt::build_dynamic_array_constructor(
 
   return body;
 }
+
+bool recursive_initializationt::needs_freeing(const exprt &expr) const
+{
+  if(expr.type().id() != ID_pointer || expr.type().id() == ID_code)
```
hannes-steffenhagen-diffblue: This could just be `return expr.type().id() = ID_pointer && expr.type().id() != ID_code;`

src/goto-harness/recursive_initialization.cpp
```diff
@@ -280,6 +321,19 @@ bool recursive_initializationt::should_be_treated_as_array(
          initialization_config.pointers_to_treat_as_arrays.end();
 }
 
+optionalt<std::size_t>
```
hannes-steffenhagen-diffblue: Small preference to have a (strong?) typedef for this `size_t` so we know that it refers to an “equal cluster”

src/goto-harness/recursive_initialization.cpp
```diff
@@ -763,3 +817,52 @@ code_blockt recursive_initializationt::build_dynamic_array_constructor(
...
+    return;
+  }
+
+  if(
+    to_symbol_expr(*common_arguments_origins[*maybe_cluster_index])
+        .get_identifier() != expr_id &&
+    initialization_config.arguments_may_be_equal)
+  {
+    // in equality cluster but not common origin -> free if equal to origin
+    const auto condition =
```
hannes-steffenhagen-diffblue, line 847: I don’t quite understand the logic, shouldn’t it be the other way round (i.e. don’t free if not common origin, unless your value is different than origin)?

src/goto-harness/recursive_initialization.cpp
xbauch: Done.

src/goto-harness/recursive_initialization.cpp
xbauch: Done.

src/goto-harness/recursive_initialization.cpp
xbauch: Off by negation: good catch. Fixed.

hannes-steffenhagen-diffblue (approved):

PR #5182, opened 2019-11-11
Selective havocing in generated function bodies

xbauch:
Adds option `--generate-havocing-body 'fun,1,params:0;2,2,params:1` for
body-less function `fun` generates two bodies `fun.1` and `fun.2` and re-names
the respective call-sites. The first body havoc(s) the 0th and 2nd parameter,
the second body havoc(s) the 1st parameter.

The ability to selectively havoc parameter by their numbers was added to the
existing option `generate-function-body`.

### DIFF
--- a/doc/cprover-manual/goto-instrument.md
+++ b/doc/cprover-manual/goto-instrument.md
@@ -1,0 +1,127 @@
+[CPROVER Manual TOC](../../)
+
+## Goto Instrument
+
+This is a tool that implements a variety of instrumentation passes modifying a
+pre-existing GOTO binary, or rather producing the modified binary as output.
+
+```sh
+$ goto-instrument in.gb out.gb <instrumentation-options>
+```
+
+### Diagnosis Options
+
+### Safety Checks
+
+### Semantic Transformations
+
+#### Generate Function Body
+
+This transformation inserts implementation of functions without definition, i.e.
+a body. The user is allowed to choose the behavior to be implemented by those
+functions from a predefined list. The command-line option to select the behavior
+is `--generate-function-body-options <option>`.
+
+- `assert-false`: The body consists of a single command: `assert(0);`
+- `assume-false`: The body consists of a single command: `assume(0);`
+- `nondet-return`: The body only returns a non-deterministic value of it's
+  return type.
+- `assert-false-assume-false`: Two commands as above.
+- `havoc[,params:<p-regex>][,globals:<g-regex>]`: Assign non-deterministic
+  values to the targets of pointer-to-non-constant parameters matching the
+  `p-regex`, and non-constant globals matching the `g-regex` and then (in case
+  of non-void function) returning as with `nondet-return`.
+
+Let's demonstrate using the havoc on a simple example:
+
+```C
+// main.c
+int global;
+const int c_global;
+
+int function(int *param, const int *c_param);
+```
+
+Often we want to avoid overwriting internal symbols, i.e. those with `__`
+prefix, which is done using the pattern `(?!__)`.
+
+```sh
+$ goto-cc main.c -o main.gb
+$ goto-instrument main.gb main-out.gb \
+  --generate-function-body-options 'havoc,params:(?!__).*,globals:(?!__).*' \
+  --generate-funtion-body function
+```
+
+Leads to a binary equivalent to the following C code:
+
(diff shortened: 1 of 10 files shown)


allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 4d090089fb943532d142c0e8a0f44871e474236b).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/135982385

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 373a2a0f5e6a4f4a462bd08e0996a499a66ace05).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/135988814

src/goto-instrument/generate_function_bodies.h
```diff
@@ -70,21 +70,36 @@ void generate_function_bodies(
...
 #define OPT_REPLACE_FUNCTION_BODY \
   "(generate-function-body):" \
+  "(generate-havocing-body):" \
   "(generate-function-body-options):"
 
 #define HELP_REPLACE_FUNCTION_BODY \
   " --generate-function-body <regex>\n" \
   /* NOLINTNEXTLINE(whitespace/line_length) */ \
   "                              Generate bodies for functions matching regex\n" \
+  " --generate-havocing-body <option>\n" \
```
hannes-steffenhagen-diffblue, line 97: There is a `havoc` in generate function body options, is there a reason why we need a separate option for this?

doc/cprover-manual/goto-instrument.md
```diff
@@ -0,0 +1,127 @@
...
+}
+```
+
+Which parameters should be modified can be specified either by a regular
+expression (as above) or by a semicolon-separated list of their numbers. For
+example `havoc,params:0;3;4` will assign non-deterministic values to the 1st,
+4th, and 5th parameter.
+
+Finally, the user can also request different implementation for a number of
+call-sites of a single function. The option `--generate-havocing-body` inserts
```
hannes-steffenhagen-diffblue, line 72: Ah, that makes more sense then! But isn’t the name `generate-havocing-body` a bit misleading then? I don’t know.

hannes-steffenhagen-diffblue (approved):
Seems reasonable to me I guess

doc/cprover-manual/goto-instrument.md
```diff
@@ -0,0 +1,127 @@
...
+  *param = nondet_int();
+  global = nondet_int();
+  return nondet_int();
+}
+```
+
+Which parameters should be modified can be specified either by a regular
+expression (as above) or by a semicolon-separated list of their numbers. For
+example `havoc,params:0;3;4` will assign non-deterministic values to the 1st,
+4th, and 5th parameter.
```
chrisr-diffblue, line 69: Is that 1st, 4th and 5th parameters regardless of type, or 1st, 4th and 5th *pointer* parameters? e.g. is it type aware or not?

chrisr-diffblue (commented):
Looks reasonable, I just have a query about how typesafe this is - for instance, if the users asks to havoc parameter 1, but it turns out that parameter 1 is not a pointer type, what happens? Do they get a nice user error message, or does '1' mean 'first pointer parameter, even if there are 5 int parameters before it' ?

doc/cprover-manual/goto-instrument.md
xbauch: So the implementation is type-safe in that it silently ignores the user-selected numbers if the respective parameter is not a pointer. Neither does it produce any error even if the user number parameter is higher than the number of parameters a given function takes. Do you think it should explicitly fail, or should we just extend the documentation?

chrisr-diffblue: From an end-user point of view "silently ignore and not do what I meant" is rather annoying - so I actually think both should be done - i.e. documentation should be extended a bit, and then the tool should give an explicit error if the user has specified something that "doesn't make sense".

xbauch: Done.

chrisr-diffblue (approved):
Looks good - many thanks for the additional documentation and user error messages - and for being diligent about adding additional regression tests for those new error messages :-)

PR #5198, opened 2019-11-27
Disable GCC 8's class-memaccess warning so that MiniSAT builds

martin-cs:
MiniSAT does a number of performance critical but unusual things
with memory allocation and layout.  These trigger a new warning
that has been added to GCC 8, class-memaccess and as we are using
-Werror, this means that CPROVER doesn't build using GCC 8.
I think it is better to disable this warning (which only seems to
be triggered in MiniSAT and by code that wouldn't pass review)
than to change the way MiniSAT works with memory.

### DIFF
--- a/src/config.inc
+++ b/src/config.inc
@@ -5,7 +5,7 @@ BUILD_ENV = AUTO
 ifeq ($(BUILD_ENV),MSVC)
   #CXXFLAGS += /Wall /WX
 else
-  CXXFLAGS += -Wall -pedantic -Werror -Wno-deprecated-declarations -Wswitch-enum
+  CXXFLAGS += -Wall -pedantic -Werror -Wno-deprecated-declarations -Wswitch-enum -Wno-class-memaccess
 endif
 
 ifeq ($(CPROVER_WITH_PROFILING),1)


martin-cs:
OK, so this patch is a laughably "works for me" attempt as it looks like clang is having none of this.  What do people think about a way forward?

martin-cs:
It is just the clang builds that are failing.  Is there a way to get clang to silently ignore this flag or do we have to have separate flag strings for GCC and clang?

hannes-steffenhagen-diffblue:
You’ll want to put adding this flag behind a check; I don’t know the best way to do it, but maybe something like (didn’t try, so maybe I got the syntax slightly wrong)

```
# as before
CXXFLAGS += -Wall -pedantic -Werror -Wno-deprecated-declarations -Wswitch-enum
ifeq ($($(CXX) -v 2>&1 | grep -q clang && echo 1 || echo 0), 1)
  CXXFLAGS += -Wno-class-memaccess
endif
```

But really, this is the problem with blanket `-Werror`, there’s really no good reason why upgrading to a compiler with better diagnostics should suddenly make your code not compile...

chrisr-diffblue:
I guess one strategy (maybe controversial) would be to have -Werror on by default for CI builds, but off for local builds.... That may be an argument that has already been had though?

martin-cs:
@hannes-steffenhagen-diffblue : thanks, this sounds like a way forward.

@chrisr-diffblue : possible but eventually we might want to move CI on to GCC 8.0 and we don't know what compiler users have so I would rather something that worked as consistently and similarly as possible on all platforms.

martin-cs:
I can't seem to reproduce the issues that first caused this.  `g++-8` does definitely perform this check:

```
$ cat g++-8-class-memaccess.cpp 
#include <string>
#include <cstring>

int main (int argc, char **argv){
  std::string str = "abc";
  memset (&str, 0, sizeof str);
  return 0;
}
$ g++ --version
g++ (Debian 8.3.0-6) 8.3.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

martin@michael:~/tmp/can-delete$ g++ g++-8-class-memaccess.cpp -Wall
g++-8-class-memaccess.cpp: In function ‘int main(int, char**)’:
g++-8-class-memaccess.cpp:6:30: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘std::__cxx11::string’ {aka ‘class std::__cxx11::basic_string<char>’} with no trivial copy-assignment; use assignment or value-initialization instead [-Wclass-memaccess]
   memset (&str, 0, sizeof str);
                              ^
In file included from /usr/include/c++/8/string:52,
                 from g++-8-class-memaccess.cpp:1:
/usr/include/c++/8/bits/basic_string.h:77:11: note: ‘std::__cxx11::string’ {aka ‘class std::__cxx11::basic_string<char>’} declared here
     class basic_string
           ^~~~~~~~~~~~
$ 
```
https://gcc.gnu.org/onlinedocs/gcc-8.4.0/gcc/C_002b_002b-Dialect-Options.html#index-Wclass-memaccess

and the build system hasn't been changed, nor has the Minisat patch so I am somewhat at a loss as to how this is no longer happening.

PR #5200, opened 2019-12-05
Instrument strchr to use the string refinement solver

xbauch:
Reusing and extending the Java-based `indexOf`. The string refinement assumes that certain associations are made between the char-pointer input, (some) array string (, and it's length). There are also assumption about exactly what kinds of expressions can be passed to the string solver representing string. We added some special handling for `byte-extract` expressions and `string-constants`. These modification should not change the java part.

`strchr` choice is just for demonstration, the intent is to implement all/most of the `string.h` functions.


regression/cbmc/string-refinement-strchr2/test.c
```diff
@@ -0,0 +1,13 @@
+#include <assert.h>
+#include <string.h>
+
+int main()
+{
+  int i;
+  char *test = i ? "fo\0tis" : "notfoti";
```
hannes-steffenhagen-diffblue, line 7: There should be a test for the case when a character can’t be found as well, e.g. in this case `'t'` may or may not be found depending on `i`.

thk123:
Note, this currently fails CI on Travis on the following test:

```
Running string-refinement-strchr2/test.desc  [FAILED]
```

Specifically fails to find:
```
^\[main.assertion.\d+\] line \d+ assertion first_o != NULL: SUCCESS$
```
Instead:

```
[main.assertion.1] line 10 assertion first_o != ((void *)0): SUCCESS
```

I guess different platform, different way of printing null?

src/solvers/strings/string_refinement_util.cpp
```diff
@@ -14,19 +14,27 @@ Author: Diffblue Ltd.
...
 bool is_char_type(const typet &type)
 {
-  return type.id() == ID_unsignedbv && to_unsignedbv_type(type).get_width() <=
-                                         STRING_REFINEMENT_MAX_CHAR_WIDTH;
+  if(type.id() == ID_unsignedbv)
+    return to_unsignedbv_type(type).get_width() <=
+           STRING_REFINEMENT_MAX_CHAR_WIDTH;
+
+  if(type.id() == ID_signedbv)
+    return to_signedbv_type(type).get_width() <= 8;
```
thk123, line 35: :no_entry_sign: Magic number here should have some explanation

src/solvers/strings/string_refinement_util.cpp
```diff
@@ -192,3 +200,68 @@ array_exprt interval_sparse_arrayt::concretize(
...
 }
+
+exprt maybe_byte_extract_expr(const exprt &expr)
+{
+  if(!can_cast_expr<byte_extract_exprt>(expr))
+    return expr;
+
+  const auto &byte_extract_expr = to_byte_extract_expr(expr);
+  const auto &offset = byte_extract_expr.offset();
+  PRECONDITION(offset.id() == ID_constant);
```
thk123, line 211: :grey_question: how come this is a precondition? Seems like should be documented at least

src/solvers/strings/string_refinement_util.cpp
```diff
@@ -192,3 +200,68 @@ array_exprt interval_sparse_arrayt::concretize(
...
+  if(offset.id() != ID_constant)
```
thk123, line 212: :question: Unreachable? You've just asserted this isn't true

src/solvers/strings/string_refinement_util.cpp
```diff
@@ -192,3 +200,68 @@ array_exprt interval_sparse_arrayt::concretize(
     size, it == entries.end() ? default_value : it->second);
   return array;
 }
+
+exprt maybe_byte_extract_expr(const exprt &expr)
```
thk123, line 204: :question: What's "maybe" about this? Would returning an optional make sense here?

src/solvers/strings/string_refinement_util.cpp
```diff
@@ -192,3 +200,68 @@ array_exprt interval_sparse_arrayt::concretize(
...
+    op.operands().end());
+  const auto extracted_array_suffix =
+    array_exprt{offset_operands, to_array_type(op.type())};
+
+  return extracted_array_suffix;
+}
+
+exprt maybe_remove_string_exprs(const exprt &expr)
+{
+  return [&]() {
```
thk123, line 239: :question: What's the purpose of this lambda...

src/solvers/strings/string_refinement_util.cpp
thk123, line 204: :bulb: This function is non-trivial, I'd consider writing a unit test for it to check it is doing everything you expect

src/solvers/strings/string_refinement_util.cpp
```diff
@@ -192,3 +200,68 @@ array_exprt interval_sparse_arrayt::concretize(
...
+      return static_cast<const exprt &>(maybe_string_constant->to_array_expr());
+    }
+    else
+    {
+      return expr;
+    }
+  }();
+}
+
+exprt massage_weird_arrays_into_non_weird_arrays(const exprt &expr)
```
thk123, line 253: :no_entry_sign: This of course needs documentation, or a much better name! And again, a unit test seems wise as a technical documentation of what qualifies as "weird" and "not weird" arrays

src/solvers/strings/string_constraint_generator_indexof.cpp
```diff
@@ -82,6 +82,90 @@ string_constraint_generatort::add_axioms_for_index_of(
   return {index, std::move(constraints)};
 }
 
+/// Add axioms stating that the returned value is the index within `haystack`
```
thk123, line 85: :bulb: Why not call the variables `haystack` and `needle rather than `str` and `c`

src/solvers/strings/string_constraint_generator_indexof.cpp
```diff
@@ -82,6 +82,90 @@ string_constraint_generatort::add_axioms_for_index_of(
...
+/// (`str`) of the first occurrence of `needle` (`c`) starting the search at
+/// `from_index`, or is `-1` if no such character occurs at or after position
+/// `from_index`.
+/// \todo Make argument names match whose of add_axioms_for_index_of_string
```
thk123, line 89: :no_entry_sign: un done todo I guess

src/solvers/strings/string_constraint_generator_indexof.cpp
```diff
@@ -82,6 +82,90 @@ string_constraint_generatort::add_axioms_for_index_of(
...
+{
+  string_constraintst constraints;
+  const typet &index_type = str.length_type();
+  symbol_exprt index = fresh_symbol("index_of", index_type);
+  symbol_exprt contains = fresh_symbol("contains_in_index_of");
+  symbol_exprt terminating_zero = fresh_symbol("zero_in_index_of", index_type);
+
+  exprt minus1 = from_integer(-1, index_type);
+  and_exprt a1(
+    binary_relation_exprt(index, ID_ge, minus1),
```
thk123, line 121: Is this right? `index >= -1`, so `index = -1` is valid?

src/solvers/strings/string_constraint_generator_indexof.cpp
```diff
@@ -82,6 +82,90 @@ string_constraint_generatort::add_axioms_for_index_of(
...
+    binary_relation_exprt(index, ID_le, terminating_zero),
+    binary_relation_exprt(
```
thk123, line 123: I think it would be clearer to introduce temporaies for each one of these conditions

```
const auto valid_index = binary_relation_exprt(index, ID_ge, minus1);
const auto index_in_bounds = binary_relation_exprt(index, ID_le, terminating_zero)
const auto str_terminates = binary_relation_exprt(
  terminating_zero, ID_lt, array_pool.get_or_create_length(str))
```

src/solvers/strings/string_constraint_generator_indexof.cpp
```diff
@@ -82,6 +82,90 @@ string_constraint_generatort::add_axioms_for_index_of(
...
+  const exprt &from_index)
+{
+  string_constraintst constraints;
+  const typet &index_type = str.length_type();
+  symbol_exprt index = fresh_symbol("index_of", index_type);
+  symbol_exprt contains = fresh_symbol("contains_in_index_of");
+  symbol_exprt terminating_zero = fresh_symbol("zero_in_index_of", index_type);
+
+  exprt minus1 = from_integer(-1, index_type);
+  and_exprt a1(
```
thk123, line 120: I think we can have clearer names (I guess this is cross referencing the axioms above?) Given that it is hard to read the LaTeX inline, I'd choose

`valid_index`

And reference this in the axioms above

src/solvers/strings/string_constraint_generator_indexof.cpp
```diff
@@ -82,6 +82,90 @@ string_constraint_generatort::add_axioms_for_index_of(
...
+    binary_relation_exprt(index, ID_ge, minus1),
+    binary_relation_exprt(index, ID_le, terminating_zero),
+    binary_relation_exprt(
+      terminating_zero, ID_lt, array_pool.get_or_create_length(str)));
+  constraints.existential.push_back(a1);
+
+  equal_exprt a2(not_exprt(contains), equal_exprt(index, minus1));
```
thk123, line 127: Wait, what is `a2` a reference to 2a? Where is `b1`? 

Perhaps `contains_if_negative_index`

src/solvers/strings/string_constraint_generator_indexof.cpp
```diff
@@ -82,6 +82,90 @@ string_constraint_generatort::add_axioms_for_index_of(
...
+  constraints.existential.push_back(a2);
+
+  const exprt lower_bound(zero_if_negative(from_index));
```
thk123, line 130: :bulb: Is there a way to write a unit test that "solves" simple logical formula. E.g. it'd be nice if you could write a test in the form:

```
const exprt axiom_1 = ...
values1 = {
  "index_of" : 1
  "terminating_zero": 5
  "length": 6
}
values1 = {
  "index_of" : 1
  "terminating_zero": 100
  "length": 6
}
REQUIRE(solve(axion_1, values1) == false)
```

Maybe using simplify with some substitution of the expr with the values? Would be a nice way to verify that the axioms are encoded correctly, given how fiddly it is to read/write the exprs

src/solvers/strings/string_constraint_generator_indexof.cpp
```diff
@@ -82,6 +82,90 @@ string_constraint_generatort::add_axioms_for_index_of(
...
+  // make sure that terminating zero exists (and is the smallest index after from
+  //  that has a 0 character)
+  constraints.existential.push_back(
+    equal_exprt{str[terminating_zero], from_integer(0, c.type())});
+  symbol_exprt k = fresh_symbol("QA_index_of", index_type);
```
thk123, line 135: I don't know what `QA_index_of` means, but it is surely a slightly better name than `k`

src/solvers/strings/string_constraint_generator_indexof.cpp
thk123: I think this can just be done with simplify_expr actually:

````
const exprt axiom_1 = build_axiom_1(from_integer(1), from_integer(5), from_integer(6)
REQUIRE(simplify_expr(axiom_1) == true_exprt{});

const exprt axiom_1 = build_axiom_1(from_integer(1), from_integer(100), from_integer(6)
REQUIRE(simplify_expr(axiom_1) == false_exprt{});
```

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -362,6 +362,9 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
   {
     options.set_option("refine-strings", true);
     options.set_option("string-printable", cmdline.isset("string-printable"));
+    options.set_option(
+      "max-nondet-string-length",
+      cmdline.get_value("max-nondet-string-length"));
```
thk123, line 367: Does this need a default value as is in jbmc? `MAX_CONCRETE_STRING_SIZE - 1`

thk123: Also note jbmc throws an error if you use this option without `--refine-strings`

src/goto-programs/string_instrumentation.cpp
```diff
@@ -189,6 +226,10 @@ void string_instrumentationt::operator()(goto_functionst &dest)
       it!=dest.function_map.end();
       it++)
   {
+    if(it->first == "strchr")
```
thk123, line 229: :question: This looks like it belongs in in `instrument`?

src/goto-programs/string_instrumentation.cpp
```diff
@@ -228,8 +269,6 @@ void string_instrumentationt::do_function_call(
     else if(identifier=="strxfrm")
     {
     }
-    else if(identifier=="strchr")
```
thk123, line 231: :question: Why does this move?

src/goto-programs/string_instrumentation.cpp
```diff
@@ -610,6 +649,199 @@ void string_instrumentationt::do_strncmp(
 {
 }
 
+auxiliary_symbolt string_instrumentationt::new_aux_symbol(
```
thk123, line 652: :bulb: I feel like this is duplication of something, can we reuse existing `get_fresh_aux_symbol`?

src/goto-programs/string_instrumentation.cpp
```diff
@@ -610,6 +649,199 @@ void string_instrumentationt::do_strncmp(
...
+  const symbolt return_sym_pointer =
+    new_aux_symbol("return_array_pointer", index_type(), symbol_table);
+  const auto cprover_associate_array_to_pointer_func_symbol =
+    symbol_exprt{ID_cprover_associate_array_to_pointer_func,
+                 mathematical_function_typet{
+                   {nondet_array_expr.type(), array_pointer.type()},
+                   return_sym_pointer.type}};
+  const auto apply_associate_pointer_to_array =
+    function_application_exprt{cprover_associate_array_to_pointer_func_symbol,
+                               {nondet_array_expr, char_ptr}};
```
thk123, line 753: :bulb: These concepts I assume are shared with other string stuff, perhaps we could extract a utility for producing these exprs?

src/goto-programs/string_instrumentation.cpp
```diff
@@ -610,6 +649,199 @@ void string_instrumentationt::do_strncmp(
...
+//     return NULL:
+// }
+void string_instrumentationt::do_strchr_via_c_index_of(
+  goto_functiont &strchr_function)
+{
+  //   struct refined_string_struct {
+  //     int length = string_length;
+  //     char *content = src;
+  //   } refined_string;
+  const auto &params = to_code_type(strchr_function.type).parameters();
```
thk123, line 794: :bulb: Feels like some invariants on the parameters might save some tricky debugging

src/goto-programs/string_instrumentation.cpp
```diff
@@ -625,11 +857,50 @@ void string_instrumentationt::do_strchr(
 
   goto_programt tmp;
 
-  goto_programt::targett assertion = tmp.add(goto_programt::make_assertion(
-    is_zero_string(arguments[0]), target->source_location));
-  assertion->source_location.set_property_class("string");
-  assertion->source_location.set_comment(
-    "zero-termination of string argument of strchr");
+  auto new_aux_symbol =
```
thk123, line 860: :question: Don't you have a function that does this?

src/goto-programs/string_instrumentation.h
```diff
@@ -41,16 +41,19 @@ class incorrect_source_program_exceptiont : public cprover_exception_baset
 void string_instrumentation(
   symbol_tablet &,
   message_handlert &,
-  goto_programt &);
+  goto_programt &,
+  size_t);
```
thk123, line 45: :pick: naming variables in the header (particularly ones where the type tells you nothing) helps users of the function to figure out what they should be passing in

src/solvers/strings/array_pool.cpp
```diff
@@ -7,6 +7,8 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 \*******************************************************************/
 
 #include "array_pool.h"
+#include "string_refinement_util.h"
```
thk123, line 10: :pick: ` etc.` in the commit message is less than ideal, perhaps it should be multiple commits, but should almost certainly list all the changes

src/solvers/strings/string_refinement.cpp
```diff
@@ -327,6 +329,16 @@ static void add_equations_for_symbol_resolution(
       continue;
     }
 
+    if(
+      rhs.id() == ID_symbol &&
+      has_prefix(
+        id2string(to_symbol_expr(rhs).get_identifier()), "symex::args::"))
```
thk123, line 335: hard coded string seems like a bad idea, is it defined anywhere?

src/util/pointer_offset_size.cpp
```diff
@@ -632,6 +632,17 @@ optionalt<exprt> get_subexpression_at_offset(
     return typecast_exprt(expr, target_type_raw);
   }
 
+  if(
```
thk123, line 635: :no_entry_sign: Please add why you need to do this (particularly in light of the comment yesterday about perf cost of byte_extract

thk123 (changes requested):
Neat! I have a number of suggestions of ways we can make the quite dense string solver code a bit more readable, happy to help with applying them. 

I would like to see the coverage report as there ends up being lots of new code, and I don't really know when it would be triggered, so would be good to check we're testing it all. Hopefully will just appear, but maybe needs a rebase?

src/solvers/strings/array_pool.cpp
```diff
@@ -135,12 +139,19 @@ static void attempt_assign_length_from_type(
...
+  {
+    const auto &if_expr = to_if_expr(array_expr);
+    attempt_assign_length_from_type(
+      to_array_string_expr(if_expr.true_case()),
+      length_of_array,
+      symbol_generator);
+    attempt_assign_length_from_type(
+      to_array_string_expr(if_expr.false_case()),
+      length_of_array,
+      symbol_generator);
```
romainbrenguier, line 152: that doesn't seem correct, the length could be assigned twice with different value, while it only has to be equal to one of these depending on the condition

src/solvers/strings/string_refinement_util.cpp
romainbrenguier: documentation of the function needs to be updated, and does it make sense to have signed characters?

hannes-steffenhagen-diffblue: @romainbrenguier `char` is allowed to be either `signed` or `unsigned` in C.

xbauch: Turns out the magic is not needed at all.

src/solvers/strings/string_refinement_util.cpp
hannes-steffenhagen-diffblue: I am fairly sure there isn’t one. Probably happened by accident during a refactor.

src/solvers/strings/string_refinement_util.cpp
xbauch: Well it always returns some `expr`. But it acts as identity for expressions that are not `byte-extract`.

majakusber:
# [Codecov](https://codecov.io/gh/diffblue/cbmc/pull/5200?src=pr&el=h1) Report
> Merging [#5200](https://codecov.io/gh/diffblue/cbmc/pull/5200?src=pr&el=desc) into [develop](https://codecov.io/gh/diffblue/cbmc/commit/cccdfd9ba3c1033b3634823884f21dbea030bf33&el=desc) will **increase** coverage by `0.07%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/diffblue/cbmc/pull/5200?src=pr&el=tree)

```diff
@@             Coverage Diff             @@
##           develop    #5200      +/-   ##
===========================================
+ Coverage    67.43%   67.51%   +0.07%     
===========================================
  Files         1157     1157              
  Lines        95354    95536     +182     
===========================================
+ Hits         64306    64504     +198     
+ Misses       31048    31032      -16     
```

| Flag | Coverage Δ | |
|---|---|---|
| #cproversmt2 | `42.55% <2.97%> (-0.13%)` | :arrow_down: |
| #regression | `64.05% <87.12%> (+0.09%)` | :arrow_up: |
| #unit | `31.90% <13.36%> (-0.05%)` | :arrow_down: |

| [Impacted Files](https://codecov.io/gh/diffblue/cbmc/pull/5200?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [src/goto-programs/string\_instrumentation.cpp](https://codecov.io/gh/diffblue/cbmc/pull/5200/diff?src=pr&el=tree#diff-c3JjL2dvdG8tcHJvZ3JhbXMvc3RyaW5nX2luc3RydW1lbnRhdGlvbi5jcHA=) | `28.49% <0.00%> (+28.49%)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/diffblue/cbmc/pull/5200?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/diffblue/cbmc/pull/5200?src=pr&el=footer). Last update [cccdfd9...0879ad0](https://codecov.io/gh/diffblue/cbmc/pull/5200?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

thk123:
Looks like two big areas not currently exercised:
1. `maybe_byte_extract_expr` never gets a non-zero offset ([here](https://github.com/diffblue/cbmc/pull/5200/files#diff-75a7c8067d4c7620b24df5995a81e2d9R226-R234))
2. Any of the new code for `string_instrumentationt::do_strchr` ([here](https://github.com/diffblue/cbmc/pull/5200/files#diff-46457de72391719da95a1f39445092f5R878-R903))

Not sure exactly what they'd require (possibly the 2nd one is just the code isn't needed?)

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -830,8 +833,13 @@ bool cbmc_parse_optionst::process_goto_program(
   link_to_library(
     goto_model, log.get_message_handler(), cprover_c_library_factory);
 
-  if(options.get_bool_option("string-abstraction"))
-    string_instrumentation(goto_model, log.get_message_handler());
+  if(
+    options.get_bool_option("string-abstraction") ||
```
martin-cs, line 837: Maybe I am getting mixed up but I thought the "string-abstraction" options did something very different?  Maybe that was in goto-instrument.

martin-cs (commented):
I'm not the expert on the string solver so perhaps it is not my place to say but this doesn't feel like a particularly good design for supporting a large number of string functions.  My experience suggests that you want a minimal number of (ideally orthogonal) primitives that are supported by the solver and the rest of the complexity and modelling can be in the library models.  That is a lot easier to edit, test (given an executable form of the primitives), understand and modify.

src/solvers/strings/string_constraint_generator_indexof.cpp
xbauch: Done.

src/solvers/strings/string_constraint_generator_indexof.cpp
xbauch: Done. (This was actually a copy-paste from the original `add_axioms_for_index_of` above. It still has this todo.)

src/solvers/strings/string_constraint_generator_indexof.cpp
xbauch: Yes. `index=-1` is for the case when `needle` is not in `haystack`.

romainbrenguier:
@martin-cs 
> I'm not the expert on the string solver so perhaps it is not my place to say but this doesn't feel like a particularly good design for supporting a large number of string functions. My experience suggests that you want a minimal number of (ideally orthogonal) primitives that are supported by the solver and the rest of the complexity and modelling can be in the library models. That is a lot easier to edit, test (given an executable form of the primitives), understand and modify.

Yes, I agree it would be better to avoid adding new primitives but rather reuse the existing ones and use models to deal with the differences between languages.

thk123:
When trying this branch out, I noticed the following example on this branch causes an invariant to be hit:

```C
#include <assert.h>
#include <string.h>

void testFunction() {
  int size;
  __CPROVER_assume(size >= 10 && size <= 100);
  char *str = malloc(sizeof(char) * size);
  for(int i = 0; i < size-1; ++i)
  {
    char nondet_char;
    str[i] = nondet_char;
  }
  str[size - 1] = '\0';

  char* result = strchr(str, 'z');
  assert(result - str != 15);
}
```

xbauch:
@martin-cs and @romainbrenguier there's a draft PR (#5239) which implements `strchr` by modelling the functionality inside `string.h` library and then instrumenting it to call the `index-of` directly. Have a look if you think that approach would be preferable to separating Java and C implementation. Note that there are some incompatible assumptions about types of things (array-size-type, char-type, etc.) between Java and C, so some changes to the originally java-only primitives are necessary.

PR #5202, opened 2019-12-08
Changes to sharing-map needed to merge VSD

martin-cs:
This PR consists of a series of commits taken from the variable-sensitivity-domain branch originally by @hannes-steffenhagen-diffblue and @chrisr-diffblue .  I have separated out the commits (and parts of commits) that just affect the sharing map so these can be reviewed and merged separately.

### DIFF
--- a/src/goto-checker/symex_coverage.cpp
+++ b/src/goto-checker/symex_coverage.cpp
@@ -24,6 +24,7 @@ Date: March 2016
 #include <linking/static_lifetime_init.h>
 
 #include <chrono> // IWYU pragma: keep
+#include <ctime>  // IWYU pragma: keep - For std::time_t
 #include <fstream> // IWYU pragma: keep
 #include <iostream>
 


martin-cs:
Some of these changes might now be redundant and I think it might be best to squash the commits as the messages aren't particularly meaningful now de-contextualised but I figured looking at the code might be the place to start.

martin-cs:
Note that the linter failures are unavoidable.

src/util/sharing_map.h
```diff
@@ -54,6 +54,16 @@ Author: Daniel Poetzl
     typename equalT> \
   type sharing_mapt<keyT, valueT, fail_if_equal, hashT, equalT>
 
+#define SHARING_MAPTV(return_type, V) \
```
hannes-steffenhagen-diffblue: Given that this is used in exactly one place I’d like to voice a strong preference for just inlining this macro.

src/util/sharing_map.h
```diff
@@ -413,7 +447,14 @@ class sharing_mapt
   /// - Best case: O(N + H)
   ///
   /// \param [out] view: Empty view
-  void get_view(viewt &view) const;
+  template <class V>
+  void get_view(V &) const;
+  viewt get_view() const
```
hannes-steffenhagen-diffblue: I suppose technically this should be a template as well now.

src/util/std_expr.h
```diff
@@ -121,6 +121,18 @@ class symbol_exprt : public nullary_exprt
   }
 };
 
+namespace std
```
hannes-steffenhagen-diffblue: This is probably harmless but I think it’d nevertheless be preferable if we could find a way to do without this.

(I.e. remove it going forward after VSD is merged, not necessarily here)

hannes-steffenhagen-diffblue:
Linter failures should be addressed with judicious use of `// NOLINT` to explicitly supress linting.

src/util/sharing_map.h
danpoe: The already existing macro `SHARING_MAPT4` does just the same, so I'd vote for using that one instead.

src/util/sharing_map.h
```diff
@@ -363,6 +386,17 @@ class sharing_mapt
...
+
+  static void insert_view_item(viewt &v, view_itemt &&vi)
+  {
+    v.push_back(vi);
+  }
+
+  static void insert_view_item(sorted_viewt &v, view_itemt &&vi)
+  {
+    v.insert(vi);
+  }
```
danpoe: Can we use forwarding references for these two functions? I.e.:
```
template <typename itemT>
static void insert_view_item(viewt &v, itemT &&vi)
{
  v.push_back(std::forward<itemT>(vi));
}
```

src/util/sharing_map.h
```diff
@@ -779,15 +824,15 @@ ::get_sharing_stats_map(Iterator begin, Iterator end)
 }
 #endif
 
-SHARING_MAPT(void)::get_view(viewt &view) const
+SHARING_MAPTV(void, view_type)::get_view(view_type &view) const
```
danpoe: Use `SHARING_MAPT4`

unit/util/sharing_map.cpp
```diff
@@ -492,7 +492,7 @@ TEST_CASE("Sharing map views and iteration", "[core][util]")
 
   SECTION("View")
   {
-    typedef std::pair<std::string, std::string> pt;
+    typedef std::pair<irep_idt, std::string> pt;
```
danpoe: Suggest changing this back to `std::string`. The intention was to sort the vector in lexicographic order. However, `<` for `irep_idt` sorts according to their sequence number AFAIK.

PR #5203, opened 2019-12-08
Changes to intervals needed to merge VSD

martin-cs:
Similar to https://github.com/diffblue/cbmc/pull/5202 this is a series of commits pulled out of the variable-sensitivity-domain branch.  As some of the patches have been squashed and cherry-picked multiple times I am not 100% sure of who wrote what but I think the full set of authors is correct.  Again, squashing and editing of the commit messages are probably needed but let's look at the code first.

### DIFF
--- a/src/analyses/interval_domain.cpp
+++ b/src/analyses/interval_domain.cpp
@@ -329,10 +329,10 @@ void interval_domaint::assume_rec(
     {
       integer_intervalt &lhs_i=int_map[lhs_identifier];
       integer_intervalt &rhs_i=int_map[rhs_identifier];
-      lhs_i.meet(rhs_i);
-      rhs_i=lhs_i;
-      if(rhs_i.is_bottom())
-        make_bottom();
+      if(id == ID_lt && !lhs_i.is_less_than(rhs_i))
+        lhs_i.make_less_than(rhs_i);
+      if(id == ID_le && !lhs_i.is_less_than_eq(rhs_i))
+        lhs_i.make_less_than_eq(rhs_i);
     }
     else if(is_float(lhs.type()) && is_float(rhs.type()))
     {
--- a/src/util/interval.cpp
+++ b/src/util/interval.cpp
@@ -1090,17 +1090,20 @@ bool constant_interval_exprt::is_float(const constant_interval_exprt &interval)
 bool constant_interval_exprt::is_bitvector(const typet &t)
 {
   return t.id() == ID_bv || t.id() == ID_signedbv || t.id() == ID_unsignedbv
-    || t.id() == ID_c_bool;
+    || t.id() == ID_c_bool
+    || (t.id() == ID_c_bit_field && is_bitvector(t.subtype()));
 }
 
 bool constant_interval_exprt::is_signed(const typet &t)
 {
-  return t.id() == ID_signedbv;
+  return t.id() == ID_signedbv
+    || (t.id() == ID_c_bit_field && is_signed(t.subtype()));
 }
 
 bool constant_interval_exprt::is_unsigned(const typet &t)
 {
-  return t.id() == ID_bv || t.id() == ID_unsignedbv || t.id() == ID_c_bool;
+  return t.id() == ID_bv || t.id() == ID_unsignedbv || t.id() == ID_c_bool
+    || t.id() == ID_c_enum || (t.id() == ID_c_bit_field && is_unsigned(t.subtype()));
 }
 
 bool constant_interval_exprt::is_signed(const constant_interval_exprt &interval)
--- a/src/util/interval.cpp
+++ b/src/util/interval.cpp
@@ -455,15 +455,10 @@ constant_interval_exprt constant_interval_exprt::get_extremes(
 
   std::vector<exprt> results;
 
-  results.push_back(
-      generate_expression(a.get_lower(), b.get_lower(), operation));
-  results.push_back(
-      generate_expression(a.get_lower(), b.get_upper(), operation));
-  results.push_back(
-      generate_expression(a.get_upper(), b.get_lower(), operation));
-  results.push_back(
-      generate_expression(a.get_upper(), b.get_upper(), operation));
-
+  generate_expression(a.get_lower(), b.get_lower(), operation, results);
+  generate_expression(a.get_lower(), b.get_upper(), operation, results);
+  generate_expression(a.get_upper(), b.get_lower(), operation, results);
+  generate_expression(a.get_upper(), b.get_upper(), operation, results);
 
   for(auto result: results)
   {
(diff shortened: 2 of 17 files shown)


unit/util/interval/add.cpp
```diff
@@ -0,0 +1,95 @@
+/*******************************************************************\
+ Module: Unit tests for variable/sensitivity/abstract_object::merge
+ Author: DiffBlue Limited
```
hannes-steffenhagen-diffblue, line 3: So it turns out it’s spelled _Diffblue_ actually, oops.

hannes-steffenhagen-diffblue:
Worth pointing out that this has... _limitations_ w.r.t. to floating point numbers.

unit/util/interval/add.cpp
martin-cs: Will fix.

martin-cs:
@hannes-steffenhagen-diffblue Thanks for the note about floating-point.  Yes, there are still things to do on this code but I would like to get VSD merged and then improve.

src/analyses/interval_domain.cpp
```diff
@@ -329,10 +329,10 @@ void interval_domaint::assume_rec(
     {
       integer_intervalt &lhs_i=int_map[lhs_identifier];
       integer_intervalt &rhs_i=int_map[rhs_identifier];
-      lhs_i.meet(rhs_i);
-      rhs_i=lhs_i;
-      if(rhs_i.is_bottom())
-        make_bottom();
+      if(id == ID_lt && !lhs_i.is_less_than(rhs_i))
```
thk123, line 332: :question: I'd expect to see an equivalent for `ID_gt` and `ID_ge`?

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+
+/// -∞ upper bound for intervals
+class min_exprt : public exprt
+{
+public:
+  explicit min_exprt(const typet &_type) : exprt(ID_min, _type)
+  {
+  }
+
+  explicit min_exprt(const exprt &_expr) : exprt(ID_min, _expr.type())
```
thk123, line 45: :question: What is the reason for this constructor?

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  constant_interval_exprt(
+    const exprt &lower,
+    const exprt &upper,
+    const typet type)
+    : binary_exprt(lower, ID_constant_interval, upper, type)
+  {
+    PRECONDITION(is_well_formed());
+  }
+
+  constant_interval_exprt(const constant_interval_exprt &x)
```
thk123, line 67: :question: Why the manual copy constructor? Seems like the default would work

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  }
+
+  constant_interval_exprt(const exprt &lower, const exprt &upper)
+    : constant_interval_exprt(lower, upper, lower.type())
+  {
+  }
+
+  bool is_well_formed() const
+  {
+    bool b = true;
```
thk123, line 89: :pick: `b`-> `well_formed`;
:bulb: I'd probably just ditch the overall local variable and introduce a few locals:

```
const bool valid_type = is_numeric() || type.id() == ID_bool  ...
...
retur valid_type && matching_type && valid_bounds
```

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+
+    b &= is_valid_bound(lower);
+    b &= is_valid_bound(upper);
+
+    b &= !is_numeric() || is_bottom() || less_than_or_equal(lower, upper);
+
+    return b;
+  }
+
+  bool is_valid_bound(const exprt &expr) const
```
thk123, line 108: :pick: This can be static if `expr.type()` is used

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  /** SET OF ARITHMETIC OPERATORS */
+  constant_interval_exprt
+  handle_constant_unary_expression(const irep_idt &op) const;
+  constant_interval_exprt handle_constant_binary_expression(
+    const constant_interval_exprt &other,
+    const irep_idt &) const;
+
+  constant_interval_exprt eval(const irep_idt &unary_operator) const;
+  constant_interval_exprt
+  eval(const irep_idt &binary_operator, const constant_interval_exprt &o) const;
```
thk123, line 146: These don't look like they belong to this class

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+
+  /* Unary arithmetic */
+  constant_interval_exprt unary_plus() const;
```
thk123, line 149: Slightly more natural name
```suggestion
  constant_interval_exprt increment() const;
```

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  constant_interval_exprt unary_minus() const;
+
+  constant_interval_exprt typecast(const typet &type) const;
```
thk123, line 152: :question: This doesn't look like it is required

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+
+  /* Logical */
+  tvt is_definitely_true() const;
+  tvt is_definitely_false() const;
+
+  tvt logical_and(const constant_interval_exprt &o) const;
+  tvt logical_or(const constant_interval_exprt &o) const;
+  tvt logical_xor(const constant_interval_exprt &o) const;
+  tvt logical_not() const;
```
thk123, line 161: Looks like should be pulled out into a separate function that deals with intervals of truth types rather than cluttering the interface of the interval

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  constant_interval_exprt bitwise_and(const constant_interval_exprt &o) const;
+
+  tvt less_than(const constant_interval_exprt &o) const;
+  tvt greater_than(const constant_interval_exprt &o) const;
+  tvt less_than_or_equal(const constant_interval_exprt &o) const;
+  tvt greater_than_or_equal(const constant_interval_exprt &o) const;
+  tvt equal(const constant_interval_exprt &o) const;
+  tvt not_equal(const constant_interval_exprt &o) const;
+
+  constant_interval_exprt increment() const;
```
thk123, line 189: :question: What's the difference between this and unary_plus

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  constant_interval_exprt decrement() const;
+
+  bool is_empty() const;
+  bool is_single_value_interval() const;
+  /** END SET OF ARITHMETIC OPERATORS */
+
+  //  tvt contains(constant_interval_exprt &o) const;
```
thk123, line 196: Commented out code should probably just be deleted

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+
+  /* SET OF EXPR COMPARATORS */
+  static bool equal(const exprt &a, const exprt &b);
+  static bool not_equal(const exprt &a, const exprt &b);
+  static bool less_than(const exprt &a, const exprt &b);
+  static bool less_than_or_equal(const exprt &a, const exprt &b);
+  static bool greater_than(const exprt &a, const exprt &b);
+  static bool greater_than_or_equal(const exprt &a, const exprt &b);
```
thk123, line 204: :question: Not sure what these do yet - but they don't seem like they should be part of the interface

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  /* END SET OF EXPR COMPS */
+
+  /* INTERVAL COMPARISONS, returns tvt.is_true().  False cannot be trusted
```
thk123, line 207: I think I'd just remove these - seems liable to be misused, for the sake of a few characters

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+    const constant_interval_exprt &lhs,
+    const constant_interval_exprt &rhs);
+  friend const constant_interval_exprt operator!(
+    const constant_interval_exprt &lhs);
+  friend const constant_interval_exprt operator&(
+    const constant_interval_exprt &lhs,
+    const constant_interval_exprt &rhs);
+  friend const constant_interval_exprt operator|(
+    const constant_interval_exprt &lhs,
+    const constant_interval_exprt &rhs);
```
thk123, line 252: Since these are implemented precisely I guess they can stay - but I'd pick _either_ these or the explicitly named functions, rather than providing two interfaces.

Also, since this header is huge, and these can be implemented in terms of public interface, I'd probably move into a separate header

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  friend const constant_interval_exprt operator>>(
+    const constant_interval_exprt &lhs,
+    const constant_interval_exprt &rhs);
+
+  friend std::ostream &
+  operator<<(std::ostream &out, const constant_interval_exprt &i);
+  std::string to_string() const;
+
+  /* Now static equivalents! */
```
thk123, line 267: :question: Why have equivalents? Just clutters the interface

src/util/interval.h
thk123: I'm not sure if it is a good idea, but the operator< etc can be implemented to _not_ return a boolean. I think this would be better than returning is true or unknown, but might be better to just remove

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+    const constant_interval_exprt &b);
+  static tvt
+  equal(const constant_interval_exprt &a, const constant_interval_exprt &b);
+  static tvt
+  not_equal(const constant_interval_exprt &a, const constant_interval_exprt &b);
+
+  static constant_interval_exprt increment(const constant_interval_exprt &a);
+  static constant_interval_exprt decrement(const constant_interval_exprt &a);
+
+  static bool is_empty(const constant_interval_exprt &a);
```
thk123, line 338: :no_entry_sign: We should surely pick whether to have a const method or a static method. I think it is probably more natural to have them as member methods, but easier to protect invariant as static methods. But surely both is the worst of both worlds!

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  bool is_bottom() const;
+  static constant_interval_exprt top(const typet &type);
+  static constant_interval_exprt bottom(const typet &type);
+  constant_interval_exprt top() const;
+  constant_interval_exprt bottom() const;
+
+  bool has_no_lower_bound() const;
+  bool has_no_upper_bound() const;
+  static bool is_min(const exprt &expr);
+  static bool is_max(const exprt &expr);
```
thk123, line 358: These helper methods belong with the min_expr and max_expr, if they even need to exist

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+
+  /* Generate min and max exprt according to current type */
+  min_exprt min() const;
+  max_exprt max() const;
```
thk123, line 362: No need for these methods IMO

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+
+  constant_exprt zero() const;
```
thk123, line 364: No need for this method

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  static constant_exprt zero(const typet &type);
+  static constant_exprt zero(const exprt &expr);
+  static constant_exprt zero(const constant_interval_exprt &interval);
```
thk123, line 367: Or these

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  /* Private? */
+  static constant_interval_exprt get_extremes(
+    const constant_interval_exprt &lhs,
+    const constant_interval_exprt &rhs,
+    const irep_idt &operation);
+  static exprt get_extreme(std::vector<exprt> values, bool min = true);
+  static exprt get_max(const exprt &a, const exprt &b);
+  static exprt get_min(const exprt &a, const exprt &b);
+  static exprt get_min(std::vector<exprt> &values);
+  static exprt get_max(std::vector<exprt> &values);
```
thk123, line 378: Not sure about these helper functions

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+
+  /* we don't simplify in the constructor otherwise */
+  static constant_interval_exprt simplified_interval(exprt &l, exprt &r);
+  static exprt simplified_expr(exprt expr);
+
+  /* Helpers */
+  /* Four common params: self, static: type, expr, interval */
```
thk123, line 385: Same comment - I believe all these helpers make the interface much harder to maintain, for tiny convenience.

unit/util/interval/bitwise.cpp
```diff
@@ -0,0 +1,64 @@
...
+        five.bitwise_xor(nine) ==
+        constant_interval_exprt(from_integer(12, unsigned_int)));
+    }
+
+    THEN("Left shift on the 5 should produce 10")
+    {
+      REQUIRE(
+        five.left_shift(
+          constant_interval_exprt(from_integer(1, unsigned_int))) ==
+        constant_interval_exprt(from_integer(10, unsigned_int)));
```
thk123, line 53: I'd add a test for shifting more than one

unit/util/interval/comparisons.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+#include <util/arith_tools.h>
+#include <util/interval.h>
+#include <util/std_expr.h>
+#include <util/std_types.h>
+#include <util/symbol_table.h>
+
+#define V(X) (bvrep2integer(X.get(ID_value).c_str(), 32, true))
+#define V_(X) (bvrep2integer(X.c_str(), 32, true))
+#define CEV(X) (from_integer(mp_integer(X), signedbv_typet(32)))
```
thk123, line 16: :no_entry_sign: I believe there are existing functions thats to do this, and will make it clearer, but this is incomprehensibke

thk123: at the very least:
V -> int_value
CEV -> int_expr
and delete the unused V_

unit/util/interval/comparisons.cpp
```diff
@@ -0,0 +1,208 @@
...
+SCENARIO("comparison interval domain", "[core][analyses][interval][comparison]")
+{
+  GIVEN("Two simple signed intervals")
+  {
+    symbol_tablet symbol_table;
+    namespacet ns(symbol_table);
+
+    WHEN("<, >, <=, >=, ==, != are tested on expressions")
+    {
+      THEN("Require correctness")
```
thk123, line 27:
```suggestion
      THEN("Less than implemented correctly")
```

unit/util/interval/comparisons.cpp
```diff
@@ -0,0 +1,208 @@
...
+          min_exprt(signedbv_typet(32)), min_exprt(signedbv_typet(32))));
+        REQUIRE(constant_interval_exprt::not_equal(
+          max_exprt(signedbv_typet(32)), min_exprt(signedbv_typet(32))));
+        REQUIRE(constant_interval_exprt::equal(
+          max_exprt(signedbv_typet(32)), max_exprt(signedbv_typet(32))));
+        REQUIRE(constant_interval_exprt::not_equal(
+          min_exprt(signedbv_typet(32)), max_exprt(signedbv_typet(32))));
+      }
+
+      THEN("")
```
thk123, line 69:
```suggestion
      THEN("Greater than implemented correctly")
```

unit/util/interval/comparisons.cpp
```diff
@@ -0,0 +1,208 @@
...
+            .greater_than_or_equal(constant_interval_exprt(CEV(30), CEV(40))) ==
+          tvt::unknown());
+        CHECK(
+          constant_interval_exprt(CEV(10), CEV(31))
+            .greater_than_or_equal(constant_interval_exprt(CEV(30), CEV(40))) ==
+          tvt::unknown());
+      }
+    }
+  }
+}
```
thk123, line 208: :no_entry_sign: appears to check in empty file `divide.cpp`

unit/util/interval/get_extreme.cpp
```diff
@@ -0,0 +1,161 @@
...
+#define CEV(X) (from_integer(mp_integer(X), signedbv_typet(32)))
+
+SCENARIO("get extreme exprt value", "[core][analyses][interval][get_extreme]")
+{
+  GIVEN("A selection of constant_exprts in a std::vector and map")
+  {
+    symbol_tablet symbol_table;
+    namespacet ns(symbol_table);
+
+    std::vector<exprt> ve;
```
thk123, line 26: :pick: This is unused

unit/util/interval/multiply.cpp
```diff
@@ -0,0 +1,144 @@
...
+    WHEN("One is entirely negative [-2,-5]*[7,11]")
+    {
+      constant_interval_exprt a(CEV(-5), CEV(-2));
+      constant_interval_exprt b(CEV(7), CEV(11));
+
+      constant_interval_exprt result = a.multiply(b);
+
+      THEN("Domain is consistent")
+      {
+        CHECK(V(a.get_lower()) == -5);
```
thk123, line 58: :no_entry_sign: I'm pretty sure CHECK here means the test won't fail if this is wrong, should be REQUIRE

unit/util/interval/shift.cpp
```diff
@@ -0,0 +1,36 @@
...
+    WHEN("Something")
+    {
+      THEN("Something else")
+      {
+        REQUIRE(
+          constant_interval_exprt(CEV(4), CEV(8))
+            .left_shift(constant_interval_exprt(CEV(1))) ==
+          constant_interval_exprt(CEV(8), CEV(16)));
+      }
+    }
```
thk123, line 34: Looks incomplete and maybe tested elsewhere?

src/util/interval_template.h
```diff
@@ -143,6 +143,47 @@ template<class T> class interval_templatet
     }
   }
 
+  void make_bottom()
```
thk123, line 146: :question: I need to see why these changes are introduced - they seem unrelated to the rest of the PR>

src/util/interval.h
```diff
@@ -0,0 +1,472 @@
...
+  {
+  }
+};
+
+/// Represents an interval of values.
+/// Bounds should be constant expressions
+/// or min_exprt for the lower bound
+/// or max_exprt for the upper bound
+/// Also, lower bound should always be <= upper bound
+class constant_interval_exprt : public binary_exprt
```
thk123, line 55: :pick: from an "is-a" relation, I don't think an interval really is a binary expr? Not sure...

src/util/interval.cpp
```diff
@@ -0,0 +1,1975 @@
...
+{
+  if(is_single_value_interval() && o.is_single_value_interval())
+  {
+    return tvt(equal(get_lower(), o.get_lower()));
+  }
+
+  if(equal(get_upper(), o.get_upper()) && equal(get_upper(), o.get_upper()))
+  {
+    return tvt(true);
+  }
```
thk123, line 439: This seems wrong. For one, it appears to be `if(A && A)` and if the intention is whether upper == o.upper && lower ==o.lower, this is wrong. I think this should always be unknown unless the intersection is empty, then false, or one element, then true

thk123 (commented):
There are a number of ways the intervals code could be improved to help maintainability, but in the interest of getting something merged I suggest focusing on two priorities and address the other points in due course:

 - ensure that testing covers 100% of the file - there is so much code here it seems like there will be a few typos that a complete set of unit tests will surely catch. 
 - minimize and tidy the public API as this will be hard to do after the fact (as code gets added, removing it later will be harder). 

I will rebase this now so that we can see the codecov report, and add a commit eliminating parts of the interface that probably aren't required.

src/util/interval.h
thk123: nope - this is the + operator, as in `+x` not `x++` which is a no-op

thk123:
However, given that the VSD is based on this interface - I think it actually makes more sense to merge this, and then tidy up the interface along with changes required to the VSD in subsequent PR. Instead i will dimiss comments to this effect - and will focus on the testing component

src/util/interval.cpp
thk123: TBH the semantics of this are confusing, do you expect [a, b] == [a, b] to be true or unknown The docs appear to suggest unknown, this behaviour appears to suggest true. I think depending on context you will want one or the other, so probably much safer to not guess and implement two clearly different methods: `range_equal` which returns true or false if the ranges are exactly the same, and value_equal which returns a tvt - but this can be left for now and fixed in the follow up PR

unit/util/interval/multiply.cpp
thk123: This is wrong - just allows all checks to be tested, and is probably _more_ correct as the results are indepedent. Certainly doesn't need to be fixed.

thk123:
#5361, a rebased version of this, has been merged. #5362 still needs to be merged (a small fix from this PR), but I'm going to close this PR in any case.

thk123:
Just noting here #5362 has now also been merged.

PR #5209, opened 2020-01-03
Correct nondet init of function pointer arguments

### LINKED ISSUE #4937: Non-deterministic initialisation of function pointers is broken
```main.c
#include <assert.h>

typedef int (*other_function_type)(int n);

void foo(other_function_type other_function)
{
  assert(other_function(4) > 5);
}
```

CBMC version: 5.12 (cbmc-5.12-d8598f8-559-gfaf21d043-dirty)
Operating system: Ubuntu 18.04
Exact command line resulting in the issue: `$ cbmc main.c --function foo`
What behaviour did you expect: VERIFICATION FAILED
What happened instead: `assignment to 'symbol' not handled`

xbauch:
At least until handling of function pointers is transferred to symex.

1. The nondet symbol factory only assigns nondet to function pointers, i.e. does
not build the `depth` deep pointer tree.
2. The function pointer removal checks if the pointer is an argument to the
entry point and if so, skips adding the `assume(false)` for cutting short the
execution in case on function target was found.

add 2: On the GOTO level, the function call is replaced with `return_value`
which is never assigned to.

Notes:
- this is an alternative/extension to #4940 which only changed the factory.
- both PRs try to fix #4937.

### DIFF
--- a/regression/cbmc/pointer-function-origin-argument/main.c
+++ b/regression/cbmc/pointer-function-origin-argument/main.c
@@ -1,0 +1,8 @@
+#include <assert.h>
+
+typedef int (*other_function_type)(int n);
+
+void foo(other_function_type other_function)
+{
+  assert(other_function(4) > 5);
+}
--- a/regression/cbmc/pointer-function-origin-argument/test.desc
+++ b/regression/cbmc/pointer-function-origin-argument/test.desc
@@ -1,0 +1,9 @@
+CORE
+main.c
+--function foo
+^EXIT=10$
+^SIGNAL=0$
+^\[foo.assertion.\d+\] line \d+ assertion other_function\(4\) > 5: FAILURE$
+VERIFICATION FAILED
+--
+^warning: ignoring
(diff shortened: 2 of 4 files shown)


hannes-steffenhagen-diffblue:
Can you explain here why this is better than #4940, which seems to be the more intuitive way to do this on the surface level at least? I guess this doesn’t handle functions where we don’t have functions with matching signatures, so you’re adding a new one here that can be called instead? That doesn’t seem like an unreasonable thing to do here either, but I suspect that the behaviour may be surprising for some people (it’s a bit hard to say what the correct behaviour for an arbitrary function pointer should be, and this makes this happen silently).

Does the function pointer removal currently add assertions to check that it can resolve the function pointer? If not, this could be one thing to “fix” to alert users to such a problem and maybe tell them to just write a function implementing the behaviour they want to test for themselves.

xbauch:
Superseded by #4940 and to be closed, eventually.

xbauch:
#4940 was merged, and solved the issue.

Issue #4604, opened 2019-05-01
CBMC malloc should fail if called for a size CBMC cannot represent


danielsn:
I believe that CBMC can only represent allocations of size 2**(64-number-object-bits). And yet 

```
    int *p = malloc(SIZE_MAX);
    assert(p);
```
succeeds. This should either return NULL, or block with an assume that the input to malloc is a representable size.

xbauch:
@danielsn Now that #5210 was merged, this may be resolved. Please have a look at `doc/cprover-manual/properties.md` for description of the user-options that enable `malloc` to fail here.

hannes-steffenhagen-diffblue:
THis has been implemented with the `--malloc-may-fail` and `--malloc-fail-null` flags.

Issue #5096, opened 2019-09-07
__CPROVER_w_ok is wrong with large object offsets


markrtuttle:
CBMC version: CBMC version 5.12 (cbmc-5.12-d8598f8-979-gfc0a0cd57) 64-bit x86_64 macos
Operating system: MacOS
Exact command line resulting in the issue: cbmc  --object-bits 7 --32 --trace demo.c
What behaviour did you expect: verification successful
What happened instead: verification failure

The final ```__CPROVER_w_ok``` assertion in the attached [demo.c](https://github.com/diffblue/cbmc/files/3586067/demo.c.zip) fails when run with ```cbmc --object-bits 7 --32 demo.c```.

```
#include <stdlib.h>

void main() {
  char* buffer;
  char* buffer_next;
  char* buffer_end;

  // Running cbmc with --object-bits 7 --32
  // Top 7 bits of pointer are object id
  // Bottom 32-7 bits of pointer are object offset
  // Ensure all valid offsets fit within 32-7 bits.

  uint32_t buffer_length;
  __CPROVER_assume(buffer_length < 0x02000000);
  buffer = malloc(buffer_length);

  size_t offset1;
  size_t offset2;
  __CPROVER_assume(offset1 < offset2);
  __CPROVER_assume(offset1 < buffer_length);
  __CPROVER_assume(offset2 < buffer_length);
  buffer_next = buffer + offset1;
  buffer_end = buffer + offset2;

  uint32_t length;
  __CPROVER_assume(length < (buffer_end - buffer_next));

  __CPROVER_size_t show_buffer_object = __CPROVER_POINTER_OBJECT(buffer_next);
  __CPROVER_ssize_t show_buffer_offset = __CPROVER_POINTER_OFFSET(buffer_next);
  __CPROVER_size_t show_buffer_size = __CPROVER_OBJECT_SIZE(buffer_next);
  __CPROVER_assert(__CPROVER_w_ok(buffer_next, length), "buffer length");
}
```

The arguments to cbmc ask for the top 7 bits of a pointer to be used for the object id, and the bottom 25 bits to be used for the object offset.

The trace (summarized below) shows that the bytes ```buffer_next``` through ```buffer_next + length``` fit within the buffer, so the assertion should be true: offset1 + length = 20125093u + 668989u = 20794082 < 25545442u = buffer_length

```
cbmc  --object-bits 7 --32 --trace demo.c
[...]

PR #5210, opened 2020-01-06
Malloc should fail on too large allocations

xbauch:
via goto-checker with a dedicated cmdl option and a property that acknowledges pointer-width and object-size.

### DIFF
--- a/regression/cbmc/array_constraints1/test.desc
+++ b/regression/cbmc/array_constraints1/test.desc
@@ -4,6 +4,6 @@ main.c
 ^EXIT=10$
 ^SIGNAL=0$
 ^VERIFICATION FAILED$
-^\*\* 2 of 14
+^\*\* 2 of 15
 --
 ^warning: ignoring
--- a/regression/cbmc/pointer-overflow1/test.desc
+++ b/regression/cbmc/pointer-overflow1/test.desc
@@ -5,7 +5,7 @@ main.c
 ^SIGNAL=0$
 ^\[main\.overflow\.\d+\] line \d+ (pointer )?arithmetic overflow on .*sizeof\(signed int\) .* : SUCCESS
 ^VERIFICATION FAILED$
-^\*\* 8 of 11 failed
+^\*\* 8 of 12 failed
 --
 ^\[main\.overflow\.\d+\] line \d+ (pointer )?arithmetic overflow on .*sizeof\(signed int\) .* : FAILURE
 ^warning: ignoring
--- a/regression/cbmc/r_w_ok1/test.desc
+++ b/regression/cbmc/r_w_ok1/test.desc
@@ -2,7 +2,7 @@ CORE
 main.c
 
 __CPROVER_[rw]_ok\(arbitrary_size, n \+ 1\): FAILURE$
-^\*\* 2 of 10 failed
+^\*\* 2 of 11 failed
 ^VERIFICATION FAILED$
 ^EXIT=10$
 ^SIGNAL=0$
--- a/src/ansi-c/ansi_c_internal_additions.cpp
+++ b/src/ansi-c/ansi_c_internal_additions.cpp
@@ -155,6 +155,15 @@ void ansi_c_internal_additions(std::string &code)
     "void *" CPROVER_PREFIX "allocate("
       CPROVER_PREFIX "size_t size, " CPROVER_PREFIX "bool zero);\n"
     "const void *" CPROVER_PREFIX "alloca_object = 0;\n"
+    "int " CPROVER_PREFIX "malloc_failure_mode="+
+      std::to_string(config.ansi_c.malloc_failure_mode)+";\n"
+    "int " CPROVER_PREFIX "malloc_failure_mode_return_null="+
+    std::to_string(config.ansi_c.malloc_failure_mode_return_null)+";\n"
+    "int " CPROVER_PREFIX "malloc_failure_mode_assert_then_assume="+
+    std::to_string(config.ansi_c.malloc_failure_mode_assert_then_assume)+";\n"
+    CPROVER_PREFIX "size_t " CPROVER_PREFIX "max_malloc_size="+
+    std::to_string(1 << (config.ansi_c.pointer_width -
+                         config.bv_encoding.object_bits - 1))+";\n"
 
     // this is ANSI-C
     "extern " CPROVER_PREFIX "thread_local const char __func__["
--- a/src/ansi-c/library/stdlib.c
+++ b/src/ansi-c/library/stdlib.c
@@ -115,23 +115,47 @@ inline void *malloc(__CPROVER_size_t malloc_size)
   // realistically, malloc may return NULL,
   // and __CPROVER_allocate doesn't, but no one cares
   __CPROVER_HIDE:;
-  void *malloc_res;
-  malloc_res = __CPROVER_allocate(malloc_size, 0);
 
-  // make sure it's not recorded as deallocated
-  __CPROVER_deallocated=(malloc_res==__CPROVER_deallocated)?0:__CPROVER_deallocated;
+    if(
+      __CPROVER_malloc_failure_mode ==
+      __CPROVER_malloc_failure_mode_return_null)
+    {
+      if(malloc_size > __CPROVER_max_malloc_size)
+      {
+        return (void *)0;
+      }
+    }
+    else if(
+      __CPROVER_malloc_failure_mode ==
+      __CPROVER_malloc_failure_mode_assert_then_assume)
+    {
+      __CPROVER_assert(
+        malloc_size <= __CPROVER_max_malloc_size,
+        "max allocation size exceeded");
+      __CPROVER_assume(malloc_size <= __CPROVER_max_malloc_size);
+    }
 
-  // record the object size for non-determistic bounds checking
-  __CPROVER_bool record_malloc=__VERIFIER_nondet___CPROVER_bool();
-  __CPROVER_malloc_object=record_malloc?malloc_res:__CPROVER_malloc_object;
-  __CPROVER_malloc_size=record_malloc?malloc_size:__CPROVER_malloc_size;
-  __CPROVER_malloc_is_new_array=record_malloc?0:__CPROVER_malloc_is_new_array;
+    void *malloc_res;
+    malloc_res = __CPROVER_allocate(malloc_size, 0);
 
-  // detect memory leaks
-  __CPROVER_bool record_may_leak=__VERIFIER_nondet___CPROVER_bool();
-  __CPROVER_memory_leak=record_may_leak?malloc_res:__CPROVER_memory_leak;
+    // make sure it's not recorded as deallocated
+    __CPROVER_deallocated =
+      (malloc_res == __CPROVER_deallocated) ? 0 : __CPROVER_deallocated;
 
-  return malloc_res;
+    // record the object size for non-determistic bounds checking
+    __CPROVER_bool record_malloc = __VERIFIER_nondet___CPROVER_bool();
+    __CPROVER_malloc_object =
+      record_malloc ? malloc_res : __CPROVER_malloc_object;
+    __CPROVER_malloc_size = record_malloc ? malloc_size : __CPROVER_malloc_size;
+    __CPROVER_malloc_is_new_array =
+      record_malloc ? 0 : __CPROVER_malloc_is_new_array;
+
+    // detect memory leaks
+    __CPROVER_bool record_may_leak = __VERIFIER_nondet___CPROVER_bool();
+    __CPROVER_memory_leak =
+      record_may_leak ? malloc_res : __CPROVER_memory_leak;
+
+    return malloc_res;
 }
 
 /* FUNCTION: __builtin_alloca */
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -1067,6 +1067,9 @@ void cbmc_parse_optionst::help()
     " --error-label label          check that label is unreachable\n"
     " --cover CC                   create test-suite with coverage criterion CC\n" // NOLINT(*)
     " --mm MM                      memory consistency model for concurrent programs\n" // NOLINT(*)
+    // NOLINTNEXTLINE(whitespace/line_length)
+    " --malloc-fail-assert         set malloc failure mode to assert-then-assume\n"
+    " --malloc-fail-null           set malloc failure mode to return null\n"
     HELP_REACHABILITY_SLICER
     HELP_REACHABILITY_SLICER_FB
     " --full-slice                 run full slicer (experimental)\n" // NOLINT(*)
--- a/src/util/config.cpp
+++ b/src/util/config.cpp
@@ -1093,6 +1093,16 @@ bool configt::set(const cmdlinet &cmdline)
     bv_encoding.is_object_bits_default = false;
   }
 
+  if(cmdline.isset("malloc-fail-assert") && cmdline.isset("malloc-fail-null"))
+  {
+    throw invalid_command_line_argument_exceptiont{
+      "at most one malloc failure mode is acceptable", "--malloc-fail-null"};
+  }
+  if(cmdline.isset("malloc-fail-null"))
+    ansi_c.malloc_failure_mode = ansi_c.malloc_failure_mode_return_null;
+  if(cmdline.isset("malloc-fail-assert"))
+    ansi_c.malloc_failure_mode = ansi_c.malloc_failure_mode_assert_then_assume;
+
   return false;
 }
 
--- a/doc/cprover-manual/properties.md
+++ b/doc/cprover-manual/properties.md
@@ -319,3 +319,19 @@ example, replacing functions or setting global variables with the `__CPROVER`
 prefix might make analysis impossible. To avoid doing this by accident, negative
 lookahead can be used. For example, `(?!__).*` matches all names not starting
 with `__`.
+
+### Malloc failure mode
+
+|Flag                    |  Check                                          |
+|------------------------|-------------------------------------------------|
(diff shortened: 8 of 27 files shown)


xbauch:
This may fix #4604.

xbauch:
It seems that we have a few regression tests that call `malloc` with nondeterministic value (of type `size_t`) without assuming reasonable upper bound.

src/ansi-c/library/stdlib.c
```diff
@@ -115,6 +115,9 @@ inline void *malloc(__CPROVER_size_t malloc_size)
   // realistically, malloc may return NULL,
   // and __CPROVER_allocate doesn't, but no one cares
   __CPROVER_HIDE:;
+  // 8 is the default object bits for ANSI-C and C++
+  if (malloc_size >= ((__CPROVER_size_t)1<<(64-8)))
```
chrisr-diffblue: Can't the object bits size be changed on the command line with `--object-bits` ? and if so, this probably needs to accommodate that. Ditto for the hardcoded `64`, which would probably be wrong if `--32` is specified to cbmc :-)

regression/cbmc/malloc-too-large/main.c
```diff
@@ -0,0 +1,11 @@
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+int main()
+{
+  int *p = malloc(SIZE_MAX);
```
chrisr-diffblue: It would also be nice to have a test for a size that is 1 byte larger than the maximum CBMC can support, and also for a size that is exactly the same as the largest CBMC can support - just to test the boundary condition.

src/ansi-c/library/stdlib.c
xbauch: I moved the check into `goto-checker`, i.e. the `stdlib.c` implemetation is unchanged, we can use the pointer-width and object-bits from config, and the whole thing is behind an option, so it won't affect the existing tests.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 640cdccbe1b06545e7cbb7f5b615c1f39505ee81).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/143393509

regression/cbmc/malloc-too-large/main.c
xbauch: Added.

regression/cbmc/malloc-too-large-almost/main.c
```diff
@@ -0,0 +1,13 @@
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+int main()
+{
+  size_t pointer_width = sizeof(void *) * 8;
+  size_t object_bits = 8; // by default
+  size_t cbmc_max_alloc_size = (size_t)1 << (pointer_width - object_bits);
```
chrisr-diffblue: A really tiny nit pick, so feel free to ignore me :-) But I'd personally write this as:
```
size_t cbmc_max_alloc_size = ((size_t)1 << (pointer_width - object_bits))-1;
int *p2 = malloc(cbmc_max_alloc_size); // try to allocate exactly the max
```
just so that the variable names really do mean what they say. Ditto for the other test case - but like I say, it's a really minor nit for me to pick...

src/analyses/goto_check.h
```diff
@@ -55,6 +56,7 @@ void goto_check(
   " --nan-check                  check floating-point for NaN\n" \
   " --no-built-in-assertions     ignore assertions in built-in library\n" \
   " --enum-range-check           checks that all enum type expressions have values in the enum range\n" /* NOLINT(whitespace/line_length) */ \
+  " --malloc-max-check           check that CBMC can represent the allocated object of requested size" /* NOLINT(whitespace/line_length) */ \
```
chrisr-diffblue: Is a better name for this option "malloc-size-check" ? Don't have a strong opinion.

chrisr-diffblue (approved):
Generally looks good to me, and I like that it's moved into goto-check. Apologies for the niggle comments - take them as suggestions, and not absolute requirements.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: b73f9689ebadceadb4fe1c4d59a963b5a2b4ef9d).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/143440844

src/analyses/goto_check.cpp
```diff
@@ -1941,6 +1943,30 @@ void goto_checkt::goto_check(
...
+            from_integer(config.ansi_c.pointer_width, alloc_size_type);
+          const auto object_bits =
+            from_integer(config.bv_encoding.object_bits, alloc_size_type);
+          const auto max_alloc_size =
+            binary_exprt{from_integer(2, alloc_size_type),
+                         ID_power,
+                         minus_exprt{pointer_bits, object_bits}};
+          add_guarded_property(
+            binary_relation_exprt{rhs_side_effect.op0(), ID_lt, max_alloc_size},
+            "malloc_size < 2 ^ (pointer_width - object_size)",
```
smowton: `object_size` -> `object_id_width`?

smowton (approved):

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: 0c2c362bb817a7c47212b49c726b39ffdac02767).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/143549138

martin-cs:
On Tue, 2020-01-07 at 06:40 -0800, Chris Ryder wrote:
> Generally looks good to me, and I like that it's moved into goto-
> check.

I'm not sure I do.  Keeping the modelling of malloc in one place, in
the library models seems, to me, to be important.  Setting CPROVER
globals to control max limits, if malloc can fail, etc. seems a more
flexible option and keeps all the modelling together.

danielsn:
> It seems that we have a few regression tests that call `malloc` with nondeterministic value (of type `size_t`) without assuming reasonable upper bound.

I see two ways to handle `malloc(size)` where `size` is too large:
1. `assert(size <= MAX_MALLOC_SIZE)`
2. `if(size > MAX_MALLOC_SIZE) return NULL`

hannes-steffenhagen-diffblue (approved):

danpoe (approved):

regression/goto-instrument/insert-final-assert-false1/main.c
```diff
@@ -3,6 +3,12 @@
 int main()
 {
   size_t n;
+  // alloc size is arbitrary but no larger than what CBMC can represent
+  size_t pointer_width = sizeof(void *) * 8;
+  size_t object_bits = 8; // by default
+  size_t cbmc_max_alloc_size = ((size_t)1 << (pointer_width - object_bits)) - 1;
+  __CPROVER_assume(n <= cbmc_max_alloc_size);
```
hannes-steffenhagen-diffblue: May be worth adding this as an exposed constant `__CPROVER_max_alloc_size`? Seems to me this might come up fairly regularly in user scenarios as well, and it’d probably be a bit of a PITA to keep redefining this.

hannes-steffenhagen-diffblue (approved):
LGTM, @martin-cs are you happy with this iteration (changes moved to `stdlib.c`)?

chrisr-diffblue:
@xbauch Did you see @danielsn comment? I wonder if its worth adding a bit more user control so that the end user could decided which model (assert, versus return null) they wish to use? I guess that could even potentially be exposed via another global, e.g. `__CPROVER_malloc_failure_mode` or something like that? Much like the rounding mode variable... I guess that having a global like that could even mean that a user could write a harness function that could even change the malloc failure mode mid-harness...

xbauch:
@chrisr-diffblue I did see it and we experimented with `malloc` returning `NULL`, but the rest of CBMC does not like that.

src/ansi-c/library/stdlib.c
```diff
@@ -115,6 +115,15 @@ inline void *malloc(__CPROVER_size_t malloc_size)
...
   // and __CPROVER_allocate doesn't, but no one cares
   __CPROVER_HIDE:;
+
+  __CPROVER_assert(
+    malloc_size < ((__CPROVER_size_t)1
+                   << (__CPROVER_pointer_width - __CPROVER_object_bits)),
+    "max allocation size exceeded");
+  __CPROVER_assume(
+    malloc_size < ((__CPROVER_size_t)1
+                   << (__CPROVER_pointer_width - __CPROVER_object_bits)));
```
danpoe: Given that object offsets are signed (see #5096), I think this should be `malloc_size <= ((__CPROVER_size_t)1 << (__CPROVER_pointer_width - __CPROVER_object_bits - 1))`. Also we can use `<=` as the maximum allocation size is one larger than the largest offset we can represent.

xbauch: Nice catch, thanks.

xbauch:
@chrisr-diffblue What do you think now? The first commit (19fcd57f385e8740b9ad84be9e2f8236c15cdf72) now contains two options to choose how malloc should fail.

regression/cbmc/array_constraints1/test.desc
```diff
@@ -4,6 +4,6 @@ main.c
 ^EXIT=10$
 ^SIGNAL=0$
 ^VERIFICATION FAILED$
-^\*\* 2 of 14
+^\*\* 2 of \d+
```
martin-cs: Why `\d+` ? shouldn't the number of VCs be known and static?

regression/cbmc/pointer-overflow1/test.desc
```diff
@@ -5,7 +5,7 @@ main.c
 ^SIGNAL=0$
 ^\[main\.overflow\.\d+\] line \d+ (pointer )?arithmetic overflow on .*sizeof\(signed int\) .* : SUCCESS
 ^VERIFICATION FAILED$
-^\*\* 8 of 11 failed
+^\*\* 8 of \d+ failed
```
martin-cs: Same again.

regression/cbmc/r_w_ok1/test.desc
```diff
@@ -2,7 +2,7 @@ CORE
 main.c
 
 __CPROVER_[rw]_ok\(arbitrary_size, n \+ 1\): FAILURE$
-^\*\* 2 of 10 failed
+^\*\* 2 of \d+ failed
```
martin-cs: Another one.

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1041,6 +1041,9 @@ void cbmc_parse_optionst::help()
     "\n"
     "Backend options:\n"
     " --object-bits n              number of bits used for object addresses\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
+    " --malloc-fail-assert         set malloc failure mode to assert-then-assume\n"
```
martin-cs: I think these should really be in `Program instrumentation options` because they are controlling what properties are checked.

src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -155,6 +155,11 @@ void ansi_c_internal_additions(std::string &code)
     "void *" CPROVER_PREFIX "allocate("
       CPROVER_PREFIX "size_t size, " CPROVER_PREFIX "bool zero);\n"
     "const void *" CPROVER_PREFIX "alloca_object = 0;\n"
+    "int " CPROVER_PREFIX "malloc_failure_mode="+
+      std::to_string(config.ansi_c.malloc_failure_mode)+";\n"
+    CPROVER_PREFIX "size_t " CPROVER_PREFIX "max_malloc_size="+
+    std::to_string(1 << (config.ansi_c.pointer_width -
+                         config.bv_encoding.object_bits - 1))+";\n"
```
martin-cs, line 166: Yes! Nice!

martin-cs (approved):
Thank you for reworking this.  I think this is a much simpler, compact and flexible way of getting this functionality.

src/ansi-c/library/stdlib.c
```diff
@@ -115,23 +115,43 @@ inline void *malloc(__CPROVER_size_t malloc_size)
   // realistically, malloc may return NULL,
   // and __CPROVER_allocate doesn't, but no one cares
   __CPROVER_HIDE:;
-  void *malloc_res;
-  malloc_res = __CPROVER_allocate(malloc_size, 0);
 
-  // make sure it's not recorded as deallocated
-  __CPROVER_deallocated=(malloc_res==__CPROVER_deallocated)?0:__CPROVER_deallocated;
+    if(__CPROVER_malloc_failure_mode == 1)
```
chrisr-diffblue: In an ideal world :-) rather than magic numbers here for the failure modes, these would be static named constants. Bonus points of those named constants were exposed to user programs as __CPROVER_blah constants.

regression/cbmc/array_constraints1/test.desc
chrisr-diffblue: Agree with Martin here - it's probably good that the number of VC's should be checked in the tests. I appreciate that is an added maintenance issue but we wouldn't want to silently regress the number of VCs either

src/cbmc/cbmc_parse_options.cpp
chrisr-diffblue: Agree with Martin here too.

src/util/config.cpp
```diff
@@ -1093,6 +1093,16 @@ bool configt::set(const cmdlinet &cmdline)
     bv_encoding.is_object_bits_default = false;
   }
 
+  if(cmdline.isset("malloc-fail-assert") && cmdline.isset("malloc-fail-null"))
+  {
+    throw invalid_command_line_argument_exceptiont{
+      "at most one malloc failure mode is acceptable", "--malloc-fail-null"};
+  }
+  if(cmdline.isset("malloc-fail-null"))
+    ansi_c.malloc_failure_mode = 1;
```
chrisr-diffblue: Again, would be nice to avoid the magic numbers here.

chrisr-diffblue (changes requested):
Looks really good - two main things I'd like to see before merging:
- Don't replace the VC numbers with patterns in regression tests
- Don't use magic numbers for malloc_failure_mode

chrisr-diffblue:
Oh, and one final thing that would make me really happy :-) is if the user docs were updated to describe this new feature....

src/ansi-c/library/stdlib.c
xbauch: Done.

xbauch:
@chrisr-diffblue 

> Looks really good - two main things I'd like to see before merging:
> 
> * Don't replace the VC numbers with patterns in regression tests

Went to exact numbers

> * Don't use magic numbers for malloc_failure_mode

There are now constants in `cprover.h` and an enum in `config.ansi_c`.
Also added some cprover-manual docu.

It's in a separate commit (https://github.com/diffblue/cbmc/commit/ed741ecd7b8353a22d46193e8cb8dbacb5e352af) for reviewing -- will be squashed before merging (if approved).

chrisr-diffblue (approved):
That looks great - many thanks for fixing all the nitpicks!

xbauch:
test group | before | after | slowdown
-- | -- | -- | --
cbmc-CORE | 127.9 | 146.88 | 1.14839718530102
cbmc-paths-lifo-CORE | 159.87 | 182.46 | 1.14130230812535
cbmc-library-CORE | 19.11 | 22.46 | 1.1753008895866
goto-analyzer-CORE | 15.89 | 17.75 | 1.11705475141598
ansi-c-CORE | 70.67 | 72.89 | 1.03141361256545
cbmc-cprover-smt2-CORE | 143.8 | 159.32 | 1.10792767732962
goto-instrument-CORE | 67.93 | 69.95 | 1.02973649344914
jbmc-CORE | 179.93 | 196.51 | 1.09214694603457
strings-smoke-tests-CORE | 68.95 | 69.65 | 1.01015228426396
strings-smoke-tests-symex-driven-lazy-loading-CORE | 260.05 | 300.16 | 1.15423956931359
jbmc-symex-driven-lazy-loading-CORE | 725.48 | 802.43 | 1.1060677068975
jbmc-strings-CORE | 317.44 | 327.24 | 1.03087197580645
jbmc-strings-symex-driven-lazy-loading-CORE | 950.72 | 1018.15 | 1.07092519353753
total | 2053.73 | 2233.66 | 1.0876113218388

There is some time regression but the CI is suddenly takes >1800s to compile (before it was ~50s).

smowton:
That's ccache.

xbauch:
> That's ccache.

Do you mean that it's ccache that's timing out or that the difference between build times is caused by ccache?

chrisr-diffblue:
Difference in build times, I suspect. Because you've touched a few header files that are reasonably widely included, ccache is going to end up thinking it does actually need to rebuild a lot of things, rather than just pulling them from the cache.

chrisr-diffblue:
Since the failing Travis job did actually get as far as saving it's ccache, there's a reasonably high (by Travis standards...) that simply re-running the failed job will make it pass.

xbauch:
> Since the failing Travis job did actually get as far as saving it's ccache, there's a reasonably high (by Travis standards...) that simply re-running the failed job will make it pass.

This was my 4th attempt.

chrisr-diffblue:
:-s Well that sucks...

chrisr-diffblue:
Looking a little deeper... the *previous* Travis run had two of the OSX configs timeout... *one* of those configs got as far as saving its cache, the other didnt. This most recent Travis run has only a single Travis failure - the config that previously timed out before saving it's cache - while the config that did manage to save it's cache has now passed. So I think it is worth one more attempt to re-run the Travis jobs since I think all the configs have now saved updated caches.

xbauch:
The changes in this PR cause ~7% slow down (the CORE test suite from 1560s to 1674s). It might be avoided if we figured out how to introduce `#ifdef`s in the library models based on user options. Any ideas @peterschrammel? Is it worth the trouble?

Issue #4605, opened 2019-05-01
CBMC malloc should be able to non-determinstically fail


danielsn:
```
#include <assert.h>
#include <stdlib.h>
#include <stdint.h>

int main() {
    char*p = malloc(100);
    assert(p);
}
```
the assertion never triggers, tho it ought to be able to, given that real malloc can fail.

hannes-steffenhagen-diffblue:
This has been implemented with the `--malloc-may-fail` and the `--malloc-fail-null` flags.

PR #5212, opened 2020-01-07
Add malloc-may-fail option to goto-check

xbauch:
And append an `assert(false)` property to every `malloc` call.

### DIFF
--- a/regression/cbmc/Pointer_byte_extract5/no-simplify.desc
+++ b/regression/cbmc/Pointer_byte_extract5/no-simplify.desc
@@ -4,7 +4,7 @@ main.i
 ^EXIT=10$
 ^SIGNAL=0$
 array\.List dynamic object upper bound in p->List\[2\]: FAILURE
-\*\* 1 of 15 failed
+\*\* 1 of 16 failed
 --
 ^warning: ignoring
 --
--- a/regression/cbmc/Pointer_byte_extract5/test.desc
+++ b/regression/cbmc/Pointer_byte_extract5/test.desc
@@ -4,7 +4,7 @@ main.i
 ^EXIT=10$
 ^SIGNAL=0$
 array\.List dynamic object upper bound in p->List\[2\]: FAILURE
-\*\* 1 of 13 failed
+\*\* 1 of 14 failed
 --
 ^warning: ignoring
 --
--- a/regression/cbmc/malloc-may-fail/test.desc
+++ b/regression/cbmc/malloc-may-fail/test.desc
@@ -1,0 +1,9 @@
+CORE
+main.c
+--malloc-may-fail --malloc-fail-null
+^EXIT=10$
+^SIGNAL=0$
+^\[main.assertion.\d+\] line \d+ assertion p: FAILURE$
+^VERIFICATION FAILED$
+--
+^warning: ignoring
--- a/regression/goto-analyzer/constant_propagation_02/test.desc
+++ b/regression/goto-analyzer/constant_propagation_02/test.desc
@@ -3,7 +3,7 @@ main.c
 --constants --simplify out.gb
 ^EXIT=0$
 ^SIGNAL=0$
-^Simplified:  assert: 1, assume: 0, goto: 1, assigns: 8, function calls: 0$
+^Simplified:  assert: 1, assume: 0, goto: 1, assigns: 9, function calls: 0$
 ^Unmodified:  assert: 0, assume: 0, goto: 0, assigns: 17, function calls: 2$
 --
 ^warning: ignoring
--- a/src/ansi-c/library/stdlib.c
+++ b/src/ansi-c/library/stdlib.c
@@ -112,50 +112,55 @@ __CPROVER_bool __VERIFIER_nondet___CPROVER_bool();
 
 inline void *malloc(__CPROVER_size_t malloc_size)
 {
-  // realistically, malloc may return NULL,
-  // and __CPROVER_allocate doesn't, but no one cares
-  __CPROVER_HIDE:;
+// realistically, malloc may return NULL,
+// but we only do so if `--malloc-may-fail` is set
+__CPROVER_HIDE:;
 
+  if(__CPROVER_malloc_failure_mode == __CPROVER_malloc_failure_mode_return_null)
+  {
+    __CPROVER_bool should_malloc_fail = __VERIFIER_nondet___CPROVER_bool();
     if(
-      __CPROVER_malloc_failure_mode ==
-      __CPROVER_malloc_failure_mode_return_null)
+      malloc_size > __CPROVER_max_malloc_size ||
+      (__CPROVER_malloc_may_fail && should_malloc_fail))
     {
-      if(malloc_size > __CPROVER_max_malloc_size)
-      {
-        return (void *)0;
-      }
+      return (void *)0;
     }
-    else if(
-      __CPROVER_malloc_failure_mode ==
-      __CPROVER_malloc_failure_mode_assert_then_assume)
-    {
-      __CPROVER_assert(
-        malloc_size <= __CPROVER_max_malloc_size,
-        "max allocation size exceeded");
-      __CPROVER_assume(malloc_size <= __CPROVER_max_malloc_size);
-    }
+  }
+  else if(
+    __CPROVER_malloc_failure_mode ==
+    __CPROVER_malloc_failure_mode_assert_then_assume)
+  {
+    __CPROVER_assert(
+      malloc_size <= __CPROVER_max_malloc_size, "max allocation size exceeded");
+    __CPROVER_assume(malloc_size <= __CPROVER_max_malloc_size);
 
-    void *malloc_res;
-    malloc_res = __CPROVER_allocate(malloc_size, 0);
+    __CPROVER_bool should_malloc_fail = __VERIFIER_nondet___CPROVER_bool();
+    __CPROVER_assert(
+      !__CPROVER_malloc_may_fail || !should_malloc_fail,
+      "max allocation may fail");
+    __CPROVER_assume(!__CPROVER_malloc_may_fail || !should_malloc_fail);
+  }
 
-    // make sure it's not recorded as deallocated
-    __CPROVER_deallocated =
-      (malloc_res == __CPROVER_deallocated) ? 0 : __CPROVER_deallocated;
+  void *malloc_res;
+  malloc_res = __CPROVER_allocate(malloc_size, 0);
 
-    // record the object size for non-determistic bounds checking
-    __CPROVER_bool record_malloc = __VERIFIER_nondet___CPROVER_bool();
-    __CPROVER_malloc_object =
-      record_malloc ? malloc_res : __CPROVER_malloc_object;
-    __CPROVER_malloc_size = record_malloc ? malloc_size : __CPROVER_malloc_size;
-    __CPROVER_malloc_is_new_array =
-      record_malloc ? 0 : __CPROVER_malloc_is_new_array;
+  // make sure it's not recorded as deallocated
+  __CPROVER_deallocated =
+    (malloc_res == __CPROVER_deallocated) ? 0 : __CPROVER_deallocated;
 
-    // detect memory leaks
-    __CPROVER_bool record_may_leak = __VERIFIER_nondet___CPROVER_bool();
-    __CPROVER_memory_leak =
-      record_may_leak ? malloc_res : __CPROVER_memory_leak;
+  // record the object size for non-determistic bounds checking
+  __CPROVER_bool record_malloc = __VERIFIER_nondet___CPROVER_bool();
+  __CPROVER_malloc_object =
+    record_malloc ? malloc_res : __CPROVER_malloc_object;
+  __CPROVER_malloc_size = record_malloc ? malloc_size : __CPROVER_malloc_size;
+  __CPROVER_malloc_is_new_array =
+    record_malloc ? 0 : __CPROVER_malloc_is_new_array;
 
-    return malloc_res;
+  // detect memory leaks
+  __CPROVER_bool record_may_leak = __VERIFIER_nondet___CPROVER_bool();
+  __CPROVER_memory_leak = record_may_leak ? malloc_res : __CPROVER_memory_leak;
+
+  return malloc_res;
 }
 
 /* FUNCTION: __builtin_alloca */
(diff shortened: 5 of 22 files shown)


xbauch:
This may fix #4605.

allredj (commented):
:heavy_check_mark:
Passed Diffblue compatibility checks (cbmc commit: effdd300e96d1165003a86502e8120e48b388db3).
Build URL: https://travis-ci.com/diffblue/test-gen/builds/143378007

martin-cs (commented):
Why not just change the model of what malloc does?  That seems to keep all of the functionality in one place?

https://github.com/diffblue/cbmc/blob/develop/src/ansi-c/library/stdlib.c#L113

Maybe it is time to change https://github.com/diffblue/cbmc/blob/develop/src/ansi-c/library/stdlib.c#L115 specifically.

xbauch:
> Why not just change the model of what malloc does?

I think the reasoning here is to avoid "false-positive" alarms with every `malloc` call by default, and only to be correctly reporting the potential of malloc to fail if requested.

martin-cs:
> I think the reasoning here is to avoid "false-positive" alarms with
> every `malloc` call by default, and only to be correctly reporting
> the potential of malloc to fail if requested.

Then have a global as part of the model that allows it to be turned on
/ off / linked to a symbolic value during analysis.  It doesn't have to
be, and shouldn't be, hard-coded to one way or the other.

src/ansi-c/library/stdlib.c
```diff
@@ -115,23 +115,31 @@ inline void *malloc(__CPROVER_size_t malloc_size)
   // realistically, malloc may return NULL,
   // and __CPROVER_allocate doesn't, but no one cares
   __CPROVER_HIDE:;
-  void *malloc_res;
-  malloc_res = __CPROVER_allocate(malloc_size, 0);
 
-  // make sure it's not recorded as deallocated
-  __CPROVER_deallocated=(malloc_res==__CPROVER_deallocated)?0:__CPROVER_deallocated;
+    __CPROVER_assert(
+      __CPROVER_malloc_may_fail == (__CPROVER_bool)0, "malloc may fail");
```
hannes-steffenhagen-diffblue: ? Why is there an assert that this is false? I thought this would be more like `if (__CPROVER_malloc_may_fail && __VERIFIER_nondet___CPROVER_bool()) return NULL;`

xbauch: Because conditionally returning `NULL` breaks too many things down the line in CBMC.

xbauch: I take it back. Returning `NULL` seems to work here just fine.

src/ansi-c/library/stdlib.c
```diff
@@ -115,23 +115,31 @@ inline void *malloc(__CPROVER_size_t malloc_size)
   // realistically, malloc may return NULL,
   // and __CPROVER_allocate doesn't, but no one cares
   __CPROVER_HIDE:;
-  void *malloc_res;
-  malloc_res = __CPROVER_allocate(malloc_size, 0);
 
-  // make sure it's not recorded as deallocated
-  __CPROVER_deallocated=(malloc_res==__CPROVER_deallocated)?0:__CPROVER_deallocated;
+    if(__CPROVER_malloc_may_fail && __VERIFIER_nondet___CPROVER_bool())
```
thk123: :no_entry_sign: Comment above needs updating

regression/cbmc/Pointer_byte_extract5/main.i
```diff
@@ -11,13 +11,13 @@ typedef union
 typedef struct
 {
   int Count;
-  Union List[1];
+  Union List[0];
```
thk123: :question: Why does this change? The change should go in a separate commit with a why

regression/cbmc/Pointer_byte_extract5/no-simplify.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+CORE broken-smt-backend
```
thk123: :no_entry_sign: Please add an explanation / link to relevant issue as to why this is broken with smt-backend at the bottom of the desc file

regression/cbmc/Pointer_byte_extract5/test.desc
```diff
@@ -4,7 +4,7 @@ main.i
 ^EXIT=10$
 ^SIGNAL=0$
 array\.List dynamic object upper bound in p->List\[2\]: FAILURE
-\*\* 1 of 11 failed 
+\*\* 1 of 12 failed
```
thk123: :question: Why does this change? 
:pick: Probably best to set to `\d+` since number of goals isn't what is being tested

regression/cbmc/malloc-may-fail/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
```
thk123, line 1: :no_entry_sign: Please add `.desc` file _without_ the option, that checks for `VERIFICATION SUCCESSFUL`

regression/goto-analyzer/constant_propagation_02/test.desc
```diff
@@ -3,7 +3,7 @@ main.c
 --constants --simplify out.gb
 ^EXIT=0$
 ^SIGNAL=0$
-^Simplified:  assert: 1, assume: 0, goto: 1, assigns: 7, function calls: 0$
+^Simplified:  assert: 1, assume: 0, goto: 1, assigns: 8, function calls: 0$
```
thk123: :pick: Please group these changes into a separate commit with why these numbers are changing

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1041,6 +1041,7 @@ void cbmc_parse_optionst::help()
     "\n"
     "Backend options:\n"
     " --object-bits n              number of bits used for object addresses\n"
+    " --malloc-may-fail            enable failures for malloc calls\n"
```
thk123: :bulb: I think it'd be better to be a little more explicit. 

```
allow malloc calls to return a null pointer
```

src/cbmc/cbmc_parse_options.h
```diff
@@ -48,6 +48,7 @@ class optionst;
   "(document-subgoals)(outfile):(test-preprocessor)" \
   "D:I:(c89)(c99)(c11)(cpp98)(cpp03)(cpp11)" \
   "(object-bits):" \
+  "(malloc-may-fail)" \
```
thk123: :grey_question: should this option be available in other C based tools, like goto-analyzer, or does that not really make sense?

src/util/config.cpp
```diff
@@ -1093,6 +1093,11 @@ bool configt::set(const cmdlinet &cmdline)
     bv_encoding.is_object_bits_default = false;
   }
 
+  if(cmdline.isset("malloc-may-fail"))
```
thk123: :pick: ?

```suggestion
  ansi_c.malloc_may_fail = cmdline.isset("malloc-may-fail");
```

src/ansi-c/library/stdlib.c
xbauch: Updated.

regression/cbmc/Pointer_byte_extract5/main.i
xbauch: This was undefined behaviour (and the changed in malloc exposed that CBMC does not really support that). It now has separate commit with explanation.

regression/cbmc/Pointer_byte_extract5/no-simplify.desc
xbauch: Added.

regression/cbmc/Pointer_byte_extract5/test.desc
xbauch: Explained and changed.

regression/cbmc/malloc-may-fail/test.desc
xbauch: Added.

regression/goto-analyzer/constant_propagation_02/test.desc
xbauch: In separate commit now.

src/cbmc/cbmc_parse_options.cpp
xbauch: Changed.

src/cbmc/cbmc_parse_options.h
xbauch: Ambivalent.

src/util/config.cpp
xbauch: Done.

regression/cbmc/malloc-may-fail/test.desc
```diff
@@ -0,0 +1,9 @@
...
+main.c
+--malloc-may-fail
+^EXIT=10$
+^SIGNAL=0$
+^\[main.assertion.\d+\] line \d+ assertion p: FAILURE$
+^VERIFICATION FAILED$
+--
+^warning: ignoring
```
thk123, line 9: :pick: in future, we've found it helpful in test-gen to include a description explaining the purpose of the test below the line. for all tests. It helps when a test fails to understand why does it even exist. E.g

```
--
^warning: ignoring
--
Test that if --malloc-may-fail is enabled, that assert can return a null pointer
```

thk123 (approved):

smowton (approved):

src/ansi-c/library/stdlib.c
```diff
@@ -115,6 +115,9 @@ inline void *malloc(__CPROVER_size_t malloc_size)
   // realistically, malloc may return NULL,
   // and __CPROVER_allocate doesn't, but no one cares
   __CPROVER_HIDE:;
+  if(__CPROVER_malloc_may_fail && __VERIFIER_nondet___CPROVER_bool())
+    return (void *)0;
```
tautschnig: 1) It may be more consistent to just declare a local, uninitialised variable of type `__CPROVER_bool` to avoid the side-effect of a function call.
2) The non-determinism also needs to be introduced in `calloc`.
3) `realloc` now needs to handle the case of a failing `malloc`.

regression/cbmc/Pointer_byte_extract5/main.i
tautschnig: I find it rather surprising that the changes in this PR expose the lack of support for this C90-hack. What exactly was going wrong here?

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1041,6 +1041,7 @@ void cbmc_parse_optionst::help()
     "\n"
     "Backend options:\n"
     " --object-bits n              number of bits used for object addresses\n"
+    // NOLINTNEXTLINE(whitespace/line_length)
```
tautschnig: The commit message says "will be squashed" so presumably that should happen ;-)

tautschnig (commented):
The changes to extend support to calloc, realloc need to happen before this can be merged. The modified regression test I'd not insist as much on, but still I'd like to see clarification.

regression/cbmc/Pointer_byte_extract5/main.i
xbauch: Why this particular PR exposes the problem is a bit more complicated. But the problem itself can be easily reproduced, e.g. (CMBC reports the assertion as failing):
```
#include <malloc.h>
#include <assert.h>

struct Foo
{
  int i;
  char data[1];
};

int main()
{
  struct Foo* foo = malloc(sizeof(struct Foo) + sizeof(char));
  foo->data[1]='b'; // set data[1]
  assert(foo->data[1]=='b'); // check data[1] -- should succeed

  return 0;
}
```

src/ansi-c/library/stdlib.c
xbauch: Add 1. No can do: using uninitialised variables doesn't even compile with the current flag set. I can initialise it to `nondet_bool`.

xbauch: 2,3 are done.

src/cbmc/cbmc_parse_options.cpp
xbauch: Squashed.

martin-cs:
There seems to be some duplication w.r.t. https://github.com/diffblue/cbmc/pull/5210  which is preferred?

chrisr-diffblue:
@martin-cs I believe the difference between this PR and #5210 is around the behaviour they are modelling. #5210 is modelling the very specific case of what happens if the user code attempts to alloc a chunk of memory that is _larger than CBMC can represent_. Where as this PR is modelling the case where "the system" (e.g. OS) cannot satisfy a malloc request because of insufficient memory availability. This is a non-deterministic situation, so this PR models the situation where malloc may non-deterministically fail.

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1041,6 +1041,8 @@ void cbmc_parse_optionst::help()
...
+    " --malloc-may-fail            allow malloc calls to return a null pointer\n"
```
chrisr-diffblue: Similar to #5210, should this help message move up into the "Program instrumentation options" section?

chrisr-diffblue (commented):
Looks pretty good, but would also be super nice if you could extend the user docs at the same time ;-)

src/cbmc/cbmc_parse_options.cpp
xbauch: Moved to the instrumentation section + the documentation was also added.

src/ansi-c/library/stdlib.c
```diff
@@ -112,50 +137,55 @@ __CPROVER_bool __VERIFIER_nondet___CPROVER_bool();
...
+    __CPROVER_malloc_failure_mode ==
+    __CPROVER_malloc_failure_mode_assert_then_assume)
+  {
+    __CPROVER_assert(
+      malloc_size <= __CPROVER_max_malloc_size, "max allocation size exceeded");
+    __CPROVER_assume(malloc_size <= __CPROVER_max_malloc_size);
+
+    __CPROVER_bool should_malloc_fail = __VERIFIER_nondet___CPROVER_bool();
+    __CPROVER_assert(
+      !__CPROVER_malloc_may_fail || !should_malloc_fail,
```
hannes-steffenhagen-diffblue, line 164: I don’t think it makes a lot of sense to have the assertion here in this form, this will _always_ fail the way it is written.

hannes-steffenhagen-diffblue (commented):
This should have an implementation for `realloc` as well, also I don’t think we should mix up the cases where malloc fails due to some nondeterministic condition and when it fails because we reach cprover internal limits.

chrisr-diffblue (approved):

martin-cs (approved):
I like this approach a lot more than the previous one.

PR #5218, opened 2020-01-28
Feature/source location hint on symex failure

hannes-steffenhagen-diffblue:
Note that due to the nature of the change there aren’t any tests for the behaviour changes, though I’m going to add some unit tests for the new utility type.

### DIFF
--- a/src/goto-checker/symex_bmc.cpp
+++ b/src/goto-checker/symex_bmc.cpp
@@ -44,10 +44,12 @@ void symex_bmct::symex_step(
   statet &state)
 {
   const source_locationt &source_location = state.source.pc->source_location;
-  // Turn on throwing invariants so we can safely log invariant failure during Symex
+  // Turn on throwing invariants so we can safely log invariant failure during
+  // Symex
   auto const invariants_should_throw_during_symex_step =
     cbmc_invariants_should_throwt{};
-  // If an exception is thrown during symex step, log the last relevant source location
+  // If an exception is thrown during symex step, log the last relevant
+  // source location
   auto const print_step_location_on_error =
     on_exit_scope_with_exception([source_location, this]() {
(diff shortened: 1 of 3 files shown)


src/goto-checker/symex_bmc.cpp
```diff
@@ -43,15 +44,32 @@ void symex_bmct::symex_step(
   statet &state)
 {
   const source_locationt &source_location = state.source.pc->source_location;
-
+  // Turn on throwing invariants so we can safely log invariant failure during Symex
```
thk123: :question: why not just wrap the whole thing in a try catch? Isn't that a more idiomatic way to accomplish the same thing?

hannes-steffenhagen-diffblue: @thk123 I don’t know about idiomatic, CPPCoreGuidelines recommends use of RAII ([1](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Re-catch), [2](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Re-finally)), though in this case the problem is this just doesn’t quite work with our setup because `uncaught_exceptions` is C++17 apparently. I can fix the scope guard thing (it can still be done, logic is just slightly different) or I can change this to try catch. I do have a preference for the RAII based approach but I’m not married to it if other people dislike it very much.

NlightNFotis (approved):
This looks good to me, from what I've seen. I'm also in agreement with the rational for adding this.

But please squash the WIP commits first.

martin-cs (commented):
This feels like there could be a simpler way of doing this.  What exactly do you mean by "can safely log invariant failure during Symex"?  What is the unsafe thing?

PR #5220, opened 2020-01-30
Revamp function contracts command line interface

karkhaz:
This PR is the first in a patch series attempting to revive the function contracts & loop invariants work.

The first commit adds flags to goto-instrument to use the function contracts feature. Future commits will build on these flags, but I wanted to get a review of these first. The second commit fixes a bug where CBMC would crash when analyzing code where a function call has been replaced with its contract, and the contract mentioned the function return value, but the function call had ignored the return value.


karkhaz:
Not sure why code coverage has dropped significantly. It might be because I've changed a command line flag, so some of the tests (which were marked as KNOWNBUG) fail much earlier than they used to (this would be all the tests under `regression/contracts`, apart from the ones that I've specifically grouped under CORE). Not sure if this is a problem... @allredj do you own codecov?

src/goto-instrument/code_contracts.cpp
```diff
@@ -257,14 +229,74 @@ const symbolt &code_contractst::new_tmp_symbol(
...
+  // Insert wrapper function into goto_functions
+  auto nexist_old_fun = goto_functions.function_map.find(original);
+  INVARIANT(
+    nexist_old_fun == goto_functions.function_map.end(),
+    "There should be no function called " + fun_to_enforce +
+      " in the function map because that function should have had its"
+      " name mangled");
+
+  auto mangled_fun = goto_functions.function_map.find(mangled);
+  INVARIANT(
```
chrisr-diffblue, line 329: Many thanks for all the INVARIANT's - makes this a lot more readable/documented.

src/goto-instrument/code_contracts.cpp
```diff
@@ -350,7 +382,7 @@ void code_contractst::add_contract_check(
       requires_cond, requires.source_location()));
   }
 
-  // ret=function(parameter1, ...)
+  // ret=mangled_fun(parameter1, ...)
```
chrisr-diffblue, line 438: Also thanks for keeping comments updated...

src/goto-instrument/code_contracts.cpp
```diff
@@ -370,25 +402,74 @@ void code_contractst::add_contract_check(
...
+  const std::list<std::string> &funs_to_replace)
 {
-  Forall_goto_functions(it, goto_functions)
-    code_contracts(it->second);
-
-  goto_functionst::function_mapt::iterator i_it=
-    goto_functions.function_map.find(INITIALIZE_FUNCTION);
-  assert(i_it!=goto_functions.function_map.end());
+  bool fail = false;
+  Forall_goto_functions(fit, goto_functions)
```
chrisr-diffblue: Do we still need to use the Forall_ macros, or do range-based loops now work? I had a feeling the did after some of the various cleanups that have gone into CBMC.

src/goto-instrument/code_contracts.cpp
```diff
@@ -370,25 +402,74 @@ void code_contractst::add_contract_check(
...
+      {
+        const code_function_callt &call = it->get_function_call();
+
+        // TODO we don't handle function pointers
+        if(call.function().id() != ID_symbol)
+          continue;
+
+        const irep_idt &fun_name =
+          to_symbol_expr(call.function()).get_identifier();
+        auto found = std::find(
```
chrisr-diffblue, line 490: Rather than using a std::list for funs_to_replace, is there any efficiency gain to use a std::set? Just wondering how large you expect funs_to_replace to be, given that you'll be doing a search through it for every call site in the program?

src/goto-instrument/code_contracts.cpp
```diff
@@ -370,25 +402,74 @@ void code_contractst::add_contract_check(
...
+          funs_to_replace.begin(), funs_to_replace.end(), id2string(fun_name));
+        if(found == funs_to_replace.end())
+          continue;
+
+        if(!has_contract(fun_name))
```
chrisr-diffblue: Given that the only way to get to this block of code is if fun_name is in funs_to_replace, it would be better from a performance point of view to move this out of the nest loops and just do an initial iteration through funs_to_replace before even attempting to do the callsite scanning. Of course, that leads to slightly different user behaviour... If this is done outside the loops they get the error message for any function in funs_to_replace with no contract (regardless of whether they get called anywhere or not), whereas if it's done here inside the loops, they only get the error for any of those that are actually _called_.

chrisr-diffblue (commented):
Generally looks really nice - just a couple of comments/queries.

allredj:
@karkhaz No I'm not in charge of that one! :-)

src/goto-instrument/code_contracts.cpp
hannes-steffenhagen-diffblue: Regular `for(auto &goto_function : goto_functions.function_map)` should work here.

src/goto-instrument/code_contracts.cpp
```diff
@@ -370,25 +402,74 @@ void code_contractst::add_contract_check(
...
   goto_functions.update();
+
+  return false;
 }
 
-void code_contracts(goto_modelt &goto_model)
+bool code_contractst::replace_calls()
 {
-  code_contractst(goto_model.symbol_table, goto_model.goto_functions)();
+  std::list<std::string> funs_to_replace;
```
hannes-steffenhagen-diffblue: Is there a good reason for this to be a `std::list`? It has a pretty big memory and performance overhead, unless you need stable iterators it’s probably better to just use a `vector`. (Same with the other instances of `list`).

src/goto-instrument/code_contracts.cpp
```diff
@@ -370,25 +402,74 @@ void code_contractst::add_contract_check(
...
+  Forall_goto_functions(fit, goto_functions)
+  {
+    if(has_contract(fit->first))
+      funs_to_replace.push_back(id2string(fit->first));
```
hannes-steffenhagen-diffblue: Why are you converting the `irep_idt`s to string? Seems to me like the follow up logic would work just as well with `irep_idt`?

chrisr-diffblue:
@yumibagge Do you have any idea why @karkhaz might be seeing these drop in coverage?

yumibagge:
> @yumibagge Do you have any idea why @karkhaz might be seeing these drop in coverage?

Believe regression tests are not run?

src/goto-instrument/goto_instrument_parse_options.h
```diff
@@ -96,7 +97,11 @@ Author: Daniel Kroening, kroening@kroening.com
   "(interpreter)(show-reaching-definitions)" \
   "(list-symbols)(list-undefined-functions)" \
   "(z3)(add-library)(show-dependence-graph)" \
-  "(horn)(skip-loops):(apply-code-contracts)(model-argc-argv):" \
+  "(horn)(skip-loops):(model-argc-argv):" \
+  "(" FLAG_REPLACE_CALL "):" \
+  "(" FLAG_REPLACE_ALL_CALLS ")" \
+  "(" FLAG_ENFORCE_CONTRACT "):" \
+  "(" FLAG_ENFORCE_ALL_CONTRACTS ")" \
```
martin-cs, line 104: Nice pattern.

martin-cs (approved):
Yes!

src/goto-instrument/code_contracts.h
```diff
@@ -14,8 +14,124 @@ Date: February 2016
...
+  /// with an assertion that the `requires` clause holds followed by an
+  /// assumption that the `ensures` clause holds. In order to ensure that `F`
+  /// actually abides by its `ensures` and `requires` clauses, you should
+  /// separately call `code_constractst::enforce_contracts()` on `F` and verify
+  /// it using `cbmc --function F`.
+  ///
+  /// \return `true` on failure, `false` otherwise
+  bool replace_calls(const std::list<std::string> &);
+
+  /// \brief Turn requires & assumes into assumptions and assertions for each of
```
karkhaz: Turn requires & **ensures**

src/goto-instrument/code_contracts.cpp
karkhaz: The follow-up logic does work just as well with `irep_idt`, but the `replace_calls` function also gets called directly from `goto_instrument_parseoptionst`---which passes in command-line arguments as `std::string`. I don't really have a strong opinion, but some conversion is necessary anyway, and my general preference is to write as little code as possible in `goto_instrument_parseoptionst` and do any necessary conversions in the class that does the work. But happy to do it a different way if you prefer.

karkhaz:
All comments addressed except by one from @hannes-steffenhagen-diffblue, thanks!

chrisr-diffblue (approved):
Looks good - I'm happy for this to go in.

hannes-steffenhagen-diffblue (approved):
I don’t have any major problems with this.

PR #5221, opened 2020-01-31
Don't evaluate arbitrary python

karkhaz:
CBMC's linker script parser did not contain a real parser for arithmetic expressions. It passed such expressions to python's `eval`, since linker script expressions are similar enough to Python's ones. This permits the possibility of executing arbitrary python code using a maliciously-formed linker script.

This commit uses python's ast module to safely parse well-formed arithmetic expressions in linker scripts and reject anything else.

### DIFF
--- a/scripts/ls_parse.py
+++ b/scripts/ls_parse.py
@@ -182,7 +182,7 @@ def get_linker_script_data(script):
                 match = m
                 matched_str = buf[m.span()[0]:m.span()[1]]
                 for s, v in locals().items():
-                    if v is regex and s is not "regex":
+                    if v is regex and s != "regex":
                         matched_re = s
         if jump_fun is not None:
             info("regex '%s' matched '%s'", matched_re, matched_str)
@@ -321,9 +321,11 @@ def memory_cmd_fun(state, _, buf):
     asrt(not state["MEM"], "encountered MEMORY twice", buf)
     state["MEM"] = True
 
+
 def unknown_cmd_fun(state, _, buf):
     asrt(not state["MEM"], "encountered UNKNOWN twice", buf)
     state["UNKNOWN"] = True
+
 
 def match_up_expr_assigns(state):
     blocks = set([data["origin"] for data in state["expr-assigns"]])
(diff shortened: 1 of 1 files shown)


chrisr-diffblue (approved):
Ouch! Good catch. This looks reasonable and I'm happy for it to go in. It might be nice to have a regression test (especially one that shows a malformed linker script being rejected), but if this is covered by existing tests somehow, thats fine.

scripts/ls_parse.py
```diff
@@ -6,17 +6,26 @@
...
+import platform
 import re
 import subprocess
 import sys
 import textwrap
 import traceback
 
+# ast.Num was deprecated in python 3.8
+_plat = platform.python_version().split(".")
+if _plat[0] == "3" and int(_plat[1]) > 7:
```
hannes-steffenhagen-diffblue: Kind of a nitpick, but shouldn’t this `int(_plat[0] > 3) or (_plat[0] == "3" and int(_plat[1]) > 7)`? This way a future Python 4 would suddenly have `ast.Num` reintroduced.

scripts/ls_parse.py
```diff
@@ -6,17 +6,26 @@
...
 import os
+import platform
 import re
 import subprocess
 import sys
 import textwrap
 import traceback
 
+# ast.Num was deprecated in python 3.8
+_plat = platform.python_version().split(".")
```
hannes-steffenhagen-diffblue: You can just use `sys.version_info.{major,minor,micro}` for this.

hannes-steffenhagen-diffblue (approved):
I think the second commit needs a better commit message.

scripts/ls_parse.py
```diff
@@ -6,17 +6,25 @@
...
+import operator
 import os
 import re
 import subprocess
 import sys
 import textwrap
 import traceback
 
+# ast.Num was deprecated in python 3.8
+if sys.version_info.major > "3" or \
```
hannes-steffenhagen-diffblue: `sys.version_info.major` is a `int` so no need for the quotes.

hannes-steffenhagen-diffblue:
`AWS CodeBuild us-east-1 (cbmc-linux-clang)` seems to be failing due to network failure.

peterschrammel (approved):

PR #5226, opened 2020-02-06
String solver: avoid cubic formulas for string-to-int conversions

smowton:
Previously when the string radix was unknown we generated a conjunction of no-overflow checks
which was (in total) cubic in the number of digits in the string being converted to an integer.

For instance, it would define a series of 'sum's, like:
```
sum0 = digit0
sum1 = digit0 * radix + digit1
sum2 = (digit0 * radix + digit1) * radix + digit2
```

Where the definitions are truly inlined like this and don't use symbols to represent sum0...n,
so |sumn| is O(n), and the sum of all sums is O(n^2)

Then for each digit it would emit an overflow check:

```
constraint2 = size = 2 => nooverflow(sum1)
constraint3 = size = 3 => nooverflow(sum1) && nooverflow(sum2)
constraint4 = size = 4 => nooverflow(sum1) && nooverflow(sum2) && nooverflow(sum3)
```

However those sums are again expanded inline, meaning the sum of all constraints is O(n^3)

The obvious fix is to define symbols for sum0... sumn and then constraint0... constraintn,
but this slows the solver down considerably for small formulas (e.g. Byte.parseByte, with
max size 3 digits), so I compromise here by replacing the constraints with:

```
constraint2 = size >= 2 => nooverflow(sum1)
constraint3 = size >= 3 => nooverflow(sum2)
constraint4 = size >= 4 => nooverflow(sum3)
```

Thus the formula size is quadratic not cubic but still fast for cases where the radix is known.


src/solvers/strings/string_constraint_generator_valueof.cpp
```diff
@@ -424,31 +423,38 @@ string_constraint_generatort::add_axioms_for_characters_in_integer_string(
...
+      digit_constraints.push_back(result);*/
     }
     sum = new_sum;
 
-    const equal_exprt premise =
-      equal_to(array_pool.get_or_create_length(str), size);
+    exprt length_expr = array_pool.get_or_create_length(str);
 
-    if(!digit_constraints.empty())
+    if(no_overflow.has_value())
```
xbauch, line 435: Why not have the `size - 1 >= max_string_length - 2 || radix_ul == 0` check here and build the `no_overflow` conjunction inside `a5`?

smowton: AFAICT it wants to assert about the old sum, I'd guess that's why it was this way to begin with.

martin-cs (approved):
Fix sounds reasonable but 1. I don't know the string solver internals that well and 2. this doesn't sound like a very good design and the "unknown radix" part has me concerned, so, YMMV.

src/solvers/strings/string_constraint_generator_valueof.cpp
```diff
@@ -424,31 +423,38 @@ string_constraint_generatort::add_axioms_for_characters_in_integer_string(
...
     {
-      const implies_exprt a5(premise, conjunction(digit_constraints));
+      const binary_predicate_exprt string_length_gte_size{
```
romainbrenguier: :pick: why gte and not ge?

src/solvers/strings/string_constraint_generator_valueof.cpp
```diff
@@ -424,31 +423,38 @@ string_constraint_generatort::add_axioms_for_characters_in_integer_string(
...
-      const and_exprt no_overflow(
+      no_overflow = and_exprt{
         equal_exprt(sum, div_exprt(radix_sum, radix)),
-        binary_relation_exprt(new_sum, ID_ge, radix_sum));
-      digit_constraints.push_back(no_overflow);
+        binary_relation_exprt(new_sum, ID_ge, radix_sum)};
+      /*symbol_exprt result =
+        fresh_symbol("string_to_int_no_overflow_check", no_overflow.type());
+      constraints.existential.push_back(equal_exprt(result, no_overflow));
+      digit_constraints.push_back(result);*/
```
romainbrenguier: don't leave commented code

src/solvers/strings/string_constraint_generator_valueof.cpp
```diff
@@ -424,31 +423,38 @@ string_constraint_generatort::add_axioms_for_characters_in_integer_string(
...
+    exprt length_expr = array_pool.get_or_create_length(str);
 
-    if(!digit_constraints.empty())
+    if(no_overflow.has_value())
     {
-      const implies_exprt a5(premise, conjunction(digit_constraints));
+      const binary_predicate_exprt string_length_gte_size{
+        length_expr, ID_ge, from_integer(size, length_expr.type())};
+      const implies_exprt a5(string_length_gte_size, *no_overflow);
       constraints.existential.push_back(a5);
```
romainbrenguier, line 440: the corresponding documentation needs to be updated

romainbrenguier (approved):

src/solvers/strings/string_constraint_generator_valueof.cpp
smowton: Just habit to call this "Greater-Than or Equal", but I see we use ge elsewhere

src/solvers/strings/string_constraint_generator_valueof.cpp
smowton: Ooops will fix

src/solvers/strings/string_constraint_generator_valueof.cpp
smowton: Will do

jbmc/regression/strings-smoke-tests/java_parseint_unknown_radix/Test.java
```diff
@@ -0,0 +1,18 @@
+public class Test {
+
+    public static void testMe(String input, int radix) {
```
peterschrammel, line 3: weird formatting

jbmc/regression/strings-smoke-tests/java_parselong_unknown_radix/Test.java
```diff
@@ -0,0 +1,18 @@
+public class Test {
+
+    public static void testMe(String input, int radix) {
```
peterschrammel, line 3: weird formatting

jbmc/src/java_bytecode/java_string_library_preprocess.cpp
```diff
@@ -1574,6 +1574,12 @@ void java_string_library_preprocesst::initialize_conversion_table()
         std::placeholders::_2,
         std::placeholders::_3,
         std::placeholders::_4);
+  cprover_equivalent_to_java_function
```
peterschrammel, line 1577: 👍

peterschrammel (approved):

Issue #4701, opened 2019-05-23
Implement gcc integer overflow builtins.


danielsn:
Now that we have the overflow primitives, it would make sense to implement all the gcc builtins https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

danielsn:
@tautschnig may have solved this already; if so, feel free to close the ticket

danielsn:
Note: the current overflow primitives lead to a significant performance cost when used in the naïve way, since they require the user to do a multiplication, keep the overflow bit but throw away the result, and do it again to get the result.  Fixing this would help performance.

martin-cs:
Maybe.  In some cases the back-end will cache the two things back together.  More importantly though, there are way better encodings of multiply overflow check than doing the full multiply.

hannes-steffenhagen-diffblue:
@martin-cs What are the way better encodings for multiply overflow check?

martin-cs:
@hannes-steffenhagen-diffblue Somewhere there is a Z3 paper from ... '09 or so with a more compact encoding but, as it's Friday, let's see what we can do in a comment.

First do the unsigned case as it's easier.  Assume bit-vectors are length n.  Consider the easy cases first:
*. If either `x` or `y` is 1 or 0 then it trivially it doesn't overflow.
*. If both are under `sqrt(2^n - 1)` then it trivially doesn't overflow.  This is just a simple bit test and removes a number of bits from consideration outright.
*. Consider the `i`th and `j`th bits of x and y (apart from the ones above) with `i + j >= n` if they are both 1 then it overflows.  This removes a maybe half of the partial product terms.
*.  You can sum all of the remaining bits in the usual way but I think we can do better than that.  I'd start building a tree (well, DAG, so cache) of ways that top bit can overflow.  Either of it is a 1 and one of the lower log(n) bits carry a 1 up to it's position or it is a 0 and 2 or more of the lower log(n) bits carry up a 1 ...  apply this recursively and you have a tree.  This could probably be built stepwise as a refinement for the best encoding as I suspect it is rarely needed.

@danielsn / @hannes-steffenhagen-diffblue  If either of you have interest and a set of real-world performance critical benchmarks that depend on this I can try to turn this into code and / or talk one of you through doing it.

danpoe:
[...]

PR #5233, opened 2020-02-13
Add __builtin_add_overflow primitives

hannes-steffenhagen-diffblue:
This addresses part of https://github.com/diffblue/cbmc/issues/4701. Only add primitives in this PR. Note `gcc` also has a generic `__builtin_add_overflow` which isn’t implemented here. The issue also has some discussion about making overflow checks faster, this also isn’t implemented here.

### DIFF
--- a/regression/cbmc/gcc_builtin_add_overflow/main.c
+++ b/regression/cbmc/gcc_builtin_add_overflow/main.c
@@ -1,0 +1,103 @@
+#include <assert.h>
+#include <limits.h>
+
+void check_int(void)
+{
+  int const one = 1;
+  int result;
+
+  assert(!__builtin_sadd_overflow(one, one, &result));
+  assert(result == 2);
+  assert(__builtin_sadd_overflow(one, INT_MAX, &result));
+  assert(!__builtin_sadd_overflow(INT_MAX / 2, INT_MAX / 2, &result));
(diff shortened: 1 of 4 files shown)


xbauch (approved):
Looks good.

danpoe (approved):

regression/cbmc/gcc_builtin_add_overflow/main.c
```diff
@@ -0,0 +1,80 @@
+#include <limits.h>
+#include <assert.h>
+
```
chrisr-diffblue, line 3: Tests look good, but could we add a few more cases? In particular, these tests only cover some of the "easily optimised" cases (e.g value of 1, adding/subtracting to MAX/MIN), but it would be great to see some tests that also cover "mid-range value x + mid-range value y == overflow" and  "slightly smaller mid-range value x + mid-range value y != overflow" cases.

hannes-steffenhagen-diffblue: @chrisr-diffblue I’ve added some more cases to that end

chrisr-diffblue (approved):
Looks good, and thanks for expanding the tests.

NlightNFotis (approved):

PR #5237, opened 2020-02-20
Refinement iteration printing

thk123:

~I'm not totally convinced by my approach for finding out whether XML is turned on. I'm not really clear on why you can't always just pipe some xml to the message handler, and let it figure out what to do with it. (Ideally, for example, if you send xml to json-ui, then it outputs equivalent json)~
Went for something much simpler, might raise a seperate PR to make the handling of this stuff less ad-hoc.

### DIFF
--- a/src/goto-checker/symex_bmc_incremental_one_loop.cpp
+++ b/src/goto-checker/symex_bmc_incremental_one_loop.cpp
@@ -88,14 +88,7 @@ bool symex_bmc_incremental_one_loopt::should_stop_unwind(
     abort_unwind_decision.is_known(), "unwind decision should be taken by now");
   bool abort = abort_unwind_decision.is_true();
 
-
-  if(output_ui == ui_message_handlert::uit::XML_UI)
-  {
-    xmlt xml("current-unwinding");
-    xml.data = std::to_string(unwind);
-    log.statistics() << xml;
-  }
-
+  log_unwinding(unwind);
   log.statistics() << (abort ? "Not unwinding" : "Unwinding") << " loop " << id
                    << " iteration " << unwind;
 
@@ -141,4 +134,30 @@ bool symex_bmc_incremental_one_loopt::resume(
   symex_with_state(*state, get_goto_function, state->symbol_table);
 
   return should_pause_symex;
+}
+void symex_bmc_incremental_one_loopt::log_unwinding(unsigned unwind)
+{
+  const std::string unwind_num = std::to_string(unwind);
+  switch(output_ui)
+  {
+  case ui_message_handlert::uit::PLAIN:
+  {
+    log.statistics() << "Current unwinding: " << unwind_num << messaget::eom;
+    break;
+  }
+  case ui_message_handlert::uit::XML_UI:
+  {
+    xmlt xml("current-unwinding");
+    xml.data = unwind_num;
+    log.statistics() << xml << messaget::eom;
+    break;
+  }
+  case ui_message_handlert::uit::JSON_UI:
+  {
+    json_objectt json;
+    json["currentUnwinding"] = json_numbert(unwind_num);
+    log.statistics() << json << messaget::eom;
+    break;
+  }
+  }
 }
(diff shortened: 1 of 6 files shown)


smowton (approved):

src/goto-checker/symex_bmc_incremental_one_loop.cpp
```diff
@@ -19,7 +19,8 @@ symex_bmc_incremental_one_loopt::symex_bmc_incremental_one_loopt(
   symex_target_equationt &target,
   const optionst &options,
   path_storaget &path_storage,
-  guard_managert &guard_manager)
+  guard_managert &guard_manager,
+  ui_message_handlert::uit output_xml)
```
xbauch: I guess the name could be made more accurate.

src/goto-checker/symex_bmc_incremental_one_loop.h
```diff
@@ -10,6 +10,7 @@
 #define CPROVER_GOTO_CHECKER_SYMEX_BMC_INCREMENTAL_ONE_LOOP_H
 
 #include "symex_bmc.h"
+#include <util/ui_message.h>
```
xbauch, line 13: Guess you don't need this one anymore.

thk123: Needed for `ui_message_handlert::uit` declaration

thk123:
@xbauch comments addressed

src/goto-checker/symex_bmc_incremental_one_loop.cpp
```diff
@@ -85,6 +88,14 @@ bool symex_bmc_incremental_one_loopt::should_stop_unwind(
     abort_unwind_decision.is_known(), "unwind decision should be taken by now");
   bool abort = abort_unwind_decision.is_true();
 
+
+  if(output_ui == ui_message_handlert::uit::XML_UI)
+  {
+    xmlt xml("current-unwinding");
+    xml.data = std::to_string(unwind);
+    log.statistics() << xml;
+  }
```
danpoe: Could we also output a `current-unwinding` item for `PLAIN` and `JSON_UI`? So the same information is output independent of the  ui.

danpoe (approved):

xbauch (approved):
Looks good.

src/goto-checker/symex_bmc_incremental_one_loop.cpp
thk123: Done.

src/goto-checker/symex_bmc_incremental_one_loop.cpp
```diff
@@ -131,3 +135,29 @@ bool symex_bmc_incremental_one_loopt::resume(
 
   return should_pause_symex;
 }
+void symex_bmc_incremental_one_loopt::log_unwinding(unsigned unwind)
+{
+  const std::string unwind_num = std::to_string(unwind);
+  switch(output_ui)
+  {
+  case ui_message_handlert::uit::PLAIN:
```
chrisr-diffblue, line 143: I know there are a few other places in the codebase where this sort of switch happens, so I'm not going to block this PR on this issue :-) But it's really ugly in general... I'm wondering if we could come up with an API extension to message handlers. Perhaps some pseudo code like this:
```
/// Individual implementations inside message handlers
keyValueMessagePair(string[] words, string value) {
/// words == ["current","unwinding"]
/// output would be:
/// "Current unwinding: value" for PLAIN
/// xmlt xml("current-unwinding"); xml.data = value; for XML
/// json_objectt json; json["currentUnwinding"] = json_numbert(value); for JSON
}

// then as a client, we could just use:
log.statistics() << output_ui.keyValueMessagePair(["current","unwinding"],unwind) << messaget::eom;
```
Obviously suitable C++ plumbing needed :-) - but hopefully that explains the idea

chrisr-diffblue (approved):
It's a shame we have to do all the message handler faff, but thats an API gap, rather than anything specific to this PR, so ignoring that, I'm happy for this to go in.

src/goto-checker/symex_bmc_incremental_one_loop.cpp
thk123: Yes I started having a go at making just a general way of logging structured data, but didn't want to hold up this PR for t

PR #5238, opened 2020-02-20
File rename

kroening:

### DIFF
--- a/src/util/file_util.cpp
+++ b/src/util/file_util.cpp
@@ -221,3 +221,41 @@ bool file_remove(const std::string &path)
   return unlink(path.c_str()) == 0;
 #endif
 }
+
+void file_rename(const std::string &old_path, const std::string &new_path)
+{
+#ifdef _WIN32
+  if(is_directory(old_path))
+  {
+    // rename() only renames directories, but does not move them.
+    // MoveFile is not atomic.
+    auto MoveFile_result =
+      MoveFileW(widen(old_path).c_str(), widen(new_path).c_str());
+
+    if(MoveFile_result == 0)
+      throw system_exceptiont("MoveFile failed");
+  }
+  else
+  {
+    // C++17 requires this to be atomic, which is delivered by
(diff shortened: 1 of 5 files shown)


src/util/file_util.cpp
```diff
@@ -221,3 +221,41 @@ bool file_remove(const std::string &path)
...
+    // MoveFile is not atomic.
+    auto MoveFile_result =
+      MoveFileW(widen(old_path).c_str(), widen(new_path).c_str());
+
+    if(MoveFile_result == 0)
+      throw system_exceptiont("MoveFile failed");
+  }
+  else
+  {
+    // C++17 requires this to be atomic, which is delivered by
```
smowton, line 240: Requires what to be atomic?

smowton (approved):

src/util/file_util.cpp
chrisr-diffblue: I assume the rename of the file? A lot of unix software depends on rename being atomic. e.g. for creating lock files.

smowton: Sure, I was just perplexed since we're not implementing any particular C++-y thing here. Presumably it requires things of `std::filesystem` implementations, but since we're implementing our own file_rename function it doesn't require anything of anyone

chrisr-diffblue: From the commit message:
```
This avoids semantic mismatch between POSIX rename(...) and the _rename()
function offered in the Win32 API.  The proposed function is meant to offer
the semantics of the corresponding C++17 function.
```
and basically the C++17 definition says "the semantics of posix rename" - so effectively what this is all doing is making sure we have the same rename semantics on unix and windows, with the bonus points that at some point in the future we can move to C++17 and then trivially replace file_rename with std::filesystem::rename without changing semantics.

chrisr-diffblue (approved):
LGTM

src/util/file_util.cpp
smowton: Aha, didn't see that, thanks

PR #5240, opened 2020-02-21
Logging structured data

thk123:
On #5237 it was pointed out it is a bit clunky to have to switch between message handler UIs. This PR attempts to simplify this by allowing sending XML to the message handler, and it will convert it to JSON or plain text if those UIs are in play. 

Thoughts?

### DIFF
--- a/src/goto-checker/symex_bmc_incremental_one_loop.cpp
+++ b/src/goto-checker/symex_bmc_incremental_one_loop.cpp
@@ -137,27 +137,8 @@ bool symex_bmc_incremental_one_loopt::resume(
 }
 void symex_bmc_incremental_one_loopt::log_unwinding(unsigned unwind)
 {
-  const std::string unwind_num = std::to_string(unwind);
-  switch(output_ui)
-  {
-  case ui_message_handlert::uit::PLAIN:
-  {
-    log.statistics() << "Current unwinding: " << unwind_num << messaget::eom;
-    break;
-  }
-  case ui_message_handlert::uit::XML_UI:
-  {
-    xmlt xml("current-unwinding");
-    xml.data = unwind_num;
-    log.statistics() << xml << messaget::eom;
-    break;
-  }
-  case ui_message_handlert::uit::JSON_UI:
-  {
-    json_objectt json;
-    json["currentUnwinding"] = json_numbert(unwind_num);
-    log.statistics() << json << messaget::eom;
-    break;
-  }
-  }
+  structured_datat data{{{labelt({"current", "unwinding"}),
+                          structured_data_entryt::data_node(
+                            json_numbert{std::to_string(unwind)})}}};
+  log.statistics() << data;
 }
--- a/src/util/json.h
+++ b/src/util/json.h
@@ -467,6 +467,27 @@ inline const json_stringt &to_json_string(const jsont &json)
 
 bool operator==(const jsont &left, const jsont &right);
 
+/// Convert the structured_datat into an json object. For example, the
+/// structured data:
+/// structured_datat data{
+///   {{labelt{{"my", "data"}},
+///      structured_data_entryt::entry(
+///        {{labelt{{"my", "number"}},
+///           structured_data_entryt::data_node(json_numbert("10"))},
+///         {labelt{{"my", "string"}},
+///           structured_data_entryt::data_node(json_stringt("hi"))}})}}};
+///
+/// Will produce:
+/// ```json
+/// {
+///   "myData": {
+///      "myNumber": 10
+///      "myString": "hi"
+///    }
+///  }
+/// ```
+/// \param data: The structured data to convert.
+/// \return The json object as specified.
 jsont to_json(const structured_datat &data);
 
 #endif // CPROVER_UTIL_JSON_H
--- a/src/util/json.h
+++ b/src/util/json.h
@@ -18,6 +18,8 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "irep.h"
 #include "range.h"
 
+class structured_datat;
+
 class json_objectt;
 class json_arrayt;
 
@@ -464,5 +466,7 @@ inline const json_stringt &to_json_string(const jsont &json)
 }
 
 bool operator==(const jsont &left, const jsont &right);
+
+jsont to_json(const structured_datat &data);
 
 #endif // CPROVER_UTIL_JSON_H
--- a/src/util/structured_data.cpp
+++ b/src/util/structured_data.cpp
@@ -71,3 +71,161 @@ bool labelt::operator<(const labelt &other) const
 {
   return camel_case() < other.camel_case();
 }
+structured_data_entryt structured_data_entryt::data_node(const jsont &data)
+{
+  // Structured data (e.g. arrays and objects) should use an entry
+  PRECONDITION(!(data.is_array() || data.is_object()));
+  return structured_data_entryt(data);
+}
+structured_data_entryt
+structured_data_entryt::entry(std::map<labelt, structured_data_entryt> children)
+{
+  return structured_data_entryt(children);
+}
+structured_data_entryt::structured_data_entryt(const jsont &data) : data(data)
+{
+}
+structured_data_entryt::structured_data_entryt(
+  std::map<labelt, structured_data_entryt> children)
+  : children(std::move(children))
+{
+}
+bool structured_data_entryt::is_leaf() const
+{
+  return children.empty();
+}
+std::string structured_data_entryt::leaf_data() const
+{
+  return data.value;
+}
+std::map<labelt, structured_data_entryt>
+structured_data_entryt::get_children() const
+{
+  return children;
+}
+jsont structured_data_entryt::leaf_object() const
+{
+  return data;
+}
+structured_datat::structured_datat(
+  std::map<labelt, structured_data_entryt> data)
+  : data(std::move(data))
+{
+}
+
+xmlt xml_node(const std::pair<labelt, structured_data_entryt> &entry)
+{
+  const labelt &label = entry.first;
+  const structured_data_entryt &data = entry.second;
+  xmlt output_data{label.kebab_case()};
+  if(data.is_leaf())
+  {
+    output_data.data = data.leaf_data();
+  }
+  else
+  {
+    const auto &children = data.get_children();
+    output_data.elements =
+      make_range(children).map(xml_node).collect<std::list<xmlt>>();
+  }
+  return output_data;
+}
+
+xmlt structured_datat::to_xml() const
+{
+  if(data.size() == 0)
+    return xmlt{};
+  if(data.size() == 1)
+  {
+    return xml_node(*data.begin());
+  }
+  else
+  {
+    xmlt root{"root"};
+    root.elements = make_range(data).map(xml_node).collect<std::list<xmlt>>();
+    return root;
+  }
+}
+
+jsont json_node(const structured_data_entryt &entry)
+{
+  if(entry.is_leaf())
+    return entry.leaf_object();
+  else
+  {
+    json_objectt result;
+    for(const auto sub_entry : entry.get_children())
+    {
+      result[sub_entry.first.camel_case()] = json_node(sub_entry.second);
+    }
+    return std::move(result);
+  }
+}
+
+jsont structured_datat::to_json() const
+{
+  if(data.size() == 0)
+    return jsont{};
+
+  json_objectt result;
+  for(const auto sub_entry : data)
+  {
+    result[sub_entry.first.camel_case()] = json_node(sub_entry.second);
+  }
+  return std::move(result);
+}
+
+std::vector<std::string>
+pretty_node(const std::pair<labelt, structured_data_entryt> &entry)
+{
+  const labelt &label = entry.first;
+  const structured_data_entryt &data = entry.second;
+  if(data.is_leaf())
+  {
+    std::ostringstream line;
+    line << label.pretty() << ": " << data.leaf_data();
+    return {line.str()};
+  }
+  else
+  {
+    const auto indent = [](const std::string line) { return "\t" + line; };
+
+    const auto &children = data.get_children();
+    std::vector<std::vector<std::string>> lines =
+      make_range(children)
+        .map(pretty_node)
+        .map([&](std::vector<std::string> sub_lines) {
+          return make_range(sub_lines)
+            .map(indent)
+            .collect<std::vector<std::string>>();
+        })
+        .collect<std::vector<std::vector<std::string>>>();
+
+    std::vector<std::string> result;
+    for(const auto &sub_lines : lines)
+    {
+      result.insert(result.end(), sub_lines.begin(), sub_lines.end());
+    }
+    return result;
+  }
+}
+
+std::string structured_datat::to_pretty() const
+{
+  if(data.empty())
+    return "";
+
+  std::vector<std::vector<std::string>> lines =
+    make_range(data)
+      .map(pretty_node)
+      .collect<std::vector<std::vector<std::string>>>();
+  std::vector<std::string> flattened_lines;
+  for(const auto &line_section : lines)
+  {
+    flattened_lines.insert(
+      flattened_lines.end(), line_section.begin(), line_section.end());
+  }
+  std::ostringstream output;
+  join_strings(output, flattened_lines.begin(), flattened_lines.end(), "\n");
+  return output.str();
+}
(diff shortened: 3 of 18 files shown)


smowton (approved):
Perhaps rename `xmlt` to `structured_messaget` or something now it's the all-purpose structured-log-message type

src/util/convert_xml_json.h
```diff
@@ -0,0 +1,13 @@
...
+
+#ifndef CPROVER__CONVERT_XML_JSON_H
+#define CPROVER__CONVERT_XML_JSON_H
+
+#include "json.h"
+#include "xml.h"
+#include <algorithm>
+json_objectt from_xml(xmlt data);
+
+std::string xml_to_pretty(const xmlt &data);
```
chrisr-diffblue: Given that `xml_to_pretty` isn't outputting json, I'd be tempted to call this new module something slightly different, e.g. `xml_conversion.{cpp,h}` or `xml_output_conversion.{cpp,h}` or something similar? Also, given that both these two functions are (at a high level) "reproduce XML in some other output format", why are the names of the two so different, and why does one take a `const xmlt &`, but the other takes a plain `xmlt` ? I'd ideally like to see both of these as `const xmlt &` inputs.

regression/cbmc-incr-oneloop/alarm2/test-json-output.desc
```diff
@@ -4,6 +4,6 @@ main.c
 ^EXIT=10$
 ^SIGNAL=0$
 "messageText": "VERIFICATION FAILED"
-"currentUnwinding": 1
+"current-unwinding": "1"
```
peterschrammel: Everything-is-a-string is problematic.

chrisr-diffblue (changes requested):
Overall I like this a lot - it's a relatively "simple" change to the interfaces and opens the scope for a whole lot of other code to get cleaned up - many thanks :-)

I'd like to see the naming and type signatures cleaned up a bit if possible, just to try to keep things as consistent as possible, so that's the only reason I'm requesting changes on this PR.

For additional bonus karma though, it would be super cool if the xml -> {whatever} conversion could *also* do a bit of name formatting conversion, e.g. converting element names from XML style (e.g. hyphenated words, like `current-unwinding`)to JSON style formatting (e.g. camel-case `currentUnwinding`) when converting to JSON. I appreciate that might be hard/impossible to do robustly, but if it is possible it would be really helpful in keeping the XML/JSON/plaintext output looking stylistically consistent with other parts of code that output XML/JSON/etc. This part is really not a blocking comment though - definitely in the unreasonable wish category :-)

peterschrammel (commented):
This might remove a few cases, but not all. Eg the json api supports streaming of arrays and objects.

src/goto-checker/symex_bmc_incremental_one_loop.cpp
```diff
@@ -137,27 +137,7 @@ bool symex_bmc_incremental_one_loopt::resume(
...
-  {
-    json_objectt json;
-    json["currentUnwinding"] = json_numbert(unwind_num);
-    log.statistics() << json << messaget::eom;
-    break;
-  }
-  }
+  xmlt data{"current-unwinding"};
+  data.data = std::to_string(unwind);
+  log.statistics() << data;
```
danpoe, line 143: It could be a bit surprising that the result here is converted to a different format. Could we introduce a `structured_messaget` type as suggested by @smowton? It could well be implemented via `xmlt` for the time being (i.e., contain a member `xmlt` and just delegate to that). Then we could add a `print(const structured_messaget &msg)` method to `ui_message_handlert` which converts a `structured_messaget` to the appropriate format. The existing `print()` methods could be left unchanged.

danpoe (commented):
Looks like a good step towards cleaning up the logging code. As mentioned by @chrisr-diffblue, it would be nice if names could be output according to xml and json conventions.

thk123:
Based on the broadly positive reception, I have tidied this up. It now uses an intermediate structure (essentially nested key/value pairs). It also addresses:
 - naming (xml now uses kebab case, pretty printing uses regular spacing, json uses camelCase) 
 - numbers as numbers not strings for json 

Hope this is still of use, ready for re-review @chrisr-diffblue @peterschrammel

src/util/structured_data.h
```diff
@@ -0,0 +1,80 @@
...
+  std::string pretty() const;
+
+  bool operator<(const labelt &other) const
+  {
+    return camel_case() < other.camel_case();
+  }
+
+  bool operator==(const labelt &other) const
+  {
+    return camel_case() == camel_case();
```
thk123: :no_entry_sign: This is dumb and should have been tested :man_facepalming:

src/util/convert_xml_json.h
thk123: Resolved by removing this file

regression/cbmc-incr-oneloop/alarm2/test-json-output.desc
thk123: Resolved: data is now a `jsont` so in particular can be a `json_numbert`

src/util/structured_data.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+#include "structured_data.h"
+#include "range.h"
+#include "string_utils.h"
+#include <algorithm>
+
+std::string capitalize(const std::string &str)
+{
+  std::string capitalized = str;
+  capitalized[0] = toupper(capitalized[0]);
```
thk123: This errors on the empty string

thk123: This is a little fun btw - accessing the 0th element is not undefined behaviour, but modifying it is!

src/util/structured_data.h
```diff
@@ -0,0 +1,99 @@
...
+///    My string: hi
+/// ```
+class structured_datat
+{
+public:
+  explicit structured_datat(std::map<labelt, structured_data_entryt> data);
+
+  xmlt to_xml() const;
+  jsont to_json() const;
+  std::string to_pretty() const;
```
chrisr-diffblue: Is it more consistent to call this `to_plain` ?

src/util/structured_data.cpp
```diff
@@ -0,0 +1,218 @@
...
+  else
+  {
+    const auto &children = data.get_children();
+    output_data.elements =
+      make_range(children).map(xml_node).collect<std::list<xmlt>>();
+  }
+  return output_data;
+}
+
+xmlt structured_datat::to_xml() const
```
chrisr-diffblue: One design question I have (which I honestly don't know what the right answer is :-) ) is - who's responsibility is it to do the conversion from `structured_data_entryt` to {xml,json,plain} ? It feels like the knowledge of what (say) xml needs to look like (e.g. kebab_case) should be more closely located with the various XML utility functions. So perhaps the design should be that structure_data.{h,cpp} has a `to_xm()` `to_json()` and `to_plain()`, and those basically just be simple wrappers around other functions like `xml_node(...)` in the XML utility modules? Then all the knowledge of what XML should look like would be locate in the XML utility modules. Of course that means xml/json/etc modules need to know a little bit about structured_data... so it's not a clear choice either way.

chrisr-diffblue (commented):
Looking good, just a few queries 👍

src/util/structured_data.h
```diff
@@ -0,0 +1,99 @@
...
+
+private:
+  std::vector<std::string> components;
+};
+
+struct structured_data_entryt
+{
+  static structured_data_entryt data_node(const jsont &data);
+  static structured_data_entryt
+  entry(std::map<labelt, structured_data_entryt> children);
```
danpoe, line 36: I think `structured_data_entryt` unnecessarily exposes the fact that `jsont` and `std::map` are used to store the data and children. Could we instead have constructors which take string and int, and a method to add children? :bee:

src/util/structured_data.h
```diff
@@ -0,0 +1,99 @@
...
+/// Plain:
+/// ```
+/// My data:
+///    My number: 10
+///    My string: hi
+/// ```
+class structured_datat
+{
+public:
+  explicit structured_datat(std::map<labelt, structured_data_entryt> data);
```
danpoe, line 76: :bee:

src/util/structured_data.cpp
```diff
@@ -0,0 +1,218 @@
...
+bool structured_data_entryt::is_leaf() const
+{
+  return children.empty();
+}
+std::string structured_data_entryt::leaf_data() const
+{
+  return data.value;
+}
+std::map<labelt, structured_data_entryt>
+structured_data_entryt::get_children() const
```
danpoe: Could return a const reference

src/util/structured_data.cpp
```diff
@@ -0,0 +1,218 @@
...
+        .map(capitalize)
+        .concat(make_range(std::next(components.begin()), components.end()));
+
+    join_strings(output, range.begin(), range.end(), ' ');
+  }
+  return output.str();
+}
+bool labelt::operator<(const labelt &other) const
+{
+  return camel_case() < other.camel_case();
```
danpoe: Could directly compare vectors

src/util/structured_data.cpp
```diff
@@ -0,0 +1,218 @@
...
+    std::string lower_case = s;
+    std::for_each(
+      lower_case.begin(), lower_case.end(), [](char &c) { c = ::tolower(c); });
+    return lower_case;
+  };
+  this->components = make_range(components)
+                       .map(to_lower_string)
+                       .collect<std::vector<std::string>>();
+}
+std::string labelt::camel_case() const
```
danpoe, line 24: Newlines between methods (here and at other places)

src/util/structured_data.cpp
```diff
@@ -0,0 +1,218 @@
...
+  auto to_lower_string = [](const std::string &s) -> std::string {
+    std::string lower_case = s;
+    std::for_each(
+      lower_case.begin(), lower_case.end(), [](char &c) { c = ::tolower(c); });
+    return lower_case;
+  };
+  this->components = make_range(components)
+                       .map(to_lower_string)
+                       .collect<std::vector<std::string>>();
+}
```
danpoe, line 22: Could we make it a precondition that the inputs need to be in all lowercase? That would ensure better consistency in the uses of the API. Additionally we could add a precondition checking that there is at least one component (then we can get rid of the `!components.empty()` checks), and a precondition checking that each component is non-empty.

src/goto-checker/symex_bmc_incremental_one_loop.cpp
```diff
@@ -137,27 +137,8 @@ bool symex_bmc_incremental_one_loopt::resume(
...
-  {
-    json_objectt json;
-    json["currentUnwinding"] = json_numbert(unwind_num);
-    log.statistics() << json << messaget::eom;
-    break;
-  }
-  }
+  structured_datat data{{{labelt({"current", "unwinding"}),
+                          structured_data_entryt::data_node(
+                            json_numbert{std::to_string(unwind)})}}};
```
danpoe, line 142: That seems a bit too verbose for this common use case. Could we have some convenience methods for the most common cases (keys with 1-2 components, values of type string or int)?

danpoe (commented):
Looks very good, some additional suggestions

src/util/structured_data.cpp
thk123: Any reason to prefer this? Here I rely on the library implemented alphabetical sort of strings, which seems simpler than writing a sort for vector

src/util/structured_data.h
thk123: Not sure? `.pretty` is the terminology used for converting irep to a plain text representation? Happy to change if you are sure.

src/util/structured_data.cpp
danpoe: I think it's already implemented in the standard library for vectors

src/util/structured_data.h
thk123: Don't agree with this - this allows you to easily create a constant structured_data using the existing type initialization C++ provides. Further, I'd either need to recreate constructors for all the json types, or limit what kind of json data you can output.

src/util/structured_data.cpp
chrisr-diffblue: One possible thing to be aware of, though not entirely sure if it applies here, and if so, whether it even matters or not... is that a lot of standard library sort routines on strings are actually locale sensitive... meaning you may/will get different sort orders on different platforms, or different users on the same platform if they change locale. If you want this to be at all stable in its sort order, you need to be aware of that. Locale sensitive sorting has caused issues with CI in the past (e.g. output from CBMC in different orders between CI and local runs)

src/util/structured_data.h
chrisr-diffblue: Honestly don't mind :-) Just asking the question, and your answer seems equally valid.

src/goto-checker/symex_bmc_incremental_one_loop.cpp
thk123: This seems like a nice to have - happy for you to do in a follow up PR but won't as part of this PR.

thk123:
@chrisr-diffblue @danpoe comments [addressed](https://github.com/diffblue/cbmc/compare/4ca6b5769f769ba1f32d154fc30da491887714f4..9652a79bfead4651263e7ac59e608df9babce580)

src/util/structured_data.h
chrisr-diffblue: Third option: a very small handful of specialized constructors/helper functions for the "common" cases, but leaving the full exposure of jsont/std::map for any of the less common cases?

src/util/json.h
```diff
@@ -465,4 +467,18 @@ inline const json_stringt &to_json_string(const jsont &json)
 
 bool operator==(const jsont &left, const jsont &right);
 
+/// Convert the structured_data into an json object. For the example structured
```
chrisr-diffblue: `structured_data` -> `structured_datat` - also there's no example structured data here now, as this comment has moved from it's original location (sorry...).

src/util/xml.h
```diff
@@ -135,4 +137,20 @@ inline std::ostream &operator<<(
   return out;
 }
 
+bool operator==(const xmlt &a, const xmlt &b);
+bool operator!=(const xmlt &a, const xmlt &b);
+
+/// Convert the structured_data into an xml object. For the example structured
```
chrisr-diffblue: As with the json version, `structured_data` -> `structured_datat` and the example structure data will likely need copying into this comment as well. (sorry again...)

src/util/structured_data.cpp
```diff
@@ -0,0 +1,167 @@
+// Copyright 2016-2020 Diffblue Limited. All Rights Reserved.
```
chrisr-diffblue: Please use the proper CBMC comment header, and no Diffblue copyright.

src/util/structured_data.h
```diff
@@ -0,0 +1,86 @@
+// Copyright 2016-2020 Diffblue Limited. All Rights Reserved.
```
chrisr-diffblue: Please use the proper CBMC comment header, and no Diffblue copyright.

unit/util/string_utils/capitalize.cpp
```diff
@@ -0,0 +1,14 @@
+// Copyright 2016-2020 Diffblue Limited. All Rights Reserved.
```
chrisr-diffblue: Shouldn't be a Diffblue copyright, just authorship.

unit/util/structured_data.cpp
```diff
@@ -0,0 +1,179 @@
+// Copyright 2016-2020 Diffblue Limited. All Rights Reserved.
```
chrisr-diffblue: Shouldn't be a Diffblue copyright, just authorship.

unit/util/xml.cpp
```diff
@@ -0,0 +1,93 @@
+// Copyright 2016-2020 Diffblue Limited. All Rights Reserved.
```
chrisr-diffblue: Shouldn't be a Diffblue copyright, just authorship.

chrisr-diffblue (approved):
Looks great! Very many thanks for persevering with this! It sucks to volunteer a solution and then have lots of nits picked as a thanks, but you've stuck with it. I'm happy for this to go in if you could just fix up the comment headers/copyrights and update a couple of comments.

danpoe (approved):

thk123:
@chrisr-diffblue done if you want to [re-check](https://github.com/diffblue/cbmc/compare/077e545107cbcd6700047b327e52d23ed45a01f8..95d0c362785650e5cb5c0db0c5b49494184c0396) otherwise will merge on green CI

chrisr-diffblue:
Thanks @thk123 - could I be a real pain and ask for proper Module titles in the comment headers, rather than blanks? a `git grep Module:` in your CBMC checkout should give your a good flavour of what's been done else where. If thats too much hassle after all this, then feel free to merge.

thk123:
apt-get failure - restarting cbmc-linux-clang

PR #5244, opened 2020-02-25
Fix over-writing library prologue

xbauch:
when linking the cprover library. The default mode for this constructor of
stringstream is to over-write the string used in construction, which presumably
was not the intention here.

Note: until now there wasn't anything too interesting in the prologue except for
the string-abstraction define, which was ignored. But we might want more defines
in the future.

### DIFF
--- a/src/ansi-c/cprover_library.cpp
+++ b/src/ansi-c/cprover_library.cpp
@@ -45,7 +45,9 @@ std::string get_cprover_library_text(
   const struct cprover_library_entryt cprover_library[],
   const std::string &prologue)
 {
-  std::ostringstream library_text(prologue);
+  // the default mode is ios_base::out which means subsequent write to the
+  // stream will overwrite the original content
+  std::ostringstream library_text(prologue, std::ios_base::ate);
 
   std::size_t count=0;
 


src/ansi-c/cprover_library.cpp
```diff
@@ -45,7 +45,9 @@ std::string get_cprover_library_text(
   const struct cprover_library_entryt cprover_library[],
   const std::string &prologue)
 {
-  std::ostringstream library_text(prologue);
+  // the default mode is ios_base::out which means subsequent write to the
+  // stream will overwrite the original content
+  std::ostringstream library_text(prologue, std::ios_base::ate);
```
hannes-steffenhagen-diffblue:
```
#include <iostream>
#include <sstream>
int main(void) {
  std::ostringstream send_help{"SOS", std::ios::ate};
  send_help << "WAT";
  std::cout << send_help.str() << '\n';
}
```

hannes-steffenhagen-diffblue, line 50:
```
#include <iostream>
#include <sstream>
int main(void) {
  std::ostringstream send_help{"SOS"};
  send_help << "WAT";
  std::cout << send_help.str() << '\n';
}
```

[Try it online!][TIO-k724hxz2]

[C++ (gcc)]: https://gcc.gnu.org/
[TIO-k724hxz2]: https://tio.run/##Sy4o0E1PTv7/XzkzLzmnNCVVwSYzv7ikKDUx144LIVYME8rMK1HITczM0yjLz0zRVKjmUlAoLkmxsgLpycxLhyhTKE7NS4nPSM0pqFYK9g9WqrUGKYOJKdjYKCiFO4YoWcM0J@eXloBE4Ur0gOZoaIKE1GPy1K25arn@/wcA "C++ (gcc) – Try It Online"

hannes-steffenhagen-diffblue (approved):

danpoe (approved):

src/ansi-c/cprover_library.cpp
thk123: WAT is the point of this constructor... 
A potentially more obvious fix might be 
```
std::ostringstream library_text;
library_text << prologue
```

hannes-steffenhagen-diffblue: @thk123 In a slightly less humorous note, I _can_ see uses for this (for example, if you need to guarantee some minimum width for the output so you add in a bunch of filler characters to override first), the “wtf” here is that this is the default, which I think few sane people would expect.

xbauch:
```
std::ostringstream library_text(prologue);
library_text.seekp(0, library_text.end);
```
would also work.

chrisr-diffblue: I'd personally lean either towards the solution @thk123 suggested, or to the original solution @xbauch had in the PR. I think the alternative using `seekp` just looks even more odd :-) Though I appreciate the oddness comes from the C++ spec here.

hannes-steffenhagen-diffblue:
Turning it off

hannes-steffenhagen-diffblue:
And turning it on again

thk123:
Superseded by #5285

PR #5246, opened 2020-02-26
String solver: support bounded-length intermediate strings

smowton:
This adds support for the string solver to use constant-sized arrays to store
variable-length strings (which may be cheaper since smaller formulas are generally needed
to represent them and their constraints to the solver)

This is off-by-default for now since it has not been heavily studied: I expect constant-sized
arrays would be expensive in the case where strings are numerous and sparsely constrained,
but cheaper when there are many index expressions leading to an explosion of Ackermann
constraints.

I originally wrote this to try to avoid the huge explosion of Ackermann constraints that arose when the string solver created a large index set. That is successfully achieved, as constant-sized arrays compile with cost proportional to the number of accesses rather than the quadratic complexity of the unbounded arrays it usually uses; however in practice any formula that featured a large enough index set that the Ackermann constraints were prohibitively numerous also generates a huge array of instances of universal string constraints, so this does not solve the problem by itself.

I successfully prototyped a followup that compresses the index sets by only instantiating universal constraints at distinct indices for a particular model, having observed hundreds of constraints that usually evaluated to address the same concrete indices in any viable model. However this requires more work, so I'm submitting this groundwork in the meantime.

### DIFF
--- a/src/solvers/prop/prop_conv_solver.h
+++ b/src/solvers/prop/prop_conv_solver.h
@@ -126,7 +126,7 @@ protected:
 
   propt &prop;
 
-  messaget log;
+  mutable messaget log;
 
   static const char *context_prefix;
 
--- a/src/solvers/strings/max_concrete_char_array.cpp
+++ b/src/solvers/strings/max_concrete_char_array.cpp
@@ -1,0 +1,34 @@
+/*******************************************************************\
+
+Module: Size limit to choose between variable- and fixed-size char arrays
+
+Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+/// \file
+///  Size limit to choose between variable- and fixed-size char arrays
+
+#include "max_concrete_char_array.h"
+
+#include <util/arith_tools.h>
+#include <util/std_expr.h>
+
+/// Pure guess; TODO study what this should be.
+/// Char arrays with max length <= this will be allocated with a fixed size
+/// (assuming the perhaps-wasted space is worth the less-complex formula
+/// generated for fixed-size arrays), while larger arrays will be allocated
+/// with indefinite size assuming generally the wasted space is not worth it.
+const std::size_t max_concrete_char_array_length = 256;
+
+array_typet make_char_array_type(
+  const typet &element_type,
+  const typet &index_type,
+  optionalt<std::size_t> max_length)
+{
+  exprt array_size = max_length && *max_length <= max_concrete_char_array_length
+                       ? (exprt)from_integer(*max_length, index_type)
+                       : (exprt)infinity_exprt{index_type};
+
+  return array_typet{element_type, array_size};
+}
--- a/src/solvers/strings/string_refinement.cpp
+++ b/src/solvers/strings/string_refinement.cpp
@@ -157,7 +157,10 @@ string_refinementt::string_refinementt(const infot &info, bool)
   : supert(info),
     config_(info),
     loop_bound_(info.refinement_bound),
-    generator(*info.ns)
+    generator(
+      *info.ns,
+      info.maximum_intermediate_string_length,
+      info.use_fixed_size_arrays_for_bounded_strings)
 {
 }
 
@@ -704,7 +707,6 @@ decision_proceduret::resultt string_refinementt::dec_solve()
 
   log.debug() << "dec_solve: add constraints" << messaget::eom;
   merge(constraints, dependencies.add_constraints(generator));
-
 #ifdef DEBUG
   output_equations(log.debug(), equations);
 #endif
@@ -768,12 +770,21 @@ decision_proceduret::resultt string_refinementt::dec_solve()
     add_lemma(substitute_array_access(lemma, generator.fresh_symbol, true));
   }
 
-  // All generated strings should have non-negative length
+  // All generated strings should have non-negative length, and perhaps a
+  // maximum length
   for(const auto &pair : generator.array_pool.created_strings())
   {
     exprt length = generator.array_pool.get_or_create_length(pair.first);
     add_lemma(
       binary_relation_exprt{length, ID_ge, from_integer(0, length.type())});
+    if(config_.maximum_intermediate_string_length)
+    {
+      add_lemma(binary_relation_exprt{
+        length,
+        ID_le,
+        from_integer(
+          *config_.maximum_intermediate_string_length, length.type())});
+    }
   }
 
   // Initial try without index set
@@ -1235,12 +1246,16 @@ static void substitute_array_access_in_place(
   {
     if(const auto index_expr = expr_try_dynamic_cast<index_exprt>(*it))
     {
-      optionalt<exprt> result =
-        substitute_array_access(*index_expr, symbol_generator, left_propagate);
+      if(!can_cast_expr<constant_exprt>(
+           to_array_type(index_expr->array().type()).size()))
+      {
+        optionalt<exprt> result = substitute_array_access(
+          *index_expr, symbol_generator, left_propagate);
 
-      // Only perform a write when we have something changed.
-      if(result)
-        it.mutate() = *result;
+        // Only perform a write when we have something changed.
+        if(result)
+          it.mutate() = *result;
+      }
     }
   }
 }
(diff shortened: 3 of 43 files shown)


src/solvers/prop/prop_conv_solver.h
```diff
@@ -126,7 +126,7 @@ class prop_conv_solvert : public conflict_providert,
 
   propt &prop;
 
-  messaget log;
+  mutable messaget log;
```
romainbrenguier, line 129: How is this related to the changes?

src/solvers/strings/max_concrete_char_array.cpp
```diff
@@ -0,0 +1,34 @@
...
+/// Char arrays with max length <= this will be allocated with a fixed size
+/// (assuming the perhaps-wasted space is worth the less-complex formula
+/// generated for fixed-size arrays), while larger arrays will be allocated
+/// with indefinite size assuming generally the wasted space is not worth it.
+const std::size_t max_concrete_char_array_length = 256;
+
+array_typet make_char_array_type(
+  const typet &element_type,
+  const typet &index_type,
+  optionalt<std::size_t> max_length)
```
romainbrenguier, line 27: :pick: const

src/solvers/strings/max_concrete_char_array.cpp
```diff
@@ -0,0 +1,34 @@
...
+{
+  exprt array_size = max_length && *max_length <= max_concrete_char_array_length
```
romainbrenguier, line 29: :pick: const

src/solvers/strings/max_concrete_char_array.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include <util/std_expr.h>
+
+/// Pure guess; TODO study what this should be.
+/// Char arrays with max length <= this will be allocated with a fixed size
+/// (assuming the perhaps-wasted space is worth the less-complex formula
+/// generated for fixed-size arrays), while larger arrays will be allocated
+/// with indefinite size assuming generally the wasted space is not worth it.
+const std::size_t max_concrete_char_array_length = 256;
+
+array_typet make_char_array_type(
```
romainbrenguier, line 24: Please document

src/solvers/strings/max_concrete_char_array.cpp
```diff
@@ -0,0 +1,34 @@
...
+Module: Size limit to choose between variable- and fixed-size char arrays
+
+Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+/// \file
+///  Size limit to choose between variable- and fixed-size char arrays
+
+#include "max_concrete_char_array.h"
```
romainbrenguier, line 12: Should the file be rather be named max_concrete_char_array_length?

src/solvers/strings/string_refinement.cpp
```diff
@@ -1235,12 +1246,16 @@ static void substitute_array_access_in_place(
   {
     if(const auto index_expr = expr_try_dynamic_cast<index_exprt>(*it))
     {
-      optionalt<exprt> result =
-        substitute_array_access(*index_expr, symbol_generator, left_propagate);
+      if(!can_cast_expr<constant_exprt>(
+           to_array_type(index_expr->array().type()).size()))
```
romainbrenguier, line 1250: I don't understand why we don't do the substitution for constant size arrays, and this should be documented.

src/solvers/strings/string_refinement.cpp
```diff
@@ -1804,6 +1819,140 @@ exprt substitute_array_lists(exprt expr, size_t string_max_length)
...
 }
 
+/// Given an index set, return a sorted vector of the concrete indices
+/// defined in the current model.
+/// \param index_set index set
+/// \return concrete indices
+std::vector<size_t> string_refinementt::get_model_defined_indices(
+  const std::set<exprt> &index_set) const
+{
+  std::vector<size_t> defined_indices;
```
romainbrenguier, line 1829: I'm not sure about the choice of size_t. mp_integer seems better. :melon:

src/solvers/strings/string_refinement.cpp
```diff
@@ -1804,6 +1819,140 @@ exprt substitute_array_lists(exprt expr, size_t string_max_length)
...
+  for(const auto &index_expr : index_set)
+  {
+    if(auto index = numeric_cast<size_t>(simplify_expr(get(index_expr), ns)))
```
romainbrenguier, line 1832: I'm a bit worried that this will silently ignore expression that cannot be simplified to a constant

src/solvers/strings/string_refinement.cpp
```diff
@@ -1804,6 +1819,140 @@ exprt substitute_array_lists(exprt expr, size_t string_max_length)
...
+
+/// Replace the given array with one where any index not defined in the given
+/// defined indices inherits its value from its right-hand defined neighbouring
+// index.
+/// For example, if the input array represents the string "abcde" and the
+/// defined indices are 1, 3 and 4, the result is "bbdde".
+/// \param array array to restrict
+/// \param defined_indices sorted index-set to restrict by
+/// \return restricted array
+static std::pair<interval_sparse_arrayt, bool> restrict_sparse_array_to_indices(
```
romainbrenguier, line 1852: :bulb: this could be a method of the `interval_sparse_arrayt` class

src/solvers/strings/string_refinement.cpp
```diff
@@ -1804,6 +1819,140 @@ exprt substitute_array_lists(exprt expr, size_t string_max_length)
...
+  interval_sparse_arrayt array,
+  const std::vector<std::size_t> &defined_indices)
```
romainbrenguier, line 1854: :melon:

src/solvers/strings/string_refinement.cpp
```diff
@@ -1804,6 +1819,140 @@ exprt substitute_array_lists(exprt expr, size_t string_max_length)
...
+      find_defined_index == defined_indices.end() ||
+      *find_defined_index != intervals_it->first)
+    {
+      any_changes = true;
+
+      // If there is no previous defined index, or the previous defined index
+      // already has an interval associated (necessarily the previous interval),
+      // just drop this interval and therefore effectively extend the one to
+      // our right to cover the whole intervening space:
+      bool should_drop_interval =
```
romainbrenguier, line 1890: const

src/solvers/strings/string_refinement.cpp
```diff
@@ -1804,6 +1819,140 @@ exprt substitute_array_lists(exprt expr, size_t string_max_length)
...
+        // these intervals are of the form "VALUE until-including INDEX".
+        exprt existing_value = intervals_it->second;
+        intervals_it = array.erase(intervals_it);
+        array.insert({*std::prev(find_defined_index), existing_value});
+      }
+    }
+    else
+    {
+      // This interval already extends to a defined index; leave it alone.
+      intervals_it++;
```
romainbrenguier, line 1912: :pick: inverting the if and else branches could make it easier to read since the else is much shorter and when reading this line we forgot what the if condition was.

src/solvers/strings/string_refinement.cpp
```diff
@@ -2034,3 +2210,57 @@ static bool is_valid_string_constraint(
 
   return true;
 }
+
+static bool check_string_limit(
```
romainbrenguier, line 2214: needs documentation

src/solvers/strings/string_refinement.h
```diff
@@ -50,7 +54,24 @@ Author: Alberto Griggio, alberto.griggio@gmail.com
   " --max-nondet-string-length n bound the length of nondet (e.g. input) strings.\n" /* NOLINT(*) */ \
   "                              Default is " + std::to_string(MAX_CONCRETE_STRING_SIZE - 1) + "; note that\n" /* NOLINT(*) */ \
   "                              setting the value higher than this does not work\n" /* NOLINT(*) */ \
-  "                              with --trace or --validate-trace.\n" /* NOLINT(*) */
+  "                              with --trace or --validate-trace.\n" /* NOLINT(*) */ \
+  " --max-intermediate-string-length n bound the length of intermediate\n"  /* NOLINT(*) */ \
```
romainbrenguier, line 58: :pick: new line after n would be easier to read

src/solvers/strings/string_refinement_util.h
```diff
@@ -133,6 +133,28 @@ class interval_sparse_arrayt final : public sparse_arrayt
   /// Complexity is logarithmic in the number of entries.
   exprt at(std::size_t index) const;
 
+  /// Iterators over interval boundaries
+  typedef std::map<std::size_t, exprt>::const_iterator const_iterator; // NOLINT
+  const_iterator begin() const
```
romainbrenguier, line 138: I think these iterators can be confusing since one could think we are iterating over all indices rather than intervals.
By making `restrict_sparse_array_to_indices` a method of this class (or sparse_array), I think you wouldn't have to define them.

Issue #5194, opened 2019-11-20
__CPROVER_w_ok has weird behaviour


danielsn:
Since `y > x`, it should be true that `!__CPROVER_w_ok(a, y)` is always false.  However, if you assert this fact, you get an assertion violation.

```
#include<assert.h>
#include<stdlib.h>
#include<stdint.h>

int main() {
  size_t x;
  size_t y;
  uint8_t* a;

  __CPROVER_assume(x > 0);
  __CPROVER_assume(y > x);

  a = malloc(sizeof(*(a)) * (x));

  assert(!__CPROVER_w_ok(a, y));
}
```

```
~/temp $ cbmc *.c --trace --pointer-check --bounds-check
CBMC version 5.12 (cbmc-5.12-d8598f8-1008-gfab409d18) 64-bit x86_64 macos
Parsing test.c
Converting
Type-checking test
Generating GOTO Program
Adding CPROVER library (x86_64)
Removal of function pointers and virtual functions
Generic Property Instrumentation
Running with 8 object bits, 56 offset bits (default)
Starting Bounded Model Checking
size of program expression: 70 steps
simple slicing removed 5 assignments
Generated 1 VCC(s), 1 remaining after simplification
Passing problem to propositional reduction
converting SSA
Running propositional reduction
Post-processing
Solving with MiniSAT 2.2.1 with simplifier
874 variables, 2035 clauses
SAT checker: instance is SATISFIABLE
Runtime decision procedure: 0.00674373s
Building error trace

** Results:
test.c function main
[main.assertion.1] line 15 assertion !__CPROVER_w_ok(a, y): FAILURE

Trace for main.assertion.1:

State 21 file test.c function main line 6 thread 0
----------------------------------------------------
  x=35235878141952ul (00000000 00000000 00100000 00001011 11111110 00000000 00000000 00000000)

State 22 file test.c function main line 7 thread 0
----------------------------------------------------
  y=36028814198833152ul (00000000 10000000 00000000 00000100 00000000 00000000 00000000 00000000)

State 23 file test.c function main line 8 thread 0
----------------------------------------------------
  a=((uint8_t *)NULL) (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

Assumption:
  file test.c line 10 function main
  x > (unsigned long int)0

Assumption:
  file test.c line 11 function main
  y > x

State 29 file test.c function main line 13 thread 0
----------------------------------------------------
[...]

PR #5247, opened 2020-02-26
Use OBJECT_SIZE(p) for dynamic memory bounds checking

danpoe:
This changes the dynamic memory bounds checking predicate in `address_check()`.
It now uses `OBJECT_SIZE()` instead of the variables `__CPROVER_malloc_size` and
`__CPROVER_malloc_object` that are set in the `malloc()` model. This also means that
the predicates `__CPROVER_r_ok()`/`__CPROVER_w_ok()` will be more precise. In the
previous approach, the following assertion could not be proven for example:

`void main() { char *c = malloc(1); assert(!__CPROVER_r_ok(c, 2)); }`

### DIFF
--- a/regression/cbmc/r_w_ok5/main.c
+++ b/regression/cbmc/r_w_ok5/main.c
@@ -1,0 +1,17 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char c[2];
+  assert(__CPROVER_r_ok(c, 2));
+  assert(!__CPROVER_r_ok(c, 2));
+  assert(__CPROVER_r_ok(c, 3));
+  assert(!__CPROVER_r_ok(c, 3));
+
+  char *p = malloc(2);
+  assert(__CPROVER_r_ok(c, 2));
+  assert(!__CPROVER_r_ok(c, 2));
+  assert(__CPROVER_r_ok(p, 3));
+  assert(!__CPROVER_r_ok(p, 3));
+}
--- a/regression/cbmc/r_w_ok5/test.desc
+++ b/regression/cbmc/r_w_ok5/test.desc
@@ -1,0 +1,15 @@
+CORE
+main.c
+
+\[main.assertion.1\] .*: SUCCESS
+\[main.assertion.2\] .*: FAILURE
+\[main.assertion.3\] .*: FAILURE
+\[main.assertion.4\] .*: SUCCESS
+\[main.assertion.5\] .*: SUCCESS
+\[main.assertion.6\] .*: FAILURE
+\[main.assertion.7\] .*: FAILURE
+\[main.assertion.8\] .*: SUCCESS
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
--- a/regression/cbmc/r_w_ok6/main.c
+++ b/regression/cbmc/r_w_ok6/main.c
@@ -1,0 +1,26 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char *p;
+  int choice;
+
+  if(choice)
+  {
+    p = malloc(2);
+  }
+  else
+  {
+    p = malloc(3);
+  }
+
+  assert(__CPROVER_r_ok(p, 2));
+  assert(!__CPROVER_r_ok(p, 2));
+
+  assert(__CPROVER_r_ok(p, 3));
+  assert(!__CPROVER_r_ok(p, 3));
+
+  assert(__CPROVER_r_ok(p, 4));
+  assert(!__CPROVER_r_ok(p, 4));
+}
--- a/regression/cbmc/r_w_ok7/main.c
+++ b/regression/cbmc/r_w_ok7/main.c
@@ -1,0 +1,18 @@
+#include <assert.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+int main()
+{
+  size_t x;
+  size_t y;
+  uint8_t *a;
+
+  __CPROVER_assume(x > 0);
+  __CPROVER_assume(y > x);
+
+  a = malloc(sizeof(uint8_t) * x);
+
+  assert(__CPROVER_w_ok(a, x));
+  assert(!__CPROVER_w_ok(a, y));
+}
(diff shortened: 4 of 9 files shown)


regression/cbmc/r_w_ok5/main.c
```diff
@@ -0,0 +1,13 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char c[2];
+  assert(__CPROVER_r_ok(c, 3));
+  assert(!__CPROVER_r_ok(c, 3));
```
thk123: Might be worth checking for valid reads too

```suggestion
  char c[2];
  assert(__CPROVER_r_ok(c, 3));
  assert(!__CPROVER_r_ok(c, 3));
  assert(__CPROVER_r_ok(c, 2)); // Success
  assert(!__CPROVER_r_ok(c, 2)); // Success
```

regression/cbmc/r_w_ok7/main.c
```diff
@@ -0,0 +1,17 @@
...
+int main()
+{
+  size_t x;
+  size_t y;
+  uint8_t *a;
+
+  __CPROVER_assume(x > 0);
+  __CPROVER_assume(y > x);
+
+  a = malloc(sizeof(*(a)) * (x));
```
thk123: :pick: I think this would be easier to read as `sizeof(uint8_t)`

regression/cbmc/r_w_ok7/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+
+VERIFICATION SUCCESSFUL
```
thk123, line 6: :pick: add the `main.assertion` stuff

regression/cbmc/r_w_ok7/main.c
```diff
@@ -0,0 +1,17 @@
...
+
+  assert(!__CPROVER_w_ok(a, y));
```
thk123, line 17: :no_entry_sign: I think you should add 
```suggestion
  assert(!__CPROVER_w_ok(a, y));
  assert(__CPROVER_w_ok(a, x));
```
to check cprover_w_ok with a dynamic offset works

regression/cbmc/r_w_ok8/main.c
```diff
@@ -0,0 +1,9 @@
+#include <assert.h>
+#include <stdlib.h>
+
+int main()
+{
+  int *x = malloc(2);
+  int *y = malloc(2);
+  assert(__CPROVER_r_ok(x, 3) == __CPROVER_r_ok(y, 3));
```
thk123, line 9: :question: Not sure I understand the purpose of the test

regression/cbmc/r_w_ok8/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+
+VERIFICATION SUCCESSFUL
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
```
thk123, line 10: :no_entry_sign: Please add explanations to each of the tests for its purpose

regression/cbmc/r_w_ok6/main.c
```diff
@@ -0,0 +1,26 @@
...
+  }
+  else
+  {
+    p = malloc(3);
+  }
+
+  assert(__CPROVER_r_ok(p, 2));
+  assert(!__CPROVER_r_ok(p, 2));
+
+  assert(__CPROVER_r_ok(p, 3));
```
thk123, line 21: Nice - not obvious so good to check it works

thk123 (commented):
I can't say I really understand much about the code change, but the tests look robust. I assume it is already tested, but it is worth having a test using `--pointer-check` and so some dynamically allocated memory?

regression/cbmc/r_w_ok5/main.c
danpoe: Added (`assert(!__CPROVER_r_ok(c, 2))` does fail though)

regression/cbmc/r_w_ok8/main.c
danpoe: This was reported in issue https://github.com/diffblue/cbmc/issues/5194. It checks that two uses of `__CPROVER_r_ok()` can return false simultaneously. Previously only one call could return false at a time.

thk123: :+1: Please add to the desc file :slightly_smiling_face:

regression/cbmc/r_w_ok5/main.c
thk123: :man_facepalming:

danpoe:
@thk123 Ok, addressed all the comments.

regression/cbmc/r_w_ok8/test.desc
```diff
@@ -0,0 +1,16 @@
...
+--
+This test checks that two usages of the primitive __CPROVER_r_ok() can be false
+simultaneously in the encoding of the program. Previously, at most one call
+could be false at a time. This was imprecise, however, it was sufficient to
+guarantee soundness when the __CPROVER_r_ok() primitive was used directly in an
+assertion (i.e., as assert(__CPROVER_r_ok(p, size)).
```
thk123: You could link to the relevant issue so people can find more context

regression/cbmc/r_w_ok5/test.desc
```diff
@@ -0,0 +1,15 @@
...
+\[main.assertion.3\] .*: FAILURE
+\[main.assertion.4\] .*: SUCCESS
+\[main.assertion.5\] .*: SUCCESS
+\[main.assertion.6\] .*: FAILURE
+\[main.assertion.7\] .*: FAILURE
+\[main.assertion.8\] .*: SUCCESS
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
```
thk123, line 15: :pick: Still missing an explanation

thk123 (approved):
Thanks - looks great.

regression/cbmc/r_w_ok5/test.desc
danpoe: I think this one is simple enough to not need an explanation

PR #2256, opened 2018-05-30
Use !shouldfail rather than !mayfail on unit tests

thk123:
May fail runs the test and if it fails prints the output. Should fail runs the test and only prints the output if it actually passes. This makes it easier when looking at build output to determine which unit test has actually failed.

PR #5248, opened 2020-02-27
Don't run failed unit tests by default

thk123:

This replaces #2256. By using the tag `[.]`, the tests are not run by default, leading to no useless output when another unit test fails. By changing `[!mayfail]` to `[!shouldfail`] means that these tests fail if they pass. This stops tests inadvertently getting fixed without anyone noticing. I used @owen-mc-diffblue suggestion of `xfail` for an expected failure. Finally, I added a separate run of these failed tests to CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -355,9 +355,13 @@ script:
   - env PATH=$PATH:`pwd`/src/solvers UBSAN_OPTIONS=print_stacktrace=1 make -C regression/cbmc test-cprover-smt2
   - make -C unit "CXX=${COMPILER} ${EXTRA_CXXFLAGS}" -j2
   - make -C unit test
+  - echo "Running expected failure tests"
+  - make TAGS="[!shouldfail]" -C unit test
   - env UBSAN_OPTIONS=print_stacktrace=1 make -C jbmc/regression test-parallel "CXX=${COMPILER} ${EXTRA_CXXFLAGS}" -j2 JOBS=2
   - make -C jbmc/unit "CXX=${COMPILER} ${EXTRA_CXXFLAGS}" -j2
   - make -C jbmc/unit test
+  - echo "Running expected failure tests"
+  - make TAGS="[!shouldfail]" -C jbmc/unit test
 
 before_cache:
   - ccache -s
--- a/.travis.yml
+++ b/.travis.yml
@@ -243,7 +243,7 @@ jobs:
         - cmake -S . -Bbuild '-DCMAKE_BUILD_TYPE=Release'  '-DCMAKE_CXX_COMPILER=/usr/bin/g++-7' '-DCMAKE_USE_CUDD=true' -DCMAKE_CXX_FLAGS="-DBDD_GUARDS" '-DWITH_MEMORY_ANALYZER=On'
         - git submodule update --init --recursive
         - cmake --build build -- -j4
-      script: (cd build; ctest -V -L CORE -j2)
+      script: (cd build; ctest -V -L CORE -j2; ctest -V -R unit-xfail -j2)
 
     # cmake build using clang++-6
     - stage: Test different OS/CXX/Flags
@@ -280,7 +280,7 @@ jobs:
         - cmake -S . -Bbuild '-DCMAKE_BUILD_TYPE=Release' '-DCMAKE_CXX_COMPILER=/usr/bin/clang++-7' '-DCMAKE_CXX_FLAGS=-Qunused-arguments' '-DWITH_MEMORY_ANALYZER=On'
         - git submodule update --init --recursive
         - cmake --build build -- -j4
-      script: (cd build; ctest -V -L CORE -j2)
+      script: (cd build; ctest -V -L CORE -j2; ctest -V -R unit-xfail -j2)
 
     # cmake build on OSX, using default clang
     - stage: Test different OS/CXX/Flags
@@ -300,7 +300,7 @@ jobs:
         - cmake -S . -Bbuild '-DCMAKE_BUILD_TYPE=Release' '-DCMAKE_OSX_ARCHITECTURES=x86_64'
         - git submodule update --init --recursive
         - cmake --build build -- -j4
-      script: (cd build; ctest -V -L CORE -j2)
+      script: (cd build; ctest -V -L CORE -j2; ctest -V -R unit-xfail -j2)
 
 
     # Run Coverity
--- a/jbmc/unit/Makefile
+++ b/jbmc/unit/Makefile
@@ -144,8 +144,9 @@ all: $(CATCH_TEST)
 clean: java-testing-utils-clean
 
 test: $(CATCH_TEST)
-	if ! ./$(CATCH_TEST) -l | grep -q "^$(N_CATCH_TESTS) test cases" ; then \
-	  ./$(CATCH_TEST) -l ; fi
+	# Include hidden tests by specifying "*,[.]" for tests to count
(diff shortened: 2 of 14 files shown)


unit/testing-utils/use_catch.h
```diff
@@ -36,4 +36,7 @@ Author: Michael Tautschnig
 #include <util/pragma_pop.def>
 #endif
 
+/// Add to the end of test tags to mark a test that is expected to fail
+#define XFAIL "[.][shouldfail]"
```
hannes-steffenhagen-diffblue: Why?

jbmc/unit/java_bytecode/ci_lazy_methods/lazy_load_lambdas.cpp
```diff
@@ -12,7 +12,7 @@ Author: Diffblue Limited.
 
 SCENARIO(
   "Lazy load lambda methods",
-  "[core][java_bytecode][ci_lazy_methods][lambdas][!mayfail]")
```
hannes-steffenhagen-diffblue, line 15: `mayfail` doesn’t stop a test from running, it just means test failure isn’t counted.

jbmc/unit/java_bytecode/java_bytecode_convert_method/convert_invoke_dynamic.cpp
```diff
@@ -188,7 +188,7 @@ void validate_local_variable_lambda_assignment(
 SCENARIO(
   "Converting invokedynamic with a local lambda",
   "[core]"
-  "[lambdas][java_bytecode][java_bytecode_convert_method][!mayfail]")
+  "[lambdas][java_bytecode][java_bytecode_convert_method]" XFAIL)
```
hannes-steffenhagen-diffblue, line 191: this should be `[!shouldfail]` (it’s `shouldfail` in the macro), and I still dont think this should be a macro.

.travis.yml
```diff
@@ -355,9 +355,13 @@ script:
   - env PATH=$PATH:`pwd`/src/solvers UBSAN_OPTIONS=print_stacktrace=1 make -C regression/cbmc test-cprover-smt2
   - make -C unit "CXX=${COMPILER} ${EXTRA_CXXFLAGS}" -j2
   - make -C unit test
+  - echo "Running expected failure tests"
+  - make TAGS="[shouldfail]" -C unit test
   - env UBSAN_OPTIONS=print_stacktrace=1 make -C jbmc/regression test-parallel "CXX=${COMPILER} ${EXTRA_CXXFLAGS}" -j2 JOBS=2
   - make -C jbmc/unit "CXX=${COMPILER} ${EXTRA_CXXFLAGS}" -j2
   - make -C jbmc/unit test
+  - echo "Running expected failure tests"
+  - make TAGS="[shouldfail]" -C jbmc/unit test xfail
```
hannes-steffenhagen-diffblue: ??? Why not just use the builtin `!shouldfail` mechanism?

jbmc/unit/Makefile
```diff
@@ -144,9 +144,9 @@ all: $(CATCH_TEST)
 clean: java-testing-utils-clean
 
 test: $(CATCH_TEST)
-	if ! ./$(CATCH_TEST) -l | grep -q "^$(N_CATCH_TESTS) test cases" ; then \
-	  ./$(CATCH_TEST) -l ; fi
-	./$(CATCH_TEST)
+	if ! ./$(CATCH_TEST) *,[.] -l | grep -q "^$(N_CATCH_TESTS) matching test cases" ; then \
```
hannes-steffenhagen-diffblue: I’m not sure what this supposed to mean.

hannes-steffenhagen-diffblue (commented):
If you don’t want to run `[!shouldfail]` test you can just provide `~[!shouldfail]` as a tag filter.

unit/testing-utils/use_catch.h
thk123: From the commit message:
> These tests won't be run by default (de-cluttering the output with expected failures). When run, they will fail if they pass

In previous PR Owen / Micahel complained that shouldfail was misleading. They _shouldn't_ fail, in the perfect world, but they do due to known bug. We felt expected fail (or XFAIL as python calls them) was clearer. Can change to `EXPECTED_FAIL` if you'd prefer?

jbmc/unit/java_bytecode/ci_lazy_methods/lazy_load_lambdas.cpp
thk123: A test that can pass or fail is bad - since no body will observe if it changes state. I believe this has been erroneously marked as mayfail, since it passes.

jbmc/unit/java_bytecode/java_bytecode_convert_method/convert_invoke_dynamic.cpp
thk123: Fixed in later commit

.travis.yml
thk123: Not sure what you mean? The trailing `xfail` was a typo, but this is me passing the appropriate tags to the unit runner

jbmc/unit/Makefile
thk123: It is listing all tests, even hidden tests, to check the number of tests == the number of tests in source. (The check is to catch people forgetting to add their unit tests to `Makefile`).

thk123:
> If you don’t want to run [!shouldfail] test you can just provide ~[!shouldfail] as a tag filter.

This is also an option. 

I prefer my approach as it makes it easier to use for the general case. I.e. to run the unit tests, you just run the unit tests binary. You neither need to provide additional flags, nor do you get misleading output about failed tests. 

Happy to change to that though, if that is preferred.

.travis.yml
```diff
@@ -355,9 +355,13 @@ script:
   - env PATH=$PATH:`pwd`/src/solvers UBSAN_OPTIONS=print_stacktrace=1 make -C regression/cbmc test-cprover-smt2
   - make -C unit "CXX=${COMPILER} ${EXTRA_CXXFLAGS}" -j2
   - make -C unit test
+  - echo "Running expected failure tests"
+  - make TAGS="[!shouldfail]" -C unit test
```
chrisr-diffblue, line 359: What about CMake configurations? Does anything need to be done for those?

thk123: Good spot, fixed

chrisr-diffblue (approved):
Looks good, thanks for checking/fixing the cmake stuff.

jbmc/unit/Makefile
thk123: @hannes-steffenhagen-diffblue how about?
```suggestion
  # Include hidden tests by specifying "*,[.]" for tests to count
	if ! ./$(CATCH_TEST) *,[.] -l | grep -q "^$(N_CATCH_TESTS) matching test cases" ; then \
```

.travis.yml
hannes-steffenhagen-diffblue: This should be `[!shouldfail]`, not `[shouldfail]`. I believe might be fixed by a later commit?

jbmc/unit/Makefile
hannes-steffenhagen-diffblue: Oh. Ok, I thought this was using a glob pattern somehow. Would you mind putting quotes around it to clarify it’s supposed to be passed as an argument to catch as-is?

hannes-steffenhagen-diffblue (approved):
I still don’t really like using a macro for this, but otherwise LGTM now.

jbmc/unit/Makefile
thk123: Done - will merge on green CI

PR #5249, opened 2020-02-27
Add generic gcc overflow builtins

hannes-steffenhagen-diffblue:
GCC has builtin operators `__builtin_add_overflow`, `__builtin_sub_overflow`
and `__builtin_mul_overflow` that perform an arithmetic operation and
check for overflow at the same time. This adds support for these
operators.

### DIFF
--- a/src/ansi-c/c_typecheck_expr.cpp
+++ b/src/ansi-c/c_typecheck_expr.cpp
@@ -12,6 +12,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "c_typecheck_base.h"
 
 #include <cassert>
+#include <sstream>
 
 #include <util/arith_tools.h>
 #include <util/c_types.h>
@@ -33,6 +34,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "builtin_factory.h"
 #include "c_qualifiers.h"
 #include "c_typecast.h"
+#include "expr2c.h"
 #include "padding.h"
 #include "type2name.h"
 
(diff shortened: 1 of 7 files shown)


hannes-steffenhagen-diffblue:
I had considered replacing the implementation of the non-generic `__builtin_*_overflow` in `math.c` with these, but unfortunately this runs afoul of the library check script in some configurations (for example, on travis we run library check against gcc 4.9, which doesn’t have these operators as they were added for gcc 5).

NlightNFotis (approved):
LGTM

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2801,6 +2803,86 @@ exprt c_typecheck_baset::do_special_functions(
...
+        return plus_exprt{lhs, rhs};
+      }
+      else if(identifier == "__builtin_sub_overflow")
+      {
+        return minus_exprt{lhs, rhs};
+      }
+      else
+      {
+        INVARIANT(
+          identifier == "__builtin_mul_overflow",
```
chrisr-diffblue, line 2872: Given this is inside the `else if(
     identifier == "__builtin_add_overflow" ||
     identifier == "__builtin_sub_overflow" ||
     identifier == "__builtin_mul_overflow")` branch, is it any easier to read to just make this final else and 'else if(identifier == blah)` and then and a final `else UNREACHABLE` ?

hannes-steffenhagen-diffblue: I don’t know about “easier”, the reason I wrote it like this is because I don’t like seeing “unreachable” whenever one of these changes. So I have a slight preference for doing it this way round, but I don’t mind doing it the other way round either.

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2801,6 +2803,86 @@ exprt c_typecheck_baset::do_special_functions(
...
+                    << " takes exactly 3 arguments, but "
+                    << expr.arguments().size() << " were provided";
+      throw invalid_source_file_exceptiont{error_message.str()};
+    }
+
+    for(std::size_t arg_index = 0; arg_index < 3; ++arg_index)
+    {
+      auto const &argument = expr.arguments()[arg_index];
+      bool const arg_has_correct_type =
+        (arg_index < 2) ? (argument.type().id() == ID_signedbv ||
```
chrisr-diffblue: Trying to do this in a single loop makes this a little complex to read. Would it be better to handle the argument type-check (seeing as their are only 3 args) as two separate blocks, one block to check arg 0 and 1, and one to check arg 2, with each block setting the arg_has_correct_type flag, then the error message block could stay the same (but not be in a loop).

src/ansi-c/c_typecheck_expr.cpp
chrisr-diffblue: Fair enough, I don't have strong leaning either way.

chrisr-diffblue (approved):
Looks good - couple of queries around code structure (but not necessarily blockers). It might also be nice if some of the test cases also explored the case where arg 0 and arg 1 are not the same type... As far as I can tell from GCC docs, those first two arguments can be arbitrary (different) integral types, whereas I think all the test cases here use the same type for both arguments. I think everything will work fine if they are different, but would be nice to have one or two test cases that show that :-)

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2801,6 +2803,86 @@ exprt c_typecheck_baset::do_special_functions(
...
+                           argument.type().id() == ID_unsignedbv)
+                        : (argument.type().id() == ID_pointer &&
+                           (argument.type().subtype().id() == ID_signedbv ||
+                            argument.type().subtype().id() == ID_unsignedbv));
+      if(!arg_has_correct_type)
+      {
+        std::ostringstream error_message;
+        error_message << expr.source_location().as_string() << ": "
+                      << identifier << " has signature " << identifier
+                      << "(integral, integral, integral*), "
```
chrisr-diffblue: GCC docs show: ` (type1 a, type2 b, type3 *res)` - implying a, b and res can all be *different* integral types, so might be worth writing this as `(integral1, integral2, integral3 *)` ? What does gcc give as an error message in this case?

src/ansi-c/c_typecheck_expr.cpp
hannes-steffenhagen-diffblue: Hm, maybe? Could you sketch out how you’d do it?

src/ansi-c/c_typecheck_expr.cpp
hannes-steffenhagen-diffblue: Version dependent I suppose. `gcc 10` says

```
test.c:4:46: error: argument 2 in call to function ‘__builtin_sub_overflow’ does not have integral type
```

I was kind of inspired by [C++ concept syntax](https://en.cppreference.com/w/cpp/concepts/integral) for this, but if you think it’s confusing I can change it to something else.

src/ansi-c/c_typecheck_expr.cpp
chrisr-diffblue: Sorry, yeah, some pseudo code would have been better :-)
```
bool correct_type=true;
// Typecheck args 0 and 1
for(int i = 0; i<1; i++) {
   correct_type &= is_correct_type(arg[i])
}
// Typecheck arg 2
correct_type &= is_correct_result_type(arg[2]);

// Error message if typecheck fails
if(!correct_type) {
 // user error message here
}
```

chrisr-diffblue: So basically the same logic you have, but without the tertiary operator which you effectively use to encode additional state in the argument index :-)

src/ansi-c/c_typecheck_expr.cpp
chrisr-diffblue: Well, I prefer what you have than what GCC has :-) No big deal.

src/ansi-c/c_typecheck_expr.cpp
chrisr-diffblue: Oh, and I think you may be able to use the existing utility functions `is_signed_or_unsigned_bitvector`, `is_c_integral_type` and  `is_c_integral_pointer_type` here too ?

hannes-steffenhagen-diffblue: I didn’t know about these, I’ll see about clarifying it using your suggestions.

PR #5250, opened 2020-03-02
Avoids arithmetic overflow on signed to unsigned type conversion

feliperodri:
Running CBMC on the following example from `regression/cbmc-library/String6` using the `conversion-check` flag reported an arithmetic overflow:
```
cbmc main.c --pointer-check --bounds-check --conversion-check
```

```
#include <assert.h>
#include <stdlib.h>
#include <string.h>

int main()
{
  char str[500] = "Hello";

  assert(strcmp(str, "Hello") == 0);
  assert(strncmp(str, "Hello", 5) == 0);

#ifndef _MSC_VER
  assert(strcasecmp(str, "HELLO") == 0);
  assert(strncasecmp(str, "HELLO", 5) == 0);
#endif

  assert(strcmp(str, "\xff") < 0);
  assert(strncmp("ASDxx", "ASDyy", 3) == 0);

  assert(strlen(str) == 5);
  char *str_cpy = strdup(str);
  assert(strcmp(str, str_cpy) == 0);
  free(str_cpy);

  return 0;
}
```

```
Violated property:
  file <builtin-library-strncmp> function strncmp line 26 thread 0
  arithmetic overflow on signed to unsigned type conversion in (unsigned char)s1[(signed long int)i]
  s1[(signed long int)i] >= 0
```

Adding pragmas to avoid spurious reports and to keep in sync with the [standard](https://pubs.opengroup.org/onlinepubs/7908799/xsh/strncmp.html),i.e., _"the sign of a non-zero return value is determined by the sign of the difference between the values of the first pair of bytes (**both interpreted as type unsigned char**) that differ in the strings being compared."_

### DIFF
--- a/src/ansi-c/library/string.c
+++ b/src/ansi-c/library/string.c
@@ -374,8 +374,11 @@ inline int strcmp(const char *s1, const char *s2)
   unsigned char ch1, ch2;
   do
   {
+#  pragma CPROVER check push
+#  pragma CPROVER check disable "conversion"
     ch1=s1[i];
     ch2=s2[i];
+#  pragma CPROVER check pop
 
     if(ch1==ch2)
     {
@@ -389,7 +392,7 @@ inline int strcmp(const char *s1, const char *s2)
   }
   while(ch1!=0 && ch2!=0);
   return 0;
-  #endif
+#endif
 }
 
 /* FUNCTION: strcasecmp */
@@ -420,8 +423,11 @@ inline int strcasecmp(const char *s1, const char *s2)
   unsigned char ch1, ch2;
   do
   {
+#  pragma CPROVER check push
+#  pragma CPROVER check disable "conversion"
     ch1=s1[i];
     ch2=s2[i];
+#  pragma CPROVER check pop
 
     if(ch1>='A' && ch1<='Z') ch1+=('a'-'A');
     if(ch2>='A' && ch2<='Z') ch2+=('a'-'A');
@@ -438,7 +444,7 @@ inline int strcasecmp(const char *s1, const char *s2)
   }
   while(ch1!=0 && ch2!=0);
   return 0;
-  #endif
+#endif
 }
 
 /* FUNCTION: strncmp */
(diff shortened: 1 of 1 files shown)


src/ansi-c/library/string.c
```diff
@@ -462,7 +462,7 @@ inline int strncmp(const char *s1, const char *s2, size_t n)
                          "strncmp zero-termination of 2nd argument");
   #else
   __CPROVER_size_t i=0;
-  unsigned char ch1, ch2;
+  char ch1, ch2;
```
tautschnig: Can you please also apply the fix to `strncasecmp`?

tautschnig: And `strcmp` and `strcasecmp`?

feliperodri: Done.

smowton (approved):

chrisr-diffblue (approved):
Looks good - thanks for addressing Michael's request too.

src/ansi-c/library/string.c
```diff
@@ -374,8 +374,11 @@ inline int strcmp(const char *s1, const char *s2)
   unsigned char ch1, ch2;
   do
   {
+#  pragma CPROVER check push
```
hannes-steffenhagen-diffblue, line 377: Why are these pragmas necessary? It seems to be the problem is rather that `ch1` and `ch2` are declared as `unsigned char` rather than `char` (which may or may not be `unsigned` depending on configuration).

hannes-steffenhagen-diffblue (changes requested):
I’d ask you to withhold merging this before confirming this doesn’t just mask a bug in the implementation.

src/ansi-c/library/string.c
feliperodri: @hannes-steffenhagen-diffblue Yep, the source of the problem is that `ch1` and `ch2` are declared as `unsigned char`; however, `\xff` doesn't work correctly with `signed char` (see the test case `String6`). Thus, we can either fix that (special) case or use `unsigned char` and add `#pragma CPROVER check disable "conversion"`. For me, the latter makes sense because the standard for these functions interpreted this characters as type `unsigned char`, which I mentioned at the description of this PR.

hannes-steffenhagen-diffblue (approved):
Still feels wrong to me, but not a blocker after explanation.

hannes-steffenhagen-diffblue:
@feliperodri why did you close this PR?

PR #5251, opened 2020-03-03
Factor out computation of maximum union member

tautschnig:
Avoids duplicate code and enables future re-use.

This is in preparation of further union support in byte-extract lowering. Refactoring is tested via existing unit tests.

### DIFF
--- a/src/solvers/lowering/byte_operators.cpp
+++ b/src/solvers/lowering/byte_operators.cpp
@@ -21,6 +21,41 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/simplify_expr.h>
 #include <util/string_constant.h>
 
+/// Determine the member of maximum fixed bit width in a union type. If no
+/// member, or no member of fixed and non-zero width can be found, return
+/// nullopt.
+/// \param union_type: Type to determine the member of.
+/// \param ns: Namespace to resolve tag types.
+/// \return Pair of a componentt pointing to the maximum fixed bit-width
+///   member of \p union_type and the bit width of that member.
+static optionalt<std::pair<struct_union_typet::componentt, mp_integer>>
+find_widest_union_component(const union_typet &union_type, const namespacet &ns)
+{
+  const union_typet::componentst &components = union_type.components();
+
+  mp_integer max_width = 0;
+  typet max_comp_type;
+  irep_idt max_comp_name;
+
+  for(const auto &comp : components)
+  {
+    auto element_width = pointer_offset_bits(comp.type(), ns);
+
+    if(!element_width.has_value() || *element_width <= max_width)
+      continue;
+
+    max_width = *element_width;
+    max_comp_type = comp.type();
+    max_comp_name = comp.get_name();
+  }
+
+  if(max_width == 0)
+    return {};
+  else
+    return std::make_pair(
+      struct_union_typet::componentt{max_comp_name, max_comp_type}, max_width);
+}
+
 static exprt bv_to_expr(
   const exprt &bitvector_expr,
   const typet &target_type,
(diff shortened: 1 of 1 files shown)


src/solvers/lowering/byte_operators.cpp
```diff
@@ -21,6 +21,46 @@ Author: Daniel Kroening, kroening@kroening.com
...
+/// member, or no member of fixed and non-zero width can be found, return
+/// nullopt.
+/// \param union_expr: Expression to use as compound operand in the resulting
+///   member_exprt.
+/// \param union_type: Type to determine the member of.
+/// \param ns: Namespace to resolve tag types.
+/// \return Pair of a member_exprt pointing to the maximum fixed bit-width
+///   member of \p union_type and the bit width of that member.
+static optionalt<std::pair<member_exprt, mp_integer>>
+find_maximum_union_component(
```
smowton: maximum -> widest?

src/solvers/lowering/byte_operators.cpp
```diff
@@ -21,6 +21,46 @@ Author: Daniel Kroening, kroening@kroening.com
...
+    max_width = *element_width;
+    max_comp_type = comp.type();
+    max_comp_name = comp.get_name();
+  }
+
+  if(max_width == 0)
+    return {};
+  else
+    return std::make_pair(
+      member_exprt{union_expr, max_comp_name, max_comp_type}, max_width);
```
smowton: A bit surprising that a utility would make a member-expression, rather than just returning the widest component

smowton (approved):
Cosmetic components

src/solvers/lowering/byte_operators.cpp
tautschnig: Fixed (and left-hand sides are renamed accordingly).

src/solvers/lowering/byte_operators.cpp
tautschnig: Good call - this makes one use slightly more work (having to construct the `member_exprt` in place), but avoids a parameter that isn't really necessary and could be used in a wrong way.

src/solvers/lowering/byte_operators.cpp
```diff
@@ -21,6 +21,41 @@ Author: Daniel Kroening, kroening@kroening.com
...
 
+/// Determine the member of maximum fixed bit width in a union type. If no
+/// member, or no member of fixed and non-zero width can be found, return
+/// nullopt.
+/// \param union_type: Type to determine the member of.
+/// \param ns: Namespace to resolve tag types.
+/// \return Pair of a componentt pointing to the maximum fixed bit-width
+///   member of \p union_type and the bit width of that member.
+static optionalt<std::pair<struct_union_typet::componentt, mp_integer>>
+find_widest_union_component(const union_typet &union_type, const namespacet &ns)
```
chrisr-diffblue, line 32: Is there any particular reason this returns an optional pair, rather than just an optional `componentt` ? Given that the componentt has the the appropriate type in it and the caller could just do `pointer_offset_bits` themselves. If it's just to simplify future clients, then fine :-) But just wondering as I don't see any current clients using the second half of the pair, and having the extra `->first` or `->second` slightly reduces readability at the call sites.

chrisr-diffblue (approved):
Happy enough for this to go in as it is, but just a mild query around the interface of `find_widest_union_component`

src/solvers/lowering/byte_operators.cpp
tautschnig: Yes, that second part is only used in #5252. I hope that's ok - I'll go ahead and merge.

tautschnig: FWIW, you're of course right that the caller could compute the size themselves, but I thought duplicating work was not a good idea.

chrisr-diffblue: Fine by me. I was reviewing in PR order so hadn't seen the followup PR at that point.

PR #5257, opened 2020-03-04
Add function pointer restriction feature

danpoe:
Replaces #5168

### DIFF
--- a/doc/cprover-manual/restrict-function-pointer.md
+++ b/doc/cprover-manual/restrict-function-pointer.md
@@ -1,0 +1,173 @@
+[CPROVER Manual TOC](../../)
+
+## Restricting function pointers
+
+### Motivation
+
+CBMC comes with a way to resolve calls to function pointers to direct function
+calls. This is needed because symbolic execution itself can't handle calls to
+function pointers. In practice, this looks something like this:
+
+Given that there are functions with these signatures available in the program:
+
+```
+int f(int x);
+int g(int x);
+int h(int x);
+```
+
+And we have a call site like this:
+
+```
+typedef int(*fptr_t)(int x);
+void call(fptr_t fptr) {
+  int r = fptr(10);
+  assert(r > 0);
+}
+```
+
+Function pointer removal will turn this into code similar to this:
+
+```
+void call(fptr_t fptr) {
+  int r;
+  if(fptr == &f) {
+    r = f(10);
+  } else if(fptr == &g) {
+    r = g(10);
+  } else if(fptr == &h) {
+    r = h(10);
+  } else {
+    // sanity check
+    assert(false);
+    assume(false);
+  }
+  return r;
+}
+```
+
+This works well enough for simple cases. However, this is a very simple
+replacement only based on the signature of the function (and whether or not they
+have their address taken somewhere in the program), so if there are many
+functions matching a particular signature, or if some of these functions are
+expensive in symex (e.g. functions with lots of loops or recursion), then this
+can be a bit cumbersome - especially if we, as a user, already know that a
+particular function pointer will only resolve to a single function or a small
+set of functions. This is what the `--restrict-function-pointer` option allows.
+
+### Example
+
+Take the motivating example. Let us assume that we know for a fact that `call`
+will always receive pointers to either `f` or `g` during actual executions of
+the program, and symbolic execution for `h` is too expensive to simply ignore
+the cost of its branch. For this, we will label the places in each function
+where function pointers are being called, to this pattern:
+
+```
+<function-name>.function_pointer_call.<N>
+```
+
+where `N` is referring to which function call it is - so the first call to a
+function pointer in a function will have `N=1`, the 5th `N=5` etc.
+
+We can call `cbmc` with `--restrict-function-pointer
+call.function_pointer_call.1/f,g`. This can be read as
+
+> For the first call to a function pointer in the function `call`, assume that
+> it can only be a call to `f` or `g`
+
+The resulting code looks similar to the original example, except now there will
+not be a call to `h`:
+
+```
+void call(fptr_t fptr) {
+  int r;
+  if(fptr == &f) {
+    r = f(10);
+  } else if(fptr == &g) {
+    r = g(10);
+  } else {
+    // sanity check
+    assert(false);
+    assume(false);
+  }
+  return r;
+}
+```
+
+Another example: Imagine we have a simple virtual filesystem API and implementation
+like this:
+
+```
+typedef struct filesystem_t filesystem_t;
+struct filesystem_t {
+  int (*open)(filesystem_t *filesystem, const char* file_name);
+};
+
+int fs_open(filesystem_t *filesystem, const char* file_name) {
+  filesystem->open(filesystem, file_name);
+}
+
+int nullfs_open(filesystem_t *filesystem, const char* file_name) {
+  return -1;
+}
+
+filesystem_t nullfs_val = {.open = nullfs_open};
+filesystem *const nullfs = &nullfs_val;
+
+filesystem_t *get_fs_impl() {
+  // some fancy logic to determine
+  // which filesystem we're getting -
+  // in-memory, backed by a database, OS file system
+  // - but in our case, we know that
+  // it always ends up being nullfs
+  // for the cases we care about
+  return nullfs;
+}
+int main(void) {
+  filesystem_t *fs = get_fs_impl();
+  assert(fs_open(fs, "hello.txt") != -1);
+}
+```
+
+In this case, the assumption is that *we* know that in our `main`, `fs` can be
+nothing other than `nullfs`; But perhaps due to the logic being too complicated
+symex ends up being unable to figure this out, so in the call to `fs_open()` we
+end up branching on all functions matching the signature of
+`filesystem_t::open`, which could be quite a few functions within the program.
+Worst of all, if it's address is ever taken in the program, as far as the "dumb"
+function pointer removal is concerned it could be `fs_open()` itself due to it
+having a matching signature, leading to symex being forced to follow a
+potentially infinite recursion until its unwind limit.
+
+In this case we can again restrict the function pointer to the value which we
+know it will have:
+
+```
+--restrict-function-pointer fs_open.function_pointer_call.1/nullfs_open
+```
+
+### Loading from file
+
+If you have many places where you want to restrict function pointers, it'd be a
+nuisance to have to specify them all on the command line. In these cases, you
+can specify a file to load the restrictions from instead, via the
+`--function-pointer-restrictions-file` option, which you can give the name of a
+JSON file with this format:
+
+```
+{
+  "function_call_site_name": ["function1", "function2", ...],
+   ...
+}
+```
+
+**Note:** If you pass in multiple files, or a mix of files and command line
+restrictions, the final restrictions will be a set union of all specified
+restrictions.
+
+**Note:** as of now, if something goes wrong during type checking (i.e. making
+sure that all function pointer replacements refer to functions in the symbol
+table that have the correct type), the error message will refer the command line
+option `--restrict-function-pointer` regardless of whether the restriction in
+question came from the command line or a file.
--- a/doc/cprover-manual/restrict-function-pointer.md
+++ b/doc/cprover-manual/restrict-function-pointer.md
@@ -2,11 +2,17 @@
 
 ## Restricting function pointers
 
+In this document, we describe the `goto-instrument` feature to replace calls
+through function pointers by case distinctions over calls to given sets of
+functions.
+
 ### Motivation
 
-CBMC comes with a way to resolve calls to function pointers to direct function
-calls. This is needed because symbolic execution itself can't handle calls to
-function pointers. In practice, this looks something like this:
+The CPROVER framework includes a goto program transformation pass
+`remove_function_pointers()` to resolve calls to function pointers to direct
+function calls. The pass is needed by `cbmc`, as symbolic execution itself can't
+handle calls to function pointers. In practice, the transformation pass works as
+follows:
 
 Given that there are functions with these signatures available in the program:
 
@@ -53,15 +59,16 @@ functions matching a particular signature, or if some of these functions are
 expensive in symex (e.g. functions with lots of loops or recursion), then this
 can be a bit cumbersome - especially if we, as a user, already know that a
 particular function pointer will only resolve to a single function or a small
-set of functions. This is what the `--restrict-function-pointer` option allows.
+set of functions. The `goto-instrument` option `--restrict-function-pointer`
+allows to manually specify this set of functions.
 
 ### Example
 
-Take the motivating example. Let us assume that we know for a fact that `call`
-will always receive pointers to either `f` or `g` during actual executions of
-the program, and symbolic execution for `h` is too expensive to simply ignore
-the cost of its branch. For this, we will label the places in each function
-where function pointers are being called, to this pattern:
+Take the motivating example above. Let us assume that we know for a fact that
+`call` will always receive pointers to either `f` or `g` during actual
+executions of the program, and symbolic execution for `h` is too expensive to
+simply ignore the cost of its branch. For this, we will label the places in each
+function where function pointers are being called, to this pattern:
 
 ```
 <function-name>.function_pointer_call.<N>
@@ -70,14 +77,14 @@ where function pointers are being called, to this pattern:
 where `N` is referring to which function call it is - so the first call to a
 function pointer in a function will have `N=1`, the 5th `N=5` etc.
 
-We can call `cbmc` with `--restrict-function-pointer
-call.function_pointer_call.1/f,g`. This can be read as
+We can call `goto-instrument --restrict-function-pointer
+call.function_pointer_call.1/f,g in.gb out.gb`. This can be read as
 
 > For the first call to a function pointer in the function `call`, assume that
 > it can only be a call to `f` or `g`
 
-The resulting code looks similar to the original example, except now there will
-not be a call to `h`:
+The resulting output (written to goto binary `out.gb`) looks similar to the
+original example, except now there will not be a call to `h`:
 
 ```
 void call(fptr_t fptr) {
--- a/doc/cprover-manual/restrict-function-pointer.md
+++ b/doc/cprover-manual/restrict-function-pointer.md
@@ -142,7 +142,7 @@ nothing other than `nullfs`; But perhaps due to the logic being too complicated
 symex ends up being unable to figure this out, so in the call to `fs_open()` we
 end up branching on all functions matching the signature of
 `filesystem_t::open`, which could be quite a few functions within the program.
-Worst of all, if it's address is ever taken in the program, as far as the "dumb"
+Worst of all, if its address is ever taken in the program, as far as the "dumb"
 function pointer removal is concerned it could be `fs_open()` itself due to it
 having a matching signature, leading to symex being forced to follow a
 potentially infinite recursion until its unwind limit.
@@ -175,6 +175,6 @@ restrictions.
 
 **Note:** as of now, if something goes wrong during type checking (i.e. making
 sure that all function pointer replacements refer to functions in the symbol
-table that have the correct type), the error message will refer the command line
-option `--restrict-function-pointer` regardless of whether the restriction in
-question came from the command line or a file.
+table that have the correct type), the error message will refer to the command
+line option `--restrict-function-pointer` regardless of whether the restriction
+in question came from the command line or a file.
--- a/regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.c
+++ b/regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.c
@@ -1,0 +1,59 @@
+#include <assert.h>
+
+typedef int (*fptr_t)(int);
+
+fptr_t get_f(void);
+fptr_t get_g(void);
+
+void use_fg(int choice, fptr_t fptr, fptr_t gptr)
+{
+  assert((choice ? fptr : gptr)(10) == 10 + choice);
+}
+
+// this is just here to distinguish the behavior from FP removal, which'd include g
+int g_always_false_cond = 0;
+
+int main(void)
+{
+  use_fg(0, get_f(), get_g());
+  use_fg(1, get_f(), get_g());
+}
+
+int f(int x)
+{
+  return x + 1;
+}
+
+int g(int x)
+{
+  return x;
+}
+
+int h(int x)
+{
+  return x / 2;
+}
+
+fptr_t get_f(void)
+{
+  if(!g_always_false_cond)
+  {
+    return f;
+  }
+  else
+  {
+    return h;
+  }
+}
+
+fptr_t get_g(void)
+{
+  if(!g_always_false_cond)
+  {
+    return g;
+  }
+  else
+  {
+    return h;
+  }
+}
--- a/regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.c
+++ b/regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.c
@@ -10,9 +10,6 @@ void use_fg(int choice, fptr_t fptr, fptr_t gptr)
   assert((choice ? fptr : gptr)(10) == 10 + choice);
 }
 
-// this is just here to distinguish the behavior from FP removal, which'd include g
-int g_always_false_cond = 0;
-
 int main(void)
 {
   use_fg(0, get_f(), get_g());
@@ -34,9 +31,14 @@ int h(int x)
   return x / 2;
 }
 
+// Below we take the address of f, g, and h. Thus remove_function_pointers()
+// would create a case distinction involving f, g, and h in the function
+// use_fg() above.
+int always_false_cond = 0;
+
 fptr_t get_f(void)
 {
-  if(!g_always_false_cond)
+  if(!always_false_cond)
   {
     return f;
   }
@@ -48,7 +50,7 @@ fptr_t get_f(void)
 
 fptr_t get_g(void)
 {
-  if(!g_always_false_cond)
+  if(!always_false_cond)
   {
     return g;
   }
--- a/regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.desc
+++ b/regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.desc
@@ -5,3 +5,7 @@ test.c
 \[use_fg.assertion.1\] line \d+ dereferenced function pointer at use_fg.function_pointer_call.1 must be one of \[(f|g), (f|g)\]: SUCCESS
 ^EXIT=0$
 ^SIGNAL=0$
+--
+--
+This test checks that the selected function pointer is replaced by a case
+distinction over both functions f and g.
--- a/regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.desc
+++ b/regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.desc
@@ -1,0 +1,7 @@
+CORE
+test.c
+--restrict-function-pointer "use_fg.function_pointer_call.1/f,g"
+\[use_fg.assertion.2\] line \d+ assertion \(choice \? fptr : gptr\)\(10\) == 10 \+ choice: SUCCESS
+\[use_fg.assertion.1\] line \d+ dereferenced function pointer at use_fg.function_pointer_call.1 must be one of \[(f|g), (f|g)\]: SUCCESS
+^EXIT=0$
+^SIGNAL=0$
--- a/regression/goto-instrument/restrict-function-pointer-to-multiple-functions-code-check/test.desc
+++ b/regression/goto-instrument/restrict-function-pointer-to-multiple-functions-code-check/test.desc
@@ -1,0 +1,9 @@
+CORE
+test.c
+--restrict-function-pointer use_f.function_pointer_call.1/f,g --show-goto-functions
+f\(10\)
+g\(10\)
+^EXIT=0$
+^SIGNAL=0$
+--
+h\(10\)
--- a/regression/goto-instrument/restrict-function-pointer-to-multiple-functions-code-check/test.desc
+++ b/regression/goto-instrument/restrict-function-pointer-to-multiple-functions-code-check/test.desc
@@ -7,3 +7,6 @@ g\(10\)
 ^SIGNAL=0$
 --
 h\(10\)
+--
+This test checks that no call to g appears in the goto program after the
+transformation by the restrict function pointer feature.
--- a/regression/goto-instrument/restrict-function-pointer-to-multiple-functions-via-file-and-command-line-options/test.desc
+++ b/regression/goto-instrument/restrict-function-pointer-to-multiple-functions-via-file-and-command-line-options/test.desc
@@ -4,3 +4,15 @@ test.c
 \[use_f\.assertion\.1\] line \d+ dereferenced function pointer at use_f.function_pointer_call.1 must be one of \[(f|g), (f|g)\]: FAILURE
 ^EXIT=10$
 ^SIGNAL=0$
+--
+--
+This test checks that the restrictions for a function pointer are the union of
+the restrictions given in a file and on the command line.
+
+The test further checks that the correct safety assertions are generated. The
+function pointer restriction feature outputs safety assertions for all calls
+that it replaces. That is, when it replaces a call with a case distinction over
+a given set of functions, it adds an assertion checking that in the original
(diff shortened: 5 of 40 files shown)


doc/cprover-manual/restrict-function-pointer.md
```diff
@@ -0,0 +1,173 @@
+[CPROVER Manual TOC](../../)
+
+## Restricting function pointers
+
+### Motivation
+
+CBMC comes with a way to resolve calls to function pointers to direct function
```
hannes-steffenhagen-diffblue: The documentation needs to be updated to reference `goto-instrument` instead of `cbmc` (also usage examples need to be updated to emphasise that this is a two-four step process source →binary→instrument→cbmc

src/goto-programs/label_function_pointer_call_sites.cpp
```diff
@@ -0,0 +1,56 @@
...
+          function_call_site_symbol.mode = goto_function_symbol_mode;
+          return function_call_site_symbol;
+        }());
+        auto const new_function_pointer =
+          goto_model.symbol_table.lookup_ref(new_symbol_name).symbol_expr();
+        auto const assign_instruction = goto_programt::make_assignment(
+          code_assignt{new_function_pointer,
+                       function_pointer_dereference.pointer()},
+          source_location);
+        goto_function.second.body.insert_before(it, assign_instruction);
```
thk123, line 59: :question: Is it safe to modify the collection we're iterating over?

src/goto-programs/label_function_pointer_call_sites.cpp
```diff
@@ -0,0 +1,56 @@
...
+        to_code_function_call(it->code).function() =
+          dereference_exprt{new_function_pointer};
+      });
```
thk123: :bulb: I find this kind of code difficult to work out what it is doing as it involves a lot of straight line code. 

```C++
const auto is_function_pointer_call = [](const goto_programt::targett it) {
        return it->is_function_call() && can_cast_expr<dereference_exprt>(
                                           it->get_function_call().function());
      };
struct required_labelt
{
  symbolt new_symbol;
  goto_programt::targett function_pointer_call;
}
  

std::vector<required_labelt> fp_labels = make_range(goto_function.body.instruction)
  .filter(is_function_pointer_call)
  .map([&](goto_programt::targett function_pointer_call) {
    .. 
   const label_symbol = [&] {
          symbolt function_call_site_symbol{};
          function_call_site_symbol.name = function_call_site_symbol.base_name =
            function_call_site_symbol.pretty_name = new_symbol_name;
          function_call_site_symbol.type =
            function_pointer_dereference.pointer().type();
          function_call_site_symbol.location = function_call.source_location();
          function_call_site_symbol.is_lvalue = true;
          function_call_site_symbol.mode = goto_function_symbol_mode;
          return function_call_site_symbol;
        }();
   return required_labelt{label_symbol, function_pointer_call};).collect();

for(const auto new_label : required_labels)
{
  goto_model.symbol_table.insert(new_label.label_symbol);
  goto_function.second.body.insert_before(new_label.function_pointer_call, new_label.assign_instruction);
}
```

or something like that? It breaks up the extracting the data from the instruction, and the mutations that we need to perform. It also means your not modifying a collection your iterating over, which always makes me nervous

src/goto-programs/label_function_pointer_call_sites.cpp
```diff
@@ -0,0 +1,56 @@
...
+                                           it->get_function_call().function());
+      },
+      [&](goto_programt::targett it) {
+        auto const &function_call = it->get_function_call();
+        auto const &function_pointer_dereference =
+          to_dereference_expr(function_call.function());
+        auto const &source_location = function_call.source_location();
+        auto const &goto_function_symbol_mode =
+          goto_model.symbol_table.lookup_ref(goto_function.first).mode;
+        auto const new_symbol_name =
```
thk123: `new_symbol_name` could be better - what is the symbol for? Is it the label? `call_site_symbol_name`?

src/goto-programs/label_function_pointer_call_sites.cpp
```diff
@@ -0,0 +1,56 @@
...
+\*******************************************************************/
+
+/// \file
+/// Label function pointer call sites across a goto model
+
+#include "label_function_pointer_call_sites.h"
+
+#include <util/fresh_symbol.h>
+
+void label_function_pointer_call_sites(goto_modelt &goto_model)
```
thk123, line 13: :no_entry_sign: A comment here demonstrating exactly the transformation might be helpful

```
///  Transform goto code of function pointer calls of the form:
/// (*fp)(...)
/// Into:
/// function_call_site_symbol = fp;
/// (*function_call_site_symbol)(...)
```

src/util/irep.h
```diff
@@ -50,6 +50,11 @@ inline const std::string &id2string(const irep_idt &d)
   #endif
 }
 
+inline irep_idt string2id(const std::string &id_string)
```
thk123: :question: I believe this is already an existing implicit conversion

src/goto-programs/label_function_pointer_call_sites.h
```diff
@@ -0,0 +1,29 @@
...
+/// 2. A dereference of a symbol expression of a function pointer type
+///
+/// This makes following stages dealing with function calls easier, because they
+/// only need to be able to handle these two cases.
+///
+/// It does this by replacing all CALL instructions to function pointers with an
+/// assignment to a function pointer variable with a name following the pattern
+/// [function_name].function_pointer_call.[N], where "N" is the nth call to a
+/// function pointer in the function "function_name".
+void label_function_pointer_call_sites(goto_modelt &goto_model);
```
thk123, line 28: :pick: This looks quite stand alone - it would be easier to review if this were in a separate commit. 
:bulb: As it is so isolated, I think a unit test with a few different kinds calls that get transformed would both make the code more robust and somewhat document the transformation

src/goto-programs/restrict_function_pointers.h
```diff
@@ -0,0 +1,87 @@
...
+
+#define OPT_RESTRICT_FUNCTION_POINTER                                          \
+  "(" RESTRICT_FUNCTION_POINTER_OPT                                            \
+  "):"                                                                         \
+  "(" RESTRICT_FUNCTION_POINTER_FROM_FILE_OPT "):"
+
+#define RESTRICT_FUNCTION_POINTER_HELP                                         \
+  "--" RESTRICT_FUNCTION_POINTER_OPT                                           \
+  " <pointer_name>/<target[,targets]*>\n"                                      \
+  "           restrict a function pointer to a set of possible targets\n"      \
```
thk123, line 41: This would definitely benefit with some examples in some documentation somewhere

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+          }
+          return irep_idt{to_json_string(array_element).value};
+        });
+      return possible_targets;
+    }());
+  }
+  return function_pointer_restrictionst{restrictions};
+}
+
+void function_pointer_restrictionst::write_to_file(
```
thk123: :bulb: if this returned a json object then it would be possible to unit test this

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+    message_handler);
+  return {merge_function_pointer_restrictions(
+    file_restrictions, commandline_restrictions)};
+}
+
+function_pointer_restrictionst function_pointer_restrictionst::read_from_file(
+  const std::string &filename,
+  message_handlert &message_handler)
+{
+  auto failed = [](bool failFlag) { return failFlag; };
```
thk123: ~:question:~ :pick:  This name seems a little misleading - is this not just the identity function?

thk123: Ah I see what is being done with this... A comment might not be an awful idea, but not blocking

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+  if(failed(parse_json(inFile, filename, message_handler, json)))
+  {
+    throw system_exceptiont{
+      "failed to read function pointer restrictions from " + filename};
+  }
+  if(!json.is_object())
+  {
+    throw system_exceptiont{"top level item is not an object"};
+  }
+  for(auto const &kv_pair : to_json_object(json))
```
thk123: :bulb: Better naming: `read_restriction`?

thk123: :bulb: Ditto - if this read a json_objectt it would be easier to unit test

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+    auto const restrictions =
+      function_pointer_restrictionst::read_from_file(filename, message_handler);
+    merged_restrictions = merge_function_pointer_restrictions(
+      merged_restrictions, restrictions.restrictions);
+  }
+  return merged_restrictions;
+}
+} // namespace
+
+function_pointer_restrictionst function_pointer_restrictionst::from_options(
```
thk123, line 357: :bulb: A great candidate for unit tests (which would document expected format for free)

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+#include <ansi-c/expr2c.h>
+#include <json/json_parser.h>
+#include <util/expr_iterator.h>
+#include <util/string_utils.h>
+
+#include <algorithm>
+#include <fstream>
+#include <iostream>
+
+namespace
```
thk123, line 21: :question: What is the purpose of this `namespace`?

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+parse_function_pointer_restrictions_from_command_line(
+  const std::list<std::string> &restriction_opts)
+{
+  auto function_pointer_restrictions =
+    function_pointer_restrictionst::restrictionst{};
+  for(auto const &restriction_opt : restriction_opts)
+  {
+    // the format for restrictions is <pointer_name>/<target[,more_targets]*>
+    // exactly one pointer and at least one target
+    auto const pointer_name_end = restriction_opt.find('/');
```
thk123: :question: Could regex be used here? Rather than manually parsing the string

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+      }
+    }
+    if(!function_pointer_restrictions
+          .emplace(irep_idt{pointer_name}, target_names)
+          .second)
+    {
+      throw invalid_command_line_argument_exceptiont{
+        "function pointer restriction for `" + pointer_name +
+          "' was specified twice",
+        "--" RESTRICT_FUNCTION_POINTER_OPT};
```
thk123, line 268: :no_entry_sign: Pull out into a function that takes a string and returns a data structure containing pointer_name, target_names and unit test to cover all the awkward cases here (before swapping it to use regex if you do choose to do that)

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+    cmdline.get_values(RESTRICT_FUNCTION_POINTER_OPT));
+  options.set_option(
+    RESTRICT_FUNCTION_POINTER_FROM_FILE_OPT,
+    cmdline.get_values(RESTRICT_FUNCTION_POINTER_FROM_FILE_OPT));
+}
+
+namespace
+{
+function_pointer_restrictionst::restrictionst
+merge_function_pointer_restrictions(
```
thk123: might be worth a quick unit test for this

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+  const function_pointer_restrictionst::restrictionst &lhs,
+  const function_pointer_restrictionst::restrictionst &rhs)
+{
+  auto result = lhs;
```
thk123: :pick: Since you're immediately copying it, you could change `lhs` to be taken by value. That way if the user is doing

```
x = merge(x, y)
```
Then can do
```
x = merge(std::move(x), y)
```
and save a copy. 

It is also one less line of code!

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+void restrict_function_pointers(
+  goto_modelt &goto_model,
+  const function_pointer_restrictionst &restrictions)
+{
+  typecheck_function_pointer_restrictions(goto_model, restrictions);
+
+  for(auto &goto_function : goto_model.goto_functions.function_map)
+  {
+    auto &goto_function_body = goto_function.second.body;
```
thk123: This gets pretty indented - consider breakin gout the body of the loop into a function that deals with bodies to save one level of indentiation

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+              auto const replaced_instruction_location =
+                goto_function.second.body.insert_before(
+                  goto_end_if_location, new_instruction);
+              else_location = goto_function.second.body.insert_before(
+                replaced_instruction_location,
+                goto_programt::make_goto(
+                  else_location,
+                  notequal_exprt{
+                    pointer_symbol,
+                    address_of_exprt{function_pointer_target_symbol_expr}}));
```
thk123: Hard to read, suggestions to improve:

 - pull out bodies of loops to reduce indentation
 - use make_range / filter rather than lots of nested ifs
 - invert the ifs with a `continue` to reduce indentation
 - use smaller functions rather than comments to explain blocks of code

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+    // restrictions for, and if so branch on its value and insert concrete calls
+    // to the restriction functions
+    for_each_function_call(
+      goto_function.second, [&](const goto_programt::targett location) {
+        // Check if this is calling a function pointer, and if so if it is one
+        // we have a restriction for
+        const auto &original_function_call = location->get_function_call();
+        if(can_cast_expr<dereference_exprt>(original_function_call.function()))
+        {
+          // because we simplify before this stage a dereference can only
```
thk123: Probably worth an INVARIANT here incase someone removes the turning into canoncical form step (and ideally the comment would reference the simplification you are using. 

I also wouldn't call it `simplify` since that is a very specific thing.Instead ideally name that transformation `normalize_function_pointer_calls`? (Since we normalize function pointer calls)

thk123: Or using your existing words `since we already label function pointer call sites`

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+          auto const &called_function_pointer =
+            to_dereference_expr(original_function_call.function()).pointer();
+          PRECONDITION(can_cast_expr<symbol_exprt>(called_function_pointer));
+          auto const &pointer_symbol = to_symbol_expr(called_function_pointer);
+          auto const restriction_iterator =
+            restrictions.restrictions.find(pointer_symbol.get_identifier());
+          if(restriction_iterator != restrictions.restrictions.end())
+          {
+            auto const &restriction = *restriction_iterator;
+            // if we can, we will replace uses of it by the
```
thk123: :question: This comment seems to stop mid sentence?

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+            *location = goto_programt::make_assertion(
+              false_exprt{},
+              make_function_pointer_restriction_assertion_source_location(
+                original_function_call_instruction.source_location,
+                restriction));
+            auto const assume_false_location = goto_function_body.insert_after(
+              location,
+              goto_programt::make_assumption(
+                false_exprt{},
+                original_function_call_instruction.source_location));
```
thk123: :question: A why comment here would be helpful

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+            // this is mutable because we'll update this at the end of each
+            // loop iteration to always point at the start of the branch
+            // we created
+            auto else_location = location;
+            auto const end_if_location = goto_function_body.insert_after(
+              assume_false_location, goto_programt::make_skip());
+            for(auto const &restriction_target : restriction.second)
+            {
+              auto new_instruction = original_function_call_instruction;
```
thk123: :bulb: This definitively should be pulled out into a separate function, independently unit tested to get an idea of what goto code it produces.

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+    }
+    comment << ']';
+  }
+  source_location.set_comment(string2id(comment.str()));
+  source_location.set_property_class(ID_assertion);
+  return source_location;
+}
+
+template <typename Handler>
+void for_each_function_call(goto_functiont &goto_function, Handler handler)
```
thk123, line 108: :pick: I'd much prefer using the range framework rather than introducing these helper functions

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+      if(!first)
+      {
+        comment << ", ";
+      }
+      else
+      {
+        first = false;
+      }
+      comment << target;
+    }
```
thk123: :idea: `join_strings` handles this for you
```suggestion
    join_strings(comment, restrictions.second.begin(), restrictions.second.end(), ',');
```

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+#include <algorithm>
+#include <fstream>
+#include <iostream>
+
+namespace
+{
+void typecheck_function_pointer_restrictions(
+  const goto_modelt &goto_model,
+  const function_pointer_restrictionst &restrictions)
```
thk123, line 25: :green_heart: awesome checks - this is exactly the kind of things that would be super frustrating from user perspective to only find out how letting the analysis run!
Be great to have a unit test for each of the error checks here

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,415 @@
+/*******************************************************************\
```
thk123, line 1: This file is pretty big - I'd be tempted to split up the command line option parsing / error checking from the goto code transformation stuff

thk123 (commented):
On the whole - I think this is a really good PR, lots of clear names, functions do what they say. It is easy to see how it all plugs together. Also a big fan of what seems to be some very robust user error handling :+1: 

Got a bunch of comments that basically come in two flavours:
 - lots of opportunities for unit tests which would give much greater confidence on the edge cases in the code and saves 10s of regression tests! 
 - some large blocks of code that are quite intimidating to start reading that I think a few simple refactors could greatly improve readability

I haven't yet reviewed the regression tests or documentation, but thought I'd leave the code comments so you could begin incorporating the ones you agree with

src/util/irep.h
thk123: This gets removed in later commit

regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.desc
```diff
@@ -0,0 +1,7 @@
+CORE
+test.c
+--restrict-function-pointer "use_fg.function_pointer_call.1/f,g"
+\[use_fg.assertion.2\] line \d+ assertion \(choice \? fptr : gptr\)\(10\) == 10 \+ choice: SUCCESS
+\[use_fg.assertion.1\] line \d+ dereferenced function pointer at use_fg.function_pointer_call.1 must be one of \[(f|g), (f|g)\]: SUCCESS
+^EXIT=0$
+^SIGNAL=0$
```
thk123, line 7: :no_entry_sign: Please add explanations for each test into the desc file

doc/cprover-manual/restrict-function-pointer.md
```diff
@@ -0,0 +1,180 @@
...
+through function pointers by case distinctions over calls to given sets of
+functions.
+
+### Motivation
+
+The CPROVER framework includes a goto program transformation pass
+`remove_function_pointers()` to resolve calls to function pointers to direct
+function calls. The pass is needed by `cbmc`, as symbolic execution itself can't
+handle calls to function pointers. In practice, the transformation pass works as
+follows:
```
thk123, line 15: :bulb: this is kind of an impl detail - and as such I'd be tempted to bury it at the _bottom_ of the docs

regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.desc
```diff
@@ -0,0 +1,7 @@
+CORE
+test.c
+--restrict-function-pointer "use_fg.function_pointer_call.1/f,g"
```
thk123, line 3: :question: What is the point of `function_pointer_call`, why isn't the syntax just `use_fg.1/f,g`?
:question: Can I specify more all calls? e.g. `use_fg/f,g` for all calls inside `use_fg`?

regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.c
```diff
@@ -0,0 +1,59 @@
...
+{
+  assert((choice ? fptr : gptr)(10) == 10 + choice);
+}
+
+// this is just here to distinguish the behavior from FP removal, which'd include g
+int g_always_false_cond = 0;
+
+int main(void)
+{
+  use_fg(0, get_f(), get_g());
```
thk123, line 15: I'm a bit wary that constant propogation will make this example work for free? Consider using a nondet int for the `choice` param

regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.c
```diff
@@ -0,0 +1,59 @@
...
+
+fptr_t get_f(void);
+fptr_t get_g(void);
+
+void use_fg(int choice, fptr_t fptr, fptr_t gptr)
+{
+  assert((choice ? fptr : gptr)(10) == 10 + choice);
+}
+
+// this is just here to distinguish the behavior from FP removal, which'd include g
```
thk123: :question: I don't understand this comment, or this variable

regression/goto-instrument/restrict-function-pointer-to-multiple-functions-code-check/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+test.c
+--restrict-function-pointer use_f.function_pointer_call.1/f,g --show-goto-functions
+f\(10\)
```
thk123, line 4: :green_heart: this test seems a precise approach

regression/goto-instrument/restrict-function-pointer-to-multiple-functions-via-file-and-command-line-options/test.desc
```diff
@@ -0,0 +1,6 @@
+CORE
+test.c
+--function-pointer-restrictions-file restrictions.json --restrict-function-pointer use_f.function_pointer_call.1/g
```
thk123, line 3: :question: Wouldn't this test work if it was ignoring the command line arguments too?

thk123 (commented):
It is hard to see exactly what the tests are testing, I think a short sentence at the foot of each desc describing it's purpose, so appologies if there are tests for the following:

 - writing to a json file (I saw the code - not sure how it is to be used in this PR)
 - multiple distinct function pointer calls replaced
 - a function pointer call that _isn't_ replaced
 - multiple calls to the same function pointer being replaced by different fps
 - parsing a mismatched type (though think we should test all of the error checks fully through unit tests - this would just be to check the validation code is indeed being used)

regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.desc
hannes-steffenhagen-diffblue: 1. This could be shortened, yes.
2. No, but something similar is in a follow up PR

regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.c
hannes-steffenhagen-diffblue: Argh, I swear I had a version of this that fixed this. The reason this is here is to trick function pointer removal (the old one) basically.

regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.c
hannes-steffenhagen-diffblue: The FP removal here happens before symex, so if constant propagation works for this we did our job right.

regression/goto-instrument/restrict-function-pointer-to-multiple-functions-code-check/test.desc
hannes-steffenhagen-diffblue: What does the green heart mean?

regression/goto-instrument/restrict-function-pointer-to-multiple-functions-via-file-and-command-line-options/test.desc
hannes-steffenhagen-diffblue: This is testing the error message we get from this pass, if the command line arguments were ignored you’d get a different error message or none at all.

src/goto-programs/label_function_pointer_call_sites.cpp
hannes-steffenhagen-diffblue: This documentation is in the header.

src/goto-programs/label_function_pointer_call_sites.cpp
hannes-steffenhagen-diffblue: Not in general, but goto function bodies are `std::list` which is safe to modify while iterating. We have more code that assumes this. May be worth putting a comment here that says as much though.

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: This is just to hide implementation-specific symbols, same as writing `static` in front of them. I just hide functions that don’t need to be global from the global namespace out of habit.

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: I didn’t know we had that

src/goto-programs/restrict_function_pointers.cpp
hannes-steffenhagen-diffblue: I didn’t even know we had that in cbmc!

src/goto-programs/restrict_function_pointers.h
hannes-steffenhagen-diffblue: I believe this is explained in the cprover-manual

hannes-steffenhagen-diffblue (commented):
Addressing some of Thomas’s comments.

smowton:
Looks like there's a ton of comments to look at-- let me know when this / its successor PR are in final-ish state and waiting on code owners?

src/goto-programs/label_function_pointer_call_sites.h
danpoe: Put it into a separate commit and added unit tests as well.

src/goto-programs/restrict_function_pointers.cpp
danpoe: We now have functions `to_json()` and `from_json()`, and `write_to_file()` and `read_from_file()` use those. I've also added unit tests for the former two.

src/goto-programs/restrict_function_pointers.cpp
danpoe: Removed this lambda.

src/goto-programs/restrict_function_pointers.cpp
danpoe: Will defer this to the next PR which changes `from_options()`.

src/goto-programs/restrict_function_pointers.cpp
danpoe: Done

src/goto-programs/restrict_function_pointers.cpp
danpoe: Added

src/goto-programs/restrict_function_pointers.cpp
danpoe: Done

src/goto-programs/restrict_function_pointers.cpp
danpoe: I would prefer those functions actually. Kept them.

src/goto-programs/restrict_function_pointers.cpp
danpoe: 450 lines is still within reasonable bounds IMO

regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.desc
danpoe: Done

regression/goto-instrument/restrict-function-pointer-to-complex-expression/test.desc
danpoe: Will defer 1 to the next PR.

src/goto-programs/restrict_function_pointers.cpp
danpoe: It could be, though it probably wouldn't improve things much. The current version has the advantage that specific error messages are printed for different format errors.

src/goto-programs/restrict_function_pointers.cpp
danpoe: Moved the body of the loop into a separate function

src/goto-programs/restrict_function_pointers.cpp
danpoe: Inverted the ifs :green_heart: and put the body of the loop into a separate function

doc/cprover-manual/restrict-function-pointer.md
danpoe: Will leave refactoring the documentation to the next PR as it changes some things.

src/goto-programs/restrict_function_pointers.cpp
danpoe: Put this into a separate function. I think it is not necessary to add unit tests as this functionality is very well tested by the regression tests already.

src/goto-programs/label_function_pointer_call_sites.cpp
danpoe: I think the original version is fine. I've added more visual breaks to make it easier to read.

src/goto-programs/label_function_pointer_call_sites.cpp
```diff
@@ -0,0 +1,65 @@
...
+      {
+        auto const &function_call = it->get_function_call();
+        auto const &function_pointer_dereference =
+          to_dereference_expr(function_call.function());
+        auto const &source_location = function_call.source_location();
+        auto const &goto_function_symbol_mode =
+          goto_model.symbol_table.lookup_ref(goto_function.first).mode;
+
+        auto const call_site_symbol_name =
+          irep_idt{id2string(goto_function.first) + ".function_pointer_call." +
```
smowton, line 33: May be best not to use `.` since it delimits members, and this isn't one ($ has been used elsewhere for a reserved character that doesn't mean anything in C)

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -0,0 +1,478 @@
...
+{
+  typecheck_function_pointer_restrictions(goto_model, restrictions);
+
+  for(auto &function_item : goto_model.goto_functions.function_map)
+  {
+    goto_functiont &goto_function = function_item.second;
+
+    for_each_function_call(
+      goto_function,
+      std::bind(
```
smowton: This is one verbose way to write
```
[&](const goto_programt::targett &location) { handle_call(goto_function, goto_model, restrictions, location); }
```

smowton (approved):
Only skimmed the code, didn't review the docs or tests

src/goto-programs/label_function_pointer_call_sites.cpp
danpoe: Agreed. This will change in the next PR anyways so I'll leave this as is for now.

PR #5258, opened 2020-03-05
Avoids arithmetic overflow on signed to unsigned type conversions

feliperodri:

**TL;DR:** The source of the problem is that `ch1` and `ch2` are declared as `unsigned char`; however, `\xff` doesn't work correctly with `signed char` (see the test case below). Thus, we can either fix that (special) case or use `unsigned char` and add `#pragma CPROVER check disable "conversion"`. For me, the latter makes more sense because the [standard for these functions](https://pubs.opengroup.org/onlinepubs/7908799/xsh/strncmp.html) interpreted these characters as type `unsigned char`.

------------------------------------------------------------------------------------------------

Running CBMC on the following example from `regression/cbmc-library/String6` using the `conversion-check` flag reported an arithmetic overflow:
```
cbmc main.c --pointer-check --bounds-check --conversion-check
```

```
#include <assert.h>
#include <stdlib.h>
#include <string.h>

int main()
{
  char str[500] = "Hello";

  assert(strcmp(str, "Hello") == 0);
  assert(strncmp(str, "Hello", 5) == 0);

#ifndef _MSC_VER
  assert(strcasecmp(str, "HELLO") == 0);
  assert(strncasecmp(str, "HELLO", 5) == 0);
#endif

  assert(strcmp(str, "\xff") < 0);
  assert(strncmp("ASDxx", "ASDyy", 3) == 0);

  assert(strlen(str) == 5);
  char *str_cpy = strdup(str);
  assert(strcmp(str, str_cpy) == 0);
  free(str_cpy);

  return 0;
}
```

```
Violated property:
  file <builtin-library-strncmp> function strncmp line 26 thread 0
  arithmetic overflow on signed to unsigned type conversion
  in (unsigned char)s1[(signed long int)i]
  s1[(signed long int)i] >= 0
```

Adding pragmas to avoid spurious reports and to keep in sync with the [standard](https://pubs.opengroup.org/onlinepubs/7908799/xsh/strncmp.html),i.e., _"the sign of a non-zero return value is determined by the sign of the difference between the values of the first pair of bytes (**both interpreted as type unsigned char**) that differ in the strings being compared."_

### DIFF
--- a/src/ansi-c/library/string.c
+++ b/src/ansi-c/library/string.c
@@ -357,7 +357,7 @@ __CPROVER_HIDE:;
 
 inline int strcmp(const char *s1, const char *s2)
 {
-  __CPROVER_HIDE:;
+__CPROVER_HIDE:;
 #ifdef __CPROVER_STRING_ABSTRACTION
   int retval;
   __CPROVER_precondition(__CPROVER_is_zero_string(s1),
@@ -406,7 +406,7 @@ inline int strcmp(const char *s1, const char *s2)
 
 inline int strcasecmp(const char *s1, const char *s2)
 {
-  __CPROVER_HIDE:;
+__CPROVER_HIDE:;
 #ifdef __CPROVER_STRING_ABSTRACTION
   int retval;
   __CPROVER_precondition(__CPROVER_is_zero_string(s1),
@@ -458,7 +458,7 @@ inline int strcasecmp(const char *s1, const char *s2)
 
 inline int strncmp(const char *s1, const char *s2, size_t n)
 {
-  __CPROVER_HIDE:;
+__CPROVER_HIDE:;
 #ifdef __CPROVER_STRING_ABSTRACTION
   __CPROVER_precondition(__CPROVER_is_zero_string(s1) ||
                          __CPROVER_buffer_size(s1)>=n,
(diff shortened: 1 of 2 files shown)


src/ansi-c/library/string.c
```diff
@@ -389,7 +392,7 @@ inline int strcmp(const char *s1, const char *s2)
   }
   while(ch1!=0 && ch2!=0);
   return 0;
-  #endif
+#endif
```
tautschnig, line 395: Can you please also adjust the matching `#if` to avoid a weirdly looking source?

tautschnig (commented):
Looks good to me, with two requests for additional changes:
1) Please include a proper commit message body summarising _how_ you avoid the overflow (well, in fact the commit message subject may need to change: you are no longer avoiding it, just not reporting a failing assertion).
2) Please include a test case.

src/ansi-c/library/string.c
feliperodri: Done.

feliperodri:
> Looks good to me, with two requests for additional changes:
> 
>     1. Please include a proper commit message body summarising _how_ you avoid the overflow (well, in fact the commit message subject may need to change: you are no longer avoiding it, just not reporting a failing assertion).
> 
>     2. Please include a test case.

I didn't include a test case, but I added the conversion-check on the existing test case, so we ensure CBMC does not report the arithmetic overflows. Is that enough? @tautschnig

tautschnig (approved):

feliperodri:
@tautschnig ready to merge.

PR #5259, opened 2020-03-05
SMT Switch API

thk123:
Adds the [smt-switch](https://github.com/makaimann/smt-switch) dependency if you build with `cmake and -DSMT_SWITCH=ON -DSMT_CVC4=ON`. Currently not used, outside of a unit test to prove compiling correctly. 

There are still a few things I need to do
 - [x] ~Need to add to `make` build as well I suppose.~
 - [x] Update documentation for how to enable the build
 - [x] Enable the codebuild job

But at least the CMake changes can be reviewed 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS true)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 10.9)
 
 include(GNUInstallDirs)
+include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/DownloadProject.cmake")
 
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
@@ -30,7 +31,7 @@ elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     # This would be the place to enable warnings for Windows builds, although
     # config.inc doesn't seem to do that currently
 
-    # Include Git Bash Environment (rqeuired for download_project (patch))
+    # Include Git Bash Environment (required for download_project (patch))
     find_package(Git)
     if(GIT_FOUND)
         get_filename_component(git_root ${GIT_EXECUTABLE} DIRECTORY)
@@ -50,8 +51,17 @@ if(${enable_cbmc_tests})
     enable_testing()
 endif()
 
+option(use_smt_switch "use generic smt interface")
+if (use_smt_switch)
+    download_project(
+            PROJ smt-switch
+            URL https://github.com/makaimann/smt-switch/archive/hwmcc19.tar.gz
+            URL_MD5 c3c40be8b1294017cd129b9a3d99ff30
+    )
+    add_subdirectory(${smt-switch_SOURCE_DIR})
+endif ()
+
 if(DEFINED CMAKE_USE_CUDD)
-    include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/DownloadProject.cmake")
     message(STATUS "Downloading Cudd-3.0.0")
     download_project(PROJ cudd
             URL https://sourceforge.net/projects/cudd-mirror/files/cudd-3.0.0.tar.gz/download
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -51,8 +52,22 @@ if(${enable_cbmc_tests})
     enable_testing()
 endif()
 
+option(use_smt_switch "use generic smt interface")
+if(use_smt_switch)
+  download_project(
+    PROJ smt-switch
+    GIT_REPOSITORY  https://github.com/thk123/smt-switch.git
+    GIT_TAG patch-1
```
thk123: This resolves a `-pedantic` error that is currently got by downloading the master or last release version. I have a PR to fix this in trunk: makaimann/smt-switch#43

If that doesn't get merged soon, the alternative would be to disable `-pedantic` for this project (hopefully @hannes-steffenhagen-diffblue knows a way to do this)

CMakeLists.txt
```diff
@@ -51,8 +52,22 @@ if(${enable_cbmc_tests})
...
+    )
+  add_subdirectory(${smt-switch_SOURCE_DIR})
+  add_compile_options(-DHAVE_SMT_SWITCH)
+#  if(BUILD_CVC4)
+    execute_process(COMMAND ./contrib/setup-cvc4.sh WORKING_DIRECTORY ${smt-switch_SOURCE_DIR})
+#  endif()
```
thk123: I will add sections for each of the supported modules here, but might be worth having a think about whether we want one on by default, if you use smt-switch).

unit/solvers/smt_switch.cpp
```diff
@@ -0,0 +1,38 @@
...
+// only enable the tests if compiling with SMT switch
+#ifdef HAVE_SMT_SWITCH
+#  include <cvc4_factory.h>
+#  include <smt.h>
+
+#endif
+
+#include <vector>
+
+TEST_CASE("Trivial SMT switch integration", "[core]")
```
thk123: :no_entry_sign: not appropriately tagged

thk123:
Can't add make since smt-switch can't be built with make anyway.

CMakeLists.txt
```diff
@@ -136,6 +163,12 @@ function(cprover_default_properties)
     set(CBMC_XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY
             "Developer ID Application: Daniel Kroening")
 
+    foreach (target IN LISTS ARGN)
+        message(${target})
+        target_compile_options(${target} PRIVATE -Wall -Wpedantic -Werror -Wno-deprecated-declarations -Wswitch-enum)
```
hannes-steffenhagen-diffblue, line 168: For other reviewers: This means we only set our warning/werror flags for targets that have `cprover_default_properties` called for them. We should do a check to make sure this is the case for all our own targets. Not setting werror stuff globally is so we can avoid having to patch dependencies just so we can avoid generating a warning.

unit/solvers/smt_switch.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+#include <testing-utils/use_catch.h>
+
+// only enable the tests if compiling with SMT switch
+#ifdef HAVE_SMT_SWITCH
+#  include <cvc4_factory.h>
```
hannes-steffenhagen-diffblue, line 13: What’s with the spaces after `#`?

thk123: This is how my clang-format wants to format this :man_shrugging:

thk123:
@peterschrammel I'm not sure if we want to merge this before it has a user. 

Advantages:
 - ensures it is tested by CI so not inadvertently broken

Downsides:
 - Code whose purpose isn't clear 

I guess if we do merge it, we should remove the travis job since that is the bottleneck on CI and it is tested on AWS codebuild anyway.

thk123:
Fixed clang-format - I believe this will now pass CI 

@peterschrammel expressed a preference for merging this, perhaps him and @hannes-steffenhagen-diffblue could you review this?

thk123:
Have manually confirmed that the smt switch unit test passes in CI (does not appear in the xfail step), but is in the xfail for non-smt build.

NlightNFotis:
What's the status on this one? Is this still going in, or is this now going to be closed?

thk123:
Given we're not pushing this forward I'll just close this PR for now.

PR #5174, opened 2019-11-06
Feature/restrict function pointer by name [depends-on: #5168]

hannes-steffenhagen-diffblue:
This works similar to restrict-function-pointer, but for names of individual
function pointer variables (globals, locals, parameters) rather than call sites.
This isn't applicable to all situations (for example, calling function pointers
in structs or function pointers returned from functions), but is more readily
applicable to some common use scenarios (e.g. global function pointers loaded at
start time like in OpenGL).

- [ ] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #5262, opened 2020-03-07
Restrict function pointers by name

danpoe:
Replaces #5174.


src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -258,21 +271,21 @@ parse_function_pointer_restrictions_from_command_line(
...
-        "--" RESTRICT_FUNCTION_POINTER_OPT,
+        option,
         restriction_format_message};
     }
     if(pointer_name_end == 0)
     {
       throw invalid_command_line_argument_exceptiont{
         "couldn't find target name before '/' in `" + restriction_opt + "'",
-        "--" RESTRICT_FUNCTION_POINTER_OPT};
+        option_name};
```
hannes-steffenhagen-diffblue: :no_entry_sign: `option`

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -328,22 +349,83 @@ parse_function_pointer_restrictions_from_file(
   }
   return merged_restrictions;
 }
+
+function_pointer_restrictionst::restrictionst
```
hannes-steffenhagen-diffblue: The purpose of this needs to be documented, I am fairly sure I wrote this and it took me like 5 minutes trying to figure out what this was doing and why

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1037,13 +1037,20 @@ void goto_instrument_parse_optionst::instrument_goto_program()
...
+      !options.get_list_option(RESTRICT_FUNCTION_POINTER_FROM_FILE_OPT).empty())
     {
       label_function_pointer_call_sites(goto_model);
+
+      auto const by_name_restrictions =
+        get_function_pointer_by_name_restrictions(goto_model, options);
+
+      const auto function_pointer_restrictions =
+        by_name_restrictions.merge(function_pointer_restrictionst::from_options(
+          options, log.get_message_handler()));
```
thk123:
```suggestion
      const auto by_callsite_restrictions = function_pointer_restrictionst::from_options(options, log.get_message_handler());
      const auto function_pointer_restrictions =  by_name_restrictions.merge(by_callsite_restrictions);
```

Just to make it clear what the two things that are being merged are.

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -459,3 +466,61 @@ void function_pointer_restrictionst::write_to_file(
 
   function_pointer_restrictions_json.output(outFile);
 }
+function_pointer_restrictionst function_pointer_restrictionst::merge(
```
thk123: :bulb: Given that is const, I think a free function that takes two and returns result would be clearer

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -104,12 +104,13 @@ source_locationt make_function_pointer_restriction_assertion_source_location(
   return source_location;
 }
 
-template <typename Handler>
-void for_each_function_call(goto_functiont &goto_function, Handler handler)
+template <typename Handler, typename GotoFunctionT>
+void for_each_function_call(GotoFunctionT &&goto_function, Handler handler)
```
thk123, line 53: :question: Why does this become generic?

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -244,9 +248,11 @@ function_pointer_restrictionst::merge_function_pointer_restrictions(
...
 }
 
-function_pointer_restrictionst::restrictionst function_pointer_restrictionst::
-  parse_function_pointer_restrictions_from_command_line(
-    const std::list<std::string> &restriction_opts)
+function_pointer_restrictionst::restrictionst
+function_pointer_restrictionst::
+parse_function_pointer_restrictions_from_command_line(
+  const std::list<std::string> &restriction_opts,
+  const std::string &option_name)
```
thk123: :no_entry_sign: This variable doesn't appear to be used

thk123: I guess it should be used on line 274

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -459,3 +466,61 @@ void function_pointer_restrictionst::write_to_file(
...
+function_pointer_restrictionst function_pointer_restrictionst::merge(
+  const function_pointer_restrictionst &other) const
+{
+  return function_pointer_restrictionst{
+    merge_function_pointer_restrictions(restrictions, other.restrictions)};
+}
+
+function_pointer_restrictionst get_function_pointer_by_name_restrictions(
+  const goto_modelt &goto_model,
+  const optionst &options)
```
thk123: :bulb: I think it would be easier to see what is going on if you have this take the result of ` options.get_list_option(RESTRICT_FUNCTION_POINTER_BY_NAME_OPT)` to decouple the CLI from the parsing of the lsit

src/goto-programs/restrict_function_pointers.cpp
thk123: Though I think it would be better to have this throw an exception `invalid_restrictionst` or something, and have the callsite catch it and throw it as a command line exception, as this code doesn't need to know that it is coming from the command line or whatevre

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -459,3 +466,61 @@ void function_pointer_restrictionst::write_to_file(
...
+      parse_function_pointer_restrictions_from_command_line(
+        options.get_list_option(RESTRICT_FUNCTION_POINTER_BY_NAME_OPT),
+        RESTRICT_FUNCTION_POINTER_BY_NAME_OPT);
+  function_pointer_restrictionst::restrictionst restrictions;
+  for(auto const &goto_function : goto_model.goto_functions.function_map)
+  {
+    for_each_function_call(
+      goto_function.second, [&](goto_programt::const_targett location) {
+        PRECONDITION(location->is_function_call());
+        if(can_cast_expr<dereference_exprt>(
```
thk123: :bulb: Use early return to avoid nesting
:bulb: :bulb: Use make_range & filter to express this as a series of transformations which breaks down the loop into understandable chunks

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -459,3 +466,61 @@ void function_pointer_restrictionst::write_to_file(
...
+             location->get_function_call().function()))
+        {
+          PRECONDITION(can_cast_expr<symbol_exprt>(
+            to_dereference_expr(location->get_function_call().function())
+              .pointer()));
+          auto const &function_pointer_call_site = to_symbol_expr(
+            to_dereference_expr(location->get_function_call().function())
+              .pointer());
+          auto const &body = goto_function.second.body;
```
thk123: This block of code is hard for me to understand - perhaps some named helper functions would make it clear what is going on here

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -459,3 +466,61 @@ void function_pointer_restrictionst::write_to_file(
...
+          for(auto it = std::prev(location); it != body.instructions.end();
```
thk123: :question: Wait what? We start before the location, and then walk forward looking for an assignment? Shouldn't we be going backwards?

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1038,18 +1038,15 @@ void goto_instrument_parse_optionst::instrument_goto_program()
     parse_function_pointer_restriction_options_from_cmdline(cmdline, options);
 
     if(
-      !options.get_list_option(RESTRICT_FUNCTION_POINTER_OPT).empty() ||
```
thk123: :pick: please don't but a bunch of unrelated refactors into the same commit as it makes it harder to review

thk123: And means you can't explain each one in the commit message

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -193,6 +198,51 @@ void handle_call(
                        address_of_exprt{function_pointer_target_symbol_expr}}));
   }
 }
+
+void get_by_name_restriction(
```
thk123: :no_entry_sign: I think it is suprising that a function called `get_...` doesn't return anything, can this be structured in a way so that it actually does get something? Otherwise, a comment would help

thk123: Yeah I think it'd be much clearer to return a new restrictions (and append at the callsite if required)

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -193,6 +198,51 @@ void handle_call(
...
+  PRECONDITION(location->is_function_call());
+
+  const exprt &function = location->get_function_call().function();
+
+  if(!can_cast_expr<dereference_exprt>(function))
+    return;
+
+  auto const &function_pointer_call_site = to_symbol_expr(
+    to_dereference_expr(function)
+      .pointer());
```
thk123: :question: Is it guaranteed that the `pointer()` result will be a symbol expr? I think it either needs a PRECONDITION or an early return as this isn't obvious to me

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -193,6 +198,51 @@ void handle_call(
...
+      by_name_restrictions.find(rhs.get_identifier());
+
+    if(
+      restriction != by_name_restrictions.end() &&
+      restriction->first == rhs.get_identifier())
+    {
+      restrictions.emplace(
+        function_pointer_call_site.get_identifier(),
+        restriction->second);
+    }
```
thk123: My understanding is if we have

```
(*fp)()
```

Then we look for lines that look like
```
fp = <something>
```

On the proceeding line, or any line after.

I still think we should be searching backwards, but this still seems off. What if the call is on the actual function we have a named restriction for? I realize from docs that this is a simple case data flow analysis, but is there not already some code that does a data flow analysis that provides this?

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -24,6 +24,11 @@ void typecheck_function_pointer_restrictions(
   const goto_modelt &goto_model,
   const function_pointer_restrictionst &restrictions)
 {
+  const std::string options =
+    "--" RESTRICT_FUNCTION_POINTER_OPT "/"
+    "--" RESTRICT_FUNCTION_POINTER_FROM_FILE_OPT "/"
+    "--" RESTRICT_FUNCTION_POINTER_BY_NAME_OPT;
```
thk123: :question: Is it not possible to resitrct by name from file?

thk123: Since this is constant - perhaps worth declaring as global constant and then not having to pass into each of the functions

thk123 (commented):
Will review tests after lunch. 

Note to other reviews: it is easier to review all at once rather than by commit

regression/goto-instrument/restrict-function-pointer-by-name-global/test.c
```diff
@@ -0,0 +1,28 @@
...
+}
+
+typedef int (*fp_t)(void);
+
+fp_t fp;
+
+void main()
+{
+  int cond;
+  fp = cond ? f : g;
```
thk123, line 23: :question: In my experience, cbmc can be smart enough to spot that `h` is unused, I guess the most precise way to test this is fixed, is have a `test-no-restriction.desc` that is the same, but check that the verification fails on the assertion :sheep:

regression/goto-instrument/restrict-function-pointer-by-name-local/test.c
```diff
@@ -0,0 +1,19 @@
...
+int g(void)
+{
+  return 2;
+}
+
+typedef int (*fp_t)(void);
+
+void main()
+{
+  fp_t fp = f;
```
thk123, line 17: :sheep:

regression/goto-instrument/restrict-function-pointer-to-multiple-functions-via-file-and-command-line-options/test.c
```diff
@@ -35,38 +19,19 @@ int g(int x)
...
+int main(void)
 {
-  select_function = 2;
+  use_f(f);
+  use_f(g);
+  use_f(h);
+  use_f(other);
 }
 
-fptr_t get_f(void)
```
thk123, line 58: I guess all this deletion would be clearer if put in its own commit explaining why

regression/goto-instrument/restrict-function-pointer-by-name-local/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+test.c
+--restrict-function-pointer-by-name main::1::fp/f
+\[main\.assertion\.1\] line \d+ dereferenced function pointer at main\.function_pointer_call\.1 must be f: SUCCESS
+\[main\.assertion\.2\] line \d+ assertion fp\(\) == 1: SUCCESS
+^EXIT=0$
+^SIGNAL=0$
+--
+--
```
thk123, line 11: :no_entry_sign: please make sure _all_ tests have an explanation of what they are supposed to be testing. The name is necessarily terse, and in 6 months we definitely will get confused.

unit/goto-programs/restrict_function_pointers.cpp
```diff
@@ -88,6 +95,67 @@ void merge_restrictions_test()
...
+      fp();
+      fp(); // second call to same function pointer
+    }
+  )";
+
+  goto_modelt goto_model = get_goto_model_from_c(code);
+
+  label_function_pointer_call_sites(goto_model);
+
+  const auto restrictions =
```
thk123: :bulb: Ideally a unit test would test less components at once (e.g. this tests both the parsing of the options, and the restriction stuff) - if you split up the parsing, and have the result of the parsing being passed into the restrictions, the test would be simpler

unit/goto-programs/restrict_function_pointers.cpp
```diff
@@ -148,3 +216,8 @@ TEST_CASE("Json conversion", "[core]")
     function_pointer_restrictions1.restrictions ==
     function_pointer_restrictions2.restrictions);
 }
+
+TEST_CASE("Get function pointer by name restrictions", "[core]")
```
thk123: :no_entry_sign: The test should be tagged with `[goto-programs][restrict-function-pointers]`

unit/goto-programs/restrict_function_pointers.cpp
```diff
@@ -148,3 +216,8 @@ TEST_CASE("Json conversion", "[core]")
...
+{
+  get_function_pointer_by_name_restrictions_test();
```
thk123, line 292: :grey_question: Why not just put the function in the test? Doing so makes it easier to find the test, and the CLion integration of catch unit tests will work better

thk123 (commented):
:no_entry_sign: need a test that assigns a fp on a line a few lines before the call

unit/goto-programs/restrict_function_pointers.cpp
thk123: Now I understand, I think the test would have been easier to follow if was a couple of simpler examples:


```
void func1(fp_t fp)
    {
      fp();
    }
void func1(fp_t fp)
    {
      fp();
    }

TEST_CASE("Check function scope is correctly applied to variable names")
fp_restrictionst::get_function_pointer_by_name_restrictions(
      {"func1::fp/fa", "func1::fp/fa"},
      goto_model);
 const auto set1 = restrictions.at("func.function_pointer_call.1");
REQUIRE(restrictions == {{"func2.function_pointer_call.1", "fa"}, {"func2.function_pointer_call.1", "fb"})
...
``

then have a separate test case for checking the scope thing

File: src/goto-programs/restrict_function_pointers.cpp (hunk 5 above)
Comment (danpoe) [reply src/goto-programs/restrict_function_pointers.cpp:53]: This is so that the function can take both const and non-const goto function objects.

File: src/goto-programs/restrict_function_pointers.cpp (hunk 8 above)
Comment (danpoe) [reply src/goto-programs/restrict_function_pointers.cpp]: Pulled out the loop body into a separate method and using early return.

File: src/goto-instrument/goto_instrument_parse_options.cpp (hunk 11 above)
Comment (danpoe) [reply src/goto-instrument/goto_instrument_parse_options.cpp]: Agreed. Will try to be more careful in the future.

File: src/goto-programs/restrict_function_pointers.cpp (hunk 13 above)
Comment (danpoe) [reply src/goto-programs/restrict_function_pointers.cpp]: Yes, this is guaranteed due to function pointer labelling having been run before. I think the precondition is redundant here as `to_symbol_expr()` (and all other such conversion functions) already include a precondition as their first statement.

File: src/goto-programs/restrict_function_pointers.cpp (hunk 10 above)
Comment (danpoe) [reply src/goto-programs/restrict_function_pointers.cpp]: Ah yes. The loop is unnecessary as the instruction we look for will always be at `std::prev(location)`. Have removed the loop.

File: src/goto-programs/restrict_function_pointers.cpp (hunk 12 above)
Comment (danpoe) [reply src/goto-programs/restrict_function_pointers.cpp]: The function now returns an `optionalt<restrictiont>`.

File: src/goto-programs/restrict_function_pointers.cpp (hunk 15 above)
Comment (danpoe) [reply src/goto-programs/restrict_function_pointers.cpp]: Currently that's not supported as the from file option was mostly intended for consuming the output from goto-analyzer which will refer to the calls via indices rather than names. Though it would also make sense to let a user supply manually written json files at some point.

Comment (danpoe) [reply src/goto-programs/restrict_function_pointers.cpp]: The string is gone now since the error handling has changed.

File: src/goto-programs/restrict_function_pointers.cpp (hunk 14 above)
Comment (danpoe) [reply src/goto-programs/restrict_function_pointers.cpp]: The loop was actually unnecessary as the assignment will always be on the previous line. I've removed the loop.

File: src/goto-programs/restrict_function_pointers.cpp (hunk 6 above)
Comment (danpoe) [reply src/goto-programs/restrict_function_pointers.cpp]: Good idea. I've reworked the error handling to now throw an `invalid_restriction_exceptiont`.

File: unit/goto-programs/restrict_function_pointers.cpp (hunk 22 above)
Comment (danpoe) [reply unit/goto-programs/restrict_function_pointers.cpp:292]: The test accesses some protected methods of `function_pointer_restrictionst`. So to access those we declare `get_function_pointer_by_name_restrictions_test()` as a friend (see at the top of the file).

File: unit/goto-programs/restrict_function_pointers.cpp (hunk 21 above)
Comment (danpoe) [reply unit/goto-programs/restrict_function_pointers.cpp]: Done

File: unit/goto-programs/restrict_function_pointers.cpp (hunk 20 above)
Comment (danpoe) [reply unit/goto-programs/restrict_function_pointers.cpp]: Split it up into several simpler cases.

File: regression/goto-instrument/restrict-function-pointer-by-name-local/test.desc (hunk 19 above)
Comment (danpoe) [reply regression/goto-instrument/restrict-function-pointer-by-name-local/test.desc:11]: Added

File: regression/goto-instrument/restrict-function-pointer-by-name-global/test.c (hunk 16 above)
Comment (danpoe) [reply regression/goto-instrument/restrict-function-pointer-by-name-global/test.c:23]: cbmc should always be able to figure out which functions are really called, so the verification result without using `--restrict-function-pointer` would be the same. Just the `remove_function_pointers()` pass would probably produce a superset of the functions that a user might specify via `--restrict-function-pointer`. For testing `--restrict-function-pointer`, I think it does not matter what `remove_function_pointers()` might do, as the user could also specify a superset of the functions identified by that pass.

To check that really the correct case distinction is produced, I've now also added regexes that check that precisely the correct functions are called in the transformed goto program.

File: src/goto-programs/restrict_function_pointers.cpp (hunk 5 above)
Comment (thk123) [reply src/goto-programs/restrict_function_pointers.cpp:53]: I don't follow - you can pass a non-const version of something to a function that takes a const object? I'll try undoing this change locally and see if then it clicks!

**Comment by danpoe** [comment]
@thk123 A test that assigns a function pointer a few lines before the call is not necessary. The reason we look for those assignments is because the `label_function_pointer_calls` pass inserts them, and it always does so immediately before the call. (I've also removed the loops that search for these assignments and replaced them by a check of just the previous instruction)

File: src/goto-programs/restrict_function_pointers.cpp (hunk 13 above)
Comment (thk123) [reply src/goto-programs/restrict_function_pointers.cpp]: Fair - however, if I'm debugging this and I get an error saying isn't a symbol I don't have any clue why you'd expect this, whereas if you provide a specific one, you can add a comment saying must be symbol because pointer_labelling has been run (and therefore, hitting this suggests someone has forgot to run pointer labelling).

File: src/goto-programs/restrict_function_pointers.cpp (hunk 5 above)
Comment (thk123) [reply src/goto-programs/restrict_function_pointers.cpp:53]: Oh presumably because the handler might require a non-const version.

File: src/goto-programs/restrict_function_pointers.cpp [hunk 23]
```diff
@@ -344,29 +382,112 @@ function_pointer_restrictionst::parse_function_pointer_restriction(
...
+  const goto_programt::const_targett &location)
+{
+  PRECONDITION(location->is_function_call());
+
+  const exprt &function = location->get_function_call().function();
+
+  if(!can_cast_expr<dereference_exprt>(function))
+    return {};
+
+  auto const &function_pointer_call_site = to_symbol_expr(
```
thk123: :bulb: Even a comment here:

```C++
// function pointer guarateed to be a symbol by the function-pointer-label pass
```

Where function-pointer-label should be an actual file/folder/function that can be searched for!

unit/goto-programs/restrict_function_pointers.cpp
thk123: Much clearer thanks!

thk123 (commented):
Looks good from a code perspective. 

Call me a Luddite, but even knowing the code doesn't rely on an assignment from a few lines before, I think it would be good to add the test, since not obvious to me that this is true. 

I also would feel better about the test without it having distinct regexes so we know it remains doing something

Other than that, lgtm :+1: thanks for addressing comments

smowton (approved):
Only reviewed the main changed, not the tests

thk123 (approved):
Dan pointed out that all the tests that don't have a local variable are testing what I was referring to.

PR #5265, opened 2020-03-10
Function to get a goto model from a C program for use in unit tests

danpoe:
The function takes a C program (as a string or an input stream) and converts it
to a goto model.

### DIFF
--- a/unit/testing-utils/get_goto_model_from_c.cpp
+++ b/unit/testing-utils/get_goto_model_from_c.cpp
@@ -1,0 +1,95 @@
+/*******************************************************************\
+
+Module: Get goto model
+
+Author: Daniel Poetzl
+
+\*******************************************************************/
+
+#include "get_goto_model_from_c.h"
+
+#include <ansi-c/ansi_c_language.h>
+
+#include <goto-programs/goto_convert_functions.h>
+
+#include <langapi/language_file.h>
+#include <langapi/mode.h>
+
+#include <util/cmdline.h>
+#include <util/config.h>
+#include <util/exception_utils.h>
+#include <util/invariant.h>
+#include <util/message.h>
+#include <util/options.h>
+#include <util/symbol_table.h>
+
+#include <testing-utils/message.h>
+
+goto_modelt get_goto_model_from_c(std::istream &in)
+{
+  {
+    const bool has_language = get_language_from_mode(ID_C) != nullptr;
+
+    if(!has_language)
+    {
+      register_language(new_ansi_c_language);
+    }
+  }
+
+  {
+    cmdlinet cmdline;
+
+    config = configt{};
+    config.main = std::string("main");
+    config.set(cmdline);
+  }
+
+  language_filest language_files;
+  language_files.set_message_handler(null_message_handler);
+
+  language_filet &language_file = language_files.add_file("");
+
+  language_file.language = get_language_from_mode(ID_C);
+  CHECK_RETURN(language_file.language);
+
+  languaget &language = *language_file.language;
+  language.set_message_handler(null_message_handler);
+
+  {
+    const bool error = language.parse(in, "");
+
+    if(error)
+      throw invalid_source_file_exceptiont("parsing failed");
+  }
+
+  language_file.get_modules();
+
+  goto_modelt goto_model;
+
+  {
+    const bool error = language_files.typecheck(goto_model.symbol_table);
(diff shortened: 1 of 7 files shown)


unit/testing-utils/get_goto_model_from_c.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  register_language(new_ansi_c_language);
+
+  cmdlinet cmdline;
+
+  config.main = std::string("main");
+  config.set(cmdline);
+
+  optionst options;
+  cbmc_parse_optionst::set_default_options(options);
```
hannes-steffenhagen-diffblue: This doesn’t actually do anything though? `set_default_options` just sets default values on the `options`, which you then discard.

unit/testing-utils/get_goto_model_from_c.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  languaget &language = *language_file.language;
+  language.set_message_handler(null_message_handler);
+  language.set_language_options(options);
+
+  {
+    const bool error = language.parse(in, "");
+
+    if(error)
+      return {};
```
hannes-steffenhagen-diffblue: I thinki it’d be better to just throw an exception rather than return optional, because presumably you’d expect for unit tests that their input programs do compile (ditto for all the other error cases)

unit/testing-utils/get_goto_model_from_c.cpp
```diff
@@ -0,0 +1,105 @@
...
+    initialize();
+    initialized = true;
+  }
+
+  language_filest language_files;
+  language_files.set_message_handler(null_message_handler);
+
+  language_filet &language_file = language_files.add_file("");
+
+  language_file.language = get_default_language();
```
hannes-steffenhagen-diffblue: This function is called `get_goto_model_from_c`, so I think this should be something like `get_language_from_mode(ID_C)`.

unit/testing-utils/get_goto_model_from_c.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+void initialize()
+{
+  PRECONDITION(!initialized);
+
+  register_language(new_ansi_c_language);
+
+  cmdlinet cmdline;
+
+  config.main = std::string("main");
```
hannes-steffenhagen-diffblue: Actually, I think this shouldn’t be called only once... because `config` is a global variable and could be messed with by other tests, ideally this would start with something like `config = configt{};` so we can be sure we’re starting from a “clean” config and run every time (it’s not like this is going to cost us much test performance).

unit/testing-utils/get_goto_model_from_c.cpp
```diff
@@ -0,0 +1,105 @@
...
+  config.set(cmdline);
+
+  optionst options;
+  cbmc_parse_optionst::set_default_options(options);
+}
+} // namespace
+
+optionalt<goto_modelt> get_goto_model_from_c(std::istream &in)
```
hannes-steffenhagen-diffblue: Not sure if it’s needed right now, but we’ll probably eventually want to set options for `config` and such in here as well (e.g. by passing in `std::function<void(configt&)>` which gets to set some options between resetting everything to default and parsing the C.

unit/testing-utils/get_goto_model_from_c.cpp
danpoe: Removed

unit/testing-utils/get_goto_model_from_c.cpp
danpoe: Done

unit/testing-utils/get_goto_model_from_c.cpp
danpoe: Done

unit/testing-utils/get_goto_model_from_c.cpp
danpoe: Good point. I've now gotten rid of the initialize function altogether.

unit/testing-utils/get_goto_model_from_c.cpp
danpoe: I'll leave this for a future PR.

hannes-steffenhagen-diffblue (approved):
LGTM

PR #5273, opened 2020-03-18
Refactor leaf node handling in smt2_conv

NlightNFotis:
This is for internal reviewing of the refactoring of the smt2_conv leaf nodes, in anticipation of integration with smt-switch.

### DIFF
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -1257,7 +1257,13 @@ void smt2_convt::convert_expr(const exprt &expr)
   }
   else if(expr.id()==ID_mod)
   {
-    convert_mod(to_mod_expr(expr));
+    // TODO: change to use term in some capacity
+    (void)convert_mod(to_mod_expr(expr));
+
+    convert_expr(expr.op0());
+    out << " ";
+    convert_expr(expr.op1());
+    out << ")";
   }
   else if(expr.id()==ID_mult)
   {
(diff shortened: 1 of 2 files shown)


kroening:
Please remember the promise that CBMC needs to be compilable without SMT-Switch!

NlightNFotis:
It will be. This PR will just host a refactoring of smt2-conv.cpp - without anything to do with smt-switch. That will come at a later PR.

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -1257,7 +1257,13 @@ void smt2_convt::convert_expr(const exprt &expr)
   }
   else if(expr.id()==ID_mod)
   {
-    convert_mod(to_mod_expr(expr));
+    // TODO: change to use term in some capacity
+    (void)convert_mod(to_mod_expr(expr));
+
+    convert_expr(expr.op0());
```
hannes-steffenhagen-diffblue, line 1263: There shouldn't be any need for this, we should just print the term we get from `convert_mod`

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -2890,23 +2896,15 @@ void smt2_convt::convert_constant(const constant_exprt &expr)
...
-      out << "(bvsrem ";
+  PRECONDITION(
+    expr.type().id() == ID_unsignedbv || expr.type().id() == ID_signedbv);
 
-    convert_expr(expr.op0());
-    out << " ";
-    convert_expr(expr.op1());
-    out << ")";
-  }
+  if(expr.type().id() == ID_unsignedbv)
```
hannes-steffenhagen-diffblue, line 2904: The fact that this has operands that need to be `convert_expr`'d means it's not a leaf right?

src/solvers/smt2/smt2_conv.h
```diff
@@ -31,6 +31,22 @@ class constant_exprt;
...
+  // Bitvector arithmetic
+  bvurem,
+  bvsrem,
+  bvudiv,
+  bvsdiv
+};
+
+struct SMT_Term
+{
+  SMT_Ops op;
```
hannes-steffenhagen-diffblue, line 47: This doesn't make a lot of sense ATM, I guess this is just a placeholder?

src/solvers/smt2/smt2_conv.h
```diff
@@ -31,6 +31,22 @@ class constant_exprt;
 class index_exprt;
 class member_exprt;
 
+// TODO: This should be replaced on the spot with the
+//  smt-switch ops.h class
```
hannes-steffenhagen-diffblue, line 35: Why though? We should just return SMT switch `Term`, which doesn't seem to reference ops? For what we're doing with these dummy terms I don't think we need anything more complicated than rose-tree of string, or even just string if it comes down to it.

NlightNFotis:
There has been a change in priorities and this work is now parked.

PR #5276, opened 2020-03-19
Do not review: on going Variable Sensitivity Domain

thk123:
Just for simplicity - I've taken @chrisr-diffblue up-to-date VSD branch and moved it onto this fork as a central place. That way, we can raise PRs adding features to it to one central place, rather than a random fork.


thk123:
PR doesn't need to open - made PR to make sure that the branch doesn't get deleted!

PR #5277, opened 2020-03-21
unicode.h: introduce utf8_to_utf32

kroening:

### DIFF
--- a/src/util/unicode.cpp
+++ b/src/util/unicode.cpp
@@ -18,124 +18,122 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "invariant.h"
 
 #ifdef _WIN32
-#include <util/pragma_push.def>
-#ifdef _MSC_VER
-#pragma warning(disable:4668)
-  // using #if/#elif on undefined macro
-#pragma warning(disable : 5039)
+#  include <util/pragma_push.def>
+#  ifdef _MSC_VER
+#    pragma warning(disable : 4668)
+// using #if/#elif on undefined macro
+#    pragma warning(disable : 5039)
 // pointer or reference to potentially throwing function passed to extern C
-#endif
-#include <windows.h>
-#include <util/pragma_pop.def>
+#  endif
+#  include <util/pragma_pop.def>
+#  include <windows.h>
 #endif
 
 std::string narrow(const wchar_t *s)
 {
-  #ifdef _WIN32
+#ifdef _WIN32
 
-  int slength=static_cast<int>(wcslen(s));
-  int rlength=
+  int slength = static_cast<int>(wcslen(s));
+  int rlength =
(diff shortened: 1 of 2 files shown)


src/util/unicode.cpp
```diff
@@ -18,124 +18,117 @@ Author: Daniel Kroening, kroening@kroening.com
...
 
 std::wstring widen(const char *s)
 {
-  #ifdef _WIN32
+#ifdef _WIN32
 
-  int slength=static_cast<int>(strlen(s));
-  int rlength=
-    MultiByteToWideChar(CP_UTF8, 0, s, slength, NULL, 0);
+  int slength = static_cast<int>(strlen(s));
```
hannes-steffenhagen-diffblue, line 54: I realise these are just formatted, but why the cast to `int`? (also [here](https://github.com/diffblue/cbmc/pull/5277/files#diff-3a00d1851f83125704f22191e7dd2075R38))

src/util/unicode.cpp
```diff
@@ -199,10 +199,24 @@ std::wstring utf8_to_utf16_native_endian(const std::string &in)
...
+  for(auto codepoint : utf8_to_utf32(in))
+    utf16_append_code(codepoint, result);
+
+  return result;
+}
+
+/// Convert UTF8-encoded string to UTF-32 with architecture-native endianness.
+/// \par parameters: String in UTF-8 format
+/// \return String in UTF-32 format.
+std::basic_string<unsigned int> utf8_to_utf32(const std::string &utf8_str)
```
hannes-steffenhagen-diffblue: I'd recommend [`char32_t`](https://en.cppreference.com/w/cpp/keyword/char32_t) instead of `unsigned int`, because it is specifically meant for this purpose.

src/util/unicode.cpp
kroening: Likely because that's what `MultiByteToWideChar` expects

src/util/unicode.cpp
kroening: Ok, now using `std::u32string`

src/util/unicode.h
```diff
@@ -25,6 +24,10 @@ std::wstring widen(const std::string &s);
 std::string
 utf32_native_endian_to_utf8(const std::basic_string<unsigned int> &s);
 
+/// \param utf8_str: UTF-8 string
+/// \return UTF-32 encoding of the string
+std::u32string utf8_to_utf32(const std::string &utf8_str);
```
smowton, line 29: Name `_native_endian` for consistency

smowton (approved):

src/util/unicode.h
kroening: Note commit message -- all functions use native endianness, thus no need make this part of the function name. It may make sense to remove the 'native endianness' everywhere.

PR #5278, opened 2020-03-23
Add validation tests for trace XSD

hannes-steffenhagen-diffblue:
We have a XSD for cbmc --trace --xml-ui in our documentation, this adds
a new test run which essentially validates the XML output of all tests
in regression/cbmc against that XSD. Not all tests in there actually
produce traces, but most do so we just exclude the few that don't.

### DIFF
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -32,7 +32,8 @@ DIRS = cbmc \
        symtab2gb \
        solver-hardness \
        goto-ld \
-       # Empty last line
+       validate-trace-xml-schema \
+			 # Empty last line
 
 ifeq ($(OS),Windows_NT)
     detected_OS := Windows
--- a/regression/validate-trace-xml-schema/CMakeLists.txt
+++ b/regression/validate-trace-xml-schema/CMakeLists.txt
@@ -1,0 +1,16 @@
+find_package(PythonInterp 3.5)
+if(${PYTHONINTERP_FOUND})
+  find_program(XMLLINT xmllint)
+  if(NOT XMLLINT)
+    message(WARNING "xmllint not found, skipping trace schema tests")
+  else()
+    add_test(NAME validate-trace-xml-schema
+      COMMAND ${PYTHON_EXECUTABLE} check.py $<TARGET_FILE:cbmc> ${XMLLINT}
+      WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR})
+    set_tests_properties(validate-trace-xml-schema
+      PROPERTIES
+      LABELS "CBMC;CORE")
+  endif()
+else()
+    message(WARNING "Python >= 3.5 not found, skipping trace schema tests")
+endif()
--- a/regression/validate-trace-xml-schema/check.py
+++ b/regression/validate-trace-xml-schema/check.py
@@ -1,0 +1,166 @@
+import os
+import subprocess
+import sys
+import tempfile
+import glob
+
+this_script_dir = os.path.abspath(os.path.dirname(__file__))
+
+TraceXsd = os.path.abspath(os.path.join(this_script_dir, '..', '..', 'doc/assets/xml_spec.xsd'))
+
+test_base_dir = os.path.abspath(os.path.join(this_script_dir, '..', 'cbmc'))
+
+# some tests in the cbmc suite don't work for the trace checks for one reason or another
+ExcludedTests = list(map(lambda s: os.path.join(test_base_dir, s), [
+    # these tests except input from stdin
+    'json-interface1/test_wrong_option.desc',
+    'json-interface1/test.desc',
+    'json-interface1/test_wrong_flag.desc',
+    'xml-interface1/test_wrong_option.desc',
+    'xml-interface1/test.desc',
+    'xml-interface1/test_wrong_flag.desc',
+    # these want --show-goto-functions instead of producing a trace
+    'destructors/compound_literal.desc',
+    'destructors/enter_lexical_block.desc',
+    'reachability-slice-interproc2/test.desc',
+    # this one wants show-properties instead producing a trace
+    'show_properties1/test.desc',
+    # program-only instead of trace
+    'vla1/program-only.desc',
+    'Quantifiers-simplify/simplify_not_forall.desc',
+    # this one doesn't work for me locally at all
+    # 'integer-assignments1/test.desc',
+    # these test for invalid command line handling
+    'bad_option/test_multiple.desc',
+    'bad_option/test.desc',
+    # this one produces XML intermingled with main XML output when used with --xml-ui
+    'graphml_witness2/test.desc',
+    # produces intermingled XML on the command line
+    'coverage_report1/test.desc',
+    'coverage_report1/paths.desc',
+    'graphml_witness1/test.desc',
+    'switch8/program-only.desc',
+    # this uses json-ui (fails for a different reason actually, but should also
+    #   fail because of command line incompatibility)
+    'json1/test.desc',
+    # uses show-goto-functions
+    'reachability-slice/test.desc',
+    'reachability-slice/test2.desc',
+    'reachability-slice/test3.desc',
+    'reachability-slice-interproc/test.desc',
+    'unsigned1/test.desc',
+    'reachability-slice-interproc3/test.desc',
+    'sync_lock_release-1/symbol_per_type.desc',
+    # this test is marked broken-smt-backend and doesn't work for me locally
+    'integer-assignments1/test.desc'
+]))
+
+# TODO maybe consider looking them up on PATH, but direct paths are
+#      harder to get wrong
+if len(sys.argv) != 3:
+    sys.stderr.write('Usage: check.py <path-to-cbmc> <path-to-xmllint>')
+CbmcPath = os.path.abspath(sys.argv[1])
+XmlLintPath = os.path.abspath(sys.argv[2])
+
+
+class ChangeDir:
+    def __init__(self, change_to_wd):
+        self.current_wd = os.getcwd()
+        self.change_to_wd = change_to_wd
+
+    def __enter__(self):
+        os.chdir(self.change_to_wd)
+
+    def __exit__(self, _type, _value, _tb):
+        os.chdir(self.current_wd)
+
+
+class TestSpec:
+    def __init__(self, args, is_knownbug, is_future, is_thorough):
+        self.args = args
+        self.is_knownbug = is_knownbug
+        self.is_future = is_future
+        self.is_thorough = is_thorough
+
+
+class Validator:
+    def __init__(self, total_desc_count):
+        self.failed_tests = []
+        self.total_desc_count = total_desc_count
+        self.checked_desc_count = 0
+
+    def check_test_desc(self, test_desc_path):
+        self.checked_desc_count += 1
+        sys.stdout.write('*** Checking [{}/{}] {}... '.format(
+            self.checked_desc_count, self.total_desc_count, test_desc_path))
+        sys.stdout.flush()
+        spec = self.read_spec(test_desc_path)
+        if spec.is_knownbug:
+            print('[Skipping KNOWNBUG]')
+            return
+        elif spec.is_future:
+            print('[Skipping FUTURE]')
+            return
+        elif spec.is_thorough:
+            print('[Skipping THOROUGH]')
+            return
+        test_desc_dir = os.path.dirname(test_desc_path)
+        with ChangeDir(test_desc_dir):
+            with tempfile.TemporaryFile() as trace_file:
+                self.read_trace_into(trace_file, spec.args)
+                trace_file.seek(0)
+                self.check_trace(test_desc_path, trace_file)
+
+    def read_trace_into(self, trace_file, args):
+        subprocess.run([CbmcPath, '--trace', '--xml-ui'] + args,
+                       stdout=trace_file)
+
+    def check_trace(self, test_desc_path, trace_file):
+        validate_result = subprocess.run([XmlLintPath,
+                                          '--noout',  # we don't want it to print the XML we pipe in
+                                          '--schema', TraceXsd,
+                                          '-'  # read from STDIN
+                                          ],
+                                         stdin=trace_file,
+                                         stdout=subprocess.PIPE,
+                                         stderr=subprocess.PIPE)
+        if validate_result.returncode == 0:
+            sys.stdout.buffer.write(b'[SUCCESS]\n')
+        else:
+            sys.stdout.buffer.write(b'[FAILURE]\n')
+            self.failed_tests.append(test_desc_path)
+            sys.stdout.buffer.write(validate_result.stdout)
+            sys.stdout.buffer.write(validate_result.stderr)
+
+    def read_spec(self, test_desc_path):
+        with open(test_desc_path) as test_desc_file:
+            test_tags = test_desc_file.readline().split()
+            source_file = test_desc_file.readline().strip()
+            extra_args = test_desc_file.readline().split()
+        is_future = 'FUTURE' in test_tags
+        is_thorough = 'THOROUGH' in test_tags
+        is_knownbug = 'KNOWNBUG' in test_tags
+        return TestSpec(args=[source_file] + extra_args,
+                        is_thorough=is_thorough,
+                        is_future=is_future,
+                        is_knownbug=is_knownbug)
+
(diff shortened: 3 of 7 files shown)


NlightNFotis (commented):
LGTM, but cannot comment on the gradle build files and the .bat files.

regression/validate-trace-xml-schema/validate-xsd/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
```
hannes-steffenhagen-diffblue: Everything in the `gradle` directory as well as the `gradlew` and `gradlew.bat` scripts are autogenerated by gradle so the build can work without a user having to have a compatible gradle version locally. So they're meant to be here, but shouldn't be reviewed.

regression/Makefile
```diff
@@ -2,35 +2,36 @@
 # running tests in parallel, it is important that these directories are
 # listed with decreasing runtimes (i.e. longest running at the top)
 DIRS = cbmc \
-       cbmc-library \
```
thk123: :question: I _assume_ this is just a reordering of the Makefile? If it is, it'd be best pulled out into its own commit so can be clearly stated as such

regression/validate-trace-xml-schema/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+if(${PYTHONINTERP_FOUND}
+  AND ${Java_FOUND} AND ${Java_VERSION_MAJOR} VERSION_GREATER 8)
+    add_test(NAME validate-trace-xml-schema
+            COMMAND ${PYTHON_EXECUTABLE} check.py $<TARGET_FILE:cbmc>
+            WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR})
+    set_tests_properties(validate-trace-xml-schema
+            PROPERTIES
+            LABELS "CBMC;CORE")
+else()
+    message(WARNING "JDK not found, skipping trace schema tests")
```
thk123: :question: This stops the tests being configured? I wonder if we should configure a test that always fails? That way, if someone configures it locally, they don't need JDK, but if CI fails on this step, it doesn't silently pass.

regression/validate-trace-xml-schema/check.sh
```diff
@@ -0,0 +1,29 @@
+if ! command -v python3 > /dev/null; then
+    echo "python3 not found, skipping XSD tests"
+    exit
+fi
+
+python_minor_version=$(python3 --version | cut -d ' ' -f 2 | cut -d '.' -f 2)
+
+if [ $python_minor_version -lt 5 ]; then
+    echo "python version less than 3.5, skipping XSD tests"
```
thk123, line 9: :bulb: I think I would exit with a non-zero exit code here, to prevent silent failures on CI :ghost:

regression/validate-trace-xml-schema/check.sh
```diff
@@ -0,0 +1,29 @@
...
+    exit
+fi
+
+if ! command -v javac > /dev/null; then
+    echo "javac not found, skipping XSD tests"
+    exit
```
thk123: :ghost:

regression/validate-trace-xml-schema/check.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Java versions are complicated
+
+java_version=$(javac -version | cut -d ' ' -f 2)
+javac_major_version=$(echo $java_version | cut -d '.' -f 1)
+javac_minor_version=$(echo $java_version | cut -d '.' -f 2)
+
+if [ $javac_major_version -eq 1 ] || [ $javac_minor_version -ge 8 ]; then
+    echo "JDK version less than 8, skipping XSD tests"
+    exit
```
thk123: :ghost:

regression/validate-trace-xml-schema/validate-xsd/src/main/java/com/diffblue/validatexsd/Main.java
```diff
@@ -0,0 +1,50 @@
...
+import javax.xml.validation.Schema;
+import javax.xml.validation.SchemaFactory;
+import javax.xml.validation.Validator;
+import java.io.FileInputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.concurrent.Callable;
+
+@Command(name = "validate-xsd", mixinStandardHelpOptions = true, version = "0.1",
```
thk123: It makes me a little sad we weren't able to find an xsd validator that exists on apt, but there is not much code here anyway so I guess it's OK

regression/validate-trace-xml-schema/validate-xsd/src/main/java/com/diffblue/validatexsd/Main.java
```diff
@@ -0,0 +1,50 @@
...
+  description = "tells you if given XML (read from stdin) matches a given XSD spec")
```
thk123: :bulb: An example usage would be great:

```
Usage:

$ cbmc test.c --trace | java -jar validator.jar 
```

regression/validate-trace-xml-schema/validate-xsd/src/main/java/com/diffblue/validatexsd/Main.java
```diff
@@ -0,0 +1,50 @@
...
+public class Main implements Callable<Integer> {
+  @Parameters(index = "0")
+  private Path xsdSchemaFilePath;
+
+  @Override
+  public Integer call() throws Exception {
```
thk123: :no_entry_sign: Sorry to be a pain, but can we have 4 simple tests just to check the errors get thrown correctly:

 - no schema
 - invalid schema
 - invalid xml
 - valid xml

Hopefully these don't take long to write.

regression/validate-trace-xml-schema/validate-xsd/src/main/java/com/diffblue/validatexsd/LoggingErrorHandler.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.xml.sax.ErrorHandler;
+import org.xml.sax.SAXException;
+import org.xml.sax.SAXParseException;
+
+public class LoggingErrorHandler implements ErrorHandler {
+  private boolean errorOccurred = false;
+
+  @Override
+  public void warning(SAXParseException e) throws SAXException {
```
thk123: Should you be re-throwing these execptions? Seems weird the signature would throw otherwise?

regression/validate-trace-xml-schema/check.py
```diff
@@ -0,0 +1,172 @@
...
+
+this_script_dir = os.path.abspath(os.path.dirname(__file__))
+
+TraceXsdSpec = os.path.abspath(os.path.join(this_script_dir, '..', '..', 'doc/assets/xml_spec.xsd'))
+
+test_base_dir = os.path.abspath(os.path.join(this_script_dir, '..', 'cbmc'))
+
+# some tests in the cbmc suite don't work for the trace checks for one reason or another
+ExcludedTests = list(map(lambda s: os.path.join(test_base_dir, s), [
+    # these tests except input from stdin
```
thk123, line 15: expect? accept? I don't think you mean except here though

regression/validate-trace-xml-schema/check.py
```diff
@@ -0,0 +1,172 @@
...
+    # these want --show-goto-functions instead of producing a trace
+    'destructors/compound_literal.desc',
+    'destructors/enter_lexical_block.desc',
+    'reachability-slice-interproc2/test.desc',
+    # this one wants show-properties instead producing a trace
+    'show_properties1/test.desc',
+    # program-only instead of trace
+    'vla1/program-only.desc',
+    # this one doesn't work for me locally at all
+    'integer-assignments1/test.desc',
```
thk123: :question: Huh? You mean the test fails locally? Perhaps we could raise an issue to look into that?

regression/validate-trace-xml-schema/check.py
```diff
@@ -0,0 +1,172 @@
...
+    # these test for invalid command line handling
+    'bad_option/test_multiple.desc',
+    'bad_option/test.desc',
+    # this one produces XML intermingled with main XML output when used with --xml-ui
+    'graphml_witness2/test.desc',
+    # this one uses program-only again
+    'Quantifiers-simplify/simplify_not_forall.desc',
```
thk123: :bulb: I guess these are in alphabetical order, but might be better to group by reason?

regression/validate-trace-xml-schema/check.py
```diff
@@ -0,0 +1,172 @@
...
+    'reachability-slice/test.desc',
+    'reachability-slice/test2.desc',
+    'reachability-slice/test3.desc',
+    'reachability-slice-interproc/test.desc',
+    # again, intermingles several XML documents
+    'graphml_witness1/test.desc',
+    'switch8/program-only.desc',
+    # show-goto-functions
+    'unsigned1/test.desc',
+    # show-goto-functions
```
thk123: :pick: Combine these groups

regression/validate-trace-xml-schema/check.py
```diff
@@ -0,0 +1,172 @@
...
+    def __exit__(self, _type, _value, _tb):
+        os.chdir(self.current_wd)
+
+
+class TestSpec:
+    def __init__(self, args, is_knownbug, is_future, is_thorough):
+        self.args = args
+        self.is_knownbug = is_knownbug
+        self.is_future = is_future
+        self.is_thorough = is_thorough
```
thk123, line 83: I think we've written a `.desc` python parser before (possibly for test-gen) :bulb: Pull out that into a separate python module so that it may be reused

regression/validate-trace-xml-schema/check.py
```diff
@@ -0,0 +1,172 @@
...
+            print('[SUCCESS]')
+        else:
+            print('[FAILURE]')
+            self.failed_tests.append(spec_path)
+        sys.stdout.buffer.write(validate_result.stdout)
+        sys.stdout.buffer.write(validate_result.stderr)
+
+    def read_spec(self, spec_path):
+        with open(spec_path) as spec_file:
+            spec = spec_file.readline().split()
```
thk123: :bulb: I guess I would use the word `test_kind` rather than spec, which has 3 means in this file:
1. the XSD specificaiton
2. the entire .desc file
3. The first line of the desc file

regression/validate-trace-xml-schema/check.py
```diff
@@ -0,0 +1,172 @@
...
+                                         stderr=subprocess.PIPE)
+        if validate_result.returncode == 0:
+            print('[SUCCESS]')
+        else:
+            print('[FAILURE]')
+            self.failed_tests.append(spec_path)
+        sys.stdout.buffer.write(validate_result.stdout)
+        sys.stdout.buffer.write(validate_result.stderr)
+
+    def read_spec(self, spec_path):
```
thk123: :bulb: Here your spec is the `.desc` file right? I would call this `read_test_desc`

thk123 (changes requested):
This seems good - some naming stuff in the python file I'd really like to see tidied up. Also, I sadly must insist on some tests for the program than runs the validation.

regression/validate-trace-xml-schema/CMakeLists.txt
hannes-steffenhagen-diffblue: Not all our CI passes have both Java and Python 3 available.

regression/validate-trace-xml-schema/check.sh
hannes-steffenhagen-diffblue: This is not supposed to fail CI

regression/validate-trace-xml-schema/validate-xsd/src/main/java/com/diffblue/validatexsd/Main.java
hannes-steffenhagen-diffblue: So after doing some more digging apparently there's xmllint? I think I can remove the java bits and just use that.

regression/validate-trace-xml-schema/validate-xsd/src/main/java/com/diffblue/validatexsd/LoggingErrorHandler.java
hannes-steffenhagen-diffblue: You can rethrow, but you don't have to.

regression/validate-trace-xml-schema/validate-xsd/src/main/java/com/diffblue/validatexsd/Main.java
thk123: Sounds good -> saves writing the tests! :tada:

regression/validate-trace-xml-schema/CMakeLists.txt
thk123: I guess I'd rather we explicitly _skip_ this test on the jobs that don't have the requirements.

thk123:
I've checked the test appears in the output for one job, but just to be safe, could you break the output trace to ensure the CI fails (e.g. just change one of the fields or something so is nonsense, ideally one that doesn't cause any other tests to fail) (maybe add a field?)

regression/validate-trace-xml-schema/check.py
thk123: > expect? accept? I don't think you mean except here though

:pick: This seems outstanding

regression/validate-trace-xml-schema/check.py
```diff
@@ -0,0 +1,166 @@
...
+    'destructors/compound_literal.desc',
+    'destructors/enter_lexical_block.desc',
+    'reachability-slice-interproc2/test.desc',
+    # this one wants show-properties instead producing a trace
+    'show_properties1/test.desc',
+    # program-only instead of trace
+    'vla1/program-only.desc',
+    'Quantifiers-simplify/simplify_not_forall.desc',
+    # this one doesn't work for me locally at all
+    # 'integer-assignments1/test.desc',
```
thk123, line 32: >  Huh? You mean the test fails locally? Perhaps we could raise an issue to look into that?

:question: outstanding, this needs to be addressed IMO

regression/validate-trace-xml-schema/check.sh
thk123: What I'm suggesting is it should fail CI (and the CIs where this doesn't work, we disable it in the configuration file). Currently if a config breaks in a way that means we have the wrong version of python, this test will just be disabled and we'll never know.

regression/validate-trace-xml-schema/check.py
thk123: Would like to see this addressed, as I keep thinking `is_knownbug = 'KNOWNBUG' in spec` is a bug, since KNOWNBUG might appear elsewhere in the spec, except here `spec` is just the first line.

Where "this" is:

Rename `spec` to `spec_kind` since `spec` makes it seem like it is the whole contents of the `desc` file.

thk123 (commented):
One remaining blocking comment (answering my question about the test you've labelled as failing locally).

I also strongly believe scripts that we run in CI should silently fail (e.g. when python not installed)

regression/validate-trace-xml-schema/check.py
thk123: To be explicit about what needs addressing:
 - shouldn't check in "me" since no way of knowing who "me" is as soon as this line gets moved. 

Further, I'd like to understand what you mean "doesn't work locally at all" means!

regression/validate-trace-xml-schema/check.py
thk123: For reference, we [did](https://github.com/diffblue/cover/blob/develop/test-gen/regression/common.py#L50 :lock:),  but not trivial to unify these two at the moment.

thk123 (approved):
I give up on trying to get the CI to fail without dependencies:
 - for make builds on windows, can't get all the dependencies, and `make test` doesn't offer the opportunity to filter _out_ tests AFAICT
 - for cmake builds, though you can filter out the tests, the info about what dependency gets lost which makes it a pain to debug, so I'll leave it

I'll raise a separate PR addressing my other comments, since I've fixed them, but let's get this in first (needs a rebase which is probably easier for you to do?

peterschrammel (approved):

PR #5279, opened 2020-03-24
value-set domain for variable-sensitivity-domain

owen-mc-diffblue:
Note this PR is not to `develop`, it is to `variable-sensitivity-domain`.


src/analyses/variable-sensitivity/abstract_object.h
```diff
@@ -224,6 +224,16 @@ class abstract_objectt:public std::enable_shared_from_this<abstract_objectt>
     const abstract_object_visitort &visitor) const
   { return shared_from_this(); }
 
+  virtual size_t internal_hash() const
```
hannes-steffenhagen-diffblue, line 226: Where are these used?

regression/goto-analyzer/value-set-function-pointers-arrays/test.desc
```diff
@@ -0,0 +1,13 @@
+CORE
+main.c
+--variable --structs --arrays --pointers --value-set --show --pointer-check
+^file main.c line 29 function main: replacing function pointer by 2 possible targets$
+^file main.c line 40 function main: replacing function pointer by 2 possible targets$
+^main::1::fun1 \(\) -> value-set-begin: ptr ->\(f\), ptr ->\(g\), :value-set-end$
+^main::1::fun_array3 \(\) -> \{\[0\] = value-set-begin: ptr ->\(f\), ptr ->\(g\), :value-set-end$
```
hannes-steffenhagen-diffblue: where are `fun_array1` and 2?

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.h
```diff
@@ -125,9 +125,19 @@ abstract_object_pointert variable_sensitivity_object_factoryt::
       abstract_object_classt, write_location_contextt>(
         type, top, bottom, e, enviroment, ns);
   else
-    return initialize_context_abstract_object<
-      abstract_object_classt, context_abstract_objectt>(
-        type, top, bottom, e, enviroment, ns);
+  {
```
hannes-steffenhagen-diffblue, line 209: The message in this commit is confusing. Top/bottom are unrelated to context afaict.

src/analyses/variable-sensitivity/value_set_abstract_object.h
```diff
@@ -0,0 +1,228 @@
...
+
+#include <unordered_set>
+
+#include <analyses/variable-sensitivity/abstract_value.h>
+
+class value_set_abstract_objectt : public abstract_valuet
+{
+public:
+  using abstract_object_sett = std::unordered_set<
+    abstract_object_pointert,
```
hannes-steffenhagen-diffblue, line 23: I believe this is incorrect? This should just be `exprt`.

hannes-steffenhagen-diffblue (changes requested):

src/analyses/variable-sensitivity/interval_array_abstract_object.cpp
```diff
@@ -38,6 +39,19 @@ static constant_interval_exprt eval_and_get_as_interval(
   const namespacet &ns)
 {
   auto evaluated_index = environment.eval(expr, ns);
+  auto maybe_interval = evaluated_index->unwrap_context();
```
hannes-steffenhagen-diffblue: I'm not really sure why we're changing interval array abstract object here? This should be unrelated to value set abstract objects.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -0,0 +1,305 @@
...
+abstract_object_pointert value_set_abstract_objectt::expression_transform(
+  const exprt &expr,
+  const std::vector<abstract_object_pointert> &operands,
+  const abstract_environmentt &environment,
+  const namespacet &ns) const
+{
+  std::size_t num_operands = expr.operands().size();
+  PRECONDITION(operands.size() == num_operands);
+
+  std::vector<abstract_object_sett> collective_operands;
```
hannes-steffenhagen-diffblue, line 92: These should just be `value_set_abstract_objectt`

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -0,0 +1,305 @@
...
+  for(const auto &op : operands)
+  {
+    auto vsab = std::dynamic_pointer_cast<const value_set_abstract_objectt>(
+      maybe_unwrap_context(op));
+    INVARIANT(vsab, "should be a value set abstract object");
+    collective_operands.push_back(vsab->get_values());
+  }
+
+  abstract_object_sett resulting_objects;
+  for_each_comb(
```
hannes-steffenhagen-diffblue, line 103: This bit needs to account for the possibility of an operand being top (or bottom I guess, although in that case I don't think we'd even evaluate this?).

src/analyses/variable-sensitivity/abstract_object.h
owen-mc-diffblue: They are used in `abstract_hashert` and `abstract_equalert`, further down in this file, which are in turn used for the definition of `abstract_object_sett` at the top of value_set_abstract_object.h, which is in the next commit.

regression/goto-analyzer/value-set-function-pointers-arrays/test.desc
owen-mc-diffblue: `fun_array1` is read from when assigning a value to `fun1`. `fun_array2` is read from when assigning a value to `fun_array3`. There is little benefit to asserting what values they contain.

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.h
owen-mc-diffblue: The point is that the commit makes it return an unwrapped object. It just so happens that this is done differently for top and bottom.

src/analyses/variable-sensitivity/value_set_abstract_object.h
NlightNFotis: Yes, I believe so too.

NlightNFotis (approved):
Overall, it looks good to me.

The only thing I would ask for is a squash of the small inconsequential commits (like copyright changes, or removal of files or the tests being together in one commit if possible).

src/analyses/variable-sensitivity/interval_array_abstract_object.cpp
owen-mc-diffblue: I don't know because I didn't write the code. All the tests pass without this chunk of code, so it seems you're probably right.

src/analyses/variable-sensitivity/value_set_abstract_object.h
owen-mc-diffblue: I don't know the ins and outs, but that would be a major change, because `abstract_object_pointert` doesn't inherit from `exprt`

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
owen-mc-diffblue: Because of the recursive nature of `apply_comb`, we need to be able to access the `n`th entry of `collective_operands`, and hence it can't just be a set.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
owen-mc-diffblue: It makes sense to me that that this should cope with top, but I have no idea how to implement it

hannes-steffenhagen-diffblue:
We'll be taking this over

PR #5281, opened 2020-03-26
doc/properties: #pragma disable requires version >=5.12

amosr:
Hi,
Just a documentation change: I noticed on the website that the documentation (http://www.cprover.org/cprover-manual/properties/) describes the `#pragma CPROVER check disable` feature, but the one you can download doesn't support that. (It took me a while to realise that's why it wasn't working.) So I just made sure the documentation explicitly stated it's only in the unreleased 5.12.

### DIFF
--- a/doc/cprover-manual/properties.md
+++ b/doc/cprover-manual/properties.md
@@ -132,9 +132,10 @@ The goto-instrument program supports these checks:
 As all of these checks apply across the entire input program, we may wish to
 disable them for selected statements in the program. For example, unsigned
 overflows can be expected and acceptable in certain instructions even when
-elsewhere we do not expect them. To selectively disable automatically generated
-properties use `#pragma CPROVER check disable "<name_of_check>"`, which remains
-in effect until a `#pragma CPROVER check pop` (to re-enable all properties
+elsewhere we do not expect them. As of version 5.12, CBMC supports selectively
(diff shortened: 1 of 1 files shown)


doc/cprover-manual/properties.md
```diff
@@ -132,9 +132,10 @@ The goto-instrument program supports these checks:
 As all of these checks apply across the entire input program, we may wish to
 disable them for selected statements in the program. For example, unsigned
 overflows can be expected and acceptable in certain instructions even when
-elsewhere we do not expect them. To selectively disable automatically generated
-properties use `#pragma CPROVER check disable "<name_of_check>"`, which remains
-in effect until a `#pragma CPROVER check pop` (to re-enable all properties
+elsewhere we do not expect them. The latest development version of CBMC (version 5.12)
```
peterschrammel: 5.12 has been released and is available from https://github.com/diffblue/cbmc/releases, but @kroening hasn't updated the cprover website yet.

peterschrammel (approved):

doc/cprover-manual/properties.md
tautschnig: I don't think we should be referring to the "latest" version in documentation, we should instead say: "CBMC 5.12 (or any later version) supports ..."

PR #5285, opened 2020-04-01
Fix CPP models when using CMake Build

thk123:
In trying to track down why the defines don't work in the library, I discovered that the cpp models (that includes definitions for new) do not work on CMake builds. I'm working on a fix, but just to confirm my suspicion I've added a test that I believe will pass on non-cmake builds but will fail on cmake builds. 

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -1,14 +1,8 @@
 file(GLOB cpp_library_sources "library/*.c")
 
-add_custom_command(OUTPUT converter_input.txt
-    COMMAND cat ${cpp_library_sources} > converter_input.txt
-    DEPENDS ${cpp_library_sources}
-)
-
 add_custom_command(OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/cprover_library.inc"
-    COMMAND ../ansi-c/converter < "converter_input.txt" > "${CMAKE_CURRENT_BINARY_DIR}/cprover_library.inc"
-    DEPENDS "converter_input.txt" ../ansi-c/converter
-)
+        COMMAND cat ${cpp_library_sources} | $<TARGET_FILE:converter> > "${CMAKE_CURRENT_BINARY_DIR}/cprover_library.inc"
+        DEPENDS converter ${cpp_library_sources})
 
 ################################################################################
 
@@ -32,6 +26,12 @@ endif()
 
 file(GLOB_RECURSE sources "*.cpp" "*.h")
 add_library(cpp ${sources})
+
+set_source_files_properties(
+        ${sources}
+        PROPERTIES
+        OBJECT_DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/cprover_library.inc"
+)
 
 generic_includes(cpp)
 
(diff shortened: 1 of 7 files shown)


thk123:
This PR shouldn't pass:
 - the cpp new test should fail on all cmake builds 
 - the string abstraction tests should also fail on cmake builds

The reason is the cmake file for the cpp folder does not add the dependency so the `cprover_library.inc` file is never generated, which means only the C library is imported, but it is imported without the ability to turn string abstraction on. I have a fix for the CMake which I'll apply once this is failing CI. 

The CI _isn't_ failing:
 - the windows cmake build bases regardless of build outcome (and has done since the start)
 - the linux gcc cmake build has disappeared from the checks

Therefore we have no CI for CMake builds :sob:

thk123:
Rebased on top of #5287 to ensure the cbmc-linux-cmake-gcc job now fails.

smowton (approved):

thk123:
Confirmed both the cpp-new test (no model for __new) and the string-abstraction tests are failing as expected on CMake job.  And indeed CI fails :tada:

thk123:
Now added commit that should fix the tests for the CMake build 
@smowton a review of the last commit (that actually fixes the problem) would be useful.

smowton:
Write -> append? Seen it already, LGTM

thk123:
Ah sorry @smowton, I think you're looking at the wrong commit, I added this commit to fix the cmake file: 9c5a5c420bb94fe4d69631b70b436bb98b18e547

smowton:
Also looks sane except nitpick above

thk123:
@smowton in this case though `${cpp_library_sources}` is the result of `file(GLOB cpp_library_sources "library/*.c")` which is a list of files, I think you need to use cat to join them together? That said - this doesn't (and didn't) work on the Windows CMake build - so an alternative to cat will be needed in a subsequent PR.

smowton:
Doh right, sorry

src/cpp/CMakeLists.txt
```diff
@@ -33,6 +27,12 @@ endif()
 file(GLOB_RECURSE sources "*.cpp" "*.h")
 add_library(cpp ${sources})
 
+set_source_files_properties(
+        ${sources}
+        PROPERTIES
+        OBJECT_DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/cprover_library.inc"
+)
```
hannes-steffenhagen-diffblue, line 34: This should just be `add_dependencies(cpp "${CMAKE_CURRENT_BINARY_DIR}/cprover_library.inc")` I think?

thk123: Copied from ansi-c, will try updating both.

thk123: This doesn't work:

```
CMake Error at src/cpp/CMakeLists.txt:30 (add_dependencies):
  The dependency target
  "/home/tkiley/workspace/cbmc/build/debug/src/cpp/cprover_library.inc" of
  target "cpp" does not exist.
```

You're more than welcome to raise a follow up PR making this nicer, but I'd rather just merge this as is, rather than spend more time on it, since it does fix the problem of the C++ library and macros in the C library not working on CMake.

tautschnig (approved):

PR #5290, opened 2020-04-08
Add support for void* pointers in goto-harness

thk123:

### DIFF
--- a/regression/goto-harness/void-star-pointer/main.c
+++ b/regression/goto-harness/void-star-pointer/main.c
@@ -1,0 +1,6 @@
+#include <assert.h>
+
+void test_function(void *input)
+{
+  assert(input == 0);
+}
--- a/regression/goto-harness/void-star-pointer/main.c
+++ b/regression/goto-harness/void-star-pointer/main.c
@@ -12,3 +12,10 @@ void test_void_array(void *input_array[10])
   assert(input_array[0] == 0);
   assert(false);
 }
+
+void test_ptr_array(void **input_array)
+{
+  __CPROVER_assume(input_array != 0);
+  assert(input_array[0] == 0);
+  assert(false);
+}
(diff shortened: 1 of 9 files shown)


src/goto-harness/recursive_initialization.cpp
```diff
@@ -249,6 +258,11 @@ code_blockt recursive_initializationt::build_constructor_body(
     {
       return build_function_pointer_constructor(result_symbol, is_nullable);
     }
+    if(type.subtype().id() == ID_empty)
+    {
+      // always initalize void* pointers as NULL
+      return build_null_pointer(result_symbol);
```
hannes-steffenhagen-diffblue, line 264: This is OK for now, but longer term we'll have to think of a better solution for this (although meaningful void* arguments are tricky, and tbh shouldn't be appearing very often in the sort of harnesses we're trying to generate right now).

thk123: Feel free to raise a ticket - though tbh I'm not really sure what other useful behaviour exists. A non-det switch over all types? I guess we could do something similar to JBMC where we look at the subsequent casts to pick a type.

regression/goto-harness/void-star-pointer/main.c
```diff
@@ -1,6 +1,14 @@
...
 void test_function(void *input)
 {
   assert(input == 0);
 }
+
+void test_void_array(void *input_array[10])
+{
+  __CPROVER_assume(input_array != 0);
+  assert(input_array[0] == 0);
+  assert(false);
```
hannes-steffenhagen-diffblue, line 13: 👍 A lot of tests forget this step

hannes-steffenhagen-diffblue (approved):

danpoe (approved):

PR #5294, opened 2020-04-14
Add newline at the end of function pointer restriction json file.

NlightNFotis:
This is fixing a small inconsistency at the output in the json files produced by `function_pointers_restrictionst` that was causing some tests to fail.

### DIFF
--- a/src/goto-programs/restrict_function_pointers.cpp
+++ b/src/goto-programs/restrict_function_pointers.cpp
@@ -579,6 +579,8 @@ void function_pointer_restrictionst::write_to_file(
   }
 
   function_pointer_restrictions_json.output(outFile);
+  // Ensure output file ends with a newline character.
+  outFile << '\n';
 }
 
 function_pointer_restrictionst::restrictionst


hannes-steffenhagen-diffblue (approved):

thk123 (approved):
Any reason not to fix it in the `output` function?

hannes-steffenhagen-diffblue:
@thk123 
The output is the output function from `jsont`; Didn't want to change that because I wasn't convinced this not having an endline at the end was incorrect. It is certainly incorrect to not have an endline at the end of a file though, which is why we put it here.

src/goto-programs/restrict_function_pointers.cpp
```diff
@@ -579,6 +579,8 @@ void function_pointer_restrictionst::write_to_file(
   }
 
   function_pointer_restrictions_json.output(outFile);
+  // output does not have a newline at the end
```
chrisr-diffblue: ⛏️ I'd say a better comment here would be "Ensure the outFile ends with a newline character"

chrisr-diffblue (approved):

src/goto-programs/restrict_function_pointers.cpp
NlightNFotis: Good stuff. I've implemented this.

pkesseli (approved):

PR #5298, opened 2020-04-16
Fix max malloc size

thk123:
Previously, to work out the max allocation size, we we working out how many bits there were for the offset (`pointer_width - object_bits - 1`) and then raising 2 to that power. However, we were doing this with int literals, and so the calculation was done in 32bits.

Therefore, for the default object bits (8) gives you:

1 << (64 - 8 - 1)

Which depends on your compiler / OS might give you [zero](https://ideone.com/H7LRLb), or `8388608` (since overflow on signed integral types is undefined behavior, both are valid). The correct number is `36028797018963968`

This PR changes to use a 64 bit type on all platforms, and invariant if the result is not going to fit. 

Also adds unit tests showing this, and documentation explaining the calculations. 

### DIFF
--- a/src/ansi-c/ansi_c_internal_additions.cpp
+++ b/src/ansi-c/ansi_c_internal_additions.cpp
@@ -122,7 +122,6 @@ static std::string architecture_string(T value, const char *s)
          std::string(s) + "=" + std::to_string(value) + ";\n";
 }
 
-using max_alloc_sizet = uint64_t;
 /// The maximum allocation size is determined by the number of bits that
 /// are left in the pointer of width \p pointer_width.
 ///
@@ -137,7 +136,7 @@ using max_alloc_sizet = uint64_t;
 /// \param pointer_width: The width of the pointer
 /// \param object_bits : The number of bits used to represent the ID
 /// \return The size in bytes of the maximum allocation supported.
-static max_alloc_sizet
+static mp_integer
 max_malloc_size(std::size_t pointer_width, std::size_t object_bits)
 {
   PRECONDITION(pointer_width >= 1);
@@ -148,9 +147,7 @@ max_malloc_size(std::size_t pointer_width, std::size_t object_bits)
   // but also down to -allocation_size, therefore the size is allowable
   // is number of bits, less the signed bit.
   const auto bits_for_positive_offset = offset_bits - 1;
-  PRECONDITION(
-    bits_for_positive_offset < std::numeric_limits<max_alloc_sizet>::digits);
-  return ((max_alloc_sizet)1) << (max_alloc_sizet)bits_for_positive_offset;
+  return ((mp_integer)1) << (mp_integer)bits_for_positive_offset;
 }
 
 void ansi_c_internal_additions(std::string &code)
@@ -195,7 +192,7 @@ void ansi_c_internal_additions(std::string &code)
     "int " CPROVER_PREFIX "malloc_failure_mode_assert_then_assume="+
     std::to_string(config.ansi_c.malloc_failure_mode_assert_then_assume)+";\n"
     CPROVER_PREFIX "size_t " CPROVER_PREFIX "max_malloc_size="+
-    std::to_string(max_malloc_size(config.ansi_c.pointer_width, config
+      integer2string(max_malloc_size(config.ansi_c.pointer_width, config
     .bv_encoding.object_bits))+";\n"
 
     // this is ANSI-C
--- a/src/ansi-c/ansi_c_internal_additions.cpp
+++ b/src/ansi-c/ansi_c_internal_additions.cpp
@@ -133,6 +133,7 @@ using max_alloc_sizet = uint64_t;
 static max_alloc_sizet
 max_malloc_size(std::size_t pointer_width, std::size_t object_bits)
 {
+  PRECONDITION(pointer_width >= 1);
   PRECONDITION(object_bits < pointer_width - 1);
   PRECONDITION(object_bits >= 1);
   const auto bits_for_offset = pointer_width - object_bits - 1;
--- a/src/ansi-c/ansi_c_internal_additions.cpp
+++ b/src/ansi-c/ansi_c_internal_additions.cpp
@@ -124,9 +124,16 @@ static std::string architecture_string(T value, const char *s)
 
 using max_alloc_sizet = uint64_t;
 /// The maximum allocation size is determined by the number of bits that
-/// are left in the pointer of width \p pointer_width after allowing for the
-/// signed bit, and the bits used for the objects ID (determined by
-/// \p object_bits).
+/// are left in the pointer of width \p pointer_width.
+///
+/// The allocation size cannot exceed the number represented by the (signed)
+/// offset, otherwise it would not be possible to store a pointer into a
+/// valid bit of memory. Therefore, the max allocation size is
+/// 2^(offset_bits - 1), where the offset bits is the number of bits left in the
+/// pointer after the object bits.
+///
+/// The offset must be signed, as a pointer can point to the end of the memory
+/// block, and needs to be able to point back to the start.
 /// \param pointer_width: The width of the pointer
 /// \param object_bits : The number of bits used to represent the ID
 /// \return The size in bytes of the maximum allocation supported.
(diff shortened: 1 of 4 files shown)


src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -120,6 +122,25 @@ static std::string architecture_string(T value, const char *s)
...
+/// signed bit, and the bits used for the objects ID (determined by
+/// \p object_bits).
+/// \param pointer_width: The width of the pointer
+/// \param object_bits : The number of bits used to represent the ID
+/// \return The size in bytes of the maximum allocation supported.
+static max_alloc_sizet
+max_malloc_size(std::size_t pointer_width, std::size_t object_bits)
+{
+  PRECONDITION(pointer_width >= 1);
+  PRECONDITION(object_bits < pointer_width - 1);
```
danpoe: I think this should be `object_bits < pointer_width`. Just one bit for the offset would still be ok.

src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -120,6 +122,25 @@ static std::string architecture_string(T value, const char *s)
...
+  PRECONDITION(object_bits >= 1);
+  const auto bits_for_offset = pointer_width - object_bits - 1;
```
danpoe: The naming is not quite right as the bits for the offset is `pointer_width - object_bits`. The offset is signed and the sign bit is considered to be part of the offset.

src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -120,6 +122,25 @@ static std::string architecture_string(T value, const char *s)
          std::string(s) + "=" + std::to_string(value) + ";\n";
 }
 
+using max_alloc_sizet = uint64_t;
```
danpoe: Shouldn't this be `size_t`? That's the type with which `__CPROVER_max_malloc_size` is declared.

src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -120,6 +122,25 @@ static std::string architecture_string(T value, const char *s)
...
+/// \return The size in bytes of the maximum allocation supported.
+static max_alloc_sizet
+max_malloc_size(std::size_t pointer_width, std::size_t object_bits)
+{
+  PRECONDITION(pointer_width >= 1);
+  PRECONDITION(object_bits < pointer_width - 1);
+  PRECONDITION(object_bits >= 1);
+  const auto bits_for_offset = pointer_width - object_bits - 1;
+  PRECONDITION(bits_for_offset < std::numeric_limits<max_alloc_sizet>::digits);
+  return ((max_alloc_sizet)1) << (max_alloc_sizet)bits_for_offset;
```
danpoe: I think casting `bits_for_offset` would be unnecessary.

src/ansi-c/ansi_c_internal_additions.cpp
thk123: I think not: you need one bit for the sign? 
Pointer width = 8
Object bits = 7
1-1 = 0 bits for the offset?

thk123: Just saw second comment - OK I misunderstood what you said - I thought the signed bit was used for something else, rather than the offset?

danpoe: The offset is signed. So with just one bit, 0 (in binary) would be 0 (in decimal), and 1 (in binary) would be -1 (in decimal).

src/ansi-c/ansi_c_internal_additions.cpp
thk123: Yes this makes sense!

thk123: Though the advantage of using a platform agnostic width is writing the unit tests is easier

hannes-steffenhagen-diffblue: @danpoe The problem is that cbmcs `size_t` that we analyze with isn't necessarily the same as that which we get when compiling cbmc itself.

src/ansi-c/ansi_c_internal_additions.cpp
thk123: I'd prefer to be explicit here, since the specific rules of what type gets promoted are fiddly.

danpoe (dismissed):

src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -120,6 +122,29 @@ static std::string architecture_string(T value, const char *s)
          std::string(s) + "=" + std::to_string(value) + ";\n";
 }
 
+/// The maximum allocation size is determined by the number of bits that
+/// are left in the pointer of width \p pointer_width.
+///
+/// The offset needs to be able to represent up to the allocation_size, and
+/// down to -allocation_size, hence 2^(pointer_width - object_bits - 1)
```
danpoe: I think the description has it backwards. The allocation size is a result of how we represent the offset, and not the other way round.

src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -120,6 +122,29 @@ static std::string architecture_string(T value, const char *s)
...
+/// \return The size in bytes of the maximum allocation supported.
+static size_t
+max_malloc_size(std::size_t pointer_width, std::size_t object_bits)
+{
+  PRECONDITION(pointer_width >= 1);
+  PRECONDITION(object_bits < pointer_width);
+  PRECONDITION(object_bits >= 1);
+  const auto offset_bits = pointer_width - object_bits;
+  // We require the offset to be able to express upto allocation_size - 1,
+  // but also down to -allocation_size + 1, therefore the size is allowable
```
danpoe: Should this be just `-allocation_size`? Since we can represent one more negative value than positive values.

src/ansi-c/ansi_c_internal_additions.cpp
danpoe: As pointed out by @thk123 in private communication, when using `size_t` instead of `uint64_t`, the computation would overflow when running on a 32-bit machine and using e.g. `--LLP64`.

src/ansi-c/ansi_c_internal_additions.cpp
thk123: In my head - they are simply intertwined via the following relationship

```
max_alloc_size = 2^(bits_for_offset - 1)
```

That said I didn't like how the function returns max_alloc_size, but the description implies you input the max allocation size. I just found this a clearer way to think about the relationship - I'll have another go.

src/ansi-c/ansi_c_internal_additions.cpp
thk123: Yes you're right - a pointer to the end of the memory would be valid, and you'd need to subtract the full allocation size to get back to the start.

thk123:
@danpoe I've made the suggestde changes: https://github.com/diffblue/cbmc/compare/566b57ce6f60f8c5d80499e9113bb1dc8473a470..8789d7eb42391db0e8a1e4a645256320b93cb5e4#diff-1975841ce4287c4605974098716678fc

src/ansi-c/ansi_c_internal_additions.cpp
danpoe: Yes that makes sense. The current description looks good to me. My reasoning for why I preferred offset -> allocation size was that this is the direction of the computation in the method.

danpoe (approved):

src/ansi-c/ansi_c_internal_additions.cpp
tautschnig: Could we please use `mp_integer` here instead of any architecturally limited type? We do not want to run into limitations of the host or compiler with this computation.

unit/ansi-c/max_malloc_size.cpp
```diff
@@ -0,0 +1,42 @@
+/*******************************************************************\
+
+Module: Unit test for max_malloc_size
+
+Author: Thomas Kiley
+
+\*******************************************************************/
+
+#include <ansi-c/ansi_c_internal_additions.cpp>
+#include <iostream>
```
tautschnig: Is this include actually needed?

tautschnig (approved):
Thank you for fixing the overflow bug, but as said in my comment I believe that the arithmetic should be done on `mp_integer`.

thk123:
@tautschnig done (which of course removes the possibility of any overflow). Will merge when passes CI / tomorrow

unit/ansi-c/max_malloc_size.cpp
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  SECTION("Not enough bits in the pointer")
+  {
+    REQUIRE_THROWS_AS(max_malloc_size(0, 0), invariant_failedt);
+  }
+
+  SECTION("Max allocation size overflow")
+  {
+  }
```
thk123, line 36: Gah - fixed this locally and evidently didn't push :facepalm: will raise a subsequent PR...

thk123: #5317

PR #5301, opened 2020-04-17
Fix overload resolution in options processing

tautschnig:
The changes in 688b90a52e resulted in wrong overload resolution as char*
would resolve to bool, with goto-ld (and possibly specific calls of
goto-cc) producing segmentation faults as isset would no longer imply
that a value was present.

### DIFF
--- a/regression/CMakeLists.txt
+++ b/regression/CMakeLists.txt
@@ -62,3 +62,6 @@ if(WITH_MEMORY_ANALYZER)
 endif()
 
 add_subdirectory(solver-hardness)
+if(NOT WIN32)
+  add_subdirectory(goto-ld)
+endif()
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -31,6 +31,7 @@ DIRS = cbmc \
        linking-goto-binaries \
        symtab2gb \
        solver-hardness \
+       goto-ld \
        # Empty last line
 
 ifeq ($(OS),Windows_NT)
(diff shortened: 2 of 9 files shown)


smowton:
If `const char *` preferentially resolves to bool not `std::string` I suggest adding an explicit overload at the receiving end to resolve that in the preferred direction, otherwise this will surely catch more people in the future

tautschnig:
> If `const char *` preferentially resolves to bool not `std::string` I suggest adding an explicit overload at the receiving end to resolve that in the preferred direction, otherwise this will surely catch more people in the future

I'm not sure I understand what an "explicit overload" is? FWIW, pointer-to-bool should strictly be considered worse than a user-defined conversion sequence (which the `std::string` constructor should provide), but still I've seen this happening.

smowton:
I mean overload the function that currently had `f(bool)` and `f(const std::string &)` and provide an `f(const char *)` to catch this case and direct it towards the string version

tautschnig:
> I mean overload the function that currently had `f(bool)` and `f(const std::string &)` and provide an `f(const char *)` to catch this case and direct it towards the string version

Isn't this what this patch is doing? With "f" == "set?"

smowton (approved):
Doh, right yes, I read it as installing a `std::string{op}` at the use site... actually that's in the exact overload I'm suggesting. Apologies for the noise.

peterschrammel (approved):

kroening:
I would be tempted to remove the `virtual void set(const std::string &option, bool value = true);` and, if we really need it, have
`virtual void set(const std::string &option);`
`virtual void unset(const std::string &option);`

kroening (approved):

tautschnig:
> I would be tempted to remove the `virtual void set(const std::string &option, bool value = true);` and, if we really need it, have
> `virtual void set(const std::string &option);`
> `virtual void unset(const std::string &option);`

Well, that was what 688b90a introduced. While I agree that changes to this API would be preferable, I was afraid it would break some use of this API in another code base.

kroening:
Ok, we can go this the slow way. I'll make a PR that announces these will be deprecated.
The fixes are very easy.

PR #5302, opened 2020-04-17
C front-end: clarify string initialiser control flow

tautschnig:
Moving a special case out of a nested if to make the early return more
obvious. This makes it easier to understand where a zero_initializer may
need to be handled.


martin-cs (approved):
Seems reasonable.

chrisr-diffblue (approved):
Thanks - yes, that's clearer :-)

PR #5303, opened 2020-04-17
C front-end: fix initialisation of large arrays

tautschnig:
In 375e9a8c the construction of zero initialisers was switched to
array_of expressions for arrays with more than 1000 elements. This is
beneficial to avoid running out of memory, but didn't take into account
arrays with designated initialisers, which this patch fixes.


kroening:
Now tempted to always use the `array_of`, even for small arrays!

kroening (approved):

PR #5305, opened 2020-04-18
Satcheck need not implement hardness_collectort

peterschrammel:
Otherwise compilation fails for SAT solvers that don't implement hardness_collectort yet, e.g. Glucose.

- [n/a] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

### DIFF
--- a/src/goto-checker/solver_factory.cpp
+++ b/src/goto-checker/solver_factory.cpp
@@ -183,10 +183,22 @@ make_satcheck_prop(message_handlert &message_handler, const optionst &options)
   auto satcheck = util_make_unique<SatcheckT>(message_handler);
   if(options.is_set("write-solver-stats-to"))
   {
-    satcheck->enable_hardness_collection();
-    satcheck->with_solver_hardness([&options](solver_hardnesst &hardness) {
-      hardness.set_outfile(options.get_option("write-solver-stats-to"));
-    });
+    if(
+      auto hardness_collector = dynamic_cast<hardness_collectort *>(&*satcheck))
+    {
+      hardness_collector->enable_hardness_collection();
+      hardness_collector->with_solver_hardness(
+        [&options](solver_hardnesst &hardness) {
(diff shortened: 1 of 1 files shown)


src/goto-checker/solver_factory.cpp
```diff
@@ -178,12 +178,16 @@ static std::unique_ptr<SatcheckT>
 make_satcheck_prop(message_handlert &message_handler, const optionst &options)
 {
   auto satcheck = util_make_unique<SatcheckT>(message_handler);
-  if(options.is_set("write-solver-stats-to"))
+  if(auto hardness_collector = dynamic_cast<hardness_collectort *>(&*satcheck))
```
hannes-steffenhagen-diffblue: Hm, the idea was that satchecks that don't want to do hardness collection could just implement these functions as no-ops?

peterschrammel: Then you should have implemented it in all existing satchecks, which are broken now. 
Besides, noop functions defeat the purpose of extension interfaces. Thus, I would rather merge this PR.

src/goto-checker/solver_factory.cpp
```diff
@@ -178,12 +178,16 @@ static std::unique_ptr<SatcheckT>
...
   {
-    satcheck->enable_hardness_collection();
-    satcheck->with_solver_hardness([&options](solver_hardnesst &hardness) {
-      hardness.set_outfile(options.get_option("write-solver-stats-to"));
-    });
+    if(options.is_set("write-solver-stats-to"))
```
tautschnig: I think we should provide an error message to the user in case `--write-solver-stats-to` is used with a SAT solver that doesn't support hardness collection.

tautschnig (approved):

martin-cs (approved):
Seems reasonable and probably more intuitive than simply giving zero stats for all other solvers.

PR #5312, opened 2020-04-21
Fix depth_iterator_baset move assignment to return a value.

markrtuttle:
Make the depth_iterator_baset assignment operator return a value.

The type says it should return a value, and VS2019 fails to builds without a return value.


smowton (approved):
Pretty amazed this ever compiled anywhere!

PR #5313, opened 2020-04-21
Use native path separators to make cmake build under vs2019 work.

markrtuttle:
Make cmake use native path separators in cat commands so that the cmake build will work from the Visual Studio command prompt.

I'd like cbmc to build on windows in the native build environment with minimal dependency on external tools, eg requiring chocolatey and cygwin, etc.  Visual Studio comes with cmake and ninja, and Visual Studio can be installed with git, which will give you access to git bash and all the bash shell tools you might need.

My method is

* Install Visual Studio Community 2019
    * Desktop development with C++
    * Individual components -> Code tools -> Git for windows, Git extension for Visual Studio
* Open x64 Native Tools Command Prompt for VS 2019
* PATH="c:\Program Files\Git\usr\bin";%PATH%
* curl -L https://downloads.sourceforge.net/project/winflexbison/win_flex_bison-latest.zip > bison.zip
* unzip bison.zip
* cmake -S. -Bbuild -GNinja 
* cmake --build build

PR #5314, opened 2020-04-22
Build system: make C string converters read files

tautschnig:
Instead of converting stdin, the converters now open and read the files
they are to convert to C strings. This avoids invoking `cat` in build
steps, which helps portability to non-Unix systems.

### DIFF
--- a/src/ansi-c/file_converter.cpp
+++ b/src/ansi-c/file_converter.cpp
@@ -12,34 +12,35 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <iostream>
 #include <string>
 
+static void convert_line(const std::string &line)
+{
+  std::cout << "\"";
+
+  for(std::size_t i = 0; i < line.size(); i++)
+  {
+    const char ch = line[i];
+    if(ch == '\\')
+      std::cout << "\\\\";
+    else if(ch == '"')
+      std::cout << "\\\"";
+    else if(ch == '\r' || ch == '\n')
+    {
+    }
+    else if((ch & 0x80) != 0)
+    {
+      std::cout << "\\x" << std::hex << (unsigned(ch) & 0xff) << std::dec;
+    }
+    else
+      std::cout << ch;
+  }
+
+  std::cout << "\\n\"\n";
+}
+
 int main()
 {
   std::string line;
 
   while(getline(std::cin, line))
-  {
-    std::cout << "\"";
-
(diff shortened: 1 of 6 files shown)


tautschnig:
See discussion in #5306 and #5313 for why this is needed.

kroening (approved):

src/ansi-c/library/converter.cpp
```diff
@@ -15,50 +15,66 @@ bool has_prefix(const std::string &s, const std::string &prefix)
...
+    if(!input_file)
+    {
+      std::cerr << "Failed to open " << argv[i] << '\n';
+      return 1;
+    }
+
+    while(getline(input_file, line))
+    {
+      convert_line(line, first);
+      if(first)
```
hannes-steffenhagen-diffblue: I think you could just remove the if and just make this `first = false;`

hannes-steffenhagen-diffblue (approved):
Looks like the right change to me, thanks for that!

thk123:
FWIW the windows build does now look to pass (though the issue where is can silently fail remains I believe).

src/ansi-c/library/converter.cpp
```diff
@@ -15,50 +15,66 @@ bool has_prefix(const std::string &s, const std::string &prefix)
...
+}
 
-      std::cout << "{ \"" << function << "\",\n";
-      std::cout << "  \"#line 1 \\\"<builtin-library-"
-                << function << ">\\\"\\n\"\n";
-    }
-    else if(!first)
-    {
-      std::cout << "  \"";
+int main(int argc, char* argv[])
```
hannes-steffenhagen-diffblue: clang-format:
```suggestion
int main(int argc, char *argv[])
```

src/ansi-c/file_converter.cpp
```diff
@@ -9,37 +9,52 @@ Author: Daniel Kroening, kroening@kroening.com
...
+      std::cout << "\\x" << std::hex << (unsigned(ch) & 0xff) << std::dec;
+    }
+    else
+      std::cout << ch;
+  }
+
+  std::cout << "\\n\"\n";
+}
+
+int main(int argc, char* argv[])
```
hannes-steffenhagen-diffblue: clang-format:
```suggestion
int main(int argc, char *argv[])
```

markrtuttle:
@tautschnig @hannes-steffenhagen-diffblue Can we merge this pull request now?

PR #5315, opened 2020-04-22
Fix typo in trace validation script

thk123:
The only useful work from this branch is fixing one typo - but hey, might as well.


thk123:
Looks like the two failing jobs that need to turn this off are:
 - (cbmc-linux-cmake-gcc)
 - (cbmc-windows) 

I realise now that it is going to be hard to disable the tests for `make` which makes me sad - so probably the make ones will have to spuriously pass, but make the cmake one at least be safe (and use `-E ` to exclude the xsd tests)

thk123:
cbmc-linux-cmake-gcc: missing xml-line:
```
xmllint not found, trace schema tests will fail 
```

cbmc-windows: 

```
python3 not found: traces not validated against XSD 
```
Last failure (`cbmc-windows-cmake`) looks spurious so have hit retry

hannes-steffenhagen-diffblue (approved):
I think this PR should have a less scary name.

PR #5321, opened 2020-04-24
Tests to check inconsistent assumptions involving __CPROVER_r_ok

danpoe:

### DIFF
--- a/regression/cbmc-primitives/r_w_ok_inconsistent_integer/main.c
+++ b/regression/cbmc-primitives/r_w_ok_inconsistent_integer/main.c
@@ -1,0 +1,14 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  // pointer with object bits = 123 and offset = 123
+  // this is to test with a pointer that does not point to valid memory, but the
+  // value of which is otherwise not special in any way (like for example a null
+  // pointer)
+  char *p = (size_t)123 << (sizeof(char *) * 8 - 8) | 123;
+  __CPROVER_assume(__CPROVER_r_ok(p, 10));
+  assert(0); // fails
+  *p;        // succeeds
+}
--- a/regression/cbmc-primitives/r_w_ok_valid_negated/main.c
+++ b/regression/cbmc-primitives/r_w_ok_valid_negated/main.c
@@ -1,0 +1,34 @@
+#include <assert.h>
+#include <stdlib.h>
+
+char c1;
+
+void main()
+{
+  // We check the conditions for a pointer to global static memory (p1), a
+  // pointer to local memory (p2), and a pointer to dynamic memory (p3).
+  // For each pointer, we check:
+  // (1) whether it is unsafe to access one byte more than was allocated
+  // (2) whether it is unsafe to access one byte before the allocated memory
+  // (3) whether it is unsafe to access one byte after the allocated memory
+
+  char c2;
+
+  char *p1 = &c1;
+
+  assert(!__CPROVER_r_ok(p1, 2));
+  assert(!__CPROVER_r_ok(p1 - 1, 1));
+  assert(!__CPROVER_r_ok(p1 + 1, 1));
+
+  char *p2 = &c2;
+
+  assert(!__CPROVER_r_ok(p2, 2));
+  assert(!__CPROVER_r_ok(p2 - 1, 1));
+  assert(!__CPROVER_r_ok(p2 + 1, 1));
+
+  char *p3 = malloc(1);
+
+  assert(!__CPROVER_r_ok(p3, 2));
+  assert(!__CPROVER_r_ok(p3 - 1, 1));
+  assert(!__CPROVER_r_ok(p3 + 1, 1));
+}
--- a/regression/CMakeLists.txt
+++ b/regression/CMakeLists.txt
@@ -56,6 +56,7 @@ add_subdirectory(goto-cc-file-local)
 add_subdirectory(linking-goto-binaries)
 add_subdirectory(symtab2gb)
 add_subdirectory(validate-trace-xml-schema)
+add_subdirectory(cbmc-primitives)
 
 if(WITH_MEMORY_ANALYZER)
   add_subdirectory(snapshot-harness)
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -33,7 +33,8 @@ DIRS = cbmc \
        solver-hardness \
        goto-ld \
        validate-trace-xml-schema \
-			 # Empty last line
+       cbmc-primitives \
+       # Empty last line
 
 ifeq ($(OS),Windows_NT)
     detected_OS := Windows
--- a/regression/cbmc-primitives/CMakeLists.txt
+++ b/regression/cbmc-primitives/CMakeLists.txt
@@ -1,0 +1,3 @@
+add_test_pl_tests(
+    "$<TARGET_FILE:cbmc>"
+)
--- a/regression/cbmc-primitives/Makefile
+++ b/regression/cbmc-primitives/Makefile
@@ -1,0 +1,18 @@
+default: tests.log
+
+test:
+	@../test.pl -e -p -c ../../../src/cbmc/cbmc
+
+tests.log: ../test.pl
+	@../test.pl -e -p -c ../../../src/cbmc/cbmc
+
+show:
+	@for dir in *; do \
+		if [ -d "$$dir" ]; then \
+			vim -o "$$dir/*.c" "$$dir/*.out"; \
+		fi; \
+	done;
+
+clean:
+	find -name '*.out' -execdir $(RM) '{}' \;
+	$(RM) tests.log
--- a/regression/cbmc-primitives/r_w_ok_bug/main.c
+++ b/regression/cbmc-primitives/r_w_ok_bug/main.c
@@ -1,0 +1,9 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char *p1;
+  __CPROVER_assume(__CPROVER_r_ok(p1 - 1, 1));
+  *p1;
+}
--- a/regression/cbmc-primitives/r_w_ok_bug/test.desc
+++ b/regression/cbmc-primitives/r_w_ok_bug/test.desc
@@ -1,0 +1,9 @@
+KNOWNBUG
+main.c
+--pointer-check --no-simplify --no-propagation
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+Crashes during the flattening, issue #5328
--- a/regression/cbmc-primitives/r_w_ok_inconsistent_dead/main.c
+++ b/regression/cbmc-primitives/r_w_ok_inconsistent_dead/main.c
@@ -1,0 +1,16 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char *p;
+
+  {
+    char c;
+    p = &c;
+  }
+
+  __CPROVER_assume(__CPROVER_r_ok(p, 1));
+  assert(0); // fails
+  *p;        // succeeds
+}
--- a/regression/cbmc-primitives/r_w_ok_inconsistent_dead/test-no-cp.desc
+++ b/regression/cbmc-primitives/r_w_ok_inconsistent_dead/test-no-cp.desc
@@ -1,0 +1,13 @@
+FUTURE
+main.c
+--pointer-check --no-simplify --no-propagation
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^warning: ignoring
+--
+Tests that an inconsistent assumption involving a pointer to an out of scope
+variable and __CPROVER_r_ok() can be detected via assert(0). We use
+--no-simplify and --no-propagation to ensure that the case is not solved by the
+constant propagation and thus tests the constraint encoding.
--- a/regression/cbmc-primitives/r_w_ok_inconsistent_dead/test.desc
+++ b/regression/cbmc-primitives/r_w_ok_inconsistent_dead/test.desc
@@ -1,0 +1,11 @@
+FUTURE
+main.c
+--function test_global_pointer --pointer-check
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^warning: ignoring
+--
+Tests that an inconsistent assumption involving a pointer to an out of scope
+variable and __CPROVER_r_ok() can be detected via assert(0).
--- a/regression/cbmc-primitives/r_w_ok_inconsistent_deallocated/main.c
+++ b/regression/cbmc-primitives/r_w_ok_inconsistent_deallocated/main.c
@@ -1,0 +1,12 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char *p = malloc(1);
+  free(p);
+
+  __CPROVER_assume(__CPROVER_r_ok(p, 1));
+  assert(0); // fails
+  *p;        // succeeds
+}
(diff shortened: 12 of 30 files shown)


regression/cbmc/r_w_ok_inconsistent_dead/test-global-no-cp.desc
```diff
@@ -0,0 +1,10 @@
+FUTURE
+main.c
+--function test_global_pointer --pointer-check --no-simplify --no-propagation
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+Tests that an inconsistent assumption involving a pointer to an out of scope
```
thk123: :green_heart: Thanks

regression/cbmc/r_w_ok_inconsistent_dead/test-global-no-cp.desc
```diff
@@ -0,0 +1,10 @@
+FUTURE
+main.c
+--function test_global_pointer --pointer-check --no-simplify --no-propagation
+^EXIT=0$
+^SIGNAL=0$
```
thk123: :bulb: I think some more agressive checks (e.g verification successful?) would be good

regression/cbmc/r_w_ok_inconsistent_dead/test-global-no-cp.desc
thk123: Though an explanation of why FUTURE (and link to relevant ticket) would be good

regression/cbmc/r_w_ok_inconsistent_dead/test-global-no-cp.desc
```diff
@@ -0,0 +1,10 @@
+FUTURE
+main.c
+--function test_global_pointer --pointer-check --no-simplify --no-propagation
```
thk123: :question: Could we include somewhere an explanation why the `no-propagation` flag is needed to be tested separately?

regression/cbmc/r_w_ok_inconsistent_integer/main.c
```diff
@@ -0,0 +1,22 @@
+#include <assert.h>
+#include <stdlib.h>
+
+char *p;
+
+void test_global_pointer()
+{
+  p = (size_t)123 << (sizeof(char *) * 8 - 1) | 123;
```
thk123: :question: An explanation in line as to what this is trying to achieve would be useful. Does it need to be this complicated? Can you just go `p = 1234;`?

regression/cbmc/r_w_ok_inconsistent_integer/test-global.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--function test_global_pointer --pointer-check
+^EXIT=0$
```
thk123: Please check verification failed and that the assert fails :tophat:

regression/cbmc/r_w_ok_inconsistent_integer/test-local.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--function test_local_pointer --pointer-check
+^EXIT=0$
```
thk123: :tophat:

regression/cbmc/r_w_ok_inconsistent_invalid/main.c
```diff
@@ -0,0 +1,24 @@
+#include <assert.h>
+#include <stdlib.h>
+
+char *p;
+
+void test_global_pointer()
+{
+  // special value of invalid pointer
```
thk123: :question: Why is it special?

regression/cbmc/r_w_ok_inconsistent_nondet/main.c
```diff
@@ -0,0 +1,21 @@
+#include <assert.h>
+
+char *p;
+
+void test_global_pointer()
+{
+  p = nondet();
```
thk123: :pick: to supress warnings, perhaps forward declare `char* nondet();`?

regression/cbmc/r_w_ok_inconsistent_valid/main.c
```diff
@@ -0,0 +1,62 @@
...
+  __CPROVER_assume(
+    __CPROVER_r_ok(p1, 2) ||
+    __CPROVER_r_ok(p1 - 1, 1) ||
+    __CPROVER_r_ok(p1 + 1, 1) ||
+    __CPROVER_r_ok(p2, 2) ||
+    __CPROVER_r_ok(p2 - 1, 1) ||
+    __CPROVER_r_ok(p2 + 1, 1) ||
+    __CPROVER_r_ok(p3, 2) ||
+    __CPROVER_r_ok(p3 - 1, 1) ||
+    __CPROVER_r_ok(p3 + 1, 1));
```
thk123: An english summary of what is being assumed here would be helpful (like a high level intention)

thk123: I think it's just checking a bunch of invalid assumes and seeing that none of them are true?

thk123: So add comment:

```
// Validate pointers with off by one validation assumptions
// (just before, just after the valid memory)
 // one byte to much
```

And perhaps, if it still works, group by pointer

regression/cbmc/r_w_ok_inconsistent_valid/test-global-no-cp.desc
```diff
@@ -0,0 +1,11 @@
+KNOWNBUG
```
thk123: Why is this knownbug rather than future?

regression/cbmc/r_w_ok_inconsistent_valid/test-global-no-cp.desc
```diff
@@ -0,0 +1,11 @@
...
+main.c
+--function test_global_pointer --pointer-check --no-simplify --no-propagation
+^EXIT=0$
```
thk123: When this works it should be exit=10 right? I'd right the test correct so that someone doesn't think it works and turns it on

regression/cbmc/r_w_ok_inconsistent_valid_negated/main.c
```diff
@@ -0,0 +1,54 @@
...
+  char c2;
+
+  p1 = &c1;
+  p2 = &c2;
+
+  p3 = malloc(1);
+  __CPROVER_assume(p1 != NULL);
+  
+  __CPROVER_assume(
+    !__CPROVER_r_ok(p1, 1) ||
```
thk123: I'd highlight the negation here with a comment explaining the purpose

thk123: Ehh only if you think worth it - it is of course also explained in the description

thk123 (commented):
Looks good - some general comments, also I note these are only for r_ok, we presumably need them for the other primitives. 

Overall comments:
 - Do we need to test the local and global pointer every time? Produces a lot of duplication, do they end up looking very different in the formulas?
 - Should we be testing pointers to memory mallocd with > size of the object (e.g. arrays)
 - Should we be testing pointers to statically allocated arrays 
 - Should we best testing pointers that are passed in from another function 

Also, this is a lot of quite specific tests. I'd be tempted to move them out of `regression/cbmc` (perhaps into a folder called `cprover-primitives`, or perhaps tag them as THROUGH when they are enabled

regression/cbmc/r_w_ok_inconsistent_dead/main.c
```diff
@@ -0,0 +1,30 @@
...
+
+void test_global_pointer()
+{
+  {
+    char c;
+    p = &c;
+  }
+
+  __CPROVER_assume(__CPROVER_r_ok(p, 1));
+  assert(0); // fails
```
danielsn: What does "fails" mean.  Do you really mean unreachable?

regression/cbmc/r_w_ok_inconsistent_dead/test-global-no-cp.desc
```diff
@@ -0,0 +1,10 @@
+FUTURE
```
danielsn: What does FUTURE mean?

regression/cbmc/r_w_ok_inconsistent_dead/main.c
danpoe: It means that the assertion fails, i.e. is reachable.

regression/cbmc/r_w_ok_inconsistent_dead/test-global-no-cp.desc
danpoe: Tests that currently don't pass but are not bugs are marked as FUTURE. This is typically used for tests for unimplemented features. Here the test doesn't pass due to the overapproximate behaviour of `__CPROVER_r_ok()`. If it will be made more pecise in the future it should pass.

danpoe:
@thk123 Indeed the separate tests for global and local pointers were unnecessary. The intention was to account for differences in the result of the local bitvector analysis but it gives the same result for both cases.

regression/cbmc/r_w_ok_inconsistent_valid/test-global-no-cp.desc
danpoe: This particular test made cbmc crash. I'll add a comment to say that, and will add a simpler test too that shows the same bug.

regression/cbmc/r_w_ok_inconsistent_integer/test-global.desc
danpoe: This test should succeed :rabbit:

regression/cbmc/r_w_ok_inconsistent_integer/test-local.desc
danpoe: :rabbit:

regression/cbmc/r_w_ok_inconsistent_invalid/main.c
danpoe: Internally,`1 << 56` (when `sizeof(char *) = 8` and `object_bits = 8`) is a special value which at various points is used to denote an invalid pointer (i.e., the object with value 1). `__CPROVER_is_invalid_pointer()` just checks for this value.

regression/cbmc/r_w_ok_inconsistent_valid/test-global-no-cp.desc
danpoe: No when it works it should be `EXIT=0`.

regression/cbmc/r_w_ok_inconsistent_nondet/main.c
danpoe: Done

regression/cbmc/r_w_ok_inconsistent_invalid/main.c
thk123: This explanation should therefore be in the code, perhaps as:
```
size_t object_bits = 8;
// special value of invalid pointer see __CPROVER_is_invalid_pointer
char *p = (size_t)1 << (sizeof(char *) * object_bits - object_bits);
```

regression/cbmc/r_w_ok_bug/test.desc
```diff
@@ -0,0 +1,9 @@
+KNOWNBUG
+main.c
+--pointer-check --no-simplify --no-propagation
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+Crashes during the flattening
```
thk123: This is pretty bad! I guess we should raise a bug?

regression/cbmc/r_w_ok_inconsistent_dead/test-global-no-cp.desc
danpoe: Added

regression/cbmc/r_w_ok_inconsistent_integer/main.c
danpoe: Added a description. It's creating a pointer with object bits being 123 and the offset being 123 (also noticed that it should be -8 instead of -1).

regression/cbmc/r_w_ok_inconsistent_invalid/main.c
danpoe: Added

regression/cbmc/r_w_ok_inconsistent_valid/main.c
danpoe: Added

regression/cbmc/r_w_ok_inconsistent_dead/test-global-no-cp.desc
danpoe: Do you think I should add an explanation to every .desc file? Basically we want to test with `--no-simplify --no-propagation` (which turns off expression simplification and constant propagation) and without those two, so as to check that both the constraint encoding and the simplification/constant propagation work correctly.

danpoe:
@thk123 I've now put the tests in a new directory `cbmc-primitives`. The tests are very quick so don't need to be tagged as THOROUGH.

Regarding your points (2) and (3) above, we don't necessarily need to separately test these cases as the behaviour is exactly the same for malloced objects of size 1 or > 1 and for static objects of size 1 or size > 1 (i.e. statically allocated arrays). Also the behaviour will be the same for pointers passed into functions.

regression/cbmc/r_w_ok_inconsistent_dead/test-global-no-cp.desc
danpoe: I think the test and the test description already provide this explanation. The FUTURE just indicates that the test does not currently pass.

regression/cbmc-primitives/r_w_ok_inconsistent_integer/main.c
```diff
@@ -0,0 +1,11 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  // pointer with object bits = 123 and offset = 123
```
thk123, line 6: Thanks for the improvement - as the reader, I'm still left with the question of _why_ you're constructing such a pointer. I think this would still be improved with a comment as to why this is an interesting pointer. Is it because it tricks cbmc into thinking it is valid?

regression/cbmc-primitives/r_w_ok_valid_negated/main.c
```diff
@@ -0,0 +1,34 @@
...
+
+  assert(!__CPROVER_r_ok(p2, 2));
+  assert(!__CPROVER_r_ok(p2 - 1, 1));
+  assert(!__CPROVER_r_ok(p2 + 1, 1));
+
+  char *p3 = malloc(1);
+
+  assert(!__CPROVER_r_ok(p3, 2));
+  assert(!__CPROVER_r_ok(p3 - 1, 1));
+  assert(!__CPROVER_r_ok(p3 + 1, 1));
```
thk123, line 33: This is much better imo :green_heart:

regression/cbmc/r_w_ok_bug/test.desc
thk123: We should raise a bug/cbmc issue for this and mention it in the .desc

thk123 (approved):
Probably what CI is telling you, but your new folder will need a `Makefile` and `CMakeLists.txt` to tell it how to behave. (A short `README.md` explaining the purpose of the folder would be appreciated, but not required). 

> Do you think I should add an explanation to every .desc file? Basically we want to test with --no-simplify --no-propagation (which turns off expression simplification and constant propagation) and without those two, so as to check that both the constraint encoding and the simplification/constant propagation work correctly.

Frankly yes - since no way to know what test will be opened, best to put it in all of them. Should of course just be the same sentence copy-pasted! 

Anyways, thanks for the tweaks, blocking comments addressed - though would appreciate raising and referencing bugs for the crashes to be sure we turn the tests back on when they're fixed. Just need to add the Make stuff to the folder and we should be good.

regression/cbmc-primitives/r_w_ok_inconsistent_integer/main.c
danpoe: Yes, it's a pointer that doesn't point to valid memory but isn't special in any other way. Added a description.

regression/cbmc/r_w_ok_bug/test.desc
danpoe: Issue recorded as #5328, and added the number to the .desc file.

PR #5324, opened 2020-04-27
Add tests to check semantics of pointer primitives

danpoe:


regression/cbmc/dynamic-object-02/test-no-cp.desc
```diff
@@ -0,0 +1,9 @@
+FUTURE
+main.c
+--no-simplify --no-propagation
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+Check that uninitialized pointers are not considered pointing to dynamic objects
```
danielsn: Is this what we want? Or should it be non-det whether this is true?  

In particular, is there a doc where all these expectations are laid out?

regression/cbmc/same-object-01/test-no-cp.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--no-simplify --no-propagation
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^warning: ignoring
+--
+Check that __CPROVER_same_object() is true when given a pointer variable, and an
```
danielsn: Would it be worth it check that an offset into the object also maintains the same object property?

regression/cbmc/same-object-03/test-no-cp.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--no-simplify --no-propagation
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^warning: ignoring
+--
+Check that __CPROVER_same_object() is false when given two pointers, with one
```
danielsn: Is that expected? malloc is allowed to (and often does) reuse objects

danielsn:
My overall question: is there a doc explaining the expected semantics of each primitive? When given a reference to (list not necessarily complete)

1. a global
1. a stack variable
1. a malloced variable
1. an unconstrained variable
1. a freed / out of scope variable
1. an out of bounds access on a valid pointer

danielsn:
Also, what happens if we semi-constrain a variable? e.g. 
```
int *x;
__CPROVER_assume(x != null); // or __CPROVER_assume(x == null)
assert(__CPROVER_dynamic_object(x));
```

regression/cbmc/dynamic-object-01/test-no-cp.desc
```diff
@@ -0,0 +1,10 @@
+FUTURE
+main.c
+--no-simplify --no-propagation
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+Check that the dynamic object property is havoc'd when dynamic objects are
```
thk123: As usual - I'd suggest all tests with FUTURE or KNOWNBUG should have some kind of JIRA reference for when they are to be turned on, otherwise people will just forget about them

thk123: In this case - shouldn't `EXIT=10`?

regression/cbmc/dynamic-object-02/test-no-cp.desc
thk123: I think I agree with @danielsn, though I'm not clear what I as a user am trying to get out asserting on this property. But in general, seems safest for it to fail if you assert either way (in my mental model - you're asserting on UB, so whatever you assert, it should be false)

regression/cbmc/object-size-01/main.c
```diff
@@ -0,0 +1,17 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char *p = malloc(1);
+  free(p);
+
+  assert(__CPROVER_OBJECT_SIZE(p) == 1);
```
thk123: :question: Should you also `assert(... != 1` and get a failed for that too?

regression/cbmc/pointer-offset-01/test-no-cp.desc
```diff
@@ -0,0 +1,11 @@
+FUTURE
+main.c
+--no-simplify --no-propagation
+^EXIT=10$
+^SIGNAL=0$
+\[main.assertion.1\] line \d+ assertion __CPROVER_POINTER_OFFSET\(p\) >= 0: FAILURE
+--
+^warning: ignoring
+--
+Check that a negative offset can be chosen for uninitialized pointers. Currently
```
thk123: Again - I sort of feel like the correct behavior for something like this is, asserts always fail? Like the offset isn't definitely positive, but also isn't definitely negative - it should be havoc'd to be any value

regression/cbmc/same-object-01/main.c
```diff
@@ -0,0 +1,11 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char *p = malloc(1);
+  assert(__CPROVER_POINTER_OBJECT(p) == 2);
+
+  assert(
```
thk123: Probably worth asserting that it _isn't_ the same object as something else?

regression/cbmc/same-object-03/test-no-cp.desc
thk123: I think this depends on how you plan to use `_same_object`? `assert(same_object(p1, p2)` certainly must fail. I think since this is an internal, it should be more exact, in this case, we know they are _not_ the same object, they are the result of two separate mallocs. However, I would expect `assert(p1 != p2)` to fail, since is possible they could have the same address since the memory has been freed.

regression/cbmc/dynamic-object-01/main.c
```diff
@@ -0,0 +1,10 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char *p = malloc(1);
+  free(p);
+
+  assert(!__CPROVER_DYNAMIC_OBJECT(p));
```
thk123: to check it is havoc'd you should add an `assert(dynamic_object(p))`
```suggestion
  assert(!__CPROVER_DYNAMIC_OBJECT(p));
  assert(__CPROVER_DYNAMIC_OBJECT(p));
```

regression/cbmc/dynamic-object-02/main.c
```diff
@@ -0,0 +1,8 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char *p;
+  assert(!__CPROVER_DYNAMIC_OBJECT(p));
```
thk123:
```suggestion
  assert(!__CPROVER_DYNAMIC_OBJECT(p));
  assert(__CPROVER_DYNAMIC_OBJECT(p));
```

thk123 (commented):
Looks good - I think we need to check some of them are actually havoicing by adding asserts for either direction

regression/cbmc/dynamic-object-01/test-no-cp.desc
danpoe: Added JIRA ticket references. Also changed it to `EXIT=0`.

regression/cbmc/dynamic-object-02/test-no-cp.desc
danpoe: @thk123 cbmc doesn't necessarily behave like that on UB. In general, one needs the appropriate check flags to be enabled to catch certain cases. For example, signed integer overflow is UB, though one needs to enable `--signed-overflow-check` to catch that case. Otherwise, the values will just wrap around.

In general though, I think it's a good idea for operations to be nondet in such cases as far as possible. So yes, I've changed the test to check for both the negated and the non-negated condition, and have specified that both should fail in the test.desc file.

regression/cbmc/same-object-01/test-no-cp.desc
danpoe: Added

regression/cbmc/same-object-03/test-no-cp.desc
danpoe: Yes, `assert(p1 != p2)` should fail, though currently the way `malloc` is modelled in cbmc it doesn't reuse addresses.

regression/cbmc/object-size-01/main.c
danpoe: Added

regression/cbmc/pointer-offset-01/test-no-cp.desc
danpoe: Added

regression/cbmc/same-object-01/main.c
danpoe: Added

regression/cbmc/dynamic-object-01/main.c
danpoe: Added

regression/cbmc/dynamic-object-02/main.c
danpoe: Added

thk123 (approved):
Thanks seems good (clang-format is complaining on travis)

PR #5325, opened 2020-04-28
Add a copy constructor for goto_program

hannes-steffenhagen-diffblue:
The absence of a copy constructor for `goto_programt` is currently preventing us from copying `goto_model`s. As far as I can the only problem with addying the copy constructor were the internal targets, which _should_ be local to each `goto_programt` and thus can be simply reconstructed from the offset to the start of the program.

### DIFF
--- a/src/goto-programs/goto_program.cpp
+++ b/src/goto-programs/goto_program.cpp
@@ -11,8 +11,9 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "goto_program.h"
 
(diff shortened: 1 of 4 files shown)


pkesseli:
This seems to be duplicating `goto_programt::copy_from`? I think the reason `copy_from` was used instead of a copy constructor was to prevent accidental copying, since copying entire GOTO models is expensive. At any rate, you can probably implement copying `goto_modelt`s by just relying on `copy_from`.

Issue #4956, opened 2019-07-29
CMake build failure on macOS: __libcheck.c:4:10: fatal error: 'err.h' file not found


karkhaz:
[Edit] **Current status:** the workaround to this bug is to set the absolute path to the compiler when configuring CBMC, like this:
```sh
> cmake -DCMAKE_C_COMPILER=/usr/bin/clang ...
```
____
original bug report follows
____

CBMC fails to build with CMake on macOS. For some reason, building using the Makefile in the src directory works.

I've been unable to find a commit that broke this, and I also recently updated my system (both macOS itself and also the command-line tools). So I'm wondering if that was to blame. I haven't been able to find any information online about moved headers, though. `err.h` still exists:
```sh
> find /Library/Developer -name err.h
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/err.h
```
but I don't have an older system to compare against.

Command:
```sh
> rm -rf build/*
> cmake -H. -Bbuild -GNinja -DWITH_JBMC=0 -DCMAKE_CXX_FLAGS="-fcolor-diagnostics"
> cmake --build build
# ...
[18/803] Generating library-check.stamp
FAILED: src/ansi-c/library-check.stamp
[...]

PR #5339, opened 2020-05-13
Document specifying compiler path when using CMake

karkhaz:
This commit adds a note to COMPILING.md, explaining that the full path
to the compiler needs to be given to CMake in order to build on macOS.
This is a workaround for the issue in https://github.com/diffblue/cbmc/issues/4956

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -255,6 +255,14 @@ require manual modification of build files.
    to generate IDE projects by supplying the `-G` flag.  Run `cmake -G` for a
    comprehensive list of supported back-ends.
 
+   On macOS >10.14, the build will fail unless you explicitly specify
+   the full path to the compiler. This issue is being tracked
+   [here](https://github.com/diffblue/cbmc/issues/4956). The invocation thus
+   looks like this:
+   ```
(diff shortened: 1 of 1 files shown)


smowton (approved):
Better than nothing, but can we have cmake make a configure-time check and emit a helpful error if it notices we're on OSX and err.h isn't found?

karkhaz:
I did try this back when I reported the bug, but it didn't work reliably. The problem isn't that err.h doesn't exist, it's that the clang that CMake uses (sometimes) doesn't have err.h in the include path. So you would either need to run clang on a test program at configure-time (I didn't try doing that, not sure whether it's possible or idiomatic), or you would need to detect the macOS minor version number.

I was focusing on trying to fix the bug (e.g. if the minor version is >14, then override CMAKE_C_COMPILER), but that's also bad if the user wants to set a different compiler entirely. You seem to be suggesting to let it fail but to emit a warning, which could work...as long as we can reliably detect when the build would actually fail.

So in summary: I'm not sure what the fix is, it isn't as obvious as it might seem. But I'd like to document the workaround meanwhile. I can't count the number of times that I've had to point people to #4956 when they've tried to compile CBMC by following the instructions, we can remove this from the documentation when it's actually fixed.

PR #4928, opened 2019-07-19
Stop asserting positive array size

xbauch:
when generating non-deterministic arrays. Zero-sized arrays are (unfortunately)
legal C construct (as for example the regression test shows).

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #5342, opened 2020-05-13
gen_nondet_array_init: arrays may have size 0

kroening:
The various C standards clearly allow arrays of size 0.

### DIFF
--- a/src/ansi-c/c_nondet_symbol_factory.cpp
+++ b/src/ansi-c/c_nondet_symbol_factory.cpp
@@ -172,9 +172,9 @@ void symbol_factoryt::gen_nondet_array_init(
   auto const &array_type = to_array_type(expr.type());
   const auto &size = array_type.size();
   PRECONDITION(size.id() == ID_constant);
-  auto const array_size = numeric_cast_v<size_t>(to_constant_expr(size));
+  auto const array_size = numeric_cast_v<mp_integer>(to_constant_expr(size));
   DATA_INVARIANT(array_size >= 0, "Arrays must have non-negative size");
-  for(size_t index = 0; index < array_size; ++index)
+  for(mp_integer index = 0; index < array_size; ++index)
   {
     gen_nondet_init(
       assignments,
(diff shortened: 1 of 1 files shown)


hannes-steffenhagen-diffblue:
Zero-length arrays are _not_ allowed in standard C. It's a common enough extension to allow 0 as the size, but in that case it's normally used to denote arrays of variable length rather than zero length (https://gcc.gnu.org/onlinedocs/gcc-4.7.2/gcc/Zero-Length.html).

Since gcc allows it we probably should too though.

src/ansi-c/c_nondet_symbol_factory.cpp
```diff
@@ -173,7 +173,7 @@ void symbol_factoryt::gen_nondet_array_init(
   const auto &size = array_type.size();
   PRECONDITION(size.id() == ID_constant);
   auto const array_size = numeric_cast_v<size_t>(to_constant_expr(size));
-  DATA_INVARIANT(array_size > 0, "Arrays should have positive size");
+  DATA_INVARIANT(array_size >= 0, "Arrays must have non-negative size");
```
tautschnig, line 176: Could we please have a test that covers the case of size 0?

src/ansi-c/c_nondet_symbol_factory.cpp
```diff
@@ -172,9 +172,9 @@ void symbol_factoryt::gen_nondet_array_init(
   auto const &array_type = to_array_type(expr.type());
   const auto &size = array_type.size();
   PRECONDITION(size.id() == ID_constant);
-  auto const array_size = numeric_cast_v<size_t>(to_constant_expr(size));
+  auto const array_size = numeric_cast_v<mp_integer>(to_constant_expr(size));
   DATA_INVARIANT(array_size >= 0, "Arrays must have non-negative size");
-  for(size_t index = 0; index < array_size; ++index)
+  for(mp_integer index = 0; index < array_size; ++index)
```
tautschnig, line 177: While this makes sense in spirit, does this actually make a difference in practice (other than slightly worse performance)? We wouldn't be able to allocate memory for an object larger than max(size_t) anyway.

src/ansi-c/c_nondet_symbol_factory.cpp
tautschnig: Just a note to perhaps help make progress on this one: I believe #4928 has such a test.

PR #5343, opened 2020-05-14
Output C code from harness generation tool.

NlightNFotis:

### DIFF
--- a/regression/goto-harness/havoc-global-int-02/test.desc
+++ b/regression/goto-harness/havoc-global-int-02/test.desc
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
 main.c
 --harness-type initialise-with-memory-snapshot --memory-snapshot ../load-snapshot-json-snapshots/global-int-x-y-snapshot.json --initial-goto-location main:7 --havoc-variables y
 ^\[main.assertion.1\] line \d+ assertion y \+ 2 > y: FAILURE$
@@ -7,3 +7,5 @@ main.c
 ^SIGNAL=0$
 --
 ^warning: ignoring
+--
+More information can be found in github#5351.
--- a/src/Makefile
+++ b/src/Makefile
@@ -8,8 +8,8 @@ DIRS = analyses \
        goto-cc \
        goto-checker \
        goto-diff \
-       goto-harness \
        goto-instrument \
+       goto-harness \
        goto-programs \
        goto-symex \
        jsil \
@@ -66,7 +66,8 @@ languages: util.dir langapi.dir \
 solvers.dir: util.dir
 
 goto-harness.dir: util.dir goto-programs.dir langapi.dir linking.dir \
-                  json.dir json-symtab-language.dir
+                  json.dir json-symtab-language.dir \
+                  goto-instrument.dir
 
 goto-instrument.dir: languages goto-programs.dir pointer-analysis.dir \
                      goto-symex.dir linking.dir analyses.dir solvers.dir
--- a/src/goto-harness/Makefile
+++ b/src/goto-harness/Makefile
@@ -16,6 +16,10 @@ OBJ += \
   ../linking/linking$(LIBEXT) \
   ../json/json$(LIBEXT) \
   ../json-symtab-language/json-symtab-language$(LIBEXT) \
+  ../ansi-c/ansi-c$(LIBEXT) \
+  ../cpp/cpp$(LIBEXT) \
+  ../goto-instrument/dump_c$(OBJEXT) \
+  ../goto-instrument/goto_program2code$(OBJEXT) \
   # Empty last line
 
 INCLUDES= -I ..
--- a/src/goto-harness/function_call_harness_generator.cpp
+++ b/src/goto-harness/function_call_harness_generator.cpp
@@ -478,7 +478,7 @@ function_call_harness_generatort::implt::declare_arguments(
   for(const auto &parameter : parameters)
   {
     auto argument = allocate_objects.allocate_automatic_local_object(
-      parameter.type(), parameter.get_base_name());
+      remove_const(parameter.type()), parameter.get_base_name());
     parameter_name_to_argument_name.insert(
       {parameter.get_base_name(), argument.get_identifier()});
     arguments.push_back(argument);
--- a/src/goto-harness/function_call_harness_generator.cpp
+++ b/src/goto-harness/function_call_harness_generator.cpp
@@ -281,7 +281,7 @@ void function_call_harness_generatort::implt::generate_nondet_globals(
     for(const auto &symbol_table_entry : *symbol_table)
     {
       const auto &symbol = symbol_table_entry.second;
-      if(recursive_initialization->is_initialization_allowed(symbol))
+      if(recursive_initializationt::is_initialization_allowed(symbol))
       {
         globals.push_back(symbol.symbol_expr());
       }
--- a/src/goto-harness/goto_harness_parse_options.cpp
+++ b/src/goto-harness/goto_harness_parse_options.cpp
@@ -9,23 +9,56 @@ Author: Diffblue Ltd.
 #include "goto_harness_parse_options.h"
 
 #include <cstddef>
+#include <fstream>
 #include <set>
 #include <string>
+#include <unordered_set>
 #include <utility>
 
+#include <goto-instrument/dump_c.h>
 #include <goto-programs/goto_convert.h>
 #include <goto-programs/goto_model.h>
 #include <goto-programs/read_goto_binary.h>
+#include <goto-programs/show_symbol_table.h>
 #include <goto-programs/write_goto_binary.h>
+#include <langapi/mode.h>
 #include <util/config.h>
 #include <util/exception_utils.h>
 #include <util/exit_codes.h>
+#include <util/expr_iterator.h>
 #include <util/invariant.h>
 #include <util/version.h>
 
 #include "function_call_harness_generator.h"
 #include "goto_harness_generator_factory.h"
 #include "memory_snapshot_harness_generator.h"
+
+std::unordered_set<irep_idt>
+get_symbol_names_from_goto_model(const goto_modelt &goto_model)
+{
+  auto symbols = std::unordered_set<irep_idt>{};
+  std::transform(
+    goto_model.get_symbol_table().begin(),
+    goto_model.get_symbol_table().end(),
+    std::inserter(symbols, symbols.end()),
+    [](const std::pair<const irep_idt, symbolt> &key_value_pair) {
+      return key_value_pair.first;
+    });
+  return symbols;
+}
+
+static void filter_goto_model(
+  goto_modelt &goto_model_with_harness,
+  const std::unordered_set<irep_idt> &goto_model_without_harness_symbols)
+{
+  for(auto &symbol : goto_model_with_harness.get_symbol_table())
+  {
+    if(
+      goto_model_without_harness_symbols.count(symbol.first) == 1 &&
+      symbol.second.is_function())
+      goto_model_with_harness.unload(symbol.first);
+  }
+}
 
 // The basic idea is that this module is handling the following
 // sequence of events:
@@ -62,6 +95,8 @@ int goto_harness_parse_optionst::doit()
                                      got_harness_config.in_file + "'"};
   }
   auto goto_model = std::move(read_goto_binary_result.value());
+  auto const goto_model_without_harness_symbols =
+    get_symbol_names_from_goto_model(goto_model);
 
   // This has to be called after the defaults are set up (as per the
   // config.set(cmdline) above) otherwise, e.g. the architecture specification
@@ -87,13 +122,15 @@ int goto_harness_parse_optionst::doit()
   harness_generator->generate(
     goto_model, got_harness_config.harness_function_name);
 
-  // Write end result to new goto-binary
-  if(write_goto_binary(
-       got_harness_config.out_file, goto_model, ui_message_handler))
-  {
-    throw system_exceptiont{"failed to write goto program from file '" +
-                            got_harness_config.out_file + "'"};
-  }
+  filter_goto_model(goto_model, goto_model_without_harness_symbols);
+  auto harness_out = std::ofstream{got_harness_config.out_file};
+  dump_c(
+    goto_model.goto_functions,
+    true,
+    true,
+    false,
+    namespacet{goto_model.get_symbol_table()},
+    harness_out);
 
   return CPROVER_EXIT_SUCCESS;
 }
--- a/src/goto-harness/recursive_initialization.cpp
+++ b/src/goto-harness/recursive_initialization.cpp
@@ -479,10 +479,10 @@ static symbolt &get_fresh_global_symbol(
   irep_idt mode)
 {
   source_locationt source_location{};
-  source_location.set_file(CPROVER_PREFIX "harness.c");
+  source_location.set_file(GOTO_HARNESS_PREFIX "harness.c");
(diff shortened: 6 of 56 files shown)


regression/goto-harness/havoc-global-int-02/main.c
```diff
@@ -3,7 +3,7 @@
 unsigned int x;
 unsigned int y;
 
-unsigned int nondet_int()
+unsigned int some_random_unsigned_integer()
```
hannes-steffenhagen-diffblue: This was changed because of a `dump_c()` bug; It assumes that `nondet_*` names are available with specific meanings.

regression/goto-harness/havoc-global-int-02/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
hannes-steffenhagen-diffblue, line 1: We can work around this for now by re-enabling the old behaviour for these, but we've marked them KNOWNBUG for now because we're first trying to get the call-function harness to behave like we want it to.

src/goto-harness/common_harness_generator_options.h
```diff
@@ -18,6 +18,7 @@ Author: Diffblue Ltd.
   "function-pointer-can-be-null"
 #define COMMON_HARNESS_GENERATOR_HAVOC_MEMBER_OPT "havoc-member"
 
+
```
hannes-steffenhagen-diffblue: 🚫 Whitespace

src/goto-harness/function_call_harness_generator.cpp
```diff
@@ -215,6 +215,8 @@ void function_call_harness_generatort::handle_option(
   }
 }
 
+#include <iostream>
```
hannes-steffenhagen-diffblue: 🚫 debug code changes

src/goto-harness/function_call_harness_generator.cpp
```diff
@@ -478,7 +480,7 @@ function_call_harness_generatort::implt::declare_arguments(
   for(const auto &parameter : parameters)
   {
     auto argument = allocate_objects.allocate_automatic_local_object(
-      parameter.type(), parameter.get_base_name());
+      remove_const(parameter.type()), parameter.get_base_name());
```
hannes-steffenhagen-diffblue, line 481: The `remove_const` changes should probably all be in one separate commit.

src/goto-harness/recursive_initialization.cpp
```diff
@@ -1002,7 +1021,7 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
     }
     ++function_pointer_index;
   }
-
+
```
hannes-steffenhagen-diffblue: stray whitespace change

src/util/type.h
```diff
@@ -224,4 +224,8 @@ inline type_with_subtypest &to_type_with_subtypes(typet &type)
     for(type_with_subtypest::subtypest::iterator it=to_type_with_subtypes(type).subtypes().begin(); \
         it!=to_type_with_subtypes(type).subtypes().end(); ++it)
 
+/// Remove const qualifier from type (if any).
+/// Returns type as is if there is no const qualifier.
+typet remove_const(typet type);
```
thk123, line 229: :bulb: I guess I'd probably expect this function to recurse... perhaps it should just be a static util function in this place where it is needed?

src/goto-harness/recursive_initialization.cpp
```diff
@@ -690,12 +690,17 @@ code_blockt recursive_initializationt::build_pointer_constructor(
     seen_assign_prev = code_assignt{*has_seen, has_seen_prev};
   }
 
-  const symbol_exprt &local_result =
-    get_fresh_local_typed_symexpr("local_result", type, nullptr_expr);
+  // we want to initialize the pointee as non-const even for pointer to const
```
thk123, line 712: :pick: Add why to the comment

src/goto-harness/recursive_initialization.cpp
```diff
@@ -797,9 +823,9 @@ code_blockt recursive_initializationt::build_dynamic_array_constructor(
   const optionalt<irep_idt> &lhs_name)
 {
   PRECONDITION(result.type().id() == ID_pointer);
-  const typet &pointer_type = result.type().subtype();
```
thk123, line 800: :pick: This renaming should have gone in a separate commt

src/goto-harness/recursive_initialization.cpp
```diff
@@ -1008,6 +1008,7 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
 
   const auto function_pointer_selector =
     get_fresh_local_symexpr("function_pointer_selector");
+  body.add(code_declt{function_pointer_selector});
```
thk123, line 1031: :question: Is there a test for this bug? That is, a test that passes if and only if this line is present

regression/goto-harness/havoc-global-int-02/test.desc
thk123, line 1: :no_entry_sign: As a rule - tests marked as knownbug should reference either a jira ticket or github issue that explains how they should be turned back on - otherwise we might as well delete them. Could you raise a Jira ticket to this effect and put it in the comment section of each desc file

src/goto-harness/goto_harness_parse_options.cpp
```diff
@@ -9,24 +9,58 @@ Author: Diffblue Ltd.
...
+    goto_model.get_symbol_table().begin(),
+    goto_model.get_symbol_table().end(),
+    std::inserter(symbols, symbols.end()),
+    [](const std::pair<const irep_idt, symbolt> &key_value_pair) {
+      return key_value_pair.first;
+    });
+  return symbols;
+}
+
+static void filter_goto_model(
```
thk123, line 50: An ideal candidate for a simple unit test

src/goto-harness/goto_harness_parse_options.cpp
```diff
@@ -50,6 +84,7 @@ int goto_harness_parse_optionst::doit()
 
   auto factory_options = collect_generate_factory_options();
 
+  register_language(new_ansi_c_language);
```
thk123: :no_entry_sign: Could a why comment explaining why this is needed be added here

src/goto-harness/goto_harness_parse_options.cpp
```diff
@@ -87,13 +124,15 @@ int goto_harness_parse_optionst::doit()
...
-    throw system_exceptiont{"failed to write goto program from file '" +
-                            got_harness_config.out_file + "'"};
-  }
+  filter_goto_model(goto_model, goto_model_without_harness_symbols);
+  auto harness_out = std::ofstream{got_harness_config.out_file};
+  dump_c(
+    goto_model.goto_functions,
+    true,
+    true,
+    false,
```
thk123, line 131: :pick: Pull these out into named local variables so there is some clue as to what they match up to. E.g. 

```C++
const bool use_system_headers = true;
...
dump_c(
  goto_model.goto_functions, 
  use_system_headers, 
...
```

src/goto-harness/recursive_initialization.cpp
```diff
@@ -472,35 +472,54 @@ std::string recursive_initialization_configt::to_string() const
...
+}
+
+irep_idt recursive_initializationt::get_fresh_global_name(
+  const std::string &symbol_name,
+  const exprt &initial_value) const
+{
+  auto &fresh_symbol = get_fresh_global_symbol(
+    goto_model.symbol_table,
+    symbol_name,
+    signed_int_type(), // FIXME why always signed_int_type???
```
thk123, line 509: :question: Noting this FIXME - it does seem odd?

src/goto-harness/recursive_initialization.cpp
```diff
@@ -472,35 +472,54 @@ std::string recursive_initialization_configt::to_string() const
...
-  const std::string &symbol_name,
-  const exprt &initial_value) const
+static symbolt &get_fresh_global_symbol(
+  symbol_tablet &symbol_table,
+  const std::string &symbol_base_name,
+  typet symbol_type,
+  irep_idt mode)
 {
+  source_locationt source_location{};
+  source_location.set_file(CPROVER_PREFIX "harness.c");
```
thk123: :pick: magic string "harness" should probably be turned into a macro or constexpr

src/goto-harness/recursive_initialization.h
```diff
@@ -22,6 +22,7 @@ Author: Diffblue Ltd.
 #include "function_harness_generator_options.h"
 #include "goto_harness_generator.h"
 
+#define GOTO_HARNESS_PREFIX "GOTO_HARNESS"
```
thk123: :grey_question: Why not 
```suggestion
#define GOTO_HARNESS_PREFIX CPROVER_PREFIX "_GOTO_HARNESS"
```

src/goto-harness/recursive_initialization.h
```diff
@@ -103,12 +103,14 @@ class recursive_initializationt
   /// \return the symbol expression for the `free` function
   symbol_exprt get_free_function();
 
-  bool is_initialization_allowed(const symbolt &symbol)
+  static bool is_initialization_allowed(const symbolt &symbol)
```
thk123, line 106: :pick: Great candidate for a unit test

thk123 (commented):
Yeah this looks mostly good to go to me. Just a few minor :no_entry_sign: comments, and a couple of :question: I'd like answered, but otherwise :+1: and of course figuring out the CI issue!

src/goto-harness/recursive_initialization.h
hannes-steffenhagen-diffblue: @thk123 anything starting with `CPROVER_PREFIX` is filtered out and ignored by `dump_c`, that's the only reason we added a goto harness specific prefix.

src/util/type.h
hannes-steffenhagen-diffblue: I was modelling this after [`std::remove_const`](https://en.cppreference.com/w/cpp/types/remove_cv) which doesn't recurse

src/goto-harness/recursive_initialization.cpp
hannes-steffenhagen-diffblue: The existing function pointer tests in goto harness wouldn't really work (anymore) without this.

src/goto-harness/recursive_initialization.cpp
hannes-steffenhagen-diffblue: @thk123 basically this is because of a somewhat odd design decision that was made at some point because someone was adding a bunch of global variables that were all integers apparently... I don't like it, but I considered it out of scope to actually fix this in this PR.

src/goto-harness/Makefile
```diff
@@ -16,6 +16,10 @@ OBJ += \
   ../linking/linking$(LIBEXT) \
   ../json/json$(LIBEXT) \
   ../json-symtab-language/json-symtab-language$(LIBEXT) \
+  ../ansi-c/ansi-c$(LIBEXT) \
+  ../cpp/cpp$(LIBEXT) \
+  ../goto-instrument/dump_c$(OBJEXT) \
+  ../goto-instrument/goto_program2code$(OBJEXT) \
```
hannes-steffenhagen-diffblue, line 22: these are all needed for `dump_c`

src/goto-harness/goto_harness_parse_options.cpp
hannes-steffenhagen-diffblue: @thk123 This snuck in from debugging earlier, it's not needed at all.

danpoe (approved):
LGTM

hannes-steffenhagen-diffblue (approved):

src/goto-harness/recursive_initialization.h
chrisr-diffblue: I would have thought that as this is an "implementation" prefix, it should be `__GOTO_HARNESS` ? (i.e. double underscore at the front) ?

src/util/type.cpp
```diff
@@ -28,3 +28,9 @@ void type_with_subtypest::move_to_subtypes(typet &type)
   sub.push_back(static_cast<const typet &>(get_nil_irep()));
   sub.back().swap(type);
 }
+
+typet remove_const(typet type)
```
chrisr-diffblue, line 32: const argument? Or if this is modifying the argument, then why return ?

src/util/type.h
chrisr-diffblue: Yeah, I'm not sure why this is a member function rather than a static?

chrisr-diffblue (commented):
Mostly looks OK, but a couple of queries I'd like answered if possible?

src/util/type.cpp
hannes-steffenhagen-diffblue: It's returning a copy with the modification. If the argument was const we'd have to make two copies, if it was const reference we'd still have to make a copy so we don't gain anything in either case.

src/util/type.h
hannes-steffenhagen-diffblue: @chrisr-diffblue it is free standing, not a member. Thomas was suggesting moving it to goto harness where it was used at the time, but this is the sort of thing I could see being useful elsewhere as well and I figured we don't want to have too much proliferation of duplicated utility functions.

chrisr-diffblue: Ah, sorry - I failed to read the diff context properly :-)

src/util/type.cpp
chrisr-diffblue: OK, I like the const just to make it explicit - and of course depending on how much this is called, an extra copy probably isn't the end of the world... But its a nitpick really on my part.

src/goto-harness/recursive_initialization.h
hannes-steffenhagen-diffblue: We could certainly add a `__` in front.

chrisr-diffblue (approved):
LGTM

src/goto-harness/recursive_initialization.cpp
```diff
@@ -774,7 +779,28 @@ code_blockt recursive_initializationt::build_struct_constructor(
...
+    // if the struct component is const we need to cast away the const
+    // for initialisation purposes.
+    // As far as I'm aware that's the closest thing to a 'correct' way
+    // to initialize dynamically allocated structs with const components
+    exprt component_initialisation_lhs = [&result, &component]() -> exprt {
+      auto member_expr = member_exprt{dereference_exprt{result}, component};
+      if(component.type().get_bool(ID_C_constant))
+      {
+        return dereference_exprt{
+          typecast_exprt{address_of_exprt{std::move(member_expr)},
```
smowton, line 815: Comment that this pile of stuff amounts to `*(X*)&field`, where `field` has type `const X*`

smowton: Also I take it you've tried using braced initialiser style for such things?

src/Makefile
```diff
@@ -8,8 +8,8 @@ DIRS = analyses \
        goto-cc \
        goto-checker \
        goto-diff \
-       goto-harness \
        goto-instrument \
+       goto-harness \
```
smowton, line 12: Sort order was correct before

smowton (approved):
Seems ok except the hundred or so tests that get marked KNOWNBUG here -- what was the memory-snapshot thingy that got trashed?

NlightNFotis:
@smowton We are tracking the broken tests internally and through github issue https://github.com/diffblue/cbmc/issues/5351. The memory snapshot harness was one of the modes that the `goto-harness` tool supported.

PR #5346, opened 2020-05-15
Treat uninitialized pointers like unknown pointers in address_check()

### LINKED ISSUE #5332: Pointer check on null pointer succeeds
```
void main()
{
  char *p;
  __CPROVER_assume(p == 0);
  *p;
}
```

CBMC version: 93b9e97980a01b9431917c8c129e10eae420e1d7
Operating system: Ubuntu
Exact command line resulting in the issue: cbmc --pointer-check test.c
What behaviour did you expect: verification failure
What happened instead: verification successful

This bug might be due to the pruning of pointer check constraints by the local bitvector analysis. In this case, only the check `!IS_INVALID_POINTER(p)` is retained, which is true for the nullpointer.

danpoe:
Fixes #5332

### DIFF
--- a/regression/cbmc/pointer-extra-checks/test.desc
+++ b/regression/cbmc/pointer-extra-checks/test.desc
@@ -8,7 +8,13 @@ main.c
 ^\[main.pointer_dereference.3\] .* dereference failure: pointer outside object bounds in \*q: SUCCESS$
 ^\[main.pointer_dereference.4\] .* dereference failure: deallocated dynamic object in \*r: SUCCESS$
 ^\[main.pointer_dereference.5\] .* dereference failure: pointer outside dynamic object bounds in \*r: SUCCESS$
-^\[main.pointer_dereference.6\] .* dereference failure: pointer uninitialized in \*s: FAILURE$
+^\[main.pointer_dereference.6\] .* dereference failure: pointer invalid in \*s: FAILURE$
+^\[main.pointer_dereference.7\] .* dereference failure: pointer NULL in \*s: FAILURE$
+^\[main.pointer_dereference.8\] .* dereference failure: deallocated dynamic object in \*s: FAILURE$
+^\[main.pointer_dereference.9\] .* dereference failure: dead object in \*s: FAILURE$
+^\[main.pointer_dereference.10\] .* dereference failure: pointer outside dynamic object bounds in \*s: FAILURE$
+^\[main.pointer_dereference.11\] .* dereference failure: pointer outside object bounds in \*s: FAILURE$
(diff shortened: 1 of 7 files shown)


regression/cbmc/pointer-extra-checks/test.desc
```diff
@@ -8,7 +8,15 @@ main.c
 ^\[main.pointer_dereference.3\] .* dereference failure: pointer outside object bounds in \*q: SUCCESS$
 ^\[main.pointer_dereference.4\] .* dereference failure: deallocated dynamic object in \*r: SUCCESS$
 ^\[main.pointer_dereference.5\] .* dereference failure: pointer outside dynamic object bounds in \*r: SUCCESS$
-^\[main.pointer_dereference.6\] .* dereference failure: pointer uninitialized in \*s: FAILURE$
+^\[main.pointer_dereference.6\] .* dereference failure: pointer invalid in \*s: FAILURE
+^\[main.pointer_dereference.7\] .* dereference failure: pointer NULL in \*s: FAILURE
```
smowton: These are quite weak checks because they don't pin down what caused the failure of each condition -- consider a separate test case per expected failure

src/analyses/goto_check.cpp
```diff
@@ -1282,30 +1282,24 @@ goto_checkt::address_check(const exprt &address, const exprt &size)
...
+    if(unknown || flags.is_null())
     {
-      conditions.push_back(
-        conditiont{or_exprt{in_bounds_of_some_explicit_allocation,
-                            not_exprt{is_invalid_pointer_exprt{address}}},
-                   "pointer uninitialized"});
+      conditions.push_back(conditiont(
+        or_exprt(
+          in_bounds_of_some_explicit_allocation,
+          not_exprt(null_pointer(address))),
```
smowton, line 1298: Not changed in this PR, but is the condition `in_bounds_of_some_explicit_allocation` redundant (or are we explicitly considering an address space where the zero pointer might point at allocated memory)?

smowton (approved):
Actually that comment about the test might be nonsense, happy to approve now as long as it is

regression/cbmc/pointer-extra-checks/test.desc
danpoe: As you probably already noticed given your comment below, these lines are there to verify that all the necessary checks are added by `address_check()` (and all of them will fail for a plain uninitialized pointer).

src/analyses/goto_check.cpp
danpoe: Good point. Yes, I think that should be redundant.

PR #5350, opened 2020-05-20
Complete Java lambda support [TG-11760]

smowton:
This adds the various missing pieces of Java lambda support:

* Allow direct method references (javac uses these when no adaptation is necessary, there are no captured parameters to pass on, etc)
* Support lambdas that use invokevirtual / invokeinterface (these are generated similar to above but for instance and interface methods, e.g. `this::f` or `someInterface::method`)
* Support the `newinvokespecial` lambda binding, used for `SomeType::new`, which means we must both instantiate and call the constructor
* Support functional interfaces whose defined method is inherited
* Support functional interfaces that also define `equals` or `hashCode`
* Support functional interfaces featuring default methods


jbmc/src/java_bytecode/java_bytecode_convert_class.cpp
```diff
@@ -426,7 +426,8 @@ void java_bytecode_convert_classt::convert(
     lambda_entry.second.is_unknown_handle()
       ? class_type.add_unknown_lambda_method_handle()
       : class_type.add_lambda_method_handle(
-          "java::" + id2string(lambda_entry.second.lambda_method_ref));
+          "java::" + id2string(lambda_entry.second.lambda_method_ref),
```
peterschrammel: ⛏️ `ID_java + "::"` ?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -92,6 +92,83 @@ lambda_method_handle(
...
+
+static optionalt<irep_idt> get_unique_abstract_method(
+  const java_class_typet::methodst &methods,
+  const namespacet &ns)
+{
+  optionalt<irep_idt> result;
+  for(const auto &method : methods)
+  {
+    const auto &name = method.get_name();
+    static irep_idt equals = "equals";
```
peterschrammel: `const` ?

peterschrammel (approved):

jbmc/src/java_bytecode/java_bytecode_convert_class.cpp
thk123: not sure - "java::" is quite a common string - probably makes more sense to pull out a ID_java_prefix than hide a few of them with this

thk123: Also unchanged by this PR

jbmc/src/java_bytecode/java_bytecode_parser.cpp
```diff
@@ -330,6 +329,11 @@ class method_handle_infot : public structured_pool_entryt
     reference_index = entry.ref2;
   }
 
+  method_handle_kindt get_reference_kind() const
```
thk123: :grey_question: `get_method_handle_kind`? Not sure why the change in nomenclature?

jbmc/src/java_bytecode/java_types.h
```diff
@@ -450,34 +450,62 @@ class java_class_typet:public class_typet
...
-  //   {
-  //   }
-  //
-  //   const irep_idt &get_lambda_method_handle() const
-  //   {
-  //     return id();
-  //   }
-  // };
-  using java_lambda_method_handlest = irept::subt;
+  enum class method_handle_typet
```
thk123: :bulb: `method_handle_kindt`?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -92,6 +92,83 @@ lambda_method_handle(
...
+    if(method.get_base_name() == equals || method.get_base_name() == hashCode)
+    {
+      // equals and hashCode methods can't be the implemented method of a
+      // functional interface, even if the interface re-declares them as
+      // abstract.
+      continue;
+    }
+    if(!ns.lookup(name).type.get_bool(ID_C_abstract))
+      continue;
+    if(result.has_value() && name != *result)
```
thk123: Is it allowed for a functional interface to have two unimplemented methods with the same name?

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -92,6 +92,83 @@ lambda_method_handle(
...
+    result = name;
+  }
+  return result;
+}
+
+static optionalt<irep_idt> get_unique_unimplemented_method(
+  const irep_idt &interface_id,
+  const namespacet &ns)
+{
+  static irep_idt jlo = "java::java.lang.Object";
```
thk123: const here too - leaving mutable global state lying around is a recipe for bugs

jbmc/src/java_bytecode/java_bytecode_parser.cpp
```diff
@@ -1911,22 +1911,12 @@ java_bytecode_parsert::parse_method_handle(const method_handle_infot &entry)
 
   lambda_method_handlet lambda_method_handle;
 
-  if(has_prefix(name_and_type.get_name(pool_entry_lambda), "lambda$"))
```
thk123, line 1910: Could commit message include an example of when this happens:

> but sometimes it can use a direct reference to a method, which we would previously have failed to spot.

jbmc/src/java_bytecode/java_bytecode_parser.cpp
```diff
@@ -1905,6 +1905,20 @@ java_bytecode_parsert::parse_method_handle(const method_handle_infot &entry)
...
+    entry.get_reference_kind() ==
+      method_handle_infot::method_handle_kindt::REF_getField ||
+    entry.get_reference_kind() ==
+      method_handle_infot::method_handle_kindt::REF_putField ||
+    entry.get_reference_kind() ==
+      method_handle_infot::method_handle_kindt::REF_getStatic ||
+    entry.get_reference_kind() ==
+      method_handle_infot::method_handle_kindt::REF_putStatic)
+  {
+    // Field accessors not implemented yet
```
thk123: :question: I don't understand is this if you assign a lambda from a field containing a lambda?

jbmc/regression/jbmc/lambda-interface-declaring-equals/InterfaceDeclaringEquals.java
```diff
@@ -0,0 +1,6 @@
+interface InterfaceDeclaringEquals {
+
+  public int f(int x);
+
+  public boolean equals(Object other);
```
thk123, line 5: Could we have a test for a functional interface that also declares hashcode

jbmc/regression/jbmc/lambda-reference-kinds/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+Test
+--function Test.main
+^VERIFICATION SUCCESSFUL$
+line 19 assertion at file Test\.java line 19 function java::Test\.main:\(\)V bytecode-index 28: SUCCESS
+line 20 assertion at file Test\.java line 20 function java::Test\.main:\(\)V bytecode-index 39: SUCCESS
+line 21 assertion at file Test\.java line 21 function java::Test\.main:\(\)V bytecode-index 50: SUCCESS
+^EXIT=0$
+^SIGNAL=0$
```
thk123, line 10: :no_entry_sign: Please add an explanation as to the purpose of this test (I believe it is checking when the method reference is directly inlined?

jbmc/src/java_bytecode/java_bytecode_parser.cpp
```diff
@@ -1881,6 +1881,12 @@ void java_bytecode_parsert::parse_local_variable_type_table(methodt &method)
   }
 }
 
+/// Translate the lambda method reference kind in a class file into the kind
+/// of handling the method requires. invokestatic/special translate into direct
+/// method calls; invokevirtual/interface translate into virtual dispatch,
+/// newinvokespecial translates into a special instantiate-and-construct
+/// sequence. The field-manipulation reference kinds appear never to happen in
+/// reality and don't have syntax in the Java language.
```
thk123, line 1889: I think this info would be good to put nearer the block that handles this

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -102,6 +102,10 @@ class no_unique_unimplemented_method_exceptiont : public std::exception
   const std::string message;
 };
 
+/// Find the unique abstract method that isn't equals(...) or hashCode() in a
+/// kist of methods. Returns the method's symbol-table identifier if one is
```
thk123:
```suggestion
/// list of methods. Returns the method's symbol-table identifier if one is
```

thk123 (commented):
In a couple of places you allow the name of the method to match the name of the method found in the functional interface, but afaict you don't have any tests for this scenario, can they be added. 

Couple of minor points but looks good

jbmc/src/java_bytecode/lambda_synthesis.cpp
smowton: These are mangled names (since we expect to find them in `ns`); renamed to clarify.

jbmc/src/java_bytecode/java_bytecode_parser.cpp
smowton: Expanded comment; in brief these are notional "field bindings", e.g. you would be able to say `Producer<Integer> = this::someIntegerField` and get a REF_getField instead of a REF_invokeSpecial pointing at a short function that fetches the field. While it made it into the JVMS and might be used in future, AFAICT nobody has actually done that.

jbmc/src/java_bytecode/java_bytecode_parser.cpp
smowton: It's now mentioned in both places

jbmc/src/java_bytecode/lambda_synthesis.cpp
thk123: Sure - but I don't understand why an interface would be allowed to have two methods with the same mangled name? I mean - surely that is just impossible? I.e. in what circumstances is `result.has_value() && name == *result`?

jbmc/src/java_bytecode/java_bytecode_parser.cpp
thk123: lol cool feature I guess :sunglasses:

jbmc/src/java_bytecode/java_bytecode_parser.cpp
```diff
@@ -1896,33 +1926,45 @@ java_bytecode_parsert::parse_method_handle(const method_handle_infot &entry)
...
-    // where POSTFIX is FUN for a function name in which the lambda is define
-    //                "static" when it is a static member of the class
-    //                "new" when it is a class variable, instantiated in <init>
-    lambda_method_handle.lambda_method_name =
-      name_and_type.get_name(pool_entry_lambda);
-    lambda_method_handle.lambda_method_ref = method_ref;
-    lambda_method_handle.handle_type =
-      method_handle_typet::LAMBDA_METHOD_HANDLE;
-
-    return lambda_method_handle;
```
thk123, line 1922: :bulb: Given that this could happen - it might make sense to output a warning, or even an invariant

thk123 (approved):
Still a little confused about the check for the mangled name - happy to chat in slack if easier to explain

thk123 (dismissed):
Oops approved too early - could you have a look at the gaps in testing highlighted by code cov, in particular:

 - :no_entry_sign: a test for with an interface that contains two methods
 - :no_entry_sign: a test for an interface with one method, and inherits another
 - :question: a test for a constructor? - I thought I'd seen this, but there is a whole untested function: `instantiate_new_object`
 - :bulb: and ideally something to cover this line: https://codecov.io/gh/diffblue/cbmc/compare/c6082355bde3657bdfdefde8eff16a2813684b54...4b58a793061915277cf8255c628f94bff27414af/diff#D32-1244

jbmc/src/java_bytecode/java_bytecode_language.cpp
```diff
@@ -1241,7 +1241,16 @@ static void notify_static_method_calls(
...
-          needed_lazy_methods->add_needed_method(fn_sym->get_identifier());
+        {
+          INVARIANT(
+            false,
+            "Java synthetic methods are not "
+            "expected to produce side_effect_expr_function_callt. If "
+            "that has changed, remove this invariant. Also note that "
+            "as of the time of writing remove_virtual_functions did "
+            "not support this form of function call.");
+          // needed_lazy_methods->add_needed_method(fn_sym->get_identifier());
```
romainbrenguier, line 1252: :pick: remove or comment with `#if 0`

romainbrenguier (approved):

PR #5353, opened 2020-05-21
Allow file-local symbol mangling for multi-file compilation

### LINKED ISSUE #4689: export-file-local-symbols doesn't work when compiling+linking
The `--export-file-local-symbols` flag appears to have no effect when doing a combined compile+link (i.e., when not using the `-c` flag to goto-cc).

azcwagner:
Addressing issue https://github.com/diffblue/cbmc/issues/4689, this makes the goto-cc utility mangle exported function symbols when multiple files are compiled and linked in a single run of goto-cc, as it does when files are compiled separately.

A compile-and-link argument has also been added to chain.sh under regression/goto-cc-file-local to test this compilation scenario.

### DIFF
--- a/regression/goto-cc-file-local/chain.sh
+++ b/regression/goto-cc-file-local/chain.sh
@@ -54,39 +54,61 @@ else
   export_flag="--export-file-local-symbols"
 fi
 
-cnt=0
-for src in ${SRC}; do
-  cnt=$((cnt + 1))
-  out_suffix=""
-  if is_in out-file-counter "$ALL_ARGS"; then
-    out_suffix="_$cnt"
-    if is_in suffix "$ALL_ARGS"; then
-      suffix="--mangle-suffix custom_$cnt"
+OUT_FILE="main.gb"
+if is_in compile-and-link "$ALL_ARGS"; then
+    if [[ "${is_windows}" == "true" ]]; then
+      "${goto_cc}"                        \
+          ${export_flag}                  \
+          --verbosity 10                  \
+          ${wall}                         \
+          ${suffix}                       \
+          ${SRC}                          \
+          /Fe"${OUT_FILE}"
+  
+    else
+      "${goto_cc}"                        \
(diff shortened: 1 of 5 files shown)


regression/goto-cc-file-local/chain.sh
```diff
@@ -54,39 +54,61 @@ else
...
+OUT_FILE="main.gb"
+if is_in compile-and-link "$ALL_ARGS"; then
+    if [[ "${is_windows}" == "true" ]]; then
+      "${goto_cc}"                        \
+          ${export_flag}                  \
+          --verbosity 10                  \
+          ${wall}                         \
+          ${suffix}                       \
+          ${SRC}                          \
+          /Fo"${OUT_FILE}"
```
karkhaz: I believe this should be `/Fe` rather than `/Fo`, since we're compiling and linking at the same time

karkhaz (changes requested):
Thanks for this! I'm not sure why the CodeBuild job is failing, but please address my comment and force-push. If that still doesn't make CodeBuild happy, we'll ask diffblue for the CodeBuild log.

karkhaz:
Also, when you force-push, could you please use the phrase "fixes #4689" in the commit message? That will autoclose that issue as soon as this PR gets merged.

karkhaz:
@tautschnig @chrisr-diffblue do you have any insight about why the Windows CodeBuild job is hanging? This was pushed yesterday, all the other jobs have finished

karkhaz (approved):

chrisr-diffblue:
@karkhaz Sadly, no - it looks like the windows job never even got started. Probably the best thing to do would be a rebase+force push to trigger a new CI run.

karkhaz:
Looks like this is ready for review, all checks have passed!

chrisr-diffblue (approved):
Thanks for this (and apologies for the long delayed review). LGTM!

PR #5356, opened 2020-05-22
Add stub implementations for value set classes

hannes-steffenhagen-diffblue:
This is just add some stub implementation for the value set classes we'll be moving the existing value set functionality to (and fix it along the way).

### DIFF
--- a/src/analyses/variable-sensitivity/value_set_abstract_value.h
+++ b/src/analyses/variable-sensitivity/value_set_abstract_value.h
@@ -1,0 +1,27 @@
+/*******************************************************************\
+
+ Module: analyses variable-sensitivity variable-sensitivity-value-set
+
+ Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+/// \file
+/// Value sets for primitives
+
+// NOLINTNEXTLINE(whitespace/line_length)
+#ifndef CPROVER_ANALYSES_VARIABLE_SENSITIVITY_VALUE_SET_ABSTRACT_VALUE_H
+// NOLINTNEXTLINE(whitespace/line_length)
+#define CPROVER_ANALYSES_VARIABLE_SENSITIVITY_VALUE_SET_ABSTRACT_VALUE_H
+
+#include "abstract_value.h"
+
+class value_set_abstract_valuet : public abstract_valuet
+{
+public:
+  explicit value_set_abstract_valuet(const typet &type);
+  CLONE
+};
+
+// NOLINTNEXTLINE(whitespace/line_length)
+#endif // CPROVER_ANALYSES_VARIABLE_SENSITIVITY_VALUE_SET_ABSTRACT_VALUE_H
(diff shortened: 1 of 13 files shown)


unit/analyses/variable-sensitivity/value_set/pointer_abstract_object.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+TEST_CASE(
+  "A value set pointer abstract object created from type is top",
+  VALUE_SET_TEST_TAGS)
+{
+  value_set_pointer_abstract_objectt abstract_object(
+    pointer_typet(signedbv_typet{32}, 32));
+  REQUIRE(abstract_object.is_top());
+  REQUIRE(!abstract_object.is_bottom());
+}
```
smowton: I suspect CI will give you a hard time for missing end-of-line here and elsewhere

smowton (approved):
Harmless

src/analyses/variable-sensitivity/value_set_abstract_value.h
```diff
@@ -0,0 +1,27 @@
...
+
+ Author: Diffblue Ltd.
+
+\*******************************************************************/
+
+/// \file
+/// Value sets for primitives
+
+// NOLINTNEXTLINE(whitespace/line_length)
+#ifndef CPROVER_ANALYSES_VARIABLE_SENSITIVITY_VALUE_SET_ABSTRACT_VALUE_H
```
danpoe, line 13: `NOLINTNEXTLINE` seems unnecessary (here and elsewhere)

src/config.inc
```diff
@@ -5,7 +5,7 @@ BUILD_ENV = AUTO
 ifeq ($(BUILD_ENV),MSVC)
   #CXXFLAGS += /Wall /WX
 else
-  CXXFLAGS += -Wall -pedantic -Werror -Wno-deprecated-declarations -Wswitch-enum
+  CXXFLAGS += -Wall -pedantic -Wno-deprecated-declarations -Wswitch-enum
```
danpoe: I take it that this is because a new compiler version warns about things the older ones did not? Can we use `-Wno-error=warning` to only disable these new warnings being treated as errors? Also should be in a separate commit.

danpoe (commented):
Code looks good, though the change disabling `-Werror` should be changed

src/config.inc
hannes-steffenhagen-diffblue: This is an accident actually, I need this change only locally because I’m using some newer compiler versions than what we’re running on CI.

src/analyses/variable-sensitivity/value_set_abstract_value.h
hannes-steffenhagen-diffblue: cpplint complains about the line length without these.

danpoe (approved):

NlightNFotis (approved):

PR #5359, opened 2020-05-27
Java frontend: add support for lambdas that require generating boxing, unboxing, widening casts and upcasts

smowton:
The Java lambda metafactory, per https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/LambdaMetafactory.html, can require adapting the type of the referenced method to that of the target functional interface by introducing boxing, unboxing, primitive type widening and pointer upcasting conversions. This adds tests for a wide variety of such cases and implements the required suppport.


jbmc/src/java_bytecode/java_utils.cpp
```diff
@@ -38,6 +38,72 @@ bool is_java_string_type(const struct_typet &struct_type)
...
+      {"java::java.lang.Short",
+       {"java::java.lang.Short.shortValue:()S", java_short_type()}},
+    };
+
+  auto found = type_info_by_name.find(type_name);
+  return found == type_info_by_name.end() ? nullptr : &found->second;
+}
+
+const java_primitive_type_infot *
+get_java_primitive_type_info(const typet &maybe_primitive_type)
```
thk123, line 69: :pick: `optionalt` rather than a pointer

jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -610,6 +610,82 @@ static symbol_exprt instantiate_new_object(
...
+  const bool required_is_pointer = can_cast_type<pointer_typet>(required_type);
+
+  if(original_is_pointer == required_is_pointer)
+  {
+    if(original_is_pointer && original_type != required_type)
+      return typecast_exprt{expr, required_type};
+    else
+      return expr;
+  }
+
```
thk123: What's this case for?

thk123 (approved):
:pick: Commit messages frequently exceed 72 (80?) chars causing github to wrap them, otherwise looks perfect to me

jbmc/src/java_bytecode/lambda_synthesis.cpp
thk123: I think the answer is byte to int and Derived -> Base?

smowton: Gains a comment in a later commit (it's for pointer or integer type clashes without boxing, where plain casting always does the trick (forcing pointers and widening integers, so long as only valid Java casts are fed in in the first place)

PR #5361, opened 2020-05-27
Intervals with tests

thk123:
A note to reviewers, this PR, though large, is entirely an isolated addition `util/interval`. There are lots of ways this code could be improved (I've listed some in #5203) since VSD is based on top of this, I propose merging as is, and then applying fixes after - so I'd encourage not going through this with a fine tooth comb to find the ways it is wrong! 

This is a rebase of #5203 that addresses I believe the most pressing issues:

 - ensures all tests are run on Make and CMake builds
 - fixes the errors in operator== that stops the tests actually checking things properly
 - fixes error in to string
 - adds a lot of tests for the missing coverage (still some gaps)
 - highlight some areas where the behavior is wrong
 - fixes bug in `eval(ID_minus)`
Some gaps in the tests remain:

**`logical_or` / `logical_and` / `logical_xor`**
_I believe these functions should be removed, and there just be a single method to get the range as a tvt, and then have these functions implemented on tvt. As such, I didn't write the tests._

**code in the `generate_*` and `append_*` functions**
_The semantics of these aren't clear to me, and shouldn't be part of the public interface. Probably the code that is uncovered will be hard to cover if the interface is tidier_

**trivial type getters and constructors**
There are loads of these, and they mostly need to be deleted. To avoid breaking, I recommend waiting until VSD is merged before deleting. However, not worth testing in the meantime.

**wrappers**
interval.cpp implements all of the calling conventions. This wants tidying up, but in the meantime I've not bothered to test every possible calling syntax. 

**typecast**
This is worth testing, but I don't understand what it's doing and it seems fishy. However, it _is_ used in VSD so I'll leave for now, so it can be tested / removed after VSD is merged. 

TBH in the brief time I've spent with this, there are a bunch of bugs and inconsistencies here, if there wasn't a mountain of work built on top of it, I'd recommend breaking this into something smaller than works (e.g. remove the untested or broken functionality). However, this isn't really an option, so I suggest we merge like this, merge VSD and perhaps immediately remove the interval portion of the VSD, along with this, and reintroduce it gradually (allowing at least the bulk of VSD to be merged). 

### DIFF
--- a/src/util/interval.cpp
+++ b/src/util/interval.cpp
@@ -43,7 +43,7 @@ constant_interval_exprt constant_interval_exprt::unary_minus() const
 {
   if(is_single_value_interval())
   {
-    handle_constants(unary_minus_exprt());
+    handle_constant_unary_expression(ID_unary_minus);
   }
 
   exprt lower;
@@ -75,7 +75,7 @@ constant_interval_exprt::plus(const constant_interval_exprt &o) const
 {
   if(o.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, plus_exprt());
+    handle_constant_binary_expression(o, ID_plus);
   }
 
   exprt lower = min();
@@ -121,7 +121,7 @@ constant_interval_exprt::multiply(const constant_interval_exprt &o) const
 {
   if(o.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, mult_exprt());
+    handle_constant_binary_expression(o, ID_mult);
   }
 
   return get_extremes(*this, o, ID_mult);
@@ -132,7 +132,7 @@ constant_interval_exprt::divide(const constant_interval_exprt &o) const
 {
   if(o.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, div_exprt());
+    handle_constant_binary_expression(o, ID_div);
   }
 
   // If other might be division by zero, set everything to top.
@@ -151,7 +151,7 @@ constant_interval_exprt::modulo(const constant_interval_exprt &o) const
 
   if(o.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, mod_exprt());
+    handle_constant_binary_expression(o, ID_mod);
   }
 
   if(o.is_bottom())
@@ -270,7 +270,7 @@ constant_interval_exprt::left_shift(const constant_interval_exprt &o) const
 {
   if(o.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, shl_exprt());
+    handle_constant_binary_expression(o, ID_shl);
   }
 
   if(is_negative(o.get_lower()))
@@ -287,7 +287,7 @@ constant_interval_exprt::right_shift(const constant_interval_exprt &o) const
 {
   if(o.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, ashr_exprt());
+    handle_constant_binary_expression(o, ID_ashr);
   }
 
   if(is_negative(o.get_lower()))
@@ -303,7 +303,7 @@ constant_interval_exprt::bitwise_xor(const constant_interval_exprt &o) const
 {
   if(o.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, bitxor_exprt());
+    handle_constant_binary_expression(o, ID_bitxor);
   }
 
   return top();
@@ -314,7 +314,7 @@ constant_interval_exprt::bitwise_or(const constant_interval_exprt &o) const
 {
   if(o.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, bitor_exprt());
+    handle_constant_binary_expression(o, ID_bitor);
   }
 
   return top();
@@ -325,7 +325,7 @@ constant_interval_exprt::bitwise_and(const constant_interval_exprt &o) const
 {
   if(o.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, bitand_exprt());
+    handle_constant_binary_expression(o, ID_bitand);
   }
 
   return top();
@@ -335,7 +335,7 @@ constant_interval_exprt constant_interval_exprt::bitwise_not() const
 {
   if(is_single_value_interval())
   {
-    handle_constants(bitnot_exprt());
+    handle_constant_unary_expression(ID_bitnot);
   }
 
   return top();
@@ -961,31 +961,25 @@ exprt constant_interval_exprt::generate_shift_expression(
 }
 
 constant_interval_exprt
-constant_interval_exprt::handle_constants(exprt expr) const
+constant_interval_exprt::handle_constant_unary_expression(
+  const irep_idt &op) const
 {
   if(is_single_value_interval())
   {
-    expr.type() = type();
-    expr.copy_to_operands(get_lower());
-
+    auto expr = unary_exprt(op, get_lower());
     return constant_interval_exprt(simplified_expr(expr));
   }
-
   return top();
 }
 
-constant_interval_exprt constant_interval_exprt::handle_constants(
-  const constant_interval_exprt &o,
-  exprt expr) const
+constant_interval_exprt
+constant_interval_exprt::handle_constant_binary_expression(
+  const constant_interval_exprt &other,
+  const irep_idt &op) const
 {
-  if(is_single_value_interval() && o.is_single_value_interval())
-  {
-    expr.type() = type();
-    expr.copy_to_operands(get_lower(), o.get_lower());
-    return constant_interval_exprt(simplified_expr(expr));
-  }
-
-  return top();
+  PRECONDITION(is_single_value_interval() && other.is_single_value_interval());
+  auto expr = binary_exprt(get_lower(), op, other.get_lower());
+  return constant_interval_exprt(simplified_expr(expr));
 }
 
 exprt constant_interval_exprt::get_max(const exprt &a, const exprt &b)
--- a/src/util/interval.cpp
+++ b/src/util/interval.cpp
@@ -105,15 +105,16 @@ constant_interval_exprt::plus(const constant_interval_exprt &o) const
 }
 
 constant_interval_exprt
-constant_interval_exprt::minus(const constant_interval_exprt &o) const
+constant_interval_exprt::minus(const constant_interval_exprt &other) const
 {
-  if(o.is_single_value_interval() && is_single_value_interval())
+  if(other.is_single_value_interval() && is_single_value_interval())
   {
-    handle_constants(o, minus_exprt());
+    handle_constant_binary_expression(other, ID_minus);
   }
 
-  // e.g. [t.u - o.l, t.l - o.u]
-  return plus(o.unary_minus().swap());
+  // FIXME This is nonsense if the interval type is unsigned int
+  // [this.lower - other.upper, this.upper - other.lower]
+  return plus(other.unary_minus());
 }
 
 constant_interval_exprt
--- a/src/util/interval.cpp
+++ b/src/util/interval.cpp
@@ -1,9 +1,10 @@
-/*
- * interval.cpp
- *
- *  Created on: 16 Jun 2017
- *      Author: dan
- */
+/*******************************************************************\
+
+ Module: intervals
+
+ Author: Daniel Neville (2017), Diffblue Ltd
+
+\*******************************************************************/
 
 /*
  *
@@ -32,8 +33,6 @@ const exprt &constant_interval_exprt::get_upper() const
   return op1();
 }
 
-//make clean -s && make -j 7 CXX="/usr/local/bin/ccache g++" -s && ./unit_tests
-
 constant_interval_exprt constant_interval_exprt::unary_plus() const
 {
   return *this;
@@ -49,7 +48,7 @@ constant_interval_exprt constant_interval_exprt::unary_minus() const
   exprt lower;
   exprt upper;
 
-  if(is_max())
+  if(has_no_upper_bound())
   {
     lower = min();
   }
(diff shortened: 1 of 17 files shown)


thk123:
I've added test for binary `eval` (And found a [bug](https://github.com/diffblue/cbmc/pull/5361/commits/4952ac4c94b04537dc2c9ed4dac831a8b672487f) in the process :expressionless:), will add one for to_string since it also originally contained a bug and then this PR I believe is ready

thk123:
Added to_string tests and squashed clang format changes into the original commit that introduces them

thk123:
Removed the changes to interval_template, one into separate PR: #5362, the other changes essentially cancelled themselves out, so I just removed them both.

thk123:
CI failure is lint - will fix, but ready for review

unit/util/interval/to_string.cpp
```diff
@@ -0,0 +1,52 @@
+/*******************************************************************\
+ Module: Unit tests for variable/sensitivity/abstract_object::merge
```
smowton: Wrong header

unit/util/interval/eval.cpp
```diff
@@ -103,3 +103,62 @@ SCENARIO("Unary eval on intervals", "[core][analyses][interval][eval]")
...
+  constant_interval_exprt five = interval_of(5);
+  constant_interval_exprt eight = interval_of(8);
+
+  SECTION("Numeric operations")
+  {
+    REQUIRE(five.eval(ID_plus, eight) == interval_of(13));
+    REQUIRE(five.eval(ID_minus, eight) == interval_of(-3));
+    REQUIRE(five.eval(ID_mult, eight) == interval_of(40));
+    REQUIRE(five.eval(ID_div, eight) == interval_of(0));
+    REQUIRE(five.eval(ID_mod, eight) == interval_from_to(0, 5));
```
smowton, line 135: Checking this oddity is known

unit/util/interval/eval.cpp
```diff
@@ -103,3 +103,62 @@ SCENARIO("Unary eval on intervals", "[core][analyses][interval][eval]")
     }
   }
 }
+
+TEST_CASE("binary eval switch", "[core][analyses][interval]")
```
smowton, line 104: Comment that this only considers singletons?

unit/util/interval/comparisons.cpp
```diff
@@ -237,3 +237,32 @@ SCENARIO("comparison interval domain", "[core][analyses][interval][comparison]")
...
+    constant_interval_exprt two_to_four(CEV(2), CEV(4));
+    constant_interval_exprt six_to_eight(CEV(6), CEV(8));
+    constant_interval_exprt five_to_ten(CEV(5), CEV(10));
+
+    // TODO: wrongly concludes that these are unequal
+    // ADA-537
+    // REQUIRE(two_to_four.equal(six_to_eight).is_unknown());
+    REQUIRE_FALSE(two_to_four.equal(five_to_ten).is_true());
+
+    REQUIRE_FALSE(two_to_four.not_equal(six_to_eight).is_unknown());
```
smowton: Why does the not-unknown one use six_to_eight but the others use five_to_ten?

unit/util/interval/modulo.cpp
```diff
@@ -17,6 +17,25 @@
...
+  GIVEN("Two single element internvals")
+  {
+    constant_interval_exprt a(CEV(5));
+    constant_interval_exprt b(CEV(10));
+    constant_interval_exprt zero(CEV(0));
+    const auto a_mod_b = a.modulo(b);
+    REQUIRE(V(a_mod_b.get_upper()) == 5);
+
+    const auto b_mod_a = b.modulo(a);
+    // TODO: precision with single element modulo
```
smowton, line 29: Perhaps add commented-out the results you would expect when fixed. Got a ticket for this?

unit/util/interval/modulo.cpp
```diff
@@ -52,6 +71,19 @@ SCENARIO("modulo interval domain", "[core][analyses][interval][modulo]")
             constant_interval_exprt(CEV(-10), CEV(20)),
             constant_interval_exprt(CEV(0), CEV(5))) ==
           constant_interval_exprt::top(signedbv_typet(32)));
+
+        REQUIRE(
+          constant_interval_exprt::modulo(
+            constant_interval_exprt(CEV(-10), CEV(20)),
```
smowton, line 78: Maybe comment on why the behaviour for this two differs so much?

src/util/interval.cpp
```diff
@@ -433,7 +433,7 @@ tvt constant_interval_exprt::equal(const constant_interval_exprt &o) const
     return tvt(equal(get_lower(), o.get_lower()));
   }
 
-  if(equal(get_upper(), o.get_upper()) && equal(get_upper(), o.get_upper()))
+  if(equal(get_upper(), o.get_upper()) && equal(get_lower(), o.get_lower()))
```
smowton, line 436: lol

unit/util/interval/multiply.cpp
```diff
@@ -22,6 +22,14 @@ SCENARIO("multiply interval domain", "[core][analyses][interval][multiply]")
     symbol_tablet symbol_table;
     namespacet ns(symbol_table);
 
+    WHEN("Single element multiplication")
+    {
+      constant_interval_exprt a(CEV(5));
+      constant_interval_exprt b(CEV(10));
+      const auto a_times_b = a.multiply(b);
+      REQUIRE(V(a_times_b.get_upper()) == 50);
```
smowton: and the lower? Even if it's crap might as well assert it for documentation's sake

unit/util/interval/eval.cpp
```diff
@@ -36,14 +36,20 @@ SCENARIO("Unary eval on intervals", "[core][analyses][interval][eval]")
...
       constant_interval_exprt(from_integer(5, signedbv_typet(32)));
+    const constant_interval_exprt &max_interval =
+      constant_interval_exprt(max_exprt{signedbv_typet(32)});
+    const constant_interval_exprt &min_interval =
+      constant_interval_exprt(min_exprt{signedbv_typet(32)});
 
     THEN("When we apply unary addition to it, nothing should happen")
     {
       REQUIRE(five.eval(ID_unary_plus) == five);
+      REQUIRE(min_interval.eval(ID_unary_minus).has_no_lower_bound());
```
smowton: typo unary_minus vs. plus

unit/util/interval/eval.cpp
```diff
@@ -52,6 +58,9 @@ SCENARIO("Unary eval on intervals", "[core][analyses][interval][eval]")
         numeric_cast<mp_integer>(five.eval(ID_unary_minus).get_lower());
       REQUIRE(negated_val.has_value());
       REQUIRE(negated_val.value() == -5);
+
+      REQUIRE(max_interval.eval(ID_unary_minus).has_no_lower_bound());
```
smowton: Maybe compare them for equality with the expected interval as is done elsewhere rather than check one property?

unit/util/interval/eval.cpp
```diff
@@ -62,4 +71,40 @@ SCENARIO("Unary eval on intervals", "[core][analyses][interval][eval]")
...
+  {
+    constant_interval_exprt five_to_eight = constant_interval_exprt(
+      from_integer(5, signedbv_typet(32)), from_integer(8, signedbv_typet(32)));
+
+    THEN("When we apply unary addition to it, nothing should happen")
+    {
+      REQUIRE(five_to_eight.eval(ID_unary_plus) == five_to_eight);
+    }
+
+    THEN("When we apply unary subtraction to it, it should be negated")
```
smowton: unary subtraction -> unary minus or negation

unit/util/interval/eval.cpp
```diff
@@ -62,4 +71,40 @@ SCENARIO("Unary eval on intervals", "[core][analyses][interval][eval]")
...
+    {
+      auto negated_val = numeric_cast<mp_integer>(
+        five_to_eight.eval(ID_unary_minus).get_lower());
+      REQUIRE(negated_val.has_value());
+      REQUIRE(negated_val.value() == -8);
+
+      auto upper_value = numeric_cast<mp_integer>(
+        five_to_eight.eval(ID_unary_minus).get_upper());
+      REQUIRE(upper_value.has_value());
+      REQUIRE(upper_value.value() == -8);
```
smowton: Known bug that -((5, 8)) == (-8, -8)?

unit/util/interval/eval.cpp
```diff
@@ -62,4 +71,40 @@ SCENARIO("Unary eval on intervals", "[core][analyses][interval][eval]")
...
+
+    THEN("When we apply unary addition to it, nothing should happen")
+    {
+      REQUIRE(five_to_eight.eval(ID_unary_plus) == five_to_eight);
+    }
+
+    THEN("When we apply unary subtraction to it, it should be negated")
+    {
+      auto negated_val = numeric_cast<mp_integer>(
+        five_to_eight.eval(ID_unary_minus).get_lower());
```
smowton: Compare equality against expected interval?

src/util/interval.cpp
```diff
@@ -1660,13 +1659,28 @@ constant_interval_exprt::unary_minus(const constant_interval_exprt &a)
 constant_interval_exprt
 constant_interval_exprt::typecast(const typet &type) const
 {
-  exprt lower = get_lower();
-  lower.type() = type;
+  if(this->type().id() == ID_bool && is_int(type))
+  {
+    unsigned lower_num = !has_no_lower_bound() && get_lower().is_true();
```
smowton: has-no-lower-bound but get-lower returns something?

unit/util/interval/subtract.cpp
```diff
@@ -87,18 +87,73 @@ SCENARIO("subtract interval domain", "[core][analyses][interval][subtract]")
...
     }
   }
 }
+
+SCENARIO("Subtracting unsigned integers")
+{
+  WHEN("Subtracting two constant intervals")
+  {
+    auto lhs =
+      constant_interval_exprt(constant_exprt("1010", unsignedbv_typet(32)));
```
smowton: Use from_integer and perhaps binary numeric literals if you want to emphasise the bitpattern you intend. I think these might be broken too because constant_exprt expects hex representation?

src/util/interval.cpp
```diff
@@ -105,15 +105,16 @@ constant_interval_exprt::plus(const constant_interval_exprt &o) const
 }
 
 constant_interval_exprt
-constant_interval_exprt::minus(const constant_interval_exprt &o) const
+constant_interval_exprt::minus(const constant_interval_exprt &other) const
```
smowton, line 111: Split commit "Remove swap function in intervals Also, refactor minus and correct a typo in the comments" to do one thing at a time

smowton (approved):
Some things to check, but as you say happy to get this in then fix it up subsequently. I didn't look at the big "fix crap after rebase" commit or the huge 2000-line "add intervals" one due to lack of time.

unit/util/interval/eval.cpp
thk123: I guess - the reason for only considering singletons is each of the opertions is tested seperatly with both singleton and ranges, this test is really to check eval dispatches to the correct function

src/util/interval.cpp
thk123: _all_ the tests were based on using this broken equality :cry:

unit/util/interval/eval.cpp
thk123: It is but i've added a comment to highlight it.

unit/util/interval/modulo.cpp
thk123: No - it isn't wrong and we have no intention to fix it any time soon - a ticket will just sit not getting done. I would only raise it once someone actually wanted this to be more precise. Perhaps TODO is wrong, perhaps should just say note modulo is very imprecise.

unit/util/interval/comparisons.cpp
thk123: Yeah this seems like a confusing mess - I've broken it into 3 sections - same interval, overlapping interval, disjoint interval and used consistent intervals for equal and not equal.

unit/util/interval/eval.cpp
thk123: Fixed in a later commit

unit/util/interval/eval.cpp
thk123: Good idea - but this ends up being commented out as it doesn't work (ADA-535), I'll put a comment on the ticket suggested that'd be a good idea

unit/util/interval/eval.cpp
thk123: Typo that gets fixed in later commit

src/util/interval.cpp
thk123: Well it has a `not` before the `has_no_lower_bound` -This function is one of the functions I don't really get, but would want to tidy up only after the code that is already using it is merged so I can change both at the same time if necessary

unit/util/interval/subtract.cpp
thk123: This thankfully ends up going away

unit/util/interval/modulo.cpp
thk123: Do you mean why [-10 20] % [0, 5] = TOP but [-10 20] % [1, 5] = [-10, 20]? I've added a comment that taking modulo on something containing zero gives top since divide by zero

unit/util/interval/multiply.cpp
thk123: Swapped for a full equality comparison

src/util/interval.cpp
thk123: Go on then - since it's your last day :stuck_out_tongue:

danpoe (approved):

Issue #179, opened 2016-07-18
Error with shorts in Java frontend on SSA


mgudemann:
The following program:

```
public class short2
{
  short a;
  void f()
  {
    a++;
  }
}
```

fails with

```
converting SSA
+
  * type: signedbv
      * width: 16
  0: byte_extract_little_endian
      * type: signedbv
          * width: 16
      0: symbol
          * type: empty
          * identifier: tmp_object_factory$1!0#1
          * expression: symbol
              * type: empty
              * identifier: tmp_object_factory$1
          * L0: 0
          * L2: 1
          * #SSA_symbol: 1
      1: constant
          * type: signedbv
              * width: 64
              * #c_type: signed_long_int
          * value: 0000000000000000000000000000000000000000000000000000000000000100
  1: constant
      * type: signedbv
          * width: 32
          * #c_type: signed_int
      * value: 00000000000000000000000000000001
add/sub with mixed types
```

when using `cbmc --cover location --function short2.f short2.class`

thk123:
Just noting, for the avoidance of ambiguity, the commit that repeatedly references this issue is not a fix for this issue. (Commit: `(INT) Assume relation between two int vars...`)

peterschrammel:
fixed

PR #5362, opened 2020-05-28
Assume relation between two symbols

thk123:
Pulled out of #5361 as appears to be an unrelated bug fix. Not sure what bug it is fixing yet - will try and add a test! 

### DIFF
--- a/regression/goto-analyzer/intervals_simple-loops/test.desc
+++ b/regression/goto-analyzer/intervals_simple-loops/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+--intervals
+^EXIT=0$
+^SIGNAL=0$
+\[main\.assertion\.1\] line 8 assertion 0: UNKNOWN
+\[main\.assertion\.2\] line 13 assertion 0: UNKNOWN
+--
+--
+Test comparision between two symbols does not mark the
+result as bottom
--- a/src/util/interval_template.h
+++ b/src/util/interval_template.h
@@ -143,6 +143,47 @@ public:
     }
   }
 
+  void make_bottom()
+  {
+    lower_set = upper_set = true;
+    upper = T();
+    lower = upper + 1;
+  }
+
+  void make_less_than_eq(interval_templatet &i)
+  {
+    if(upper_set && i.upper_set)
+      upper = std::min(upper, i.upper);
+    if(lower_set && i.lower_set)
+      i.lower = std::max(lower, i.lower);
+  }
+
+  void make_less_than(interval_templatet &i)
+  {
+    make_less_than_eq(i);
+    if(singleton() && i.singleton() && lower == i.lower)
+    {
+      make_bottom();
+      i.make_bottom();
+    }
+  }
+
+  bool is_less_than_eq(const interval_templatet &i)
+  {
+    if(i.lower_set && upper_set && upper <= i.lower)
+      return true;
+    else
+      return false;
+  }
+
+  bool is_less_than(const interval_templatet &i)
+  {
+    if(i.lower_set && upper_set && upper < i.lower)
+      return true;
+    else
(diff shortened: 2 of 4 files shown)


thk123:
This doesn't fix #179. but a different issue on a private fork - adding a test now

regression/goto-analyzer/intervals_simple-loops/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--intervals
+^EXIT=0$
+^SIGNAL=0$
+\[main\.assertion\.1\] line 8 assertion 0: UNKNOWN
+\[main\.assertion\.2\] line 13 assertion 0: UNKNOWN
+--
+Test comparision between two symbols does not mark the
```
smowton, line 10: This is a do-not-match line, not a comment

src/util/interval_template.h
```diff
@@ -143,6 +143,47 @@ template<class T> class interval_templatet
...
+    if(singleton() && i.singleton() && lower == i.lower)
+    {
+      make_bottom();
+      i.make_bottom();
+    }
+  }
+
+  bool is_less_than_eq(const interval_templatet &i)
+  {
+    if(i.lower_set && upper_set && upper <= i.lower)
```
smowton, line 173: if(x) return true; else return false; --> return x

smowton, line 173: Cross-check since there are no comments to this effect in `interval_template.h`: these intervals are inclusive at both ends?

src/util/interval_template.h
```diff
@@ -143,6 +143,47 @@ template<class T> class interval_templatet
...
+  void make_bottom()
+  {
+    lower_set = upper_set = true;
+    upper = T();
+    lower = upper + 1;
+  }
+
+  void make_less_than_eq(interval_templatet &i)
+  {
+    if(upper_set && i.upper_set)
```
smowton, line 155: If my own upper bound is not set but i's is, shouldn't it now become set, and the same for i's lower bound inheriting mine?

src/util/interval_template.h
```diff
@@ -143,6 +143,47 @@ template<class T> class interval_templatet
...
+      upper = std::min(upper, i.upper);
+    if(lower_set && i.lower_set)
+      i.lower = std::max(lower, i.lower);
+  }
+
+  void make_less_than(interval_templatet &i)
+  {
+    make_less_than_eq(i);
+    if(singleton() && i.singleton() && lower == i.lower)
```
smowton, line 164: Shouldn't this do more, too? For example, if i has max 5 and this is now always less than it, this should have max value 4, and similarly i's lower bound should be max(i.lower, lower + 1)?

thk123:
@smowton all your comments seem correct, but are a question of precision. I'm reluctant to apply since:
 - this is a valid over-approximation 
 - this isn't my code
 - it should be deprecated by the new intervals in VSD 

Will fix the test desc problem

smowton (approved):

src/util/interval_template.h
hannes-steffenhagen-diffblue: @smowton It _can_ do a lot more, but the nice thing about AI is that as long as we don’t add impossible behaviour we don’t necessarily have to do everything as precise as possible from the start (we should make it as precise as reasonable obviously, and there’s a lot of things that still could be done here).

hannes-steffenhagen-diffblue:
IIRC this wasn’t a bug fix, this was there to implement some amount of narrowing for intervals.

thk123:
@hannes-steffenhagen-diffblue I believe it is a bug - previously that assert in there would be marked as success (unreachable) which is not true!

hannes-steffenhagen-diffblue (approved):

PR #5363, opened 2020-05-28
Java frontend: support lambda autoboxing involving generics

smowton:
Based on #5359, extending that implementation to also account for generics


jbmc/src/java_bytecode/lambda_synthesis.cpp
```diff
@@ -635,44 +701,40 @@ exprt box_or_unbox_type_if_necessary(
...
-      fresh_local, boxed_type_factory_method.symbol_expr(), {expr}});
-
-    return std::move(fresh_local);
-  }
+
+  const auto *primitive_type_info = get_java_primitive_type_info(
+    original_is_pointer ? required_type : original_type);
+  INVARIANT(
+    primitive_type_info != nullptr,
+    "A Java non-pointer type involved in a type disagreement should"
```
thk123, line 709: :question: What non-pointer java types are not primitives?

thk123 (approved):
Review of final commit only
lgtm - one question, but non-blocking

jbmc/src/java_bytecode/lambda_synthesis.cpp
smowton: Void! Which should have been excluded earlier. I guess this would also fire if you screwed up and fed it a deref'd pointer type, the string_typet @class_identifier, or similar.

PR #5365, opened 2020-05-28
Allow selective havocking of volatile variables

danpoe:


src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1622,13 +1622,8 @@ void goto_instrument_parse_optionst::instrument_goto_program()
...
 
-  // nondet volatile?
-  if(cmdline.isset("nondet-volatile"))
-  {
-    log.status() << "Making volatile variables non-deterministic"
-                 << messaget::eom;
-    nondet_volatile(goto_model);
-  }
+  parse_nondet_volatile_options(cmdline, options);
+  nondet_volatile(goto_model, options);
```
thk123, line 1627: :bulb: clearly this predates the PR - but performing the operation next to parsing the option seems wrong. Either the parsing should happen earlier, or the transformation later, so user gets typos in syntax quickly

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -13,9 +13,13 @@ Date: September 2011
 
 #include "nondet_volatile.h"
 
+#include <util/cmdline.h>
+#include <util/options.h>
 #include <util/std_expr.h>
 #include <util/symbol_table.h>
 
+using havoc_predt = std::function<bool(const exprt &)>;
```
thk123: :pick: avoid contractions:
```suggestion
using havoc_predicatet = std::function<bool(const exprt &)>;
```

src/goto-instrument/goto_instrument_parse_options.cpp
thk123, line 1627: :bulb: Rather than parsing the cmdline into the equally unsafe optionst, consider using a separate data structure (`nondet_volitalie_optionst`) that contains the type safe options only

src/goto-instrument/nondet_volatile.h
```diff
@@ -14,6 +14,41 @@ Author: Daniel Kroening, kroening@kroening.com
...
+#define OPT_NONDET_VOLATILE \
+  "(" NONDET_VOLATILE_OPT ")" \
+  "(" NONDET_VOLATILE_VARIABLE_OPT "):"
+
+#define HELP_NONDET_VOLATILE \
+  " --" NONDET_VOLATILE_OPT "            makes reads from volatile variables " \
+  "non-deterministic\n" \
+  " --" NONDET_VOLATILE_VARIABLE_OPT " <variable>\n" \
+  "                              makes reads from given volatile variable " \
+  "non-deterministic\n"
```
thk123, line 33: :bulb: Not sure if even possible with the way parse options work, but I think a nicer syntax from a user perspective might be:

--nondet-volatile -> all volatile readss nondet
 --nondet-volatile x -> reads from x nondet 

Rather than having separate flags

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -101,25 +121,102 @@ nondet_volatile(symbol_tablet &symbol_table, goto_programt &goto_program)
...
+  }
+}
+
+void nondet_volatile(goto_modelt &goto_model, const optionst &options)
+{
+  if(options.get_bool_option(NONDET_VOLATILE_OPT))
+  {
+    nondet_volatile(goto_model);
+  }
+  else if(options.is_set(NONDET_VOLATILE_VARIABLE_OPT))
```
thk123, line 183: :pick: For future - PR would have been easier to review if you'd had one commit that introduced the flag, but had it do nothing (just introduced the code scaffaloding)

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -101,25 +121,102 @@ nondet_volatile(symbol_tablet &symbol_table, goto_programt &goto_program)
...
+      const symbolt *symbol;
+
+      if(ns.lookup(id, symbol))
+      {
+        throw invalid_command_line_argument_exceptiont(
+          "given name " + id2string(id) + " not found in symbol table",
+          NONDET_VOLATILE_VARIABLE_OPT);
+      }
+
+      if(symbol->is_type || !symbol->type.get_bool(ID_C_volatile))
```
thk123: :question: Is the `->is_type` checking anything? Can types by volatile? If not, I'd remove it, as I think it's confusing (you don't for example check if it is a function, which I think are also in the symbol table)

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -101,25 +121,102 @@ nondet_volatile(symbol_tablet &symbol_table, goto_programt &goto_program)
...
+      {
+        throw invalid_command_line_argument_exceptiont(
+          "given name " + id2string(id) + " does not represent a volatile "
+          "variable",
+          NONDET_VOLATILE_VARIABLE_OPT);
+      }
+
+      INVARIANT(
```
thk123: :question: Indeed - why is this an invariant? Is it possible to trigger this through invalid input - in which case should throw an exception

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -101,25 +121,102 @@ nondet_volatile(symbol_tablet &symbol_table, goto_programt &goto_program)
...
+        symbol->type.id() != ID_code,
+        "symbol must not represent a function");
+    }
+
+    auto should_havoc = [&variables](const exprt &expr) {
```
thk123, line 218: :no_entry_sign: This looks like a dangling reference..? You capture variables by reference, but they are a local variable that goes out of scope, but you return the lambda. I think `=variables` is needed here, or convince me why I'm wrong!

regression/goto-instrument/nondet-volatile-02/test.c
```diff
@@ -0,0 +1,25 @@
+#include <assert.h>
+
+volatile int a;
```
thk123, line 3: :no_entry_sign: These are already nondet since you never assign to them. I believe main should start with assignments to 0 for all?

regression/goto-instrument/nondet-volatile-02/test.c
```diff
@@ -0,0 +1,25 @@
...
+{
+  assert(c == 0);
+  assert(c != 0);
+}
+
+void main()
+{
+  volatile int d = 0;
+  
+  assert(a == 0);
```
thk123, line 8: :question: symetric assert a != 0?

regression/goto-instrument/nondet-volatile-02/test.desc
```diff
@@ -0,0 +1,16 @@
...
+\[main.assertion.2\] line \d+ assertion b == 0: FAILURE
+\[main.assertion.3\] line \d+ assertion b != 0: FAILURE
+\[main.assertion.4\] line \d+ assertion d == 0: FAILURE
+\[main.assertion.5\] line \d+ assertion d != 0: FAILURE
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+Check that volatile global, local, and parameter variables are havocked when
+given via the --nondet-volatile-variable option
```
thk123: :pick: add to description - `a is not havoc-d so the assert succeeds`

regression/goto-instrument/nondet-volatile-02/test.c
thk123: No - because we _don't_ havoc a - I don't understand why this passes, shouldn't a be nondet

regression/goto-instrument/nondet-volatile-02/test.c
thk123: The point of the test is that a is not havoc'd

thk123 (commented):
I think I'd add tests for:

 - nodet variable after deref'ing a pointer
 - nondet variable after indexing into an array

But the dangling reference needs to be checked

src/goto-instrument/nondet_volatile.h
danpoe: Currently not possible as the command line parsing would need to disambiguate whether the thing following `--nondet-volatile` is an argument to it, or an argument to goto-instrument or another option.

src/goto-instrument/nondet_volatile.cpp
danpoe: The lambda is not returned. The declaration of `variables`, the lambda definition, and the call to `nondet_volatile()` (which uses the lambda) are all in the same scope.

src/goto-instrument/nondet_volatile.cpp
danpoe: Yes, the type symbols can have a volatile flag in `symbol->type`. That's because of how non-type symbols and type symbols are represented. For non-type symbols, `symbol->type` gives the type of the symbol. For type symbols (i.e., type aliases), `symbol->type` gives the aliased type. For example, if type symbol `x` is an alias for `volatile int`, then `x.type` would have a volatile flag.

src/goto-instrument/goto_instrument_parse_options.cpp
smowton: Looks like goto-instrument never separated command-line validation from processing, but now is probably the time to start

src/goto-instrument/nondet_volatile.cpp
smowton: Perhaps should_havoc_predicatet? To disambiguate from is_havocked or some other havocking-related question

src/goto-instrument/nondet_volatile.cpp
danpoe: Assuming well-formed goto programs, this should always hold as the type of functions cannot be volatile.

src/goto-instrument/nondet_volatile.cpp
danpoe: Done

src/goto-instrument/nondet_volatile.cpp
smowton: Are you looking to specifically catch global variables, or are locals reasonable targets for this flag? If the former consider checking `is_static_lifetime`

smowton (approved):

regression/goto-instrument/nondet-volatile-02/test.c
danpoe: Variables with static storage duration are initialized to zero (including volatile ones). :peach:

regression/goto-instrument/nondet-volatile-02/test.c
danpoe: :peach:

regression/goto-instrument/nondet-volatile-02/test.desc
danpoe: Added

src/goto-instrument/goto_instrument_parse_options.cpp
danpoe: I've moved this call all the way up to the top of the function (which is the best we can do at this point). Unfortunately, unlike cbmc and goto-analyzer, goto-instrument doesn't have a separate `get_command_line_options()` phase at the moment.

danpoe: This follows the cbmc convention of parsing `cmdlinet` into `optionst` at first. The actual typechecking (which requires a `goto_model`) is done in `nondet_volatile()`.

src/goto-instrument/nondet_volatile.cpp
danpoe: Locals would be reasonable targets as well.

danpoe: Good point. Yes, local volatiles probably wouldn't make much sense. Changed it.

src/goto-instrument/nondet_volatile.cpp
thk123: Hmm - must have been half asleep - right you are.

PR #5367, opened 2020-05-29
Java lambda synthetic classes: don't modify the symtab while iterating over it

smowton:


peterschrammel (approved):

thk123 (commented):
A test that would have revealed this would be a good addition.

thk123 (approved):
But the change is sensible so I won't block on it

PR #5370, opened 2020-05-30
add stack dumps for Windows

kroening:
This adds the missing code for dumping stacks on Windows.

### DIFF
--- a/src/util/invariant.cpp
+++ b/src/util/invariant.cpp
@@ -10,11 +10,20 @@ Author: Martin Brain, martin.brain@diffblue.com
 
 #include "freer.h"
 
+#include <iomanip>
 #include <memory>
-#include <string>
 #include <sstream>
+#include <string>
 
 #include <iostream>
+
+#ifdef _WIN32
+// the ordering of includes is required
+// clang-format off
+#include <windows.h>
+#include <dbghelp.h>
+// clang-format on
+#endif
 
 bool cbmc_invariants_should_throw = false;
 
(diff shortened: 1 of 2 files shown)


src/util/invariant.cpp
```diff
@@ -93,6 +102,29 @@ void print_backtrace(
...
     }
 
+#elif defined(_WIN32)
+
+  void *stack[50];
+  HANDLE process = GetCurrentProcess();
+
+  SymInitialize(process, NULL, TRUE);
+
+  auto number_of_frames = CaptureStackBackTrace(0, 50, stack, NULL);
```
smowton: Use a sizeof or a define for 50 to avoid the obvious mistake when someone wants more frames

src/util/invariant.cpp
```diff
@@ -93,6 +102,29 @@ void print_backtrace(
...
+
+  auto symbol = static_cast<SYMBOL_INFO *>(calloc(sizeof SYMBOL_INFO + 256, 1));
```
smowton: Similarly for 256 here and 255 below

smowton: This allocation procedure is mad enough you should probably cite https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/ns-dbghelp-symbol_info for the future reader who thinks that surely can't be how you do it :)

smowton (approved):

src/util/invariant.cpp
kroening: Now using sizeof.

src/util/invariant.cpp
kroening: added

kroening: now using a variable.

PR #5372, opened 2020-06-02
Implement basic operations for VSD Value Sets

hannes-steffenhagen-diffblue:
This implements some of the basic functions for abstract values, such as creation, merge, and converting single-value sets to a constant (which is needed for simplification later on).

Nothing is hooked up to goto-analyser yet, but the functionality is unit tested.

### DIFF
--- a/src/analyses/variable-sensitivity/value_set_abstract_value.cpp
+++ b/src/analyses/variable-sensitivity/value_set_abstract_value.cpp
@@ -9,6 +9,70 @@
 #include "value_set_abstract_value.h"
 
 value_set_abstract_valuet::value_set_abstract_valuet(const typet &type)
-  : abstract_valuet{type}
+  : abstract_valuet{type}, values{}
+{
+}
+
+const value_set_abstract_valuet::valuest &
+value_set_abstract_valuet::get_values() const
+{
+  PRECONDITION(!is_top());
+  PRECONDITION(!is_bottom());
+  return values;
+}
+
+abstract_object_pointert
+value_set_abstract_valuet::merge(abstract_object_pointert other) const
+{
+  if(is_top())
+  {
+    return shared_from_this();
+  }
+
+  if(other->is_top())
+  {
+    return other;
+  }
+
+  if(is_bottom())
+  {
+    return other;
+  }
+
+  if(other->is_bottom())
+  {
+    return shared_from_this();
+  }
+
+  if(
+    auto other_value_set =
+      dynamic_cast<const value_set_abstract_valuet *>(other.get()))
+  {
+    valuest merged_values{values};
+    auto const &other_values = other_value_set->get_values();
+    merged_values.insert(other_values.begin(), other_values.end());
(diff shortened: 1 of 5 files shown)


src/analyses/variable-sensitivity/value_set_abstract_value.cpp
```diff
@@ -9,6 +9,54 @@
...
 value_set_abstract_valuet::value_set_abstract_valuet(const typet &type)
-  : abstract_valuet{type}
+  : abstract_valuet{type}, values{}
 {
 }
+
+const value_set_abstract_valuet::valuest&
+value_set_abstract_valuet::get_values() const
+{
+  PRECONDITION(!is_top() && !is_bottom());
```
danpoe: Nit: Split precondition in two.

src/analyses/variable-sensitivity/value_set_abstract_value.cpp
```diff
@@ -9,6 +9,54 @@
...
+      merged_values.insert(other_values.begin(), other_values.end());
+    }
+    return std::make_shared<value_set_abstract_valuet>(type(), std::move(merged_values));
+  }
+  return abstract_objectt::merge(other);
+}
+
+
+value_set_abstract_valuet::value_set_abstract_valuet(const typet &type, valuest values)
+  : abstract_valuet{type, values.size() > max_value_set_size, values.empty()}, values{std::move(values)}
```
danpoe: Could use a forwarding reference and then only copy `values` when `values.size() <= max_value_set_size`.

src/analyses/variable-sensitivity/value_set_abstract_value.cpp
```diff
@@ -9,6 +9,54 @@
...
+  }
+
+  if(other->is_bottom()) {
+    return shared_from_this();
+  }
+
+  if(auto other_value_set = dynamic_cast<const value_set_abstract_valuet*>(other.get()))
+  {
+    valuest merged_values{values};
+    if(!other_value_set->is_bottom()) {
```
danpoe: Shouldn't this have been checked by the if above already?

src/analyses/variable-sensitivity/value_set_abstract_value.cpp
```diff
@@ -9,6 +9,54 @@
...
+      auto const other_values = other_value_set->get_values();
```
danpoe: Should be `const auto &` to avoid a copy.

danpoe (approved):
Looks good, and thanks for the thorough set of tests as well

NlightNFotis (approved):
LGTM

src/analyses/variable-sensitivity/value_set_abstract_value.cpp
hannes-steffenhagen-diffblue: I don’t think there’s a huge amount of value in using a forwarding reference here (these aren’t expensive to copy once), but I think I’ll clear these if they’re too big.

PR #5373, opened 2020-06-04
Filter out illegal characters from identifiers in typecheck_expr.

NlightNFotis:
This change removes some characters, like "*{}", from identifier names that were autogenerated. The reason for this is that these names are now being passed on to other tools, and they cause these to fail with parsing errors.


src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -1849,6 +1849,34 @@ void c_typecheck_baset::typecheck_expr_side_effect(side_effect_exprt &expr)
   }
 }
 
+/// type2name generates strings that aren't valid C identifiers
+/// If we want utilities like dump_c to work properly identifiers
+/// should ideally always be valid C identifiers
+/// This replaces some invalid characters that can appear in type2name output.
+static std::string clean_type_name(const std::string &type_name)
```
thk123: :grey_question: Could (should?) this be fixed in type_to_name? Seems odd that the parsing would be concerned about valid C names. 

I think the answer is no: type_to_name is making valid C stuff like `int*`, and then this code is misusing that to put straight back into a C identifier.

regression/goto-cc-instrument-cbmc/identifier_names_illegal_characters/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test.c
+
```
thk123, line 3: :no_entry_sign: Positive regexes for how the function declarations are printed and negative regexes for what they shouldn't be?

NlightNFotis: Could they be better left in the comment section? :thinking: 

In this particular case, for this pipeline (goto-cc -> goto-instrument -> cbmc), the end result if the function declarations are wrong is that `cbmc` results in a parsing error.

To implement your change, I imagine I have to remove cbmc from there, and do regex pattern matching on the `--dump-c` output of `goto-instrument`, which I think is less ideal for our usecase.

thk123: Good point - this is fine

thk123 (approved):
Don't know why the codecov report has a cross - it seems fine (100% on new code, no losses in old code...)

src/ansi-c/c_typecheck_expr.cpp
NlightNFotis: We didn't want to touch `type_to_name`, especially for this fix, because we don't know how much depends on the output it produces, and we were afraid of breaking a lot more than we could fix.

chrisr-diffblue (commented):
This feels like the wrong place to fix this, and likely too narrow a fix... I'd be more inclined to push this into type2name.cpp/cpp_type2name.cpp as a new function 'type2identifier' - it could probably still do something morally equivalent to whats happening in this PR, just better named (I don't like functions called 'clean...' because its unclear what that means) and better located. Its probably also worth a skim read through the existing type2name functions to see what they can produce because I don't think this this current 'clean_type_name' function covers all the other invalid identifiers that type2name can produce.

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -1890,17 +1891,19 @@ void c_typecheck_baset::typecheck_side_effect_function_call(
...
-            type2name(parameters.front().type().subtype(), *this);
+            type2identifier(parameters.front().type().subtype(), *this);
         }
         else
         {
           identifier_with_type = id2string(identifier) + "_" +
-                                 type2name(parameters.front().type(), *this);
+                                 type2identifier(parameters.front().type(), *this);
         }
+        identifier_with_type = id2string(identifier_with_type);
```
chrisr-diffblue: This looks wrong - `identifier_with_type` is an `irep_idt`, so I can understand doing id2string... but then you assign that back to itself...

NlightNFotis: I looked into this. Indeed it looks extraneous. I've removed it.

chrisr-diffblue (approved):

PR #5374, opened 2020-06-04
Selective modelling of volatile variables

danpoe:

### DIFF
--- a/regression/goto-instrument/nondet-volatile-model-01/test.c
+++ b/regression/goto-instrument/nondet-volatile-model-01/test.c
@@ -1,0 +1,20 @@
+#include <assert.h>
+
+volatile int a;
+volatile int b;
+
+int model()
+{
+  int value;
+  __CPROVER_assume(value >= 0);
+  return value;
+}
+
+void main()
+{
+  assert(a == 0);
+
+  assert(b >= 0);
+  assert(b == 0);
+  assert(b != 0);
+}
--- a/regression/goto-instrument/nondet-volatile-model-03/test.c
+++ b/regression/goto-instrument/nondet-volatile-model-03/test.c
@@ -1,0 +1,20 @@
+#include <assert.h>
+
+volatile int a;
+volatile int b;
+
+int model_a()
+{
+  return 1;
+}
+
+int model_b()
+{
+  return 2;
+}
+
+void main()
+{
+  int result = a + b;
+  assert(result == 3);
+}
--- a/src/goto-instrument/nondet_volatile.cpp
+++ b/src/goto-instrument/nondet_volatile.cpp
@@ -19,39 +19,52 @@ Date: September 2011
 #include <util/string_utils.h>
 #include <util/symbol_table.h>
 
-using havoc_predicatet = std::function<bool(const exprt &)>;
-
 class nondet_volatilet
 {
 public:
-  static void nondet_volatile(
-    symbol_tablet &symbol_table,
-    goto_programt &goto_program,
-    havoc_predicatet should_havoc);
+  nondet_volatilet(goto_modelt &goto_model, const optionst &options)
+    : goto_model(goto_model), all_nondet(false)
+  {
+    typecheck_options(options);
+  }
 
-  static const symbolt &
-  typecheck_variable(const irep_idt &id, const namespacet &ns);
+  void operator()()
+  {
+    if(!all_nondet && nondet_variables.empty() && variable_models.empty())
+    {
+      return;
+    }
+
+    for(auto &f : goto_model.goto_functions.function_map)
+    {
+      nondet_volatile(goto_model.symbol_table, f.second.body);
+    }
+
+    goto_model.goto_functions.update();
+  }
 
 private:
   static bool is_volatile(const namespacet &ns, const typet &src);
 
-  static void nondet_volatile_rhs(
-    const symbol_tablet &symbol_table,
-    exprt &expr,
-    havoc_predicatet should_havoc);
+  void handle_volatile_expression(exprt &expr, const namespacet &ns);
 
-  static void nondet_volatile_lhs(
-    const symbol_tablet &symbol_table,
-    exprt &expr,
-    havoc_predicatet should_havoc);
+  void nondet_volatile_rhs(const symbol_tablet &symbol_table, exprt &expr);
 
-  static void typecheck_model(
+  void nondet_volatile_lhs(const symbol_tablet &symbol_table, exprt &expr);
+
+  void
+  nondet_volatile(symbol_tablet &symbol_table, goto_programt &goto_program);
+
+  const symbolt &typecheck_variable(const irep_idt &id, const namespacet &ns);
+
+  void typecheck_model(
     const irep_idt &id,
     const symbolt &variable,
     const namespacet &ns);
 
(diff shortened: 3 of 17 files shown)


regression/goto-instrument/nondet-volatile-model-01/test.c
```diff
@@ -0,0 +1,20 @@
+#include <assert.h>
+
+volatile int a;
+volatile int b;
```
hannes-steffenhagen-diffblue, line 4: I think one of these should have a regression test involving the interaction with non-volatile variables; E.g.

```
volatile int x;
int y = x;
assert(x == y); // fails because volatile
```

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -20,7 +20,29 @@ Date: September 2011
 
 using havoc_predicatet = std::function<bool(const exprt &)>;
 
-static bool is_volatile(const namespacet &ns, const typet &src)
+class nondet_volatilet
```
hannes-steffenhagen-diffblue, line 23: So... this is just a namespace? Maybe the commit should say why you’re doing this.

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -158,6 +177,146 @@ void nondet_volatilet::nondet_volatile(
...
+const symbolt &nondet_volatilet::typecheck_variable(
+  const irep_idt &id,
+  const namespacet &ns)
+{
+  const symbolt *symbol;
+
+  if(ns.lookup(id, symbol))
+  {
+    throw invalid_command_line_argument_exceptiont(
+      "given name `" + id2string(id) + "` not found in symbol table",
```
hannes-steffenhagen-diffblue: :pick:  Should this also be "symbol" rather than "given name"?

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -158,6 +177,146 @@ void nondet_volatilet::nondet_volatile(
...
+      NONDET_VOLATILE_VARIABLE_OPT);
```
hannes-steffenhagen-diffblue: :pick: Usually we start these with `--` because the `_OPT` things don’t have it; Without the `--` it might be slightly harder to recognise that this is referring to a command line option.

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -158,6 +177,146 @@ void nondet_volatilet::nondet_volatile(
...
+  }
+
+  if(!symbol->is_static_lifetime || !symbol->type.get_bool(ID_C_volatile))
```
hannes-steffenhagen-diffblue, line 279: It's a bit unclear to me why this only works with global variables.

regression/goto-instrument/nondet-volatile-model-03/test.c
```diff
@@ -0,0 +1,20 @@
...
+}
+
+int model_b()
+{
+  return 2;
+}
+
+void main()
+{
+  int result = a + b;
```
hannes-steffenhagen-diffblue, line 18: Also, does this work if the global is read via volatile pointer? (I'm assuming no, just for clarification).

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -246,9 +279,9 @@ void nondet_volatilet::typecheck_model(
 }
 
 void nondet_volatilet::typecheck_options(
-  const goto_modelt &goto_model,
   const optionst &options)
 {
+  PRECONDITION(!all_nondet);
```
hannes-steffenhagen-diffblue, line 338: :pick:  There's a lot happening in 782395c0457cd970c07266a337f5d0dde3b3281d which makes it a bit hard to recognise as a refactoring (i.e. there's so much code moving around its a bit hard to see where what is moving).

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -101,27 +108,67 @@ bool nondet_volatilet::is_volatile(const namespacet &ns, const typet &src)
 
 void nondet_volatilet::handle_volatile_expression(
   exprt &expr,
-  const namespacet &ns)
+  const namespacet &ns,
```
hannes-steffenhagen-diffblue, line 105: :pick: bd81144c0d2fb0347497adf3fa1df20b021a4cc7

Says it adds a new `--nondet-volatile-model` option, but this option is actually added in a different commit.

src/goto-instrument/nondet_volatile.cpp
```diff
@@ -76,66 +99,76 @@ bool nondet_volatilet::is_volatile(const namespacet &ns, const typet &src)
...
 void nondet_volatilet::nondet_volatile_lhs(
   const symbol_tablet &symbol_table,
-  exprt &expr,
-  havoc_predicatet should_havoc)
+  exprt &expr)
 {
   if(expr.id()==ID_if)
   {
-    nondet_volatile_rhs(symbol_table, to_if_expr(expr).cond(), should_havoc);
+    nondet_volatile_rhs(symbol_table, to_if_expr(expr).cond());
```
hannes-steffenhagen-diffblue: :question:  Is this case match already covered by existing tests somewhere?

hannes-steffenhagen-diffblue (approved):
I'm OK with the code changes, if you have the time I think the commits themselves could be cleaned up a bit.

src/goto-instrument/nondet_volatile.cpp
danpoe: This is currently untested as the original nondet volatile code on which this builds was untested. I'll add more tests for the existing code in a separate PR.

src/goto-instrument/nondet_volatile.cpp
danpoe: Changed commit message.

regression/goto-instrument/nondet-volatile-model-03/test.c
danpoe: Yes that's right, it currently only works for direct variable reads.

(One could also mark a pointer as nondet, in which case the pointer dereferences would be nondet as well. Though when `--pointer-check` would be enabled it would fail.)

src/goto-instrument/nondet_volatile.cpp
danpoe: Since the volatile variables will typically model hardware resources, it would probably make less sense to have volatile variables with automatic storage duration.

src/goto-instrument/nondet_volatile.cpp
danpoe: Added

src/goto-instrument/nondet_volatile.cpp
danpoe: Changed

src/goto-instrument/nondet_volatile.cpp
danpoe: Mentioned in the commit message that this is in preparation for changes to come.

regression/goto-instrument/nondet-volatile-model-01/test.c
danpoe: Added

PR #5205, opened 2019-12-09
Changes to util needed to merge VSD

martin-cs:
Another PR from the variable-sensitivity-domain branch.  A few minor additions to core things in util/.  I have kept the old commit messages to preserve authorship but TBH they should probably be reworded at least.

- [ ] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #5375, opened 2020-06-05
Util changes needed for VSD

thk123:
Rebased version of #5205, with two additions:

 - tests for the const iterators added to base symbol table
 - stop definitions in symbol_table hiding the implemention in symbol_table_baset (risks these defintions going out of sync) 

### DIFF
--- a/src/util/symbol_table.h
+++ b/src/util/symbol_table.h
@@ -120,17 +120,8 @@ public:
     return iteratort(internal_symbols.end());
   }
 
-  typedef symbolst::const_iterator const_iteratort;
-
-  virtual const_iteratort begin() const
-  {
-    return internal_symbols.begin();
-  }
-
-  virtual const_iteratort end() const
-  {
-    return internal_symbols.end();
-  }
+  using symbol_table_baset::begin;
+  using symbol_table_baset::end;
 
   /// Check that the symbol table is well-formed
   void validate(const validation_modet vm = validation_modet::INVARIANT) const;
--- a/src/util/symbol_table_base.cpp
+++ b/src/util/symbol_table_base.cpp
@@ -60,6 +60,16 @@ void symbol_table_baset::show(std::ostream &out) const
     out << symbols.at(name);
 }
 
+symbol_table_baset::const_iteratort symbol_table_baset::end() const
+{
+  return symbols.end();
+}
+
+symbol_table_baset::const_iteratort symbol_table_baset::begin() const
+{
+  return symbols.begin();
+}
+
 /// Print the contents of the symbol table.
 /// \param out: The ostream to direct output to
 /// \param symbol_table: The symbol table to print out
--- a/unit/util/irep.cpp
+++ b/unit/util/irep.cpp
@@ -22,7 +22,7 @@ SCENARIO("irept_memory", "[core][utils][irept]")
       const std::size_t ref_count_size = 0;
 #endif
 
-#ifndef USE_STRING
+#ifndef USE_STD_STRING
       const std::size_t data_size = sizeof(dstringt);
       REQUIRE(sizeof(dstringt) == sizeof(unsigned));
 #else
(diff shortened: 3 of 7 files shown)


unit/util/irep.cpp
```diff
@@ -22,7 +22,7 @@ SCENARIO("irept_memory", "[core][utils][irept]")
       const std::size_t ref_count_size = 0;
 #endif
 
-#ifndef USE_STRING
+#ifndef USE_STD_STRING
```
hannes-steffenhagen-diffblue, line 25: Side note: I’d really love if we could just get rid of `USE_STD_STRING` already. I’m not entirely convinced we keep it around for anything other than nostalgia at this point.

src/util/symbol_table_base.cpp
```diff
@@ -60,6 +60,16 @@ void symbol_table_baset::show(std::ostream &out) const
     out << symbols.at(name);
 }
 
+symbol_table_baset::const_iteratort symbol_table_baset::end() const
+{
+  return symbols.end();
```
hannes-steffenhagen-diffblue, line 65: 96bd3ae8e86ac8712a662e80d710c05a64b8442c `s/interate/iterate` in the commit message.

Also this commit should be removed; Current develop already has begin/end for `symbol_tablet`.

unit/util/symbol_table.cpp
```diff
@@ -17,14 +17,55 @@ TEST_CASE("Iterating through a symbol table", "[core][utils][symbol_tablet]")
 
   symbol_table.insert(symbol);
 
-  int counter = 0;
-  for(auto &entry : symbol_table)
+  SECTION("Non const iterator")
```
hannes-steffenhagen-diffblue, line 20: This commit (a9afb9a25f3fd32e5e227bc9a91d7d5228bf38f6) can also be removed; symbol table iterators are tested in `unit/util/symbol_table.cpp`

src/util/symbol_table.h
```diff
@@ -122,15 +122,8 @@ class symbol_tablet : public symbol_table_baset
...
-  {
-    return internal_symbols.begin();
-  }
-
-  virtual const_iteratort end() const
-  {
-    return internal_symbols.end();
-  }
+  using symbol_table_baset::begin;
+  using symbol_table_baset::end;
```
hannes-steffenhagen-diffblue, line 124: develop `symbol_table_baset::begin/end` are both purely virtual, so can not be used in this way.

hannes-steffenhagen-diffblue (changes requested):
Other than the `USE_STD_STRING` change I don’t believe any of these changes should be merged as they already exist in some form or another on develop.

unit/util/symbol_table.cpp
thk123: I can see a single class to `begin()` but no call to end. This also tests the correct dispatching when you have a `symbol_tablet` but the compile time type is `symbol_table_baset`

src/util/symbol_table.h
thk123: Discussed out of band: more consistent to have both const and non-const iterators be virtual and defined in the base class.

thk123:
@hannes-steffenhagen-diffblue I've made the begin/end virtual in the base class as agreed, ready for [re-review](https://github.com/diffblue/cbmc/compare/acc46d626ecb385499286a86194eaa1527d1c09f..dca07dda7bb9a504cef9aca5d302d6f542f4267e)

unit/util/irep.cpp
martin-cs: I believe the only use case for it is / was debugging problems in the `dtringt` implementation.  I am not sure how many issues it will find that have not already been found.

martin-cs (approved):
Am happy once @hannes-steffenhagen-diffblue 's comments have been addressed.  I think that it will still work with the patch set.

hannes-steffenhagen-diffblue (approved):
I still think it’s a bit inconsistent we have a pure virtual mut `begin/end` but an implementation for const `begin/end`. But I’d not block on that if that’s the only complaint.

thk123:
@peterschrammel if you get a min, would you be able to provide the code owner review for this short PR? You reviewed this originally in #5205, Martin addressed your comments (this is a rebased version)

peterschrammel (approved):

PR #5383, opened 2020-06-18
Make filter_goto_model remove file local symbols

hannes-steffenhagen-diffblue:
Since we want goto_harness to print out as little as possible to be
useful this is helpful, and this also lets us avoid bugs with duplicate
definitions (e.g. a static variable that’s now in the produced harness.c
– and thus a *different* static variable!).

Also marks non-file-local globals as extern so we avoid double
definitions.


martin-cs (approved):
Obviously some kind of testing would be nice but I appreciate that this is more "desired behaviour" rather than right / wrong and also that testing negatives is kind of awkward.  If there is a test somewhere that exercises this behaviour that will be fine.

thk123 (commented):
Code changes LGTM :+1:

src/goto-harness/goto_harness_parse_options.cpp
```diff
@@ -51,12 +51,35 @@ static void filter_goto_model(
...
-      symbol.second.is_function())
-      goto_model_with_harness.unload(symbol.first);
+    auto &symbol =
+      goto_model_with_harness.symbol_table.get_writeable_ref(symbol_id);
+    if(symbol.is_function())
+    {
+      // We don’t want bodies of functions that already existed in the
+      // symbol table (i.e. not generated by us)
+      goto_model_with_harness.unload(symbol_id);
+      if(symbol.is_file_local)
```
hannes-steffenhagen-diffblue, line 63: Heads up this had to be changed because my stupid brain forgot that you can only take one branch in an if-else chain.

NlightNFotis (approved):
LGTM

PR #5387, opened 2020-06-18
Add dbghelp.lib to windows cmake build of util.lib

markrtuttle:
This pull requests add DgbHelp.lib to the list of libraries for util.lib to link against.  Methods from the Visual Studio library DbgHelp like SymInitialize were added to print_backtrace in src/util/invariant.cpp to produce stack traces on windows.

### DIFF
--- a/src/util/CMakeLists.txt
+++ b/src/util/CMakeLists.txt
@@ -49,3 +49,6 @@ add_dependencies(util generate_version_cpp)
 generic_includes(util)
 
 target_link_libraries(util big-int langapi)
+if(WIN32)
+  target_link_libraries(util dbghelp)
+endif()


martin-cs (approved):
Seems reasonable.  There should be at least one regression test that purposely generates an assertion failure and thus a backtrace.  If this could be made to work on Windows as well then we could check fo the problems caused by the lack of this library.

markrtuttle:
@martin-cs The problem is that cbmc fails to build with cmake in the visual studio command shell without this change.  This was not caught, I suspect, because CI has no cmake within the visual studio command shell.  The windows cmake codebuild job appears to be running under powershell, based on the syntax of the shell commands.  It would be reasonable to add another build to CI, but I don't know what regression test would make sense.

martin-cs:
Thanks for the clarrification @markrtuttle

peterschrammel (approved):

PR #5391, opened 2020-06-23
Add control dependencies on function call

martin-cs:
This is another shard from the VSD branch.  Originally by @danpoe it adds control dependency edges to the traditional dependency graph.  This will affect the output of goto-analyzer when generating the graph and could affect `full-slice` but I think it handles calls independently and this shouldn't make it any less safe.

### DIFF
--- a/src/analyses/dependence_graph.cpp
+++ b/src/analyses/dependence_graph.cpp
@@ -214,8 +214,9 @@ void dep_graph_domaint::transform(
         std::all_of(
           std::next(control_deps.begin()),
           control_deps.end(),
-          [](const goto_programt::const_targett &d)
-            { return d->is_function_call(); }),
+          [](const goto_programt::const_targett &d) {
+            return d->is_function_call();
+          }),
         "All entries must be function calls");
 
       control_deps.clear();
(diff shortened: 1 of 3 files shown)


src/analyses/dependence_graph.cpp
```diff
@@ -203,6 +203,26 @@ void dep_graph_domaint::transform(
...
+          std::next(control_deps.begin()),
+          control_deps.end(),
+          [](const goto_programt::const_targett &d) {
+            return d->is_function_call();
+          }),
+        "All entries must be function calls");
+
+      control_deps.clear();
+    }
+  }
```
danpoe, line 224: I remember that I wrote this piece of code, though I'm not convinced anymore that it does what I intended. I think it was supposed to check that either all of the control deps are function calls, or none of them is. Though this only checks that if the first dep is a function call, then all of them are.

danpoe (approved):

src/analyses/dependence_graph.cpp
martin-cs: It is not critical for me that this gets merged.  My plan is to trying to break parts of the VSD branch off and get them merged separately.  If they are no longer needed or should stay a branch that is fine.  I'm willing to follow your lead / preference on this one.

danpoe: Ah I was referring to only lines 206-223 not to the whole PR. Nevermind though, I checked again and also those changes are correct actually. So the PR can go in as is.

martin-cs: Great!

PR #5394, opened 2020-06-25
Specify exactly what the comment at the top of the file should be

thk123:
It is confusing to reference existing files, as existing files are a little inconsistent (some use a module specifier, more or less at
random). This provides a single source of truth for what should be there.

### DIFF
--- a/CODING_STANDARD.md
+++ b/CODING_STANDARD.md
@@ -61,12 +61,19 @@ Formatting is enforced using clang-format. For more information about this, see
 # Comments
 - Do not use `/* */`
 - Each source and header file must start with a comment block stating the
-  author. See existing source for an example of the format of this block. This
-  should be followed by a Doxygen `\file` comment:
+  author, followed by a Doxygen comment about the file:
+
   ```c++
+  /*******************************************************************\
+
+  Author: Author Name
+
+  \*******************************************************************/
+
   /// \file
   /// <Some information about this file goes here>
   ```
+
   Note that the `\file` tag must be immediately followed by a newline in order
   for Doxygen to relate the comment to the current file.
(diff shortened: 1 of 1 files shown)


martin-cs (approved):
Approve because it is an improvement.

BUT...
A. should this be checked by the linter?
B. in the age of git blame / praise, do we really need the author?  It tells us about the authorship of pre-git code though, so maybe useful.

CODING_STANDARD.md
```diff
@@ -61,12 +61,19 @@ Formatting is enforced using clang-format. For more information about this, see
 # Comments
 - Do not use `/* */`
 - Each source and header file must start with a comment block stating the
-  author. See existing source for an example of the format of this block. This
-  should be followed by a Doxygen `\file` comment:
+  author, followed by a Doxygen comment about the file:
```
peterschrammel, line 64: 💡 Doxygen has `\author`

peterschrammel (approved):

hannes-steffenhagen-diffblue:
@martin-cs The problem is that git commiter and author don’t necessarily preserve authorship as work is moved around (and also in the face of things like pair programming). Github looks at Co-Authored-By but I don’t believe this is widely used

@peterschrammel Should we perhaps just get rid of the module header comments entirely? Seems like we can encode everything that we want in the Doxygen already, and keeping two different header comment styles for every file seems like it induces unnecessary friction.

peterschrammel:
⛏️ IMO, we should get rid of the old headers and use doxygen ones.

thk123:
@peterschrammel I completely agree - I'm happy to update this PR to include this, but of course this won't change the code base (though this can of course be done incrementally). LMK one way or the other

Issue #5380, opened 2020-06-12
Name mangling changes break goto-cc


NlightNFotis:
Recent multi-file mangling changes have caused goto-cc to fail with an invariant violation.

We have traced the change that breaks it to https://github.com/diffblue/cbmc/commit/fcd1083c6ba318720607d85413cf2a7bafbf5521

CBMC version: `develop`
Operating system: `ubuntu 18.04.4 LTS`
Exact command line resulting in the issue: 
  * I am running this against jobs in https://github.com/awslabs/aws-c-common/, in particular `.cbmc_batch/jobs/linked_list_pop_front`
  * `make goto`

What behaviour did you expect: Compilation finishes without an issue.
What happened instead: Compilation crashed with an invariant violation:

```
--- begin invariant violation report ---
Invariant check failed
/home/fkoutoulakis/Devel/cbmc/src/goto-programs/name_mangler.h:127 function: mangle
Condition: inserted.second
Reason: The mangled name '__CPROVER_file_local_byteswap_h___bswap_32' should not already exist in the codebase
Backtrace:
goto-cc(print_backtrace(std::ostream&)+0x4d) [0x55a708004b98]
goto-cc(get_backtrace[abi:cxx11]()+0x45) [0x55a708004d9f]
goto-cc(std::enable_if<std::is_base_of<invariant_failedt, invariant_failedt>::value, void>::type invariant_violated_structured<invariant_failedt, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)+0x4e) [0x55a707d19a59]
goto-cc(+0x31e4bb) [0x55a707d194bb]
goto-cc(function_name_manglert<file_name_manglert>::mangle()+0xb61) [0x55a707d2fef5]
goto-cc(compilet::link()+0x2d1) [0x55a707d27451]
goto-cc(compilet::doit()+0x421) [0x55a707d259bd]
goto-cc(gcc_modet::doit()+0x2a1a) [0x55a707d69382]
goto-cc(goto_cc_modet::main(int, char const**)+0xaf) [0x55a707d7402f]
[...]

PR #5398, opened 2020-06-29
Add test for the name mangler regression in #5380.

NlightNFotis:
This PR is submitted as a preliminary change to the bug described in #5380. This adds a regression test that guards against it being introduced again.


regression/Makefile
```diff
@@ -28,6 +28,7 @@ DIRS = cbmc \
        systemc \
        contracts \
        goto-cc-file-local \
+       goto-cc-regression \
```
hannes-steffenhagen-diffblue: :pick: Everything here is “regression” tests, so the "regression" here doesn’t give us any information. Might be better to give this a more descriptive name – either what kind of testing it is doing if you think we’re likely to add different tests with the same test script in the future, or otherwise maybe just include the github issue ID as part of the name (e.g. `-gh-issue-XXXX`).

(Well technically most of them are integration tests, but we’re calling them regression tests regardless).

hannes-steffenhagen-diffblue (approved):
Mostly LGTM

regression/Makefile
NlightNFotis: Hm, you are right, I'm not perfectly happy with the `-regression` suffix, but my original one was `-crash` and I didn't think that was much better. 

TBH having it very issue specific is something that could easily be done, but for a strange kind of fashion I don't like it.

Let me ponder on that.

NlightNFotis: Also might be worth mentioning that originally I wanted the test to go into `goto-cc-file-local` but there was just no way I could bend the `chain.sh` to do exactly what I wanted it to do without causing issues.

danpoe (approved):

regression/Makefile
martin-cs: +1 what @hannes-steffenhagen-diffblue said.

( I think they are a mix of integration tests and regressions tests really but let's not get into that can of worms right now. )

regression/CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ add_subdirectory(contracts)
 add_subdirectory(goto-harness)
 add_subdirectory(goto-harness-multi-file-project)
 add_subdirectory(goto-cc-file-local)
+add_subdirectory(goto-cc-regression)
```
martin-cs: Is there really no way this can fit into any of the existing collections of tests?

regression/goto-cc-regression/static-name-clash/test.desc
```diff
@@ -0,0 +1,14 @@
...
+^SIGNAL=0$
+--
+^warning: ignoring
+^Invariant check failed$
+--
+This test guards against a regression added when the local file
+name mangler would not add the required name into the symbol
+table. When the regression was present, we were met with an
+invariant violation during goto-cc compilation. Without it,
+nothing spectacular happens - compilation proceeds as expected.
```
martin-cs: Any chance of some more detail or a link to the relevant issue on github?

martin-cs (approved):
Approve the revert, would like some changes to the test but those should be simple enough.

regression/CMakeLists.txt
NlightNFotis: Hi Martin, I spent some time this morning to make sure I can't get the test into another folder using the existing scaffolding.

I can't find a way myself, as this requires a very specific sequence in the `chain.sh`. Everyone of these folders have a specific `chain.sh` file that acts as scaffolding for that folder. Changing them causes issues which I can't fix easily. 

Hence, I think it's best to leave it as it is, unless you strongly object to that.

regression/goto-cc-regression/static-name-clash/test.desc
NlightNFotis: Yes, I have changed it to contain a link to the issue as well.

karkhaz:
Can I suggest #5402 instead @NlightNFotis? I believe the assertion is safe to remove.

NlightNFotis:
@karkhaz I merged your PR and repurposed this to remove the commit that backs out your previous change - so this is now just the test, as an enhancement to #5402. 

I have already merged your PR, and I will merge this to get in the test as well, as soon as it passes the CI. Thanks for the quick fix!

karkhaz:
@NlightNFotis sounds great, thank you!

PR #5121, opened 2019-09-20
Extremely minor code improvement

owen-mc-diffblue:
Only review the last commit. This is based on https://github.com/diffblue/cbmc/pull/5115, and should be merged after it.

- [ ] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #5401, opened 2020-06-30
Rework memory allocation failure checks and options

danpoe:

### DIFF
--- a/src/analyses/goto_check.h
+++ b/src/analyses/goto_check.h
@@ -39,7 +39,8 @@ void goto_check(
   "overflow-check)"                                                            \
   "(pointer-overflow-check)(conversion-check)(undefined-shift-check)"          \
   "(float-overflow-check)(nan-check)(no-built-in-assertions)"                  \
-  "(pointer-primitive-check)"
+  "(pointer-primitive-check)"                                                  \
+  "(overly-large-allocation-check)"
 
 // clang-format off
 #define HELP_GOTO_CHECK \
@@ -56,7 +57,10 @@ void goto_check(
   " --nan-check                  check floating-point for NaN\n" \
   " --no-built-in-assertions     ignore assertions in built-in library\n" \
   " --enum-range-check           checks that all enum type expressions have values in the enum range\n" /* NOLINT(whitespace/line_length) */ \
-  " --pointer-primitive-check    checks that all pointers in pointer primitives are valid or null\n" /* NOLINT(whitespace/line_length) */
+  " --pointer-primitive-check    checks that all pointers in pointer primitives are valid or null\n" /* NOLINT(whitespace/line_length) */ \
+  << help_entry( \
+    "--overly-large-allocation-check", \
+    "checks that no individual allocation exceeds CBMC's address space limit")
 
 #define PARSE_OPTIONS_GOTO_CHECK(cmdline, options) \
   options.set_option("bounds-check", cmdline.isset("bounds-check")); \
@@ -72,7 +76,8 @@ void goto_check(
   options.set_option("float-overflow-check", cmdline.isset("float-overflow-check")); /* NOLINT(whitespace/line_length) */  \
   options.set_option("nan-check", cmdline.isset("nan-check")); \
   options.set_option("built-in-assertions", !cmdline.isset("no-built-in-assertions")); /* NOLINT(whitespace/line_length) */ \
-  options.set_option("pointer-primitive-check", cmdline.isset("pointer-primitive-check")) /* NOLINT(whitespace/line_length) */
+  options.set_option("pointer-primitive-check", cmdline.isset("pointer-primitive-check")); /* NOLINT(whitespace/line_length) */ \
+  options.set_option("overly-large-allocation-check", cmdline.isset("overly-large-allocation-check")) /* NOLINT(whitespace/line_length) */
 // clang-format on
 
 #endif // CPROVER_ANALYSES_GOTO_CHECK_H
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -1067,7 +1067,7 @@ void cbmc_parse_optionst::help()
     HELP_SHOW_GOTO_FUNCTIONS
     "\n"
     "Program instrumentation options:\n"
-    HELP_GOTO_CHECK
+    << HELP_GOTO_CHECK <<
     " --no-assertions              ignore user assertions\n"
     " --no-assumptions             ignore user assumptions\n"
     " --error-label label          check that label is unreachable\n"
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -1073,11 +1073,12 @@ void cbmc_parse_optionst::help()
     " --error-label label          check that label is unreachable\n"
     " --cover CC                   create test-suite with coverage criterion CC\n" // NOLINT(*)
     " --mm MM                      memory consistency model for concurrent programs\n" // NOLINT(*)
-    // NOLINTNEXTLINE(whitespace/line_length)
-    " --malloc-fail-assert         set malloc failure mode to assert-then-assume\n"
-    " --malloc-fail-null           set malloc failure mode to return null\n"
-    // NOLINTNEXTLINE(whitespace/line_length)
-    " --malloc-may-fail            allow malloc calls to return a null pointer\n"
+    << help_entry(
+      "--overly-large-allocation-returns-null",
+      "memory allocation functions return null when more memory is allocated than can be addressed by cbmc") // NOLINT(*)
+    << help_entry(
+      "--allocation-may-fail",
+      "allow memory allocation functions to return null") <<
     HELP_REACHABILITY_SLICER
     HELP_REACHABILITY_SLICER_FB
     " --full-slice                 run full slicer (experimental)\n" // NOLINT(*)
--- a/regression/cbmc/Pointer_byte_extract5/no-simplify.desc
+++ b/regression/cbmc/Pointer_byte_extract5/no-simplify.desc
@@ -4,7 +4,7 @@ main.i
 ^EXIT=10$
 ^SIGNAL=0$
 array\.List dynamic object upper bound in p->List\[2\]: FAILURE
-\*\* 1 of 16 failed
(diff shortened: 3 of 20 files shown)


src/analyses/goto_check.h
```diff
@@ -58,7 +58,7 @@ void goto_check(
   " --no-built-in-assertions     ignore assertions in built-in library\n" \
   " --enum-range-check           checks that all enum type expressions have values in the enum range\n" /* NOLINT(whitespace/line_length) */ \
   " --pointer-primitive-check    checks that all pointers in pointer primitives are valid or null\n" /* NOLINT(whitespace/line_length) */ \
-  " --allocate-size-check        checks that not more memory is allocated than can be addressed by cbmc" /* NOLINT(whitespace/line_length) */
+  << help_entry("--allocate-size-check", "checks that not more memory is allocated than can be addressed by cbmc") /* NOLINT(whitespace/line_length) */
```
NlightNFotis: What is the `help_entry` function. Where was it introduced? Is there a CBMC equivalent?

NlightNFotis (approved):
LGTM so far, only one question

src/analyses/goto_check.h
danpoe: It's defined in `parse_options.cpp`. I've introduced it a few weeks ago. It's supposed to make writing the help entries easier, by automatically wrapping the description at 80 characters, and indenting it by 30 characters.

martin-cs (commented):
I believe this works and is a good implementation of the idea but I don't see why it is needed.  To my mind, having the modelling of malloc, including the checks in the library makes way more sense.  Why is this PR an improvement?

danpoe:
@martin-cs: The PR essentially does two things:

1. It fixes the issue raised by @hannes-steffenhagen-diffblue's on the previous PR (see #5212)
2. The checking of the allocated size (flag `--allocate-size-check`) is now done for `__CPROVER_allocate()`, which is used by the models. This means the models are simpler, the check is only present when the flag is given, and it will also work for plain `__CPROVER_allocate()` statements. That this is implemented in `goto_checkt` is consistent with how other primitives like `__CPROVER_r_ok`, etc. are handled.

src/analyses/goto_check.h
```diff
@@ -56,7 +57,8 @@ void goto_check(
   " --nan-check                  check floating-point for NaN\n" \
   " --no-built-in-assertions     ignore assertions in built-in library\n" \
   " --enum-range-check           checks that all enum type expressions have values in the enum range\n" /* NOLINT(whitespace/line_length) */ \
-  " --pointer-primitive-check    checks that all pointers in pointer primitives are valid or null\n" /* NOLINT(whitespace/line_length) */
+  " --pointer-primitive-check    checks that all pointers in pointer primitives are valid or null\n" /* NOLINT(whitespace/line_length) */ \
+  " --allocate-size-check        checks that not more memory is allocated than can be addressed by cbmc" /* NOLINT(whitespace/line_length) */
```
smowton: checks that no individual allocation exceeds CBMC's addresss space limit? After all total alloc can still be very large

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1073,11 +1073,12 @@ void cbmc_parse_optionst::help()
     " --error-label label          check that label is unreachable\n"
     " --cover CC                   create test-suite with coverage criterion CC\n" // NOLINT(*)
     " --mm MM                      memory consistency model for concurrent programs\n" // NOLINT(*)
-    // NOLINTNEXTLINE(whitespace/line_length)
-    " --malloc-fail-assert         set malloc failure mode to assert-then-assume\n"
-    " --malloc-fail-null           set malloc failure mode to return null\n"
-    // NOLINTNEXTLINE(whitespace/line_length)
-    " --malloc-may-fail            allow malloc calls to return a null pointer\n"
+    << help_entry(
+      "--allocate-size-null",
```
smowton: overly-large-allocation-returns-null? The other option would then be overly-large-allocation-asserts

smowton (approved):
You should reword the PR, this isn't a refactor, which I would expect to change no behaviour. Approving to remove my codeowner block (I don't mind either way), but you should certainly win the argument with @martin-cs before merging :)

martin-cs:
Is #5121 the right issue?  I don't see the link here.

I think the model way of checking these is better because it is more flexible.  It can be turned on and off during analysis, the limits can be varied, etc.  But I appreciate that this is just my opinion.  More importantly than where it is, having all of the modelling / instrumentation for memory allocation in one place is a good thing IMHO.

danpoe:
Ah, it's #5212. I think it depends on whether one views this as a modelling issue or not. The flag `--allocate-size-check` checks that an implementation limit of cbmc isn't hit. The most straightforward way of doing this is to instrument a precondition before every `__CPROVER_allocate()` invocation. On the other hand, whether `malloc()` can nondeterministically return null is more of a modelling question, and the code for this is still part of the model.

martin-cs:
I appreciate the distinction and it is a useful thing to note.

https://github.com/diffblue/cbmc/pull/5210 looks more relevant to me.

Sorry to be awkward but this feels like something that needs a bit more discussion.  If this is checking an internal limit of the 'base + offset' pointer model so that we don't generate spurious models then...

A. Why should it be an option?  Why would you want it turned off?
B. Wouldn't 'assert then assume' be the right way to handle this because we can't reasonably proceed with an allocation that is too large.
C. Doesn't this also affect global (or even stack allocated!) objects above a certain size?

danpoe:
Yes, #5210 is the first of the two related PRs. The issue though that this one fixes was pointed out by @hannes-steffenhagen-diffblue on #5212.

A. If you know the limit will not be hit you don't need to turn on the option (or if you're using cbmc for bug finding). Currently cbmc adds no checks by default, and all of them need to be explicitely enabled (e.g., `--pointer-check`, etc.)
B. Not necessarily. It's done like that for all `goto_checkt` checks, e.g. array bounds checks, even though one could question how much sense it makes to continue to execute after an out of bounds access.
C. Yes, I think it also affects those.

There's also other cbmc primitives that have preconditions for which checks need to be explicitely enabled (see http://cprover.diffblue.com/memory-primitives.html).

src/analyses/goto_check.h
danpoe: Done

src/cbmc/cbmc_parse_options.cpp
danpoe: Done

chrisr-diffblue (commented):
Looks generally good to me, but I have two requests before I'm happy to approve :-)
1) Get an 'OK' from @martin-cs 
2) Update the user docs - I think some of the removed command line options were possibly documented, if so, they'll need removing from the docs too, plus add user docs for the new options too.

danpoe:
@martin-cs Would you be ok with this going in for now? I agree it's not a perfect or complete solution, it just would help us as it fixes an issue for us.

martin-cs:
@danpoe Sorry for the lag.  I think it is possible to sort this out but let's just check we are talking about the same thing.  This program shows the problem:

```
#define BIG_NUMBER   // something bigger than 1 << (sizeof(void*)*CHAR_BITS - (--object-bits n))

char big_array[BIG_NUMBER];
char other_array[256];

  char *b = big_array;
  char *c = other_array;

  *(b + (BIG_NUMBER - 1) = '\0';  // Should modify big_array not other_array

  char *p = malloc(BIG_NUMBER);
  char *q = malloc(256);

  *(p + (BIG_NUMBER - 1)) = '\0'; // Should modify the first allocated block, not the second one

  return 0;
}
```

So the root cause of the problem is the split between object and offset in the representation of pointers in the default back-end?  If the offset is too high then it can overflow into the object, resulting in incorrect answers.

The patch you are proposing will add a command-line option to CBMC which will insert assertions everywhere __CPROVER_alloc is used.  When the flag is used it will indicate when the program is able to dynamically allocate a block of memory that /might/ have this problem.

Have I understood the problem?

1. Ideally we need to solve the actual root of the problem which is the overflow (or underflow) from offset to object.  The back-end effectively already tracks when a pointer is becoming invalid as it is one of the overflow bits in the adder when modifying the offset.  It would take a little thought and care to expose this to the front-end / instrumentation and you would need to decide if the problem was A. creating an invalid pointer or using an invalid pointer, B. whether making the pointer valid again is possible.

2. A quicker fix would be to prevent objects that are too large being created assuming that bounds checks are used.  This doesn't actually solve the problem as offset overflow can modify a pointer so that it is still in-bounds for another object.  Regardless, this should be possible by checking global and local variables and putting assert/assume before calls to __CPROVER_alloc.  As it is an internal function and only used in a few place I think this makes more sense to do in the modelled C library.  As long as --object-bits is converted into a global variable this should be a trivial patch.

3. I think there is an important question about what the default should be.  I can't see a reason not to have this on by default.  Ordered compare with constants, particularly powers of 2, are about as cheap as you can get.  If it is on by default then in the model makes more sense to me.

HTH

danpoe:
Yes, I agree with your analysis, the allocation of a block for which not all parts can be addressed (in the realm of cbmc) is not in itself a problem. So yes the best fix would probably be to catch the pointer overflows instead.

It turns out that `--pointer-overflow-check` only checks for overflows of the complete bitvector, as opposed to overflows of the offset part only. The offset part does not overflow into the the object bits part though but wraps around.

Btw, if the result of pointer arithmetic points outside the pointed-to memory block, the behavior is undefined (except pointing to one past the last element which is allowed). See, e.g., here: https://blog.regehr.org/archives/1395.

So a possible solution would be:

- add an always on check which verifies that the offset part of a pointer does not overflow
  
- modify `--pointer-overflow-check` such that it verifies that the result of pointer arithmetic does not point outside the pointed-to object

martin-cs:
I like the proposed course of action.  Thanks for giving this the extra consideration.  Limitations in how we handle pointers should not conceal bugs, so I like the always on check for offset overflow.  I can't see what the use of checking the overflow of a whole pointer is, so modifying `--pointer-overflow-check` sounds good to me.  The proposed semantics seem good to me because it catches an extra kind of UB.  I think having it as an option still makes sense because I can see cases (sort of like 1.B. above) where people assume that any kind of pointer arithmetic is fine as long as they only access in the object.

martin-cs:
@danpoe Might your suggestion also address https://github.com/diffblue/cbmc/issues/5096 ?

PR #5402, opened 2020-06-30
Remove multiple function assertion

### LINKED ISSUE #5380: Name mangling changes break goto-cc

karkhaz:
This commit removes an assertion that triggers when multiple functions
get mangled to the same name when using the --export-file-local-symbols
flag to goto-cc.

This assertion was previously valid when compiling files one at a time, but
commit fcd1083 introduced the possibility of compiling and linking several
files at once. This is a problem when two source files both include an
implementation file, like [1], because the implementation may contain static
functions that get mangled twice.

This commit replaces the assertion with a debug message in case users would
like to check how often this happens.

This commit fixes #5380.

[1] https://github.com/awslabs/aws-c-common/blob/master/include/aws/common/atomics_gnu.inl

### DIFF
--- a/src/goto-programs/name_mangler.h
+++ b/src/goto-programs/name_mangler.h
@@ -121,10 +121,9 @@ public:
 
       auto inserted = model.goto_functions.function_map.emplace(
         pair.second, std::move(found->second));
-      INVARIANT(
-        inserted.second,
-        "The mangled name '" + std::string(pair.second.c_str()) +
-          "' should not already exist in the codebase");
+      if(!inserted.second)
+        log.debug() << "Found a mangled name that already exists: "
+                    << std::string(pair.second.c_str()) << log.eom;
 
       model.goto_functions.function_map.erase(found);
     }


martin-cs (approved):
As I understand emplace, the log will only be triggered if you have two things that mangle to the same name and are identical, right?  If so and we have a valid use-case where this can happen (and it seems that https://github.com/diffblue/cbmc/pull/5398 has one) then this should be fine.

karkhaz:
@martin-cs that's all correct. Thank you!

smowton (commented):
Add a test?

NlightNFotis (approved):
LGTM

NlightNFotis:
@karkhaz I think that's alright. I'd rather keep the test from the previous PR however, but now I can remove the backing out of the commit.

NlightNFotis:
@smowton There's a test in #5398 that's relevant I believe.

smowton (approved):
ok

smowton:
Hang on, surely in that case that test should change from broken to working or similar? Can we add a test that would previously trip the invariant?

NlightNFotis:
@smowton The test is working because the #5398 is removing the commit that caused the issue in the first place. Assuming this goes in, the test should be passing.

If the test gets merged first, without first backing out the previous commit or just removing the assertion like in here, then it should fail. But the test is associated, it's just the merging sequence that should matter as to whether it passes or fails.

NlightNFotis:
I am going ahead and merging this one. For any issues with this, ping me.

PR #5404, opened 2020-07-01
Add an abstract interpreter 'history' that gives (local) per-path analysis and / or loop unwinding

martin-cs:
In the past users have had to modify the program before analysis to get this kind of precision.  Doing so is expensive in terms of memory for both the program and the analysis.  This gives a more efficient route.  The is an equivalent of an unwinding limit that controls the number of histories that can be generated for each location.  This is a better measure as it should scale memory linearly and time linearly(ish) while loop unwinding can have seriously non-linear effects when there are nested loops.

This is the enabling technology for a "loop unwinding" analysis that can automatically over-approximate loop bounds.  It is also probably the last "core" change to the AI infrastructure before VSD can be merged in full.

### DIFF
--- a/src/analyses/ai.cpp
+++ b/src/analyses/ai.cpp
@@ -303,8 +303,14 @@ bool ai_baset::visit(
       if(to_l == goto_program.instructions.end())
         continue;
 
-      new_data |=
-        visit_edge(function_id, p, function_id, to_l, ns, working_set);
+      new_data |= visit_edge(
+        function_id,
+        p,
+        function_id,
+        to_l,
+        ai_history_baset::no_caller_history,
+        ns,
+        working_set); // Local steps so no caller history needed
     }
   }
 
@@ -316,11 +322,13 @@ bool ai_baset::visit_edge(
   trace_ptrt p,
   const irep_idt &to_function_id,
   locationt to_l,
+  trace_ptrt caller_history,
   const namespacet &ns,
   working_sett &working_set)
 {
   // Has history taught us not to step here...
-  auto next = p->step(to_l, *(storage->abstract_traces_before(to_l)));
+  auto next =
+    p->step(to_l, *(storage->abstract_traces_before(to_l)), caller_history);
   if(next.first == ai_history_baset::step_statust::BLOCKED)
     return false;
   trace_ptrt to_p = next.second;
@@ -366,6 +374,8 @@ bool ai_baset::visit_edge_function_call(
     p_call,
     calling_function_id,
     l_return,
+    ai_history_baset::
+      no_caller_history, // Not needed as we are skipping the function call
     ns,
     working_set);
 }
@@ -439,6 +449,7 @@ bool ai_baset::visit_function_call(
     p_call,
     calling_function_id,
     l_return,
+    ai_history_baset::no_caller_history, // Would be the same as p_call...
     ns,
     working_set);
 }
@@ -480,6 +491,8 @@ bool ai_recursive_interproceduralt::visit_edge_function_call(
       p_call,
       callee_function_id,
       l_begin,
+      ai_history_baset::
+        no_caller_history, // Not needed as p_call already has the info
       ns,
       catch_working_set);
 
@@ -525,6 +538,7 @@ bool ai_recursive_interproceduralt::visit_edge_function_call(
           p_end,
           calling_function_id,
           l_return,
+          p_call, // To allow function-local history
           ns,
           working_set);
       }
--- a/src/analyses/ai.h
+++ b/src/analyses/ai.h
@@ -252,7 +252,11 @@ public:
 
     locationt n = std::next(l);
 
-    auto step_return = p->step(n, *(storage->abstract_traces_before(n)));
+    auto step_return = p->step(
+      n,
+      *(storage->abstract_traces_before(n)),
+      ai_history_baset::no_caller_history);
+    // Caller history not needed as this is a local step
 
     return storage->abstract_state_before(step_return.second, *domain_factory);
   }
@@ -469,6 +473,7 @@ protected:
     trace_ptrt p,
     const irep_idt &to_function_id,
     locationt to_l,
+    trace_ptrt caller_history,
     const namespacet &ns,
     working_sett &working_set);
 
--- a/src/analyses/call_stack_history.cpp
+++ b/src/analyses/call_stack_history.cpp
@@ -11,8 +11,10 @@ Author: Martin Brain, martin.brain@cs.ox.ac.uk
 
 #include "call_stack_history.h"
 
-ai_history_baset::step_returnt
-call_stack_historyt::step(locationt to, const trace_sett &others) const
+ai_history_baset::step_returnt call_stack_historyt::step(
+  locationt to,
+  const trace_sett &others,
+  trace_ptrt caller_hist) const
 {
   DATA_INVARIANT(current_stack != nullptr, "current_stack must exist");
 
@@ -79,12 +81,29 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
     }
     else
     {
+      INVARIANT(
+        caller_hist != ai_history_baset::no_caller_history,
+        "return from function should have a caller");
+
       // The expected call return site...
       locationt l_caller_return =
         std::next(current_stack->caller->current_location);
 
       if(l_caller_return->location_number == to->location_number)
       {
+        INVARIANT(
+          std::next(caller_hist->current_location())->location_number ==
+            l_caller_return->location_number,
+          "caller and caller_hist should be consistent");
+        // It is tempting to think that...
+        // INVARIANT(*(current_stack->caller) ==
+        //           *(std::dynamic_pointer_cast<const call_stack_historyt>
+        //                                       (caller_hist)->current_stack),
+        //           "call stacks should match");
+        // and that we could just use caller_hist->current_stack here.
+        // This fails when you hit the recursion limit so that the
+        // caller_hist has a deep stack but *this has a shallow one.
+
         // ... which is where we are going
         next_stack = cse_ptrt(std::make_shared<call_stack_entryt>(
           to, current_stack->caller->caller));
--- a/src/analyses/local_control_flow_history.cpp
+++ b/src/analyses/local_control_flow_history.cpp
@@ -1,0 +1,382 @@
+/*******************************************************************\
+
+Module: Abstract Interpretation
+
+Author: Martin Brain, martin.brain@cs.ox.ac.uk
+
+\*******************************************************************/
+
+/// \file
+/// Track function-local control flow for loop unwinding and path senstivity
+
+#include "local_control_flow_history.h"
+
+template <bool track_forward_jumps, bool track_backward_jumps>
+ai_history_baset::step_returnt
+local_control_flow_historyt<track_forward_jumps, track_backward_jumps>::step(
+  locationt to,
+  const trace_sett &others,
+  trace_ptrt caller_hist) const
+{
+  // First compute what the new history could be,
+  // then find it in others or possibly merge it if the limit is hit
+
+  std::shared_ptr<const local_control_flow_historyt<
+    track_forward_jumps,
+    track_backward_jumps>>
+    next_step = nullptr;
+
+  // Local branching is the key step
+  if(
+    current_loc->is_goto() &&
+    (current_loc->is_backwards_goto() ? track_backward_jumps
+                                      : track_forward_jumps))
+  {
+    bool branch_taken = (current_loc->get_target() == to) &&
+                        !(current_loc->get_target() == std::next(current_loc));
+    // Slight oddity -- we regard branches to the next instruction as not taken
+    // regardless of their guard condition.  This is slightly more general than
+    // is_skip() but without changing the step() interface to have a "taken"
+    // flag we will have to assume taken or not taken and not taken seems safer.
+
+    auto decision = std::make_shared<local_control_flow_decisiont>(
+      current_loc, branch_taken, control_flow_decision_history);
+    next_step = std::make_shared<
+      local_control_flow_historyt<track_forward_jumps, track_backward_jumps>>(
+      to, decision, max_histories_per_location);
+  }
+  else if(
+    current_loc->is_function_call() &&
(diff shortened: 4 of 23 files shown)


martin-cs:
Please can someone with access to the AWS CodeBuild outputs let me know what the problem is?  Last time it was @chrisr-diffblue or @thk123 who could do this...

chrisr-diffblue:
Looks like all of the failed AWS codebuild jobs failed really early due to infrastructure problems (failed to apt-get install dependencies). I've restarted all those failed jobs.

martin-cs:
Thanks @chrisr-diffblue  they are listed as "failed" rather than "pending" for me at the moment.  I'll give them a minute...

martin-cs:
These still look failed to me, plus the code coverage numbers are kinda wacky.

chrisr-diffblue:
Unfortunately, it looks like although the restarted jobs seem to have passed, they haven't posted status back to github, sigh. You might have to go for the rebase/dummy commit + force push to make github start new jobs.

smowton:
@chrisr-diffblue IIRC updating Github status after a rerun is a Codebuild option, @rjmunro knows more

rjmunro:
@smowton @chrisr-diffblue Yes, there is an option "Report build statuses to source provider when your builds start and finish" in the "Edit" -> "Source" menu, hidden in "Additional configuration (Git clone depth, Git submodules)" (even though it is neither of those things). Without this, codebuild will report status to github only where the build was triggered by a github webhook. I can see it's not turned on, but I'd rather not change the setting myself as I'm not involved in this project. I'm happy to do it if someone officially asks me to.

Finding that setting was a huge win for us because it let us trigger codebuild jobs from other codebuild jobs, rather than from webhooks. This meant we could build a CI pipeline.

thk123:
@rjmunro I'm happy to do it - any  reason to _not_ have this on? Seems like a no-brainer...

rjmunro:
@thk123 The guess the main reason not to have it on would be if you were testing your codebuild config and didn't want failures to block PRs or something, but in general, I would turn it on.

martin-cs:
Thanks for the help all but ... I still can't see what is actually causing the failures.  I'd really like to get this PR moving and I'm keen to fix the issues but without the info...

thk123:
@martin-cs as Chris said - the failures seem spurious - you should just be able to rebase this and push to retrigger the builds. I'll try and get the updated status thing working today.

martin-cs:
Thank you @thk123 that seems to have done the trick.

Do any of the maintainers have time for a review of this PR?

src/analyses/ai.cpp
```diff
@@ -366,6 +368,7 @@ bool ai_baset::visit_edge_function_call(
     p_call,
     calling_function_id,
     l_return,
+    nullptr, // Not needed as we are skipping the function call
```
smowton: This and other `nullptr` args are pretty opaque -- how about `#define CALLER_UNSPECIFIED nullptr`?

src/analyses/ai.cpp
```diff
@@ -480,6 +484,7 @@ bool ai_recursive_interproceduralt::visit_edge_function_call(
       p_call,
       callee_function_id,
       l_begin,
+      nullptr,
```
smowton: Why is this one not given?

src/analyses/ai.h
```diff
@@ -252,7 +252,8 @@ class ai_baset
 
     locationt n = std::next(l);
 
-    auto step_return = p->step(n, *(storage->abstract_traces_before(n)));
+    auto step_return =
+      p->step(n, *(storage->abstract_traces_before(n)), nullptr);
```
smowton: Does the nullptr here imply we're conflating some states that we could potentially distinguish?

src/analyses/local_control_flow_history.h
```diff
@@ -0,0 +1,203 @@
...
+  std::shared_ptr<const local_control_flow_historyt<
+    track_forward_jumps,
+    track_backward_jumps>>
+  down_cast(trace_ptrt in) const
+  {
+    PRECONDITION(in != nullptr);
+    auto tmp = std::dynamic_pointer_cast<const local_control_flow_historyt<
+      track_forward_jumps,
+      track_backward_jumps>>(in);
+    PRECONDITION(tmp != nullptr);
```
smowton, line 65: POSTCONDITION?

src/analyses/local_control_flow_history.h
```diff
@@ -0,0 +1,203 @@
...
+  locationt current_loc;
+  lcfd_ptrt control_flow_decision_history;
+  size_t max_histories_per_location;
+
+  // A repeating idiom -- can cast from interface type to this type
+  // caller's responsibility to ensure correct types
+  std::shared_ptr<const local_control_flow_historyt<
+    track_forward_jumps,
+    track_backward_jumps>>
+  down_cast(trace_ptrt in) const
```
smowton: rename `to_local_control_flow_history`?

src/analyses/local_control_flow_history.cpp
```diff
@@ -0,0 +1,380 @@
...
+    {
+      out << " not taken";
+    }
+    working = working->previous;
+  }
+  return;
+}
+
+bool local_control_flow_decisiont::
+operator<(const local_control_flow_decisiont &op) const
```
smowton, line 272: Looks like you intentionally want more informative orderings to be lower in the sort order? If so say so up top

src/analyses/local_control_flow_history.cpp
```diff
@@ -0,0 +1,380 @@
...
+        {
+          return *(this->previous) == *(op.previous);
+        }
+      }
+    }
+  }
+  return false;
+}
+
+// Instantiate the versions we need
```
smowton, line 359: Being all possible versions :)

src/goto-analyzer/goto_analyzer_parse_options.cpp
```diff
@@ -295,6 +296,34 @@ void goto_analyzer_parse_optionst::get_command_line_options(optionst &options)
         "call-stack-recursion-limit", cmdline.get_value("call-stack"));
       options.set_option("history set", true);
     }
+    else if(cmdline.isset("loop-unwind"))
+    {
+      options.set_option("local-control-flow-history", true);
```
smowton, line 301: Considering there's some trickiness here, perhaps print a line describing the options you've inferred from the higher-level options provided

src/goto-analyzer/goto_analyzer_parse_options.cpp
```diff
@@ -295,6 +296,34 @@ void goto_analyzer_parse_optionst::get_command_line_options(optionst &options)
...
+      options.set_option("local-control-flow-history-forward", false);
+      options.set_option("local-control-flow-history-backward", true);
+      options.set_option(
+        "local-control-flow-history-limit", cmdline.get_value("loop-unwind"));
+      options.set_option("history set", true);
+    }
+    else if(cmdline.isset("branching"))
```
smowton, line 308: The else-ifs here imply `--loop-unwind --branching` won't work as expected, right? If so check for this case.

smowton (approved):
Mostly lgtm, some requests for clarity, particularly clarifying either by naming a pass-through function or using a named `nullptr` when histories are being omitted because they're irrelevant, and when they're being omitted due to a strategic choice to ignore history and merge all states.

src/analyses/ai.cpp
hannes-steffenhagen-diffblue: I don't really agree with using a define for this. How about `constexpr whatever* UnspecifiedCaller = nullptr;`?

src/analyses/ai.cpp
```diff
@@ -316,11 +316,13 @@ bool ai_baset::visit_edge(
   trace_ptrt p,
   const irep_idt &to_function_id,
   locationt to_l,
+  trace_ptrt caller_history,
```
hannes-steffenhagen-diffblue, line 325: I do have a pretty strong preference for not using `nullptr` to indicate optional arguments – this isn't obvious from the signature, and frequently makes the logic inside more complicated if you need to disambiguate between pointers you're allowed to dereference and ones you aren't...

Also, are these `trace_ptrts` all `shared_ptr`? In that case it'd seem a bit wasteful to me,  unless these are stored somewhere (I can't quite see it, maybe somewhere further down in the call stack)?

src/analyses/ai.cpp
martin-cs: Added to ai_historyt as a member variable as @hannes-steffenhagen-diffblue suggests.

src/analyses/ai.cpp
martin-cs: Because it is only used on function call return edges, this is a call, so `p_call` already has the caller info.

src/analyses/ai.h
martin-cs: Good question but in this case no.  The step is local so the history is not needed.  I have added a comment to this effect.

( There is a separate conversation about whether such an interface is even meaningful w.r.t. function calls and goto instructions. )

src/analyses/local_control_flow_history.h
martin-cs: I think it is a precondition of `down_cast` that this conversion works.  To quote the comment above `caller's responsibility to ensure correct types`.

src/analyses/local_control_flow_history.h
martin-cs: Done.

src/analyses/local_control_flow_history.cpp
martin-cs: Ummm... I am not entirely sure I have an idea of what constitutes informative in this context.  Where the choice of ordering matters (both in this and in `local_control_flow_historyt<track_forward_jumps, track_backward_jumps>::
operator<`) it is commented and in one case checked by assertion (if there is a merge history it should sort last).  Specifically this includes:
```
// Lower locations first, generally this means depth first
// Branch not taken takes priority
// This prioritises complete histories over merged ones
```

src/analyses/local_control_flow_history.cpp
martin-cs: That too!  (although now I think about it maybe the version that doesn't track either isn't actually used... )

src/goto-analyzer/goto_analyzer_parse_options.cpp
martin-cs: The options that are set are internal only and are intended to make the handling of these options simpler.  I am reluctant to print anything here because the user can't set any of these options directly.

src/goto-analyzer/goto_analyzer_parse_options.cpp
martin-cs: That is correct and in part why the combined option is known as `--loop-unwind-and-branching`.  I am reluctant to address this issue in this PR because it affects almost all options to goto-analyze`, all of the major options have an implicit ordering and check for absence but not duplication.

src/analyses/ai.cpp
martin-cs: I have renamed as you suggested earlier.  I hope the invariants in `ai_history.h` show when the optional argument is present.  `step` kind of has to be a case switch because of the different types of instruction.

Yes, they are all `shared_ptr`s.  They are stored in something that implements the `ai_storaget` interface, allowing you to store states and histories with varying degrees of sensitivity (per history, per location, soon per function and just once).  It's not great but I had a solution with `unique_ptr` but it involved way too many copies.

src/analyses/ai.cpp
martin-cs: ( Unfortunately `constexpr` doesn't actually work because `ai_history_baset` has a virtual destructor but a `static const` variable does. )

martin-cs:
@smowton and @hannes-steffenhagen-diffblue I have responded to and (mostly) implemented your review comments.  They are two separate commits at the moment for simplicity.  If you are happy I will squash them into the relevant earlier commits.

martin-cs:
@smowton and @hannes-steffenhagen-diffblue sorry to hassle but please can you have a look at the changes I have made and see if they address the issues you raised?

smowton:
I stand by my green tick :) (meaning I'm happy with this as-is but have suggestions)

smowton:
Apologies for slowness btw, was on holiday the past week

martin-cs:
@smowton Thanks!  Hope you had a nice holiday and sorry for hassling.

@hannes-steffenhagen-diffblue @danpoe @peterschrammel @chrisr-diffblue Any chance you could help with a second green tick?  This is kind of blocking for me and the next PR is something that might make your lives easier...

src/analyses/call_stack_history.cpp
```diff
@@ -79,12 +81,28 @@ call_stack_historyt::step(locationt to, const trace_sett &others) const
     }
     else
     {
+      INVARIANT(
+        caller_hist != nullptr, "return from function should have a caller");
```
chrisr-diffblue: Can you use `ai_history_baset::no_caller_history` rather in this INVARIANT, rather than `nullprt`?

src/analyses/local_control_flow_history.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+      return std::make_pair(
+        ai_history_baset::step_statust::MERGED, *(others.begin()));
+    }
+  }
+  else if(current_loc->is_end_function())
+  {
+    // This is slightly complicated as we have to drop the current,
+    // callee history and pick up the caller history.
+    PRECONDITION(caller_hist != nullptr);
```
chrisr-diffblue: Ditto for the `nullptr` here too

chrisr-diffblue (approved):
A couple of extremely minor nitpicks, but nothing blocking - apologies for the massive delay in me getting around to reviewing....

src/analyses/call_stack_history.cpp
martin-cs: Done.

src/analyses/local_control_flow_history.cpp
martin-cs: Done.

martin-cs:
I have addressed @chrisr-diffblue comments ( thanks for the review and no worries about the lag ) and squashed the changes in.  I will merge once CI has passed, @hannes-steffenhagen-diffblue shout now if you aren't happy with how your comments have been resolved.

PR #5409, opened 2020-07-03
Do not use side_effect_expr_nondett in recursive_initialization

hannes-steffenhagen-diffblue:
If we use side_effect_expr_nondett then dump_c will generation function calls
to functions with names like `nondet_int` without declaration.

This is a problem because it’ll lead the C frontend to complain about
undeclared identifiers and such.

To fix this, we changed goto-harness to just not use side_effect_expr_nondett
at all (better long term fix would be to make sure `dump_c` outputs
declarations too).

### DIFF
--- a/regression/goto-harness/do-not-use-nondet-for-primitives/test.desc
+++ b/regression/goto-harness/do-not-use-nondet-for-primitives/test.desc
@@ -1,0 +1,17 @@
+CORE
+test.c
+--function test_function --harness-type call-function
+\[test_function\.assertion\.1\] line \d+ assertion test != 42
+^EXIT=10$
+^SIGNAL=0$
+--
+nondet_signed_int
+--
+If we use side_effect_expr_nondett then dump_c will generation function calls
+to functions with names like nondet_signed_int without declaration.
+This can cause problems around usability.
+
+To fix this, we changed goto-harness to just not use side_effect_expr_nondett
+at all.
+
+This test tests for the absence the use for these things.
--- a/src/goto-harness/recursive_initialization.cpp
+++ b/src/goto-harness/recursive_initialization.cpp
@@ -169,11 +169,11 @@ void recursive_initializationt::initialize(
             goto_model.symbol_table.lookup_ref(fun_name);
           const auto proper_init_case = code_function_callt{
             fun_symbol.symbol_expr(), {depth, address_of_exprt{lhs}}};
-
+          const auto should_make_equal =
+            get_fresh_local_typed_symexpr("should_make_equal", bool_typet{});
+          body.add(code_declt{should_make_equal});
           body.add(code_ifthenelset{
-            side_effect_expr_nondett{bool_typet{}, source_locationt{}},
-            set_equal_case,
-            proper_init_case});
+            should_make_equal, set_equal_case, proper_init_case});
         }
         else
         {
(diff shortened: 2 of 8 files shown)


regression/goto-harness/do-not-use-nondet-for-primitives/test.desc
```diff
@@ -0,0 +1,17 @@
+CORE
+test.c
+--function test_function --harness-type call-function
+\[test_function\.assertion\.1\] line \d+ assertion test != 42
+^EXIT=10$
```
thk123, line 5: :pick: It'd be good to have a positive regex for what it now contains (e.g. a declaration or whatever) since that documents the behaviour, and makes the test less likely to spuriously pass in the future

regression/goto-harness/do-not-use-nondet-for-primitives/test.desc
```diff
@@ -0,0 +1,17 @@
...
+nondet_signed_int
+--
+If we use side_effect_expr_nondett then dump_c will generation function calls
+to functions with names like nondet_signed_int without declaration.
+This can cause problems around usability.
+
+To fix this, we changed goto-harness to just not use side_effect_expr_nondett
+at all.
+
+This test tests for the absence the use for these things.
```
thk123, line 17: :pick: This seems like great info for a commit message (since it is talking about why we moved from one solution to another) but in a test, where the reader won't have the context that we used to use `side_effect_expr_nondett` it is maybe confusing. I'd focus instead on what the test is testing - in this case not using undeclared functions.

src/goto-harness/recursive_initialization.cpp
```diff
@@ -169,11 +169,11 @@ void recursive_initializationt::initialize(
             goto_model.symbol_table.lookup_ref(fun_name);
           const auto proper_init_case = code_function_callt{
             fun_symbol.symbol_expr(), {depth, address_of_exprt{lhs}}};
-
+          const auto should_make_equal =
```
thk123, line 172: :no_entry_sign: Looks like this is a separate case and therefore should have a separate test? :safety_pin:

src/goto-harness/recursive_initialization.cpp
```diff
@@ -693,9 +693,12 @@ code_blockt recursive_initializationt::build_pointer_constructor(
   code_blockt null_and_return{{assign_null, code_returnt{}}};
   body.add(code_ifthenelset{conjunction(should_not_recurse), null_and_return});
 
+  const auto should_recurse_nondet =
```
thk123, line 694: :no_entry_sign: :safety_pin:

src/goto-harness/recursive_initialization.cpp
```diff
@@ -753,8 +756,10 @@ code_blockt recursive_initializationt::build_array_string_constructor(
   {
     index_exprt index_expr{dereference_exprt{result},
                            from_integer(index, size_type())};
-    body.add(code_assignt{
```
thk123, line 756: :no_entry_sign: :safety_pin:

src/goto-harness/recursive_initialization.cpp
```diff
@@ -1029,10 +1035,6 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
   const auto function_pointer_selector =
     get_fresh_local_symexpr("function_pointer_selector");
   body.add(code_declt{function_pointer_selector});
-  body.add(
```
thk123, line 1032: :safety_pin:

thk123 (commented):
There are a bunch of changes - it'd be good to have a test for each one to ensure the change has worked as expected. Other than that, lgtm

regression/goto-harness/do-not-use-nondet-for-primitives/test.desc
thk123: I note this info _is_ in the commit too :+1: I still think probably more confusing than helpful in the test, but up to you

NlightNFotis (approved):
Good enough for me, can we have a ticket for the better long term fix?

martin-cs (approved):
Fine modulo @thk123 's comments.

src/goto-harness/recursive_initialization.cpp
hannes-steffenhagen-diffblue: Ehem, upon trying to come up with a test for this I actually found that there was a mess up in some earlier refactoring leading to this function never ever getting called (instead implementation of what this used to do has been moved into `build_dynamic_array_constructor`.

That is a tad unfortunate. I will add a commit that removes this function to show that it doesn’t work right now, and then we can re-refactor it later if we like.

hannes-steffenhagen-diffblue:
Note that there’s a second commit removing a dead (as in, impossible to reach) function now.

PR #5403, opened 2020-06-30
Added scalar assigns clause to code contracts

azcwagner:
- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [x] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [x] My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

This PR expands code contracts functionality with an 'assigns' clause for defining the writable memory frame of a function. At this time, only scalars are supported, with more complex types to follow.

PR #5410, opened 2020-07-03
Report conversion error on void pointer dereferences

danpoe:

### DIFF
--- a/regression/ansi-c/void_dereference/main.c
+++ b/regression/ansi-c/void_dereference/main.c
@@ -1,0 +1,5 @@
+int main()
+{
+  void *p;
+  *p;
+}
--- a/regression/ansi-c/void_dereference/test.desc
+++ b/regression/ansi-c/void_dereference/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+
+^EXIT=1$
+^SIGNAL=0$
+^CONVERSION ERROR$
+cannot dereference void pointer
+--
+^warning: ignoring
+--
+Checks that cbmc reports an error for void pointer dereferences
--- a/src/ansi-c/c_typecheck_expr.cpp
+++ b/src/ansi-c/c_typecheck_expr.cpp
@@ -1730,6 +1730,14 @@ void c_typecheck_baset::typecheck_expr_dereference(exprt &expr)
   }
   else if(op_type.id()==ID_pointer)
   {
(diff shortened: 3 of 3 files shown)


martin-cs (commented):
I mean, yes, but ...

I feel kinda mean saying this on a Friday afternoon but I can't find in the C standard where is says you shouldn't dereference a void pointer.  You definitely can't /use/ the result:

```
6.3.2.2 void
The (nonexistent) value of a void expression (an expression that has type void) shall not be used in any way, and implicit or explicit conversions (except to void) shall not be applied to such an expression.  If an expression of any other type is evaluated as a void expression, its value or designator is discarded. (A void expression is evaluated for its side effects.)
```

Section 6.3.2.3 goes on to talk about pointers but doesn't say you can't dereference a void pointer.  Section 6.5.3.2 (all from ISO 9899:1999) gives the semantics of the unary * ;indirection' operator.

```
4 The unary * operator denotes indirection.  If the operand points to a function, the results is a function designator; if it points to an object, the result is an lvalue designating the object.  If the operand has type "pointer to type", the result has type "type".  If an invalid value has been assigned to the pointer, the behaviour of the unary * operator is undefined. (83)

`(83)` ...  Among the invalid values for dereferencing a pointer by the unary * operator are a null pointe, an address inappropriately aligned for the type of object pointed to, and the address of an object after the end of its lifetime.
```

So I think that:

```
void f00 (void *p) {
  *p;

 return;
}
```

Is actually valid C but it's semantics are "dereference the address but discard the value, undefined behaviour if the address is not valid".

I know this seems like pedantry but I have written code including:

```
  (void)argc;  // Avoid the complaint about unused variables
  ...
}
```

So it seems that `f00` above is legit if idiomatic C for saying the value must be valid and try using it.

This is, in a way what the https://github.com/diffblue/cbmc/pull/5403#pullrequestreview-440715391 does.

danpoe:
Ah I see, thanks for this. In that case I think I'll convert it to a warning.

(Btw, one other thing the PR doesn't address yet are cases like `&(*p)` or `__typeof__(*p)` where even a warning shouldn't be printed.)

martin-cs:
No worries.

If I recall the standard correctly there is something weird about `&(*p)`.  I think it is semantically equivalent to `p` with a type cast and you are allowed to skip the "is this dereferencable" check.

PR #5411, opened 2020-07-06
Havoc undefined functions during symbolic execution

polgreen:
This havocs any modifiable arguments passed to undefined functions
at symex time. It is enabled by --havoc-undefined-functions

It is possible to achieve a similar result by using goto-instrument generate-function-bodies with the correct command line options but only if the binaries you are using this on are small. goto-instrument generate-function-bodies fails to finish within 1 hour on binaries taken from Xen. Havoc-ing during symex time allows us to run CBMC without generating the function bodies, and CBMC finishes within ten minutes.

Where am I supposed to document this? None of the symex configuration options seem to be documented in the user guide?

### DIFF
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -240,6 +240,9 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
   if(cmdline.isset("drop-unused-functions"))
     options.set_option("drop-unused-functions", true);
 
+  if(cmdline.isset("havoc-undefined-functions"))
+    options.set_option("havoc-undefined-functions", true);
+
   if(cmdline.isset("string-abstraction"))
     options.set_option("string-abstraction", true);
 
@@ -1082,6 +1085,9 @@ void cbmc_parse_optionst::help()
     HELP_REACHABILITY_SLICER_FB
     " --full-slice                 run full slicer (experimental)\n" // NOLINT(*)
     " --drop-unused-functions      drop functions trivially unreachable from main function\n" // NOLINT(*)
+    " --havoc-undefined-functions\n"
+    "                              for any function that has no body, assign non-deterministic values to\n" // NOLINT(*)
+    "                              any parameters passed as non-const pointers and the return value\n" // NOLINT(*)
     "\n"
     "Semantic transformations:\n"
     // NOLINTNEXTLINE(whitespace/line_length)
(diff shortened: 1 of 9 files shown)


src/goto-symex/symex_function_call.cpp
```diff
@@ -293,6 +294,26 @@ void goto_symext::symex_function_call_code(
...
+      {
+        if(
+          arg.type().id() == ID_pointer &&
+          !arg.type().subtype().get_bool(ID_C_constant) &&
+          arg.type().subtype().id() != ID_code)
+        {
+          exprt object = dereference_exprt(arg, arg.type().subtype());
+          exprt cleaned_object = clean_expr(object, state, true);
+          guard_expr_managert tmp;
+          const guard_exprt guard(true_exprt(), tmp);
```
romainbrenguier: `state` already has a guard manager, why not use that one?

romainbrenguier (approved):

src/goto-checker/symex_bmc.cpp
```diff
@@ -212,5 +214,12 @@ void symex_bmct::no_body(const irep_idt &identifier)
   {
     log.warning() << "**** WARNING: no body for function " << identifier
                   << log.eom;
```
martin-cs: Did you mean to delete this one because this will give a duplicate `log.eom` in the normal case.

martin-cs (approved):
The code looks fine but I have so many unrelated questions.

1. The goto-instrument route does seem better and generally in line with the idea of "every function should have a body, generated or not" but given the problem you are having with it, I can see how this fix is worth having.

2. How can goto-instrument take this long?  @hannes-steffenhagen-diffblue @NlightNFotis am I wrong thinking that you wrote this code?  Is there something you can do here.

3. Should this just be the default to minimise surprise?

src/goto-checker/symex_bmc.cpp
polgreen: deleted

src/goto-symex/symex_function_call.cpp
polgreen: now using that

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1082,6 +1085,9 @@ void cbmc_parse_optionst::help()
     HELP_REACHABILITY_SLICER_FB
     " --full-slice                 run full slicer (experimental)\n" // NOLINT(*)
     " --drop-unused-functions      drop functions trivially unreachable from main function\n" // NOLINT(*)
+    " --havoc-undefined-functions\n"
+    "                              for any function that has no body, assign non-deterministic values to\n" // NOLINT(*)
+    "                              any parameters passed as non-const pointers and the return value\n" // NOLINT(*)
```
smowton, line 1090: Might want to clarify that the return value is nondet with or without this option

smowton (approved):

PR #5412, opened 2020-07-07
Fix static symbols referencing in harness file

NlightNFotis:
This is fixing an issue we came across when trying to run `goto-harness` against the batch jobs in the https://github.com/awslabs/aws-c-common/tree/master/.cbmc-batch/jobs files, where the tool would generate harnesses that referenced static symbols in different files.

### DIFF
--- a/regression/goto-harness-multi-file-project/static_symbols_referencing/first_one.c
+++ b/regression/goto-harness-multi-file-project/static_symbols_referencing/first_one.c
@@ -1,0 +1,25 @@
+#include "first_one.h"
+#include <assert.h>
+
+static int static_function(int a)
+{
+  assert(a == 0);
+  return a;
+}
+
+int non_static_function(int a)
+{
+  return static_function(a);
+}
+
+static int with_matching_signature(int a)
+{
+  assert(0 && "this is not reachable as far as goto-harness is concerned");
+  return 0;
+}
+
+void non_static_with_non_matching_signature(void)
+{
+  // this is just here so `static_with_matching_signature` has a non-file-local use
+  assert(static_with_matching_signature(10) == 10);
+}
--- a/src/goto-harness/recursive_initialization.cpp
+++ b/src/goto-harness/recursive_initialization.cpp
@@ -8,6 +8,7 @@ Author: Diffblue Ltd.
 
 #include "recursive_initialization.h"
 
+#include <goto-programs/name_mangler.h>
 #include <util/allocate_objects.h>
 #include <util/arith_tools.h>
 #include <util/c_types.h>
@@ -1008,6 +1009,27 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
   for(const auto &target : targets)
   {
     auto const assign = code_assignt{dereference_exprt{result}, target};
+    auto const sym_to_lookup =
+      target.id() == ID_address_of
+        ?
+        // This is either address of or pointer; in pointer case, we don't
+        // need to do anything. In the address of case, the operand is
+        // a symbol representing a target function.
+        to_symbol_expr(to_address_of_expr(target).object()).get_identifier()
+        : "";
+    // skip referencing globals because the corresponding symbols in the symbol
+    // table are no longer marked as file local.
+    if(has_prefix(id2string(sym_to_lookup), FILE_LOCAL_PREFIX))
+    {
+      continue;
+    }
+    else if(
+      goto_model.get_symbol_table().lookup(sym_to_lookup) &&
+      goto_model.get_symbol_table().lookup(sym_to_lookup)->is_file_local)
+    {
+      continue;
+    }
+
     if(function_pointer_index != targets.size() - 1)
     {
       auto const condition = equal_exprt{
--- a/regression/goto-harness-multi-file-project/static_symbols_referencing/first_one.h
+++ b/regression/goto-harness-multi-file-project/static_symbols_referencing/first_one.h
@@ -1,0 +1,1 @@
+int non_static_function(int a);
--- a/regression/goto-harness-multi-file-project/static_symbols_referencing/second_one.c
+++ b/regression/goto-harness-multi-file-project/static_symbols_referencing/second_one.c
@@ -1,0 +1,8 @@
+#include "first_one.h"
+
+int another(int (*fun_ptr)(int), int c)
+{
+  int a = (*fun_ptr)(c);
+
+  return a;
+}
--- a/regression/goto-harness-multi-file-project/static_symbols_referencing/test.desc
+++ b/regression/goto-harness-multi-file-project/static_symbols_referencing/test.desc
@@ -1,0 +1,15 @@
+CORE
+dummy.c
+--function another --harness-type call-function
+^EXIT=10$
+^SIGNAL=0$
+\[static_function\.assertion\.1\] line \d assertion a == 0: FAILURE
+\[non_static_with_non_matching_signature\.assertion\.1] line \d+ assertion static_with_matching_signature\(10\) == 10: SUCCESS
+
+^VERIFICATION FAILED$
+--
+^CONVERSION ERROR$
+--
+For this particular error, we care mostly that goto-harness
+doesn't reference static symbols in other files, which would
+cause analysis through CBMC to fail with a conversion error.
--- a/src/goto-programs/name_mangler.cpp
+++ b/src/goto-programs/name_mangler.cpp
@@ -19,7 +19,7 @@ operator()(const symbolt &src, const std::string &extra_info)
   std::string basename = get_base_name(src.location.get_file().c_str(), false);
 
   std::stringstream ss;
-  ss << CPROVER_PREFIX << "file_local_";
+  ss << FILE_LOCAL_PREFIX;
   ss << std::regex_replace(
           std::regex_replace(basename, forbidden, "_"), multi_under, "_")
      << "_";
@@ -42,7 +42,7 @@ operator()(const symbolt &src, const std::string &extra_info)
   uint32_t eight_nibble_hash = (uint32_t)hash;
 
   std::stringstream ss;
-  ss << CPROVER_PREFIX << "file_local_" << std::setfill('0') << std::setw(8)
-     << std::hex << eight_nibble_hash << "_" << extra_info << "_" << src.name;
+  ss << FILE_LOCAL_PREFIX << std::setfill('0') << std::setw(8) << std::hex
+     << eight_nibble_hash << "_" << extra_info << "_" << src.name;
   return irep_idt(ss.str());
 }
--- a/src/goto-programs/name_mangler.h
+++ b/src/goto-programs/name_mangler.h
@@ -13,6 +13,8 @@
 #include <regex>
 #include <vector>
 
+#define FILE_LOCAL_PREFIX CPROVER_PREFIX "file_local_"
+
 /// \brief Mangles the names in an entire program and its symbol table
 ///
 /// The type parameter to this class should be a functor that has a no-arg


src/goto-harness/recursive_initialization.cpp
```diff
@@ -1008,6 +1008,12 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
   for(const auto &target : targets)
   {
     auto const assign = code_assignt{dereference_exprt{result}, target};
+    auto sym_to_lookup = to_symbol_expr(to_address_of_expr(target).operands()[0]).get_identifier();
+    // skip referencing globals
```
hannes-steffenhagen-diffblue: Please note down here that the corresponding symbols in the symbol table are no longer marked as file local, which is why you are looking them up by name I believe?

src/goto-harness/recursive_initialization.cpp
```diff
@@ -1008,6 +1008,12 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
...
+    if (has_prefix(id2string(sym_to_lookup), "__CPROVER_file_local_"))
```
hannes-steffenhagen-diffblue: `CPROVER_PREFIX "file_local_"`

regression/goto-harness-gb-input/chain.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -e
+
+goto_harness=$1
+cbmc=$2
+
+input_goto_binary=${*:$#}
+args=${*:3:$#-3}
+echo "BURRITO"
```
hannes-steffenhagen-diffblue: :no_entry_sign: No burritos

regression/goto-harness-gb-input/static-symbol-referencing/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+aws_byte_buf_append_harness.goto
+--harness-type call-function --harness-function-name generated_harness_test_function --function aws_byte_buf_append
+^EXIT=6$
+^SIGNAL=0$
+--
+^CONVERSION ERROR$
+--
+For this particular error, we care mostly that goto-harness
```
thk123: :pick: Probably worth adding a positive regex (e.g. verification successful) just in case the error message for conversion error ever changes

src/goto-harness/recursive_initialization.cpp
```diff
@@ -1008,6 +1008,21 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
   for(const auto &target : targets)
   {
     auto const assign = code_assignt{dereference_exprt{result}, target};
+    auto sym_to_lookup =
```
thk123: :pick: `const`
:bulb: Invariant on the first operand being a symbol_exprt with specific justification for why should be a symbol

src/goto-harness/recursive_initialization.cpp
```diff
@@ -1008,6 +1008,21 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
...
+      target.id() == ID_address_of
+        ?
+        // This is either address of or pointer; in pointer case, we don't
+        // need to do anything
+        to_symbol_expr(to_address_of_expr(target).operands()[0])
+          .get_identifier()
+        : "";
+    // skip referencing globals because the corresponding symbols in the symbol
+    // table are no longer marked as file local.
+    if(has_prefix(id2string(sym_to_lookup), CPROVER_PREFIX "file_local_"))
```
thk123: :no_entry_sign: Please introduce a global constant for `file_local_` and use it here and in the mangling.

src/goto-harness/recursive_initialization.cpp
```diff
@@ -1008,6 +1008,21 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
   for(const auto &target : targets)
   {
     auto const assign = code_assignt{dereference_exprt{result}, target};
+    auto sym_to_lookup =
+      target.id() == ID_address_of
+        ?
+        // This is either address of or pointer; in pointer case, we don't
+        // need to do anything
+        to_symbol_expr(to_address_of_expr(target).operands()[0])
```
thk123: :no_entry_sign: Having converted to `address_of_exprt`, you should use `.object()`
```suggestion
        to_symbol_expr(to_address_of_expr(target).object())
```

regression/goto-harness-gb-input/Makefile
```diff
@@ -0,0 +1,18 @@
+default: tests.log
```
thk123: :question: How come we need to introduce a whole new test directory?

thk123 (commented):
Even when relying on magic strings, it is important to not duplicate them, as this makes them even more fragile and prone to error.

regression/goto-harness-gb-input/Makefile
NlightNFotis: The reason for this is that the other ones do not have the same compile->test pipeline.

The `goto-harness` directory takes as input `.c` files, passes them to `goto-cc`, gets a binary out, and then feeds that to `goto-harness`.

In this directory I just pass a goto binary directly to `goto-harness`, and the result of that straight into cbmc to make sure there's no conversion error (i.e. erroneous output in the c file produced by the harness).

src/goto-harness/recursive_initialization.cpp
NlightNFotis: I played around with introducing an INVARIANT, but to be honest it was very awkward - because it's a ternary operator, the whole thing needs to be expression, and for that to happen, I need to introduce a lambda if I want to introduce more than one statements followed by the expression.

I also feel that the INVARIANT is being covered by the `PRECONDITION` inside `to_symbol_expr`, but I can definitely add the justification for it.

regression/goto-harness-gb-input/static-symbol-referencing/test.desc
NlightNFotis: I'm not sure I agree with this to be honest.

I see your point, but the whole essence of this test is to test the parser - or more accurately, that the parsing of this file is successful and the file doesn't contain any erroneous input.

If I add a positive regex (i.e. `VERIFICATION SUCCESSFUL`), then I'm also depended on the analysis part to have a specific output, something that I feel is extraneous for this specific issue.

I'll have a think about improving the test though.

NlightNFotis:
@thk123 @hannes-steffenhagen-diffblue All of your comments have now been addressed. Can I get another review please?

regression/goto-harness-gb-input/Makefile
thk123: Oh I see, meaning you're checking in goto binary files. Do you have to? Can you not run the C file through goto-cc as part of the test - checking in binary files is a recipe for them going out of date when someone changes the C file but doesn't realize they need to run goto-cc manually.

regression/goto-harness-gb-input/static-symbol-referencing/test.desc
thk123: Well maybe not verification successful, but some check that the parsing has been successful. Much better to have a test that needs periodic updating when it fails, than a test that starts spuriously passing

thk123 (dismissed):
Blocking comment addressed - but feel quite strongly that a test without much positive regexes is not a great test.

thk123:
Done - looks like you've got a formatting problem btw

regression/goto-harness-gb-input/static-symbol-referencing/test.desc
NlightNFotis: I think I found a way: I am including some of the pipeline status messages in the `test.desc`, like `converting`, `type-checking`, etc..

This way, it will fail both on the assertion failure, and if it didn't go through all the appropriate stages.

hannes-steffenhagen-diffblue (approved):
I don’t particularly like that this includes a binary, but I suppose it’s better than not having a test at all.

regression/goto-harness-gb-input/Makefile
NlightNFotis: This is a test on `goto-harness` only: We don't need a `.c` file because the input to `goto-harness` is a goto-binary (hence why it's unneeded to run `goto-cc`).

Also, the reason this is based on a binary test is that because this was very hard to reproduce based on a c file (as this was based on the compilation pipeline of the aws-c-common repository), I figured I could only use the binary (which is the output of the `make goto` step), which is what I need to feed to `goto-harness` for it to manifest the problem.

So in a sense, we don't run the risk of it getting stale (unless we change the internal representation of goto-binaries and the binary attached becomes stale - in which case it's easy enough to reproduce).

thk123: Hmm.. I didn't initially spot the source C file is missing. :no_entry_sign: At a min, there needs to be a readme of how someone would go about recreating this. I'd rather a little more work on coming up with a simpler reproducing example (didn't we have one that admittedly relied on using goto-cc, but at least was simpler). This will also I guess fix the windows CI problem

martin-cs: Yes; binary without source and build instructions is a hard block.  Even if it is not what the test is for, I would *really* prefer it if the test compiled up the goto-binaries it needs.

regression/goto-harness-gb-input/static-symbol-referencing/test.desc
martin-cs: Agree with @thk123 

It may be testing for the parsing but `VERIFICATION SUCCESSFUL` turns up in so many of our other tests that it is hardly an unreasonable extra condition.

martin-cs: It is better to test more than you need and to have a more fragile test than to have something that may just start silently passing even when wrong.

martin-cs (changes requested):
Code change looks fine.  Agree with @thk123 about the issue of binary tests.  The binary format is not intended to be stable or portable.

NlightNFotis:
Hi @martin-cs @thk123 @hannes-steffenhagen-diffblue I have re-requested approvals on this - The PR was rewritten (at least the last commit that contains the test), to now address your concerns and contains a c source file as a test, and with some other shortcomings addressed.

src/goto-harness/recursive_initialization.cpp
```diff
@@ -1008,6 +1009,27 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
   for(const auto &target : targets)
   {
     auto const assign = code_assignt{dereference_exprt{result}, target};
+    auto const sym_to_lookup =
```
hannes-steffenhagen-diffblue, line 1012: Aside comment:

FWIW I think we should _probably_  do something cleverer for function pointers, this is very similar to function pointer removal.

src/goto-harness/recursive_initialization.cpp
```diff
@@ -1008,6 +1009,27 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
...
+      target.id() == ID_address_of
+        ?
+        // This is either address of or pointer; in pointer case, we don't
+        // need to do anything. In the address of case, the operand is
+        // a symbol representing a target function.
+        to_symbol_expr(to_address_of_expr(target).object()).get_identifier()
+        : "";
+    // skip referencing globals because the corresponding symbols in the symbol
+    // table are no longer marked as file local.
+    if(has_prefix(id2string(sym_to_lookup), FILE_LOCAL_PREFIX))
```
hannes-steffenhagen-diffblue, line 1022: The thinking here is that symbols prefixed this way might actually be referenceable, but we decided to push off potential different handling for this case off as a different feature. This is coming from `--export-file-local-symbols` apparently.

hannes-steffenhagen-diffblue (approved):
I think this is probably addressing @martin-cs and @thk123 comments now

regression/goto-harness-multi-file-project/static_symbols_referencing/first_one.c
```diff
@@ -0,0 +1,13 @@
+#include "first_one.h"
+#include <assert.h>
+
+static int static_function(int a)
+{
+  assert(a == 0);
```
hannes-steffenhagen-diffblue, line 6: Sorry to be annoying, but in the interest of having good tests it might actually make sense to have a couple more assertions here. This is reachable because `non_static_function` is, but to contrast I think it’d be good to also have another static function with this signature that"s called by a non-static function with a _different_ signature whose assertion we don’t reach, demonstrating that we currently do not consider these functions for goto-harness.

NlightNFotis: Done.

thk123 (approved):
Blocking problems addressed

NlightNFotis:
Hi @martin-cs, can we get another review on this one? Thanks :)

martin-cs (approved):
@NlightNFotis Looks much better -- thank you!

danpoe (approved):
LGTM

peterschrammel (approved):

PR #5414, opened 2020-07-07
show flow-insensitive value set

polgreen:
Adds command line argument to goto-instrument to show the flow insensitive value set. Previously only possible to show the flow sensitive value set. Does not add this command line argument to help file, to be consistent with the original flow sensitive value set

None of the existing similar goto-instrument options are documented, or in the help file so I assume these are not meant to be publicly advertised features

I haven't added new features, just the ability to print the result

### DIFF
--- a/src/goto-instrument/goto_instrument_parse_options.cpp
+++ b/src/goto-instrument/goto_instrument_parse_options.cpp
@@ -54,6 +54,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <pointer-analysis/goto_program_dereference.h>
 #include <pointer-analysis/show_value_sets.h>
 #include <pointer-analysis/value_set_analysis.h>
+#include <pointer-analysis/value_set_analysis_fi.h>
 
 #include <analyses/call_graph.h>
 #include <analyses/constant_propagator.h>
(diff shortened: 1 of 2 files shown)


peterschrammel (approved):

hannes-steffenhagen-diffblue:
:question: 

> Does not add this command line argument to help file, to be consistent with the original flow sensitive value set

I really don’t believe this is the sort of thing we want to be consistent about. Having one undocumented option is already bad, adding another and not documenting that either just because the other one isn’t is just making it worse.

Also, would you mind adding some tests that show what kind of output you’re expecting from this option to `regression/goto-instrument`?

martin-cs (changes requested):
Code change is fine.  Please get the consistency in the help file by documenting more not less!  Also a regression test would be good; `goto-instrument`'s functionality is woefully under tested.

martin-cs:
General aspirational ramble.  I would like to get us down to just one abstract interpretation framework, rather than the current three.  As part of that I would like to move this kind of functionality out of `goto-instrument` and into `goto-analyser --show` because ... it's really not instrumentation or even program transformation.  To do this we need to:

1. refactor how function pointer removal works.
2. be able to do return removal with function pointers in.
3. modify the `ait` framework to tolerate function pointers.
4. flow-insensitive `ait_storage`.
5. migrate over the pointer analysis and other bits.

3-5 I am happy to do and are not actually much work.  I have been working with @chrisr-diffblue to do 1 and 2 for ... a year or two but it has stalled a number of times.  @NlightNFotis has written at least two full PRs to do this.

If you care about this kind of functionality, maybe we could get this moving again?  @chrisr-diffblue ?

Issue #5415, opened 2020-07-08
CBMC dumps core on simple program


danielsn:
```
#include <assert.h>
#include <stddef.h>

struct foo {
  int x;
  int y;
};


int main(struct foo* p)
{
}
```

CBMC version: 5.12 (cbmc-5.12.1)
Operating system: osx
Exact command line resulting in the issue: cbmc test.c
What behaviour did you expect: no crash. Perhaps an error message
What happened instead: 
```
CBMC version 5.12 (cbmc-5.12.1) 64-bit x86_64 macos
Parsing be.c
Converting
Type-checking be
--- begin invariant violation report ---
Invariant check failed
/tmp/cbmc-20200620-32000-1h7ts1b/src/ansi-c/ansi_c_entry_point.cpp:476 function: generate_ansi_c_start_function
Condition: false
Reason: Unreachable
Backtrace:
0   cbmc                                0x00000001007bb91a _Z15print_backtraceRNSt3__113basic_ostreamIcNS_11char_traitsIcEEEE + 74
1   cbmc                                0x00000001007bbdd9 _Z13get_backtracev + 167
2   cbmc                                0x00000001005e9e9c _Z29invariant_violated_structuredI17invariant_failedtJRKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEEEENS1_9enable_ifIXsr3std10is_base_ofIS0_T_EE5valueEvE4typeES9_S9_iS9_DpOT0_ + 44
3   cbmc                                0x00000001007cc459 _Z25invariant_violated_stringRKNSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEES7_iS7_S7_ + 9
4   cbmc                                0x00000001004f8813 _Z30generate_ansi_c_start_functionRK7symboltR13symbol_tabletR16message_handlertRK28c_object_factory_parameterst + 7689
5   cbmc                                0x00000001004f67ff _Z18ansi_c_entry_pointR13symbol_tabletR16message_handlertRK28c_object_factory_parameterst + 993
6   cbmc                                0x000000010078f0d3 _ZN15language_filest26generate_support_functionsER13symbol_tablet + 135
7   cbmc                                0x00000001005c01f5 _Z21initialize_goto_modelRKNSt3__16vectorINS_12basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEENS4_IS6_EEEER16message_handlertRK8optionst + 2277
[...]

PR #5416, opened 2020-07-09
Add human readable error message for invalid main signatures

hannes-steffenhagen-diffblue:
Currently we have special handling if the entry point for analysis is called
`main`, in particular we assume it is one of the following:

    int main(void)
    int main(int argc, char *argv[], char *envp[])

If it is not so far we have been failing with an UNREACHABLE invariant failure.
Since it is completely possible to have a main signature like that, we should
fail with a proper human readable error message that tells people what is wrong
and maybe how they can work around it.

For now, addresses https://github.com/diffblue/cbmc/issues/5415. As I say in that ticket maybe we want to handle non-standard `main` signatures in a more graceful way, but until we decide on that having a better error message should take away some of the pain.

### DIFF
--- a/src/ansi-c/ansi_c_entry_point.cpp
+++ b/src/ansi-c/ansi_c_entry_point.cpp
@@ -18,6 +18,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <linking/static_lifetime_init.h>
 
 #include "c_nondet_symbol_factory.h"
+#include "expr2c.h"
 
 exprt::operandst build_function_environment(
   const code_typet::parameterst &parameters,
@@ -473,7 +474,19 @@ bool generate_ansi_c_start_function(
       }
     }
     else
-      UNREACHABLE;
+    {
+      const namespacet ns{symbol_table};
+      const std::string main_signature = type2c(symbol.type, ns);
+      throw invalid_source_file_exceptiont{
+        "'main' with signature '" + main_signature +
+        "' found,"
+        " but expecting one of:\n"
+        "   int main(void)\n"
+        "   int main(int argc, char *argv[])\n"
+        "   int main(int argc, char *argv[], char *envp[])\n"
+        "If this is a non-standard main entry point please provide a custom\n"
+        "entry function and point to it via cbmc --function instead"};
+    }
   }
   else
   {
--- a/regression/cbmc/human-readable-error-on-wrong-main-signature/main.c
+++ b/regression/cbmc/human-readable-error-on-wrong-main-signature/main.c
@@ -1,0 +1,12 @@
+#include <assert.h>
+#include <stddef.h>
+
+struct foo
+{
+  int x;
+  int y;
+};
+
+int main(struct foo *p)
+{
+}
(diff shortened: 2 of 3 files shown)


NlightNFotis (approved):
Okay

NlightNFotis:
Is there a long term plan for better handling of the situation? Or is it how we're going to work with this?

hannes-steffenhagen-diffblue:
@NlightNFotis Like I said, I am not sure what the ‘better’ version looks like.

src/ansi-c/ansi_c_entry_point.cpp
```diff
@@ -473,7 +474,19 @@ bool generate_ansi_c_start_function(
...
+      const std::string main_signature = type2c(symbol.type, ns);
+      throw invalid_source_file_exceptiont{
+        "'main' with signature '" + main_signature +
+        "' found,"
+        " but expecting one of:\n"
+        "   int main(void)\n"
+        "   int main(int argc, char *argv[])\n"
+        "   int main(int argc, char *argv[], char *envp[])\n"
+        "If this is a non-standard main entry point please provide a custom\n"
+        "entry function and point to it via cbmc --function instead"};
```
thk123, line 488: This is a great error message :green_heart: 
:grey_question: If I do `cbmc main.c --function main` does this work? No worries if not, but if it does, I'd be half tempted to give that as the advice

thk123 (approved):

src/ansi-c/ansi_c_entry_point.cpp
```diff
@@ -473,7 +474,19 @@ bool generate_ansi_c_start_function(
       }
     }
     else
-      UNREACHABLE;
+    {
+      const namespacet ns{symbol_table};
+      const std::string main_signature = type2c(symbol.type, ns);
+      throw invalid_source_file_exceptiont{
+        "'main' with signature '" + main_signature +
```
chrisr-diffblue, line 481: Should it explicitly say "with an invalid signature" ?

src/ansi-c/ansi_c_entry_point.cpp
```diff
@@ -473,7 +474,19 @@ bool generate_ansi_c_start_function(
...
+        "' found,"
+        " but expecting one of:\n"
+        "   int main(void)\n"
+        "   int main(int argc, char *argv[])\n"
+        "   int main(int argc, char *argv[], char *envp[])\n"
+        "If this is a non-standard main entry point please provide a custom\n"
```
chrisr-diffblue, line 487: Thanks for the extra user help in the message :-)

chrisr-diffblue (approved):

src/ansi-c/ansi_c_entry_point.cpp
hannes-steffenhagen-diffblue: It doesn’t unfortunately. What does work is adding something like

```
void harness(void) {
  main(NULL);
}
```

and pointing at that.

thomasspriggs (approved):
:+1:

src/ansi-c/ansi_c_entry_point.cpp
hannes-steffenhagen-diffblue: Well, I feel like ‘invalid’ is a matter of opinion. Maybe this is coming from code that does have a non-standard entry point. It is certainly not what we expect though, so saying that much is value neutral.

Issue #5397, opened 2020-06-29
Field sensitivity is not applied for pointer to struct member


andreast271:
CBMC version: 5.12 (cbmc-5.12.2-3-g9b5f761d5)
Operating system: Linux x86_64  
Exact command line resulting in the issue: cbmc --program-only access.c
What behaviour did you expect: field sensitivity is applied for the assignment to s.n[i]
What happened instead: field sensitivity is not applied

Test case:
[access.c.txt](https://github.com/diffblue/cbmc/files/4844617/access.c.txt)

I have encountered an example where field sensitivity is not applied, although as far as I can see it could be applied. I have attached the full test case above. The relevant code is as follows:

Good case:

```
  struct st s;
  s.n[i] = k;
```

Here field sensitivity works as expected and the assignment to s.n[i] becomes the equation

```
(37) s!0@1#1..n == { s!0@1#1..n[[0]], s!0@1#1..n[[1]], s!0@1#1..n[[2]], s!0@1#1..n[[3]] } WITH [(signed long int)i!0@1#2:=k!0@1#1]
```

However, with an intermediate pointer field sensitivity is no longer applied:

```
  struct st s;
  int *p = s.n;
  p[i] = k;
```

Now the assignment updates the whole struct:

```
(37) s!0@1#1 == byte_update_little_endian({ .c=s!0@1#1..c, .$pad1=s!0@1#1..$pad1, .n={ s!0@1#1..n[[0]], s!0@1#1..n[[1]], s!0@1#1..n[[2]], s!0@1#1..n[[3]] } }, 4l * (signed long int)i!0@1#2 + 4l, k!0@1#1, signed int)
```

There is also the dual case of reading from an index of the array s.n. The situation is similar in that field sensitivity is applied if the access happens directly ( s.n[i] ), but not if an intermediate pointer is inserted ( p = s.n, p[i] ). In this case a simplification rule can get the same effect as field sensitivity, but having a solution involving field sensitivity would be preferable.

martin-cs:
I suspect this is happening because the pointer analysis is not field sensitive and so says that `p` can point to `s`.  If I am right then this won't be fixed without a major change to the internals of the symbolic execution engine.  VSD might give a way of doing field sensitive pointer analysis that could be used.  Alternatively I know @tautschnig did have plans at one point to improve symex's handling of pointers so that the preliminary pointer analysis was not needed.

smowton:
[...]

PR #5418, opened 2020-07-12
Generate index expressions for deref(p + offset), where p points to an array in a struct

smowton:
Currently if offset is unknown, value_set_dereference will resolve p + offset to [(o1, unknown), (o2, unknown), ...], ultimately leading to a byte-update operation that may touch any of o1, o2 etc and specifically knocks them both out of the constant propagator.

With this change if p is known to point to arrays or array-members of structs then we will produce an expression like (c1 ? array1[offset] : c2 ? someobj.array2[offset] : ...)

Field-sensitivity may then kick in and update only the array within someobj, rather than the whole struct (this is unsound, but mirrors our behaviour when an array-in-struct is directly referenced. We may want to make this behaviour optional if the user is intentionally searching for solutions involving out-of-bounds accesses)

Note this is quite similar to value_set_dereferencet::build_reference_to's behaviour when referring to an array (but not an array within a struct), which builds an index expression instead of a byte-extract op. We can't extend that code however as the (constant + nonconstant) form of the pointer being dereferenced is lost by then.

TODOs to get this to usable standard: add tests, check particular if there's any bad interaction between array-type / index-type casts, such as deref'ing an int* whose underlying objects are char arrays.

The interplay of `value_set`, the `value_set_dereferencet` wrapper around it, the `goto_symext::dereference_rec` wrapper around *that*, and `build_reference_to`, `memory_model` and `memory_model_bytes` all attempting to salvage what would otherwise be byte-extract operations has become an unholy mess. Possible suggestion any forthcoming sponsored work on CBMC: document / clean them up?

### DIFF
--- a/regression/cbmc/array-cell-sensitivity15/access.c
+++ b/regression/cbmc/array-cell-sensitivity15/access.c
@@ -1,0 +1,16 @@
+struct st
+{
+  char c;
+  int n[4];
+};
+
+int main()
+{
+  unsigned i;
+  int k;
+  i %= 4;
+  struct st s;
+  int *p = s.n;
+  p[i] = k;
+  return 0;
+}
--- a/regression/cbmc/array-cell-sensitivity15/test.desc
+++ b/regression/cbmc/array-cell-sensitivity15/test.desc
@@ -1,0 +1,11 @@
+CORE paths-lifo-expected-failure
+access.c
+--program-only
+^EXIT=0$
+^SIGNAL=0$
+s!0@1#1\.\.n == \{ s!0@1#1\.\.n\[\[0\]\], s!0@1#1\.\.n\[\[1\]\], s!0@1#1\.\.n\[\[2\]\], s!0@1#1\.\.n\[\[3\]\] } WITH \[\(.*\)i!0@1#2:=k!0@1#1\]
+--
+byte_update
+--
+This tests applying field-sensitivity to a pointer to an array that is part of a struct. See cbmc issue #5397 and PR #5418 for more detail.
+Disabled for paths-lifo mode, which does not support --program-only.
--- a/src/pointer-analysis/value_set_dereference.cpp
+++ b/src/pointer-analysis/value_set_dereference.cpp
@@ -99,6 +99,36 @@ static json_objectt value_set_dereference_stats_to_json(
   return json_result;
 }
 
+optionalt<exprt> value_set_dereferencet::try_add_offset_to_indices(
+  const exprt &expr,
+  const exprt &offset_elements)
+{
+  if(const auto *index_expr = expr_try_dynamic_cast<index_exprt>(expr))
+  {
+    return index_exprt{
+      index_expr->array(),
+      plus_exprt{index_expr->index(),
+                 typecast_exprt::conditional_cast(
+                   offset_elements, index_expr->index().type())}};
+  }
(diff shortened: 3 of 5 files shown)


tautschnig:
@smowton Would you be willing to revive this one? Perhaps all that it takes is a rebase plus adding "Fixes: #5397" to the commit message? The example included in that issue might well serve as a test case!

smowton:
I'm on holiday from the weekend; I'll take care of this + the other one you'd asked me to revive then :)

smowton:
So actually what about the other TODO I mentioned-- are you confident this won't break anything mentioned there?

tautschnig:
> So actually what about the other TODO I mentioned-- are you confident this won't break anything mentioned there?

Let's just try it out! It should be fairly easy to revert the patch if the need arises? What I'm perhaps most worried about are large arrays, where the resulting expression may cause performance problems.

smowton:
OK, test case added and rebased as requested.

regression/cbmc/array-cell-sensitivity15/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+access.c
+--program-only
```
tautschnig, line 3: I believe you need to add this test to the exclusion list in `regression/validate-trace-xml-schema/check.py` as `--program-only` isn't compatible with `--xml-ui`. That should make the regression tests pass.

regression/cbmc/array-cell-sensitivity15/access.c
```diff
@@ -0,0 +1,16 @@
...
+int main()
+{
+  unsigned i;
+  int k;
+  i %= 4;
+  struct st s;
+  int *p = s.n;
+  p[i] = k;
+  return 0;
+}
```
tautschnig: Nit pick: missing final newline.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -99,6 +99,33 @@ static json_objectt value_set_dereference_stats_to_json(
...
 }
 
+optionalt<exprt> value_set_dereferencet::try_add_offset_to_indices(
+  const exprt &expr,
+  const exprt &offset)
+{
+  if(const auto *index_expr = expr_try_dynamic_cast<index_exprt>(expr))
+  {
+    return index_exprt{index_expr->array(),
+                       plus_exprt{index_expr->index(), offset}};
```
tautschnig: Checking: 1) Is the `offset` here in terms of array elements, and not bytes? 2) Is it guaranteed to be of the same type as `index_expr->index().type()` (else do `typecast_exprt::conditional_cast(offset, index_expr->index().type())`)?

regression/cbmc/array-cell-sensitivity15/test.desc
smowton: Done

src/pointer-analysis/value_set_dereference.cpp
smowton: Elements: yes, since this comes from an expression of the form (ptr + offset), i.e. advancing a number of elements. Clarified by renaming the variable to include the units.

Cast: perhaps not, added conditional cast out of caution.

smowton:
@tautschnig rebased, passing tests, ready for review

tautschnig (approved):

martin-cs (approved):
Seems reasonable and feels like he kind of tweak that actually makes a big difference on larger code-bases.

PR #5420, opened 2020-07-13
Robust type name to type identifier conversion for C harnesses

thomasspriggs:
This PR makes the type name to type identifier conversion used in generating C harnesses more robust.

- [x] n/a Because this improves existing functionality, to reduce potential compile errors in generated harnesses and I assume that these potential errors are not documented. ~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~
- [x] n/a - Non claimed ~My commit message includes data points confirming performance improvements (if claimed).~

### DIFF
--- a/src/ansi-c/type2name.cpp
+++ b/src/ansi-c/type2name.cpp
@@ -324,7 +324,7 @@ std::string type_name2type_identifier(const std::string &name)
       replace_special_characters(name))));
 }
 
-std::string type2identifier(const typet &type, const namespacet &ns)
+std::string type_to_partial_identifier(const typet &type, const namespacet &ns)
 {
   return type_name2type_identifier(type2name(type, ns));
 }
--- a/src/ansi-c/type2name.cpp
+++ b/src/ansi-c/type2name.cpp
@@ -305,11 +305,12 @@ std::string type_name2type_identifier(const std::string &name)
   };
   const auto replace_invalid_characters_with_underscore =
     [](const std::string &identifier) {
-      static const std::regex non_alpha_numeric{"[^A-Za-z0-9]+"};
+      static const std::regex non_alpha_numeric{"[^A-Za-z0-9\x80-\xff]+"};
       return std::regex_replace(identifier, non_alpha_numeric, "_");
     };
   const auto strip_leading_non_letters = [](const std::string &identifier) {
-    static const std::regex identifier_regex{"[A-Za-z][A-Za-z0-9_]*"};
+    static const std::regex identifier_regex{
+      "[A-Za-z\x80-\xff][A-Za-z0-9_\x80-\xff]*"};
     std::smatch match_results;
     bool found = std::regex_search(identifier, match_results, identifier_regex);
     POSTCONDITION(found);
--- a/src/ansi-c/type2name.cpp
+++ b/src/ansi-c/type2name.cpp
@@ -277,11 +277,10 @@ std::string type2name(const typet &type, const namespacet &ns)
 /// If we want utilities like dump_c to work properly identifiers
 /// should ideally always be valid C identifiers
 /// This replaces some invalid characters that can appear in type2name output.
-std::string type2identifier(const typet &type, const namespacet &ns)
+std::string type_name2type_identifier(const std::string &name)
 {
-  auto type2name_res = type2name(type, ns);
   std::string result{};
-  for(char c : type2name_res)
+  for(char c : name)
   {
     switch(c)
     {
@@ -300,4 +299,9 @@ std::string type2identifier(const typet &type, const namespacet &ns)
     }
   }
   return result;
+}
+
+std::string type2identifier(const typet &type, const namespacet &ns)
+{
+  return type_name2type_identifier(type2name(type, ns));
 }
--- a/src/ansi-c/type2name.cpp
+++ b/src/ansi-c/type2name.cpp
@@ -308,15 +308,15 @@ std::string type_name2type_identifier(const std::string &name)
       static const std::regex non_alpha_numeric{"[^A-Za-z0-9\x80-\xff]+"};
       return std::regex_replace(identifier, non_alpha_numeric, "_");
     };
-  const auto strip_leading_non_letters = [](const std::string &identifier) {
+  const auto strip_leading_digits = [](const std::string &identifier) {
     static const std::regex identifier_regex{
-      "[A-Za-z\x80-\xff][A-Za-z0-9_\x80-\xff]*"};
+      "[A-Za-z\x80-\xff_][A-Za-z0-9_\x80-\xff]*"};
     std::smatch match_results;
     bool found = std::regex_search(identifier, match_results, identifier_regex);
     POSTCONDITION(found);
     return match_results.str(0);
   };
-  return strip_leading_non_letters(replace_invalid_characters_with_underscore(
+  return strip_leading_digits(replace_invalid_characters_with_underscore(
     replace_special_characters(name)));
 }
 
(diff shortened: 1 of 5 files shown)


src/ansi-c/type2name.cpp
```diff
@@ -301,3 +300,8 @@ std::string type2identifier(const typet &type, const namespacet &ns)
   }
   return result;
 }
+
+std::string type2identifier(const typet &type, const namespacet &ns)
+{
+  return type_name2type_identifier(type2name(type, ns));
```
NlightNFotis, line 325: Nice :+1:

NlightNFotis (approved):
Solid!

src/ansi-c/type2name.cpp
```diff
@@ -277,27 +279,52 @@ std::string type2name(const typet &type, const namespacet &ns)
 /// If we want utilities like dump_c to work properly identifiers
 /// should ideally always be valid C identifiers
 /// This replaces some invalid characters that can appear in type2name output.
-std::string type2identifier(const typet &type, const namespacet &ns)
+std::string type_name2type_identifier(const std::string &name)
```
hannes-steffenhagen-diffblue, line 282: I still don’t like the `2` naming convention here, but changing that isn’t really in scope for this PR.

src/ansi-c/type2name.cpp
```diff
@@ -277,27 +279,52 @@ std::string type2name(const typet &type, const namespacet &ns)
...
+    return result;
+  };
+  const auto replace_invalid_characters_with_underscore =
+    [](const std::string &identifier) {
+      static const std::regex non_alpha_numeric{"[^A-Za-z0-9]"};
+      return std::regex_replace(identifier, non_alpha_numeric, "_");
+    };
+  const auto remove_duplicate_underscores = [](const std::string &identifier) {
+    static const std::regex duplicate_underscore{"_+"};
+    return std::regex_replace(identifier, duplicate_underscore, "_");
```
hannes-steffenhagen-diffblue: I’m not completely sure this is always ‘the right thing’ to do, for example some compilers (including `gcc`) have a builtin `__int128` type, i.e. double underscores in names are a relatively normal thing to see.

That being said, since we only really need these to be related to the original names and not _necessarily_ completely match up with them I suppose this is fine.

hannes-steffenhagen-diffblue (approved):

src/ansi-c/type2name.cpp
chrisr-diffblue: Yeah, I'm not sure this is the right thing to do either - double underscore is a perfectly valid identifier for internal symbols - which you may well get in things like header files.

src/ansi-c/type2name.cpp
```diff
@@ -277,27 +279,52 @@ std::string type2name(const typet &type, const namespacet &ns)
...
+      }
     }
-  }
-  return result;
+    return result;
+  };
+  const auto replace_invalid_characters_with_underscore =
+    [](const std::string &identifier) {
+      static const std::regex non_alpha_numeric{"[^A-Za-z0-9]"};
+      return std::regex_replace(identifier, non_alpha_numeric, "_");
```
chrisr-diffblue, line 309: This, I assume, risks generating clashing names?

src/ansi-c/type2name.h
```diff
@@ -20,6 +20,15 @@ class namespacet;
 
 std::string type2name(const typet &type, const namespacet &ns);
 
+/**
+ * Constructs a string describing the given type, which can be used as part of
+ * a `C` identifier. The resulting identifier is not guaranteed to uniquely
+ * identify the type in all cases.
```
chrisr-diffblue, line 26: This might need clarifying  - in fact, it might actually be worth making this explicit in the function name, e.g. rename it to something like `type_to_approximate_identifier` or `type_to_nonunique_identifier`.

src/ansi-c/type2name.cpp
hannes-steffenhagen-diffblue: @chrisr-diffblue Yes, in general this isn’t generating a unique identifier. The idea would be to use this as an identifier fragment useful for generating a human-readable-but-unique identifier elsewhere.

src/ansi-c/type2name.h
thomasspriggs: I get that "identifier is a bit of a misnomer in this case. However `nonunique_identifier` seems like a bit of an oxymoron to me. How does the term `partial_identifier` sound to you?

src/ansi-c/type2name.cpp
thomasspriggs: We were already risking clashing names, rather than guaranteeing uniqueness. For example `fruit*` will be replaced with `fruit_ptr_`, but there was no guarantee that something else in user code wasn't already called `fruit_ptr`.

src/ansi-c/type2name.cpp
thomasspriggs: Superflous double underscores could be generated in the preceding section of code. For example `char**` would be replaced with `char_ptr__ptr`, which is what I wanted to tidy up.

However I would also argue that we should not emit identifiers which begin with an underscore in the harness code. This is because identifiers which start with an underscore are reserved for use by the compiler / standard libraries. See section 7.1.3 of the standard - http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf
I would argue that because we are outputting the  harnesses as C code for user modification, the harnesses are therefore user code. This means that use of identifiers with leading underscores would invoke "undefined behaviour" in terms of the language standards.

hannes-steffenhagen-diffblue: @thomasspriggs note that this is currently not necessarily being used only for the harness – that’s where we found the issue, but the reason it occurred was actually because of the way we generated names for specialisations of generic builtin functions.

chrisr-diffblue: I know that double-underscore is reserved for compiler/library use, and partly thats why I think it may be ok to emit them - because if you are generating a C file from a goto-binary, that goto-binary may well have legitimate typenames from include files (e.g. from things like #defines, etc) and as such, they will be at least semi-legitimate to emit (assuming the emitted code is then compiled against the same headers/libraries). 

I may of course be missing something about the context for this change, so if you have an example of where this fails and you need this, do feel free to point me at it :-)

src/ansi-c/type2name.h
chrisr-diffblue: `partial_identifer` sounds reasonable to me - it makes it clearer to an unsuspecting caller that they need to understand what is returning :-)

thomasspriggs: Done.

src/ansi-c/type2name.cpp
thomasspriggs: Ok. I have updated this, in order to retain more underscores.

thomasspriggs:
@chrisr-diffblue Please can you re-review? Comments are addressed in new commits at the end of the PR.

thomasspriggs:
I have pushed to restart CI only. The windows job appeared to fail for reasons unrelated to my changes.

chrisr-diffblue (approved):
Looks good, thanks. Approving, with a slight note of caution that if CBMC ever supports (or claims to support) internationalised source code (i.e. getting into the whole input file encoding/internal encoding question) then this might need revisiting. It won't be the only place that'll need revisiting though :-)

thomasspriggs:
I did take a look at the implementation of the lexer and there does seem to be some UTF-8 support in there. But it is reasonable to assume that there will be incompatibilities, if there are no end-to-end regression tests for such functionality.

PR #5421, opened 2020-07-14
Fix goto-cc same filename bug

### LINKED ISSUE #5417: goto-cc cannot compile two files with the same name together
When compiling two files with the same name in a single invocation of goto-cc, only one of the files' functions are included in the resulting goto-binary.
```c
// 1/c.c
int foo1(){ return 0; }
```
```c
// 2/c.c
int foo2(){ return 0; }
```
What happens:
```sh
> goto-cc -o a.out 1/c.c 2/c.c
> goto-instrument --show-goto-functions a.out
Reading GOTO program from 'a.out'
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

foo2 /* foo2 */
        // 0 file 2/c.c line 1 function foo2
        return 0;
        // 1 file 2/c.c line 1 function foo2
        GOTO 1
        // 2 file 2/c.c line 1 function foo2
     1: END_FUNCTION

>
```
What I expected to happen: the binary should also contain an implementation of `foo1()`.
CBMC version: cd3c603d8fe2d5b0572dd761532dd76d37f89d33

danpoe:
Fixes #5417.


src/goto-cc/gcc_mode.cpp
```diff
@@ -688,7 +689,9 @@ int gcc_modet::doit()
 
           std::string new_name=
             get_base_name(arg_it->arg, true)+new_suffix;
-          std::string dest=temp_dir(new_name);
+
+          temp_dirs.emplace_front("goto-cc-XXXXXX");
+          std::string dest = temp_dirs.front()(new_name);
```
hannes-steffenhagen-diffblue: If we’re already doing temp dirs and what not, couldn’t we create proper temp files for the destinations of these things instead of maintaining a list of temp directories? Because it seems like now we’re creating a temp directory for each file. I don’t think this is a big problem practically speaking, but it seems a bit odd to me.

src/goto-cc/gcc_mode.cpp
```diff
@@ -655,7 +656,7 @@ int gcc_modet::doit()
 
   // We now iterate over any input files
 
-  temp_dirt temp_dir("goto-cc-XXXXXX");
+  std::forward_list<temp_dirt> temp_dirs;
```
hannes-steffenhagen-diffblue: Why not just a vector?

danpoe: This was done since `temp_dirt` deletes the temporary dir upon destruction. So when the vector elements are relocated the directories would be deleted. Though thinking about it the proper fix would be to add a move constructor to `temp_dirt` and delete the copy constructor.

src/goto-cc/gcc_mode.cpp
danpoe: Ah yes there's already `temporary_filet` to create a temporary file (and it also handles the copying and moving correctly). I'll use that.

danpoe: Ok, switched this back to using temporary directories, as the name of the preprocessed file is used at various places (such as when deducing the name of the object file when goto-cc is invoked with a single source file, i.e., `goto-cc -c test.c`).

chrisr-diffblue (approved):
LGTM, thanks!

peterschrammel (approved):

PR #5425, opened 2020-07-16
 Add binary representation to all bitvector values in the xml trace

thomasspriggs:
Add binary representation to all bitvector values in the xml trace. The purpose of this is to provide a consistent interface for getting the values of bitvector-ed types out of the trace.

- [x] N/A - None claimed - ~My commit message includes data points confirming performance improvements (if claimed).~

### DIFF
--- a/doc/assets/xml_spec.tex
+++ b/doc/assets/xml_spec.tex
@@ -149,8 +149,10 @@ always present
 \item \texttt{full\_lhs}: original lhs expression (after dereferencing)
 \item \texttt{full\_lhs\_value}: a constant with the new value.
   \begin{itemize}
-    \item if the type is a floating point number type, there will be an attribute
-    \texttt{binary} showing its value.
+    \item If the type of data can be represented as a fixed width sequence of bits
+     then, there will be an attribute \texttt{`binary`} containing the binary
+     representation. If the data type is signed and the value is negative
+     then the binary will be encoded using two's complement.
   \end{itemize}
 
 \end{itemize}
@@ -164,7 +166,7 @@ always present
   <location .. />
   <type>signed int</type>
   <full_lhs>b</full_lhs>
-  <full_lhs_value>-1879048192</full_lhs_value>
+  <full_lhs_value binary="10010000000000000000000000000000">-1879048192</full_lhs_value>
 </assignment>
 <assignment assignment_type="state" base_name="d" display_name="main::1::d" hidden="false" identifier="main::1::d" mode="C" step_nr="26" thread="0">
   <location ../>
--- a/regression/cbmc/integral-trace/test.c
+++ b/regression/cbmc/integral-trace/test.c
@@ -1,0 +1,30 @@
+#include <assert.h>
+#include <inttypes.h>
+#include <math.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdlib.h>
+
+int main()
+{
+  int8_t i1 = 2;
+  int16_t i2 = 3;
+  int32_t i3 = 4;
+  int64_t i4 = 5;
+
+  int8_t in1 = -2;
+  int16_t in2 = -3;
+  int32_t in3 = -4;
+  int64_t in4 = -5;
+
+  uint8_t u8 = 'a';
+  uint16_t u16 = 8;
+  uint32_t u32 = 16;
+  uint64_t u64 = 32;
+
+  void *ptr = malloc(1);
+
+  assert(false);
+
+  return 0;
+}
--- a/regression/cbmc/integral-trace/test.desc
+++ b/regression/cbmc/integral-trace/test.desc
@@ -1,0 +1,22 @@
+CORE
+test.c
+--trace --xml-ui
+activate-multi-line-match
+^EXIT=10$
+^SIGNAL=0$
+VERIFICATION FAILED
+<full_lhs>u8</full_lhs>\s*<full_lhs_value binary="01100001">97</full_lhs_value>
+<full_lhs>u16</full_lhs>\s*<full_lhs_value binary="0{12}1000">8</full_lhs_value>
+<full_lhs>u32</full_lhs>\s*<full_lhs_value binary="0{27}10000">16ul?</full_lhs_value>
+<full_lhs>u64</full_lhs>\s*<full_lhs_value binary="0{58}100000">32ull?</full_lhs_value>
+<full_lhs>i1</full_lhs>\s*<full_lhs_value binary="0{6}10">2</full_lhs_value>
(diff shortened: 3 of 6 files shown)


regression/cbmc/integral-trace/test.c
```diff
@@ -0,0 +1,30 @@
...
+  uint16_t u16 = 8;
+  uint32_t u32 = 16;
+  uint64_t u64 = 32;
+
+  void *ptr = malloc(1);
+
+  assert(false);
+
+  return 0;
+}
```
hannes-steffenhagen-diffblue: missing newline

hannes-steffenhagen-diffblue:
Looks like the new regression test is crashing with an invariant violation? Also, you should probably update [the actual XSD](https://github.com/diffblue/cbmc/blob/develop/doc/assets/xml_spec.xsd) with this.

src/goto-programs/xml_goto_trace.cpp
```diff
@@ -25,31 +25,32 @@ Author: Daniel Kroening
...
-  const goto_trace_stept &step,
-  const namespacet &ns)
-{
-  return can_cast_type<floatbv_typet>(step.full_lhs_value.type());
-}
-
 xmlt full_lhs_value(const goto_trace_stept &step, const namespacet &ns)
 {
-  xmlt full_lhs_value{"full_lhs_value"};
+  xmlt xml{"full_lhs_value"};
```
thk123: :pick: Personally don't love this name either just cos the type tells you its xml /shrug

thk123 (approved):

peterschrammel (approved):

doc/assets/xml_spec.tex
```diff
@@ -164,7 +166,7 @@ \subsection{Trace Steps in XML}
   <location .. />
   <type>signed int</type>
   <full_lhs>b</full_lhs>
-  <full_lhs_value>-1879048192</full_lhs_value>
+  <full_lhs_value binary="10010000000000000000000000000000">-1879048192</full_lhs_value>
```
martin-cs, line 169: This seems like a bit of a weird way of giving this.  As there are multiple possible representations of the value, wouldn't it be better as `<full_lhs_value_binary></>` `<full_lhs_value_decimal></>` etc.?  Or possibly with `full_lhs_value` be the default human-readable.  Then again, I am not an XML user so maybe this is the appropriate way of doing it.

regression/cbmc/integral-trace/test.desc
```diff
@@ -0,0 +1,22 @@
...
+test.c
+--trace --xml-ui
+activate-multi-line-match
+^EXIT=10$
+^SIGNAL=0$
+VERIFICATION FAILED
+<full_lhs>u8</full_lhs>\n\s*<full_lhs_value binary="01100001">97</full_lhs_value>
+<full_lhs>u16</full_lhs>\n\s*<full_lhs_value binary="0{12}1000">8</full_lhs_value>
+<full_lhs>u32</full_lhs>\n\s*<full_lhs_value binary="0{16}0{11}10000">16u</full_lhs_value>
+<full_lhs>u64</full_lhs>\n\s*<full_lhs_value binary="0{32}0{16}0{10}100000">32ul</full_lhs_value>
```
martin-cs: FWIW : This is not what would expect to find in a field marked `binary`; I would just expect `0` and `1`.

martin-cs (approved):
Just thoughts; no action required.

regression/cbmc/integral-trace/test.desc
hannes-steffenhagen-diffblue: @martin-cs this: `0{32}` is regex notation for repetition, it means 32 0's. I assume this was chosen here to make the structure of the matched expressions somewhat easier to see. This isn’t actually in the output. Though I’m not sure what the rationale was of doing this over just `0{58}100000`

martin-cs: Ah; yes.  I missed that it is a regex in the test pattern rather than a literal output.  Thanks.

doc/assets/xml_spec.tex
thomasspriggs: This is consistent with the way that the binary value is already given for floating point values. My intention is to add this feature in a way which will require fewer code changes for anyone who is processing the xml traces. If I was writing this representation from scratch, I would have probably chosen to given equal prominence to the binary and the decimal. This could looks something like -
```
<full_lhs_value><binary>10010000000000000000000000000000</binary><decimal>-1879048192</decimal></full_lhs_value>
```

thomasspriggs:
@hannes-steffenhagen-diffblue The xsd does not need to be updated, because it already contains the binary attribute. This was already present in the xsd because it is already present for floating point values.

regression/cbmc/integral-trace/test.c
thomasspriggs: Fixed.

src/goto-programs/xml_goto_trace.cpp
thomasspriggs: Ok. I am going to make this `value_xml` instead.

src/goto-programs/xml_goto_trace.cpp
```diff
@@ -27,15 +27,15 @@ Author: Daniel Kroening
 
 xmlt full_lhs_value(const goto_trace_stept &step, const namespacet &ns)
 {
-  xmlt full_lhs_value{"full_lhs_value"};
+  xmlt value_xml{"full_lhs_value"};
```
hannes-steffenhagen-diffblue, line 30: I am not sure if I agree with the reasoning of this rename. I’d probably have gone for `full_lhs_value_xml` instead.

src/goto-programs/xml_goto_trace.cpp
```diff
@@ -25,31 +25,34 @@ Author: Daniel Kroening
...
   const irep_idt identifier =
     lhs_object.has_value() ? lhs_object->get_identifier() : irep_idt();
+  value_xml.data = from_expr(ns, identifier, value);
 
-  if(step.full_lhs_value.is_not_nil())
-    full_lhs_value.data = from_expr(ns, identifier, step.full_lhs_value);
-  if(full_lhs_value_includes_binary(step, ns))
+  const auto &bv_type = type_try_dynamic_cast<bitvector_typet>(value.type());
+  const auto &constant = expr_try_dynamic_cast<constant_exprt>(value);
+  if(bv_type && constant)
```
hannes-steffenhagen-diffblue, line 43: I did like the `full_lhs_value_includes_binary` name, but I suppose if the intention is to use all bitvectors for this then this is more honest.

hannes-steffenhagen-diffblue (approved):
LGTM, thanks for addressing comments

PR #5427, opened 2020-07-20
Fix cbmc crash on pointer checks of void pointer dereferences

danpoe:

### DIFF
--- a/regression/cbmc/void_pointer6/test.desc
+++ b/regression/cbmc/void_pointer6/test.desc
@@ -1,0 +1,10 @@
+CORE
+main.c
+--pointer-check
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+Checks that cbmc does not crash with --pointer-check on dereference of an
+invalid void pointer
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -1164,10 +1164,24 @@ void goto_checkt::pointer_validity_check(
 
   const exprt &pointer=expr.pointer();
 
-  auto size_of_expr_opt = size_of_expr(expr.type(), ns);
-  CHECK_RETURN(size_of_expr_opt.has_value());
+  exprt size;
 
-  auto conditions = address_check(pointer, size_of_expr_opt.value());
+  if(expr.type().id() == ID_empty)
+  {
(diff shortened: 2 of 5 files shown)


regression/cbmc/void_pointer6/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--pointer-check
+^EXIT=10$
```
peterschrammel, line 4: + verification failed?

src/analyses/goto_check.cpp
```diff
@@ -1164,10 +1164,20 @@ void goto_checkt::pointer_validity_check(
...
   const exprt &pointer=expr.pointer();
 
-  auto size_of_expr_opt = size_of_expr(expr.type(), ns);
-  CHECK_RETURN(size_of_expr_opt.has_value());
+  exprt size;
 
-  auto conditions = address_check(pointer, size_of_expr_opt.value());
+  if(expr.type().id() == ID_empty)
+  {
+    size = from_integer(1, size_type());
```
hannes-steffenhagen-diffblue, line 1175: Should we change `size_of_expr` to return `1` here instead?

danpoe: Not sure. There are quite a number of uses in the codebase that check that the return value of `size_of_expr()` is not empty, and if it is throw an exception or have some other handling. Probably we should review those at some point, though that's out of scope of this PR.

peterschrammel: Is the purpose of this to create an address check that fails?

danpoe: No the address check should succeed. Basically `*p` with `p` being a pointer to void is valid as long as `p` points to valid memory, though then using the value would be undefined behavior.

regression/cbmc/void_pointer6/test.desc
danpoe: Not necessary I think as `EXIT=10` already indicates that cbmc exited without error and verification failed.

src/analyses/goto_check.cpp
peterschrammel: Please explain the magic `1` in a comment.

peterschrammel (approved):

PR #5431, opened 2020-07-22
Add support for enums with underlying type specifications to the parser

danpoe:
This adds support for parsing enums that specify the underlying type (a clang extension) to cbmc. For example, a usage would look like `enum enum1 : unsigned {X, Y}`.

The change to the grammar however introduces an additional shift/reduce conflict. The issue seems to be that the new syntax leads to ambiguity with the syntax for bitfields (where the bit portion also starts with a `:`). That is, for `enum enum1 : unsigned`, when `enum enum1` has been read and `:` becomes the lookahead token, there is a choice between reducing (thus `enum enum1` becomes the portion matched by the `enum_name` rule), or shifting, in which case `: unsigned` will become part of the portion matched by the rule as well. Any suggestions on how or if this can be fixed would be appreciated.


regression/ansi-c/enum9/main.c
```diff
@@ -0,0 +1,20 @@
...
+
+struct S
+{
+  enum my_enum2 : unsigned a;
+  enum my_enum2 : unsigned b : 2;
+};
+
+int main()
+{
+  enum my_enum2 : unsigned enum_var1;
```
hannes-steffenhagen-diffblue, line 19: Huh? Any ideas on what this means? Seems like clang allows this notation, but it appears to not actually mean anything, for instance

```
#include <stdio.h>

enum my_enum : signed char {
  A, B
};


int main(void) {
  enum my_enum : unsigned long long  X = 255;
  printf("%d\n", X);
}
```

Seems to compile just fine and works as if the second `: unsigned long long` wasn’t there.

danpoe: Yeah it seems to just be ignored by clang. My guess would be that it might have been overly complicated for clang to exclude this case in the grammar, which would also be true for cbmc. I think we should report this as an error during conversion/typechecking.

chrisr-diffblue (commented):
I _think_ this looks OK, though its not my absolute area of expertise. I assume you've already taken a look at `Parser::ParseEnumSpecifier` in https://clang.llvm.org/doxygen/ParseDecl_8cpp_source.html to check what Clang is doing - though of course they have a hand-written parser which gives them a bit more flexibility.

src/util/irep_ids.def
```diff
@@ -212,6 +212,7 @@ IREP_ID_ONE(NULL)
 IREP_ID_ONE(null)
 IREP_ID_ONE(nullptr)
 IREP_ID_ONE(c_enum)
+IREP_ID_ONE(underlying_type)
```
chrisr-diffblue: Should this be something like `enum_underlying_type` or is there a reason is should not be specific to enums?

chrisr-diffblue (approved):
LGTM, with one small query around naming - but don't want to bike shed :-)

src/util/irep_ids.def
danpoe: No particular reason, I've changed it to `enum_underlying_type`.

PR #5432, opened 2020-07-24
CBMC additional profiling info: byte extract, update metrics

natasha-jeppu:

### DIFF
--- a/regression/cbmc/byte-op-metric/test.desc
+++ b/regression/cbmc/byte-op-metric/test.desc
@@ -1,0 +1,16 @@
+CORE
+main.c
+--show-byte-ops
+^EXIT=0$
+^SIGNAL=0$
+^Byte Extracts:$
+^.*main\.c line 8 function main$
+^.* byte\_extract\_little\_endian\(.*\)$
+^Number of byte extracts: 1$
+^Byte Updates:$
+^.*main\.c line 8 function main$
+^.* byte\_update\_little\_endian\(.*\)$
+^Number of byte updates: 1$
+--
+--
+To test --show-byte-ops option that displays byte extract and update metrics.
--- a/src/goto-symex/show_program.cpp
+++ b/src/goto-symex/show_program.cpp
@@ -11,11 +11,17 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "show_program.h"
 
+#include <fstream>
 #include <iostream>
 
 #include <goto-symex/symex_target_equation.h>
 
 #include <langapi/language_util.h>
+
+#include <util/byte_operators.h>
+#include <util/exit_codes.h>
+#include <util/json_irep.h>
+#include <util/ui_message.h>
 
 /// Output a single SSA step
 /// \param ns: Namespace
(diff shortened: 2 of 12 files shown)


src/goto-symex/show_program.cpp
```diff
@@ -162,6 +162,7 @@ std::string json_get_key_byte_op_stats(const byte_op_type type)
   case byte_op_type::BYTE_UPDATE:
     return "byteUpdateStats";
   }
```
natasha-jeppu: Thought this was legal for switch case over enums! But throws "control reaches end of non-void function" error.

hannes-steffenhagen-diffblue: This is actually fine! What was probably happening here was that you didn’t have a `default` case. This is not actually an error either, just a warning that if there are any cases you haven’t handled, then you might not reach a return statement which is invalid in a non-`void` function.

This can unfortunately happen even if you explicitly cover all possible cases for the enum, for example:

```
enum class ButtonState { Off = 0, On = 1 };
```

you’d think that only `On` and `Off` are valid values here... unfortunately that’s not how C++ works, so for example in C++11 you can write `ButtonState someInvalidState = static_cast<ButtonState>(10);` – this is somewhat awkward, but is completely intended for use as a way to define new integral types without implicit conversions to existing ones, like for example C++17’s `std::byte`. If values other than the named ones aren’t intended to be used the best way to handle this is probably something along the lines of:

```
switch(yourValue) {
  case YourSwitch::FirstCase: return <something>;
  case YourSwitch::SecondCase: return <something>;
  // ...
  case YourSwitch::NthCase: return <something>;
}
UNREACHABLE; // this is a macro we have in CBMC for marking unreachable locations, this triggers an assertion failure at runtime
```

This way you’ll not silence the `-Wswitch` warning that warns about missing cases (in case anyone adds more enum values later on), but won’t (or at least shouldn’t) get the warning about a missing return here.

In any case, the `switch` statement isn’t the actual problem here!

natasha-jeppu: Thank you! 
Do you have any comments regarding the code in general?

src/util/byte_operators.h
```diff
@@ -123,4 +130,35 @@ inline byte_update_exprt &to_byte_update_expr(exprt &expr)
...
+
+inline std::size_t get_byte_op_count(const byte_op_type type, const exprt &expr)
+{
+  std::size_t count = 0;
+  if(!expr.id().empty() && byte_expr_is_of_type(type, expr))
+    count++;
+
+  forall_expr(it, expr.operands())
+  {
+    count += get_byte_op_count(type, *it);
```
thomasspriggs: :no_entry_sign: I am concerned that this recursive implementation may run the risk of exceeding the maximum stack size. An `exprt` tree could be arbitrarily deep, depending on the input program. The maximum stack size is platform dependant and does differ depending on a raft of factors.
:bulb: This should be straight forward to resolve based on implementing this algorithm using `exprt::visit_pre` instead. Using `exprt::visit_pre` would resolve this comment as it is implemented without using recursion.

src/util/byte_operators.h
```diff
@@ -25,7 +25,7 @@ Author: Daniel Kroening, kroening@kroening.com
 /// The object can either be interpreted in big endian or little endian, which
 /// is reflected by the \c id of the expression which is either
 /// \c ID_byte_extract_big_endian or \c ID_byte_extract_little_endian
-class byte_extract_exprt:public binary_exprt
+class byte_extract_exprt : public binary_exprt
```
thomasspriggs, line 28: :pick: Just a nit pick, but our usual practice is to reformat lines where other changes are being made only. Based on this practise, this line should not be updated.

src/util/byte_operators.h
```diff
@@ -111,6 +111,13 @@ class byte_update_exprt : public ternary_exprt
   const exprt &value() const { return op2(); }
 };
 
+template <>
+inline bool can_cast_expr<byte_update_exprt>(const exprt &base)
+{
+  return base.id() == ID_byte_update_little_endian ||
+         base.id() == ID_byte_update_big_endian;
```
thomasspriggs, line 124: :pick: It would be nice to have a corresponding invariant in the constructor of `byte_update_exprt` to confirm that the `_id` argument is always `ID_byte_update_little_endian` or `ID_byte_update_big_endian` when it is constructed.

src/util/byte_operators.h
```diff
@@ -111,6 +111,13 @@ class byte_update_exprt : public ternary_exprt
   const exprt &value() const { return op2(); }
 };
 
+template <>
+inline bool can_cast_expr<byte_update_exprt>(const exprt &base)
```
thomasspriggs, line 121: :+1: Looks useful.

src/util/byte_operators.h
```diff
@@ -123,4 +130,35 @@ inline byte_update_exprt &to_byte_update_expr(exprt &expr)
   return static_cast<byte_update_exprt &>(expr);
 }
 
+enum byte_op_type
```
thomasspriggs: :pick: `enum class` should be preferred over `enum` for new code. Please see point `Enum.3` in the C++ core guidelines for more information. These are not part of part of the CBMC programming standards, but they do provide a good reference for the current C++ best practice.

src/util/byte_operators.h
```diff
@@ -123,4 +130,35 @@ inline byte_update_exprt &to_byte_update_expr(exprt &expr)
...
+{
+  BYTE_EXTRACT,
+  BYTE_UPDATE
+};
+inline bool byte_expr_is_of_type(const byte_op_type type, const exprt &expr)
```
thomasspriggs: :no_entry_sign: It is better not to keep function implementations in header files, because it increases compile times. I know there are plenty of implementations in header files already, but lets not add to the problem. I suggest moving these functions which use `byte_op_type` to `show_program.cpp` where they are used.

src/util/byte_operators.h
thomasspriggs: :no_entry_sign: Our coding standards require that identifiers for user defined types end in the letter `t`. Therefore `byte_op_type` must be renamed to `byte_op_typet`.

src/util/byte_operators.h
```diff
@@ -123,4 +130,35 @@ inline byte_update_exprt &to_byte_update_expr(exprt &expr)
   return static_cast<byte_update_exprt &>(expr);
 }
 
+enum byte_op_type
+{
+  BYTE_EXTRACT,
```
thomasspriggs: :pick: I would prefer it if this PR could be implemented without introducing new names for existing types. To be able to read your code I need to keep in mind that `byte_op_type::BYTE_EXTRACT` correlates to `byte_extract_exprt`.

:bulb: As an example of the techniques which could be employed, your `get_byte_op_count` function could be implemented without these new names like this -
```
template <typename expr_typet>
std::size_t count_expr_typed(const exprt &expr)
{
  static_assert(
    std::is_base_of<exprt, expr_typet>::value,
    "`expr_typet` is expected to be a type of `exprt`.");
  std::size_t count = 0;
  expr.visit_pre([&](const exprt &expr) {
    if(can_cast_expr<expr_typet>(expr))
    {
      count++;
    }
  });
  return count;
}
```
This could then be called like - `count_expr_typed<byte_extract_exprt>(expr)`

regression/cbmc/byte-op-metric/test.desc
```diff
@@ -0,0 +1,16 @@
+CORE
+main.c
+--show-byte-ops
```
thomasspriggs, line 3: :no_entry_sign: There should really be a duplicate equivalent test run with `--json-ui`. If it has been implemented then they should be tested. Without these tests it will be more difficult to maintain the printing code which you have implemented.

src/goto-symex/show_program.cpp
```diff
@@ -78,3 +84,243 @@ void show_program(const namespacet &ns, const symex_target_equationt &equation)
...
+  const std::string ssa_expr_as_string = from_expr(ns, function_id, ssa_expr);
+
+  out << messaget::faint << "// " << ssa_step.source.pc->location_number << " ";
+  out << ssa_step.source.pc->source_location.as_string() << "\n"
+      << messaget::reset;
+  out << ssa_expr_as_string << "\n";
+}
+
+void show_ssa_step_json(
+  std::ostream &out,
```
thomasspriggs: :pick: The `out` parameter isn't used in `show_ssa_step_json` or `show_byte_op_json`. So it should be removed from these two functions.

src/goto-symex/show_program.cpp
```diff
@@ -78,3 +84,243 @@ void show_program(const namespacet &ns, const symex_target_equationt &equation)
...
+  const namespacet &ns,
+  const SSA_stept &ssa_step,
+  const exprt &ssa_expr,
+  json_arrayt &byte_op_list)
+{
+  const std::string key_srcloc = "sourceLocation";
+  const std::string key_ssa_expr = "ssaExpr";
+  const std::string key_ssa_expr_as_string = "ssaExprString";
+
+  json_objectt &json_ssa_step = byte_op_list.push_back(jsont()).make_object();
```
thomasspriggs: :pick: The way you are creating this is a little mutation heavy. I'd prefer if the `json_objectt` didn't need to be mutated after it was constructed.
:bulb: The `json_objectt` class supports list initialisation. For example the `show_ssa_step_json` could be written like this -
```
void show_ssa_step_json(
  const namespacet &ns,
  const SSA_stept &ssa_step,
  const exprt &ssa_expr,
  json_arrayt &byte_op_list)
{
  const irep_idt &function_id = ssa_step.source.function_id;
  const std::string ssa_expr_as_string = from_expr(ns, function_id, ssa_expr);
  json_objectt json_ssa_step{
    {"sourceLocation", json(ssa_step.source.pc->source_location)},
    {"ssaExprString", json_stringt(ssa_expr_as_string)},
    {"ssaExpr", json_irept(false).convert_from_irep(ssa_expr)}};
  byte_op_list.push_back(std::move(json_ssa_step));
}
```
:information_source: Note that classes derived from `jsont` can be handled/copied by value. This does not risk object slicing with this particular hierarchy. This is because all the fields are defined in `jsont` and none of the derived classes add extra fields.

src/goto-symex/show_program.cpp
```diff
@@ -78,3 +84,243 @@ void show_program(const namespacet &ns, const symex_target_equationt &equation)
...
+      << messaget::reset;
+  out << ssa_expr_as_string << "\n";
+}
+
+void show_ssa_step_json(
+  std::ostream &out,
+  const namespacet &ns,
+  const SSA_stept &ssa_step,
+  const exprt &ssa_expr,
+  json_arrayt &byte_op_list)
```
thomasspriggs: :pick: As a point of best-practice, I would prefer to see a function's return value used for its result rather than an in/out parameter. For example with this function you could return the `json_objectt` for a step, rather than mutating the `json_arrayt &byte_op_list` parameter. Adding the object to the array would then be done in the calling code.

src/goto-symex/show_program.cpp
```diff
@@ -78,3 +84,243 @@ void show_program(const namespacet &ns, const symex_target_equationt &equation)
...
+  switch(ui_message_handler.get_ui())
+  {
+  case ui_message_handlert::uit::XML_UI:
+    msg.error()
+      << "XML UI not supported for displaying byte extracts and updates."
+      << " Try --json-ui instead" << messaget::eom;
+    return;
+
+  case ui_message_handlert::uit::JSON_UI:
+  {
```
thomasspriggs: :pick: I would factor out some of the blocks of code for these cases into a separate functions, in order to keep the size of the `show_byte_ops` function smaller.

thomasspriggs (commented):
Thanks for your efforts. This seems broadly reasonable. Please address the 4 comments I have marked with a :no_entry_sign: emoji first. Please don't take the overall number of comments I have made as heavy criticism, I just had a lot I wanted to share.

Key to the emoji short hands I've used in my review comments - 

- :no_entry_sign: - Blocking comments. I don't intend to approve this PR, until these have been addressed.

- :pick: - Nitpicks. I am willing to approve the PR without corresponding updates being made to the commits. I would be happy if these are addressed. I am ok with it, if these comments are just taken as things to remember next time, if you have insufficient time to deal with them right now.

- :bulb: - Just suggested ideas for making things better. I am ok with it if you don't resolve a comment using the precise way specified next to this emoji.

- :information_source: - Information which might be useful, if you don't know it already.

:pick: I have one general comment on the bodies of you commit messages. Some of them read like a list of separate changes. The commit body should really explain why a given change is being made, rather than list changes. I am guessing that some of this may have arisen due to squashing commits. The body should really be edited when you squash the commits so that it still explains the why of the commit, rather than listing the commits which were squashed into it.

src/util/byte_operators.h
hannes-steffenhagen-diffblue: Also I believe our coding standard says to end all types with `t`.

src/util/byte_operators.h
natasha-jeppu: This came up as an error with cpplint, hence the change. I will make a note for future PRs.
Would you suggest I revert this back or just leave it as it is?

src/util/byte_operators.h
natasha-jeppu: Comment addressed. Pushed changes.

src/util/byte_operators.h
natasha-jeppu: Comment addressed. Pushed changes.

src/util/byte_operators.h
natasha-jeppu: Use of template functions eliminates this altogether. Pushed changes.

regression/cbmc/byte-op-metric/test.desc
natasha-jeppu: Done.

src/goto-symex/show_program.cpp
natasha-jeppu: Done.

src/goto-symex/show_program.cpp
natasha-jeppu: Comment addressed. Pushed changes.

src/goto-symex/show_program.cpp
natasha-jeppu: Comment addressed. Pushed changes.
All json_objectt are now returned from functions.

src/goto-symex/show_program.cpp
natasha-jeppu: Comment addressed. Pushed changes.

thomasspriggs (approved):
Much better. :+1:

src/util/byte_operators.h
thomasspriggs: I am not too concerned either way. The ideal is to only fix the style where the linter CI job in Travis complains. This is because this job only checks the style of the lines which have been updated as part of the PR. So the ideal would be to revert this change if you can do so without the CI job failing. That way when we check the git blame for any given line, we have a better chance of seeing the last functional change. You can do the equivalent check to the CI job, but locally, by checking out your branch and running `./scripts/run_lint.sh develop`, from the root of the repository. You may need to install the python unidiff package first with `pip install --user unidiff`.

peterschrammel (approved):

peterschrammel:
Additional metrics that would be interesting performance-wise could e.g. related to constraints added for arrays in the postprocessing.

NlightNFotis:
@natasha-jeppu Is there anything outstanding on this?

Otherwise, I can merge it immediately, it looks to me 👍

natasha-jeppu:
> @natasha-jeppu Is there anything outstanding on this?
> 
> Otherwise, I can merge it immediately, it looks to me 👍

This is ready to be merged. All comments have been addressed.

PR #5436, opened 2020-07-29
goto-instrument: function pointer removal with value_set_fi

polgreen:
This adds a new option to goto-instrument for removing function pointers.
The points-to analysis is done using flow-insensitive value sets, which is
more precise than using the signature of the function to identify the
points-to set.

- [n/a ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

### DIFF
--- a/regression/goto-instrument/value-set-fi-fp-removal/test.c
+++ b/regression/goto-instrument/value-set-fi-fp-removal/test.c
@@ -1,0 +1,20 @@
+#include <assert.h>
+
+typedef void (*fp_t)();
+
+void f()
+{
+}
+
+void g()
+{
+}
+
+int main(void)
+{
+  fp_t fp = f;
+  fp();
+
+  // this would fool an analysis that looks for functions whose address is taken
+  fp_t other_fp = g;
+}
--- a/regression/goto-instrument/value-set-fi-fp-removal/test.desc
+++ b/regression/goto-instrument/value-set-fi-fp-removal/test.desc
@@ -1,0 +1,12 @@
+CORE
+test.c
+--value-set-fi-fp-removal
+^EXIT=0$
+^SIGNAL=0$
+^  function: f$
+--
+^  function: g$
+--
+This test checks that the value-set-fi-based function pointer removal
+precisely identifies the function to call for a particular function pointer
+call.
--- a/regression/goto-instrument/value-set-fi-fp-removal4/test.c
+++ b/regression/goto-instrument/value-set-fi-fp-removal4/test.c
@@ -1,0 +1,21 @@
+#include <assert.h>
+
+typedef void (*fp_t)();
+
+void f()
+{
+}
+
+void g()
+{
+}
+
+int main(void)
+{
+  // the value set for fp is empty, defaults to standard function pointer removal behaviour
+  fp_t other_fp = g;
+  other_fp = f;
+
+  fp_t fp;
+  fp();
+}
--- a/regression/goto-instrument/value-set-fi-fp-removal5/test.c
+++ b/regression/goto-instrument/value-set-fi-fp-removal5/test.c
@@ -1,0 +1,20 @@
+#include <assert.h>
+
+typedef void (*fp_t)();
+
+void f(int x)
+{
+}
+
+void g(int y)
+{
+}
+
+int main(void)
+{
+  // the value set is empty, defaults to standard function pointer removal behaviour
+  fp_t other_fp = g;
+
+  fp_t fp;
+  fp();
+}
--- a/src/goto-instrument/goto_instrument_parse_options.cpp
+++ b/src/goto-instrument/goto_instrument_parse_options.cpp
@@ -54,6 +54,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <pointer-analysis/goto_program_dereference.h>
 #include <pointer-analysis/show_value_sets.h>
 #include <pointer-analysis/value_set_analysis.h>
+#include <pointer-analysis/value_set_analysis_fi.h>
 
 #include <analyses/call_graph.h>
 #include <analyses/constant_propagator.h>
@@ -111,6 +112,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "undefined_functions.h"
 #include "uninitialized.h"
 #include "unwind.h"
+#include "value_set_fi_fp_removal.h"
 #include "wmm/weak_memory.h"
 
 /// invoke main modules
@@ -1198,6 +1200,12 @@ void goto_instrument_parse_optionst::instrument_goto_program()
         exit(CPROVER_EXIT_USAGE_ERROR);
   }
 
+  if(cmdline.isset("value-set-fi-fp-removal"))
+  {
+    value_set_fi_fp_removal(goto_model, ui_message_handler);
+    do_indirect_call_and_rtti_removal();
+  }
+
   // replace function pointers, if explicitly requested
   if(cmdline.isset("remove-function-pointers"))
   {
@@ -1867,6 +1875,9 @@ void goto_instrument_parse_optionst::help()
     " --no-caching                 disable caching of intermediate results during transitive function inlining\n" // NOLINT(*)
     " --log <file>                 log in json format which code segments were inlined, use with --function-inline\n" // NOLINT(*)
     " --remove-function-pointers   replace function pointers by case statement over function calls\n" // NOLINT(*)
+    " --value-set-fi-fp-removal    build flow-insensitive value set and replace function pointers by a case statement\n" // NOLINT(*)
+    "                              over the possible assignments. If the set of possible assignments is empty the function pointer\n" // NOLINT(*)
+    "                              is removed using the standard remove-function-pointers pass. \n" // NOLINT(*)
     HELP_RESTRICT_FUNCTION_POINTER
     HELP_REMOVE_CALLS_NO_BODY
     HELP_REMOVE_CONST_FUNCTION_POINTERS
--- a/src/goto-instrument/goto_instrument_parse_options.h
+++ b/src/goto-instrument/goto_instrument_parse_options.h
@@ -83,6 +83,7 @@ Author: Daniel Kroening, kroening@kroening.com
   "(full-slice)(reachability-slice)(slice-global-inits)" \
   "(fp-reachability-slice):" \
   "(inline)(partial-inline)(function-inline):(log):(no-caching)" \
+  "(value-set-fi-fp-removal)" \
   OPT_REMOVE_CONST_FUNCTION_POINTERS \
   "(print-internal-representation)" \
   "(remove-function-pointers)" \
--- a/src/goto-instrument/value_set_fi_fp_removal.cpp
+++ b/src/goto-instrument/value_set_fi_fp_removal.cpp
@@ -1,0 +1,247 @@
+/*******************************************************************\
+
+Module: value_set_fi_Fp_removal
+
+Author: Daniel Kroening, kroening@kroening.com
+
+\*******************************************************************/
+
+#include "value_set_fi_fp_removal.h"
+
+#include <goto-programs/goto_model.h>
+#include <goto-programs/remove_function_pointers.h>
+
+#include <pointer-analysis/value_set_analysis_fi.h>
+
+#include <util/base_type.h>
+#include <util/c_types.h>
+#include <util/expanding_vector.h>
+#include <util/fresh_symbol.h>
+#include <util/message.h>
+#include <util/namespace.h>
+#include <util/std_code.h>
+#include <util/union_find.h>
+
+#ifdef USE_STD_STRING
+#  include <util/dstring.h>
+#endif
+
+void fix_argument_types(
+  code_function_callt &function_call,
+  const namespacet &ns)
+{
+  const code_typet &code_type =
+    to_code_type(ns.follow(function_call.function().type()));
+
+  const code_typet::parameterst &function_parameters = code_type.parameters();
+
+  code_function_callt::argumentst &call_arguments = function_call.arguments();
+
+  for(std::size_t i = 0; i < function_parameters.size(); i++)
+  {
+    // casting pointers might result in more arguments than function parameters
+    if(i < call_arguments.size())
+    {
+      call_arguments[i] = typecast_exprt::conditional_cast(
+        call_arguments[i], function_parameters[i].type());
+    }
+  }
+}
+
+void fix_return_type(
+  code_function_callt &function_call,
+  goto_programt &dest,
+  goto_modelt &goto_model)
+{
+  // are we returning anything at all?
+  if(function_call.lhs().is_nil())
+    return;
+
+  const namespacet ns(goto_model.symbol_table);
+
+  const code_typet &code_type =
+    to_code_type(ns.follow(function_call.function().type()));
+
+  // type already ok?
+  if(function_call.lhs().type() == code_type.return_type())
+    return;
+
+  const symbolt &function_symbol =
+    ns.lookup(to_symbol_expr(function_call.function()).get_identifier());
+
+  symbolt &tmp_symbol = get_fresh_aux_symbol(
+    code_type.return_type(),
+    id2string(function_call.source_location().get_function()),
+    "tmp_return_val_" + id2string(function_symbol.base_name),
+    function_call.source_location(),
+    function_symbol.mode,
+    goto_model.symbol_table);
+
+  const symbol_exprt tmp_symbol_expr = tmp_symbol.symbol_expr();
+
+  exprt old_lhs = function_call.lhs();
+  function_call.lhs() = tmp_symbol_expr;
+
+  dest.add(goto_programt::make_assignment(
+    code_assignt(old_lhs, typecast_exprt(tmp_symbol_expr, old_lhs.type()))));
+}
+
+void remove_function_pointer(
+  goto_programt &goto_program,
+  goto_programt::targett target,
+  const std::set<symbol_exprt> &functions,
+  goto_modelt &goto_model)
+{
+  const code_function_callt &code = to_code_function_call(target->code);
+
+  const exprt &function = code.function();
+  PRECONDITION(function.id() == ID_dereference);
+
+  // this better have the right type
+  code_typet call_type = to_code_type(function.type());
+
+  // refine the type in case the forward declaration was incomplete
+  if(call_type.has_ellipsis() && call_type.parameters().empty())
+  {
+    call_type.remove_ellipsis();
+    forall_expr(it, code.arguments())
+      call_type.parameters().push_back(code_typet::parametert(it->type()));
+  }
+
+  const exprt &pointer = to_dereference_expr(function).pointer();
+
+  // the final target is a skip
+  goto_programt final_skip;
+
+  goto_programt::targett t_final = final_skip.add_instruction();
+  t_final->make_skip();
+
+  // build the calls and gotos
+
+  goto_programt new_code_calls;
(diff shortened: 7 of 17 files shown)


regression/goto-instrument/value-set-fi-fp-removal/test.c
```diff
@@ -0,0 +1,20 @@
...
+{
+}
+
+int main(void)
+{
+  fp_t fp = f;
+  fp();
+
+  // this would fool an analysis that looks for functions whose address is taken
+  fp_t other_fp = g;
```
hannes-steffenhagen-diffblue, line 19: Can we have more expansive tests for this? What about function pointers inside structs, inside arrays, or behind pointers? Also cases where the value set would be empty (with and without functions with matching signature) and cases where there should be multiple elements in the value set?

polgreen: Have added more tests for all these cases. When the value set is empty it defaults to the behaviour of the standard function pointer removal

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+#include <util/base_type.h>
+#include <util/c_types.h>
+#include <util/expanding_vector.h>
+#include <util/fresh_symbol.h>
+#include <util/namespace.h>
+#include <util/std_code.h>
+#include <util/union_find.h>
+
+#ifdef USE_STD_STRING
```
hannes-steffenhagen-diffblue, line 25: Not quite sure what this is for?

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+  // further pointer-related errors.
+  code_expressiont code_expression(function);
+  code_expression.add_source_location() = function.source_location();
+  target->code.swap(code_expression);
+  target->type = OTHER;
+}
+
+void value_set_fi_fp_removal(goto_modelt &goto_model)
+{
+  std::cout << "Doing FI value set analysis\n";
```
hannes-steffenhagen-diffblue: Should take a `messaget` and use that instead I’d think?

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+  goto_programt new_code;
+
+  // patch them all together
+  new_code.destructive_append(new_code_gotos);
+  new_code.destructive_append(new_code_calls);
+  new_code.destructive_append(final_skip);
+
+  // set locations
+  Forall_goto_program_instructions(it, new_code)
```
hannes-steffenhagen-diffblue: This could be done without macros:

```
for(auto const &instruction : new_code.instructions)
```

polgreen:
I think this is now ready to merge, is anyone available to review it?

smowton:
I'll get to this this weekend

polgreen:
Is anyone free to review this please?

regression/goto-instrument/value-set-fi-fp-removal/test.c
polgreen: @hannes-steffenhagen-diffblue I've added more tests, would you have chance to re-review this? Thanks

martin-cs:
@polgreen sorry for the lag, will try to look tomorrow.

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -906,7 +908,6 @@ int goto_instrument_parse_optionst::doit()
         "goto-instrument needs one input and one output file, aside from other "
         "flags");
     }
-
```
smowton: Accidental change?

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+  const code_typet::parameterst &function_parameters = code_type.parameters();
+
+  code_function_callt::argumentst &call_arguments = function_call.arguments();
+
+  for(std::size_t i = 0; i < function_parameters.size(); i++)
+  {
+    if(i < call_arguments.size())
+    {
+      if(!base_type_eq(
```
smowton: base_type_eq is deprecated, recommending to use == instead

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+           call_arguments[i].type(), function_parameters[i].type(), ns))
+      {
+        call_arguments[i] =
+          typecast_exprt(call_arguments[i], function_parameters[i].type());
```
smowton: Use `typecast_exprt::conditional_cast`

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+  if(function_call.lhs().is_nil())
+    return;
+
+  const namespacet ns(goto_model.symbol_table);
+
+  const code_typet &code_type =
+    to_code_type(ns.follow(function_call.function().type()));
+
+  // type already ok?
+  if(base_type_eq(function_call.lhs().type(), code_type.return_type(), ns))
```
smowton: Similarly, use `==`

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+  const code_typet &code_type =
+    to_code_type(ns.follow(function_call.function().type()));
+
+  const code_typet::parameterst &function_parameters = code_type.parameters();
+
+  code_function_callt::argumentst &call_arguments = function_call.arguments();
+
+  for(std::size_t i = 0; i < function_parameters.size(); i++)
+  {
+    if(i < call_arguments.size())
```
smowton, line 43: Add a comment mentioning how/why the call can have too few args?

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+    goto_programt::targett t4 = new_code_gotos.add_instruction();
+    t4->make_goto(
+      t1,
+      equal_exprt(
+        pointer, typecast_exprt::conditional_cast(address_of, pointer.type())));
+  }
+
+  // fall-through
+  // if(add_safety_assertion)
```
smowton: Remove or uncomment

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+  next_target++;
+
+  goto_program.destructive_insert(next_target, new_code);
+
+  // We preserve the original dereferencing to possibly catch
+  // further pointer-related errors.
+  code_expressiont code_expression(function);
+  code_expression.add_source_location() = function.source_location();
+  target->code.swap(code_expression);
+  target->type = OTHER;
```
smowton, line 185: what does an OTHER instruction even do? Maybe use an assign targeting a dummy symbol?

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+            {
+              const auto &od = to_object_descriptor_expr(address);
+              const auto &object = od.object();
+              if(object.type().id() == ID_code && object.id() == ID_symbol)
+                functions.insert(to_symbol_expr(object));
+            }
+          }
+
+          for(const auto &f : functions)
+            std::cout << "  function: " << f.get_identifier() << '\n';
```
smowton: Use a `messaget` not `cout`

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+          if(functions.size() > 0)
+            remove_function_pointer(
```
smowton, line 240: So if it has no ideas, we leave a function pointer behind? That seems problematic for downstream passes, probably better to turn it into an assume

polgreen: What do you mean turn it into an assume? 

The standard remove function pointers pass is run after this pass, so leaving the function pointer behind seems like a sensible thing to do because we then default to the standard behaviour.

smowton: Probably an assume false, but the default is fine too; in that case in the doc comment for the pass mention that it can leave function pointers behind and some other removal pass must run after it.

src/goto-instrument/value_set_fi_fp_removal.cpp
polgreen: An OTHER instruction is used for anything that isn't one of the other instruction types

An assign would be treated differently when symexed: OTHER will essentially be ignored, since it's not one of the specific cases handled by symex_other and ASSIGN will be handled by symex_assign, and we don't want this to be symexed.

Could just remove it since there currently isn't any code that consumes it at the moment, if that's preferred?

src/goto-instrument/value_set_fi_fp_removal.cpp
polgreen: assume false would potentially hide bugs

src/goto-instrument/value_set_fi_fp_removal.cpp
smowton: What's wrong with just executing it with an assign? The comment sounds like we do want to deref the pointer "for real" to obtain any side-effects of that action (i.e., a crash if it's null) as might be obtained from later passes adding pointer-sanity assertions, and that puts it on less shaky ground than using an OTHER instruction that is liable to be ignored in a way that an ASSIGN rhs probably won't be

src/goto-instrument/value_set_fi_fp_removal.cpp
smowton: Sorry should have said assert, but the fall-through is fine, happy to stick with that so long as the pass is documented as needing a subsequent cleanup pass to take care of residual function pointers

src/goto-instrument/value_set_fi_fp_removal.cpp
polgreen: How do later passes that add pointer-sanity assertions handle assigns? Or are these hypothetical future passes? Obviously we should make this work for existing passes, but I don't think it's obvious that a dummy variable is less easily ignored by a future pass writer than an "OTHER" instruction

smowton: OK never mind, the particular pass I was thinking about was `goto_check` and that appears to expect these instructions: https://github.com/diffblue/cbmc/blob/30244047f309d73bdeef4aec9d8bcb7342f82b73/src/analyses/goto_check.cpp#L1979

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1867,6 +1875,10 @@ void goto_instrument_parse_optionst::help()
     " --no-caching                 disable caching of intermediate results during transitive function inlining\n" // NOLINT(*)
     " --log <file>                 log in json format which code segments were inlined, use with --function-inline\n" // NOLINT(*)
     " --remove-function-pointers   replace function pointers by case statement over function calls\n" // NOLINT(*)
+    " --value-set-fi-fp-removal    build flow-insensitive value set and replace function pointers by a case statement\n" // NOLINT(*)
+    "                              over the possible assignments. If the set of possible assignments is empty the function pointer\n" // NOLINT(*)
+    "                              is removed using the standard remove-function-pointers pass. \n" // NOLINT(*)
+    " --remove-function-pointers   replace function pointers by case statement over all function calls with same signature\n" // NOLINT(*)
```
smowton: Option also documented on line 1877

src/goto-instrument/goto_instrument_parse_options.h
```diff
@@ -107,6 +108,8 @@ Author: Daniel Kroening, kroening@kroening.com
   "(show-threaded)(list-calls-args)" \
   "(undefined-function-is-assume-false)" \
   "(remove-function-body):"\
+  "(moderate-function-pointer-removal)"\
+  "(extreme-function-pointer-removal)"\
```
smowton: Options added but not referred to in the PR?

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,249 @@
...
+  const code_typet::parameterst &function_parameters = code_type.parameters();
+
+  code_function_callt::argumentst &call_arguments = function_call.arguments();
+
+  for(std::size_t i = 0; i < function_parameters.size(); i++)
+  {
+    // casting pointers might result in more arguments than function parameters
+    if(i < call_arguments.size())
+    {
+      if(call_arguments[i].type() != function_parameters[i].type())
```
smowton: Can remove this if statement now (conditional_cast does this)

smowton: To be clear, I mean that `conditional_cast(a, b)` expands to `return a.type() != b ? typecast(a, b) : a`, so there's no need to surround it in an additional `if(a.type() != b)`

polgreen:
I think all changes requested are now made

src/goto-instrument/value_set_fi_fp_removal.cpp
polgreen: that's outdated, it's changed in the latest commit

src/goto-instrument/value_set_fi_fp_removal.cpp
polgreen: Needed for one of the CI tests, which compiles with this flag and needs this include to compile

src/goto-instrument/value_set_fi_fp_removal.cpp
polgreen: Would need to drop the const in that.

Are the macros now deprecated? Because couldn't everything that is written using the macro also be written without?

martin-cs:
@polgreen I will review next.

regression/goto-instrument/value-set-fi-fp-removal/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+test.c
+--value-set-fi-fp-removal
+^EXIT=0$
+^SIGNAL=0$
+^  function: f$
+--
+^  function: g$
```
martin-cs, line 8: Negative as well as positive tests are good.

regression/goto-instrument/value-set-fi-fp-removal4/test.c
```diff
@@ -0,0 +1,21 @@
...
+}
+
+void g()
+{
+}
+
+int main(void)
+{
+  fp_t fp;
+  fp();
```
martin-cs: This looks like undefined behaviour to me.  Is there any way to test this functionality without triggering UB?

regression/goto-instrument/value-set-fi-fp-removal5/test.c
```diff
@@ -0,0 +1,20 @@
...
+}
+
+void g(int y)
+{
+}
+
+int main(void)
+{
+  fp_t fp;
+  fp();
```
martin-cs: Again, anything that happens after this seems a bit dubious to me.

regression/goto-instrument/value-set-fi-fp-removal5/test.c
```diff
@@ -0,0 +1,20 @@
...
+
+  // the value set is empty, defaults to standard function pointer removal behaviour
```
martin-cs, line 15: ( If it is an over-approximation it shouldn't be empty here but that is a separate issue. )
Thanks for documenting this as it is an unexpected (but not necessarily unreasonable) behaviour.

src/goto-instrument/value_set_fi_fp_removal.cpp
martin-cs: They have been deprecated for some time now in favour of the new-style `for` loops.  As to why...

src/goto-instrument/value_set_fi_fp_removal.cpp
```diff
@@ -0,0 +1,247 @@
...
+  const symbol_exprt tmp_symbol_expr = tmp_symbol.symbol_expr();
+
+  exprt old_lhs = function_call.lhs();
+  function_call.lhs() = tmp_symbol_expr;
+
+  dest.add(goto_programt::make_assignment(
+    code_assignt(old_lhs, typecast_exprt(tmp_symbol_expr, old_lhs.type()))));
+}
+
+void remove_function_pointer(
```
martin-cs, line 89: It is good to see this bit of functionality broken out.

martin-cs (approved):
I have Opinions on how function pointer removal should be done but I will spare you those.  I think this is a step in the right direction.  Thank you especially for the detailed test cases.

src/goto-instrument/value_set_fi_fp_removal.cpp
polgreen: If the macros are deprecated, perhaps they should be marked as deprecated in the goto_program.h?

regression/goto-instrument/value-set-fi-fp-removal5/test.c
polgreen: Shouldn't it be empty? fp is never assigned to?

src/pointer-analysis/value_set_fi.cpp
```diff
@@ -1197,6 +1197,9 @@ void value_set_fit::assign_rec(
   {
     const typet &type = to_index_expr(lhs).array().type();
 
+    if(type.id() != ID_array)
```
peterschrammel: ❓ This makes the second part of the condition in the data invariant below unreachable.

smowton (approved):

regression/goto-instrument/value-set-fi-fp-removal5/test.c
martin-cs: `fp` is never assigned to so frankly; it could be anything, so I think it should be \top.  Whether or not it is, is not an issue for your PR though.

src/goto-instrument/value_set_fi_fp_removal.cpp
martin-cs: There was a move to simply replace them all in one go which would have basically caused a number of forks, I think the compromise was the linter would flag them in new code and they would be deprecated 'at some point'.  I guess the linter doesn't so... yeah, yeah it would be a good idea.

regression/goto-instrument/value-set-fi-fp-removal4/test.c
polgreen: it's testing for occasions when the value set is empty, so that should be the case is if the initialisation code is skipped, which is then undefined behaviour

martin-cs: Fair enough then.

src/pointer-analysis/value_set_fi.cpp
polgreen: I've removed this from the PR

regression/goto-instrument/value-set-fi-fp-removal5/test.c
polgreen: i've flipped the order of the delcarations now

polgreen:
I've now addressed all of the comments in the reviews

martin-cs:
Thanks @polgreen !  @peterschrammel and @hannes-steffenhagen-diffblue do you have further comments or can this be merged as it is?

PR #5437, opened 2020-07-29
Fix CMake build on Windows

NlightNFotis:
Attempting to build `cbmc` with `cmake` on `Visual Studio 2019/Win 10` fails. 

This PR introduces some needed changes to two `CMakeLists.txt` files to make
it work.

### DIFF
--- a/jbmc/CMakeLists.txt
+++ b/jbmc/CMakeLists.txt
@@ -4,7 +4,7 @@ add_subdirectory(unit)
 
 add_custom_target(java-models-library ALL
     COMMAND mvn --quiet -Dmaven.test.skip=true package
-    COMMAND cp target/core-models.jar ${CMAKE_CURRENT_SOURCE_DIR}/src/java_bytecode/library/
+    COMMAND ${CMAKE_COMMAND} -E copy target/core-models.jar ${CMAKE_CURRENT_SOURCE_DIR}/src/java_bytecode/library/
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/lib/java-models-library
 )
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,7 +1,7 @@
 project(CBMC)
 
-find_package(BISON)
-find_package(FLEX)
+find_package(BISON REQUIRED)
+find_package(FLEX REQUIRED)
 
 find_package(Doxygen)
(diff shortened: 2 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 project(CBMC)
 
-find_package(BISON)
-find_package(FLEX)
+find_package(BISON REQUIRED)
```
hannes-steffenhagen-diffblue, line 3: 👍 it's a non-optional dependency and absolutely should be marked as `REQUIRED`

jbmc/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ add_subdirectory(unit)
 
 add_custom_target(java-models-library ALL
     COMMAND mvn --quiet -Dmaven.test.skip=true package
-    COMMAND cp target/core-models.jar ${CMAKE_CURRENT_SOURCE_DIR}/src/java_bytecode/library/
+    COMMAND ${CMAKE_COMMAND} -E copy target/core-models.jar ${CMAKE_CURRENT_SOURCE_DIR}/src/java_bytecode/library/
```
hannes-steffenhagen-diffblue, line 7: Yep, we should avoid using unix-y commands in our... commands.

I'd recommend putting all the variable stuff in `"` quotes as well... I don't actually know off the top of my head how CMake handles variable expansions with whitespace in them.

jbmc/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ add_subdirectory(unit)
 
 add_custom_target(java-models-library ALL
     COMMAND mvn --quiet -Dmaven.test.skip=true package
```
hannes-steffenhagen-diffblue, line 6: Not an action for this PR, but we should also avoid hardcoding the maven executable. By adding a configuration for this we could support cases where people want to build with `~/.local/bin/maven-experimental` or whatever.

hannes-steffenhagen-diffblue (approved):

jbmc/CMakeLists.txt
chrisr-diffblue: It's no worse than the _current_ situation so I'm happy for this to be merged as is - but It probably is worth trying a build on a checkout of CBMC that has spaces in its file path. On Mac/Windows it's not _that_ uncommon for this to happen.

chrisr-diffblue (approved):

PR #5438, opened 2020-07-30
Fix Bison grammar issues preventing us from parsing CoreFoundation

hannes-steffenhagen-diffblue:

There are currently two issues preventing us from parsing Apple's CoreFoundation framework: For one, we still don't quite parse underlying enum types correctly, so for example

    enum A : unsigned int { AV };

and 

    enum B : uint32_t { BV };

would previously fail to parse, and we didn't allow gcc-style attributes after a functions parameter list:

    int some_function(int argument) __attribute__((this_previously_wasn't_allowed));

which gcc itself doesn't permit either but is valid in Clang's C dialect, and used this way by Apple across their frameworks for their `API_AVAILABLE` and `API_UNAVAILABLE` macros.

This fixes both of these issues.

### DIFF
--- a/regression/ansi-c/function_with_postfix_attributes/test.desc
+++ b/regression/ansi-c/function_with_postfix_attributes/test.desc
@@ -1,0 +1,11 @@
+CORE gcc-only
+test.c
+
+^EXIT=0$
+^SIGNAL=0$
+--
+^PARSING ERROR$
+--
+gcc doesn't support postfix attributes but clang does, and Apples CoreFoundation framework makes use of this for availability macros.
+
+Testing both the inline and non-inline variants because these are different cases in our parser.
--- a/src/ansi-c/parser.y
+++ b/src/ansi-c/parser.y
@@ -1837,13 +1837,21 @@ enum_name:
         }
         ;
 
+basic_type_name_list:
+    basic_type_name
+  | basic_type_name_list basic_type_name
+
+enum_underlying_type:
+    basic_type_name_list
+  | typedef_name
+
 enum_underlying_type_opt:
         /* empty */
         {
           init($$);
           parser_stack($$).make_nil();
         }
-        | ':' basic_type_name
+        | ':' enum_underlying_type
         {
           $$=$2;
         }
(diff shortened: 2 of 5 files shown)


danpoe (approved):

regression/ansi-c/function_with_postfix_attributes/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
```
thomasspriggs: :exclamation: Your test is broken because you committed a `test.c` but this file is specifying `main.c`

regression/ansi-c/function_with_postfix_attributes/test.desc
```diff
@@ -0,0 +1,11 @@
...
+
+^EXIT=0$
+^SIGNAL=0$
+--
+^PARSING ERROR$
+--
+gcc doesn't support postfix attributes but clang does, and Apples CoreFoundation framework makes use of this for availability macros.
+
+Testing both the inline and non-inline variants because these are different cases in our parser.
```
thomasspriggs: :pick: Missing newline.

src/ansi-c/parser.y
```diff
@@ -1837,13 +1837,21 @@ enum_name:
         }
         ;
 
+basic_type_name_list:
+    basic_type_name
+  | basic_type_name_list basic_type_name
```
thomasspriggs, line 1842: :information_source: We probably want something like the following, in order to actually build the list - 
```
  | basic_type_name_list basic_type_name
  {
    $$=merge($1, $2);
  }
  ;
```
I have added this :information_source: comment as a note about future work which will likely be needed eventually. This is not a blocking comment.

hannes-steffenhagen-diffblue: For onlookers: actually fully supporting this feature is in the pipeline, just getting this. to parse is higher priority though so it comes first.

thomasspriggs (approved):

chrisr-diffblue (approved):
My Yacc skills are quite rusty.... but I think this looks OK to me.

Issue #5272, opened 2020-03-16
goto-cc isn't aware of cbmc's --object-bits setting


karkhaz:
This seems to be a CBMC bug introduced by [PR 5210](https://github.com/diffblue/cbmc/pull/5210).

The issue exhibits when running the [s2n proof for `stuffer_resize_if_empty`](https://github.com/awslabs/s2n/pull/1408). The command is:
```
cbmc --verbosity 4  --bounds-check --div-by-zero-check --float-overflow-check --nan-check --pointer-check --pointer-overflow-check --signed-overflow-check --undefined-shift-check --unsigned-overflow-check --unwind 1 --unwinding-assertions  --object-bits 6 --trace /Users/karkhaz/code/s2n/tests/cbmc/proofs/s2n_stuffer_resize_if_empty/gotos/s2n_stuffer_resize_if_empty_harness.goto 2>&1 | tee  logs/cbmc.log; if [ ${PIPESTATUS[0]} -ne 10 ]; then exit ${PIPESTATUS[0]}; fi
```
which emits this warning:
```
file <built-in-additions> line 24: warning: conflicting initializers for variable "__CPROVER_max_malloc_size"
using old value in module abort_override_assert_false file <built-in-additions> line 24
8388608ul
ignoring new value in module <built-in-library> file <built-in-additions> line 24
33554432ul
```
The problem is that we're running CBMC with `--object-bits 6`, which sets `max_malloc_size` to `8399608ul` (`0x800 000`). However, when goto-cc built one of the files, it had set `max_malloc_size` to `33554432ul` (`0x2 000 000`), which is what it would be if object-bits were 8. We can confirm this by running CBMC with `--object-bits 10`, in which case we see the opposite problem:
```
file <built-in-additions> line 24: warning: conflicting initializers for variable "__CPROVER_max_malloc_size"
using old value in module abort_override_assert_false file <built-in-additions> line 24
8388608ul
ignoring new value in module <built-in-library> file <built-in-additions> line 24
2097152ul
```

I will next try to reduce this to a minimal example.

hannes-steffenhagen-diffblue:
This is internally tracked as ADA-568

thomasspriggs:
This issue can now be avoided by passing `--object-bits` to `goto-cc`. This was done in this PR - https://github.com/diffblue/cbmc/pull/5440

PR #5440, opened 2020-07-31
Add `--object-bits` option to goto-cc

thomasspriggs:
So that we can support running with non-default settings of `--object-bits` in `goto-cc`. This should address the issue seen in https://github.com/diffblue/cbmc/issues/5272


regression/goto-cc-cbmc-shared-options/object-bits/test.c
```diff
@@ -0,0 +1,30 @@
+#include <assert.h>
+#include <stdlib.h>
+
+extern __CPROVER_size_t __CPROVER_malloc_size;
+
+size_t
+find_first_set(const size_t max_malloc_size, const size_t bits_accumulator)
+{
+  // Implemented using recursion to avoid excessive unwinding or specifying an
+  // unwind limit.
```
hannes-steffenhagen-diffblue: how does recursion stop excessive unwinding?

src/goto-cc/gcc_cmdline.cpp
```diff
@@ -28,6 +28,7 @@ const char *goto_cc_options_with_separated_argument[]=
   "--native-linker",
   "--print-rejected-preprocessed-source",
   "--mangle-suffix",
+  "--object-bits",
```
hannes-steffenhagen-diffblue, line 31: Shouldn't this be added to the other `cmdlinet`s as well?

regression/goto-cc-cbmc-shared-options/object-bits/default_bits.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+test.c
+--function main
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+assertion object_bits != 8: FAILURE
```
hannes-steffenhagen-diffblue, line 8: can you add the expected results for the other assertions in here as well? Clearly it's supposed to be always 2 successes and one failure but that might be easier to see if its spelt out.

hannes-steffenhagen-diffblue (commented):
So adding the option to the cmdlinet was all that was needed to make this work? Neat.

regression/goto-cc-cbmc-shared-options/object-bits/test.c
thomasspriggs: I discovered this by experimentation. The same algorithm written as a loop spent a long time unwinding if no unwind limit was set. The recursive version only unwinds a sufficient number of times for the value of `max_malloc_size` which is actually specified.

I am not familiar with the unwinding code. However I would hazard a guess that loop unwinding happens early on, using a only a local analysis. Whereas recursion unwinding happens later as part of symmex. The constant being passed to the function is been propagated in this case, leading to unwinding the exact number of times required.

src/goto-cc/gcc_cmdline.cpp
thomasspriggs: The original ticket referred to `goto-cc` only, fixing `goto-cc` only will get a fix ready for use sooner. Yes, the same problem and fix probably applies to the other entry points as well. The time consuming part of fixing the other entry points would be writing tests for each of them.

NlightNFotis (approved):

regression/goto-cc-cbmc-shared-options/object-bits/default_bits.desc
thomasspriggs: Done.

regression/goto-cc-cbmc-shared-options/object-bits/test.c
```diff
@@ -0,0 +1,30 @@
...
+  if(max_malloc_size & 1)
+    return bits_accumulator;
+  return find_first_set(max_malloc_size >> 1, bits_accumulator + 1);
+}
+
+size_t calculate_object_bits()
```
danpoe, line 12: IMO it would be unnecessary to reverse engineer the object bits here or even use the value of `__CPROVER_max_malloc_size`. It's probably sufficient to just run `cbmc` with a non-default object bits setting after `goto-cc` and check that it suceeds.

danpoe (commented):
I think for now this solution is fine. At some point though, the `__CPROVER_max_malloc_size` variable will go away anyways (when ADA-576 is implemented) and the conflict will disappear. Then `goto-cc` should be able to be oblivious to the object bits setting again.

The reason for `__CPROVER_max_malloc_size` going away is that it is currently used to check that not more memory is allocated than can be addressed by cbmc. However, it turns out that a better way to implement this is to check for pointer overflows instead (see discussion on #5401).

regression/goto-cc-cbmc-shared-options/object-bits/test.c
thomasspriggs: The reason I chose to do it this way is that the previous incorrect behaviour resulted in a warning, but might not actually cause a failure without giving an input program which depends on the implications of the `object-bits` configuration. I could have used a negative regex to check that the warning message does not appear in the output. However this would not be a robust test because if the text of the message were to be updated then the corresponding test could easily be missed. This may be a little over-engineered but I am happy that this is a robust test and it is clear what the positive regexes are actually checking for.

regression/goto-cc-cbmc-shared-options/object-bits/test.c
thomasspriggs: I haven't been able to reproduce the problem I saw with the iterative version and I never committed that version, so I can't get it back even with `git reflog`. Therefore I am going to assume there is no problem with unwinding loops / constant propagation. I shall delete this comment in order to avoid being misleading, but leave the recursive version in place because it doesn't seem worth rewriting this.

src/goto-cc/gcc_cmdline.cpp
thomasspriggs: There is no standard list of options in the base class. So it would need to be manually added and tested for each `cmdlinet` in the heirarchy.

chrisr-diffblue (commented):
Have you tested what happens with linking? I'm thinking of the case where you compile one translation unit with one value for object-bits, and another translation unit with a different value for object-bits. How does the linker handle that? (Hopefully with a nice user-friendly error message... :-) )

regression/goto-cc-cbmc-shared-options/object-bits/test.c
danpoe: Yes, that makes sense.

danpoe (approved):

thomasspriggs:
@chrisr-diffblue If I feed two different goto binaries into cbmc which were generated using different `--object-bits` configurations then it generates appropriate warning messages -
```
$ cbmc ./temp_func.gb ./temp_main.gb
CBMC version 5.12 (cbmc-5.12.4-9-gdae95d970-dirty) 64-bit x86_64 linux
Reading: ./temp_func.gb
Reading: ./temp_main.gb
file <built-in-additions> line 24: warning: conflicting initializers for variable "__CPROVER_max_malloc_size"
using old value in module temp_func file <built-in-additions> line 24
9007199254740992ul
ignoring new value in module temp_main file <built-in-additions> line 24
144115188075855872ul
Generating GOTO Program
file <built-in-additions> line 24: warning: conflicting initializers for variable "__CPROVER_max_malloc_size"
using old value in module temp_func file <built-in-additions> line 24
9007199254740992ul
ignoring new value in module <built-in-library> file <built-in-additions> line 24
36028797018963968ul
```

chrisr-diffblue (approved):
I'm glad theres a diagnostic, thanks for checking/demonstrating. It's probably an open question whether that should be an error rather than a warning, but that doesn't have to be fixed in this PR, so feel free to raise a Github bug ticket to cover that. It'd be super nice if there was a test case to demonstrate the diagnostic gets triggered (just so we don't end up accidentally losing that at some point :-) ) but again, I won't block the PR just on that.

PR #5441, opened 2020-08-03
Correctly interpret underlying type specification of enums

danpoe:

### DIFF
--- a/regression/cbmc/enum_underlying_type_02/main.c
+++ b/regression/cbmc/enum_underlying_type_02/main.c
@@ -1,0 +1,10 @@
+#include <assert.h>
+
+enum enum1 : unsigned char
+{
+  E1 = 256
+};
+
+int main()
+{
+}
--- a/src/ansi-c/c_typecheck_type.cpp
+++ b/src/ansi-c/c_typecheck_type.cpp
@@ -1160,10 +1160,24 @@ void c_typecheck_baset::typecheck_c_enum_type(typet &type)
     throw 0;
   }
 
-  if(as_expr.find(ID_enum_underlying_type).is_not_nil())
+  const bool have_underlying_type =
+    type.find_type(ID_enum_underlying_type).is_not_nil();
+
+  if(have_underlying_type)
   {
-    warning().source_location = source_location;
-    warning() << "ignoring specification of underlying type for enum" << eom;
+    typecheck_type(type.add_type(ID_enum_underlying_type));
+
+    const typet &underlying_type =
+      static_cast<const typet &>(type.find(ID_enum_underlying_type));
+
+    if(!is_signed_or_unsigned_bitvector(underlying_type))
+    {
+      std::ostringstream msg;
+      msg << source_location << ": non-integral type '"
+          << underlying_type.get(ID_C_c_type)
+          << "' is an invalid underlying type";
+      throw invalid_source_file_exceptiont{msg.str()};
+    }
   }
 
   // enums start at zero;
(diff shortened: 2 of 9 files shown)


src/ansi-c/c_typecheck_type.cpp
```diff
@@ -1160,10 +1160,20 @@ void c_typecheck_baset::typecheck_c_enum_type(typet &type)
...
 
-  if(as_expr.find(ID_enum_underlying_type).is_not_nil())
+  const bool have_underlying_type =
+    type.find_type(ID_enum_underlying_type).is_not_nil();
+
+  if(have_underlying_type)
   {
-    warning().source_location = source_location;
-    warning() << "ignoring specification of underlying type for enum" << eom;
+    typecheck_type(type.add_type(ID_enum_underlying_type));
```
hannes-steffenhagen-diffblue, line 1168: Using `add_something` to refer to something that's already there is pretty oof.

regression/cbmc/enum_underlying_type_02/main.c
```diff
@@ -0,0 +1,11 @@
+#include <assert.h>
+
+enum enum1 : unsigned char
+{
+  E1 = 256
```
hannes-steffenhagen-diffblue, line 5: Good test case to include!

src/ansi-c/c_typecheck_type.cpp
```diff
@@ -1160,10 +1160,20 @@ void c_typecheck_baset::typecheck_c_enum_type(typet &type)
...
+
+    if(!is_signed_or_unsigned_bitvector(
+         static_cast<const typet &>(type.find(ID_enum_underlying_type))))
+    {
+      error().source_location = source_location;
+      error() << "underlying type for enum must be an integral type" << eom;
+      throw 0;
```
chrisr-diffblue: Do we not have a structured exception to throw here? (and in the other cases below)

chrisr-diffblue (commented):
Looks good, and thanks for the comprehensive test cases. A couple of minor nitpicks really:
1) Error handling - we put quite a big effort into cleaning up error handling, adding structured exceptions, things like that - so it would be good to avoid the `throw 0` if possible.
2) What error messages does clang emit for the "non-integral underlying type" and "underlying type too narrow for constant" cases? It would be good to make our error messages broadly similar just for ease of users googling for help if they hit them, etc.

src/ansi-c/c_typecheck_type.cpp
hannes-steffenhagen-diffblue: Yes this should an `invalid_source_file_exceptiont`

hannes-steffenhagen-diffblue:
@chrisr-diffblue Current errors for these cases in clang look like this:

    test.c:1:10: error: non-integral type 'float' is an invalid underlying type

and

    test.c:1:24: error: enumerator value is not representable in the underlying type 'signed char'

I think the existing error messages we have here are sufficiently close to that. We could include some extra information like what we think the underlying type is and why it doesn't fit and such, but I wouldn't call this a blocker.

hannes-steffenhagen-diffblue:
I'm OK with these changes though I'd also prefer if we didn't proliferate the `throw 0` thing.

hannes-steffenhagen-diffblue (approved):

chrisr-diffblue:
> @chrisr-diffblue Current errors for these cases in clang look like this:
> 
> ```
> test.c:1:10: error: non-integral type 'float' is an invalid underlying type
> ```
> 
> and
> 
> ```
> test.c:1:24: error: enumerator value is not representable in the underlying type 'signed char'
> ```
> 
> I think the existing error messages we have here are sufficiently close to that. We could include some extra information like what we think the underlying type is and why it doesn't fit and such, but I wouldn't call this a blocker.

Thanks @hannes-steffenhagen-diffblue - @danpoe If it's not too much of a pain, could you model the error messages on the above? In particular, showing the user the type like that its quite helpful to them.

src/ansi-c/c_typecheck_type.cpp
danpoe: Done

danpoe:
> > @chrisr-diffblue Current errors for these cases in clang look like this:
> > ```
> > test.c:1:10: error: non-integral type 'float' is an invalid underlying type
> > ```
> > 
> > 
> > and
> > ```
> > test.c:1:24: error: enumerator value is not representable in the underlying type 'signed char'
> > ```
> > 
> > 
> > I think the existing error messages we have here are sufficiently close to that. We could include some extra information like what we think the underlying type is and why it doesn't fit and such, but I wouldn't call this a blocker.
> 
> Thanks @hannes-steffenhagen-diffblue - @danpoe If it's not too much of a pain, could you model the error messages on the above? In particular, showing the user the type like that its quite helpful to them.

Ok, changed the error messages to a text similar to the clang messages and including the type.

chrisr-diffblue (approved):
Thanks for addressing my nitpicks :-)

PR #5442, opened 2020-08-03
Refactoring around the goto-cc entry point.

thomasspriggs:
As I was working on https://github.com/diffblue/cbmc/pull/5440, I found the entry point code less than straight forward to follow. I carried out this refactorings locally in order to make it easier for me to understand. I am raising this separate PR containing the refactoring work only as it was not required in order to implement the feature in the original PR, but should still improve maintainability.

- [x] N/A - No user visible behaviour changes. ~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~
- [x] N/A - None claimed ~My commit message includes data points confirming performance improvements (if claimed).~

### DIFF
--- a/src/goto-cc/armcc_cmdline.cpp
+++ b/src/goto-cc/armcc_cmdline.cpp
@@ -265,6 +265,20 @@ static const char *options_with_arg[]=
   nullptr
 };
 
+bool prefix_in_list(const char *option, const char **list, std::string &prefix)
+{
+  for(std::size_t i = 0; list[i] != nullptr; i++)
+  {
+    if(strncmp(option, list[i], strlen(list[i])) == 0)
+    {
+      prefix = std::string(list[i]);
+      return true;
+    }
+  }
+
+  return false;
+}
+
 bool armcc_cmdlinet::parse(int argc, const char **argv)
 {
   for(int i=1; i<argc; i++)
--- a/src/goto-cc/armcc_cmdline.cpp
+++ b/src/goto-cc/armcc_cmdline.cpp
@@ -12,9 +12,13 @@ Author: Daniel Kroening
 #include "armcc_cmdline.h"
 
 #include <util/optional.h>
+#include <util/prefix.h>
 
+#include <algorithm>
 #include <cstring>
 #include <iostream>
+#include <string>
+#include <vector>
 
 /// parses the command line options into a cmdlinet
 /// \par parameters: argument count, argument strings
@@ -197,7 +201,8 @@ static const char *options_no_arg[]=
   nullptr
 };
 
-static const char *options_with_prefix[]=
+// clang-format off
+static const std::vector<std::string> options_with_prefix
 {
   "--project=",
   "--workdir=",
@@ -243,11 +248,10 @@ static const char *options_with_prefix[]=
   "--configure_sysroot=",
   "--configure_cpp_headers=",
   "--configure_extra_includes=",
-  "--configure_extra_libraries=",
-  nullptr
+  "--configure_extra_libraries="
 };
 
-static const char *options_with_arg[]=
+static const std::vector<std::string> options_with_arg
 {
   // goto-cc specific
   "--verbosity",
@@ -263,21 +267,20 @@ static const char *options_with_arg[]=
   "-Warmcc,",
   "-o",
   "--cpu",
-  "--apcs",
-  nullptr
+  "--apcs"
 };
+// clang-format on
 
-optionalt<std::string> prefix_in_list(const char *option, const char **list)
+optionalt<std::string>
+prefix_in_list(const std::string &option, const std::vector<std::string> &list)
 {
-  for(std::size_t i = 0; list[i] != nullptr; i++)
-  {
-    if(strncmp(option, list[i], strlen(list[i])) == 0)
-    {
-      return {list[i]};
-    }
-  }
-
-  return {};
+  const auto found =
+    std::find_if(list.cbegin(), list.cend(), [&](const std::string &argument) {
+      return has_prefix(argument, option);
+    });
+  if(found == list.cend())
+    return {};
+  return {*found};
 }
 
 bool armcc_cmdlinet::parse(int argc, const char **argv)
--- a/src/goto-cc/armcc_cmdline.cpp
+++ b/src/goto-cc/armcc_cmdline.cpp
@@ -11,6 +11,8 @@ Author: Daniel Kroening
 
 #include "armcc_cmdline.h"
 
+#include <util/optional.h>
+
 #include <cstring>
 #include <iostream>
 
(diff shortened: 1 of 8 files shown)


src/goto-cc/armcc_cmdline.cpp
```diff
@@ -265,6 +265,21 @@ static const char *options_with_arg[]=
   nullptr
 };
 
+static bool
+prefix_in_list(const char *option, const char **list, std::string &prefix)
```
hannes-steffenhagen-diffblue: I think we could simplify a lot of this if we stored the args in the mode as `std::vector<std::string>` instead of `char **`. Then we could also just use regular `std::find` or something like that instead of this.

src/goto-cc/armcc_cmdline.cpp
```diff
@@ -14,6 +14,8 @@ Author: Daniel Kroening
 #include <cstring>
 #include <iostream>
 
+#include "optional.h"
```
hannes-steffenhagen-diffblue: should be `<util/optional.h>`

src/goto-cc/armcc_cmdline.cpp
```diff
@@ -265,19 +267,18 @@ static const char *options_with_arg[]=
   nullptr
 };
 
-static bool
-prefix_in_list(const char *option, const char **list, std::string &prefix)
+static optionalt<std::string>
```
hannes-steffenhagen-diffblue: Like I said above, I'd prefer not touching this like this for now and instead changing how we store command line arguments, then we don't have to invent our own finds at all.

src/goto-cc/goto_cc_cmdline.cpp
```diff
@@ -117,3 +117,13 @@ void goto_cc_cmdlinet::add_infile_arg(const std::string &arg)
     fclose(tmp);
   }
 }
+
+bool goto_cc_cmdlinet::have_infile_arg() const
```
hannes-steffenhagen-diffblue, line 122: I think we should probably just make sure to put definitions in cpp files in general.

src/util/bv_encoding.h
```diff
@@ -0,0 +1,23 @@
+/*******************************************************************\
+
+Module: bv_encoding configuration
+
+Author: Diffblue Ltd
+
+\*******************************************************************/
```
hannes-steffenhagen-diffblue: I think I mentioned this before but we want to move away from this style of module header... we might have to adjust our linter for that though.

src/util/bv_encoding.h
```diff
@@ -0,0 +1,23 @@
...
+
+#ifndef CPROVER_UTIL_BV_ENCODING_H
+#define CPROVER_UTIL_BV_ENCODING_H
+
+#include <cstddef>
+
+struct bv_encodingt
```
hannes-steffenhagen-diffblue: Moves it into a different namespace though

src/util/bv_encoding.h
```diff
@@ -11,7 +11,7 @@ Author: Diffblue Ltd
 
 #include <cstddef>
 
-struct bv_encodingt
+struct bv_encodingt final
```
hannes-steffenhagen-diffblue: I'm not sure that's necessary, I find `final` to be a bit noisy. To me `final` actually suggests the opposite of what you're saying here: This *is* part of a class hierarchy, it's just supposed to be at the bottom of it.

src/util/bv_encoding.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "bv_encoding.h"
+
+#include "exception_utils.h"
+#include "string2int.h"
+
+bv_encodingt parse_object_bits_encoding(
+  const std::string &argument,
+  const std::size_t pointer_width)
+{
+  const unsigned int object_bits = unsafe_string2unsigned(argument);
```
hannes-steffenhagen-diffblue: Use `string2optional` here so you can handle a parse error with a proper error message. The `unsafe_` functions are for when you know the parse will succeed.

thomasspriggs: Yep. A non-integer argument will cause an invariant violation, both before and after this PR. This is expected as the PR is intended to be refactoring only, not bug fixing. I agree that we should improve the error handling here.

src/util/bv_encoding.h
thomasspriggs: You know that isn't part of a hierarchy from reading this line of code, because no base classes are specified. Having it specified as `final` also means that you are always safe to pass / return / assign instances of this class by value, without risk of slicing occurring. I do take your point about noise however. I think I may drop this commit as I don't really want to include anything in this PR, where the pros and cons could be argued either way. Suffice to say that my current habit is to make new classes/structs `final` by default, in the same way as I would make variables and member functions `const` by default.

src/goto-cc/goto_cc_cmdline.cpp
thomasspriggs: :+1:

src/goto-cc/armcc_cmdline.cpp
thomasspriggs: That sounds like a good follow up refactoring.

src/goto-cc/armcc_cmdline.cpp
thomasspriggs: Yes, I think that using a standard container would be a great follow-up refactoring. However I would argue that removing the need for an output parameter is already a step in the right direction, which makes the usages of this function more readable.

src/goto-cc/armcc_cmdline.cpp
thomasspriggs: Done.

src/goto-cc/armcc_cmdline.cpp
thomasspriggs: Fixed.

src/goto-cc/armcc_cmdline.cpp
thomasspriggs: Done.

src/util/bv_encoding.h
thomasspriggs: I have removed this commit from this PR.

src/util/bv_encoding.h
thomasspriggs: Ok. I'll leave it where it was then.

unit/goto-cc/armcc_cmdline.cpp
```diff
@@ -0,0 +1,32 @@
+// Copyright 2020 Diffblue Limited. All Rights Reserved.
```
chrisr-diffblue: I think the licence header is wrong - thats the internal Diffblue header - CBMC contributions require a different header.

chrisr-diffblue (commented):
All looks good apart from the copyright header - happy to approve once that is fixed up.

chrisr-diffblue (approved):

PR #5446, opened 2020-08-10
Github action testing job for cbmc under VS2019

NlightNFotis:
This PR is introducing testing of cbmc and jbmc on Windows CI, under VS2019.

A number of tests that couldn't immediately be made to work under windows,
have been excluded from the tests run by addition of an additional label, `winbug`.

These tests will be revisited in the future, for further exploration of the mechanisms
that were making them fail under windows, and they are not in scope of this PR,
as the scope was limited in getting cbmc to perform a test run under the targeted
configuration.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -39,6 +39,7 @@ jobs:
           make -C regression/cbmc test-paths-lifo
           env PATH=$PATH:`pwd`/src/solvers make -C regression/cbmc test-cprover-smt2
           make -C jbmc/regression test
+
   check-ubuntu-20_04-cmake-gcc:
     runs-on: ubuntu-20.04
     steps:
@@ -111,13 +112,11 @@ jobs:
         run: cd build; ninja
       - name: Run CTest
         run: cd build; ctest -V -L CORE .
-  check-vs-2019-build-only:
+
+  check-vs-2019-build-and-test:
     runs-on: windows-2019
     env:
       SCRIPT_DIR: .github/workflows/vs2019
-    defaults:
-      run:
-        shell: cmd
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v2
@@ -125,10 +124,13 @@ jobs:
           submodules: recursive
 
       - name: Install bison and flex
-        run: "%SCRIPT_DIR%\\install-bison.bat"
+        run: "${{env.SCRIPT_DIR}}\\install-bison.bat"
 
       - name: Build cbmc
-        run: "%SCRIPT_DIR%\\build-cbmc.bat"
+        run: "${{env.SCRIPT_DIR}}\\build-cbmc.bat"
+
+      - name: Test cbmc
+        run: "${{env.SCRIPT_DIR}}\\run-tests.bat"
 
   check-clang-format:
     runs-on: ubuntu-20.04
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -126,6 +126,16 @@ jobs:
       - name: Install bison and flex
         run: "${{env.SCRIPT_DIR}}\\install-bison.bat"
 
+      - name: Install z3
+        run: |
+          Invoke-WebRequest "https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-win.zip" -OutFile z3.zip
+          Expand-Archive z3.zip
+          # We need Get-ChildItem.Name here because the path includes
+          # the z3 version number, which we expect to change.
+          $z3_path = "$(Get-Location)\z3\$((Get-ChildItem z3).Name)"
+          $env:PATH += ";$z3_path"
+          Write-Output "::set-env name=PATH::$env:PATH"
+
       - name: Build cbmc
         run: "${{env.SCRIPT_DIR}}\\build-cbmc.bat"
 
--- a/.github/workflows/vs2019/build-cbmc.bat
+++ b/.github/workflows/vs2019/build-cbmc.bat
@@ -16,7 +16,7 @@ echo "Remove strawberry"
 del /s /q "c:/Strawberry/c/bin"
 
 echo "Configure CBMC with cmake"
-cmake -S. -Bbuild -GNinja
+cmake --config Release -S. -Bbuild -GNinja
 
 echo "Build CBMC with cmake"
 cmake --build build
--- a/.github/workflows/vs2019/build-cbmc.bat
+++ b/.github/workflows/vs2019/build-cbmc.bat
@@ -16,7 +16,7 @@ echo "Remove strawberry"
 del /s /q "c:/Strawberry/c/bin"
 
 echo "Configure CBMC with cmake"
-cmake --config Release -S. -Bbuild -GNinja
+cmake -S. -Bbuild
 
 echo "Build CBMC with cmake"
-cmake --build build
+cmake --build build --config Release
--- a/.github/workflows/vs2019/run-tests.bat
+++ b/.github/workflows/vs2019/run-tests.bat
@@ -1,0 +1,12 @@
+@REM Run tests on visual studio 2019 generated build
+
+@echo "Set up PATH to find Visual Studio tools"
+call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+
+@echo PATH:
+@echo
+@echo %PATH%
+
+@echo Run tests with ctest
+cd build
+ctest -j2 -V -L CORE -C Release .
--- a/regression/CMakeLists.txt
+++ b/regression/CMakeLists.txt
@@ -3,7 +3,7 @@ set(test_pl_path "${CMAKE_CURRENT_SOURCE_DIR}/test.pl")
 macro(add_test_pl_profile name cmdline flag profile)
     add_test(
         NAME "${name}-${profile}"
-        COMMAND ${test_pl_path} -e -p -c "${cmdline}" ${flag} ${ARGN}
+        COMMAND perl ${test_pl_path} -e -p -c "${cmdline}" ${flag} ${ARGN}
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
     )
     set_tests_properties("${name}-${profile}" PROPERTIES
--- a/regression/cbmc-cpp/CMakeLists.txt
+++ b/regression/cbmc-cpp/CMakeLists.txt
@@ -1,9 +1,15 @@
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-add_test_pl_tests(
-    "$<TARGET_FILE:cbmc>" -X gcc-only
-)
+  set(gcc_only -X gcc-only)
 else()
-add_test_pl_tests(
-    "$<TARGET_FILE:cbmc> --validate-goto-model --validate-ssa-equation"
-)
+  set(gcc_only "")
 endif()
+
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  set(exclude_win_broken_tests -X winbug)
+else()
+  set(exclude_win_broken_tests "")
+endif()
+
+add_test_pl_tests(
+    "$<TARGET_FILE:cbmc> --validate-goto-model --validate-ssa-equation" ${gcc_only} ${exclude_win_broken_tests}
+)
--- a/regression/cbmc-incr-oneloop/CMakeLists.txt
+++ b/regression/cbmc-incr-oneloop/CMakeLists.txt
@@ -1,3 +1,3 @@
 add_test_pl_tests(
-    "perl -e 'alarm shift @ARGV; exec @ARGV' 8 $<TARGET_FILE:cbmc> --slice-formula"
+    "perl ../timeout.pl 8 $<TARGET_FILE:cbmc> --slice-formula"
 )
--- a/regression/cbmc-incr-oneloop/CMakeLists.txt
+++ b/regression/cbmc-incr-oneloop/CMakeLists.txt
@@ -1,3 +1,9 @@
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  set(exclude_win_broken_tests -X winbug)
+else()
+  set(exclude_win_broken_tests "")
+endif()
+
 add_test_pl_tests(
-    "perl ../timeout.pl 8 $<TARGET_FILE:cbmc> --slice-formula"
+    "perl ../timeout.pl 8 $<TARGET_FILE:cbmc> --slice-formula" ${exclude_win_broken_tests}
 )
--- a/regression/cbmc-incr/CMakeLists.txt
+++ b/regression/cbmc-incr/CMakeLists.txt
@@ -1,0 +1,9 @@
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  set(exclude_win_broken_tests -X winbug)
+else()
+  set(exclude_win_broken_tests "")
+endif()
+
+add_test_pl_tests(
+    "perl ../timeout.pl 30 $<TARGET_FILE:cbmc> --incremental --magic-numbers" ${exclude_win_broken_tests}
+)
--- a/regression/cbmc-incr/timeout.pl
+++ b/regression/cbmc-incr/timeout.pl
@@ -1,0 +1,2 @@
+alarm shift @ARGV;
+exec @ARGV;
--- a/regression/cbmc/CMakeLists.txt
+++ b/regression/cbmc/CMakeLists.txt
@@ -4,8 +4,14 @@ else()
   set(gcc_only "")
 endif()
 
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  set(exclude_win_broken_tests -X winbug)
+else()
+  set(exclude_win_broken_tests "")
+endif()
+
 add_test_pl_tests(
-    "$<TARGET_FILE:cbmc> --validate-goto-model --validate-ssa-equation" -X smt-backend ${gcc_only}
+    "$<TARGET_FILE:cbmc> --validate-goto-model --validate-ssa-equation" -X smt-backend ${gcc_only} ${exclude_win_broken_tests}
 )
 
 add_test_pl_profile(
--- a/regression/cbmc/CMakeLists.txt
+++ b/regression/cbmc/CMakeLists.txt
@@ -1,5 +1,5 @@
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-  set(gcc_only "-X gcc-only")
+  set(gcc_only -X gcc-only)
 else()
   set(gcc_only "")
 endif()
@@ -10,16 +10,18 @@ add_test_pl_tests(
 
 add_test_pl_profile(
     "cbmc-paths-lifo"
-    "$<TARGET_FILE:cbmc> --paths lifo ${gcc_only}"
+    "$<TARGET_FILE:cbmc> --paths lifo"
     "-C;-X;thorough-paths;-X;smt-backend;-X;paths-lifo-expected-failure;-s;paths-lifo"
     "CORE"
+    ${gcc_only}
 )
 
 add_test_pl_profile(
     "cbmc-cprover-smt2"
-    "$<TARGET_FILE:cbmc> --cprover-smt2 ${gcc_only}"
+    "$<TARGET_FILE:cbmc> --cprover-smt2"
     "-C;-X;broken-smt-backend;-s;cprover-smt2"
     "CORE"
+    ${gcc_only}
 )
 
 set_property(
--- a/regression/cbmc/CMakeLists.txt
+++ b/regression/cbmc/CMakeLists.txt
@@ -22,16 +22,20 @@ add_test_pl_profile(
     ${gcc_only}
 )
 
-add_test_pl_profile(
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  message(WARNING "Tests are failing on windows under this config, for reasons unknown. Investigation is pending.")
+else()
+  add_test_pl_profile(
     "cbmc-cprover-smt2"
     "$<TARGET_FILE:cbmc> --cprover-smt2"
     "-C;-X;broken-smt-backend;-s;cprover-smt2"
     "CORE"
     ${gcc_only}
-)
+    )
 
-set_property(
-  TEST "cbmc-cprover-smt2-CORE"
-  PROPERTY ENVIRONMENT
-    "PATH=$ENV{PATH}:${CMAKE_BINARY_DIR}/bin"
-)
+  set_property(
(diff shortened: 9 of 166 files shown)


hannes-steffenhagen-diffblue:
It seems like you accidentally committed a bunch of build files in here?

NlightNFotis:
@hannes-steffenhagen-diffblue Yes, that would have been the github client on my machine automatically checking in newly-generated files 🤦 

In any case, I want to see if github actions passes with this change (the binary files should not affect it), and if so, I'm going to edit this whole PR completely, to just keep the necessary parts (there's a lot of intermediate commits)

.github/workflows/vs2019/build-cbmc.bat
```diff
@@ -16,7 +16,7 @@ echo "Remove strawberry"
 del /s /q "c:/Strawberry/c/bin"
 
 echo "Configure CBMC with cmake"
-cmake -S. -Bbuild -GNinja
+cmake -DCMAKE_BUILD_TYPE=Release -S. -Bbuild -GNinja
```
hannes-steffenhagen-diffblue: This doesn’t quite work on windows because windows creates a multi config build. Instead you need to pass  `--config Release` to `cmake --build build`

.github/workflows/vs2019/run-tests.bat
```diff
@@ -0,0 +1,12 @@
...
+@echo "Set up PATH to find Visual Studio tools"
+call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+
+@echo PATH:
+@echo
+@echo %PATH%
+
+@echo Run tests with ctest
+cd build
+ctest -j2 -V -L CORE -C DEBUG .
```
hannes-steffenhagen-diffblue: This should be `-C Release` if you’re trying to test the release build

regression/CMakeLists.txt
```diff
@@ -1,9 +1,11 @@
 set(test_pl_path "${CMAKE_CURRENT_SOURCE_DIR}/test.pl")
 
 macro(add_test_pl_profile name cmdline flag profile)
+    set(__TEST_PL_TEST_COMMAND__ perl "${test_pl_path}" -e -p -c "${cmdline}" ${flag} ${ARGN})
+    message("${name}-${profile} command: ${__TEST_PL_TEST_COMMAND__}")
```
thomasspriggs: :+1: This looks incredibly useful for helping see what command is actually being run. It should make the whole process more transparent.

regression/cbmc/CMakeLists.txt
```diff
@@ -10,16 +10,18 @@ add_test_pl_tests(
 
 add_test_pl_profile(
     "cbmc-paths-lifo"
-    "$<TARGET_FILE:cbmc> --paths lifo ${gcc_only}"
+    "$<TARGET_FILE:cbmc> --paths lifo"
```
thomasspriggs, line 19: :pick: I am assuming this was done, so that `-X gcc-only` is passed to `test.pl` rather than `cbmc`. It would be great to have that information in the commit message, if that is indeed the case.

.github/workflows/pull-request-checks.yaml
```diff
@@ -117,9 +117,6 @@ jobs:
     runs-on: windows-2019
     env:
       SCRIPT_DIR: .github/workflows/vs2019
-    defaults:
```
thomasspriggs, line 118: :grey_question: Is this just for the sake of modernisation, or does it actually fix something as well?

regression/goto-instrument/CMakeLists.txt
```diff
@@ -4,6 +4,12 @@ else()
...
 
+if("{CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  set(exclude_win_broken_tests -X winbug)
+else()
+  set(exclude_win_broken_tests "")
+endif()
+
 add_test_pl_tests(
-    "${CMAKE_CURRENT_SOURCE_DIR}/chain.sh $<TARGET_FILE:goto-cc> $<TARGET_FILE:goto-instrument> $<TARGET_FILE:cbmc> ${is_windows}"
+    "${CMAKE_CURRENT_SOURCE_DIR}/chain.sh $<TARGET_FILE:goto-cc> $<TARGET_FILE:goto-instrument> $<TARGET_FILE:cbmc> ${is_windows}" ${exclude_win_broken_tests}
```
thomasspriggs, line 14: :bulb: It might be possible to reduce the amount of duplication by adding the `-X winbug` stuff in the definition of `add_test_pl_profile`, rather than adding it in each sub directory.

regression/cbmc-incr/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  set(exclude_win_broken_tests -X winbug)
+else()
+  set(exclude_win_broken_tests "")
+endif()
+
+add_test_pl_tests(
+    "perl ../timeout.pl 30 $<TARGET_FILE:cbmc> --incremental --magic-numbers" ${exclude_win_broken_tests}
```
thomasspriggs, line 8: :grey_question: Were these tests completely missing from the `cmake/ctest` run on all platforms previously? Would be good it have that in the commit message, if that is the case.

regression/cbmc/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ else()
   set(gcc_only "")
 endif()
 
-if("{CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
```
thomasspriggs, line 7: :grey_question: Should this fix have been squashed into a different commit? It looks like it belongs in the commit titles "Add exclusion mechanism for windows broken tests to cbmc/".

regression/goto-cc-file-local/CMakeLists.txt
```diff
@@ -4,6 +4,12 @@ else()
...
-add_test_pl_tests(
-    "${CMAKE_CURRENT_SOURCE_DIR}/chain.sh $<TARGET_FILE:goto-cc> $<TARGET_FILE:goto-instrument> $<TARGET_FILE:cbmc> ${is_windows}"
-)
+
+if(NOT WIN32)
+    add_test_pl_tests(
+        "${CMAKE_CURRENT_SOURCE_DIR}/chain.sh $<TARGET_FILE:goto-cc> $<TARGET_FILE:goto-instrument> $<TARGET_FILE:cbmc> ${is_windows}"
+    )
+else()
+    message(WARNING "Skipping known problematic windows tests under regression/goto-cc-file/local/")
```
thomasspriggs: :grey_question: Is the problem here that `chain.sh` for this folder doesn't work properly on windows or that the test functionality in `goto-cc` / `cbmc` doesn't work properly on Windows? It would be nice if there was a git-hub issue for this, which was referenced in the commit message.

.github/workflows/vs2019/build-cbmc.bat
```diff
@@ -16,7 +16,7 @@ echo "Remove strawberry"
 del /s /q "c:/Strawberry/c/bin"
 
 echo "Configure CBMC with cmake"
-cmake -S. -Bbuild -GNinja
+cmake -S. -Bbuild
```
thomasspriggs, line 19: :pick: It would be nice if the commit title didn't wrap. To that end I would just state the intended new functionality in the title and put any further info in the commit body. Something like title of `Use msbuild to build on Windows`, body of `Because ninja is not the native way of doing builds on windows.`.

thomasspriggs (approved):
This has actually been relatively easy to review. The small commits have made things easier, for this relatively big PR. It will be a good step forward to have the tests which are working running on Windows, rather than the current status of running none of them.

regression/CMakeLists.txt
hannes-steffenhagen-diffblue: :-1: This should be removed. (As in, the entire f52add784ecee6682f219c97ace39fbed4cd2f7d commit).

@thomasspriggs the problem is this doesn't print the test command when it's being run, this prints test commands at _configure_ time. This is for debugging yes, but throwaway code I suggested with the intention of it being temporary.

.github/workflows/pull-request-checks.yaml
```diff
@@ -39,6 +39,7 @@ jobs:
           make -C regression/cbmc test-paths-lifo
           env PATH=$PATH:`pwd`/src/solvers make -C regression/cbmc test-cprover-smt2
           make -C jbmc/regression test
+
```
hannes-steffenhagen-diffblue, line 42: Why is there a commit that's adding something called a "draft" in here if this is no longer a draft PR? It also shouldn't add this empty line here.

SHA: cfdcdca5b7091d0ea18c918f8f1159c4ef217b61

.github/workflows/pull-request-checks.yaml
```diff
@@ -125,10 +127,13 @@ jobs:
           submodules: recursive
 
       - name: Install bison and flex
-        run: "%SCRIPT_DIR%\\install-bison.bat"
+        run: "${{env.SCRIPT_DIR}}\\install-bison.bat"
```
hannes-steffenhagen-diffblue, line 127: Why this change? Should be explained.

.github/workflows/vs2019/run-tests.bat
```diff
@@ -0,0 +1,12 @@
+@REM Run tests on visual studio 2019 generated build
```
hannes-steffenhagen-diffblue, line 1: I don’t like putting the logic into bat files at all, this makes it really hard to see what's going on when looking at the job on github.

This isn't practically reusable anyway because there are so many details in here that you'd want to do different in different jobs. Strong preference for job logic to be all contained in the job script unless it's either doing an actually usefully reusable thing or is so long/complicated that it doesn't reasonably fit there.

regression/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ set(test_pl_path "${CMAKE_CURRENT_SOURCE_DIR}/test.pl")
 macro(add_test_pl_profile name cmdline flag profile)
     add_test(
         NAME "${name}-${profile}"
-        COMMAND ${test_pl_path} -e -p -c "${cmdline}" ${flag} ${ARGN}
+        COMMAND perl ${test_pl_path} -e -p -c "${cmdline}" ${flag} ${ARGN}
```
hannes-steffenhagen-diffblue, line 6: The issue here is that apparently on windows the shebang thing isn't really a thing. It's possible to hack something similar together by associating file endings with runners which is what python does, but apparently this doesn't work for perl.

Commit message should mention this (4edec5a5cf78aaaece8367d30b3ccb38d18ad7bc)

regression/cbmc/CMakeLists.txt
hannes-steffenhagen-diffblue: :+1: Yeah you’d be right @thomasspriggs, this was always the wrong place to pass in that flag and it just never became a problem before because gcc_only is empty on non-windows platforms.

regression/cbmc/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-  set(gcc_only "-X gcc-only")
+  set(gcc_only -X gcc-only)
```
hannes-steffenhagen-diffblue, line 2: Should have a short explanation of what the difference here is in the commit message (list vs. string).

SHA: ff54008eb8595c961bdecee02c95273dee0e2d53

.github/workflows/pull-request-checks.yaml
```diff
@@ -129,6 +129,14 @@ jobs:
       - name: Install bison and flex
         run: "${{env.SCRIPT_DIR}}\\install-bison.bat"
 
+      - name: Install z3
+        run: |
+          Invoke-WebRequest "https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-win.zip" -OutFile z3.zip
+          Expand-Archive z3.zip
+          $z3_path = "$(Get-Location)\z3\$((Get-ChildItem z3).Name)"
```
hannes-steffenhagen-diffblue, line 135: Can you add an explanation for why you need the get-childitem.name here? IIRC this is because the path includes the z3 version number which we expect to change, but this is weird enough to deserve a comment here.

SHA: 1dc52122584b6df8ddfe742a6f69c30333cc96de

regression/cbmc-incr-oneloop/CMakeLists.txt
```diff
@@ -1,3 +1,3 @@
 add_test_pl_tests(
-    "perl -e 'alarm shift @ARGV; exec @ARGV' 8 $<TARGET_FILE:cbmc> --slice-formula"
+    "perl ../timeout.pl 8 $<TARGET_FILE:cbmc> --slice-formula"
```
hannes-steffenhagen-diffblue: "Wrongful passing" sounds a bit like something you'd say at an eulogy for someone who was run over by a drunk driver. :smile: 

Worth pointing out here is that this is again CMake issue; Semicolons are list-item separators in CMake. Passing strings with semicolons in them around in CMake is incredibly error prone, it'd be possible to work around this here with some extra escapes but this approach is better!

SHA: aaa8425c8d210fdf9fb645721c34b5861c878dfe

regression/cbmc/CMakeLists.txt
```diff
@@ -4,8 +4,14 @@ else()
   set(gcc_only "")
 endif()
 
+if("{CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  set(exclude_win_broken_tests -X winbug)
```
hannes-steffenhagen-diffblue, line 8: :pick: :pick: :pick: 
Super nitpicky, but I'm not sure I like the name "winbug" for this. Other than that, no complains here.

regression/goto-instrument/CMakeLists.txt
hannes-steffenhagen-diffblue: Possible yes, not sure I like it. At any rate we're definitely going to refactor these macros at some point, as this work has proven they're dangerously brittle and error prone.

regression/goto-instrument/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ else()
     set(is_windows false)
 endif()
 
-if("{CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
```
hannes-steffenhagen-diffblue, line 7: This looks like a bugfix which should be squashed into the commit that introduced it.

hannes-steffenhagen-diffblue (changes requested):

regression/solver-hardness/CMakeLists.txt
```diff
@@ -1,2 +1,7 @@
-add_test_pl_tests(
-  "../chain.sh $<TARGET_FILE:cbmc>")
+if(NOT "${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  add_test_pl_tests(
+    "../chain.sh $<TARGET_FILE:cbmc>")
+else()
+  message(WARNING "skipping known problematic test on windows")
```
hannes-steffenhagen-diffblue: :no_entry:  Needs an explanation for _why_ they are being skipped.

regression/solver-hardness/solver-hardness-simple/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+CORE winbug
```
hannes-steffenhagen-diffblue, line 1: If these are being skipped why this change?

regression/goto-cc-file-local/CMakeLists.txt
hannes-steffenhagen-diffblue: Same, needs explanation

regression/cbmc-cpp/CMakeLists.txt
```diff
@@ -1,9 +1,15 @@
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-add_test_pl_tests(
-    "$<TARGET_FILE:cbmc>" -X gcc-only
-)
+  set(gcc_only -X gcc-only)
```
hannes-steffenhagen-diffblue, line 2: :+1: on changing this to match the way the other test scripts are formatted

regression/cbmc/CMakeLists.txt
```diff
@@ -22,16 +22,20 @@ add_test_pl_profile(
     ${gcc_only}
 )
 
-add_test_pl_profile(
+if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
+  message(WARNING "skipping broken cprover-smt2 tests on Windows")
```
hannes-steffenhagen-diffblue: This is confusing, the `-X broken-smt-backend` already excludes the "broken" cprover-smt2 tests. Why do we need to exclude the rest of them as well. Short explanation will do.

.github/workflows/vs2019/build-cbmc.bat
```diff
@@ -16,7 +16,7 @@ echo "Remove strawberry"
 del /s /q "c:/Strawberry/c/bin"
 
 echo "Configure CBMC with cmake"
-cmake --config Release -S. -Bbuild -GNinja
+cmake -S. -Bbuild -GNinja
 
 echo "Build CBMC with cmake"
-cmake --build build
+cmake --build build --config Release
```
hannes-steffenhagen-diffblue, line 22: This does nothing when building with ninja...

.github/workflows/vs2019/build-cbmc.bat
hannes-steffenhagen-diffblue, line 19: ... which is rectified here, these two things should be one commit, with an explanation of what's going on.

hannes-steffenhagen-diffblue (changes requested):
Thanks for doing all this work!

But there's a bunch of small things that definitely should be fixed before merging.

.github/workflows/vs2019/run-tests.bat
thomasspriggs: I have to disagree here somewhat. I like the `.bat` file for this. Having the logic in `.bat` files makes it much easier to try out locally. Otherwise more copy and pasting from the `.yml` file is required.

hannes-steffenhagen-diffblue:
Also, if it's not too much to ask could you start a README.md in the top level regression directory? Just to drop a line in there about what "winbug" means for now, but for later work we can add an explanation for how our regression tests work in there, we really need to start writing these things down.

regression/cbmc/CMakeLists.txt
thomasspriggs: I agree. I would have chosen something more verbose. Maybe `broken-on-windows`, to be similar to the wording of `broken-smt-backend`. Or maybe `known-windows-bug` to be more similar to original wording.

.github/workflows/pull-request-checks.yaml
NlightNFotis: No - later on in the same file we are using powershell features not available to cmd.

thomasspriggs: It would be good to add that explanation to the commit body and/or squash this commit into the commit which starts using powershell features.

regression/cbmc/CMakeLists.txt
NlightNFotis: I could change this, but I have a strong preference on doing this in a different PR, not now.

Unless there are strong objections to this.

.github/workflows/pull-request-checks.yaml
NlightNFotis: Because this is *literally* a first draft: This got the infrastructure in place, but it gets nailed down later, for example, with the commit that installs z3.

hannes-steffenhagen-diffblue: Right, but now that you've done all that I don't see a good reason not to consolidate these changes. Also, this doesn't change the fact there's a spurious newline here.

.github/workflows/vs2019/run-tests.bat
NlightNFotis: I'm somewhere in the middle with this - I see Hannes' point, and I actually prefer that - but for consistency's sake with the current infrastructure, I'd rather it stays a batch script.

I could put it on my todo if people insist to port both of these to the github action workflow file, but I would much rather this happens at a later PR.

.github/workflows/pull-request-checks.yaml
NlightNFotis: I much prefer they are not consolidated - this was fleshed out in bits, and I prefer that the history reflects that.

For the empty line, it looks spurious, I agree, but this was introduced because without it the file looks badly formatted - there's no new line between two workflows, and at least to me, that looks weird, especially since it's inconsistent with the others.

regression/cbmc-incr/timeout.pl
```diff
@@ -0,0 +1,2 @@
+alarm shift @ARGV;
```
peterschrammel, line 1: ⛏️ Maybe this script can be moved to a more central place to avoid duplicating it.

regression/readme.md
```diff
@@ -0,0 +1,12 @@
+# CProver regression tests
+
+This folder contains the CProver regression test-suite.
+
+## Notes
+
+* Tests marked as `winbug` are currently known to be failing
+  on windows, but passing on other platforms. The reason for
```
peterschrammel: ⛏️ ⛏️ ⛏️ 
```suggestion
  on Windows, but passing on other platforms. The reason for
```

regression/readme.md
```diff
@@ -0,0 +1,12 @@
...
+  this is not known, and it's currently being investigated.
+  This was discovered during work done to port CI from travis
+  and codebuild to github actions. Worth noting that those tests
+  were not being run on windows before.
```
peterschrammel: ⛏️ ⛏️ ⛏️ 
```suggestion
  were not being run on Windows before.
```

peterschrammel (approved):

.github/workflows/pull-request-checks.yaml
NlightNFotis: I played around a little bit, and it appears this is the correct syntax.

`env`, in the context of github actions, is setting a *workflow environment variable*, not a shell one. If this is changed to a powershell syntax variable reference, we get all sorts of syntax errors, and this stack overflow answer seems to be supporting the previous statement: https://stackoverflow.com/questions/63467662/github-action-define-workflow-level-environment-variable-between-jobs

regression/readme.md
```diff
@@ -0,0 +1,12 @@
...
+This folder contains the CProver regression test-suite.
+
+## Notes
+
+* Tests marked as `winbug` are currently known to be failing
+  on Windows, but passing on other platforms. The reason for
+  this is not known, and it's currently being investigated.
+  This was discovered during work done to port CI from travis
+  and codebuild to github actions. Worth noting that those tests
+  were not being run on Windows before.
```
hannes-steffenhagen-diffblue, line 12: :+1: :100:

hannes-steffenhagen-diffblue (approved):
My preference would've been to merge all the commits that change the github actions script but if you feel strongly that keeping those separate is better I'm not going to block because of that.

PR #5447, opened 2020-08-10
Refactoring of the parsing of the `--object-bits` command line argument

thomasspriggs:
As I was working on #5440, I found the entry point code less than straight forward to follow. I carried out this refactorings locally in order to make it easier for me to understand. I am raising this separate PR containing the refactoring work only as it was not required in order to implement the feature in the original PR, but should still improve maintainability.

- [x] N/A - No user visible changes ~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~
- [x] N/A - non claimed - ~My commit message includes data points confirming performance improvements (if claimed).~

### DIFF
--- a/regression/cbmc/object-bits-parsing/non_numeric.desc
+++ b/regression/cbmc/object-bits-parsing/non_numeric.desc
@@ -1,6 +1,7 @@
 CORE
 test.c
 --object-bits foobar
+Value of "foobar" given for object-bits is not a valid unsigned integer.
 object-bits must be positive and less than the pointer width
 ^EXIT=1$
 ^SIGNAL=0$
--- a/regression/cbmc/object-bits-parsing/non_numeric.desc
+++ b/regression/cbmc/object-bits-parsing/non_numeric.desc
@@ -1,0 +1,9 @@
+CORE
+test.c
+--object-bits foobar
+object-bits must be positive and less than the pointer width
+^EXIT=1$
+^SIGNAL=0$
+--
+--
+Test parsing of an invalid non numeric object-bits setting.
--- a/src/util/config.cpp
+++ b/src/util/config.cpp
@@ -786,7 +786,7 @@ configt::bv_encodingt parse_object_bits_encoding(
   const auto object_bits = string2optional<unsigned int>(argument);
   if(!object_bits)
     throw_for_reason("not a valid unsigned integer");
-  if(*object_bits <= 0 || *object_bits >= pointer_width)
+  if(*object_bits == 0 || *object_bits >= pointer_width)
     throw_for_reason("out of range");
 
   configt::bv_encodingt bv_encoding;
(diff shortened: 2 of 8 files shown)


src/util/config.cpp
```diff
@@ -776,8 +776,8 @@ configt::bv_encodingt parse_object_bits_encoding(
   const std::string &argument,
   const std::size_t pointer_width)
 {
-  const unsigned int object_bits = unsafe_string2unsigned(argument);
-  if(object_bits <= 0 || object_bits >= pointer_width)
```
hannes-steffenhagen-diffblue: :pick: Would be better if the case where parsing fails and the `object_bits` are out of range were separated, also it would be helpful for debugging to put `argument` in the error message (this makes it easier to find if for example you've messed up a command line when copy&pasting, e.g. `--object-bits --function hello world 8`)

hannes-steffenhagen-diffblue (approved):
Looks OK to me

hannes-steffenhagen-diffblue (approved):
Looks OK to me

regression/cbmc/object-bits-parsing/non_numeric.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+test.c
+--object-bits foobar
+object-bits must be positive and less than the pointer width
+^EXIT=1$
+^SIGNAL=0$
+--
+Test parsing of an invalid non numeric object-bits setting.
```
smowton, line 10: This is a must-not-match, not a comment (similarly others)

src/util/config.cpp
```diff
@@ -765,16 +765,38 @@ void configt::set_arch(const irep_idt &arch)
...
+///   the value of object_bits is within the valid range.
+/// \return A `bv_encodingt` on successful parsing. In the case where an invalid
+///    argument is specified, an `invalid_command_line_argument_exceptiont` will
+///    be thrown.
+configt::bv_encodingt parse_object_bits_encoding(
+  const std::string &argument,
+  const std::size_t pointer_width)
+{
+  const auto object_bits = string2optional<unsigned int>(argument);
+  if(!object_bits || *object_bits <= 0 || *object_bits >= pointer_width)
```
smowton: Note `<= 0` on an unsigned type

smowton (approved):
Couple of simple fixes otherwise lgtm

regression/cbmc/object-bits-parsing/non_numeric.desc
thomasspriggs: Fixed.

src/util/config.cpp
thomasspriggs: I have swapped this to use `== 0`, as this should be less misleading.

src/util/config.cpp
thomasspriggs: Done.

chrisr-diffblue (approved):

PR #5449, opened 2020-08-11
Add github actions ubuntu-20.04 make build

thomasspriggs:
So that we have testing of the make build on Ubuntu-20.04.

- [x] n/a No code modified - ~Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.~
- [x] n/a No features or user visible behaviour changes - ~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~
- [x] n/a No code modified - ~Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).~
- [x] n/a None claimed - ~My commit message includes data points confirming performance improvements (if claimed).~
- [x] n/a ~White-space or formatting changes outside the feature-related changed lines are in commits of their own.~

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -4,6 +4,42 @@ on:
     branches: [ develop ]
 
 jobs:
+  check-ubuntu-20_04-make-gcc:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Fetch dependencies
+        env:
+          # This is needed in addition to -yq to prevent apt-get from asking for
+          # user input
+          DEBIAN_FRONTEND: noninteractive
+          TESTPL_JOBS: 4
+        run: |
+          sudo apt-get install -yq gcc gdb g++ maven jq flex bison libxml2-utils cpanminus
+          make -C src minisat2-download
+          cpanm Thread::Pool::Simple
+      - name: Build with make
+        run: |
+          make -C src CXX='/usr/bin/g++' -j2
+          make -C unit CXX='/usr/bin/g++' -j2
+          make -C jbmc/src  CXX='/usr/bin/g++' -j2
+          make -C jbmc/unit CXX='/usr/bin/g++' -j2
+      - name: Run unit tests
+        run: |
+          make -C unit test
+          make -C jbmc/unit test
+          echo "Running expected failure tests"
+          make TAGS="[!shouldfail]" -C unit test
+          make TAGS="[!shouldfail]" -C jbmc/unit test
+      - name: Run regression tests
+        run: |
+          make -C regression test
+          make -C regression/cbmc test-paths-lifo
+          env PATH=$PATH:`pwd`/src/solvers make -C regression/cbmc test-cprover-smt2
+          make -C jbmc/regression test
+
   check-macos-10_15-make-clang:
     runs-on: macos-10.15
     steps:


.github/workflows/pull-request-checks.yaml
```diff
@@ -4,6 +4,40 @@ on:
...
+          make -C src minisat2-download
+      - name: Build with make
+        run: |
+          make -C src CXX='/usr/bin/g++' -j2
+          make -C unit CXX='/usr/bin/g++' -j2
+          make -C jbmc/src  CXX='/usr/bin/g++' -j2
+          make -C jbmc/unit CXX='/usr/bin/g++' -j2
+      - name: Run unit tests
+        run: |
+          make -C unit test
```
hannes-steffenhagen-diffblue, line 31: I believe the unit tests, at least the jbmc ones, actually need to be run from the `unit` directory

thomasspriggs: The `-C unit` argument to `make` changes into the `unit` directory and changes back before exiting `make`. This is the same way as the `make` build is implemented in Travis right now.

hannes-steffenhagen-diffblue (approved):

hannes-steffenhagen-diffblue (approved):
LGTM with explanation

hannes-steffenhagen-diffblue:
So yeah, the `TESTPL_JOBS` optimisation here leads to a roughly 2x speedup in test runtime. This is expected, but good to have confirmed.

.github/workflows/pull-request-checks.yaml
```diff
@@ -4,6 +4,42 @@ on:
...
+          submodules: true
+      - name: Fetch dependencies
+        env:
+          # This is needed in addition to -yq to prevent apt-get from asking for
+          # user input
+          DEBIAN_FRONTEND: noninteractive
+          TESTPL_JOBS: 4
+        run: |
+          sudo apt-get install -yq gcc gdb g++ maven jq flex bison libxml2-utils cpanminus
+          make -C src minisat2-download
```
chrisr-diffblue, line 21: I'd be more inclined to move this `make` invocation into the build section - it's part of the system-under-tests (i.e. the Makefiles) rather than general environment setup (the apt-get)

NlightNFotis: Yes, but afaik this doesn't really build minisat, does it? It only downloads the sources and applies a patch as far as I can recall?

chrisr-diffblue: All of which is defined in the Makefile, which is what is being tested here.

NlightNFotis: Yes, I recall correctly (https://github.com/diffblue/cbmc/blob/develop/src/Makefile#L120).

I would argue this is part of the configuration, as without minisat being downloaded, the build fails - but it's really because the sources are missing, not because of a compilation issue or anything else. And seeing as minisat is one of the SAT solvers we support, I would argue even more this is part of the configuration, even if it's just the default right now.

NlightNFotis (approved):
LGTM

.github/workflows/pull-request-checks.yaml
chrisr-diffblue: I don't have a massive issue with this - so I'm happy to approve if everyone else is happy. My logic was:

1) Cmake does the minisat download as part of its configure/build steps, so my suggestion would be more consistent with that
2) If the apt-get fails, or the cpanm fails, its mostly just a case of saying "shrug, infrastructure failed - hit retry", but if the minisat-download fails you _do_ have to double check whether our own stuff needs to be modified. If you have the "setup our CI host" as a separate section to "run our own stuff" then it makes a clear distinction between stuff we may need to manage/modify and stuff that is just pure environment setup.

chrisr-diffblue (approved):
A minor quibble, but not going to block the PR considering other approvals.

PR #5451, opened 2020-08-12
WIP add version bump on release

hannes-steffenhagen-diffblue:
We are using the `CBMC_VERSION` setting in `config.inc` to display a version for `cbmc --version`. Since we’re doing biweekly releases now this field should be updated, ideally automatically, on release. This is an attempt to get that done.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,33 @@
+name: cprover-release
+on:
+  # Only on pull request so I can see what's happening
+  pull_request:
+    branches: [develop]
+  release:
+    types: [created]
+
+jobs:
+  update-cbmc-version:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set CBMC_VERSION environment variable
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,33 @@
+name: cprover-release
+on:
+  # Only on pull request so I can see what's happening
+  pull_request:
+    branches: [develop]
+  release:
+    types: [created]
```
hannes-steffenhagen-diffblue, line 7: This is currently triggered "on release", but more likely it would be better to have an automatically scheduled released (via the schedule event + release action). Otherwise we get into the awkward spot of where we set the release version one commit _after_ what the release tag points to.

* Release action: https://github.com/marketplace/actions/create-release
* Scheduled actions: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events

hannes-steffenhagen-diffblue:
I don’t think this is the right way to do it; Instead we should create a bump and then a release within the action.

PR #5452, opened 2020-08-12
Don’t try to validate trace XML when xmllint not found

hannes-steffenhagen-diffblue:
This was the intended behaviour to begin with, I just forgot
to put the ‘exit’ there.

### DIFF
--- a/regression/validate-trace-xml-schema/check.sh
+++ b/regression/validate-trace-xml-schema/check.sh
@@ -13,6 +13,7 @@ fi
 xmllint="$(command -v xmllint)"
 if [ $? -ne 0 ] > /dev/null; then
     echo "xmllint not found, skipping XSD tests"
+    exit
 fi
 
 python3 check.py ../../src/cbmc/cbmc "$xmllint"


regression/validate-trace-xml-schema/check.sh
```diff
@@ -13,6 +13,7 @@ fi
 xmllint="$(command -v xmllint)"
 if [ $? -ne 0 ] > /dev/null; then
     echo "xmllint not found, skipping XSD tests"
+    exit
```
thomasspriggs, line 16: It seems to me that this should exit with an error code, as we haven't tested successfully if `xmllint` is missing.

NlightNFotis: I'm a fan of this idea - even though it looks to me like we have checked if `xmllint` was found, which is what we care for here, isn't it?

regression/validate-trace-xml-schema/check.sh
```diff
@@ -13,6 +13,7 @@ fi
 xmllint="$(command -v xmllint)"
```
NlightNFotis, line 13: Also, for something unrelated to this PR, but why is `command -v` used here instead of `whereis`? Are there any benefits? 

I don't mind it, it was just surprising to me - never seen this before, and for this functionality, I've always seen `whereis` being used instead.

NlightNFotis (approved):
Looks good overall, only would ask if it would be possible to get an error code returned before it gets merged.

hannes-steffenhagen-diffblue:
@NlightNFotis I'd prefer to keep this consistent with the other cases right now. Having this error out if the prereqs aren't there would be a much bigger change, as it would involve basically changing most of our CI scripts.

regression/validate-trace-xml-schema/check.sh
hannes-steffenhagen-diffblue: `command` is very simple. It just checks if something is on `$PATH`, optionally returns the path to it and gives me an error code if can’t be found. `whereis` seems to be a more complicated tool used for actually locating things, which is not what I’m trying to do.

It also doesn’t fail by default if it can’t find anything and gives me ton of different things if it does. For me the question is more the other way round, is there a reason to use `whereis` when `command` will do?

PR #5454, opened 2020-08-12
Bump CBMC_VERSION to 5.12.6

hannes-steffenhagen-diffblue:
We use this version number to make debugging easier, it should stay up to date with the current release version. Updating it automatically is WIP.

### DIFF
--- a/src/config.inc
+++ b/src/config.inc
@@ -76,7 +76,7 @@ endif
 OSX_IDENTITY="Developer ID Application: Daniel Kroening"
 
 # Detailed version information
-CBMC_VERSION = 5.12
+CBMC_VERSION = 5.12.6
 
 # Use the CUDD library for BDDs, can be installed using `make -C src cudd-download`
 # CUDD = ../../cudd-3.0.0


src/config.inc
```diff
@@ -76,7 +76,7 @@ endif
 OSX_IDENTITY="Developer ID Application: Daniel Kroening"
 
 # Detailed version information
-CBMC_VERSION = 5.12
+CBMC_VERSION = 5.12.6
```
peterschrammel, line 79: Shouldn't the git tag be the single source of truth?

hannes-steffenhagen-diffblue: @peterschrammel We currently use both this and the git tag. We could alternatively change how we do versioning to only take the git tag into consideration; The only problem is then what version do we put on code built e.g. from a source tarball (though I suppose that could be fixed by making sure `version.cpp` is included with that?).

thomasspriggs: I think we need to keep the version number in `src/config.inc`, as long as we want the ability to build cbmc from a downloaded tar ball. Automatically fetching the version number from git as the single source of truth would mean that we can only build from a git checkout.

thomasspriggs (approved):
Approving as this makes `config.inc` consistent with the current state of the git tags.

peterschrammel (approved):
Yes, given that we distribute source tarballs, the version number must be somewhere inside.

PR #5455, opened 2020-08-12
Fix loop-to-assume transformation in goto-symex

### LINKED ISSUE #5450: CBMC smybolic execution soundness issue; backwards goto missed in do ... while loop
CBMC version: 5.12 (cbmc-5.12.6-17-g7d30335)
Operating system: GNU/Linux x86_64 
Exact command line resulting in the issue: cbmc top.c
What behaviour did you expect: VERIFICATION SUCCESSFUL
What happened instead: VERIFICATION FAILED

Test case `top.c`:
```
#include "assert.h"
int main()
{
   int count = 0;
   do { 
     count = count +1 ; 
   } while(count < 5);

   assert(count == 5);
   assert(count == 17);
}
```
The first assertion should pass, the second should fail. However, CBMC reports SUCCESS for both assertions.
Looking at the program functions shows that simulating the loop is aborted via `ASSUME(false)` after one iteration:
```
cbmc top.c --program-only
[...]
// 1 file top.c line 5 function main
(36) count!0@1#2 == 0
// 2 file top.c line 7 function main
(37) count!0@1#3 == 1
// 3 file top.c line 8 function main
// 3 file top.c line 8 function main
(38) ASSUME(FALSE)
```
The problem is also present in cbmc-5.12.6
The last major version of CBMC where this test case works correctly is 5.7.

tautschnig:
The previous implementation (from e672e0d4) would not account for
loop heads with side effects.

Fixes: #5450

### DIFF
--- a/regression/cbmc/while2/main.c
+++ b/regression/cbmc/while2/main.c
@@ -1,0 +1,20 @@
+#include "assert.h"
+
+int main()
+{
+  int count = 0;
+  do
+  {
+    count = count + 1;
+  } while(count < 5);
+
+  do
+  {
+  } while(count < 5);
+
+  while(count < 5)
+    ;
+
+  assert(count == 5);
+  assert(count == 17);
+}
--- a/regression/cbmc/while2/requires-transform.c
+++ b/regression/cbmc/while2/requires-transform.c
@@ -1,0 +1,15 @@
+#include "assert.h"
+
+int main()
+{
+  int count;
+
+  do
+  {
+  } while(count < 5);
+
+  while(count < 5)
+    ;
+
+  assert(count >= 5);
+}
--- a/regression/cbmc/while2/requires-transform.desc
+++ b/regression/cbmc/while2/requires-transform.desc
@@ -1,0 +1,11 @@
+CORE
+requires-transform.c
+
+^EXIT=0$
+^SIGNAL=0$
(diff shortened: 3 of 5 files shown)


martin-cs (approved):
Yep; that looks like the kind of fix that is needed.  Are there already tests for the kind of loop it is supposed to target?

smowton (commented):
Are you sure your test case exercises the new code? I'd expect the first loop to leave `count == 5` in the constant propagator. Then the second loop (the empty do-while) would use the empty-loop special case, but the third loop would probably be skipped by plain constant propagation evaluating the conditional jump statically before we consider if it's an empty loop?

smowton:
(Either way it'd be worth adding comments to the test-case indicating what you expect to happen)

hannes-steffenhagen-diffblue:
@smowton This definitely fixes the reported issue

hannes-steffenhagen-diffblue (approved):
LGTM (the way I understood this this previously replaced all empty or seemingly “empty” loops with assume false, and now only does if the loop condition can be simplified to true?).

smowton (approved):
OK, if you're sure it works as advertised

tautschnig:
> OK, if you're sure it works as advertised

Well, now at least it is being tested...

peterschrammel (approved):

PR #5456, opened 2020-08-12
Add an action to perform automatic releases

hannes-steffenhagen-diffblue:
This action automatically does the following every thursday at 9:00 AM:

* Create a commit that increments the last number of CBMC_VERSION
  in src/config.inc
* Perform a release tagged with the new CBMC_VERSION number

This only saves us a tiny bit of work right now, but we’ll add actions to
automatically create and upload binary artifacts for these releases soon.

Depends on https://github.com/diffblue/cbmc/pull/5454 because that sets the `CBMC_VERSION` to what this script expects.

### DIFF
--- a/.github/workflows/regular-release.yaml
+++ b/.github/workflows/regular-release.yaml
@@ -1,0 +1,66 @@
+on:
+  schedule:
+#            ┌───────────── minute (0 - 59)
+#            │ ┌───────────── hour (0 - 23)
+#            │ │ ┌───────────── day of the month (1 - 31)
+#            │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+#            │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+#            │ │ │ │ │
+#            │ │ │ │ │
+#            │ │ │ │ │
+    - cron: '0 9 * * THU'
+      # ^ this means 9:00 AM every thursday
+      # I can’t figure out the right syntax
+      # for ‘every other thursday’
+
+jobs:
+  bump_cbmc_version:
+    runs-on: ubuntu-20.04
+    outputs:
+      cbmc_version: ${{ steps.cbmc_version_number.outputs.CBMC_VERSION }}
+      bump_git_sha: ${{ steps.commit_bump.outputs.bump_git_sha }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Get new CBMC version number
+        id: cbmc_version_number
+        run: |
+          NEW_CBMC_VERSION=$(grep '^CBMC_VERSION =' src/config.inc | cut -d = -f 2 | scripts/increment_version.sh)
+          echo ::set-env name=CBMC_VERSION::$NEW_CBMC_VERSION
+          echo ::set-output name=CBMC_VERSION::$NEW_CBMC_VERSION
+      - name: Update CBMC version
+        run: |
+          sed -i "s/CBMC_VERSION = .*/CBMC_VERSION = $CBMC_VERSION/" src/config.inc
+      - name: Set git identity to github bot
+        run: |
+          git config --local user.name db-ci-cprover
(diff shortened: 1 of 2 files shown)


NlightNFotis (approved):
LGTM

.github/workflows/regular-release.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Get new CBMC version number
+        id: cbmc_version_number
+        run: |
+          NEW_CBMC_VERSION=$(grep '^CBMC_VERSION =' src/config.inc | cut -d = -f 2 | scripts/increment_version.sh)
+          echo ::set-env name=CBMC_VERSION::$NEW_CBMC_VERSION
+          echo ::set-output name=CBMC_VERSION::$NEW_CBMC_VERSION
+      - name: Update CBMC version
+        run: |
+          sed -i "s/CBMC_VERSION = .*/CBMC_VERSION = $CBMC_VERSION/" src/config.inc
+      - name: Set git identity to github bot
```
NlightNFotis, line 35: Is this name still valid after the change?

.github/workflows/regular-release.yaml
```diff
@@ -0,0 +1,66 @@
...
+#            │ │ ┌───────────── day of the month (1 - 31)
+#            │ │ │ ┌───────────── month (1 - 12 or JAN-DEC)
+#            │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
+#            │ │ │ │ │
+#            │ │ │ │ │
+#            │ │ │ │ │
+    - cron: '0 9 * * THU'
+      # ^ this means 9:00 AM every thursday
+      # I can’t figure out the right syntax
+      # for ‘every other thursday’
```
hannes-steffenhagen-diffblue, line 14: Actually having looked into this a bit more I don't think there *is* a right syntax for "every other thursday". The cron syntax we have available here is very limited (see: https://crontab.guru/#0_9_*_*_THU)

.github/workflows/regular-release.yaml
hannes-steffenhagen-diffblue: err, no I need to update the name of this step.

*edit* github bot is probably the right name for this

.github/workflows/regular-release.yaml
```diff
@@ -0,0 +1,66 @@
...
+          git config --local user.name db-ci-cprover
+          git config --local user.email "db-ci-cprover@diffblue.com"
+      - name: Commit changes
+        id: commit_bump
+        run: |
+          git commit -a -m "Bump version to $CBMC_VERSION"
+          echo "::set-output name=bump_git_sha::$(git rev-parse HEAD)"
+      - name: Push changes
+        run: |
+          git push
```
thomasspriggs, line 46: :grey_question: Which branch is this pushing to? Might be nice to have a comment with the answer.

thomasspriggs (approved):

.github/workflows/regular-release.yaml
hannes-steffenhagen-diffblue: We didn’t specify a branch in `checkout`, so we pull from the default (`develop`), which automatically also sets the upstream of the branch to `develop`. So this will push against develop.

.github/workflows/regular-release.yaml
chrisr-diffblue: Yeah, I don't think Cron syntax supports that - best you can probably do is to run this job every thursday, and then in the job check if it's a release Thursday or not (and short circuit if its not). Might want to think about timezones as well?

chrisr-diffblue (commented):
I think this mostly looks fine. If I'm reading this correctly, this looks like it directly commits to `develop` ? If so, does this create a risk that we might tag a build as a 'release' when it's actually failing CI due to a previous bad merge?

chrisr-diffblue (approved):
Discussed offline with @hannes-steffenhagen-diffblue - happy for this to go in as is.

peterschrammel (approved):

PR #5458, opened 2020-08-13
Set CMake project version based on config.inc

hannes-steffenhagen-diffblue:
Currently we set the CBMC version in config.inc.

It would be useful to have this information in our CMakeLists.txt as
well (especially for e.g. CPack packaging), and to avoid duplication it
would be better to have it in just one place. It’s somewhat easier to
parse config.inc from CMake than doing it the other way round (and we
already do it in src/util/CMakeLists.txt anyway), we do it here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,16 @@
 cmake_minimum_required(VERSION 3.2)
 
-project(CBMC)
+
+# Grab the current CBMC version from config.inc
+file(
+  STRINGS src/config.inc CBMC_VERSION
+  REGEX "CBMC_VERSION = (.*)")
+string(REGEX REPLACE "CBMC_VERSION = (.*)" "\\1" CBMC_VERSION ${CBMC_VERSION})
+
+project(CBMC VERSION ${CBMC_VERSION})
+
+# when config.inc changes we’ll need to reconfigure to check if the version changed
+set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/config.inc")
 
 find_program(CCACHE_PROGRAM ccache)
 if(CCACHE_PROGRAM)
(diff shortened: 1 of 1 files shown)


NlightNFotis (approved):
LGTM, but I would appreciate a small comment on why we do it in the source - I know it's in the commit message, but something similar (smaller preferably) would be great to have in the code itself.

CMakeLists.txt
```diff
@@ -1,6 +1,20 @@
...
+# filenames)
+file(
+  STRINGS src/config.inc CBMC_VERSION
+  REGEX "CBMC_VERSION = (.*)")
+string(REGEX REPLACE "CBMC_VERSION = (.*)" "\\1" CBMC_VERSION ${CBMC_VERSION})
+
+project(CBMC VERSION ${CBMC_VERSION})
+
+# when config.inc changes we’ll need to reconfigure to check if the version changed
+set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/config.inc")
```
chrisr-diffblue, line 17: Just a general query about this - when does config.inc get generated? IIRC it gets auto generated using the git info? If so, when does that change, and what is the effect here? I'm slightly worried about whether this means every time you make a local change and re-build, it _also_ causes a fresh re-configure? And if so, doesn't that also mean a fresh re-download of minisat?

chrisr-diffblue (approved):
Generally looks good to me, and good to see this made consistent with the Makefile builds. A small query but not blocking.

CMakeLists.txt
hannes-steffenhagen-diffblue: @chrisr-diffblue This string is static and currently only user set (plan is to autoincrement on release going forward). The git-tag string is _also_ used but only for use in `util/version.h`, which uses _both_ this info _and_ the git-tag.

The reason we keep the two separate is to support builds from source packages and shallow clones. So this will not trigger frequent reconfigures.

As I believe I’ve mentioned before as well a long term goal of mine is for us to stop downloading things off the internet in our `CMakeLists.txt` at all and instead provide the dependency downloading/building stuff as a separate script.

PR #5459, opened 2020-08-13
Build CBMC in Visual Studio 2019, in Github Actions.

NlightNFotis:

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -21,3 +21,21 @@ jobs:
         run: cd build; ninja
       - name: Run CTest
         run: cd build; ctest -V -L CORE .
+  check-vs-2019-build-only:
+    runs-on: windows-2019
+    env:
+      SCRIPT_DIR: .github/workflows/vs2019
+    defaults:
+      run:
+        shell: cmd
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Install bison and flex
+        run: "%SCRIPT_DIR%\\install-bison.bat"
+
+      - name: Build cbmc
+        run: "%SCRIPT_DIR%\\build-cbmc.bat"


.github/workflows/pull-request-checks.yaml
```diff
@@ -21,3 +21,22 @@ jobs:
         run: cd build; ninja
       - name: Run CTest
         run: cd build; ctest -V -L CORE .
+  check-vs-2019-build-and-test:
```
NlightNFotis: The name of the action is `build-and-test` because I'm already working on building and running tests on Github Actions under vs2019, but the running tests part is not there yet - however, since the building CBMC is already done, we figured we could publish this part of the work, and get the rest of it merged as it's finished.

hannes-steffenhagen-diffblue: I don’t think this needs to be called ‘build-and-test’, the other jobs in here are called ‘check’-somethingsomething and I think building _and_ testing should be implied.

If you’re doing something _other_ than building and testing _that_ should be called out.

.github/workflows/pull-request-checks.yaml
```diff
@@ -21,3 +21,22 @@ jobs:
...
+    runs-on: windows-2019
+    env:
+      SCRIPT_DIR: .github/workflows/vs2019
+    defaults:
+      run:
+        shell: cmd
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v2
```
hannes-steffenhagen-diffblue, line 33:
```suggestion
        uses: actions/checkout@v2
        with:
          submodules: recursive
```

.github/workflows/pull-request-checks.yaml
```diff
@@ -21,3 +21,22 @@ jobs:
...
+
+      - name: Update the submodules
+        run: git submodule update --init
```
hannes-steffenhagen-diffblue: Apply above suggestion then this step isn’t needed anymore

.github/workflows/pull-request-checks.yaml
```diff
@@ -21,3 +21,22 @@ jobs:
...
+
+      - name: Install bison and flex
+        run: ${{env.SCRIPT_DIR}}\install-bison.bat
```
hannes-steffenhagen-diffblue: It’s in env, so this really should be %SCRIPT_DIR%

.github/workflows/pull-request-checks.yaml
```diff
@@ -21,3 +21,22 @@ jobs:
...
+
+      - name: Build cbmc
+        run: ${{env.SCRIPT_DIR}}\build-cbmc.bat
```
hannes-steffenhagen-diffblue: ditto

.github/workflows/pull-request-checks.yaml
thomasspriggs: :pick: Given that testing isn't part of this PR, we should probably name this `check-vs-2019-build-only`, so that there is no confusion in the interim between the two PRs. It can just be renamed again, when the second PR comes.

thomasspriggs (approved):
:pick: When I was checking through the build log I noticed that it stated "Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE) ". So It may be worth adding a doxygen install.

.github/workflows/pull-request-checks.yaml
```diff
@@ -21,3 +21,21 @@ jobs:
...
+      run:
+        shell: cmd
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Install bison and flex
+        run: %SCRIPT_DIR%\install-bison.bat
```
hannes-steffenhagen-diffblue: :no_entry_sign: This doesn’t run right now due to a yaml parsing error (apparently % is not allowed in plain context). Write this as either

```suggestion
        run: "%SCRIPT_DIR%\\install-bison.bat"
```

or



```suggestion
        run: |
          %SCRIPT_DIR%\install-bison.bat
```

hannes-steffenhagen-diffblue (approved):
LGTM

peterschrammel (approved):

PR #5461, opened 2020-08-14
Add spell correction for CLI options

hannes-steffenhagen-diffblue:
This gives you some suggestions if you’ve mistyped a command line parameter.

E.g. on `cbmc --traec` it will suggest `--trace`, if you type `cbmc --cvc` it will suggest trying `--cvc3` or `--cvc4`, that sort of thing.

### DIFF
--- a/regression/cbmc/unknown-argument-suggestion/test.desc
+++ b/regression/cbmc/unknown-argument-suggestion/test.desc
@@ -1,0 +1,10 @@
+CORE
+dummy.c
+--traec
+did you mean --trace
+^EXIT=64$
+^SIGNAL=0$
+--
+--
+This checks that we get a useful suggestion when we make a typo on the
+commandline
--- a/regression/cbmc/unknown-argument-suggestion/test.desc
+++ b/regression/cbmc/unknown-argument-suggestion/test.desc
@@ -2,9 +2,11 @@ CORE
 dummy.c
 --traec
 did you mean --trace
-^EXIT=[123456789]\d*$
+^EXIT=(1|64)$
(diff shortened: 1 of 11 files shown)


regression/cbmc/unknown-argument-suggestion/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+dummy.c
+--traec
+did you mean --trace
+^EXIT=[123456789]\d*$
```
hannes-steffenhagen-diffblue: A right nuisance: Apparently the error codes on windows and linux are different, but our test.pl script insist on very particular exit code checks. I'm tempted to just remove this test to be honest.

src/util/nfa.h
```diff
@@ -0,0 +1,168 @@
+/// \file
+/// \author Diffblue Ltd.
+///
+/// A simple NFA implementation
```
martin-cs: Please could you add what it is intended for.  There have been other automaton implementations in CPROVER (for SATAbs for example) and I wouldn't like future people to get the wrong idea of what this is.

martin-cs (approved):
This is a really cute feature -- we should have this.

src/util/nfa.h
hannes-steffenhagen-diffblue: Good point, will do

chrisr-diffblue (approved):
While this is a bit of feature creep - it is indeed cute, but also really helpful to new users (and old users that typo :-) ) so I'm happy for it to go in.

peterschrammel (approved):

PR #5464, opened 2020-08-17
Cd/create debian package on release

hannes-steffenhagen-diffblue:

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -51,7 +51,7 @@ jobs:
           # user input
           DEBIAN_FRONTEND: noninteractive
         run: |
-          sudo apt-get install -yq cmake ninja-build gcc g++ maven flex bison libxml2-utils cpanminus
+          sudo apt-get install -yq cmake ninja-build gcc g++ maven flex bison libxml2-utils cpanminus dpkg-dev
           cpanm Thread::Pool::Simple
       - name: Configure using CMake
         run: |
@@ -60,6 +60,11 @@ jobs:
           cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++
       - name: Build with Ninja
         run: cd build; ninja
+      - name: Check if package building works
+        run: |
+          cd build
+          ninja package
+          ls *.deb
       - name: Run tests
         run: cd build; ctest . -V -L CORE
         env:
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -1,0 +1,45 @@
+on:
+  release:
+    types: [created]
+
+name: Upload additional release assets
+jobs:
+  ubuntu-package:
+    runs-on: ubuntu-20.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Fetch dependencies
+        run: sudo apt install g++ flex bison cmake ninja-build maven jq xmllint dpkg-dev
+      - name: Configure CMake
+        run: |
+          mkdir build
+          cd build
+          cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release
+      - name: Build using Ninja
+        run: |
+          cd build
+          ninja
+      - name: Run CTest
+        run: cd build; ctest -V .
+      - name: Create packages
+        id: create_packages
+        run: |
+          cd build
+          ninja package
+          deb_package_name="$(ls *.deb)"
+          echo "::set-output name=deb_package::./build/$deb_package_name"
+          echo "::set-output name=deb_package_name::$deb_package_name"
+      - name: Get release info
+        id: get_release_info
+        uses: bruceadams/get-release@v1.2.0
+      - name: Upload binary packages
+        uses: actions/upload-release-asset@v1
+        with:
+          upload_url: ${{ steps.get_release_info.outputs.upload_url }}
+          asset_path: ${{ steps.create_packages.outputs.deb_package }}
+          asset_name: ${{ steps.create_packages.outputs.deb_package_name }}
+          asset_content_type: application/x-deb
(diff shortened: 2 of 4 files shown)


hannes-steffenhagen-diffblue:
The windows build here is broken not because of any of these changes, but because sourceforge links we were downloading packages from went stale.

NlightNFotis:
This is a sick PR by the way - this has the potential to make all our packaging super easy if I understood correctly?

So it's possible now to build a package in all platforms just by doing a `build package` (or `ninja package`)?

Nicely done in any case 💯

martin-cs (approved):
This would be a good piece of functionality.  I assume you have spoken to @tautschnig about this as he is the current maintainer for the Debian package : https://tracker.debian.org/pkg/cbmc

hannes-steffenhagen-diffblue:
@NlightNFotis Not _all_ our packaging, unfortunately. But we can produce debian binary packages, WiX installers, simple binary zips etc from this.

For an Ubuntu PPA we would need to produce debian _source_ packages apparently and I am not completely sure what's involved with that yet.

@martin-cs I wasn't actually aware of that.

martin-cs:
@hannes-steffenhagen-diffblue he has the power to get a package into Debian (and thus, I think, downstream, to Ubuntu) plus has all of the control.tar.gz meta-information you need to get through the Debian linter.

hannes-steffenhagen-diffblue:
@martin-cs Our short term goal right now is actually to have an ‘unstable’ Ubuntu PPA so that people can keep their cbmc up-to-date through that. If I understand the process correctly – and I am not completely sure I do – this wouldn’t really work with the “normal” mainstream Debian or Ubuntu release process. Checking out your link it seems that @tautschnig has some scripts to keep debian-unstable up to date, but if I understood correctly using that would mean also pulling in a lot of other debian-unstable dependencies. We’re trying to get it working with stable Ubuntu dependencies, but rolling releases for cbmc (if that makes sense).

I think we should talk to Michael regardless because of course long term making sure cbmc stays up to date in Ubuntu/Debian mainline is also important.

martin-cs:
@hannes-steffenhagen-diffblue having your own package repo and getting people to pull from there seems like a positive step and a good first target.  Longer-term I believe it might be possible (with the assistance of @tautschnig or at least @tautschnig 's scripts) to have the unstable package up-to-date and (as long as it doesn't have any dependencies beyond the current stable) set the repository priority so that you get just that one and not accidentally upgrade everything to unstable.

cmake/packaging.cmake
```diff
@@ -0,0 +1,34 @@
+set(CPACK_PACKAGE_NAME "cbmc") # XXX should this be cprover instead?
+set(CPACK_PACKAGE_VENDOR "Diffblue Ltd.")
+set(CPACK_PACKAGE_CONTACT "kroening@cs.ox.ac.uk") # XXX should this be a Diffblue email address instead?
```
hannes-steffenhagen-diffblue: Clarification on this point would be great, this is the address used in in `src/cbmc/dist-linux`, the current Ubuntu package lists “Ubuntu Developers” as the maintainer and @tautschnig as the ‘Original Maintainer’, so I’m genuinely unsure what to put here.

.github/workflows/pull-request-checks.yaml
```diff
@@ -60,6 +60,11 @@ jobs:
           cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++
       - name: Build with Ninja
         run: cd build; ninja
+      - name: Check if package building works
+        run: |
+          cd build
+          ninja package
+          ls *.deb
```
chrisr-diffblue, line 67: What does the `ls *.deb` achieve, other than a nice output? Will it fail the build in someway? Or will it mask any failures from the `ninja package` because the exit code of the `ls` will be success?

cmake/packaging.cmake
chrisr-diffblue: Can you put multiple contacts in? Might also be worth asking @peterschrammel for an opinion too.

chrisr-diffblue (approved):
Generally looks good - as you say, you need to clarify some details but I'm not sure I have answers for you there.

.github/workflows/pull-request-checks.yaml
thomasspriggs: My experience of debugging the run sections of github actions suggests that it will stop on the first line where an error code is returned. Similar to writing a bash script where `set -e` has been specified. Therefore this extra line should not mask any other failures.

.github/workflows/release-packages.yaml
```diff
@@ -0,0 +1,45 @@
...
+        run: sudo apt install g++ flex bison cmake ninja-build maven jq xmllint dpkg-dev
+      - name: Configure CMake
+        run: |
+          mkdir build
+          cd build
+          cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release
+      - name: Build using Ninja
+        run: |
+          cd build
+          ninja
```
thomasspriggs, line 25: :bulb: This could be done as a single line -
`run: ninja -C build`

thomasspriggs (approved):
LGTM.

It may be worth us investigating whether we can use `cpack` to make the `.msi` for the windows packaging as well. @markrtuttle added a handwritten xml file with the configuration to create it with `WiX`. But there is a `WiX` plugin for cpack which could potentially generate it for us and reduce the amount of packaging setup which would otherwise be duplicated.

cmake/packaging.cmake
```diff
@@ -0,0 +1,34 @@
+set(CPACK_PACKAGE_NAME "cbmc") # XXX should this be cprover instead?
```
peterschrammel: I think `cbmc` is more well-known.

cmake/packaging.cmake
peterschrammel: If possible I'd put `info@diffblue.com` to have a stable address.

peterschrammel (approved):

.github/workflows/pull-request-checks.yaml
hannes-steffenhagen-diffblue: @chrisr-diffblue `ls` returns a non-zero exit code if called on non-existent files... so this is just checking whether or not there is a file ending with `.deb` in the directory after `ninja package`.  See [here](https://github.com/hannes-steffenhagen-diffblue/github-actions-test/pull/4/checks?check_run_id=1011772782) for a demonstration of how this works.

Failures propagate in github action scripts because I believe `set -e` is set by default.

hannes-steffenhagen-diffblue:
@thomasspriggs Yes, my plan was to try the `WiX` generation of CPack to reduce our maintenance burden on these things.

PR #5475, opened 2020-08-26
CBMC additional profiling info: Pointer dereference, points-to set metric

natasha-jeppu:
Parse option to track size and contents of points-to sets during pointer dereferencing.

### DIFF
--- a/src/pointer-analysis/goto_program_dereference.h
+++ b/src/pointer-analysis/goto_program_dereference.h
@@ -36,7 +36,7 @@ public:
     : options(_options),
       ns(_ns),
       value_sets(_value_sets),
-      dereference(_ns, _new_symbol_table, *this, ID_nil, false)
+      dereference(_ns, _new_symbol_table, *this, ID_nil, false, messaget())
   {
   }
 
--- a/src/pointer-analysis/goto_program_dereference.h
+++ b/src/pointer-analysis/goto_program_dereference.h
@@ -28,15 +28,24 @@ public:
   // for the final argument to value_set_dereferencet.
   // This means that language-inappropriate values such as
   // (struct A*)some_integer_value in Java, may be returned.
+  // Note: value_set_dereferencet requires a messaget instance
+  // as on of its arguments to display the points-to set
+  // during symex. Display is not done during goto-program
+  // conversion. To ensure this the display_points_to_sets
+  // parameter in value_set_dereferencet::dereference()
+  // is set to false by default and is not changed by the
+  // goto program conversion modules. Similarly, here we set
+  // _log to be a default messaget instance.
   goto_program_dereferencet(
     const namespacet &_ns,
     symbol_tablet &_new_symbol_table,
     const optionst &_options,
-    value_setst &_value_sets)
+    value_setst &_value_sets,
+    const messaget &_log = messaget())
     : options(_options),
       ns(_ns),
       value_sets(_value_sets),
-      dereference(_ns, _new_symbol_table, *this, ID_nil, false, messaget())
+      dereference(_ns, _new_symbol_table, *this, ID_nil, false, _log)
   {
   }
 
--- a/src/pointer-analysis/value_set_dereference.cpp
+++ b/src/pointer-analysis/value_set_dereference.cpp
@@ -11,10 +11,9 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "value_set_dereference.h"
 
-#ifdef DEBUG
 #include <iostream>
 #include <util/format_expr.h>
-#endif
+#include <util/json_irep.h>
 
 #include <util/arith_tools.h>
 #include <util/array_name.h>
@@ -60,7 +59,7 @@ static bool should_use_local_definition_for(const exprt &expr)
   return false;
 }
 
-exprt value_set_dereferencet::dereference(const exprt &pointer)
+exprt value_set_dereferencet::dereference(const exprt &pointer, bool display_points_to_sets)
 {
   if(pointer.type().id()!=ID_pointer)
     throw "dereference expected pointer type, but got "+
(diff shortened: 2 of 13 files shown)


src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -224,6 +261,15 @@ exprt value_set_dereferencet::dereference(const exprt &pointer)
   if(compare_against_pointer != pointer)
     value = let_exprt(to_symbol_expr(compare_against_pointer), pointer, value);
 
+  if(display_points_to_sets)
+  {
+    std::stringstream ss;
+    ss << format(value);
+    json_result["Value"] = json_stringt(ss.str());
+
+    std::cout << ",\n" << json_result;
```
peterschrammel: 🚫 This is not allowed. Pass a message handler to the class instead, as done everywhere else.

martin-cs (commented):
1. There are already a number of ways of showing the points-to sets in general.
2. If it must specifically be done during symbolic execution, why not just add it as a call to the right level of logging.  I don't see why we need extra options.

FYI : there has been a long-standing goal of minimising the number of options that CPROVER tools have or at least preventing the proliferation of them.  We haven't done so well of late but it is still a worthwhile goal to pursue.

natasha-jeppu:
> 1. There are already a number of ways of showing the points-to sets in general.

Are you referring to the --show-value-sets option? As I understand, the CBMC call just looks at the points-to set information which is recorded during symex and this is what we are interested in.

danielsn:
> 1. There are already a number of ways of showing the points-to sets in general.
> 2. If it must specifically be done during symbolic execution, why not just add it as a call to the right level of logging.  I don't see why we need extra options.
> 
> FYI : there has been a long-standing goal of minimising the number of options that CPROVER tools have or at least preventing the proliferation of them. We haven't done so well of late but it is still a worthwhile goal to pursue.

I wonder if it would make sense to have a mechanism for defining what statistics to collect / what output to generate, that could go under a single commandline flag.
Ideas are: 
1. A json config file
2. A `--show-metric-in-log metric` flag that takes a metric, or list of metrics.

src/pointer-analysis/value_set_dereference.cpp
natasha-jeppu: Fixed this.

peterschrammel:
If it's just about showing _metrics_ that are cheap to collect (e.g. points to set size) then I would just always output them - no flag needed. Such metrics are indeed very useful.
If you are also going to show the points to sets themselves then I'd prefer to put them behind a flag, as this produces _massive_ output... and I'm not sure what it is good for except for very special debug situations.

natasha-jeppu:
> If it's just about showing _metrics_ that are cheap to collect (e.g. points to set size) then I would just always output them - no flag needed. Such metrics are indeed very useful.
> If you are also going to show the points to sets themselves then I'd prefer to put them behind a flag, as this produces _massive_ output... and I'm not sure what it is good for except for very special debug situations.

For now both are behind a flag. We prefer to keep this behind a flag because even with just the points-to set size, we can get a really big output (multiple pointer dereferences and number of pointers may be large). But if you prefer to have it as part of the standard output we could make the change.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -11,10 +11,9 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "value_set_dereference.h"
 
-#ifdef DEBUG
```
peterschrammel, line 14: ❗Please don't remove the ifdef here. `iostream` must not be included outside `DEBUG` here.

You can use `iosfwd` outside `DEBUG` - that's what you need here, as far as I can see.

src/pointer-analysis/goto_program_dereference.h
```diff
@@ -36,7 +36,7 @@ class goto_program_dereferencet:protected dereference_callbackt
     : options(_options),
       ns(_ns),
       value_sets(_value_sets),
-      dereference(_ns, _new_symbol_table, *this, ID_nil, false)
+      dereference(_ns, _new_symbol_table, *this, ID_nil, false, messaget())
```
peterschrammel: ❗ Passing a default `messaget` instance doesn't look right here. The instance should come from the caller.

natasha-jeppu: Is this a valid construct? The points-to sets are displayed during symex only. `exprt dereference(const exprt &pointer, bool display_points_to_sets = false);` sets `display_points_to_set` to false by default so any calls to `value_set_dereferencet::dereference()` from the goto program conversion phase does not display the points-to sets. Similarly, is it okay to set the `messaget` here to a default message instance for `goto_program_dereferencet`?
```
goto_program_dereferencet(
    const namespacet &_ns,
    symbol_tablet &_new_symbol_table,
    const optionst &_options,
    value_setst &_value_sets,
    const messaget &_log = messaget())
    : options(_options),
      ns(_ns),
      value_sets(_value_sets),
      dereference(_ns, _new_symbol_table, *this, ID_nil, false, _log)
  {
  }
```

Passing a valid messaget object here`goto_program_dereferencet` will need a lot of additional refactoring across the goto program conversion code modules.

peterschrammel: Ok, thanks for the explanation. Please put a code comment to clarify why a default messaget instance is passed in this place. The next person stumbling over that code should not have to redo your reasoning. Many thanks!

peterschrammel (approved):

peterschrammel:
Approved assuming that the last two comments will be fixed before merging.

src/pointer-analysis/goto_program_dereference.h
natasha-jeppu: Should I retain the current implementation i.e `dereference(_ns, _new_symbol_table, *this, ID_nil, false, messaget())` or replace it with the construct above? 
I will add in the comment.

peterschrammel: The one above is probably slightly cleaner as it pushes the messaget further up. But, still, please put a comment in goto_program_dereferencet then.

natasha-jeppu: Fixed. Comment added.

src/pointer-analysis/value_set_dereference.cpp
natasha-jeppu: Fixed.

PR #5476, opened 2020-08-26
CBMC additional profiling info: Track memcmp, memcpy and memset function calls

natasha-jeppu:
Memset, memcmp and memcpy functions often increase proof runtime and need to be replaced by stubs. Here we track these calls for ease of identification and subsequent replacement by stubs.

### DIFF
--- a/regression/cbmc/memop-calls/main.c
+++ b/regression/cbmc/memop-calls/main.c
@@ -1,0 +1,9 @@
+#include<string.h>
+int main()
+{
+	const int a[2] = {0, 0};
+	int b[2];
+
+	memcpy(a, b, 3);
+	return 0;
+}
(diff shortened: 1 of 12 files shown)


martin-cs (commented):
1. I don't think this needs to be added to all executables, just 'goto-instrument' should be fine.
2. Can't this be handled by post-processing the output of `--show-goto-programs` or similar?

peterschrammel (commented):
If you want to list the calls then goto-instrument is probably the better place for this functionality. If you want to report metrics (number of various calls) then this can be done in CBMC, but probably doesn't require an extra flag.

natasha-jeppu:
> If you want to list the calls then goto-instrument is probably the better place for this functionality. If you want to report metrics (number of various calls) then this can be done in CBMC, but probably doesn't require an extra flag.

So now I'm just using the --show-goto-functions output to list these. This PR is no longer valid, so I have closed it.

PR #5478, opened 2020-08-26
CBMC additional profiling info: Count and report all array constraints added during postprocessing

natasha-jeppu:
Track constraints added for arrays in post processing as a potentially useful metric for performance evaluation.


src/solvers/flattening/arrays.h
```diff
@@ -77,6 +82,27 @@ class arrayst:public equalityt
...
+    ARRAY_CONSTANT,
+    ARRAY_NON_CONSTANT,
+    ARRAY_COMPREHENSION
+  };
+
+  typedef std::list<exprt> array_constraintst;
+  typedef std::map<constraint_typet, array_constraintst> array_constraints_mapt;
+  array_constraints_mapt array_constraints_map;
+  void display_constraints();
+  std::string enum_to_string(constraint_typet);
```
natasha-jeppu: Is there a better way of doing this? Wanted to include more specific information regarding the type of constraint. For example split constraints added for array_with to constraints added for the specific index and for other indices.

src/solvers/flattening/arrays.h
```diff
@@ -77,6 +82,27 @@ class arrayst:public equalityt
...
+    ARRAY_OF,
+    ARRAY_TYPECAST,
+    ARRAY_CONSTANT,
+    ARRAY_NON_CONSTANT,
+    ARRAY_COMPREHENSION
+  };
+
+  typedef std::list<exprt> array_constraintst;
+  typedef std::map<constraint_typet, array_constraintst> array_constraints_mapt;
+  array_constraints_mapt array_constraints_map;
```
peterschrammel: ❗Storing copies of these constraints is not ideal wrt memory consumption.

peterschrammel (commented):
I thought this was about reporting metrics, but this PR stores copies of constraints and outputs them - what is the use case for this very detailed information?

natasha-jeppu:
> I thought this was about reporting metrics, but this PR stores copies of constraints and outputs them - what is the use case for this very detailed information?

The intention was to investigate how array constraints added during post processing affects CBMC performance. 

In order to see if the constraints themselves give any information, I tracked them here. But this is not very helpful indeed. I will be changing this to track just the _number_ of constraints rather than the constraints themselves.

Let me know if you have any other metric in mind pertaining to this that could be useful.

src/solvers/flattening/arrays.h
```diff
@@ -77,6 +82,27 @@ class arrayst:public equalityt
...
+    ARRAY_WITH,
+    ARRAY_WITH_OTHER,
+    ARRAY_IF_TRUE,
+    ARRAY_IF_FALSE,
+    ARRAY_OF,
+    ARRAY_TYPECAST,
+    ARRAY_CONSTANT,
+    ARRAY_NON_CONSTANT,
+    ARRAY_COMPREHENSION
+  };
```
natasha-jeppu: Is there a clean way of extending the existing `lazy_typet` enum class into a new enum class to include `ARRAY_EQUALITY`? Do not want to change the meaning of the `lazy_typet` by adding `ARRAY_EQUALITY` to the existing enum class.

src/solvers/flattening/arrays.cpp
```diff
@@ -825,9 +853,75 @@ void arrayst::add_array_constraints_if(
...
+      constraint_it++;
+      num_constraints++;
+    }
+    it++;
+  }
+
+  json_result["numOfConstraints"] =
+    json_numbert(std::to_string(num_constraints));
+  log.status() << ",\n" << json_result;
+}
```
natasha-jeppu, line 904: What is an elegant way to exit CBMC once the display is done? Adding `exit(CPROVER_EXIT_SUCCESS)` here makes the json output messy.

tautschnig: You must not exit directly from here. This code may be called in a number of ways, and it will be up to the caller to exit in a clean manner. As far as I understand, currently `post_process()` is the only place calling this. Now I'd question whether this should really be called _from_ `post_process()`, or rather be called from some place _after_ `post_process()` has been called.

src/solvers/flattening/arrays.h
natasha-jeppu: @peterschrammel

src/solvers/flattening/arrays.h
natasha-jeppu: No longer applicable, changed to store only number of constraints, don't store the constraints themselves.

src/solvers/flattening/arrays.cpp
natasha-jeppu: Noted.

regression/cbmc/array-constraint/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+--show-array-constraints
+^EXIT=1$
+^SIGNAL=0$
+--
+--
+To test output for --show-array-constraints option that displays the count and list of all array constraints added during post processing in json format.
```
peterschrammel: ⛏️  missing newline at end of file

regression/cbmc/array-constraint/test_xml.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+--show-array-constraints --xml-ui
+^EXIT=1$
+^SIGNAL=0$
+--
+--
+To test output for --show-array-constraints option that displays the count and list of all array constraints added during post processing in json format.
```
peterschrammel: ⛏️  missing newline at end of file

src/solvers/flattening/arrays.cpp
peterschrammel: If you want to terminate without sat-solving then the intended mechanism is to do that by using the symex-only checkers. Those run the pipeline until after symex and then terminate.
You could possibly do that in the `src/goto-checker/multi_path_symex_only_checker.cpp` which is used to handles similar cases such as `--show-vcc`, `--show-program`, etc.

natasha-jeppu: The array constraints are added during post-processing right, so this would not help here. We need it to complete symex, convert SSA and then post process as well before terminating.

peterschrammel: True we would need to introduce an intermediate pipeline that stops later. Not sure whether this is worth doing...

peterschrammel (approved):

regression/cbmc/array-constraint/test.desc
natasha-jeppu: Fixed

regression/cbmc/array-constraint/test_xml.desc
natasha-jeppu: Fixed

src/solvers/flattening/arrays.cpp
natasha-jeppu: This might be useful for `--write-solver-stats-to` as well. When you have very slow proofs, you may not want to wait to get solver query statistics. I'm assuming here that all relevant info is already collected before the call to the solver.

PR #5480, opened 2020-08-26
CBMC additional profiling info: Extend solver hardness to track clauses mapped to an instruction

natasha-jeppu:
Currently solver hardness tracks number of clauses. This is an extension of solver hardness to track the clauses themselves. Provides a mapping between clauses and instructions.

### DIFF
--- a/src/solvers/sat/satcheck_minisat2.cpp
+++ b/src/solvers/sat/satcheck_minisat2.cpp
@@ -140,8 +140,10 @@ void satcheck_minisat2_baset<T>::lcnf(const bvt &bv)
 
     solver->addClause_(c);
 
-    with_solver_hardness(
-      [&bv](solver_hardnesst &hardness) { hardness.register_clause(bv); });
+    size_t solver_clause_num = clause_counter;
+    with_solver_hardness([&bv, &solver_clause_num](solver_hardnesst &hardness) {
+      hardness.register_clause(bv, solver_clause_num);
+    });
     clause_counter++;
   }
   catch(const Minisat::OutOfMemoryException &)
--- a/src/solvers/sat/satcheck_minisat2.cpp
+++ b/src/solvers/sat/satcheck_minisat2.cpp
@@ -140,10 +140,22 @@ void satcheck_minisat2_baset<T>::lcnf(const bvt &bv)
 
     solver->addClause_(c);
 
-    size_t solver_clause_num = clause_counter;
-    with_solver_hardness([&bv, &solver_clause_num](solver_hardnesst &hardness) {
-      hardness.register_clause(bv, solver_clause_num);
-    });
+    // To map clauses to lines of program code, track clause indices in the
+    // dimacs cnf output. Dimacs output is generated after processing clauses
+    // to remove duplicates and clauses that are trivially true. Here, a clause
+    // is checked to see if it can be thus eliminated. If not, add the clause
+    // index to list of clauses in solver_hardnesst::register_clause().
+    static size_t cnf_clause_index = 0;
+    bvt cnf;
+    bool clause_removed = process_clause(bv, cnf);
+
+    if(!clause_removed)
+      cnf_clause_index++;
+
+    with_solver_hardness(
+      [&bv, &cnf, &clause_removed](solver_hardnesst &hardness) {
+        hardness.register_clause(bv, cnf, cnf_clause_index, !clause_removed);
+      });
     clause_counter++;
   }
   catch(const Minisat::OutOfMemoryException &)
--- a/src/solvers/solver_hardness.cpp
+++ b/src/solvers/solver_hardness.cpp
@@ -108,6 +108,8 @@ void solver_hardnesst::register_clause(
   std::cout << "0\n";
 #endif
 
+  // The clause_counter is incremented after the call to register_clause.
+  // Here we add 1 to solver_clause_num to incorporate this increment.
   current_hardness.clause_set.push_back(solver_clause_num + 1);
   std::sort(
     current_hardness.clause_set.begin(), current_hardness.clause_set.end());
--- a/src/solvers/solver_hardness.cpp
+++ b/src/solvers/solver_hardness.cpp
@@ -26,8 +26,8 @@ operator+=(const solver_hardnesst::sat_hardnesst &other)
   clauses += other.clauses;
   literals += other.literals;
   variables.insert(other.variables.begin(), other.variables.end());
-  clause_set.insert(clause_set.end(),
-    other.clause_set.begin(), other.clause_set.end());
+  clause_set.insert(
+    clause_set.end(), other.clause_set.begin(), other.clause_set.end());
   return *this;
 }
 
--- a/src/solvers/solver_hardness.cpp
+++ b/src/solvers/solver_hardness.cpp
@@ -26,6 +26,8 @@ operator+=(const solver_hardnesst::sat_hardnesst &other)
   clauses += other.clauses;
   literals += other.literals;
   variables.insert(other.variables.begin(), other.variables.end());
+  clause_set.insert(clause_set.end(),
+    other.clause_set.begin(), other.clause_set.end());
   return *this;
 }
 
@@ -91,10 +93,20 @@ void solver_hardnesst::register_clause(const bvt &bv)
 {
   current_hardness.clauses++;
   current_hardness.literals += bv.size();
+  std::vector<int> clause;
+
   for(const auto &literal : bv)
   {
     current_hardness.variables.insert(literal.var_no());
+
+    int signed_literal = literal.var_no();
+    if(literal.sign())
+      signed_literal = -signed_literal;
+    clause.push_back(signed_literal);
   }
+  clause.push_back(0);
+  std::sort(clause.begin(), clause.end());
+  current_hardness.clause_set.push_back(clause);
 }
 
 void solver_hardnesst::set_outfile(const std::string &file_name)
@@ -151,6 +163,16 @@ void solver_hardnesst::produce_report()
       sat_hardness_json["Variables"] =
         json_numbert{std::to_string(hardness.variables.size())};
 
+      json_arrayt sat_hardness_clause_set_json;
+      for(auto const &clause : hardness.clause_set)
+      {
+        json_arrayt clause_json;
+        for(auto const &lit : clause)
+          clause_json.push_back(json_numbert{std::to_string(lit)});
+        sat_hardness_clause_set_json.push_back(clause_json);
+      }
+      sat_hardness_json["ClauseSet"] = sat_hardness_clause_set_json;
+
       hardness_stats_json["SAT_hardness"] = sat_hardness_json;
 
       json_stream_array.push_back(hardness_stats_json);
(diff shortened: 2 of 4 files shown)


src/solvers/solver_hardness.cpp
```diff
@@ -91,10 +93,20 @@ void solver_hardnesst::register_clause(const bvt &bv)
 {
   current_hardness.clauses++;
   current_hardness.literals += bv.size();
+  std::vector<int> clause;
```
danielsn: `int`? or do we want a type with a defined length like `int_64`?

src/solvers/solver_hardness.cpp
```diff
@@ -91,10 +93,20 @@ void solver_hardnesst::register_clause(const bvt &bv)
...
   {
     current_hardness.variables.insert(literal.var_no());
+
+    int signed_literal = literal.var_no();
+    if(literal.sign())
+      signed_literal = -signed_literal;
+    clause.push_back(signed_literal);
   }
+  clause.push_back(0);
+  std::sort(clause.begin(), clause.end());
```
danielsn: Do we want to sort before or after we add the `0`?

src/solvers/solver_hardness.cpp
```diff
@@ -91,10 +93,20 @@ void solver_hardnesst::register_clause(const bvt &bv)
...
   current_hardness.literals += bv.size();
+  std::vector<int> clause;
+
   for(const auto &literal : bv)
   {
     current_hardness.variables.insert(literal.var_no());
+
+    int signed_literal = literal.var_no();
+    if(literal.sign())
+      signed_literal = -signed_literal;
```
danielsn: Any chance this could overflow?

src/solvers/solver_hardness.cpp
```diff
@@ -151,6 +163,16 @@ void solver_hardnesst::produce_report()
       sat_hardness_json["Variables"] =
         json_numbert{std::to_string(hardness.variables.size())};
 
+      json_arrayt sat_hardness_clause_set_json;
+      for(auto const &clause : hardness.clause_set)
+      {
+        json_arrayt clause_json;
+        for(auto const &lit : clause)
+          clause_json.push_back(json_numbert{std::to_string(lit)});
```
danielsn: `json_numbert` takes a string?

src/solvers/solver_hardness.cpp
```diff
@@ -26,6 +26,8 @@ operator+=(const solver_hardnesst::sat_hardnesst &other)
   clauses += other.clauses;
   literals += other.literals;
   variables.insert(other.variables.begin(), other.variables.end());
+  clause_set.insert(
```
danielsn, line 29: Should we sort this?

src/solvers/solver_hardness.cpp
danielsn: Should we typedef this?

src/solvers/solver_hardness.cpp
natasha-jeppu: Post processing of the SAT core cnf runs sort for the entire clause, including the zero. So might as well sort it here too.

src/solvers/solver_hardness.cpp
natasha-jeppu: This is what is used inside CBMC already for the literal.dimacs() function. I'm assuming here that they have already taken into consideration any potential overflow issues. And also, they define the return type as int.

src/solvers/solver_hardness.cpp
natasha-jeppu: Yes, takes a string but the output is formatted json number.

src/solvers/solver_hardness.cpp
natasha-jeppu: I was worried about too many calls to sort().

src/solvers/solver_hardness.cpp
peterschrammel: ❓ What are the memory and performance implications of storing a copy of the clauses?

natasha-jeppu: I haven't explored this yet. Although a much more efficient approach would be to store some form of index (specifically an index interval) to some existing clause storage, but I am not sure if this can be done.

peterschrammel: What are the metrics that you want to compute based on the clauses?

natasha-jeppu: To identify which program lines contribute to the SAT core. So given this infrastructure we can map clauses from the SAT core directly to LOC.

peterschrammel: What is a SAT core?

natasha-jeppu: Referring to unsatisfiable core here.
https://en.wikipedia.org/wiki/Unsatisfiable_core

peterschrammel: How are you going to compute the UNSAT core?

natasha-jeppu: There are tools which compute the UNSAT core given a CNF formula and DRAT proof of unsatisfiability (these proofs have become a standard in the SAT competitions). CBMC is used to get the dimacs CNF formula (--dimacs) and the clause to LOC mapping with --write-solver-stats-to option.

peterschrammel: If this can only be used wiith dimacs couldn't you just store the index (eg. line of the clause) in the dimacs output?

natasha-jeppu: How do I access the dimacs clause index here? Also is the dimacs output ordered?
I will need to map lines of program code to indices in the dimacs output.

peterschrammel: You can make `dimacs_cnft` (or probably `cnf_clause_listt`) implement `hardness_collectort`(in the same way other sat solvers implement that interface). It seems that the clauses are stored in order in `cnf_clause_listt`, so `dimacs_cnft` will print them in the same order as you get the `register_clause` calls. Of course, you'll have to test and confirm that.

src/solvers/sat/satcheck_minisat2.cpp
```diff
@@ -140,8 +140,10 @@ void satcheck_minisat2_baset<T>::lcnf(const bvt &bv)
 
     solver->addClause_(c);
 
-    with_solver_hardness(
-      [&bv](solver_hardnesst &hardness) { hardness.register_clause(bv); });
+    size_t solver_clause_num = clause_counter;
+    with_solver_hardness([&bv, &solver_clause_num](solver_hardnesst &hardness) {
+      hardness.register_clause(bv, solver_clause_num);
```
natasha-jeppu: Pass the clause counter to hardness.register_clause(). Store the clause counter value rather than the clause themselves.

src/solvers/solver_hardness.cpp
```diff
@@ -89,24 +89,28 @@ void solver_hardnesst::register_assertion_ssas(
...
-  clause.push_back(0);
-  std::sort(clause.begin(), clause.end());
-  current_hardness.clause_set.push_back(clause);
+
+#ifdef DEBUG
+  std::cout << solver_clause_num << ": ";
+  for(const auto &literal : bv)
+    std::cout << literal.dimacs() << " ";
+  std::cout << "0\n";
+#endif
```
natasha-jeppu: Debug statements to output 'clause_counter_value: clause' to cross check if they match the corresponding lines in dimacs output.

src/solvers/solver_hardness.cpp
natasha-jeppu: I now keep track of the clause counter in `satcheck_minisat2_baset<T>::lcnf` that calls `register_clause` and store the counter value rather than the clause itself in solver hardness. Have added optional DEBUG statements to verify if these match corresponding lines in the dimacs output.

src/solvers/sat/satcheck_minisat2.cpp
```diff
@@ -140,8 +140,10 @@ void satcheck_minisat2_baset<T>::lcnf(const bvt &bv)
 
     solver->addClause_(c);
 
-    with_solver_hardness(
-      [&bv](solver_hardnesst &hardness) { hardness.register_clause(bv); });
+    size_t solver_clause_num = clause_counter;
```
peterschrammel: Assignment seems redundant, just use `clause_counter` in line 145.

src/solvers/solver_hardness.h
```diff
@@ -50,6 +50,7 @@ struct solver_hardnesst
     size_t clauses = 0;
     size_t literals = 0;
     std::unordered_set<size_t> variables = {};
+    std::vector<int> clause_set = {};
```
peterschrammel:
```suggestion
    std::vector<size_t> clause_set = {};
```

src/solvers/solver_hardness.cpp
```diff
@@ -87,14 +89,28 @@ void solver_hardnesst::register_assertion_ssas(
...
   }
+
+#ifdef DEBUG
+  std::cout << solver_clause_num << ": ";
+  for(const auto &literal : bv)
+    std::cout << literal.dimacs() << " ";
+  std::cout << "0\n";
+#endif
+
+  current_hardness.clause_set.push_back(solver_clause_num + 1);
```
peterschrammel: Please add a comment to explain why +1.

src/solvers/sat/satcheck_minisat2.cpp
natasha-jeppu: `clause_counter` is an inherited data member of `satcheck_minisat2_baset<T>` from `cnf_solvert` so I cannot pass it directly to the lambda capture list (not in scope). Will need to pass `this` instead. Thought it would be better to just pass a copy.

src/solvers/solver_hardness.cpp
natasha-jeppu: Done.

src/solvers/solver_hardness.h
natasha-jeppu: Done.

peterschrammel (approved):

src/solvers/sat/satcheck_minisat2.cpp
```diff
@@ -140,8 +140,22 @@ void satcheck_minisat2_baset<T>::lcnf(const bvt &bv)
 
     solver->addClause_(c);
 
+    // To map clauses to lines of program code, track clause indices in the
```
martin-cs, line 143: I don't want to be a downer on this but ...

1. There is not necessarily a 1-to-1 correlations between clauses and lines of code.  For example, rewriting and normalisation may mean that multiple lines reduce to the same shared term.  Which line should this be credited to?  This is worse when it is things like multiplication where additions from other operations can appear as sub-parts of the operator.

2. Number of variables and clauses is a *very* crude measure of SAT difficulty.  Reducing the number of clauses can make the problem harder; I have seen multiple instances of this.

src/solvers/solver_hardness.h
```diff
@@ -105,7 +106,15 @@ struct solver_hardnesst
   /// Called e.g. from the `satcheck_minisat2::lcnf`, this function adds the
   ///   complexity statistics from the last SAT query to the `current_ssa_key`.
   /// \param bv: the clause (vector of literals)
-  void register_clause(const bvt &bv);
+  /// \param cnf: processed clause
+  /// \param cnf_clause_index: index of clause in dimacs output
+  /// \param register_cnf: negation of boolean variable tracking if the clause
```
martin-cs, line 111: Please clarify this comment.

src/solvers/solver_hardness.h
```diff
@@ -50,6 +50,7 @@ struct solver_hardnesst
     size_t clauses = 0;
     size_t literals = 0;
     std::unordered_set<size_t> variables = {};
+    std::vector<size_t> clause_set = {};
```
martin-cs, line 53: What is the performance cost of this?

martin-cs (commented):
I have done quite a bit of performance optimisation work on CBMC.  If you want a chat about it then drop me a line.

PR #5482, opened 2020-08-27
Add check of making `.deb` source package on pull requests

thomasspriggs:
Add check of making `.deb` source package on pull requests


cmake/CPackGeneratorConfig.cmake
```diff
@@ -0,0 +1,9 @@
+# This is run once per generator, which means that generator specific logic can
+# go here.
```
hannes-steffenhagen-diffblue, line 2: 👍 sounds like the right approach on principle

hannes-steffenhagen-diffblue:
@thomasspriggs looks like sources for source packages need to be one level deeper, e.g. at the top level of the source package there should only be a single directory containing the source code, I think the name of that is supposed to be `<package name>-<version name>`

cmake/CPackGeneratorConfig.cmake
```diff
@@ -5,5 +5,5 @@ message(STATUS "Configuring for generator ${CPACK_GENERATOR} and TOPLEVEL_TAG ${
 
 if("${CPACK_TOPLEVEL_TAG}" STREQUAL "Linux-Source")
   message(STATUS "Adding python dependency")
-  set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS},python3")
+  set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS},python")
```
hannes-steffenhagen-diffblue: Can you explain this one? Is this installing both `python3` and `python2`?

thomasspriggs: The `python` dependency is for python 2.7. The `python3` dependency is for python 3.8. We actually need both to pass lintian, as we have some python scripts which start with `#!/usr/bin/env python` and with `#!/usr/bin/env python3`.

PR #5483, opened 2020-08-27
CBMC additional profiling info: Extend standard CBMC output to display runtime for different stages of CBMC

natasha-jeppu:
Extend CBMC output to include runtime for different stages of CBMC:
1. symex
2. convert SSA
3. post processing
4. solver

### DIFF
--- a/regression/cbmc/runtime-profiling/test.desc
+++ b/regression/cbmc/runtime-profiling/test.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+
+^EXIT=0$
+^SIGNAL=0$
+^Runtime Symex:.*$
+^Runtime Convert SSA:.*$
+^Runtime Post-process:.*$
+^Runtime Solver:.*$
+--
+--
+To test runtime profiling output.
--- a/regression/cbmc/runtime-profiling/test.desc
+++ b/regression/cbmc/runtime-profiling/test.desc
@@ -3,10 +3,7 @@ main.c
 
 ^EXIT=0$
 ^SIGNAL=0$
-^Runtime Symex:.*$
-^Runtime Convert SSA:.*$
-^Runtime Post-process:.*$
-^Runtime Solver:.*$
+^VERIFICATION SUCCESSFUL$
 --
 --
 To test runtime profiling output.
--- a/regression/cbmc/runtime-profiling/test.desc
+++ b/regression/cbmc/runtime-profiling/test.desc
@@ -1,9 +1,12 @@
-CORE
+KNOWNBUG test-paths-lifo test-cprover-smt2
 main.c
 
 ^EXIT=0$
 ^SIGNAL=0$
-^VERIFICATION SUCCESSFUL$
+^Runtime Symex:.*$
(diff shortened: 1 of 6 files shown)


peterschrammel (approved):

regression/cbmc/runtime-profiling/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+To test runtime profiling output.
```
martin-cs, line 12: I don't see anything in this test that actually checks this.

martin-cs (commented):
How does this differ to `--timestamp` or piping the output through `ts` or similar?

regression/cbmc/runtime-profiling/test.desc
natasha-jeppu: > How does this differ to `--timestamp` or piping the output through `ts` or similar?

With the --timestamp option, we get timestamps for every line of output from CBMC, but this may not give the correct duration of a specific stage in CBMC as it may include more than just a call to the function that performs operations for that specific stage. So here we look at the duration of a phase in CBMC, for eg post processing, by tracking the monotonic clock time right before and right after the call to `post_process()`. Similarly, for other phases: symex, ssa conversion and solver.

natasha-jeppu:
> How does this differ to `--timestamp` or piping the output through `ts` or similar?

With the --timestamp option, we get timestamps for every line of output from CBMC, but this may not give the correct duration of a specific stage in CBMC as it may include more than just a call to the function that performs operations for that specific stage. So here we look at the duration of a phase in CBMC, for eg post processing, by tracking the monotonic clock time right before and right after the call to post_process(). Similarly, for other phases: symex, ssa conversion and solver.

regression/cbmc/runtime-profiling/test.desc
natasha-jeppu: Wasn't sure what to add here. It's not always the case that all stages of CBMC are executed for different tests so the output varies and only the lines specified here are common.

natasha-jeppu: I can provide a general test along the lines of ("Runtime Symex: .\*" | "Runtime Convert SSA: .\*" ...)

hannes-steffenhagen-diffblue: @natasha-jeppu but this only testing one file with one set of arguments? Surely we'd expect the output here to be somewhat consistent?

hannes-steffenhagen-diffblue (approved):
LGTM, not _too_ fussed on the testing considering this is just profiling code

regression/cbmc/runtime-profiling/test.desc
natasha-jeppu: The only output that was consistent across all tests was "Runtime Solver: .*". Is there a way to access the CI checks done for previous commits? It is evident there.

PR #5488, opened 2020-09-04
introduce dstringt::starts_with

kroening:

### DIFF
--- a/src/util/dstring.h
+++ b/src/util/dstring.h
@@ -90,6 +90,22 @@ public:
     return no==0; // string 0 is exactly the empty string
   }
 
+  /// equivalent of as_string().starts_with(s)
+  bool starts_with(const char *s)
+  {
+    for(const char *t = c_str(); *s != 0; s++, t++)
+      if(*t != *s)
+        return false;
+
+    return true;
+  }
+
+  /// equivalent of as_string().starts_with(s)
+  bool starts_with(const std::string &prefix)
+  {
+    return as_string().compare(0, prefix.size(), prefix) == 0;
+  }
+
   char operator[](size_t i) const
   {
     return as_string()[i];


NlightNFotis (commented):
There's a weird build failure on Travis, on osx, but aside from that, this is good stuff!

I very much like the readability of this compared to `has_prefix`. Does it mean that `has_prefix` is now deprecated? If yes, should it be marked so?

hannes-steffenhagen-diffblue:
Does that mean we no longer care about being able to swap in `std::string` for `dstringt`? No objection to that personally, but if we do that we should also remove `USE_STD_STRING` (that’d have to happen before this PR because we do actually test if cbmc builds with `USE_STD_STRING`).

tautschnig:
> Does that mean we no longer care about being able to swap in `std::string` for `dstringt`? No objection to that personally, but if we do that we should also remove `USE_STD_STRING` (that’d have to happen before this PR because we do actually test if cbmc builds with `USE_STD_STRING`).

I would believe we maintain full compatibility: with this patch, `.starts_with()` can be used the same way with `dstringt` and `std::string`.

@kroening I'd +1 the suggestion to deprecate `has_prefix`, and in a lot of files the `#include <util/prefix.h>` should be removed.

kroening:
We can either wait for C++20 (which adds `.starts_with()` to `std::string`) or remove `USE_STD_STRING`.

kroening:
I've dropped the second commit until we do C++20. The first commit is useful in preparation.

src/util/dstring.h
```diff
@@ -90,6 +90,22 @@ class dstringt final
     return no==0; // string 0 is exactly the empty string
   }
 
+  /// equivalent of as_string().starts_with(s)
+  bool starts_with(const char *s)
+  {
+    for(const char *t = c_str(); *s != 0; s++, t++)
+      if(*t != *s)
```
tautschnig, line 97: This is out-of-bounds when `this->size()` is less than `strlen(s)`.

tautschnig (changes requested):

src/util/dstring.h
kroening: `c_str` guarantees zero-termination.

tautschnig (approved):

PR #5500, opened 2020-09-21
Fix bug that prevented --malloc-may-fail from working

### LINKED ISSUE #5492: goto-cc incompatible with cbmc's --malloc-may-fail and --malloc-fail-null
CBMC version: 5.13.0
Operating system: Linux 64-bit

**Minimal example:**
```
#include <assert.h>
#include <stdlib.h>

int main (void) {
  int*p = malloc(sizeof(*p));
  assert(p);
  return 0;
}
```
`cbmc --malloc-may-fail --malloc-fail-null test.c` correctly detects a bug in this .c file.
When we run `goto-cc test.c -o test.goto`, and then run `cbmc --malloc-may-fail --malloc-fail-null test.goto`, cbmc incorrectly reports successful verification.

Currently, `goto-cc` does not accept these two malloc flags, so we cannot add them to the `goto-cc` invocation.

```
$ cbmc --malloc-may-fail --malloc-fail-null test.c
CBMC version 5.13.0 (cbmc-5.13.1) 64-bit x86_64 linux
Parsing test.c
Converting
Type-checking test
Generating GOTO Program
size of program expression: 89 steps
simple slicing removed 5 assignments
Generated 1 VCC(s), 1 remaining after simplification
Passing problem to propositional reduction
converting SSA
Running propositional reduction
Post-processing
Solving with MiniSAT 2.2.1 with simplifier
324 variables, 48 clauses
SAT checker: instance is SATISFIABLE
Runtime decision procedure: 0.0016344s

** Results:
<builtin-library-malloc> function malloc
[malloc.assertion.1] line 26 max allocation size exceeded: SUCCESS
[malloc.assertion.2] line 31 max allocation may fail: SUCCESS

test.c function main
[main.assertion.1] line 6 assertion p: FAILURE

** 1 of 3 failed (2 iterations)
VERIFICATION FAILED
```

**What happened instead:**
```
$ goto-cc test.c -o test.goto; cbmc --malloc-may-fail --malloc-fail-null test.goto
CBMC version 5.13.0 (cbmc-5.13.1) 64-bit x86_64 linux
Reading: test.goto
Generating GOTO Program
file <built-in-additions> line 21: warning: conflicting initializers for variable "__CPROVER_malloc_failure_mode"
using old value in module test file <built-in-additions> line 21
0
ignoring new value in module <built-in-library> file <built-in-additions> line 21
1
file <built-in-additions> line 25: warning: conflicting initializers for variable "__CPROVER_malloc_may_fail"
using old value in module test file <built-in-additions> line 25
FALSE
ignoring new value in module <built-in-library> file <built-in-additions> line 25
TRUE
size of program expression: 76 steps
simple slicing removed 0 assignments
Generated 1 VCC(s), 0 remaining after simplification

** Results:
<builtin-library-malloc> function malloc
[malloc.assertion.1] line 26 max allocation size exceeded: SUCCESS
[malloc.assertion.2] line 31 max allocation may fail: SUCCESS

test.c function main
[main.assertion.1] line 6 assertion p: SUCCESS

VERIFICATION SUCCESSFUL
```

hannes-steffenhagen-diffblue:
We have code in our stdlib models that depends on two global variables,
__CPROVER_malloc_may_fail and __CPROVER_malloc_failure_mode. Previously
these were set based on the ansi-c config in builtin_additions by
generating C code containing definitions for these variables.

The problem with that is that if you compile an executable using goto-cc
we also load the builtin_additions, so this executable already contains
definitions for these variables which can't be overridden.

This works around this issue by instead declaring these variables as
extern in builtin_additions, and setting their values dynamically as
last-minute instrumentation at the end of __CPROVER_initialize.  Giving
these no/nondet initial values is important because otherwise
preprocessing like constant folding on goto-binaries before cbmc
actually runs could remove branches we can, in fact, take with different
values set by --malloc-fail-* flags.

An alternative would be to instead dynamically change the code for the
models functions as instrumentation, but this has a couple of problems;
For instance this means we might change the behaviour of custom stdlib
models (which we allow via --no-library) or custom goto instrumentation
in unexpected ways. I believe this way is the least likely to cause any
friction like that.


src/ansi-c/ansi_c_internal_additions.cpp
```diff
@@ -186,17 +186,15 @@ void ansi_c_internal_additions(std::string &code)
       CPROVER_PREFIX "size_t size, " CPROVER_PREFIX "bool zero);\n"
     "const void *" CPROVER_PREFIX "alloca_object = 0;\n"
 
-    "int " CPROVER_PREFIX "malloc_failure_mode="+
-      std::to_string(config.ansi_c.malloc_failure_mode)+";\n"
+    "extern int " CPROVER_PREFIX "malloc_failure_mode;\n"
```
peterschrammel, line 189: Can the malloc-X options be removed from goto-cc now?

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -778,6 +779,47 @@ bool cbmc_parse_optionst::set_properties()
...
+  T &&value)
+{
+  const auto &symbol = symbol_table.lookup_ref(symbol_name).symbol_expr();
+  return code_assignt{symbol, from_integer(value, symbol.type())};
+}
+
+/// Some variables have different initial values based on what flags are being passed to cbmc;
+/// since the _start function might've been generated by goto-cc before getting passed over here
+/// we need to add the initialisation of these variables here as a post-processing step
+static void
```
peterschrammel: I'm not sure this code should be in this class - better move it into a separate file in goto-programs or goto-instrument.

src/ansi-c/ansi_c_internal_additions.cpp
hannes-steffenhagen-diffblue: @peterschrammel we never added them to `goto-cc`. What we did add to `goto-cc` was `--object-bits`, which could be removed with a similar approach (or by doing the check the object bits are used for in a different way entirely the way @danpoe was suggesting).

src/cbmc/cbmc_parse_options.cpp
hannes-steffenhagen-diffblue: Done

peterschrammel (approved):

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -879,6 +880,8 @@ bool cbmc_parse_optionst::process_goto_program(
   link_to_library(
     goto_model, log.get_message_handler(), cprover_c_library_factory);
 
+  add_malloc_may_fail_variable_initializations(goto_model);
```
martin-cs, line 883: If this is now part of the correct set up for goto programs can you either A. add this to all such set-up code ( goto-analyzer, goto-diff, goto-instrument at least) or B. add it to one of the existing methods -- I would not oppose it being part of `link_to_library` as it is part of the handling of the library.

martin-cs (changes requested):

src/cbmc/cbmc_parse_options.cpp
hannes-steffenhagen-diffblue: It definitely can’t go into `link_to_library` and shouldn’t go into any of the instrumentation tools (i.e. it can’t go into `link_to_library` _because_ it shouldn’t go into the instrumentation tools). The reason this is put in this location is specifically to _avoid_ instrumentation code setting these variables, they should _only_ be set by cbmc. Arguably jbmc and goto-analyzer too, but _definitely_ not goto-instrument etc.

The thing with `goto-analyzer` though is also that because the stdlib models are cbmc models, I don't think they'd even make sense for goto-analyzer (and I don't think they're loaded). Similar story with jbmc, I don't think it loads the C models (that would make more sense like to verify C code interacting with java code I guess, but I don't know if anyone's actually doing that).

hannes-steffenhagen-diffblue: To summarise above, I don’t really agree from what I understand of the situation, but maybe I’m missing something.

martin-cs: They shouldn't just be `cbmc` models; `goto-analyzer` will need them as well.  They may not work that well at the moment but let's not make them any more broken than they currently are.  In `jbmc` I don't know what their models are written in, Java or C.  It would be good if it was possible for it to work.

My concern is avoiding divergence in how the tools set up and configure goto programs.  This has been a source of subtle bugs in the past.

hannes-steffenhagen-diffblue: @martin-cs Again, I _could_ put this code in goto-analyzer but I still don’t quite understand what good it’ll do. The models library as a matter of fact is currently not being used for goto-analyzer (and as far as I know this has been the case for _at least_ 5 years), so if I add this code it’ll naturally go untested. So the problem is that then I’m adding code to goto-analyzer that is vestigial to begin with (i.e. literally a no-op in all circumstances). While I do understand that we want to keep goto transforms consistent between different tools in _this_ particular case I don’t see the value in it. Right now goto-analyzer  doesn’t even understand things like `__CPROVER_allocate` etc (or rather, I don’t think any of the domains do). For me the problem here is I’m adding code to places that I _know_ does nothing right now, and that I can’t be sure will do the right thing whenever someone changes the surrounding context in such a way that would allow it to do something. Whenever someone in the future would read that goto-analyzer code and asks "what is this bit of the code for", the true answer would be "it does absolutely nothing because the preconditions for it doing anything are not there". If it ever starts doing anything we’ll have no way of knowing it’s doing the right thing because we didn’t (and couldn’t) test it until then.

I think we’re both concerned about maintainability here, but we seem to disagree on which way round is better for maintainability.

@peterschrammel @piotr-grabalski @NlightNFotis @thomasspriggs would you mind chipping in? Very much willing to be overruled here, but I want to be sure we’re doing the right thing here.

hannes-steffenhagen-diffblue: @martin-cs Would you be OK with it if I made a different PR adding these changes to goto-analyzer and we can have the discussion on whether or not it _should_ go there on that PR? Getting this fix in for cbmc at least is somewhat time critical for us.

peterschrammel: Yes, I think it should be called by goto-analyzer too.
IMO, this shouldn't be implemented through the models library at all, but by a pass over the goto model - but that may follow afterwards. This PR is for urgently fixing a major bug in CBMC when used with goto-cc. I.e. priority is getting the bug fixed today. Tomorrow we can look how to improve the solution.

hannes-steffenhagen-diffblue: The main reason I’d want it to be in the models library is because IMHO whether or not this can return NULL is part of the model, and you should be able to figure out what behaviour the models library has from looking at the library code. Instrumenting logic means distributing the logic between the models library source files and different parts of the C++ code base, but like you said we should discuss how to move forward with this when we’ve actually fixed the problem.

martin-cs: @hannes-steffenhagen-diffblue would you be OK adding a comments (and commented out code?) in the other locations saying why it shouldn't / isn't / may be needed?  Would be happy to approve that and get it merged in an hour or so when the CI passes?

@peterschrammel I feel the library solution has a number of advantages; should we talk about this at some point?

src/goto-programs/add_malloc_may_fail_variable_initializations.cpp
```diff
@@ -0,0 +1,62 @@
...
+      initialize_function_name) == 1);
+  auto &initialize_function =
+    goto_model.goto_functions.function_map.find(initialize_function_name)
+      ->second;
+  const auto initialize_function_end =
+    --initialize_function.body.instructions.end();
+
+  initialize_function.body.insert_before(
+    initialize_function_end,
+    goto_programt::make_assignment(make_integer_assignment(
```
thomasspriggs, line 51: :pick: I think I would have replaced the existing assignment in cprover_initialise, rather than adding extra assignments. Your approach will make the resulting code in `cprover_initialise` more difficult to read, when it is printed for debugging purposes. I think that replacing the existing assignment would also have reduced the need to modify the regression tests.

src/cbmc/cbmc_parse_options.cpp
thomasspriggs: :thinking: There is kind of a design decision to be made, as to whether these checks should be part of the model or to be an instrumentation pass. I liked the idea of this being an instrumentation pass.

:pick: If we are sticking with them being part of the implementation of the model, then I think the approach outlined in this PR isn't good in the long term either. The reason being that this is separating correctly configuring the model/library from adding it. This leaves things in an invalid state in between the two stages, where it is added but not correctly configured. However I am willing to approve as is because this is the simplest way to keep the existing user interface and fix the issue in a short period of time.

src/goto-programs/add_malloc_may_fail_variable_initializations.cpp
hannes-steffenhagen-diffblue: @thomasspriggs I don’t understand how having an extra assignment is more difficult to read/debug than adding extra logic to functions at runtime

martin-cs (approved):
Thanks.

If possible, a comment in `goto-instrument` saying "Don't do this" and something for `goto-diff` ( @peterschrammel : what is `goto-diff`'s status) would be great but those could be in another PR if you need to get this moving ASAP.

src/cbmc/cbmc_parse_options.cpp
martin-cs: @thomasspriggs agreed -- we do need to discuss this.  Adding the models and configuring them are two different things and yes, we do need to think about it.  Do you want to open an issue to discuss this?

jbmc/src/jbmc/jbmc_parse_options.cpp
```diff
@@ -726,6 +726,10 @@ int jbmc_parse_optionst::get_goto_program(
 
     goto_modelt &goto_model = dynamic_cast<goto_modelt &>(*goto_model_ptr);
 
+    // if we added the ansi-c library models here this should also call
+    // add_malloc_may_fail_variable_initializations(goto_model);
+    // here
```
thomasspriggs, line 731: :no_entry_sign: I am not sure this comment is helpful. JBMC uses its own models rather than the ansi-c models. Also the java language uses `new`, not malloc. This is parsed from the class file into `ID_java_new` / `ID_java_new_array`, which is then lowered directly into `ID_allocate`. This bypasses the use of a malloc function entirely. See `jbmc/src/java_bytecode/remove_java_new.cpp` for details.

hannes-steffenhagen-diffblue: @thomasspriggs there are valid reasons why you might want to add the ansi-c library here; e.g. if there was JNI support and you wanted to link your java models to C models. Not a thing as of now as far as I know, but it's not inconceivable that his would be a concern in the future.

thomasspriggs: Ok, that make a little more sense, if you were thinking about JNI support. AFAIK JNI support isn't planned for the near future. I also suspect that it would be tricky to implement given the complexities of lazy loading.

hannes-steffenhagen-diffblue: Yeah, that’s why I didn’t want to put it here in the first place! But if you wanted to load the C library, you’d have to add this code and the flags as well, which I believe was @martin-cs's concern and why he asked for these comments.

martin-cs: @thomasspriggs : Yes, as @hannes-steffenhagen-diffblue observed JNI is one thing.  Another is that there are a lot of implicit assumptions in the modelling and instrumentation that the C library is present.  It is likely that future modelling of more sophisticated things may well make use of / depend on the C library and I would hate to see the intermediate representation diverge because of this.

@hannes-steffenhagen-diffblue : yes, my concern is in divergence of how we handle things and the "well it works if you use X but not if you use Y first, then X".  This way the divergence is clear and documented in the relevant place (ideally it would be in `goto-instrument` and `goto-diff` too).

chrisr-diffblue (approved):
Adding my approval to the others. Though I agree it looks like some broader/longer term discussion is also needed.

PR #5504, opened 2020-09-24
Build fail clang 12

### LINKED ISSUE #5501: Build fails with Apple Clang 12 (clang-1200.0.32.2) due to use of reference types
Clang build fails due to the -WError flag, resulting in -Wrange-loop-analysis warnings being shown as errors. 
The issue arises in multiple files due to the use of `auto` instead of `auto&` since the compiler expects a non reference type for places where one iterates over a loop in such a way that creates a proxy/copy of the elements instead of a reference to the elements themselves.
e.g.

`
src/util/graph.h:635:20: error: loop variable 'succ' of type 'const std::__1::pair<const unsigned long, empty_edget>' creates a
      copy from type 'const std::__1::pair<const unsigned long, empty_edget>' [-Werror,-Wrange-loop-analysis]
    for(const auto succ : succs)`

CBMC version: 5.14.1 (cbmc-5.14.1-dirty)
Operating system: MacOS Catalina 10.15.6
Exact command line resulting in the issue: cmake --build build
What behaviour did you expect: building without errors
What happened instead: build fails

mina1604:
fix issue #5501

### DIFF
--- a/src/goto-programs/goto_program.cpp
+++ b/src/goto-programs/goto_program.cpp
@@ -1093,7 +1093,7 @@ bool goto_programt::equals(const goto_programt &other) const
 
     // the number of targets is the same as instructiont::equals returned "true"
     auto other_target_it = other_it->targets.begin();
-    for(const auto t : ins.targets)
+    for(const auto &t : ins.targets)
     {
       if(
         t->location_number - ins.location_number !=
--- a/src/goto-symex/build_goto_trace.cpp
+++ b/src/goto-symex/build_goto_trace.cpp
@@ -329,7 +329,7 @@ void build_goto_trace(
 
   for(const auto &time_and_ssa_steps : time_map)
   {
-    for(const auto ssa_step_it : time_and_ssa_steps.second)
+    for(const auto &ssa_step_it : time_and_ssa_steps.second)
     {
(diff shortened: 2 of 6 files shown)


martin-cs (approved):
Looks like the kind of patch we should have regardless of the clang issue, even better if it fixes the thing.

NlightNFotis (approved):

NlightNFotis:
Thanks for your contribution.

Can I ask that the last commit is removed, and the `clang-format` changes be squashed into the previous commit? LGTM otherwise.

martin-cs:
Yes; good catch @NlightNFotis : @mina1604 we normally work by rebasing branches and then merging into `develop` rather than merging `develop` into the branches.

hannes-steffenhagen-diffblue (approved):
LGTM modulo previous comments (i.e. rebase instead of merge, would be better if clang-format changes were done in the same commit as the code changes).

mina1604:
Should be one commit now.

NlightNFotis:
Thanks for the prompt change.

We'll take it from here - we'll chase reviews internally and merge it once it has an adequate number of reviews.

peterschrammel (approved):

PR #5506, opened 2020-09-25
Fix JBMC crash on windows in `NondetEnumOpaqueReturn` test

thomasspriggs:
String literals with multibyte utf-8 characters in java cause `ch` to be
negative. Passing a negative value into `std::isalnum` results in
undefined behaviour. On windows the observed behaviour is a popup error
message about a failed assertion in the run time library. This can stall
the CI process as none of the buttons on the popup message ever get
pressed. Escaping these characters is sufficient to fix the stalling
test.

- [x] N/A - Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [x] N/A - The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] N/A - My commit message includes data points confirming performance improvements (if claimed).


src/util/string_utils.cpp
```diff
@@ -158,7 +158,7 @@ std::string escape_non_alnum(const std::string &to_escape)
   {
     if(ch == '_')
       escaped << "__";
-    else if(isalnum(ch))
+    else if(ch >= 0 && isalnum(ch))
```
martin-cs: I'm a little worried about this because certainly in C whether `char` is signed or not is implementation dependent.

martin-cs (approved):
If the test passes on all platforms then I'm happy.

martin-cs (approved):
This seems like a better way of doing things.  If possible it might be good to squash some of the commits together.

src/util/string_utils.cpp
```diff
@@ -156,13 +156,16 @@ std::string escape_non_alnum(const std::string &to_escape)
   std::ostringstream escaped;
   for(auto &ch : to_escape)
   {
+    const auto uch = static_cast<unsigned char>(ch);
```
hannes-steffenhagen-diffblue: Actually had to look this up, the standard actually guarantees this is well defined and behaves like a sane person would expect! A surprise to be sure, but a welcome one.

unit/util/string_utils/escape_non_alnum.cpp
```diff
@@ -0,0 +1,278 @@
...
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xF6)}) == "_f6");
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xF7)}) == "_f7");
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xF8)}) == "_f8");
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xF9)}) == "_f9");
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xFA)}) == "_fa");
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xFB)}) == "_fb");
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xFC)}) == "_fc");
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xFD)}) == "_fd");
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xFE)}) == "_fe");
+  CHECK(escape_non_alnum(std::string{static_cast<signed char>(0xFF)}) == "_ff");
```
hannes-steffenhagen-diffblue, line 277: I hope this is generated instead of handwritten!

hannes-steffenhagen-diffblue (approved):

src/util/string_utils.cpp
```diff
@@ -156,13 +156,16 @@ std::string escape_non_alnum(const std::string &to_escape)
...
     if(ch == '_')
       escaped << "__";
-    else if(isalnum(ch))
+    else if(isalnum(uch))
       escaped << ch;
     else
+    {
       escaped << '_' << std::hex << std::setfill('0') << std::setw(2)
-              << (unsigned int)ch;
+              << (unsigned int)uch;
```
hannes-steffenhagen-diffblue: Since you’re changing this line anyway could you turn this into `static_cast` please?

thomasspriggs: Now done as a `narrow` style cast instead as that is the preferred way to cast numerics according to the C++ core guidelines.

thomasspriggs:
> This seems like a better way of doing things. If possible it might be good to squash some of the commits together.

Now squashed. This was still in draft status as I was still tidying things up.

src/util/string_utils.cpp
thomasspriggs: Now done through a `narrow_cast` cast, in order to avoid explicitly checking for negative values.

src/util/string_utils.cpp
hannes-steffenhagen-diffblue: @thomasspriggs I don't think this is a good in this instance as this is a widening cast, not a narrowing cast, so it's a bit confusing. Not a blocker, just an odd thing.

thomasspriggs: Ok. I have swapped the conversion to `unsigned char` back to `static_cast`, because reinterpretation is the desired behaviour here. I have swapped the conversion from `unsigned char` to `int`, such that it is done using `{}` braced initialisation because this form of initialisation supports widening, but not narrowing. Therefore it makes it explicit that we are not performing narrowing.

peterschrammel (approved):

src/util/string_utils.cpp
```diff
@@ -156,13 +156,13 @@ std::string escape_non_alnum(const std::string &to_escape)
   std::ostringstream escaped;
   for(auto &ch : to_escape)
   {
+    const int uch{static_cast<unsigned char>(ch)};
```
hannes-steffenhagen-diffblue, line 166: Nit: This is now complicated enough to warrant an explanation I'd think. Would be great if you could drop a line here of why this is done.

thomasspriggs: Done.

NlightNFotis (approved):
LGTM

PR #5508, opened 2020-09-29
xen-cbmc: fix ci script

nmanthey:
This script has been broken, and not been picked up in CI until now.

This change fixes the script, as well as makes sure that the relevant
steps are executed, and the resulting build artefacts have the required
goto-cc section in them.

This CI script has been refered to as part of the paper:
"Using model checking tools to triage the severity of security bugs in
the Xen hypervisor", that has been presented at FMCAD 2020.

Signed-off-by: Norbert Manthey <nmanthey@amazon.de>


kroening (approved):

NlightNFotis (changes requested):
Thanks for the fix! I would say it's much needed 👍 

May I ask why the file was renamed from `.yml` to `.yaml`? This has the unfortunate side effect of breaking file revision history, which makes it hard to detect what the actual change here is (the one that actually fixed the broken CI issue).

.github/workflows/build-and-test-Xen.yaml
```diff
@@ -0,0 +1,45 @@
...
+        run: git clone git://xenbits.xen.org/xen.git xen_4_13 && cd xen_4_13 && git reset --hard RELEASE-4.13.0 && pwd
+
+      - name: Prepare compile Xen with CBMC via one-line-scan
+        run: |
+          ln -s goto-cc src/goto-cc/goto-ld
+          ln -s goto-cc src/goto-cc/goto-as
+          ln -s goto-cc src/goto-cc/goto-g++
+
+      - name: Compile Xen with CBMC via one-line-scan, and check for goto-cc section
+        run: one-line-scan/one-line-scan --add-to-path $(pwd)/src/cbmc --add-to-path $(pwd)/src/goto-diff --add-to-path $(pwd)/src/goto-cc --no-analysis --trunc-existing --extra-cflags -Wno-error -o CPROVER -j 3 -- make -C xen_4_13 xen -j $(nproc) -k || true
```
hannes-steffenhagen-diffblue: Could you split this over multiple lines? This is a bit hard to read.

martin-cs (approved):
Does this mean I will stop getting e-mails saying "Run failed: Build and Test Xen" whenever I push to a PR?

.github/workflows/build-and-test-Xen.yaml
nmanthey: Yes, will split.

nmanthey:
> Thanks for the fix! I would say it's much needed +1
> 
> May I ask why the file was renamed from `.yml` to `.yaml`? This has the unfortunate side effect of breaking file revision history, which makes it hard to detect what the actual change here is (the one that actually fixed the broken CI issue).

I will split the commit into (1) file rename, and (2) actual diff. The rename is basically for consistency.

Yes, I can split that over multiple lines.

Yes, this fix should stop those PR emails (I hope, I did not push PRs recently ;) )

NlightNFotis (approved):
Thanks for the explanation. 

Good to go from me 👍

PR #5509, opened 2020-10-01
Add additional documentation of --nan-check flag

piotr-grabalski:
Additionally added a regression test for --nan-check flag because it
seemed like there is no test for this functionality.

### DIFF
--- a/doc/cprover-manual/properties.md
+++ b/doc/cprover-manual/properties.md
@@ -147,6 +147,21 @@ unsigned foo(unsigned x)
   x = x + 2; // unsigned overflow checks are generated here
 ```
 
+#### Flag --nan-check limitations
+
+Please note that `--nan-check` flag is adding not-a-number checks only for
+generation of NaN value. Current implementation of `--nan-check` flag is not
+providing checks for propagation of NaN values. Generating assertions on type
+casting or structure/union member access is unsupported and such operation
+will not be examined.
+
+For example:
+
+```
+float f = 0.0/0.0; // will generate NaN - CBMC will add assertion
+float g = NAN+0.0; // propagation of NaN value - no assertion generated
+```
+
 #### Generating function bodies
 
 Sometimes implementations for called functions are not available in the goto
--- a/regression/cbmc-library/float-nan-check/main.c
+++ b/regression/cbmc-library/float-nan-check/main.c
@@ -1,0 +1,76 @@
+#include <assert.h>
+#include <math.h>
+#include <stdint.h>
+
+// This test contains a set of operations performed on a floating point numbers
+// to verify --nan-check flag. The result of test may change when/if support
+// of NaN value propagation checking is supported. Current behaviour of
+// --nan-check flag is that it checks whether an arithmetic operation performed
+// on a floating point value(s) is generating new not-a-number value. If NaN is
+// generated because one of the values is already NaN the assertion is not
+// produced.
+
+union container {
+  uint32_t u;
+  float f;
+};
+
+int main(void)
+{
+  // set special values
+  float mynan;
+  __CPROVER_assume(isnan(mynan));
+  float myinf;
+  __CPROVER_assume(isinf(myinf) && myinf > 0.0);
+  float myzero;
+  __CPROVER_assume(myzero == 0.0f);
+
+  // variables
+  union container c = {UINT32_MAX};
+  int n;
+
+  // should generate assertion if --nan-check flag will support
+  // NaN propagation checking
+  float g = (float)(c.u);
+  float f = c.f;
+  f = c.f + myzero;
+
+  // these operations should generate assertions
+  // 0.0 / 0.0 = NaN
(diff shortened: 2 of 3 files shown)


doc/cprover-manual/properties.md
```diff
@@ -147,6 +147,19 @@ unsigned foo(unsigned x)
   x = x + 2; // unsigned overflow checks are generated here
 ```
 
+#### Flag --nan-check limitations
+
+Please note that `--nan-check` flag is adding not-a-number checks only for
+generation of NaN value. Current implementation of `--nan-check` flag is not
+providing checks for propagation of NaN values.
```
peterschrammel: Clarify casts and unions

peterschrammel (approved):

regression/cbmc-library/float-nan-check/test.desc
```diff
@@ -0,0 +1,24 @@
...
+\[main.NaN.9\] line \d+ NaN on / in \(double\)byte_extract_little_endian\(c, 0l, float\) / 0.0: SUCCESS
+\[main.NaN.10\] line \d+ NaN on \* in return_value___builtin_nanf \* return_value___builtin_nanf\$0: SUCCESS
+\[main.NaN.11\] line \d+ NaN on \+ in return_value___builtin_nanf\$1 \+ return_value___builtin_nanf\$2: SUCCESS
+\[main.NaN.12\] line \d+ NaN on - in return_value___builtin_nanf\$3 - return_value___builtin_nanf\$4: SUCCESS
+\[main.NaN.13\] line \d+ NaN on / in return_value___builtin_nanf\$5 / return_value___builtin_nanf\$6: SUCCESS
+\[main.NaN.14\] line \d+ NaN on \+ in return_value___builtin_nanf\$7 \+ \(float\)n: SUCCESS
+\[main.NaN.15\] line \d+ NaN on - in return_value___builtin_nanf\$8 - \(float\)n: SUCCESS
+\[main.NaN.16\] line \d+ NaN on \* in return_value___builtin_nanf\$9 \* \(float\)n: SUCCESS
+\[main.NaN.17\] line \d+ NaN on / in return_value___builtin_nanf\$10 / \(float\)n: SUCCESS
+^EXIT=10$
```
martin-cs, line 21: Can you add a negative test for `main.NaN.18` so that adding additional checks will break this test.

regression/cbmc-library/float-nan-check/main.c
```diff
@@ -0,0 +1,66 @@
...
+
+union container {
+  uint32_t u;
+  float f;
+};
+
+int main(void)
+{
+  union container c = {UINT32_MAX};
+  int n;
```
martin-cs, line 30: `n` `int` or `float`?

regression/cbmc-library/float-nan-check/main.c
```diff
@@ -0,0 +1,66 @@
...
+
+  // should generate assertion if --nan-check flag will support
+  // NaN propagation checking
+  float f = c.f;
+  f = c.f + 0.0;
+
```
martin-cs, line 37: Rather than use constants (and so test just the constant propagation code), how about:

```
float my_NaN;
__CPROVER_assume(isnan(my_NaN));
float my_inf;
__CPROVER_assume(isinf(my_inf) && my_inf > 0.0);
float my_zero;
__CPROVER_assume(my_zero == 0.0f);
```

The results should be the same (well `my_zero` might be negative but I don't think it changes anything) but because they are non-deterministic, the whole solver will be run.

martin-cs (approved):
Basically, yes.

doc/cprover-manual/properties.md
piotr-grabalski: OK, I will add the following sentence:

> Generating assertions on type casting or structure/union member access is unsupported and such operation will not be examined.

so that the full text will look like:

> Please note that `--nan-check` flag is adding not-a-number checks only for generation of NaN value. Current implementation of `--nan-check` flag is not providing checks for propagation of NaN values. Generating assertions on type casting or structure/union member access is unsupported and such operation will not be examined.

regression/cbmc-library/float-nan-check/test.desc
piotr-grabalski: OK

regression/cbmc-library/float-nan-check/main.c
piotr-grabalski: I don't think it really matters in my test, my thinking was that `n` as `int` should trigger implicit conversion to `float` on mixed type arithmetic, like `f = NAN + n;` but when I look at this from one day perspective it does not seem to prove anything.

regression/cbmc-library/float-nan-check/main.c
piotr-grabalski: Thanks for suggestions, I will have a play with non-deterministic approach.

PR #5512, opened 2020-10-01
Updates posix_memalign to consider malloc may fail 

feliperodri:

### DIFF
--- a/regression/cbmc-library/posix_memalign-01/test.desc
+++ b/regression/cbmc-library/posix_memalign-01/test.desc
@@ -1,8 +1,8 @@
 CORE
 main.c
---pointer-check --bounds-check
+--pointer-check --bounds-check --malloc-may-fail --malloc-fail-null
 VERIFICATION SUCCESSFUL
 ^EXIT=0$
 ^SIGNAL=0$
 --
-^\*\*\*\* WARNING: no body for function posix_memalign
+^warning: ignoring
--- a/regression/cbmc-library/posix_memalign-02/main.c
+++ b/regression/cbmc-library/posix_memalign-02/main.c
@@ -1,0 +1,13 @@
+#include <stdlib.h>
+#include <string.h>
+
+int main()
+{
+  size_t size = 4;
+  size_t page_size = 4096;
+  void *src = "testing";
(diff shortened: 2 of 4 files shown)


martin-cs (approved):
Looks good; thanks.

regression/cbmc-library/posix_memalign-01/test.desc
```diff
@@ -1,8 +1,8 @@
 CORE
 main.c
---pointer-check --bounds-check
+--pointer-check --bounds-check --malloc-may-fail --malloc-fail-null
```
hannes-steffenhagen-diffblue, line 3: I don't completely understand the changes for this test. You are changing the flags but not the expected output?

hannes-steffenhagen-diffblue (approved):
Looks OK to me, just don't quite understand the changes to one of these tests.

regression/cbmc-library/posix_memalign-01/test.desc
feliperodri: I added the flags, so now the test considers malloc can fail. However, this doesn't change the expected output of this test, since it's resilient to a failed malloc. @hannes-steffenhagen-diffblue

feliperodri:
Could I get one more review @tautschnig @kroening @chrisr-diffblue?

regression/cbmc-library/posix_memalign-01/test.desc
hannes-steffenhagen-diffblue: @feliperodri I'd recommend to reverse the changes here. Tests should be self contained and not contain flags that aren’t relevant for the test. This test is unrelated to `--malloc-may-fail` and hence shouldn’t use the flag.

chrisr-diffblue (approved):
Looks sensible, thanks!

regression/cbmc-library/posix_memalign-01/test.desc
feliperodri: @hannes-steffenhagen-diffblue I'll send a new PR that update these flags. I'll keep the failed regression test with the malloc flags and I'll remove them from the this test case.

PR #5515, opened 2020-10-05
Create .deb package for ubuntu-18.04 on release

hannes-steffenhagen-diffblue:
Also makes sure to include the name of the distribution in the .deb
packages to make it easier to see which is which.

### DIFF
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -4,7 +4,7 @@ on:
 
 name: Upload additional release assets
 jobs:
-  ubuntu-package:
+  ubuntu-20_04-package:
     runs-on: ubuntu-20.04
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


NlightNFotis (approved):
Looks good to me.

.github/workflows/release-packages.yaml
```diff
@@ -44,6 +44,47 @@ jobs:
...
+        uses: bruceadams/get-release@v1.2.0
+      - name: Upload binary packages
+        uses: actions/upload-release-asset@v1
+        with:
+          upload_url: ${{ steps.get_release_info.outputs.upload_url }}
+          asset_path: ${{ steps.create_packages.outputs.deb_package }}
+          asset_name: ${{ steps.create_packages.outputs.deb_package_name }}
+          asset_content_type: application/x-deb
+
+
```
chrisr-diffblue, line 87: Seems a bit of a shame that so much of this is a duplication of the the ubuntu-20_04 target - is there any way to re-use this (I don't know github actions, sorry! - but was thinking about either some way of 'including' comment parts, or moving those out into a script/cmake action?)

chrisr-diffblue (approved):
Looks good to me, and happy for it to go in as is - just a minor query about duplication.

.github/workflows/release-packages.yaml
hannes-steffenhagen-diffblue: @chrisr-diffblue There is, technically. Although to be honest I've been deliberately avoiding it, I find these short _enough_ and I find logic gets harder to follow as you split it over multiple scripts. This isn't enough script for this to really save us any work, so the main advantage would be it's easier to see that these are doing mostly the same thing – at the cost of hiding what that thing is behind possibly multiple files.

PR #5516, opened 2020-10-05
Change path for executables in package builder schema.

NlightNFotis:
Previously, we were building using the default config.
This has now been changed to be explicitly a release
configuration, so the binaries are now being built
into build/bin/Release/<binary.exe>


NlightNFotis:
During discussion in the standup today (6-10-2020), it emerged that this is not useful, given that the broken script `packages.yaml` is bound to be removed in the near future.

PR #5521, opened 2020-10-09
Add concurrency examples to regression tests

thomasspriggs:
This PR adds concurrency examples to the regression tests in preparation for the planned sequentialization work.

### DIFF
--- a/regression/cbmc-sequentialization/CMakeLists.txt
+++ b/regression/cbmc-sequentialization/CMakeLists.txt
@@ -1,0 +1,10 @@
+if(NOT WIN32)
+  add_test_pl_tests(
+          "$<TARGET_FILE:cbmc> --validate-goto-model --validate-ssa-equation"
+  )
+else()
+  add_test_pl_tests(
+          "$<TARGET_FILE:cbmc> --validate-goto-model --validate-ssa-equation"
+          -X requires_posix_only_headers
+  )
+endif()
--- a/regression/cbmc-sequentialization/posix_semaphores/blocking_queue.c
+++ b/regression/cbmc-sequentialization/posix_semaphores/blocking_queue.c
@@ -1,0 +1,153 @@
+#include <assert.h>
+#include <pthread.h>
+#include <semaphore.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+pthread_t start_worker_thread(void *(*start_routine)(void *))
+{
+  pthread_t worker_thread;
+  const pthread_attr_t *const attributes = NULL;
+  void *const worker_argument = NULL;
+  const int create_status =
+    pthread_create(&worker_thread, attributes, start_routine, worker_argument);
+  assert(create_status == 0);
+  return worker_thread;
+}
+
+void join_thread(const pthread_t thread)
+{
+  const int join_status = pthread_join(thread, NULL);
+  assert(join_status == 0);
+}
+
+sem_t create_semaphore(int initial_value)
+{
+  const int shared_between_processes = false;
+  sem_t semaphore;
+  const int init_error =
+    sem_init(&semaphore, shared_between_processes, initial_value);
+  assert(init_error == false);
+  return semaphore;
+}
+
+void destroy_semaphore(sem_t *const semaphore)
+{
+  int destroy_error = sem_destroy(semaphore);
+  assert(destroy_error == false);
+}
+
+// This blocking queue structure supports waiting for free space before
+// enqueuing and waiting for data before dequeuing, using a pair of semaphores.
+// These functions do not employ a mutex, so a data race may occur if multiple
+// threads attempt to enqueue at the same time or multiple threads attempt to
+// dequeue at the same time. The expected valid use case is with two threads,
+// where one of the two enqueues and the other thread dequeues.
+struct blocking_queuet
+{
+  size_t size;
+  int *elements;
+  size_t begin;
+  size_t end;
+  sem_t free_space;
+  sem_t data_waiting;
+};
+
+struct blocking_queuet initialise_blocking_queue(const size_t size)
+{
+  printf("init queue begin\n");
+  struct blocking_queuet queue;
+  queue.size = size;
+  queue.elements = calloc(size, sizeof(int));
+  queue.begin = 0;
+  queue.end = 0;
+  queue.free_space = create_semaphore(size);
+  queue.data_waiting = create_semaphore(0);
+  printf("init queue end\n");
+  return queue;
+}
+
+void free_blocking_queue(struct blocking_queuet *queue)
+{
+  free(queue->elements);
+  queue->elements = NULL;
+  destroy_semaphore(&queue->free_space);
+  destroy_semaphore(&queue->data_waiting);
+}
+
+void enqueue(struct blocking_queuet *queue, const int data)
+{
+  printf(
+    "enqueue begin:%lu end:%lu data:%d \n", queue->begin, queue->end, data);
+  const int free_wait_error = sem_wait(&queue->free_space);
+  assert(free_wait_error == false);
+  queue->elements[queue->end] = data;
+  if(++queue->end == queue->size)
+  {
+    queue->end = 0;
+  }
+  const int post_error = sem_post(&queue->data_waiting);
+  assert(post_error == false);
+  printf("enqueue done\n");
+}
+
+int dequeue(struct blocking_queuet *queue)
+{
+  printf("dequeue begin:%lu end:%lu\n", queue->begin, queue->end);
+  const int data_wait_error = sem_wait(&queue->data_waiting);
+  assert(data_wait_error == false);
+  int result = queue->elements[queue->begin];
+  if(++queue->begin == queue->size)
+  {
+    queue->begin = 0;
+  }
+  const int free_error = sem_post(&queue->free_space);
+  assert(free_error == false);
+  printf("dequeue done. data:%d \n", result);
+  return result;
+}
+
+struct blocking_queuet input_queue;
+struct blocking_queuet output_queue;
+
+void *worker(void *arguments)
+{
+  int data;
+  while(data = dequeue(&input_queue))
+  {
+    enqueue(&output_queue, data * data);
+  }
+  pthread_exit(NULL);
+}
+
+int main(void)
+{
+  input_queue = initialise_blocking_queue(3);
+  output_queue = initialise_blocking_queue(10);
+  const pthread_t worker_thread1 = start_worker_thread(&worker);
+  printf("worker_started\n");
+  enqueue(&input_queue, 1);
+  enqueue(&input_queue, 2);
+  enqueue(&input_queue, 3);
+  enqueue(&input_queue, 4);
+  enqueue(&input_queue, 5);
+  enqueue(&input_queue, 6);
+  enqueue(&input_queue, 7);
+  enqueue(&input_queue, 8);
+  enqueue(&input_queue, 9);
+  enqueue(&input_queue, 0);
+  join_thread(worker_thread1);
+  free_blocking_queue(&input_queue);
+  assert(dequeue(&output_queue) == 1);
+  assert(dequeue(&output_queue) == 4);
+  assert(dequeue(&output_queue) == 9);
+  assert(dequeue(&output_queue) == 16);
+  assert(dequeue(&output_queue) == 25);
+  assert(dequeue(&output_queue) == 36);
+  assert(dequeue(&output_queue) == 49);
+  assert(dequeue(&output_queue) == 64);
+  assert(dequeue(&output_queue) == 81);
+  free_blocking_queue(&output_queue);
+  return EXIT_SUCCESS;
+}
--- a/regression/cbmc-sequentialization/posix_semaphores/blocking_wait.desc
+++ b/regression/cbmc-sequentialization/posix_semaphores/blocking_wait.desc
@@ -1,0 +1,12 @@
+FUTURE requires_posix_only_headers
+blocking_wait.c
+
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^warning: ignoring
+pointer handling for concurrency is unsound
+--
+Test that sem_wait will block when the value of a semaphore is 0, until it is
+increased via sem_post.
--- a/regression/cbmc-sequentialization/posix_semaphores/create_destroy.c
+++ b/regression/cbmc-sequentialization/posix_semaphores/create_destroy.c
@@ -1,0 +1,30 @@
+#include <assert.h>
+#include <pthread.h>
+#include <semaphore.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+sem_t create_semaphore(int initial_value)
+{
+  const int shared_between_processes = false;
+  sem_t semaphore;
+  const int init_error =
+    sem_init(&semaphore, shared_between_processes, initial_value);
+  assert(init_error == false);
+  return semaphore;
+}
+
+void destroy_semaphore(sem_t *const semaphore)
+{
+  int destroy_error = sem_destroy(semaphore);
+  assert(destroy_error == false);
+}
+
+sem_t global_semaphore;
+
+int main(void)
+{
+  global_semaphore = create_semaphore(0);
+  destroy_semaphore(&global_semaphore);
+  return EXIT_SUCCESS;
+}
--- a/regression/cbmc-sequentialization/pthread_mutex/no_mutex.c
+++ b/regression/cbmc-sequentialization/pthread_mutex/no_mutex.c
@@ -1,0 +1,52 @@
+#include <assert.h>
+#include <pthread.h>
+#include <sched.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+int shared_count = 0;
+
+void *worker(void *arguments)
+{
+  for(int i = 0; i < 100; ++i)
+  {
+    int shared_count_copy = shared_count;
+    // The following call to yield is here in order to increase the chance of
+    // thread swaps during concrete execution in order to show unsoundness.
+    sched_yield();
+    ++shared_count_copy;
+    shared_count = shared_count_copy;
+  }
+  pthread_exit(NULL);
+}
+
+pthread_t start_worker_thread(void)
+{
+  pthread_t worker_thread;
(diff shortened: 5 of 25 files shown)


regression/cbmc-sequentialization/posix_semaphores/blocking_queue.c
```diff
@@ -0,0 +1,118 @@
...
+  enqueue(&input_queue, 2);
+  enqueue(&input_queue, 3);
+  enqueue(&input_queue, 4);
+  enqueue(&input_queue, 5);
+  enqueue(&input_queue, 6);
+  enqueue(&input_queue, 7);
+  enqueue(&input_queue, 8);
+  enqueue(&input_queue, 9);
+  enqueue(&input_queue, 0);
+  join_thread(worker_thread1);
```
martin-cs, line 140: I like this example and it should sequentialize well.

regression/cbmc-sequentialization/pthread_mutex/no_mutex.c
```diff
@@ -0,0 +1,52 @@
...
+#include <stdlib.h>
+
+int shared_count = 0;
+
+void *worker(void *arguments)
+{
+  for(int i = 0; i < 100; ++i)
+  {
+    int shared_count_copy = shared_count;
+    // The following call to yield is here in order to cause a thread swap
```
martin-cs: The wording of this sounds a bit certain.  I can believe that 99.99% of the time it does but I don't think it is certain to do so.

regression/cbmc-sequentialization/pthread_mutex/with_mutex.c
```diff
@@ -0,0 +1,63 @@
...
+pthread_mutex_t mutex;
+
+void *worker(void *arguments)
+{
+  for(int i = 0; i < 100; ++i)
+  {
+    const int lock_error = pthread_mutex_lock(&mutex);
+    assert(!lock_error);
+    int shared_count_copy = shared_count;
+    // The following call to yield is here in order to cause a thread swap
```
martin-cs: As above.

regression/cbmc-sequentialization/pthread_mutex/no_mutex.c
martin-cs: I would feel better if the comment was reworded so that it captured this kind of "likely but not certain" but it is not a blocking issue.

piotr-grabalski (approved):
Lots of code but it looks good to me, 

_Personally I am not a big fan of including *.c files though_ ;)

regression/cbmc-sequentialization/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+  set(gcc_only "-X gcc-only")
```
NlightNFotis: This is incorrect I believe - it should be: `set(gcc_only -X gcc-only)`. 

The difference is that the code as it stands now passes this as a single string, which passed to `test.pl` creates a problem.

For a more detailed explanation, take a look at the commit message here: https://github.com/diffblue/cbmc/pull/5446/commits/741c303064ab837778087c047b8f1df56d65e549

hannes-steffenhagen-diffblue: No quotes, this is the same bug fotis was struggling on windows with :)

If you add quotes here this'll be passed as a single argument to test.pl, i.e. `test.pl "-X gcc only"` vs `test.pl -X gcc-only`.

regression/cbmc-sequentialization/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+  set(gcc_only "")
+endif()
+
+add_test_pl_tests(
+    "$<TARGET_FILE:cbmc> --validate-goto-model --validate-ssa-equation" -X smt-backend ${gcc_only}
+)
+
+add_test_pl_profile(
+    "cbmc-paths-lifo"
+    "$<TARGET_FILE:cbmc> --paths lifo ${gcc_only}"
```
NlightNFotis: I believe `${gcc_only}` should be outside the quotes here, but I'm not sure.

@hannes-steffenhagen-diffblue ?

regression/cbmc-sequentialization/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+    "-C;-X;thorough-paths;-X;smt-backend;-X;paths-lifo-expected-failure;-s;paths-lifo"
+    "CORE"
+)
+
+add_test_pl_profile(
+    "cbmc-cprover-smt2"
+    "$<TARGET_FILE:cbmc> --cprover-smt2 ${gcc_only}"
```
NlightNFotis: Same as above.

regression/cbmc-sequentialization/CMakeLists.txt
hannes-steffenhagen-diffblue: This gcc_only needs to be in the next line, its an argument to test.pl not cbmc.

regression/cbmc-sequentialization/CMakeLists.txt
hannes-steffenhagen-diffblue: same

NlightNFotis (changes requested):
Can we get changes on the `CMakeLists.txt`. I think they are correctness related and pretty important.

martin-cs (approved):
Seems good.  Thanks for adding the tests first.  Will you have these run with and without the sequentialisation?  Also, I take it whoever is doing the sequentialisation is aware of the considerable amount of literature on "seqeuntialise and then run CBMC"?

regression/cbmc-sequentialization/pthread_cond/test.c
```diff
@@ -0,0 +1,82 @@
...
+  assert(!pthread_mutex_init_error);
+
+  pthread_condattr_t *condition_attributes = NULL;
+  int condition_init_error =
+    pthread_cond_init(&condition, condition_attributes);
+  assert(!condition_init_error);
+}
+
+// Wait for signal. Note that pthreads requires us to lock and unlock the mutex
+// around the call to `pthread_cond_wait`.
```
hannes-steffenhagen-diffblue: Maybe worth pointing out if you want to clarify behavior unfamiliar with pthreads that in the period between calling `pthread_cond_wait` and it returning the mutex is actually unlocked, just in case anyone is wondering if this doesn't block everything else.

regression/cbmc-sequentialization/pthread_cond/test.c
```diff
@@ -0,0 +1,82 @@
...
+void wait(void)
+{
+  const int lock_error = pthread_mutex_lock(&signal_mutex);
+  assert(!lock_error);
+  const int wait_error = pthread_cond_wait(&condition, &signal_mutex);
```
hannes-steffenhagen-diffblue: Writing these comments as I go – note that POSIX requires an associated boolean condition with the condition variable, and correct wake-up behaviour should always be a loop that's waiting and checking the condition after the wait call. Even if the condition you want to wait on is just "someone called signal() at some point" this does, at minimum, still require a flag to operate correctly.

> When using condition variables there is always a Boolean predicate involving shared variables associated with each condition wait that is true if the thread should proceed. Spurious wakeups from the pthread_cond_timedwait() or pthread_cond_wait() functions may occur. Since the return from pthread_cond_timedwait() or pthread_cond_wait() does not imply anything about the value of this predicate, the predicate should be re-evaluated upon such return. 

Without this the expected result for this benchmark should actually be a failure, as a spurious wakeup is possible (this also means that actually we literally don't need to do anything to model `signal` and `wait` correctly other than unlocking and then locking the mutex...). Which to be clear isn't a bad thing, "there is a bug here and we can find it" is a more interesting result (and therefore test) than "we can't find anything here", but this should be documented as such so a future developer doesn't scratch their head why this test is failing despite everything seemingly working as intended.

regression/cbmc-sequentialization/pthread_mutex/no_mutex.c
hannes-steffenhagen-diffblue: That'd make more sense to me as well, practically this should work but IMHO it’s probably not a good idea to be overly reliant on specific thread scheduling behaviour.

regression/cbmc-sequentialization/pthread_mutex/no_mutex.c
```diff
@@ -0,0 +1,52 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+int shared_count = 0;
+
+void *worker(void *arguments)
+{
+  for(int i = 0; i < 100; ++i)
+  {
+    int shared_count_copy = shared_count;
```
hannes-steffenhagen-diffblue, line 14: might be worth pointing out here that the odd construction with the copy-then-increment is because our intended approach (sequentialisation) will give us sequential consistency at the statement level, so an increment operation is going to be atomic for us.

regression/cbmc-sequentialization/CMakeLists.txt
thomasspriggs: Should be fixed now.

regression/cbmc-sequentialization/CMakeLists.txt
thomasspriggs: Should be fixed now.

regression/cbmc-sequentialization/CMakeLists.txt
thomasspriggs: Should be fixed now.

regression/cbmc-sequentialization/CMakeLists.txt
thomasspriggs: Should be fixed now. Sorry I reintroduced this error, it was a result of cut and pasting from before your fix was merged.

regression/cbmc-sequentialization/CMakeLists.txt
thomasspriggs: It was actually on completely the wrong line.

NlightNFotis (approved):
Thanks for the changes. LGTM! 🎉

regression/cbmc-sequentialization/CMakeLists.txt
thomasspriggs: Actually none of this is going to work on Windows as-is, because pthreads isn't natively supported on that platform. I'll either need to disable all these tests, or add a windows test-time dependency for a suitable compatibility layer.

regression/cbmc-sequentialization/pthread_mutex/no_mutex.c
thomasspriggs: I have reworded these comments slightly.

thomasspriggs:
> Seems good. Thanks for adding the tests first. Will you have these run with and without the sequentialisation? Also, I take it whoever is doing the sequentialisation is aware of the considerable amount of literature on "seqeuntialise and then run CBMC"?

We have been looking into the existing literature and implementations. My thought is to run these specific tests with sequentialisation only, once it is implemented. This is because these test run up against existing limitations without the sequentialisation step. My thought is that the new workflow will look something like -

1. Run goto-cc on the source.
2. Run goto-sequentialise on the goto-binary.
3. Run CBMC against the sequentialised goto-binary. 

Therefore this directory will need some kind of `chain` script, once `goto-sequentialise` exists. Presuming that we keep the sequentialisation in a separate binary.

martin-cs:
@thomasspriggs : Some thoughts

- Last I looked CSeq and Yogar-CBMC were the best sequentialisation tools.
- `goto-instrument` already has some concurrent instrumentation (for weak memory), so you could add it there if you want it to implement it as part of a pipeline.
- Most of the sequentialisations I have seen have been underapproximate and slowly increasing the number of interleavings, thread interactions, etc.  It feels like it should even be possible to use the SAT model to tell you where needs more freedom (as with BMC and finding which loops need more unwinding).  As such, you may want to do multiple sequentialisations and have the parameters of the later ones depend on the output of the earlier ones.  If so, do you want to do that with multiple program calls and a script or have one program?

HTH

hannes-steffenhagen-diffblue (commented):
Not done reviewing yet just to get these out of the way:

regression/cbmc-sequentialization/posix_semaphores/util.c
```diff
@@ -0,0 +1,33 @@
+
```
hannes-steffenhagen-diffblue: I do have to agree the `util.c` is _slightly_ sketchy, generally I prefer if tests stay self contained, even if that means duplicating some code. If you think it's worth it to give more emphasis to the parts that are actually different between tests we can leave it as is though.

regression/cbmc-sequentialization/posix_semaphores/create_destroy.c
```diff
@@ -0,0 +1,20 @@
+#include <assert.h>
+#include <pthread.h>
+#include <semaphore.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+sem_t global_semaphore;
+
+void *worker(void *arguments)
```
hannes-steffenhagen-diffblue: :pick: Unused function?

regression/cbmc-sequentialization/posix_semaphores/util.c
```diff
@@ -0,0 +1,33 @@
...
+pthread_t start_worker_thread(void)
+{
+  pthread_t worker_thread;
+  const pthread_attr_t *const attributes = NULL;
+  void *const worker_argument = NULL;
+  const int create_status =
+    pthread_create(&worker_thread, attributes, &worker, worker_argument);
```
hannes-steffenhagen-diffblue: I really don’t like that this assumes the existence of a function with exactly this name, what's wrong with passing this in as an argument?

regression/cbmc-sequentialization/posix_semaphores/blocking_queue.c
```diff
@@ -0,0 +1,118 @@
...
+  destroy_semaphore(&queue->data_waiting);
+}
+
+void enqueue(struct blocking_queuet *queue, const int data)
+{
+  printf(
+    "enqueue begin:%lu end:%lu data:%d \n", queue->begin, queue->end, data);
+  const int free_wait_error = sem_wait(&queue->free_space);
+  assert(free_wait_error == false);
+  queue->elements[queue->end] = data;
```
hannes-steffenhagen-diffblue, line 85: :question: 

Is the intention here that `enqueue` and `dequeue` are only allowed to be called from one thread respectively? Otherwise this would be a data race. If so, can you document this here?

regression/cbmc-sequentialization/rw_lock/race_test.c
```diff
@@ -0,0 +1,2 @@
+#define NO_LOCKS
```
hannes-steffenhagen-diffblue: Again, I don't like having tests split up over multiple files like this too much if we don't have to..

regression/cbmc-sequentialization/rw_lock/with_lock.c
```diff
@@ -0,0 +1,138 @@
...
+  sched_yield();
+  shared->c = 6;
+}
+
+void *writer1(void *arguments)
+{
+  struct sharedt *shared = (struct sharedt *)arguments;
+  for(int i = 0; i < 1000; ++i)
+  {
+#ifndef NO_LOCKS
```
hannes-steffenhagen-diffblue: I think these would be easier to follow if you had just copy&pasted the code.

regression/cbmc-sequentialization/posix_semaphores/blocking_wait.desc
```diff
@@ -1,4 +1,4 @@
-FUTURE
+FUTURE requires_posix_only_headers
```
hannes-steffenhagen-diffblue, line 1: I'd be careful with naming like this, AFAIK Windows is (or at least has been at some point) POSIX compliant.

hannes-steffenhagen-diffblue (approved):
Couple nitpicks here and there, nothing blocking I think.

regression/cbmc-sequentialization/posix_semaphores/blocking_queue.c
thomasspriggs: Yes. This queue implementation is only safe with a single enqueuing and dequeuing thread. I will add some comments to detail this limitation.

regression/cbmc-sequentialization/posix_semaphores/blocking_wait.desc
thomasspriggs: A standard install of Visual Studio does not make the POSIX header files we need available as part of its include path. Without these the build environment is does not have the POSIX compliance level we need. The windows API calls for threading are available through separate header and will have different naming conventions entirely. If a Windows based build environment were to be setup with the appropriate POSIX header files made available, then it should be possible to make these tests work on Windows.

Technically I believe these these would need a minimum of POSIX.1b for semaphores and POSIX.1c for thread creation and joining. The original POSIX.1 version would not be sufficient. But version POSIX.1-2001 onward would have everything we need.

regression/cbmc-sequentialization/posix_semaphores/blocking_queue.c
thomasspriggs: Added.

regression/cbmc-sequentialization/posix_semaphores/create_destroy.c
thomasspriggs: Removed.

regression/cbmc-sequentialization/posix_semaphores/util.c
thomasspriggs: I have now fixed this for the semaphores examples.

regression/cbmc-sequentialization/rw_lock/race_test.c
thomasspriggs: Ok. I have pushed a commit which duplicates this instead.

regression/cbmc-sequentialization/rw_lock/with_lock.c
thomasspriggs: Done.

regression/cbmc-sequentialization/posix_semaphores/util.c
thomasspriggs: I have removed `util.c` and duplicated the code. I instinctively dislike cut and pasting code. But I am willing to accept that different ways of doing things may be better for test code.

hannes-steffenhagen-diffblue (approved):
Thanks for indulging my nitpicks here

PR #5523, opened 2020-10-13
Draft/goto unwind

hannes-steffenhagen-diffblue:
This is supposed to be a tool that ‘unwinds’ goto-programs, by unwinding loops,
inlining and following function calls (including recursively) from some starting
point, similar to what CBMC does in its symex phase but producing a goto-program
instead of an SSA program.

For now this is just a skeleton that only does loop-unwinding like the existing
`--unwind` (and related) flags in goto-instrument, but these don’t quite fit our
needs on this and so we decided that putting this funcitonality into a separate
tool would be easier to manage (rather than adding even more configuration
options to the already fairly crowded `goto-instrument`).

---

Note that for now I'd prefer to avoid putting any of this into the manual, as we haven’t decided on a stable UI for this yet, this is very much WIP.

---


src/CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ endmacro(generic_flex)
 
 #   Set the public include locations for a target.
 macro(generic_includes name)
+    message("GENERIC INCLUDES: ${name}; CBMC_SOURCE_DIR: ${CBMC_SOURCE_DIR}")
```
thomasspriggs: :question: Is this debugging code? It doesn't seem to fit with the stated purpose of the commit.

martin-cs (commented):
Thoughts on the PR itself:

- It would be good to have `src/goto-unwind/src/goto_unwind_parse_options.h` with the options
- We should also have CODEOWNERS assigned for the tool.

General thoughts:

I don't see the need for a separate tool.  Transforming goto-programs, including unwinding and inlining is *literally* what `goto-instrument` does.  Rather than duplicating not only the functionality *and* the tool, why not modify or extend what already exists?  What is inadequate or insufficient about the existing way of doing this?

src/goto-unwind/src/goto_unwind_main.cpp
```diff
@@ -0,0 +1,10 @@
+/// \file
+/// \author Diffblue Ltd.
+
+#include "goto_unwind_parse_options.h"
+
+int main(int argc, char **argv)
+{
+  goto_unwind_parse_optionst goto_unwind(argc, const_cast<const char **>(argv));
```
thomasspriggs, line 8: :question: Why is this `const_cast` being used instead of just defining main as `int main(const int argc, const char **argv)`?

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+  /// passed, returns a filled-in option with the value that was passed if the
+  /// option was passed, throws an exception if the option was passed but the
+  /// option value couldn't be parsed.
+  template <typename T>
+  static optionalt<T> parse_int_option(
+    const cmdlinet &cmdline,
+    const std::string &option_name,
+    const std::string &error_message)
+  {
+    if(cmdline.isset(option_name.c_str()))
```
thomasspriggs, line 56: :bulb: It would be great if `cmdlinet::isset` and `cmdlinet::get_value` had overloads which took a `std::string` typed parameter, so we didn't need this conversion back to a C string.

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+      }
+      else
+      {
+        throw invalid_command_line_argument_exceptiont{
+          error_message +
+            "(value was: " + cmdline.get_value(option_name.c_str()) + ")",
+          "--" + option_name};
+      }
+    }
+    return nullopt;
```
thomasspriggs, line 72: :pick: `return nullopt;` and `return {};` are equivalent. However I would prefer `return {};`, as I have come across that form more frequently in our code base.

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  static goto_unwind_optionst from_cmdline(const cmdlinet &cmdline)
+  {
+    goto_unwind_optionst goto_unwind_options{};
+
+    if(cmdline.args.size() == 1)
+    {
+      goto_unwind_options.source_goto_binary_filepath =
```
thomasspriggs: :pick: Taking the binary file path from the first argument is duplicated in the bodies of code in these 2 if conditions. Factoring it out would help conform to the DRY (Don't Repeat Yourself) principal better.

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+    {
+      throw invalid_command_line_argument_exceptiont{
+        "Expecting exactly one or two arguments",
+        "<Source binary> [<Destination binary>]"};
+    }
+
+    goto_unwind_options.unwind_limit =
+      parse_int_option<std::size_t>(
+        cmdline, unwind_limit_option_name, "couldn't parse as size_t")
+        .value_or(10);
```
thomasspriggs: :pick: It would be nice to have this default stored in a separate constant somewhere and included in the help text.

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+            "(value was: " + cmdline.get_value(option_name.c_str()) + ")",
+          "--" + option_name};
+      }
+    }
+    return nullopt;
+  }
+
+  static goto_unwind_optionst from_cmdline(const cmdlinet &cmdline)
+  {
+    goto_unwind_optionst goto_unwind_options{};
```
thomasspriggs: :pick: I'd rename `goto_unwind_options` to just `options`. It is clear what kind of options this from the context and it would help make the rest of this function less verbose.

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+    }
+    else if(cmdline.args.size() == 2)
+    {
+      goto_unwind_options.source_goto_binary_filepath = cmdline.args[0];
+      goto_unwind_options.destination_goto_binary_filepath = cmdline.args[1];
+    }
+    else
+    {
+      throw invalid_command_line_argument_exceptiont{
+        "Expecting exactly one or two arguments",
```
thomasspriggs: :pick: This message is a little confusing from a user perspective. Some thing like the following would be a valid invocation with 4 arguments - `goto-unwind ./main.gb main_unwound.gb --unwind-limit 10`.

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+int goto_unwind_parse_optionst::doit()
+{
+  const auto options = goto_unwind_optionst::from_cmdline(cmdline);
+  auto goto_model = get_goto_model(options.source_goto_binary_filepath);
+  do_unwind(goto_model, options.unwind_limit);
+  write_goto_binary(
+    options.destination_goto_binary_filepath,
+    goto_model,
+    log.get_message_handler());
+  return 0;
```
thomasspriggs: :no_entry_sign: We should use `CPROVER_EXIT_SUCCESS` instead of this magic value.

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+  log.status() << messaget::eom;
+}
+
+goto_modelt
+goto_unwind_parse_optionst::get_goto_model(const std::string &goto_binary_path)
+{
+  auto goto_model = [&goto_binary_path, this]() {
+    if(
+      auto goto_model =
+        read_goto_binary(goto_binary_path, log.get_message_handler()))
```
thomasspriggs: :bulb: Can we use `ui_message_handler` instead of `log.get_message_handler()`? That way this condition would fit on one line instead of three.

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+      return std::move(*goto_model);
+    }
+    throw invalid_source_file_exceptiont{"Failed to read goto model from " +
+                                         goto_binary_path};
+  }();
+
+  // function pointers need to be removed for unwinding to take place
+  const bool add_safety_assertion = true;
+  const bool only_remove_const_fps = false;
+  remove_function_pointers(
```
thomasspriggs, line 151: :pick: This is kind of a surprising place to find this. Could it be moved into a separate function which is then called from `goto_unwind_parse_optionst::doit`? That way we keep all the calls to different passes in the same place.

src/goto-unwind/src/goto_unwind_parse_options.h
```diff
@@ -0,0 +1,23 @@
...
+
+class goto_unwind_parse_optionst : public parse_options_baset
+{
+public:
+  goto_unwind_parse_optionst(int argc, const char **argv);
+
+  int doit() override;
+  void help() override;
+  goto_modelt get_goto_model(const std::string &goto_binary_path);
+  void do_unwind(goto_modelt &model, size_t limit);
```
thomasspriggs, line 22: :pick: The `get_goto_model` and `do_unwind` functions have been made part of the public interface, but they are only used from inside the `goto_unwind_parse_optionst` class and they are missing doxygen. I think these should be made `protected` at the least, but I would probably have written them as free functions myself, in order to avoid implicitly passing them the entire state of the class.

src/goto-unwind/src/goto_unwind_parse_options.cpp
```diff
@@ -0,0 +1,157 @@
...
+    log.get_message_handler(),
+    goto_model,
+    add_safety_assertion,
+    only_remove_const_fps);
+
+  return goto_model;
+}
+
+void goto_unwind_parse_optionst::do_unwind(goto_modelt &model, size_t limit)
```
thomasspriggs: :pick: `limit` should be `const`, for reasons of const-correctness.

thomasspriggs (changes requested):
This covers the first commit only. I haven't reviewed the tests yet.

src/goto-unwind/src/goto_unwind_parse_options.cpp
hannes-steffenhagen-diffblue: Different strokes I guess, I prefer `nullopt` somewhat because it makes it more clear that that’s what what’s intended.

src/goto-unwind/src/goto_unwind_parse_options.cpp
hannes-steffenhagen-diffblue: For me that’s "2 arguments and one option", but I get what you mean, I'll have a think about how to reword this.

regression/goto-unwind/chain.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+set -e
+
+goto_cc=$1
+goto_unwind=$2
+cbmc=$3
+is_windows=$4
+
+name=${*:$#}
```
thomasspriggs, line 31: :pick: I am finding it difficult to follow what this bash script is doing. Some of this is because I am not sufficiently practised in bash. But this script also has no documentation, no comments and many special cases for different platforms. Just having a `readme.md` in this directory explaining the intended overall process flow would help a great deal. Do you think we would have ended up with fewer special cases if this had been written as a python based `chain.py` instead of a bash based `chain.sh`?

regression/goto-unwind/basic-test/test_unwind_not_enough.desc
```diff
@@ -0,0 +1,24 @@
+CORE
+test.c
+--unwind-limit 10
```
thomasspriggs, line 3: :no_entry_sign: We really need to document which binary these arguments are being passed to. Without this documentation it takes some detective work either reading the `chain.sh`, or searching the code base looking what entry point `--unwind-limit` is valid for. I would put this documentation in a `regression/goto-unwind/readme.md` file.

src/CMakeLists.txt
hannes-steffenhagen-diffblue: I actually have no idea where this is coming from, will remove. Must've added that weeks ago for something else.

src/goto-unwind/src/goto_unwind_main.cpp
hannes-steffenhagen-diffblue: @thomasspriggs Technically speaking `()` and `(int, char*[])` are the only standard defined main signatures. `const char**` is a different sort of thing, which is why the cast is necessary in the first place (`char **const` would be a different story, but also wouldn’t be helpful here). Would I expect it to work regardless? Sure. But it’s not technically standards conformant anymore.

regression/goto-unwind/chain.sh
```diff
@@ -0,0 +1,52 @@
...
+
+if [[ x$name == x ]];  then
+  name=${name%.c}
+fi
+
+args=${*:5:$#-5}
+
+if [[ "${is_windows}" == "true" && x$name != x ]]; then
```
thomasspriggs: :grey_question: What is `x$name`?

thomasspriggs (changes requested):

src/goto-unwind/src/goto_unwind_main.cpp
thomasspriggs: Ah. Thanks for the explanation. I guess ideally we would update all the main functions to have type `(int, char*[])` and update `parse_options_baset` to make the type of its argv be `const char *const *const argv`. That way complete constness could be added with an implicit cast. I now understand why the `const_cast` is currently needed here.

src/goto-unwind/src/goto_unwind_parse_options.cpp
hannes-steffenhagen-diffblue: @thomasspriggs if we ever upgrade to C++17 we should just use `string_view` for this purpose as that’ll work frictionless-ly with both.

src/goto-unwind/src/goto_unwind_parse_options.cpp
hannes-steffenhagen-diffblue: I kinda disagree on that one? For me they’re two separate things, not one thing where a bit extra happens sometimes.

src/goto-unwind/src/goto_unwind_parse_options.cpp
hannes-steffenhagen-diffblue: Main reason I put `goto_unwind_options` there is to avoid confusion with `optionst`, I’ll try renaming and see if I like it.

src/goto-unwind/src/goto_unwind_parse_options.cpp
hannes-steffenhagen-diffblue: ok

src/goto-unwind/src/goto_unwind_parse_options.cpp
hannes-steffenhagen-diffblue: I did not actually know this member existed.

src/goto-unwind/src/goto_unwind_parse_options.h
hannes-steffenhagen-diffblue: Well, the main advantage of having them be member functions is to have `log` immediately available and have a namespace for them. But I’ll make them private.

regression/goto-unwind/chain.sh
hannes-steffenhagen-diffblue: This is a really hacky way to check if `$name` is empty (the problem being that this otherwise might not parse correctly). This is basically saying "is x with $name appended still x?".

I've removed these cases because they’re not really needed here (I've taken this script from `goto-instrument` and just changed some bits).

regression/goto-unwind/chain.sh
hannes-steffenhagen-diffblue: It wouldn’t be much simpler in python, though it might be more familiar looking. In any case I've somewhat simplified and documented the file.

src/goto-unwind/src/goto_unwind_parse_options.cpp
hannes-steffenhagen-diffblue: That’s not what I’d use the term const-correctness for, but sure I think that’s a good idea.

src/goto-unwind/src/goto_unwind_parse_options.cpp
thomasspriggs: :shrug: I am not going to argue this any further as both approaches are workable. However this is how I would do this -
```
    if(cmdline.args.empty() || cmdline.args.size() > 2)
    {
        "Expecting either a single binary to read from and modify, or both a "
        "source and a destination binary",
        "<Source binary> [<Destination binary>]"};
    }
    options.source_goto_binary_filepath = cmdline.args[0];
    options.destination_goto_binary_filepath =
      cmdline.args.size() >= 2 ? cmdline.args[1]
                               : options.source_goto_binary_filepath;
````

File: regression/goto-unwind/chain.sh (hunk 16 above)
Comment (thomasspriggs) [reply regression/goto-unwind/chain.sh]: Ah, that makes sense. Thanks for the simplification.

**Review by thomasspriggs — APPROVED** [review]
Thanks for the updates. This is looking much better.

File: src/goto-unwind/src/goto_unwind_parse_options.cpp [hunk 17]
```diff
@@ -102,6 +102,13 @@ struct goto_unwind_optionst
   }
 };
 
+// due to an interesting quirk with C++11 we do need to define these
+// even if they are static constexpr... C++17 removed this limitation
+// so if we ever manage to upgrade please remove these lines and forget
+// they ever existed.
+constexpr const char *goto_unwind_optionst::unwind_limit_option_name;
```
thomasspriggs, line 109: :grey_question: Could we avoid the need for this by using `const` instead of `constexpr`? It isn't like this is performance sensitive or like we are adding static asserts based on it.

hannes-steffenhagen-diffblue: @thomasspriggs it's more of a semantics thing for me, it’s a constexpr because it’s a constant expression. The only difference in making this `const` would be that this being required wouldn't be weird anymore, it’d still be required.

hannes-steffenhagen-diffblue:
Closing for now because it seems it'll be a while until we do something in this direction again.

PR #5524, opened 2020-10-14
Introduce reference to avoid unneeded copy.

NlightNFotis:
Previously this broke the build on macOS, because the compiler was complaining about creating a copy of a `const std::pair` - which
registered as an error (because of `-Werror`) - on Apple Clang 12.0.0.


hannes-steffenhagen-diffblue (approved):
LGTM. Reports like this make me wonder though if there's a way to tell compilers you’re _intentionally_ making a copy. I guess you wouldn't do that too often and having an extra assignment statements in those cases should be sufficient to signal this to developers and compilers?

peterschrammel (approved):

PR #1997, opened 2018-04-03
Do not sort operands as part of simplification [blocks: #3486]

tautschnig:
~Do-not-merge as above: this needs proper profiling to confirm the effect with data.~

PR #5539, opened 2020-10-27
Performance optimisation - Use `unordered_map` in `numbering` instead of `map`

thomasspriggs:
The commit which specifically gives the performance improvement is titled "Make `union_find` use hashes for numbering". The other commits are related refactoring.

Using `unordered_map` in `numbering` instead of `map`, changes `union_find`, which changes `arrays`, where the performance issue was seen. This change avoids slow lookups into `std::map` caused by long running `irept::operator<` calls. I have an example where cbmc spent approximately 70% of runtime in `irept::operator<` before this change and approximately 1% of runtime in hashing/ordering after this change.

Run time on example exercising arrays code before this PR -
real	2m39.018s
user	2m37.394s
sys	0m1.608s

Run time on example exercising arrays code after this PR -
real	0m30.897s
user	0m29.476s
sys	0m1.404s

Gprof before (top results up to `has_byte_operator` + `irept::hash()`) -
```
Each sample counts as 0.01 seconds.
  %   cumulative   self              self     total           
 time   seconds   seconds    calls   s/call   s/call  name    
 44.90     21.08    21.08  2025113     0.00     0.00  irept::compare(irept const&) const
 26.84     33.68    12.60 4483321688     0.00     0.00  irept::number_of_non_comments(std::map<dstringt, irept, std::less<dstringt>, std::allocator<std::pair<dstringt const, irept> > > const&)
  6.54     36.75     3.07 52261322     0.00     0.00  Minisat::SimpSolver::addClause_(Minisat::vec<Minisat::Lit>&)
  3.43     38.36     1.61 52261322     0.00     0.00  Minisat::Solver::addClause_(Minisat::vec<Minisat::Lit>&)
  2.26     39.42     1.06    58881     0.00     0.00  has_byte_operator(exprt const&)
  0.06     46.03     0.03  3365106     0.00     0.00  irept::hash() const
```

Gprof after (top results up to `has_byte_operator` + `irept::hash()`) -
```
Each sample counts as 0.01 seconds.
  %   cumulative   self              self     total           
 time   seconds   seconds    calls   s/call   s/call  name    
 27.40      3.11     3.11 52261322     0.00     0.00  Minisat::SimpSolver::addClause_(Minisat::vec<Minisat::Lit>&)
 13.74      4.67     1.56 52261322     0.00     0.00  Minisat::Solver::addClause_(Minisat::vec<Minisat::Lit>&)
  5.73      5.32     0.65 10600597     0.00     0.00  Minisat::SimpSolver::newVar(Minisat::lbool, bool)
  5.29      5.92     0.60    58881     0.00     0.00  has_byte_operator(exprt const&)
  0.18     10.51     0.02  3424407     0.00     0.00  irept::hash() const
```

### DIFF
--- a/src/analyses/custom_bitvector_analysis.h
+++ b/src/analyses/custom_bitvector_analysis.h
@@ -163,7 +163,7 @@ public:
 
   unsigned get_bit_nr(const exprt &);
 
-  typedef numbering<irep_idt> bitst;
+  typedef numberingt<irep_idt> bitst;
   bitst bits;
 
 protected:
--- a/src/analyses/escape_analysis.h
+++ b/src/analyses/escape_analysis.h
@@ -120,7 +120,7 @@ protected:
   {
   }
 
-  numbering<irep_idt> bits;
+  numberingt<irep_idt> bits;
 
   void insert_cleanup(
     goto_functionst::goto_functiont &,
--- a/src/analyses/invariant_set.h
+++ b/src/analyses/invariant_set.h
@@ -55,7 +55,7 @@ public:
 protected:
   const namespacet &ns;
 
-  typedef hash_numbering<irep_idt, irep_id_hash> mapt;
+  typedef numbering<irep_idt> mapt;
   mapt map;
 
   struct entryt
--- a/src/analyses/invariant_set.h
+++ b/src/analyses/invariant_set.h
@@ -55,7 +55,7 @@ public:
 protected:
   const namespacet &ns;
 
-  typedef numbering<irep_idt> mapt;
+  typedef numberingt<irep_idt> mapt;
   mapt map;
 
   struct entryt
--- a/src/analyses/local_bitvector_analysis.h
+++ b/src/analyses/local_bitvector_analysis.h
@@ -183,7 +183,7 @@ protected:
 
   typedef std::stack<unsigned> work_queuet;
 
-  numbering<irep_idt> pointers;
+  numberingt<irep_idt> pointers;
 
   // pointers -> flagst
   // This is a vector, so it's fast.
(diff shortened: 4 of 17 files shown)


martin-cs (approved):
Good catch.  One thing to have a look at is the status of caching of the hashes in `dt`.  IIRC @tautschnig and @peterschrammel worked on this some time ago and I'm not sure if all of the options were enabled.

kroening:
It's odd that irep.h is a dependency of numbering.h.

I'd prefer an 'unordered_numbering' and an 'ordered_numbering', following the naming style of the STL. There may not be a use-case for the ordered numbering.

hannes-steffenhagen-diffblue:
I agree with @kroening here, this can probably be done without special casing ireps. I believe @thomasspriggs is investigating if we can basically just use `hash_numbering` everywhere we can use `numbering` right now (I'd imagine so?).

thomasspriggs:
I have now refactored this PR to avoid either `union_find` or `numbering` depending on `irep.h`. This PR removes the possibility of using an ordered map as the backing for numbering on the basis that this this variation is both unneeded and a potential performance foot-gun. This has the added benefit that we no longer need 3 different names for different numbering variations.

tautschnig:
@thomasspriggs What is the source of those `irept::operator<` calls? Is #1997 perhaps a way to actually address the root cause?

thomasspriggs:
@tautschnig For the particular example I have been examining, much of the overall runtime is spent in `src/solvers/flattening/arrays.cpp`. The `arrayst` class has a `union_find<exprt> arrays;`  field. This instance of `union_find<exprt>` was based on a `numbering<exprt>`, which in turn uses a `std::map<exprt, std::size_t>`, which in turn uses the `irept::operator<`.

From looking at the code I would say that the improvement in this PR looks unrelated to the improvement in #1997. As far as I can see, `arrays.cpp` is not using `sort_and_join`.

tautschnig:
@thomasspriggs Thank you for clarifying!

peterschrammel (approved):

tautschnig (approved):

PR #5541, opened 2020-11-02
Deactivate broken tests that depend on assert on MacOS.

NlightNFotis:
The reason we are deactivating the tests is that there
was a change to the `assert.h` header under MacOS that
makes it now include some other C++ headers which include
`type_traits` causing template errors that have broken
CI under that platform.

### DIFF
--- a/regression/ansi-c/CMakeLists.txt
+++ b/regression/ansi-c/CMakeLists.txt
@@ -1,15 +1,32 @@
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-add_test_pl_tests(
-    "$<TARGET_FILE:goto-cc>" -X gcc-only
-)
+    add_test_pl_tests(
+        "$<TARGET_FILE:goto-cc>" -X gcc-only
+    )
+elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
+    add_test_pl_tests(
+        "$<TARGET_FILE:goto-cc>"
+    )
+
+    # In MacOS, a change in the assert.h header file
+    # is causing template errors when exercising the
+    # C++ front end (because of a transitive include
+    # of <type_traits>) for files that include the
+    # <assert.h> or <cassert> headers.
+    add_test_pl_profile(
+        "ansi-c-c++-front-end"
+        "$<TARGET_FILE:goto-cc> -xc++ -D_Bool=bool"
+        "-C;-I;test-c++-front-end;-s;c++-front-end-X;macos-assert-broken"
+        "CORE"
+    )
 else()
-add_test_pl_tests(
-    "$<TARGET_FILE:goto-cc>"
-)
-add_test_pl_profile(
-    "ansi-c-c++-front-end"
-    "$<TARGET_FILE:goto-cc> -xc++ -D_Bool=bool"
-    "-C;-I;test-c++-front-end;-s;c++-front-end"
-    "CORE"
-)
+    add_test_pl_tests(
+        "$<TARGET_FILE:goto-cc>"
+    )
+
+    add_test_pl_profile(
+        "ansi-c-c++-front-end"
+        "$<TARGET_FILE:goto-cc> -xc++ -D_Bool=bool"
+        "-C;-I;test-c++-front-end;-s;c++-front-end"
+        "CORE"
+    )
 endif()
--- a/regression/ansi-c/Makefile
+++ b/regression/ansi-c/Makefile
@@ -4,21 +4,35 @@ include ../../src/config.inc
 include ../../src/common
 
 ifeq ($(BUILD_ENV_),MSVC)
-	exe=../../../src/goto-cc/goto-cl -X gcc-only
+	exe = ../../../src/goto-cc/goto-cl
 else
-	exe=../../../src/goto-cc/goto-cc
+	exe = ../../../src/goto-cc/goto-cc
+endif
+
+ifeq ($(BUILD_ENV_),MSVC)
+	excluded_tests = -X gcc-only
+else
+	# In MacOS, a change in the assert.h header file
+	# is causing template errors when exercising the
+	# C++ front end (because of a transitive include
+	# of <type_traits>) for files that include the
+	# <assert.h> or <cassert> headers.
+	OS := $(shell uname)
+	ifeq ($(OS),Darwin)
+		excluded_tests = -X macos-assert-broken
+	endif
 endif
 
 test:
-	@../test.pl -e -p -c $(exe)
+	@../test.pl -e -p -c $(exe) $(excluded_tests)
 ifneq ($(BUILD_ENV_),MSVC)
-	@../test.pl -e -p -c "$(exe) -xc++ -D_Bool=bool" -I test-c++-front-end -s c++-front-end
+	@../test.pl -e -p -c "$(exe) -xc++ -D_Bool=bool" -I test-c++-front-end -s c++-front-end $(excluded_tests)
 endif
 
 tests.log: ../test.pl
-	@../test.pl -e -p -c $(exe)
+	@../test.pl -e -p -c $(exe) $(excluded_tests)
 ifneq ($(BUILD_ENV_),MSVC)
-	@../test.pl -e -p -c "$(exe) -xc++ -D_Bool=bool" -I test-c++-front-end -s c++-front-end
+	@../test.pl -e -p -c "$(exe) -xc++ -D_Bool=bool" -I test-c++-front-end -s c++-front-end $(excluded_tests)
 endif
 
 show:
--- a/regression/ansi-c/array_initialization2/test.desc
+++ b/regression/ansi-c/array_initialization2/test.desc
@@ -1,4 +1,4 @@
-CORE test-c++-front-end
+CORE test-c++-front-end macos-assert-broken
 main.c
 
 ^EXIT=0$
--- a/regression/ansi-c/float_constant2/test.desc
+++ b/regression/ansi-c/float_constant2/test.desc
@@ -1,4 +1,4 @@
-CORE test-c++-front-end
+CORE test-c++-front-end macos-assert-broken
 main.c
 
 ^EXIT=0$
--- a/regression/ansi-c/goto_convert_switch_range_case_valid/test.desc
+++ b/regression/ansi-c/goto_convert_switch_range_case_valid/test.desc
@@ -1,4 +1,4 @@
-CORE test-c++-front-end
+CORE test-c++-front-end macos-assert-broken
 main.c
 
 ^EXIT=0$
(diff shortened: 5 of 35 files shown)


thomasspriggs (approved):

regression/ansi-c/CMakeLists.txt
```diff
@@ -1,15 +1,20 @@
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-add_test_pl_tests(
-    "$<TARGET_FILE:goto-cc>" -X gcc-only
-)
+    add_test_pl_tests(
+        "$<TARGET_FILE:goto-cc>" -X gcc-only
+    )
+elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
+    add_test_pl_tests(
+        "$<TARGET_FILE:goto-cc>" -X mac-frontend-broken
```
hannes-steffenhagen-diffblue: These changes need to go in the Makefile too

‘mac-frontend-broken’ is a bad name. What’s broken about the frontend? Which frontend are we even talking about? In fact this is basically a lie, the C++ frontend on mac isn’t more broken than anywhere else.

martin-cs (commented):
I'm a bit concerned by this.  `assert.h` is a C header and should not be including C++.  The C++ header should be `cassert`.  If `assert.h` doesn't work on Mac then I doubt the tool will be usable at all on Mac.  Are you sure about this?  If it is really a problem with `assert.h` it should break hundreds of regression tests, not just three.

hannes-steffenhagen-diffblue:
@martin-cs Basically what’s happening is that `assert.h` has something along the lines of

```
#ifdef __cplusplus
#include <type_traits>
#define assert(cond) some fancy thing that probably allows for better error reporting or whatever
#else
#define assert(cond) the normal/straightforward way
#endif
```

A lot of standard libraries just have their `cassert` header consist of `#include "assert.h"`

<details>
<summary>here’s the full text of my cassert header for instance:</summary>

```
// -*- C++ -*- forwarding header.

// Copyright (C) 1997-2020 Free Software Foundation, Inc.
//
// This file is part of the GNU ISO C++ Library.  This library is free
// software; you can redistribute it and/or modify it under the
// terms of the GNU General Public License as published by the
// Free Software Foundation; either version 3, or (at your option)
// any later version.

// This library is distributed in the hope that it will be useful,
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
// GNU General Public License for more details.

// Under Section 7 of GPL version 3, you are granted additional
// permissions described in the GCC Runtime Library Exception, version
// 3.1, as published by the Free Software Foundation.

// You should have received a copy of the GNU General Public License and
// a copy of the GCC Runtime Library Exception along with this program;
// see the files COPYING3 and COPYING.RUNTIME respectively.  If not, see
// <http://www.gnu.org/licenses/>.

/** @file cassert
 *  This is a Standard C++ Library file.  You should @c \#include this file
 *  in your programs, rather than any of the @a *.h implementation files.
 *
 *  This is the C++ version of the Standard C Library header @c assert.h,
 *  and its contents are (mostly) the same as that header, but are all
 *  contained in the namespace @c std (except for names which are defined
 *  as macros in C).
 */

//
// ISO C++ 14882: 19.2  Assertions
//

// No include guards on this header...

#pragma GCC system_header

#include <bits/c++config.h>
#include <assert.h>
```

</details>



 and handle all the language differing logic (if any) there (`libstdc++` also has some C++ specific logic in there, but apparently nothing that trips up the C++ frontend – note that on Mac the default is `libc++`, not `libstdc++` though). As long as you don’t try to compile with `-x c++` there’s no problem.

martin-cs:
OK... but then why are these tests, in `ansi-c` being compiled with the C++ flag?  I've had a look an these are not C++ tests.  Please look into this more deeply and fix a different way.  Just disabling tests is not a solution.

martin-cs:
Also, while I am at it, @NlightNFotis the code coverage is broken.  I remember commenting on this when you were changing the CI infrastructure and you said it was temporary and would go away.  It hasn't.  Please can you fix it?

hannes-steffenhagen-diffblue:
> OK... but then why are these tests, in ansi-c being compiled with the C++ flag? 

The reason seems to be related to something @tautschnig did in 2019, apparently the idea was that these tests should also pass if the C++ frontend is used to compile them. I think perhaps a better way would be to move them over to the C++ frontend tests, please note that this is currently blocking our CI (and thus everything) though. The correct solution here is going to involve disabling these tests on macOS or deleting them (for C++, not for C!) though, as the reason they’re broken is not a regression – it’s just that the C++ frontend is flaky.

martin-cs:
They are not C++ tests!  Have a look at them!  The correct fix is to work out why they are being compiled as C++ and have them compiled as C instead.

hannes-steffenhagen-diffblue:
> They are not C++ tests! Have a look at them! The correct fix is to work out why they are being compiled as C++ and have them compiled as C instead.

They are compiled as C++ because they are tagged as "this should be compiled with C++". Take a look at 723620dd6c585cfd77776b806bb40d7a7ecc6ca0

Basically what this commit did was it changed the C tests so they would be tested with both the C _and_ the C++ frontend, the idea being that most C tests should also work for a C++ frontend. Whether or not that’s a good idea is neither here nor there, but that’s what’s causing the issue we have right now.

martin-cs:
This raises more questions than it answers...

1. Why is it only these three that cause issues?  Why not the others?
2. Why are the rest of the C++ tests passing on Mac?
3. This should be run as C++ as well as being run as C.

If `assert.h` on Mac is broken when compiled as C++ then I am fine disabling the testing of the C++ front-end on Mac as a solution.

thomasspriggs:
> They are not C++ tests! Have a look at them! The correct fix is to work out why they are being compiled as C++ and have them compiled as C instead.

My understanding is that these tests are run as _both_ C and C++ tests. They are C tests which are also used to check the C++ front-end's compatibility with C code. The problem is that the latest Mac versions of `assert.h` are detecting that they are being used by a C++ front end and then using C++ features, which we don't support.

martin-cs:
@thomasspriggs Thanks.  Do you know why this is not affecting other C++ tests, like the ones in `cpp` or `cbmc-cpp`?

hannes-steffenhagen-diffblue:
> Why is it only these three that cause issues? Why not the others?

I’ve not looked at _all_ of the other tests for which C++ testing is enabled but none of the ones I looked at that aren’t disabled by this PR include `assert.h`.

> Why are the rest of the C++ tests passing on Mac?

Now that’s an excellent question. For the most part it seems the C++ tests also don’t use assertions, but it’s possible the ansi-c tests (C++ version) are somehow treated slightly differently than the C++ proper tests.

martin-cs:
On Mon, 2020-11-02 at 08:41 -0800, hannes-steffenhagen-diffblue wrote:
> > Why is it only these three that cause issues? Why not the others?
> 
> I’ve not looked at _all_ of the other tests for which C++ testing is
> enabled but none of the ones I looked at that aren’t disabled by this
> PR include `assert.h`.

Right; that seems like a good answer.


> > Why are the rest of the C++ tests passing on Mac?
> 
> Now that’s an excellent question. For the most part it seems the C++
> tests also don’t use assertions, but it’s possible the ansi-c tests
> (C++ version) are somehow treated slightly differently than the C++
> proper tests.

Many don't but enough do to make this concerning.

$ find ../regression/ -name '*.cpp' -print | wc -l
405
$ find ../regression/ -name '*.cpp' -exec grep '<assert.h>' {} \; | wc
-l
11
$ find ../regression/ -name '*.cpp' -exec grep '<cassert>' {} \; | wc
-l
49

I don't get why these don't break as well?

../regression/cbmc-cpp/MethodParam1/main.cpp
#include <assert.h>
  assert(x<4);
  assert(x<5);
../regression/cbmc-cpp/lvalue1/main.cpp
#include <cassert>
  assert(a == ASD);
  assert(a == ABC);

NlightNFotis:
One thing to note here is that you are right @martin-cs, more tests are broken than the ones excluded here. I'm in the process of excluding them as well.

NlightNFotis:
Right, so now there should be a lot more tests excluded, which should paint a more accurate picture of what was broken.

There might be one or two missing, but I'm waiting CI to see what's being reported and fix it later.

martin-cs:
Thanks @NlightNFotis this is starting to look like something that will actually fix the problem.  One more thing; please could you add a comment, to the Makefiles or to the test cases, explaining what is broken and why.

regression/ansi-c/CMakeLists.txt
```diff
@@ -1,15 +1,20 @@
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-add_test_pl_tests(
-    "$<TARGET_FILE:goto-cc>" -X gcc-only
-)
+    add_test_pl_tests(
+        "$<TARGET_FILE:goto-cc>" -X gcc-only
+    )
+elseif("${CMAKE_SYSTEM_NAME}" STREQUAL "Darwin")
+    add_test_pl_tests(
+        "$<TARGET_FILE:goto-cc>" -X macos-assert-broken
```
thomasspriggs: Can you double check this? I am not sure the logic is quite correct. As I read this change, the new behaviour on mac will be to not run the c++ front end tests at all, and to skip the C tests, which are marked with `macos-assert-broken`. Or am I reading this incorrectly? I thought we just needed to skip the C++ front end tests on mac which are tagged with `macos-assert-broken` and still run all the C tests on mac.

regression/ansi-c/Makefile
```diff
@@ -9,16 +9,22 @@ else
...
-	@../test.pl -e -p -c $(exe)
+	@../test.pl -e -p -c $(exe) $(exclude_mac_broken_tests)
 ifneq ($(BUILD_ENV_),MSVC)
-	@../test.pl -e -p -c "$(exe) -xc++ -D_Bool=bool" -I test-c++-front-end -s c++-front-end
+	@../test.pl -e -p -c "$(exe) -xc++ -D_Bool=bool" -I test-c++-front-end -s c++-front-end $(exclude_mac_broken_tests)
 endif
 
 tests.log: ../test.pl
-	@../test.pl -e -p -c $(exe)
+	@../test.pl -e -p -c $(exe) $(exclude_mac_broken_tests)
```
thomasspriggs: :no_entry_sign: Likewise, I thought that the C tests were still fine on mac?

tautschnig (approved):

martin-cs (approved):
This seems more consistent and understandable.   Thanks for the discussion and extra work on sorting this out, I think it has really improved it.

regression/ansi-c/CMakeLists.txt
thomasspriggs: Yep the cmake build looks sorted.

PR #5544, opened 2020-11-05
Use bitvector types to implement __builtin_{add,sub,mul}_overflow

tautschnig:
The use of mathematical integers is not supported by the SAT back-end.
Although GCC's specification describes the operations over mathematical
integers, the same precision can be achieved by using bitvector types of
sufficient width.

The implementation now also introduces a fresh symbol to store the
result of the full-precision operation. This ensures that the arguments
to __builtin_{add,sub,mul}_overflow are evaluated exactly once.

### DIFF
--- a/regression/cbmc/gcc_builtin_add_overflow/main.c
+++ b/regression/cbmc/gcc_builtin_add_overflow/main.c
@@ -110,6 +110,15 @@ void check_generic(void)
   assert(0 && "reachability");
 }
 
+void check_non_const()
+{
+  int a, b, c, d, r;
+  // this may overflow (negated assertion fails)
+  assert(!__builtin_add_overflow(a, b, &r));
+  // but need not overflow (assertion fails)
+  assert(__builtin_add_overflow(c, d, &c));
+}
+
 int main(void)
 {
   check_int();
@@ -119,4 +128,5 @@ int main(void)
   check_unsigned_long();
   check_unsigned_long_long();
   check_generic();
+  check_non_const();
 }
--- a/regression/cbmc/gcc_builtin_add_overflow/main.c
+++ b/regression/cbmc/gcc_builtin_add_overflow/main.c
@@ -110,6 +110,17 @@ void check_generic(void)
   assert(0 && "reachability");
 }
 
+void check_generic_p(void)
+{
+  unsigned char small_result = 0;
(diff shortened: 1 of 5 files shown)


kroening:
Note that we've got expressions ID_overflow_plus, ID_overflow_minus and ID_overflow_mult -- these make attempts at optimizing the encoding required for these predicates.

tautschnig:
> Note that we've got expressions ID_overflow_plus, ID_overflow_minus and ID_overflow_mult -- these make attempts at optimizing the encoding required for these predicates.

I am well aware of those, and did wonder whether we can just use them instead of making a big fuzz here. My concern was twofold, and I'd appreciate further thoughts on this:

1) What type should be used for the operands? `__builtin_{add,sub,mul}_overflow` permit using three different types.
2) `__builtin_{add,sub,mul}_overflow` require computing/storing the result, not just performing the overflow check. Thus, preferably, the actual arithmetic operation should only be done once.

kroening:
1. `ID_overflow_{plus, minus, mult}` support the three types needed for `__builtin_{add,sub,mul}_overflow`.

2. There is a bit of savings potential when doing the overflow check for multiplication given the way it's implemented now. No significant savings when doing plus/minus. There is the option to do something iterative:

https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/z3prefix.pdf

tautschnig:
> `ID_overflow_{plus, minus, mult}` support the three types needed for `__builtin_{add,sub,mul}_overflow`.

Where would that be supported? `boolbvt::convert_overflow` appears to assume that the two operands have the same type.

kroening:
https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html appears to suggest that the two operands do indeed have the same type.

tautschnig:
> https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html appears to suggest that the two operands do indeed have the same type.

I'm not sure I can see this? The signature `bool __builtin_add_overflow (type1 a, type2 b, type3 *res)` doesn't seem to require this? The comment also says "The first built-in function allows arbitrary integral types for operands and the result type must be pointer to some integral type other than enumerated or boolean type, [...]"

kroening (approved):

PR #5552, opened 2020-11-09
Add code sign of Windows installer

piotr-grabalski:
Windows installer will now get automatically signed using Diffblue
code sign certificate so that it will no longer appear as from
unknown publisher when installer is executed.
Windows package release action has 4 tasks added:
1. Decoding Personal Information Exchange file (*.pfx) from Github
Secrets into a file.
2. Signing the Windows Installer package with decoded certificate
using signtool from Windows SDK.
3. Deleting decoded certificate file.
4. Verifying signature.

### DIFF
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -112,6 +112,10 @@ jobs:
           choco install winflexbison3
       - uses: microsoft/setup-msbuild@v1.0.1
         name: Setup Visual Studio environment
+      - name: Setup code sign environment
+        run: | 
+          echo "$(Split-Path -Path $(Get-ChildItem -Path ${env:ProgramFiles(x86)} -Recurse -Filter 'signtool.exe' | Where-Object FullName -like '*10.0.19041.0\x64\signtool.exe').FullName)" >> $env:GITHUB_PATH
+          echo "pfxcert=$([string](Get-Location)+'\CodeSignCertificate.pfx')" >> $env:GITHUB_ENV
       - name: Configure with cmake
         run: |
           New-Item -ItemType Directory -Path build
@@ -130,6 +134,23 @@ jobs:
           $msi_name = Get-ChildItem -Filter *.msi -Name
           Write-Output "::set-output name=msi_installer::build/$msi_name"
           Write-Output "::set-output name=msi_name::$msi_name"
+      - name: Decode signing certificate
+        id: decode_certificate
+        run: |
+          $pfx_bytes=[System.Convert]::FromBase64String("${{ secrets.CODESIGNCERTPFX }}")
+          [IO.File]::WriteAllBytes($env:pfxcert, $pfx_bytes)
+      - name: Sign the installer
+        id: code_sign
+        run: |
+          & signtool.exe sign /f $env:pfxcert /p "${{ secrets.CODESIGNCERTPASSWORD }}" /tr http://tsa.starfieldtech.com ${{ steps.create_packages.outputs.msi_installer }}
+      - name: Remove signing certificate
+        id: remove_certificate
+        run: |
+          Remove-Item $env:pfxcert
+      - name: Verify installer signature
+        id: verify_codesign
+        run: |
+          & signtool.exe verify /pa ${{ steps.create_packages.outputs.msi_installer }}
       - name: Get release info
         id: get_release_info
         uses: bruceadams/get-release@v1.2.0


hannes-steffenhagen-diffblue:
Code looks fine to me, but did you check if this actually fails CI if code signing fails? I see no explicit checks for status codes here, and IIRC PowerShell doesn’t automatically fail on non-zero exit codes (nor does cmd for that matter, reason apparently being that status codes were never standardised on Windows so there are some windows applications that return 1 on _success_ instead).

piotr-grabalski:
> Code looks fine to me, but did you check if this actually fails CI if code signing fails? I see no explicit checks for status codes here, and IIRC PowerShell doesn’t automatically fail on non-zero exit codes (nor does cmd for that matter, reason apparently being that status codes were never standardised on Windows so there are some windows applications that return 1 on _success_ instead).

Thanks Hannes for your comment, the current situation is as follows:
Task 1: Decoding Personal Information Exchange file - should never fail - even if there is no secrets the file is successfully created but it's empty.
Task 2: Signing using signtool - action will fail when wrong/no certificate is used or if password is incorrect (for example when secrets are missing). Action will also fail if given version of signtool.exe cannot be found in the system or if there is more than one signtool that can be used - unlikely since the specific version is selected (10.0.19041.0).
Task 3: Removing file created in Task 1 - have not seen this one failing.
Task 4: Signature verification - This can fail only if certificate is no longer valid - have not seen this one failing.

Please let me know if above is not satisfactory and I will try to add check for specific errors.

.github/workflows/release-packages.yaml
```diff
@@ -130,6 +130,28 @@ jobs:
...
+        id: decode_certificate
+        run: |
+          $pfx_filename=[string](Get-Location)+"\CodeSignCertificate.pfx"
+          $pfx_bytes=[System.Convert]::FromBase64String("${{ secrets.CODESIGNCERTPFX }}")
+          [IO.File]::WriteAllBytes($pfx_filename, $pfx_bytes)
+      - name: Sign the installer
+        id: code_sign
+        run: |
+          $pfx_filename=[string](Get-Location)+"\CodeSignCertificate.pfx"
+          $signtool = $(Get-ChildItem -Path ${env:ProgramFiles(x86)} -Recurse -Filter 'signtool.exe' | Where-Object FullName -like '*10.0.19041.0\x64\signtool.exe').FullName
```
hannes-steffenhagen-diffblue: I don’t like that this simultaneously hardcodes a path but then still does a search for it. Are you sure there’s no easier way to get at this path? I’m thinking in particular of something like `vswhere` here, which we already use to find `msbuild` in the `setup-msbuild` action above. I’m actually surprised that doesn’t already add the `signtool` to the path given the VS dev command prompt should do it, and the action advertises itself as working like that.

piotr-grabalski: To be absolutely honest I have not checked singtool by running only signtool.exe, there are multiple versions of Windows Software Development Kit installed on the Github Actions Windows agent with signtool.exe for each of them and since I was testing it offline on latest SDK (version 10.0.19041.0) I decided to go for this approach with search for specific version. Additionally there are multiple platforms supported by signtool (x64, x86, arm, etc.). Maybe the default singtool.exe will be the correct one - I will check if signtool is in the path and in which version but I am afraid that the default one may not always be the correct one.

I have tried to find a way of detecting signtool with vswhere but I could not get anywhere with it - signtool.exe comes from Windows SDK not directly from VS package so maybe it is not possible to locate it with vswhere.

piotr-grabalski: Unfortunately there is no `signtool.exe` in the path, I can either leave it as is (with the search), or hard-code the "Program Files" path - I personally would prefer to give it a little bit of flexibility and let the search sort the path out. Also I could not find anything about getting `signtool` path through `vswere`.

piotr-grabalski:
> > Code looks fine to me, but did you check if this actually fails CI if code signing fails? I see no explicit checks for status codes here, and IIRC PowerShell doesn’t automatically fail on non-zero exit codes (nor does cmd for that matter, reason apparently being that status codes were never standardised on Windows so there are some windows applications that return 1 on _success_ instead).
> 
> Thanks Hannes for your comment, the current situation is as follows:
> Task 1: Decoding Personal Information Exchange file - should never fail - even if there is no secrets the file is successfully created but it's empty.
> Task 2: Signing using signtool - action will fail when wrong/no certificate is used or if password is incorrect (for example when secrets are missing). Action will also fail if given version of signtool.exe cannot be found in the system or if there is more than one signtool that can be used - unlikely since the specific version is selected (10.0.19041.0).
> Task 3: Removing file created in Task 1 - have not seen this one failing.
> Task 4: Signature verification - This can fail only if certificate is no longer valid - have not seen this one failing.
> 
> Please let me know if above is not satisfactory and I will try to add check for specific errors.

A bit more on this: 
Task 3: Remove file created in Task 1 - Fails if file does not exist - I can add `continue-on-error` for this one since a failure at this point is not very relevant - if the file was not created for whatever reason the singning process will fail either way, but if it was and got deleted in the meantime the signing process might have been successful so we should continue to the next job.
Task 4: Signature verification - Fails if wrong, invalid or not-trusted certificate is used

tautschnig:
Would it be possible to use an existing GitHub action: https://github.com/dlemstra/code-sign-action ?

piotr-grabalski:
> Would it be possible to use an existing GitHub action: https://github.com/dlemstra/code-sign-action ?

Hi @tautschnig - thanks for your comment! Long story short - I have tried.

I have tried using an existing GitHub action from https://github.com/DanaBear/code-sign-action (which seems to be evolved version of the action suggested by you) but with no luck - I actually did not try hard after @hannes-steffenhagen-diffblue noticed that the action was reporting a success even on a code sign failure. The action you suggested does not use password anywhere (at least I cannot see password option at the first glance) and that would be a big problem since we are using Personal Information Exchange file (`.pfx`) which requires password to be provided.

hannes-steffenhagen-diffblue (approved):
I was still hoping that there’d be an easier/prettier way to do this, but I’m broadly OK with this.

.github/workflows/release-packages.yaml
tautschnig: While I'd strongly favour a keeping-it-simple approach, could we at bare minimum only do this once? I think https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files should help here.

NlightNFotis (approved):
LGTM

.github/workflows/release-packages.yaml
piotr-grabalski: @tautschnig Thanks for your suggestion! it was not super straight-forward but I have managed to get the workflow simplified a little bit with the Github actions environmental files. 
There is a new job in the workflow: "Setup code sign environment" which is appending the path for `signtool.exe` to `env:GITHUB_PATH` file, and storing PFX certificate file name as "`pfxcert`" variable in `env:GITHUB_ENV` file. Following jobs are now referring to `$env:pfxcert` instead of concatenating the PFX file name every time it is needed and the `signtool.exe` is accessible without the previously required search.

PR #5553, opened 2020-11-10
forall_operands should not trigger any non-const operations

tautschnig:
If .operands() is used in a non-const context, an unnecessary
(temporary) copy of the irep is created. Use as_const to avoid this.


kroening:
It's likely beneficial to remove the `if((expr).has_operands())` line.

The reason why this was added was that `.operands()` might have added a named `operands` sub-node, which no longer exists.

tautschnig:
> It's likely beneficial to remove the `if((expr).has_operands())` line.
> 

Indeed, thank you!

kroening (approved):

PR #5559, opened 2020-11-11
[CI] Add a linux+clang+make github actions job.

NlightNFotis:
This replaces a codebuild job that had not been ported yet,
so that codebuild can be safely deactivated as a CI provider,
having all the configurations ported so far.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -59,6 +59,45 @@ jobs:
           env PATH=$PATH:`pwd`/src/solvers make -C regression/cbmc test-cprover-smt2
           make -C jbmc/regression test-parallel JOBS=2
 
+  check-ubuntu-20_04-make-clang:
+    runs-on: ubuntu-20.04
+    env:
+      CC: "/usr/bin/clang"
+      CXX: "/usr/bin/clang++"
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Fetch dependencies
+        env:
+          # This is needed in addition to -yq to prevent apt-get from asking for
+          # user input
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-checks.yaml
```diff
@@ -40,6 +40,42 @@ jobs:
...
+        with:
+          submodules: true
+      - name: Fetch dependencies
+        env:
+          # This is needed in addition to -yq to prevent apt-get from asking for
+          # user input
+          DEBIAN_FRONTEND: noninteractive
+          TESTPL_JOBS: 4
+        run: |
+          sudo apt-get install -yq gcc gdb g++ maven jq flex bison libxml2-utils cpanminus
```
thomasspriggs: :grey_question: Do we need `gcc gdb g++` for a `clang` build?

thomasspriggs (approved):
:+1:

.github/workflows/pull-request-checks.yaml
NlightNFotis: I thought about it - I opted for keeping the job's environment in line with the `linux+make+gcc` one.

Though these are arguably not very useful given that they are preinstalled in Github runners. But I suggest this stays for now and we have a separate discussion on whether we should remove them from the other build as well.

piotr-grabalski (approved):
looks good to me

tautschnig:
Please also make sure we port all the features of the CodeBuild jobs. Specifically, the "cbmc" CodeBuild job is running with `_GLIBCXX_DEBUG`, which we don't seem to be doing in any of the GitHub actions just yet.

~Also, I am wondering about the use of `ccache` to speed up builds?~ (done)

kroening (approved):

PR #5429, opened 2020-07-21
Vsd/new value set/expression transform

hannes-steffenhagen-diffblue:
- [ ] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #5568, opened 2020-11-12
Expression transform

jezhiggins:
This PR picks out Hannes' expression transform code from https://github.com/diffblue/cbmc/pull/5429 so we can apply to the mainline. 

I've reinstated the -Werror flag in CMakeLists.txt and applied clang-format fixes, but otherwise the code is as it comes.

### DIFF
--- a/src/analyses/variable-sensitivity/value_set_abstract_value.cpp
+++ b/src/analyses/variable-sensitivity/value_set_abstract_value.cpp
@@ -9,6 +9,7 @@
 #include "value_set_abstract_value.h"
 
 #include <ansi-c/expr2c.h>
+#include <util/simplify_expr.h>
 
 value_set_abstract_valuet::value_set_abstract_valuet(
   const typet &type,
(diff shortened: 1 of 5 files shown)


martin-cs:
Paging @hannes-steffenhagen-diffblue : hopefully you don't mind us picking up this PR?

martin-cs (approved):
Looks reasonable.  If you could squash the reverted change to the CMake file back into the relevant commit so that it doesn't touch the file at all that would be nice, otherwise this is OK.

For other reviewers : one of @jezhiggins 's next tasks is refactoring the handling of command-line arguments for VSD so it will be 'connected up' by a later PR.

jezhiggins:
@martin-cs No problem. I'll do the same for the clang-format fixes.

hannes-steffenhagen-diffblue:
Hang on a minute this is just far enough in the past for me to give this a fair review I think.

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.h
```diff
@@ -83,6 +84,8 @@ struct vsd_configt
       options.get_bool_option("interval");
     config.advanced_sensitivities.value_set =
       options.get_bool_option("value-set");
+    config.advanced_sensitivities.new_value_set =
```
hannes-steffenhagen-diffblue, line 87: This `new_value_set` shenanigans probably requires an explanation. The reason here basically is we had an initial implementation of `value_set_abstract_objectt` that just contained a set of `abstract_object_pointert` – and that _looked_ like it worked for a couple of example programs, unfortunately that just gave us set semantics over _pointers_ rather than the actual values in them. It also didn’t have a separate class for pointers/arrays which may require different handling.

At the time we thought it’d be bad to just throw all these changes at once in there (especially for reviewers), so we essentially came up with this approach to add a flag to turn on the new behaviour instead and just rename that to "regular value set" once we’d be done implementing it. Unfortunately we didn’t get done with it before other priorities became more pressing.

unit/analyses/variable-sensitivity/value_set/abstract_value.cpp
```diff
@@ -57,6 +58,7 @@ class ContainsAllOf
       {
         oss << ", ";
       }
+      first = false;
```
hannes-steffenhagen-diffblue, line 61: That was just a bug in the original implementation, ouch

unit/analyses/variable-sensitivity/value_set/abstract_value.cpp
```diff
@@ -359,3 +361,117 @@ TEST_CASE(
 
   REQUIRE(value_set.to_constant() == value);
 }
+
+static abstract_environmentt get_value_set_abstract_environment()
+{
```
hannes-steffenhagen-diffblue, line 366: One thing I really like about variable sensitivity that at least in principle most things here are much easier to test in isolation than other parts of the codebase.

src/analyses/variable-sensitivity/value_set_abstract_value.cpp
```diff
@@ -125,3 +126,92 @@ void value_set_abstract_valuet::output(
...
+    if(expr_with_evaluated_operands_filled_in.is_constant())
+    {
+      bool out_modifications;
+      auto post_merge = abstract_objectt::merge(
+        out_value,
+        std::make_shared<value_set_abstract_valuet>(
+          expr.type(),
+          value_set_abstract_valuet::valuest{
+            expr_with_evaluated_operands_filled_in}),
+        out_modifications);
```
hannes-steffenhagen-diffblue, line 151: For very large expressions there might be some value in exiting early here when value sets get too large and just become TOP. Not super necessary right now I don’t think, but may be worth keeping in mind if it ever turns out we’re spending a ton of time here.

hannes-steffenhagen-diffblue (approved):
> For other reviewers : one of @jezhiggins 's next tasks is refactoring the handling of command-line arguments for VSD so it will be 'connected up' by a later PR.

Yeah, this is probably necessary. But clearly doesn’t need to happen here.

unit/analyses/variable-sensitivity/value_set/abstract_value.cpp
```diff
@@ -359,3 +361,117 @@ TEST_CASE(
...
+  namespacet ns{symbol_tablet{}};
+  const auto s32_type = signedbv_typet{32};
+  const auto three = from_integer(3, s32_type);
+  const auto five = from_integer(5, s32_type);
+  const auto three_plus_five = plus_exprt{three, five};
+
+  auto eval_result = environment.eval(three_plus_five, ns);
+  REQUIRE(eval_result != nullptr);
+  REQUIRE(!eval_result->is_top());
+  REQUIRE(!eval_result->is_bottom());
```
hannes-steffenhagen-diffblue, line 409: This is failing apparently? Strange, I could swear I remember this working.

jezhiggins: I'm investigate this now. Back soon :)

jezhiggins: They pass on your original branch. I'm trying to investigate further ...

jezhiggins: Change in the abstract_environment default constructor 

This branch has 
`abstract_environmentt::abstract_environmentt() : bottom(true) { }`
while the branch this change was originally targetting has 
`abstract_environmentt::abstract_environmentt() : bottom(false) { }`

jezhiggins: Comes from this commit https://github.com/diffblue/cbmc/commit/d9cc8ac3b4a9a54fa643e5cd7724dd5cc1846d13

jezhiggins: If I patch the abstract_value.cpp unit test so instead of the default environment, it explicitly makes it top 

```
  auto environment = abstract_environmentt{};
  environment.make_top();
  return environment;
```

then the unit tests pass. @hannes-steffenhagen-diffblue Is this ok with you? If so, I'll fixup that last commit, and we should be good.

hannes-steffenhagen-diffblue: Sounds reasonable to me.

PR #5569, opened 2020-11-12
Update cbmc config to version 5.18.0.

NlightNFotis:

### DIFF
--- a/src/config.inc
+++ b/src/config.inc
@@ -76,7 +76,7 @@ endif
 OSX_IDENTITY="Developer ID Application: Daniel Kroening"
 
 # Detailed version information
-CBMC_VERSION = 5.17.0
+CBMC_VERSION = 5.18.0
 
 # Use the CUDD library for BDDs, can be installed using `make -C src cudd-download`
 # CUDD = ../../cudd-3.0.0


hannes-steffenhagen-diffblue (approved):

thomasspriggs (approved):

tautschnig:
~Can someone explain to me why the codecov/project job is failing here (well, actually I'm seeing it failing on all other PRs as well, unless I'm mistaken)?~  Edit: seemingly that's not longer failing now. Whatever the magic.

tautschnig (approved):
It would be nice if the release commit could at least include some statement of why marking this particular state of develop as the next release is the right thing. If it's "It's Thursday, so we release" then that's ok as well, but right now it feels entirely random.

NlightNFotis:
Hi @tautschnig. The regular release process is indeed every two weeks - it's documented in https://github.com/diffblue/cbmc/blob/develop/doc/ADR/release_process.md.

We felt that every two weeks was a good compromise between a release cadence that's often enough to make sure we ship new features, but it's not very tight (for example every week) so that we come out with releases that don't contain new features or bug releases. Do you think this information should go somewhere else?

tautschnig:
> Hi @tautschnig. The regular release process is indeed every two weeks - it's documented in https://github.com/diffblue/cbmc/blob/develop/doc/ADR/release_process.md.
> 
> We felt that every two weeks was a good compromise between a release cadence that's often enough to make sure we ship new features, but it's not very tight (for example every week) so that we come out with releases that don't contain new features or bug releases. Do you think this information should go somewhere else?

I think that's perfectly good, but it would be nice if doing `git log` could in future also provide some insight, for example by not having an empty body of the commit message :-) Really, "Bi-weekly release" could even do. But it's not a big deal, just nice to have.

PR #1086, opened 2017-07-04
Make pointer encoding sound wrt integer address translation (and lift object limits)

tautschnig:
Instead of encoding pointers with the same bit width a pointer on a given
platform has, just widen the bit-blasted encoding to include both the previous
object/offset encoding as well as an (integer) address. The encoding is thus
also trivially be extended to handle larger numbers of objects and offsets of
the same width as the address.

Furthermore clean up the code to encapsulate encoding properly, and make in-code
layout of pointer encoding more natural (it's now object, offset,
integer-address).

Fixes: #436
Fixes: #311
Fixes: #94

PR #5572, opened 2020-11-12
Synchronise goto-cc command-line options to re-enable Windows regression tests

tautschnig:
object-bits was not supported by all of goto-cc's command-line front-end
variants. While at it, also make other options consistently available.

### DIFF
--- a/regression/goto-cc-cbmc-shared-options/CMakeLists.txt
+++ b/regression/goto-cc-cbmc-shared-options/CMakeLists.txt
@@ -1,11 +1,9 @@
 if(WIN32)
     set(is_windows true)
-    set(exclude_broken_windows_tests -X winbug)
 else()
     set(is_windows false)
-    set(exclude_broken_windows_tests "")
 endif()
 
 add_test_pl_tests(
-    "${CMAKE_CURRENT_SOURCE_DIR}/chain.sh $<TARGET_FILE:goto-cc> $<TARGET_FILE:cbmc> ${is_windows}" ${exclude_broken_windows_tests}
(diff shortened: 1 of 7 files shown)


peterschrammel (approved):
I don't quite understand/remember why object-bits is required at all here. It should only have an influence on symbolic execution, but not appear in a goto-program. Going forward this should probably be cleaned up.

kroening:
Agreed, goto-cc (or goto-gcc or goto-cl) should not have analysis options.

tautschnig:
@peterschrammel said:
> I don't quite understand/remember why object-bits is required at all here. It should only have an influence on symbolic execution, but not appear in a goto-program. Going forward this should probably be cleaned up.

@kroening said:
> Agreed, goto-cc (or goto-gcc or goto-cl) should not have analysis options.

As of 531b856ace714d6f63c9d7971aafb07be7a97b06 (#5210) the C front-end needs to determine the value of `__CPROVER_max_malloc_size` based on this configuration information. Would be unnecessary with #1086, but that's also quite far off.

kroening:
`__CPROVER_max_malloc_size` can just as well be set by the analyser.

martin-cs (approved):
In principle, yes, but like others I have concerns about `--object-bits`.

tautschnig:
> `__CPROVER_max_malloc_size` can just as well be set by the analyser.

@kroening Any suggestions of what would be the cleanest way to do so? We could certainly generate an assignment in `goto_symext::initialize_entry_point_state`, but that does feel like a hack.

Edit: another option would be to introduce a function `__CPROVER_max_malloc_size()` and have `goto-symex` give special treatment to the function call. That, however, would require some awareness of return-value removal.

tautschnig:
Closing in favour of #7858.

PR #5573, opened 2020-11-12
Fix PATH separator and CMake syntax to make cbmc-cprover-smt2 tests work on Windows

tautschnig:
Windows separates PATH components by semicolons. add_test_pl_profile does not take more than 4 arguments.

### DIFF
--- a/regression/cbmc/CMakeLists.txt
+++ b/regression/cbmc/CMakeLists.txt
@@ -1,41 +1,40 @@
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
   set(gcc_only -X gcc-only)
+  set(gcc_only_string "-X;gcc-only;")
 else()
   set(gcc_only "")
-endif()
-
-if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
-  set(exclude_win_broken_tests -X winbug)
-else()
-  set(exclude_win_broken_tests "")
+  set(gcc_only_string "")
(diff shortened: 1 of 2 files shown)


regression/cbmc/CMakeLists.txt
```diff
@@ -1,41 +1,40 @@
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
   set(gcc_only -X gcc-only)
+  set(gcc_only_string ";-X;gcc-only")
```
NlightNFotis: I have two questions about this PR:

1. Why is this being added? I can see that this is moved to be back into the quotes part in line 16, but I don't understand why this being done.
2. I hope I'm not being overly nitpicky, but this string reads a little bit weird to me. May I suggest that this is turned into something like `-X;gcc-only;` so that it looks more naturally delimited here and in line 16? Unless there's a specific reason for it to be like this.

NlightNFotis (approved):
Looks good, but I would also like @hannes-steffenhagen-diffblue to take a look since he's more familiar with cmake than I am.

regression/cbmc/CMakeLists.txt
tautschnig: Re 1) The macro `add_test_pl_profile` takes exactly four arguments when previously six were passed (an additional "-X" and "gcc-only"). So `gcc_only` was simply ignored.
Re 2) Fair point, I've fixed this.

PR #5578, opened 2020-11-13
Make goto-harness-multi-file-project regression test work on Windows

tautschnig:
The test script assumed goto-cc command-line option syntax, which
wouldn't work with goto-cl in Windows.

### DIFF
--- a/regression/goto-harness-multi-file-project/chain.sh
+++ b/regression/goto-harness-multi-file-project/chain.sh
@@ -5,7 +5,8 @@ set -e
 goto_cc="$1"
 goto_harness="$2"
 cbmc="$3"
-goto_harness_args="${@:4:$#-4}"
+is_windows=$4
+goto_harness_args="${@:5:$#-5}"
 
 cleanup()
 {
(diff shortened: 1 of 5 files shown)


martin-cs (approved):

regression/goto-harness-multi-file-project/chain.sh
```diff
@@ -5,21 +5,30 @@ set -e
...
+    "$goto_cc" -c "$file" -o "$target_dir"/"${file_name%.c}.o"
+  fi
 done
 
 # Linking
 
 main_exe="$target_dir/main.gb"
-"$goto_cc" "$target_dir"/*.o -o "$main_exe"
+if [[ "${is_windows}" == "true" ]]; then
+  "$goto_cc" "$target_dir"/*.obj "/Fe$main_exe"
```
NlightNFotis, line 37: This seems to be tripping up the windows build with the following message:

```
2020-11-13T12:52:04.8110199Z 107: + D:/a/cbmc/cbmc/build/bin/Release/goto-cl.exe '/tmp/tmp.aEaZLyQI0l/*.obj' /Fe/tmp/tmp.aEaZLyQI0l/main.gb
2020-11-13T12:52:04.8121175Z 107: Warning: uninterpreted CL option '/tmp/tmp.aEaZLyQI0l/*.obj'
2020-11-13T12:52:04.8167232Z 107: no input files
```

I'm not sure why it falters now though, given that the syntax is very similar to what was there before.

NlightNFotis (approved):
Hi Michael,

Thanks for reactivating the tests on windows. That was much needed.

This should take one small fix and should be good to go. Unfortunately, I'm not very familiar with either windows scripting or the windows toolchain itself, so I can't make any further suggestions.

regression/goto-harness-multi-file-project/chain.sh
tautschnig: I believe the problem is that using quotes (intentionally) avoids pathname conversion, i.e., replacing the Unix-style path by a Windows path (something like "C:\..."). The Unix-style path (file names starting with "/"), however, confuses the command-line parser of goto-cl, which treats anything following a "/" as a command-line parameter. I've pushed a new variant, hoping this will now work.

PR #5580, opened 2020-11-13
Use single quotes to make regression tests work on Windows

tautschnig:
test.pl cannot cope with double quotes on some Windows-based platforms
(including GitHub's Windows environment).

### DIFF
--- a/jbmc/regression/jbmc-strings/CMakeLists.txt
+++ b/jbmc/regression/jbmc-strings/CMakeLists.txt
@@ -1,20 +1,10 @@
-if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
-  set(exclude_win_broken_tests -X winbug)
-else()
-  set(exclude_win_broken_tests "")
-endif()
-
 add_test_pl_tests(
-    "$<TARGET_FILE:jbmc> --validate-goto-model --validate-ssa-equation" ${exclude_win_broken_tests}
+    "$<TARGET_FILE:jbmc> --validate-goto-model --validate-ssa-equation"
 )
 
-if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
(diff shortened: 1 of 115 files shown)


martin-cs (approved):
Changes as described.  I will take your word that they are necessary.

jbmc/regression/jbmc-json-ui/pointer-simplification/test.desc
```diff
@@ -1,6 +1,6 @@
 CORE
 test
-'.[] | select(has("result")) | .result | .[] | select(has("trace")) | .trace | .[] | select(has("value")) | .value | select(.name == "pointer") | {has_data: has("data")}' --json-ui --trace
+'.[] | select(has('result')) | .result | .[] | select(has('trace')) | .trace | .[] | select(has('value')) | .value | select(.name == 'pointer') | {has_data: has('data')}' --json-ui --trace
```
NlightNFotis: This seems to be tripping up all the UNIX builds now (ubuntu and osx). 

Is there a way perhaps to have two of these commands, and dispatch based on OS detection at test runtime? I'm not sure we have one, but maybe it wouldn't be hard to come up with one?

If that's not possible/convenient for the time being, maybe this change could be reverted and the test marked as `KNOWNBUG` on windows?

NlightNFotis (approved):
Hi Michael,

Again many thanks for reactivating the tests on windows. 

Good to go, just one problem with the `jbmc-json-ui/pointer-simplification` test passing the wrong syntax to `jq` on UNIX systems, causing `jq` to throw syntax errors.

jbmc/regression/jbmc-json-ui/pointer-simplification/test.desc
tautschnig: Thank you for digging into this one, I was quite confused.

PR #5581, opened 2020-11-13
Update subclasses of ai_domain_baset to deprecate function

martin-cs:
transform's signature changed during the addition of
ai_historyt.  This is removing the old signature in favour
of the newer, more flexible one.

### DIFF
--- a/src/analyses/constant_propagator.cpp
+++ b/src/analyses/constant_propagator.cpp
@@ -122,12 +122,15 @@ void constant_propagator_domaint::assign_rec(
 
 void constant_propagator_domaint::transform(
   const irep_idt &function_from,
-  locationt from,
+  trace_ptrt trace_from,
   const irep_idt &function_to,
-  locationt to,
+  trace_ptrt trace_to,
   ai_baset &ai,
   const namespacet &ns)
 {
+  locationt from{trace_from->current_location()};
+  locationt to{trace_to->current_location()};
+
 #ifdef DEBUG
   std::cout << "Transform from/to:\n";
   std::cout << from->location_number << " --> "
--- a/src/analyses/variable-sensitivity/three_way_merge_abstract_interpreter.cpp
+++ b/src/analyses/variable-sensitivity/three_way_merge_abstract_interpreter.cpp
@@ -125,9 +125,9 @@ bool ai_three_way_merget::visit_edge_function_call(
     // but in VSD it does, so this cannot be omitted.
     s_working.transform(
       callee_function_id,
-      l_callee_end,
+      p_callee_end,
       calling_function_id,
-      l_return_site,
+      p_return_site,
       *this,
       ns);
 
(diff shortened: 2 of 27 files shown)


martin-cs:
Related to https://github.com/diffblue/cbmc/pull/5576

martin-cs:
My thanks to @smowton for `unit/analyses/ai/ai.cpp` which is a nice unit test.

smowton:
sfx: ghostly, though appreciative, rattling of chains

martin-cs:
The one CI failure appears to be a code coverage failure `69.31% (+-0.01%) compared to aa069d3 ` which I think is because I have deleted a method.  But apparently `88.88% of diff hit (target 69.31%)`.  Weird.

src/analyses/constant_propagator.cpp
```diff
@@ -122,12 +122,15 @@ void constant_propagator_domaint::assign_rec(
...
   const irep_idt &function_from,
-  locationt from,
+  trace_ptrt trace_from,
   const irep_idt &function_to,
-  locationt to,
+  trace_ptrt trace_to,
   ai_baset &ai,
   const namespacet &ns)
 {
+  locationt from(trace_from->current_location());
```
tautschnig: Nit picking: as stated in `CODING_STANDARD.md`, braces are preferred over parentheses for constructor calls.

src/analyses/variable-sensitivity/three_way_merge_abstract_interpreter.cpp
```diff
@@ -125,9 +125,9 @@ bool ai_three_way_merget::visit_edge_function_call(
     // but in VSD it does, so this cannot be omitted.
     s_working.transform(
       callee_function_id,
-      l_callee_end,
+      p_callee_end,
       calling_function_id,
-      l_return_site,
+      p_return_site,
```
tautschnig, line 130: Is "p" short for "trace pointer?" I think "l" I could see as "location," but this one seemed less clear. Saying this while I'm aware that this isn't really related to this PR.

tautschnig (approved):

src/analyses/constant_propagator.cpp
martin-cs: OK; will fix.

src/analyses/variable-sensitivity/three_way_merge_abstract_interpreter.cpp
martin-cs: Yes and maybe `t_` would be better.  Having similar names makes it much more readable IMHO.

martin-cs:
Also paging @peterschrammel w.r.t. 2LS and @chrisr-diffblue (?) w.r.t. Diffblue's use of goto-analyzer.  This clean up and the ones to follow will break any out-of-tree patches that have their own abstract domains.

chrisr-diffblue (approved):
Looks fine to me - I don't personally have out-of-tree patches that would depend on this now. @hannes-steffenhagen-diffblue Anything that you might have?

peterschrammel (approved):
LGTM

src/analyses/constant_propagator.cpp
martin-cs: Fixed.

martin-cs:
Will merge when CI passes.

Issue #5531, opened 2020-10-15
Improve error message for pointer primitive check


feliperodri:
CBMC version: 5.15.0 (cbmc-5.14.3-16-ga0d09f8e8-dirty)
Operating system: macOS Mojave 10.14.6
Exact command line resulting in the issue: `cbmc main.c --pointer-primitive-check --pointer-check --trace`
What behaviour did you expect: An error message that suggest which part of the conjunction leads to failure.
What happened instead:

```c
#include <stdlib.h>

int main() {
  size_t size = 10;
  void* mem = malloc(size);
  assert(__CPROVER_r_ok(mem, size));
  assert(__CPROVER_r_ok(mem, 0));
  size_t offset;
  __CPROVER_assume(offset <= size);
  void* ptr = mem + offset;
  __CPROVER_r_ok(ptr, 0);
}
```

```
[main.pointer_primitives.2] line 11 pointer in pointer primitive is neither null nor valid in R_OK(ptr, (__CPROVER_size_t)0): FAILURE

Trace for main.pointer_primitives.2:

State 26 file main.c function main line 4 thread 0
----------------------------------------------------
  size=0ul (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 27 file main.c function main line 4 thread 0
----------------------------------------------------
  size=10ul (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00001010)

State 28 file main.c function main line 5 thread 0
----------------------------------------------------
  mem=NULL (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 31 file main.c function main line 5 thread 0
----------------------------------------------------
  malloc_size=10ul (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00001010)

State 60 file main.c function main line 5 thread 0
----------------------------------------------------
  mem=(const void *)dynamic_object1 (00000010 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 61 file main.c function main line 8 thread 0
----------------------------------------------------
  offset=10ul (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00001010)

Assumption:
  file main.c line 9 function main
  offset <= size

State 63 file main.c function main line 10 thread 0
----------------------------------------------------
  ptr=NULL (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 64 file main.c function main line 10 thread 0
----------------------------------------------------
[...]

PR #5582, opened 2020-11-13
Usability/split up pointer primitive checks

hannes-steffenhagen-diffblue:
The pointer-primitive-checks require an address check. Previous we’ve just done these all in in one go, but this makes it hard to see _which_ of our address checks failed.

See https://github.com/diffblue/cbmc/issues/5531

### DIFF
--- a/regression/cbmc/pointer-primitive-check-01/test.desc
+++ b/regression/cbmc/pointer-primitive-check-01/test.desc
@@ -1,15 +1,43 @@
-CORE
+CORE broken-smt-backend
 main.c
 --pointer-primitive-check
 ^EXIT=10$
 ^SIGNAL=0$
-\[main.pointer_primitives.1\] line \d+ pointer in pointer primitive is neither null nor valid in POINTER_OBJECT.*: FAILURE
-\[main.pointer_primitives.2\] line \d+ pointer in pointer primitive is neither null nor valid in POINTER_OFFSET.*: FAILURE
-\[main.pointer_primitives.3\] line \d+ pointer in pointer primitive is neither null nor valid in POINTER_OBJECT.*: FAILURE
-\[main.pointer_primitives.4\] line \d+ pointer in pointer primitive is neither null nor valid in OBJECT_SIZE.*: FAILURE
-\[main.pointer_primitives.5\] line \d+ pointer in pointer primitive is neither null nor valid in R_OK.*: FAILURE
-\[main.pointer_primitives.6\] line \d+ pointer in pointer primitive is neither null nor valid in W_OK.*: FAILURE
-\*\* 7 of \d+ failed
+\[main.pointer_primitives.1\] line \d+ pointer invalid in POINTER_OBJECT\(\(const void \*\)p1\): FAILURE
+\[main.pointer_primitives.2\] line \d+ deallocated dynamic object in POINTER_OBJECT\(\(const void \*\)p1\): SUCCESS
+\[main.pointer_primitives.3\] line \d+ dead object in POINTER_OBJECT\(\(const void \*\)p1\): SUCCESS
+\[main.pointer_primitives.4\] line \d+ pointer outside dynamic object bounds in POINTER_OBJECT\(\(const void \*\)p1\): FAILURE
+\[main.pointer_primitives.5\] line \d+ pointer outside object bounds in POINTER_OBJECT\(\(const void \*\)p1\): FAILURE
+\[main.pointer_primitives.6\] line \d+ pointer invalid in POINTER_OFFSET\(\(const void \*\)p2\): FAILURE
+\[main.pointer_primitives.7\] line \d+ deallocated dynamic object in POINTER_OFFSET\(\(const void \*\)p2\): SUCCESS
+\[main.pointer_primitives.8\] line \d+ dead object in POINTER_OFFSET\(\(const void \*\)p2\): SUCCESS
+\[main.pointer_primitives.9\] line \d+ pointer outside dynamic object bounds in POINTER_OFFSET\(\(const void \*\)p2\): FAILURE
+\[main.pointer_primitives.10\] line \d+ pointer outside object bounds in POINTER_OFFSET\(\(const void \*\)p2\): FAILURE
+\[main.pointer_primitives.11\] line \d+ pointer invalid in POINTER_OBJECT\(\(const void \*\)p3\): FAILURE
+\[main.pointer_primitives.12\] line \d+ deallocated dynamic object in POINTER_OBJECT\(\(const void \*\)p3\): SUCCESS
+\[main.pointer_primitives.13\] line \d+ dead object in POINTER_OBJECT\(\(const void \*\)p3\): SUCCESS
+\[main.pointer_primitives.14\] line \d+ pointer outside dynamic object bounds in POINTER_OBJECT\(\(const void \*\)p3\): FAILURE
+\[main.pointer_primitives.15\] line \d+ pointer outside object bounds in POINTER_OBJECT\(\(const void \*\)p3\): FAILURE
+\[main.pointer_primitives.16\] line \d+ pointer invalid in OBJECT_SIZE\(\(const void \*\)p4\): FAILURE
+\[main.pointer_primitives.17\] line \d+ deallocated dynamic object in OBJECT_SIZE\(\(const void \*\)p4\): SUCCESS
+\[main.pointer_primitives.18\] line \d+ dead object in OBJECT_SIZE\(\(const void \*\)p4\): SUCCESS
+\[main.pointer_primitives.19\] line \d+ pointer outside dynamic object bounds in OBJECT_SIZE\(\(const void \*\)p4\): FAILURE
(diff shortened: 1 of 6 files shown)


regression/cbmc/pointer-primitive-check-01/test.desc
```diff
@@ -1,15 +1,43 @@
-CORE
+CORE broken-smt-backend
```
martin-cs, line 1: Why `broken-smt-backend` when it wasn't before?

src/analyses/goto_check.cpp
```diff
@@ -1226,21 +1230,21 @@ void goto_checkt::pointer_primitive_check(
...
-    or_expr,
-    "pointer in pointer primitive is neither null nor valid",
-    "pointer primitives",
-    expr.source_location(),
-    expr,
-    guard);
+  {
+    // we want all address checks *except* for the null check because these
+    // should only fail if the assumption is on in invalid non-null pointer.
+    if(c.description != pointer_is_null_check_description)
```
martin-cs: This seems to have less checks?  Thanks for adding the comment but I think I am struggling to understand it, possibly because I am missing the context.  Please can you explain it a bit more to me?

src/analyses/goto_check.cpp
```diff
@@ -1226,21 +1230,21 @@ void goto_checkt::pointer_primitive_check(
...
-  add_guarded_property(
-    or_expr,
-    "pointer in pointer primitive is neither null nor valid",
-    "pointer primitives",
-    expr.source_location(),
-    expr,
-    guard);
+  {
+    // we want all address checks *except* for the null check because these
+    // should only fail if the assumption is on in invalid non-null pointer.
```
tautschnig: NIt picking: s/on in/on an/

src/analyses/goto_check.cpp
tautschnig: Adding to this: is a string comparison really the best solution? Couldn't we instead compare `c.assertion` to `null_object(pointer)`?

tautschnig (commented):
As said in my comment below, I'm not convinced the string comparison is the right approach. Apart from this: Please also add a "Fixes: #5531" line to the commit message.

regression/cbmc/pointer-primitive-check-01/test.desc
hannes-steffenhagen-diffblue: @martin-cs I am not sure _why_ but some individual pointer checks seem to have different results in the smt backend. Because in the previous version of this we AND-ed them all together this didn’t matter before.

src/analyses/goto_check.cpp
hannes-steffenhagen-diffblue: @tautschnig I don’t like it either, but I don’t like repeating the implementation of `c.assertion` much better. My preferred thing to do would be to add stable IDs to the checks we generate so we can check that, e.g. this is "the address null check", something along those lines. This would be a bigger change though, not sure if people would be OK with that approach?

tautschnig: How about adding further members to `conditiont` that track which kinds of properties are already asserted? We don't have to be complete right away, you could just start with a single one: `bool has_null_check`.

hannes-steffenhagen-diffblue: @tautschnig One option to get this through right now would be to park the idea of how to filter out checks for later and just accept the redundant null-check for now, but that doesn’t make me happy either.

tautschnig: Well, why not try to route of adding, e.g., `has_null_check`?

regression/cbmc/pointer-primitive-check-01/test.desc
martin-cs: @hannes-steffenhagen-diffblue Thanks for the explanation but this does leave me a little concerned.  They really shouldn't do.  This is kind of important functionality.

src/analyses/goto_check.cpp
```diff
@@ -1226,21 +1230,21 @@ void goto_checkt::pointer_primitive_check(
...
+    if(c.description != pointer_is_null_check_description)
+    {
+      add_guarded_property(
+        or_exprt{null_object(pointer), c.assertion},
```
thomasspriggs: :no_entry_sign: This doesn't actually appear to be adding the equivalent checks in the case where `mode == ID_java` or where `flags.is_null()` is false.

There must be a way to implement this which is both more readable and more straight forward to get correct. Besides the issue of the string comparison not being great, I think that filtering the list of conditions after generating them is fundamentally a sub-optimal approach. I would suggest starting by looking at splitting the `goto_checkt::address_check` function into multiple functions. At least one of which should generate the list of conditions without the null check. That way the check never has to be filtered out after being unnecessarily added.

src/analyses/goto_check.cpp
hannes-steffenhagen-diffblue: @tautschnig I don’t completely understand how you’d want that to work – I want to remove the null check, not ask whether it has one.

tautschnig: I'd basically use a `bool` member instead of the string. But then what @thomasspriggs [said](https://github.com/diffblue/cbmc/pull/5582/files#r525275170) seems a lot better.

tautschnig:
@hannes-steffenhagen-diffblue Please let us know when this is ready for review - I like the direction this is going, but 1) some cleanup of the previous approach is needed and 2) tests seem to be failing.

src/analyses/goto_check.cpp
```diff
@@ -39,6 +39,8 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "guard.h"
 #include "local_bitvector_analysis.h"
 
+static constexpr const char *pointer_is_null_check_description = "pointer NULL";
```
thomasspriggs: :no_entry_sign: Adding this when we don't need it is now unnecessary churn. It should be removed so that we are treating all the descriptions in the same way.

src/analyses/goto_check.cpp
```diff
@@ -164,6 +166,8 @@ class goto_checkt
     }
 
     exprt assertion;
+    // pointer-primitive-check assumes that these are fixed. If we ever add I18N
```
thomasspriggs: :question: What do you mean by "fixed"? Is this an assumption that is still being made with the current refactor? Is `pointer_validity_check` making the same assumption?

thomasspriggs: :pick: I am not sure the comment about internationalisation is useful. Surely that would affect strings throughout the code base, not just this one. That also seems like the implementation choices for that are something which would be decided when we get to it rather than something which is a useful comment.

src/analyses/goto_check.cpp
```diff
@@ -2228,6 +2133,134 @@ void goto_checkt::goto_check(
...
+
+  return conditions;
+}
+
+optionalt<goto_checkt::conditiont> goto_checkt::get_pointer_is_null_condition(
+  const exprt &address,
+  const exprt &size)
+{
+  PRECONDITION(local_bitvector_analysis);
+  PRECONDITION(address.type().id() == ID_pointer);
```
thomasspriggs, line 2224: :bulb: You could just do `const auto &pointer_type = type_checked_cast<pointer_typet>(address.type());` rather than duplicating the same check inside `to_pointer_type` and taking 2 lines to do this.

src/analyses/goto_check.cpp
```diff
@@ -202,7 +206,18 @@ class goto_checkt
   /// \return true if the given expression is a pointer primitive
   bool is_pointer_primitive(const exprt &expr);
 
-  conditionst address_check(const exprt &address, const exprt &size);
+  optionalt<goto_checkt::conditiont>
+  get_pointer_is_null_condition(const exprt &address, const exprt &size);
```
thomasspriggs, line 206: :pick: It would be nice to have some doxygen for these functions.

thomasspriggs (commented):
This is looking better. I like that we no longer need the filtering. :+1: . However I think this needs to be split into more commits before merging (and maybe before fixing the tests). This is because we are now doing multiple distinct different things within the same commit. Keeping them in separate commits would help with both debugging the tests right now and making the commit history make sense which could help with tracking down future issues. I think I would make the commit history look something like this -

1. Reorder existing checks in `address_check`. As in - put the null check first only. The tests can then be updated for this functionality only which will help with testing the next commit.
2. Refactor to split the `address_check` function. This commit should include no functionality changes. That way we can use the tests from commit 1 to test that this refactor is sound.
3. Separate calls to `add_guarded_property` in `pointer_primitive_check`. Leave the null condition in place.
4. Remove the null condition in `pointer_primitive_check`. I am assuming you did this removal in order to avoid adding a condition which would otherwise be checking "x is null or x is not null". However the commit history does not currently make this clear and putting this in a separate commit will allow for putting the motivation in the commit message body.

I am a little on the fence about the ordering 3/4.

thomasspriggs (approved):
Thank you for putting the work into refactoring this and cleaning up the commit history. This version was _much_ easier to review and this implementation should be easier to maintain.

chrisr-diffblue (approved):
Generally looks good, thanks - I'd recommend getting a re-review from @tautschnig too, just to check he's happy with the updates.

hannes-steffenhagen-diffblue:
@tautschnig are you OK with this now?

PR #5595, opened 2020-11-15
interpretert isn't a messaget

tautschnig:
Use a messaget member instead as there is no is-a relationship between
interpretert and messaget.

### DIFF
--- a/src/goto-programs/interpreter_class.h
+++ b/src/goto-programs/interpreter_class.h
@@ -24,14 +24,14 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "goto_trace.h"
 #include "json_goto_trace.h"
 
-class interpretert:public messaget
+class interpretert
 {
 public:
   interpretert(
     const symbol_tablet &_symbol_table,
     const goto_functionst &_goto_functions,
     message_handlert &_message_handler)
-    : messaget(_message_handler),
+    : output(_message_handler),
       symbol_table(_symbol_table),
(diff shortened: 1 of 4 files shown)


src/goto-programs/interpreter_class.h
```diff
@@ -24,14 +24,14 @@ Author: Daniel Kroening, kroening@kroening.com
...
-class interpretert:public messaget
+class interpretert
 {
 public:
   interpretert(
     const symbol_tablet &_symbol_table,
     const goto_functionst &_goto_functions,
     message_handlert &_message_handler)
-    : messaget(_message_handler),
+    : output(_message_handler),
```
hannes-steffenhagen-diffblue, line 34: Why is this called `output` rather than `log` here?

hannes-steffenhagen-diffblue (approved):
Naming seems inconsistent, but otherwise OK

hannes-steffenhagen-diffblue:
GH Actions failed due to network error, restarting (all of them because apparently there's no way to restart only a single one).

src/goto-programs/interpreter_class.h
tautschnig: The interpreter necessarily provides user-facing output, i.e., is not useful if not printing some of the messages. So I thought "output" may be a better choice than "log" on this occasion. But I might well be wrong?

peterschrammel (approved):
Fine with output or log.

PR #5599, opened 2020-11-15
Clean up ansi_c_convert_typet's interface

tautschnig:
The need to invoke read() was only specified in a comment. Make the
constructor take care of this. Also, make the constructor initialise all
members rather than leaving this to clear(), which is thus no longer
necessary. Finally, do not unnecessarily inherit from messaget (there is
no is-a relationship).


peterschrammel (approved):
Sounds like there should have been 3 commits, but ok.

tautschnig:
> Sounds like there should have been 3 commits, but ok.

I turned it into 2 commits, breaking up the smaller one of the two turned out to be rather cumbersome and it wasn't really clear anymore where to put which change.

PR #5604, opened 2020-11-16
Fix float to int conversion bugs in JBMC

peterschrammel:


romainbrenguier (approved):

tautschnig:
@peterschrammel Needs a rebase to make CI pass.

jbmc/regression/jbmc/FloatToIntConversion/Test.java
```diff
@@ -0,0 +1,220 @@
...
+        int result = (int)(a / b);
+        assert result == -1;
+      }
+      break;
+    case 2: 
+      if (a == 1.0f) {
+        // use constant here, -0.0f reverses cases 2 and 3.
+        int result = (int)(a / 0.0f);
+        // a/b = +inf
+        assert result == 2147483647;
```
martin-cs, line 24: I'm going to trust you that this is the JVM semantics but ... I am surprised by this.

jbmc/regression/jbmc/FloatToIntConversion/Test.java
```diff
@@ -0,0 +1,220 @@
...
+      break;
+    case 3: 
+      if (a == -1.0f) {
+        int result = (int)(a / 0.0f);
+        // a/b = -inf
+        assert result == -2147483648;
+      }
+      break;
+    case 4: 
+      if (a == 2147483646.0f && b == 1.0f) {
```
martin-cs: There is quite a lot going on here and I feel like, without the comment it is easy to miss some of it.  Here is my understanding:

`2147483646` is `0x7FFFFFFE` which has 30 significant bits so `2147483646.0f` will round /in the compiler/ to `0x1p+31f` a.k.a. 2^31 which is the same as `2147483648.0f`.  The divide is effectively a no-op and so this tests the conversion to `int` but also the compiler's handling of constants.

( There are some really horrific examples like this where you can write code that will, at run-time, tell you whether the compiler correctly parsed the number as a float or whether it parsed it first as a double and cast down. )

jbmc/regression/jbmc/FloatToIntConversion/Test.java
```diff
@@ -0,0 +1,220 @@
...
+      break;
+    case 3: 
+      if (a == -1.0f) {
+        long result = (long)(a / 0.0f);
+        // a/b = -inf
+        assert result == -9223372036854775808L;
+      }
+      break;
+    case 4: 
+      if (a == 2147483649.0f && b == 1.0f) {
```
martin-cs: A similar thing happens here.

martin-cs (approved):
Looks good.  Some more comments on some of the rounding effects would be helpful IMHO.

jbmc/regression/jbmc/FloatToIntConversion/Test.java
peterschrammel: I was also surprised until I ran it through the JVM.

jbmc/regression/jbmc/FloatToIntConversion/Test.java
peterschrammel: There are quite a lot of subtle things going on. I can comment in more detail what is happening in the test, and maybe choose constants that are not going to be rounded. I should probably also explain the effect of +1 (resp. -1) in java_bytecode_convert_method.cpp:1681, which are subject to similar rounding effects (i.e. it may not be totally obvious why this works).

martin-cs: With this kind of thing, more documentation is better than less.  I know you get it and intend what happens but it should be commented so that others can learn that.

jbmc/regression/jbmc/FloatToIntConversion/Test.java
peterschrammel: @martin-cs, improved.

jbmc/regression/jbmc/FloatToIntConversion/Test.java
martin-cs: Cool.

jbmc/regression/jbmc/FloatToIntConversion/Test.java
martin-cs: Thanks.

PR #5606, opened 2020-11-17
Upgrade setup-msbuild workflow to 1.0.2 to avoid deprecated add-path

tautschnig:
The add-path command has now been disabled by GitHub for security
reasons. setup-msbuild fixed the problem on their end in
https://github.com/microsoft/setup-msbuild/pull/25, which hopefully is
part of their 1.0.2 release.

### DIFF
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -110,7 +110,7 @@ jobs:
       - name: Fetch dependencies
         run: |
           choco install winflexbison3
-      - uses: microsoft/setup-msbuild@v1.0.1
+      - uses: microsoft/setup-msbuild@v1.0.2
         name: Setup Visual Studio environment
       - name: Setup code sign environment
         run: | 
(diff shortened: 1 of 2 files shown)


.github/workflows/release-packages.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
       - name: Get release tag name
         # The GITHUB_REF we get has refs/tags/ in front of the tag name so we
         # strip that here
-        run: echo "::set-env name=RELEASE_TAG::${GITHUB_REF/refs\/tags\/}"
+        run: echo "RELEASE_TAG=${GITHUB_REF/refs\/tags\/}" >> $GITHUB_ENV
```
piotr-grabalski, line 94: I think I have tried to refer to Github environmental variables file through `$GITHUB_ENV` and I was unsuccessful, does it work for you? I had to use `$env:GITHUB_ENV` instead of `$GITHUB_ENV`.

tautschnig: Oh, that just depends on whether that's Windows or Unix, I believe. Powershell requires `$env:VARIABLE_NAME`, while bash should work this way.

NlightNFotis (approved):
LGTM

PR #5608, opened 2020-11-17
Use both cores of GitHub runners

tautschnig:
According to
https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources,
all GitHub runners feature two CPU cores. Let's use them, as we used to
do in Travis and CodeBuild configurations.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -33,10 +33,10 @@ jobs:
           make TAGS="[!shouldfail]" -C jbmc/unit test
       - name: Run regression tests
         run: |
-          make -C regression test
+          make -C regression test-parallel JOBS=2
           make -C regression/cbmc test-paths-lifo
           env PATH=$PATH:`pwd`/src/solvers make -C regression/cbmc test-cprover-smt2
-          make -C jbmc/regression test
+          make -C jbmc/regression test-parallel JOBS=2
 
   check-ubuntu-20_04-cmake-gcc:
     runs-on: ubuntu-20.04
@@ -57,14 +57,14 @@ jobs:
           cd build
           cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++
       - name: Build with Ninja
-        run: cd build; ninja
+        run: cd build; ninja -j2
       - name: Check if package building works
         run: |
           cd build
           ninja package
           ls *.deb
       - name: Run tests
-        run: cd build; ctest . -V -L CORE
+        run: cd build; ctest . -V -L CORE -j2
 
   check-macos-10_15-make-clang:
     runs-on: macos-10.15
@@ -73,12 +73,12 @@ jobs:
         with:
           submodules: true
       - name: Fetch dependencies
-        run: brew install maven flex bison
+        run: brew install maven flex bison parallel
       - name: Build using Make
         run: |
           make -C src minisat2-download
-          make -C src
-          make -C jbmc/src
+          make -C src -j2
+          make -C jbmc/src -j2
           make -C unit
           make -C jbmc/unit
       - name: Run unit tests
@@ -86,9 +86,9 @@ jobs:
       - name: Run JBMC unit tests
         run: cd jbmc/unit; ./unit_tests
       - name: Run regression tests
-        run: cd regression; make
+        run: make -C regression test-parallel JOBS=2
       - name: Run JBMC regression tests
-        run: cd jbmc/regression; make
+        run: make -C jbmc/regression test-parallel JOBS=2
 
   check-macos-10_15-cmake-clang:
     runs-on: macos-10.15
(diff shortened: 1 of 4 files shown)


hannes-steffenhagen-diffblue:
Because most of our time is spent in regression tests and some regression test suites take longer than others, rather than passing `-j` to make or ctest for regression testing I’d recommend install `Thread::Pool::Simple` and setting `TESTPL_JOBS=2` before running the tests, this will make the actual tests run in parallel instead. `cbmc` tests take especially long so I’d expect a greater speedup that way.

NlightNFotis (approved):
Excellent. :+1:

How much of a speedup do we get with this? Last time I remember builds where taking between 50mins-1hour across all platforms. Hopefully this drops it to about 30-40mins.

.github/workflows/pull-request-checks.yaml
```diff
@@ -52,22 +51,21 @@ jobs:
           # user input
           DEBIAN_FRONTEND: noninteractive
         run: |
-          sudo apt-get install -yq cmake ninja-build gcc g++ maven flex bison libxml2-utils cpanminus dpkg-dev
-          cpanm Thread::Pool::Simple
```
thomasspriggs, line 56: Why is this perl module being removed? It is required for the multi threading functionality of `test.pl`. This should currently be enabled via the `TESTPL_JOBS` environment variable.

thomasspriggs (commented):
Does CI actually finish in less time with these changes?

.github/workflows/pull-request-checks.yaml
NlightNFotis: Yes, sorry, I overlooked it - please don't delete this, this is actually helpful.

tautschnig:
> Excellent. 👍
> 
> How much of a speedup do we get with this? Last time I remember builds where taking between 50mins-1hour across all platforms. Hopefully this drops it to about 30-40mins.

I have added the info to the commit message: on Ubuntu it's down to 30-40 minutes, and MacOS it's under 1 hour. Windows I didn't yet have data for #5606 had not been merged yet.

.github/workflows/pull-request-checks.yaml
tautschnig: Except ... we never set TESTPL_JOBS anywhere? I indeed was pondering to set this variable, but then we had been successful on Travis in the past with the use of GNU parallel, hence I opted for using the `test-parallel` target instead. But I'm fine either way. FWIW, the time for running the regression tests was almost exactly halved.

tautschnig:
> Because most of our time is spent in regression tests and some regression test suites take longer than others, rather than passing `-j` to make or ctest for regression testing I’d recommend install `Thread::Pool::Simple` and setting `TESTPL_JOBS=2` before running the tests, this will make the actual tests run in parallel instead. `cbmc` tests take especially long so I’d expect a greater speedup that way.

As said in https://github.com/diffblue/cbmc/pull/5608#discussion_r525086706, I'd be happy to go down that route. I just picked the somewhat more conservative approach that seemed to work fine in the past.

.github/workflows/pull-request-checks.yaml
NlightNFotis: `TESTPL_JOBS` at least for this particular action, is being set in line 71 of this file as an environment variable, to my understanding.

thomasspriggs: I think the `test.pl` level of multi threading is included in the existing `.yaml` but isn't actually working correctly. Looking at the CI log from another PR there are existing warnings.
<details>
<summary> `cpanm` output </summary>

```
!
! Can't write to /usr/local/share/perl/5.30.0 and /usr/local/bin: Installing modules to /home/runner/perl5
! To turn off this warning, you have to do one of the following:
!   - run me as a root or with --sudo option (to install to /usr/local/share/perl/5.30.0 and /usr/local/bin)
!   - Configure local::lib in your existing shell to set PERL_MM_OPT etc.
!   - Install local::lib by running the following commands
!
!         cpanm --local-lib=~/perl5 local::lib && eval $(perl -I ~/perl5/lib/perl5/ -Mlocal::lib)
!
--> Working on Thread::Pool::Simple
Fetching http://www.cpan.org/authors/id/J/JW/JWU/Thread-Pool-Simple/Thread-Pool-Simple-0.25.tar.gz ... OK
Configuring Thread-Pool-Simple-0.25 ... OK
Building and testing Thread-Pool-Simple-0.25 ... OK
Successfully installed Thread-Pool-Simple-0.25
```

</details>  


<details>
<summary> `test.pl` output </summary>

```
1: Jobs set but thread pool module not found,
1: install with 'cpan -i Thread::Pool::Simple'
```

</details> 

I am fine with removing this, as it isn't currently working. However I would prefer the removal to be in a separate commit with the "because it doesn't work" explanation. That way if we ever want to re-instate it we can revert that commit only.

.github/workflows/pull-request-checks.yaml
```diff
@@ -52,22 +51,21 @@ jobs:
...
       - name: Check if package building works
         run: |
           cd build
           ninja package
           ls *.deb
       - name: Run tests
-        run: cd build; ctest . -V -L CORE
+        run: cd build; ctest . -V -L CORE -j2
         env:
           TESTPL_JOBS: 2
```
thomasspriggs: :no_entry_sign: If we are removing the installation of `Thread::Pool::Simple` we should also remove the setting of this environment variable. That will keep things consistent and avoid `test.pl` attempting to use a missing module.

.github/workflows/pull-request-checks.yaml
tautschnig: Thanks, will move this to a commit of its own!

thomasspriggs (changes requested):
I intend to approve once the issue of partial removal of the `test.pl` level multi-threading support is resolved.

tautschnig:
> I intend to approve once the issue of partial removal of the `test.pl` level multi-threading support is resolved.

Should now be done.

thomasspriggs (approved):
:+1:

PR #5609, opened 2020-11-17
Use CCache to speed up build times in GitHub actions

tautschnig:
We used to do this on Travis and CodeBuild, let's also port this to
GitHub actions.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -251,7 +251,21 @@ jobs:
       - name: Fetch dependencies
         run: |
           choco install winflexbison3
+          nuget install clcache -OutputDirectory "c:\tools" -ExcludeVersion -Version 4.1.0
+          echo "c:\tools\clcache\clcache-4.1.0" >> $env:GITHUB_PATH
       - uses: microsoft/setup-msbuild@v1.0.2
+      - name: Prepare ccache
+        uses: actions/cache@v2
+        with:
+          path: .ccache
+          key: ${{ runner.os }}-msbuild-${{ github.ref }}-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-msbuild-${{ github.ref }}
+            ${{ runner.os }}-msbuild
+      - name: ccache environment
+        run: |
+          echo "CLCACHE_BASEDIR=$((Get-Item -Path '.\').FullName)" >> $env:GITHUB_ENV
+          echo "CLCACHE_DIR=$pwd\.ccache" >> $env:GITHUB_ENV
       - name: Configure with cmake
         run: |
           New-Item -ItemType Directory -Path build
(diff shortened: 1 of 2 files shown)


tautschnig:
FWIW, I am planning to do a separate PR to clean up the vs-2019-build-and-test configuration and, once done, also make it use clcache.

NlightNFotis (approved):
For me this looks good, but I'd be wary of merging something that's actively breaking CI, even if it promises to fix it later.

NlightNFotis:
Also, this won't be able to be merged because the `check-vs-2019-build-and-test` check is marked as required. Can the two PRs not be one?

hannes-steffenhagen-diffblue:
Have you checked if this has a measurable performance improvement in subsequent builds?

**EDIT** NVM apparently it’s in the commit messages

hannes-steffenhagen-diffblue:
Looks like one java unit test is failing due to what looks like a pretty sketchy pretty function. Relies on a lot of platform specific behaviour which seems unnecessary.

tautschnig:
@NlightNFotis 
> Also, this won't be able to be merged because the `check-vs-2019-build-and-test` check is marked as required. Can the two PRs not be one?

@hannes-steffenhagen-diffblue 
> Looks like one java unit test is failing due to what looks like a pretty sketchy pretty function. Relies on a lot of platform specific behaviour which seems unnecessary.

Argh, this was fixed in #5613. I should have rebased this earlier today, my apologies.

thomasspriggs:
The issue with the test actually appears to be a long-standing issue with the `floating_point_to_java_string` function when compiled under Visual Studio. I would say that the issue is with the function rather than the test. This is because 1) Java is usually intended to be platform independent and 2) because the doxygen for this function mentions not printing unnecessary zeros, but it is returning more zeros when compiled under visual studio than on other platforms. Therefore I would prefer to see the function under test fixed, rather than in increasing number of exceptions for different compilers added to the unit test.

tautschnig:
> The issue with the test actually appears to be a long-standing issue with the `floating_point_to_java_string` function when compiled under Visual Studio. I would say that the issue is with the function rather than the test. This is because 1) Java is usually intended to be platform independent and 2) because the doxygen for this function mentions not printing unnecessary zeros, but it is returning more zeros when compiled under visual studio than on other platforms. Therefore I would prefer to see the function under test fixed, rather than in increasing number of exceptions for different compilers added to the unit test.

+100, but #5613 was the quick way to stop the bleeding.

thomasspriggs:
It is worth noting that the total cache size for the whole repository is limited to 5GB. Source - https://github.com/actions/cache#cache-limits

tautschnig:
> It is worth noting that the total cache size for the whole repository is limited to 5GB. Source - https://github.com/actions/cache#cache-limits

Yes, that's why I limited each job to 500 MB, I should perhaps have noted that in the commit message. GitHub also compresses the tar ball, making it around 50 MB it seems, so we can have quite a lot of cache keys before old ones are expired.

PR #5610, opened 2020-11-17
Variable sensitivity object factory rework

jezhiggins:
This PR removes the variable_sensitivity_object_factoryt static instance. 

Now, a configured variable_sensitivity_object_factoryt instance is passed into the variable_sensitivity_domain_factoryt, which in turn passes down into the variable_sensitivity_domaint objects it creates. Those pass the object factory to their abstract_environments and, finally, they use the factory to
build the objects they need.

The variable_sensitivity_object_factoryt are created and passed as shared pointers.

### DIFF
--- a/src/analyses/variable-sensitivity/variable_sensitivity_domain.h
+++ b/src/analyses/variable-sensitivity/variable_sensitivity_domain.h
@@ -218,14 +218,11 @@ public:
 #endif
 };
 
-template <>
-class ai_domain_factory_default_constructort<variable_sensitivity_domaint>
+class variable_sensitivity_domain_factoryt
   : public ai_domain_factoryt<variable_sensitivity_domaint>
 {
 public:
-  ai_domain_factory_default_constructort() = delete;
-
-  explicit ai_domain_factory_default_constructort(
+  explicit variable_sensitivity_domain_factoryt(
     variable_sensitivity_object_factory_ptrt _object_factory)
     : object_factory(_object_factory)
   {
@@ -241,9 +238,6 @@ public:
 private:
   variable_sensitivity_object_factory_ptrt object_factory;
 };
-
-using variable_sensitivity_domain_factoryt =
-  ai_domain_factory_default_constructort<variable_sensitivity_domaint>;
 
 #ifdef ENABLE_STATS
 template <>
--- a/src/analyses/variable-sensitivity/variable_sensitivity_domain.h
+++ b/src/analyses/variable-sensitivity/variable_sensitivity_domain.h
@@ -76,7 +76,8 @@ class variable_sensitivity_domaint : public ai_domain_baset
 public:
   explicit variable_sensitivity_domaint(
     variable_sensitivity_object_factory_ptrt _object_factory)
-  : abstract_state(_object_factory) {
+    : abstract_state(_object_factory)
+  {
   }
 
   /// Compute the abstract transformer for a single instruction
@@ -217,7 +218,7 @@ public:
 #endif
 };
 
-template<>
+template <>
 class ai_domain_factory_default_constructort<variable_sensitivity_domaint>
   : public ai_domain_factoryt<variable_sensitivity_domaint>
 {
@@ -226,7 +227,8 @@ public:
 
   explicit ai_domain_factory_default_constructort(
     variable_sensitivity_object_factory_ptrt _object_factory)
-  : object_factory(_object_factory) {
+    : object_factory(_object_factory)
+  {
   }
 
   std::unique_ptr<statet> make(locationt l) const override
--- a/src/analyses/variable-sensitivity/variable_sensitivity_domain.h
+++ b/src/analyses/variable-sensitivity/variable_sensitivity_domain.h
@@ -71,15 +71,12 @@
 #include <analyses/ai.h>
 #include <analyses/variable-sensitivity/abstract_enviroment.h>
 
-class variable_sensitivity_object_factoryt;
-using variable_sensitivity_object_factory_ptrt =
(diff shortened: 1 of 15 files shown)


src/analyses/variable-sensitivity/variable_sensitivity_domain.h
```diff
@@ -212,6 +218,33 @@ class variable_sensitivity_domaint : public ai_domain_baset
 #endif
 };
 
+template <>
+class ai_domain_factory_default_constructort<variable_sensitivity_domaint>
```
martin-cs: All that `ai_domain_factory_default_constructort` adds over `ai_domain_factoryt` is a implementation of `make` which calls the default constructor of the domain.  As you are implementing your own `make` it should be fine to simply have this as `class variable_sensitivity_domain_factoryt : public ai_domain_factoryt<variable_sensitivity_domaint>` which may simplify things a little.

martin-cs (approved):
Could be merged as-is.  I think the simplification of avoiding `ai_domain_factory_default_constructort` is worth doing but I won't insist on it.

src/analyses/variable-sensitivity/variable_sensitivity_domain.h
jezhiggins: I think it's a bit more involved than that because the default constructor for class ait is is 

```
  ait()
    : ai_recursive_interproceduralt(
        util_make_unique<
          ai_history_factory_default_constructort<ahistoricalt>>(),
        util_make_unique<ai_domain_factory_default_constructort<domainT>>(),
        util_make_unique<location_sensitive_storaget>())
  {
  }
```

In the variable sensitivity domain we call 
```
  explicit ait(std::unique_ptr<ai_domain_factory_baset> &&df)
    : ai_recursive_interproceduralt(
        util_make_unique<
          ai_history_factory_default_constructort<ahistoricalt>>(),
        std::move(df),
        util_make_unique<location_sensitive_storaget>())
  {
  }

```
but the default constructor still gets generated, and without that explicit specialisation variable_sensitivity_domaint has to be default constructable.  However a consequence of this change is that variable_sensitivity_domaint is no longer default constructable :) It can simplified, but I'd like to handle that as a separate PR because it's a bit more than just changing the variable_sensitivity_domain_factoryt definition.

tautschnig:
I was almost going to hit merge on this one, but I'd like to ask for the churn in those commits to be cleaned up first. There are several changes in later commits that just fix code only introduced in this PR. Also, it would be nice if each commit did still fully compile, i.e., updates to the unit test should be merged into the commit that breaks the unit test.

src/analyses/variable-sensitivity/variable_sensitivity_domain.h
martin-cs: Hmmm... yes; good point.  There are a number of ways of solving this...

For context, `goto-analyzer` has options for which algorithm you want to use to do the abstract interpretation.  Options are:

`--recursive-interprocedural` / `ai_recursive_interproceduralt` -- this is the current best choice for most applications as it allows you to pick the domain, history and storage independently.

`--three-way-merge` / `ai_three_way_merget` -- a specific kind of abstract interpretation that only works with VSD, it does a more sophisticated merge for function calls.

`--legacy-ait` / `ait` -- basically the backwards compatible interface to `ai_recursive_interproceduralt` it fixes a bunch of the options so that it is code compatible with older code.  At the moment this is the default if you don't specify

`--legacy-concurrent` / `concurrency_aware_ait` -- a very specific kind of analysis for concurrent code that only works with one specific domain.  Used by the dependence graph code.


The specific problem is that:
A. `variable_sensitivity_domain_factoryt` should not be default constructable, it should take options.
B. Instantiating `ait` requires that the factory is default constructable (even if it is not used).
C. At the moment `--legacy-ait --variable-sensitivity` is supported, which requires instantiating `ait<variable_sensitivity_domaint>` even thought the default constructor is not used.
D. Because the default abstract interpreter is `--legacy-ait`, supporting `--legacy-ait --variable-sensitivity` allows backwards compatibility.

So I think the options are:
A/1. make `variable_sensitivity_domain_factoryt` default constructable with default options or a `INVARIANT(false,"Don't use this, it's here because...");`
B/1. the existing clever solution of specialising `ai_domain_factory_default_constructort` and deleting the default constructor so that the instantiation works as long as it is not used.
B/2. remove the default constructor from `ait` and always require a domain factory.  This breaks backwards compatibility but maybe it's not that bad given that we have just merged a clean-up PR that would also break things.
C/1. remove `else if(options.get_bool_option("vsd"))` from `legacy-ait`.  This requires that users of VSD give the option `--recursive-interprocedural` or `--three-way-merge` (which is a good idea anyway).
D/1. switch the default abstract interpreter to `--recursive-interprocedural` and remove as C/1.  This should happen at some point but it means that all of the users of the dependency graph have to give `--legacy-ait`.
D/2. D/1 plus making the abstract interpreter default to `--legacy-ait` when the dependency graph domains are requested.

Let me try some stuff...

martin-cs: The following seems to compile for me and I think it is a simpler version of B/1:

https://github.com/jezhiggins/cbmc/pull/1

I think this is probably the most compact solution if it compiles on all platforms.

jezhiggins:
@tautschnig I've cleaned up the history around those last steps - you're right about the unit tests and that was an oversight on my part. I'm happy to do more, but I'd like to keep the earlier history as is because collapsing that feels like it makes the change a bit too magical - it was like this, now it's like that - when it was a little bit of a voyage of discovery.

martin-cs (approved):
Thanks for the fixes.

PR #5611, opened 2020-11-18
C front-end: printing enum type declarations with bitvector awareness

tautschnig:
The constants are no longer stored as string literals, and instead use
the bitvector representation (currently: hexadecimal values).

### DIFF
--- a/regression/goto-instrument/dump-enum/test.desc
+++ b/regression/goto-instrument/dump-enum/test.desc
@@ -2,6 +2,7 @@ CORE
 main.c
 --dump-c
 enum hex \{ .* \};
+h=/\*enum\*/V10;
 ^EXIT=0$
 ^SIGNAL=0$
 --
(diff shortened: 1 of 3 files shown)


regression/goto-instrument/dump-enum/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+--dump-c
+enum hex \{ .* \};
```
martin-cs, line 4: This feels like a fairly minimal check given the contents of the test file.  Could this be strengthened or, alternatively, documented in the test case why this is sufficient.

martin-cs (approved):

regression/goto-instrument/dump-enum/test.desc
tautschnig: Oh, indeed what was obvious to me as I saw the failures first-hand wasn't at all obvious here! I've added a comment to the test.desc file.

martin-cs: Right!  I get it now!

kroening (approved):

PR #5615, opened 2020-11-18
Remove {f,F}orall_goto_functions

tautschnig:
This was useful in the past, but with C++-11 we can use a ranged-for to
avoid the iterator altogether (in all but the cases in goto-diff, where
we now make the iterator explicit).

### DIFF
--- a/src/analyses/flow_insensitive_analysis.cpp
+++ b/src/analyses/flow_insensitive_analysis.cpp
@@ -60,11 +60,12 @@ void flow_insensitive_analysis_baset::output(
   const goto_functionst &goto_functions,
   std::ostream &out)
 {
-  forall_goto_functions(f_it, goto_functions)
+  for(const auto &gf_entry : goto_functions.function_map)
   {
-    out << "////\n" << "//// Function: " << f_it->first << "\n////\n\n";
+    out << "////\n"
+        << "//// Function: " << gf_entry.first << "\n////\n\n";
 
-    output(f_it->first, f_it->second.body, out);
+    output(gf_entry.first, gf_entry.second.body, out);
   }
 }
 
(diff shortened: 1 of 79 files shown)


hannes-steffenhagen-diffblue (approved):

hannes-steffenhagen-diffblue:
Yes please. Though I think it’d probably be a bit too ambitious to do them all in one PR

src/analyses/flow_insensitive_analysis.h
```diff
@@ -173,7 +173,7 @@ class flow_insensitive_analysis_baset
     const goto_functionst &goto_functions);
 
   bool fixedpoint(
-    goto_functionst::function_mapt::const_iterator it,
+    const goto_functionst::function_mapt::value_type &gf_entry,
```
martin-cs: This feels like it might be better as a `goto_programt`?

src/analyses/flow_insensitive_analysis.cpp
```diff
@@ -383,20 +384,21 @@ void flow_insensitive_analysis_baset::fixedpoint(
...
     {
-      fixedpoint(it, goto_functions);
+      fixedpoint(gf_entry, goto_functions);
     }
+  }
 }
 
 bool flow_insensitive_analysis_baset::fixedpoint(
-  const goto_functionst::function_mapt::const_iterator it,
+  const goto_functionst::function_mapt::value_type &gf_entry,
```
martin-cs: `goto_programt` ?

src/goto-checker/symex_coverage.cpp
```diff
@@ -52,7 +52,7 @@ class goto_program_coverage_recordt : public coverage_recordt
 public:
   goto_program_coverage_recordt(
     const namespacet &ns,
-    goto_functionst::function_mapt::const_iterator gf_it,
+    const goto_functionst::function_mapt::value_type &gf_entry,
```
martin-cs: Feels a little cumbersome.  Isn't this a `goto_functiont`?

martin-cs (approved):
Thanks for these patches; sorry I missed this one.

peterschrammel (approved):

src/goto-checker/symex_coverage.cpp
tautschnig: These also require the name of the `goto_functiont`. What could be done is passing that information as two arguments? That would probably make for are clearer interface. Let me fix this.

tautschnig: Now done.

PR #5618, opened 2020-11-19
goto_cc_modet isn't a messaget

tautschnig:
Use local messaget instances instead as there is no is-a relationship
between goto_cc_modet (nor any class deriving from it) and messaget.

### DIFF
--- a/src/goto-cc/ld_mode.cpp
+++ b/src/goto-cc/ld_mode.cpp
@@ -88,7 +88,7 @@ int ld_modet::doit()
   if(cmdline.isset("version") || cmdline.isset("print-sysroot"))
     return run_ld();
 
-  eval_verbosity(
+  messaget::eval_verbosity(
     cmdline.get_value("verbosity"), messaget::M_ERROR, gcc_message_handler);
 
   compilet compiler(cmdline, gcc_message_handler, false);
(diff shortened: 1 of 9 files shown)


martin-cs (approved):

src/goto-cc/ld_mode.cpp
```diff
@@ -88,7 +88,7 @@ int ld_modet::doit()
   if(cmdline.isset("version") || cmdline.isset("print-sysroot"))
     return run_ld();
 
-  eval_verbosity(
+  messaget::eval_verbosity(
     cmdline.get_value("verbosity"), messaget::M_ERROR, gcc_message_handler);
```
kroening, line 93: Does that do anything?

src/goto-cc/ms_link_mode.cpp
```diff
@@ -30,7 +30,7 @@ int ms_link_modet::doit()
     return 0;
   }
 
-  eval_verbosity(
+  messaget::eval_verbosity(
     cmdline.get_value("verbosity"), messaget::M_ERROR, message_handler);
```
kroening, line 35: Same here

tautschnig: I had wondered as well, but yes, it does: the `message_handler` argument will have its verbosity set. We might want to change this interface though as that makes two people who were surprised by this code...

kroening (approved):

PR #5625, opened 2020-11-19
GitHub actions: Make all Windows checks consistent

tautschnig:
With the exceptions of whether or not to create packages, run tests, or
do signing, make all three Windows configurations exactly the same.
Apart from cleanliness, the main benefit is the use of clcache for all
Windows build jobs, making them consistently faster.


tautschnig:
@NlightNFotis @hannes-steffenhagen-diffblue @thomasspriggs This is finally ready for review.

.github/workflows/pull-request-checks.yaml
```diff
@@ -239,21 +251,21 @@ jobs:
...
-  # I just added this to check if my stuff is working on github actions
   windows-msi-package:
     runs-on: windows-2019
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v2
         with:
-          submodules: recursive
+          submodules: true
```
hannes-steffenhagen-diffblue: I’d prefer if this stayed at recursive. It’s not needed right now becaue we don’t have submodules with submodules, but if we ever do then we won’t run into mysterious CI bugs then (well, not this particular one at any rate).

regression/validate-trace-xml-schema/check.py
```diff
@@ -11,49 +11,47 @@
...
-    # this test is marked broken-smt-backend and doesn't work for me locally
-    'integer-assignments1/test.desc'
+    ['reachability-slice', 'test.desc'],
+    ['reachability-slice', 'test2.desc'],
+    ['reachability-slice', 'test3.desc'],
+    ['reachability-slice-interproc', 'test.desc'],
+    ['unsigned1', 'test.desc'],
+    ['reachability-slice-interproc3', 'test.desc'],
+    ['sync_lock_release-1', 'symbol_per_type.desc'],
+    # this test is marked smt-backend
```
hannes-steffenhagen-diffblue: No, I meant `broken-smt-backend`. I don’t think regular smt backend should be a problem. I actually don’t remember why the broken smt backend is a problem either because these shouldn’t pass the extra smt flag we do for the smt backend tests?

hannes-steffenhagen-diffblue (approved):
FWIW Github actions has a cache size limit of 5GB, one build of cbmc takes up ~120MB on my local machine. We need to keep an eye on the cache, for now it's probably fine but if gets thrashed all the time (which it easily may on days where we get a lot of PRs) caching is probably more of a pessimisation than optimisation.

.github/workflows/pull-request-checks.yaml
tautschnig: I have added an additional commit to make all actions `recursive` as we previously had a rather arbitrary mix.

regression/validate-trace-xml-schema/check.py
tautschnig: My comment was misleading, I've fixed this now. What I meant to say (and the comment now does say): we cannot expect to run this test successfully as we don't use the SMT solver in the XML validation tests.

PR #5629, opened 2020-11-23
goto_analyzer --verify now prints histories in the XML and JSON output

martin-cs:
This patch adds to the `--json` and `--xml` output to give the `ai_historyt`s that show that an assertion is either false or unknown.

To do this I had to refactor the output handling a bit.  I think it is now simpler.  In the process I have improved the `properties` output in the case where a history other than `ahistoricalt` is used.

I haven't add the traces to the text output because it wasn't obvious to how to do it without breaking the format (maybe there is something to be said for JSON and even XML!).  I'm working on a new task that will do a lot more with outputting the histories for failed verifications so that might be the right way of handling it.

### DIFF
--- a/src/goto-analyzer/static_verifier.cpp
+++ b/src/goto-analyzer/static_verifier.cpp
@@ -115,24 +115,7 @@ static void static_verifier_xml(
   {
     xmlt &x = xml_result.new_element("result");
 
-    switch(result.status)
-    {
-    case static_verifier_resultt::TRUE:
-      x.set_attribute("status", "SUCCESS");
-      break;
-
-    case static_verifier_resultt::FALSE:
-      x.set_attribute("status", "FAILURE (if reachable)");
-      break;
-
-    case static_verifier_resultt::BOTTOM:
-      x.set_attribute("status", "SUCCESS (unreachable)");
-      break;
-
-    case static_verifier_resultt::UNKNOWN:
-      x.set_attribute("status", "UNKNOWN");
-    }
-
+    x.set_attribute("status", message(result.status));
     x.set_attribute("file", id2string(result.source_location.get_file()));
     x.set_attribute("line", id2string(result.source_location.get_line()));
     x.set_attribute(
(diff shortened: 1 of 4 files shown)


martin-cs:
( Review commit by commit; it makes more sense like that. )

src/goto-analyzer/static_verifier.cpp
```diff
@@ -27,6 +27,31 @@ struct static_verifier_resultt
   irep_idt function_id;
 };
 
+static static_verifier_resultt::statust
```
tautschnig: Wouldn't a declaration also do the trick? (Just to reduce churn.)

tautschnig (approved):

src/goto-analyzer/static_verifier.cpp
martin-cs: Alas no.  The problem is...

1. I want to move the JSON and XML generation code for a single result out of the methods.  Having them as member functions of `static_verifier_resultt` seems consistent with the code base.
2. They use the `message` function, so it must appear be declared after the `statust` enum but before `output_(xml|json)`.
3. While the `enum` is declared in `static_verifier_result` this is not possible without making it a static member function which doesn't really reduce churn and misses the point of the separation really.

I see your point about churn but i have tried to minimise it and I don't think it is too bad.

martin-cs:
@peterschrammel @chrisr-diffblue @hannes-steffenhagen-diffblue Any chance of a maintainer review?  Reviewed by commit it should be pretty straight forward.

src/goto-analyzer/static_verifier.cpp
```diff
@@ -46,6 +46,45 @@ struct static_verifier_resultt
...
+      {"falseHistories", false_json},
+    };
+  }
+
+  xmlt output_xml(void) const
+  {
+    xmlt x("result");
+
+    x.set_attribute("status", message(this->status));
+    x.set_attribute("file", id2string(this->source_location.get_file()));
```
peterschrammel, line 77: Why not `xml(source_location)`?

peterschrammel (approved):

src/goto-analyzer/static_verifier.cpp
martin-cs: ... good question ... will change.

PR #5631, opened 2020-11-25
Variable sensitivity option passing

jezhiggins:
Rework goto-analyser command line options so the domain object types can be configured individually. Adds/changes the following 

```
--vsd-values constants|intervals|set-of-constants
--vsd-arrays top-bottom|every-element
--vsd-pointers top-bottom|constants|value-set
--vsd-structs top-bottom|every-field
--vsd-unions top-bottom
```

and removes 

```
--vsd-value-sets
--vsd-intervals
```


src/analyses/variable-sensitivity/variable_sensitivity_object_factory.cpp
```diff
@@ -10,8 +10,217 @@
...
+      data_dependency_contextt>(type, top, bottom, e, environment, ns);
+  if(configuration.context_tracking.last_write_context)
+    return initialize_context_abstract_object<
+      abstract_object_classt,
+      write_location_contextt>(type, top, bottom, e, environment, ns);
+  else
+  {
+    if(top || bottom)
+    {
+      return abstract_object_pointert(
```
tautschnig, line 209: Nit picking (here and elsewhere): use curly braces for constructors as indicated in the coding guidelines.

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.cpp
```diff
@@ -10,8 +10,217 @@
...
+
+  // This should always be on (for efficeny with 3-way merge)
+  // Does not work with value set
+  config.context_tracking.last_write_context =
+    (config.value_abstract_type != VALUE_SET) &&
+    (config.pointer_abstract_type != VALUE_SET);
+  config.context_tracking.data_dependency_context =
+    options.get_bool_option("data-dependencies");
+  config.advanced_sensitivities.new_value_set =
+    options.get_bool_option("new-value-set");
```
tautschnig, line 50: I know this is a refactor only, but is this option documented in the help output actually?

src/goto-analyzer/goto_analyzer_parse_options.cpp
```diff
@@ -991,12 +993,12 @@ void goto_analyzer_parse_optionst::help()
...
     "Variable sensitivity domain (VSD) options:\n"
-    " --vsd-structs                struct field sensitive analysis\n"
-    " --vsd-arrays                 array entry sensitive analysis\n"
-    " --vsd-pointers               pointer sensitive analysis\n"
-    " --vsd-value-sets             use value sets\n"
+    " --vsd-values                 value tracking - constants|intervals|set-of-constants\n"
+    " --vsd-structs                struct field sensitive analysis - top-bottom|every-field\n"
+    " --vsd-arrays                 array entry sensitive analysis - top-bottom|every-element\n"
+    " --vsd-pointers               pointer sensitive analysis - top-bottom|constants|value-set\n"
+    " --vsd-unions                 union sensitive analysis - top-bottom\n"
```
tautschnig, line 1000: Had discussed this with @martin-cs recently: there is a fair abundance of options, and to noobs like myself it's not obvious which ones to choose. As a first step, it would actually help to indicate the default choices in the help output. Could you please add that information?

tautschnig (approved):

src/goto-analyzer/goto_analyzer_parse_options.cpp
martin-cs: I have a goal of writing a PR that improves the documentation in the manual (to explain the options), the code that picks the defaults and the `--help` text to show them.  At the moment it is 2nd to write on my pile after the successor to https://github.com/diffblue/cbmc/pull/5629 .  Will you accept a "promise this will be fixed but not in this PR"?

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.cpp
jezhiggins: It isn't, because it isn't exposed as a command line option. 

There are unit tests that use this option, but in due course this will go away in favour of additional possible arguments to vsd-values and vsd-pointers.

regression/goto-analyzer/variable-sensitivity-bug-03/test.desc
```diff
@@ -1,6 +1,6 @@
 KNOWNBUG
 main.c
---variable-sensitivity --vsd-structs --vsd-arrays --vsd-pointers --show
+--variable-sensitivity --vsd-structs every-field --vsd-arrays every-element --vsd-pointers --show
```
martin-cs: Isn't `--vsd-pointers` missing an argument?

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.h
```diff
@@ -31,14 +31,29 @@
 #include <util/namespace.h>
 #include <util/options.h>
 
+enum ABSTRACT_OBJECT_TYPET
+{
+  TWO_VALUE,
+  CONSTANT,
+  INTERVAL,
+  ARRAY_SENSITIVE,
```
martin-cs, line 39: If you want to rename any of these to match the command-line names / make them clearer then I am fine with that.  Likewise, if you want to keep them as-is for clarity, that is also fine.

regression/goto-analyzer/variable-sensitivity-bug-02/test.desc
```diff
@@ -1,6 +1,6 @@
 KNOWNBUG
 main.c
---variable-sensitivity --vsd-structs --vsd-arrays --vsd-pointers --show
+--variable-sensitivity --vsd-structs every-field --vsd-arrays every-element --vsd-pointers --show
```
martin-cs: I think `--vsd-pointers` should have an argument; it looks lonely without one.

martin-cs (approved):
Yes; looks good and ready to merge after any minor tweaks based on this review.

regression/goto-analyzer/variable-sensitivity-bug-03/test.desc
jezhiggins: Both this test and the one you picked up below are currently skipped in the regression tests. I'll make the change nonetheless.

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.h
jezhiggins: I'll keep them as they are for the moment so we can follow the changes through.

martin-cs:
There is a very small drop in coverage.  I can't work out exactly why because the coverage of the diff is actually higher than the average.  I'm going to merge anyway.

PR #5633, opened 2020-11-25
Restore member-of-union expressions in trace output

tautschnig:
byte_extract operations precisely capture the semantics, but aren't
meaningful to users.

### DIFF
--- a/src/goto-programs/goto_trace.h
+++ b/src/goto-programs/goto_trace.h
@@ -117,6 +117,7 @@ public:
   // for assume, assert, goto
   bool cond_value;
   exprt cond_expr;
+  exprt original_condition;
 
   // for assert
   irep_idt property_id;
@@ -161,6 +162,7 @@ public:
     full_lhs.make_nil();
     full_lhs_value.make_nil();
     cond_expr.make_nil();
+    original_condition.make_nil();
   }
 
   /// Use \p dest to establish sharing among ireps.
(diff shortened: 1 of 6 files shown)


kroening:
It seems a bit odd to do the 'restoring' in the output of a trace; wouldn't it be better to do that when that is built?

tautschnig:
> It seems a bit odd to do the 'restoring' in the output of a trace; wouldn't it be better to do that when that is built?

Indeed! I have reworked this, the patch is much smaller now.

src/goto-programs/goto_trace.h
```diff
@@ -117,6 +117,7 @@ class goto_trace_stept
   // for assume, assert, goto
   bool cond_value;
   exprt cond_expr;
+  exprt original_condition;
```
kroening, line 121: The difference between `original_condition` and `cond_expr` is non-obvious, and does deserve a comment.
Do we really need both?

tautschnig: I was tempted to just re-purpose `cond_expr`, but it turns out it is actually used by `goto-checker/bmc_util.cpp` (I believe that is the only use). `cond_expr` is the SSA-renamed expression, while we'd want to provide an expression that is meaningful to the user. So maybe adding a comment is the right approach?

martin-cs (approved):

peterschrammel (approved):

Issue #5543, opened 2020-11-05
Omitting property checking flags changes coverage checking


markrtuttle:
CBMC version: 5.17.0 (cbmc-5.17.0-12-ga46f12d66
Operating system: Ubuntu 20
Exact command line resulting in the issue: see below
What behaviour did you expect: see below
What happened instead: see below

## Summary

My coverage checking results are affected by whether or not I include property checking flags on the command line.

I had been in the habit of doing property checking with flags like --pointer-check, and doing coverage checking with the same set of flags.  I was advised that I could accelerate coverage checking by omitting the property checking flags when I'm doing coverage checking.  But omitting the flags changes the coverage results.

## Decreased coverage

Omitting property checking flags when doing coverage checking results in lower coverage (lower percentage of lines hit).

Consider the attached file [example2.c](https://github.com/diffblue/cbmc/files/5491205/example2.c.gz) containing

```
#include <stdlib.h>

int foo(int x, int y) { return x; }

int main() {
  int *ptr = malloc(sizeof(*ptr));

  foo(1,
      *ptr);

  if (0) {
    *ptr = 0;
  }

  return 0;
}
```

My results are

```
cbmc example2.c --cover location --malloc-may-fail --malloc-fail-null --pointer-check

** coverage results:
[foo.coverage.1] file example2.c line 3 function foo block 1 (lines example2.c:foo:3): SATISFIED
[main.coverage.5] file example2.c line 15 function main block 5 (lines example2.c:main:15,16): SATISFIED
[main.coverage.4] file example2.c line 12 function main block 4 (lines example2.c:main:12): FAILED
[main.coverage.3] file example2.c line 11 function main block 3 (lines example2.c:main:11): SATISFIED
[main.coverage.2] file example2.c line 6 function main block 2 (lines example2.c:main:6,8,9): SATISFIED
[main.coverage.1] file example2.c line 6 function main block 1 (lines example2.c:main:6): SATISFIED
```

and

```
cbmc example2.c --cover location --malloc-may-fail --malloc-fail-null

** coverage results:
[foo.coverage.1] file example2.c line 3 function foo block 1 (lines example2.c:foo:3): SATISFIED
[main.coverage.5] file example2.c line 15 function main block 5 (lines example2.c:main:15,16): SATISFIED
[main.coverage.4] file example2.c line 12 function main block 4 (lines example2.c:main:12): FAILED
[...]

PR #5636, opened 2020-11-27
Cover failed assertions

hannes-steffenhagen-diffblue:
This adds an option `--cover-failed-assertions` that prevents the
default behaviour of coverage stopping at failed assertions by turning
assertions into skips rather than assumes (which is the default
behaviour for coverage criteria other than `assertion`, which behaves
the same with or without the flag).

This is one way to address https://github.com/diffblue/cbmc/issues/5543

### DIFF
--- a/regression/cbmc/cover-failed-assertions/test-no-failed-assertions.desc
+++ b/regression/cbmc/cover-failed-assertions/test-no-failed-assertions.desc
@@ -1,10 +1,10 @@
 CORE
 test.c
 --cover location --pointer-check --malloc-may-fail --malloc-fail-null
-\[main.coverage.4\] file test.c line \d+ function main block 4 \(lines test.c:main:12,13\): SATISFIED
-\[main.coverage.3\] file test.c line \d+ function main block 3 \(lines test.c:main:10\): FAILED
-\[main.coverage.2\] file test.c line \d+ function main block 2 \(lines test.c:main:4,5,7,9\): SATISFIED
-\[main.coverage.1\] file test.c line \d+ function main block 1 \(lines test.c:main:4\): SATISFIED
+\[main\.coverage\.4\] file test\.c line \d+ function main block 4 \(lines test\.c:main:13,14\): SATISFIED
+\[main\.coverage\.3\] file test\.c line \d+ function main block 3 \(lines test\.c:main:11\): FAILED
+\[main\.coverage\.2\] file test\.c line \d+ function main block 2 \(lines test\.c:main:5,6,8,10\): SATISFIED
+\[main\.coverage\.1\] file test\.c line \d+ function main block 1 \(lines test\.c:main:5\): SATISFIED
 ^EXIT=0$
 ^SIGNAL=0$
 --
--- a/regression/cbmc/cover-failed-assertions/test-no-failed-assertions.desc
+++ b/regression/cbmc/cover-failed-assertions/test-no-failed-assertions.desc
@@ -1,4 +1,4 @@
-CORE
+CORE paths-lifo-expected-failure
 test.c
 --cover location --pointer-check --malloc-may-fail --malloc-fail-null
 \[main\.coverage\.4\] file test\.c line \d+ function main block 4 \(lines test\.c:main:13,14\): SATISFIED
--- a/regression/cbmc/cover-failed-assertions/test-no-failed-assertions.desc
+++ b/regression/cbmc/cover-failed-assertions/test-no-failed-assertions.desc
@@ -1,0 +1,11 @@
+CORE
+test.c
+--cover location --pointer-check --malloc-may-fail --malloc-fail-null
+\[main.coverage.4\] file test.c line \d+ function main block 4 \(lines test.c:main:12,13\): SATISFIED
+\[main.coverage.3\] file test.c line \d+ function main block 3 \(lines test.c:main:10\): FAILED
+\[main.coverage.2\] file test.c line \d+ function main block 2 \(lines test.c:main:4,5,7,9\): SATISFIED
+\[main.coverage.1\] file test.c line \d+ function main block 1 \(lines test.c:main:4\): SATISFIED
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
--- a/regression/cbmc/cover-failed-assertions/test-no-failed-assertions.desc
+++ b/regression/cbmc/cover-failed-assertions/test-no-failed-assertions.desc
@@ -1,11 +1,14 @@
 CORE paths-lifo-expected-failure
 test.c
 --cover location --pointer-check --malloc-may-fail --malloc-fail-null
-\[main\.coverage\.4\] file test\.c line \d+ function main block 4 \(lines test\.c:main:13,14\): SATISFIED
-\[main\.coverage\.3\] file test\.c line \d+ function main block 3 \(lines test\.c:main:11\): FAILED
-\[main\.coverage\.2\] file test\.c line \d+ function main block 2 \(lines test\.c:main:5,6,8,10\): SATISFIED
+\[main\.coverage\.4\] file test\.c line \d+ function main block 4 \(lines test\.c:main:17,18\): SATISFIED
+\[main\.coverage\.3\] file test\.c line \d+ function main block 3 \(lines test\.c:main:15\): FAILED
+\[main\.coverage\.2\] file test\.c line \d+ function main block 2 \(lines test\.c:main:5,6,12,14\): SATISFIED
 \[main\.coverage\.1\] file test\.c line \d+ function main block 1 \(lines test\.c:main:5\): SATISFIED
 ^EXIT=0$
 ^SIGNAL=0$
 --
 ^warning: ignoring
+--
+This is checking that without the --cover-failed-assertions flag we still get the same result as we did before adding
+it in this example.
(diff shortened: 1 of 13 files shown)


src/goto-instrument/cover.cpp
```diff
@@ -285,16 +291,22 @@ static void instrument_cover_goals(
...
+            successor->is_assume() &&
+            successor->get_condition() == i_it->get_condition())
+          {
+            successor->turn_into_skip();
+          }
+          i_it->type = goto_program_instruction_typet::ASSUME;
+        }
+        else
+        {
+          i_it->turn_into_skip();
```
hannes-steffenhagen-diffblue, line 307: Btw we have to skip them rather than just leaving them be because otherwise cbmc would consider these assertions coverage goals.

src/cbmc/README.md
```diff
@@ -76,13 +76,15 @@ by other mechanism may categorise their properties differently.
...
-passed, assertions in the program text are converted into assumptions (these
+There is one further BMC mode that differs more fundamentally: when `--cover`
+is passed, assertions in the program text are converted into assumptions (these
 must hold for control flow to proceed past them, but are not goals for the
-equation solver), while new `assert(false)` statements are added throughout the
-source program representing coverage goals. The equation solving process then
-proceeds the same as in all-properties mode. Coverage solving is implemented by
-`bmc_covert`, but is structurally practically identical to
+equation solver; In cases where this behaviour is undesirable you can pass the
+`--cover-failed-assertions` which makes coverage checking continue even for
```
hannes-steffenhagen-diffblue, line 83: BTW shouldn’t this file be under `doc` rather than `src`? I don’t really understand our documentation structure...

peterschrammel: The READMEs which are in each module are pulled into here: http://cprover.diffblue.com/folder-walkthrough.html (click on `cbmc`, for example).
We decided to keep them in the source folders so that they are as close as possible to the code.

src/goto-instrument/cover.h
```diff
@@ -22,9 +22,17 @@ class message_handlert;
 class cmdlinet;
 class optionst;
 
-#define OPT_COVER "(cover):"
-#define HELP_COVER \
-  " --cover CC                   create test-suite with coverage criterion CC\n"
+#define OPT_COVER                                                              \
+  "(cover):"                                                                   \
+  "(cover-failed-assertions)"
```
martin-cs, line 27: Is it possible to have this as an argument to `--cover`?

src/goto-instrument/cover.h
```diff
@@ -22,6 +22,10 @@ class message_handlert;
 class cmdlinet;
 class optionst;
 
+#define OPT_COVER "(cover):"
```
martin-cs: Thank you for this refactoring.  Finding an removing unnecessary and redundant repetition is always good.

src/cbmc/README.md
```diff
@@ -76,13 +76,15 @@ by other mechanism may categorise their properties differently.
 
 ### Coverage mode
 
-There is one further BMC mode that differs more fundamentally: when `--cover` is
-passed, assertions in the program text are converted into assumptions (these
+There is one further BMC mode that differs more fundamentally: when `--cover`
+is passed, assertions in the program text are converted into assumptions (these
```
martin-cs, line 80: I don't want to open a massive can of worms here but ... why is this the default?  Is this actually useful?  Given that there are options for `--no-assertions`, `--no-assumptions` and `--assert-to-assume` (whether they work or even do anything at all ... is an open question) could we make the transformation of assertions and assumptions orthogonal to `--cover`.  It feels like this should be possible and even desirable...

martin-cs (approved):
Approve because the refactor is good.  I would prefer if the new option could be added in to the argument for `--cover` and I would prefer it even more if we could make assertion transformation orthogonal to and preceeding coverage assertions or at least have a conversation about it but I also realise that you need to get stuff done so approve.

src/cbmc/README.md
hannes-steffenhagen-diffblue: @martin-cs I think this is the default because whoever wrote this had the expectation that programs should stop at failed assertions. I suppose we could change it, but I’m always a bit wary about changing defaults.

hannes-steffenhagen-diffblue: @martin-cs FWIW because cover treats each failed goal as something relating to coverage right now we have to do _something_ with asserts. That said, I think the alternative (just making cover cleverer about how it counts coverage) would be more desirable.

I'd consider that as something we can take a look at in terms of future work though.

martin-cs: Yes, if you are doing a naive count then, I see the need to do something with pre-existing assertions or otherwise true assertions (generally a good thing) will count against your coverage.  Smarter counting would certainly be one way to address this.

It feels to me that making the "doing something about asserts" independent might be a desirable option as well.  I can see use-cases for converting to SKIP, to ASSUME and leaving as they are.

Then again, I am not a user of this code at the moment and do not have time to work on it so my views maybe not so important.

regression/validate-trace-xml-schema/check.py
```diff
@@ -34,6 +34,9 @@
     ['unknown-argument-suggestion', 'test.desc'],
     # this one produces XML intermingled with main XML output when used with --xml-ui
     ['graphml_witness2', 'test.desc'],
+    # these are producing coverage goals which aren't including in the schema
```
peterschrammel: including -> included?

src/goto-instrument/cover.cpp
peterschrammel: Why not make `--no-assertions` (aka turn-assertions-into-skips) useable with `--cover` then (as @martin-cs  suggests)? That would make the behaviour more transparent to the user.

src/cbmc/cbmc_parse_options.h
```diff
@@ -73,8 +75,9 @@ class optionst;
   "(property):(stop-on-fail)(trace)" \
   "(error-label):(verbosity):(no-library)" \
   "(nondet-static)" \
-  "(version)" \
-  "(cover):(symex-coverage-report):" \
+  "(version)"        \
```
thomasspriggs: :grey_question: Is this additional spacing accidental?

hannes-steffenhagen-diffblue:
@peterschrammel `--no-assertions` applies to `goto_check` assertions, it only turns user provided assertions into skips. What we'd want for cover is to turn _all_ non-cover related assertions into skips, user added or not. Making it work differently depending on whether or not `--cover` is passed would be confusing.

That said, maybe the better solution would be to make sure cover only checks cover-added assertions for coverage metrics instead of this skip-workaround...

peterschrammel:
...unless you use `--cover assertion`.

hannes-steffenhagen-diffblue:
I don’t think the behaviour is any different for `--cover assertion`? It still only removes user assertions (from what I can tell `--cover assertions` doesn’t care specifically about this option at all, this is all happening in goto check), and the original complaint this is meant to address is that coverage changes based on whether or not you pass in the various `*-check` flags (which I think may well be expected behaviour in some circumstances, but is clearly not the only reasonable behaviour).

It’s just that (only) removing user assertions is actually what we want for `--cover assertion` I guess.

regression/cbmc/cover-failed-assertions/test-no-failed-assertions.desc
```diff
@@ -0,0 +1,11 @@
...
+test.c
+--cover location --pointer-check --malloc-may-fail --malloc-fail-null
+\[main.coverage.4\] file test.c line \d+ function main block 4 \(lines test.c:main:12,13\): SATISFIED
+\[main.coverage.3\] file test.c line \d+ function main block 3 \(lines test.c:main:10\): FAILED
+\[main.coverage.2\] file test.c line \d+ function main block 2 \(lines test.c:main:4,5,7,9\): SATISFIED
+\[main.coverage.1\] file test.c line \d+ function main block 1 \(lines test.c:main:4\): SATISFIED
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
```
thomasspriggs, line 11: :pick: It would be nice to have a comment in each of `.desc` files explaining what the test is intended to test. Even something pretty short could make this a while lot easier to maintain.

src/goto-instrument/cover.h
```diff
@@ -22,9 +22,17 @@ class message_handlert;
...
-#define HELP_COVER \
-  " --cover CC                   create test-suite with coverage criterion CC\n"
+#define OPT_COVER                                                              \
+  "(cover):"                                                                   \
+  "(cover-failed-assertions)"
+
+#define HELP_COVER                                                             \
+  " --cover CC                   create test-suite with coverage criterion "   \
+  "CC\n"                                                                       \
+  " --cover-failed-assertions    do not stop coverage checking at failed "     \
```
thomasspriggs, line 32: :pick: Because of the preceding refactor, this option now applies to `jdiff` and `goto-diff` as well as `cbmc`. This means that we should probably test that it works with those entry points.

src/goto-instrument/cover.h
thomasspriggs, line 27: :pick: If we keep this as a separate option from `--cover` then we should check that `--cover-failed-assertions` is not passed without `--cover` and exit with a suitable error message in the case of that particular invalid usage.

thomasspriggs (approved):
Seems reasonable. If it was up to me, I would squash the commits fixing the tests back into the commit where they were added. This is because it could make use of `git bisect` easier, if we ever need to use it. Please fix the spelling error noted by Peter before this is merged.

peterschrammel (approved):
I understand that this needs a quick fix. But I don't think that the current solution will be the last. The options around goto check assertions, user assertions, cover, the various -no-X options, etc need to be made consistent so that they can be used in a clearly understandable and orthogonal way without subtle side effects.

martin-cs:
I realise this is easy for me to say as 1. it's not me doing the work and 2. it's not me in need of the urgent fix but... +1 agree with @peterschrammel .  This has been a mess for too long.  I will leave it up to others (those that 1 and 2 apply to) to decide how this PR should relate to that work.

PR #5638, opened 2020-11-30
Refactor value set dereference debug output

hannes-steffenhagen-diffblue:
This moves the output functionality for `--show-points-to-sets` out of the main logic for `value_set_dereference` and removes the `#ifdef DEBUG` output which is essentially the same information as that flag gives. This makes it somewhat easier to follow along with the logic of the function.

### DIFF
--- a/src/pointer-analysis/value_set_dereference.cpp
+++ b/src/pointer-analysis/value_set_dereference.cpp
@@ -61,6 +61,42 @@ static bool should_use_local_definition_for(const exprt &expr)
   return false;
 }
 
+static json_objectt value_set_dereference_stats_to_json(
+  const exprt &pointer,
+  const std::vector<exprt> &points_to_set,
+  const std::vector<exprt> &retained_values,
+  const exprt &value)
+{
+  json_objectt json_result;
+  json_result["Pointer"] = json_stringt{format_to_string(pointer)};
+  json_result["PointsToSetSize"] =
+    json_numbert{std::to_string(points_to_set.size())};
+
+  json_arrayt points_to_set_json;
+  for(const auto &object : points_to_set)
+  {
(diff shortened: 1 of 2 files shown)


peterschrammel (approved):

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -61,6 +61,43 @@ static bool should_use_local_definition_for(const exprt &expr)
   return false;
 }
 
+static void do_display_points_to_sets(
```
chrisr-diffblue: Minor quibble about the naming :-) Its not the most descriptive name in the world, so perhaps something like `display_points_to_sets_json`. No major preference though, so just take it as a suggestion but nothing more than that.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -61,6 +61,43 @@ static bool should_use_local_definition_for(const exprt &expr)
...
+  const messaget &log,
```
chrisr-diffblue: And another small quibble around the architecture here... Should this function take a `messaget`, or should it return a `json_result` and leave it to the caller to do what it will with that? Obviously this choice has an impact on the naming of the function. In the implementation here, this function has two purposes - formatting json, then writing it to a (provided) output destination. Alternatively, it could simply focus on the formatting aspect, and then make the caller just do `log.status() << this_function(blah)` 
Again, I'm honestly not sure which is the best approach, and I'm happy with either - just floating it as an option to consider.

chrisr-diffblue (approved):
Happy for this to go in as is - and I agree, it does make this easier to read!

chrisr-diffblue (approved):
Looks good :-)

PR #5639, opened 2020-11-30
Value set dereference misc refactor

hannes-steffenhagen-diffblue:
Various miscellaneous refactorings to value_set_dereference. Goal here is mainly to make the logic easier to follow.

### DIFF
--- a/src/pointer-analysis/value_set_dereference.cpp
+++ b/src/pointer-analysis/value_set_dereference.cpp
@@ -184,40 +184,7 @@ exprt value_set_dereferencet::dereference(
 
   if(may_fail)
   {
-    // first see if we have a "failed object" for this pointer
-
-    exprt failure_value;
-
-    if(
-      const symbolt *failed_symbol =
-        dereference_callback.get_or_create_failed_symbol(pointer))
-    {
-      // yes!
-      failure_value=failed_symbol->symbol_expr();
-      failure_value.set(ID_C_invalid_object, true);
-    }
-    else
-    {
-      // else: produce new symbol
-      symbolt &symbol = get_fresh_aux_symbol(
-        type,
-        "symex",
-        "invalid_object",
-        pointer.source_location(),
-        language_mode,
-        new_symbol_table);
-
-      // make it a lvalue, so we can assign to it
-      symbol.is_lvalue=true;
-
-      failure_value=symbol.symbol_expr();
-      failure_value.set(ID_C_invalid_object, true);
-    }
-
-    valuet value;
-    value.value=failure_value;
-    value.pointer_guard=true_exprt();
-    values.push_front(value);
+    values.push_front(get_failure_value(pointer, type));
   }
 
   // now build big case split, but we only do "good" objects
@@ -246,6 +213,45 @@ exprt value_set_dereferencet::dereference(
   }
 
   return result_value;
+}
+
+value_set_dereferencet::valuet value_set_dereferencet::get_failure_value(
+  const exprt &pointer,
+  const typet &type)
+{
+  // first see if we have a "failed object" for this pointer
+  exprt failure_value;
+
+  if(
+    const symbolt *failed_symbol =
+      dereference_callback.get_or_create_failed_symbol(pointer))
+  {
+    // yes!
+    failure_value = failed_symbol->symbol_expr();
+    failure_value.set(ID_C_invalid_object, true);
+  }
+  else
+  {
+    // else: produce new symbol
+    symbolt &symbol = get_fresh_aux_symbol(
+      type,
+      "symex",
+      "invalid_object",
+      pointer.source_location(),
+      language_mode,
+      new_symbol_table);
+
+    // make it a lvalue, so we can assign to it
+    symbol.is_lvalue = true;
+
+    failure_value = symbol.symbol_expr();
+    failure_value.set(ID_C_invalid_object, true);
+  }
+
+  valuet result{};
+  result.value = failure_value;
+  result.pointer_guard = true_exprt();
+  return result;
 }
 
 /// Check if the two types have matching number of ID_pointer levels, with
--- a/src/pointer-analysis/value_set_dereference.cpp
+++ b/src/pointer-analysis/value_set_dereference.cpp
@@ -172,23 +172,11 @@ exprt value_set_dereferencet::dereference(
       return build_reference_to(value, compare_against_pointer, ns);
     });
 
-  // can this fail?
-  bool may_fail;
-
-  if(values.empty())
-  {
-    may_fail=true;
-  }
-  else
-  {
-    may_fail=false;
-    for(std::list<valuet>::const_iterator
-        it=values.begin();
-        it!=values.end();
-        it++)
-      if(it->value.is_nil())
-        may_fail=true;
-  }
+  const bool may_fail =
+    values.empty() ||
+    std::any_of(values.begin(), values.end(), [](const valuet &value) {
+      return value.value.is_nil();
+    });
 
   if(may_fail)
   {
--- a/src/pointer-analysis/value_set_dereference.cpp
+++ b/src/pointer-analysis/value_set_dereference.cpp
@@ -34,6 +34,8 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/simplify_expr.h>
 #include <util/ssa_expr.h>
 
+#include <deque>
+
 /// Returns true if \p expr is complicated enough that a local definition (using
 /// a let expression) is preferable to repeating it, potentially many times.
 /// Of course this is just a heuristic -- currently we allow any expression that
@@ -167,10 +169,12 @@ exprt value_set_dereferencet::dereference(
     compare_against_pointer = fresh_binder.symbol_expr();
   }
 
-  std::list<valuet> values =
-    make_range(retained_values).map([&](const exprt &value) {
-      return build_reference_to(value, compare_against_pointer, ns);
-    });
+  auto values =
+    make_range(retained_values)
+      .map([&](const exprt &value) {
+        return build_reference_to(value, compare_against_pointer, ns);
+      })
+      .collect<std::deque<valuet>>();
 
   const bool may_fail =
     values.empty() ||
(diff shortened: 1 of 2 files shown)


peterschrammel (commented):
Would be great to have unit tests for that to check that the refactoring is actually correct.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -167,10 +169,12 @@ exprt value_set_dereferencet::dereference(
...
-  std::list<valuet> values =
-    make_range(retained_values).map([&](const exprt &value) {
-      return build_reference_to(value, compare_against_pointer, ns);
-    });
+  auto values =
+    make_range(retained_values)
+      .map([&](const exprt &value) {
+        return build_reference_to(value, compare_against_pointer, ns);
+      })
+      .collect<std::deque<valuet>>();
```
hannes-steffenhagen-diffblue, line 183: This one is the one exception here, this doesn’t make it easier to follow it just bothers me to see a `std::list` if we don’t need the flexibility.

hannes-steffenhagen-diffblue:
@peterschrammel I agree. I am still in the process of understanding this file though, I was editing while I was reading through it. I thought these would be covered by [unit/util/pointer-analysis/value_set.cpp](https://github.com/diffblue/cbmc/blob/fee703e3a5ee942de052e6f1bcab316ef46f0367/unit/pointer-analysis/value_set.cpp)

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -184,40 +184,7 @@ exprt value_set_dereferencet::dereference(
...
-
-      failure_value=symbol.symbol_expr();
-      failure_value.set(ID_C_invalid_object, true);
-    }
-
-    valuet value;
-    value.value=failure_value;
-    value.pointer_guard=true_exprt();
-    values.push_front(value);
+    values.push_front(get_failure_value(pointer, type));
```
hannes-steffenhagen-diffblue, line 193: This `push_front` is the only reason this is a `deque`. I haven’t checked yet if there’s any significance to the order here, if not I’d just change this to a `vector` and `push_back` instead.

**EDIT** Actually, just double checked and I was misremembering how `deque` works. I think this should probably just be a `vector`. Looks like expression is only appended at the front so it ends up at the end in the latter expression, but this could easily be reversed.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -224,16 +224,14 @@ exprt value_set_dereferencet::dereference(
...
-      it != values.end();
-      it++)
+  for(const auto &value : values)
   {
-    if(it->value.is_not_nil())
+    if(value.value.is_not_nil())
     {
       if(result_value.is_nil()) // first?
-        result_value = it->value;
+        result_value = value.value;
```
NlightNFotis, line 205: `value.value` reads a bit weird in all places. Is there perhaps a better name that could apply? 🤔

src/pointer-analysis/value_set_dereference.h
```diff
@@ -112,6 +112,9 @@ class value_set_dereferencet final
   const bool exclude_null_derefs;
   const messaget &log;
   valuet get_failure_value(const exprt &pointer, const typet &type);
+  exprt handle_dereference_base_case(
```
NlightNFotis, line 115: Should there be some documentation for the new method? 🤔

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -248,6 +215,45 @@ exprt value_set_dereferencet::dereference(
...
 
+value_set_dereferencet::valuet value_set_dereferencet::get_failure_value(
+  const exprt &pointer,
+  const typet &type)
+{
+  // first see if we have a "failed object" for this pointer
+  exprt failure_value;
+
+  if(
+    const symbolt *failed_symbol =
```
NlightNFotis, line 232: I really, *really* do not like assignment statements inside the condition of if statements. I know you haven't introduced this, but could you perhaps clean it up? 🙏

NlightNFotis (commented):
Looks better already, thanks 👍 

Just some minor comments.

src/pointer-analysis/value_set_dereference.cpp
hannes-steffenhagen-diffblue: @NlightNFotis Actually, I really prefer this style so I wouldn’t consider that a cleanup. It’s a bargain bin pattern match.

src/pointer-analysis/value_set_dereference.cpp
thomasspriggs: :pick: It would be good to have some performance checks to make sure the change from `std::list` -> `std::deque` doesn't have any performance issues.

thomasspriggs: Do you prefer the explicit call to `collect` over the implicit conversion which was used before?

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -172,23 +172,11 @@ exprt value_set_dereferencet::dereference(
...
-        it!=values.end();
-        it++)
-      if(it->value.is_nil())
-        may_fail=true;
-  }
+  const bool may_fail =
+    values.empty() ||
+    std::any_of(values.begin(), values.end(), [](const valuet &value) {
+      return value.value.is_nil();
+    });
```
thomasspriggs, line 189: :+1: I find this is much more readable than what we had before.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -222,31 +222,32 @@ exprt value_set_dereferencet::dereference(
...
+        result_value = it->value;
       else
-        value=if_exprt(it->pointer_guard, it->value, value);
+        result_value = if_exprt(it->pointer_guard, it->value, result_value);
     }
   }
 
   if(compare_against_pointer != pointer)
-    value = let_exprt(to_symbol_expr(compare_against_pointer), pointer, value);
+    result_value =
```
thomasspriggs, line 212: :pick: Can we have some `{}` brackets, if this is being reformatted across multiple lines?

src/pointer-analysis/value_set_dereference.cpp
thomasspriggs: Maybe `value_set_dereferencet::valuet` should be renamed `value_set_dereferencet::reference_tot`, given that it is the return type of `build_reference_to`. The instances could then be named `reference_to`.

src/pointer-analysis/value_set_dereference.cpp
thomasspriggs: Defining and assigning variables in `if` statements is great, because it restricts the scope, such that it is only in scope if the value is not null (or not empty if using optional).

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -224,16 +224,14 @@ exprt value_set_dereferencet::dereference(
 
   exprt result_value = nil_exprt{};
 
-  for(std::deque<valuet>::const_iterator it = values.begin();
-      it != values.end();
-      it++)
+  for(const auto &value : values)
```
thomasspriggs, line 200: :pick: I might have used `valuet` here rather than `auto`. Having the type especially when it is that short can make things more readable when reviewing in github without a full IDE.

src/pointer-analysis/value_set_dereference.cpp
thomasspriggs: :pick: Consider removing the `type` variable and parameter and just using `pointer.type().subtype()` inside `get_failure_value`.

src/pointer-analysis/value_set_dereference.h
```diff
@@ -111,6 +111,7 @@ class value_set_dereferencet final
   /// disregard an apparent attempt to dereference NULL
   const bool exclude_null_derefs;
   const messaget &log;
+  valuet get_failure_value(const exprt &pointer, const typet &type);
```
thomasspriggs, line 114: :pick: Missing doxygen.

src/pointer-analysis/value_set_dereference.cpp
```diff
@@ -141,7 +141,13 @@ exprt value_set_dereferencet::dereference(
     }
   }
 
-  // type of the object
+  return handle_dereference_base_case(pointer, display_points_to_sets);
+}
+
+exprt value_set_dereferencet::handle_dereference_base_case(
```
thomasspriggs, line 147: :pick: I think I would have done this split the opposite way round. As in leave the base case in `dereference`, but factor out the special cases for `ID_if` and `ID_typecast`. That way you don't end up with quite such a verbose function name :wink:
:bulb: Your "bargin bin pattern match" could be used to split out overloads. For example -
```
  if(const auto if_expr = expr_try_dynamic_cast<if_exprt>(pointer))
    return dereference(*if_expr, display_points_to_sets);
```
in place of -
```
  if(pointer.id()==ID_if)
  {
    const if_exprt &if_expr=to_if_expr(pointer);
```

thomasspriggs (commented):
The refactoring in this PR looks to be sound as far as I can see. However I'd like to work out exactly what tests cover the refactoring before I approve. Note that the PR check list states "Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).", which doesn't look to have been done.

src/pointer-analysis/value_set_dereference.cpp
hannes-steffenhagen-diffblue: Didn’t put this here because the name `valuet` doesn’t really add any information the names `value` and `values` don’t also provide.

src/pointer-analysis/value_set_dereference.cpp
hannes-steffenhagen-diffblue: @thomasspriggs The bodies of the branches could also be factored out, I just didn't bother for now since they’re short. The reason why I factored this out was to separate the decision whether or not to recurse from the handling of the case where we don’t recurse, because these are different... functions.

src/pointer-analysis/value_set_dereference.cpp
martin-cs: I'm with @NlightNFotis on this one, intentional assignments in conditions make it much harder to check for `=` vs. `==` bugs.

martin-cs: This feels rather like the kind of thing that should be in the coding standard.

martin-cs (approved):
I think the refactorings do not affect the functionality.  Approve subject to other reviewers comments being addressed.

PR #5640, opened 2020-12-02
abstract_objectt interface

jezhiggins:
Tightens up the abstract_objectt interface 

* removes public `abstract_objectt::clone` member function. It's unnecessary as abstract_objectts are immutable after construction
* made `abstract_objectt::type()` member function virtual. `context_abstract_object` wanted to override it, but because it wasn't virtual was just hiding it.
* pass all immutable parameters to `abstract_objectt::write` as const-refs. There were a couple of parameters passed by const value. Corrected implementation member functions in derived classes to match. 

There are no functional changes in this PR.

### DIFF
--- a/src/analyses/variable-sensitivity/abstract_object.h
+++ b/src/analyses/variable-sensitivity/abstract_object.h
@@ -121,7 +121,7 @@ public:
   /// Get the real type of the variable this abstract object is representing.
   ///
   /// \return The program type this abstract object represents
-  const typet &type() const;
+  virtual const typet &type() const;
 
   /// Find out if the abstract object is top
   ///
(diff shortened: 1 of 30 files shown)


src/analyses/variable-sensitivity/abstract_object.h
```diff
@@ -121,7 +121,7 @@ class abstract_objectt : public std::enable_shared_from_this<abstract_objectt>
   /// Get the real type of the variable this abstract object is representing.
   ///
   /// \return The program type this abstract object represents
-  const typet &type() const;
+  virtual const typet &type() const;
```
martin-cs, line 124: Good catch

martin-cs (approved):
Looks good to me.

martin-cs:
CI failures seem to be code coverage which are a fraction of a percent out.  I'm not sure they are a great concern.

jezhiggins:
> CI failures seem to be code coverage which are a fraction of a percent out. I'm not sure they are a great concern.

I think that's an artefact of there being ever so slightly less code than before.

PR #5644, opened 2020-12-09
abstract_environmentt::eval rework

jezhiggins:
Refactoring work centred on abstract_environmentt::eval, trying to reduce the volume of special case handling. 

Moved implementation of indexing, member access, and pointer dereferencing out into the target object types. This allowed us to combine several of the if-ladder cases together.


regression/goto-analyzer/simple-union-access/test.desc
```diff
@@ -0,0 +1,15 @@
+CORE
+union.c
+--variable-sensitivity --vsd --verify
```
martin-cs: I think `--vsd` and `--variable-sensitivity` should do the same thing.  One was for backwards compatibility.

regression/goto-analyzer/simple-union-access/test.desc
```diff
@@ -0,0 +1,15 @@
...
+^EXIT=0$
+^SIGNAL=0$
+^\[main.assertion.1\] line 13 x.a==0: UNKNOWN$
+^\[main.assertion.2\] line 14 x.a==1: UNKNOWN$
+^\[main.assertion.3\] line 15 x.b==0: UNKNOWN$
+^\[main.assertion.4\] line 16 x.b==1: UNKNOWN$
+^\[main.assertion.5\] line 19 x.a==0: UNKNOWN$
+^\[main.assertion.6\] line 20 x.a==1: UNKNOWN$
+^\[main.assertion.7\] line 21 x.b==0: UNKNOWN$
+^\[main.assertion.8\] line 22 x.b==1: UNKNOWN$
```
martin-cs, line 13: With a "most recently written" abstraction for unions it should be possible to turn about half of these into successes but for now all unknown is the best that can be done.

unit/analyses/variable-sensitivity/eval-member-access.cpp
```diff
@@ -0,0 +1,111 @@
...
+);
+
+exprt fetch_element(
+  int index,
+  exprt &array,
+  abstract_environmentt &environment,
+  namespacet &ns
+);
+
+SCENARIO(
```
martin-cs, line 17: Unit tests are good.  In theory much of VSD should be testable via unit tests so having more is a good thing.

unit/analyses/variable-sensitivity/eval-member-access.cpp
```diff
@@ -0,0 +1,158 @@
...
+{
+  return (i == TOP) ? top_expression() : from_integer(i, integer_typet());
+}
+
+exprt top_expression()
+{
+  auto top_value =
+    std::make_shared<abstract_valuet>(integer_typet(), true, false);
+  return top_value->to_constant();
+}
```
martin-cs: Missing new line.  Only worth fixing if you have other things to fix.

src/analyses/variable-sensitivity/abstract_enviroment.cpp
```diff
@@ -68,35 +74,12 @@ abstract_environmentt::eval(const exprt &expr, const namespacet &ns) const
...
-  else if(simplified_id == ID_index)
-  {
-    index_exprt index_expr(to_index_expr(simplified_expr));
-    abstract_object_pointert array_abstract_object =
-      eval(index_expr.array(), ns);
-
-    return array_abstract_object->read(*this, index_expr, ns);
-  }
-  else if(simplified_id == ID_array)
+  else if(
```
martin-cs, line 77: Yes; these are all basically the same constant case.

src/analyses/variable-sensitivity/abstract_enviroment.cpp
```diff
@@ -51,14 +56,15 @@ abstract_environmentt::eval(const exprt &expr, const namespacet &ns) const
       return found_symbol_value;
     }
   }
-  else if(simplified_id == ID_member)
+  else if(
```
martin-cs, line 59: Can this case be folded into the general `else` case?

src/analyses/variable-sensitivity/array_abstract_object.cpp
```diff
@@ -35,6 +35,21 @@ array_abstract_objectt::array_abstract_objectt(
...
 
+abstract_object_pointert array_abstract_objectt::expression_transform(
+  const exprt &expr,
+  const std::vector<abstract_object_pointert> &operands,
+  const abstract_environmentt &environment,
+  const namespacet &ns) const
+{
+  if(expr.id() == ID_index)
+  {
+    return read(environment, expr, ns);
```
martin-cs: Given that `read` just calls `read_index` can you call `read_index` directly here (and the corresponding functions in struct, pointer, etc.) and then remove `read` from the interface of `abstract_objectt` completely.

martin-cs (changes requested):
Yes; this is directly in the right direction and does simplify things.

Request changes because it would be good if it could go a little bit further and remove `abstract_objectt::read` completely by calling the internal functions `read_*` of the relevant classes.

martin-cs:
Reflecting on this, I think the only uses of `read_index`, `read_component`, etc. are in `read` so you could inline the contents of these into the `expression_transform` and remove everything `read` related.

martin-cs:
Better -- thank you.

regression/goto-analyzer/simple-union-access/test.desc
martin-cs: Please can you go down to just one of these two?

martin-cs (approved):
One tweak to the union regression test and should be good to merge.

PR #5646, opened 2020-12-10
CMake builds: set a default build type (Release)

tautschnig:
In absence of a default build type, CMake would not use any pre-defined
build configuration. As a consequence, any user running CMake without
setting -DCMAKE_BUILD_TYPE would end up with builds without
optimisation, and thus (surprisingly!) poor performance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,12 @@
 cmake_minimum_required(VERSION 3.2)
 
+# Build a Release version by default (default build flags for each build type
+# are configured below).
+if (NOT EXISTS ${CMAKE_BINARY_DIR}/CMakeCache.txt)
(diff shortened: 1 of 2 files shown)


martin-cs (approved):
This seems like a reasonable change for a reasonable reason.  I don't know CMake well enough to know if this is the best way of doing it.  Can we simplify the build instructions using CMake?

kroening (approved):

hannes-steffenhagen-diffblue (approved):
This is harmless, but normally a user should specify a build type and if our documentation doesn't say so we should fix that.

tautschnig:
> This is harmless, but normally a user should specify a build type and if our documentation doesn't say so we should fix that.

We didn't clearly state this in our documentation. I have now tried to fix that as well.

PR #5649, opened 2020-12-10
Automatically create release tags via GitHub actions

tautschnig:
~The action is currently marked as DRY_RUN: true to perform testing, and
will be enabled in a subsequent PR.~ Edit: I have performed testing in my fork (see https://github.com/tautschnig/cbmc/runs/1533820084), and am thus inclined to immediately enable this over here.

### DIFF
--- a/.github/workflows/create-release-tag.yaml
+++ b/.github/workflows/create-release-tag.yaml
@@ -1,0 +1,34 @@
+name: Create release tag
+on:
+  push:
+    branches: [ develop ]
+
+jobs:
+  create-tag:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: '0'
+    - name: Compute new tag
+      uses: anothrNick/github-tag-action@1.33.0
+      id: dry_run
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        DEFAULT_BUMP: none
+        RELEASE_BRANCHES: develop
+        VERBOSE: false
+        DRY_RUN: true
+    - if: ${{ steps.dry_run.new_tag != "" }}
+      run: |
+        v=$(grep ^CBMC_VERSION src/config.inc | perl -p -e 's/^CBMC_VERSION\s*=\s*//')
+        echo "CBMC_VERSION: $v"
+        echo "New tag: ${{ steps.dry_run.new_tag }}"
+        test "cbmc-$v" == ${{ steps.dry_run.new_tag }}
+    - name: Bump version and push tag
+      uses: anothrNick/github-tag-action@1.33.0
+      env:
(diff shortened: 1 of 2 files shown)


.github/workflows/create-release-tag.yaml
```diff
@@ -0,0 +1,19 @@
...
+  create-tag:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: '0'
+    - name: Bump version and push tag
+      uses: anothrNick/github-tag-action@1.33.0
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
hannes-steffenhagen-diffblue, line 31: Have you checked if the default `GITHUB_TOKEN` has the rights to create a tag? It not being allowed to do some things is something that has bitten me in the past, that's why I’m asking.

.github/workflows/create-release-tag.yaml
```diff
@@ -0,0 +1,19 @@
...
+        DEFAULT_BUMP: none
+        RELEASE_BRANCHES: develop
+        VERBOSE: false
```
piotr-grabalski, line 34: Does this read the latest version from the existing tags?
If not we may need INITIAL_VERSION too:
> INITIAL_VERSION (optional) - Set initial version before bump. Default 0.0.0.

Probably something like: `INITIAL_VERSION: 5.20.1`?

tautschnig: Supposedly the bash script behind this GitHub action does check for existing tags. But it may be worth flipping `DRY_RUN` back on to make sure this does actually work as expected.

.github/workflows/create-release-tag.yaml
tautschnig: Good call out. I don't know whether this works, but also I'm not sure how to figure it out other than trying it out? (It seems safe to do so: in the worst case, it does not create a tag, and we'd have to go back to the previous, manual approach.)

.github/workflows/create-release-tag.yaml
tautschnig: I have now added a commit to keep in in `DRY_RUN` mode.

tautschnig:
A further comment: we should likely wait for https://github.com/anothrNick/github-tag-action/pull/95 to be merged to maintain the `cbmc-` prefix.

NlightNFotis:
Hi Michael, I'd love to get this in and experiment with this a bit - it would save us quite a bit of time!

Are you happy to take this forward (rebase on `develop`) or if you don't have the bandwidth for this, are you happy with me taking this over?*

*: (I'll close this PR and make a new one, but `cherry-pick` your commits on top of the new branch, and any changes by me will go in another commit, so your authorship is going to be preserved)

tautschnig:
Closing: automatic tagging won't save much time as the tag can be produced at the same time as publishing the changelog.

PR #5657, opened 2020-12-14
goto-symex: nil array size must not have a type added

tautschnig:
An expression has a type() member, which the nil irept lacks. Trying to
access (in a non-const context) the type() member would thus create it,
which in turn means that it no longer compares equal to a nil_exprt.  As
SSA renaming did access the type() member in such a way, the type of an
array without specified size would no longer compare equal to the irept
describing the type as generated by the C front-end, which in turn made
simplification fail.

The problem was surfaced by running cbmc --unwind 2 --pointer-check
--bounds-check on
c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-32_7a-drivers--gpu--drm--i915--i915.ko-entry_point.cil.out.i
from SV-COMP.

### DIFF
--- a/src/goto-symex/renaming_level.cpp
+++ b/src/goto-symex/renaming_level.cpp
@@ -263,6 +263,10 @@ bool check_renaming(const exprt &expr)
     if(to_ssa_expr(expr).get_original_expr().type() != type)
       return true;
   }
+  else if(expr.id() == ID_nil)
+  {
+    return expr != nil_exprt{};
+  }
   else
   {
     forall_operands(it, expr)
--- a/src/util/std_types.cpp
+++ b/src/util/std_types.cpp
@@ -18,6 +18,19 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "std_expr.h"
 #include "string2int.h"
 
+void array_typet::check(const typet &type, const validation_modet vm)
+{
+  PRECONDITION(type.id() == ID_array);
+  const array_typet &array_type = static_cast<const array_typet &>(type);
+  if(array_type.size().is_nil())
+  {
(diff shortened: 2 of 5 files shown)


src/goto-symex/renaming_level.cpp
```diff
@@ -263,6 +263,10 @@ bool check_renaming(const exprt &expr)
     if(to_ssa_expr(expr).get_original_expr().type() != type)
       return true;
   }
+  else if(expr.id() == ID_nil)
+  {
```
kroening, line 267: Is that still needed after the change in `goto_symex_state.cpp`?

src/util/std_types.cpp
```diff
@@ -18,6 +18,17 @@ Author: Daniel Kroening, kroening@kroening.com
...
 
+void array_typet::check(const typet &type, const validation_modet vm)
+{
+  PRECONDITION(type.id() == ID_array);
+  const array_typet &array_type = static_cast<const array_typet &>(type);
+  if(array_type.size().is_nil())
+  {
+    DATA_CHECK(
+      vm, array_type.size() == nil_exprt{}, "array size must be an expression");
+  }
```
kroening, line 31: The message in that check could be clearer.

src/goto-symex/renaming_level.cpp
tautschnig: This is just the sanity check (when using `--validate-ssa-equation`) - hopefully it never fails...

src/util/std_types.cpp
tautschnig: How about the updated message?

kroening (approved):

PR #5659, opened 2020-12-14
value sets: do not try to access components of an empty struct/union

tautschnig:
This avoids a segmentation fault while running on
c/ldv-linux-4.2-rc1/linux-4.2-rc1.tar.xz-32_7a-drivers--staging--lustre--lustre--mdc--mdc.ko-entry_point.cil.out.i
with options --pointer-check --bounds-check --unwind 2. This benchmark
includes the empty `union __anonunion_u_rpc_386`. While the patch does
address the problem on this benchmark, I failed to come up with a small
regression test.

### DIFF
--- a/regression/cbmc/empty_compound_type1/main.c
+++ b/regression/cbmc/empty_compound_type1/main.c
@@ -1,0 +1,19 @@
+union a {
+};
+struct
+{
+  union a;
+} b;
+struct c
+{
+  int cli;
+};
+void e(struct c *f)
+{
+  int a = f->cli;
+}
+main()
+{
+  // pass a pointer to an incompatible type
+  e(&b);
+}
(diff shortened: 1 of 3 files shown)


hannes-steffenhagen-diffblue (approved):
Could you come up with a test case for this for our regression tests?

tautschnig:
> Could you come up with a test case for this for our regression tests?

I tried, and I'll keep trying, but it seems one has to craft code that does invalid pointer dereferences and perhaps other questionable things.

martin-cs (approved):
Change seems reasonable.  I agree with @hannes-steffenhagen-diffblue that a test case would be good.  Have you tried https://github.com/csmith-project/creduce ?

tautschnig:
> Have you tried https://github.com/csmith-project/creduce ?

Ah, of course! Thank you for reminding me!! That did the trick, a tiny regression test is now included. (And, yes, it does involve invalid pointer dereferencing.)

martin-cs (approved):
Much better!  Thanks.

PR #5661, opened 2020-12-14
C library: Windows defines time to _time64 (or _time32)

tautschnig:
Windows uses time as a wrapper to _time64 by default, or _time32 when
_USE_32BIT_TIME_T is defined. See
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/time-time32-time64?view=msvc-160.

This is a fix-up to c6f1ed5d37, which introduced a broken regression
test for Windows because of the then-undefined _time64 function.

### DIFF
--- a/src/ansi-c/library/time.c
+++ b/src/ansi-c/library/time.c
@@ -17,6 +17,48 @@ time_t time(time_t *tloc)
   return res;
 }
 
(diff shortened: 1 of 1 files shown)


kroening (approved):

martin-cs (approved):
Seems reasonable.

PR #5664, opened 2020-12-15
Do not fail an internal invariant when initialising an empty union

tautschnig:
The C front-end should instead report a meaningful error message. The
test is based on an example originally created by CSmith and then
reduced using C-Reduce.

### DIFF
--- a/regression/ansi-c/Union_Initialization2/main.c
+++ b/regression/ansi-c/Union_Initialization2/main.c
@@ -1,0 +1,5 @@
+union a {
+} b()
+{
+  union a c = {0};
+}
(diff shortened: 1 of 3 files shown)


kroening (approved):

kroening:
The message is a bit misleading -- shouldn't that index be always zero?

kroening:
So I am thinking s.th. like "attempt to initialize an empty union"

martin-cs (approved):

tautschnig:
> So I am thinking s.th. like "attempt to initialize an empty union"

Thank you, I've done a slight variation on this.

PR #5669, opened 2020-12-15
Cleanup side effects in declaration initialisers

tautschnig:
Declarations with initialisers generate a declaration and an assignment.
The side effect in the rhs of the assignment was never cleaned up,
resulting in a failing invariant during symex for the regression test
included.

The test was built by C-Reduce on a example originating from CSmith.

### DIFF
--- a/regression/cbmc/Sideeffects7/main.c
+++ b/regression/cbmc/Sideeffects7/main.c
@@ -1,0 +1,20 @@
+int foo()
+{
+  __CPROVER_assert(0, "");
+}
+
+int main()
+{
+  struct
+  {
+    int a : 2;
+  } b;
+  int c;
+  short d;
+  // this is to make sure the fix for the below assignment does not introduce a
+  // new bug
+  d |= b.a = c >= 0;
+
+  // we previously weren't able to handle the below
+  short a = (a ^= 0) || foo();
+}
--- a/src/goto-programs/goto_convert_side_effect.cpp
+++ b/src/goto-programs/goto_convert_side_effect.cpp
@@ -128,8 +128,24 @@ void goto_convertt::remove_assignment(
   // revert assignment in the expression to its LHS
   if(result_is_used)
   {
-    exprt lhs;
-    lhs.swap(to_binary_expr(expr).op0());
+    exprt lhs = to_binary_expr(expr).op0();
+    // assign_* statements can have an lhs operand with a different type than
+    // that of the overall expression, because of integer promotion (which may
+    // have introduced casts to the lhs).
+    if(expr.type() != lhs.type())
+    {
+      // Skip over those type casts, but also
+      // make sure the resulting expression has the same type as before.
+      DATA_INVARIANT(
+        lhs.id() == ID_typecast,
+        id2string(expr.id()) +
+          " expression with different operand type expected to have a "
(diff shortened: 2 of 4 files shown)


martin-cs (approved):
Seems reasonable.

src/goto-programs/goto_convert_side_effect.cpp
```diff
@@ -128,8 +128,12 @@ void goto_convertt::remove_assignment(
...
   {
-    exprt lhs;
-    lhs.swap(to_binary_expr(expr).op0());
+    // assign_* statements can have an lhs operand with a different type than
+    // that of the overall expression, because of integer promotion (which may
+    // have introduced casts to the lhs). skip over those type casts, but also
+    // make sure the resulting expression has the same type as before.
+    exprt lhs = typecast_exprt::conditional_cast(
+      skip_typecast(to_binary_expr(expr).op0()), expr.type());
     expr.swap(lhs);
```
kroening, line 149: I am a bit worried about the `skip_typecast`, since that may remove any number of casts.

Say think of `(int)(char)i`, where `i` is an integer.

tautschnig: Thank you for calling this out! I have now added sanity checks and refrained from using `skip_typecast`.

regression/cbmc/Sideeffects7/main.c
```diff
@@ -0,0 +1,15 @@
...
+  } b;
+  int c;
+  short d;
+  // this is to make sure the fix for the below assignment does not introduce a
+  // new bug
+  d |= b.a = c >= 0;
+
+  // we previously weren't able to handle the below
+  short a = (a ^= 0) || __assert_fail("", 2, __PRETTY_FUNCTION__);
+}
```
kroening, line 20: I would replace the architecture-specific `__assert_fail` by some other side effect, say `x++` or the like.

src/goto-programs/goto_convert_side_effect.cpp
```diff
@@ -128,8 +128,17 @@ void goto_convertt::remove_assignment(
...
-    lhs.swap(to_binary_expr(expr).op0());
+    exprt lhs = to_binary_expr(expr).op0();
+    // assign_* statements can have an lhs operand with a different type than
+    // that of the overall expression, because of integer promotion (which may
+    // have introduced casts to the lhs).
+    if(expr.type() != lhs.type())
+    {
+      // Skip over those type casts, but also
+      // make sure the resulting expression has the same type as before.
+      PRECONDITION(to_typecast_expr(lhs).op().type() == expr.type());
```
kroening: I would add a DATA_INVARIANT here to check that the typecast is there, and turn that PRECONDITION into a DATA_INVARIANT.

regression/cbmc/Sideeffects7/main.c
tautschnig: Done (I've added a call to a newly added function, which in turn calls `__CPROVER_assert`).

src/goto-programs/goto_convert_side_effect.cpp
tautschnig: Done.

kroening (approved):

PR #5671, opened 2020-12-16
Return value removal: handle missing declarations more gracefully

tautschnig:
If a function is used before it is defined, a signature of int f(void)
is assumed. Then trying to use the (possibly non-existent) return value
fails during return-statement removal. In such cases, just assume a
non-deterministic value is being returned.

Found by running C-Reduce on a CSmith-generated example.

### DIFF
--- a/regression/cbmc/return7/main.c
+++ b/regression/cbmc/return7/main.c
@@ -1,0 +1,18 @@
+void a()
+{
+  // Uses the implicit signature of undefined functions: int c(void)
+  int b;
+  b = c();
+  __CPROVER_assert(b == 0, "expected to fail");
+}
+void c(void)
+{
+  // Actually... don't return anything
+  // So the results will be non-deterministic
+}
+
+int main(int argc, char **argv)
+{
+  a();
+  return 0;
+}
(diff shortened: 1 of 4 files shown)


martin-cs (approved):
I feel like the test could be further developed to show what happens when you actually call this / where this goes wrong.  The "VERIFICATION SUCCESSFUL" is technically correct but I feel like it is almost a bit in bad faith.  How about:

```
void a()
{
  int b = c();  // Uses the implicit signature of undefined functions: int c(void)
  assert(b == 0);
}
void c(void)
{
  // Actually... don't return anything
  // So the results will be non-deterministic
}

  a();
  return 0;
}
```

tautschnig:
> I feel like the test could be further developed to show what happens when you actually call this / where this goes wrong. The "VERIFICATION SUCCESSFUL" is technically correct but I feel like it is almost a bit in bad faith. How about:
> 
> [...]

Thank you - shamelessly took this code verbatim to replace the regression test.

peterschrammel (approved):

kroening:
Should this become a hard error in C99/C11? Clang won't have it.

PR #5672, opened 2020-12-16
dump-c: when removing const-ness, also handle bit fields

tautschnig:
We need to remove the const qualifier when the assignment is no longer
performed as part of the declaration. This worked for various types
already, but bit fields had not been considered.

### DIFF
--- a/regression/goto-instrument/chain.sh
+++ b/regression/goto-instrument/chain.sh
@@ -38,6 +38,7 @@ elif echo $args | grep -q -- "--dump-c-type-header" ; then
   cat "${name}-mod.gb"
   mv "${name}.gb" "${name}-mod.gb"
 elif echo $args | grep -q -- "--dump-c" ; then
+  cat "${name}-mod.gb"
   mv "${name}-mod.gb" "${name}-mod.c"
 
   if [[ "${is_windows}" == "true" ]]; then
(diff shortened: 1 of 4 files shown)


regression/goto-instrument/chain.sh
```diff
@@ -38,6 +38,7 @@ elif echo $args | grep -q -- "--dump-c-type-header" ; then
   cat "${name}-mod.gb"
   mv "${name}.gb" "${name}-mod.gb"
 elif echo $args | grep -q -- "--dump-c" ; then
+  cat "${name}-mod.gb"
```
martin-cs, line 41: Is this going to cause issues with other tests / increase the amount of output drastically?

regression/goto-instrument/dump-decl-const/test.desc
```diff
@@ -0,0 +1,16 @@
...
+^[[:space:]]*s1 = \(struct S\)\{ .* \};
+^EXIT=0$
+^SIGNAL=0$
+--
+const signed int x
+const signed int y
+const p
+--
+This test demonstrates that the constness of struct members has been removed,
+which is necessary as the initialisation is not performed in the declaration.
```
martin-cs, line 16: I guess this is just one of those unavoidable things as we split an initialisation into `DECL` and `ASSIGN` instructions.

martin-cs (approved):
Thanks for the test cases esp.!

regression/goto-instrument/chain.sh
tautschnig: Since this output is only generated when `--dump-c` is included in the test options, I think only very few cases (with small source files) will be affected. Thus, we should be fine.

PR #5674, opened 2020-12-17
C/C++ front-ends: report a meaningful error for invalid function definition

tautschnig:
The parser happily accepts the syntax shown in the regression test
included, but type checking strictly expects code instead of an
initializer list. Report a proper error message, like GCC does, instead
of failing an invariant.

Test generated using C-Reduce starting from an SV-COMP task.


martin-cs (approved):
"Giving a helpful error message rather than an assertion failure on user input" is a great look!

kroening (approved):

PR #5676, opened 2020-12-17
C/C++ front-end: do not accept void-typed objects or members

tautschnig:
We silently accepted these, as many cases would work fine in goto
programs. The included regression test, however, demonstrates where this
would fail an invariant during Boolean flattening (if run with
--pointer-check). GCC refuses void typed variables, and so should we.

Test originates from an SV-COMP task, reduced to this minimal example
using C-Reduce.

- n/a  Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.

### DIFF
--- a/regression/ansi-c/invalid_use_of_void1/main.c
+++ b/regression/ansi-c/invalid_use_of_void1/main.c
@@ -1,0 +1,21 @@
+struct a
+{
+#ifdef void_member
+  void b;
+#else
+  int b;
+#endif
+};
+struct a *ae;
+void d()
+{
+  ae->b;
+}
+#ifdef void_global
+void x;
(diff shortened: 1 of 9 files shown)


martin-cs (approved):
If you added copies of the test to `regression/cpp/` then you could get 100% diff coverage.

tautschnig:
> If you added copies of the test to `regression/cpp/` then you could get 100% diff coverage.

Indeed, done! (Which of course required me expanding my changes in the C++ front-end - testing does help, after all ;-) )

tautschnig:
#5679 needs to be merged first to make CI happy again.

kroening (approved):

PR #5677, opened 2020-12-17
_Static_assert failures should be reported in the front-end

tautschnig:
We left it to a failing invariant in goto-program conversion to detect a
failing (local) _Static_assert. Global _Static_assert statements were
not verified at all.

### DIFF
--- a/regression/cbmc-library/Float_lib1/main.c
+++ b/regression/cbmc-library/Float_lib1/main.c
@@ -43,10 +43,10 @@ int main()
 
   _Static_assert(__builtin_isnormal(DBL_MIN), "__builtin_isnormal is constant");
 
-  _Static_assert(!__builtin_isinf(0), "__builtin_isinf is constant");
+  _Static_assert(!__builtin_isinf(0.0), "__builtin_isinf is constant");
 
   _Static_assert(
-    __builtin_isinf_sign(0) == 0, "__builtin_isinf_sign is constant");
+    __builtin_isinf_sign(0.0) == 0, "__builtin_isinf_sign is constant");
 
 #endif
 
--- a/src/ansi-c/c_typecheck_code.cpp
+++ b/src/ansi-c/c_typecheck_code.cpp
@@ -15,6 +15,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/config.h>
 #include <util/expr_initializer.h>
 #include <util/range.h>
+#include <util/string_constant.h>
 
 #include "ansi_c_declaration.h"
 
@@ -102,9 +103,24 @@ void c_typecheck_baset::typecheck_code(codet &code)
   }
   else if(statement==ID_static_assert)
   {
-    assert(code.operands().size()==2);
+    PRECONDITION(code.operands().size() == 2);
+
     typecheck_expr(code.op0());
     typecheck_expr(code.op1());
+
+    implicit_typecast_bool(code.op0());
+    make_constant(code.op0());
+
+    if(code.op0().is_false())
+    {
+      // failed
+      error().source_location = code.find_source_location();
+      error() << "static assertion failed";
+      if(code.op1().id() == ID_string_constant)
+        error() << ": " << to_string_constant(code.op1()).get_value();
+      error() << eom;
+      throw 0;
+    }
   }
   else if(statement==ID_CPROVER_try_catch ||
           statement==ID_CPROVER_try_finally)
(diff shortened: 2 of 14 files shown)


martin-cs (approved):
OK but I do wonder whether we should treat these like `assert`s.  I guess the point is that they are supposed to be build time and if they fail then the code is explicitly marked as being wrong.

regression/cbmc-library/Float_lib1/main.c
```diff
@@ -43,10 +43,10 @@ int main()
 
   _Static_assert(__builtin_isnormal(DBL_MIN), "__builtin_isnormal is constant");
 
-  _Static_assert(!__builtin_isinf(0), "__builtin_isinf is constant");
+  _Static_assert(!__builtin_isinf(0.0), "__builtin_isinf is constant");
 
   _Static_assert(
-    __builtin_isinf_sign(0) == 0, "__builtin_isinf_sign is constant");
+    __builtin_isinf_sign(0.0) == 0, "__builtin_isinf_sign is constant");
```
kroening, line 49: Shouldn't that be fixed rather than changing the test?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -102,9 +103,24 @@ void c_typecheck_baset::typecheck_code(codet &code)
   }
   else if(statement==ID_static_assert)
   {
-    assert(code.operands().size()==2);
+    PRECONDITION(code.operands().size() == 2);
+
     typecheck_expr(code.op0());
     typecheck_expr(code.op1());
+
+    code.op0() = typecast_exprt::conditional_cast(code.op0(), bool_typet{});
```
kroening: That might not work, consider `implicit_typecast_bool`?

regression/cbmc-library/Float_lib1/main.c
tautschnig: Well, it's the test that was broken: running GCC on this test before the change said:
```
$ gcc regression/cbmc-library/Float_lib1/main.c
regression/cbmc-library/Float_lib1/main.c: In function ‘main’:
regression/cbmc-library/Float_lib1/main.c:46:35: error: non-floating-point argument in call to function ‘__builtin_isinf’
   _Static_assert(!__builtin_isinf(0), "__builtin_isinf is constant");
                                   ^
regression/cbmc-library/Float_lib1/main.c:46:18: error: expression in static assertion is not an integer
   _Static_assert(!__builtin_isinf(0), "__builtin_isinf is constant");
                  ^
regression/cbmc-library/Float_lib1/main.c:49:26: error: non-floating-point argument in call to function ‘__builtin_isinf_sign’
     __builtin_isinf_sign(0) == 0, "__builtin_isinf_sign is constant");
                          ^
regression/cbmc-library/Float_lib1/main.c:49:5: error: expression in static assertion is not an integer
     __builtin_isinf_sign(0) == 0, "__builtin_isinf_sign is constant");
     ^~~~~~~~~~~~~~~~~~~~
```

kroening: Ah, ok, gcc/clang also error this.

src/ansi-c/c_typecheck_code.cpp
tautschnig: Done!

tautschnig:
#5679 needs to be merged first to make CI happy again.

src/util/simplify_expr_floatbv.cpp
```diff
@@ -20,9 +20,6 @@ Author: Daniel Kroening, kroening@kroening.com
 simplify_exprt::resultt<>
 simplify_exprt::simplify_isinf(const unary_exprt &expr)
 {
-  if(expr.op().type().id() != ID_floatbv)
```
kroening, line 23: I would leave that as a PRECONDITION.

tautschnig: Could do, but would be redundant with `to_floatbv_type(expr.type())` that `ieee_floatt::from_expr` does. I'm happy to add it for clarity, but if so I'd suggest to add it to `simplify_isnan` and `simplify_isnormal` as well?

kroening: Yes, please add, the idea being "fail early" and "clarity".

kroening (approved):

src/util/simplify_expr_floatbv.cpp
tautschnig: Done!

PR #5678, opened 2020-12-19
VSD pointer arithmetic, differencing, and comparisons

jezhiggins:
Pointer arithmetic fixes 

- Evaluates `(p - q)` correctly, when `p` and `q` point to the same base object
- Pointer manipulation evaluates correctly when the result points to the base object. eg 
``` 
int* p = array;
++p;
--p;
assert (p == array);
```
Assertion is now SUCCESS, when previously it was UNKNOWN.

----
A pointer difference or comparison expression is slightly unusual - the type of the expression is a number (or boolean), but the two operands are pointer types. Generally, the expression type is the same as the first operand (if not all of them). Because of this type mismatch, the abstract environment was dispatching the expression to the wrong abstract type for evaluation, so the result was TOP and therefore unhelpful.

The changes to evaluate pointer differences are small - an additional case in `abstract_environmentt::eval` to pick up a pointer difference and dispatch it to `abstract_pointer_object`, and then the corresponding changes in `abstract_pointer_object` to calculate the difference. 

Pointer comparisons are a little more involved. Again, we have an additional case in `abstract_environmentt::eval` to dispatch the expression to `abstract_pointer_object`. If the two pointers point to the same base object we can rewrite the comparison in terms of the pointer offset and evaluate that. If the two pointers point to different objects, we can evaluate == and !=, but other comparisons go top.


jezhiggins:
**Build Xen with CPROVER tools / CompileXen (pull_request)** is failing for unrelated reasons.

```
Fetched 6178 kB in 1s (5261 kB/s)
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/g/glibc/libc6-dbg_2.27-3ubuntu1.3_amd64.deb  404  Not Found [IP: 52.147.219.192 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

regression/goto-analyzer/pointer-arithmetic-simple/test.desc
```diff
@@ -0,0 +1,7 @@
+CORE
+main.c
+--variable-sensitivity --vsd-pointers constants --vsd-values constants --verify --pointer-check
```
martin-cs: Why is `--pointer-check` needed?

regression/goto-analyzer/pointer-comparison/main.c
```diff
@@ -0,0 +1,15 @@
...
+  int a[2];
+  int b[2];
+
+  int *p = a;
+  int *q = a;
+
+  assert(p - q == 0);
+
+  q = b;
+  assert(p - q == 0);
```
martin-cs, line 14: I have a feeling that the C standard says this is undefined behaviour.  We probably should support it and so it is reasonable to test for it but until we have a more principled way of handling UB, we should be careful.  If you have to modify the code then a comment noting that this is intentionally UB would be good.

regression/goto-analyzer/pointer-difference-after-decrement/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+--variable-sensitivity --vsd-pointers constants --vsd-values constants --verify
+^EXIT=0$
+^SIGNAL=0$
+^\[main.assertion.1\] line 10 assertion q - p == 1: SUCCESS
+^\[main.assertion.2\] line 11 assertion q != p: UNKNOWN
```
martin-cs: Why is this unknown?  If the previous one is resolvable it feels like it should be.

regression/goto-analyzer/pointer-difference-after-decrements/main.c
```diff
@@ -0,0 +1,29 @@
...
+
+  q -= 1;
+  offset = (q - p);
+  assert(offset == 1);
+
+  q = q - 1;
+  offset = (q - p);
+  assert(offset == 0);
+  assert(q == p);
+}
```
martin-cs, line 29: Nice test!

regression/goto-analyzer/pointer-comparison/test.desc
```diff
@@ -0,0 +1,7 @@
+CORE
+main.c
+--variable-sensitivity --vsd-pointers constants --vsd-values constants --verify --pointer-check
```
martin-cs: Why `--pointer-check`?

regression/goto-analyzer/pointer-difference-after-increment/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+--variable-sensitivity --vsd-pointers constants --vsd-values constants --verify
+^EXIT=0$
+^SIGNAL=0$
+^\[main.assertion.1\] line 10 assertion q - p == 1: SUCCESS
+^\[main.assertion.2\] line 11 assertion q != p: UNKNOWN
```
martin-cs: Again, somewhat unclear why this is unknown.

regression/goto-analyzer/pointer-differences/test.desc
```diff
@@ -0,0 +1,12 @@
...
+main.c
+--variable-sensitivity --vsd-pointers constants --vsd-values constants --verify
+^EXIT=0$
+^SIGNAL=0$
+^\[main.assertion.1\] line 16 assertion offset == 4: SUCCESS$
+^\[main.assertion.2\] line 18 assertion offset == 3: SUCCESS$
+^\[main.assertion.3\] line 20 assertion offset == 1: SUCCESS$
+^\[main.assertion.4\] line 22 assertion offset == 0: SUCCESS$
+^\[main.assertion.5\] line 26 assertion offset == 2: SUCCESS$
+^\[main.assertion.6\] line 27 assertion q == p: UNKNOWN$
```
martin-cs: Another unclear one; equality should be resolvable to a difference check if nothing else.

src/analyses/variable-sensitivity/abstract_enviroment.cpp
```diff
@@ -517,3 +519,18 @@ std::vector<abstract_object_pointert> eval_operands(
 
   return operands;
 }
+
+class simplify_vsd_exprt : public simplify_exprt
```
martin-cs: Having a custom simplification for VSD is not unreasonable but I am not immediately seeing what it adds in this case.

src/analyses/variable-sensitivity/constant_pointer_abstract_object.cpp
```diff
@@ -70,6 +71,35 @@ constant_pointer_abstract_objectt::merge(abstract_object_pointert other) const
...
+}
+
+exprt constant_pointer_abstract_objectt::offset_from(
+  abstract_object_pointert other) const
+{
+  auto cast_other =
+    std::dynamic_pointer_cast<const constant_pointer_abstract_objectt>(other);
+
+  if(value_stack.is_top_value() || cast_other->value_stack.is_top_value())
+    return nil_exprt();
```
martin-cs, line 110: I would really like to get rid of the use of `nil_exprt()` as it feels very kludged.  Can this method return an `abstract_objectt` so it has a way of returning top.  Failing that?  `optionalt`?

src/analyses/variable-sensitivity/pointer_abstract_object.h
```diff
@@ -97,6 +97,15 @@ class pointer_abstract_objectt : public abstract_objectt
 protected:
   CLONE
 
+  virtual bool same_target(abstract_object_pointert other) const
```
martin-cs: I am not entirely opposed to adding to the interface for pointers if it is necessary.  However there is a general move towards making the interface of `abstract_objectt` and children more uniform so I wonder if there is another way.

src/util/irep_ids.def
```diff
@@ -842,6 +842,7 @@ IREP_ID_ONE(statement_list_instructions)
 IREP_ID_ONE(max)
 IREP_ID_ONE(min)
 IREP_ID_ONE(constant_interval)
+IREP_ID_ONE(ptr_diff)
```
martin-cs: I realise it is quite an elegant solution to the problem but I would ideally like to avoid adding an `irept` type that is a restricted version of another type / echos some of the type information.  I haven't got that far through the PR, maybe there is something that is simply not possible without it but, could we just have a utility method for `is this subtract a pointer difference`?

martin-cs (changes requested):
I like what this is doing, there are a few places where I feel how it is doing it could be better.  I suspect "better" here means deeper and more intrusive changes to the `write_stack` parts and `constant_pointer_abstract_objectt` which are only obviously "better" when taking into account a load of other things so why don't we have a chat and see what can be done?

src/analyses/variable-sensitivity/abstract_enviroment.cpp
jezhiggins: It's gone!

src/util/irep_ids.def
jezhiggins: Yep - this and all the scaffolding to support it is gone.

src/analyses/variable-sensitivity/constant_pointer_abstract_object.cpp
jezhiggins: This is calling site is `return environment.eval(offset_from(rhs), ns);` so the `nil_exprt()` does end up as a TOP value.

regression/goto-analyzer/pointer-difference-after-decrement/test.desc
jezhiggins: Turns out the `q-p == 1` case evaluating to success was an artefact. I'm implementing proper pointer comparisons and `q != p` will evaluate properly.

regression/goto-analyzer/pointer-comparison-equality-struct-members/main.c
```diff
@@ -0,0 +1,28 @@
+#include <assert.h>
+
+int main()
+{
+  struct int_pair
+  {
+    int a;
+    int b;
+  };
+  struct int_pair x = {0, 1.0};
```
martin-cs: Is the use of a double here significant?

regression/goto-analyzer/pointer-comparison-inequality-different-target/main.c
```diff
@@ -0,0 +1,20 @@
+#include <assert.h>
+
+int main()
+{
+  int a[2];
+  int b[2];
+
+  int *p = &(a[0]);
+  int *q = &(b[1]);
+
```
martin-cs, line 10: Can we have a comment in either in the test case or in the `.desc` files noting that this is undefined behaviour and so `UNKNOWN` is the expected behaviour.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -824,3 +854,18 @@ exprt assume_greater_than(
 
   return assume_less_than(env, symmetric_expr, ns);
 }
+
+class simplify_vsd_exprt : public simplify_exprt
```
martin-cs: I would rather not do this, if possible.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -53,59 +54,88 @@ std::vector<abstract_object_pointert> eval_operands(
...
+{
+  return (expr.id() == ID_minus) &&
+         (expr.operands()[0].type().id() == ID_pointer) &&
+         (expr.operands()[1].type().id() == ID_pointer);
+}
+
+bool is_ptr_comparison(const exprt &expr)
+{
+  auto const &id = expr.id();
+  if(id == ID_not)
```
martin-cs: Are you handling `ID_and` elsewhere?  If so, should you handle `ID_not` there rather than here?

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -53,59 +54,88 @@ std::vector<abstract_object_pointert> eval_operands(
...
+
 abstract_object_pointert
 abstract_environmentt::eval(const exprt &expr, const namespacet &ns) const
 {
   if(bottom)
     return abstract_object_factory(expr.type(), ns, false, true);
 
   // first try to canonicalise, including constant folding
-  const exprt &simplified_expr = simplify_expr(expr, ns);
+  const exprt &simplified_expr = simplify_vsd_expr(expr, ns);
```
martin-cs: Pondering : do we have to simplify here?  Could we do special case recognition first?

src/analyses/variable-sensitivity/abstract_pointer_object.cpp
```diff
@@ -74,48 +73,79 @@ abstract_object_pointert abstract_pointer_objectt::write(
   return write_dereference(environment, ns, stack, value, merging_write);
 }
 
-abstract_object_pointert abstract_pointer_objectt::read_dereference(
+void abstract_pointer_objectt::get_statistics(
```
martin-cs, line 76: FYI : I am not sure whether this infrastructure is used at the moment.

src/analyses/variable-sensitivity/abstract_pointer_object.cpp
```diff
@@ -74,48 +73,79 @@ abstract_object_pointert abstract_pointer_objectt::write(
...
-  const abstract_environmentt &env,
+exprt abstract_pointer_objectt::eval_ptr_comparison_as_expr(
+  const exprt &expr,
+  const std::vector<abstract_object_pointert> &operands,
+  const abstract_environmentt &environment,
   const namespacet &ns) const
 {
-  abstract_objectt::get_statistics(statistics, visited, env, ns);
-  ++statistics.number_of_pointers;
+  if(expr.id() == ID_not)
```
martin-cs: It feels rather like breaking conditions into atoms should be done elsewhere.

src/analyses/variable-sensitivity/constant_pointer_abstract_object.cpp
```diff
@@ -72,6 +72,48 @@ abstract_object_pointert constant_pointer_abstract_objectt::merge(
...
+  auto cast_other =
+    std::dynamic_pointer_cast<const constant_pointer_abstract_objectt>(other);
+
+  if(value_stack.is_top_value() || cast_other->value_stack.is_top_value())
+    return false;
+
+  if(value_stack.depth() != cast_other->value_stack.depth())
+    return false;
+
+  for(size_t d = 0; d != value_stack.depth() - 1; ++d)
```
martin-cs, line 87: Looks reasonable, the -1 is needed because you don't want the most detailed part of the offset.

src/analyses/variable-sensitivity/constant_pointer_abstract_object.cpp
```diff
@@ -72,6 +72,48 @@ abstract_object_pointert constant_pointer_abstract_objectt::merge(
...
+{
+  auto cast_other =
+    std::dynamic_pointer_cast<const constant_pointer_abstract_objectt>(other);
+
+  if(value_stack.is_top_value() || cast_other->value_stack.is_top_value())
+    return nil_exprt();
+
+  return minus_exprt(
+    value_stack.offset_expression(),
+    cast_other->value_stack.offset_expression());
```
martin-cs, line 114: Why convert back to exprt's when you could just subtract the two abstract objects?

src/analyses/variable-sensitivity/context_abstract_object.cpp
```diff
@@ -86,8 +86,6 @@ abstract_object_pointert context_abstract_objectt::expression_transform(
   const abstract_environmentt &environment,
   const namespacet &ns) const
 {
-  PRECONDITION(expr.operands().size() == operands.size());
```
martin-cs, line 89: Why?

src/util/std_expr.h
```diff
@@ -922,22 +914,31 @@ inline const minus_exprt &to_minus_expr(const exprt &expr)
...
-class mult_exprt:public multi_ary_exprt
+class mult_exprt : public multi_ary_exprt
 {
 public:
   mult_exprt(exprt _lhs, exprt _rhs)
     : multi_ary_exprt(std::move(_lhs), ID_mult, std::move(_rhs))
   {
   }
+
+  const exprt &lhs() const
```
martin-cs: This is potentially a bit misleading as `mult_exprt` can have more than 2 operands.

martin-cs (changes requested):
The actual content of the PR looks good.

Reformatting of things in `util/` will need to go in a separate PR if they are needed.

regression/goto-analyzer/pointer-comparison-equality-struct-members/main.c
jezhiggins: No. I originally wrote `struct { int a, float b; };` but then changed it. Can#t remember why.

src/analyses/variable-sensitivity/abstract_pointer_object.cpp
jezhiggins: Maybe I got carried away here. I can unwind this.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: We have to customise the simplification, because it recurses right down the expression tree. If we just let the simplification run, then the special case ends up being eliminated before we get a chance to try and evaluate it.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: We do have to do it, but I've made further changes to try and make things less intrusive.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: Yep, sorted

src/analyses/variable-sensitivity/constant_pointer_abstract_object.cpp
jezhiggins: I'm not :)

martin-cs (approved):
Thanks for removing the changes to `util/` I think it is a better PR for the extra passes.

PR #5681, opened 2020-12-21
Enable reading NULL_is_zero from symbol table

tautschnig:
This was left commented out ever since introducing it in 5952f6aeb7c,
which is reasonable for it required bumping the goto-program version. We
have incremented the version since 2014, and thus it is safe to enable
this now.


martin-cs (approved):
OK but ... for those developers who weren't paying attention 7 years ago... any chance we could have some context or maybe a test (unit or regression) that this actually effects?

peterschrammel (approved):

PR #5682, opened 2020-12-21
goto_symext::apply_goto_condition must honour --no-propagation

tautschnig:
goto_symext::apply_goto_condition unconditionally wrote to the
propagation map (via goto_statet::apply_condition). Although
--no-propagation might not be really useful in practice, but is used for
testing (including some of our existing regression tests) or debugging.

### DIFF
--- a/regression/cbmc/no-propagation/main.c
+++ b/regression/cbmc/no-propagation/main.c
@@ -1,0 +1,6 @@
+int main()
+{
+  int x;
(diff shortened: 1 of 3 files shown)


martin-cs (approved):
I can see why this would be useful, yes.  It is also more consistent with what the `--no-propagation` does.

peterschrammel (approved):

PR #5683, opened 2020-12-21
Concurrency: ensure apply_condition's value is used

tautschnig:
goto_statet::apply_condition generates an entry (with fresh L2 index) in
the propagation map with generating an assignment. We did not previously
use the propagation map when handling atomic sections, and thus had a
missing link between values, causing spurious verification failures.

The regression test is based on SV-COMP's
pthread-ext/23_lu-fig2.fixed.c.

There is further room for improvement, as shown in the
KNOWNBUG-regression test. This will be addressed in a separate commit.

### DIFF
--- a/regression/cbmc-concurrency/atomic_section_sc7/test-guard.desc
+++ b/regression/cbmc-concurrency/atomic_section_sc7/test-guard.desc
@@ -1,0 +1,12 @@
+KNOWNBUG
+main.c
+--program-only
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+__COUNT__#\d+ == (1 \+ __COUNT__#\d+|__COUNT__#\d+ \+ 1)
+--
+__COUNT__ is guaranteed to have been read within the atomic section, there
+should not be a need to produce a conditional assignment, and therefore the
+value of __COUNT__ can be constant propagated.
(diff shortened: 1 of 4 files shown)


regression/cbmc-concurrency/atomic_section_sc7/test-guard.desc
```diff
@@ -0,0 +1,12 @@
+KNOWNBUG
+main.c
+--program-only
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+__COUNT__#\d+ == (1 \+ __COUNT__#\d+|__COUNT__#\d+ \+ 1)
```
peterschrammel, line 8: Isn't the same expression in thr2 where it cannot be propagated?

peterschrammel (approved):

regression/cbmc-concurrency/atomic_section_sc7/test-guard.desc
tautschnig: It actually can be propagated, because that entire function is an atomic section and the increment is guarded by `if(__COUNT__ == 1)` (in `thr2`, and `if(__COUNT__ == 0)` in `thr1`). As we nowadays constant-propagate such conditions, it ought to be fully constant.

PR #5688, opened 2020-12-23
C front-end: initializer lists do not initialise anonymous members

tautschnig:
We already took care of this in another place in
c_typecheck_initializer, but did not handle initializer lists correctly.

The test is based on a case found by C-Reduce when starting from an
SV-COMP task.

### DIFF
--- a/regression/cbmc-library/memcpy-06/test.desc
+++ b/regression/cbmc-library/memcpy-06/test.desc
@@ -1,4 +1,4 @@
-CORE
+CORE gcc-only
 main.c
 
 function 'memcpy' is not declared
@@ -8,3 +8,6 @@ parameter "memcpy::dst" type mismatch
 --
 ^warning: ignoring
 Invariant check failed
+--
+This test requires support for empty structs, which aren't supported by Visual
+Studio.
--- a/src/ansi-c/c_typecheck_type.cpp
+++ b/src/ansi-c/c_typecheck_type.cpp
@@ -947,6 +947,55 @@ void c_typecheck_baset::typecheck_compound_body(
 
         typecheck_type(new_component.type());
 
+        // the rules for anonymous members depend on the type and the compiler,
+        // just bit fields are accepted everywhere
+        if(
+          new_component.get_name().empty() &&
+          new_component.type().id() != ID_c_bit_field)
+        {
+          if(config.ansi_c.mode == configt::ansi_ct::flavourt::VISUAL_STUDIO)
+          {
+            // Visual Studio rejects anything other than a struct or union
+            // declaration, but also accepts those when they introduce a tag
+            if(
+              new_component.type().id() != ID_struct_tag &&
+              new_component.type().id() != ID_union_tag)
+            {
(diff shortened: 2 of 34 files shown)


martin-cs (approved):
The small part that touches area I maintain looks OK.  I'm not sure the tests have enough context for me to really check what is going on with them.

regression/cbmc-library/memcpy-06/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+CORE pthread
```
kroening: Use `gcc-only` instead?

regression/ansi-c/struct5/test.desc
```diff
@@ -1,6 +1,6 @@
 CORE
 main.c
-
+--i386-win32
 ^EXIT=0$
```
kroening, line 4: Seems unrelated?

tautschnig: No, this is a relevant change. We previously wrongly supported the code in this test on all platforms: it uses anonymous (but tagged) union members.

regression/cbmc-library/memcpy-06/test.desc
tautschnig: Fixed.

kroening:
Are zero-width bit-fields considered when processing initialiser lists?

tautschnig:
@kroening:
> Are zero-width bit-fields considered when processing initialiser lists?

The C standard requires "The expression that specifies the width of a bit-field [...]. If the value is zero, the declaration shall have no declarator." GCC and Clang enforce this, we did not. I'll put together a PR to ensure our C front-end becomes equally strict.

src/ansi-c/c_typecheck_type.cpp
```diff
@@ -948,6 +948,41 @@ void c_typecheck_baset::typecheck_compound_body(
...
+          else
+          {
+            // GCC and Clang ignore anything other than an untagged struct or
+            // union; we could print a warning, but there isn't any ambiguity in
+            // semantics here. Printing a warning could elevate this to an error
+            // when compiling code with goto-cc with -Werror.
+            if(
+              (new_component.type().id() != ID_struct_tag &&
+               new_component.type().id() != ID_union_tag) ||
+              follow(new_component.type()).find(ID_tag).is_not_nil())
```
kroening: What does the follow do here?

tautschnig: This is to distinguish untagged (anonymous) structs/unions from ones that are named.

tautschnig: I have extended the comment above this code now.

kroening: May I suggest using `to_tag_type` to make this clear.

tautschnig: I've rewritten it to use `to_struct_tag_type` and `to_union_tag_type`.

peterschrammel (approved):

PR #5689, opened 2020-12-23
Simplification of pointer comparison: do not assume numeric types

tautschnig:
As we cannot (or chose not to) catch all type inconsistencies in the
language front-end, we may pass expressions to simplification that do
not adhere to typing expectations. The included example is such a case,
where the missing declaration of `memcpy` makes the struct-to-pointer
cast get all the way to the middle end.

Test is based on a sample generated by C-Reduce when starting from an
SV-COMP task.

### DIFF
--- a/regression/cbmc-library/memcpy-06/test.desc
+++ b/regression/cbmc-library/memcpy-06/test.desc
@@ -1,0 +1,10 @@
+CORE
+main.c
+
+function 'memcpy' is not declared
+parameter "memcpy::dst" type mismatch
+^EXIT=6$
+^SIGNAL=0$
+--
+^warning: ignoring
+Invariant check failed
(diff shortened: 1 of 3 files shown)


regression/cbmc-library/memcpy-06/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+
+function 'memcpy' is not declared
+parameter "memcpy::dst" type mismatch
```
peterschrammel, line 5: ⛏️ mixture of " and '

peterschrammel (approved):

regression/cbmc-library/memcpy-06/test.desc
tautschnig: It seems there's some more work after #4875 left to be done. I'll create a separate PR for this.

PR #5692, opened 2020-12-24
C front-end/KnR: generate nondet for missing arguments

tautschnig:
The back-end does not know how to handle `nil`, and symex would also
generate nondet in such cases, so do the same in the front-end.

### DIFF
--- a/regression/cbmc/KnR1/main.c
+++ b/regression/cbmc/KnR1/main.c
@@ -1,0 +1,8 @@
+void a(b)
+{
+  __CPROVER_assert(0, "");
+}
+int main()
+{
+  a();
+}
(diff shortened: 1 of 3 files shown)


kroening:
This deviates quite a lot from the usual pattern: this would be the only place in the front-end where we pick nondeterministic semantics.

tautschnig:
> This deviates quite a lot from the usual pattern: this would be the only place in the front-end where we pick nondeterministic semantics.

Should we instead handle this case in goto-symex, translating `nil` to nondet?

peterschrammel (approved):
At some point it needs to be replaced (frontend, goto, or symex). Having nil in the goto model doesn't seem right, and symex is certainly too late, IMO.

PR #5694, opened 2020-12-24
C front-end: ignore or reject conflicting function redeclaration

tautschnig:
A declaration of a function currently being type checked should not
replace the existing declaration. To accomplish this, make sure a
definition `type name() { ... }` fixes the type of `name` to have no
parameters and no ellipsis. This, more precise, typing highlighted a bug
in the goto-harness/parameter_and_global_variable_distinction regression
test (and the underlying implementation!), which assumed that the type
of `int (*fptr_t)(void)` did not match `int call_f()`.

Test generated by C-Reduce, starting from an SV-COMP task.

### DIFF
--- a/regression/ansi-c/invalid_redeclaration1/test-same_type.desc
+++ b/regression/ansi-c/invalid_redeclaration1/test-same_type.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+-Dsame_return_type
+^EXIT=0$
+^SIGNAL=0$
+--
+Invariant check failed
+^CONVERSION ERROR$
+--
+This previously triggered the following (undesired) invariant violation:
+File: ../src/goto-programs/goto_convert_functions.cpp:167 function: convert_function
+Condition: parameter identifier should not be empty
--- a/regression/ansi-c/invalid_redeclaration1/test.desc
+++ b/regression/ansi-c/invalid_redeclaration1/test.desc
@@ -1,0 +1,9 @@
+CORE test-c++-front-end
+main.c
+
+^EXIT=(64|1)$
+^SIGNAL=0$
+(function symbol 'a' redefined with a different type|symbol 'c' is unknown)
+^CONVERSION ERROR$
+--
+Invariant check failed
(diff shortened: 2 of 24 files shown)


tautschnig:
Requires #7305 for tests to pass.

regression/ansi-c/invalid_redeclaration1/test-same_type.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+-Dsame_return_type
+^EXIT=0$
+^SIGNAL=0$
+--
+Invariant check failed
```
feliperodri, line 7: How do you check that is falling on the intended invariant violation here?

regression/ansi-c/invalid_redeclaration1/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE test-c++-front-end
+main.c
+
+^EXIT=(64|1)$
+^SIGNAL=0$
+(function symbol 'a' redefined with a different type|symbol 'c' is unknown)
```
feliperodri, line 6: FYI, gcc reports `error: a parameter list without types is only allowed in a function definition`.

feliperodri (approved):
LGTM, only minor comments outside contracts module.

regression/ansi-c/invalid_redeclaration1/test.desc
tautschnig: That will be true with more recent GCC versions (which have become stricter about the use of function declarations without parameters), but GCC 9 still only says "error: conflicting types for ‘a’". I don't think we want to become as strict as more recent GCC versions are to keep supporting old, non-standards-compliant code.

regression/ansi-c/invalid_redeclaration1/test-same_type.desc
tautschnig: This is a statement in the second of patterns not to occur, so this is checking that there is no invariant violation. I have added a comment to the file to explain which previous invariant violation this was.

peterschrammel (approved):

PR #5696, opened 2020-12-25
Message handler: add quote_begin, quote_end commands

tautschnig:
These will permit custom formatting of quoting for each message handler, and for a start introduce <quote> tags for XML. See #4875 for discussion.

Please review commit-by-commit.

### DIFF
--- a/src/util/message.cpp
+++ b/src/util/message.cpp
@@ -95,6 +95,16 @@ const messaget::commandt messaget::bright_yellow(93);
 const messaget::commandt messaget::bright_blue(94);
 const messaget::commandt messaget::bright_magenta(95);
 const messaget::commandt messaget::bright_cyan(96);
+// quote_begin and quote_end use the ASCII codes for '<' (60) and '>'
+// (62), which happen to fall inside ECMA-48's SGR ideogram-styling
+// parameter range (60..65). The overload is safe today because nothing
+// uses ideogram modes via this same mechanism, but if a future styling
+// command is added with a code in 60..65 there will be a clash with no
+// compile-time signal. Either re-encode quote_begin/quote_end to a
+// value above the SGR range (e.g. >= 256) or pick non-overlapping
+// codes for any new ideogram styles.
+const messaget::commandt messaget::quote_begin('<');
+const messaget::commandt messaget::quote_end('>');
 
 /// Parse a (user-)provided string as a verbosity level and set it as the
 /// verbosity of dest.
--- a/src/util/ui_message.h
+++ b/src/util/ui_message.h
@@ -96,10 +96,74 @@ protected:
 
   std::string command(unsigned c) const override
   {
-    if(message_handler)
-      return message_handler->command(c);
-    else
+    // quote_begin / quote_end render as a single quote on every currently
+    // supported UI. Handled before the message_handler null-check so this
+    // also applies when the handler is constructed via the cmdlinet ctor,
+    // which leaves message_handler null for XML_UI / JSON_UI. We
+    // considered emitting `<quote>` / `</quote>` tags for XML output (see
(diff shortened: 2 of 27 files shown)


martin-cs (approved):
Seems good.  It would be nice if this also had the implementation of this for XML so that it was clear why it was valueable and no-one will accidentally 'simplify' it out.

src/util/ui_message.h
```diff
@@ -94,10 +94,75 @@ class ui_message_handlert : public message_handlert
...
+      case 36: // cyan
+      case 91: // bright_red
+      case 92: // bright_green
+      case 93: // bright_yellow
+      case 94: // bright_blue
+      case 95: // bright_magenta
+      case 96: // bright_cyan
+        return std::string();
+      case '<': // quote_begin
+        return "<quote>";
```
peterschrammel: Not sure this is a good idea as it changes the XML output format. I'd stick to single quotes for now and introduce a different format in a separate PR if desired.

tautschnig: I have now factored this into a commit of its own, which could thus easily be dropped from the series.

peterschrammel: I would really prefer to drop this commit. IMO, it causes unnecessary hassle to XML output users.

tautschnig: Removed, just left behind a comment that this could be done.

peterschrammel (approved):

TGWDB (approved):

TGWDB (approved):

src/util/ui_message.h
```diff
@@ -96,10 +96,80 @@ class ui_message_handlert : public message_handlert
...
+    case 92: // bright_green
+    case 93: // bright_yellow
+    case 94: // bright_blue
+    case 95: // bright_magenta
+    case 96: // bright_cyan
+      return true;
+    default:
+      return false;
+    }
   }
```
tautschnig: Two minor maintainability points on the new helper:

1. The body is large enough to be worth moving to `ui_message.cpp`. Today it is inlined into every TU that includes `ui_message.h`, which is a lot. Same comment applies to `command()` above.
2. There is no unit test exercising this helper, even though the whole point of the refactor was to make it easier to add/remove SGR codes. A quick `TEST_CASE("is_sgr_style_command covers all messaget styling codes", "[util][ui_message]")` that walks the `messaget::commandt` constants and asserts the predicate matches (and conversely that `quote_begin`/`quote_end` do *not* match) would lock the contract down. It would also catch the case where the doc-string range disagrees with the `case` list (they currently agree, but they're maintained independently).

src/util/message.cpp
```diff
@@ -95,6 +95,8 @@ const messaget::commandt messaget::bright_yellow(93);
 const messaget::commandt messaget::bright_blue(94);
 const messaget::commandt messaget::bright_magenta(95);
 const messaget::commandt messaget::bright_cyan(96);
+const messaget::commandt messaget::quote_begin('<');
+const messaget::commandt messaget::quote_end('>');
```
tautschnig: Encoding-choice nit: `'<'` and `'>'` happen to be ASCII 60 and 62, which are valid ECMA-48 SGR parameter slots reserved for ideogram-related styling (60–65). Nothing currently uses them, but if anyone ever adds ideogram modes via this same mechanism we'll have a runtime clash that's easy to miss because it produces no compile-time signal.

A cleaner alternative is to give non-SGR commands values outside the SGR range, e.g. `static const commandt quote_begin(256);`. The `console_message_handlert::command` and `ui_message_handlert::command` already special-case these by value, so widening is essentially free. At minimum, please add a comment here explaining the overload so future readers don't accidentally pick another colliding code.

PR #5700, opened 2020-12-27
Add CSmith GitHub action

tautschnig:
Run 10 randomly generated CSmith tests as a GitHub action to perform an
integration test involving CBMC and goto-instrument --dump-c. Each test
is first compiled and run using GCC to obtain a checksum, which is then
included as an assertion for CBMC to verify.


hannes-steffenhagen-diffblue:
This seems like a good idea but the script could use some more explanation for what is going on there.

tautschnig:
> This seems like a good idea but the script could use some more explanation for what is going on there.

Indeed. I've added comments!

martin-cs (approved):
I /really/ like the concept but ... non-deterministic things make me nervous.  I would be happier if the `scripts/csmith.sh` took a random seed and `csmith.yml` set it to the hash of the diff or something similar that could be locally reproduced.

I'm also a little unsure about what you feel should happen if you find issues of this form.  Is it the job of the person who submitted the PR to fix them?  Raise issues?

hannes-steffenhagen-diffblue:
@martin-cs I'd say failure for these is not PR raiser responsibility, but if something does come up we should definitely take note.

hannes-steffenhagen-diffblue (approved):
IMHO "cbmc matches csmith test behaviour" and "goto-cc --> dump-c should yield program with same behaviour" are two different tests and probably shouldn't be in the same script, but otherwise I think this is a good addition for finding potential problems hopefully before users run into them.

NlightNFotis (approved):
This is great!

tautschnig:
> I /really/ like the concept but ... non-deterministic things make me nervous. I would be happier if the `scripts/csmith.sh` took a random seed and `csmith.yml` set it to the hash of the diff or something similar that could be locally reproduced.
> 

Very good point - I've now modified the script to:
1) Generate a seed (using `date +%s`) and print that seed.
2) Add the ability to re-run with a particular seed.

> I'm also a little unsure about what you feel should happen if you find issues of this form. Is it the job of the person who submitted the PR to fix them? Raise issues?

I've done multiple thousands of runs of CSmith in recent weeks, and am thus fairly confident that the likelihood of a PR tripping this up by a bug outside the changes introduced in said PR are fairly low. Thus I'd argue it is the job of the person who submitted the PR to get this back in working order.

tautschnig:
> IMHO "cbmc matches csmith test behaviour" and "goto-cc --> dump-c should yield program with same behaviour" are two different tests and probably shouldn't be in the same script, but otherwise I think this is a good addition for finding potential problems hopefully before users run into them.

Yes, having two separate jobs might be nicer, but I'm a bit hesitant to have almost-the-same-script twice in the repository. So I'd for now stick with the current solution. If we find that we regularly only run into bugs in one or the other part then that'll provide an incentive to improve this.

martin-cs:
@tautschnig : sounds great; let's do it!

PR #5701, opened 2020-12-27
Ensure unique GitHub action cache keys

tautschnig:
Concurrently running jobs using the same cache key result in GitHub
actions not caching, reporting "Unable to reserve cache with key
<cache-key>, another job may be creating this cache."

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: .ccache
-          key: ${{ runner.os }}-20.04-make-${{ github.ref }}-${{ github.sha }}
+          key: ${{ runner.os }}-20.04-make-${{ github.ref }}-${{ github.sha }}-PR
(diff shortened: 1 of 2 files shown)


hannes-steffenhagen-diffblue (approved):
Looks OK to me. Is there any way we can see cache utilisation? With so many cache keys I'm somewhat afraid we'll eventually just constantly be hitting the cache limit if that's not the case already.

tautschnig:
> Looks OK to me. Is there any way we can see cache utilisation? With so many cache keys I'm somewhat afraid we'll eventually just constantly be hitting the cache limit if that's not the case already.

I haven't found a way to see this. Each of the cache entries is fairly small, and GitHub evicts unused keys of 7 days (https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy), so I _think_ we're fine. The only indication that I have is that most builds on GitHub actions are fairly quick, the CodeBuild jobs are usually what we need to wait for.

PR #5704, opened 2020-12-29
Union regression test of unspecified behaviour

tautschnig:
The C standard leaves the values of bytes not belonging to the object
representation of the member last stored unspecified, but GCC, Clang,
and Visual Studio all seem to behave just as CBMC currently does. Ensure
we maintain this behaviour via a regression test.

### DIFF
--- a/regression/cbmc/union15/main.c
+++ b/regression/cbmc/union15/main.c
@@ -1,0 +1,62 @@
+#include <assert.h>
+
+struct S
+{
+  union U {
+    int x;
+    int y;
+  } u;
+  int z;
+} s = {1, 2};
+
+union B {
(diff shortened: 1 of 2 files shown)


kroening:
That one may be worth mentioning in the manual.

kroening (approved):

tautschnig:
> That one may be worth mentioning in the manual.

Any proposals what section this could go into? That said, I'd view it as a temporary fix as an upcoming change will make it really easy to make this behaviour configurable.

tautschnig:
I'll merge this one, and will leave working on documentation and/or making this behaviour configurable to #5705, which requires further work anyway.

PR #5705, opened 2020-12-29
Initialise union of static lifetime with zeros

tautschnig:
The C standard does not specify the value of bytes that do not belong to
the object representation of a member referred to in an initializer
list, but compilers seem to ensure those bytes are zero (possibly by
using the .bss section).

### DIFF
--- a/src/util/simplify_utils.cpp
+++ b/src/util/simplify_utils.cpp
@@ -185,8 +185,15 @@ optionalt<exprt> bits2expr(
   // bits start at lowest memory address
   auto type_bits = pointer_offset_bits(type, ns);
 
-  if(!type_bits.has_value() || *type_bits != bits.size())
+  if(
+    !type_bits.has_value() ||
+    (type.id() != ID_union && type.id() != ID_union_tag &&
+     *type_bits != bits.size()) ||
+    ((type.id() == ID_union || type.id() == ID_union_tag) &&
+     *type_bits < bits.size()))
+  {
     return {};
+  }
 
   if(
     type.id() == ID_unsignedbv || type.id() == ID_signedbv ||
--- a/src/util/simplify_utils.cpp
+++ b/src/util/simplify_utils.cpp
@@ -201,6 +201,13 @@ optionalt<exprt> bits2expr(
     type.id() == ID_c_bit_field || type.id() == ID_pointer ||
     type.id() == ID_bv || type.id() == ID_c_bool)
   {
+    if(
+      type.id() == ID_pointer && config.ansi_c.NULL_is_zero &&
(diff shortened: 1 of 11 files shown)


kroening:
It would make sense to have a warning in the manual about this one as well.

kroening (approved):

tautschnig:
Marking do-not-merge as this requires fixes in dump-c (or some approach not involving `byte_update`).

kroening:
Looking at this, I am wondering whether or not we might want to have a 'strict mode', which would not make compiler or architecture-specific assumptions.

tautschnig:
> Looking at this, I am wondering whether or not we might want to have a 'strict mode', which would not make compiler or architecture-specific assumptions.

Well, at least we should make sure we can configure all those architecture/compiler-specific assumptions. We have quite a lot of this in `configt` already, but handling of unions is not among that yet. I'll work on adding that union-configuration part, but I'd be happy to hear about more aspects that come to mind.

tautschnig:
Trying to make sure we can successfully dump union expressions after fixing zero initialisation via byte updates became quite the bug squashing party. I started out by creating a test for dump-c, which I then noticed CBMC would not produce the correct verification result for. Here is the history as it may help to understand the flood of commits:

1. Added a test that dump-c ought to reproduce correctly (42b4740).
2. Figured out that we could also check for "VERIFICATION SUCCESSFUL" as part of this test, but that required a bitvector flattening bugfix (completely unrelated to to dump-c, the original source would not yield the correct verification result) (7e02b53).
3. Wanted to finally add support to dump-c (eventually done in db4f25c), but that required working `ID_designated_initializer` output. Thus fixed that first (a835405).
4. Now it is possible to use `byte_update` in initialising unions without breaking dump-c's output thereof (a8592a7).
5. Wanted to use `simplify_expr` to resolve this `byte_update` in common cases. This required extending the simplifier (9a51a92) and fixing bugs in `bits2expr`/`expr2bits` (e8ac4d6, ed75598).
6. Finally the simplifier can be used successfully for the desired `byte_update` removal (ac31213).

src/util/simplify_utils.cpp
```diff
@@ -410,7 +417,7 @@ expr2bits(const exprt &expr, bool little_endian, const namespacet &ns)
     else if(type.id() == ID_pointer)
     {
       if(value == ID_NULL && config.ansi_c.NULL_is_zero)
-        return std::string('0', to_bitvector_type(type).get_width());
+        return std::string(to_bitvector_type(type).get_width(), '0');
```
peterschrammel, line 420: Good catch!

peterschrammel (approved):

PR #5707, opened 2020-12-31
Cleanup Java library build

tautschnig:
Please review commit-by-commit.


peterschrammel (approved):

CMakeLists.txt
```diff
@@ -211,6 +211,7 @@ cprover_default_properties(
 
 option(WITH_JBMC "Build the JBMC Java front-end" ON)
 if(WITH_JBMC)
+    find_program(MAVEN_PROGRAM mvn)
```
hannes-steffenhagen-diffblue: This change seems to be completely supplanted by the second commit? Maybe it'd make more sense to squash those.

jbmc/CMakeLists.txt
```diff
@@ -21,3 +14,25 @@ cprover_default_properties(
     java-unit
     miniz
 )
+
+# java models library
+find_program(MAVEN_PROGRAM mvn REQUIRED)
+
+file(GLOB_RECURSE java_models_sources "lib/java-models-library/src/*.java")
```
hannes-steffenhagen-diffblue, line 21: FWIW if ergonomics is a concern then I wouldn't recommend using globs. The problem is that they're only re-evaluated on a reconfigure, so if you add another file it won't be picked up, and removing a file is an error. This is particularly annoying when switching branches, or when doing a git-bisect.

See also [this SO Q&A](https://stackoverflow.com/questions/1027247/is-it-better-to-specify-source-files-with-glob-or-each-file-individually-in-cmak).

Now I realise I'm kind of late to this because we already use globbing elsewhere, and the models library doesn't change that much anyway, so not a blocker, I'll just mention this whenever I see this.

jbmc/src/java_bytecode/library/CMakeLists.txt
```diff
@@ -1,2 +0,0 @@
-# create a target for the executable performing the .jar -> .inc conversion
-add_executable(java-converter converter.cpp)
```
hannes-steffenhagen-diffblue, line 2: Well, I'm certainly not complaining if we're removing code that's no longer relevant. Maybe anyone more familiar with jbmc might want to chime in if that's actually the case?

hannes-steffenhagen-diffblue (approved):
OK with the general direction here.

jbmc/CMakeLists.txt
tautschnig: Yep, I'm regularly running into this with `git-bisect`. But I think that ship has sailed, we're doing it all over the place, and indeed I'm least concerned with the slow-moving target that the java models library is. So if we, at some point, decide to expand all the globs into explicit list then of course we should do this here as well. But for now it seems we can happily use globbing without making things much worse...

jbmc/src/java_bytecode/library/CMakeLists.txt
tautschnig: @peterschrammel perhaps?

CMakeLists.txt
tautschnig: Done!

jbmc/src/java_bytecode/library/CMakeLists.txt
peterschrammel: That code is not used anymore and it's unlikely to be used anytime soon.

PR #5710, opened 2021-01-01
Make IPASIR configuration build after solver hardness changes

tautschnig:
In #5480 the solver hardness interface was updated, but changes were
only implemented in the Minisat2 interface. Update the IPASIR interface
to match these changes.

To avoid future regressions, make the check-macos-10_15-make-clang
GitHub action use IPASIR with Riss as the back-end solver.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -16,8 +16,7 @@ jobs:
           # user input
           DEBIAN_FRONTEND: noninteractive
         run: |
-          sudo apt-get install --no-install-recommends -yq gcc gdb g++ maven jq flex bison libxml2-utils ccache
-          make -C src minisat2-download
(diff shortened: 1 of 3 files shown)


martin-cs (approved):
Again, good to have this in CI.

src/solvers/refinement/bv_refinement_loop.cpp
```diff
@@ -123,6 +123,10 @@ void bv_refinementt::check_SAT()
 
   arrays_overapproximated();
 
+  // get values before modifying the formula
```
peterschrammel: How did you notice this? Is there a test for that or does it fail quite broadly?

peterschrammel (approved):

src/solvers/refinement/bv_refinement_loop.cpp
tautschnig: I actually noticed this while working on #4758, because Cadical complained. For the IPASIR interface this commit was just an attempt to make things work, but we still have failing regression tests, so I need to dig deeper.

PR #5713, opened 2021-01-02
Value sets must not be field sensitive for unions

### LINKED ISSUE #5263: Writing to a value via a pointer fails in some circumstances
CBMC version: 9ee5b9d68324791c00a0493492eae5e3ebff6cdb (develop), also reproducible on 5.11 and 5.10 (from Debian) to an extent 
Operating system: Debian amd64

The program below is a minimal reproducer of a problem I encountered while trying to verify a uC program of mine. I have found three "fixes" to get the program to verify one of which only works on develop, see the comments.

Test program:

```
typedef void (*callback_t)(void);

struct xfer {
        union { /*< FIX: union -> struct */
                struct {
                        int * buf;
                        callback_t callback; /*< FIX: remove */
                } a;
        };
} g_xfer;

int g_buf;

int main()
{
        g_xfer = (struct xfer) {{{ &g_buf }}};

        /* FIX, uncomment (only on cbmc develop 9ee5b9d6): */
        // g_xfer.a.buf = &g_buf;

        /* check the pointer was initialized properly */
        assert(g_xfer.a.buf == &g_buf);

        g_xfer.a.buf[0] = 42; /* write a value via the pointer */
        assert(g_xfer.a.buf[0] == 42); /* check it was written */
        assert(g_buf == 42); /* the underlying value should also be updated */
}
```

Exact command line resulting in the issue: `cbmc test.c`
What behaviour did you expect: All three properties should succeed. 
What happened instead: The last two assertions fail. Doing what any one of the "FIX" comments say will get all properties to succeed.

tautschnig:
Distinct fields in a union need not refer to distinct memory locations.
Thus treat all fields of a union the same, with pointers assigned to any
field being dereferenceable via another field.

Fixes: #5263

### DIFF
--- a/regression/cbmc/union14/test.desc
+++ b/regression/cbmc/union14/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
 
 ^EXIT=0$
@@ -7,6 +7,5 @@ main.c
 --
 ^warning: ignoring
 --
-Value sets do not properly track pointers through byte-extract operations. Thus
-derferencing yields __CPROVER_memory, which results in a spurious verification
-failure.
+Value sets did not properly track pointers through unions. Thus derferencing
+yields __CPROVER_memory, which results in a spurious verification failure.
--- a/regression/cbmc/union16/main.c
+++ b/regression/cbmc/union16/main.c
@@ -1,0 +1,29 @@
+typedef void (*callback_t)(void);
+
+struct xfer
+{
+  union { /*< FIX: union -> struct */
+    struct
+    {
+      int *buf;
+      callback_t callback; /*< FIX: remove */
+    } a;
+  };
+} g_xfer;
+
+int g_buf;
+
+int main()
+{
+  g_xfer = (struct xfer){{{&g_buf}}};
+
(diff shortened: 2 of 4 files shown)


kroening:
It might be worth adding a comment in line 1286 to say that this case also covers unions.

kroening (approved):

thomasspriggs (approved):
Looks good to me. It makes sense to prioritise soundness over speed. Do you know how much of a performance difference this change makes? Do you have any ideas for doing field sensitivity for unions in a sound fashion?

peterschrammel (approved):

tautschnig:
> Looks good to me. It makes sense to prioritise soundness over speed. Do you know how much of a performance difference this change makes? Do you have any ideas for doing field sensitivity for unions in a sound fashion?

I haven't seen a performance penalty so far, and have more far-reaching union-related changes in the queue that do improve performance for me. I believe that, at present, a fair comparison is not possible, because we're simply missing the right entries from the value set and thus performance is, well, rather arbitrary. Performance might even be improved by this patch, because pointers can now be dereferenced and symex can work with actual objects when previously we'd try to work with `__CPROVER_memory`.

PR #5714, opened 2021-01-02
CBMC/JBMC --cover: only store traces with --trace to avoid memory exhaustion

tautschnig:
Not everyone needs test inputs, especially when programs lack INPUT
instructions anyway. This helped me avoid running out of memory on a
768GB system for a particular verification problem. The same problem can
now be run with just 75GB of memory.

### DIFF
--- a/regression/cbmc-cover/branch-loop1/test.desc
+++ b/regression/cbmc-cover/branch-loop1/test.desc
@@ -1,6 +1,6 @@
 CORE
 main.c
---xml-ui --cover branch
+--xml-ui --cover branch --show-test-suite
 activate-multi-line-match
 ^EXIT=0$
 ^SIGNAL=0$
--- a/doc/cprover-manual/test-suite.md
+++ b/doc/cprover-manual/test-suite.md
@@ -128,10 +128,11 @@ and outputs of interest for generating test suites.
 To demonstrate the automatic test suite generation in CBMC, we call the
 following command:
 
-    cbmc pid.c --cover mcdc --unwind 6 --xml-ui
+    cbmc pid.c --cover mcdc --show-test-suite --unwind 6 --xml-ui
 
 We'll describe those command line options one by one. The option `--cover mcdc`
-specifies the code coverage criterion. There
+specifies the code coverage criterion, and --show-test-suite requests that a
+test suite be printed. There
 are four conditional statements in the PID function: in lines 41,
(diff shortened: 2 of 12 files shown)


peterschrammel (approved):
Good workaround in the meanwhile. Storing traces and inputs should be decoupled, though, I think.
Besides, I should probably revive the PR that enables trace output as traces are produced rather than storing them at all (but that's quite a UI change which needs to be phased in properly).

regression/cbmc-cover/branch-loop1/test.desc
```diff
@@ -1,6 +1,6 @@
 CORE
 main.c
---xml-ui --cover branch
+--xml-ui --cover branch --trace
```
hannes-steffenhagen-diffblue: Seems like this is broken somehow?

martin-cs (commented):
Given the performance numbers this sounds very much like the kind of thing we want and I am in favour of not storing things we don't need.  But I am a little confused over the use-case for this. `--cover` creates a bunch of assertions for the relevant coverage metric and then solves for all of them.  Then, orthogonally(? is this what @peterschrammel is suggesting?) we could output for each location:

A. reached / unreachable.
B. inputs that would cause the relevant path to be taken.
C. the execution trace that reaches each one.

So this is only fro case C?  Are there users of this?  Also, can't we output these as we compute them?  I'm not sure I see why they need to be stored at all.

regression/cbmc-cover/branch-loop1/test.desc
tautschnig: Yes, it was broken as the code generating XML output is used for both trace and test-suite output. I have therefore changed the patch to introduce a new command-line option "show-test-suite."

tautschnig:
@peterschrammel @martin-cs I'd appreciate you taking another look at this for I've made substantial changes over the version you previously reviewed/approved:

1. I've introduced `merge_irept` to effectively compress the amount of memory consumed by a trace. This alone would have solved my out-of-memory problem as I've tested, but it seems the computational cost of building all the traces is still substantial (and often unnecessary).
2. Instead of using (abusing?) the "trace" command-line option, I've instead added a new option "show-test-suite."

It may still be desirable to print test inputs as-you-go, but that's not something on my priority queue (as I don't care about those inputs values at the moment).

martin-cs:
@tautschnig Apologies for the lag; first week of term.  I'm guessing by the merge that you are happy with this.

PR #5715, opened 2021-01-04
Introduction of Dockerfile for building and running CBMC in a Docker container.

NlightNFotis:
Hi,

This PR adds support for building and running CBMC in a docker container.

I would appreciate all feedback.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -349,3 +349,24 @@ jobs:
       - uses: actions/checkout@v2
       - name: Check for unused irep ids
         run: ./scripts/string_table_check.sh
+
+  check-docker-image:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Download test dependencies
+        run: sudo apt install openjdk-11-jdk-headless
+      - name: Build docker image
+        run: docker build -t cbmc .
+      - name: Smoke test goto-cc
+        run: docker run -v ${PWD}/.github/workflows/smoke_test_assets:/mnt/smoke -t cbmc goto-cc -o /mnt/smoke/test.goto /mnt/smoke/test.c
+      - name: Smoke test cbmc
+        run: docker run -v ${PWD}/.github/workflows/smoke_test_assets:/mnt/smoke -t cbmc cbmc /mnt/smoke/test.goto
+      - name: Smoke test jbmc
+        run: |
+          javac .github/workflows/smoke_test_assets/Test.java
+          docker run -v ${PWD}/.github/workflows/smoke_test_assets:/mnt/smoke -t cbmc jbmc --classpath /mnt/smoke Test
+      - name: Smoke test goto-analyzer
+        run: docker run -v ${PWD}/.github/workflows/smoke_test_assets:/mnt/smoke -t cbmc goto-analyzer /mnt/smoke/test.goto --unreachable-functions
--- a/.github/workflows/smoke_test_assets/test.c
+++ b/.github/workflows/smoke_test_assets/test.c
@@ -1,0 +1,9 @@
+#include <assert.h>
+
+#define return_val 0;
+
+int main(void)
+{
+  assert(1 == 1);
+  return return_val;
+}
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -300,6 +300,21 @@ the need to integrate JBMC as a separate project. Be aware that you need to
 change the build location (Select project in Eclipse -> Properties -> C/C++ 
 Build) to one of the src directories.
 
+# WORKING WITH DOCKER
+
+To compile and run the tools in a Docker container, do the following:
+
+1. From the root folder of the project, run `$ docker build -t cbmc .`
+2. After the building phase has finished, there should be a new 
+   image with the CProver binaries installed under `/usr/local/bin/`.
+
+   To start a container using that image as a base, run `$ docker run -i -t cbmc`
+   This will result in dropping you to a new terminal inside the container. To
+   load files for analysis into the container, one way is by mounting the folder
+   that contains the tests to the container. A possible invocation that does that
+   is: `$ docker run --mount type=bind,source=local/path/with/files,target=/mnt/analysis -i t cbmc`. In the
+   resulting container, the files present in the local file system under
+   `local/path/with/files` will be present under `/mnt/analysis`. 
 
 # OPTIONS AND VARIABLES
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+FROM ubuntu:20.04 as builder
+ENV DEBIAN_FRONTEND noninteractive
+ENV DEBCONF_NONINTERACTIVE_SEEN true
+# Timezone data is needed during the installation of dependencies,
+# since cmake depends on the tzdata package. In an interactive terminal,
+# the user selects the timezone at installation time. Since this needs
+# to be a non-interactive terminal, we need to setup some sort of default.
+# The UTC one seemed the most suitable one.
+RUN echo 'tzdata tzdata/Areas select Etc' | debconf-set-selections; \
+    echo 'tzdata tzdata/Zones/Etc select UTC' | debconf-set-selections; \
+    apt-get update && apt-get upgrade -y && apt-get install --no-install-recommends -y \
+    cmake \
+    ninja-build \
+    gcc \
+    g++ \
+    maven \
+    flex \
+    bison \
+    libxml2-utils \
+    patch
+COPY . /tmp/cbmc
+WORKDIR /tmp/cbmc
+RUN cmake -S . -Bbuild -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ && cd build; ninja -j2
+
+FROM ubuntu:20.04 as runner
+COPY --from=builder /tmp/cbmc/build/bin/* /usr/local/bin/
+RUN apt-get update && apt-get install -y gcc
--- a/.github/workflows/smoke_test_assets/Test.java
+++ b/.github/workflows/smoke_test_assets/Test.java
@@ -1,0 +1,6 @@
+public class Test {
+    public static void main(String[] args) {
+        System.out.println("Hi");
+        assert(1 == 1);
+    }
+}


Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ubuntu:21.04
+
+COPY /build/bin/* /home/cbmc/
```
hannes-steffenhagen-diffblue: You should probably just copy `/usr/bin`

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ubuntu:21.04
```
hannes-steffenhagen-diffblue: 21.04 isn't out yet

Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+
+COPY /build/bin/* /home/cbmc/
+
+CMD /bin/bash
```
hannes-steffenhagen-diffblue: I don't think this is necessary, I'd say this is what `docker run`

NlightNFotis:
So, updates on some of the points raised so far, and the future direction of it:

1. After a discussion with Hannes, it became obvious that the image shouldn't be that big - it is simply because the binaries I copied from my system are debug binaries, some of them more than 100MiB+. There are two ways I want to take this: first, it should be that copying non-debug binaries should have a direct effect on the image size. But this leaves us dependent on the platform the binaries are built (they need to be version compatible with the container distro). So a better approach overall might be to build on the container using staged builds. I will be investigating this next.
2. I will also try to find what's missing from relevant tools and install them in the container image (gcc, gdb, binutils, buildutils, etc)

NlightNFotis:
Update: yesterday we (me and @thomasspriggs) managed to get multi-stage builds enabled. The PR is more or less good for proper reviews now, but there's at least one more thing I would like to do to feel like this is ready (that is, add documentation for using the Dockerfile).

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    flex \
+    bison \
+    libxml2-utils \
+    dpkg-dev
+COPY . /tmp/cbmc
+WORKDIR /tmp/cbmc
+RUN cmake -S . -Bbuild -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ && cd build; ninja -j2
+
+FROM ubuntu:20.04 as runner
+COPY --from=builder /tmp/cbmc/build/bin/* /tmp/cbmc/
```
hannes-steffenhagen-diffblue: I still don't completely understand why this goes into `/tmp` rather than `/usr/bin`.

Dockerfile
NlightNFotis: This is from an older version of the commits of this PR? It's not like this now - it targets `20.04`

Dockerfile
NlightNFotis: This is not reflective of the file at its current state.

Dockerfile
NlightNFotis: Ditto.

Dockerfile
NlightNFotis: Fair point - I can make that happen. I started with putting the binaries under `/home` but this was weird looking, and then it was decided it might be better to be put under `/tmp`.

I agree `/usr/bin` is a more appropriate location - I have this on my todo as part of finalising this PR.

Dockerfile
thomasspriggs: We should probably copy the built binaries to `/usr/local/bin` as this is usually the standard directory on Linux for binaries which are not managed by the package manager. It also has the advantage of being in the path. (I think we were using `/tmp/cbmc` for testing purposes yesterday.)

Dockerfile
thomasspriggs: We should probably copy the built binaries to /usr/local/bin as this is usually the standard directory on Linux for binaries which are not managed by the package manager. It also has the advantage of being in the path. (I think we were using /tmp/cbmc for testing purposes yesterday.)

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    flex \
+    bison \
+    libxml2-utils \
+    dpkg-dev
+COPY . /tmp/cbmc
+WORKDIR /tmp/cbmc
+RUN cmake -S . -Bbuild -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ && cd build; ninja -j2
+
+FROM ubuntu:20.04 as runner
+COPY --from=builder /tmp/cbmc/build/bin/* /usr/local/bin/
```
piotr-grabalski, line 26: Are there any permission issues possible with copying binaries to /usr/local/bin? If it is done as root are there any issues with executing the binaries later on?

hannes-steffenhagen-diffblue: @piotr-grabalski docker provisioning always runs as root

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /tmp/cbmc/
```
hannes-steffenhagen-diffblue: Seems unnecessary?

Dockerfile
piotr-grabalski: cool, thanks @hannes-steffenhagen-diffblue!

NlightNFotis: Good question: The default user in a docker container is the `root` user so we don't have any problems with that.

COMPILING.md
```diff
@@ -300,6 +300,16 @@ the need to integrate JBMC as a separate project. Be aware that you need to
 change the build location (Select project in Eclipse -> Properties -> C/C++ 
 Build) to one of the src directories.
 
+# WORKING WITH DOCKER
+
+To compile and run the tools in a Docker container, do the following:
+
+1. From the root folder of the project, run `$ docker built -t cbmc .`
```
hannes-steffenhagen-diffblue: `docker build`?

COMPILING.md
```diff
@@ -300,6 +300,16 @@ the need to integrate JBMC as a separate project. Be aware that you need to
...
+2. After the building phase has finished, there should be a new 
+   image with the CProver binaries installed under `/usr/bin/`.
+
+   To start a container using that image as a base, run `$ docker run -i -t cbmc`
+   This will result in dropping you to a new terminal inside the container.
```
hannes-steffenhagen-diffblue: You probably want to mention how to make use of that here - by default docker only has access to the containers own filesystem, so you either want to mount e.g. the directory where your files are (via [`--mount`](https://docs.docker.com/storage/bind-mounts/)) or some other way.

hannes-steffenhagen-diffblue (commented):
I thought you'd add the CI tests for this in this PR, I know we track these as separate work items but that doesn't mean we need to create a PR for each right? Problem with merging as is is that right now it's untested code.

Dockerfile
NlightNFotis: It is definitely needed it seems - without it we get a CMake errors that the current directory doesn't contain a `CMakeLists.txt`.

It is also recommended by the [docker documentation to always use `WORKDIR` instead of `cd` or something else](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#workdir)

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:20.04 as builder
+ENV DEBIAN_FRONTEND noninteractive
+ENV DEBCONF_NONINTERACTIVE_SEEN true
+RUN echo 'tzdata tzdata/Areas select Etc' | debconf-set-selections; \
+    echo 'tzdata tzdata/Zones/Etc select UTC' | debconf-set-selections; \
```
hannes-steffenhagen-diffblue, line 10: What's this for? I feel like this could use an explanation (in the Dockerfile).

Dockerfile
hannes-steffenhagen-diffblue: @NlightNFotis I don't understand how that could be the case given this command is on the container that doesn't even have cmake installed, or invokes cmake in any way.

NlightNFotis: @hannes-steffenhagen-diffblue You are right. Apologies for that, I got confused with the one in line 17. This definitely seems extraneous.

Dockerfile
NlightNFotis: It has now been explained 👍🏻

.github/workflows/pull-request-checks.yaml
```diff
@@ -349,3 +349,24 @@ jobs:
       - uses: actions/checkout@v2
       - name: Check for unused irep ids
         run: ./scripts/string_table_check.sh
+
```
thomasspriggs, line 352: :pick: Might be worth stating in the commit title that this is PR check. Otherwise it might be confused with the release actions when looking through the commit history later. Something like - "Add github action on PRs to check the Dockerfile"
:pick: The commit title should not end with a full stop. See point "4. Do not end the subject line with a period", here - [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/)

COMPILING.md
```diff
@@ -300,6 +300,21 @@ the need to integrate JBMC as a separate project. Be aware that you need to
...
+
+1. From the root folder of the project, run `$ docker build -t cbmc .`
+2. After the building phase has finished, there should be a new 
+   image with the CProver binaries installed under `/usr/bin/`.
+
+   To start a container using that image as a base, run `$ docker run -i -t cbmc`
+   This will result in dropping you to a new terminal inside the container. To
+   load files for analysis into the container, one way is by mounting the folder
+   that contains the tests to the container. A possible invocation that does that
+   is: `$ docker run -v local/path/with/files:/mnt/analysis -i t cbmc`. In the
```
hannes-steffenhagen-diffblue: The Docker documentation strongly suggests to avoid using `-v` in favour of the slightly different `--mount` [here](https://docs.docker.com/storage/bind-mounts/)

> **Tip:** New users should use the --mount syntax. Experienced users may be more familiar with the -v or --volume syntax, but are encouraged to use --mount, because research has shown it to be easier to use.

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:20.04 as builder
+ENV DEBIAN_FRONTEND noninteractive
+ENV DEBCONF_NONINTERACTIVE_SEEN true
+# Timezone data is needed during the installation of dependencies,
+# since one of them depends on tzdata. In an interactive terminal,
```
hannes-steffenhagen-diffblue: which one?

.github/workflows/smoke_test_assets/test.c
```diff
@@ -0,0 +1,9 @@
+#include <assert.h>
```
hannes-steffenhagen-diffblue, line 1: ❓ Is the `smoke_tests` thing in any way docker specific? I don't necessarily think so, but if it is the name should reflect that.

hannes-steffenhagen-diffblue (approved):
LGTM

.github/workflows/smoke_test_assets/test.c
NlightNFotis: Well, it's being used only for testing the docker image soundness for now, but there's nothing that excludes any other usage.

COMPILING.md
NlightNFotis: I don't think it's easier than `-v`, especially for such a simple use case.

Apart from that, this is indicative usage, and not a docker tutorial. If a user prefers a different invocation, he can use that. Right now this is massively preferred by me, as it's much simpler than the `--mount` equivalent.

thomasspriggs: If the `--mount` syntax is easier to read then I am all for switching it over. As this is, it isn't necessarily clear which path is in the container and which is outside, without consulting the docker manual.

Dockerfile
NlightNFotis: It's in the list of additional packages installed by `cmake`. So the `cmake` package depends on it, either directly or indirectly.

tautschnig:
> So, updates on some of the points raised so far, and the future direction of it:
> 
> 1. After a discussion with Hannes, it became obvious that the image shouldn't be that big - it is simply because the binaries I copied from my system are debug binaries, some of them more than 100MiB+. There are two ways I want to take this: first, it should be that copying non-debug binaries should have a direct effect on the image size. But this leaves us dependent on the platform the binaries are built (they need to be version compatible with the container distro). So a better approach overall might be to build on the container using staged builds. I will be investigating this next.

You might be interested in #5646.

COMPILING.md
```diff
@@ -300,6 +300,21 @@ the need to integrate JBMC as a separate project. Be aware that you need to
 change the build location (Select project in Eclipse -> Properties -> C/C++ 
 Build) to one of the src directories.
 
+# WORKING WITH DOCKER
+
+To compile and run the tools in a Docker container, do the following:
+
+1. From the root folder of the project, run `$ docker build -t cbmc .`
+2. After the building phase has finished, there should be a new 
+   image with the CProver binaries installed under `/usr/bin/`.
```
tautschnig: They're in `/usr/local/bin/`.

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    gcc \
+    g++ \
+    maven \
+    flex \
+    bison \
+    libxml2-utils \
+    dpkg-dev
+COPY . /tmp/cbmc
+WORKDIR /tmp/cbmc
+RUN cmake -S . -Bbuild -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++ && cd build; ninja -j2
```
tautschnig, line 23: I don't think we need the `-DCMAKE_*_COMPILER` settings as those will be no different from the system defaults. And `ninja -C build -j2` would spare you the `cd build`.

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    apt-get update && apt-get upgrade -y && apt-get install --no-install-recommends -y \
+    cmake \
+    ninja-build \
+    gcc \
+    g++ \
+    maven \
+    flex \
+    bison \
+    libxml2-utils \
+    dpkg-dev
```
tautschnig: What is `dpkg-dev` needed for?

Dockerfile
hannes-steffenhagen-diffblue: Or `cmake --build build`

Dockerfile
hannes-steffenhagen-diffblue: I don't think it is needed for anything, I think this list might've been copied from one of our packaging builds.

Dockerfile
NlightNFotis: Hm, when it comes to the `WORKDIR` I think it might be better to actually keep it. From the [docker documentation](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers) it seems there's no longer any massive need to avoid extraneous instructions, as only a few introduce new layers, and I think it's cleaner this way.

What do you think about this?

Dockerfile
NlightNFotis: We need `patch` that this depends on and pulls as a dependency - not `dpkg-dev`. I will update this now.

martin-cs:
@NlightNFotis : what is the use-case for this?  If it is intended to be a way to distribute CBMC then should creating it be added to the CI or release automation?

NlightNFotis:
Hi @martin-cs 

Yes, my understanding is that this is one potential distribution method for us to maintain. The plan for this is is as follows:

1. There's going to be a github action check, run per PR, which is just building CBMC under docker and runs some basic image sanity checks (this is just to ensure that the image itself isn't broken by the changes introduced), and
2. There's going to be follow up work (which is going to come in a different PR most likely), whose job is to push a newly built image (corresponding to the newly-released version) to dockerhub.

Step 1 is present in this PR if you want to review it (it's the last commit).

martin-cs:
@NlightNFotis Thanks for the fast response.  At the risk of showing my ignorance about docker...

1. Why not run the regression and unit tests in the docker container?  They are our usual criteria for "is the system working".
2. If so, why not simply have one of the CI builds be in a docker container and then export the relevant container once everything has run?  That way we treat docker as 'just another platform' and have the same levels of testing and assurance on it.

How often will the docker image be produced and 'released' to dockerhub?  Every PR?  Every merge?  Every release?

NlightNFotis:
Hi @martin-cs, no problem all of these are good questions. Regarding:

1. We thought it was unneeded. The reasoning behind it is that the docker container is nothing special but an ubuntu 20.04 build (of which we are already building and testing in CI), so for reasons of economy of time (and a simplified first implementation) we felt that we could just make do with a very simple suite of smoke tests that tell us if the build is broken. We thought that an extensive test suite under a docker container wouldn't be terribly useful as it would just prolong the build time and not offer much more information than what we already get from the other jobs.
2. This is what the next PR is aiming for. However, I believe it's not a good idea to export the container at every PR (by export I'm assuming we both mean push the image to dockerhub) because of the possibility of it being broken and the fact that there's going to be a huge (i.e. spammy) collection of low quality releases - so a better idea would be to have an action that at a new CBMC release makes a new build of the docker container and pushes that, in a similar way to how we do update the homebrew package for CBMC at release.

martin-cs:
1. Could you move the Ubuntu 20.04 build /into/ the docker container or is that not how it works?  Or take the binaries from that build that have been tested?
2. Given the discussion about rolling releases, I think a docker container per release makes much more sense than per PR.

Out of interest and a desire to keep a vague track of how people use CBMC, is this driven by specific user requests or because "we" feel like it is something that should be available?

NlightNFotis:
Hi @martin-cs,

1. Well, in theory yes. Taking the binaries out might be a little bit weird, but I think it's doable. My preference though would be for the existing CI infrastructure to not change for one reason: consistency. I believe it would be weird if mac and windows builds are happening outside containers and for the linux build to be inside.

Moving the binaries from the other build into the container is a lot harder though, as that would involve different VMs (the ones spawned by the different github actions) so we would need to find a different way to store these binaries before they are either copied to the container or mounted to it from another directory.

2. Yes, this is the plan now.
3. Yes, this is driven by a specific user request for these images.

As an aside, we were experimenting with potentially speeding up the build inside the container by using `ccache`, so if we manage to do that, then it might be easier to also include more elaborate testing (by virtue of dropping the total build time). But that is still heavily experimental so we would much rather we proceed without `ccache` or heavier testing for now, as it can be added at a later point.

martin-cs:
@NlightNFotis This all sounds good, thanks.

I would dispute a little bit that consistency is a good thing in the CI builds.  I think we want diverse builds.  As it would be infeasibly expensive to try all possible combinations but I think having the set of build cover all of the options (but not all of the combinations).  Michael has tow commits for this happening with SAT solvers, I believe some of the build options have this kind of diversity as well.  So I would be OK with one of the Linux builds being in a docker container we then kept.

NlightNFotis:
@martin-cs Would that be okay if we do it at a later time? 

My plan for this is to get `ccache` to work reliably inside the container, and if we manage to get this to work, then we could port bits of the existing Ubuntu job to ensure that the container based one and the vm based one are in parity. Otherwise I'm slightly nervous to make this change now (looks like a relatively big one, with a lot of scope for breakage of the existing CI infrastructure).

martin-cs:
@NlightNFotis Sure!  I'm not in charge of CI or release, so, do as you wish. :smile:

PR #5716, opened 2021-01-06
Fix bug with function pointer initialisation in presence of statics

hannes-steffenhagen-diffblue:
Previously the code used some logic to skip generating selector cases
for static functions, which is what we want because we can't see
functions in the harness. Problem is if that happens to be the *last*
function we skip generating the "else" case at the end, so we end up
leaving the function pointer potentially uninitialised.

This just removes the functions we want to skip from the targets list.
Bonus: Loops without skips are easier to read.

Draft for now because I haven't had time to make a small reproducer for our regression tests yet.

### DIFF
--- a/src/goto-harness/recursive_initialization.cpp
+++ b/src/goto-harness/recursive_initialization.cpp
@@ -994,6 +994,29 @@ code_blockt recursive_initializationt::build_function_pointer_constructor(
     }
   }
 
+  targets.erase(
+    std::remove_if(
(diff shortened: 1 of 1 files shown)


hannes-steffenhagen-diffblue:
This seems to be an improvement but it seems that even with this handling of statics is still broken

tautschnig:
> This seems to be an improvement but it seems that even with this handling of statics is still broken

Could we have a test for this?

PR #5723, opened 2021-01-08
C front-end: hide our use of single-bit bool within sizeof

tautschnig:
In C, Boolean operations are of type int. Rejecting sizeof expressions
of Boolean operations breaks building the Linux kernel.

### DIFF
--- a/src/ansi-c/c_typecheck_expr.cpp
+++ b/src/ansi-c/c_typecheck_expr.cpp
@@ -950,7 +950,14 @@ void c_typecheck_baset::typecheck_expr_sizeof(exprt &expr)
   }
   else
   {
-    type.swap(to_unary_expr(expr).op().type());
+    const exprt &op = to_unary_expr(as_const(expr)).op();
+    // This is one of the few places where it's detectable
+    // that we are using "bool" for boolean operators instead
+    // of "int". We convert for this reason.
+    if(op.type().id() == ID_bool)
+      type = signed_int_type();
+    else
+      type = op.type();
   }
 
   exprt new_expr;
--- a/regression/ansi-c/sizeof4/test.desc
+++ b/regression/ansi-c/sizeof4/test.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 main.c
 
 ^EXIT=0$


martin-cs (commented):
Wasn't this why there was a distinction between `bool` and `c_bool`.  I thought they were 1 bit and 8 bit respectively and C defaulted to generating `c_bool` with `__CPROVER_bool` needed to generate actual `ID_bool`.  Perhaps I have got very confused.

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -931,7 +931,21 @@ void c_typecheck_baset::typecheck_expr_sizeof(exprt &expr)
...
+    const exprt &op = to_unary_expr(as_const(expr)).op();
+    // This is one of the few places where it's detectable
+    // that we are using "bool" for boolean operators instead
+    // of "int". We convert for this reason.
+    if(
+      op.type().id() == ID_bool &&
+      (op.id() == ID_not || op.id() == ID_and || op.id() == ID_or ||
+       op.id() == ID_equal || op.id() == ID_notequal || op.id() == ID_lt ||
+       op.id() == ID_le || op.id() == ID_gt || op.id() == ID_ge ||
+       op.id() == ID_if))
```
kroening: I would probably drop the check of the operator.
Say ?: is missing.

tautschnig: There's `ID_if` in the above, but I agree this could easily be missing some other operator. When writing this code I was wondering whether to just always cast, or whether some cases should be considered illformed input and be rejected by the type checker?

tautschnig:
> Wasn't this why there was a distinction between `bool` and `c_bool`. I thought they were 1 bit and 8 bit respectively and C defaulted to generating `c_bool` with `__CPROVER_bool` needed to generate actual `ID_bool`. Perhaps I have got very confused.

You are absolutely right, except the C front-end nowadays readily makes various "Boolean" operations (such as comparisons) a `bool` rather than an `int`. In some distant past we used to generate type casts in all sorts of places, but I think we stopped doing that. Or maybe I'm wrong and we never did that.

src/ansi-c/c_typecheck_expr.cpp
kroening: I think it's the right thing to always use `int` whenever "proper bool" is encountered in `sizeof`.

tautschnig: Ok, done!

kroening (approved):

PR #5726, opened 2021-01-08
Move expr2bits, bits2expr, try_get_string_data_array to simplify_utils

tautschnig:
They do not have any dependency on the simplify_exprt class (and thus
don't need to be class members). Similarly, try_get_string_data_array
has nothing to do with the simplify_exprt functions.


martin-cs (approved):
Seems reasonable.

kroening (approved):

PR #5731, opened 2021-01-08
GCC attribute alias should not trigger warning

tautschnig:
We spuriously warned about an extern symbol with value, where it was
only our handling of the "alias" attribute that introduced that value in
the first place. While at it, make that warning more useful.


martin-cs (approved):
I've had a look at the regression test and I think this is correct and an improvement.

peterschrammel (approved):

PR #5734, opened 2021-01-10
CBMC/SMT2 regression tests: distinguish slow from broken ones

tautschnig:
We should not conflate these two classes of problems.


kroening:
We did have THOROUGH as a top-level category, simply use that?

tautschnig:
> We did have THOROUGH as a top-level category, simply use that?

Could do, but then we'd have to have a separate .desc file for SMT (with tag `smt-backend`) as the test terminates quickly with the SAT back-end.

martin-cs (approved):
Yes; this is a very important distinction.

PR #5735, opened 2021-01-10
C front-end: evaluate __builtin_clz over constants

tautschnig:
The Linux kernel uses __builtin_clz(ll) in expressions that are expected
to be compile-time constants. To support this, evaluate such expressions
during type checking.

### DIFF
--- a/regression/cbmc-library/__builtin_clz-02/test.desc
+++ b/regression/cbmc-library/__builtin_clz-02/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+For this example, __builtin_clz is fully evaluated in the front-end. The
+_Static_assert ensures this is the case, as type checking fails when
+_Static_assert does not have a compile-time constant value.
--- a/src/ansi-c/c_typecheck_expr.cpp
+++ b/src/ansi-c/c_typecheck_expr.cpp
@@ -2591,6 +2591,40 @@ exprt c_typecheck_baset::do_special_functions(
 
     return std::move(popcount_expr);
   }
+  else if(
+    identifier == "__builtin_clz" || identifier == "__builtin_clzl" ||
+    identifier == "__builtin_clzll" || identifier == "__lzcnt16" ||
+    identifier == "__lzcnt" || identifier == "__lzcnt64")
+  {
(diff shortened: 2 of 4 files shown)


kroening:
We seem to have a variety of ways to assert that something can be simplified, and perhaps we should consolidate. This test, I believe, introduces a new way.
Would using `static_assert` be a good, standard-compilant alternative here?

tautschnig:
> We seem to have a variety of ways to assert that something can be simplified, and perhaps we should consolidate. This test, I believe, introduces a new way.
> Would using `static_assert` be a good, standard-compilant alternative here?

You're of course right, this is a much better approach. Done.

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2591,6 +2591,40 @@ exprt c_typecheck_baset::do_special_functions(
...
+    typecheck_function_call_arguments(try_constant);
+    exprt argument = try_constant.arguments().front();
+    simplify(argument, *this);
+    const auto int_constant = numeric_cast<mp_integer>(argument);
+
+    if(
+      !int_constant.has_value() || *int_constant == 0 ||
+      argument.type().id() != ID_unsignedbv)
+    {
+      return nil_exprt{};
```
kroening, line 2616: What happens here?

tautschnig: That just means it will be treated as a regular function call, not a "special function." So we fall back to using the library implementation.

regression/cbmc-library/__builtin_clz-02/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+For this example, __builtin_clz is fully evaluated in the front-end. This test
+ensures that the function call no longer shows up.
```
martin-cs: Does it actually check that the function call is gone?

martin-cs (approved):

regression/cbmc-library/__builtin_clz-02/test.desc
tautschnig: Thank you for pointing out this outdated comment! This is now fixed.

kroening (approved):

PR #5738, opened 2021-01-12
[RFC] Rewrite unions such that all members are of the same size

tautschnig:
In the C front-end, rewrite union components with types smaller than the
union's size to anonymous structs. Each such struct contains the
original union component plus padding. Assignments to union members thus
always assign all bytes that make up the object representation of a
union. The use of an anonymous struct ensures that member accesses can
still be resolved.

As this is a change in the semantics of goto programs, the goto binary
version is incremented. rewrite_union is no longer necessary, but bugs
(hidden by rewrite_union) in handling endianness in simplify_expr_member
and convert_member surfaced and had to be fixed.

Although this is stable and works well in all the testing I have done over the last weeks [edit: in a branch that has other bugfixes, most of which can be found in pull requests already], I am marking this [RFC] as this is a substantial change that could do with input especially from @kroening @martin-cs. A very natural further step would be making configurable how we handle the behaviour left unspecified by the C standard (cf. #5704 and #5705). Performance appears to be ~10% better on some large benchmarks, but I cannot currently claim having done a precise performance evaluation.


src/ansi-c/c_typecheck_initializer.cpp
```diff
@@ -526,19 +526,42 @@ exprt::operandst::const_iterator c_typecheck_baset::do_designated_initializer(
         // The first component is not the maximum member, which the (default)
         // zero initializer prepared. Replace this by a component-specific
         // initializer; other bytes have an unspecified value (C Standard
-        // 6.2.6.1(7)).
-        const auto zero =
-          zero_initializer(component.type(), value.source_location(), *this);
-        if(!zero.has_value())
+        // 6.2.6.1(7)). In practice, objects of static lifetime are fully zero
+        // initialized.
```
tautschnig, line 532: This is part one we could make configurable to cover all behaviours permitted by the C standard.

src/goto-symex/symex_assign.cpp
```diff
@@ -343,6 +341,30 @@ void symex_assignt::assign_struct_member(
...
+
+  // lhs must be member operand, which takes one operand, which must be a union.
+  exprt lhs_union = lhs.op();
+  const irep_idt &component_name = lhs.get_component_name();
+
+  // turn
+  //   a.c=e
+  // into
+  //   a'== { .c=e }
+  // as the front-end guarantees that .c will have the full size of the union
```
tautschnig, line 360: This is part two to make configurable: `symex_assignt::assign_struct_member` will have generated a `with_exprt` that selectively updates a single member (of what now is a struct with padding), keeping the rest intact. We could instead populate the rest of the struct with nondets.

martin-cs:
I haven't had a chance to look in detail yet but in principle this sounds like a good idea as it should simplify some of the handling of padding bits and how they are modified or not.  I think this may need a major version number change though.  Do you have a particular use-case in mind for this or is it a general "I think this would be simpler"?

tautschnig:
> I haven't had a chance to look in detail yet but in principle this sounds like a good idea as it should simplify some of the handling of padding bits and how they are modified or not. I think this may need a major version number change though. Do you have a particular use-case in mind for this or is it a general "I think this would be simpler"?

There probably isn't a need for doing in-depth looks, I'm happy to discuss the high-level idea first. What I have done is bumping the goto-program version, because previously compiled programs wouldn't have seen the changes now done in the front-end.

Why am I proposing to make this change? `union` expressions are a major source of `byte_extract` expressions being introduced, which are very general and ensure we get the semantics right, but often block constant propagation. It is my understanding that verification tasks involving unions are deemed performance bottlenecks for this very reason. I had also thought we could thereby get a grip on unsoundness when involving pointers, but that turned out to be much simpler (cf. #5713).

martin-cs:
Fair point; `byte_extract` and the like are cases where the memory typing model breaks and for performance for anything it's best if we avoid them.  I can see how it could help during symex.

I guess one question is how common are the cases where this actually helps.  It seems there are two (kinds?) of use cases for unions, one where you actually want type-unsafe access because you want to manipulate the representation of some data in a way that is not normally possible and one where you have a some kind of enum or discriminant that fixes which view of the union you will take.  It seems that only the later would benefit from this.

I presume during symex you are tracking which 'view' of the union is being used and then only using `byte_extract` when changing view?  If so then unions should have the same 'cost' as a struct.

tautschnig:
[...]
> I presume during symex you are tracking which 'view' of the union is being used and then only using `byte_extract` when changing view? If so then unions should have the same 'cost' as a struct.

symex treats the entire union as one object, but `WITH` expressions will effectively yield that concept of a "view."

tautschnig:
Closing in favour of #7230.

PR #5739, opened 2021-01-13
rol and ror expressions

kroening:

### DIFF
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -1362,6 +1362,46 @@ void smt2_convt::convert_expr(const exprt &expr)
         "unsupported type for " + shift_expr.id_string() + ": " +
         type.id_string());
   }
+  else if(expr.id() == ID_rol || expr.id() == ID_ror)
+  {
+    const shift_exprt &shift_expr = to_shift_expr(expr);
+    const typet &type = shift_expr.type();
+
+    if(
+      type.id() == ID_unsignedbv || type.id() == ID_signedbv ||
+      type.id() == ID_bv)
+    {
+      // SMT-LIB offers rotate_left and rotate_right, but these require a
+      // constant distance.
+      if(shift_expr.id() == ID_rol)
+        out << "((_ rotate_left";
+      else if(shift_expr.id() == ID_ror)
+        out << "((_ rotate_right";
+      else
+        UNREACHABLE;
+
+      out << ' ';
+
+      auto distance_int_op = numeric_cast<mp_integer>(shift_expr.distance());
+
+      if(distance_int_op.has_value())
+      {
+        out << distance_int_op.value();
+      }
+      else
+        UNEXPECTEDCASE(
+          "distance type for " + shift_expr.id_string() + "must be constant");
+
+      out << ") ";
+      convert_expr(shift_expr.op());
+
+      out << ")"; // rotate_*
+    }
+    else
+      UNEXPECTEDCASE(
+        "unsupported type for " + shift_expr.id_string() + ": " +
+        type.id_string());
+  }
   else if(expr.id()==ID_with)
   {
     convert_with(to_with_expr(expr));
--- a/src/util/std_expr.h
+++ b/src/util/std_expr.h
@@ -2503,8 +2503,7 @@ inline bitand_exprt &to_bitand_expr(exprt &expr)
   return static_cast<bitand_exprt &>(expr);
 }
 
-
-/// \brief A base class for shift operators
+/// \brief A base class for shift and rotate operators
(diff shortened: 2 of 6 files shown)


src/solvers/smt2/smt2_conv.cpp
```diff
@@ -1362,6 +1362,46 @@ void smt2_convt::convert_expr(const exprt &expr)
...
+
+      auto distance_int_op = numeric_cast<mp_integer>(shift_expr.distance());
+
+      if(distance_int_op.has_value())
+      {
+        out << distance_int_op.value();
+      }
+      else
+        UNEXPECTEDCASE(
+          "distance type for " + shift_expr.id_string() + "must be constant");
```
martin-cs, line 1393: Can't this be implemented as a pair of shifts and an or?

src/util/std_expr.h
```diff
@@ -2539,7 +2538,8 @@ class shift_exprt:public binary_exprt
 template <>
 inline bool can_cast_expr<shift_exprt>(const exprt &base)
 {
-  return base.id() == ID_shl || base.id() == ID_ashr || base.id() == ID_lshr;
+  return base.id() == ID_shl || base.id() == ID_ashr || base.id() == ID_lshr ||
+         base.id() == ID_ror || base.id() == ID_rol;
```
martin-cs, line 2542: There are `lshr_exprt` et al. so it would be good to have `ror_exprt` and `rol_exprt` for consistency.

martin-cs (commented):
What is here looks right (it would be nice to handle the roll-by-variable case) but what worries me is that not only is there no test, there is also no way to test as none of the front-ends actually generate these.

tautschnig:
> What is here looks right (it would be nice to handle the roll-by-variable case) but what worries me is that not only is there no test, there is also no way to test as none of the front-ends actually generate these.

Perhaps we could have a unit test?

kroening:
> Perhaps we could have a unit test?

That would introduce a dependency on an SMT solver, which could be quite inconvenient. And a dependency like that doesn't really fit the idea of a unit test.

The best I can think of is adding `__CPROVER_rotate_left` and `__CPROVER_rotate_right`.

tautschnig:
> > Perhaps we could have a unit test?
> 
> That would introduce a dependency on an SMT solver, which could be quite inconvenient. And a dependency like that doesn't really fit the idea of a unit test.
> 
> The best I can think of is adding `__CPROVER_rotate_left` and `__CPROVER_rotate_right`.

I don't think we'd have to go as far as that? The unit test could generate a goto program including rotate expressions, could trigger symbolic execution, and could then request an SMT-LIB 2 output file to be generated.

martin-cs:
On Wed, 2021-01-13 at 13:05 -0800, Michael Tautschnig wrote:
> > > Perhaps we could have a unit test?
> > 
> > That would introduce a dependency on an SMT solver, which could be
> > quite inconvenient. And a dependency like that doesn't really fit
> > the idea of a unit test.
> > 
> > The best I can think of is adding `__CPROVER_rotate_left` and
> > `__CPROVER_rotate_right`.
> 
> I don't think we'd have to go as far as that? The unit test could
> generate a goto program including rotate expressions, could trigger
> symbolic execution, and could then request an SMT-LIB 2 output file
> to be generated.

Either would be good.  Without something actually using rolls it feels
like it could very easily bit-rot.

Cheers,
 - Martin

tautschnig:
> > Perhaps we could have a unit test?
> 
> That would introduce a dependency on an SMT solver, which could be quite inconvenient. And a dependency like that doesn't really fit the idea of a unit test.
> 
> The best I can think of is adding `__CPROVER_rotate_left` and `__CPROVER_rotate_right`.

How about us adding support for Clang's `__builtin_rotate*` intrinsics (cf. https://releases.llvm.org/8.0.0/tools/clang/docs/LanguageExtensions.html)?

tautschnig (approved):

tautschnig:
@kroening I'm not entirely sure why only a single one of our regression tests is failing (check-macos-10_15-cmake-clang), but that does lead to two observations:

1. The following warnings are surprising:
```
file rotate.c line 42 function check_left64: function '__builtin_rotateleft64' is not declared
file rotate.c line 51 function check_right8: function '__builtin_rotateright8' is not declared
file rotate.c line 60 function check_right16: function '__builtin_rotateright16' is not declared
file rotate.c line 69 function check_right32: function '__builtin_rotateright32' is not declared
file rotate.c line 78 function check_right64: function '__builtin_rotateright64' is not declared
```
2. It actually because of
```
Running SMT2 QF_AUFBV using CPROVER SMT2
SMT2 solver returned error message:
	"line 107: expected ')' after rotate_left index"
```

PR #5740, opened 2021-01-13
Evaluate __builtin_ffs{,l,ll} over constants at compile time

tautschnig:
GCC evaluates this built-in at compile time, and the Linux kernel
assumes this is the case.


martin-cs (approved):
Yes but...

1. Do we have any tests for what happens when it is not a compile-time constant?
2. Given that this has quite a few a few friends that need to be (partially, constant only, etc.) evaluated before / during type checking, is it time to start thinking about an abstraction / pass / organisation of these?

tautschnig:
> Yes but...
> 
> 1. Do we have any tests for what happens when it is not a compile-time constant?

We do - the existing regression test (in which the constant cases have now been turned into `_Static_assert`s) already included that case.

> 2. Given that this has quite a few a few friends that need to be (partially, constant only, etc.) evaluated before / during type checking, is it time to start thinking about an abstraction / pass / organisation of these?

Fair point. One way of doing this is introducing `exprt`s for each of those functions (like we already have for `popcount`) and have the simplifier do all this work.

martin-cs:
I like the idea of the simplifier doing this because that way it can be used elsewhere, constant folded during symex, checked during trace construction, etc.  Not entirely sure each one needs an `ID`, but, then again, what would be better?

tautschnig:
> I like the idea of the simplifier doing this because that way it can be used elsewhere, constant folded during symex, checked during trace construction, etc. Not entirely sure each one needs an `ID`, but, then again, what would be better?

This is now built on top of #5879, introducing a new `find_first_set_exprt`.

kroening (approved):

PR #5742, opened 2021-01-13
Make simplify_exprt::get_values file-local

tautschnig:
This procedure is used in exactly one method, and does not use any class
members. Thus, move it to the file that method resides in. Also remove
the now-unnecessary forall_ macro and replace the previous use by ranged
for.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -55,7 +55,6 @@ ForEachMacros: [
   'Forall_irep',
   'forall_named_irep',
   'Forall_named_irep',
-  'forall_value_list',
   'forall_symbol_base_map',
   'forall_subtypes',
   'Forall_subtypes']
--- a/regression/cbmc/inequality-with-constant-normalisation1/main.c
+++ b/regression/cbmc/inequality-with-constant-normalisation1/main.c
@@ -1,0 +1,20 @@
+#include <assert.h>
+
+#ifdef _MSC_VER
+#  define _Static_assert(x, m) static_assert(x, m)
+#endif
+
+int main()
+{
+  _Bool b1, b2;
+
+  int nc = (b1 ? 1 : 2) == (b2 ? 1 : 2);
+  assert(b1 != b2 || nc != 0);
+
+  // The following are true for all values of b1, b2, and the simplifier should
+  // be able to figure this out.
+  _Static_assert((b1 ? 1 : 1) == (b2 ? 1 : 1), "");
+  _Static_assert(((b1 ? 2 : 3) >= (b2 ? 1 : 2)) != 0, "");
+  _Static_assert(((b1 ? 0 : 1) >= (b2 ? 2 : 3)) == 0, "");
+  _Static_assert(((b1 ? 0 : 1) >= 2) == 0, "");
+}
--- a/regression/cbmc/inequality-with-constant-normalisation1/test.desc
+++ b/regression/cbmc/inequality-with-constant-normalisation1/test.desc
@@ -1,0 +1,8 @@
+CORE
+main.c
+
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^warning: ignoring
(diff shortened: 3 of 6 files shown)


martin-cs (approved):
I believe the changes are as described and are advantageous, so approve...

.... but ...

Would you consider a more radical change.  There are some other code quality issues here.

*. The name is kind of ... generic, can we have something more specific?
*. No documentation of what it should do.
*. Why it does what it does is ... not entirely clear to me.  I guess you are supposed to iterate over the children of an arithmetic node and pull out all the constants used (for constant folding?) but ... then why the recursion on `ID_if`.
*. It does naive structural induction rather than a visit.  I suspect that for 99.999% of use cases that is fine but if we have another lot of auto-generated code with `() ? :` nested 700 deep, it could get very slow and this kind of issue is a pain to find.
*. The review notes on github say that it isn't covered.  If that isn't just a UI or CI race condition issue then ... maybe this is not needed at all?  Which might actually be more worrying.
*. The "`to_integer` returns false on success" threw me for a moment, maybe a comment?
*.

peterschrammel (approved):

kroening:
BTW, this kind of change is clearly a pattern.
I would like to question whether this is a good idea -- the coupling between the function and the class is lost. It only remains via the fact that they are in the same file, which is very indirect.
If the goal is to document that the function/method does not read or write the members of the class, then it could be made a static member function.

peterschrammel:
I've fought too many fights on this topic so that I've probably become less religious about it being a private static class member. The only rather minor advantage of the file-local variant I recognise is that there is no member declaration cluttering the header file and requiring maintenance.

tautschnig:
> BTW, this kind of change is clearly a pattern.
> I would like to question whether this is a good idea -- the coupling between the function and the class is lost. It only remains via the fact that they are in the same file, which is very indirect.
> If the goal is to document that the function/method does not read or write the members of the class, then it could be made a static member function.

In addition to what @peterschrammel said: I'm generally happy either way, except I'd expect a class member to be sufficiently well documented or otherwise clear in its purpose. The functions I'm making file-local in these PRs really can only be used in exactly the way they are being used right now, and even using them from other class members runs the risk of violating invariants. If that is resolved and the functions become more generally usable, then I'm all for making them static members.

tautschnig:
> I believe the changes are as described and are advantageous, so approve...
> 
> .... but ...
> 
> Would you consider a more radical change. There are some other code quality issues here.
> 

Thank you for raising this. I've now added a second commit that mostly covers your concerns, I believe:

> *. The name is kind of ... generic, can we have something more specific?

Renamed to `collect_constants` and ...

> *. No documentation of what it should do.

... documented.

> *. Why it does what it does is ... not entirely clear to me. I guess you are supposed to iterate over the children of an arithmetic node and pull out all the constants used (for constant folding?) but ... then why the recursion on `ID_if`.

Hopefully that has now become clearer.

> *. It does naive structural induction rather than a visit. I suspect that for 99.999% of use cases that is fine but if we have another lot of auto-generated code with `() ? :` nested 700 deep, it could get very slow and this kind of issue is a pain to find.

Ack, but not actually changed. It is possible that we have such expressions, but I am not sure I can get the restriction only-work-on-if-statements right with an iterator.

> *. The review notes on github say that it isn't covered. If that isn't just a UI or CI race condition issue then ... maybe this is not needed at all? Which might actually be more worrying.

I've now added a test that exercises this code. One aspect of the code, I believe, is actually dead (and I have added a comment saying as much) for other simplification rules already catch this beforehand.

> *. The "`to_integer` returns false on success" threw me for a moment, maybe a comment?
> *.

Replaced by `numeric_cast` (somewhere I've got a commit lying around that cleans out a lot of the `to_integer` uses already).

martin-cs (approved):
Much better -- thanks for making a second pass.

tautschnig:
> Much better -- thanks for making a second pass.

I've made a third pass: the test confirmed that this simplification rule doesn't even fully trigger, making the code either dead or useless. I have thus added a further commit that completely removes the code.

PR #5749, opened 2021-01-15
Dockerhub image push second set of fixes.

NlightNFotis:
This should fix docker complaining about invalid reference format passed to --tag
during `docker build`.

### DIFF
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -221,10 +221,13 @@ jobs:
         id: get_release_info
         uses: bruceadams/get-release@v1.2.0
       - name: Set Image Tag
-        run: echo "IMAGE_TAG=diffblue/cbmc:${{ steps.get_release_info.outputs.tag_name }}" >> $GITHUB_ENV
+        run: |
+          VERSION = ${{ steps.get_release_info.outputs.tag_name }}
+          arrVERSION=(${VERSION//-/ })
+          echo "IMAGE_TAG=diffblue/cbmc:${arrVERSION[1]}" >> $GITHUB_ENV
       - name: Build docker image
         run: docker build -t "$IMAGE_TAG" .
       - name: Push docker image to DockerHub
         run: |
           docker login --username=dbcicprover --password=${{ secrets.DOCKERHUB_ACCESS_DB_CI_CPROVER }}
-          docker image push $IMAGE_TAG
+          docker image push "$IMAGE_TAG"
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -222,6 +222,7 @@ jobs:
         uses: bruceadams/get-release@v1.2.0
       - name: Set Image Tag
         run: |
+          echo ${{  steps.get_release_info.outputs.tag_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-packages.yaml
```diff
@@ -223,7 +223,7 @@ jobs:
       - name: Set Image Tag
         run: echo "IMAGE_TAG=diffblue/cbmc:${{ steps.get_release_info.outputs.tag_name }}" >> $GITHUB_ENV
       - name: Build docker image
-        run: docker build -t $IMAGE_TAG .
+        run: docker build -t "$IMAGE_TAG" .
```
hannes-steffenhagen-diffblue, line 230: 🚫 What's it explaining about right now? What's the value of `$IMAGE_TAG` here? Image tags should be easy to reference by developers so shouldn't contain things like spaces or other weirdness

NlightNFotis: I think that might have been a mistake from my part as well when I made the release. Because of the release version action we are using, and the release names I chose which were kind of elaborate, like `cbmc-5.21.2-exp`, it might be using the whole tag as a name.

NlightNFotis: I mean the `" "` should certainly help in any case, but part of it was the release name I chose made the tag name more convoluted than it is allowed to be.

.github/workflows/release-packages.yaml
```diff
@@ -221,10 +221,13 @@ jobs:
...
-        run: echo "IMAGE_TAG=diffblue/cbmc:${{ steps.get_release_info.outputs.tag_name }}" >> $GITHUB_ENV
+        run: |
+          VERSION = ${{ steps.get_release_info.outputs.tag_name }}
+          arrVERSION=(${VERSION//-/ })
+          echo "IMAGE_TAG=diffblue/cbmc:${arrVERSION[1]}" >> $GITHUB_ENV
       - name: Build docker image
         run: docker build -t "$IMAGE_TAG" .
       - name: Push docker image to DockerHub
         run: |
           docker login --username=dbcicprover --password=${{ secrets.DOCKERHUB_ACCESS_DB_CI_CPROVER }}
```
piotr-grabalski, line 233: Would there be any overheads of keeping username in github secrets too? I think it might be easier to maintain if we ever change the username if it is (or is going to be) used in multiple places. It would make the credentials more secure too.

piotr-grabalski (approved):

.github/workflows/release-packages.yaml
NlightNFotis: I asked Peter about it, but he said that this doesn't need to be a secret.

piotr-grabalski: OK, do you think you will need this in any other place so that it would be worth of keeping it as a define somewhere above?

NlightNFotis: I don't think it's going to be needed anywhere else to be honest.

And if it is, I can do the refactoring later when there is evidence it's needed.

thomasspriggs: The benefit of keeping the username as a secret would be that forks of CBMC could supply different secrets rather than diverge the code. However it shouldn't make any difference to us.

PR #5753, opened 2021-01-15
Make guard_exprt::operator-= work with individual conditions

tautschnig:
Before, operator-= would only apply when both of the operands were
conjunctions. Yet it is perfectly safe use when one or both operands
aren't conjunctions.


martin-cs (approved):
Obviously I wonder "could this go further" but I think the point is to keep it light weight.  The BDD patches et al. are probably a better solution than much more explicit case-by-case handling.

peterschrammel (approved):

PR #5755, opened 2021-01-16
VSD: Simplifying abstract_objectt::write, collapsing inheritance hierarchy

jezhiggins:
This PR simplifies the `write` member function implementation in `abstract_objectt` subclasses. It collapses the inheritance hierarchy for classes modelling aggregate values (arrays, structs, unions), eliminating duplication. 

Several classes were named to better reflect their purpose - eg `union_abstract_objectt` became `two_value_union_abstract_objectt`, and some new classes introduced to in place of a previously existing base class, eg `two_value_array_abstract_objectt` in place of the eliminated `array_abtract_objectt`. 

This allowed further work to combine `interval_array_abstract_objectt` and `constant_array_abstract_objectt` together. The combined implementation has been renamed `full_array_abstract_objectt`, and the implementation refactored for reading comprehension. 

This PR is behaviour preserving. No need functionality is introduced.


src/analyses/variable-sensitivity/abstract_enviroment.cpp
```diff
@@ -66,17 +66,9 @@ abstract_environmentt::eval(const exprt &expr, const namespacet &ns) const
...
-    abstract_object_pointert pointer_object =
-      abstract_object_factory(simplified_expr.type(), simplified_expr, ns);
-
-    // Store the abstract object in the pointer
-    return pointer_object;
-  }
   else if(
     simplified_id == ID_array || simplified_id == ID_struct ||
-    simplified_id == ID_constant)
+    simplified_id == ID_constant || simplified_id == ID_address_of)
```
martin-cs, line 62: It is true that it is handled the same but it is for quite different reasons.  This is not necessarily a problem, but it is something to be aware of.

src/analyses/variable-sensitivity/abstract_aggregate_object.h
```diff
@@ -0,0 +1,107 @@
+#ifndef CBMC_ABSTRACT_AGGREGATE_OBJECT_H
+#define CBMC_ABSTRACT_AGGREGATE_OBJECT_H
+
+#include <analyses/variable-sensitivity/abstract_object.h>
+#include <analyses/variable-sensitivity/abstract_environment.h>
+#include <stack>
+
+class member_exprt;
+
+template<class aggregate_typet, class aggregate_traitst>
```
martin-cs: This is probably a bit more template-y than we normally write but under the circumstances I can see this reduces an amount of duplication.

src/analyses/variable-sensitivity/interval_array_abstract_object.cpp
```diff
@@ -60,27 +60,15 @@ abstract_object_pointert interval_array_abstract_objectt::write_component(
...
-      stack,
-      index_exprt(index_expr.array(), index_interval.get_lower()),
-      value,
-      merging_write);
-  }
-  else if(
+  if(
     !index_interval.is_top() && !index_interval.is_bottom() &&
     index_interval.get_lower().id() != ID_min &&
     index_interval.get_upper().id() != ID_max)
```
martin-cs: Although not touched by this PR, this case really should result in the interval being top.

src/analyses/variable-sensitivity/abstract_value_object.h
```diff
@@ -0,0 +1,32 @@
...
+  abstract_value_objectt(
+    const exprt &expr,
+    const abstract_environmentt &environment,
+    const namespacet &ns)
+    : abstract_objectt(expr, environment, ns)
+  {
+  }
+};
+
+#endif
```
martin-cs, line 72: New line?

src/analyses/variable-sensitivity/abstract_value_object.h
```diff
@@ -0,0 +1,32 @@
+#ifndef CPROVER_ANALYSES_VARIABLE_SENSITIVITY_ABSTRACT_VALUE_OBJECT_H
+#define CPROVER_ANALYSES_VARIABLE_SENSITIVITY_ABSTRACT_VALUE_OBJECT_H
+
+#include <analyses/variable-sensitivity/abstract_object.h>
+
+class abstract_value_tag { };
```
martin-cs: I'm not sure I follow why this is needed...

src/analyses/variable-sensitivity/interval_abstract_value.cpp
```diff
@@ -537,6 +569,17 @@ interval_abstract_valuet::interval_abstract_valuet(
 {
 }
 
+index_range_ptrt interval_abstract_valuet::index_range() const
+{
+  if (is_top() || is_bottom() || interval.is_top() || interval.is_bottom())
+    return std::make_shared<empty_index_ranget>();
+  if (interval.get_lower().id() == ID_min || interval.get_upper().id() == ID_max)
+    return std::make_shared<empty_index_ranget>();
```
martin-cs: I get that if the interval is bottom you should get an empty range but I'm not sure I get why this is the case for top or if the interval is open at one end.

unit/analyses/variable-sensitivity/abstract_object/index_range.cpp
```diff
@@ -40,14 +40,18 @@ SCENARIO(
     }
   }
 
-  GIVEN("a top constant's range is empty")
+  GIVEN("a top constant's range is has a single nil expression")
```
martin-cs, line 46: If it tis top then it should be everything so this is definitely an improvement but I'm not quite sure what the nil is supposed to describe.

src/analyses/variable-sensitivity/abstract_object.h
```diff
@@ -443,18 +443,6 @@ class abstract_objectt : public std::enable_shared_from_this<abstract_objectt>
   /// abstraction anyway
   bool should_use_base_meet(const abstract_object_pointert &other) const;
 
-  template <class keyt>
```
martin-cs, line 446: Good tidy up.

src/analyses/variable-sensitivity/full_array_abstract_object.cpp
```diff
@@ -272,113 +270,137 @@ abstract_object_pointert full_array_abstract_objectt::write_element(
       environment, ns, stack, expr, value, merging_write);
   }
 
+  if(!stack.empty())
```
martin-cs, line 251: It's not entirely clear to me why this is needed.  Can't you just delegate the write to the relevant `abstract_object` and then update the relevant entry with whatever it returns?

src/analyses/variable-sensitivity/full_array_abstract_object.cpp
```diff
@@ -376,13 +376,10 @@ abstract_object_pointert full_array_abstract_objectt::write_leaf_element(
     }
     else
     {
-      auto const old_value = result->map.find(index_value);
+      auto old_value = result->map.find(index_value);
       if(old_value.has_value())
```
martin-cs, line 354: It may not change the value that is stored but breaking sharing is a serious concern.  Please check that the sharing map does actually handle that case correctly.

martin-cs (approved):
I don't see any immediate show stoppers but this is quite a substantial refactor and it would be good to work through the code-base with you before we merge this PR.

src/analyses/variable-sensitivity/abstract_aggregate_object.h
jezhiggins: It's compile type polymorphism rather than run-time. We don't need, actually don't want, run-time polymorphism here.

src/analyses/variable-sensitivity/abstract_value_object.h
jezhiggins: I'd initially expected `abstract_value_objectt` to be a template in a similar way to `abstract_aggregate_objectt`. Didn't turn out that way, although may yet with the next bit of work. Happy to remove this now if you'd prefer.

src/analyses/variable-sensitivity/interval_abstract_value.cpp
jezhiggins: It doesn't make sense to me either, but this is the existing behaviour.

unit/analyses/variable-sensitivity/abstract_object/index_range.cpp
jezhiggins: The index_range is only used to during array access, and this supports the existing behaviour. Returning a nil expression here takes us down to https://github.com/jezhiggins/cbmc/blob/b618d2e3703434aa0cfb4d1c991c867b1dea9955/src/analyses/variable-sensitivity/full_array_abstract_object.cpp#L298 
It is a special case, but it's isolated here rather than having to be explicitly handed in potentially several places elsewhere. As we look to expand other value representations, the more we can keep within those representations, the better.

src/analyses/variable-sensitivity/full_array_abstract_object.cpp
jezhiggins: No, because once we get to the end of the stack we need to worry about whether it's a merging write or not. Higher up the stack, we don't need to do that. 

We have a similar decision point in `full_struct_abstract_objectt`, and I assume we'll have one in our union representation when we fill that out. When we get there, it's entirely possible we'll be able to collapse the implementations.

martin-cs:
I've had a discussion with @jezhiggins and I am in favour of merging this because:

1. It is at least an intermediate step towards sorting this code out.
2. It is code that I requested we merge "to be improved in tree".
3. It makes things simpler.
4. It is a pure refactoring.  Jez has very intentionally preserved functionality, even when questionable.

martin-cs:
@tautschnig @peterschrammel This is part of the refactor and tidy up of VSD.  I don't expect you to look at that part of the commit, i think it is a net improvement and there is more to do.  Please could you have a look at the changes in other parts of the tree, notably `util/' which are blocking the merge.

src/analyses/variable-sensitivity/abstract_enviroment.cpp
```diff
@@ -17,8 +17,8 @@
 #include <analyses/variable-sensitivity/pointer_abstract_object.h>
 #include <analyses/variable-sensitivity/struct_abstract_object.h>
 #include <analyses/variable-sensitivity/variable_sensitivity_object_factory.h>
-
 #include <util/pointer_expr.h>
+#include <language_util.h>
```
tautschnig: Seems surprising?

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -14,7 +14,7 @@
 #include <analyses/variable-sensitivity/two_value_struct_abstract_object.h>
 #include <analyses/variable-sensitivity/variable_sensitivity_object_factory.h>
 #include <util/pointer_expr.h>
-#include <language_util.h>
+#include <langapi/language_util.h>
```
tautschnig: Why is this file being included?

unit/analyses/variable-sensitivity/constant_abstract_value/constant_index_range.cpp
```diff
@@ -0,0 +1,53 @@
+#include <analyses/variable-sensitivity/abstract_environment.h>
```
tautschnig: Quoting `CODING_GUIDELINES.md`: "Each source and header file must start with a comment block stating the author."

src/analyses/variable-sensitivity/abstract_value_object.cpp
```diff
@@ -0,0 +1,47 @@
+#include <analyses/variable-sensitivity/abstract_value_object.h>
```
tautschnig, line 9: See earlier comment: please make sure all files include a header stating the author.

src/analyses/variable-sensitivity/full_array_abstract_object.cpp
```diff
@@ -14,6 +14,12 @@
 
 #include "full_array_abstract_object.h"
 
+bool eval_index(
```
tautschnig, line 17: Should likely be `static` as only required within this file.

src/util/sharing_map.h
```diff
@@ -455,6 +455,12 @@ class sharing_mapt
     get_view(result);
     return result;
   }
+  sorted_viewt get_sorted_view() const
+  {
+    sorted_viewt result;
+    get_view(result);
+    return result;
+  }
```
tautschnig, line 467: Would you mind adding documentation? It may be ok to put the three methods in a group, but there should be some sort of documentation.

src/analyses/variable-sensitivity/abstract_object.h
```diff
@@ -258,6 +258,9 @@ class abstract_objectt : public std::enable_shared_from_this<abstract_objectt>
     abstract_object_pointert op1,
     abstract_object_pointert op2,
     bool &out_modifications);
+  static abstract_object_pointert merge(
+    abstract_object_pointert op1,
+    abstract_object_pointert op2);
```
tautschnig: How about changing the interface to make `merge` return a pair?

tautschnig (approved):
Some administrative comments only, I have browsed (but not properly reviewed) the code in analyses/ and focussed on the single change in src/util/ only.

src/analyses/variable-sensitivity/abstract_object.h
jezhiggins: I did try it, but it didn't really make the code any clearer. If structured bindings were available, I'd absolutely have gone for it.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: Good question - looking at now, I probably included for some debug/dev code which didn't need to be committed. I'll tidy this up.

PR #5758, opened 2021-01-17
Handle bool expressions in implicit_typecast_arithmetic

### LINKED ISSUE #5103: goto-cc crashes on softfloat code using __int128
CBMC version: gcc (goto-cc 5.12 (crest.merged-48-gabe78fc77-dirty)) 7.4.0
Operating system: 4.15.0-45-generic x86_64 GNU/Linux
Exact command line resulting in the issue: goto-cc sf.c
What behaviour did you expect: no crash
What happened instead: invariant failure

The following is a reduced version of the softfloat code causing the crash:
file sf.c:
```
void softfloat_mul128To256M(unsigned a)
{
  unsigned __int128 z;
  z = (unsigned __int128) a;
  z += (a < 1);
}
```
command: `goto-cc sf.c`

output:
```
Invariant check failed
File: ../util/std_types.h:1091 function: to_bitvector_type
Condition: can_cast_type<bitvector_typet>(type)
Reason: Precondition
...
```

tautschnig:
We use Booleans instead of int as type of binary predicates. This must
not preclude mixing binary predicates with bitvector-typed expressions.

Fixes: #5103

### DIFF
--- a/src/ansi-c/c_typecast.cpp
+++ b/src/ansi-c/c_typecast.cpp
@@ -650,24 +650,8 @@ void c_typecastt::implicit_typecast_arithmetic(
 
   if(max_type==LARGE_SIGNED_INT || max_type==LARGE_UNSIGNED_INT)
   {
-    // get the biggest width of both
-    std::size_t width1 = to_bitvector_type(type1).get_width();
-    std::size_t width2 = to_bitvector_type(type2).get_width();
-
     // produce type
-    typet result_type;
-
-    if(width1==width2)
-    {
-      if(max_type==LARGE_SIGNED_INT)
-        result_type=signedbv_typet(width1);
-      else
-        result_type=unsignedbv_typet(width1);
-    }
-    else if(width1>width2)
-      result_type=type1;
-    else // width1<width2
-      result_type=type2;
+    typet result_type = (max_type == c_type1) ? type1 : type2;
 
     do_typecast(expr1, result_type);
     do_typecast(expr2, result_type);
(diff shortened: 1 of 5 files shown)


src/ansi-c/c_typecast.cpp
```diff
@@ -619,8 +619,10 @@ void c_typecastt::implicit_typecast_arithmetic(
   if(max_type==LARGE_SIGNED_INT || max_type==LARGE_UNSIGNED_INT)
   {
     // get the biggest width of both
-    std::size_t width1 = to_bitvector_type(type1).get_width();
-    std::size_t width2 = to_bitvector_type(type2).get_width();
+    std::size_t width1 =
```
kroening: Would it make sense to use c_type1 and c_type2 instead of type1 and type2?

martin-cs: Agreed; this should be post-promotion.

martin-cs (commented):
I think @kroening 's suggestion is correct.

src/ansi-c/c_typecast.cpp
tautschnig: But then we'd need new facilities to turn a `c_typet` into a `typet`. It's kind-of done by c_typecastt::implicit_typecast_arithmetic, but that doesn't do anything about a `LARGE_SIGNED_INT` (or `LARGE_UNSIGNED_INT`).

tautschnig: I've now implemented this without a need for `to_bitvector_type` at all. Maybe I'm missing something? The new implementation seems almost too simple.

chris-ryder (approved):
Yeah, that does seem rather simple :-) but I can't at the moment see why it'd need to be more complex 🤷

PR #5759, opened 2021-01-17
move bitvector-related expression classes into separate header file

kroening:
`std_expr.h` is by far the largest header file, which impedes on readability.
This commit moves the bitvector-related expression classes into a separate
header file, following the pattern in `mathematical_expr.h`.


martin-cs (approved):
If we are going the route of splitting up `std_expr.h`, then, yes, clearly.

src/util/bitvector_expr.h
```diff
@@ -0,0 +1,676 @@
...
+Module: API to expression classes for bitvectors
+
+Author: Daniel Kroening, kroening@kroening.com
+
+\*******************************************************************/
+
+#ifndef CPROVER_UTIL_BITVECTOR_EXPR_H
+#define CPROVER_UTIL_BITVECTOR_EXPR_H
+
+/// \file util/std_expr.h
```
tautschnig: Copy&paste error.

src/util/bitvector_expr.h
```diff
@@ -0,0 +1,676 @@
...
+/// API to expression classes for bitvectors
+
+#include "std_expr.h"
```
tautschnig, line 15: I understand that we need this right now, but could we consider adding something like `abstract_expr.h` that declares the likes of `unary_exprt` so that we can eventually speed up compile time a bit?

tautschnig (approved):
I believe this is a great first step to make things more manageable for humans. Now I'm wondering whether we can also do some work to help the compiler having to do a little less work, see my comment below.

src/util/bitvector_expr.h
kroening: Yes, that's the plan. I am considering to have `unary_exprt`, `binary_exprt` etc. in `expr.h`.

src/util/bitvector_expr.h
kroening: Fixed.

peterschrammel (approved):

PR #5760, opened 2021-01-18
move floating-point-related expression classes into separate header file

kroening:
`std_expr.h` is by far the largest header file, which impedes on readability.
This commit moves the floating-point-related expression classes into a
separate header file, following the pattern in `mathematical_expr.h`.


src/solvers/flattening/boolbv.h
```diff
@@ -31,6 +31,8 @@ class bswap_exprt;
 class concatenation_exprt;
 class extractbit_exprt;
 class extractbits_exprt;
+class floatbv_typecast_exprt;
+class ieee_float_op_exprt;
```
martin-cs, line 35: Why are these forward declared when other types are not.  I can see cases for forward declaration or inclusion but the mix seems unclear.

martin-cs (approved):

src/solvers/flattening/boolbv.h
kroening: The algorithm is to count the .cpp files that includes this .h file that include the omitted header file. If that is a high percentage, then use the include, otherwise use the forward declaration.

In this case, a lot of .cpp files (46) include boolbv.h but do not need the full floatbv_expr.h (32).

peterschrammel (approved):

tautschnig (approved):

PR #5762, opened 2021-01-19
[CI] Dockerhub image push action: commit to one syntax for string manipulation.

NlightNFotis:
Commit to one syntax for isolating the version
number, remove the debug statements and add an
explanatory comment.

### DIFF
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -217,11 +217,10 @@ jobs:
           submodules: recursive
       - name: Set Image Tag
         run: |
-          echo ${{ github.ref }}
-          echo $GITHUB_REF
-          echo $GITHUB_REF | cut -d "/" -f 3 >> $VERSION
-          arrVERSION=(${VERSION//-/ })
-          echo "IMAGE_TAG=diffblue/cbmc:${arrVERSION[1]}" >> $GITHUB_ENV
+          # Isolate the version number from a reference to a tag, for example,
+          # '5.20.3' from a string like 'refs/tags/cbmc-5.20.3-exp'
+          VERSION=$(echo ${{ github.ref }} | cut -d "/" -f 3 | cut -d "-" -f 2)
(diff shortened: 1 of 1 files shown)


.github/workflows/release-packages.yaml
```diff
@@ -217,11 +217,10 @@ jobs:
...
       - name: Set Image Tag
         run: |
-          echo ${{ github.ref }}
-          echo $GITHUB_REF
-          echo $GITHUB_REF | cut -d "/" -f 3 >> $VERSION
-          arrVERSION=(${VERSION//-/ })
-          echo "IMAGE_TAG=diffblue/cbmc:${arrVERSION[1]}" >> $GITHUB_ENV
+          # Isolate the version number from a reference to a tag, for example,
+          # '5.20.3' from a string like 'refs/tags/cbmc-5.20.3-exp'
+          VERSION=$(echo ${{ github.ref }} | cut -d "/" -f 3 | cut -d "-" -f 2)
```
piotr-grabalski, line 222: is the tag reference always in that form? I wonder if that can go wrong in the future if tag is placed under different path

piotr-grabalski (approved):

.github/workflows/release-packages.yaml
piotr-grabalski: Maybe use `basename`? 
something like:
`VERSION=$(basename ${{ github.ref }}  | cut -d "-" -f 2)`

NlightNFotis: This is related to git internals, so I doubt it's going to change anytime soon https://git-scm.com/book/en/v2/Git-Internals-Git-References

hannes-steffenhagen-diffblue: @piotr-grabalski As to the tagname starting with `refs/tags/`, yes, it'll be in that form. This is how it's handled in git and unlikely to change, certainly not without a major version change in git itself. As to the tagname itself, this should always be `cbmc-$VERSION`, if we're changing that of course we need to change our scripts but I'm not sure why we would change it (also we only trigger releases if the tag has that form in the first place).

PR #5764, opened 2021-01-19
Refine recursive calls within the simplifier

### LINKED ISSUE #4988: Various uses of simplify_node are unnecessary or wrong
See #4982 for comments, this issue is just for tracking purposes.

tautschnig:
1) When we know the id of an expression, use the corresponding simplify_*
method rather than the generic simplify_node.
2) Avoid simplify_rec when all operands have been simplified already.
3) Don't call simplify_rec (or simplify_node) when there is nothing to
simplify.

Fixes: #4988

### DIFF
--- a/src/util/simplify_expr_boolean.cpp
+++ b/src/util/simplify_expr_boolean.cpp
@@ -38,7 +38,7 @@ simplify_exprt::resultt<> simplify_exprt::simplify_boolean(const exprt &expr)
     binary_exprt new_expr = implies_expr;
     new_expr.id(ID_or);
     new_expr.op0() = simplify_not(not_exprt(new_expr.op0()));
-    return changed(simplify_node(new_expr));
+    return changed(simplify_boolean(new_expr));
   }
   else if(expr.id()==ID_xor)
   {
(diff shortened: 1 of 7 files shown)


src/util/simplify_expr_boolean.cpp
```diff
@@ -207,18 +207,14 @@ simplify_exprt::resultt<> simplify_exprt::simplify_not(const not_exprt &expr)
...
   }
   else if(op.id() == ID_forall) // !(forall: a) <-> exists: not a
   {
     auto const &op_as_forall = to_forall_expr(op);
-    exists_exprt rewritten_op(
-      op_as_forall.symbol(), not_exprt(op_as_forall.where()));
-    rewritten_op.where() = simplify_node(rewritten_op.where());
-    return std::move(rewritten_op);
+    return exists_exprt{op_as_forall.symbol(),
+                        simplify_not(not_exprt(op_as_forall.where()))};
```
martin-cs, line 214: Nice!

martin-cs (approved):
Yes but...

I believe this PR is a correct implementation of what it claims and is one way of improving the code; this is definitely faster than redoing the dispatch when the ID is known.

But... I'm not sure it's the only way of improving it.  This makes it harder to implement solutions where `simplify_node` is caching or enforces some kind of sharing-aware graph visit rather than potentially exponential recursion.  I don't have the bandwidth to do anything like that at the moment (I don't even know if it is a good / widely accepted idea) but I wouldn't want to make it harder than it already is.

tautschnig:
> Yes but...
> 
> I believe this PR is a correct implementation of what it claims and is one way of improving the code; this is definitely faster than redoing the dispatch when the ID is known.
> 
> But... I'm not sure it's the only way of improving it. This makes it harder to implement solutions where `simplify_node` is caching or enforces some kind of sharing-aware graph visit rather than potentially exponential recursion. I don't have the bandwidth to do anything like that at the moment (I don't even know if it is a good / widely accepted idea) but I wouldn't want to make it harder than it already is.

I agree that it may be a good idea to always go through `simplify_node`, for the reasons you called out. But then we should indeed _always_ do that, and not just sometimes (which is what we did before this patch). So this patch makes use uniformly _not_ use `simplify_node` when the ID is known. I believe it should be fairly easy to flip all calls to `simplify_*` to `simplify_node` if so desired. But then we should really do it for _all_ of them. (I am also happy to make this happen in this PR instead of what's being proposed right now.)

martin-cs:
Agreed; it should be one way or the other.  As I am not an "owner" for this area of code and don't have time to work on it... I don't feel like I should have a strong say.  I think this is an improvement over the current state-of-affairs and until we have someone who can do something fancy (graph traversal, caching, etc.) with `simplify_node` then this is probably the way to go.

peterschrammel (approved):
Yes, this makes sense to call the right case directly when you already know which case it is going to be.

PR #5768, opened 2021-01-19
Fix doxygen in cmake

hannes-steffenhagen-diffblue:
This has been broken for 3 years, ever since doxygen.cfg was renamed to
doxyfile. This lets us build documentation in from cmake with the 'doc'
target. One difference to the previous version (i.e. the one that was
broken before) is that this creates the documentation in the cmake
binary directory now instead of the source directory. Reasoning for this
is that cmake targets should not change files in the source directory
ideally.

---

Not addressed here: I think this is probably the wrong place to do doxygen stuff anyway, we're referring to files outside this directory which is somewhat unfortunate.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -66,7 +66,7 @@ jobs:
           # user input
           DEBIAN_FRONTEND: noninteractive
         run: |
-          sudo apt-get install --no-install-recommends -yq cmake ninja-build gcc g++ maven flex bison libxml2-utils dpkg-dev ccache
+          sudo apt-get install --no-install-recommends -yq cmake ninja-build gcc g++ maven flex bison libxml2-utils dpkg-dev ccache doxygen
       - name: Prepare ccache
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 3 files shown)


tautschnig:
@hannes-steffenhagen-diffblue Would you mind adding a test by invoking the target from one of the GitHub actions that use `cmake`?

hannes-steffenhagen-diffblue:
@tautschnig how about that?

.github/workflows/pull-request-checks.yaml
```diff
@@ -84,6 +84,10 @@ jobs:
           mkdir build
           cd build
           cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=/usr/bin/gcc -DCMAKE_CXX_COMPILER=/usr/bin/g++
+      - name: Check that doc task works
+        run: |
+          cd build
+          ninja doc
```
tautschnig, line 90: Nit pick: `ninja -C build doc` will do the same in a single line.

tautschnig (approved):

piotr-grabalski (approved):

NlightNFotis (approved):
Looks good to me

PR #5772, opened 2021-01-19
Replace boolbv_mapt::map_bitt by bvt

tautschnig:
The Boolean flag is_set does not need to be stored for each individual
bit as all entries will either be false (initial state) or be true. It
is not possible for only some of the bits to be set.

On the benchmark that triggered me looking into this, memory consumption
caused by boolbv_mapt::get_map_entry was 4.50 GB, and is now 2.25 GB
(which is reasonable, because the bool-typed struct member likely
resulted in padding being generated that would effectively double the
memory consumption).

In addition to memory savings, some of the code can now be simplified as
the entire vector can be copied at once.

### DIFF
--- a/src/solvers/flattening/boolbv_map.cpp
+++ b/src/solvers/flattening/boolbv_map.cpp
@@ -37,10 +37,20 @@ std::string boolbv_mapt::map_entryt::get_value(const propt &prop) const
   return result;
 }
 
-boolbv_mapt::map_entryt &boolbv_mapt::get_map_entry(
-  const irep_idt &identifier,
-  const typet &type)
+void boolbv_mapt::show(std::ostream &out) const
 {
+  for(const auto &pair : mapping)
+    out << pair.first << "=" << pair.second.get_value(prop) << '\n';
+}
+
+void boolbv_mapt::get_literals(
+  const irep_idt &identifier,
+  const typet &type,
+  const std::size_t width,
+  bvt &literals)
+{
+  PRECONDITION(literals.size() == width);
+
   std::pair<mappingt::iterator, bool> result=
     mapping.insert(std::pair<irep_idt, map_entryt>(
       identifier, map_entryt()));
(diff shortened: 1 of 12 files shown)


peterschrammel (approved):

martin-cs (approved):
Seems reasonable.

tautschnig:
Thanks @martin-cs @peterschrammel for your approvals! @kroening I believe you will have authored this code originally and might have extra insight as to why having an `is_set` for each bit was the right thing to do. I'll thus hold off on merging this for a moment to give @kroening a chance to comment as well.

src/solvers/flattening/boolbv_map.cpp
```diff
@@ -20,27 +20,18 @@ Author: Daniel Kroening, kroening@kroening.com
...
-      if(value.is_true())
-        ch='1';
-      else if(value.is_false())
-        ch='0';
-      else
-        ch='?';
-    }
-
-    result=result+ch;
+    result += (value.is_true() ? '1' : (value.is_false() ? '0' : '?'));
```
hannes-steffenhagen-diffblue, line 29: With the way this is written right now I'd expect something like `10?001` to be possible result, but from the commit message it should either be all `?`s or none of them should be right?

src/solvers/flattening/bv_dimacs.cpp
```diff
@@ -49,21 +49,24 @@ bool bv_dimacst::write_dimacs(std::ostream &out)
...
     for(const auto &lit : literal_map)
-      if(!lit.is_set)
-        out << " "
-            << "?";
-      else if(lit.l.is_constant())
-        out << " " << (lit.l.is_true() ? "TRUE" : "FALSE");
+    {
+      out << ' ';
+
+      if(!m.second.is_set)
```
hannes-steffenhagen-diffblue: Same thing here, I guess. With the changes to me it looks like now the case where the whole map is not set is "separate" logically.

hannes-steffenhagen-diffblue (commented):
Don't know enough about this to be confident here, just leave my 2c.

src/solvers/flattening/boolbv_map.cpp
tautschnig: Not quite: the bits not being set would be printed as a string of '*'. The '?' refers to an individual variable not having a value under a particular assignment computed by a solver. That said: it seems the above is dead code, so perhaps the distinction between `is_set` (which perhaps should be "is_unconstrained" (!)) and is-not-set-under-an-assignment isn't all that important.

src/solvers/flattening/bv_dimacs.cpp
tautschnig: I first wanted to keep the changes to a minimum, but ultimately that wasn't possible anyway. I have now clarified the code structure.

tautschnig:
@kroening I have extended this as discussed, which really permitted a number of further simplifications.

PR #5776, opened 2021-01-21
[HOTFIX] Rename version tag during release.

NlightNFotis:
Because of a recent refactoring during to our
regular release process script, we misnamed
the release tag, causing a number of automated
systems that were dependent on a specific form
of a release tag (e.g `cbmc-<version>`) to fall
over.

### DIFF
--- a/.github/workflows/regular-release.yaml
+++ b/.github/workflows/regular-release.yaml
@@ -33,12 +33,12 @@ jobs:
         uses: actions/create-release@v1
         env:
           GITHUB_TOKEN: ${{ secrets.DB_CI_CPROVER_ACCESS_TOKEN }}
-          CBMC_TAG: ${{ needs.get-version-information.outputs.version }}
+          CBMC_VERSION: ${{ needs.get-version-information.outputs.version }}
         with:
-          tag_name: ${{ env.CBMC_TAG }}
-          release_name: ${{ env.CBMC_TAG }}
+          tag_name: cbmc-${{ env.CBMC_VERSION }}
+          release_name: cbmc-${{ env.CBMC_VERSION }}
           body: |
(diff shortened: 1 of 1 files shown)


.github/workflows/regular-release.yaml
```diff
@@ -35,8 +35,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.DB_CI_CPROVER_ACCESS_TOKEN }}
           CBMC_TAG: ${{ needs.get-version-information.outputs.version }}
         with:
-          tag_name: ${{ env.CBMC_TAG }}
-          release_name: ${{ env.CBMC_TAG }}
+          tag_name: cbmc-${{ env.CBMC_TAG }}
```
hannes-steffenhagen-diffblue: ❓ why not just have two outputs, one with the version number and one with the full tag name?

.github/workflows/regular-release.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
             set up correctly, and then issue:
 
             ```sh
-            $ docker run -it diffblue/cbmc:${{ env.CBMC-TAG }}
+            $ docker run -it diffblue/cbmc:${{ env.CBMC_TAG }}
```
hannes-steffenhagen-diffblue: 🚫 This is not the tag, this is the version after your changes

hannes-steffenhagen-diffblue (changes requested):
Also, related: When we do a sort-of experimental realease this should be `cbmc-$VERSION[.extra-stuff]` so the `.extra-stuff` is included in the version. If you push `cbmc-x.x.x` that should be an actual release that people are meant to use, if you push up something that's not meant to be used by end users that should be obvious from the version name.

hannes-steffenhagen-diffblue (approved):

PR #5778, opened 2021-01-22
Remove unnecessary copy and loss of state information

TGWDB:
This removes a constructor that was not necessary
because it implemented the default behaviour of
the copy constructor (and thus created two places
to edit if changes were made to the class).

### DIFF
--- a/src/goto-symex/goto_state.h
+++ b/src/goto-symex/goto_state.h
@@ -20,6 +20,10 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 
 #include "renaming_level.h"
 
+// Forward declaration required since subclass is used explicitly
+// by the parent class.
+class goto_symex_statet;
+
 /// Container for data that varies per program point, e.g. the constant
 /// propagator state, when state needs to branch. This is copied out of
 /// goto_symex_statet at a control-flow fork and then back into it at a
@@ -82,8 +86,6 @@ public:
   goto_statet &operator=(goto_statet &&other) = default;
   goto_statet(const goto_statet &other) = default;
   goto_statet(goto_statet &&other) = default;
-
-  explicit goto_statet(const class goto_symex_statet &s);
 
   explicit goto_statet(guard_managert &guard_manager)
     : guard(true_exprt(), guard_manager), reachable(true)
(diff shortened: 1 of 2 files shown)


src/goto-symex/goto_state.h
```diff
@@ -20,6 +20,9 @@ Author: Romain Brenguier, romain.brenguier@diffblue.com
 
 #include "renaming_level.h"
 
+// 20210122 Forward declaration required
```
martin-cs: We generally don't date comments, in favour of using git blame / praise.  Maybe we should but at the moment we don't

martin-cs: Also, why is it required?  It doesn't look necessary if you are deleting the code that uses it...

tautschnig:
It may also be useful to add some history to the commit message: this constructor became obsolete as of 7938cd19f3d.

src/goto-symex/goto_state.h
TGWDB: Updated the comment to remove the date stamp. Also added more detail to the comment about the declaration of the child class inside the parent class. The parent class explicitly uses the child class, this was resolved before by declaring the constructor for the child class inline. I don't know why the child class needs to be used in the parent, this just removes the duplicate constructor declaration and handles the required (forward) declaration.

tautschnig: Thank you for these updates, but can we please actually dig a little deeper here? It seems the `apply_condition` function is what uses `goto_symex_statet`, but then either 1) `apply_condition` could safely be used with a `goto_statet` instead of a `goto_symex_statet`, or 2) `apply_condition` should become a member function of `goto_symex_statet` instead of being a member of `goto_statet`. Thanks!

TGWDB: The `apply_condition` code in `goto_statet` has special cases related to `goto_symex_statet` that are non-trivial to split from the parent class. To do this properly is a larger refactor that impacts several locations in the code and there are already comments related to this (e.g. see `symex_goto.cpp` lines 35-46).
I'm not comfortable taking on this larger refactor at the moment, specially since this PR is just to remove some redundant code.

tautschnig (approved):
Ok, but we really need to clean up this mess of a parent class restricting itself to a particular form of inheritance.

PR #5782, opened 2021-01-25
Prevent users from misusing CMake

hannes-steffenhagen-diffblue:
When CMake is invoked from the source directory it will overwrite files
in there. In general CMake expects to be used for out-of-source builds.
Previously it was still possible to get this wrong by accident, this
just adds a check to prevent this from happening and tells people what
to do instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,20 @@
 cmake_minimum_required(VERSION 3.2)
 
+# If cmake generates files inside of the cbmc source directory this can lead to important files being overwritten, so we prevent that here
+if("${CMAKE_BINARY_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
+  message(FATAL_ERROR
+    "We have detected that you were trying to invoke cmake with its output"
+    " directory being equal to the source directory.  You may have done this by"
+    " accident by invoking cmake from the cbmc top level directory without"
+    " specifying a build directory.  Since this can lead to cmake overwriting"
+    " files in the source tree, this is prohibited.  Please perform an"
+    " out-of-source build by either creating a separate build directory and"
+    " invoking cmake from there, or specifying a build directory with -B.\n"
+    "(see also: COMPILING.md)\n"
+    "Note that running this command will have created CMakeCache.txt"
+    " and a CMakeFiles directory; You should delete both of them.")
+endif()
+
 # Build a Release version by default (default build flags for each build type
 # are configured below).
 if (NOT EXISTS ${CMAKE_BINARY_DIR}/CMakeCache.txt)


CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
 cmake_minimum_required(VERSION 3.2)
 
+# If cbmc generates files inside of the cbmc source directory this can lead to important files being overwritten, so we prevent that here
+if("${CMAKE_BINARY_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
+  message(FATAL_ERROR
+    "Do not generate CMake files inside of the cbmc source directory."
+    " Please either create a separate build directory and invoke cbmc from there, or specify a build directory with -B."
```
tautschnig: s/cbmc/cmake/ ?

CMakeLists.txt
```diff
@@ -8,6 +16,7 @@ if (NOT EXISTS ${CMAKE_BINARY_DIR}/CMakeCache.txt)
     endif()
 endif()
 
+
```
tautschnig: Necessary change?

CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
 cmake_minimum_required(VERSION 3.2)
 
+# If cbmc generates files inside of the cbmc source directory this can lead to important files being overwritten, so we prevent that here
```
tautschnig: Why does "cbmc" generate anything (at build time!)?

CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
...
+if("${CMAKE_BINARY_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
+  message(FATAL_ERROR
+    "Do not generate CMake files inside of the cbmc source directory."
```
tautschnig: I'm not sure about the wording here: you are talking to the user here, who just invoked `cmake` - they didn't actually (intend to) create any files at all, and might thus be confused by this error message.

tautschnig (approved):
Thank you for adding those changes (and @TGWDB thank you for raising this topic!) The wording could do with some tweaking as indicated by my comments below.

hannes-steffenhagen-diffblue:
@tautschnig I updated the wording

tautschnig:
> @tautschnig I updated the wording

That's great, thank you very much!

CMakeLists.txt
```diff
@@ -1,5 +1,20 @@
...
+# If cmake generates files inside of the cbmc source directory this can lead to important files being overwritten, so we prevent that here
+if("${CMAKE_BINARY_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
+  message(FATAL_ERROR
+    "We have detected that you were trying to invoke cmake with its output"
+    " directory being equal to the source directory.  You may have done this by"
+    " accident by invoking cmake from the cbmc top level directory without"
+    " specifying a build directory.  Since this can lead to cmake overwriting"
+    " files in the source tree, this is prohibited.  Please perform an"
+    " out-of-source build by either creating a separate build directory and"
+    " invoking cmake from there, or specifying a build directory with -B.\n"
```
piotr-grabalski, line 12: Maybe add 'flag' or 'option' after `-B`?
> ..., or specifying a build directory using -B option.

TGWDB:
Looks fine for checking the single top level directory, do we care about mistakes related to subdirectories as well?
`tgw@DB0907:~/github/hannes/cbmc$ cmake . -Bsrc`
will continue fine and only over-write a couple of files.

hannes-steffenhagen-diffblue:
@TGWDB I'm trying to prevent accidental misuse, it seems unlikely to me that someone would do that by accident?

TGWDB:
@hannes-steffenhagen-diffblue Probably not a concern then, there is always a limit to how much safety/checking we can do.

NlightNFotis (approved):
LGTM, only one question, why are there 2 spaces in some lines (for example line 10)?

tautschnig:
> @hannes-steffenhagen-diffblue Probably not a concern then, there is always a limit to how much safety/checking we can do.

But perhaps that's still easy to add by replacing the `STREQUAL` by a prefix check?

TGWDB:
> > @hannes-steffenhagen-diffblue Probably not a concern then, there is always a limit to how much safety/checking we can do.
> 
> But perhaps that's still easy to add by replacing the `STREQUAL` by a prefix check?

@tautschnig Except that this then prevents building in a subdirectory called "build" since the prefix will match.

tautschnig:
> > > @hannes-steffenhagen-diffblue Probably not a concern then, there is always a limit to how much safety/checking we can do.
> > 
> > 
> > But perhaps that's still easy to add by replacing the `STREQUAL` by a prefix check?
> 
> @tautschnig Except that this then prevents building in a subdirectory called "build" since the prefix will match.

Err, of course true. Let's stick with the current approach for now.

PR #5788, opened 2021-01-26
Replace some uses of {F,f}orall_expr by ranged-for

tautschnig:
In some (but not all!) cases of {F,}orall_expr we do not actually need the iterator.

### DIFF
--- a/src/analyses/goto_rw.cpp
+++ b/src/analyses/goto_rw.cpp
@@ -745,9 +745,11 @@ static void goto_rw(
   rw_set.get_objects_rec(
     function, target, rw_range_sett::get_modet::READ, function_call.function());
 
-  forall_expr(it, function_call.arguments())
+  for(const auto &argument : function_call.arguments())
+  {
     rw_set.get_objects_rec(
-      function, target, rw_range_sett::get_modet::READ, *it);
+      function, target, rw_range_sett::get_modet::READ, argument);
+  }
 }
 
 void goto_rw(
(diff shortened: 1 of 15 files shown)


src/ansi-c/expr2c.cpp
```diff
@@ -2358,7 +2358,7 @@ std::string expr2ct::convert_side_effect_expr_function_call(
 
   dest+='(';
 
-  forall_expr(it, src.arguments())
+  for(auto it = src.arguments().begin(); it != src.arguments().end(); ++it)
```
hannes-steffenhagen-diffblue: Think this should probably be a separate refactor, but the only thing the iterator is being used for here is checking whether this is the first iteration, which we normally just do with a flag + for-each. Same is true for at least some of the other `for(auto it = something.begin(); it != something.end(); ++it)` loops. I was thinking it'd be nice to have something like `for(auto const &it : iterators(collection))` or `for(auto const it& : iterators(collection.begin(), collection.end())` for some cases that can't be so easily transformed too.

hannes-steffenhagen-diffblue (approved):

src/cpp/cpp_instantiate_template.cpp
```diff
@@ -289,7 +291,9 @@ const symbolt &cpp_typecheckt::instantiate_template(
 
 #ifdef DEBUG
   std::cout << "A: <";
-  forall_expr(it, specialization_template_args.arguments())
+  for(auto it = specialization_template_args.arguments().begin();
```
martin-cs: At the risk of sounding ungrateful for the amazing clean up PRs... the ones that do still need iterators, I think they are best still with the macro.

martin-cs (approved):

tautschnig:
@hannes-steffenhagen-diffblue @martin-cs Thanks a lot for the feedback. I have modified to commit such as to leave in place the macro where the iterator was required. The refactoring suggested by @hannes-steffenhagen-diffblue probably should be done as well, but seems right for a separate PR.

kroening (approved):

PR #5791, opened 2021-01-29
SMT2 uninterpreted functions

kroening:

### DIFF
--- a/src/solvers/flattening/boolbv.cpp
+++ b/src/solvers/flattening/boolbv.cpp
@@ -565,6 +565,11 @@ literalt boolbvt::convert_rest(const exprt &expr)
     else if(op.type().id() == ID_fixedbv)
       return const_literal(true);
   }
+  else if(expr.id() == ID_function_application)
+  {
+    functions.record(to_function_application_expr(expr));
+    return prop.new_variable();
+  }
 
   return SUB::convert_rest(expr);
 }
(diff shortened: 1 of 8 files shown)


src/solvers/flattening/boolbv.cpp
```diff
@@ -565,6 +565,11 @@ literalt boolbvt::convert_rest(const exprt &expr)
     else if(op.type().id() == ID_fixedbv)
       return const_literal(true);
   }
+  else if(expr.id() == ID_function_application)
+  {
+    functions.record(to_function_application_expr(expr));
+    return prop.new_variable();
+  }
```
tautschnig, line 572: Does this also have implications outside SMT2, i.e., were uninterpreted functions previously broken in some way?

tautschnig (approved):
Looks ok to me, but I can't claim particular knowledge with the SMT2 back-end or solver.

src/solvers/flattening/boolbv.cpp
martin-cs: Yeah; I agree with @tautschnig this seems like it should have a visible / testable effect on the usual back-end.

martin-cs (approved):
Approve modulo @tautschnig 's concern.

src/solvers/flattening/boolbv.cpp
kroening: > Does this also have implications outside SMT2, i.e., were uninterpreted functions previously broken in some way?

I've added a further test that uses the C front-end.

PR #5472, opened 2020-08-24
Variable sensitivity domain

martin-cs:
This PR merges in several years of work on a new domain for the abstract interpreter.  It should solve a number of the long-standing problems with soundness and precision.  The key feature of the domain is that the abstraction used for each type can be selected at run-time, allowing a wide range of scalability / precision trade-offs and easy specialisation without having to re-write all of the array, field, pointer sensitivity, etc.

1. This is the work of a lot of people, notably @thk123 @chrisr-diffblue @hannes-steffenhagen-diffblue @danpoe @NlightNFotis and others.  I have tried to preserve authorship where-ever possible.

2. I don't think it is reasonable to try to review this in the normal way, nor do I think it is viable to split it up further.  Whether it gets merged or not has to be a policy question really.  Over the years this has been going on I have asked @peterschrammel and @tautschnig about this as an approach and they seemed supportive because...

3. The PR should have very little impact on existing functionality, it is enabled with an option in goto-analyzer.  It has also been deployed in at least two industrial applications.

4. There is still work to be done in tidying up the code but I think it can be and would be best done as PRs to it after merging.  I will try to do some during the discussion / review.

5. This includes https://github.com/diffblue/cbmc/pull/5202 and https://github.com/diffblue/cbmc/pull/5204 which were previous attempts at splitting this out a bit.

6. I don't know who is currently looking after
https://github.com/diffblue/cbmc/tree/variable-sensitivity-domain
and
https://github.com/diffblue/cbmc/tree/variable-sensitivity-with-get-function-pointers
( @chrisr-diffblue ? @danpoe ? @hannes-steffenhagen-diffblue ? )
but this PR includes the majority of the content of these.  I have a branch of the remainder and am happy to work with people to rebase that on top of this PR.

Let me know what you think!

Issue #5307, opened 2020-04-20
Variable Sensitivity Value Set implementation should be split into separate objects for "normal" values and pointers


hannes-steffenhagen-diffblue:
Note: This issue references functionality that's currently not merged into develop.

This is how other subdomains of VSD work (see e.g. `constant_abstract_objectt` vs `constant_pointer_abstract_objectt`). As of now this is causing problems with merging, because the set management uses pointer comparisons to determine if two objects are equal (e.g. if you have two value sets each consisting only of pointer-to-three, you'll end up with a value set with two different pointer-to-three's in it).

What should be done instead is that `value_set_abstract_objectt` should only manage `exprt`s, and for pointers we either use `constant_pointer_abstract_objectt` or add a `set_pointer_abstract_objectt` that manages a set of possible pointees to complement the value set. Possibly similar considerations for arrays.

martin-cs:
https://github.com/diffblue/cbmc/pull/5793 touches on some of this functionality.

PR #5793, opened 2021-01-30
Vsd value set implementation

jezhiggins:
The focus of this PR is filling out value-set behaviour. Up to now you could assign to a value-set, but any kind of operation would result in the value-set going to top. This PR fills out value-set behaviour for unary & n-ary operations, and the ternary operator too because that's a bit of an edge case. 

This PR also adds array indexing operations with value-sets, and makes value-sets work with context tracking. Large value-sets are correctly converted to intervals, previously they just went top.

Finally, the value-set output is now sorted. value-set intervan ordering is unstable in the face of code changes, which made tests fragile. Sorting output fixes that and is also easier on the human eye.

In addition, there's are some refactoring work in interval_value_objectt, mainly to clarify the implementation so I could draw parallels with value-set. 

This PR looks larger that it really is. Of the 70 files listed, 55 of them are part of the regression tests and all but a handful are new tests covering value-set behaviour. The rest are renamed to better express what they test, or have minor fixes now value-set behaviour has been filled out. The substantial changes are all in [value_set_abstract_object.cpp](https://github.com/diffblue/cbmc/pull/5793/files#diff-3937ebb4cbcb3ee7ecf9b820b6e2ae3ce7c64253dff0c1cf60e061a920feb7ab)


regression/goto-analyzer/ternary-operator-intervals/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--variable-sensitivity --vsd-values intervals --show
```
martin-cs: Something that works but is not widely documented is having multiple `.desc` files in the same directory.  Thus rather than having a bunch of different directories for the ternary code, you can have `intervals.desc`, `constants.desc` etc.

regression/goto-analyzer/ternary-operator-constants/main.c
```diff
@@ -0,0 +1,19 @@
...
+  bool b2;
+
+  b1 = true;
+  b2 = !b1;
+
+  bool b3 = x ? true : false;
+  int i = b1 ? 10 : 20;
+  int j = b2 ? 10 : 20;
+  int k = b3 ? 10 : 20;
+
```
martin-cs: If the test is demonstrating what values are represented by the abstract domain then I have a mild preference for having `assert`s in the test cases, ideally passing and failing.  I tend to use `--show` in tests for when I want to actually check the displayed output.  This is not wrong and you don't need to change it, more of a point of style really.

regression/goto-analyzer/ternary-operator-intervals/test.desc
```diff
@@ -0,0 +1,12 @@
...
+^EXIT=0$
+^SIGNAL=0$
+main::1::b1 \(\) -> \[1, 1\] @ \[2\]
+main::1::b2 \(\) -> \[0, 0\] @ \[3\]
+main::1::b3 \(\) -> \[0, 1\] @ \[5\]
+main::1::i \(\) -> \[A, A\] @ \[7\]
```
martin-cs: I realise this is probably not your doing but if the output is in hex, it should have a leading `0x`.

regression/goto-analyzer/value-set-addition-05/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+--variable-sensitivity --vsd-values set-of-constants --show
+^EXIT=0$
+^SIGNAL=0$
+main::1::p .* value-set-begin: 2, 3 :value-set-end
+main::1::q .* value-set-begin: 5, 10 :value-set-end
+main::1::t .* value-set-begin: 7, 8, 12, 13 :value-set-end
```
martin-cs, line 8: This line is especially interesting as it shows that the abstraction is not relational; the link between `p` and `q` is not maintained after the merge at the end of the `if`, so you wind up with false alarms like `12`.

regression/goto-analyzer/value-set-function-pointers-incremented-02/main.c
```diff
@@ -0,0 +1,34 @@
...
+{
+  int i;
+  // This line is needed so that g is considered as a possibility for the TOP
+  // value
+  fptr_t dummy = g;
+
+  // function pointer incremented should be top
+  fptr_t fun_incremented = f;
+  if(i)
+    ++fun_incremented;
```
martin-cs, line 29: Ummm... is this incrementing a function pointer?  I'm not sure that is going to work...

regression/goto-analyzer/value-set-function-pointers-incremented-02/main.c
```diff
@@ -0,0 +1,34 @@
+#include <assert.h>
+
+typedef int (*fptr_t)(int);
```
martin-cs, line 3: Note that, as things stand, function pointers are removed before analysis.

regression/goto-analyzer/value-set-function-pointers-simple/test.desc
```diff
@@ -6,20 +6,20 @@ main.c
...
-^main::1::fun3_show \(\) -> \{\[0\] = value-set-begin: .*ptr ->\(h\).* :value-set-end$
-^fun_global_show \(\) -> value-set-begin: .*ptr ->\(h\).* :value-set-end$
+^main::1::fun0_show \(\) -> value-set-begin: .* :value-set-end
+^main::1::fun1_show \(\) -> value-set-begin: .*ptr ->\(h\).* :value-set-end
+^main::1::fun2_show \(\) -> value-set-begin: .*ptr ->\(h\).* :value-set-end
+^main::1::fun3_show \(\) -> \{\[0\] = value-set-begin: .*ptr ->\(h\).* :value-set-end
+^fun_global_show \(\) -> value-set-begin: .*ptr ->\(h\).* :value-set-end
 --
 These TOP values in the sets shouldn't exist.
 They're caused by a quirk in the implementation, see
```
martin-cs: The next line references:

https://github.com/diffblue/cbmc/issues/5307 

if most of the erronous `TOP` values are gone then maybe the comment should go too?  Also possibly the issue?

src/analyses/variable-sensitivity/constant_abstract_value.cpp
```diff
@@ -33,18 +33,21 @@ index_range_ptrt make_constant_index_range(const exprt &val)
   return std::make_shared<constant_index_ranget>(val);
 }
 
-constant_abstract_valuet::constant_abstract_valuet(typet t)
+constant_abstract_valuet::constant_abstract_valuet(const typet &t)
```
martin-cs, line 36: Good catch!

src/analyses/variable-sensitivity/abstract_environment.h
```diff
@@ -170,6 +170,19 @@ class abstract_environmentt
...
+  /// \param abstract_object: The object to be wrapped
+  ///
+  /// \return The wrapped abstract object
+  ///
+  /// Look at the configuration context dependency, and constructs
+  /// the appropriate wrapper object around the supplied object
+  /// If no such configuration is enabled, the supplied object will be
+  /// returned unchanged
+  virtual abstract_object_pointert
+  add_object_context(const abstract_object_pointert &abstract_object) const;
```
martin-cs, line 184: I have a feeling I know what this does but it is not 100% clear to me how it fits in in the work-flow.

src/analyses/variable-sensitivity/interval_abstract_value.h
```diff
@@ -27,15 +27,14 @@ class interval_abstract_valuet : public abstract_value_objectt
...
-  explicit interval_abstract_valuet(typet t);
-  interval_abstract_valuet(typet t, bool tp, bool bttm);
+  explicit interval_abstract_valuet(const typet &t);
+  interval_abstract_valuet(const typet &t, bool tp, bool bttm);
 
-  interval_abstract_valuet(const constant_interval_exprt e, int merge_count);
-
-  explicit interval_abstract_valuet(const constant_interval_exprt e);
+  explicit interval_abstract_valuet(const constant_interval_exprt &e);
+  interval_abstract_valuet(const constant_interval_exprt &e, int merge_count);
```
martin-cs: What does `merge_count` do?

src/analyses/variable-sensitivity/variable_sensitivity_configuration.cpp
```diff
@@ -41,9 +41,7 @@ vsd_configt vsd_configt::from_options(const optionst &options)
 
   // This should always be on (for efficeny with 3-way merge)
   // Does not work with value set
```
martin-cs: If it works, remove the comment!

src/analyses/variable-sensitivity/variable_sensitivity_configuration.cpp
```diff
@@ -41,9 +41,7 @@ vsd_configt vsd_configt::from_options(const optionst &options)
 
   // This should always be on (for efficeny with 3-way merge)
```
martin-cs: I'm ... not 100% sure this is actually true, perhaps we should re-visit this bit of the config.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -46,12 +47,83 @@ class value_set_index_ranget : public index_ranget
...
+/// \param other: the abstract object to get the type of
+/// \return the abstract-type of \p other
+value_set_abstract_objectt::abstract_typet
+get_type(const abstract_object_pointert &other);
+
+/// Determine abstract-type of an expression-type \p type.
+/// \param type: the expression type to get the abstract-type of
+/// \return the abstract-type of \p type
+value_set_abstract_objectt::abstract_typet
+type_to_abstract_type(const typet &type);
```
martin-cs, line 66: Am I correct in thinking that this is what picks which `abstract_objectt` subtype is used for the abstractions inside value set?

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -130,29 +202,56 @@ abstract_object_pointert value_set_abstract_objectt::expression_transform(
...
+    return evaluate_conditional(
+      expr.type(), collective_operands, environment, ns);
 
   abstract_object_sett resulting_objects;
+
+  auto dispatcher = *values.begin();
   for_each_comb(
     collective_operands,
-    [&resulting_objects, this, &expr, &environment, &ns](
+    [&resulting_objects, &dispatcher, &expr, &environment, &ns](
```
martin-cs, line 219: Given that it will either be unary or binary, this feels like over-kill.  Is there any reason not to replace this with:

```
if (expr.children() == 1) {
  for (...) { recursively eval }
} else if (expr.children() == 2) {
  for (...) {
    for (...) {
      recursively eval;
    }
  }
```

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -169,23 +268,26 @@ abstract_object_pointert value_set_abstract_objectt::write(
...
   }
-  return resolve_new_values(new_values);
+  return resolve_new_values(new_values, environment);
 }
 
 abstract_object_pointert value_set_abstract_objectt::resolve_new_values(
+  const abstract_object_sett &new_values,
+  const abstract_environmentt &environment) const
+{
+  auto result = resolve_values(new_values);
```
martin-cs, line 279: I think we probably need to have a chat about this function.

martin-cs (changes requested):
Thanks for the amazing test cases and many fixes.  I would say this is 95% of the way there.  Let's have a chat about the remaining 5%, I don't think it will be hard to fix up.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
martin-cs: Hmmm... perhaps not, `a + b + c + ... + z` would not be handled like this.

regression/goto-analyzer/ternary-operator-constants/main.c
jezhiggins: I wasn't sure how to write an assert that would show that the value-set did indeed contain all the values I expected it to.

regression/goto-analyzer/ternary-operator-intervals/test.desc
jezhiggins: It did take me a little while to realise that the interval output was hex. I would be delighted to change everything to decimal or everything to hex :)

regression/goto-analyzer/value-set-function-pointers-incremented-02/main.c
jezhiggins: It is and it's not supposed to. I took this from an existing test to ensure the behaviour was consistent with it.

src/analyses/variable-sensitivity/constant_abstract_value.cpp
jezhiggins: The IDE told me :)

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
jezhiggins: There can be an arbitrary number of operands, yes, and I've added tests explicitly to exercise for cases where there are more than 2.

src/analyses/variable-sensitivity/variable_sensitivity_configuration.cpp
jezhiggins: Turns out there's another chunk that should also go 
```
  if(
    options.get_bool_option("value-set") &&
    options.get_bool_option("data-dependencies"))
  {
      "Value set is not currently supported with data dependency analysis",
      "--value-set --data-dependencies",
      "--data-dependencies"};
  }
```

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
jezhiggins: Not exactly - it affects some runtime behaviour based on whether the value-set is holding pointers or values. This is related to the issue you highlighted above.

src/analyses/variable-sensitivity/interval_abstract_value.h
jezhiggins: It's used to drive a repeated merged interval to TOP, but for nothing else. It's not clear to me why we would do this, though. 
```
interval_abstract_valuet::interval_abstract_valuet(
  const constant_interval_exprt &e,
  int merge_count)
  : abstract_value_objectt(
      e.type(),
      e.is_top() || merge_count > 10,
      e.is_bottom()),
    interval(e),
    merge_count(merge_count)
{
}
```

regression/goto-analyzer/value-set-function-pointers-simple/test.desc
jezhiggins: I'm looking now at splitting value-sets of constants from value-sets-of-pointers at the moment. I'm not sure it will necessarily solve everything, but it should at least help separate concerns and highlight where the holes are.

src/analyses/variable-sensitivity/interval_abstract_value.h
hannes-steffenhagen-diffblue: @jezhiggins IIRC this was for loop bodies, e.g. consider

```
for(int i = 0; i < 2; ++i)
{
  // do something
}
```

We start out with `i in {0}`, then we get `i in {0,1}`, `i in {0,1,2}` etc, because we keep merging with the loop body rather than inlining the loop like we do in cbmc, so we never reach a natural termination condition. There's probably a better way to do this than this hack though.

regression/goto-analyzer/ternary-operator-constants/main.c
martin-cs: It's a bit of a faff so I wouldn't require it but the following should give you the exact values in the value set via the assertion checking results:

```
if (k == value1) { assert(1); }
else if (k == value2) { assert(1); }
...
else (k == valueN) { assert(1); }
else { assert(0); }
```

If you have enumerated all of the values then each of the `assert(1)` should be marked as a success.  If it is unreachable then the value has been omitted.  Correspondingly, if `assert(0)` is unreachable then there are no extra values, if it fails then there are additional, possibly unexpected, values.

Ofcourse, having written that, I am not 100% sure that `abstract_environmentt::assume()` will work well enough with `value_sett` for this to work.  At some point it should but that is not necessary for this PR.

regression/goto-analyzer/ternary-operator-intervals/test.desc
martin-cs: All decimal or all hex (or even, all both) are valid design choices and arguable.  The current mix and lack of `0x` is definitely wrong.  I would be very happy to have PRs that fix this.  Again, out of scope for this PR.

regression/goto-analyzer/value-set-function-pointers-incremented-02/main.c
martin-cs: OK.

regression/goto-analyzer/value-set-function-pointers-simple/test.desc
martin-cs: OK.  I am open to both split and non-split solutions, also I'm fine with it being in a subsequent PR.  However the comment about TOP should probably leave with the commit that removes them.

regression/goto-analyzer/ternary-operator-constants/main.c
jezhiggins: Thanks. Noted for the future.

src/analyses/variable-sensitivity/interval_abstract_value.h
martin-cs: Right!  It is a way of implementing widening!  Ofcourse!

@hannes-steffenhagen-diffblue thanks for the input.  I believe there is a better way and it is on the tasks I've asked @jezhiggins to look at.  Briefly... `ai_historyt` and it's children have a `bool should_widen() const` method which allows CFG aware decisions on when to widen.  So we can widen on the 10th iteration or on recursive function calls.  Domains that have non-trivial height / want to do fancy widening things can then query this in their `merge` method and take appropriate action.  VSD needs to do this and then pass a "widen" parameter down to the `abstract_objectt::merge` method.

Note that this PR and other changes to `value_sett` are working on adding a size limit that is independent of widening so that the domain will be space limited not just at merge but also at transform.

TL;DR : this can go; there are better ways of doing this.

martin-cs:
@jezhiggins Thanks for the changes.  Let's have a call at some point to sort out the last bits.

src/analyses/variable-sensitivity/interval_abstract_value.h
jezhiggins: It's gone!

regression/goto-analyzer/value-set-function-pointers-simple/test.desc
jezhiggins: Fixed

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.cpp
```diff
@@ -29,12 +29,10 @@ abstract_object_pointert create_abstract_object(
   const namespacet &ns)
 {
   if(top || bottom)
-    return abstract_object_pointert(
-      new abstract_object_classt { type, top, bottom });
+    return std::make_shared<abstract_object_classt>(type, top, bottom);
```
thomasspriggs, line 30: :+1: Nice clean up.
:pick: Could you add a commit message body justifying the rational for the change? E.G. - "Because it is best practise to create shared pointers using `make_shared`. See C++ Core Guidelines points C.151 and R.22".

regression/goto-analyzer/value-set-addition-08/test.desc
```diff
@@ -6,5 +6,5 @@ main.c
 main::1::p .* value-set-begin: 3, 2, 4, :value-set-end
 main::1::q .* value-set-begin: 5, 10, :value-set-end
 main::1::r .* value-set-begin: 20, :value-set-end
-main::1::t .* value-set-begin: 28, 27, 32, 33, 29, 34, :value-set-end
+main::1::t .* value-set-begin: 28, 27, 34, 32, 33, 29, :value-set-end
```
thomasspriggs: Why does using `make_shared` change the ordering in this output?

regression/goto-analyzer/value-set-addition-01/main.c
```diff
@@ -1,4 +1,4 @@
-int main(int argc, char argv[]) 
+int main(int argc, char argv[])
```
thomasspriggs, line 1: Formatting fixes to code which has been added as part of this PR should ideally be merged into the commit where it is added. Only fixes to code which isn't changed/added by your PR really needs to be in separate commits. This fixing of formatting in these commits can be automated using - `git filter-branch --tree-filter 'git-clang-format develop' -- develop..HEAD`. Assuming that your PR is already based on `develop`.

src/analyses/variable-sensitivity/interval_abstract_value.cpp
```diff
@@ -321,23 +320,43 @@ abstract_object_pointert interval_abstract_valuet::expression_transform(
   if(num_operands == 0)
     return environment.abstract_object_factory(type, ns, true);
 
-  if(expr.id() == ID_plus)
+  if(expr.id() == ID_if)
```
thomasspriggs, line 336: :information_source: There is a set of function templates to provide an abstraction over checking/casting to different types of expression. For example this check could be written as `if(can_cast_expr<if_exprt>(expr))`. You can also cast complete with validity checks by using `expr_try_dynamic_cast<if_exprt>(expr)`.

regression/goto-analyzer/value-set-addition-02/main.c
```diff
@@ -0,0 +1,14 @@
+int unknown();
```
thomasspriggs: :pick: Could you make the commit title more descriptive? Something like `Add more value-set behaviour tests`. It wasn't clear to me from reading the commit title only that this only added more tests rather than more functionality.

src/analyses/variable-sensitivity/interval_abstract_value.cpp
```diff
@@ -365,59 +383,21 @@ abstract_object_pointert interval_abstract_valuet::expression_transform(
...
   }
-  else if(num_operands == 2)
+  else
   {
-    const constant_interval_exprt &interval0 = interval_operands[0]->interval;
-    const constant_interval_exprt &interval1 = interval_operands[1]->interval;
+    constant_interval_exprt result = interval_operands[0]->interval;
 
-    constant_interval_exprt interval = interval0.eval(expr.id(), interval1);
+    for (size_t opIndex = 1; opIndex != interval_operands.size(); ++opIndex)
```
thomasspriggs: :bulb: This loop could be done using a ranged for loop -
```
  for(const auto &interval_operand : make_range(interval_operands).drop(1))
  {
    result = result.eval(expr.id(), interval_operand->get_interval());
  }
```
This would require including `util/range.h`.

src/analyses/variable-sensitivity/value_set_abstract_object.h
```diff
@@ -67,6 +67,7 @@ class value_set_abstract_objectt : public abstract_value_objectt
   void set_values(const abstract_object_sett &other_values)
   {
     PRECONDITION(!other_values.empty());
+    set_not_top();
```
thomasspriggs: :no_entry_sign: This commit does not pass the regression tests. All commits should compile and pass all tests.

regression/goto-analyzer/value-set-function-pointers-incremented-01/test.desc
```diff
@@ -2,9 +2,8 @@ CORE
 main.c
 --variable-sensitivity --vsd-values set-of-constants --vsd-structs every-field --vsd-arrays every-element --vsd-pointers value-set --show --pointer-check
 ^file main.c line 28 function main: replacing function pointer by 2 possible targets$
-^main::1::fun_incremented_show \(\) -> TOP$
+^main::1::fun_incremented_show \(\) -> value-set-begin: TOP, :value-set-end$
```
thomasspriggs: :no_entry_sign: This commit appears to fix problems with the previous commit. It should be merged into the previous commit in order to make future use of git bisect easier when tracking down other issues. If the tests in the previous commit are testing previously correct functionality and you want to keep a record of this previous functionality, then it might be better to separate these tests into a separate preceding commit.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -148,8 +148,12 @@ abstract_object_pointert value_set_abstract_objectt::expression_transform(
     collective_operands,
     [&resulting_objects, this, &expr, &environment, &ns](
       const std::vector<abstract_object_pointert> &ops) {
+      auto operands_expr = exprt::operandst { };
```
thomasspriggs: :pick: Seems like a slightly long winded way of writing this declaration. Why not just `exprt::operandst operands_expr;`?

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -148,8 +148,12 @@ abstract_object_pointert value_set_abstract_objectt::expression_transform(
...
+      for (auto v : ops)
```
thomasspriggs: :pick: Can we use `for(const auto &v : ops)` in order to make it clear that `v` is not mutated and to remove the copying?

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -148,8 +148,12 @@ abstract_object_pointert value_set_abstract_objectt::expression_transform(
...
+        operands_expr.push_back(v->to_constant());
+      auto rewritten_expr = exprt(expr.id(), expr.type(), std::move(operands_expr));
```
thomasspriggs: :pick: I suggest using `const exprt rewritten_expr{expr.id(), expr.type(), std::move(operands_expr)};`, in order to avoid the unnecessary use of `auto`.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -392,12 +401,24 @@ void value_set_abstract_objectt::output(
...
+    for(const auto &value : values)
+    {
+      std::ostringstream ss;
+      value->output(ss, ai, ns);
+      output_values.emplace_back(ss.str());
+    }
+    std::sort(output_values.begin(), output_values.end(), by_length);
+
+    bool comma = false;
+    for(auto const &value : output_values)
```
thomasspriggs: :pick: There are existing unit tested functions for joining a collection of strings together with a delimiter and outputting the result to a stream. These are called `join_strings` and they can be found in `src/util/string_utils.h`. It is preferable to reuse the existing functions so that we don't have duplicate versions of the same functionality to maintain and because the existing functions are unit tested.

regression/goto-analyzer/ternary-operator-value-sets/test.desc
```diff
@@ -3,10 +3,10 @@ main.c
 --variable-sensitivity --vsd-values set-of-constants --show
 ^EXIT=0$
 ^SIGNAL=0$
-main::1::b1 \(\) -> value-set-begin: TRUE, :value-set-end
```
thomasspriggs: :+1: Good to see the formatting being improved and the tests being made more stable.

thomasspriggs (commented):
This is a partial review. I had a number of comments I wanted to feed back, before continuing to review this PR in its in entirety. It is good to see more functionality added and problems fixed though. So thank-you for your contribution.

I have one general comment - I have found this PR relatively hard going to review. Part of that is that I am currently learning about this area of the code base. But some of this has to do with the way this PR has been organised. So here are some suggestions which could make this easier to review.
1. Firstly even if a lot of the lines changed are in the tests rather than the code, those tests still need to be eyeballed to check them over. So I think this could have been more digestible for reviewers if this was 2-4 smaller PRs rather than a single large PR. This can also benefit you as a PR author, because it gives you the opportunity to learn from the first PR and anticipate/fix issues with the follow up PRs before you submit them.
2. All commits should have a title which states what is being changed and a commit message body which explains and justifies why that change is being made. If it is fairly straight forward then the why can be a single sentence. The why of a commit may be obvious to you as an author, but it can take time as a reviewer or as anyone looking back at the commit history to look at the changes to infer the why; if it hasn't been included. The PR template links to https://chris.beams.io/posts/git-commit/ which is a great guide to good commit messages if you haven't read it.
3. Every commit should both compile and pass all tests when it is checked out. When regressions find their way into the code base then we would want to be able to use `git-bisect` to go back through the commit history to find the commit which introduced the regression, so that we can fix it more easily. Commits which don't pass the tests can add noise to this process and make it very difficult to track down the cause of a regression. This is the case even if the regression is entirely unrelated to the work in your PR. Commits which contain problems which are fixed later on in a PR also add cognitive load to the reviewer. When reviewing a PR in commit order a reviewer will likely comment on a problem when they spot it. Then they will need to make the connection to the later commit where the problem is fixed in order to better understand the PR as a whole. They may find that there initial comment was somewhat redundant. If instead, the commit which fixes a problem is squashed into the commit which adds it then the reviewer doesn't need to comment on that and they don't need to mentally connect the two commits together.

I have labelled some of my review comments with emojis. This is the key to what I mean by them -
 * :information_source: An FYI. Might be useful if you didn't already know about that piece of information.
 * :no_entry_sign: A comment for where I do not intend to approve the PR, until this has been addressed.
 * :pick: A nit pick. It would be good to get these addressed, if possible. However if you don't have time to address these or you disagree with them, then I am OK with getting this PR merged without any corresponding changes.
 * :bulb: Just an idea. Implement if you like the idea and you have the time.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
jezhiggins: Change made.

regression/goto-analyzer/value-set-addition-08/test.desc
jezhiggins: The value set's hashing function is ultimately resolves down to `std::hash<abstract_object_pointert>{}(shared_from_this());` so presumably it's related to changes in memory location. It's would assume it's possible that the order might change from invocation to invocation, but I can't recall if I saw that. Anyway, that instability is what prompted me to order the output.

thomasspriggs: :+1: Makes sense. Ordering the commits so the output order was fixed first would have been super nice. But I am ok with the current order + the explanation.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
jezhiggins: Everything is else in that little lambda is declared `auto`. For me, it's easier to read if everything is `auto` or everything isn't `auto`. Obviously there are times when that's not possible, because the type deduction doesn't do what we want, but within a scope I generally try to go with one or the other.

src/analyses/variable-sensitivity/interval_abstract_value.cpp
jezhiggins: Thanks - these aren't something I've encountered in the code I've worked in. I'll bear them in mind.

src/analyses/variable-sensitivity/value_set_abstract_object.h
jezhiggins: The focus of this work was filling out new behaviour in value-sets. In this commit, I've changed existing behaviour, which is something I'm cautious about doing (hence also why I didn't change the output order until later on). Consequently, I chose to change the behaviour in one commit, and the update the test in separate one.

martin-cs:
@thomasspriggs Thanks for reviewing.  Please forgive me if I am telling you things you already know but some of the social context for this PR:

- The variable sensitivity domain (VSD) has quite a long and a complex history with many different developers involved.  Along that path the design and design intent was not always communicated as well as it could have been.  As a consequence there are some parts that need some re-architecting.  Jez and I are working on doing that and this is one step towards it.
- After several years being developed out of tree, the bulk of VSD was merged in https://github.com/diffblue/cbmc/pull/5472 following a substantial team effort.  One thing we discussed informally that I don't know if it made to the ticket was that it was better to merge it partially sorted and fix-up on tree rather than continue to track develop until it was perfect.
- As a consequence of these two, when I have been reviewing I have not been as strict about following all of the rules for PRs; as long as it passes all of CI and is a net improvement then it could be mergable.  There is still a lot of tidy up to do and some of it is overlapping so I am accepting that not everything will get done on the first pass.
- As far as I know, my team / project are the only users of this code.  I know at one point Diffblue had another user of this but despite my best attempts I haven't been able to find out who is currently responsible for that.  If you know I would *love* to be able to co-ordinate.
- If you have time / energy / use cases for this or any of the rest of goto-analyzer then it is most appreciated and I'm happy to include you / anyone else in The Plan.

chrisr-diffblue:
@martin-cs To my knowledge, the only other user of VSD that we might have in common is no longer using it. We are no longer actively involved in working with VSD at the moment, other than as part of a general role in looking after the health of the CBMC project.

martin-cs:
@chrisr-diffblue Thanks for the update.

martin-cs (approved):
@jezhiggins and I have had a long discussion about this PR.  I am happy for it to be merged as he says he will address the following issues in future PRs:

1. Making the limit on the number of elements configurable via the factory.
2. Consolidating the "what to do when we reach the maximum number of elements" code from merge and eval and making sure that this works for all contained types (this includes not automatically converting to intervals).
3. Try to avoiding wrapping and unwrapping context and last-modified-information.
4. Looking at how value sets of compound objects (arrays, structs, etc.) and intervals work.

I'll raises these internally on one of our issues.

@thomasspriggs are you happy for this to be merged on this basis (along with the fixes that @jezhiggins has already made)?

thomasspriggs:
> @jezhiggins and I have had a long discussion about this PR. I am happy for it to be merged as he says he will address the following issues in future PRs:
> 
> 1. Making the limit on the number of elements configurable via the factory.
> 2. Consolidating the "what to do when we reach the maximum number of elements" code from merge and eval and making sure that this works for all contained types (this includes not automatically converting to intervals).
> 3. Try to avoiding wrapping and unwrapping context and last-modified-information.
> 4. Looking at how value sets of compound objects (arrays, structs, etc.) and intervals work.
> 
> I'll raises these internally on one of our issues.
> 
> @thomasspriggs are you happy for this to be merged on this basis (along with the fixes that @jezhiggins has already made)?

I'd really like the issues with commits that don't pass tests to be resolved first. Some of the non-compiling commits in the previous work which was merged in that huge PR have already caused us problems when bisecting the history. I'm OK with the non-compiling commits in that PR of old work, because its better than leaving to rot forever. However I think we should be trying to do better for new work. Especially when all that is required is squashing a fix commit in to the correct preceding commit. I'm perfectly fine with the rest being at the discretion of those who know this part of the code base best.

thomasspriggs:
> @thomasspriggs Thanks for reviewing. Please forgive me if I am telling you things you already know but some of the social context for this PR:
> 
> * The variable sensitivity domain (VSD) has quite a long and a complex history with many different developers involved.  Along that path the design and design intent was not always communicated as well as it could have been.  As a consequence there are some parts that need some re-architecting.  Jez and I are working on doing that and this is one step towards it.
> * After several years being developed out of tree, the bulk of VSD was merged in #5472 following a substantial team effort.  One thing we discussed informally that I don't know if it made to the ticket was that it was better to merge it partially sorted and fix-up on tree rather than continue to track develop until it was perfect.
> * As a consequence of these two, when I have been reviewing I have not been as strict about following all of the rules for PRs; as long as it passes all of CI and is a net improvement then it could be mergable.  There is still a lot of tidy up to do and some of it is overlapping so I am accepting that not everything will get done on the first pass.
> * As far as I know, my team / project are the only users of this code.  I know at one point Diffblue had another user of this but despite my best attempts I haven't been able to find out who is currently responsible for that.  If you know I would _love_ to be able to co-ordinate.
> * If you have time / energy / use cases for this or any of the rest of goto-analyzer then it is most appreciated and I'm happy to include you / anyone else in The Plan.

Thank you for the wider context. I am interested in broadening my knowledge of the code base, as there are a limited number of parts that I know well at the moment. I was aware of the large PR of historical work being merged, even if I don't really know its complete history. It is good to see more work being done to bring it up to a higher standard.

I heard that the value set work could potentially be useful for improving the `remove_function_pointers` pass in cbmc, because it could be used to reduce the number of different cases added to the switch blocks. However to my knowledge this isn't a piece of work which we are currently planning to undertake.

martin-cs:
@thomasspriggs At the moment the only in-tree use of VSD is in `goto-analyzer`.  However CBMC does pointer analysis and I have had a long-term goal of unifying this code with the `ait` abstract intepreter.  Doing that would mean we could use VSD for pointer analysis and that would improve a number of performance issues with pointers.

Function pointers are a separate but related question.  At the moment function pointer removal is a very syntactic process.  For a number of years I have wanted (and tried) to make it more semantic.  Doing so would mean we could use VSD for function pointer removal and then, yes, it would reduce the number of alternatives which could have a significant performance improvement on some code.

jezhiggins:
> I'd really like the issues with commits that don't pass tests to be resolved first. Some of the non-compiling commits in the previous work which was merged in that huge PR have already caused us problems when bisecting the history. I'm OK with the non-compiling commits in that PR of old work, because its better than leaving to rot forever. However I think we should be trying to do better for new work. Especially when all that is required is squashing a fix commit in to the correct preceding commit. I'm perfectly fine with the rest being at the discretion of those who know this part of the code base best.

I understand :+1: I've gone through the PR, fixing up those problematic commits. It should be good to merge now.

Issue #5797, opened 2021-02-05
Regression: Recent changes to byte-operators break proofs in aws-c-common


hannes-steffenhagen-diffblue:
CBMC version: 5.23.0
Operating system: Linux
Exact command line resulting in the issue:

With the CProver tools on PATH and the project setup for aws-c-common done, when inside the `verification/cbmc/proofs/aws_byte_buf_write_be16` directory:

```
make result
```

What behaviour did you expect: Should verify without errors
What happened instead: Crashes with an invariant violation in bitvector conversion when adding array constraints.

Here's the invariant violation report:
<details>
<summary>Reason: source bitvector width shall equal the destination bitvector width </summary>

```
--- begin invariant violation report ---
Invariant check failed
/home/hannes/Documents/Diffblue/cbmc/src/solvers/flattening/boolbv_typecast.cpp:505 function: type_conversion
Condition: src_width == dest_width
Reason: source bitvector width shall equal the destination bitvector width
Backtrace:
/home/hannes/Documents/Diffblue/cbmc-builds/cmake-build-debug/bin/cbmc(print_backtrace(std::ostream&)+0x51) [0x55e0e0ff8210]
/home/hannes/Documents/Diffblue/cbmc-builds/cmake-build-debug/bin/cbmc(get_backtrace[abi:cxx11]()+0x44) [0x55e0e0ff841c]
/home/hannes/Documents/Diffblue/cbmc-builds/cmake-build-debug/bin/cbmc(std::enable_if<std::is_base_of<invariant_failedt, invariant_failedt>::value, void>::type invariant_violated_structured<invariant_failedt, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)+0x52) [0x55e0e087d5e5]
/home/hannes/Documents/Diffblue/cbmc-builds/cmake-build-debug/bin/cbmc(+0x7becb3) [0x55e0e087ccb3]
[...]

PR #5798, opened 2021-02-05
Fix/gh issue 5797

hannes-steffenhagen-diffblue:
This fixes a small logic issue in `unpack_rec`. See https://github.com/diffblue/cbmc/issues/5797 for details.

### DIFF
--- a/src/solvers/lowering/byte_operators.cpp
+++ b/src/solvers/lowering/byte_operators.cpp
@@ -913,15 +913,17 @@ static exprt unpack_rec(
       }
     }
 
+    auto const src_as_bitvector = typecast_exprt::conditional_cast(
+      src, bv_typet{numeric_cast_v<std::size_t>(last_bit)});
+    auto const byte_type = bv_typet{8};
     exprt::operandst byte_operands;
     for(; bit_offset < last_bit; bit_offset += 8)
     {
       extractbits_exprt extractbits(
-        typecast_exprt::conditional_cast(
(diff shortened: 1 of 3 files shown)


src/solvers/lowering/byte_operators.cpp
```diff
@@ -918,7 +919,7 @@ static exprt unpack_rec(
     {
       extractbits_exprt extractbits(
         typecast_exprt::conditional_cast(
-          src, bv_typet{numeric_cast_v<std::size_t>(last_bit)}),
+          src, bv_typet{numeric_cast_v<std::size_t>(total_bits)}),
```
tautschnig: How about factoring this out of the loop completely? It's the same expression for each iteration of the loop, so only constructing it once would actually be beneficial.

tautschnig (approved):
Thank you for the additional cleanup! I'm not entirely sure, but should the commit message actually say "Fixes: #5797" instead of "See #5797" to automatically close the issue once this PR is merged?

src/solvers/lowering/byte_operators.cpp
hannes-steffenhagen-diffblue: @tautschnig I can pull that change into the first refactoring commit, that would probably be an improvement yes.

tautschnig: Just make sure you don't hoist it too high up so that you incidentally fix the issue by mere refactoring :-)

NlightNFotis (approved):
LGTM

PR #5799, opened 2021-02-05
dump-c: handle further byte_update union initializers

tautschnig:
The changes of a8592a73 made the C front-end introduce byte updates to
ensure compliance with the behaviour of popular compilers when
initialising unions. As byte updates aren't part the C language, dump-c
must make sure such union initialisers using byte updates do not end up
in the (re-)generated output.

db4f25cb was the first step in dump-c in this direction, and 3a24545
introduced further rules. The CSmith test generated with random seed
1612048908, however, demonstrated that dump-c's handling still didn't
cover all cases of byte_update expressions introduced by the C
front-end. This commit now handles those further cases.

The regression test is a minified version of that CSmith test.

### DIFF
--- a/regression/goto-instrument/dump-union2/main.c
+++ b/regression/goto-instrument/dump-union2/main.c
@@ -19,6 +19,13 @@ union U6 {
 
 union U6 g_1197 = {1L};
 
+union U4 {
+  signed f0 : 6;
+  int f3;
+};
+
+union U4 g_1408 = {-1L};
+
 int main()
 {
   assert(g_2110.f0 == 53747);
--- a/src/goto-instrument/dump_c.cpp
+++ b/src/goto-instrument/dump_c.cpp
@@ -1450,7 +1450,8 @@ void dump_ct::cleanup_expr(exprt &expr)
                                         ID_initializer_list,
                                         std::move(designated_initializer2)};
           expr.swap(initializer_list);
-          break;
+
+          return;
         }
       }
     }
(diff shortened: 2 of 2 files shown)


NlightNFotis:
Thanks for this! We had a similar issue crop up in https://github.com/diffblue/cbmc/runs/1830186932. Is that also a related issue?

On a more general note, I'm not very familiar with CSmith. Is there any material to quickly familiarise myself with it to that I can interpret those reports better, or is it better if I get to spend more time diving deeper on it?

tautschnig:
> Thanks for this! We had a similar issue crop up in https://github.com/diffblue/cbmc/runs/1830186932. Is that also a related issue?

It's the very same issue and should be fixed by this PR.

> On a more general note, I'm not very familiar with CSmith. Is there any material to quickly familiarise myself with it to that I can interpret those reports better, or is it better if I get to spend more time diving deeper on it?

CSmith lives here: https://embed.cs.utah.edu/csmith/ - but our `scripts/csmith.sh` might be all that you'd need to look at.

martin-cs (commented):
I'm not sure if I see how this fixes the issue and the commit message doesn't say but it seems like the kind of change that would.

tautschnig:
> I'm not sure if I see how this fixes the issue and the commit message doesn't say but it seems like the kind of change that would.

Thank you for calling this out! I have updated the text of the commit message (and also the pull request summary) to hopefully clarify what is going on here.

martin-cs (approved):
Much clearer.  Thank you!

PR #5800, opened 2021-02-05
Array-element sensitive SSA: avoid string-to-integer-to-string

tautschnig:
There is no need to use decimal indices in field-sensitive SSA
identifiers: we just need pairwise distinct names. Thus, just use the
irep-stored value instead of converting the value to an integer and back
to a string.

On one (array-heavy) benchmark, this reduced the time constructing the
symex target equation from 74 to 60 seconds.


martin-cs (approved):
Seems reasonable.  I guess this is found by profiling?  Normally I would want some kind of evidence for things that claim to be performance improvements but this looks clear cut enough that it won't produce performance regressions.

kroening (approved):

PR #5801, opened 2021-02-05
goto-checker properties: do not ignore ERROR

tautschnig:
The assignment of the value of b to overwrite the known-ERROR status of
a resulted in not reporting an overall result of "ERROR" when only some
of the properties had "ERROR" status.

Added a unit test of this code to ensure this keeps working as expected.


martin-cs (approved):
This rather feels like it needs some kind of automata for what are acceptable status transitions.

peterschrammel (approved):
Thanks for turning the documentation into unit tests.

PR #5803, opened 2021-02-05
boolbvt::convert_bv: avoid unnecessary vector copy

tautschnig:
We can safely move the vector, there is no need to keep the intermediate return
value constant.


kroening:
Note that the copy is still happening -- `convert_bv` returns a reference.

tautschnig:
> Note that the copy is still happening -- `convert_bv` returns a reference.

Are you referring to copies that the callers of `convert_bv` might create? Yes, those will continue to exist. This PR just tries to avoid (unnecessary) copying within `convert_bv`.

kroening (approved):

PR #5807, opened 2021-02-08
Refactor process_goto_program so that all tools use common processing code

martin-cs:
`*_parse_optionst::process_goto_program` appears in `cbmc`, `goto-analyze` and `goto-diff` (and maybe the Java tools, @peterschrammel @hannes-steffenhagen-diffblue who is handling these at the moment?) with slight variations, bugs, inconsistencies and omissions.  This PR carefully makes them the same (so that we can `git-bisect` to find any issues this might introduce) and then pulls the common code out into a helper function.  This is another step towards a common format for goto-programs and stronger invariants.

This does not break tested functionality but it will have an impact on untested functionality.  Mostly, as far as I can see, it goes from 'silently doing the wrong thing' to 'maybe works, maybe breaks in an obvious way'.  i consider this a net win.

### DIFF
--- a/regression/goto-analyzer/approx-array-variable-const-fp/test.desc
+++ b/regression/goto-analyzer/approx-array-variable-const-fp/test.desc
@@ -1,7 +1,7 @@
 CORE
 main.c
 --show-goto-functions --pointer-check
-^Removing function pointers and virtual functions$
+^Removal of function pointers and virtual functions$
 ^\s*IF fp_tbl\[.*i\] == f2 THEN GOTO [0-9]$
 ^\s*IF fp_tbl\[.*i\] == f3 THEN GOTO [0-9]$
 ^\s*IF fp_tbl\[.*i\] == f4 THEN GOTO [0-9]$
--- a/src/goto-analyzer/goto_analyzer_parse_options.cpp
+++ b/src/goto-analyzer/goto_analyzer_parse_options.cpp
@@ -913,6 +913,12 @@ bool goto_analyzer_parse_optionst::process_goto_program(
     // checks don't know about adjusted float expressions
     adjust_float_expressions(goto_model);
 
+    if(options.get_bool_option("string-abstraction"))
+    {
+      log.status() << "String Abstraction" << messaget::eom;
+      string_abstraction(goto_model, log.get_message_handler());
+    }
+
     // recalculate numbers, etc.
     goto_model.goto_functions.update();
 
--- a/src/goto-analyzer/goto_analyzer_parse_options.cpp
+++ b/src/goto-analyzer/goto_analyzer_parse_options.cpp
@@ -147,6 +147,9 @@ void goto_analyzer_parse_optionst::get_command_line_options(optionst &options)
     options.set_option("error-label", cmdline.get_values("error-label"));
 #endif
 
+  // all checks supported by goto_check
+  PARSE_OPTIONS_GOTO_CHECK(cmdline, options);
+
   // The user should either select:
   //  1. a specific analysis, or
   //  2. a tuple of task / analyser options / outputs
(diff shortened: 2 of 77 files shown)


martin-cs:
Commit by commit reviewing strongly encouraged.

src/goto-programs/process_goto_program.h
```diff
@@ -0,0 +1,29 @@
...
+
+\*******************************************************************/
+
+/// \file
+/// Common processing and simplification of goto_programts.
+/// This includes linking libraries and removing a number of
+/// more complex types (vectors, complex, etc.) and constructs
+/// (returns, function pointers, etc.).
+/// This is can be used after initialize_goto_model but before
+/// analysis.  It is not madatory but is used by most tools.
```
tautschnig: Nit picking: s/madatory/mandatory/
But also: could/should this comment be moved towards the function documentation rather than the file's documentation?

src/goto-analyzer/goto_analyzer_parse_options.cpp
```diff
@@ -146,6 +146,9 @@ void goto_analyzer_parse_optionst::get_command_line_options(optionst &options)
     options.set_option("error-label", cmdline.get_values("error-label"));
 #endif
 
+  // all checks supported by goto_check
+  PARSE_OPTIONS_GOTO_CHECK(cmdline, options);
```
tautschnig, line 125: Ah! I had noticed at some point that I had to use `goto-instrument --*-check` before using `goto-analyzer`. Debugging `goto-analyzer` is still on my backlog...

src/goto-analyzer/goto_analyzer_parse_options.cpp
```diff
@@ -898,6 +900,11 @@ bool goto_analyzer_parse_optionst::process_goto_program(
     remove_function_pointers(
       ui_message_handler, goto_model, options.get_bool_option("pointer-check"));
 
+    mm_io(goto_model);
+
+    // instrument library preconditions
+    instrument_preconditions(goto_model);
```
tautschnig: Just noting that this isn't noted in the commit message. On a second thought, this change should probably go with the earlier commit that enabled library support in goto-analyzer.

regression/goto-analyzer/approx-array-variable-const-fp/test.desc
```diff
@@ -1,7 +1,7 @@
 CORE
 main.c
 --show-goto-functions --pointer-check
-^Removing function pointers and virtual functions$
+^Removal of function pointers and virtual functions$
```
tautschnig, line 4: Perhaps put that change of the status message in a commit of its own so that the biggest change in this series does come without any observable impact on tests?

tautschnig (approved):
Thank you for this clean-up and doing it in such an easy-to-review way!

src/goto-analyzer/goto_analyzer_parse_options.cpp
martin-cs: Yeah, a historical fuck up, esp. as the `GOTO_CHECK` options where given in `--help` and the options were parsed...

src/goto-programs/process_goto_program.h
martin-cs: Fixed.

src/goto-analyzer/goto_analyzer_parse_options.cpp
martin-cs: Fixed.

regression/goto-analyzer/approx-array-variable-const-fp/test.desc
martin-cs: Resolved

martin-cs:
@hannes-steffenhagen-diffblue and / or @peterschrammel any chance of a review?  It looks big but commit by commit it is actually simple and should be easy to review quickly.

src/goto-analyzer/goto_analyzer_parse_options.cpp
```diff
@@ -600,6 +581,10 @@ int goto_analyzer_parse_optionst::doit()
 
   goto_model = initialize_goto_model(cmdline.args, ui_message_handler, options);
 
+  // Perserve backwards compatability in processing
```
peterschrammel:
```suggestion
  // Preserve backwards compatibility in processing
```

peterschrammel (approved):

Issue #5473, opened 2020-08-25
Cprover should have an `__enum_in_range` primitive


danielsn:
It is currently difficult to use the `--enum-range-check` option, because it is hard to ensure in a harness that an enum is within range. IF there was a 
`__CPROVER_enum_is_in_range(e)` primitive, we could use it like 
`__CPROVER_assume(__CPROVER_enum_is_in_range(e))`
and 
`assert(__CPROVER_enum_is_in_range(e))`

feliperodri:
@hannes-steffenhagen-diffblue is there anyone currently working on this issue? Any updates?

TGWDB:
See [pull request here](https://github.com/diffblue/cbmc/pull/5808).

TGWDB:
Fixed in https://github.com/diffblue/cbmc/pull/5808

PR #5808, opened 2021-02-09
Cprover enum is in range

TGWDB:

### DIFF
--- a/doc/cprover-manual/api.md
+++ b/doc/cprover-manual/api.md
@@ -125,6 +125,38 @@ stored in the given arrays are equal. The function
 the array **dest**. The function **\_\_CPROVER\_array\_set** initializes
 the array **dest** with the given value.
 
+
+#### \_\_CPROVER\_enum\_is\_in\_range
+
+```C
+__CPROVER_bool __CPROVER_enum_is_in_range();
+```
+
+The function **\_\_CPROVER\_enum\_is\_in\_range** is used to check
+that an enumeration has one of the defined enumeration values. In
+the following example `__CPROVER_enum_is_in_range(ev1)` will return
+true and the assertion will pass
+```C
+enum my_enum { first, second };
+
+int main()
+{
+  enum my_enum ev1 = second;
+  assert(__CPROVER_enum_is_in_range(ev1));
+}
+```
+However, in the example below the assertion will fail
+```C
+enum my_enum { first, second };
+
+int main()
+{
+  enum my_enum ev1 = second + 1;
+  assert(__CPROVER_enum_is_in_range(ev1));
+}
+```
+
+
 #### Uninterpreted Functions
 
 Uninterpreted functions are documented [here](./modeling-nondeterminism.md)).
--- a/regression/ansi-c/enum_is_in_range/enum_test10.desc
+++ b/regression/ansi-c/enum_is_in_range/enum_test10.desc
@@ -1,0 +1,12 @@
+CORE
+enum_test10.c
+
+^EXIT=(6|70)$
+^SIGNAL=0$
+^file enum_test10.c line \d+ function main: __CPROVER_enum_is_in_range expects enum, but \(i\) has type `signed int`$
+--
+^\[main.assertion.1\] line \d+ assertion __CPROVER_enum_is_in_range\(i\): SUCCESS$
+^\[main.assertion.1\] line \d+ assertion __CPROVER_enum_is_in_range\(i\): FAILURE$
+^\*\*\*\* WARNING: no body for function __CPROVER_enum_is_in_range$
+--
+This test the type checking of argument for __CPROVER_enum_is_in_range
--- a/regression/cbmc/enum_is_in_range/enum_test3.c
+++ b/regression/cbmc/enum_is_in_range/enum_test3.c
@@ -1,0 +1,24 @@
+#include <assert.h>
+
+enum my_enum
+{
+  first,
+  second,
+  third,
+  fourth,
+  fifth
+};
+
+int main()
+{
+  enum my_enum ev1;
+  enum my_enum ev2;
+  ev1 = first;
+  ev2 = fifth;
+  while(ev1 != ev2)
+  {
+    ev1++;
+    assert(__CPROVER_enum_is_in_range(ev1));
+  }
+  return 0;
+}
--- a/src/ansi-c/c_typecheck_expr.cpp
+++ b/src/ansi-c/c_typecheck_expr.cpp
@@ -2841,6 +2841,33 @@ exprt c_typecheck_baset::do_special_functions(
     overflow.type() = bool_typet{};
     return overflow;
   }
+  else if(identifier == CPROVER_PREFIX "enum_is_in_range")
+  {
+    // Check correct number of arguments
+    if(expr.arguments().size() != 1)
+    {
+      std::ostringstream error_message;
+      error_message << expr.source_location().as_string() << ": " << identifier
+                    << " takes exactly 1 argument, but "
+                    << expr.arguments().size() << " were provided";
+      throw invalid_source_file_exceptiont{error_message.str()};
+    }
+    auto arg1 = expr.arguments()[0];
+    typecheck_expr(arg1);
+    if(!can_cast_type<c_enum_tag_typet>(arg1.type()))
+    {
+      // Can't enum range check a non-enum
+      std::ostringstream error_message;
+      error_message << expr.source_location().as_string() << ": " << identifier
+                    << " expects enum, but (" << expr2c(arg1, *this)
+                    << ") has type `" << type2c(arg1.type(), *this) << '`';
+      throw invalid_source_file_exceptiont{error_message.str()};
+    }
+    else
+    {
+      return expr;
+    }
+  }
   else if(
     identifier == "__builtin_add_overflow" ||
     identifier == "__builtin_sub_overflow" ||
--- a/src/goto-programs/builtin_functions.cpp
+++ b/src/goto-programs/builtin_functions.cpp
@@ -610,6 +610,32 @@ exprt make_va_list(const exprt &expr)
   return result;
 }
 
+void goto_convertt::do_enum_is_in_range(
+  const exprt &lhs,
+  const symbol_exprt &function,
+  const exprt::operandst &arguments,
+  goto_programt &dest)
+{
+  PRECONDITION(arguments.size() == 1);
+  const auto enum_expr = arguments.front();
+  const auto enum_type =
+    type_try_dynamic_cast<c_enum_tag_typet>(enum_expr.type());
+  PRECONDITION(enum_type);
+  const c_enum_typet &c_enum_type = ns.follow_tag(*enum_type);
+  const c_enum_typet::memberst enum_values = c_enum_type.members();
+
+  exprt::operandst disjuncts;
+  for(const auto &enum_value : enum_values)
+  {
+    constant_exprt val{enum_value.get_value(), *enum_type};
+    disjuncts.push_back(equal_exprt(enum_expr, std::move(val)));
+  }
+
+  code_assignt assignment(lhs, disjunction(disjuncts));
+  assignment.add_source_location() = function.source_location();
+  copy(assignment, ASSIGN, dest);
+}
+
 /// add function calls to function queue for later processing
 void goto_convertt::do_function_call_symbol(
   const exprt &lhs,
@@ -733,6 +759,10 @@ void goto_convertt::do_function_call_symbol(
       error() << identifier << " expected not to have LHS" << eom;
       throw 0;
     }
+  }
+  else if(identifier == CPROVER_PREFIX "enum_is_in_range")
+  {
+    do_enum_is_in_range(lhs, function, arguments, dest);
   }
   else if(
     identifier == CPROVER_PREFIX "assert" ||
--- a/src/goto-programs/goto_convert_class.h
+++ b/src/goto-programs/goto_convert_class.h
@@ -588,6 +588,26 @@ protected:
   bool get_string_constant(const exprt &expr, irep_idt &);
   exprt get_constant(const exprt &expr);
 
+  /// \brief Converts calls to the built in `enum_is_in_range` function to a
+  ///    test that the given enum value is in the valid range of values for that
+  ///    enum type.
+  ///
+  /// Note that the check for the range of values is done by creating a
+  /// disjunction comparing the expression with each possible valid value.
+  /// \param lhs: The destination for the generated assignment.
+  /// \param function: The `enum_is_in_range` symbol of the source function call.
+  /// \param arguments: A collection of arguments, which is expected to contain a
+  ///    single argument which is an expression that resolves to a value of
+  ///    enum type. The arguments are expected to have been prevalidated by the
+  ///    typechecking process.
+  /// \param dest: The goto program into which the generated assignment is
+  ///    copied.
+  void do_enum_is_in_range(
+    const exprt &lhs,
+    const symbol_exprt &function,
+    const exprt::operandst &arguments,
+    goto_programt &dest);
+
   // some built-in functions
   void do_atomic_begin(
     const exprt &lhs,
--- a/regression/ansi-c/enum_is_in_range/enum_test10.c
+++ b/regression/ansi-c/enum_is_in_range/enum_test10.c
@@ -1,0 +1,26 @@
+#include <assert.h>
+
+enum my_enum
+{
+  first,
+  second,
+  third,
+  fourth,
+  fifth
+};
+
+int main()
+{
+  enum my_enum ev1;
+  enum my_enum ev2;
+  ev1 = first;
+  ev2 = fifth;
+  int i = 0;
+  while(i < 10)
+  {
+    ev1++;
+    assert(__CPROVER_enum_is_in_range(i));
+    i++;
+  }
+  return 0;
+}
--- a/regression/ansi-c/enum_is_in_range/enum_test12.c
+++ b/regression/ansi-c/enum_is_in_range/enum_test12.c
@@ -1,0 +1,20 @@
+#include <assert.h>
+
+enum my_enum
+{
+  first,
+  second,
+  third,
+  fourth,
+  fifth
+};
+
+int main()
+{
+  enum my_enum ev1;
+  enum my_enum ev2;
+  ev1 = first;
+  ev2 = fifth;
+  assert(__CPROVER_enum_is_in_range(ev1, ev2));
+  return 0;
+}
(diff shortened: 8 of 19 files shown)


regression/cbmc/enum_is_in_range/enum_test3.c
```diff
@@ -0,0 +1,24 @@
...
+  fifth
+};
+
+int main()
+{
+  enum my_enum ev1;
+  enum my_enum ev2;
+  ev1 = first;
+  ev2 = fifth;
+  while(!(ev1 == ev2))
```
peterschrammel: ⛏️ 
```suggestion
  while(ev1 != ev2) 
```

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2841,6 +2841,50 @@ exprt c_typecheck_baset::do_special_functions(
     overflow.type() = bool_typet{};
     return overflow;
   }
+  else if(identifier == CPROVER_PREFIX "enum_is_in_range")
+  {
+    // Check correct number of arguments
+    if(expr.arguments().size() != 1)
+    {
+      error().source_location = expr.source_location();
+      error() << "__enum_is_in_range expects one argument" << eom;
```
hannes-steffenhagen-diffblue: Just throw an exception.

doc/cprover-manual/api.md
```diff
@@ -125,6 +125,37 @@ stored in the given arrays are equal. The function
...
+
+#### \_\_CPROVER\_enum\_is\_in\_range
+
+```C
+__CPROVER_bool __CPROVER_enum_is_in_range();
+```
+
+The function **\_\_CPROVER\_enum\_is\_in\_range** returns true
+if their one argument (an `enum`) is one of the defined values for
+the enumeration. For example
```
hannes-steffenhagen-diffblue: I would probably rewrite that as

> in this example, `__CPROVER_enum_is_in_range(ev1)` would return true, while in that example...

Because otherwise we have a block of code and it's a bit unclear 'what' returns true/false. Otherwise, maybe just say something like

> In this example the assertion will pass

> while in this example it will fail.

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2841,6 +2841,50 @@ exprt c_typecheck_baset::do_special_functions(
...
+      std::ostringstream error_message;
+      error_message << expr.source_location().as_string() << ": "
+                    << "__enum_is_in_range expects one argument, "
+                    << expr.arguments().size() << " were provided";
+      throw invalid_source_file_exceptiont{error_message.str()};
+    }
+    // Replace expr with giant boolean that checks all the values
```
peterschrammel: I don't think the translation into a condition should be done here, but rather later as a pass on the goto program.

tautschnig: +1

src/ansi-c/c_typecheck_expr.cpp
tautschnig: It should be `CPROVER_PREFIX` "enum_is_in_range`

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2841,6 +2841,50 @@ exprt c_typecheck_baset::do_special_functions(
...
+    // Check correct number of arguments
+    if(expr.arguments().size() != 1)
+    {
+      error().source_location = expr.source_location();
+      error() << "__enum_is_in_range expects one argument" << eom;
+      std::ostringstream error_message;
+      error_message << expr.source_location().as_string() << ": "
+                    << "__enum_is_in_range expects one argument, "
+                    << expr.arguments().size() << " were provided";
+      throw invalid_source_file_exceptiont{error_message.str()};
```
tautschnig, line 2853: I believe that in this code we would just use `throw 0;`

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2841,6 +2841,50 @@ exprt c_typecheck_baset::do_special_functions(
...
+    // of the enumeration as disjunction
+    // This code is similar to goto_checkt::enum_range_check
+    auto arg1 = expr.arguments()[0];
+    typecheck_expr(arg1);
+    if(
+      auto c_enum_tag_type =
+        type_try_dynamic_cast<c_enum_tag_typet>(arg1.type()))
+    {
+      symbolt enum_type = this->lookup(c_enum_tag_type->get_identifier());
+      const c_enum_typet &c_enum_type = to_c_enum_type(enum_type.type);
```
tautschnig: Please use `const c_enum_typet &c_enum_type = follow_tag(*c_enum_tag_type);` instead of explicitly going via the symbol.

TGWDB: Fixed.

src/ansi-c/c_typecheck_expr.cpp
TGWDB: Fixed in both places.

src/ansi-c/c_typecheck_expr.cpp
TGWDB: Fixed in both places.

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2841,6 +2841,45 @@ exprt c_typecheck_baset::do_special_functions(
...
+      auto c_enum_tag_type =
+        type_try_dynamic_cast<c_enum_tag_typet>(arg1.type()))
+    {
+      const c_enum_typet &c_enum_type = follow_tag(*c_enum_tag_type);
+      const c_enum_typet::memberst enum_values = c_enum_type.members();
+
+      std::vector<exprt> disjuncts;
+      for(const auto &enum_value : enum_values)
+      {
+        const constant_exprt val{enum_value.get_value(), *c_enum_tag_type};
```
tautschnig: Rather than making this `const` I'd even opt to `std::move` it in the construction of the `equal_exprt` below.

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2841,6 +2841,45 @@ exprt c_typecheck_baset::do_special_functions(
...
   }
+  else if(identifier == CPROVER_PREFIX "enum_is_in_range")
+  {
+    // Check correct number of arguments
+    if(expr.arguments().size() != 1)
+    {
+      error().source_location = expr.source_location();
+      error() << identifier << " expects one argument, "
+              << expr.arguments().size() << " were provided" << eom;
+      throw 0;
```
tautschnig: Is it worth adding a test also covering this error condition? You've got a bunch of tests already, adding one more would help tick another box :-)

tautschnig (commented):
I think this is great work, and would just ask for
1) @peterschrammel's comment to be addressed and
2) the commit pile to be cleaned up.

src/ansi-c/c_typecheck_expr.cpp
hannes-steffenhagen-diffblue: Don't throw integers please!

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2841,6 +2841,45 @@ exprt c_typecheck_baset::do_special_functions(
...
+      }
+
+      return disjunction(disjuncts);
+    }
+    else
+    {
+      // Can't enum range check a non-enum
+      error().source_location = expr.source_location();
+      error() << identifier << " expects enum argument type" << eom;
+      throw 0;
```
hannes-steffenhagen-diffblue: Again, please don't throw integers. This is a `invalid_source_file_exceptiont`

src/ansi-c/c_typecheck_expr.cpp
hannes-steffenhagen-diffblue: @peterschrammel Could you elaborate where exactly you'd like this to happen?

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2841,6 +2841,45 @@ exprt c_typecheck_baset::do_special_functions(
...
+        disjuncts.push_back(equal_exprt(arg1, val));
+      }
+
+      return disjunction(disjuncts);
+    }
+    else
+    {
+      // Can't enum range check a non-enum
+      error().source_location = expr.source_location();
+      error() << identifier << " expects enum argument type" << eom;
```
hannes-steffenhagen-diffblue: Again, would be better if the error message also stated what we got _instead_ of what it expected.

hannes-steffenhagen-diffblue (commented):
I think the code is mostly OK, withholding approval for commit / error handling cleanup.

src/ansi-c/c_typecheck_expr.cpp
hannes-steffenhagen-diffblue: @tautschnig we _do_ do that, but just as a historical artifact. We'd agreed in the past we wouldn't do that for new code anymore.

src/ansi-c/c_typecheck_expr.cpp
peterschrammel: I think it would fit well into the goto_check pass. Then you can even reuse the logic for generating the conditons that is already there.

hannes-steffenhagen-diffblue:
@tautschnig We seem to be disagreeing on the error handling. To avoid us giving conflicting advice in the future, we should probably come to an agreement on what we do with error handling. My view was that we should avoid adding additional unstructured exceptions/remove unstructured exceptions when we come across them. Do you disagree with that on principle, or do you only have a problem with it in this particular case?

src/ansi-c/c_typecheck_expr.cpp
hannes-steffenhagen-diffblue: @peterschrammel This should work without enum-range-check being enabled as it's more generally useful though. Or do you mean just have it as something that happens during goto check? Because I feel that'd be slightly confusing as it's not a check.

src/ansi-c/c_typecheck_expr.cpp
TGWDB: Fixed.

src/ansi-c/c_typecheck_expr.cpp
TGWDB: Back to using the exception now.

src/ansi-c/c_typecheck_expr.cpp
TGWDB: Fixed.

doc/cprover-manual/api.md
TGWDB: Improved this.

src/ansi-c/c_typecheck_expr.cpp
TGWDB: Done

src/ansi-c/c_typecheck_expr.cpp
TGWDB: Done

tautschnig:
> @tautschnig We seem to be disagreeing on the error handling. To avoid us giving conflicting advice in the future, we should probably come to an agreement on what we do with error handling. My view was that we should avoid adding additional unstructured exceptions/remove unstructured exceptions when we come across them. Do you disagree with that on principle, or do you only have a problem with it in this particular case?

My apologies if I caused confusion here. I'm all for structured exceptions in general, except the C front-end does things differently (but is consistent in doing it in such a way). I'd thus ask for this to remain consistent, not ruling out that we change this at a later point - but then we should make a wholesale change.

Edit: to be very explicit: my comment is just about this particular case.

tautschnig:
To also explain at a more technical level why the C front-end (rightly) does `throw 0`, though only for handling type checking errors that can be triggered by syntactically valid input files: the message handler will ensure that the error message is formatted in a style matching the compiler (GCC or Visual Studio). That `throw 0` will be caught by https://github.com/diffblue/cbmc/blob/develop/src/ansi-c/ansi_c_typecheck.cpp#L53. If we were to use structured exceptions here, we would need to make sure the behaviour of printing a well-formatted message is preserved.

src/ansi-c/c_typecheck_expr.cpp
thomasspriggs: @peterschrammel It would be nice to be able to reuse the same check logic for both kinds of check. However we don't think it fits into `goto_check` because it is replacing an existing expression rather than adding new checks. Would you be OK with us adding it in `src/goto-programs/builtin_functions.cpp`, function `goto_convertt::do_function_call_symbol`? This seems to be the place where the other built in functions are implemented / replaced.

peterschrammel: `builtin_functions` is certainly a better place than `c_typecheck_expr`.

TGWDB: @peterschrammel This is addressed now.

src/goto-programs/builtin_functions.cpp
```diff
@@ -610,6 +610,34 @@ exprt make_va_list(const exprt &expr)
...
+{
+  PRECONDITION(arguments.size() == 1);
+  const auto enum_expr = arguments.front();
+  const auto enum_type =
+    type_try_dynamic_cast<c_enum_tag_typet>(enum_expr.type());
+  PRECONDITION(enum_type);
+  const c_enum_typet &c_enum_type = ns.follow_tag(*enum_type);
+  const c_enum_typet::memberst enum_values = c_enum_type.members();
+
+  std::vector<exprt> disjuncts;
```
peterschrammel: ⛏️ 
```suggestion
  exprt::operandst disjuncts;
```

src/goto-programs/goto_convert_class.h
```diff
@@ -589,6 +589,11 @@ class goto_convertt:public messaget
   exprt get_constant(const exprt &expr);
 
   // some built-in functions
+  void do_enum_is_in_range(
```
peterschrammel, line 605: ⛏️  Some docs would be appreciated.

peterschrammel (approved):

TGWDB:
@hannes-steffenhagen-diffblue Can you please (re)review this PR now it's been redone taking into account above comments (and also tidied up).

src/goto-programs/builtin_functions.cpp
```diff
@@ -610,6 +610,34 @@ exprt make_va_list(const exprt &expr)
...
+  exprt::operandst disjuncts;
+  for(const auto &enum_value : enum_values)
+  {
+    constant_exprt val{enum_value.get_value(), *enum_type};
+    disjuncts.push_back(equal_exprt(enum_expr, std::move(val)));
+  }
+
+  const auto disjunction = ::disjunction(disjuncts);
+
+  code_assignt assignment(lhs, disjunction);
```
tautschnig: Nit picking: I'd do `code_assignt assignment{lhs, disjunction(disjuncts)};` to avoid constructing (and then deleting) a temporary.

src/goto-programs/goto_convert_class.h
```diff
@@ -588,6 +588,26 @@ class goto_convertt:public messaget
   bool get_string_constant(const exprt &expr, irep_idt &);
   exprt get_constant(const exprt &expr);
 
+  /// \brief Converts calls to the built in `enum_is_in_range` function to a
+  ///    test that the given enum value is in the valid range of values for that
+  ///    enum type.
+  ///
+  /// Note that the check for the range of values is done by creating a
+  /// disjunction comparing the expression with each possible valid value.
+  /// \param lhs The destination for the generated assignment.
```
tautschnig: Nit picking: I believe the coding guidelines suggest to use a colon after the parameter's name.

regression/cbmc/enum_is_in_range/enum_test10.desc
```diff
@@ -0,0 +1,12 @@
...
+
+^EXIT=6$
+^SIGNAL=0$
+^file enum_test10.c line \d+ function main: __CPROVER_enum_is_in_range expects enum, but \(i\) has type `signed int`$
+--
+^\[main.assertion.1\] line \d+ assertion __CPROVER_enum_is_in_range\(i\): SUCCESS$
+^\[main.assertion.1\] line \d+ assertion __CPROVER_enum_is_in_range\(i\): FAILURE$
+^\*\*\*\* WARNING: no body for function __CPROVER_enum_is_in_range$
+--
+This test the type checking of argument for __CPROVER_num_is_in_range
```
tautschnig: `s/_num_/_enum_/`

regression/cbmc/enum_is_in_range/enum_test12.desc
```diff
@@ -0,0 +1,12 @@
...
+
+^EXIT=6$
+^SIGNAL=0$
+^file enum_test12.c line \d+ function main: __CPROVER_enum_is_in_range takes exactly 1 argument, but 2 were provided$
+--
+^\[main.assertion.1\] line \d+ assertion __CPROVER_enum_is_in_range\(.*\): SUCCESS$
+^\[main.assertion.1\] line \d+ assertion __CPROVER_enum_is_in_range\(.*\): FAILURE$
+^\*\*\*\* WARNING: no body for function __CPROVER_enum_is_in_range$
+--
+This test the type checking of argument for __CPROVER_num_is_in_range
```
tautschnig: `s/_num_/_enum_`

regression/cbmc/enum_is_in_range/enum_test10.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+enum_test10.c
+
+^EXIT=6$
+^SIGNAL=0$
+^file enum_test10.c line \d+ function main: __CPROVER_enum_is_in_range expects enum, but \(i\) has type `signed int`$
```
tautschnig: This test ought to live in the `regression/ansi-c/` folder, I'd say. This will require modifying `EXIT=6` to `EXIT=(1|64)`

regression/cbmc/enum_is_in_range/enum_test12.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+enum_test12.c
+
+^EXIT=6$
+^SIGNAL=0$
+^file enum_test12.c line \d+ function main: __CPROVER_enum_is_in_range takes exactly 1 argument, but 2 were provided$
```
tautschnig: Should also go in `regression/ansi-c/`.

tautschnig (approved):
I would ask for the changes I've suggested to be addressed, but no longer have blocking concerns. Also, please add `Fixes: #5473` to one of the commit messages.

regression/ansi-c/enum_is_in_range/enum_test10.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+enum_test10.c
+
+^EXIT=(6|70)$
```
tautschnig, line 4: This isn't the right exit code. Will be fixed by #5837.

PR #5809, opened 2021-02-09
[CI] Port CMake Codecov job from Codebuild to Github Actions.

NlightNFotis:
This is part of our streamlining and cleanup of our CI infrastructure.

This change is supposed to bring functional parity between what
we support for coverage in Codebuild and GitHub Actions.


.github/workflows/pull-request-checks.yaml
```diff
@@ -378,3 +378,28 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Download testing and coverage dependencies
+        run: sudo apt-get install --no-install-recommends -y g++ flex bison cmake maven jq libxml2-utils dpkg-dev openjdk-11-jdk-headless lcov
+      - name: Configure CMake CBMC build with coverage instrumentation parameters
+        run: cmake -S . -Bbuild '-Denable_coverage=1' '-Dparallel_tests=2' '-DCMAKE_CXX_COMPILER=/usr/bin/g++'
```
tautschnig: Nit picking: we don't actually need the quotes here, do we?

.github/workflows/pull-request-checks.yaml
```diff
@@ -378,3 +378,28 @@ jobs:
...
+
+  codecov-coverage-report:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Download testing and coverage dependencies
+        run: sudo apt-get install --no-install-recommends -y g++ flex bison cmake maven jq libxml2-utils dpkg-dev openjdk-11-jdk-headless lcov
```
tautschnig: I don't think we need `dpkg-dev` for this build?

tautschnig (approved):
Do we have any way of knowing whether this genuinely works as expected?

tautschnig:
#5799 needs to be merged to address the CSmith failure (which can also be safely ignored for this PR).

.github/workflows/pull-request-checks.yaml
```diff
@@ -378,3 +378,28 @@ jobs:
...
+      - name: Configure CMake CBMC build with coverage instrumentation parameters
+        run: cmake -S . -Bbuild '-Denable_coverage=1' '-Dparallel_tests=2' '-DCMAKE_CXX_COMPILER=/usr/bin/g++'
+      - name: Execute CMake CBMC build
+        run: cmake --build build --target coverage -- -j2
+      - name: Collect coverage statistics
+        run: |
+          lcov --capture --directory build --output-file lcov.info
+          lcov --remove lcov.info '/usr/*' --output-file lcov.info
```
piotr-grabalski, line 411: Is there anything else that could be removed from trace file (e.g. miniSAT/or other SAT solver files)?

piotr-grabalski (approved):
Looks good

.github/workflows/pull-request-checks.yaml
NlightNFotis: No, we don't. You are right :)

.github/workflows/pull-request-checks.yaml
NlightNFotis: You are right on this as well.

NlightNFotis:
@tautschnig Addressing you question as to how we would know if this is working, I intend to push a change to this PR deactivating the other coverage jobs (the ones on code build), so they don't submit coverage statistics. In this case, if we see coverage report, it should be coming from this change, validating it.

.github/workflows/pull-request-checks.yaml
NlightNFotis: Hm, I'm not sure to be honest, and I'd rather not play around with the contents of the files for now, as this is supposed to be just a port with functionality parity.

Attempting any changes like this would cause us to need extra time to validate/test those changes. I am happy however to add them on a todo list for later inspection/implementation. Does that work for you, @piotr-grabalski ?

piotr-grabalski: Yes, that is OK for me.

I've been wondering if the report could be more accurate and display the test coverage of the code that is a part of CBMC rather than some third-party stuff that we are not involved in. I don't know how much of external code is going into the statistics so, as you said, more investigation will be required.

peterschrammel (approved):

.github/workflows/pull-request-checks.yaml
```diff
@@ -391,7 +391,7 @@ jobs:
           # This is needed in addition to -yq to prevent apt-get from asking for
           # user input
           DEBIAN_FRONTEND: noninteractive
-        run: sudo apt-get install --no-install-recommends -y g++ binutils flex bison cmake maven jq libxml2-utils openjdk-11-jdk-headless lcov
+        run: sudo apt-get install -y g++ gcc binutils flex bison cmake maven jq libxml2-utils openjdk-11-jdk-headless lcov
```
tautschnig: I think you're doing yourself a disservice in removing `--no-install-recommends`. Perhaps use `build-essential` to get all of `g++`, `gcc`, `binutils` and any other packages really required for building.

NlightNFotis: Hi Michael, why do you think it's a disservice to ourselves to remove `--no-install-recommends`? My rationale for removing it is that by keeping it here we may be subtly missing dependencies that are recommended by the system, but they really are essential for our build process, and by missing them we have all sorts of errors.

I guess this makes the process slightly more opaque (in terms of what dependencies will be installed, because they can be changed under our feet) but it's making it safer in terms of making the build much less likely to fail because of missed dependencies. What do you think?

tautschnig: I'm not entirely sure I understand whether you argue for or against `--no-install-recommends`, but let me try to clarify my position:
1) We should (ourselves) have a very clear understanding of the dependencies so that we can correctly document these in files like `COMPILING.md`. We cannot know whether our customers use/don't use `--no-install-recommends` (or their equivalents in APT configuration files). That said, a GitHub action need not be the one place to test our own understanding of dependencies, but it might well be a good one.
2) Installing recommended packages might make it easier to get the package lists right, but will likely also install all sorts of packages that we do not need. Installing packages takes time, and CI jobs shouldn't be wasting time.

NlightNFotis: Hm, I see your point. I'm convinced by point 1., so I'll re-introduce the option as soon as I stabilise an issue I'm having with CI.

Re 2., I can also see this point, but any time savings are marginal. As things are, it downloads about 300MB of packages in about 7 seconds, and it takes another 10-15 seconds to unpack those. At the same time, we take more than 30 minutes to compile and test.

chrisr-diffblue: Though the more separate things there are to download, the greater the chance of random network gliches killing a build. GitHub actions seems a lot more reliable than, say, Travis was when we used that, but still worth minimising your 'surface area' for glitches :-)

.github/workflows/pull-request-checks.yaml
```diff
@@ -385,3 +385,34 @@ jobs:
...
+      - name: Collect coverage statistics
+        run: |
+          lcov --capture --directory build --output-file lcov.info
+          lcov --remove lcov.info '/usr/*' --output-file lcov.info
+      - name: Upload coverage statistics to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: ./lcov.info
+          fail_ci_if_error: true
```
hannes-steffenhagen-diffblue, line 417: ❓ Last concern here: Is this a required status check? I don't think it should be, given how codecov has been a bit flaky in the past. IMHO codecov checks should be advisory.

CMakeLists.txt
```diff
@@ -149,6 +149,8 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
               "$<TARGET_FILE:jbmc>"
               "$<TARGET_FILE:jdiff>"
               "$<TARGET_FILE:smt2_solver>"
+              "$<TARGET_FILE:symtab2gb>"
```
hannes-steffenhagen-diffblue, line 152: FWIW all this "$<TARGET_FILE:` stuff in here isn't _invalid_ per se, but it also doesn't help. Just listing the target names here would work just as well and look less confusing.

hannes-steffenhagen-diffblue (approved):
Apparently this is not a required check and it seems to work, so LGTM

.github/workflows/pull-request-checks.yaml
NlightNFotis: I've marked it `fail_ci_if_error` so that less coverage (or an error) will cause the specific check to fail, marking it with a red X, so that it's behaviourally consistent with the previous codecov jobs we had.

However, it's not marked as a required check in Github, and as a result, a possible failure of this job doesn't block the merging of a PR.

CMakeLists.txt
tautschnig: @hannes-steffenhagen-diffblue Would it be worth doing a cleanup across the code base? Would you mind driving this as my CMake expertise is, well, not really existent?

PR #5815, opened 2021-02-11
Make componentt::{base,pretty}_name comments

tautschnig:
The only semantically relevant name information is the ID_name entry.
Two components should not be considered different when they only differ
in their base_name or pretty_name. Thus, use ID_C_base_name and
ID_C_pretty_name, respectively, to store these.

The goto binary version is incremented as goto binaries compiled before
this patch are incompatible with the changes introduced here.


src/goto-programs/write_goto_binary.h
```diff
@@ -12,7 +12,7 @@ Author: CM Wintersteiger
 #ifndef CPROVER_GOTO_PROGRAMS_WRITE_GOTO_BINARY_H
 #define CPROVER_GOTO_PROGRAMS_WRITE_GOTO_BINARY_H
 
-#define GOTO_BINARY_VERSION 5
+#define GOTO_BINARY_VERSION 6
```
martin-cs, line 15: Does this need a major or minor version number bump?

martin-cs (approved):
Changes are fine but a change to the format number feels like it is worth a major or minor version number change.

src/goto-programs/write_goto_binary.h
tautschnig: I don't know how many customers have use cases where they build goto binaries, store them, and (re-)analyse them at a later point in time. I know that's something I occasionally see, but I have no idea how common this is. So I'm with you that a version number change is desirable, but I'd lean towards minor-version-bump-is-sufficient. If others agree with this, then we could contemplate merging this PR just before one of the next regular version bumps (e.g., next Thursday).

hannes-steffenhagen-diffblue:
I concur with @martin-cs, changing the goto binary format should be a major version bump as it's a pretty clear example of a breaking change.

While hopefully this would not impact too many people, this gives us an opportunity to tell people "Here, this breaks this particular use case, watch out for that, if you're not doing that you should be fine". 

Of course in principle any change is potentially a breaking change for someone, but my view has been that if we _know_ something is a breaking change it should result in a major version bump.

tautschnig:
As side effect, this PR also sorts out the goto binary version bump that we will want anyhow with version 6 (and the breaking changes that we intend to merge as part of this).

kroening:
> As side effect, this PR also sorts out the goto binary version bump that we will want anyhow with version 6 (and the breaking changes that we intend to merge as part of this).

I haven't seen yet a convincing reason to do this -- I'd be convinced by the changes to the array type, but that needs more work.

peterschrammel (approved):

PR #5817, opened 2021-02-11
use a hash table in the expression formatter

kroening:
This replaces the linear-time if-then-else construct by a logarithmic hash
table.  The hash table could be extended at runtime.

### DIFF
--- a/src/util/format_expr.cpp
+++ b/src/util/format_expr.cpp
@@ -217,81 +217,145 @@ std::ostream &fallback_format_rec(std::ostream &os, const exprt &expr)
   return os;
 }
 
-// The below generates a string in a generic syntax
+class format_expr_configt
+{
+public:
+  format_expr_configt()
+  {
+    setup();
+  }
+
+  using formattert =
+    std::function<std::ostream &(std::ostream &, const exprt &)>;
+  using expr_mapt = std::unordered_map<irep_idt, formattert>;
+
+  expr_mapt expr_map;
+
+  /// find the formatter for a given expression
+  const formattert &find_formatter(const exprt &);
+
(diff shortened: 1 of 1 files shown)


src/util/format_expr.cpp
```diff
@@ -217,81 +217,145 @@ std::ostream &fallback_format_rec(std::ostream &os, const exprt &expr)
...
+  }
+
+  using formattert =
+    std::function<std::ostream &(std::ostream &, const exprt &)>;
+  using expr_mapt = std::unordered_map<irep_idt, formattert>;
+
+  expr_mapt expr_map;
+
+  /// find the formatter for a given expression
+  const formattert &find_formatter(const exprt &);
```
tautschnig, line 235: Shouldn't this be `const`?

src/util/format_expr.cpp
```diff
@@ -217,81 +217,145 @@ std::ostream &fallback_format_rec(std::ostream &os, const exprt &expr)
...
+
+private:
+  /// setup the expressions we can format
+  void setup();
```
tautschnig, line 239: Possibly a matter of taste: Is there value in having this extra method as opposed to just doing all work directly in the constructor?

tautschnig (approved):

src/util/format_expr.cpp
kroening: Not a whole lot, it may convey a bit of information to the reader.

src/util/format_expr.cpp
kroening: As-is now yes, but was contemplating to generate formatters on the fly.

PR #5821, opened 2021-02-11
goto-harness: create goto binary or C output

tautschnig:
It is not always necessary or desirable to produce C output. An ability
to produce goto binaries also avoids possible limitations of C and/or
dump-c (such as goto-harness creating initialisers for structs with bit
fields, which yield functions that take bit fields as parameters, which
isn't possible in C).

While at it, also clean up the help output to fix confusing language
about goto binaries when the output was actually C code.

### DIFF
--- a/regression/goto-harness/chain.sh
+++ b/regression/goto-harness/chain.sh
@@ -8,13 +8,20 @@ cbmc=$3
 is_windows=$4
 entry_point='generated_entry_function'
 
+shift 4
 name=${*:$#}
 name=${name%.c}
-args=${*:5:$#-5}
 
 input_c_file="${name}.c"
 input_goto_binary="${name}.gb"
-harness_c_file="${name}-harness.c"
+if [[ "$1" == "harness.gb" ]]; then
+  harness_file=$1
+  shift
+else
+  harness_file="${name}-harness.c"
+fi
+
+args=${*:1:$#-1}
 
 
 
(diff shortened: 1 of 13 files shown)


regression/goto-harness/chain.sh
```diff
@@ -8,13 +8,20 @@ cbmc=$3
...
 
+shift 4
 name=${*:$#}
 name=${name%.c}
-args=${*:5:$#-5}
 
 input_c_file="${name}.c"
 input_goto_binary="${name}.gb"
-harness_c_file="${name}-harness.c"
+if [[ "$1" == "harness.gb" ]]; then
```
martin-cs, line 17: One of these days we should really decide on the preferred file ending for goto-model binaries, `.gb`? `.goto`? Something else?

regression/goto-harness/chain.sh
```diff
@@ -25,16 +32,24 @@ else
...
 $cbmc --show-goto-functions "$input_goto_binary"
-$goto_harness "$input_goto_binary" "$harness_c_file" --harness-function-name $entry_point ${args}
-$cbmc --show-goto-functions "$harness_c_file"
-$cbmc --function $entry_point "$input_c_file" "$harness_c_file" \
+$goto_harness "$input_goto_binary" "$harness_file" --harness-function-name $entry_point ${args}
+$cbmc --show-goto-functions "$harness_file"
+if [[ "${harness_file}" == "harness.gb" ]];then
+  $cbmc --function $entry_point "$harness_file" \
+    --pointer-check `# because we want to see out of bounds errors` \
+    --unwind 11 `# with the way we set up arrays symex can't figure out loop bounds automatically` \
```
martin-cs, line 46: Having these hardcoded in `chain.sh` I find worrying ... but I don't see and immediate alternative that is not a lot of extra complexity.

martin-cs (approved):
Looks to be an improvement.

Unhelpful suggestion : should it only ever generate goto binaries?  If you want C then you can `goto-instrument --dump-c output.gb`.  That would make the functionality more orthogonal.  (But less user-friendly and would involve throwing away useful, working code, so, maybe not).

tautschnig:
> Unhelpful suggestion : should it only ever generate goto binaries? If you want C then you can `goto-instrument --dump-c output.gb`. That would make the functionality more orthogonal. (But less user-friendly and would involve throwing away useful, working code, so, maybe not).

Would probably be the better route forward, but we'd need to figure out a way of not breaking any customer's workflows. That seems hard to accomplish, I fear.

regression/goto-harness/chain.sh
tautschnig: Done: #5826

regression/goto-harness/chain.sh
tautschnig: Agreed, but I wanted to keep the changes to a minimum. The cleaner approach would be to include the output file name in each and every test of `goto-harness`.

martin-cs: Fair.

regression/goto-harness/chain.sh
martin-cs: Thanks.

regression/goto-harness/chain.sh
chrisr-diffblue: Its a little bit of a shame that the whole cbmc command line gets repeated here - but I guess the chain.sh is not likely to change too often, so the risk of adding options to only one of the two branches of the if is low.

src/goto-harness/goto_harness_parse_options.cpp
```diff
@@ -145,15 +146,23 @@ int goto_harness_parse_optionst::doit()
...
-  filter_goto_model(goto_model, goto_model_without_harness_symbols);
-  auto harness_out = std::ofstream{got_harness_config.out_file};
-  dump_c(
-    goto_model.goto_functions,
-    true,
-    true,
-    false,
-    namespacet{goto_model.get_symbol_table()},
-    harness_out);
+  if(!has_suffix(got_harness_config.out_file, ".c"))
```
chrisr-diffblue: Probably just my personal preference, but I'd switch the condition and branches, so thats is more of the form:
```
if(has_suffix(".c"))
{
/* stuff to dump-c */
}
else
{
/* Dump goto-binary */
}
```
As to me, that reads more 'natural' rather than having the semi-invisible negation operator in the condition :-)

chrisr-diffblue (approved):
Looks good, and I'm happy for this to go in - with a personal-preference suggestion that you are totally free to ignore.

src/goto-harness/goto_harness_parse_options.cpp
tautschnig: Good call, thanks, fixed!

PR #5823, opened 2021-02-12
Tests demonstrating pointer-encoding unsoundness

tautschnig:
The tests exercise the way the back-end encodes pointers. Upcoming
encoding changes will ensure that these can soundly be handled.

### DIFF
--- a/regression/cbmc/Pointer_Arithmetic18/main.c
+++ b/regression/cbmc/Pointer_Arithmetic18/main.c
@@ -1,0 +1,11 @@
+#define MB 0x00100000
+#define BASE (15 * MB)
+#define OFFSET (1 * MB)
+
+main()
+{
+  char *base = BASE;
+  int offset = OFFSET;
+  __CPROVER_assert(
+    &((char *)base)[offset] >= BASE + OFFSET, "no wrap-around expected");
+}
--- a/regression/cbmc/Pointer_comparison1/main.c
+++ b/regression/cbmc/Pointer_comparison1/main.c
@@ -1,0 +1,22 @@
+#include <stdlib.h>
+
+int main()
+{
+  if(sizeof(void *) != 8)
+    return 0;
+
+  char *p = malloc(1);
+  if(p == 0)
+    return 0;
+
+  if(
+    (unsigned long)p >
+    42) // unsoundly evaluates to true due to pointer encoding
+  {
+    return 0;
+  }
+
+  __CPROVER_assert(0, "");
+
+  return 0;
+}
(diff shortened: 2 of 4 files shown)


regression/cbmc/Pointer_Arithmetic18/main.c
```diff
@@ -0,0 +1,11 @@
...
+#define BASE (15 * MB)
+#define OFFSET (1 * MB)
+
+main()
+{
+  char *base = BASE;
+  int offset = OFFSET;
+  int n = 2;
+  __CPROVER_assert(&((char *)base)[offset] != 0, "no wrap-around expected");
+}
```
kroening, line 11: This one feels like it's a bit overfitted to the current encoding -- how about checking that the conversion to an integer doesn't get smaller?

regression/cbmc/Pointer_Arithmetic18/main.c
```diff
@@ -0,0 +1,11 @@
+#define MB 0x00100000
+#define BASE (15 * MB)
+#define OFFSET (1 * MB)
+
+main()
+{
+  char *base = BASE;
```
martin-cs, line 7: I'm a little concerned by this.  As I understand CBMC's memory model we do not support literal addresses.  I don't think we are required to do so by the standard either.

regression/cbmc/Pointer_comparison1/main.c
```diff
@@ -0,0 +1,22 @@
...
+  if(sizeof(void *) != 8)
+    return 0;
+
+  char *p = malloc(1);
+  if(p == 0)
+    return 0;
+
+  if(
+    (unsigned long)p >
+    42) // unsoundly evaluates to true due to pointer encoding
```
martin-cs, line 14: Is this unsound?  ISO-9899 allows conversion of pointers to integers but doesn't say what they will be.  Thinking about actual platforms, this kind of thing could be ASLR dependent!

martin-cs (commented):
Without knowing the bigger picture of what you are planning with pointer encodings it's a bit hard to say but... I'm not particularly comfortable with these.

regression/cbmc/Pointer_comparison1/main.c
tautschnig: Agreed, and I would thus argue that the assertion below ought to be considered reachable by a sound verification tool (i.e., there exists some system configuration under which `p` is less or equal to `42`).

regression/cbmc/Pointer_Arithmetic18/main.c
tautschnig: I'd say we half-support them... And I'd really like to see that fixed as well, might just happen along the way ;-)

regression/cbmc/Pointer_Arithmetic18/main.c
tautschnig: Done.

kroening (approved):

PR #5825, opened 2021-02-14
Vsd value set of pointers

jezhiggins:
This leads on from #5793, separating out value-sets of constants from value-sets of pointers.

Prior to this work, `value_set_abstract_objectt` did double duty for constants and pointers, contained a certain amount of runtime type checking to do the right thing, doing a decent job for values, but a poor job for pointers. Separating the two out isolates the constant and pointer specific behaviours, immediately increasing clarity and enabling some easy fixes for pointer dereferencing and writing through.

I could have gone further, as the pointer behaviour, while improved, is not complete, and there's still a certain amount of duplication that could be eliminated. However, I believe much of this will be addressed, and actually more easily addressed, in the upcoming working on multiple-dispatch/cross-representation evaluation and merge.


src/analyses/variable-sensitivity/abstract_pointer_object.cpp
```diff
@@ -2,23 +2,20 @@
 
  Module: analyses variable-sensitivity
 
- Author: Thomas Kiley, thomas.kiley@diffblue.com
+ Author: Jez Higgins, jez@jezuk.co.uk
```
martin-cs: Although the author fields are a bit meaningless in the age of `git praise` / `git blame`, we do prefer to have the name of the original author.

src/analyses/variable-sensitivity/abstract_pointer_object.h
```diff
@@ -2,84 +2,52 @@
 
  Module: analyses variable-sensitivity
 
- Author: Thomas Kiley, thomas.kiley@diffblue.com
```
martin-cs: Ditto.

src/analyses/variable-sensitivity/abstract_pointer_object.h
```diff
@@ -2,84 +2,52 @@
...
   /// \param environment: the environment in which the pointer is being
   ///                     created
   /// \param ns: the current namespace
-  explicit pointer_abstract_objectt(
+  explicit abstract_pointer_objectt(
     const exprt &expr,
     const abstract_environmentt &environment,
     const namespacet &ns);
 
-  /**
```
martin-cs, line 67: If the function stays then it would be good if the doxygen stays / gets updated.

src/analyses/variable-sensitivity/value_set_pointer_abstract_object.cpp
```diff
@@ -1,23 +1,256 @@
-
 /*******************************************************************\
 
- Module: analyses variable-sensitivity variable-sensitivity-value-set
+ Module: analyses variable-sensitivity
 
- Author: Diffblue Ltd.
```
martin-cs, line 6: This should remain.

martin-cs (approved):
Approve subject to minor corrections.

src/analyses/variable-sensitivity/abstract_pointer_object.cpp
jezhiggins: I have a feeling this might be something of a git artefact, but happy to make the changes.

martin-cs: If these are new files (that git has just decided are moved) then they should have you as the author.  If they are moved they should keep the original author.

jezhiggins: Honestly, I can't remember, and as you say not hugely important. However, since it obviously looked odd in the diff, I've made the changes.

PR #5827, opened 2021-02-15
fix distinct operator

polgreen:

Adds support for distinct operator in smt2 parser (which is syntactic sugar for a pairwise "not equal" predicate). Previously operator was parsed but later ignored

### DIFF
--- a/regression/smt2_solver/distinct2/distinct2.smt2
+++ b/regression/smt2_solver/distinct2/distinct2.smt2
@@ -1,0 +1,2 @@
+(assert (distinct (_ bv0 8) (_ bv1 8)(_ bv2 8) (_ bv1 8)))
+(check-sat)
--- a/src/solvers/smt2/smt2_parser.cpp
+++ b/src/solvers/smt2/smt2_parser.cpp
@@ -950,7 +950,23 @@ void smt2_parsert::setup_expressions()
 
   expressions["distinct"] = [this] {
     // pair-wise different constraint, multi-ary
-    return multi_ary("distinct", operands());
+    auto op = operands();
+    if(op.size() == 2)
+      return binary_predicate(ID_notequal, op);
+    else
+    {
+      std::vector<exprt> pairwise_constraints;
+      for(std::size_t i = 0; i < op.size(); i++)
+      {
+        for(std::size_t j = i; j < op.size(); j++)
+        {
+          if(i != j)
+            pairwise_constraints.push_back(
+              binary_exprt(op[i], ID_notequal, op[j], bool_typet()));
+        }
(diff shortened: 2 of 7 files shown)


regression/smt2_solver/distinct2/distinct2.smt2
```diff
@@ -0,0 +1,2 @@
+(assert (distinct (_ bv0 8) (_ bv1 8)(_ bv2 8) (_ bv1 8)))
```
martin-cs, line 1: `(_ bv1 8)` appears twice but the test case is marked `sat`?

src/solvers/smt2/smt2_parser.cpp
```diff
@@ -950,7 +950,22 @@ void smt2_parsert::setup_expressions()
...
-    return multi_ary("distinct", operands());
+    auto op = operands();
+    if(op.size() % 2 != 0)
+      throw error("distinct needs even number of operands");
+    if(op.size() == 2)
+      return binary_predicate(ID_notequal, op);
+    else
+    {
+      std::vector<exprt> pairwise_constraints;
+      for(std::size_t i = 0; i < op.size();)
```
martin-cs: I don't think this is the semantics of SMT-LIB `distinct`.  It is described here ( http://smtlib.cs.uiowa.edu/theories-Core.shtml ) as:

```
(par (A) (distinct A A Bool :pairwise))
```

And `:pairwise` is described on Page 34 of http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf as:

```
[pairwise annotation] (f   t1···tn) is syntactic sugar (recursively) for (and (f   t1 t2) ··· (f   t1 tn) (f   t2···tn)).
```

martin-cs (changes requested):
Sorry but I don't think this is the correct semantics.  I think that `distinct` is pair-wise inequality.

src/solvers/smt2/smt2_parser.cpp
```diff
@@ -950,7 +950,23 @@ void smt2_parsert::setup_expressions()
...
+    if(op.size() == 2)
+      return binary_predicate(ID_notequal, op);
+    else
+    {
+      std::vector<exprt> pairwise_constraints;
+      for(std::size_t i = 0; i < op.size(); i++)
+      {
+        for(std::size_t j = i; j < op.size(); j++)
+        {
+          if(i != j)
```
martin-cs, line 963: You could start at `j = i + 1` and skip this check if you wanted.

martin-cs (approved):
Much better -- thank you!

PR #5829, opened 2021-02-16
Do not rely on Clang's results to verify our GCC version handling

### LINKED ISSUE #5828: Failing tests on M1-Arm64 on develop
CBMC version: 5.23.0 (cbmc-5.21.3-exp-254-g268f8864e)
Operating system: macOS 11.0
Exact command line resulting in the issue: `cbmc/regression/ansi-c/gcc_version1 $ clang gcc-5.c`
What behaviour did you expect: Binary is compiled without issue.
What happened instead: Clang fails with error: `__float128 is not supported on this target`.

---

Running the regression test-suite under make on macOS Big Sur, on top of the new m1 platform is causing some of the regression tests to fail. Specifically, `regression/ansi-c/gcc-version1` and `regression/ansi-c/gcc-version2` are failing.

On closer inspection, invoking clang directly on the binary gives me this:

```sh
fotiskoutoulakis@Fotiss-Mac-mini gcc_version1 % clang gcc-5.c
gcc-5.c:10:1: error: __float128 is not supported on this target
__float128 f128;
^
```

Querying clang about version and target produces this:

```sh
fotiskoutoulakis@Fotiss-Mac-mini gcc_version1 % clang -v 
Apple clang version 12.0.0 (clang-1200.0.32.27)
Target: arm64-apple-darwin20.1.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

I have two questions/remarks:

* It seems like it is assumed that `__float128` is going to be supported on other platforms, but this looks like an unsafe assumption. Should we change the test/deactivate it on some platforms?
* It looks like a subset of the tests in `regression/ansi-c` are marked `gcc-only`. Do we need to run them under MacOS where the default compiler is going to be `clang/llvm`? If not we may need to make changes to our CI infrastructure, as we currently *do* run them under the `macOS-cmake` and `macOS-make` builds.

tautschnig:
We use __float128 to distinguish GCC versions, which is not a
distinguishing factor for Clang. (It's supported on several Clang
versions on x86, but not on Apple's M1/ARM.) This commit effectively
disables the gcc_version1 test on macOS (where "gcc" really is Clang).

Fixes: #5828

### DIFF
--- a/regression/ansi-c/gcc_version1/gcc-4.c
+++ b/regression/ansi-c/gcc_version1/gcc-4.c
@@ -1,4 +1,5 @@
-// None of these types should be defined when emulating gcc-4:
+// None of these types should be defined when emulating gcc-4, which clang on
+// macOS also emulates:
 
 typedef float _Float32;
 typedef double _Float32x;
--- a/regression/ansi-c/gcc_version1/gcc-5.c
+++ b/regression/ansi-c/gcc_version1/gcc-5.c
@@ -1,4 +1,6 @@
 // These types should *not* be provided when emulating gcc-5:
+
+#ifndef __clang__
 typedef float _Float32;
 typedef double _Float32x;
 typedef double _Float64;
@@ -8,3 +10,4 @@ typedef long double _Float128x;
 
 // But this type should:
 __float128 f128;
+#endif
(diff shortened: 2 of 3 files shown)


NlightNFotis (approved):
Tested on the M1 platform, and the tests are now passing. Thanks 👍🏻

PR #5830, opened 2021-02-16
bv_utilst: mark non-modifying members static

tautschnig:
Some members were marked as const already, but can really be marked as
static; others did not make any use of the (sole) "prop" member, and can
thus be marked static as well.

### DIFF
--- a/src/solvers/flattening/bv_utils.cpp
+++ b/src/solvers/flattening/bv_utils.cpp
@@ -75,7 +75,7 @@ bvt bv_utilst::extract_lsb(const bvt &a, std::size_t n)
   return result;
 }
 
-bvt bv_utilst::concatenate(const bvt &a, const bvt &b) const
+bvt bv_utilst::concatenate(const bvt &a, const bvt &b)
 {
   bvt result;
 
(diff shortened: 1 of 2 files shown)


martin-cs (approved):
Seems reasonable enough but I am left wondering why aren't there call sites that need modifying?

src/solvers/flattening/bv_utils.cpp
```diff
@@ -75,7 +75,7 @@ bvt bv_utilst::extract_lsb(const bvt &a, std::size_t n)
   return result;
 }
 
-bvt bv_utilst::concatenate(const bvt &a, const bvt &b) const
+bvt bv_utilst::concatenate(const bvt &a, const bvt &b)
 {
```
kroening, line 79: This one isn't `static`?

kroening (approved):

tautschnig:
> Seems reasonable enough but I am left wondering why aren't there call sites that need modifying?

I wouldn't have expected anything that _needs_ modifying, but it may enable const-ification of some of the call sites.

PR #5831, opened 2021-02-16
bv_pointerst: make all members return (optional) bvt

tautschnig:
Instead of having in-out non-const bvt references, return bvt by value
and rely on copy elision for performance. This simplifies the code in
that it avoids creating (and sometimes unnecessarily resizing) objects
that are then to be populated by a function.

For convert_address_of_rec use optionalt<bvt> to remove the true/false
error reporting.


martin-cs (approved):
Seems reasonable, although I wonder if we should go further and do this throughout the solver.  I suspect doing so will break patches against it though.

PR #5833, opened 2021-02-17
Bug fix using `format` function on `code_declt` containing initialization

thomasspriggs:
Bug fix using `format` function on `code_declt` containing initialization. Functionality was previously unimplemented, which resulted in part of the statement being missing from the formatted output.

- [x] N/A ~~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~~
- [x] N/A ~~My commit message includes data points confirming performance improvements (if claimed).~~

### DIFF
--- a/src/goto-programs/goto_program.h
+++ b/src/goto-programs/goto_program.h
@@ -199,13 +199,21 @@ public:
     const code_declt &get_decl() const
     {
       PRECONDITION(is_decl());
-      return to_code_decl(code);
+      const auto &decl = expr_checked_cast<code_declt>(code);
+      INVARIANT(
+        !decl.initial_value(),
+        "code_declt in goto program may not contain initialization.");
+      return decl;
     }
 
     /// Set the declaration for DECL
     void set_decl(code_declt c)
     {
       PRECONDITION(is_decl());
+      INVARIANT(
+        !c.initial_value(),
+        "Initialization must be separated from code_declt before adding to "
+        "goto_instructiont.");
       code = std::move(c);
     }
 
--- a/src/util/std_code.h
+++ b/src/util/std_code.h
@@ -422,6 +422,8 @@ public:
 
   /// Returns the initial value to which the declared variable is initialized,
   /// or empty in the case where no initialisation is included.
+  /// \note: Initial values may be present in the front end but they must be
+  ///   separated into a separate assignment when used in a `goto_instructiont`.
   optionalt<exprt> initial_value() const
   {
     if(operands().size() < 2)
@@ -431,6 +433,8 @@ public:
 
   /// Sets the value to which this declaration initializes the declared
   /// variable. Empty optional maybe passed to remove existing initialisation.
+  /// \note: Initial values may be present in the front end but they must be
+  ///   separated into a separate assignment when used in a `goto_instructiont`.
   void set_initial_value(optionalt<exprt> initial_value)
   {
     if(!initial_value)
--- a/src/util/std_code.h
+++ b/src/util/std_code.h
@@ -429,6 +429,25 @@ public:
     return {op1()};
   }
 
+  /// Sets the value to which this declaration initializes the declared
+  /// variable. Empty optional maybe passed to remove existing initialisation.
+  void set_initial_value(optionalt<exprt> initial_value)
+  {
+    if(!initial_value)
+    {
+      operands().resize(1);
+    }
+    else if(operands().size() < 2)
+    {
+      PRECONDITION(operands().size() == 1);
+      add_to_operands(std::move(*initial_value));
+    }
+    else
+    {
+      op1() = std::move(*initial_value);
+    }
+  }
+
   static void check(
     const codet &code,
     const validation_modet vm = validation_modet::INVARIANT)
--- a/src/util/std_code.h
+++ b/src/util/std_code.h
@@ -420,6 +420,15 @@ public:
     return symbol().get_identifier();
   }
 
+  /// Returns the initial value to which the declared variable is initialized,
+  /// or empty in the case where no initialisation is included.
+  optionalt<exprt> initial_value() const
+  {
+    if(operands().size() < 2)
+      return {};
+    return {op1()};
+  }
+
   static void check(
     const codet &code,
     const validation_modet vm = validation_modet::INVARIANT)
--- a/src/goto-instrument/dump_c.cpp
+++ b/src/goto-instrument/dump_c.cpp
@@ -650,19 +650,18 @@ void dump_ct::cleanup_decl(
   std::list<irep_idt> &local_static,
   std::list<irep_idt> &local_type_decls)
 {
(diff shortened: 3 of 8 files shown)


SaswatPadhi:
The weakest precondition computation here assumes no initial value for a `code_declt`:
https://github.com/diffblue/cbmc/blob/develop/src/goto-programs/wp.cpp#L224-L235

This function would also need an update -- it currently ignores the initial value, even if it's present. There might be more such cases elsewhere in the codebase?

src/util/std_code.h
```diff
@@ -420,6 +420,34 @@ class code_declt:public codet
...
+  optionalt<exprt> initial_value() const
+  {
+    if(operands().size() < 2)
+      return {};
+    return {op1()};
+  }
+
+  /// Sets the value to which this declaration initializes the declared
+  /// variable. Empty optional maybe passed to remove existing initialisation.
+  void set_initial_value(optionalt<exprt> initial_value)
```
hannes-steffenhagen-diffblue, line 438: If these can exist with optional `initial_value`s then I'd much prefer we had a constructor that took an initial value as the argument rather than having a separate setter.

hannes-steffenhagen-diffblue (approved):
I suppose I find it a bit surprising that we have initialisers in `code_declt`s to begin with.

src/util/std_code.h
thomasspriggs: All the examples I have found so far were mutating after construction. I could add the facility to set this via the constructor as well.

thomasspriggs:
> The weakest precondition computation here assumes no initial value for a `code_declt`:
> https://github.com/diffblue/cbmc/blob/develop/src/goto-programs/wp.cpp#L224-L235
> 
> This function would also need an update -- it currently ignores the initial value, even if it's present. There might be more such cases elsewhere in the codebase?

This PR doesn't make any changes to what data is present. I assume that function currently works due to the assignment/initialisation being separated out before the function is reached. I have tried adding a precondition in that function to initialisation not being in the declaration and all tests passed. So I may tack that change onto this PR, or add it as a separate follow-up PR.

martin-cs (commented):
This PR makes me concerned.  Not because of what it does as such but more because of the context.  I think you have uncovered one of the historic ambiguities over what, exactly, a well-formed `goto_programt` is.

I /guess/ what has happened is:

1. `DECL` instructions used to have an initialiser statement.
2. At some point it was decided to simplify things by having `DECL` be just declaration and initialisation being done with a following `ASSIGN`.
3. Most of the code was converted over to split up initialisation.  Some older, at-the-time untested, code was left.  Also the option was not removed from `goto_programt` for some reason (backwards compatibility, overlooked, forgotten?).

So, perhaps the question I would ask is not /does/ it have initialisers but /should/ it have initialisers.  Personally I would say no because:

- Quite a bit of the code doesn't support it, it would be a non-trivial effort to do so.
- AFAIK it is not currently widely used.
- Orthogonality and simplicity in `goto_programt` is good because it makes analysis code easier to write.

So, sorry @thomasspriggs for hijacking what seems like a good and simple tidy up but ... should we just remove this all together?

Some of the comments imply that people have found sections of the code that do / don't take these initialisers into account.  It might help the discussion if you could post links to these.

thomasspriggs:
> This PR makes me concerned. Not because of what it does as such but more because of the context. I think you have uncovered one of the historic ambiguities over what, exactly, a well-formed `goto_programt` is.
> 
> I /guess/ what has happened is:
> 
> 1. `DECL` instructions used to have an initialiser statement.
> 2. At some point it was decided to simplify things by having `DECL` be just declaration and initialisation being done with a following `ASSIGN`.
> 3. Most of the code was converted over to split up initialisation.  Some older, at-the-time untested, code was left.  Also the option was not removed from `goto_programt` for some reason (backwards compatibility, overlooked, forgotten?).
> 
> So, perhaps the question I would ask is not /does/ it have initialisers but /should/ it have initialisers. Personally I would say no because:
> 
> * Quite a bit of the code doesn't support it, it would be a non-trivial effort to do so.
> * AFAIK it is not currently widely used.
> * Orthogonality and simplicity in `goto_programt` is good because it makes analysis code easier to write.
> 

I agree that it would be good to be certain whether it is or is not valid to include this on a `code_declt` or not. My current assumption is that is valid for the parse tree in the symbol table to have such an initialisation in a declaration, but not for a declaration statement in a goto program to have it. Such a distinction would be tricky to enforce given that the same data structure is used in both contexts. I will investigate how the C parser currently handles this and reply again. I would suggest that it could be better for this to be well documented and marked as deprecated, than it would be for it to be left as an undocumented gotcha for future maintainers.

> So, sorry @thomasspriggs for hijacking what seems like a good and simple tidy up but ... should we just remove this all together?

My concern with trying to remove this altogether is that it is going to be difficult to be sure that we have found and removed all usages. `code_declt` has over 100 usages across the code base when I search for it. Any of these usages that exist could be accessing this initialisation directly through the operands. Making it potentially very time-consuming to review them all. 

> Some of the comments imply that people have found sections of the code that do / don't take these initialisers into account. It might help the discussion if you could post links to these.

The usages which I have looked at so far are either touched by this PR already, or mentioned in Saswat's comment.

tautschnig:
Just a quick thought: should we introduce something like `code_frontend_declt`?

thomasspriggs:
> Just a quick thought: should we introduce something like `code_frontend_declt`?

Or a `code_goto_declt`?

thomasspriggs:
@martin-cs I have investigated a bit further and found that declarations which initialise are produced by the C front end. I have added additional commits which add invariants/preconditions to disallow the declarations in goto-programs from containing initialisations as these should be removed by this point. Please can you re-review? If the clean-up which is in this PR is controversial then I can drop all but the first two commits from it. I do want to get the fix to `format` merged but I don't want to spend time expanding the scope of this PR to cover changes around using more separate data structures in the front end vs goto-programs.

tautschnig (approved):
I think this is a reasonable intermediate fix, but the long-term route should likely be one of having a separate `some_variant_of_declt` class that's to be used by the front-end.

src/goto-programs/goto_program.h
```diff
@@ -199,13 +199,21 @@ class goto_programt
     const code_declt &get_decl() const
     {
       PRECONDITION(is_decl());
-      return to_code_decl(code);
+      const auto &decl = expr_checked_cast<code_declt>(code);
+      INVARIANT(
+        !decl.initial_value(),
+        "code_declt in goto program may not contain initialization.");
+      return decl;
```
tautschnig, line 206: While the `INVARIANT` in `set_decl` is probably the right thing, this one may better be placed in `goto_programt::instructiont::validate`.

thomasspriggs: Given that this PR is now approved, I am going to merge as is and clean up this as a follow-up PR.

PR #5835, opened 2021-02-18
Do not use invalid_source_file_exceptiont without source location

tautschnig:
In preparation of tying invalid_source_file_exceptiont to an actual
source location, introduce an exception that is thrown when some invalid
input is encountered, but a precise source location is not known at the
time of throwing the exception.

### DIFF
--- a/jbmc/src/java_bytecode/java_bytecode_language.cpp
+++ b/jbmc/src/java_bytecode/java_bytecode_language.cpp
@@ -318,7 +318,7 @@ void java_bytecode_languaget::initialize_class_loader()
 
 static void throwMainClassLoadingError(const std::string &main_class)
 {
-  throw invalid_source_file_exceptiont(
+  throw system_exceptiont(
     "Error: Could not find or load main class " + main_class);
 }
 
@@ -392,8 +392,7 @@ bool java_bytecode_languaget::parse(
     std::ifstream jar_file(path);
     if(!jar_file.good())
     {
-      throw invalid_source_file_exceptiont(
-        "Error: Unable to access jarfile " + path);
+      throw system_exceptiont("Error: Unable to access jarfile " + path);
     }
 
     // build an object to potentially limit which classes are loaded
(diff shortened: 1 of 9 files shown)


martin-cs (approved):
Seems reasonable.

chrisr-diffblue (approved):
Thanks - thats a nice cleanup improvement. As an aside, I can't help thinking an additional intermediate class in the exception hierarchy - something like a `cprover_exception_with_descriptiont` to cover what appears to be the very common pattern of "exception is constructed from a std::string argument, and what() just returns that string" might make `exception_utils.{h,cpp}` somewhat more succinct... one for another day though!

tautschnig:
> Thanks - thats a nice cleanup improvement. As an aside, I can't help thinking an additional intermediate class in the exception hierarchy - something like a `cprover_exception_with_descriptiont` to cover what appears to be the very common pattern of "exception is constructed from a std::string argument, and what() just returns that string" might make `exception_utils.{h,cpp}` somewhat more succinct... one for another day though!

Good call! I'm about to add a commit to #5837 to sort this out. I did, however, opt to make `cprover_exception_baset` that very class: all inheriting classes want that description.

tautschnig:
> > Thanks - thats a nice cleanup improvement. As an aside, I can't help thinking an additional intermediate class in the exception hierarchy - something like a `cprover_exception_with_descriptiont` to cover what appears to be the very common pattern of "exception is constructed from a std::string argument, and what() just returns that string" might make `exception_utils.{h,cpp}` somewhat more succinct... one for another day though!
> 
> Good call! I'm about to add a commit to #5837 to sort this out. I did, however, opt to make `cprover_exception_baset` that very class: all inheriting classes want that description.

... and now also made this a PR of its own in #6608.

chrisr-diffblue:
> > > Thanks - thats a nice cleanup improvement. As an aside, I can't help thinking an additional intermediate class in the exception hierarchy - something like a `cprover_exception_with_descriptiont` to cover what appears to be the very common pattern of "exception is constructed from a std::string argument, and what() just returns that string" might make `exception_utils.{h,cpp}` somewhat more succinct... one for another day though!
> 
> > 
> 
> > Good call! I'm about to add a commit to #5837 to sort this out. I did, however, opt to make `cprover_exception_baset` that very class: all inheriting classes want that description.
> 
> 
> 
> ... and now also made this a PR of its own in #6608.

You're too good to us... :-)

chris-ryder (approved):

peterschrammel (approved):

PR #5837, opened 2021-02-18
Add source location to invalid_source_file_exceptiont

tautschnig:
This enables proper use in the C front-end in a way that is compatible
with the compiler-specific message handlers in goto-cc: those message
handlers generate source-location output that mimics that of the
compiler being emulated.

As future work it is thus possible to remove all "throw 0" instances
from the C front-end.

### DIFF
--- a/src/ansi-c/literals/convert_character_literal.cpp
+++ b/src/ansi-c/literals/convert_character_literal.cpp
@@ -20,7 +20,8 @@ Author: Daniel Kroening, kroening@kroening.com
 
 exprt convert_character_literal(
   const std::string &src,
-  bool force_integer_type)
+  bool force_integer_type,
+  const source_locationt &source_location)
 {
   assert(src.size()>=2);
 
@@ -60,8 +61,10 @@ exprt convert_character_literal(
       result=from_integer(x, type);
     }
     else
-      throw "wide literals with "+std::to_string(value.size())+
-            " characters are not supported";
+      throw invalid_source_file_exceptiont{
+        "wide literals with " + std::to_string(value.size()) +
+          " characters are not supported",
+        source_location};
   }
   else
   {
@@ -93,9 +96,12 @@ exprt convert_character_literal(
       result=from_integer(x, signed_int_type());
     }
     else
-      throw "literals with "+std::to_string(value.size())+
-            " characters are not supported";
+      throw invalid_source_file_exceptiont{
+        "literals with " + std::to_string(value.size()) +
+          " characters are not supported",
+        source_location};
   }
 
+  result.add_source_location() = source_location;
   return result;
 }
--- a/src/util/exception_utils.h
+++ b/src/util/exception_utils.h
@@ -168,10 +168,26 @@ public:
 /// Thrown when we can't handle something in an input source file.
 /// For example, if we get C source code that is not syntactically valid
 /// or that has type errors.
-class invalid_source_file_exceptiont : public cprover_exception_baset
+class invalid_source_file_exceptiont : public invalid_input_exceptiont
 {
 public:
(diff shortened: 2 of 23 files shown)


src/ansi-c/literals/convert_character_literal.cpp
```diff
@@ -32,13 +33,13 @@ exprt convert_character_literal(
 
     std::basic_string<unsigned int> value=
       unescape_wide_string(std::string(src, 2, src.size()-3));
+    // the parser rejects empty character constants
+    CHECK_RETURN(!value.empty());
```
chrisr-diffblue: Nice :-)

src/ansi-c/literals/convert_character_literal.cpp
```diff
@@ -92,9 +95,12 @@ exprt convert_character_literal(
...
     else
-      throw "literals with "+std::to_string(value.size())+
-            " characters are not supported";
+      throw invalid_source_file_exceptiont{"literals with " +
+                                             std::to_string(value.size()) +
+                                             " characters are not supported",
+                                           source_location};
   }
 
+  result.add_source_location() = source_location;
```
chrisr-diffblue, line 105: Also a nice catch!

src/goto-programs/initialize_goto_model.cpp
```diff
@@ -33,13 +33,12 @@ Author: Daniel Kroening, kroening@kroening.com
...
 static bool generate_entry_point_for_function(
-  const irep_idt &entry_function_name,
   const optionst &options,
-  goto_modelt &goto_model,
+  symbol_tablet &symbol_table,
   message_handlert &message_handler)
 {
-  auto const entry_function_sym =
-    goto_model.symbol_table.lookup(entry_function_name);
+  const irep_idt &entry_function_name = options.get_option("function");
```
chrisr-diffblue: Since this is a (slight) change in semantics for this function, such that it now _requires_ the `function` option to be set (previously the function name was passed in as an argument) - is it worth adding a CHECK_RETURN here?

src/util/exception_utils.h
```diff
@@ -164,17 +164,42 @@ class analysis_exceptiont : public cprover_exception_baset
...
+
+protected:
+  std::string m_reason;
+};
+
 /// Thrown when we can't handle something in an input source file.
 /// For example, if we get C source code that is not syntactically valid
 /// or that has type errors.
-class invalid_source_file_exceptiont : public cprover_exception_baset
+class invalid_source_file_exceptiont : public invalid_input_exceptiont
```
chrisr-diffblue, line 171: 👍

chrisr-diffblue (approved):
That looks great to me, and is what we should have done when we first added the structured exceptions. Some nice code cleanups/refactoring along the way too.

src/goto-programs/initialize_goto_model.cpp
tautschnig: Good point! I've added the suggested `CHECK_RETURN`.

chris-ryder (approved):
Previously reviewed and approved with my Diffblue account, re-approving with my non-Diffblue account to keep the CODEOWNER review.

martin-cs (approved):
Yes.  I don't think this breaks any scripting.

src/util/typecheck.cpp
```diff
@@ -37,5 +38,11 @@ bool typecheckt::typecheck_main()
     error() << e << eom;
   }
 
+  catch(const invalid_source_file_exceptiont &e)
```
peterschrammel, line 41: 🎉

peterschrammel (approved):

PR #5838, opened 2021-02-18
Fix pointer subtraction semantics

tautschnig:
(Please review commit-by-commit.)

Subtracting pointers over the same object is subtraction of the offset
(divided by the element size). Subtracting pointers over different
objects should yield a non-deterministic result.

### DIFF
--- a/regression/cbmc/Pointer_difference2/main.c
+++ b/regression/cbmc/Pointer_difference2/main.c
@@ -1,9 +1,31 @@
 int main()
 {
-  int array[2];
+  int array[4];
   int other_array[2];
 
+  __CPROVER_assert(&array[0] - &array[2] == -2, "correct");
+
   int diff = array - other_array;
+  _Bool nondet;
+  if(nondet)
+    __CPROVER_assert(diff != 42, "undefined behavior");
+  else
+    __CPROVER_assert(diff == 42, "undefined behavior");
+
+  __CPROVER_assert(
+    ((char *)(&array[3])) - ((char *)(&array[1])) == 2 * sizeof(int),
+    "casting works");
+
+  int *p = &array[3];
+  ++p;
+  __CPROVER_assert(p - &array[0] == 4, "end plus one works");
+  __CPROVER_assert(p - &array[0] != 3, "end plus one works");
+  ++p;
+  _Bool nondet_branch;
+  if(nondet_branch)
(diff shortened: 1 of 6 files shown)


kroening:
While at it, if both pointers are null, the result should also be nondeterministic.

tautschnig:
> While at it, if both pointers are null, the result should also be nondeterministic.

More generally, I believe this should be the case whenever pointers aren't valid, unless just one byte (or element?) beyond the object?

regression/cbmc/Pointer_difference2/main.c
```diff
@@ -0,0 +1,16 @@
+int main()
+{
+  int array[4];
+  int other_array[2];
+
+  __CPROVER_assert(&array[0] - &array[2] == -2, "correct");
+
+  int diff = array - other_array;
+  _Bool nondet;
+  if(nondet)
```
martin-cs, line 10: I like this test for non-determinism.

regression/cbmc/Pointer_difference2/main.c
```diff
@@ -0,0 +1,16 @@
+int main()
+{
+  int array[4];
+  int other_array[2];
+
+  __CPROVER_assert(&array[0] - &array[2] == -2, "correct");
+
```
martin-cs, line 7: Would I be awful to request...

`__CPROVER_assert(((char *)(&array[3])) - ((char *)(&array[1])) == 2*sizeof(int), "casting works");`

and

```
int *p = &array[3];
++p;
__CPROVER_assert(p - &array[0] == 4, "end plus one works");
__CPROVER_assert(p - &array[0] != 3, "end plus one works");
++p;
_Bool nondet_branch;
if (nondet_branch)
  __CPROVER_assert(p - &array[0] == 5, "end plus 2 is nondet");
else
  __CPROVER_assert(p - &array[0] != 5, "end plus 2 is nondet");
```

martin-cs (approved):
If you can add the tests it would be nice.

tautschnig:
> > While at it, if both pointers are null, the result should also be nondeterministic.
> 
> More generally, I believe this should be the case whenever pointers aren't valid, unless just one byte (or element?) beyond the object?

That's now covered, which required fixing one of our regression tests...

regression/cbmc/Pointer_difference2/main.c
tautschnig: Done, thanks!

PR #5844, opened 2021-02-19
Perform pointer validity checks when doing pointer arithmetic

### LINKED ISSUE #5426: Confusing results with --pointer-overflow-check
CBMC version: release 5.12
Operating system:  x86_64-linux (Ubuntu 18.04.4)
Exact command line resulting in the issue: `cbmc test.c --pointer-overflow-check`
What behaviour did you expect: VERIFICATION FAILED
What happened instead: VERIFICATION SUCCESSFUL

`test.c`:
```c
1  #include <stdlib.h>
2  
3
4  int main()
5  {
6      int *p = malloc(sizeof(int)*5);
7      int *p2 = p + 10;    // undefined behavior for indexing out of bounds
8      int *p3 = p - 10;    // undefined behavior for indexing out of bounds
9  
10     int arr[5];
11     int *p4 = arr + 10;  // undefined behavior for indexing out of bounds
12     int *p5 = arr - 10;  // undefined behavior for indexing out of bounds
13     return 0;
14 }                                                                                                                                                           
```

Greetings,

I am confused that CBMC reports no errors on the code above: The C standard specifies that incrementing a pointer to an array
further away than the `sizeof(array)+1` is undefined behavior [1].
The value of the pointer `p` returned by `malloc` could very well be `MAX_POINTER_ADDRESS - 7 * sizeof(int)`, so adding 10 to that value *should* trigger a pointer overflow error.

Am I making a mistake here, or is the tool not doing what I am expecting?

-----
```
** Results:
test.c function main                                                
[main.overflow.1] line 7 pointer arithmetic overflow on + in p + (signed long int)10: SUCCESS                                           
[main.overflow.2] line 8 pointer arithmetic overflow on - in p - (signed long int)10: SUCCESS                                                                                                    
[main.overflow.3] line 11 pointer arithmetic overflow on + in arr + (signed long int)10: SUCCESS
[main.overflow.4] line 12 pointer arithmetic overflow on - in arr - (signed long int)10: SUCCESS
** 0 of 4 failed (1 iterations)                                     
VERIFICATION SUCCESSFUL             
```

-----

[1] ISO/IEC 9899:1999 - Programming languages — C - ISO - §6.5.6 Additive operators [expr.add] para. 8, accessed 19.07.2020 at http://www.dii.uchile.cl/~daespino/files/Iso_C_1999_definition.pdf

Exact citation:

*When an expression that has integer type is added to or subtracted from a pointer, the result has the type of the pointer operand. If the pointer operand points to an element of an array object, and the array is large enough, the result points to an element offset from the original element such that the difference of the subscripts of the resulting and original array elements equals the integer expression. In other words, if the expression P points to the i-th element of an array object, the expressions (P)+N (equivalently, N+(P)) and (P)-N (where N has the value n) point to, respectively, the i+n-th and i−n-th elements of
the array object, provided they exist. Moreover, if the expression P points to the last element of an array object, the expression (P)+1 points one past the last element of the array object, and if the expression Q points one past the last element of an array object, the expression (Q)-1 points to the last element of the array object. If both the pointer operand and the result point to elements of the same array object, or one past the last element of the array object, the evaluation shall not produce an overflow; ***otherwise, the behavior is undefined***. If the result points one past the last element of the array object, it shall not be used as the operand of a unary \* operator that is evaluated.*

Edit: changed citation to C99 standard

tautschnig:
Arithmetic over pointers requires that they point to valid objects (or
one past the end of an object).

The test uncovered two further problems: 1) there was a typo in
subtraction handling in bv_pointerst; 2) redundant assertions are
removed, even when they refer to different expressions.

Fixes: #5426

### DIFF
--- a/regression/cbmc/pointer-overflow3/test.desc
+++ b/regression/cbmc/pointer-overflow3/test.desc
@@ -1,0 +1,16 @@
+CORE broken-smt-backend
+main.c
+--pointer-overflow-check --no-simplify
+^\[main.pointer_arithmetic.\d+\] line 6 pointer arithmetic: pointer outside dynamic object bounds in p \+ \(signed (long (long )?)?int\)10: FAILURE
+^\[main.pointer_arithmetic.\d+\] line 7 pointer arithmetic: pointer outside dynamic object bounds in p - \(signed (long (long )?)?int\)10: FAILURE
+^\[main.pointer_arithmetic.\d+\] line 10 pointer arithmetic: pointer outside object bounds in arr \+ \(signed (long (long )?)?int\)10: FAILURE
+^\[main.pointer_arithmetic.\d+\] line 11 pointer arithmetic: pointer outside object bounds in arr - \(signed (long (long )?)?int\)10: FAILURE
+^\*\* 4 of \d+ failed
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
+Invariant check failed
+--
+Uses --no-simplify to avoid removing repeated ASSERT FALSE statements.
--- a/regression/cbmc/pointer-overflow1/main.c
+++ b/regression/cbmc/pointer-overflow1/main.c
@@ -11,8 +11,8 @@ int main()
   p2 = p - other1;
   p2 = p - other2 - other1;
 
-  p2 = p + sizeof(int);
-  p2 = p + sizeof(int) + sizeof(int);
+  p2 = (char *)p + sizeof(int);
+  p2 = (char *)p + sizeof(int) + sizeof(int);
 
   return 0;
 }
--- a/regression/cbmc/pointer-overflow1/test.desc
+++ b/regression/cbmc/pointer-overflow1/test.desc
@@ -3,9 +3,12 @@ main.c
 --pointer-overflow-check --unsigned-overflow-check
 ^EXIT=10$
 ^SIGNAL=0$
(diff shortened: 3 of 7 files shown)


regression/cbmc/pointer-overflow3/test.desc
```diff
@@ -0,0 +1,16 @@
+CORE
+main.c
+--pointer-overflow-check --no-simplify
```
martin-cs, line 3: At some point are we stretching the intuitive interpretation of 'overflow' too much?

martin-cs (approved):
Seems pretty good.

regression/cbmc/pointer-overflow3/test.desc
tautschnig: I don't mind very much whether we should cover these under `--pointer-check` or `--pointer-overflow-check`. My preference towards `--pointer-overflow-check` stems from the following sentence in C11, 6.5.6 Additive operators: "If both the pointer operand and the result point to elements of the same array object, or one past the last element of the array object, the evaluation shall not produce an overflow; otherwise, the behavior is undefined." My thoughts were: if the arithmetic takes it out-of-bounds, then we could reasonably treat it as an "overflow." (If our pointer encoding were sound, we wouldn't even need any extra checks here: the arithmetic overflow already being covered would do the trick, as we have to assume that an object could be placed right at the boundary of (virtual) memory.)

martin-cs: Fair.

PR #5845, opened 2021-02-19
goto_checkt: keep same-assertion for different expressions

tautschnig:
~[Only the second commit is relevant, the first one is #5844 and only required to have the necessary test in place.]~

Purging redundant assertions results in surprising user output as it would seem that certain properties aren't being checked.

### DIFF
--- a/regression/cbmc/Pointer_byte_extract5/test.desc
+++ b/regression/cbmc/Pointer_byte_extract5/test.desc
@@ -4,7 +4,7 @@ main.i
 ^EXIT=10$
 ^SIGNAL=0$
 array\.List dynamic object upper bound in p->List\[2\]: FAILURE
-\*\* 1 of 14 failed
+\*\* 1 of \d+ failed
 --
 ^warning: ignoring
 --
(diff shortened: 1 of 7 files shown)


martin-cs (approved):
I'm going with a tentative 'yes'.  Assorted thoughts:

- When doing work on how to QA / evaluate / certify program analysis tools, I came to the conclusion that omitting VCs was actually quite a bit worse than simply getting them wrong.  If you have a VC and a status then you have a reasonable chance of being able to check that it is correct; there are a bunch of techniques.  If you don't have the VC there are many fewer ways of finding that it is missing.  As such I am suspicious of anything that removes VCs, even the ones that are "obviously" safe or redundant.
- I am working with a colleague on tracking down missing VCs.  This is a much more tedious and harder process than it should be.  Let's not have concealed cracks where VCs can just vanish.
- If a VC is trivial then we should still report it, we just need to make sure that it is minimal overhead.

tautschnig:
@martin-cs Thank you very much for your elaboration, I fully concur. It seems a further fix is in order: d974aa38 introduced a "retain-trivial" option, but this option was never made available in any command-line parser...

tautschnig:
> @martin-cs Thank you very much for your elaboration, I fully concur. It seems a further fix is in order: [d974aa3](https://github.com/diffblue/cbmc/commit/d974aa3897271d8226f1cd82bb3d279ac665f4a8) introduced a "retain-trivial" option, but this option was never made available in any command-line parser...

Such a fix is now included as well.

PR #5848, opened 2021-02-19
C library/str(n)cat: do not write spurious 0 byte

### LINKED ISSUE #3200: Buffer overflow in strcat
Hi,

I am trying to use CBMC to verify generic properties (like pointer-check) on
Openssl. I am on Ubuntu 16 and I use a x86_64 architecture.

I first compiled Openssl using goto-cc. It worked like a charm.
Now for the analysis I am using:

```sh
    cbmc --object-bits 16 --unwind 30 --pointer-check openssl
```

I obtain the following error trace:

> Violated property:
>   file <builtin-library-__builtin___strcat_chk> line 36 function __builtin___strcat_chk
>   dereference failure: pointer outside dynamic object bounds in dst[(signed long int)i]
>   POINTER_OFFSET(dst) + (signed long int)i >= 0l && __CPROVER_malloc_size >= 1ul + (unsigned long int)POINTER_OFFSET(dst) + (unsigned long int)(signed long int)i || !(POINTER_OBJECT(dst) == POINTER_OBJECT(__CPROVER_malloc_object))

After some debugging I discovered that the error was not in Openssl but in
CBMC's implementation of  `__builtin___strcat_chk`. The bug is also in CBMC's
implementation of strcat. Here is the faulty line:

https://github.com/diffblue/cbmc/blob/8b5b51d91bb6de1b7e694f97c6836099ec9f6320/src/ansi-c/library/string.c#L279

The instruction writes past the end of the dst string by 1. Down below is a
C file reproducing the bug.

As a correction I think the faulty line should simply be removed.

Best Regards,
Tristan Ninet

```C
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <assert.h>

void cbmcstrcat(char *dst, const char *src)
{
    size_t i;
    size_t j;
    char ch;

    i = 0;
    while (dst[i] != 0) i++;
    j = 0;
    ch = 1;
    assert(dst[i] == 0);
    for (; ch != (char)0; ++i, ++j)
    {
        ch = src[j];
        dst[i] = ch;
    }
    /* dst[i] = '\0'; */
    /*
     * Above instruction causes a buffer overflow. The instruction should
     * simply be removed as dst is already zero-terminated:
     */
    assert(dst[i-1] == '\0');
}

{
    const char *src;
    size_t len;
    char dst[3];

    src = "b";
    strcpy(dst, "a");
    /* strcat(dst, src); */
    cbmcstrcat(dst, src);

    printf("%x",*dst);
    printf(" %x",*(dst+1));
    printf(" %x",*(dst+2));
    printf(" %x",*(dst+3));
    /*
     * When using cbmcstrcat the last printed char is always zero. This shows
     * that cbmcstrcat is writing a zero at dst[3], which is a buffer overflow.
     */
    return 0;
}
```

tautschnig:
In strcat, the construction of the loop already ensures that the 0 byte
is written. In strncat, the 0 byte should only be written when the loop
terminated early.

Also, clean up regression tests of str(n)cat.

Fixes: #3200

### DIFF
--- a/regression/cbmc-library/strcat-01/main.c
+++ b/regression/cbmc-library/strcat-01/main.c
@@ -10,19 +10,5 @@ int main()
   assert(A1[3] == 'd');
   assert(strlen(A1) == 4);
 
-  char A2[5] = {'a', 'b', '\0'};
-  char B2[3] = {'c', 'd', '\0'};
-
-  strncat(A2, B2, 2);
-  assert(A2[3] == 'd');
-  assert(strlen(A2) == 4);
-
-  char A3[5] = {'a', 'b', '\0'};
-  char B3[3] = {'c', 'd', '\0'};
-
-  strncat(A3, B3, 1);
-  assert(A3[3] == '\0');
-  assert(strlen(A3) == 4); // expected to fail
-
   return 0;
 }
--- a/regression/cbmc-library/strcat-01/test.desc
+++ b/regression/cbmc-library/strcat-01/test.desc
@@ -1,10 +1,8 @@
 CORE
 main.c
---unwind 10
-^EXIT=10$
+--unwind 10 --pointer-check --bounds-check
+^VERIFICATION SUCCESSFUL$
+^EXIT=0$
 ^SIGNAL=0$
-^VERIFICATION FAILED$
-\[main.assertion.6\] .* assertion strlen\(A3\) == 4: FAILURE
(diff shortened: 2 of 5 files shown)


martin-cs (approved):

kroening (approved):

tautschnig:
Merging as the Xen failure is being dealt with in #5926.

PR #5849, opened 2021-02-19
Pointer arithmetic check: do not use pointer-to-int conversion

tautschnig:
We do not currently have a sound pointer-to-int conversion, and thus
arithmetic involving pointers and integers does not translate to
arithmetic over the integer representation of a pointer.

With the recently added bounds checks on pointers involved in arithmetic
we have a more reliable way of testing for overflow in pointer
arithmetic.

### DIFF
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -1177,22 +1177,6 @@ void goto_checkt::pointer_overflow_check(
     expr.operands().size() == 2,
     "pointer arithmetic expected to have exactly 2 operands");
 
-  // check for address space overflow by checking for overflow on integers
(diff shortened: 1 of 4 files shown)


martin-cs (approved):

src/analyses/goto_check.cpp
```diff
@@ -1177,22 +1177,6 @@ void goto_checkt::pointer_overflow_check(
     expr.operands().size() == 2,
     "pointer arithmetic expected to have exactly 2 operands");
 
-  // check for address space overflow by checking for overflow on integers
```
martin-cs, line 1180: :grimacing:

martin-cs (approved):

PR #5854, opened 2021-02-22
Remove arch package building script that is no longer maintained.

NlightNFotis:
Last time the file was maintained was 4 years ago. It no longer
makes sense for us to keep it around, as no one maintains it, and
for anyone that wants to use cbmc under Arch Linux, there are
packages in AUR that are better maintained than this one.


thomasspriggs (approved):
Makes sense to me. If there is a desire to keep arch package(s) in repository long term, then they would need to be tested in CI.

NlightNFotis:
In my opinion, we shouldn't be the ones maintaining packages for various distributions. 

It's the distributions' responsibility to package CBMC if they want/need it. Otherwise our packages get bit-rotten (like this one that's getting removed) and it increases our load to monitor various distributions for changes and update our build scripts/packages to facilitate them.

In terms of this particular package building script, there's already a more up to date one in AUR https://aur.archlinux.org/packages/cbmc/

tautschnig (approved):
For traceability, can I ask that the commit message initially be generated using `git revert 735c152ce8`, and then adding the commit message body that you've got right now? This is a clean revert of 735c152ce8, and we should document that.

PR #5855, opened 2021-02-22
[CI] Add a make + windows github action job.

NlightNFotis:
This is a draft (for now) PR that adds a `windows` * `make` Github Actions
job. This is using the already existing `cmake` one as a template.

### DIFF
--- a/regression/cbmc-cpp/Makefile
+++ b/regression/cbmc-cpp/Makefile
@@ -4,7 +4,7 @@ include ../../src/config.inc
 include ../../src/common
 
 ifeq ($(BUILD_ENV_),MSVC)
-	excluded_tests = -X gcc-only
+	excluded_tests = -X gcc-only -X winbug
 else
 	# In MacOS, a change in the assert.h header file
 	# is causing template errors when exercising the
--- a/regression/cpp/Makefile
+++ b/regression/cpp/Makefile
@@ -10,7 +10,7 @@ else
 endif
 
 ifeq ($(BUILD_ENV_),MSVC)
-	excluded_tests = -X gcc-only
+	excluded_tests = -X gcc-only -X winbug
 else
 	# In MacOS, a change in the assert.h header file
 	# is causing template errors when exercising the
--- a/src/common
+++ b/src/common
@@ -146,11 +146,11 @@ ifeq ($(origin CXX),default)
   CXX = cl
 endif
 ifeq ($(origin YACC),default)
-  YACC   = bison -y
+  YACC   = win_bison -y
 endif
   YFLAGS ?= -v
 ifeq ($(origin LEX),default)
-  LEX    = flex
+  LEX    = win_flex
 endif
 
 
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -277,6 +277,59 @@ jobs:
           Set-Location build
           ctest -V -L CORE -C Release . -j2
 
+  check-vs-2019-make-build-and-test:
+    runs-on: windows-2019
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Fetch dependencies
+        run: |
+          choco install winflexbison3 strawberryperl
+          nuget install clcache -OutputDirectory "c:\tools" -ExcludeVersion -Version 4.1.0
+          echo "c:\tools\clcache\clcache-4.1.0" >> $env:GITHUB_PATH
+          echo "c:\ProgramData\chocolatey\bin" >> $env:GITHUB_PATH
+          echo "c:\Strawberry\" >> $env:GITHUB_PATH
+      - name: Setup MSBuild
+        uses: microsoft/setup-msbuild@v1.0.2
+      - name: Initialise Developer Command Line
+        uses: ilammy/msvc-dev-cmd@v1
+      - name: Prepare ccache
+        uses: actions/cache@v2
+        with:
+          path: .ccache
+          key: ${{ runner.os }}-msbuild-make-${{ github.ref }}-${{ github.sha }}-PR
+          restore-keys: |
+            ${{ runner.os }}-msbuild-make-${{ github.ref }}
+            ${{ runner.os }}-msbuild-make
+      - name: ccache environment
+        run: |
+          echo "CLCACHE_BASEDIR=$((Get-Item -Path '.\').FullName)" >> $env:GITHUB_ENV
+          echo "CLCACHE_DIR=$pwd\.ccache" >> $env:GITHUB_ENV
+      - name: Download minisat with make
+        run: make -C src minisat2-download
+      - name: Build CBMC with make
+        run: make CXX=clcache BUILD_ENV=MSVC -j2 -C src
+      - name: Build unit tests with make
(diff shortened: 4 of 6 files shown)


regression/cbmc-cpp/Makefile
```diff
@@ -4,7 +4,7 @@ include ../../src/config.inc
 include ../../src/common
 
 ifeq ($(BUILD_ENV_),MSVC)
-	excluded_tests = -X gcc-only
+	excluded_tests = -X gcc-only - X winbug
```
TGWDB: Extra space in "- X"?

src/common
```diff
@@ -146,11 +146,11 @@ ifeq ($(origin CXX),default)
   CXX = cl
 endif
 ifeq ($(origin YACC),default)
-  YACC   = bison -y
+  YACC   = win_bison -y
```
thomasspriggs, line 149: :pick: I am guessing this is required due to the names used by the versions of flex/bison we are using in the new job. But can you add the rationale to the commit message body so A) I don't have to guess B) Anyone who starts seeing local failures due to using different versions, knows there was a reason for doing this. Not just to ruin their day ;)

.gitattributes
```diff
@@ -7,5 +7,5 @@
 *.html text
 *.css text
 *.inc text
-test.desc text
+*.desc text eol=lf
```
thomasspriggs: :pick: Can you add the explanation about the line endings causing problems with `test.pl` to the commit message body please?

regression/cbmc-cpp/Makefile
NlightNFotis: Good catch!

regression/cpp/Makefile
```diff
@@ -10,7 +10,7 @@ else
 endif
 
 ifeq ($(BUILD_ENV_),MSVC)
-	excluded_tests = -X gcc-only
+	excluded_tests = -X gcc-only -X winbug
```
thomasspriggs, line 13: I am kinda surprised this wasn't failing on the old codebuild job. Obviously not a blocking comment, but I wonder why that was passing.

thomasspriggs (commented):
One general nit pick here is that your commit messages titles should not end in a full stop and your commit message bodies should explain why you are making the change. See - [Chis Beams full stop](https://chris.beams.io/posts/git-commit/#end) and [Chris Beams why not how](https://chris.beams.io/posts/git-commit/#why-not-how).

Nit picks aside, congratulations on finally getting this working.

src/common
kroening: May I suggest to simply override this variable when calling make. This way, local builds won't be broken.

regression/cpp/Makefile
NlightNFotis: I have had a thought about this, and I've arrived at two theories regarding this, but I can't test them now:

1) The `makefiles` were misconfigured before, not including whole folders from `regression/` so it might just be that some tests were skipped by accident.
2) Even though the `_BUILD_ENV` was also specified to be `MSVC`, it ended up being overridden to `Cygwin` because of the build environment, which was calling `g++` which does not throw compiler errors for these tests (`cl.exe` does though, which is what `clcache` is invoking).

src/common
NlightNFotis: It's worth noting here that when someone installs `winflexbison` (https://github.com/lexxmark/winflexbison) the default binaries are called `win_flex` and `win_bison`. It takes an extra manual step to rename these to `flex` or `bison`

regression/cpp/Makefile
thomasspriggs: Hmm. I thought this was supposed to be a visual studio build. I.E. not using g++ at all. Still not a blocking comment though, as we need to get this merged rather than continuing to have no tests of Windows + make.

src/common
tautschnig: Is this the only place that people could possibly source flex/bison for Windows from? Putting it differently, I don't recall hearing from anyone that they were unhappy with our Makefiles for Windows, which either means that no one is using them or existing users have a working set-up. So I'm with @kroening: set `YACC='win_bison -y'` and `LEX=win_flex` on the command-line.

thomasspriggs: There are many ways to get Flex / Bison on Windows. If a user has a build environment based on Cygwin or MSYS2 for example then the packages for those environments may still be using the original names. I'd need to check to know for sure, but I don't have a Windows machine on hand.

regression/cpp/Makefile
NlightNFotis: This *is* a Visual Studio build only - I've gone to great lengths to ensure that the only compiler used is `cl.exe` through `clcache`.

I'm talking about the previous code build job this is replicating. The reason I said I'm assuming it ended up using `g++` is a weird explicit setup of the path to Cygwin tools before it called `make`.

thomasspriggs: Ah, now I understand. :+1: It makes sense that the tests could have been failing silently before due to incorrect setup.

src/common
NlightNFotis: @tautschnig Are we sure we want to do that? Just for clarification, there are 3 potential configurations for building on windows (that I can see in `src/common`), and they are Cygwin, MSVC and MinGW based. For them:

1. The `MinGW` configuration is inheriting `flex` and `bison` binary names from the default configuration,
2. The `Cygwin` configuration also sets them explicitly to `flex` and `bison`, and
3. The `MSVC` configuration is the one I changed to address `win_flex` and `win_bison`, which are the default binaries a user should end up with if he installs `winflexbison` (which is the default way a programmer would acquire these to use them with Visual Studio today, to my understanding).

So, I guess my question is: Does this really break anything? Alternative build environments on Windows should not be affected in any capacity (if a user is using Cygwin or MinGW) and the MSVC one I can argue that was misconfigured (or at least stale by current standards).

thomasspriggs: I have now taken another look at the rest of this file. I have now seen that there are other sets of configuration in this file for cygwin and mingw. Therefore I am less concerned about explicitly changing this for Visual Studio only. I am happy for this change to be merged as is.

thomasspriggs (approved):
LGTM.

src/common
tautschnig: > So, I guess my question is: Does this really break anything? [...] and the MSVC one I can argue that was misconfigured [...]

I simply did not know whether `winflexbison` was the (single?) most common approach. If you're confident that's the case, then, sure, let's go with it. My concern was breaking existing user's set-ups, but if it seems unlikely that anyone was (successfully) using this configuration, then there shouldn't be a problem. Well, if there is, people will tell us and we can revert :-)

chrisr-diffblue (approved):
It seems like the contentious part of the PR only affects VS builds, and not _all_ windows builds - which perhaps was a bit of context missing from the original PR/commit history. Happy for this to go in.

.gitattributes
tautschnig: Can we fix `test.pl` instead? I've read rumours that Strawberry Perl isn't doing a good job adapting to the OS, so maybe we should just accept that there are perl implementations that don't play ball. I believe all that it takes is replacing `chomp` in line 79 of `test.pl` by some use of `s/\R$//`.

tautschnig: More concretely: Replacing `chomp @data;` by `s/\R$// for @data;` should do the trick.

NlightNFotis: I'm happy with this if this works. I will wait for CI and if this is working, I am going to merge it.

PR #5859, opened 2021-02-23
Use endianness map for both big and little endian

tautschnig:
Avoid special-casing little endianness, and use endianness map for union
member access.

### DIFF
--- a/src/solvers/flattening/boolbv.h
+++ b/src/solvers/flattening/boolbv.h
@@ -104,6 +104,8 @@ public:
     return endianness_mapt{type, little_endian, ns};
   }
 
+  virtual endianness_mapt endianness_map(const typet &type) const;
+
 protected:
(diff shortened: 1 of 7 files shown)


src/solvers/flattening/boolbv.h
```diff
@@ -104,6 +104,8 @@ class boolbvt:public arrayst
     return endianness_mapt{type, little_endian, ns};
   }
 
+  virtual endianness_mapt endianness_map(const typet &type) const;
```
martin-cs, line 107: Should this be the only public interface for `endianness_map`?

martin-cs (approved):

src/solvers/flattening/boolbv.h
tautschnig: No, because byte extracts do (potentially) override the configured endianness.

PR #5866, opened 2021-02-25
Abort after failing C assert

### LINKED ISSUE #5505: Propagating information from assertions that hold to subsequent lines of code using `__CPROVER_assume`
CBMC version: `5.13.0 (cbmc-5.13.1-46-g1276c241e)`
Operating system: `macOS Catalina 10.15.6`
Exact command line resulting in the issue: `make result`
What behaviour did you expect: In functions `addHeader` and `writeRequestLine`, the `pRequestHeaders` points to only a dynamically allocated object (points-to set size for `pRequestHeaders` is 1)
What happened instead: `pRequestHeaders` points to NULL and dynamically allocated object (points-to set size is 2)

**To reproduce these results use the following:**
Code base: `awslabs/aws-iot-device-sdk-embedded-C`
Proof: `HTTPClient_InitializeRequestHeaders`
Code without optimisation: https://github.com/natasha-jeppu/aws-iot-device-sdk-embedded-C/blob/njjeppu-branch/libraries/standard/http/src/http_client.c#L1118
https://github.com/natasha-jeppu/aws-iot-device-sdk-embedded-C/blob/njjeppu-branch/libraries/standard/http/src/http_client.c#L1195
Code with optimisation: https://github.com/natasha-jeppu/aws-iot-device-sdk-embedded-C/blob/assert_then_assume/libraries/standard/http/src/http_client.c#L1136
https://github.com/natasha-jeppu/aws-iot-device-sdk-embedded-C/blob/assert_then_assume/libraries/standard/http/src/http_client.c#L1216
Command: make result

**Optimisation:** The assertion on https://github.com/natasha-jeppu/aws-iot-device-sdk-embedded-C/blob/njjeppu-branch/libraries/standard/http/src/http_client.c#L1209 and https://github.com/natasha-jeppu/aws-iot-device-sdk-embedded-C/blob/njjeppu-branch/libraries/standard/http/src/http_client.c#L1129 always holds true. So the `pRequestHeaders` is not `NULL` in the `addHeader` and `writeRequestLine` functions. We add an equivalent assume right after to propagate this information to subsequent lines of code. 

The optimisation produces about a 6% speedup (this is after applying the optimisation to a single pointer) but what's interesting here is that we can reduce the points-to set size by doing so. This also enables constant propagation thereby reducing the number of dereferences. I have attached screenshots of the points-to set reports below.

**Without Optimization:**
Point-to set size: `2`
Points-to set: `['object_descriptor(NULL-object, 0)', 'object_descriptor(symex_dynamic::dynamic_object1, 0)']`
Runtime: `957.319s`
Screenshot 2020-09-24 at 16 09 23

**With Optimization:**
Point-to set size: `1`
Points-to set: `['object_descriptor(symex_dynamic::dynamic_object1, 0)']`
Runtime: `893.876s`
Screenshot 2020-09-24 at 16 09 09

tautschnig:
The C standard specifies behaviour of the "assert" macro as resulting in
an abort when the condition does not evaluate to true. Implement this
behaviour by inserting assume(0) after assert(0).

Fixes: https://github.com/diffblue/cbmc/issues/5505

### DIFF
--- a/regression/cbmc-incr-oneloop/multiple-asserts/test.c
+++ b/regression/cbmc-incr-oneloop/multiple-asserts/test.c
@@ -2,8 +2,8 @@ int main()
 {
   for(int i = 0; i < 10; ++i)
   {
-    assert(i != 5);
-    assert(i != 8);
+    __CPROVER_assert(i != 5, "assertion i != 5");
+    __CPROVER_assert(i != 8, "assertion i != 8");
   }
   return 0;
 }
--- a/regression/cbmc-cover/simple_assert/main.c
+++ b/regression/cbmc-cover/simple_assert/main.c
@@ -1,9 +1,7 @@
-#include <assert.h>
-
 int main(int argc, char *argv[])
 {
   int x = 5;
-  assert(x == 5);
+  __CPROVER_assert(x == 5, "assertion x == 5");
 
   return 0;
 }
--- a/regression/cbmc-cover/simple_assert/test.desc
+++ b/regression/cbmc-cover/simple_assert/test.desc
@@ -9,5 +9,3 @@ main.c
 ^warning: ignoring
 ^CONVERSION ERROR$
 ^\[main\.coverage\..\] .* function main block .*: FAILED$
---
-On Windows/Visual Studio, "assert" does not introduce any branching.
--- a/regression/cbmc/Bool5/test.desc
+++ b/regression/cbmc/Bool5/test.desc
@@ -1,7 +1,7 @@
 CORE
 main.c
 
-Generated 4 VCC\(s\), 0 remaining after simplification
+Generated [34] VCC\(s\), 0 remaining after simplification
 ^VERIFICATION SUCCESSFUL$
 ^EXIT=0$
 ^SIGNAL=0$
--- a/regression/cbmc/Quantifiers-not-exists/fixed.c
+++ b/regression/cbmc/Quantifiers-not-exists/fixed.c
@@ -10,33 +10,43 @@ int main()
   // NOLINTNEXTLINE(whitespace/line_length)
   __CPROVER_assume(!__CPROVER_exists { int i; (i>=0 && i<2) && (__CPROVER_exists{int j; (j>=0 && j<2) && a[i][j]<=10}) });
 
-  assert(0);
+  __CPROVER_assert(0, "assertion 0");
 
   // NOLINTNEXTLINE(whitespace/line_length)
   __CPROVER_assume(__CPROVER_forall { int i; (i>=0 && i<2) ==> (!__CPROVER_exists{int j; (j>=0 && j<2) && b[i][j]>=1 && b[i][j]<=10}) });
 
-  assert(0);
+  __CPROVER_assert(0, "assertion 0");
 
   // NOLINTNEXTLINE(whitespace/line_length)
   __CPROVER_assume(!__CPROVER_exists { int i; (i>=0 && i<2) && (!__CPROVER_exists{int j; (j>=0 && j<2) && (c[i][j]==0 || c[i][j]<=10)}) });
 
-  assert(0);
+  __CPROVER_assert(0, "assertion 0");
 
   // NOLINTNEXTLINE(whitespace/line_length)
   __CPROVER_assume(!__CPROVER_exists { int i; (i>=0 && i<2) && (__CPROVER_forall{int j; (j>=0 && j<2) ==> d[i][j]>=1 && d[i][j]<=10}) });
   // clang-format on
 
-  assert(0);
+  __CPROVER_assert(0, "assertion 0");
 
-  assert(a[0][0] > 10);
+  __CPROVER_assert(a[0][0] > 10, "assertion a[0][0] > 10");
 
-  assert((b[0][0] < 1 || b[0][0] > 10) && (b[0][1] < 1 || b[0][1] > 10));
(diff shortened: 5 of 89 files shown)


regression/cbmc-incr-oneloop/multiple-asserts/test.c
```diff
@@ -2,8 +2,8 @@ int main()
 {
   for(int i = 0; i < 10; ++i)
   {
-    assert(i != 5);
-    assert(i != 8);
+    __CPROVER_assert(i != 5, "");
```
peterschrammel: ❓ Do we need a convenienve overload without the comment argument since `__CPROVER_assert` is now the default way to specify non-aborting properties?

peterschrammel (commented):
I think this requires a thorough documentation review since `__CPROVER_assert` is now the default way to specify non-aborting properties. Some downstream CBMC users might not operate with abort-semantics in mind.

Side note: JBMC does something similar (in one of its modes) https://github.com/diffblue/cbmc/blob/f4edb72f3db37f039c215d35bb0d75bd80540742/jbmc/src/java_bytecode/java_bytecode_convert_method.cpp#L2444-L2446

regression/cbmc-incr-oneloop/multiple-asserts/test.c
hannes-steffenhagen-diffblue: Sounds like a good idea to me

hannes-steffenhagen-diffblue:
IMHO this sounds like the right way to go forward with this. I do also agree with @peterschrammel though that this requires us going through documentation and making sure we don't say anything wrong now, and for the other I think this should result in a major version bump.

regression/goto-harness/do-not-use-nondet-for-selecting-pointers-to-treat-as-equal/test.desc
```diff
@@ -2,11 +2,11 @@ CORE
...
-\[test.assertion.1\] line 3 assertion x: SUCCESS
-\[test.assertion.2\] line 4 assertion y: SUCCESS
-\[test.assertion.3\] line 5 assertion x == y: FAILURE
-\[test.assertion.4\] line 6 assertion x != y: FAILURE
-\[test.assertion.5\] line 7 assertion \*x == \*y: FAILURE
+\[test.assertion.1\] line 3 assertion: SUCCESS
+\[test.assertion.2\] line 4 assertion: SUCCESS
+\[test.assertion.3\] line 5 assertion: FAILURE
+\[test.assertion.4\] line 6 assertion: FAILURE
+\[test.assertion.5\] line 7 assertion: FAILURE
```
TGWDB: Have we lost something useful and helpful here by not showing any detail of the assertion that failed? (I realise the empty messages may contribute to this, but with the comment of @peterschrammel above perhaps we should consider having the same behaviour in our override?)

martin-cs: +1 these should give a message...

martin-cs:
We are doing a similar assert/assume thing for Ada.

As regards updating -- this always used to be how `assert()` worked.  I guess it must have been removed at some point?  There were a few years where there was ... ambiguity over the semantics of assert.

martin-cs (approved):
A good and useful thing to do.  I won't block this but I do agree with others that this needs to be clearly communicated (or, at least, "If you want a non-terminating check, use `__CPROVER_assert()`").

One thought : it feels like `assert` should be implementable in the C library (possibly with a macro to copy the expression into the string so that we don't have the problem with loss of messages that @TGWDB flagged).  I am generally in favour of modelling going in the library in C where possible.

SaswatPadhi:
I have one small comment that I had mentioned to @tautschnig this morning:

The semantics for `ASSUME` statement currently is stated (in [`src/goto-programs/goto_program.h`](https://github.com/diffblue/cbmc/blob/develop/src/goto-programs/goto_program.h)) as:
> This thread of execution waits for `guard` to evaluate to true.
> Assume does not "retro-actively" affect the thread or any ASSERTs.

This means, if the `guard` evaluates to `false`, it would keep waiting forever as opposed to aborting (as mandated by C `assert` semantics).

It seems to me that `THROW` might be the closest we have to an aborting goto primitive, but I'm not sure.

martin-cs:
@SaswatPadhi This is a sensible and a valid concern but I think that `assume` is the right thing to do.  Our (informal) model of execution has nothing happen after termination, so in the single threaded case, an abrupt termination and an infinite loop are indistinguishable.  The multi-threaded case is a little more complex and I haven't worked on it so I am probably not the best person to ask but I don't think we have a way of distinguishing a "hung" thread from one that terminates early, so, again, this is not an issue.

`THROW`and `CATCH`... TBH their semantics are even less clear to me.  I think the clearest explanation is the code that removes them.  This is different for C++ and Java.  In practice they are removed early on so I don't think we should be introducing them, esp. in things like C that don't have an exception model at all.

tautschnig:
Marking do-not-merge as this likely wants a major-version bump. Otherwise it should be ready for (re-)review.

martin-cs (approved):

feliperodri:
Fixes: https://github.com/diffblue/cbmc/issues/5505

esteffin:
We are considering to include this PR into version 6.

However we should make sure that we do add the documentation specifying that `assert` is blocking as opposed to `__CPROVER_assert` that is not.

tautschnig:
This is a major undertaking with limited added value. Users could choose to add their own assert macro that exhibits this behaviour. Closing as we are unlikely to find time to work on this anytime soon.

PR #5867, opened 2021-02-25
dump-c: fix output of _Bool-typed constants

tautschnig:
Only __CPROVER_bool-typed constants can be used with exprt::is_true.
exprt::is_one's documentation promised that ID_c_bool was supported,
which wasn't actually the case. This commit fixes this and makes use of
is_one to fix the bug in dump-c.

### DIFF
--- a/regression/goto-instrument/dump-bool/test.desc
+++ b/regression/goto-instrument/dump-bool/test.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+--dump-c
+VERIFICATION SUCCESSFUL
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+Dump-c previously always generated 0 for _Bool or __CPROVER_bool-typed
+constants. This test ensures that the value "1" (in v = 1) is correctly
+generated, instead of a spurious (v = (__CPROVER_bool)0).
--- a/src/goto-instrument/dump_c.cpp
+++ b/src/goto-instrument/dump_c.cpp
@@ -1469,11 +1469,18 @@ void dump_ct::cleanup_expr(exprt &expr)
       }
     }
 
+    optionalt<exprt> clean_init;
     if(
       ns.follow(bu.type()).id() == ID_union &&
-      bu.source_location().get_function().empty() &&
-      bu.op() == zero_initializer(bu.op().type(), source_locationt{}, ns)
-                   .value_or(nil_exprt{}))
+      bu.source_location().get_function().empty())
+    {
+      clean_init = zero_initializer(bu.op().type(), source_locationt{}, ns)
+                     .value_or(nil_exprt{});
+      if(clean_init->id() != ID_struct || clean_init->has_operands())
+        cleanup_expr(*clean_init);
+    }
+
+    if(clean_init.has_value() && bu.op() == *clean_init)
     {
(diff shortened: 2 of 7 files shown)


kroening (approved):

regression/goto-instrument/dump-bool/test.desc
```diff
@@ -0,0 +1,12 @@
...
+--dump-c
+VERIFICATION SUCCESSFUL
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+irep
+--
+Dump-c previously always generated 0 for _Bool or __CPROVER_bool-typed
+constants. This test ensures that the value "1" is correctly generated.
```
hannes-steffenhagen-diffblue: I don't understand how this test tests this?

src/goto-instrument/dump_c.cpp
```diff
@@ -1521,22 +1521,48 @@ void dump_ct::cleanup_type(typet &type)
     !type.get(ID_tag).empty());
 }
 
+static expr2c_configurationt expr2c_configuration()
+{
+  static expr2c_configurationt configuration{
```
hannes-steffenhagen-diffblue, line 1535: IMHO would be better written as a series of assignments, then you don't need the comments.

src/goto-instrument/dump_c_class.h
```diff
@@ -156,7 +152,7 @@ class dump_ct
   symbol_tablet copied_symbol_table;
   const namespacet ns;
   const dump_c_configurationt dump_c_config;
-  std::unique_ptr<languaget> language;
+  const irep_idt mode;
```
hannes-steffenhagen-diffblue, line 155: This change should be a separate commit with an explanation for why it's done. It's not obvious to me that this is an improvement, and in any case it doesn't seem to be directly related to fixing anything with the output of booleans? Not saying I'm against this, just that the commit should do a better job of explaining it.

regression/goto-instrument/dump-bool/test.desc
martin-cs: Yeah... I'm not seeing it either.  Not quite sure where the `irep` is coming from.

src/goto-instrument/dump_c.cpp
```diff
@@ -526,7 +526,7 @@ void dump_ct::convert_compound(
       if(t.get_width()<=config.ansi_c.long_long_int_width)
         struct_body << "long long int " << comp_name
           << " : " << t.get_width();
-      else if(language->id()=="cpp")
+      else if(mode == ID_cpp)
```
martin-cs, line 529: Doesn't seem to correspond with the PR message.  I'm fine with them going in as a minor fix but just wanted to flag it in case it was a mistake.

martin-cs (approved):

peterschrammel (approved):

regression/goto-instrument/dump-bool/test.desc
tautschnig: Test description expanded: the previous (buggy) output made dump-c produce v = 0 instead of v = 1.

src/goto-instrument/dump_c_class.h
tautschnig: Broken out into a second commit.

PR #5869, opened 2021-02-25
Fix bug in constant propagator

hannes-steffenhagen-diffblue:
Before it was using the index operator. However, since there is no
initialisation, this caused an error if there was unreachable code in
the goto model. This can be fixed by simply using abstract_state_before
which will always return a valid domain state (bottom for unreachable
code) instead.

### DIFF
--- a/regression/goto-instrument/constant_propagator_unreachable/test.c
+++ b/regression/goto-instrument/constant_propagator_unreachable/test.c
@@ -1,0 +1,12 @@
+#include <assert.h>
+
+void unreachable(void)
+{
+  int x = 0;
+  assert(x != 0);
+}
+
+int main(void)
+{
+  return 0;
+}
--- a/regression/goto-instrument/constant_propagator_unreachable/test.desc
+++ b/regression/goto-instrument/constant_propagator_unreachable/test.desc
@@ -1,0 +1,11 @@
+CORE
+test.c
+--constant-propagator
+^EXIT=0$
+^SIGNAL=0$
+VERIFICATION SUCCESSFUL
+--
+failed to find state
+--
+This is testing for a previous regression that caused constant propagation to
+fail in unreachable code
--- a/src/analyses/constant_propagator.cpp
+++ b/src/analyses/constant_propagator.cpp
@@ -751,8 +751,10 @@ void constant_propagator_ait::replace(
 {
   Forall_goto_program_instructions(it, goto_function.body)
   {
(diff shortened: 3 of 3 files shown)


TGWDB (approved):

tautschnig (approved):

peterschrammel (approved):

NlightNFotis (approved):

martin-cs:
Thanks for updating this use to the new API.  It looks like this one escaped me.

( Obligatory advert : the constant propagator has some soundness issues.  `goto-analyze --simplify out.gb --variable-sensitivity-domain` should already be more correct and will hopefully continue to be better. )

peterschrammel:
> has some soundness issues

Yes, we should replace this by vsd. I'm surprised that this once late-night-hack-before-submission is still in use.

SaswatPadhi:
I was using this flag to try and shrink a huge goto file. Thanks @martin-cs and @peterschrammel for mentioning the vsd flag. I'll give it a shot.

May be the constant prop flag could be deprecated and a warning could be issued to use the vsd flag instead?

martin-cs:
@peterschrammel : I think you should take it as a complement!

@SaswatPadhi : `goto-analyzer` and VSD in particular are currently "under development" so feedback most appreciated.

SaswatPadhi:
Hi @martin-cs 

I tried using `goto-analyzer` in place of `goto-instrument` but I can't figure out the right set of command-line flags.

Previous cmdline:

```console
$ goto-instrument --constant-propagator in.goto out.goto
```

Current cmdline:

```console
$ goto-analyze --simplify in.gb --vsd
GOTO-ANALYSER version 5.24.0 (cbmc-5.24.0-71-g8122b9d35-dirty) 64-bit x86_64 macos
Invalid User Input
Option: filename
Reason: missing program argument
Suggestion: one or more paths to program files
```

Does `goto-analyzer` also need the `.c` source file? The `--help` seems to suggest that. If so, how should I use it when there are  multiple `.c` files in a large project?

martin-cs:
@SaswatPadhi I'm happy to help with questions on `goto-analyze` but we should probably find a better place to discuss; either e-mail me or open an issue?

The command-line you are looking for is:

```
$ goto-analyze --simplify out.gb --vsd in.gb
```

Alternatively, instead of `in.gb` you can use one or more C files.  It's handling of files should be the same as CBMC, you can give them all or you can use `goto-cc` to compile them first.

Issue #5841, opened 2021-02-19
Feature request: Saturating arithmatic


danielsn:
It would be really useful to have saturating arithmetic primitives in CBMC

martin-cs:
This should be reasonably do-able but it would be good to know a bit more about what you want.  Are you using the internal back-end or an SMT solver?  Do you want signed or unsigned?  Do you have a particular implementation / system in mind?  What about the edge cases involving the limit value(s); what is LIMIT * 0, what is LIMIT / 0 or 0 / LIMIT?

peterschrammel:
@danielsn, can you give a code example of a use case?

martin-cs:
Note that https://github.com/diffblue/cbmc/issues/5405 would give a relatively simple and probably pretty good route to implementing them.

danielsn:
I'm trying to implement operations like these: 
I'm trying to implement these:
https://doc.rust-lang.org/std/intrinsics/fn.saturating_add.html
https://doc.rust-lang.org/std/intrinsics/fn.saturating_sub.html
https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_mul

PR #5872, opened 2021-02-26
Simplify overflow-* expressions

tautschnig:
We can simplify these over constants and for some trivial cases when
mathematical types are used.

### DIFF
--- a/regression/ansi-c/simplify-overflow/test.desc
+++ b/regression/ansi-c/simplify-overflow/test.desc
@@ -1,0 +1,7 @@
+CORE
+main.c
+
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
--- a/src/util/simplify_expr.cpp
+++ b/src/util/simplify_expr.cpp
@@ -2047,6 +2047,117 @@ simplify_exprt::simplify_complex(const unary_exprt &expr)
   return unchanged(expr);
 }
 
+simplify_exprt::resultt<>
+simplify_exprt::simplify_overflow_binary(const binary_exprt &expr)
+{
+  // zero is a neutral element for all operations supported here
+  if(
+    expr.op1().is_zero() ||
+    (expr.op0().is_zero() && expr.id() != ID_overflow_minus))
+  {
+    return false_exprt{};
+  }
+
+  // we only handle the case of same operand types
+  if(expr.op0().type() != expr.op1().type())
+    return unchanged(expr);
+
+  // catch some cases over mathematical types
+  const irep_idt &op_type_id = expr.op0().type().id();
+  if(
+    op_type_id == ID_integer || op_type_id == ID_rational ||
+    op_type_id == ID_real)
+  {
+    return false_exprt{};
+  }
+
+  if(op_type_id == ID_natural && expr.id() != ID_overflow_minus)
+    return false_exprt{};
+
+  // we only handle constants over signedbv/unsignedbv for the remaining cases
+  if(op_type_id != ID_signedbv && op_type_id != ID_unsignedbv)
+    return unchanged(expr);
+
+  if(!expr.op0().is_constant() || !expr.op1().is_constant())
+    return unchanged(expr);
+
+  const auto op0_value = numeric_cast<mp_integer>(expr.op0());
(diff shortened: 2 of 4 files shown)


martin-cs (approved):
Useful for https://github.com/diffblue/cbmc/issues/5841

src/util/simplify_expr.cpp
```diff
@@ -2047,6 +2047,113 @@ simplify_exprt::simplify_complex(const unary_exprt &expr)
   return unchanged(expr);
 }
 
+simplify_exprt::resultt<>
+simplify_exprt::simplify_overflow_binary(const binary_exprt &expr)
+{
+  // zero is a neutral element for all operations supported here
+  if(expr.op0().is_zero() || expr.op1().is_zero())
+    return false_exprt{};
```
TGWDB: What about the following? This appears to have a zero and also an overflow.
unsigned int x = 0;
unsigned int y = 2;
x = x - y;

regression/cbmc/simplify-overflow/main.c
```diff
@@ -0,0 +1,9 @@
+int main()
+{
+  __CPROVER_assert(!__CPROVER_overflow_plus(1, 2), "");
+  __CPROVER_assert(__CPROVER_overflow_minus(1U, 2U), "");
+  __CPROVER_assert(!__CPROVER_overflow_mult(1U, 2U), "");
+  __CPROVER_assert(!__CPROVER_overflow_shl(1, 2), "");
+  __CPROVER_assert(!__CPROVER_overflow_unary_minus(1), "");
+  __CPROVER_assert(__CPROVER_overflow_unary_minus(1U), "");
+}
```
TGWDB: Maybe we should also have:
__CPROVER_assert(__CPROVER_overflow_minus(0U, 2U), "");
To check the case below where a 0 value can still have overflow?

src/util/simplify_expr_class.h
```diff
@@ -184,6 +184,8 @@ class simplify_exprt
   NODISCARD resultt<> simplify_sign(const sign_exprt &);
   NODISCARD resultt<> simplify_popcount(const popcount_exprt &);
   NODISCARD resultt<> simplify_complex(const unary_exprt &);
+  NODISCARD resultt<> simplify_overflow_binary(const binary_exprt &);
+  NODISCARD resultt<> simplify_overflow_unary(const unary_exprt &);
```
TGWDB: Doxygen documentation?

regression/cbmc/simplify-overflow/main.c
martin-cs: If so, it should probably be in a separate test case as the trick @tautschnig is using for testing that it is simplification that handles these will not work for false things.

```
while (__CPROVER_overflow_minus(0U, 2U) {}
```

could do something interesting though...

src/util/simplify_expr.cpp
martin-cs: Good catch!

tautschnig: Thanks, and added as a test!

regression/cbmc/simplify-overflow/main.c
tautschnig: Trying to cover this in the test spec: `^Generated 7 VCC\(s\), 0 remaining after simplification$`

kroening: I would say that this is both frail and also obscure.
How about using `__builtin_constant_p`?

tautschnig: Fair. I'm now using `_Static_assert` instead.

regression/ansi-c/simplify-overflow/test.desc
```diff
@@ -0,0 +1,7 @@
+CORE gcc-only
```
kroening: The `#ifdef _MSC_VER` suggests that the test might work with goto-cl?

kroening (approved):

regression/ansi-c/simplify-overflow/test.desc
tautschnig: Ah, thank you, that was a left-over from when I thought about the `__builtin_constant_p` approach. Fixed.

PR #917, opened 2017-05-13
Do not hardcode char/bytes having 8 bits

tautschnig:
The C standard does not guarantee that char is exactly 8 bits, and there are DSPs (such Texas Instruments C55x) that do not have 8-bit-bytes. Use the configuration value instead.

PR #5873, opened 2021-02-27
Simplify byte-extract from struct or union expressions

tautschnig:
With rewrite_union introducing rewriting union accesses to byte-extract
operations, simplifying such expressions must cover as many cases as
possible. Constants were mostly already being rewritten via expr2bits,
but that wouldn't be able to handle expressions involving pointers (even
when they are effectively constants when using address-of operations).
The additional simplification rules mimic what lower_byte_extract
already did at a later stage, but such late rewrites would not benefit
constant propagation done by goto-symex.

### DIFF
--- a/src/util/pointer_offset_size.cpp
+++ b/src/util/pointer_offset_size.cpp
@@ -16,6 +16,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "c_types.h"
 #include "invariant.h"
 #include "namespace.h"
+#include "pointer_expr.h"
 #include "simplify_expr.h"
 #include "ssa_expr.h"
 #include "std_expr.h"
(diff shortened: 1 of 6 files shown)


src/util/pointer_offset_size.cpp
```diff
@@ -681,6 +682,27 @@ optionalt<exprt> get_subexpression_at_offset(
...
+    const union_typet &union_type = to_union_type(source_type);
+
+    for(const auto &component : union_type.components())
+    {
+      const auto m_size_bits = pointer_offset_bits(component.type(), ns);
+      if(!m_size_bits.has_value())
+        continue;
+
+      // if this member completely contains the target, recurse into it
+      if(offset_bytes * 8 + *target_size_bits <= *m_size_bits)
```
martin-cs, line 698: `CHAR_BIT` or `config.ansi_c.char_width`  You don't want to annoy the author of https://github.com/diffblue/cbmc/pull/917/ ;-)

martin-cs (approved):
This feels like the kind of thing that could have or acquire subtle bugs.  Would it be possible to add a test case or two?  It feels like 100% patch coverage is possible and might be worth having.

src/util/pointer_offset_size.cpp
tautschnig: The author of #917 surely will want to consistently replace all occurrances of "8," so I'll leave it to him ;-P

tautschnig:
> This feels like the kind of thing that could have or acquire subtle bugs. Would it be possible to add a test case or two? It feels like 100% patch coverage is possible and might be worth having.

Indeed, test added.

src/util/pointer_offset_size.cpp
martin-cs: OK.  Not sure he's going to be happy but I'll let you fight that fight!

martin-cs:
Thanks.

peterschrammel (approved):

PR #5874, opened 2021-02-27
Move the declaration and description of options that configt handles into config.h

martin-cs:
This is a tidy up that should not have any real effect on the tools but makes the source clearer.  It raises the issue that `--function` and `--string-abstraction` are handled by `configt` and by parsing into `optionst`.  I think that should be left for tidying up of the functionality of `configt` (or even, you know, removing it all together).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,7 @@ regression/**/tests-*.log
 regression/**/*.goto-cc-saved
 regression/**/*.gb
 regression/**/*.smt2
+regression/solver-hardness/solver-hardness-simple/solver_hardness.json
 jbmc/regression/**/tests.log
 jbmc/regression/**/tests-symex-driven-loading.log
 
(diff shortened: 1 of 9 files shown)


martin-cs:
The only failed test is diff coverage which I think drops because this is a net removal of (duplicated) code.  I don't think this is a problem as such.

peterschrammel (approved):
lgtm

What about the other executables besides cbmc and goto-analyzer?

martin-cs:
@peterschrammel : `goto-cc` handles it's options in a completely different way so this code doesn't work for it. `goto-instrument` and `goto-diff` do not have these options (although, via `configt` the code is there) -- I was surprised too.  I don't know what is going on with the Java tools in general so I tend to leave them be.  This change shouldn't break anything in these tools.

tautschnig (approved):

PR #5876, opened 2021-03-01
Value set dereferencing: do not treat struct prefixes as equal

tautschnig:
Two distinct struct types cannot be cast between, even when one is a
prefix of the other. Value set dereferencing taking this approach just
resulted in propositional encoding producing a warning about invalid
type casts.

### DIFF
--- a/jbmc/regression/jbmc/simplify-classid-of-interface/test_no_simplify_vccs.desc
+++ b/jbmc/regression/jbmc/simplify-classid-of-interface/test_no_simplify_vccs.desc
@@ -3,9 +3,13 @@ Test
 --function Test.main --no-simplify --unwind 10 --show-vcc -cp target/classes
 ^EXIT=0$
 ^SIGNAL=0$
-"java::Impl1" = cast\(\{ \{ "java::Impl1" \}, 0 \}, struct (\{ struct \{ string @class_identifier \} @java.lang.Object \}|java::Intf)\)\.@java.lang.Object\.@class_identifier
+"java::Impl1" = byte_extract_little_endian\(\{ \{ "java::Impl1" \}, 0 \}, 0, struct java::Intf\)\.@java.lang.Object\.@class_identifier
 --
 ^warning: ignoring
 --
 This checks that the test generates the VCC testing the class-id that we're
-intending to simplify away in the main test.
+intending to simplify away in the main test. Following the removal of the
+struct-prefix shortcut from value_set_dereferencet::dereference_type_compare
+the encoding of the Intf/Impl1 cast in the VCC is via byte_extract_little_endian
+rather than a struct-to-struct cast (which boolbvt::conversion_failed would
+silently drop with a "warning: ignoring").
--- a/regression/cbmc/struct-prefix-deref1/main.c
+++ b/regression/cbmc/struct-prefix-deref1/main.c
@@ -1,0 +1,28 @@
+// Test that value set dereferencing correctly handles the case where a
+// pointer to a smaller struct is used to access an object of a larger
+// struct type (struct prefix relationship). The removed struct-prefix
+// check in dereference_type_compare would have produced a direct cast
+// (and a "warning: ignoring" from the solver); now it correctly uses
+// byte_extract instead.
+
+#include <assert.h>
+
+struct small
+{
+  int x;
+};
+
+struct big
+{
+  int x;
+  int y;
+};
+
+int main()
+{
+  struct big b = {1, 2};
+  void *v = &b;
+  struct small *p = (struct small *)v;
+  assert(p->x == 1);
+  return 0;
+}
--- a/regression/cbmc/struct-prefix-deref1/vcc.desc
+++ b/regression/cbmc/struct-prefix-deref1/vcc.desc
@@ -1,0 +1,17 @@
+CORE
+main.c
+--no-simplify --show-vcc
+^EXIT=0$
+^SIGNAL=0$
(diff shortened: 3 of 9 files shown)


martin-cs:
I had to go back and check my copy of ISO 9899 because I thought there was a rule about this but ... I can't find it.  I wrote the following...

```
#include <assert.h>

struct a { int x; };
struct b { struct a p; int y; };

int f00 (struct a *ptr) {
  return ptr->x;
}

int main() {
  struct b z = { {1}, 2 };
  
  assert(&z == &(z.p));
  assert(&z == &(z.p.x));

  assert(f00(&z) == z.p.x);
  
  return 0;
}
```

This compiles but I don't think it is actually standard C:

```
$ gcc  struct-pointer.c -o struct-pointer 
In file included from struct-pointer.c:1:
struct-pointer.c: In function ‘main’:
struct-pointer.c:13:13: warning: comparison of distinct pointer types lacks a cast
   assert(&z == &(z.p));
             ^~
struct-pointer.c:13:13: warning: comparison of distinct pointer types lacks a cast
   assert(&z == &(z.p));
             ^~
struct-pointer.c:14:13: warning: comparison of distinct pointer types lacks a cast
   assert(&z == &(z.p.x));
             ^~
struct-pointer.c:14:13: warning: comparison of distinct pointer types lacks a cast
   assert(&z == &(z.p.x));
             ^~
struct-pointer.c:16:14: warning: passing argument 1 of ‘f00’ from incompatible pointer type [-Wincompatible-pointer-types]
   assert(f00(&z) == z.p.x);
              ^~
struct-pointer.c:6:20: note: expected ‘struct a *’ but argument is of type ‘struct b *’
 int f00 (struct a *ptr) {
          ~~~~~~~~~~^~~
In file included from struct-pointer.c:1:
struct-pointer.c:16:14: warning: passing argument 1 of ‘f00’ from incompatible pointer type [-Wincompatible-pointer-types]
   assert(f00(&z) == z.p.x);
              ^~
struct-pointer.c:6:20: note: expected ‘struct a *’ but argument is of type ‘struct b *’
 int f00 (struct a *ptr) {
          ~~~~~~~~~~^~~
```

which we can handle:

```
$ cbmc struct-pointer.c
CBMC version 5.24.0 (cbmc-5.24.0-81-g19fca2ebfc) 64-bit x86_64 linux
Parsing struct-pointer.c
Converting
Type-checking struct-pointer
file struct-pointer.c line 16 function main: warning: conversion from 'struct b *' to 'struct a *': incompatible pointer types
file struct-pointer.c line 16 function main: warning: conversion from 'struct b *' to 'struct a *': incompatible pointer types
Generating GOTO Program
Runtime Symex: 0.00476498s
size of program expression: 54 steps
simple slicing removed 0 assignments
Generated 3 VCC(s), 0 remaining after simplification
Runtime Postprocess Equation: 3.4174e-05s

** Results:
struct-pointer.c function main
[main.assertion.1] line 13 assertion &z == &(z.p): SUCCESS
[main.assertion.2] line 14 assertion &z == &(z.p.x): SUCCESS
[main.assertion.3] line 16 assertion f00(&z) == z.p.x: SUCCESS

VERIFICATION SUCCESSFUL
```

and correctly put in the casts:

```
#include <assert.h>

struct a { int x; };
struct b { struct a p; int y; };

int f00 (struct a *ptr) {
  return ptr->x;
}

int main() {
  struct b z = { {1}, 2 };
  
  assert(&z == &(z.p));
  assert(&z == &(z.p.x));

  assert(f00(&z) == z.p.x);
  
  return 0;
}
```

jbmc/regression/jbmc/simplify-classid-of-interface/test_no_simplify_vccs.desc
```diff
@@ -1,11 +0,0 @@
-CORE
```
martin-cs, line 1: If we are removing the `.desc` file, should we remove the Java file as well or is it used for other things?

martin-cs (approved):
Seems correct but removing things like this makes me nervous.  Please could you check whether it is covered by any of the tests-- it should be dead right?  The code seems to be almost exactly a decade old but that might well be the date of the move from CVS to Subversion and the "start" of history.

jbmc/regression/jbmc/simplify-classid-of-interface/test_no_simplify_vccs.desc
tautschnig: There are other tests in this directory that are still using this Java code and that do pass as expected.

tautschnig:
> I had to go back and check my copy of ISO 9899 because I thought there was a rule about this but ... I can't find it. I wrote the following...

[...]

Thanks, I included your proposed test as a further regression test.

Also, it may be that some of the tests actually used this code, but they will now fall back to the code path that actually does the right thing and creates a byte extract.

chris-ryder (approved):

kroening:
Structs can't be casted, but pointers to structs can be casted to pointers to structs that are prefixes -- this is what this piece of code was meant to cover.

tautschnig:
> Structs can't be casted, but pointers to structs can be casted to pointers to structs that are prefixes -- this is what this piece of code was meant to cover.

But then `value_set_dereferencet::dereference_type_compare` would have to do a test of `is_prefix_of` on `object_unwrapped` and `dereference_unwrapped` with the additional check that they are not the same as `object_type` (and `dereference_type`). Is that the change we should be making?

PR #5877, opened 2021-03-01
Avoid spurious libgcc-s1 Debian package dependency

### LINKED ISSUE #5875: Ubuntu 18.04 deb depends on libgcc-s1 which isn't in 18.04, it should depend on libgcc1 instead
CBMC version: 5.24 (also older packages)
Operating system: Ubuntu 18.04
Exact command line resulting in the issue: dpkg -i ubuntu-18.04-cbmc-5.24.0-Linux.deb
What behaviour did you expect: CBMC gets installed
What happened instead: dpkg returns: cbmc : Depends: libgcc-s1 (>= 3.4) but it is not installable

libgcc-s1 isn't in Ubuntu 18.04: https://packages.ubuntu.com/search?suite=xenial&arch=amd64&keywords=libgcc-s1
Update: Sorry, Ubuntu 18.04 is bionic, not xenial: https://packages.ubuntu.com/search?suite=bionic&arch=amd64&keywords=libgcc-s1

The file libgcc_s.so.1 does exist, but is provided by the libgcc1 library in 18.04.

I tried manually extracting and running cbmc, and it does seem to work, so `dpkg -i --force-depends` should work.

tautschnig:
Marked as Draft (and do-not-merge, do-not-review), includes extra debug output.

Do not add a strict dependency on gcc, which in a GitHub container isn't
necessarily the same version as in a pristine Ubuntu system.

Fixes: #5875

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -375,6 +375,10 @@ jobs:
         run: | 
           sudo apt-get update
           sudo apt-get install --no-install-recommends -y g++ flex bison cmake ninja-build maven jq libxml2-utils dpkg-dev ccache
+          # remove libgcc-s1, which isn't normally available in Ubuntu 18.04
+          target=$(dpkg-query -W --showformat='${Version}\n' gcc-8-base | head -n 1)
+          # libgcc1 uses an epoch, thus the extra 1:
+          sudo apt-get install -y --allow-downgrades --reinstall gcc g++ libgcc-s1- libstdc++6=$target liblsan0=$target libtsan0=$target libcc1-0=$target libgcc1=1:$target
       - name: Prepare ccache
         uses: actions/cache@v2
         with:
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -69,7 +69,12 @@ jobs:
         with:
           submodules: recursive
       - name: Fetch dependencies
-        run: sudo apt-get install --no-install-recommends -y g++ flex bison cmake ninja-build maven jq libxml2-utils dpkg-dev ccache
+        run: |
+          sudo apt-get install --no-install-recommends -y g++ flex bison cmake ninja-build maven jq libxml2-utils dpkg-dev ccache
+          # remove libgcc-s1, which isn't normally available in Ubuntu 18.04
+          target=$(dpkg-query -W --showformat='${Version}\n' gcc-8-base | head -n 1)
+          # libgcc1 uses an epoch, thus the extra 1:
+          sudo apt-get install -y --allow-downgrades --reinstall gcc g++ libgcc-s1- libstdc++6=$target liblsan0=$target libtsan0=$target libcc1-0=$target libgcc1=1:$target
       - name: Prepare ccache
         uses: actions/cache@v2
         with:
--- a/cmake/packaging.cmake
+++ b/cmake/packaging.cmake
@@ -27,6 +27,10 @@ set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS YES)
 # In addition, we depend on gcc for preprocessing
 set(CPACK_DEBIAN_PACKAGE_DEPENDS gcc)
 
+# Enable debug output so that we can see the dependencies being generated in the
+# logs
+set(CPACK_DEBIAN_PACKAGE_DEBUG YES)
+
 # For windows we need to set up product and update GUID
 # See: https://docs.microsoft.com/en-us/windows/win32/msi/productcode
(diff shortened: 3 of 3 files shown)


.github/workflows/pull-request-checks.yaml
```diff
@@ -375,6 +375,7 @@ jobs:
         run: | 
           sudo apt-get update
           sudo apt-get install --no-install-recommends -y g++ flex bison cmake ninja-build maven jq libxml2-utils dpkg-dev ccache
+          sudo apt-get --purge remove -s libgcc-s1
```
hannes-steffenhagen-diffblue: Huh, that seems screwy? So the github actions ubuntu-18.04 container isn't necessarily actually ubuntu-18.04?

hannes-steffenhagen-diffblue:
For curiosities sake, how did you detemine this was the issue?

kroening:
The cleaner answer here could be do build that package in a docker container (Github actions does this). The problem with the proposed tweak is that Github may well choose to make further changes to their standard image, and then the issue is back.

hannes-steffenhagen-diffblue:
@tautschnig Following @kroening's comment here, I believe the correct thing to do would be to specify this running on the ubuntu:18.04 docker image via the [container](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer) attribute, then we don't have to remove packages which seems slightly sketchy anyway.

TBH I didn't realise that the actions images had so much nonstandard stuff in them when we came up with these actions.

tautschnig:
> @tautschnig Following @kroening's comment here, I believe the correct thing to do would be to specify this running on the ubuntu:18.04 docker image via the [container](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer) attribute, then we don't have to remove packages which seems slightly sketchy anyway.
[...]

I tried this avenue, but then our use of submodules requires git 2.18, which isn't in Ubuntu 18.04. Thus sticking with the force-downgrade option for now. Marking the PR as ready-for-review as the logs confirmed that libgcc-s1 no longer is among the dependencies being generated.

NlightNFotis (approved):
LGTM

thomasspriggs:
> > @tautschnig Following @kroening's comment here, I believe the correct thing to do would be to specify this running on the ubuntu:18.04 docker image via the [container](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer) attribute, then we don't have to remove packages which seems slightly sketchy anyway.
> > [...]
> 
> I tried this avenue, but then our use of submodules requires git 2.18, which isn't in Ubuntu 18.04. Thus sticking with the force-downgrade option for now. Marking the PR as ready-for-review as the logs confirmed that libgcc-s1 no longer is among the dependencies being generated.

Can you further explain what aspects of submodules aren't working with git 2.17? I am using ubuntu 18.04 locally and I haven't seen any submodule related issues.

tautschnig:
> > > @tautschnig Following @kroening's comment here, I believe the correct thing to do would be to specify this running on the ubuntu:18.04 docker image via the [container](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer) attribute, then we don't have to remove packages which seems slightly sketchy anyway.
> > > [...]
> > 
> > 
> > I tried this avenue, but then our use of submodules requires git 2.18, which isn't in Ubuntu 18.04. Thus sticking with the force-downgrade option for now. Marking the PR as ready-for-review as the logs confirmed that libgcc-s1 no longer is among the dependencies being generated.
> 
> Can you further explain what aspects of submodules aren't working with git 2.17? I am using ubuntu 18.04 locally and I haven't seen any submodule related issues.

See https://github.com/diffblue/cbmc/runs/2021105147 for one of the past logs. That says:
```
The repository will be downloaded using the GitHub REST API
To create a local Git repository instead, add Git 2.18 or higher to the PATH
Error: Input 'submodules' not supported when falling back to download using the GitHub REST API. To create a local Git repository instead, add Git 2.18 or higher to the PATH.
```

hannes-steffenhagen-diffblue:
@tautschnig That seems like an issue with the github action we use for fetching the code. I don't think that's a real blocker for building on 18.04 proper, worst case we can just work around that.

tautschnig:
> @tautschnig That seems like an issue with the github action we use for fetching the code. I don't think that's a real blocker for building on 18.04 proper, worst case we can just work around that.

Up to you. Can I suggest we put a fix in place before merging #5883 today? This might be merging this PR and doing a cleaner solution later on, or someone making the container solution work today.

hannes-steffenhagen-diffblue:
@tautschnig seems good to me, plan is to merge this and put building in docker in the mid-term

hannes-steffenhagen-diffblue (approved):

PR #5879, opened 2021-03-01
New expression: count_leading_zeros_exprt

tautschnig:
Rather than ad-hoc handling __builtin_clz and __lzcnt (and their
variants) in the C front-end, make counting-leading-zeros available
across the code base.


src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2689,27 +2689,22 @@ exprt c_typecheck_baset::do_special_functions(
...
 
     if(
-      !int_constant.has_value() || *int_constant == 0 ||
-      argument.type().id() != ID_unsignedbv)
+      config.ansi_c.mode == configt::ansi_ct::flavourt::GCC ||
+      config.ansi_c.mode == configt::ansi_ct::flavourt::CLANG)
     {
-      return nil_exprt{};
+      simplify(clz, *this);
     }
```
kroening: I wouldn't simplify these up-front -- the simplifier is called by the front-end whenever a constant is needed.

src/goto-programs/goto_clean_expr.cpp
```diff
@@ -429,6 +438,22 @@ void goto_convertt::clean_expr(
...
     expr = to_unary_expr(expr).op();
   }
+  else if(expr.id() == ID_count_leading_zeros)
+  {
+    count_leading_zeros_exprt &clz = to_count_leading_zeros_expr(expr);
+    if(!clz.zero_permitted())
+    {
+      clz.zero_permitted(true);
+      const source_locationt source_location = clz.source_location();
+
```
kroening: I would go stronger than the nondet: The wording in the gcc docs is that __builtin_clz is undefined when the argument is zero. Assuming they use the meaning of "undefined" as in the ISO C standard, then this is worth an assertion, to be generated in goto_check.

src/ansi-c/c_typecheck_expr.cpp
tautschnig: Ah, true, will fix.

src/goto-programs/goto_clean_expr.cpp
tautschnig: Hmm, true, then I could perhaps demote the `ID_zero_permitted` to `ID_C_zero_permitted`?

kroening: I don't think you need it at all. We don't have anything equivalent for any of the other expressions that are only partially defined.

tautschnig: But we need to distinguish whether we're handling MSVC's `__lzcnt` or `__builtin_clz`. The former is defined for all values, the latter isn't defined for `0`.

kroening: Ah, right. Then I'd make it `ID_zero_permitted`; a comment would be too weak, given that the semantics do differ.

tautschnig: Except they only sort-of do: it seems that in practice it's almost always ok, it's just that GCC doesn't provide a guarantee: https://stackoverflow.com/questions/19527897/how-undefined-are-builtin-ctz0-or-builtin-clz0. How about the solution that I just pushed?

martin-cs (approved):
Seems reasonable if we want to go that way.

kroening (approved):

PR #5880, opened 2021-03-02
VSD - Cross-representation expression evaluation

jezhiggins:
Enable the variable-sensitivity domain to evaluate expressions involving different value representations. 

For example, this enables evaluation of the expression 
  `a + b` 
where 
  `a` is a constant 3
  `b` is a value-set { 2, 4, 6 }
returning a new value-set of { 5, 7, 9 }

- any are value sets, iterate over all of the combinations and dispatch for each of these (producing a value set to contain all of the results)
- mixed constant and interval, convert all constants to intervals and handle as all intervals.
- all interval ao, have to do interval arithmetic
- all constant ao, can simply substitute (if not top) and simplify to get the constant.

All existing behaviour is retained - this is new additional behaviour. 

----
Rather than each representation having its own `expression_transform` everything is now routed through `abstract_value_objectt::expression_transform`, which then dispatches according to the operand representations. For all constants, and all interval expressions, the implementation is essentially identical `constant_abstract_valuet::expression_transform` and `interval_abstract_valuet::expression_transform` . The value_set evaluation is broadly similar to what we had before, building each combination of operands. Each representation now has a `value_range`, and the value_set evaluation uses this to build the combinations in a representation agnostic way. 

The focus of this PR has been on the interaction of value-sets and constants. There will be follow-up PR focussing on intervals - this may only be additional test-cases, but in any case will be rather less substantial. 

The commit where the action happens is https://github.com/diffblue/cbmc/pull/5880/commits/8329afd4c0980576ede3573030c06985598f68a6 Everything before that is essentially preparatory, everything after is refactoring enabled by this change.


martin-cs (approved):
First off, thank you for this large and impressive refactoring.  I've been through each of the commits and they all seem to be doing the right kind of thing.  Given the issues with the existing code and the scale of the restructuring I favour merging this and then I will do a review over all of VSD and pick up any issues that remain.  I think that is the most efficient way forward.

PR #5881, opened 2021-03-02
Implement __builtin_ctz{,l,ll} via count_trailing_zeros_exprt

tautschnig:
Marked as Draft as it depends on #5912.

__builtin_ctz returns the number of trailing zeros. Just like ffs and clz, introduce a new bitvector expression.

### DIFF
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -174,7 +174,7 @@ protected:
 
   void bounds_check(const exprt &, const guardt &);
   void bounds_check_index(const index_exprt &, const guardt &);
-  void bounds_check_clz(const count_leading_zeros_exprt &, const guardt &);
+  void bounds_check_bit_count(const unary_exprt &, const guardt &);
   void div_by_zero_check(const div_exprt &, const guardt &);
   void mod_by_zero_check(const mod_exprt &, const guardt &);
   void mod_overflow_check(const mod_exprt &, const guardt &);
@@ -1336,8 +1336,11 @@ void goto_checkt::bounds_check(const exprt &expr, const guardt &guard)
 
   if(expr.id() == ID_index)
     bounds_check_index(to_index_expr(expr), guard);
-  else if(expr.id() == ID_count_leading_zeros)
-    bounds_check_clz(to_count_leading_zeros_expr(expr), guard);
+  else if(
+    expr.id() == ID_count_leading_zeros || expr.id() == ID_count_trailing_zeros)
+  {
+    bounds_check_bit_count(to_unary_expr(expr), guard);
+  }
 }
 
 void goto_checkt::bounds_check_index(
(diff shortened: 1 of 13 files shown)


martin-cs:
How does this relate to https://github.com/diffblue/cbmc/pull/5879 ?

tautschnig:
> How does this relate to #5879 ?

There is a code overlap in `goto_checkt` that in principle I could resolve to make them entirely independent, but I'm not too keen on first effectively creating a future merge conflict and later resolving that.

src/analyses/goto_check.cpp
```diff
@@ -1519,13 +1522,22 @@ void goto_checkt::bounds_check_index(
...
   const guardt &guard)
 {
+  std::string name;
+
+  if(expr.id() == ID_count_leading_zeros)
+    name = "leading";
+  else if(expr.id() == ID_count_trailing_zeros)
+    name = "trailing";
+  else
+    UNREACHABLE;
```
kroening: I would have made that `PRECONDITION(false)`

src/analyses/goto_check.cpp
```diff
@@ -1519,13 +1522,22 @@ void goto_checkt::bounds_check_index(
...
+
   add_guarded_property(
     notequal_exprt{expr.op(), from_integer(0, expr.op().type())},
-    "count leading zeros argument",
+    "count " + name + " zeros argument",
     "bit count",
```
kroening, line 1541: It may be worth being a bit more verbose: it won't be obvious to a user that the argument of these two functions must not be zero.

src/util/simplify_expr.cpp
```diff
@@ -172,6 +172,28 @@ simplify_exprt::simplify_clz(const count_leading_zeros_exprt &expr)
   return from_integer(n_leading_zeros, expr.type());
 }
 
+simplify_exprt::resultt<>
+simplify_exprt::simplify_ctz(const count_trailing_zeros_exprt &expr)
+{
+  const auto const_bits_opt = expr2bits(
```
kroening, line 178: Does it make sense to use the lowering in the simplifier to avoid the duplicate implementation?

tautschnig: We could, but it's probably more expensive. Whether that's a practical concern, however, is a different question. I suppose ctz will be fairly rare.

src/analyses/goto_check.cpp
tautschnig: Done.

src/analyses/goto_check.cpp
tautschnig: I've made it more verbose, would you mind taking another look?

kroening (approved):

src/util/bitvector_expr.h
```diff
@@ -915,4 +915,97 @@ inline count_leading_zeros_exprt &to_count_leading_zeros_expr(exprt &expr)
   return ret;
 }
 
+/// \brief The count trailing zeros (counting the number of zero bits starting
+/// from the least-significant bit) expression. When \c zero_permitted is set to
+/// false, goto_checkt must generate an assertion that the operand does not
+/// evaluates to zero. The result is always defined, even for zero (where the
```
peterschrammel:
```suggestion
/// evaluate to zero. The result is always defined, even for zero (where the
```

peterschrammel (approved):

PR #5884, opened 2021-03-04
Enforce __CPROVER_loop_invariant contracts in goto-instrument

### LINKED ISSUE #5882: Loop invariant contracts are ignored
**CBMC version:** 5.24.0 (also applies to `develop` @ 7f998c4d)
**Operating system:** Mac OS

```
goto-cc test.c -o contracts.goto
goto-instrument --enforce-all-contracts contracts.goto asserts.goto
goto-instrument --show-goto-functions asserts.goto
```

with `test.c` being:

```c
void main()
{
    unsigned i, n;
    __CPROVER_assume(i == 0);
    while (i < n)
      __CPROVER_loop_invariant(0 <= i && i <= n)
    {
        i++;
    }
    __CPROVER_assert(i == n, "");
}
```

I expect the `__CPROVER_loop_invariant` contract to be compiled to a bunch of assertions, and those to show up in the output of `--show-goto-functions`.

**What happened instead:**
The invariant contract is simply ignored -- no new assertions are generated in the goto binary.

**More information:**
A quick scan of the [src/goto-instrument/code_contracts.cpp](https://github.com/diffblue/cbmc/blob/develop/src/goto-instrument/code_contracts.cpp) file reveals that:
- the invariant contracts are processed within `check_apply_invariants`
- `check_apply_invariants` is called from `code_contracts`
- `code_contracts` is never called within the project
- in particular, `code_contracts` is unreachable from the entry point `enforce_contracts` for the `--enforce-all-contracts` flag

SaswatPadhi:
This PR resolves #5882.

It enforces `__CPROVER_loop_invariant` contract (in addition to function contracts) when `--enforce-all-contracts` or `--enforce-contract <fun>` flag is used with `goto-instrument`.

NOTE: The `code_contracts.h` and `code_contracts.cpp` files need a thorough clean up -- the return types seem inconsistent (`void` vs `bool`) for many functions that have similar behavior, and several functions don't use the C++11 range-based for loops etc. I would make a separate PR later for refactoring and cleanup. This PR only adds the invariant checking back into the contracts flags and enables related regression tests.

- [x] The feature or user visible behaviour. I have added or modified has been documented in the User Guide in doc/cprover-manual/


regression/contracts/invar_check_01/test.desc
```diff
@@ -1,11 +1,7 @@
-KNOWNBUG
+CORE
 main.c
---apply-code-contracts
+--enforce-all-contracts
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
```
feliperodri: Could you add a small explanation about what feature this test case covers? The same goes to all others that you updated.

tautschnig (approved):

tautschnig:
[...]
> I would make a separate PR later for refactoring and cleanup. This PR only adds the invariant checking back into the contracts flags and enables related regression tests.

@SaswatPadhi Please make sure you sync with @feliperodri on this to avoid duplicate work.

regression/contracts/invar_loop_constant/test.desc
```diff
@@ -1,9 +1,10 @@
...
 main.c
---check-code-contracts
+--enforce-all-contracts
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
 --
-Loop invariant checking throws away more information than needed.
+The current implementation of `check_apply_invariants` function
```
feliperodri: What is the exact behavior you expect for this test case?

feliperodri (changes requested):
I'd aslo recommend you to break down the PR into at least two commits: one for the changes in `code_contracts.*` files and another for the updates in the regression tests.

regression/contracts/invar_check_01/test.desc
SaswatPadhi: Thanks for pointing this out! There were comments in the corresponding `.c` files, so I just moved those (with minor edits) to these `.desc` files.

regression/contracts/invar_loop_constant/test.desc
SaswatPadhi: I clarified the description for this test and added a couple more related tests to demonstrate the expected behavior.

feliperodri (approved):

regression/contracts/invar_check_break_pass/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
```
martin-cs, line 9: Is there a reason why you don't check the individual VCs as well as the over-all status?

regression/contracts/invar_check_continue/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
```
martin-cs, line 9: Again.  I prefer to check VCs as well as it catches bugs that cause VCs to be omitted.

regression/contracts/invar_check_large/test.desc
```diff
@@ -0,0 +1,11 @@
...
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+This test checks the invariant contracts on a larger problem --- in this case,
+the partition function of quicksort, applied to a fixed-length array.
+This serves as a stop-gap test until issues to do with quantifiers and
+side-effects in loop invariants are fixed.
```
martin-cs: New line?

martin-cs (approved):
Seems good.  Comments are only minor improvements.  Thanks for the contribution.

src/goto-instrument/code_contracts.cpp
```diff
@@ -158,7 +158,7 @@ bool code_contractst::has_contract(const irep_idt fun_name)
          type.find(ID_C_spec_ensures).is_not_nil();
 }
 
-bool code_contractst::apply_contract(
+bool code_contractst::apply_function_contract(
```
hannes-steffenhagen-diffblue, line 158: Comment applies to: d20bbe127aa03d8fd8d506c8c43f80871be850ce

Having this refactor take place in the same commit as other changes makes this harder to follow than necessary. This would be easier to follow if, for example, renamings happened in one commit (with an explanation for why the names are better unless obvious), one commit for replacing the loops with ranged-for (don't need an explanation for that beyond "ranged-for is better"), and then finally the actually functional change.

Reason for this is that refactorings should _not_ cause functional changes, whereas any bugfixes or features you add of course _should_ change the functionality, and it makes it a bit hard to see which is which if they're all combined in one changeset.

My preference would be if this commit was split up to that effect, although this is not a hard blocker. Worth considering for the future though!

src/goto-instrument/code_contracts.cpp
```diff
@@ -299,24 +299,16 @@ bool code_contractst::apply_contract(
...
-      l_it != natural_loops.loop_map.end();
-      l_it++)
+  for(const auto &loop : natural_loops.loop_map)
     check_apply_invariants(
-      goto_function, local_may_alias, l_it->first, l_it->second);
-
-  // look at all function calls
-  Forall_goto_program_instructions(ins, goto_function.body)
-    if(ins->is_function_call())
-      apply_contract(goto_function.body, ins);
```
hannes-steffenhagen-diffblue, line 319: Still d20bbe1

Case in point: In this case it's a bit hard for me to see why this loop was removed. It's combined with a fairly straightforward refactoring of the preceding loop so it's difficult for me see if this was an intentional change or a mistake that snuck in during refactoring.

hannes-steffenhagen-diffblue (approved):
I'm ok with the actual changes, clang-format weirdness can be resolved (see my reply to your slack message), would prefer if the first commit was split up further before merging but not a blocker.

regression/contracts/invar_check_continue/test.desc
SaswatPadhi: Good idea! I would add VC checks to all tests. Thanks!

src/goto-instrument/code_contracts.cpp
SaswatPadhi: This particular function was perhaps the old entry point for the old contracts flag. It was used to replace loops with invariant contracts and function calls with function contracts. I removed this loop intentionally because the replacement of calls with contracts now happens in `replace_calls`. You can see a very similar loop in `replace_calls`:

https://github.com/diffblue/cbmc/blob/947fb8fc11d85c21593c3947701babef78b48b45/src/goto-instrument/code_contracts.cpp#L894-L916

SaswatPadhi:
Thanks for the review, @martin-cs and @hannes-steffenhagen-diffblue. I think I have addressed all your comments now. Please take another look and let me know if I should make any more changes.

src/goto-instrument/code_contracts.cpp
SaswatPadhi: Thanks Hannes! This is a great suggestion. I just pushed this change.

martin-cs (approved):
Yep; looks good.

martin-cs:
I realise this isn't really the right place to ask this but... if you are considering a reworking / rewrite of the code contracts stuff:

1. Is there any chance you could look at / incorporate / test / obsolete the k-induction code in goto-instrument?  There were tests for this somewhere but I don't know if these are enabled.
2. Could it get added to the manual?  I think we are close to having one of the best deductive verification tools for C ( and with @danielsn and @tautschnig 's work on s2n, one of the best used ) but no-one outside would know (I am aware that `goto-analyze` also needs documentation).

SaswatPadhi:
Hi @martin-cs,

1. Yes sure, that sounds very related to what I am working on, so I should look at the k-induction code as well. I will take a look at the code and tests next week.

2. Good idea! @feliperodri is maintaining an internal Wiki page documenting the contracts stuff, but we should add these to the public manuals too. Could you please point us to where we should be adding them?

martin-cs:
@SaswatPadhi 

1. `src/goto-instrument/k-induction.{h,cpp}` and `regression/k-induction` but it is not routinely run so who knows if it still works!
2. `doc/cprover-manual` would be good.

PR #5890, opened 2021-03-07
Fully simplify (in)equalities over constants+if

tautschnig:
The case of constant-on-lhs together with an if-expression on the
right-hand side previously required multiple simplification passes as
the constant-on-both-sides case was not considered for this case.

### DIFF
--- a/unit/util/simplify_expr.cpp
+++ b/unit/util/simplify_expr.cpp
@@ -516,3 +516,27 @@ TEST_CASE("Simplify bitor", "[core][util]")
     REQUIRE(simplify_expr(bitor_exprt{b, zero}, ns) == b);
   }
 }
+
+TEST_CASE("Simplify inequality", "[core][util]")
+{
+  symbol_tablet symbol_table;
+  namespacet ns(symbol_table);
+
+  {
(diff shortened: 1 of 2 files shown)


martin-cs (approved):
Consistency w.r.t. LHS and RHS is good.  I worry a little that this could give exponential blow up but there are a number of concerns of this sort with this code.

peterschrammel (approved):

unit/util/simplify_expr.cpp
```diff
@@ -516,3 +516,27 @@ TEST_CASE("Simplify bitor", "[core][util]")
...
 }
+
+TEST_CASE("Simplify inequality", "[core][util]")
+{
+  symbol_tablet symbol_table;
+  namespacet ns(symbol_table);
+
+  {
+    // This checks that 3 < (B ? 4 : 5) simplifies to true, just like (B ? 4 :
+    // 5) > 5 simplifies to true.
```
TGWDB: (B ? 4 : 5) > 5 simplifies to false?

tautschnig: Thank you for catching this typo in the comment! (It should have said/now says "> 3")

PR #5892, opened 2021-03-07
Simplify (T)(a?b:c) to a?(T)b:(T)c

tautschnig:
The idea of such a rule existed ever since 7285dda, but was possibly too
costly when doing it post-order for one would have to re-process
multiple sub-expressions. A pre-order rule avoids such repeat
processing.

### DIFF
--- a/src/util/simplify_expr.cpp
+++ b/src/util/simplify_expr.cpp
@@ -1007,21 +1007,6 @@ simplify_exprt::simplify_typecast(const typecast_exprt &expr)
     }
   }
 
-  #if 0
-  // (T)(a?b:c) --> a?(T)b:(T)c
-  if(expr.op().id()==ID_if &&
-     expr.op().operands().size()==3)
-  {
-    typecast_exprt tmp_op1(expr.op().op1(), expr_type);
-    typecast_exprt tmp_op2(expr.op().op2(), expr_type);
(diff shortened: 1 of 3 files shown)


martin-cs (approved):
I am a little concerned about the cost of this, esp. w.r.t. non-trivial casts like `double`->`float` or `int`->`double`.  Some things like integer reduction are definitely worth doing.  Is it worth restricting what types this is applied to?

tautschnig:
> I am a little concerned about the cost of this, esp. w.r.t. non-trivial casts like `double`->`float` or `int`->`double`. Some things like integer reduction are definitely worth doing. Is it worth restricting what types this is applied to?

Done.

TGWDB (approved):

martin-cs (approved):
I think that will avoid the most expensive cases.  Thanks.

kroening:
Do we have any means to performance-benchmark?

tautschnig:
@kroening:
> Do we have any means to performance-benchmark?

I have run all of SV-COMP's benchmarks on this branch and compared it to develop. No difference beyond the expected jitter was observable. Therefore, I'd argue that this simplification step can in some cases permit additional constant propagation without adversely affecting other verification tasks that don't contain such expressions.

peterschrammel (approved):

PR #5893, opened 2021-03-08
Adds support for function calls within contracts

feliperodri:
Contracts definitions often makes use of elaborate predicates,
which could be written on a separate functions. Therefore,
function contracts should be able to accept functions
as part of their specifications.

### DIFF
--- a/src/linking/remove_internal_symbols.cpp
+++ b/src/linking/remove_internal_symbols.cpp
@@ -57,6 +57,30 @@ static void get_symbols(
         if(!ns.lookup(p.get_identifier(), s))
           working_set.push_back(s);
       }
+
+      // check for contract definitions
+      const exprt ensures =
+        static_cast<const exprt &>(code_type.find(ID_C_spec_ensures));
+      const exprt requires =
+        static_cast<const exprt &>(code_type.find(ID_C_spec_requires));
+
+      find_symbols_sett new_symbols;
+      find_type_and_expr_symbols(ensures, new_symbols);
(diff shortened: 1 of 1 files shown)


martin-cs (approved):
This seems a reasonable and necessary step to get this working.  I don't know the contract code as well as I should.  Is there an option to ignore / disregard / drop the contracts?  If so, does this add significant extra overhead?  My feeling is probably not but I am not sure.

feliperodri:
> This seems a reasonable and necessary step to get this working. I don't know the contract code as well as I should. Is there an option to ignore / disregard / drop the contracts? If so, does this add significant extra overhead? My feeling is probably not but I am not sure.

Contracts are only added if you run `goto-instrument` with one of the following options: 
```shell
Function contracts and invariants:
 --replace-call-with-contract <fun>
                              replace calls to fun with fun's contract
 --replace-all-calls-with-contracts
                              as above for all functions with a contract
 --enforce-contract <fun>     wrap fun with an assertion of its contract
 --enforce-all-contracts      as above for all functions with a contract
```

So you can choose which functions should use contracts or if they should be applied to all of them. I guess this confirms that there won't be any extra significant overhead.  Did I answer your question, @martin-cs ?

TGWDB (approved):

NlightNFotis (approved):
LGTM.

chrisr-diffblue (approved):
LGTM, thanks.

martin-cs:
@feliperodri yes and sorry for not getting back to you sooner.  One minor concern -- what happens if there is recursion via the contracts?

PR #5897, opened 2021-03-08
Add {binary,unary_overflow_exprt}

tautschnig:
Create a dedicated exprt to construct overflow expressions for improved
type checking.

### DIFF
--- a/src/goto-instrument/accelerate/overflow_instrumenter.cpp
+++ b/src/goto-instrument/accelerate/overflow_instrumenter.cpp
@@ -13,10 +13,11 @@ Author: Matt Lewis
 
 #include <iostream>
 
-#include <util/std_expr.h>
-#include <util/std_code.h>
 #include <util/arith_tools.h>
+#include <util/bitvector_expr.h>
 #include <util/simplify_expr.h>
+#include <util/std_code.h>
+#include <util/std_expr.h>
 
 #include <goto-programs/goto_program.h>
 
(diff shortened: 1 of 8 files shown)


src/util/bitvector_expr.h
```diff
@@ -677,4 +677,146 @@ inline popcount_exprt &to_popcount_expr(exprt &expr)
   return ret;
 }
 
+/// \brief A Boolean expression returning true, iff operation \c kind would
+/// result in an overflow when applied to operands \c lhs and \c rhs.
+class binary_overflow_exprt : public binary_exprt
+{
```
kroening, line 683: You could make it a `binary_predicate_exprt`

src/util/bitvector_expr.h
```diff
@@ -677,4 +677,146 @@ inline popcount_exprt &to_popcount_expr(exprt &expr)
...
+    expr.id() == ID_overflow_minus || expr.id() == ID_overflow_shl);
+  binary_overflow_exprt &ret = static_cast<binary_overflow_exprt &>(expr);
+  validate_expr(ret);
+  return ret;
+}
+
+/// \brief A Boolean expression returning true, iff operation \c kind would
+/// result in an overflow when applied to the (single) operand.
+class unary_overflow_exprt : public unary_exprt
+{
```
kroening, line 762: Likewise, `unary_predicate_exprt`

kroening (approved):

src/util/bitvector_expr.h
tautschnig: Done.

src/util/bitvector_expr.h
tautschnig: Done.

src/goto-instrument/accelerate/overflow_instrumenter.cpp
```diff
@@ -211,38 +212,25 @@ void overflow_instrumentert::overflow_expr(
           expr.id()==ID_mult)
   {
     // A generic arithmetic operation.
-    multi_ary_exprt overflow(
-      "overflow-" + expr.id_string(), expr.operands(), bool_typet());
-
-    if(expr.operands().size()>=3)
+    // The overflow checks are binary.
+    for(std::size_t i = 1; i < expr.operands().size(); i++)
```
martin-cs, line 216: This is not a request for this PR but more a general observation that this code seems to be not covered by the regression tests.  I have a feeling that at one point Matt had tests for this code.  Maybe @cristina-david could help?

martin-cs (approved):
Yeah... this has the feel of "why didn't we always do it like this" which is a good sign.

I hate to be that person but ... what about divide? Specifically `INT_MIN / -1`.

martin-cs:
I'm also having horrible thoughts about what happens with signs when taking modulo.  Is it possible to get an overflow with INT_{MAX,MIN} % INT_{MAX,MIN} ?

src/goto-instrument/accelerate/overflow_instrumenter.cpp
tautschnig: We have regression/acceleration, it just seems the script in there needs some love to make this work. Adding this to my to-do list.

tautschnig:
> I'm also having horrible thoughts about what happens with signs when taking modulo. Is it possible to get an overflow with INT_{MAX,MIN} % INT_{MAX,MIN} ?

We do generate some of these checks in `goto_checkt`, though not using `overflow_*` expressions, but by specifically testing for boundary values. I think that's the right approach, but I do wonder whether we have the necessary coverage for floating-point operands for all of these...

PR #5900, opened 2021-03-09
Add side_effect_expr_overflowt

tautschnig:
Introduce a new expression to handle GCC's __builtin_X_overflow in the
middle-end (during goto conversion) instead of directly encoding the
semantics during type checking.

This commit is a refactoring only, it does not yet fix the bugs that #5544 intends to fix. #5544 will now be rebased on top of this commit.

### DIFF
--- a/src/ansi-c/c_expr.h
+++ b/src/ansi-c/c_expr.h
@@ -107,4 +107,97 @@ inline shuffle_vector_exprt &to_shuffle_vector_expr(exprt &expr)
   return ret;
 }
 
+/// \brief A Boolean expression returning true, iff the result of performing
+/// operation \c kind on operands \c lhs and \c rhs in infinite-precision
+/// arithmetic cannot be represented in the type of the object that \c result
+/// points to.
+/// If \c result is a pointer, the result of the operation is stored in the
+/// object pointed to by \c result.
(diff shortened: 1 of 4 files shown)


src/util/std_code.h
```diff
@@ -2561,4 +2561,96 @@ inline code_function_bodyt &to_code_function_body(codet &code)
...
+/// If \c result is a pointer, the result of the operation is stored in the
+/// object pointed to by \c result.
+class side_effect_expr_overflowt : public side_effect_exprt
+{
+public:
+  side_effect_expr_overflowt(
+    const irep_idt &kind,
+    exprt _a,
+    exprt _b,
+    exprt _r,
```
TGWDB: Longer/prettier names might be nice here (and for the access functions below), e.g. "lhs", "rhs", "result".

TGWDB (approved):

src/util/std_code.h
tautschnig: I was initially not convinced that "lhs", "rhs" would make sense here, but when considered with respect to the underlying arithmetic operation they certainly do. Changed as suggested.

kroening:
Same here now -- consider ansi-c/c_expr.h

kroening (approved):

PR #5904, opened 2021-03-09
Fix a crash during SMT translation of structs

SaswatPadhi:
This PR fixes a bug with SMT translation of structs using data types (not bit vectors). CBMC `--smt2 --z3` was crashing on the attached regression test with a `key not found` exception from a `map::at` call.

### DIFF
--- a/regression/cbmc/struct15/test.desc
+++ b/regression/cbmc/struct15/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+--trace --z3
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+^map::at:
+key not found
+--
+This test checks the encoding of C `struct`s using SMT2 data types.
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -4890,8 +4890,10 @@ void smt2_convt::find_symbols_rec(
 
     if(recstack.find(id) == recstack.end())
     {
+      const auto &base_struct = ns.follow_tag(struct_tag);
       recstack.insert(id);
-      find_symbols_rec(ns.follow_tag(struct_tag), recstack);
+      find_symbols_rec(base_struct, recstack);
+      datatype_map[type] = datatype_map[base_struct];
     }
   }
   else if(type.id() == ID_union_tag)
--- a/regression/cbmc/struct15/main.c
+++ b/regression/cbmc/struct15/main.c
@@ -1,0 +1,14 @@
+#include <assert.h>
+
+struct test
+{
+  unsigned int a;
+  unsigned int b;
+};
+
+int main()
+{
+  struct test t;
+  if(t.a > 10)
+    assert(t.a > 0);
+}
--- a/src/solvers/smt2/smt2_conv.h
+++ b/src/solvers/smt2/smt2_conv.h
@@ -209,8 +209,10 @@ protected:
 
   identifier_mapt identifier_map;
 
-  // for modeling structs as Z3 datatype, enabled when
-  // use_datatype is set
+  // for modeling structs as SMT datatype when use_datatype is set
+  //
+  // it maintains a map of `struct_typet` or `struct_tag_typet`
+  // to datatype names declared in SMT
   typedef std::map<typet, std::string> datatype_mapt;
   datatype_mapt datatype_map;
 


regression/cbmc-smt2-z3/Makefile
```diff
@@ -0,0 +1,19 @@
+default: tests.log
+
+include ../../src/config.inc
+include ../../src/common
+
+test: ../test.pl
+	@../test.pl -e -p -c "../../../src/cbmc/cbmc --smt2 --z3"
```
martin-cs: Rather than adding a separate directory, can you tag the test.desc as `smt-backend` as is done in `cbmc/integer-assignments1/test.desc`

regression/cbmc-smt2-z3/struct/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+
+^EXIT=6$
+^SIGNAL=0$
+^VERIFICATION ERROR$
```
martin-cs: If this fixes the translation, why is the test case giving an error?

martin-cs (changes requested):
Please can you reformat this a little to fix with current conventions.

regression/cbmc-smt2-z3/struct/test.desc
SaswatPadhi: Oh this is because the assertion doesn't hold.

With a trivial property, CBMC was just simplifying it to `true` and was not even generating SMT queries. I can try to check a different property that won't be trivially simplified to true and would actually require an SMT solver. Or, can I disable all optimizations using some flags?

tautschnig: If the assertion doesn't hold, then the result should be `VERIFICATION FAILED`, not `VERIFICATION ERROR`?!

To disable constant propagation and/or simplification: use `--no-propagation` and `--no-simplify`, respectively.

kroening:
Should there be a similar change for union_tags?

regression/cbmc-smt2-z3/struct/test.desc
kroening: I would avoid either one of these options unless they are specifically tested.
If it bugs you that this is a satisfiable instance (which it shouldn't) then you could use s.th. like
`if(t.a>10) assert(t.a > 0);`

SaswatPadhi:
> Should there be a similar change for union_tags?

Unions are currently not encoded using data types in SMT, so this particular issue doesn't appear with unions.

But I haven't checked if we would have a similar issue with non-DT encoding of unions and structs. We do something similar for bit vector encodings -- instead of looking up the `datatype_map`, we look up `boolbv_width` of the type. I will check this and create a separate PR to fix the BV encoding, if needed.

regression/cbmc-smt2-z3/Makefile
SaswatPadhi: Hi @martin-cs,

I just tagged it with `smt-backend` and ran locally, but I see the `smt-backend` tests are run with `--cprover-smt2` not with `--smt2`.

This regression test was crashing earlier with `cbmc --smt2` and now it runs as expected with that flag. However, the test does not pass (gives `VERIFICATION ERROR`) with `cbmc --cprover-smt2` -- same without the changes introduced by this PR.

So, I am not sure what's the best way to tag this test. I have kept it as `CORE` but have added the appropriate flags in `test.desc`. What do you suggest?

regression/cbmc-smt2-z3/struct/test.desc
SaswatPadhi: Thanks everyone! I implemented Daniel's suggestion above. Marking this one as resolved.

SaswatPadhi:
I think the test is failing in CI because it cannot find `z3`.

Is there a way we can just test the SMT constraint generation with data types enabled, without actually solving them using z3? Alternatively, we can install z3 on our CI :)

thomasspriggs:
> I think the test is failing in CI because it cannot find `z3`.
> 
> Is there a way we can just test the SMT constraint generation with data types enabled, without actually solving them using z3? Alternatively, we can install z3 on our CI :)

It seems reasonable to install an SMT solver in CI in order to test the SMT output properly. Just so long as we don't end up giving it anything to solve which inflates the overall run time too much. @diffblue/diffblue-opensource What do the rest of you think?

Please note that I haven't reviewed the rest of this PR, I am just trying to help with this particular issue.

SaswatPadhi:
Thanks @thomasspriggs.

@martin-cs I think I have addressed your comment regarding conventions. There is one remaining issue with `smt-backend` tag which I have described above. Could you please take another look?

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -4869,8 +4869,10 @@ void smt2_convt::find_symbols_rec(
 
     if(recstack.find(id) == recstack.end())
     {
+      const auto &base_struct = ns.follow_tag(struct_tag);
       recstack.insert(id);
-      find_symbols_rec(ns.follow_tag(struct_tag), recstack);
+      find_symbols_rec(base_struct, recstack);
+      datatype_map[type] = datatype_map[base_struct];
     }
```
thomasspriggs, line 4897: :+1: I have tested this locally and this does indeed seem to fix a failure wherein cbmc prints `map::at` and exits.

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -4869,8 +4869,10 @@ void smt2_convt::find_symbols_rec(
 
     if(recstack.find(id) == recstack.end())
     {
+      const auto &base_struct = ns.follow_tag(struct_tag);
       recstack.insert(id);
-      find_symbols_rec(ns.follow_tag(struct_tag), recstack);
+      find_symbols_rec(base_struct, recstack);
+      datatype_map[type] = datatype_map[base_struct];
```
thomasspriggs, line 4896: :pick: It would be nice to have some more doxygen at the declaration of `datatype_map` covering exactly what it maps to/from. It would also be nice to have better doxygen above `use_datatypes` documenting what changes should be expected in the output file when this is set to true. Obviously not a blocking comment, just wishful thinking.

regression/cbmc/struct15/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+--trace --smt2
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
```
thomasspriggs, line 7: :pick: Might be nice to have a regex in the must not be matched section here checking for `map::at`.

thomasspriggs (commented):
The fix appears reasonable to me. But we need get the test to run and pass in CI, before I will be happy to approve. This so that the additional CI related changes can be reviewed and approved before merging.

src/solvers/smt2/smt2_conv.cpp
SaswatPadhi: Thanks.

src/solvers/smt2/smt2_conv.cpp
SaswatPadhi: I'm adding some comments at `datatype_map` declaration, but I think I could make a separate PR for explaining `use_datatypes` better.

There's some documentation about `use_datatypes` sprinkled throughout `smt2_conv.cpp`, e.g. at https://github.com/diffblue/cbmc/blob/develop/src/solvers/smt2/smt2_conv.cpp#L4775-L4810. Probably we could add some high-level description above its declaration.

regression/cbmc/struct15/test.desc
SaswatPadhi: Good idea, thanks! Pushed this change.

SaswatPadhi:
Blocked on #5949.

thomasspriggs:
> Blocked on #5949.

I only raised that PR to run CI, it is in Draft status and the work in known to be incomplete. We are currently giving higher priority to other pieces of work. It could be a week or so, before we come back to those CI changes. Is that going to be a problem for you?

SaswatPadhi:
Hi @thomasspriggs,

No problem. I was going through all of my PRs and rebased some of them with `develop`. This is just a note to myself. I have things running locally for demos etc. so this is not a high priority for now.

Thanks.

martin-cs (approved):
Looks good; thanks.

regression/cbmc/struct15/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--trace --smt2
```
peterschrammel: I don't think this code is even exercised without `--z3`, see https://github.com/diffblue/cbmc/blob/develop/src/solvers/smt2/smt2_conv.cpp#L101

thomasspriggs: If `--smt2` is specified and a specific solver is not specified, then z3 is selected by default. See https://github.com/diffblue/cbmc/blob/develop/src/cbmc/cbmc_parse_options.cpp#L428 and https://github.com/diffblue/cbmc/blob/develop/src/cbmc/cbmc_parse_options.cpp#L438
That said, I think this would be a more robust test if `--z3` was specified, rather than relying on the default.

SaswatPadhi: Thanks, that's a good point! I will make this change.

PR #5913, opened 2021-03-10
C front-end: __builtin_shuffle, __builtin_shufflevector

tautschnig:
These are polymorphic built-ins supported by GCC and Clang,
respectively. Introduce a new shuffle_vector expression to handle these,
with type checking addressing the differences between the GCC/Clang
APIs.


kroening:
I would put the new expressions into ansi-c/c_expr.h and .cpp -- while it's conceivable that there might be another frontend with expressions like these, it's specific dialects of C for now.

kroening (approved):

PR #5914, opened 2021-03-11
Refactor and optimise size to have approximate bound

TGWDB:
This optimises the size function that was only used when
checking for a bound. The new bounded_size takes an
explicit bound argument and reduces the search when
this bound is (approximately) reached.

Also this fixes a bug in the complexity computation where
the complexity (size) can never be zero, instead check
against one.

### DIFF
--- a/src/util/expr.cpp
+++ b/src/util/expr.cpp
@@ -24,16 +24,28 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <stack>
 
-std::size_t exprt::size() const
+/// Returns the size of the exprt added to count without searching significantly
+/// beyond the supplied limit.
+std::size_t exprt::bounded_size(std::size_t count, std::size_t limit) const
 {
-  // Initial size of 1 to count self.
-  std::size_t size = 1;
-  for(const auto &op : operands())
+  const auto &ops = operands();
+  count += ops.size();
+  for(const auto &op : ops)
   {
-    size += op.size();
+    if(count >= limit)
+    {
+      return count;
+    }
+    count = op.bounded_size(count, limit);
   }
+  return count;
+}
 
-  return size;
+/// Returns the size of the exprt without significantly searching beyond the
+/// supplied limit.
+std::size_t exprt::bounded_size(std::size_t limit) const
+{
+  return bounded_size(1, limit);
 }
 
 /// Return whether the expression is a constant.
--- a/src/util/expr.h
+++ b/src/util/expr.h
@@ -85,9 +85,9 @@ public:
     return static_cast<const typet &>(find(ID_type));
   }
 
-  /// Amount of nodes this expression tree contains. This is the size of the
-  /// actual tree, ignoring memory/sub-tree sharing.
-  std::size_t size() const;
+  /// Amount of nodes in this expression tree approximately bounded by limit.
+  /// This is the size of the actual tree, ignoring memory/sub-tree sharing.
+  std::size_t bounded_size(std::size_t limit) const;
 
   /// Return true if there is at least one operand.
   bool has_operands() const
@@ -123,6 +123,10 @@ protected:
 
   const exprt &op3() const
   { return operands()[3]; }
+
+  /// Amount of nodes this expression tree contains, with a bound on how far
+  /// to search. Starts with an existing count.
+  std::size_t bounded_size(std::size_t count, std::size_t limit) const;
 
 public:
   void reserve_operands(operandst::size_type n)
--- a/unit/util/expr.cpp
+++ b/unit/util/expr.cpp
@@ -36,3 +36,36 @@ SCENARIO("bitfield-expr-is-zero", "[core][util][expr]")
     }
   }
 }
+
+TEST_CASE("Bounded size is computer correctly", "[core][util][expr]")
+{
+  // Helper types and functions for constructing test expressions.
+  const typet type = signedbv_typet(32);
+  std::function<exprt(size_t)> make_expression;
+  make_expression = [&](size_t size) -> exprt {
+    PRECONDITION(size != 0);
+    if(size <= 1)
+      return from_integer(1, type);
+    if(size == 2)
+      return unary_minus_exprt(from_integer(1, type));
+    return plus_exprt(from_integer(1, type), make_expression(size - 2));
+  };
+  // Actual test cases.
+  REQUIRE(make_expression(1).bounded_size(10) == 1);
+  REQUIRE(make_expression(2).bounded_size(10) == 2);
+  REQUIRE(make_expression(3).bounded_size(10) == 3);
+
+  REQUIRE(make_expression(30).bounded_size(10) < 13);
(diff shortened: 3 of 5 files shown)


TGWDB:
@peterschrammel @kroening Any chance of a a code owner review/approval?

peterschrammel (commented):
Are there tests for this?

TGWDB:
> Are there tests for this?

Added unit tests now.

src/util/expr.cpp
```diff
@@ -24,16 +24,28 @@ Author: Daniel Kroening, kroening@kroening.com
...
+/// Returns the size of the exprt added to count without searching significantly
+/// beyond the supplied limit.
+std::size_t exprt::bounded_size(std::size_t count, std::size_t limit) const
 {
-  // Initial size of 1 to count self.
-  std::size_t size = 1;
-  for(const auto &op : operands())
+  auto ops = operands();
+  count += ops.size();
+  for(const auto &op : ops)
```
martin-cs, line 33: In the case of structural sharing this is (unnecessarily) exponential.  Maybe use `visit`?

thomasspriggs: Wouldn't `visit` always do a full traversal, thus removing this optimisation of only doing a partial traversal?

thomasspriggs: @TGWDB Side note, please split this commit into the correctly titled optimisation and bug fix commits. I think the current title may already have mislead Martin. So lets get it sorted.

TGWDB: @martin-cs I'm afraid I don't fully understand your comment. In particular I don't understand the terminology "structural sharing" here. There are a few things that have come up as possible, and some comments addressing them below.

1. There is a concern that `ops` is potentially copied. I've pushed a commit to make it clear that `ops` is never changed and so there should be no issues related to copying of `ops`.
2. There is some question about the exponential nature of the function? The function is depth first and linear in the size of the exprt up to the bound. Note that the `size()` call is a constant for this datatype, and then the (depth first) walk of the exprt is linear.
3. There may be a concern that the depth first walk could potentially overflow the stack with many function calls. This was also a (potential) problem in the old version. The version here should reduce this risk in two ways. Firstly, at each call the entire size of the `operands` is added, thus capturing more of the true size faster and so reducing function calls. Secondly, the bounded nature means this should (particularly when called on large structures) return without traversing the entire exprt.
4. If the concern is writing a custom traversal rather than using a visitor then I guess this could be considered, assuming there is suitable motivation?

unit/util/expr.cpp
```diff
@@ -36,3 +36,35 @@ SCENARIO("bitfield-expr-is-zero", "[core][util][expr]")
...
+  REQUIRE(make_expression(30).bounded_size(10) >= 10);
+}
+
+TEST_CASE("Bounded size versus pathological example", "[core][util][expr]")
+{
+  const typet type = signedbv_typet(32);
+  exprt pathology = plus_exprt(from_integer(1, type), from_integer(1, type));
+  // Create an infinite exprt by self reference!
+  pathology.add_to_operands(pathology);
+  // If bounded size is not checking the bound this will never terminate.
```
thomasspriggs, line 68: :bulb: I suggest adding `CHECK(pathology.bounded_size(10) >= 10)` to check that your pathological example is actually cyclical, rather than just partially copied.

thomasspriggs (approved):
Looks fine to me. @peterschrammel Can you re-review?

src/util/expr.h
```diff
@@ -85,9 +85,8 @@ class exprt:public irept
     return static_cast<const typet &>(find(ID_type));
   }
 
-  /// Amount of nodes this expression tree contains. This is the size of the
-  /// actual tree, ignoring memory/sub-tree sharing.
```
peterschrammel, line 89: I'd keep this comment. This refers what @martin-cs meant by structural sharing.

peterschrammel (approved):
Certainly an improvement (although I don't know any user of this functionality).

unit/util/expr.cpp
TGWDB: Done, thanks.

src/util/expr.h
TGWDB: Added in.

src/util/expr.cpp
martin-cs: @TGWDB CPROVER shares `irept`s where possible.  This can result in exponential compression.  Consider the following expression:

```
let b = a + a
let c = b * b
let d = c + c
let e = d * d
```

CPROVER will store a total of 5 irepts.  Four of them will have the same irept appearing twice in the `sub`.

A naive traversal of the form that you had written will visit `a` 16 times rather than just once.  As the sharing gets larger so the time taken becomes exponential in the size of the data structure rather than linear.  This happens.  It is A Problem.  In the best cases the sharing is linked enough that the pass grinds to a halt and it is obvious that the code needs to be fixed.  Worse is when it is an issue but no-one notices the 8* slow down on the average case of one pass.  Then at some point we feed in code and ... it just times out somewhere random and without careful and detailed profiling (which you almost certainly can't do with customer code) you will not find the problem.

I say this from bitter personal experience.  @tautschnig and @peterschrammel have both done great work on making things sharing aware and sharing conscious.

So, if you want to traverse an expression for an unlimited amount *PLEASE* do not write for loops and recursion.  Use a work queue and a set for what you have seen.  Or use the visitor; that is is what it is there for.

Also, in the specific use-case you want, I think you probably want the number of unique exprts, you don't want to count duplicates.

thomasspriggs: Thank you for this explanation. The details of this particular sharing issue are useful. However I have just been looking at the implementation details of `exprt::visit` and I do not see how it could solve the particular issue you describe. The implementation for this function can be found in `visit_pre_template`. It uses an `std::stack` as its work queue (it also uses a for loop which you object to). `std::stack` is a simple LIFO data structure. If the same `irept` is pushed onto the stack twice, then it will be popped twice and visited twice. It does not offer the duplication removal of `std::set` or `std::unordered_set`. The use of the `std::stack` is still worthwhile here, because storing the working data set on the heap avoids storing it in the call stack as for a recursive implementation. Recursion could overflow the call stack when processing a sufficiently deeply nested data structure. Using the visitor is good, I just don't see how it could offer the particular benefit which you purport that it does. Was there a different visitor implementation which you were thinking of instead?

martin-cs: @thomasspriggs apologies.  There used to be a visitor that used a set to handle sharing.  I would consider this a bug in `exprt::visit`.

thomasspriggs: No problem. I just wanted to get to the bottom of what best practice would actually be.

Changing `exprt::visit` to not visit shared nodes like that could break existing usages of it. Say for example if it has been used for printing code, where we actually want to print the full tree. The `exprt::unique_depth_(begin/end/cbegin/cend)` iterators look to offer the functionality you were expecting. But they also seem to be unused.

PR #5916, opened 2021-03-12
Linux ci - Compile KVM with goto-cc

nmanthey:
This change adds a github workflow that compiles parts of the Linux kernel with goto-cc, and fails in case compilation fails. Furthermore, the input fails that lead to the failure are archived, to allow easy consumption and investigation.

### Note: This is currently failing, and should help to improve goto-cc to be able to compile Linux

An example failed run with the broken input as artifact has been created as part of my testing: https://github.com/nmanthey/cbmc/pull/5/checks?check_run_id=2093209018

- [N/A] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [N/A] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [N/A] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [N/A] My commit message includes data points confirming performance improvements (if claimed).

### DIFF
--- a/.github/workflows/build-and-test-Linux.yaml
+++ b/.github/workflows/build-and-test-Linux.yaml
@@ -1,0 +1,53 @@
+name: Build Linux partially with CPROVER tools
+
+on:
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+  CompileLinux:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Install Packages
+        env:
+          # This is needed in addition to -yq to prevent apt-get from asking for
+          # user input
+          DEBIAN_FRONTEND: noninteractive
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -y g++ flex bison cmake ninja-build maven jq libxml2-utils dpkg-dev ccache
+          sudo apt-get install --no-install-recommends -y libssl-dev libelf-dev libudev-dev libpci-dev libiberty-dev autoconf
+          sudo apt-get install --no-install-recommends -y gawk jq
+
+      - name: Prepare ccache
+        uses: actions/cache@v2
+        with:
+          path: .ccache
+          key: ${{ runner.os }}-20.04-make-${{ github.ref }}-${{ github.sha }}-KERNEL
+          restore-keys: |
+            ${{ runner.os }}-20.04-make-${{ github.ref }}
+            ${{ runner.os }}-20.04-make
+      - name: ccache environment
+        run: |
+          echo "CCACHE_BASEDIR=$PWD" >> $GITHUB_ENV
+          echo "CCACHE_DIR=$PWD/.ccache" >> $GITHUB_ENV
+      - name: Zero ccache stats and limit in size
+        run: ccache -z --max-size=500M
+      - name: Build CBMC tools
+        run: |
+          make -C src minisat2-download
+          make -C src CXX='ccache /usr/bin/g++' cbmc.dir goto-cc.dir goto-diff.dir -j2
+      - name: Print ccache stats
+        run: ccache -s
+
+      - name: Run (Docker Based) Linux Build test
+        run: integration/linux/compile_linux.sh
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: CPROVER-faultyInput
+          path: CPROVER/faultyInput/*
--- a/integration/linux/compile_linux.sh
+++ b/integration/linux/compile_linux.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+# Fail on errors
+# set -e # not for now
+
+# Show steps we execute
+set -x
+
+# This script needs to operate in the root directory of the CBMC repository
+SCRIPT=$(readlink -e "$0")
+readonly SCRIPT
+SCRIPTDIR=$(dirname "$SCRIPT")
+readonly SCRIPTDIR
+cd "$SCRIPTDIR/../.."
+
+# Build CBMC tools
+make -C src minisat2-download
+make -C src CXX='ccache /usr/bin/g++' cbmc.dir goto-cc.dir goto-diff.dir -j$(nproc)
+
+# Get one-line-scan, if we do not have it already
+[ -d one-line-scan ] || git clone https://github.com/awslabs/one-line-scan.git one-line-scan
+
+# Get Linux v5.10, if we do not have it already
+[ -d linux_5_10 ] || git clone -b v5.10 --depth=1 https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git linux_5_10
+
+# Prepare compile a part of the kernel with CBMC via one-line-scan
+ln -s goto-cc src/goto-cc/goto-ld
+ln -s goto-cc src/goto-cc/goto-as
+ln -s goto-cc src/goto-cc/goto-g++
+
+
+configure_linux ()
+{
+  pushd linux_5_10
+
+  cat > kvm-config <<EOF
+CONFIG_64BIT=y
+CONFIG_X86_64=y
+CONFIG_HIGH_RES_TIMERS=y
+CONFIG_MULTIUSER=y
+CONFIG_NET=y
+CONFIG_VIRTUALIZATION=y
+CONFIG_HYPERVISOR_GUEST=y
+CONFIG_PARAVIRT=y
+CONFIG_KVM_GUEST=y
+CONFIG_KVM=y
+CONFIG_KVM_INTEL=y
+CONFIG_KVM_AMD=y
+EOF
+  # use the configuration from the generated file
+  export KCONFIG_ALLCONFIG=kvm-config
+
+  # ... and use it during configuration
+  make allnoconfig
+  popd
+}
+
+# Configure kernel, and compile all of it
+configure_linux
+make -C linux_5_10 bzImage -j $(nproc)
+
+# Clean files we want to be able to re-compile
+find linux_5_10/arch/x86/kvm/ -name "*.o" -delete
+
+# Compile Linux with CBMC via one-line-scan, and check for goto-cc section
+# Re-Compile with goto-cc, and measure disk space
+declare -i STATUS=0
+du -h --max-depth=1
+one-line-scan/one-line-scan \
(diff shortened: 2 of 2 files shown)


.github/workflows/build-and-test-Linux.yaml
```diff
@@ -70,5 +70,10 @@ jobs:
       - name: Check for faulty input
         run: ls CPROVER/faultyInput/* || true
 
+      - uses: actions/upload-artifact@v2
+        with:
+          name: CPROVER-faultyInput
+          path: CPROVER/faultyInput/*
```
tautschnig, line 53: How does one actually download these? GitHub keeps telling me that I need to have the "actions" scope, perhaps the repository permissions are set up in a way that doesn't permit non-members to do this?

.github/workflows/build-and-test-Linux.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Install Packages
+        env:
+          # This is needed in addition to -yq to prevent apt-get from asking for
+          # user input
+          DEBIAN_FRONTEND: noninteractive
+        run: |
+          sudo apt-get install --no-install-recommends -y coreutils  build-essential gcc git make flex bison software-properties-common curl python
+          sudo apt-get install --no-install-recommends -y bin86 gdb bcc liblzma-dev python-dev gettext iasl uuid-dev libncurses5-dev libncursesw5-dev pkg-config
+          sudo apt-get install --no-install-recommends -y libgtk2.0-dev libyajl-dev sudo time ccache
+          sudo apt-get install --no-install-recommends -y automake bc libssl-dev dkms libelf-dev libudev-dev libpci-dev libiberty-dev autoconf gawk jq
```
tautschnig: As far as I recall, we just need `build-essential libssl-dev` for the Linux kernel.

.github/workflows/build-and-test-Linux.yaml
```diff
@@ -0,0 +1,74 @@
...
+          make -C src minisat2-download
+          make -C src CXX='ccache /usr/bin/g++' cbmc.dir goto-cc.dir goto-diff.dir -j2
+      - name: Print ccache stats
+        run: ccache -s
+
+      - name: Get one-line-scan
+        run: git clone -b path-addition https://github.com/awslabs/one-line-scan.git
+
+      - name: Get Linux v5.10
+        run: git clone -b v5.10 --depth=1 https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git linux_5_10
```
tautschnig: I'd suggest to always use the latest release (at the risk of builds surprisingly breaking) and perhaps also using tar/zip instead of cloning?

.github/workflows/build-and-test-Linux.yaml
tautschnig: How about doing `tail -n +1 CPROVER/faultyInput/*` instead? Really simple and avoids any permissions problems.

.github/workflows/build-and-test-Linux.yaml
nmanthey: IIRC, that depends on the configuration. I can minimize this set.

.github/workflows/build-and-test-Linux.yaml
nmanthey: On the checks page (here, e.g.: https://github.com/diffblue/cbmc/pull/5916/checks there is an "artifacts" button on the top left, which allows (me) to download, I did not see any permission restrictions in the description of the action). Yes, I could also dump the content of the files, but that likely floods the output, as those are preprocessed already.

.github/workflows/build-and-test-Linux.yaml
nmanthey: If people are willing to take the risk, and address the issue (or roll this action back to the last known working release) I can build logic to always access the latest release.

On the other hand, I would like to be able to ensure that a specific version is always working. 

This is the risk I see when always using the latest release: Once the latest release fails, people might ignore the failure or disable this action to unblock themselves. Afterwards, even compiling older relases might start to break again. Hence, I decided to pick a specific release for now.

tautschnig: Maybe we should have two such actions? One with a fixed, longterm-maintenance release (5.10 seems like a reasonable candidate) and one using the "master" branch?

.github/workflows/build-and-test-Linux.yaml
tautschnig: Err, thank you, that was too easy. A first fix is in https://github.com/diffblue/cbmc/pull/5928.

tautschnig:
We will next need #5730 to be merged.

tautschnig:
> We will next need #5730 to be merged.

And then #5723, at which point the job should actually succeed.

tautschnig:
> > We will next need #5730 to be merged.
> 
> And then #5723, at which point the job should actually succeed.

Both of which are now merged, I believe this test should now succeed.

.github/workflows/build-and-test-Linux.yaml
```diff
@@ -0,0 +1,80 @@
...
+          ln -s goto-cc src/goto-cc/goto-ld
+          ln -s goto-cc src/goto-cc/goto-as
+          ln -s goto-cc src/goto-cc/goto-g++
+
+      - name: Compile Linux with CBMC via one-line-scan, and check for goto-cc section
+        run: |
+          ls
+          make -C linux_5_10 allnoconfig
+          make -C linux_5_10 kvm_guest.config
+          make -C linux_5_10 bzImage -j $(nproc)
```
tautschnig: May I suggest the following variation of this:
```
cat > kvm-config <<EOF
CONFIG_64BIT=y
CONFIG_X86_64=y
CONFIG_HIGH_RES_TIMERS=y
CONFIG_MULTIUSER=y
CONFIG_NET=y
CONFIG_VIRTUALIZATION=y
CONFIG_HYPERVISOR_GUEST=y
CONFIG_PARAVIRT=y
CONFIG_KVM_GUEST=y
CONFIG_KVM=y
CONFIG_KVM_INTEL=y
CONFIG_KVM_AMD=y
EOF
KCONFIG_ALLCONFIG=kvm-config make allnoconfig
make bzImage -j $(nproc)
```

.github/workflows/build-and-test-Linux.yaml
```diff
@@ -0,0 +1,80 @@
...
+          du -h . --max-depth=1
+          rm "linux_5_10/arch/x86/kernel/kvm.o"
```
tautschnig: With the above, perhaps instead do:
```
rm arch/x86/kernel/kvm.o
find arch/x86/kvm/ -name "*.o" -delete
```

.github/workflows/build-and-test-Linux.yaml
```diff
@@ -0,0 +1,80 @@
...
+          one-line-scan/one-line-scan --add-to-path $(pwd)/src/cbmc --add-to-path $(pwd)/src/goto-diff  --add-to-path $(pwd)/src/goto-cc --no-analysis --trunc-existing --extra-cflags -Wno-error -o CPROVER -j 3 -- make -C linux_5_10 bzImage -j $(nproc) -k || true
```
tautschnig: I don't think we we should be ignoring failures here, thus drop `-k || true`. Also, shouldn't one-line-scan use (at least) the same `-j` as the command being run?

.github/workflows/build-and-test-Linux.yaml
nmanthey: Sure, I can re-iterate

.github/workflows/build-and-test-Linux.yaml
nmanthey: Will iterate

.github/workflows/build-and-test-Linux.yaml
nmanthey: Yes, that should work.

The '-j' for one-line-scan is limited to not consume too much memory. For github, that should be good enough. For other use cases, that might be higher.

.github/workflows/build-and-test-Linux.yaml
nmanthey: Let's do the stable one for now, and iterate once that one is in.

nmanthey:
I addressed the comments, and factored out the actual compilation into a separate script, so that these steps can be executed in other environments more easily.

tautschnig:
> I addressed the comments, and factored out the actual compilation into a separate script, so that these steps can be executed in other environments more easily.

Thanks a lot, I'll review whether I already have a patch for the invariant failure.

tautschnig:
> > I addressed the comments, and factored out the actual compilation into a separate script, so that these steps can be executed in other environments more easily.
> 
> Thanks a lot, I'll review whether I already have a patch for the invariant failure.

A proposal for the patch is in #5967, but I need to reconsider whether that's the correct approach (it is borrowed from #5738, which also made a number of other changes).

NlightNFotis (commented):
This looks good to me (modulo the CI failure).

I will wait for this to be fixed then I will approve this to get in.

tautschnig:
@nmanthey #5967 is now merged, I believe this should pass once rebased.

tautschnig (approved):

.github/workflows/build-and-test-Linux.yaml
```diff
@@ -0,0 +1,53 @@
+name: Build Linux partially with CPROVER tools
+
+on:
+  pull_request:
+    branches:
+      - '**'
```
TGWDB, line 6: I'm a little hesitant about (another) CI job that runs on all PRs, specially one that we expect to report failures.

nmanthey: Feel free to suggest another branch pattern. I can also target the develop and production branch only. That might result in late surprises.

tautschnig: @TGWDB Why would you expect failures here?

TGWDB: @nmanthey I'm no git expert and don't have a better suggestion. I agree that late failures are not a good idea.

@tautschnig  Based on the original PR text:

"This change adds a github workflow that compiles parts of the Linux kernel with goto-cc, and fails in case compilation fails. Furthermore, the input fails that lead to the failure are archived, to allow easy consumption and investigation."

I'm happy if this works without failures, but was (mis?)reading that we expect this to fail more often due to the target and perhaps related aspects of `cbmc`.

If we expect `goto-cc` to never fail this except for when we break something then I have no objection.

TGWDB (approved):

PR #5922, opened 2021-03-12
Use simplification for more reliable pointer comparison

### LINKED ISSUE #5921: Wrong simplification in try_evaluate_pointer_comparison
CBMC version: latest (downloaded the tip of today)
Operating system: Linux x86_64
Exact command line resulting in the issue: ./cbmc main.c
What behaviour did you expect: SUCCESS on the assertion
What happened instead: FAILURE

Hello,

I noticed an unexpected behavior in the symbolic execution and crafted a testcase to expose the behavior.
I'm pretty convinced this is a bug, but maybe I'm missing something.

```cpp
#include <stdlib.h>
#include <assert.h>

int* get_ptr_minus_one(int* ptr)
{
    if (ptr) return ptr - 1;
    return 0;
}

int main()                                                                                                              
{
    int *begin = malloc(4*4);
    int *end = begin + 1;

    if (begin != get_ptr_minus_one(end))
        assert(0);
}
```

One more detail that indicates this being a bug is that replacing `malloc(4*4)` by `malloc(4*sizeof(int))` makes the assertion correctly pass.

Since I already debugged this a bit, I'll share the details in case it helps. (Notice I'm new to the codebase so the analysis might be off).

When we have a pointer comparison in the branch condition, the function https://github.com/diffblue/cbmc/blob/51c59e9e06e204f84e1b35bf7ad75439185300d7/src/goto-symex/symex_goto.cpp#L224 kicks in

It eventually will look over the value set of `get_ptr_minus_one(end)`
https://github.com/diffblue/cbmc/blob/51c59e9e06e204f84e1b35bf7ad75439185300d7/src/goto-symex/symex_goto.cpp#L107

 and check if it could possibly match `begin`
 https://github.com/diffblue/cbmc/blob/51c59e9e06e204f84e1b35bf7ad75439185300d7/src/goto-symex/symex_goto.cpp#L132
 
 If none of those match, then the pointer could not possibly match:
 https://github.com/diffblue/cbmc/blob/51c59e9e06e204f84e1b35bf7ad75439185300d7/src/goto-symex/symex_goto.cpp#L143
 
 That means, the two inputs being compared should be normalized to make sure that they always match if they are in fact the same symbol.
 However, I noticed scenarios in which they were not "normalized" and crafted this testcase.
 
 In this scenario, we have the comparison between the `&symex_dynamic::dynamic_object1` with `&symex_dynamic::dynamic_object1[0]`.
 
 Since they are not equal, they do not "match" (but should), so the code will say that the two pointers cannot be equal, reducing the guard to true, and always reaching the `assert(0)`.

The print of the two pointers being compared are the following:
 ```
 address_of
  * type: pointer
      * width: 64
      0: array
          * size: constant
              * type: unsignedbv
                  * width: 64
                  * #typedef: size_t
                  * #c_type: unsigned_long_int
              * value: 10
          * #dynamic: 1
          0: unsignedbv
              * width: 8
              * #c_type: unsigned_char
  0: symbol
      * type: array
          * size: constant
              * type: unsignedbv
                  * width: 64
                  * #typedef: size_t
                  * #c_type: unsigned_long_int
              * value: 4
          * #dynamic: 1
          0: unsignedbv
              * width: 8
              * #c_type: unsigned_char
      * identifier: symex_dynamic::dynamic_object1
      * expression: symbol
          * type: array
              * size: constant
                  * type: unsignedbv
                      * width: 64
                      * #typedef: size_t
                      * #c_type: unsigned_long_int
                  * value: 10
              * #dynamic: 1
              0: unsignedbv
                  * width: 8
                  * #c_type: unsigned_char
          * identifier: symex_dynamic::dynamic_object1
      * #SSA_symbol: 1
      * L1_object_identifier: symex_dynamic::dynamic_object1
```

and this:

```
address_of
  * type: pointer
      * width: 64
      0: unsignedbv                                                                                                     
          * width: 8
          * #c_type: unsigned_char
  0: index
      * type: unsignedbv
          * width: 8
          * #c_type: unsigned_char
      0: symbol
          * type: array
              * size: constant
                  * type: unsignedbv
                      * width: 64
                      * #typedef: size_t
                      * #c_type: unsigned_long_int
                  * value: 4
              * #dynamic: 1
              0: unsignedbv
                  * width: 8
                  * #c_type: unsigned_char
          * identifier: symex_dynamic::dynamic_object1 
          * expression: symbol
              * type: array
                  * size: constant
                      * type: unsignedbv
                          * width: 64
                          * #typedef: size_t
                          * #c_type: unsigned_long_int
                      * value: 10
                  * #dynamic: 1
                  0: unsignedbv
                      * width: 8
                      * #c_type: unsigned_char
              * identifier: symex_dynamic::dynamic_object1
          * #SSA_symbol: 1
          * L1_object_identifier: symex_dynamic::dynamic_object1
      1: constant
          * type: signedbv
              * width: 64
              * #c_type: signed_long_int
          * value: 0  
```

### PR DESCRIPTION
**Pull request by tautschnig**
As demonstrated in #5921, try_evaluate_pointer_comparisons must not
assume that all pointers being compared to have a canonical
representation. Use simplification to handle more cases instead of
relying on syntactic equality of expressions.

Co-authored-by: Breno Rodrigues Guimarães <brenorg@gmail.com>

Fixes: #5921

### DIFF
--- a/src/goto-symex/symex_goto.cpp
+++ b/src/goto-symex/symex_goto.cpp
@@ -85,10 +85,9 @@ static optionalt<renamedt<exprt, L2>> try_evaluate_pointer_comparison(
 {
   const constant_exprt *constant_expr =
     expr_try_dynamic_cast<constant_exprt>(other_operand);
-  const exprt other_without_typecast = skip_typecast(other_operand);
 
   if(
-    other_without_typecast.id() != ID_address_of &&
+    skip_typecast(other_operand).id() != ID_address_of &&
     (!constant_expr || constant_expr->get_value() != ID_NULL))
   {
     return {};
@@ -129,11 +128,24 @@ static optionalt<renamedt<exprt, L2>> try_evaluate_pointer_comparison(
         value_set_dereferencet::build_reference_to(
           value_set_element, symbol_expr, ns);
 
-      if(skip_typecast(value.pointer) == other_without_typecast)
+      // use the simplifier to test equality as we need to skip over typecasts
+      // and cannot rely on canonical representations, which would permit a
+      // simple syntactic equality test
+      exprt test_equal = simplify_expr(
+        equal_exprt{
+          typecast_exprt::conditional_cast(value.pointer, other_operand.type()),
+          other_operand},
+        ns);
+      if(test_equal.is_true())
       {
         constant_found = true;
         // We can't break because we have to make sure we find any instances of
         // ID_unknown or ID_invalid
+      }
+      else if(!test_equal.is_false())
+      {
+        // We can't conclude anything about the value-set
+        return {};
       }
     }
   }
--- a/regression/cbmc/Pointer_comparison2/main.c
+++ b/regression/cbmc/Pointer_comparison2/main.c
@@ -1,0 +1,18 @@
+#include <assert.h>
+#include <stdlib.h>
+
+int *get_ptr_minus_one(int *ptr)
+{
+  if(ptr)
+    return ptr - 1;
+  return 0;
+}
+
+int main()
+{
+  int *begin = malloc(4 * 4);
+  int *end = begin + 1;
+
+  if(begin != get_ptr_minus_one(end))
+    assert(0);
+}
--- a/regression/cbmc/Pointer_comparison2/test.desc
+++ b/regression/cbmc/Pointer_comparison2/test.desc
@@ -1,0 +1,8 @@
+CORE
+main.c
+
+^VERIFICATION SUCCESSFUL$
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
--- a/src/util/simplify_expr_int.cpp
+++ b/src/util/simplify_expr_int.cpp
@@ -1243,13 +1243,8 @@ simplify_exprt::simplify_inequality(const binary_relation_exprt &expr)
 
   // see if we are comparing pointers that are address_of
   if(
-    (tmp0.id() == ID_address_of ||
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Review by peterschrammel — APPROVED** [review]

File: src/goto-symex/symex_goto.cpp [hunk 1]
```diff
@@ -129,7 +128,11 @@ static optionalt<renamedt<exprt, L2>> try_evaluate_pointer_comparison(
         value_set_dereferencet::build_reference_to(
           value_set_element, symbol_expr, ns);
 
-      if(skip_typecast(value.pointer) == other_without_typecast)
+      // use the simplifier to test equality as we need to skip over typecasts
+      // and cannot rely on canonical representations, which would permit a
+      // simple syntactic equality test
+      equal_exprt test_equal{value.pointer, other_operand};
+      if(simplify_expr(test_equal, ns).is_true())
```
brenoguim: This code assumes that if this resolves to true, then it's a match, which is fine.

But it also assumes that if it does not resolves to true, then it's not a match, which is not entirely correct. If it does not resolve to true, it does not mean it's false. It could mean it's to complex to evaluate.

Does that make sense? Perhaps one should add:
```
    else if (!simplified_condition.is_false()) { return {}; /* cannot say anything */ }
```

tautschnig: You're of course right. I have amended the code, and also had to make some further fixes (mainly in the simplifier).

kroening (approved):

PR #5923, opened 2021-03-13
Introduce ansi-c/c_expr.h

kroening:
This introduces a new header file, ansi-c/c_expr.h, which is to collect
expressions that are specific to and internal to the C frontend.


peterschrammel (approved):

tautschnig:
Trying to figure out why check-cpplint isn't failing here, as it should. Working on this in #5929.

src/assembler/remove_asm.cpp
```diff
@@ -15,6 +15,8 @@ Date:   December 2014
 
 #include "remove_asm.h"
 
+#include <ansi-c/c_expr.h>
```
tautschnig: `src/assembler/module_dependencies.txt` does not currently list `ansi-c`, so we're introducing a new dependency on ansi-c here. And, really, this means that the expression is _not_ limited to the C front-end.

tautschnig (commented):
I'm not (yet) convinced: a further (assembler) and one stronger (goto-programs) dependency on ansi-c doesn't seem right.

tautschnig (approved):

PR #5924, opened 2021-03-13
introduce code_with_contract_typet

kroening:
This documents an existing, informal, variant of `code_typet`, which is
decorated with predicates for code contracts.

### DIFF
--- a/src/util/c_types.h
+++ b/src/util/c_types.h
@@ -327,6 +327,88 @@ inline c_enum_tag_typet &to_c_enum_tag_type(typet &type)
   return static_cast<c_enum_tag_typet &>(type);
 }
 
+class code_with_contract_typet : public code_typet
+{
+public:
+  /// Constructs a new 'code with contract' type, i.e., a function type
+  /// decorated with a function contract.
+  /// \param _parameters: The vector of function parameters.
+  /// \param _return_type: The return type.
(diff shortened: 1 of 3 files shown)


peterschrammel (approved):

src/util/c_types.h
```diff
@@ -327,6 +327,88 @@ inline c_enum_tag_typet &to_c_enum_tag_type(typet &type)
   return static_cast<c_enum_tag_typet &>(type);
 }
 
+class code_with_contract_typet : public code_typet
```
tautschnig, line 330: I don't think `c_types.h` is the right place for this: there is nothing C-specific about code contracts. It's just that currently only the C front-end has support for these.

tautschnig (approved):
This is a great clarification, but as said below: I don't think the header file is the right one.

src/util/c_types.h
kroening: I'll be happy to promote this once there is any other front-end that offers contracts.

kroening: src/util otherwise ends up being the union of everything, when it should be for things that are used in most places.

tautschnig: Understood. What I'm worried about is the distinction between "this is only meaningful for C" (which is what I thought this file was about) vs "this is only used in the C front-end." But that may well be a matter of taste, we have much more important issues to obsess over ;-)

PR #5929, opened 2021-03-15
Make cpplint check actually work in CI

tautschnig:
The Python script silently failed when invoked from run_diff.sh, which
redirects stdout to a file, when unidiff was not available.

~Leaving in Draft state while debugging seeming silent failures of the cpplint action.~

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -364,7 +364,8 @@ jobs:
           # user input
           DEBIAN_FRONTEND: noninteractive
         run: |
-          pip install unidiff
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -yq python3-unidiff
       - name: Check updated lines of code meet linting standards
         env:
           BASE_BRANCH: ${{ github.base_ref }}
--- a/scripts/diff_to_added_lines.py
+++ b/scripts/diff_to_added_lines.py
@@ -1,6 +1,5 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
-from __future__ import print_function
 import sys
 
 def eprint(*args, **kwargs):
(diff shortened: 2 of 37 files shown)


tautschnig:
check-cpplint is now working again, as can be seen in https://github.com/diffblue/cbmc/pull/5929/checks?check_run_id=2112286038.

tautschnig:
Regression tests are now also enabled.

.github/workflows/pull-request-checks.yaml
```diff
@@ -352,7 +352,7 @@ jobs:
         run: ./.github/workflows/pull-request-check-clang-format.sh
 
   check-cpplint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-18.04
```
NlightNFotis: I'm slightly weary of downgrading CI platforms to something that was released over 3 years ago. Do we really need to do this?

scripts/diff_to_added_lines.py
```diff
@@ -1,13 +1,17 @@
 #!/usr/bin/env python
```
NlightNFotis: Also, a general question: Should we start porting our internal scripts to Python 3 instead? I don't think we should be maintaining Python 2 scripts forever.

NlightNFotis (commented):
Two questions from me.

scripts/diff_to_added_lines.py
tautschnig: Yes, I do think that all scripts should be ported to Python 3.

.github/workflows/pull-request-checks.yaml
tautschnig: Reverted and enforcing the use of python3 for our internal script instead.

NlightNFotis (approved):
LGTM

PR #5931, opened 2021-03-15
C front-end: labelled declarations are only supported by Visual Studio

tautschnig:
In order to support GCC's label attributes even when the semicolon is
omitted, distinguish GCC and Visual Studio modes when parsing labelled
statements. See
https://github.com/gcc-mirror/gcc/blob/master/gcc/c/c-parser.cc#L6070 for
GCC's limitation, which we now mimic. (We already supported attributes
followed by a semicolon, which GCC's documentation mandates, but the
Linux kernel doesn't always follow this rule.) Fixed regression tests
that had declarations immediately following a label.

### DIFF
--- a/regression/ansi-c/label1/main.c
+++ b/regression/ansi-c/label1/main.c
@@ -1,0 +1,14 @@
+extern void foo(int *, unsigned int);
+
+int main()
+{
+#ifdef _MSC_VER
+label:
+  int x;
+#elif defined(__GNUC__)
+  int *p;
+  unsigned int u;
+label:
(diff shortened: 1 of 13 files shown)


peterschrammel:
The link https://github.com/gcc-mirror/gcc/blob/master/gcc/c/c-parser.c#L5553 doesn't work.

tautschnig:
> The link https://github.com/gcc-mirror/gcc/blob/master/gcc/c/c-parser.c#L5553 doesn't work.

I have updated the link in both the commit message and the PR description.

peterschrammel (approved):

PR #5932, opened 2021-03-15
Fix CMake syntax to enable memory analyzer tests

tautschnig:
"option" requires the help text as second, not as third argument. This
resulted in memory analyzer tests being disabled no matter what the OS
being built on.


unit/memory-analyzer/gdb_api.cpp
```diff
@@ -18,13 +18,61 @@ Author: Malte Mues <mail.mues@gmail.com>
...
+          "\n"
+          "  return 0;\n"
+          "}\n";
+    of.close();
+
+    REQUIRE(run("gcc", {"gcc", "-g", "-o", compiled(), tmp()}) == 0);
+  }
+
+  temporary_filet compiled;
+};
```
TGWDB: This seems unnecessarily ugly, is there a reason to do it this way?

unit/memory-analyzer/gdb_api.cpp
```diff
@@ -67,9 +115,34 @@ void gdb_api_internals_test()
...
+          "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+          "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+          "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+          "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+          "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+          "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+          "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+          "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n"
+          "xyz";
+    of.close();
```
TGWDB: Again, is this the best way to do this?

unit/memory-analyzer/gdb_api.cpp
tautschnig: If, as was the case before, we keep the contents in a separate file, then this file must either be included in the compiled unit test binary (e.g., via `#include`), or else the compiled unit test binary necessarily relies on the source tree being present as well. It seems reasonable to assume that the source tree will be present, but then also the exact location of the file must be known.

I made these changes as I found the above conditions not to be satisfied in my ad-hoc testing for I wasn't aware where the unit test had to be run from to make things work. That being said, we have the same problem in other unit tests. So maybe it's not all that important to fix if people are concerned about this approach.

TGWDB: I understand the motivation, I was more considering that there might be a prettier way to have the text inside the file.

TGWDB (approved):

unit/memory-analyzer/gdb_api.cpp
tautschnig: I've modified this to use `ansi-c/file_converter` and `#include` them instead.

TGWDB (approved):

chrisr-diffblue (approved):

PR #5933, opened 2021-03-15
Enable goto-analyzer-simplify regression tests

tautschnig:
Updated scripts and test specifications to work in current environments.


regression/goto-analyzer-simplify/Makefile
```diff
@@ -1,22 +1,15 @@
...
-		exit 1; \
-	fi
+tests.log: ../test.pl
+	@../test.pl -e -p -c "../chain.sh ../../../src/goto-analyzer/goto-analyzer"
 
 show:
 	@for dir in *; do \
 		if [ -d "$$dir" ]; then \
-			vim -o "$$dir/*.c" "$$dir/*.out"; \
+			vim -o "$$dir/*.java" "$$dir/*.out"; \
```
martin-cs, line 12: ??? Why change this?

martin-cs (approved):
Changes look good and preserve the original configuration.  I have also checked the tests and I think the results are correct.

regression/goto-analyzer-simplify/Makefile
tautschnig: Oh, my bad, I copied `regression/goto-analyzer/Makefile`. I suppose both should be changed, or perhaps we should just get rid of this `show` target. Will create a PR for this.

tautschnig: Finally done in https://github.com/diffblue/cbmc/pull/5957.

PR #5937, opened 2021-03-15
Enable regression tests of weak memory model instrumentation

tautschnig:
These regression tests are a selected subset (cf. README.txt) of the
full set of available tests (stored in goto-instrument-wmm-full.tgz),
and will ensure basic testing of the functionality implemented in
goto-instrument.


martin-cs (approved):
I must confess that I haven't read every line of this PR but I think I get the general picture.  One concern : how often do the THOROUGH tests get run.  Could we add it to the new version / release workflow?

tautschnig:
> I must confess that I haven't read every line of this PR but I think I get the general picture. One concern : how often do the THOROUGH tests get run. Could we add it to the new version / release workflow?

I'll create a separate PR for this.

PR #5940, opened 2021-03-15
Support ==,!=,<,<=,>,>= over vectors

### LINKED ISSUE #5930: Feature request: Support for SIMD comparison operators
Support for SIMD arithmetic operators (e.g., addition or logical operators) exists in CBMC but this is not the case for comparison operators `==`, `!=`, `<`, `>`, `<=` and `>=`. According to [GCC online docs](https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html ) vectors are compared element by element producing a 0 when the comparison is false and -1 otherwise.

I am not able to provide an example since #5903 does not allow to compile, but this is the warning I am getting from CBMC when I run it on code that we generate:
```
warning: ignoring =
  * type: array
      * #source_location: nil
      * size: constant
          * type: signedbv
              * width: 64
              * #c_type: signed_long_int
          * value: 2
      0: signedbv
          * width: 64
          * #c_type: signed_long_int
  0: array
      * type: array
          * #source_location: nil
          * size: constant
              * type: signedbv
                  * width: 64
                  * #c_type: signed_long_int
              * value: 2
          0: signedbv
              * width: 64
              * #c_type: signed_long_int
      0: constant
          * type: signedbv
              * width: 64
              * #c_type: signed_long_int
          * value: 0
      1: constant
          * type: signedbv
              * width: 64
              * #c_type: signed_long_int
          * value: 0
  1: array
      * type: array
          * #source_location: nil
          * size: constant
              * type: signedbv
                  * width: 64
                  * #c_type: signed_long_int
              * value: 2
          0: signedbv
              * width: 64
              * #c_type: signed_long_int
      0: constant
          * type: signedbv
              * width: 64
              * #c_type: signed_long_int
          * value: 0
      1: constant
          * type: signedbv
              * width: 64
              * #c_type: signed_long_int
          * value: 0
```

tautschnig:
These need to be rewritten to point-wise applications and return -1
(true) or 0 (false) per element.

Fixes: #5930

### DIFF
--- a/src/ansi-c/c_typecheck_expr.cpp
+++ b/src/ansi-c/c_typecheck_expr.cpp
@@ -1417,18 +1417,12 @@ void c_typecheck_baset::typecheck_expr_rel(
   throw 0;
 }
 
-void c_typecheck_baset::typecheck_expr_rel_vector(
-  binary_relation_exprt &expr)
+void c_typecheck_baset::typecheck_expr_rel_vector(binary_exprt &expr)
 {
-  exprt &op0=expr.op0();
-  exprt &op1=expr.op1();
+  const typet &o_type0 = as_const(expr).op0().type();
+  const typet &o_type1 = as_const(expr).op1().type();
 
-  const typet o_type0 = op0.type();
-  const typet o_type1 = op1.type();
-
-  if(
-    o_type0.id() != ID_vector || o_type1.id() != ID_vector ||
-    o_type0.subtype() != o_type1.subtype())
+  if(o_type0.id() != ID_vector || o_type0 != o_type1)
   {
     error().source_location = expr.source_location();
     error() << "vector operator '" << expr.id() << "' not defined for types '"
@@ -1437,9 +1431,18 @@ void c_typecheck_baset::typecheck_expr_rel_vector(
     throw 0;
   }
 
-  // Comparisons between vectors produce a vector
-  // of integers with the same dimension.
-  expr.type()=vector_typet(signed_int_type(), to_vector_type(o_type0).size());
+  // Comparisons between vectors produce a vector of integers of the same width
+  // with the same dimension.
+  auto subtype_width = to_bitvector_type(o_type0.subtype()).get_width();
+  expr.type() =
+    vector_typet{signedbv_typet{subtype_width}, to_vector_type(o_type0).size()};
+
+  // Replace the id as the semantics of these are point-wise application (and
+  // the result is not of bool type).
+  if(expr.id() == ID_notequal)
+    expr.id(ID_vector_notequal);
+  else
+    expr.id("vector-" + id2string(expr.id()));
 }
 
 void c_typecheck_baset::typecheck_expr_ptrmember(exprt &expr)
--- a/regression/cbmc/gcc_vector1/main.c
+++ b/regression/cbmc/gcc_vector1/main.c
@@ -105,6 +105,20 @@ int main()
   assert(z.members[2]==-x.members[2]);
   assert(z.members[3]==-x.members[3]);
 
+  // Boolean operators on vectors
+  z = x;
+  y.v = z.v == x.v;
+  assert(y.members[0]);
+  assert(y.members[1]);
+  assert(y.members[2]);
+  assert(y.members[3]);
+
+  y.v = x.v <= z.v;
+  assert(y.members[0]);
+  assert(y.members[1]);
+  assert(y.members[2]);
+  assert(y.members[3]);
+
   // build vector with typecast
   z.v=(v4si){ 0, 1, 2, 3 };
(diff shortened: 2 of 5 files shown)


kroening:
What happens when the vector subtype is unsigned?

tautschnig:
> What happens when the vector subtype is unsigned?

GCC's documentation says "constant of the appropriate type where all bits are set" (https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html), so this should work (as we use two's complement).

src/util/std_expr.h
```diff
@@ -664,8 +664,8 @@ class binary_predicate_exprt:public binary_exprt
 
     DATA_CHECK(
       vm,
-      expr.type().id() == ID_bool,
-      "result of binary predicate expression should be of type bool");
+      expr.type().id() == ID_bool || expr.type().id() == ID_vector,
+      "result of binary predicate expression should be of type bool or vector");
```
tautschnig: @kroening I quite dislike that I had to make the above change: this is really to make `--validate-goto-model` pass, which is currently applied before and (!) after calls to `remove_vector`. We should probably discuss this.

kroening:
> > What happens when the vector subtype is unsigned?
> 
> GCC's documentation says "constant of the appropriate type where all bits are set" (https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html), so this should work (as we use two's complement).

Does that operation exist for vectors of type `float`?

src/util/std_expr.h
kroening: Yes, not super. One could argue that something that returns a vector isn't a predicate.

tautschnig: Perhaps the better solution would be introducing ids along the lines of `vector_equal`, `vector_gt`, etc?

tautschnig:
> > > What happens when the vector subtype is unsigned?
> > 
> > 
> > GCC's documentation says "constant of the appropriate type where all bits are set" (https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html), so this should work (as we use two's complement).
> 
> Does that operation exist for vectors of type `float`?

The docs say: "Comparison operands can be vector expressions of integer-type or real-type."

kroening:
> > Does that operation exist for vectors of type `float`?
> 
> The docs say: "Comparison operands can be vector expressions of integer-type or real-type."

Won't the `from_integer(-1, subtype)` then fail?

src/util/std_expr.h
kroening: Not absolutely necessary -- we do overload many other IDs, but I admit that others may be surprised to see a non-boolean equality. `vector_equal` etc. may then be the best compromise.

kroening:
> > The docs say: "Comparison operands can be vector expressions of integer-type or real-type."
> 
> Won't the `from_integer(-1, subtype)` then fail?

`from_integer(-1, subtype)` actually works for `float`. However, you then do get the floating-point representation of -1, which is not the same as "constant of the appropriate type where all bits are set" -- that would be NaN.

kroening:
> `from_integer(-1, subtype)` actually works for `float`. However, you then do get the floating-point representation of -1, which is not the same as "constant of the appropriate type where all bits are set" -- that would be NaN.

A quick test confirms that clang generates NaN, not -1.

tautschnig:
> > `from_integer(-1, subtype)` actually works for `float`. However, you then do get the floating-point representation of -1, which is not the same as "constant of the appropriate type where all bits are set" -- that would be NaN.
> 
> A quick test confirms that clang generates NaN, not -1.

Ah, just noting that `subtype` (which is the type of the result, not that of the operands) wouldn't be `float` here, it would still be `signed_int`. Though that's only what the C front-end generates, I acknowledge that I omitted any checking in here.

kroening:
Ah, I overlooked "The result of the comparison is a vector of the same width and number of elements as the comparison operands with a signed integral element type".

Right now, the result type is an array type, but should be a vector type.

kroening:
Ah, sorry, wrong again -- the frontend does produce a vector type, of signed integers, but it does fail to meet the requirement "same width".

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -1437,9 +1431,18 @@ void c_typecheck_baset::typecheck_expr_rel_vector(
...
+  expr.type() =
+    vector_typet{signedbv_typet{subtype_width}, to_vector_type(o_type0).size()};
+
+  // Replace the id as the semantics of these are point-wise application (and
+  // the result is not of bool type).
+  if(expr.id() == ID_notequal)
+    expr.id(ID_vector_notequal);
+  else
+    expr.id("vector-" + id2string(expr.id()));
 }
```
kroening, line 1446: This will be vector-=, vector-<= etc?

kroening (approved):

PR #5942, opened 2021-03-16
Fix side-effect check on loop invariants

### LINKED ISSUE #5915: False positive from side-effect check in loop contracts
**CBMC version:** `develop`
**Operating system:** Mac OS 10.15

Running `--enforce-all-contracts` raises a false positive on the following file:

```c
int main()
{
  int i, N, *a = malloc(sizeof(int));

  for(i = *a; i < N; ++i)
    __CPROVER_loop_invariant(
      (*a > N) ||
      (*a <= i && i <= N)
    )
  {
    ++i;
  }
}
```

**Exact command line:**

```console
$ goto-cc test.c -o 1.gb
$ goto-instrument --enforce-all-contracts 1.gb 2.gb
Invariant check failed
File: ../src/goto-programs/goto_convert.cpp:877 function: convert_loop_invariant
Condition: loop invariant is not side-effect free
Reason: no_sideeffects.instructions.empty()
Backtrace:
...
```

`goto-instrument` shouldn't reject this file. The invariant seems free of side effects -- it only performs basic arithmetic comparisons and read-only heap access.

SaswatPadhi:
Fixes #5915.

Invariant predicates were being _cleaned_ (C -> logic via `clean_expr`) within `convert_loop_invariant` and were rejected if any intermediate goto instructions were introduced during the cleaning process. However, this check is overly conservative -- it ends up rejecting some complex, but pure, Boolean expressions that are compiled to multiple goto instructions.

We replace this logic with a call to `has_subexpr` with `ID_side_effect`.

- N/A ~~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~~
- N/A ~~My commit message includes data points confirming performance improvements (if claimed).~~
- N/A ~~White-space or formatting changes outside the feature-related changed lines are in commits of their own.~~

### DIFF
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -20,7 +20,6 @@ Date: February 2016
 
 #include <ansi-c/c_expr.h>
 
-#include <goto-programs/goto_convert_class.h>
 #include <goto-programs/remove_skip.h>
 
 #include <util/arith_tools.h>
@@ -33,8 +32,6 @@ Date: February 2016
 #include <util/pointer_offset_size.h>
 #include <util/pointer_predicates.h>
 #include <util/replace_symbol.h>
-
-#include "loop_utils.h"
 
 /// Predicate to be used with the exprt::visit() function. The function
 /// found_return_value() will return `true` iff this predicate is called on an
@@ -74,11 +71,12 @@ exprt get_size(const typet &type, const namespacet &ns, messaget &log)
   return result;
 }
 
-static void check_apply_invariants(
+void code_contractst::check_apply_invariants(
   goto_functionst::goto_functiont &goto_function,
   const local_may_aliast &local_may_alias,
   const goto_programt::targett loop_head,
-  const loopt &loop)
+  const loopt &loop,
+  const irep_idt &mode)
 {
   PRECONDITION(!loop.empty());
 
@@ -96,16 +94,18 @@ static void check_apply_invariants(
   if(invariant.is_nil())
     return;
 
-  // change H: loop; E: ...
+  // change
+  //   H: loop;
+  //   E: ...
   // to
-  // H: assert(invariant);
-  // havoc;
-  // assume(invariant);
-  // if(guard) goto E:
-  // loop;
-  // assert(invariant);
-  // assume(false);
-  // E: ...
+  //   H: assert(invariant);
+  //   havoc;
+  //   assume(invariant);
+  //   if(guard) goto E:
+  //   loop;
+  //   assert(invariant);
+  //   assume(false);
+  //   E: ...
 
   // find out what can get changed in the loop
   modifiest modifies;
@@ -116,17 +116,20 @@ static void check_apply_invariants(
 
   // assert the invariant
   {
-    goto_programt::targett a = havoc_code.add(
-      goto_programt::make_assertion(invariant, loop_head->source_location));
-    a->source_location.set_comment("Check loop invariant before entry");
+    code_assertt assertion{invariant};
+    assertion.add_source_location() = loop_head->source_location;
+    converter.goto_convert(assertion, havoc_code, mode);
+    havoc_code.instructions.back().source_location.set_comment(
+      "Check loop invariant before entry");
   }
 
   // havoc variables being written to
   build_havoc_code(loop_head, modifies, havoc_code);
 
   // assume the invariant
-  havoc_code.add(
-    goto_programt::make_assumption(invariant, loop_head->source_location));
+  code_assumet assumption{invariant};
+  assumption.add_source_location() = loop_head->source_location;
+  converter.goto_convert(assumption, havoc_code, mode);
 
   // non-deterministically skip the loop if it is a do-while loop
   if(!loop_head->is_goto())
@@ -142,11 +145,14 @@ static void check_apply_invariants(
 
   // assert the invariant at the end of the loop body
   {
-    goto_programt::instructiont a =
-      goto_programt::make_assertion(invariant, loop_end->source_location);
-    a.source_location.set_comment("Check that loop invariant is preserved");
-    goto_function.body.insert_before_swap(loop_end, a);
-    ++loop_end;
+    code_assertt assertion{invariant};
+    assertion.add_source_location() = loop_end->source_location;
+    converter.goto_convert(assertion, havoc_code, mode);
+    havoc_code.instructions.back().source_location.set_comment(
+      "Check that loop invariant is preserved");
+    auto offset = havoc_code.instructions.size();
+    goto_function.body.insert_before_swap(loop_end, havoc_code);
+    std::advance(loop_end, offset);
   }
 
   // change the back edge into assume(false) or assume(guard)
@@ -474,15 +480,21 @@ bool code_contractst::apply_function_contract(
 }
 
 void code_contractst::apply_loop_contract(
+  const irep_idt &function_name,
   goto_functionst::goto_functiont &goto_function)
 {
   local_may_aliast local_may_alias(goto_function);
   natural_loops_mutablet natural_loops(goto_function.body);
 
-  // iterate over the (natural) loops in the function
+  // Iterate over the (natural) loops in the function,
+  // and apply any invariant annotations that we find.
   for(const auto &loop : natural_loops.loop_map)
     check_apply_invariants(
-      goto_function, local_may_alias, loop.first, loop.second);
+      goto_function,
+      local_may_alias,
+      loop.first,
+      loop.second,
+      symbol_table.lookup_ref(function_name).mode);
 }
 
 const symbolt &code_contractst::new_tmp_symbol(
@@ -1069,7 +1081,7 @@ bool code_contractst::enforce_contracts()
     if(has_contract(goto_function.first))
       funs_to_enforce.insert(id2string(goto_function.first));
     else
-      apply_loop_contract(goto_function.second);
+      apply_loop_contract(goto_function.first, goto_function.second);
   }
   return enforce_contracts(funs_to_enforce);
 }
@@ -1089,7 +1101,7 @@ bool code_contractst::enforce_contracts(
                   << messaget::eom;
       continue;
     }
-    apply_loop_contract(goto_function->second);
+    apply_loop_contract(goto_function->first, goto_function->second);
 
     if(!has_contract(fun))
     {
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -158,15 +158,6 @@ static void check_apply_invariants(
     loop_end->set_condition(boolean_negate(loop_end->get_condition()));
 }
 
-void code_contractst::convert_to_goto(
-  const codet &code,
-  const irep_idt &mode,
-  goto_programt &program)
-{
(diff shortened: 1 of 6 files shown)


src/goto-programs/goto_convert.cpp
```diff
@@ -868,13 +868,11 @@ void goto_convertt::convert_loop_invariant(
...
-
-  INVARIANT_WITH_DIAGNOSTICS(
-    no_sideeffects.instructions.empty(),
-    "loop invariant is not side-effect free",
-    code.find_source_location());
+  // Loop contracts are expanded to assertions (and assumptions)
+  // at multiple code points around a loop.
+  // Rather than performing `clean_expr` during compilation,
+  // we convert the invariant to logic at these various code points
+  // only when they are enforced (using goto-instrument).
```
martin-cs: So... this means that the output of goto_convert will no longer be side effect free?  So I can now write code that has a different meaning if run with and without code contracts?  That doesn't seem like a very safe design.

tautschnig: Without having reviewed the details just yet: the short term idea is to move all invariant-related processing (as far as possible) to `code_contracts.cpp`, knowing that this will leave annotations in there that wouldn't satisfy side-effect freedom. Once this has been cleaned up (right now it's a mess with parts of the processing happening in all sorts of places), it should be fully integrated into goto-program conversion, possibly leaving little or (ideally) nothing left in `goto-instrument`.

SaswatPadhi:
@martin-cs @tautschnig 

After a lot of back and forth discussion and changes, it seemed like at least for the invariants that we currently write, it's sufficient to just check for `ID_side_effect` subexpressions. I left a TODO to revisit the `has_sideeffect` function later and allow calling functions that have an empty ASSIGNS clause from within loop invariants. If you think we might also need to handle some other cases, please suggest -- I'd be happy to include them as part of this change, or leave a TODO to revisit later (if it gets too complicated).

src/goto-programs/goto_convert_side_effect.cpp
```diff
@@ -35,6 +35,24 @@ bool goto_convertt::has_function_call(const exprt &expr)
...
+
+  for(const auto &op : expr.operands())
+  {
+    if(goto_convertt::has_sideeffect(op))
+      return true;
+  }
+
+  return false;
+}
+
```
tautschnig: Good call out that this was never implemented. That said, can you actually do this a bit different as follows:
1) Get rid of the declaration as well.
2) At the point where you will currently need it, just use `has_subexpr(expr, ID_side_effect);` (from `util/expr_util.h`).

src/goto-instrument/code_contracts.cpp
```diff
@@ -112,17 +112,20 @@ static void check_apply_invariants(
 
   // assert the invariant
   {
-    goto_programt::targett a = havoc_code.add(
-      goto_programt::make_assertion(invariant, loop_head->source_location));
-    a->source_location.set_comment("Check loop invariant before entry");
+    auto assertion = code_assertt(invariant);
+    assertion.add_source_location().swap(loop_head->source_location);
```
tautschnig: Is `swap` the right thing to do here?

src/goto-instrument/code_contracts.cpp
```diff
@@ -112,17 +112,20 @@ static void check_apply_invariants(
 
   // assert the invariant
   {
-    goto_programt::targett a = havoc_code.add(
-      goto_programt::make_assertion(invariant, loop_head->source_location));
-    a->source_location.set_comment("Check loop invariant before entry");
+    auto assertion = code_assertt(invariant);
```
tautschnig: Nit picking: is this use of `auto` adding value? Wouldn't `code_assertt assertion{invariant};` be even fewer characters and thus less cognitive load?

src/goto-instrument/code_contracts.cpp
```diff
@@ -112,17 +112,20 @@ static void check_apply_invariants(
...
+      "Check loop invariant before entry");
   }
 
   // havoc variables being written to
   build_havoc_code(loop_head, modifies, havoc_code);
 
   // assume the invariant
-  havoc_code.add(
-    goto_programt::make_assumption(invariant, loop_head->source_location));
+  auto assumption = code_assumet(invariant);
```
tautschnig: As above: does `auto` add value?

src/goto-instrument/code_contracts.cpp
```diff
@@ -138,11 +141,14 @@ static void check_apply_invariants(
 
   // assert the invariant at the end of the loop body
   {
-    goto_programt::instructiont a =
-      goto_programt::make_assertion(invariant, loop_end->source_location);
-    a.source_location.set_comment("Check that loop invariant is preserved");
-    goto_function.body.insert_before_swap(loop_end, a);
-    ++loop_end;
+    auto assertion = code_assertt(invariant);
```
tautschnig: Does `auto` add value?

src/goto-instrument/code_contracts.cpp
```diff
@@ -112,17 +112,20 @@ static void check_apply_invariants(
...
   }
 
   // havoc variables being written to
   build_havoc_code(loop_head, modifies, havoc_code);
 
   // assume the invariant
-  havoc_code.add(
-    goto_programt::make_assumption(invariant, loop_head->source_location));
+  auto assumption = code_assumet(invariant);
+  assumption.add_source_location().swap(loop_head->source_location);
```
tautschnig: Given the earlier use of `swap`, there won't be any source location in here?

src/goto-instrument/code_contracts.cpp
```diff
@@ -138,11 +141,14 @@ static void check_apply_invariants(
 
   // assert the invariant at the end of the loop body
   {
-    goto_programt::instructiont a =
-      goto_programt::make_assertion(invariant, loop_end->source_location);
-    a.source_location.set_comment("Check that loop invariant is preserved");
-    goto_function.body.insert_before_swap(loop_end, a);
-    ++loop_end;
+    auto assertion = code_assertt(invariant);
+    assertion.add_source_location().swap(loop_end->source_location);
```
tautschnig: As above: nothing left to swap in.

src/goto-instrument/code_contracts.h
```diff
@@ -18,6 +18,8 @@ Date: February 2016
 #include <string>
 #include <unordered_set>
 
+#include <analyses/local_may_alias.h>
```
tautschnig: No need to have this in the header file, a forward declaration will do.

src/goto-instrument/code_contracts.h
```diff
@@ -26,6 +28,8 @@ Date: February 2016
 #include <util/namespace.h>
 #include <util/pointer_expr.h>
 
+#include "loop_utils.h"
```
tautschnig, line 30: Wouldn't forward declarations do the trick?

src/goto-programs/goto_convert.cpp
```diff
@@ -868,13 +868,11 @@ void goto_convertt::convert_loop_invariant(
...
     return;
 
-  goto_programt no_sideeffects;
-  clean_expr(invariant, no_sideeffects, mode);
-
-  INVARIANT_WITH_DIAGNOSTICS(
-    no_sideeffects.instructions.empty(),
-    "loop invariant is not side-effect free",
-    code.find_source_location());
+  if(has_sideeffect(invariant))
```
tautschnig: See comment in the previous commit: use `has_subexpr(invariant, ID_side_effect)`

tautschnig (approved):

src/goto-instrument/code_contracts.h
SaswatPadhi: Sorry, resolved this comment by mistake without addressing it.

I `#include`d this header file for the `loopt` type, which is `typedef`ed as:

```c
typedef const natural_loops_mutablet::natural_loopt loopt;
```

Could I forward declare it here somehow? I wasn't sure how to.

tautschnig: I believe one of `typename loopt;` or `class loopt;` might still work, and likely the compiler will telly you if it doesn't :-) Not high priority, though.

src/goto-programs/goto_convert_class.h
```diff
@@ -97,7 +97,6 @@ class goto_convertt:public messaget
 
   void rewrite_boolean(exprt &dest);
 
-  static bool has_sideeffect(const exprt &expr);
```
martin-cs, line 100: Why is this removed?

src/goto-programs/goto_convert.cpp
```diff
@@ -868,13 +868,11 @@ void goto_convertt::convert_loop_invariant(
...
     return;
 
-  goto_programt no_sideeffects;
-  clean_expr(invariant, no_sideeffects, mode);
-
-  INVARIANT_WITH_DIAGNOSTICS(
-    no_sideeffects.instructions.empty(),
-    "loop invariant is not side-effect free",
-    code.find_source_location());
+  if(has_subexpr(invariant, ID_side_effect))
```
martin-cs, line 871: As I recall, this was the really contentious part last time around.  This seems much less disruptive.

I think throwing the exception is correct because the user can trigger this even if the tool is non-buggy.  I wonder if it would be more user-friendly to check this earlier but ... I guess we need to check it here as well.

I wonder if this might still be a bit too conservative but, if it works for you then fine.

martin-cs (approved):
My initial reason for objection, that we were changing a major post condition of goto-convert is gone so there is no reason for me to block this.

Some thoughts:
*. Are similar conditions checked for pre/post?
*. Are the restrictions documented in the manual?

Longer-term, do we want to ban function calls from these annotations?  One thing I found nice about SPARK was being able to define predicates (like `is_initialised` and `is_well_structured`) using Ada functions.  If there isn't a way of doing this you risk a lot of parallel maintenance when altering common invariants.  I guess macros would be one option but these don't work for things like `is_sorted_tree` and having options of how much to expand them.   This is well beyond the scope of this PR but is something worth thinking about IMHO and will touch this code.

src/goto-programs/goto_convert_class.h
SaswatPadhi: It was never defined, so I removed the declaration as well.

martin-cs: Sounds good.

src/goto-programs/goto_convert.cpp
SaswatPadhi: Thank you, yes earlier I had removed this part entirely and was doing some checking in `goto-instrument` instead. Now this check is performed at `goto-cc` compilation.

> I wonder if this might still be a bit too conservative but, if it works for you then fine.

I completely agree with you. And it's on our roadmap to allow function calls within contracts, if those functions have an empty assigns clause (so no side effects).

SaswatPadhi:
Thank you for taking a look, @martin-cs.

I just wanted to mention that @feliperodri and I are actively working on the issues that you pointed out:

> Are similar conditions checked for pre/post?

Not currently, but we are fix this during our refactoring pass over the code contracts module

> Are the restrictions documented in the manual?

Good call out! Nope, they are not. I would add it to our TODO list for documentation updates.

> Longer-term, do we want to ban function calls from these annotations?

It's on our roadmap to allow function calls within contracts, if those functions have an empty assigns clause (so no side effects).

Issue #5910, opened 2021-03-10
Some regression test runners fail to `clean`


SaswatPadhi:
**Exact command line resulting in the issue:** `make clean` within `regression/cbmc`
**What behaviour did you expect:** `.out`, `.gb`, `.log` files should be removed
**What happened instead:** `find: illegal option -- n`

Some of the `regression/<suite>/Makefile`s use a use a `find` command to recursively search and remove temporary files, e.g. in `regression/cbmc/Makefile` we have:
```bash
find -name '*.out' -execdir $(RM) '{}' \;
```
This statement raises a syntax error on Mac OS (and probably *nix in general?) because `find` is missing a path argument. The following works as expected:
```bash
find . -name '*.out' -execdir $(RM) '{}' \;
```

Some other `Makefile`s use a `for` loop to iterate over subdirectories and remove temporary files, but that would only clean up until depth 1, e.g. in `regression/goto-instrument/Makefile` we have:
```bash
	@for dir in *; do \
		$(RM) tests.log; \
		if [ -d "$$dir" ]; then \
			cd "$$dir"; \
			$(RM) *.out *.gb; \
			cd ..; \
		fi \
	done
```
If cleaning until depth 1 is sufficient, the `for` loop could be replaced by `$(RM) tests.log */*.out */*.gb`. If it's not, then the find command above could be used.

There needs to be a global pass over all the `regression/` subdirectories to fix the `clean` action in `Makefile`s.

TGWDB:
Addressed in [PR 5943](https://github.com/diffblue/cbmc/pull/5943) .

SaswatPadhi:
Thanks!

PR #5943, opened 2021-03-16
Ensure make clean of regression tests cleans everything

TGWDB:
This commit fixes many make files to ensure they properly
find and clean up the files created by the running of
regression tests. This also fixes a bug with the use of
the find command without a specified path.


NlightNFotis (approved):
LGTM

regression/acceleration/Makefile
```diff
@@ -28,5 +28,6 @@ show:
 	done;
 
 clean:
-	$(RM) tests.log
-	$(RM) */main.out
+	find -name '*.log' -execdir $(RM) '{}' \;
+	find -name '*.gb' -execdir $(RM) '{}' \;
+	find -name '*.out' -execdir $(RM) '{}' \;
```
tautschnig: In other cases you explicitly added the `.` after `find`. Both might work, but please be consistent.

regression/acceleration/Makefile
```diff
@@ -28,5 +28,6 @@ show:
 	done;
 
 clean:
-	$(RM) tests.log
-	$(RM) */main.out
+	find -name '*.log' -execdir $(RM) '{}' \;
```
tautschnig: Would `$(RM) tests.log` not do the trick here?

regression/goto-analyzer-simplify/Makefile
```diff
@@ -21,11 +21,6 @@ show:
...
-		rm -f tests.log; \
-		if [ -d "$$dir" ]; then \
-			cd "$$dir"; \
-			rm -f *.out *.gb; \
-			cd ..; \
-		fi \
-	done
+	find . -name '*.out' -execdir $(RM) '{}' \;
+	find . -name '*.gb' -execdir $(RM) '{}' \;
+	find . -name '*.log' -execdir $(RM) '{}' \;
```
tautschnig: Would `$(RM) tests.log` not suffice here?

regression/goto-ld/Makefile
```diff
@@ -27,6 +27,7 @@ show:
 	done;
 
 clean:
-	find -name '*.out' -execdir $(RM) '{}' \;
-	find -name '*.gb' -execdir $(RM) '{}' \;
-	find -name '*.goto-cc-saved' -execdir $(RM) '{}' \;
+	find . -name '*.out' -execdir $(RM) '{}' \;
+	find . -name '*.gb' -execdir $(RM) '{}' \;
+	find . -name '*.goto-cc-saved' -execdir $(RM) '{}' \;
+	find . -name '*.log' -execdir $(RM) '{}' \;
```
tautschnig: `$(RM) tests.log`?

regression/goto-instrument-typedef/Makefile
```diff
@@ -25,11 +25,6 @@ show:
...
-		rm -f tests.log; \
-		if [ -d "$$dir" ]; then \
-			cd "$$dir"; \
-			rm -f *.out *.gb; \
-			cd ..; \
-		fi \
-	done
+	find . -name '*.out' -execdir $(RM) '{}' \;
+	find . -name '*.gb' -execdir $(RM) '{}' \;
+	find . -name '*.log' -execdir $(RM) '{}' \;
```
tautschnig: `$(RM) tests.log`?

regression/goto-cc-multi-file/Makefile
```diff
@@ -26,11 +26,6 @@ show:
...
-		rm -f tests.log; \
-		if [ -d "$$dir" ]; then \
-			cd "$$dir"; \
-			rm -f *.out *.gb; \
-			cd ..; \
-		fi \
-	done
+        find . -name '*.out' -execdir $(RM) '{}' \;
+        find . -name '*.gb' -execdir $(RM) '{}' \;
+        find . -name '*.log' -execdir $(RM) '{}' \;	@for dir in *; do \
```
tautschnig: `$(RM) tests.log` and please merge the second commit into this one.

regression/goto-cc-goto-analyzer/Makefile
```diff
@@ -26,11 +26,6 @@ show:
...
-		rm -f tests.log; \
-		if [ -d "$$dir" ]; then \
-			cd "$$dir"; \
-			rm -f *.out *.gb; \
-			cd ..; \
-		fi \
-	done
+	find . -name '*.out' -execdir $(RM) '{}' \;
+	find . -name '*.gb' -execdir $(RM) '{}' \;
+	find . -name '*.log' -execdir $(RM) '{}' \;
```
tautschnig: `$(RM) tests.log`?

regression/cbmc/Makefile
```diff
@@ -28,6 +28,6 @@ show:
 	done;
 
 clean:
-	find -name '*.out' -execdir $(RM) '{}' \;
-	find -name '*.smt2' -execdir $(RM) '{}' \;
-	$(RM) tests.log
+	find . -name '*.out' -execdir $(RM) '{}' \;
+	find . -name '*.smt2' -execdir $(RM) '{}' \;
+	find . -name '*.log' -execdir $(RM) '{}' \;
```
tautschnig: `$(RM) tests.log`?

regression/Makefile
```diff
@@ -95,3 +95,4 @@ clean:
 			$(MAKE) -C "$$dir" clean; \
 		fi; \
 	done;
+	${RM} tests.log
```
tautschnig: `$(RM)` for consistency.

regression/acceleration/Makefile
SaswatPadhi: These 3 `find` invocations are missing `.`

tautschnig:
Overall comment: please something like `Fixes: #5910` in the commit message for the issue to be automatically closed as soon as this PR is merged.

regression/acceleration/Makefile
TGWDB: Thanks, I missed some in adding the `.`.

regression/acceleration/Makefile
TGWDB: I decided to do the more general and consistent approach that will remove `tests.log` as well as any others. I could change this to only the specific file if there is agreement not to use a general clean up instead?

TGWDB: I agree there is now inconsistency though since some use specific commands and others general. I made the general ones when factoring out loops with limit depth. This comment is also to note that the approach should be consistent, either all use `$(RM) tests.log` or all use `find . -name '*.log' -execdir $(RM) \;`.

tautschnig: Sounds good, I'm happy with either approach, as long as it is consistent. Doing the `find` in some places only suggests that those directories generate additional `.log` files, which may or may not be the case. That said, I am starting to wonder whether it wouldn't be much simpler (less code!) to move the common cleanup commands one level up. Each directory would then only have to care about any cleanup that's not covered by the parent-level `find` calls.

TGWDB: Chose to go with the specific and local version. Chosen so we are clear on where log files are expected and which log files are generated by the regression tests.

tautschnig (approved):
Thank you for cleaning this up!

SaswatPadhi (approved):

PR #5944, opened 2021-03-16
Add z3 package installation to CI pull request checks

thomasspriggs:
Add z3 package installation to CI pull request checks. This will enable subsequent PRs to add tests which require the z3 solver.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -17,7 +17,7 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
         run: |
           sudo apt-get update
-          sudo apt-get install --no-install-recommends -yq gcc gdb g++ maven jq flex bison libxml2-utils ccache cmake
+          sudo apt-get install --no-install-recommends -yq gcc gdb g++ maven jq flex bison libxml2-utils ccache cmake z3
       - name: Prepare ccache
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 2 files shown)


NlightNFotis (commented):
We probably also need `z3` for the coverage job as well.

SaswatPadhi:
The test case that I had added in #5904 seems to succeed now -- see: https://github.com/diffblue/cbmc/pull/5944/checks?check_run_id=2124973614#step:11:2064.

But the loop acceleration tests seem to be failing now. Checking if my PR caused this ...

SaswatPadhi:
Okay I just tried this on the `develop` branch -- ignoring the changes here (and in #5904). This is what I see:

- If z3 is installed on the system, these tests fail.
- If z3 is not installed, they pass -- they're not skipped.

This makes me think there's some code in the `acceleration` regression tests that "prefers" z3 and uses it when available (and fails because of some bug), but doesn't really "need" it because it runs fine without z3?

thomasspriggs:
> Okay I just tried this on the `develop` branch -- ignoring the changes here (and in #5904). This is what I see:
> 
> * If z3 is installed on the system, these tests fail.
> * If z3 is not installed, they pass -- they're not skipped.
> 
> This makes me think there's some code in the `acceleration` regression tests that "prefers" z3 and uses it when available (and fails because of some bug), but doesn't really "need" it because it runs fine without z3?

There is at least one bug in the implementation of `acceleration` which was causing failures when `z3` is available. I have pushed an additional commit to this branch to fix that issue.

thomasspriggs:
> We probably also need `z3` for the coverage job as well.

Added.

thomasspriggs:
I have raised a separate draft PR to run CI with the combination of installing z3 in CI + the fix from https://github.com/diffblue/cbmc/pull/5904. This CI running PR is here - https://github.com/diffblue/cbmc/pull/5949

chrisr-diffblue (approved):

NlightNFotis (approved):
LGTM

PR #5946, opened 2021-03-17
allow contracts on function declarations

kroening:
This changes the grammar of the C frontend to allow contracts on function
declarations.  This also removes the requirement that the contract clauses
appear in any particular order.


tautschnig (commented):
This looks nice, but given that tests are failing there obviously are some problems left to be fixed. Nothing blatantly obvious while reviewing the source code, thus requires more debugging.

feliperodri:
> This looks nice, but given that tests are failing there obviously are some problems left to be fixed. Nothing blatantly obvious while reviewing the source code, thus requires more debugging.

All code contracts are failing, which is expected since the regression suite also need to be updated (i.e., move all contracts to function declarations).

kroening:
The bit that is missing is to change the scope to the scope used for the parameters when parsing the contract expressions.

There should not be a need to change any of the existing tests, the existing syntax continues to work.

martin-cs:
On Wed, 2021-03-17 at 08:09 -0700, Felipe R. Monteiro wrote:
> > This looks nice, but given that tests are failing there obviously
> > are some problems left to be fixed. Nothing blatantly obvious while
> > reviewing the source code, thus requires more debugging.
> 
> All code contracts are failing, which is expected since the
> regression suite also need to be updated (i.e., move all contracts to
> function declarations).

I don't know your use-case fully but one cute feature SPARK has is that
you can put preconditions and postconditions on both the declaration
and definitions of functions.  You need that they imply each other but
it allows you to separate the pre/post you use to prove the function
and the pre/post you use as an abstraction in proving other code.  It
is really more useful when the code base gets larger but if we are in a
place where we /could/ implement this maybe we should think about if we
/should/.

Cheers,
 - Martin

src/ansi-c/expr2c.cpp
```diff
@@ -550,6 +550,19 @@ std::string expr2ct::convert_rec(
         dest.resize(dest.size()-1);
     }
 
+    // contract, if any
+    if(to_code_with_contract_type(src).requires().is_not_nil())
+      dest += " [[requires " +
+              convert(to_code_with_contract_type(src).requires()) + "]]";
```
tautschnig, line 557: Nit pick: add braces  around multi-line body.

tautschnig (approved):
Looks good to me, but it would be nice to modify some of the existing tests in regression/contract to make use of the new feature. Ideally with all the options now newly available: mixing contracts on declarations and definitions, mixing up their order. Also, what should we expect to happen when both a declaration and a definition have a contract (cf. @martin-cs' comment) - this might mean adding a KNOWNBUG test.

Issue #5947, opened 2021-03-17
SIGSEGV in loop acceleration with z3


SaswatPadhi:
**CBMC version:** `develop`
**Operating system:** Mac OS
**Exact command line resulting in the issue:** `make` in `regression/acceleration` with `z3` installed
**What behaviour did you expect:** Regression tests to pass
**What happened instead:** 3 of the regression tests crash with segmentation fault

This was exposed in #5944.

When `z3` is installed, the regression tests fail.
When `z3` is not installed, the regression tests pass!

`goto-instrument` seems to silently ignore errors in running the SMT solver, and treats errors as an UNSAT result!

SaswatPadhi:
I linked a PR that fixes the issue of solver errors being ignored, ~~and changes the default solver from Z3 to CPROVER_SMT2.~~

But I am leaving this issue open because we should debug z3 issues with loop acceleration code at some point in the future.

thomasspriggs:
This should be fixed in https://github.com/diffblue/cbmc/pull/5944/commits/fa657acfb8e43161a0cceab9c8ed61a10193b31c

thomasspriggs:
This has been fixed as part of https://github.com/diffblue/cbmc/pull/5944.

PR #5948, opened 2021-03-17
Correctly handle SMT solver errors in loop acceleration code

SaswatPadhi:
Related to #5944 and #5947.

This PR correctly handles SMT solver failures in loop acceleration routines. Earlier these errors were silently ignored, and were being treated as UNSAT.

### DIFF
--- a/src/goto-instrument/accelerate/scratch_program.cpp
+++ b/src/goto-instrument/accelerate/scratch_program.cpp
@@ -71,7 +71,21 @@ bool scratch_programt::check_sat(bool do_slice, guard_managert &guard_manager)
   std::cout << "Finished symex, invoking decision procedure.\n";
 #endif
 
-  return ((*checker)() == decision_proceduret::resultt::D_SATISFIABLE);
+  switch((*checker)())
+  {
+  case decision_proceduret::resultt::D_ERROR:
+    throw "error running the SMT solver";
+
+  case decision_proceduret::resultt::D_SATISFIABLE:
+    return true;
+
+  case decision_proceduret::resultt::D_UNSATISFIABLE:
+    return false;
+
+    UNREACHABLE;
+  }
+
+  UNREACHABLE;
 }
 
 exprt scratch_programt::eval(const exprt &e)


src/goto-instrument/accelerate/scratch_program.cpp
```diff
@@ -71,7 +71,12 @@ bool scratch_programt::check_sat(bool do_slice, guard_managert &guard_manager)
   std::cout << "Finished symex, invoking decision procedure.\n";
 #endif
 
-  return ((*checker)() == decision_proceduret::resultt::D_SATISFIABLE);
+  auto result = (*checker)();
+  INVARIANT(
```
kroening: `INVARIANT` seems wrong here -- errors do happen, and should be dealt with by giving a proper error message to the user.

Think of `INVARIANT` as something that you could prove to be true, always.

src/goto-instrument/accelerate/scratch_program.h
```diff
@@ -77,8 +77,8 @@ class scratch_programt:public goto_programt
       symex(mh, symbol_table, equation, options, path_storage, guard_manager),
       satcheck(util_make_unique<satcheckt>(mh)),
       satchecker(ns, *satcheck, mh),
-      z3(ns, "accelerate", "", "", smt2_dect::solvert::Z3, mh),
-      checker(&z3) // checker(&satchecker)
+      solver(ns, "accelerate", "", "", smt2_dect::solvert::CPROVER_SMT2, mh),
```
kroening: Do we know how this change of solver affects performance?

src/goto-instrument/accelerate/scratch_program.cpp
SaswatPadhi: Thanks!

I am not super familiar with the code base, so I am not sure what other construct I could use here to report an error to the user. The function returns a `bool`, so the return value can't capture 3 possibilities: SAT, UNSAT, ERROR.

@kroening  @thomasspriggs do you have any suggestions?

src/goto-instrument/accelerate/scratch_program.h
SaswatPadhi: Please ignore this change. I made a mistake in my local testing. This doesn't actually fix the error. @thomasspriggs made a separate change that fixes the root cause.

src/goto-instrument/accelerate/scratch_program.cpp
SaswatPadhi: Looking at the call sites for this function, I realized that there is exception handling code around it, so probably the right thing to do here is to `throw` an exception and let the caller handle it properly.

SaswatPadhi:
Build failures would be resolved after fa657ac is merged.

SaswatPadhi:
@kroening @thomasspriggs I have addressed your comments now. Could you please take another look?

Thanks.

kroening (approved):

src/goto-instrument/accelerate/scratch_program.cpp
SaswatPadhi: Added a `switch` as per Daniel's suggestion.

src/goto-instrument/accelerate/scratch_program.cpp
```diff
@@ -71,7 +71,21 @@ bool scratch_programt::check_sat(bool do_slice, guard_managert &guard_manager)
   std::cout << "Finished symex, invoking decision procedure.\n";
 #endif
 
-  return ((*checker)() == decision_proceduret::resultt::D_SATISFIABLE);
+  switch((*checker)())
+  {
+  case decision_proceduret::resultt::D_ERROR:
+    throw "error running the SMT solver";
```
chrisr-diffblue, line 77: Is there a structure exception you could throw instead of a raw string?

chrisr-diffblue (approved):

PR #5957, opened 2021-03-19
Remove "show" target from regression/*/Makefile

tautschnig:
This may have been useful to someone in the past, but a semi-interactive
(starting a new vim in a loop for each directory) approach with several
hundred files is no longer meaningful.


martin-cs (approved):
I'm fine with this and have not heard of a user for a long time (something makes me think I saw @peterschrammel using it in ... 2015?) but maybe we should give this one a little so that, in the unlikely event of anyone having strong feelings about it, they have time to object.

TGWDB (approved):

NlightNFotis (approved):

PR #5958, opened 2021-03-19
Run KNOWNBUG and THOROUGH regression tests in CI

tautschnig:
Use the check-ubuntu-20_04-cmake-gcc job to run KNOWNBUG (any test
reported as failure will tell us that a bug has unexpectedly been fixed)
an THOROUGH (tests that are expected to pass, but take longer to do so)
tests.


martin-cs (approved):
This is a nice idea.  Is it worth running this once first and PRing any of the KNOWNBUG tests that now work?  My impression is there are probably going to be some and they may need attention to figure out if they are really working on just incorrect tests.

NlightNFotis:
I'm not a huge fan of repurposing existing jobs to be honest.

Would it be possible that this gets done as a separate job?

Otherwise I agree that this is an excellent idea.

tautschnig:
> I'm not a huge fan of repurposing existing jobs to be honest.
> 
> Would it be possible that this gets done as a separate job?
> 
> Otherwise I agree that this is an excellent idea.

I guess I'm still living the times of limited Travis runners :-)  That's fixed now, added two new jobs.

tautschnig:
> This is a nice idea. Is it worth running this once first and PRing any of the KNOWNBUG tests that now work? My impression is there are probably going to be some and they may need attention to figure out if they are really working on just incorrect tests.

Belated response: yes, I'll try to `git bisect` to figure out when exactly tests were fixed. Will keep this PR in "Draft" state for the time being.

TGWDB (commented):
I like the idea of the KNOWNBUG checking to routinely check for improvements/changes. The THOROUGH on all PRs seems like it may be a lot of extra load?

.github/workflows/pull-request-checks.yaml
```diff
@@ -170,6 +170,90 @@ jobs:
...
+        with:
+          submodules: recursive
+      - name: Fetch dependencies
+        env:
+          # This is needed in addition to -yq to prevent apt-get from asking for
+          # user input
+          DEBIAN_FRONTEND: noninteractive
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -yq cmake ninja-build gcc g++ maven flex bison libxml2-utils ccache z3
```
TGWDB, line 186: The new z3 output on version is not in the new check.

.github/workflows/pull-request-checks.yaml
```diff
@@ -170,6 +170,90 @@ jobs:
...
```
TGWDB, line 233: z3 version output not here again

tautschnig:
> I like the idea of the KNOWNBUG checking to routinely check for improvements/changes. The THOROUGH on all PRs seems like it may be a lot of extra load?

I'm not sure we care? "CBMC / codecov-coverage-report" takes longer (102 vs 75 minutes), and GitHub happily runs all of the in parallel (up to 20 - according to https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits).

.github/workflows/pull-request-checks.yaml
tautschnig: Done!

.github/workflows/pull-request-checks.yaml
tautschnig: Done!

TGWDB (commented):
Broadly this PR looks good, but I would prefer that other fixes/changes that are not simply a CI change be a separate PR. For example https://github.com/diffblue/cbmc/pull/5958/commits/00847391c31f6e3fd2781eb379fa244656084420 appears to be bug fix that was found along the way (and great to have done), but not part of the titular change.

tautschnig:
> Broadly this PR looks good, but I would prefer that other fixes/changes that are not simply a CI change be a separate PR. For example [0084739](https://github.com/diffblue/cbmc/commit/00847391c31f6e3fd2781eb379fa244656084420) appears to be bug fix that was found along the way (and great to have done), but not part of the titular change.

Fair point, I've factored those out into #6102, #6103, #6104. The remaining commits could perhaps be factored out as well, but wouldn't be subject to CI-based testing until last commit of this PR is merged (which, in turn, can only be merged if fixing those issues first).

tautschnig:
@TGWDB All dependencies are now merged and this PR is rebased and should be ready for review.

NlightNFotis (approved):
I'm happy with this personally, provided we don't mark the extra CI jobs as required - they shouldn't be blocking PRs in my opinion.

PR #2278, opened 2018-06-05
Code contracts: match up introducing/removing parameter map modifications

tautschnig:
Code contracts that introduce local declarations result in nested
typecheck_declaration calls. The artificial __CPROVER_return_value symbol must
then be added/removed by a single call only.

PR #5959, opened 2021-03-19
Supports quantifiers in function contracts

ArenBabikian:
This PR adds support for quantified expressions in function contracts. The current implementation handles cases where a function contract is only composed of a single quantified expression. Additionally, nested quantified expressions are not yet supported.

This PR might be a continuation of #2278. 
Additionally, #5968 is a follow-up to this PR.

---

### DIFF
--- a/regression/contracts/function_check_02/test.desc
+++ b/regression/contracts/function_check_02/test.desc
@@ -1,6 +1,6 @@
-KNOWNBUG
+CORE
 main.c
---check-code-contracts
+--enforce-all-contracts
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
--- a/regression/contracts/quantifiers-exists-both-02/test.desc
+++ b/regression/contracts/quantifiers-exists-both-02/test.desc
@@ -1,0 +1,19 @@
+KNOWNBUG
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safely use __CPROVER_exists
+in a negative context (which is currently not always the case). By using the
+--enforce-all-contracts flag, this test assumes the statement in 
+__CPROVER_requires, then asserts the same statement (in __CPROVER_ensures),
+thus, verification should be successful.
+
+Known bug:
+The current implementation cannot handle a structure such as
+__CPROVER_assume(__CPROVER_exists(int i; pred(i))), where i is
+not explicitly bounded to a predefined range (i.e. if at least
+one of its bound is only declared and not defined).
--- a/regression/contracts/quantifiers-exists-ensures-01/main.c
+++ b/regression/contracts/quantifiers-exists-ensures-01/main.c
@@ -1,0 +1,20 @@
+// clang-format off
+int f1(int *arr) __CPROVER_ensures(__CPROVER_exists {
+  int i;
+  (0 <= i && i < 10) ==> arr[i] == i
+})
+// clang-format on
+{
+  for(int i = 0; i < 10; i++)
+  {
+    arr[i] = i;
+  }
+
+  return 0;
+}
+
+int main()
+{
+  int arr[10];
+  f1(arr);
+}
--- a/regression/contracts/quantifiers-exists-ensures-01/test.desc
+++ b/regression/contracts/quantifiers-exists-ensures-01/test.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safely use __CPROVER_exists
+in __CPROVER_ensures clauses. By using the --enforce-all-contracts flag, 
+goto-instrument will transform the __CPROVER_ensures clauses into an
+assertion and the verification remains sound when using __CPROVER_exists.
--- a/regression/contracts/quantifiers-exists-ensures-02/main.c
+++ b/regression/contracts/quantifiers-exists-ensures-02/main.c
@@ -1,0 +1,20 @@
+// clang-format off
+int f1(int *arr) __CPROVER_ensures(__CPROVER_exists {
+  int i;
+  (0 <= i && i < 10) ==> arr[i] != 0
+})
+// clang-format on
+{
+  for(int i = 0; i < 10; i++)
+  {
+    arr[i] = 0;
+  }
+
+  return 0;
+}
+
+int main()
+{
+  int arr[10];
+  f1(arr);
+}
--- a/regression/contracts/quantifiers-exists-ensures-02/test.desc
+++ b/regression/contracts/quantifiers-exists-ensures-02/test.desc
@@ -1,0 +1,19 @@
+KNOWNBUG
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+The purpose of this test is to ensure that we can safely use __CPROVER_exists
+in __CPROVER_ensures clauses. By using the --enforce-all-contracts flag, 
+goto-instrument will transform the __CPROVER_ensures clauses into an
+assertion and the verification remains sound when using __CPROVER_exists.
+
+Known Bug:
+We expect verification to fail because arr[i] is always equal to 0 for
+[0 <= i < 10]. In fact, we expect the (0 <= i && i < 10) statement to act as a
+range for i. However, in the current implementation of quantifier handling,
+the (0 <= i && i < 10) statement is not interpreted as an explicit range, but
+instead, as part of a logic formula, which causes verification to succeed.
--- a/regression/contracts/quantifiers-exists-requires-01/main.c
+++ b/regression/contracts/quantifiers-exists-requires-01/main.c
@@ -1,0 +1,15 @@
+// clang-format off
+int f1(int *arr) __CPROVER_requires(__CPROVER_exists {
+  int i;
+  (0 <= i && i < 10) ==> arr[i] == 4
+}) __CPROVER_ensures(__CPROVER_return_value == 0)
+// clang-format on
+{
+  return 0;
+}
+
+int main()
+{
+  int arr[10] = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
+  f1(arr);
+}
--- a/regression/contracts/quantifiers-exists-requires-01/test.desc
+++ b/regression/contracts/quantifiers-exists-requires-01/test.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safely use __CPROVER_exists
+in __CPROVER_requires clauses. By using the --replace-all-calls-with-contracts
+flag, goto-instrument will transform the __CPROVER_requires clauses into an
+assertion and the verification remains sound when using __CPROVER_exists.
--- a/regression/contracts/quantifiers-exists-requires-02/main.c
+++ b/regression/contracts/quantifiers-exists-requires-02/main.c
@@ -1,0 +1,15 @@
+// clang-format off
+int f1(int *arr) __CPROVER_requires(__CPROVER_exists {
+  int i;
+  (0 <= i && i < 10) ==> arr[i] == 1
+}) __CPROVER_ensures(__CPROVER_return_value == 0)
+// clang-format on
+{
+  return 0;
+}
+
+int main()
+{
+  int arr[10] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+  f1(arr);
+}
--- a/regression/contracts/quantifiers-exists-requires-02/test.desc
+++ b/regression/contracts/quantifiers-exists-requires-02/test.desc
@@ -1,0 +1,19 @@
+KNOWNBUG
+main.c
+--replace-all-calls-with-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+The purpose of this test is to ensure that we can safely use __CPROVER_exists
+in __CPROVER_requires clauses. By using the --replace-all-calls-with-contracts
+flag, goto-instrument will transform the __CPROVER_requires clauses into an
+assertion and the verification remains sound when using __CPROVER_exists.
+
+Known Bug:
+We expect verification to fail because arr[i] is never equal to 1 for
+[0 <= i < 10]. In fact, we expect the (0 <= i && i < 10) statement to act as a
+range for i. However, in the current implementation of quantifier handling,
+the (0 <= i && i < 10) statement is not interpreted as an explicit range, but
+instead, as part of a logic formula, which causes verification to succeed.
--- a/regression/contracts/quantifiers-forall-both-01/test.desc
+++ b/regression/contracts/quantifiers-forall-both-01/test.desc
@@ -1,0 +1,13 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safely use __CPROVER_forall
+in a negative context (which is currently not always the case). By using the
+--enforce-all-contracts flag, this test assumes the statement in
+__CPROVER_requires, then asserts the same statement (in __CPROVER_ensures), 
+thus, verification should be successful.
--- a/regression/contracts/quantifiers-forall-both-02/main.c
+++ b/regression/contracts/quantifiers-forall-both-02/main.c
@@ -1,0 +1,18 @@
+// clang-format off
+int f1(int *arr, int len) __CPROVER_requires(__CPROVER_forall {
+  int i;
+  (0 <= i && i < len) ==> arr[i] == 0
+}) __CPROVER_ensures(__CPROVER_forall {
+  int i;
+  (0 <= i && i < len) ==> arr[i] == 0
+})
+// clang-format on
+{
+  return 0;
+}
+
+int main()
+{
+  int len, arr[8];
+  f1(arr, len);
+}
--- a/regression/contracts/quantifiers-forall-ensures-01/test.desc
+++ b/regression/contracts/quantifiers-forall-ensures-01/test.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safely use __CPROVER_forall
+in __CPROVER_ensures clauses. By using the --enforce-all-contracts
+flag, goto-instrument will transform the __CPROVER_ensures clauses into an
+assertion and the verification remains sound when using __CPROVER_forall.
--- a/regression/contracts/quantifiers-forall-ensures-02/test.desc
+++ b/regression/contracts/quantifiers-forall-ensures-02/test.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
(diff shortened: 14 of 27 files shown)


regression/contracts/function_check_02/test.desc
```diff
@@ -1,6 +1,6 @@
 KNOWNBUG
 main.c
---check-code-contracts
+--enforce-all-contracts
```
feliperodri, line 3: If it's still a known bug, even with `--enforce-all-contracts`, does the description is still accurate? I believe this change doesn't belong to this PR.

regression/contracts/quantifiers-exists-ensures-01/main.c
```diff
@@ -0,0 +1,18 @@
+int f1(int *arr) __CPROVER_ensures(__CPROVER_exists {
+  int i;
+  (0 <= i && i < 10) && arr[i] == i
+})
```
feliperodri: We can only use quantifiers in positive context, correct? Is there anywhere where this is documented for now? Could we add a regression test to expose this problem using contracts (e.g., using in requires with enforce).

regression/contracts/quantifiers-exists-ensures-01/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test asserts the postconditions of f1.
```
feliperodri: The purpose of this test is to ensure we can safety use `__CPROVER_exists` in require clauses. The expected behavior is that using `--enforce-all-contracts` flag will transform requires into assertions and the verification remains sound when using `__CPROVER_exists`. Could you update the test description for this test (same for all the others). Also, `test.desc` lines should be < 80 characters.

regression/contracts/quantifiers-exists-ensures-02/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+Verification:
+This test asserts the postconditions of f1.
```
feliperodri: Update description as described above.

regression/contracts/quantifiers-exists-ensures-01/main.c
feliperodri: Why not `(0 <= i && i < 10) ==> arr[i] == i`?

regression/contracts/quantifiers-exists-ensures-02/main.c
```diff
@@ -0,0 +1,18 @@
+int f1(int *arr) __CPROVER_ensures(__CPROVER_exists {
+  int i;
+  (0 <= i && i < 10) && arr[i] != 0
```
feliperodri: Why not `(0 <= i && i < 10) ==> arr[i] != 0`?

regression/contracts/quantifiers-exists-requires-01/main.c
```diff
@@ -0,0 +1,13 @@
+int f1(int *arr) __CPROVER_requires(__CPROVER_exists {
+  int i;
+  (0 <= i && i < 10) && arr[i] == 4
```
feliperodri: Why not `(0 <= i && i < 10) ==> arr[i] == 4`?

regression/contracts/quantifiers-exists-requires-01/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test asserts the preconditions of f1.
```
feliperodri: Update description.

regression/contracts/quantifiers-exists-requires-02/main.c
```diff
@@ -0,0 +1,13 @@
+int f1(int *arr) __CPROVER_requires(__CPROVER_exists {
+  int i;
+  (0 <= i && i < 10) && arr[i] == 1
```
feliperodri: Why not `(0 <= i && i < 10) ==> arr[i] == 1`?

regression/contracts/quantifiers-exists-requires-02/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+Verification:
+This test asserts the preconditions of f1.
```
feliperodri: Update description.

regression/contracts/quantifiers-forall-ensures-01/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test asserts the preconditions of f1.
```
feliperodri: Update description.

regression/contracts/quantifiers-forall-ensures-02/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+Verification:
+This test asserts the postconditions of f1.
```
feliperodri: Update description.

regression/contracts/quantifiers-forall-requires-01/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test asserts the preconditions of f1.
```
feliperodri: Update description.

regression/contracts/quantifiers-forall-requires-02/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+Verification:
+This test asserts the preconditions of f1.
```
feliperodri: Update description.

src/goto-instrument/code_contracts.cpp
```diff
@@ -163,6 +163,41 @@ bool code_contractst::has_contract(const irep_idt fun_name)
...
+  replace_symbolt &replace,
+  irep_idt mode)
+{
+  // If the expression is a quantified expression, this function adds
+  // the quantified variable to the symbol table and to the expression map
+
+  // TODO Currently only works if the contract contains only a single
+  // quantified formula
+  // i.e. (1) the top-level element is a quantifier formula
+  // and (2) there are no inner quantifier formulae
```
feliperodri, line 177: Can we turn this into a `INVARIANT(...)`? Then we should add a GitHub issue to track that.

src/goto-instrument/code_contracts.cpp
```diff
@@ -163,6 +163,41 @@ bool code_contractst::has_contract(const irep_idt fun_name)
...
+  if(expression.id() == ID_exists || expression.id() == ID_forall)
```
feliperodri, line 180: Could we also turn this into an invariant instead of an if statement (e.g., `INVARIANT(expression.id() == ID_exists || expression.id() == ID_forall)`)? I think this function will only be called for these two expressions.

src/goto-instrument/code_contracts.cpp
```diff
@@ -149,7 +149,7 @@ static void check_apply_invariants(
 
   // change the back edge into assume(false) or assume(guard)
   loop_end->targets.clear();
-  loop_end->type=ASSUME;
+  loop_end->type = ASSUME;
```
feliperodri: This change doesn't belong here.

src/goto-instrument/code_contracts.cpp
```diff
@@ -230,7 +265,7 @@ bool code_contractst::apply_function_contract(
   }
 
   // Replace formal parameters
-  code_function_callt::argumentst::const_iterator a_it=
+  code_function_callt::argumentst::const_iterator a_it =
```
feliperodri: This change doesn't belong here.

src/goto-instrument/code_contracts.cpp
```diff
@@ -745,7 +785,7 @@ void code_contractst::add_contract_check(
                        .symbol_expr();
     check.add(goto_programt::make_decl(r, skip->source_location));
 
-    call.lhs()=r;
+    call.lhs() = r;
```
feliperodri: This change doesn't belong here.

src/goto-instrument/code_contracts.cpp
```diff
@@ -440,38 +480,38 @@ void code_contractst::instrument_call_statement(
...
+    assigns_clauset called_assigns_clause(
+      called_assigns, *this, function_id, log);
+    exprt compatible =
+      assigns_clause.compatible_expression(called_assigns_clause);
+    goto_programt alias_assertion;
+    alias_assertion.add(goto_programt::make_assertion(
+      compatible, instruction_iterator->source_location));
+    program.insert_before_swap(instruction_iterator, alias_assertion);
+    ++instruction_iterator;
+  }
```
feliperodri: You just moved this code, correct?

src/goto-instrument/code_contracts.h
```diff
@@ -165,6 +166,13 @@ class code_contractst
 
   void
   add_contract_check(const irep_idt &, const irep_idt &, goto_programt &dest);
+
+  // If the expression is a quantified expression, this function adds
+  // the quantified variable to the symbol table and to the expression map
```
feliperodri: Keep comment style for methods.
```cpp
  /// If the expression is a quantified expression, this function adds
  /// the quantified variable to the symbol table and to the expression map
```

feliperodri (changes requested):

regression/contracts/quantifiers-exists-ensures-01/main.c
SaswatPadhi: Yes. `forall i . (0 <= i && i < 10) && arr[i] == i` is simply false. It (and similar predicates across all regression tests) should use implication.

ArenBabikian: I am using `(0 <= i && 1 < 10) && arr[i] == i` within an existential quantifier. I believe that it is correct as is.

SaswatPadhi: The correctness would depend on what we want to check about the function. Is it:
1. If there is some value within the range then the property holds, or
2. There must be a value within the range _and_ the property should hold

Consider this example:
```
exists x : char . 1000 <= (int) x <= 5000 ==> pred(x)
```
This would evaluate to `true` and models (1) correctly because there is no char between `1000` and `5000` in the first place.

```
exists x : char . 1000 <= (int) x <= 5000 && pred(x)
```
This would evaluate to `false` and models (2) correctly.

src/goto-instrument/code_contracts.cpp
ArenBabikian: Yes, I had just fixed the indentation, but I will get rid of these changes as well

regression/contracts/function_check_02/test.desc
ArenBabikian: Sory, it is no longer a known bug, I will adjust the first line.

src/goto-instrument/code_contracts.cpp
ArenBabikian: I think using an INVARIANT would restrict this function too much, especially down the line, when this function will become recursive. In my current design, I call this function on every function contract, regardless whether they contain a quantified expression or not. If no quantified expression is found, then this function simply does nothing. However, having an INVARIANT would cause the program to fail if no quantifier is found, which is not what desired behavioor, in my opinion.

ArenBabikian:
I have assess all code review comments with the exception of [this comment](https://github.com/diffblue/cbmc/pull/5959/files/160d8873c7f842afdf533d4448f428d48a6f36f7#diff-1ff6abca57de19a68456666ed5c25ce09fcaae20fdc9c2b109cb60cb311c183a) about using `INVARIANT()` for handling  quantifiers (I believe that in the medium-to-long-term, this would not be useful).

regression/contracts/quantifiers-exists-ensures-01/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safety use __CPROVER_exists
```
feliperodri: `s/safety/safely/g`

regression/contracts/quantifiers-exists-ensures-02/test.desc
```diff
@@ -0,0 +1,19 @@
+KNOWNBUG
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+The purpose of this test is to ensure that we can safety use __CPROVER_exists
```
feliperodri: `s/safety/safely/g`

regression/contracts/quantifiers-exists-requires-01/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safety use __CPROVER_exists
```
feliperodri: `s/safety/safely/g`

regression/contracts/quantifiers-exists-requires-02/test.desc
```diff
@@ -0,0 +1,19 @@
+KNOWNBUG
+main.c
+--replace-all-calls-with-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+The purpose of this test is to ensure that we can safety use __CPROVER_exists
```
feliperodri: `s/safety/safely/g`

regression/contracts/quantifiers-forall-both-01/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safely use __CPROVER_forall
+in a negative context. By using the --enforce-all-contracts flag, this test
```
feliperodri: We cannot *always* safely use `__CPROVER_forall` in negative contexts, correct? I'd just add this here too. Same for `__CPROVER_exists` test cases.

regression/contracts/quantifiers-forall-ensures-01/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safety use __CPROVER_forall
```
feliperodri: `s/safety/safely/g`

regression/contracts/quantifiers-forall-ensures-02/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+The purpose of this test is to ensure that we can safety use __CPROVER_forall
```
feliperodri: `s/safety/safely/g`

regression/contracts/quantifiers-forall-requires-01/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+The purpose of this test is to ensure that we can safety use __CPROVER_forall
```
feliperodri: `s/safety/safely/g`

regression/contracts/quantifiers-forall-requires-02/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+The purpose of this test is to ensure that we can safety use __CPROVER_forall
```
feliperodri: `s/safety/safely/g`

src/goto-instrument/code_contracts.cpp
feliperodri: Could you tag here the GitHub issue to track this comment?

feliperodri (approved):
LGTM, I only spotted a few more typos and minor comments. Thank you for implementing all comments!!

regression/contracts/quantifiers-exists-both-02/test.desc
```diff
@@ -0,0 +1,19 @@
...
+in a negative context (which is currently not always the case). By using the
+--enforce-all-contracts flag, this test assumes the statement in 
+__CPROVER_requires, then asserts the same statement (in __CPROVER_ensures),
+thus, verification should be successful.
+
+Known bug:
+The current implementation cannot handle a structure such as
+__CPROVER_assume(__CPROVER_exists(int i; pred(i))), where i is
+not explicitly bounded to a predefined range (i.e. if at least
+one of its bound is only declared and not defined).
```
martin-cs, line 19: Have you tried the SMT back-end?

src/goto-instrument/code_contracts.cpp
martin-cs: I don't think it should be an `INVARIANT` because it can be triggered by user input.  Some kind of warning / error and an issue would be amazing.

martin-cs (approved):
Looks good.  In the perfect world, the code changes and the regression tests would be in separate commits but don't worry about changing that unless you have to push a new version.

regression/contracts/quantifiers-forall-both-02/main.c
```diff
@@ -0,0 +1,18 @@
+// clang-format off
+int f1(int *arr, int *len) __CPROVER_requires(__CPROVER_forall {
+  int i;
+  (0 <= i && i < len) ==> arr[i] == 0
```
SaswatPadhi, line 4: This looks incorrect.

Did you mean `int len` in place of `int *len` in the function signature?

SaswatPadhi (changes requested):

regression/contracts/quantifiers-forall-both-02/main.c
```diff
@@ -0,0 +1,18 @@
...
+})
+// clang-format on
+{
+  return 0;
+}
+
+int main()
+{
+  int len, arr[8];
```
SaswatPadhi, line 16: I think `len` should also be initialized (to 8).

ArenBabikian: Such a case is already tested for in the `quantifiers-forall-both-01` test, and it works with the current implementation. The purpose of this test is in fact to check wether CBMC can handle cases where the range of the quantified variable is not explicitly set a priori.

SaswatPadhi (approved):

regression/contracts/quantifiers-exists-ensures-01/main.c
SaswatPadhi: @ArenBabikian your original change above was fine in this case.

Check (1) above actually checks: there exists a value such that if it's in the range, then it satisfies property.

The proper way to handle the "if there is a value" case would be to add an implication for non-emptiness:
https://github.com/diffblue/cbmc/blob/66d9d175919787fb08107fe4d28759f284be8f46/regression/contracts/quantifiers-loop-03/main.c#L17-L24

Sorry about the confusion; fixing this (among other issues) in #6145.

PR #5961, opened 2021-03-19
Fix pointer havocing within loop invariant contracts

### LINKED ISSUE #5960: Loop invariant contract havocs a pointer instead of havocing its value
**CBMC version:** `develop`

**Operating system:** Mac OS 10.15.7

**Test case:**

```c
#include <assert.h>

void main()
{
  char *data = malloc(1);
  *data = 42;

  unsigned i;
  while(i > 0)
    __CPROVER_loop_invariant(*data == 42)
    {
      *data = 42;
      i--;
    }

  assert(*data = 42);
}
```

```
goto-cc -o test.gb test.c
goto-instrument --enforce-all-contracts test.gb test-loop.gb
cbmc --pointer-check test-loop.gb
```

CBMC should be able to verify the goto program with the loop contracts properly enforced.

**What happened instead:**

CBMC complains that `data` is assigned to `NULL` and thus `*data = 42` may be invalid.

This happens because the `__CPROVER_loop_invariant` contract incorrectly havocs the write set of the loop -- instead of havocing `*data`, it havocs `data` and non-deterministically sets it to `NULL`.

SaswatPadhi:
Fixes #5960.

Please see the linked bug report for more information.

### DIFF
--- a/regression/contracts/invar_check_pointer_modifies-01/main.c
+++ b/regression/contracts/invar_check_pointer_modifies-01/main.c
@@ -1,0 +1,18 @@
+#include <assert.h>
+#include <stdlib.h>
+
+void main()
+{
+  char *data = malloc(1);
+  *data = 42;
+
+  unsigned i;
+  while(i > 0)
+    __CPROVER_loop_invariant(*data == 42)
+    {
+      *data = 42;
+      i--;
+    }
+
+  assert(*data = 42);
+}
--- a/src/goto-instrument/loop_utils.cpp
+++ b/src/goto-instrument/loop_utils.cpp
@@ -65,10 +65,9 @@ void get_modifies_lhs(
     modifies.insert(lhs);
   else if(lhs.id()==ID_dereference)
   {
-    modifiest m=local_may_alias.get(t, to_dereference_expr(lhs).pointer());
-    for(modifiest::const_iterator m_it=m.begin();
-        m_it!=m.end(); m_it++)
-      get_modifies_lhs(local_may_alias, t, *m_it, modifies);
+    const auto &pointer = to_dereference_expr(lhs).pointer();
+    for(const auto &mod : local_may_alias.get(t, pointer))
+      modifies.insert(dereference_exprt{mod});
   }
   else if(lhs.id()==ID_member)
   {
--- a/regression/contracts/invar_check_pointer_modifies-01/test.desc
+++ b/regression/contracts/invar_check_pointer_modifies-01/test.desc
@@ -1,0 +1,14 @@
+CORE
+main.c
+--enforce-all-contracts --pointer-check
+^EXIT=0$
+^SIGNAL=0$
+^\[main.1\] .* Check loop invariant before entry: SUCCESS$
+^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
+^\[main.assertion.1\] .* assertion \*data = 42: SUCCESS$
+^VERIFICATION SUCCESSFUL$
+--
+^\[main.pointer_dereference.*\] line .* dereference failure: pointer NULL in \*data: FAILURE
+--
+This test checks that modifications to deref-ed pointers are correctly handled.
+In such cases, pointers should not be havoc'ed, only the value should be.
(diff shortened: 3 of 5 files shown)


regression/contracts/invar_check_pointer_modifies/main.c
```diff
@@ -0,0 +1,17 @@
+#include <assert.h>
```
feliperodri:
```c
#include <stdlib.h>
```
because of malloc.

regression/contracts/invar_check_pointer_modifies/test.desc
```diff
@@ -0,0 +1,14 @@
+CORE
+main.c
+--enforce-all-contracts --pointer-check
+^EXIT=0$
+^SIGNAL=0$
+^\[main.1\] .* Check loop invariant before entry: SUCCESS$
+^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
+^\[main.assertion.1\] .* assertion r == 0: SUCCESS$
+^VERIFICATION SUCCESSFUL$
```
feliperodri: Can we also add a negative test?

src/goto-instrument/loop_utils.cpp
```diff
@@ -65,10 +65,9 @@ void get_modifies_lhs(
     modifies.insert(lhs);
   else if(lhs.id()==ID_dereference)
   {
-    modifiest m=local_may_alias.get(t, to_dereference_expr(lhs).pointer());
-    for(modifiest::const_iterator m_it=m.begin();
-        m_it!=m.end(); m_it++)
-      get_modifies_lhs(local_may_alias, t, *m_it, modifies);
+    auto pointer = to_dereference_expr(lhs).pointer();
```
feliperodri, line 68: `address_of_exprt`?

feliperodri (changes requested):

regression/contracts/invar_check_pointer_modifies/main.c
SaswatPadhi: Thanks. Done.

regression/contracts/invar_check_pointer_modifies/test.desc
SaswatPadhi: Good idea! Added one for aliased pointers :)

regression/contracts/invar_check_pointer_modifies-01/main.c
```diff
@@ -0,0 +1,18 @@
+#include <assert.h>
+#include <stdlib.h>
+
```
tautschnig, line 3: Random location for comment: would you mind merging the second commit (the regression test) into the first one? When later looking at git history, this makes it much easier to understand why a particular change was made.

src/goto-instrument/loop_utils.cpp
tautschnig: `const auto &pointer` (or perhaps `const exprt &pointer`)

feliperodri (approved):
LGTM.

src/goto-instrument/loop_utils.cpp
```diff
@@ -65,10 +65,9 @@ void get_modifies_lhs(
...
+    for(const auto &mod : local_may_alias.get(t, pointer))
+      modifies.insert(dereference_exprt{mod, mod.type().subtype()});
```
tautschnig: `dereference_exprt{mod}` should work, I believe.

tautschnig (approved):

regression/contracts/invar_check_pointer_modifies-01/main.c
SaswatPadhi: Makes sense, thanks for the quick feedback!

src/goto-instrument/loop_utils.cpp
SaswatPadhi: Thanks. Done.

src/goto-instrument/loop_utils.cpp
SaswatPadhi: Thanks. Done.

peterschrammel (approved):

PR #5962, opened 2021-03-19
Do not assume that build architecture has byte/char==8 bits

tautschnig:
The C/C++ standard does not guarantee this, and limits.h/climits has the
CHAR_BIT define set properly.

All changes in this commit only refer to the platform the tool is being
built for. The analysis target requires a separate set of changes where
the byte width must be taken from the configuration or annotations
within expressions.

This is a first part of #917, now factored into a PR of its own.

### DIFF
--- a/jbmc/src/janalyzer/janalyzer_parse_options.cpp
+++ b/jbmc/src/janalyzer/janalyzer_parse_options.cpp
@@ -11,6 +11,7 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "janalyzer_parse_options.h"
 
+#include <climits>
 #include <cstdlib> // exit()
 #include <fstream>
 #include <iostream>
@@ -357,8 +358,9 @@ int janalyzer_parse_optionst::doit()
   // Print a banner
   //
   log.status() << "JANALYZER version " << CBMC_VERSION << " "
-               << sizeof(void *) * 8 << "-bit " << config.this_architecture()
-               << " " << config.this_operating_system() << messaget::eom;
+               << sizeof(void *) * CHAR_BIT << "-bit "
+               << config.this_architecture() << " "
+               << config.this_operating_system() << messaget::eom;
 
   register_languages();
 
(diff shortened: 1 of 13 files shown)


martin-cs (approved):
Glad to see you and the author of #5873 managed to sort things out ( https://github.com/diffblue/cbmc/pull/5873#discussion_r584146281 )

TGWDB (approved):

peterschrammel (approved):

kroening:
Note that the commit message isn't entirely true: The changes to ansi-c/library/gcc.c affect the analysis platform, not the build platform.

kroening (approved):

tautschnig:
> Note that the commit message isn't entirely true: The changes to ansi-c/library/gcc.c affect the analysis platform, not the build platform.

Oh, indeed, I've moved those changes to the commit in #917 instead.

jbmc/src/janalyzer/janalyzer_parse_options.cpp
```diff
@@ -357,8 +358,9 @@ int janalyzer_parse_optionst::doit()
   // Print a banner
   //
   log.status() << "JANALYZER version " << CBMC_VERSION << " "
-               << sizeof(void *) * 8 << "-bit " << config.this_architecture()
-               << " " << config.this_operating_system() << messaget::eom;
+               << sizeof(void *) * CHAR_BIT << "-bit "
```
chrisr-diffblue, line 361: Out-of-scope for this PR, but give how this exact log.status pattern is duplicated several times, it does rather feel like a utility function could be useful :-)

chrisr-diffblue (approved):
Thanks for the cleanup - much better :-)

jbmc/src/janalyzer/janalyzer_parse_options.cpp
tautschnig: Thank you for highlighting this. Now PR'ed in #6089.

PR #5963, opened 2021-03-22
Ignore *.tmp files generated by clang

SaswatPadhi:
- [ ] NA ~~Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ Release/*
 *.od
 *.d
 *.o
+*.tmp
 *.obj
 *.a
 *.lib


tautschnig:
Why would clang leave any temporary files lying around? The linked LLVM PR does explain why clang might temporarily (!) have .tmp files, but I don't see why those would need to be git-ignored for they are transient.

SaswatPadhi:
Hi Michael.
Yes, you're right that the .tmp files are transient, and clang doesn't leave them lying around after compilation. But I do think ignoring temporary files is a good idea, even if they are transient. This change will allow me to, for instance, commit my changes when a build is still running:
- I make some changes and run `make` to make sure everything compiles
- I see that the files I changed compiled fine, but the whole build is still running
- I can't do a `git add -A .` at the top level because of these .tmp files

tautschnig:
Thank you for elaborating. My concerns are twofold:
1) Baking implementation details (that can and will change outside of our control) of tool chains into our repository configuration.
2) Shadowing actual bugs.

Quite honestly: is `git add -A .` really useful? Adding new files is somewhat rare (perhaps outside `regression/`) and often limited to few files. I would have thought that a focussed `git add <file name>` would typically do a reasonable job?

SaswatPadhi:
I use `git add -A .` all the time -- it's very useful for simultaneously adding multiple edits (changed files, new files, renamed files etc). May be I am just lazy 😄, but this is exactly where .gitignore is useful. If we always add specific files using `git add <filename>` then we wouldn't need a .gitignore at all.

The concern about adding rules specific to external tools is a valid one, but `.tmp` seems standard for temporary files (which is why LLVM changed it from `file.o-XXXXX` to `file.o.tmp`).

I guess a nicer alternative could be to make LLVM store these temporary outside of the source dir -- is there a compiler option for that?

tautschnig:
> I guess a nicer alternative could be to make LLVM store these temporary outside of the source dir -- is there a compiler option for that?

Maybe you just want to do out-of-tree builds via `cmake`?

SaswatPadhi:
Thanks!! Just tried building with `cmake` :) Works perfectly.

Closing this issue.

PR #5968, opened 2021-03-22
Supports nested quantifiers in function contracts

ArenBabikian:
This PR adds support for nested quantified expressions in function contracts. It is a direct extension to #5959 

### DIFF
--- a/regression/contracts/quantifiers-nested-01/test.desc
+++ b/regression/contracts/quantifiers-nested-01/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test case checks the handling of a forall expression
+nested within another forall expression.
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -26,6 +26,7 @@ Date: February 2016
 #include <util/expr_util.h>
 #include <util/format_type.h>
 #include <util/fresh_symbol.h>
+#include <util/mathematical_expr.h>
 #include <util/mathematical_types.h>
 #include <util/message.h>
 #include <util/pointer_offset_size.h>
@@ -168,35 +169,68 @@ void code_contractst::add_quantified_variable(
   replace_symbolt &replace,
   irep_idt mode)
 {
-  // If the expression is a quantified expression, this function adds
-  // the quantified variable to the symbol table and to the expression map
-
-  // TODO Currently only works if the contract contains only a single
-  // quantified formula
-  // i.e. (1) the top-level element is a quantifier formula
-  // and (2) there are no inner quantifier formulae
-  // This TODO is handled in PR #5968
-
-  if(expression.id() == ID_exists || expression.id() == ID_forall)
+  if(expression.id() == ID_not || expression.id() == ID_typecast)
   {
-    // get quantified symbol
-    exprt tuple = expression.operands().front();
-    exprt quantified_variable = tuple.operands().front();
-    symbol_exprt quantified_symbol = to_symbol_expr(quantified_variable);
+    // For unary connectives, recursively check for
+    // nested quantified formulae in the term
+    const auto &unary_expression = to_unary_expr(expression);
+    add_quantified_variable(unary_expression.op(), replace, mode);
+  }
+  if(expression.id() == ID_notequal || expression.id() == ID_implies)
+  {
+    // For binary connectives, recursively check for
+    // nested quantified formulae in the left and right terms
+    const auto &binary_expression = to_binary_expr(expression);
+    add_quantified_variable(binary_expression.lhs(), replace, mode);
+    add_quantified_variable(binary_expression.rhs(), replace, mode);
+  }
+  if(expression.id() == ID_if)
+  {
+    // For ternary connectives, recursively check for
+    // nested quantified formulae in all three terms
+    const auto &if_expression = to_if_expr(expression);
+    add_quantified_variable(if_expression.cond(), replace, mode);
+    add_quantified_variable(if_expression.true_case(), replace, mode);
+    add_quantified_variable(if_expression.false_case(), replace, mode);
+  }
+  if(expression.id() == ID_and || expression.id() == ID_or)
+  {
+    // For multi-ary connectives, recursively check for
+    // nested quantified formulae in all terms
+    const auto &multi_ary_expression = to_multi_ary_expr(expression);
+    for(const auto &operand : multi_ary_expression.operands())
+    {
+      add_quantified_variable(operand, replace, mode);
+    }
+  }
+  else if(expression.id() == ID_exists || expression.id() == ID_forall)
+  {
+    // When a quantifier expression is found,
+    // 1. get quantified variables
+    const auto &quantifier_expression = to_quantifier_expr(expression);
+    const auto &quantified_variables = quantifier_expression.variables();
+    for(const auto &quantified_variable : quantified_variables)
+    {
+      // for each quantified variable...
+      const auto &quantified_symbol = to_symbol_expr(quantified_variable);
 
-    // create fresh symbol
-    symbolt new_symbol = get_fresh_aux_symbol(
-      quantified_symbol.type(),
-      id2string(quantified_symbol.get_identifier()),
-      "tmp",
-      quantified_symbol.source_location(),
-      mode,
-      symbol_table);
+      // 1.1 create fresh symbol
+      symbolt new_symbol = get_fresh_aux_symbol(
+        quantified_symbol.type(),
+        id2string(quantified_symbol.get_identifier()),
+        "tmp",
+        quantified_symbol.source_location(),
+        mode,
+        symbol_table);
 
-    // add created fresh symbol to expression map
-    symbol_exprt q(
-      quantified_symbol.get_identifier(), quantified_symbol.type());
-    replace.insert(q, new_symbol.symbol_expr());
+      // 1.2 add created fresh symbol to expression map
+      symbol_exprt q(
+        quantified_symbol.get_identifier(), quantified_symbol.type());
+      replace.insert(q, new_symbol.symbol_expr());
+
+      // 1.3 recursively check for nested quantified formulae
+      add_quantified_variable(quantifier_expression.where(), replace, mode);
+    }
   }
 }
 
--- a/regression/contracts/quantifiers-nested-01/main.c
+++ b/regression/contracts/quantifiers-nested-01/main.c
@@ -1,0 +1,24 @@
+// clang-format off
+int f1(int *arr) __CPROVER_ensures(__CPROVER_forall {
+  int i;
+  __CPROVER_forall
+  {
+    int j;
+    (0 <= i && i < 10 && i <= j && j < 10) ==> arr[i] <= arr[j]
+  }
+})
+// clang-format on
+{
+  for(int i = 0; i < 10; i++)
+  {
+    arr[i] = i;
+  }
+
+  return 0;
+}
+
+int main()
+{
+  int arr[10];
+  f1(arr);
+}
--- a/regression/contracts/quantifiers-nested-02/main.c
+++ b/regression/contracts/quantifiers-nested-02/main.c
@@ -1,0 +1,19 @@
+// clang-format off
+int f1(int *arr) __CPROVER_requires(__CPROVER_forall {
+  int i;
+  0 <= i && i < 9 ==> __CPROVER_forall
+  {
+    int j;
+    (i <= j && j < 10) ==> arr[i] <= arr[j]
+  }
+}) __CPROVER_ensures(__CPROVER_return_value == 0)
+// clang-format on
+{
+  return 0;
+}
+
+int main()
+{
+  int arr[10] = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
+  f1(arr);
+}
--- a/regression/contracts/quantifiers-nested-02/test.desc
+++ b/regression/contracts/quantifiers-nested-02/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+--replace-all-calls-with-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test case checks the handling of a forall expression
+nested within an implication.
--- a/regression/contracts/quantifiers-nested-03/main.c
+++ b/regression/contracts/quantifiers-nested-03/main.c
@@ -1,0 +1,19 @@
+int f1(int *arr)
+  __CPROVER_ensures(__CPROVER_return_value == 0 && __CPROVER_exists {
+    int i;
+    (0 <= i && i < 10) && arr[i] == i
+  })
+{
+  for(int i = 0; i < 10; i++)
+  {
+    arr[i] = i;
+  }
+
+  return 0;
+}
+
+int main()
+{
+  int arr[10];
+  f1(arr);
+}
--- a/regression/contracts/quantifiers-nested-03/test.desc
+++ b/regression/contracts/quantifiers-nested-03/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test case checks the handling of an exists expression
+nested within a conjunction.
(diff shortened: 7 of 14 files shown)


src/ansi-c/c_typecheck_base.cpp
```diff
@@ -748,11 +748,15 @@ void c_typecheck_baset::typecheck_declaration(
       typet ret_type = void_type();
       if(new_symbol.type.id()==ID_code)
         ret_type=to_code_type(new_symbol.type).return_type();
-      assert(parameter_map.empty());
       if(ret_type.id()!=ID_empty)
+      {
+        DATA_INVARIANT(
```
martin-cs: If `parameter_map` should always be empty then `DATA_INVARIANT`.  If it should be empty at this point then `INVARIANT`.

src/ansi-c/parser.y
```diff
@@ -101,6 +101,7 @@ extern char *yyansi_ctext;
 %token TOK_NE        "!="
 %token TOK_ANDAND    "&&"
 %token TOK_OROR      "||"
+%token TOK_RANGE     ".."
```
martin-cs: Syntax extensions are always a bit controversial.  Is there an ACSL construct you could use?

src/goto-instrument/code_contracts.cpp
```diff
@@ -961,3 +1005,713 @@ bool code_contractst::enforce_contracts(
...
+    {
+      exprt::operandst conjuncts;
+      conjuncts.push_back(result);
+      conjuncts.push_back(current_target_compatible);
+      result = conjunction(conjuncts);
+    }
+  }
+
+  return result;
+}
```
martin-cs: A. Please can these changes be broken up a bit.
B. Adding 70% more code to this file is quite a bit.  Is there an obvious way of splitting this up / out?

src/util/std_expr.h
```diff
@@ -824,6 +824,51 @@ inline multi_ary_exprt &to_multi_ary_expr(exprt &expr)
   return static_cast<multi_ary_exprt &>(expr);
 }
 
+/// \brief A base class for int-range expressions
+class range_exprt : public expr_protectedt
```
martin-cs: How do these relate to intervals?

martin-cs (commented):
Has potential but I think some changes may be needed.

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +168,30 @@ void code_contractst::add_quantified_variable(
...
-    // get quantified symbol
-    exprt tuple = expression.operands().front();
-    exprt quantified_variable = tuple.operands().front();
+  // This function recursively searches the expression to find nested or
+  // non-nested quantified expressions. When a quantified expression is found,
+  // the quantified variable is added to the symbol table
+  // and to the expression map.
+  if(
+    expression.id() == ID_and || expression.id() == ID_or ||
+    expression.id() == ID_implies)
```
SaswatPadhi: What about negations? And the ternary `?:` operator?

Could you please also add regression tests for them?

regression/contracts/quantifiers-nested-01/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test asserts the postconditions of f1.
```
SaswatPadhi: Could we update the descriptions of these test cases to indicate exactly what kind of nesting we are checking for? For instance, this test case is checking a forall inside a forall. Other tests check for forall inside conjunction / disjunction etc.

SaswatPadhi (changes requested):

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +168,30 @@ void code_contractst::add_quantified_variable(
...
-
-  if(expression.id() == ID_exists || expression.id() == ID_forall)
-  {
-    // get quantified symbol
-    exprt tuple = expression.operands().front();
-    exprt quantified_variable = tuple.operands().front();
+  // This function recursively searches the expression to find nested or
+  // non-nested quantified expressions. When a quantified expression is found,
+  // the quantified variable is added to the symbol table
+  // and to the expression map.
```
feliperodri: You don't need to repeat this here and in the header file.

feliperodri (approved):
LGTM module the requested changes from @SaswatPadhi

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +168,46 @@ void code_contractst::add_quantified_variable(
...
-
-  if(expression.id() == ID_exists || expression.id() == ID_forall)
+  if(
+    expression.id() == ID_not || expression.id() == ID_notequal ||
+    expression.id() == ID_typecast)
   {
-    // get quantified symbol
-    exprt tuple = expression.operands().front();
-    exprt quantified_variable = tuple.operands().front();
+    // For unary connectives, recursively check for
```
martin-cs, line 174: `ID_notequal` is not unary?

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +168,46 @@ void code_contractst::add_quantified_variable(
...
+    // nested quantified formulae in the term
+    exprt term = expression.operands().at(0);
```
martin-cs: This works but converting to a `unary_exprt` is better style.

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +168,46 @@ void code_contractst::add_quantified_variable(
...
+    add_quantified_variable(term, replace, mode);
+  }
+  if(
+    expression.id() == ID_and || expression.id() == ID_or ||
+    expression.id() == ID_implies)
+  {
+    // For binary connectives, recursively check for
+    // nested quantified formulae in the left and right terms
+    exprt left = expression.operands().at(0);
```
martin-cs: `ID_and` and `ID_or` are not necessarily binary.

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +168,46 @@ void code_contractst::add_quantified_variable(
...
+    add_quantified_variable(left, replace, mode);
+    exprt right = expression.operands().at(1);
+    add_quantified_variable(right, replace, mode);
+  }
+  if(expression.id() == ID_if)
+  {
+    // For ternary connectives, recursively check for
+    // nested quantified formulae in all three terms
```
martin-cs, line 190: Please cast to an `if_exprt`.

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +168,46 @@ void code_contractst::add_quantified_variable(
...
+    exprt first = expression.operands().at(1);
+    add_quantified_variable(first, replace, mode);
+    exprt second = expression.operands().at(2);
+    add_quantified_variable(second, replace, mode);
+  }
+  else if(expression.id() == ID_exists || expression.id() == ID_forall)
+  {
+    // When a quantified expression is found,
+    // 1. get quantified symbol
+    exprt tuple = expression.operands().at(0);
```
martin-cs: `quantifier_exprt`

martin-cs (changes requested):
Looks good in principle but please can you use the appropriate `exprt` types rather than just pulling things out of the operators array.

src/goto-instrument/code_contracts.cpp
ArenBabikian: Yes, `ID_notequal` is unary, at least in this [regression test](https://github.com/diffblue/cbmc/pull/5968/files#diff-d669317d65162d540ce34a83fff0edd7695fea0bc385d2d2377e8ed2e8912ed1). In fact, `ID_notequal` shows up when using the ternary ITE expression in contracts, I do not actually use it as is.

src/goto-instrument/code_contracts.cpp
ArenBabikian: According to [this regression test](https://github.com/diffblue/cbmc/blob/c6a8eefa516c6c1f439b2193b59d47766c76168f/regression/contracts/quantifiers-nested-04/main.c), it seems that conjunctions/disjunctions with more than 2 terms are represented as nested binary conjunctions/disjunctions. In fact, the regression test passes when I cast the `ID_or` to a `binary_exprt`.

src/goto-instrument/code_contracts.cpp
ArenBabikian: You are correct, I have moved it to the section for binary operators.

src/goto-instrument/code_contracts.cpp
SaswatPadhi: May be that decomposition to binary conjunctions/disjunctions is done by some other preprocessing of the contracts, not sure.

It looks like `and` and `or` are not binary though, as @martin-cs mentioned:
https://github.com/diffblue/cbmc/blob/f929ec480f3ef5ef24544c6fc2532ef322bb4511/src/util/std_expr.h#L1833-L1834

https://github.com/diffblue/cbmc/blob/f929ec480f3ef5ef24544c6fc2532ef322bb4511/src/util/std_expr.h#L1941-L1942

I think we should handle the general case of two or more operands.

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
...
+    add_quantified_variable(first, replace, mode);
+    exprt second = if_expression.operands().at(2);
+    add_quantified_variable(second, replace, mode);
+  }
+  if(expression.id() == ID_and || expression.id() == ID_or)
+  {
+    // For multi-ary connectives, recursively check for
+    // nested quantified formulae in all terms
+    multi_ary_exprt multi_ary_expression = to_multi_ary_expr(expression);
+    for(auto operand : multi_ary_expression.operands())
```
SaswatPadhi: Would `const auto &` suffice? Do we have to create copies?

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
...
+    // nested quantified formulae in the term
+    unary_exprt unary_expression = to_unary_expr(expression);
+    exprt term = unary_expression.operands().at(0);
+    add_quantified_variable(term, replace, mode);
+  }
+  if(expression.id() == ID_notequal || expression.id() == ID_implies)
+  {
+    // For binary connectives, recursively check for
+    // nested quantified formulae in the left and right terms
+    binary_exprt binary_expression = to_binary_expr(expression);
```
SaswatPadhi: Many of these types, e.g. in place of `binary_exprt`, can be `const auto &`

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
   replace_symbolt &replace,
```
SaswatPadhi, line 169: Could we use `const exprt& expression`?

SaswatPadhi (approved):
Approving since it's not a blocker, but it would be nice to use `const` references (`const T &`) in place of copies (`T`) wherever possible.

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
...
-  // and (2) there are no inner quantifier formulae
-  // This TODO is handled in PR #5968
-
-  if(expression.id() == ID_exists || expression.id() == ID_forall)
+  if(expression.id() == ID_not || expression.id() == ID_typecast)
+  {
+    // For unary connectives, recursively check for
+    // nested quantified formulae in the term
+    unary_exprt unary_expression = to_unary_expr(expression);
+    exprt term = unary_expression.operands().at(0);
```
martin-cs: `unary_expression.op()`

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
...
+    add_quantified_variable(term, replace, mode);
+  }
+  if(expression.id() == ID_notequal || expression.id() == ID_implies)
+  {
+    // For binary connectives, recursively check for
+    // nested quantified formulae in the left and right terms
+    binary_exprt binary_expression = to_binary_expr(expression);
+    exprt left = binary_expression.operands().at(0);
```
martin-cs: `binary_expression.lhs()`

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
...
+    add_quantified_variable(left, replace, mode);
+    exprt right = binary_expression.operands().at(1);
```
martin-cs: `binary_expression.rhs()`

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
...
+    add_quantified_variable(right, replace, mode);
+  }
+  if(expression.id() == ID_if)
+  {
+    // For ternary connectives, recursively check for
+    // nested quantified formulae in all three terms
+    if_exprt if_expression = to_if_expr(expression);
+    exprt condition = if_expression.operands().at(0);
```
martin-cs: `if_expression.cond()`

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
...
+    add_quantified_variable(condition, replace, mode);
+    exprt first = if_expression.operands().at(1);
```
martin-cs: `if_expression.true_case()`

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
...
+    add_quantified_variable(first, replace, mode);
+    exprt second = if_expression.operands().at(2);
```
martin-cs: `if_expression.false_case()`

src/goto-instrument/code_contracts.cpp
martin-cs: It would be better, yes.  Please change.

src/goto-instrument/code_contracts.cpp
```diff
@@ -168,23 +169,56 @@ void code_contractst::add_quantified_variable(
...
+  else if(expression.id() == ID_exists || expression.id() == ID_forall)
   {
-    // get quantified symbol
-    exprt tuple = expression.operands().front();
-    exprt quantified_variable = tuple.operands().front();
+    // When a quantifier expression is found,
+    // 1. get quantified symbol
+    quantifier_exprt quantifier_expression = to_quantifier_expr(expression);
+    exprt tuple = quantifier_expression.operands().at(0);
+    exprt quantified_variable = tuple.operands().at(0);
```
martin-cs: The other ones I have flagged are style but here it really matters, you should use `variables()` and `where()` so that you can handle the case of multiple quantified variables.

src/goto-instrument/code_contracts.cpp
```diff
@@ -193,10 +227,14 @@ void code_contractst::add_quantified_variable(
...
       symbol_table);
 
-    // add created fresh symbol to expression map
+    // 3. add created fresh symbol to expression map
     symbol_exprt q(
       quantified_symbol.get_identifier(), quantified_symbol.type());
     replace.insert(q, new_symbol.symbol_expr());
+
+    // 4. recursively check for nested quantified formulae
+    exprt quantified_expression = quantifier_expression.operands().at(1);
```
martin-cs: `where()`

martin-cs (approved):
Approve subject to the requested corrections.

Where possible use the named / most specific accessor functions.  This is to avoid the need to rummage around in the `irept` formatting.  In most cases it is just stylistic but in the case of quantifiers it is covering a bug.

src/goto-instrument/code_contracts.cpp
ArenBabikian: Currently, I do not think that it is possible. But I the medium/long term, we should get rid of this parameter entirely.

src/goto-instrument/code_contracts.cpp
ArenBabikian: I have made the modification as requested. However, I would like to point out that currently, it seems like quantifier expressions with multiple variables are not fully handled (or used, at least). This can be fixed in a separate PR.

SaswatPadhi: Yes, in my experience, most call sites use `.symbol()` in place of `.variables()` and expect a quantifier over a single variable. But thanks for the fix here. Looks good to me and we could add a regression test later once we handle multi-variable quantification properly.

PR #5971, opened 2021-03-24
[CI] Remove fixed windows SDK version lookup.

NlightNFotis:
Before this, we were looking for the codesign tool under a specific
folder, signifying a specific version of the Windows SDK. But Github
Action Runner Software is being updated every now and then, causing
this dependency to be wobbly.

This should make the tool lookup more robust for future SDK version
updates as well.

### DIFF
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -153,7 +153,7 @@ jobs:
         uses: microsoft/setup-msbuild@v1.0.2
       - name: Setup code sign environment
         run: |
-          echo "$(Split-Path -Path $(Get-ChildItem -Path ${env:ProgramFiles(x86)} -Recurse -Filter 'signtool.exe' | Where-Object FullName -like '*10.0.19041.0\x64\signtool.exe').FullName)" >> $env:GITHUB_PATH
+          echo "$(Split-Path -Path $(Get-ChildItem -Path ${env:ProgramFiles(x86)} -Recurse -Filter 'signtool.exe' | Where-Object FullName -like '*10.0.*\x64\signtool.exe').FullName)" >> $env:GITHUB_PATH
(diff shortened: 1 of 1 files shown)


.github/workflows/release-packages.yaml
```diff
@@ -153,7 +153,7 @@ jobs:
         uses: microsoft/setup-msbuild@v1.0.2
       - name: Setup code sign environment
         run: |
-          echo "$(Split-Path -Path $(Get-ChildItem -Path ${env:ProgramFiles(x86)} -Recurse -Filter 'signtool.exe' | Where-Object FullName -like '*10.0.19041.0\x64\signtool.exe').FullName)" >> $env:GITHUB_PATH
+          echo "$(Split-Path -Path $(Get-ChildItem -Path ${env:ProgramFiles(x86)} -Recurse -Filter 'signtool.exe' | Where-Object FullName -like '*10.0.*\x64\signtool.exe').FullName)" >> $env:GITHUB_PATH
```
hannes-steffenhagen-diffblue, line 156: I swear there already was an action or something along the lines that did more or less this, but I suppose this is more robust than the previous version. Might cause slight problems if there are multiple SDK versions, but then again I don't think it really matters which version of sign tool you use.

hannes-steffenhagen-diffblue (approved):

.github/workflows/release-packages.yaml
NlightNFotis: There are multiple SDK versions pre-installed (https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md), but I also figured it shouldn't matter which SDK (minor) version the code sign tool we use is attached to.

thomasspriggs (approved):

TGWDB (approved):

PR #5973, opened 2021-03-25
Fix SMT2 encoding of `array_comprehension_exprt`

SaswatPadhi:
The SMT conversion routines were missing a rule for translating `array_comprehension_exprt`.

This PR creates a new translation rule leveraging the `lambda` construct in SMTLIB2.

NOTE: There may be several other `broken-smt-backend` regression tests that may be fixed by this change; I haven't checked (there are too many!).

### DIFF
--- a/regression/cbmc/byte_update15/test.desc
+++ b/regression/cbmc/byte_update15/test.desc
@@ -1,4 +1,4 @@
-CORE broken-smt-backend
+CORE broken-cprover-smt-backend
 main.c
 
 ^EXIT=0$
@@ -6,3 +6,6 @@ main.c
 ^VERIFICATION SUCCESSFUL$
 --
 ^warning: ignoring
+--
+This test currently does not pass with CPROVER SMT2 backend
+because the solver does not fully support quantified expressions.
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -57,6 +57,7 @@ smt2_convt::smt2_convt(
     use_as_const(false),
     use_datatypes(false),
     use_array_of_bool(false),
+    use_lambda_for_array(false),
     emit_set_logic(true),
     ns(_ns),
     out(_out),
@@ -101,6 +102,7 @@ smt2_convt::smt2_convt(
   case solvert::Z3:
     use_as_const = true;
     use_array_of_bool = true;
+    use_lambda_for_array = true;
     emit_set_logic = false;
     use_datatypes = true;
     break;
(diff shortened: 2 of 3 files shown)


kroening:
If you can use an SMT-LIB expression that maps directly onto what you want to translate, you no longer need the `defined_expressions` mechanism!

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -4390,6 +4403,27 @@ void smt2_convt::find_symbols(const exprt &expr)
...
   }
+  else if(expr.id() == ID_array_comprehension)
+  {
+    if(defined_expressions.find(expr) == defined_expressions.end())
+    {
+      const auto &array_comprehension = to_array_comprehension_expr(expr);
+
+      const irep_idt id =
+        "array_comprehension." + std::to_string(defined_expressions.size());
+      out << "(define-fun " << id << " () ";
```
martin-cs: Why are you creating a function with no arguments that returns a lambda?

martin-cs (changes requested):
1. Is you think it fixes other things, please can you take the time to find out what they are?  Someone will have to do this at some point and it would be better that it is connected with the change that fixes them.
2. I'm pretty sure `lambda` isn't in SMT-LIB 2.  It has been /proposed/ for SMT-LIB 3 but we shouldn't be outputting it in an SMT-LIB 2.0 back-end.

src/solvers/smt2/smt2_conv.cpp
martin-cs: Unless I have got the wrong end of the stick completely, I think you can implement this without the lambda.

SaswatPadhi: Arrays are infinite maps in SMT and are represented just like key -> value functions. The function with no arguments returns a lambda, because we want to return an array.

This has been tested with both Z3 -- you can run the attached regression test with `cbmc --smt2`.
I am not super sure about the official status of `lambda` within SMTLIB2 vs SMTLIB3, but Z3 supports it for a while now.

AFAIK, the only other way to implement array comprehension would be to use quantifiers, but that did not work even for the simpler case of `array_of` where the lambda expression was a constant (independent of the index).

SaswatPadhi: I will make this translation solver-specific then -- only emit SMT2 output using lambda if solver is Z3. Otherwise, fall back to quantifier-based initialization.

[Also, more info about `lambda` handling here: Z3Prover/z3/issues/2221]

regression/cbmc/byte_update15/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE broken-smt-backend
+CORE smt-backend
```
tautschnig: Discussed offline: this one should just be "CORE", not restricted to "smt-backend"

SaswatPadhi: Changed as suggested.

SaswatPadhi:
I have tested this (the attached regression test) locally with both `--z3` and `--cvc4` and it works correctly.

However, with CVC4 (and other solvers that don't allow `lambda` as an array), the SMT logic needs to be set to `AUFBV` instead of `QF_AUFBV` [see #5977] because we fallback to universally quantified initialization of the array.

martin-cs (approved):
I would rather not have the default logic change to be quantified as it will block some solvers.  Beyond that, looks OK.

SaswatPadhi:
Thanks for taking a look, @martin-cs.

I agree with not changing the default logic to allow quantifiers. I opened #5977 to discuss more about this issue.

peterschrammel (approved):

PR #5974, opened 2021-03-25
Fix SMT2 encoding of `array_of_exprt`

SaswatPadhi:
The SMT conversion routines were broken for `array_comprehension_exprt`. They used quantifiers to initialize constant arrays, but that encoding didn't seem to work with existing SMT solvers.

This PR fixes the translation rule by using the `as const` construct in SMTLIB2.

NOTE: I am not sure which regression test is directly affected by this, I just noticed this issue while fixing another bug [#5973]. I'd be happy to add a new regression test, but there must be some `broken-smt-backend` regression test blocked on this one.

### DIFF
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -54,6 +54,7 @@ smt2_convt::smt2_convt(
   solvert _solver,
   std::ostream &_out)
   : use_FPA_theory(false),
+    use_as_const(false),
     use_datatypes(false),
     use_array_of_bool(false),
     emit_set_logic(true),
@@ -79,6 +80,7 @@ smt2_convt::smt2_convt(
     break;
 
   case solvert::CPROVER_SMT2:
+    use_as_const = true;
     use_array_of_bool = true;
     emit_set_logic = false;
     break;
@@ -87,6 +89,7 @@ smt2_convt::smt2_convt(
     break;
 
   case solvert::CVC4:
+    use_as_const = true;
     break;
 
   case solvert::MATHSAT:
@@ -96,6 +99,7 @@ smt2_convt::smt2_convt(
     break;
 
   case solvert::Z3:
+    use_as_const = true;
     use_array_of_bool = true;
     emit_set_logic = false;
     use_datatypes = true;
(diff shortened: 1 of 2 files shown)


SaswatPadhi:
Hmm ... CI is failing because CPROVER SMT2 doesn't understand `as const` but this is supported by Z3 and CVC4.

kroening:
Likewise here: you no longer need defined_expressions -- simply emit the required SMT-LIB expression directly.

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -4367,27 +4367,23 @@ void smt2_convt::find_symbols(const exprt &expr)
...
-      #if 0 // not really working in any solver yet!
-      out << "(assert (forall ((i ";
-      convert_type(array_index_type());
-      out << ")) (= (select " << id << " i) ";
-      convert_expr(expr.op0());
-      out << ")))" << "\n";
-      #endif
+      out << "(define-fun " << id << " () ";
+      convert_type(array_of_expr.type());
+      out << " ((as const ";
```
martin-cs: I'm pretty sure `as const` isn't in SMT-LIB.

SaswatPadhi: I am not sure about the status of `as const` in SMTLIB2 standard, but both Z3 and CVC4 accept `as const`. We anyway didn't have a working solution for this with quantifiers, and I have tested array initialization with `as const` for both solvers.

SaswatPadhi: I will make this translation solver-specific then -- only emit SMT2 output if `solver` is either Z3 or CVC4.

SaswatPadhi:
Marking as draft until #5976  is merged.

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -53,7 +53,8 @@ smt2_convt::smt2_convt(
   const std::string &_logic,
   solvert _solver,
   std::ostream &_out)
-  : use_FPA_theory(false),
+  : use_as_const(false),
+    use_FPA_theory(false),
```
tautschnig: Nit picking: in the interest of reducing churn, could you swap those two lines (and then also the declarations in the header file)?

SaswatPadhi: Sure, I'll make this change.

Thanks for taking a look!

src/solvers/smt2/smt2_conv.cpp
martin-cs: Yes, it is fine to use extensions but they must not be in the default.  Z3 and CVC4 are not the only solvers, they are not even the only solvers that people use with CBMC.

SaswatPadhi: I agree. Please take another look and see if the changes look okay. I introduced the `use_as_const` field to track solvers that support `as const`.

kroening (approved):

kroening:
It would be awesome to identify the tests that work after this change, and remove the 'broken' flag from these!

tautschnig:
> It would be awesome to identify the tests that work after this change, and remove the 'broken' flag from these!

https://github.com/diffblue/cbmc/pull/5958 will help with this, but I need to do my homework first...

martin-cs (approved):

SaswatPadhi:
I will create a separate PR for the `broken-*` tests that are fixed, and I will link to this PR so we know the root cause.

PR #5979, opened 2021-03-26
Vsd - merge values

jezhiggins:
As a precursor to work on widening, I've been filling out the merge implementation for the various value representations. 

The changes are relatively minor, generally simplifying and generalising the merge. Once change of note is the addition of `internal_hash` and `internal_equality` members function to `interval_abstract_valuet`, which means intervals can be inserted into value-sets correctly, without duplication. 

The bulk of this PR is unit tests, covering the various combinations of constants, intervals, and value-sets.

### DIFF
--- a/src/analyses/variable-sensitivity/abstract_object_set.cpp
+++ b/src/analyses/variable-sensitivity/abstract_object_set.cpp
@@ -37,7 +37,7 @@ void abstract_object_sett::output(
   join_strings(out, output_values.begin(), output_values.end(), ", ");
 }
 
-abstract_object_pointert abstract_object_sett::to_interval()
+constant_interval_exprt abstract_object_sett::to_interval() const
 {
   PRECONDITION(!values.empty());
 
@@ -49,6 +49,5 @@ abstract_object_pointert abstract_object_sett::to_interval()
     lower_expr = constant_interval_exprt::get_min(lower_expr, value_expr);
     upper_expr = constant_interval_exprt::get_max(upper_expr, value_expr);
   }
-  return std::make_shared<interval_abstract_valuet>(
-    constant_interval_exprt(lower_expr, upper_expr));
+  return constant_interval_exprt(lower_expr, upper_expr);
 }
--- a/src/analyses/variable-sensitivity/abstract_object_set.cpp
+++ b/src/analyses/variable-sensitivity/abstract_object_set.cpp
@@ -41,13 +41,20 @@ constant_interval_exprt abstract_object_sett::to_interval() const
 {
   PRECONDITION(!values.empty());
 
-  exprt lower_expr = first()->to_constant();
-  exprt upper_expr = first()->to_constant();
+  const auto &initial =
+    std::dynamic_pointer_cast<const abstract_value_objectt>(first());
+
+  exprt lower_expr = initial->to_interval().get_lower();
+  exprt upper_expr = initial->to_interval().get_upper();
   for(const auto &value : values)
   {
-    const auto &value_expr = value->to_constant();
-    lower_expr = constant_interval_exprt::get_min(lower_expr, value_expr);
-    upper_expr = constant_interval_exprt::get_max(upper_expr, value_expr);
+    const auto &v =
+      std::dynamic_pointer_cast<const abstract_value_objectt>(value);
+    const auto &value_expr = v->to_interval();
+    lower_expr =
+      constant_interval_exprt::get_min(lower_expr, value_expr.get_lower());
+    upper_expr =
+      constant_interval_exprt::get_max(upper_expr, value_expr.get_upper());
   }
   return constant_interval_exprt(lower_expr, upper_expr);
 }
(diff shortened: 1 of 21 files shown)


jezhiggins:
The check-ubuntu-20_04-make-gcc build failure appear to be unrelated to the changes on this PR.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
```diff
@@ -162,6 +163,18 @@ value_set_abstract_objectt::value_range_implementation() const
   return make_value_set_value_range(values);
 }
 
+exprt value_set_abstract_objectt::to_constant() const
+{
+  verify();
+  return values.size() == 1 ? (*values.begin())->to_constant()
+                            : abstract_objectt::to_constant();
```
martin-cs: In the case of multiple entries, merge all of them then `to_constant`?

src/analyses/variable-sensitivity/abstract_object_set.cpp
```diff
@@ -37,7 +37,7 @@ void abstract_object_sett::output(
   join_strings(out, output_values.begin(), output_values.end(), ", ");
 }
 
-abstract_object_pointert abstract_object_sett::to_interval()
+constant_interval_exprt abstract_object_sett::to_interval() const
 {
```
martin-cs, line 41: Rather than using `to_constant` this should probably call `to_interval` so that if it is a value set of intervals it winds up taking the join of all of the intervals.

unit/analyses/variable-sensitivity/value_expression_evaluation/expression_evaluation.cpp
```diff
@@ -595,9 +595,9 @@ SCENARIO(
       auto op2 = make_value_set({interval12, val1}, environment, ns);
       auto result = add_as_value_set(op1, op2, environment, ns);
 
-      THEN("= { [2,4], [2,3], [2,3], 2 }")
-      { // duplicate interval ok as first pass. Will be eliminated in due course.
```
martin-cs, line 599: Nice!

src/analyses/variable-sensitivity/abstract_object_set.cpp
martin-cs: Ah; ignore this, I have found the later commit that does this!

unit/analyses/variable-sensitivity/constant_abstract_value/#merge.cpp#
```diff
@@ -0,0 +1,468 @@
+/*******************************************************************\
+
+ Module: Unit tests for constant_abstract_valuet::merge
```
martin-cs: File name? `#merge.cpp#`?  Maybe this shouldn't be in the PR?

unit/analyses/variable-sensitivity/constant_abstract_value/#merge.cpp#
```diff
@@ -0,0 +1,468 @@
...
+    WHEN("merging BOTTOM constant with BOTTOM")
+    {
+      auto bottom1 = make_bottom_constant();
+      auto top2 = make_bottom_object();
+
+      auto merged = merge(bottom1, top2);
+
+      THEN("result is unmodified BOTTOM")
+      {
+<<<<<<< HEAD
```
martin-cs: Yeah... don't think this is supposed to be here.

unit/analyses/variable-sensitivity/interval_abstract_value/merge.cpp
```diff
@@ -453,6 +453,32 @@ SCENARIO(
...
+        make_value_set({constant_interval_exprt(val4, val6)}, environment, ns);
+
+      auto merged = merge(op1, op2);
+
+      THEN("result is modified [3, 6]")
+      {
+        EXPECT_MODIFIED(merged, val3, val6);
+      }
+    }
+    WHEN("merging [3, 4] with { 1, [3, 3] }")
```
martin-cs, line 469: Nice!

martin-cs (changes requested):
One critical fix, the rest should be good to go.  If the unit test linking issue persists we should investigate.

unit/analyses/variable-sensitivity/constant_abstract_value/#merge.cpp#
jezhiggins: Oops, sorry.

unit/analyses/variable-sensitivity/interval_abstract_value/merge.cpp
jezhiggins: We aim to please.

martin-cs (approved):
Thanks for the fixes.

src/analyses/variable-sensitivity/value_set_abstract_object.cpp
jezhiggins: Add new test case merging 1 with { 1, [1, 1] } returns 1.

PR #5980, opened 2021-03-26
Minor cleanup of some SMT2 files

SaswatPadhi:
This is a minor cleanup PR.

There were 3 SMT2 files that were:
1. being rejected by CVC4 without the `--incremental` cmdline flag
2. leading to errors in Z3 because `ALL_SUPPORTED` logic was not understood

In this PR, I:
1. added `incremental` option in the SMT2 files, which is required for using `push` and `pop` commands in CVC4.
2. replaced `ALL_SUPPORTED` logic (which is CVC4-specific) with `ALL` (which is part of the SMTLIB 2.5+ standard).

The updated SMT2 files have been tested with both solvers and they emit no warnings or errors.

- NA ~~Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).~~

### DIFF
--- a/regression/strings/test3/test-bv-to-int-onebyone.smt2
+++ b/regression/strings/test3/test-bv-to-int-onebyone.smt2
@@ -1,5 +1,6 @@
 (set-option :produce-models true)
-(set-logic ALL_SUPPORTED)
+(set-option :incremental true)
+(set-logic ALL)
 
 (declare-fun s () String)
 (declare-fun s2 () String)
--- a/regression/strings/test3/test-bv-to-int.smt2
+++ b/regression/strings/test3/test-bv-to-int.smt2
@@ -1,5 +1,5 @@
 (set-option :produce-models true)
-(set-logic ALL_SUPPORTED)
+(set-logic ALL)
 
 (declare-fun s () String)
 (declare-fun s2 () String)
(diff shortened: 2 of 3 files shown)


martin-cs (commented):
The `incremental` part; definitely.  The `set-logic`, well ... neither are actually standard nor the minimal logic in question so... unsure?

SaswatPadhi:
But this doc [<http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.5-r2015-06-28.pdf>] says (on page 11) that `ALL` is now  a standard logic since SMTLIB 2.5:

> There is now a predefined argument, `ALL`, for the `set-logic` command which refers to the most general logic supported by the solver executing the command.

martin-cs (approved):
Thanks @SaswatPadhi I had missed it's addition to SMT-LIB 2.5.  I'd prefer that the logics were minimal but `ALL` is definitely better than `ALL_SUPPORTED`.

SaswatPadhi:
Thanks for approving the changes @martin-cs.

I tried restricting the logic -- these files work with the `BVS` logic in CVC4, but Z3 doesn't recognize `BVS` and rejects these files.

PR #5981, opened 2021-03-27
Change domain merge(...) to take trace_ptrts rather than locationt.

jezhiggins:
Passing in trace_ptrts provides more context that locationt, which we
will need in the variable-sensitivity domain to determine if we need to
widen.

While this change touches code across the repository, at the moment
 however, none of the merge functions use this information at all,
so this change is a benign.


martin-cs (approved):
I love it when someone fixes up a mess I made!  Thanks for this.  It should have been changed a long time ago but at the time there were various out-of-tree PRs it would have messed with.

PR #5983, opened 2021-03-29
Fix z3 installation on PR CI jobs

thomasspriggs:
Includes updating z3 to 4.8.10 for Windows CI jobs. So that the version, which is currently most recently released is available for use in Windows regression tests. Z3 installation was previously broken, due to the chocolatey package being out of date / broken and due to the installation being missing from various jobs.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -287,7 +287,7 @@ jobs:
           submodules: recursive
       - name: Fetch dependencies
         run: |
-          choco install winflexbison3 strawberryperl
+          choco install winflexbison3 strawberryperl z3
           nuget install clcache -OutputDirectory "c:\tools" -ExcludeVersion -Version 4.1.0
           echo "c:\tools\clcache\clcache-4.1.0" >> $env:GITHUB_PATH
           echo "c:\ProgramData\chocolatey\bin" >> $env:GITHUB_PATH
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -248,9 +248,12 @@ jobs:
           submodules: recursive
       - name: Fetch dependencies
         run: |
-          choco install winflexbison3 z3
+          choco install winflexbison3
           nuget install clcache -OutputDirectory "c:\tools" -ExcludeVersion -Version 4.1.0
           echo "c:\tools\clcache\clcache-4.1.0" >> $env:GITHUB_PATH
+          Invoke-WebRequest -Uri https://github.com/Z3Prover/z3/releases/download/z3-4.8.10/z3-4.8.10-x64-win.zip -OutFile .\z3.zip
+          Expand-Archive -LiteralPath '.\z3.Zip' -DestinationPath C:\tools
+          echo "c:\tools\z3-4.8.10-x64-win\bin;" >> $env:GITHUB_PATH
       - name: Setup Visual Studio environment
         uses: microsoft/setup-msbuild@v1.0.2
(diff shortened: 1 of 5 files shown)


.github/workflows/pull-request-checks.yaml
```diff
@@ -287,7 +287,7 @@ jobs:
           submodules: recursive
       - name: Fetch dependencies
         run: |
-          choco install winflexbison3 strawberryperl
+          choco install winflexbison3 strawberryperl z3
```
NlightNFotis: Should this commit be dropped?

NlightNFotis (commented):
I have a couple of remarks:

1) This PR claims to be upgrading z3 on windows, but it also makes a number of other changes, such as probing for z3 version on mac and ubuntu jobs? Is this on purpose? If that's the case, can the PR be renamed to accurately reflect scope of changes, or submit the changes as another PR?
2) One of the CI jobs is failing because it's probing for `z3` when it's not installed.

TGWDB (changes requested):
A couple of bookkeeping changes:

1. The PR comment claims "latest version", which may be true today but will not be in future. Since the PR explicitly sets the version to 4.8.10 can we make that clear? (Specially since we could probably also use something like https://github.com/Z3Prover/z3/releases/latest for the true latest.)
2. Several places claim to "Check z3 version", but I think they only display the version, there does not appear to be a "check" that this is correct?

thomasspriggs:
After discussion with Fotis, I am going to change this PR to fix Z3 installation across all jobs and add a test. Please wait for me to push the extra changes, before reviewing.

.github/workflows/pull-request-checks.yaml
thomasspriggs: This commit exists, so that I am not doing both adding and updating in subsequent commits.

thomasspriggs:
> A couple of bookkeeping changes:
> 
> 1. The PR comment claims "latest version", which may be true today but will not be in future. Since the PR explicitly sets the version to 4.8.10 can we make that clear? (Specially since we could probably also use something like https://github.com/Z3Prover/z3/releases/latest for the true latest.)

I think this should now be fixed.

> 2. Several places claim to "Check z3 version", but I think they only display the version, there does not appear to be a "check" that this is correct?

It is useful as is for confirming that Z3 is installed, in the path and logging the version. We could in theory check that the information printed is what we expect. But I think the changes are useful as is, without adding more functionality. I'll re-word the message a little, in order to try and make this clear.

thomasspriggs:
@TGWDB Your comments should now be addressed. Please re-review when you have a minute.

SaswatPadhi:
Not a blocker if we don't have broken tests right now, but we might soon run into cases where a regression test might pass on Windows & Mac but failing on Ubuntu 18.04 or 20.04 because they're running older versions of Z3 (4.4.1 and 4.8.7 respectively).

Would it be possible to `wget` the https://github.com/Z3Prover/z3/releases/download/z3-4.8.10/z3-4.8.10-x64-ubuntu-18.04.zip binary from https://github.com/Z3Prover/z3/releases/tag/z3-4.8.10 and just move it to `/usr/local/bin` or some directory in `$PATH`?

@tautschnig is listed as the maintainer of the `z3` Ubuntu package (https://packages.ubuntu.com/bionic/z3), may be he could help too.

TGWDB (approved):

NlightNFotis (approved):
LGTM

thomasspriggs:
@SaswatPadhi Now that this is merged, you should be able to re-base and merge https://github.com/diffblue/cbmc/pull/5904

Issue #5982, opened 2021-03-28
Regression testing infrastructure for the SMT backend


SaswatPadhi:
This is based on a recent discussion I had with @tautschnig. I wanted to open this RFC and get some feedback before introducing a new tag in #5973.

Currently we seem to have 3 tags for our regression tests that relevant to the SMT backend:
- `smt-backend`: Tests that _require_ the SMT backend, and don't work with the SAT backend
- `broken-smt-backend`: Tests that pass with the SAT backend, but don't work with the SMT backend
- `thorough-smt-backend`: Tests that pass with the SAT backend, but take a long time with the SMT backend

While working on #5973, I came across a regression test that works with both Z3 and CVC4, but doesn't work with CPROVER SMT2 because it lacks support for quantifiers. So I think it would make sense to add a few new tags that are solver-specific:
- `broken-cprover-smt2-smt-backend` and `thorough-cprover-smt2-smt-backend`
- `broken-z3-smt-backend` and `thorough-z3-smt-backend`
- `broken-cvc4-smt-backend` and `thorough-cvc4-smt-backend`

The `-smt-backend` suffix for these tags would be useful for `grep`-ing all regression tests related to the SMT-backend in any way. The tags can also be combined, e.g.
- `smt-backend broken-cprover-smt2-smt-backend` would indicate that the test requires SMT backend, but doesn't work CPROVER SMT2 solver
- `thorough-smt-backend broken-cvc4-smt-backend` would indicate that the test runs slower with the SMT backend, and doesn't work at all with CVC4 backend

I also suggest running the same version of these solvers across all platforms on our CI. That way, each time we bump up the version of solver `X` on our CI, we can revisit all of the `broken-X-smt-backend` and `thorough-X-smt-backend` tests and check if any of them are now fixed.

How many solvers we want our CI to be running is a separate discussion, but I think we should definitely at least be running Z3, and preferably both Z3 and CVC4 (in future).

peterschrammel:
Sounds good to me. 

Just too comments:
- Isn't `broken-smt-backend` currenty used for "doesn't work with CPROVER SMT2" (i.e. has nothing to do with whether it works with other SMT solvers such as Z3 or CVC4)?
- I'd use `broken-cprover-smt-backend` instead of `broken-cprover-smt2-smt-backend`.

SaswatPadhi:
Thanks Peter.

[...]

PR #5985, opened 2021-03-30
Added new solver-specific tags for SMT regression tests

SaswatPadhi:
As per the proposal described in #5982 and offline discussions with @tautschnig, this PR introduces new tags that can be used to label tests that are known to be broken with specific solvers. It also updates the `Makefile` targets to use these solver-specific tags.

This PR would unblock #5973. The regression test there was tagged `broken-smt-backend` and has been fixed to work with Z3 and CVC4 (and other solvers that support quantifiers), but currently there is no viable fix for CPROVER SMT2. Once this PR is merged, I would label that regression test as `broken-cprover-smt-backend` instead.

- NA ~~Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).~~

### DIFF
--- a/regression/cbmc/CMakeLists.txt
+++ b/regression/cbmc/CMakeLists.txt
@@ -20,7 +20,7 @@ add_test_pl_profile(
 add_test_pl_profile(
     "cbmc-cprover-smt2"
     "$<TARGET_FILE:cbmc> --cprover-smt2"
-    "-C;-X;broken-smt-backend;-X;thorough-smt-backend;${gcc_only_string}-s;cprover-smt2"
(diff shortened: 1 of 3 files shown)


SaswatPadhi:
Note that I have added additional targets to the `Makefile` but those are not built as part of the `default` target (that's run on the CI). I didn't know of a way to do that with cmake -- using `add_test_pl_profile` directly adds them to the build queue.

martin-cs (approved):
This is more distinctions than I would make but ... if you are happy to work with these then, cool.

SaswatPadhi:
Thanks for taking a look @martin-cs.

I removed the cvc4 tags for now since we don't run cvc4 on CI. I will keep #5982 open for further discussion and we can add more tags later if we need to.

PR #5964, opened 2021-03-22
Cache dereferences during symex to avoid repeatedly inserting large dereferences into SSA

hannes-steffenhagen-diffblue:
Repeated large expressions within the SSA (and the resulting final formula) can be problem for both symex (because we need to e.g. simplify the same/similar large expressions over and over again) as well as for sat solvers. Especially for the latter case the performance impact of making changes to the formula are somewhat unpredictable, but in our testing seemed to average out to an improvement. However, since it can lead to performance regressions in some cases and does change what traces look like a bit (due to the extra assignments) this is opt-in behaviour.

The idea here comes from an interesting example originally observe by Natasha Jeppu and raised as an issue [here](https://github.com/diffblue/cbmc/issues/5505). The approach doesn't replicate exactly what she did in the issue because we're changing things at the symex stage, but we still produce a noticeable (if not quite as a dramatic) speedup on her example in our testing.

Here's a table with performance measured on one test machine with the default miniSAT (times in seconds). Exact numbers here are going to be hard to reproduce, but manual checking on a variety of machines showed that these are _roughly_ consistent between different machines. Since a lot of the runtime difference for these is in the sat solver swapping that out is expected to produce different results, but preliminary testing suggests that the caching is still often beneficial there. The examples are the proofs from [S2N](https://github.com/awslabs/s2n), specifically on commit bb0f4c9b3f7e7ec0192dca9521a37e72190d0d30.

The runtimes are one-shots, but random manual testing doesn't show a lot of run-to-run variation so that should be enough to evaluate a trend; However this does mean that small changes (<5 %) are mostly attributable to noise. As you can see for most examples there isn't a particularly large difference in runtime with most examples, however a couple of examples have dramatic improvements (e.g. `s2n_hash_copy` or `s2n_hmac_free`), with a couple of small but also a handful of very notable regressions (e.g. `s2n_stuffer_write_base64` or `s2n_stuffer_read_expected_str`).

Unfortunately I still don't have a great working theory to explain or predict the large performance differences observed. One thing we've tried is limiting the caching to only larger dereference sets (based on expression size), but the results from that were somewhat inconclusive (it does seem to help in some cases but make things worse in others). IMHO this is still interesting enough to ship out behind a flag for people to play around with. I imagine this could be useful for use cases where the same tests are run over and over again in CI (such as is the case in a lot of projects in awslabs), where it ought to be possible to run once with and without the flag and just set it for subsequent runs if it turns out to help.

| Proof Name | No Cache | With Cache |
[...]

PR #5988, opened 2021-03-31
Move exprt::bounded_size to complexity_limitert

tautschnig:
complexity_limitert is really the only user of this method, and may be
fine with its implementation details (cf. discussion in #5914). For
other potential users such assumptions might not be safe.

### DIFF
--- a/src/goto-symex/complexity_limiter.cpp
+++ b/src/goto-symex/complexity_limiter.cpp
@@ -136,7 +136,8 @@ complexity_limitert::check_complexity(goto_symex_statet &state)
   if(!complexity_limits_active() || !state.reachable)
     return complexity_violationt::NONE;
 
-  std::size_t complexity = state.guard.as_expr().bounded_size(max_complexity);
+  std::size_t complexity =
+    bounded_expr_size(state.guard.as_expr(), max_complexity);
   if(complexity == 1)
     return complexity_violationt::NONE;
 
@@ -210,4 +211,30 @@ void complexity_limitert::run_transformations(
   else
     for(auto transform_lambda : violation_transformations)
       transform_lambda.transform(complexity_violation, current_state);
+}
+
+/// Amount of nodes \p expr contains, with a bound on how far to search.
+/// Starts with an existing count.
+/// \return Size of \p expr added to count without searching significantly
+/// beyond the supplied limit.
+static std::size_t
+bounded_expr_size(const exprt &expr, std::size_t count, std::size_t limit)
+{
+  const auto &ops = expr.operands();
+  count += ops.size();
+  for(const auto &op : ops)
+  {
+    if(count >= limit)
+    {
+      return count;
+    }
+    count = bounded_expr_size(op, count, limit);
+  }
+  return count;
+}
+
+std::size_t
+complexity_limitert::bounded_expr_size(const exprt &expr, std::size_t limit)
+{
+  return ::bounded_expr_size(expr, 1, limit);
 }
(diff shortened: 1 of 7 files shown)


unit/goto-symex/complexity_limiter.cpp
```diff
@@ -0,0 +1,49 @@
...
+Author: Diffblue Ltd
+
+\*******************************************************************/
+
+#include <testing-utils/use_catch.h>
+
+#include <goto-symex/complexity_limiter.h>
+
+TEST_CASE(
+  "Bounded size is computed correctly",
```
TGWDB: "Bounded size" -> "Bounded expr[ession] size"

unit/goto-symex/complexity_limiter.cpp
```diff
@@ -0,0 +1,49 @@
...
+  REQUIRE(complexity_limitert::bounded_expr_size(make_expression(2), 10) == 2);
+  REQUIRE(complexity_limitert::bounded_expr_size(make_expression(3), 10) == 3);
+
+  REQUIRE(complexity_limitert::bounded_expr_size(make_expression(30), 10) < 13);
+  REQUIRE(
+    complexity_limitert::bounded_expr_size(make_expression(30), 10) >= 10);
+}
+
+TEST_CASE(
+  "Bounded size versus pathological example",
```
TGWDB: "Bounded size" -> "Bounded expr[ession] size"

unit/goto-symex/complexity_limiter.cpp
```diff
@@ -0,0 +1,49 @@
...
+  "[core][goto-symex][complexity_limiter]")
+{
+  const typet type = signedbv_typet(32);
+  exprt pathology = plus_exprt(from_integer(1, type), from_integer(1, type));
+  // Create an infinite exprt by self reference!
+  pathology.add_to_operands(pathology);
+  // If bounded size is not checking the bound this will never terminate.
```
TGWDB: "Bounded size" -> "Bounded expr[ession] size"

TGWDB (commented):
Minor tweaks on text matching new names noted.

Outside of this, there were many experiments with CSE (some info #5964) that used the expression size and these showed that this could be a useful parameter to manipulate. I note this because if we want to have a bounded size for expressions later this would involve a refactor to return this to expr and not be in complexity_limiter.

src/goto-symex/complexity_limiter.cpp
```diff
@@ -211,3 +212,29 @@ void complexity_limitert::run_transformations(
...
+{
+  const auto &ops = expr.operands();
+  count += ops.size();
+  for(const auto &op : ops)
+  {
+    if(count >= limit)
+    {
+      return count;
+    }
+    count = bounded_expr_size(op, count, limit);
```
martin-cs, line 231: If we aren't going to fix this, can we at least have a comment saying this is potentially risky?

martin-cs (approved):
This is an improvement.  I'd rather we didn't have the for/recurse without caching but this is an improvement.

src/goto-symex/complexity_limiter.cpp
tautschnig: Comment now added to the public member function.

tautschnig:
> Minor tweaks on text matching new names noted.
> 
> Outside of this, there were many experiments with CSE (some info #5964) that used the expression size and these showed that this could be a useful parameter to manipulate. I note this because if we want to have a bounded size for expressions later this would involve a refactor to return this to expr and not be in complexity_limiter.

Thanks a lot for the comments, text changes implemented.

As far as re-use is concerned: if it turns out to be useful, then we could certainly consider moving this elsewhere. If we do, however, then IMHO it should be somewhere around `irept` as there isn't anything `exprt` specific happening here: `operands()` is the same as `sub()`, and this code really counts the number of unnamed subs.

TGWDB (approved):

kroening (approved):

PR #5990, opened 2021-04-01
Support Unicode 13.0 in CBMC

karkhaz:
Prior to this PR, users have only been able to pass numerical verbosity levels to CBMC:
```sh
> cbmc --verbosity 4 hello.c
# ...
VERIFICATION SUCCESSFUL
```

This pull request adds an additional verbosity level to cater for the growing number of CBMC users on iOS and Android:
```sh
> cbmc --verbosity 💯 hello.c
😆👍🎉
> cbmc --verbosity 💯 segfault.c
😩🐜💥
```

The patchset also adds a new error stream, `messaget::clout`. This is similar to `std::cout` except that it connects to the user's social media stream, allowing one's verification results to go viral.

clout
[[xkcd.com](https://xkcd.com/1482/)]

Finally, due to the new expressivity afforded by Unicode support, I've also implemented a code-style linter that rates your code by printing an emoji–––the more delighted the emoji, the more readable your code is.
```sh
> cbmc --condescending-linter my_first_intern_project.c
😖

# Below file is included as a test case
> cbmc --condescending-linter every_CVE_since_2014.c
💩
```

### DIFF
--- a/regression/goto-instrument/every_cve/every_cve_since_2014.c
+++ b/regression/goto-instrument/every_cve/every_cve_since_2014.c
@@ -1,0 +1,82 @@
+int SSLHASHSHA1_update();
+int tls1_write_bytes();
+int speculate();
+const int TLS1_RT_HEARTBEAT;
+
+#define EFAULT 1
+#define EBUSY 3
+#define ENOENT 4
+
+int main()
+{
+  int err;
+  const int payload_len = 16;
+  const int msg_len;
+  int payload[payload_len];
+  int message[msg_len];
+  int *tsk;
+  int *mm;
+  int *page_mask;
+  unsigned long start;
+  unsigned long nr_pages;
+  unsigned int gup_flags;
+  int *nonblocking;
+
+  // 2014-0160
+  int i = 0;
+  for(; i < err; ++i)
+    tls1_write_bytes(message[i], TLS1_RT_HEARTBEAT, payload[i]);
+
+  // 2014-6271
+  system("env x='() { :;}; echo shellshock'");
+
+  // 2014-1266
+  // not a great year for the industry
+  if((err = SSLHashSHA1_update()) != 0)
+    goto fail;
+    goto fail;
+
+
+  // 2017-5753
+  if(payload_len < msg_len)
+  {
+    int val = message[payload_len];
+    if(speculate(val))
+      sleep(2);
+  }
+
+  // 2016-5195
+  do {
+    int *page;
+    unsigned int foll_flags = gup_flags;
+    int vma = find_extend_vma(mm, start);
+
+retry:
+    cond_resched();
+    page = follow_page_mask(vma, start, foll_flags, &page_mask);
+    if (!page) {
+        int ret;
+        ret = faultin_page(tsk, vma, start, &foll_flags,
+                nonblocking);
+        switch (ret) {
+        case 0:
+            goto retry;
+        case -EFAULT:
+            return i ? i : ret;
+        case -EBUSY:
+            return i;
+        case -ENOENT:
+            goto next_page;
+        }
+        BUG();
+    }
+
+next_page:
+    nr_pages -= start;
+  } while (nr_pages);
+
+fail:
+  return 1;
+
+  return 0;
+}
--- a/src/goto-instrument/opinionated_linter.cpp
+++ b/src/goto-instrument/opinionated_linter.cpp
@@ -1,0 +1,94 @@
+#include "opinionated_linter.h"
+#include "goto_program2code.h"
+
+#include <util/prefix.h>
+
+#include <iostream>
+
+int goto_hatert::score()
+{
+  int score = this->max_score();
+
+  // You only get two chances to ruin your program
+  const int decrement = this->max_score() / 2;
+
+  std::list<irep_idt> emp_l;
+  std::unordered_set<irep_idt> emp_s;
+  std::set<std::string> emp_ss;
+  for(auto &gf : this->gm.goto_functions.function_map)
+  {
+    if(has_prefix(id2string(gf.first), CPROVER_PREFIX))
+      continue;
+    if(!gf.second.body.instructions.size())
+      continue;
+    code_blockt dst;
+    goto_program2codet g2c(
+      gf.first, gf.second.body, gm.symbol_table, dst, emp_l, emp_l, emp_s,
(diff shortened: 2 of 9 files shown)


karkhaz:
@nrdg42 @polgreen @mirelap-amazon @fbbotero-aws @mww-aws @nchong-at-aws @eidelmanjonathan

src/goto-instrument/opinionated_linter.h
```diff
@@ -0,0 +1,90 @@
...
+    analyses.push_back(new goto_hatert(this->gm));
+  }
+
+  void operator()();
+};
+
+// clang-format off
+#define OPT_OPINIONATED_LINTER "(condescending-linter)"
+#define HELP_OPINIONATED_LINTER \
+  " --condecending-linter          Complain about code formatting\n"
```
SaswatPadhi, line 87: Typo: conde[s]cending

src/goto-instrument/opinionated_linter.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+  int score = this->max_score();
+
+  // You only get two chances to ruin your program
+  const int decrement = this->max_score() / 2;
+
+  std::list<irep_idt> emp_l;
+  std::unordered_set<irep_idt> emp_s;
+  std::set<std::string> emp_ss;
+  for(auto &gf : this->gm.goto_functions.function_map)
```
SaswatPadhi, line 18: May be a `const` reference would also work?

`goto_program2code` shouldn't modify the goto function

src/goto-instrument/opinionated_linter.cpp
```diff
@@ -0,0 +1,94 @@
...
+int average_fun_lengtht::max_score()
+{
+  return this->boundaries.size();
+}
+
+void considered_harmfult::operator()()
+{
+  int possible_score = 0;
+  int actual_score = 0;
+  for(const auto analysis : this->analyses)
```
SaswatPadhi, line 78: May be we don't need to copy here? A (const) reference should work as well: `const auto &analysis`.

src/goto-instrument/opinionated_linter.cpp
```diff
@@ -0,0 +1,94 @@
...
+  {
+    possible_score += analysis->max_score();
+    actual_score += analysis->score();
+  }
+
+  int normalized = actual_score * 100 / possible_score;
```
SaswatPadhi, line 84: I think we should parenthesize the multiplication, so that we perform the division after the multiplication.

Otherwise, because of integer operations, the errors might be large. `5 * 100 / 7`  = `70` but `(5 * 100) / 7` = `71`.

regression/goto-instrument/every_cve/every_cve_since_2014.c
```diff
@@ -0,0 +1,82 @@
+int SSLHASHSHA1_update();
```
SaswatPadhi, line 1: May be we should call this file _named_ CVEs or _major_ CVEs since 2014? Those are the one I see here.

The total number of CVEs reported is roughly 20K per year! Many of them are low-severity ones.

src/util/message.h
```diff
@@ -168,7 +168,7 @@ class messaget
   enum message_levelt
   {
     M_ERROR=1, M_WARNING=2, M_RESULT=4, M_STATUS=6,
-    M_STATISTICS=8, M_PROGRESS=9, M_DEBUG=10
+    M_STATISTICS=8, M_PROGRESS=9, M_DEBUG=10, M_GEN_Z=1337
```
SaswatPadhi, line 171: Haha, Gen Z is leet! 😜

SaswatPadhi (commented):
Very cool! 😃
I took a quick look at the changes and left some minor comments.

My main question is regarding the verbosity setting. From what I understand, we have to set the verbosity to 1337 (even higher than DEBUG) to see the emoji output. Is that the right way to expose this feature? What if we don't want to flood the stderr / stdout with debug level messages but still want fancy emojis on social media?

Would a separate flag(`--gen-z-mode` lol) be better?

chrisr-diffblue:
Excellent work, as ever! I might request two further feature extensions (though I'm happy for these to be followup PRs)...

* Building on the new Unicode support, I'd like to see a new option to configure CBMC in 'SV-COMP mode' - I'd suggest either --🏎 or --🏆

* Building on the social media hookup via the new messaget::clout mechanism, I'd like to see a --zero-day option, in which for each vulnerability found, CBMC will generate a catchy name and post to social media.

martin-cs (approved):
👀🤔🧐🤩🤯💥👍👍👍💯❤️❤️❤️
http://gph.is/2b6YjxW

peterschrammel (commented):
Could we have the same for the verification verdicts too? 
E.g. 
```
VERIFICATION 😛
```

polgreen:
This looks great. I'm concerned about the range of expression of the condescending linter though. If my code is 🔥, I want it to say my code is 🔥.

Building on Peter's suggestion, I'd also support integrating unicode into status messages. I think this would make it easier for the user to empathise with CBMC when they give it more challenging verification tasks. For example:
- Passing problem to propositional reduction 🤓 
- Simple slicing 🔪 🔪  
- Solving with Minisat 🤔 💪 
- XX variables, XX clauses: 😩 (if numbers are large), 😥 (if numbers are small)
- VERIFICATION SUCCESSFUL 🥳 🎉 
- VERIFICATION FAILED 😭 💀

src/goto-instrument/opinionated_linter.cpp
tautschnig: But then we should perhaps use `mp_integer` to avoid overflow?

martin-cs:
https://twitter.com/ciphernyx/status/1078422385555054593 ?

SaswatPadhi:
LOL. Okay makes a lot more sense now. April 1 😅

karkhaz:
@polgreen, your code is always 🔥. Thank you all for your excellent suggestions. I'm going to close this while I implement a CVE logo generator. I've also discovered that [U+1FAD3 FLATBREAD](https://emojipedia.org/flatbread/) isn't widely supported yet, so let's wait for that so that we can turn `--naan-check` into `--🫓-🏁`.

danielsn:
I’m a little concerned about accessibility. How will this work with screen reader apps for blind users?

PR #5993, opened 2021-04-01
Unify make and cmake behaviours

### LINKED ISSUE #5986: Several inconsistenties in regression test build scripts
**CBMC version:** abba3d7
**Operating system:** Mac OS 10.15.7

PR #5985 exposed some issues with the existing `Makefile`s and `CMakeLists.txt` in `regression/` subdirectories:

#### 1. `make` and `cmake` sometimes run different regression tests

One example is `regression/cbmc`:
- `make` only builds the `test` target. Although `test-cprover-smt2` and `test-paths-lifo` are _defined_, they are not _built_ as part of the `default` target.
- `cmake` builds (the equivalents of) all three targets: `test` and `test-cprover-smt2` and `test-paths-lifo`.

There might be similar instances in other subdirectories under `regression/`. You can notice this difference in the `*cmake*` (and Ubuntu 18.04) CI jobs vs `*make*` CI jobs. They don't run the same set of regression tests!

I am not sure if there is a good / automated solution to ensuring that `make` and `cmake` always build the same targets. An immediate step for now is to update all the `default` targets and make sure they run the same tests as added in the `CMakeLists.txt` file.

#### 2. Inconsistent behavior of `make` command within `regression/`

The `make` command under `regression/` runs the `default` target in each subdirectory:
https://github.com/diffblue/cbmc/blob/abba3d78d7a364714d109a1f9340c784b0431e6a/regression/Makefile#L72-L75

However, the `make X` command under `regression/` runs the `test` target under `X` subdirectory:
https://github.com/diffblue/cbmc/blob/abba3d78d7a364714d109a1f9340c784b0431e6a/regression/Makefile#L79-L81

Line 81 should be fixed to also build the `default` target instead.

TGWDB:
This fixes the inconsistencies between cmake and make
behaviours.

Fixes #5986 

### DIFF
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -85,7 +85,7 @@ test:
 .PHONY: $(DIRS)
 $(DIRS):
 	@echo "Running $@..." ;
-	$(MAKE) -C "$@" test || exit 1;
+	$(MAKE) -C "$@" || exit 1;
 
 # Run all test directories using GNU Parallel
 .PHONY: test-parallel
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -78,14 +78,14 @@ endif
 .PHONY: test
 test:
 	@for dir in $(DIRS); do \
-		$(MAKE) "$$dir" || exit 1; \
+		$(MAKE) "$$dir" test || exit 1; \
 	done;
 
 # Pattern to execute a single test suite directory
 .PHONY: $(DIRS)
 $(DIRS):
 	@echo "Running $@..." ;
-	$(MAKE) -C "$@" || exit 1;
+	$(MAKE) -C "$@" test || exit 1;
 
 # Run all test directories using GNU Parallel
 .PHONY: test-parallel
--- a/regression/cbmc/Makefile
+++ b/regression/cbmc/Makefile
@@ -1,5 +1,3 @@
-THIS_FILE := $(lastword $(MAKEFILE_LIST))
-
 default: test
 
 include ../../src/config.inc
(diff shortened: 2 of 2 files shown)


regression/cbmc/Makefile
```diff
@@ -11,6 +13,8 @@ endif
 
 test:
 	@../test.pl -e -p -c "../../../src/cbmc/cbmc --validate-goto-model --validate-ssa-equation" -X smt-backend $(GCC_ONLY)
+	@$(MAKE) -f $(THIS_FILE) test-paths-lifo
+	@$(MAKE) -f $(THIS_FILE) test-cprover-smt2
```
tautschnig: How about making the `test` target depend on `test-paths-lifo` and `test-cprover-smt2`?

SaswatPadhi: Yes, either that, or make `default` target depend on `test-paths-lifo` and `test-cprover-smt2`?

tautschnig:
I'm wondering whether tests would pass if this were rebased onto current develop, which includes #5985.

regression/Makefile
```diff
@@ -72,14 +72,14 @@ endif
...
-		$(MAKE) "$$dir" || exit 1; \
+		$(MAKE) "$$dir" test || exit 1; \
 	done;
 
 # Pattern to execute a single test suite directory
 .PHONY: $(DIRS)
 $(DIRS):
 	@echo "Running $@..." ;
-	$(MAKE) -C "$@" || exit 1;
+	$(MAKE) -C "$@" test || exit 1;
```
tautschnig: Wouldn't `test-parallel` (below) also need this change?

regression/cbmc/Makefile
SaswatPadhi: One issue to consider here is: if we add these additional targets as dependencies for `test`, we wouldn't be able to run "just `test`" ...

tautschnig: Good call. The right approach probably is to rename the existing `test` target and introduce a new (`PHONY`) `test` target.

PR #6001, opened 2021-04-02
apply_condition: notequal need not have constant operand

tautschnig:
Just as is done for equality, the not-equal case needs to test that the
LHS is a symbol and the RHS is a constant before attempting to constant
propagate.

The problem became apparent while trying to enable abort-after-failing-C
assert, but only on Visual Studio builds (owing to the code that the
assert macro expands to in Visual Studio).

### DIFF
--- a/regression/cbmc/apply_condition1/main.c
+++ b/regression/cbmc/apply_condition1/main.c
@@ -1,0 +1,10 @@
+int main()
+{
+  int y;
+  __CPROVER_bool x;
+  if(x == (__CPROVER_bool)y)
+  {
+    __CPROVER_assert(0, "assertion 0");
+  }
+  return 0;
+}
--- a/regression/cbmc/apply_condition1/test.desc
+++ b/regression/cbmc/apply_condition1/test.desc
@@ -1,0 +1,9 @@
+CORE
+main.c
+
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+^Invariant check failed
+^warning: ignoring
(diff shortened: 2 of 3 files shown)


TGWDB (approved):
The commentary above  says this does the same checks as the `equal_expr` case, but they use different code style, consistency here would be nice. Also some in code comments/notes for future reference.

A regression test would also be good if this can be easily isolated.

martin-cs (approved):
Seems reasonable but as @TGWDB says, if it is not too hard to isolate a test case it would be good to see the before / after difference.

tautschnig:
I've now added a test.

tautschnig:
> The commentary above says this does the same checks as the `equal_expr` case, but they use different code style, consistency here would be nice. Also some in code comments/notes for future reference.
> 
[...]

I'm wondering whether it's worth the additional code churn to make them use the same code style? I tried to keep the change to a minimum, but could surely change this if you (or maybe others) think this is a priority.

TGWDB:
> I'm wondering whether it's worth the additional code churn to make them use the same code style? I tried to keep the change to a minimum, but could surely change this if you (or maybe others) think this is a priority.

Approved regardless of this. More just an observation that it's less clear conceptually when there are two similar paths where the first one has the structure
```
if(a && b)
{
  do_stuff();
}
```
and the other that is the "same" conceptually uses
```
if(!a || !b)
  return;
do_stuff();
```
That said, I agree that smaller changes are always nicer (and less changes are less potential mistakes).

tautschnig:
> That said, I agree that smaller changes are always nicer (and less changes are less potential mistakes).

Ok, cool, I agree that similar code structure provides for fewer surprises and thus helps readability a lot. I just wasn't sure whether the cost of adding churn (by having to indent 6 lines) would be worth it. So I'll keep it as is for now.

kroening (approved):

PR #6004, opened 2021-04-03
[smt_convt] Fix usage of `use_array_of_bool`

SaswatPadhi:
Several SMT translation routines were not respecting `use_array_of_bool`, or led to a buggy SMT translation when they used it.

This change ensures consistent usage of this option across all array-related translation routines.

### DIFF
--- a/regression/cbmc/array_of_bool_as_bitvec/main.c
+++ b/regression/cbmc/array_of_bool_as_bitvec/main.c
@@ -1,0 +1,19 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+__CPROVER_bool w[8];
+
+void main()
+{
+  _Bool x[8] = {false};
+  __CPROVER_bool y[8] = {false};
+  bool *z = calloc(8, sizeof(bool));
+
+  unsigned idx;
+  __CPROVER_assume(0 <= idx && idx < 8);
+
+  assert(w[idx] == x[idx]);
+  assert(x[idx] == y[idx]);
+  assert(y[idx] == z[idx]);
+}
--- a/regression/cbmc/array_of_bool_as_bitvec/test-smt2-outfile.desc
+++ b/regression/cbmc/array_of_bool_as_bitvec/test-smt2-outfile.desc
@@ -1,0 +1,31 @@
+CORE broken-smt-backend
+main.c
+--smt2 --outfile -
+\(= \(select array_of\.2 i\) \(ite false #b1 #b0\)\)
+\(= \(select array\.3 \(\(_ zero_extend 32\) \|main::1::idx!0@1#1\|\)\) #b1\)
+\(= \(select array\.3 \(_ bv\d+ 64\)\) \(ite false #b1 #b0\)\)
+^EXIT=0$
+^SIGNAL=0$
+--
+\(= \(select array_of\.2 i\) false\)
+\(= \(select array\.3 \(\(_ zero_extend 32\) \|main::1::idx!0@1#1\|\)\) #b1 #b0\)
+\(= \(select array\.3 \(_ bv\d+ 64\)\) false\)
+--
+This test checks for correctness of BitVec-array encoding of Boolean arrays.
+
+Ideally, we shouldn't have to check the SMT output, but should run with backend
+SMT solvers. However, we currently cannot because of #5977 (also see #6005).
+
+For now, we tag this test as `broken-smt-backend` to avoid running main.c
+with `--z3` or `--cprover-smt2`, both of which encode Boolean arrays directly.
+Instead, we generate a generic SMT encoding with `--smt2` and check the output.
+
+Explanation of regexes:
+line 4,10: array.2 elements should be BitVec not Bool
+line 5,11: equality operator (=) should have 2 arguments not 3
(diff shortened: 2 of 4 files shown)


tautschnig (commented):
Seems reasonable to me, but would you have any tests that newly pass with this fix?

SaswatPadhi:
Thanks for taking a look, Michael.

The options `--z3` and `--cprover-smt2` set `use_array_of_bool` to `true` and so these fixes won't apply -- all of these fixes apply to cases when `bool` is modeled as `(_ BitVec 1)`.

I wanted to include the test case from #6005, basically use `--smt2 --z3` to run `z3` with `use_array_of_bool` set to `false`, but unfortunately that won't pass because of other issues (use of quantifiers within QF logic: see #6005 for more details).

martin-cs (approved):
+1 to what @tautschnig says, test please!

SaswatPadhi:
Thanks for the review, @martin-cs.

I am not sure how I could add a test for these changes ...
With `--smt2` CBMC would perform the SMT encoding with `use_array_of_bool = false` but the result won't run with Z3 because it would also set the logic to `QF_AUFBV`.

Any ideas? cc: @tautschnig @kroening.

Some background on why I made this PR and how I tested it:
1. with `--z3` CBMC produces (with `use_array_of_bool = true`) an SMT file that takes forever on Z3 -- no syntax error, but just really _really_ slow
2. with `--smt2 --z3` I was getting incorrect translations because `use_array_of_bool` was not handled correctly. With the changes proposed in this PR, the translation is correct and if I change the logic to `ALL` then Z3 reports `unsat` in a couple of seconds.

So there is some performance issue in Z3 that is triggered by `--z3` encoding but not `--smt2` encoding. But that should be in a separate PR. This just fixes translation errors with `--smt2` encoding.

SaswatPadhi:
After an offline discussion with @tautschnig, I have added a regression test to check the raw SMT2 output instead of running end-to-end against an SMT solver. Please take a look and see if it looks okay.

regression/cbmc/array_of_bool_as_bitvec/main.c
```diff
@@ -0,0 +1,12 @@
...
+
+__CPROVER_bool x[8];
+bool *y = malloc(8 * sizeof(bool));
+
+void main()
+{
+  unsigned idx;
+  __CPROVER_assume(0 <= idx && idx < 8);
+  assert(x[idx] == y[idx]);
+}
```
tautschnig: Missing newline at end of file.

regression/cbmc/array_of_bool_as_bitvec/main.c
```diff
@@ -0,0 +1,12 @@
+#include <stdbool.h>
+#include <stdlib.h>
+
+__CPROVER_bool x[8];
+bool *y = malloc(8 * sizeof(bool));
```
tautschnig: Use of `malloc` at global scope?

regression/cbmc/array_of_bool_as_bitvec/main.c
SaswatPadhi: Fixed

regression/cbmc/array_of_bool_as_bitvec/main.c
SaswatPadhi: Fixed

regression/cbmc/array_of_bool_as_bitvec/main.c
```diff
@@ -0,0 +1,20 @@
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+
+__CPROVER_bool x[8];
```
tautschnig: Unused?

regression/cbmc/array_of_bool_as_bitvec/main.c
```diff
@@ -0,0 +1,20 @@
...
+
+void main()
+{
+  bool y[8];
+  bool *z = malloc(8 * sizeof(bool));
+
+  memset(y, 0, sizeof(y));
+  memset(z, 0, sizeof(8 * sizeof(bool)));
```
tautschnig: Wouldn't
```
bool y[8] = { 0 };
bool *z = calloc(8, sizeof(boo));
```
also do the trick, or does this not trigger the desired code paths? Likewise, is it actually necessary to use `bool`, or could `_Bool` or `int` equally be used?

regression/cbmc/array_of_bool_as_bitvec/main.c
SaswatPadhi: Used @ line 18 (in the `assert`).

regression/cbmc/array_of_bool_as_bitvec/main.c
SaswatPadhi: Only `__CPROVER_bool` seems to trigger the desired code paths.
I have these `bool` variables (which are actually encoded as `int`) just to make sure their encoding is unaffected by the proposed changes.

I am not sure what `_Bool` is -- I will try this out and also modify the array initializations as you suggested.

SaswatPadhi: Looks like `_Bool` doesn't trigger the code path either.

I have now implemented the changes as suggested.

PR #6006, opened 2021-04-03
Add TODO in invariant_set seeking to understand magic number

tautschnig:
When trying to eliminate hard-coded uses of "8" in the code base I came
across this piece of code, which I don't actually understand the purpose
of. Should someone decipher invariant_set.cpp at some later date:
perhaps they can clean this up.


peterschrammel (approved):
I have never stumbled upon that piece of code either.

martin-cs:
FWIW I think this code was used in / intended for use in SatAbs.  I don't think it is used in the current code base and I'm not sure it has been used in a *long* time.

What it is supposed to do is pretty cool though, so I have wondered if it should be resurrected via VSD.

PR #6007, opened 2021-04-03
ansi_c_entry_point: avoid magic number

tautschnig:
The code previously hard-coded an upper bound for argc that perhaps was
suitable for 32-bit systems (albeit even questionable for that case).
Replace this calculation by one that refers to configured bit widths.

### DIFF
--- a/src/ansi-c/ansi_c_entry_point.cpp
+++ b/src/ansi-c/ansi_c_entry_point.cpp
@@ -313,12 +313,10 @@ bool generate_ansi_c_start_function(
         init_code.add(code_assumet(std::move(ge)));
       }
 
+      if(config.ansi_c.max_argc.has_value())
       {
-        // assume argc is at most MAX/8-1
-        mp_integer upper_bound=
-          power(2, config.ansi_c.int_width-4);
-
-        exprt bound_expr=from_integer(upper_bound, argc_symbol.type);
+        exprt bound_expr =
+          from_integer(*config.ansi_c.max_argc, argc_symbol.type);
 
         binary_relation_exprt le(
           argc_symbol.symbol_expr(), ID_le, std::move(bound_expr));
--- a/regression/cbmc/argv2/main.i
+++ b/regression/cbmc/argv2/main.i
@@ -1,0 +1,6 @@
+int main(int argc, char *argv[])
+{
+  __CPROVER_assert(
+    sizeof(char *) > sizeof(int) || argc < 0x7FFFFFFF,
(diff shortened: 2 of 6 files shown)


peterschrammel (approved):
On real systems this might be substantially lower, see `getconf ARG_MAX` for POSIX, for instance. Maybe we should have an instrumentation option that allows setting the value (with the system's limit as default)?

tautschnig:
> On real systems this might be substantially lower, see `getconf ARG_MAX` for POSIX, for instance. Maybe we should have an instrumentation option that allows setting the value (with the system's limit as default)?

Good call. I've now handled the Windows case, but for Unix it will indeed require an instrumentation option. I'm not sure how useful this even is, to be honest, so I'll not bother implementing this for now. We have `--model-argc-argv` in `goto-instrument`, which likely is more useful already.

TGWDB (approved):

peterschrammel (approved):

src/ansi-c/ansi_c_entry_point.cpp
```diff
@@ -313,18 +313,42 @@ bool generate_ansi_c_start_function(
...
+      // On Windows, CreateProcess accepts no more than 32767 characters, so
+      // make that a hard limit.
+      if(config.ansi_c.os == configt::ansi_ct::ost::OS_WIN)
       {
-        // assume argc is at most MAX/8-1
-        mp_integer upper_bound=
-          power(2, config.ansi_c.int_width-4);
-
-        exprt bound_expr=from_integer(upper_bound, argc_symbol.type);
+        exprt bound_expr = from_integer(32767, argc_symbol.type);
```
chrisr-diffblue: A non-blocking suggestion... but given the PR is titled "avoid magic number" this looks suspciously like a magic number 😁 How about moving this value to a hypothetical `config.ansi_c.os_max_argc`, and then just use that here without the `if(windows) {} else {}` logic? That way the platform dependent logic could be in config where it morally deserves to be, and some of the duplicated code here (creating the `binary_relation_exprt`, adding it to the `init_code`) could be de-dup'ed.

chrisr-diffblue (approved):
Looks fine to me as is. Take my one comment as a suggestion to ponder, and not a requirement.

src/ansi-c/ansi_c_entry_point.cpp
tautschnig: Thank you!! I've modified the implementation as you suggested.

chrisr-diffblue (approved):
Much cleaner to my (biased) eye - thanks for the updates :-)

PR #6009, opened 2021-04-03
Configure bits_per_byte in byte_extract/byte_update expression

tautschnig:
Semantically evaluating a byte_extract/byte_update expression requires
knowledge of the number of bits that a byte is composed of. This,
however, is a configuration parameter known to the front-end that
created the expression in the first place. The back-end should not need
to consult configuration information to evaluate the expression.

### DIFF
--- a/jbmc/unit/java_bytecode/java_trace_validation/java_trace_validation.cpp
+++ b/jbmc/unit/java_bytecode/java_trace_validation/java_trace_validation.cpp
@@ -6,17 +6,18 @@ Author: Diffblue Ltd.
 
 \*******************************************************************/
 
+#include <util/byte_operators.h>
+#include <util/pointer_expr.h>
+#include <util/symbol_table.h>
+
 #include <goto-programs/goto_trace.h>
 
 #include <java_bytecode/java_trace_validation.h>
 #include <java_bytecode/java_types.h>
-
 #include <testing-utils/message.h>
 #include <testing-utils/use_catch.h>
 
-#include <util/byte_operators.h>
-#include <util/pointer_expr.h>
-#include <util/symbol_table.h>
+#include <climits>
 
 TEST_CASE("java trace validation", "[core][java_trace_validation]")
 {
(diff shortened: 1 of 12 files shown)


jbmc/unit/java_bytecode/java_trace_validation/java_trace_validation.cpp
```diff
@@ -33,9 +33,9 @@ TEST_CASE("java trace validation", "[core][java_trace_validation]")
     index_exprt(valid_symbol_expr, valid_constant);
   const index_exprt index_plain = index_exprt(exprt(), exprt());
   const byte_extract_exprt byte_little_endian = byte_extract_exprt(
-    ID_byte_extract_little_endian, exprt(), exprt(), typet());
-  const byte_extract_exprt byte_big_endian =
-    byte_extract_exprt(ID_byte_extract_big_endian, exprt(), exprt(), typet());
+    ID_byte_extract_little_endian, exprt(), exprt(), 8, typet());
```
martin-cs: I know one of the core CPROVER developers just purged all of the "magic 8" constants.  Could you at least use a macro for these?  I really don't want to make them feel their work was for nothing.

unit/util/expr_cast/expr_cast.cpp
```diff
@@ -149,6 +149,7 @@ SCENARIO("expr_dynamic_cast",
       ID_byte_extract_little_endian,
       symbol_exprt(typet()),
       constant_exprt("0", typet()),
+      8,
```
martin-cs: At least `CHAR_BIT`?  Or some other macro?

martin-cs (approved):
I like obscure architectures but ... you have a use-case for this, right?

unit/util/expr_cast/expr_cast.cpp
tautschnig: Actually it doesn't matter, I've replaced this by `1, // arbitrary integer`.

jbmc/unit/java_bytecode/java_trace_validation/java_trace_validation.cpp
tautschnig: Using `CHAR_BIT` here.

tautschnig:
> I like obscure architectures but ... you have a use-case for this, right?

While I do know of an organisation that had a use case, I'm not sure they are still using CBMC. Either way, I'd like to get rid of all magic numbers.

peterschrammel (approved):
🎉

PR #3407, opened 2018-11-13
enable -flto for unix systems

nmanthey:
These changes might improve the performance of CBMC, as well as allow to spot compilation problems earlier.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [x] My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #6011, opened 2021-04-04
Enable IPO / LTO and -O3

rurban:
LTO and -O3 give a good performance improvement.
I would not recommend -O3 with bad compiler versions though, such
as gcc-9


rurban:
ubuntu 18 failed: 
old compilers with -O3 complain with bogus -Wmaybe-uninitialized for *begin_range (g++-7.4)
maybe disable -O3 for this function, or silence this warning.

I went with silencing this wrong warning

martin-cs (commented):
Thank you for the PR.  If we are going to make this change then `config.inc` will also need to change in sync.

tautschnig:
@rurban Would you be able to provide numbers to back up "good performance improvement?" I once tried this myself, and didn't see much of a difference, but perhaps I didn't use a good-enough compiler version or did not try all the options.

tautschnig:
FWIW, we already have #3407, which should likely be closed, but equally needed more data.

rurban:
Performance numbers: Typically 10-20% faster, which aligns to my other projects where -O3 -flto brought about the same improvements.
I usually verify only very small files. Big files are usually not solvable, or only with satabs.

For my project rurban/ctl (the C++ STL in C): make verify

* /usr/bin/cbmc -O3 -flto gcc-10
  CBMC version 5.26.1 (cbmc-5.26.1-exp-89-g214216ef7) 64-bit x86_64 linux

* /usr/bin/cbmc -O2 gcc-10
  CBMC version 5.25.0 (n/a) 64-bit x86_64 linux

```sh
$ time for f in tests/verify/*.c; do /usr/bin/cbmc --bounds-check --pointer-check --memory-leak-check --div-by-zero-check --signed-overflow-check --unsigned-overflow-check --pointer-overflow-check --conversion-check --undefined-shift-check --enum-range-check --pointer-primitive-check --compact-trace --depth 6 --unwind 6 -I. $f >/dev/null; done

real	0m3.673s
user	0m2.245s
sys	0m0.337s

$ time for f in tests/verify/*.c; do /usr/local/bin/cbmc --bounds-check --pointer-check --memory-leak-check --div-by-zero-check --signed-overflow-check --unsigned-overflow-check --pointer-overflow-check --conversion-check --undefined-shift-check --enum-range-check --pointer-primitive-check --compact-trace --depth 6 --unwind 6 -I. $f >/dev/null; done

real	0m2.819s
user	0m2.196s
sys	0m0.335s

```

when I add `-march=native` I get `real	0m1.942s`. This is valid for all gcc-compatible compilers on Intel.
```patch
diff --git CMakeLists.txt CMakeLists.txt
index fe1bd4ca7..dc0602cf5 100644
--- CMakeLists.txt
+++ CMakeLists.txt
@@ -202,6 +202,21 @@ if(ipo_supported)
   message(STATUS "IPO / LTO enabled")
   set_property(GLOBAL PROPERTY INTERPROCEDURAL_OPTIMIZATION True)
   add_definitions(-DLTO)
+  if ((CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "i386")
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "i586")
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "i686")
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64") # freebsd
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64") # windows
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86"))
+     set(PROCESSOR_FAMILY "Intel")
+     if(CMAKE_COMPILER_IS_GNUCC
+         OR (CMAKE_C_COMPILER_ID STREQUAL AppleClang)
+         OR (CMAKE_C_COMPILER_ID STREQUAL Clang)
+         OR (CMAKE_C_COMPILER_ID STREQUAL Intel))
+       add_definitions(-march=native)
+     endif()
+  endif()
 else()
   message(STATUS "IPO / LTO not supported: <${error}>")
 endif()
```

src/util/CMakeLists.txt
```diff
@@ -33,12 +33,17 @@ file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/version.cpp.in
...
     -D CBMC_SOURCE_DIR=${CBMC_SOURCE_DIR}
     -D CUR=${CMAKE_CURRENT_BINARY_DIR}
     -P ${CMAKE_BINARY_DIR}/version.cmake
-)
+    )
+
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND "${CMAKE_BUILD_TYPE}" STREQUAL "Release")
+  set_source_files_properties(format_number_range.cpp
+    PROPERTIES COMPILE_FLAGS "${CMAKE_CXX_FLAGS} -O2")
+endif()
```
tautschnig: Perhaps we should instead rewrite the code in `format_number_range.cpp`? It could probably be replaced by multiple functions that handle the case of having/not having a value. Having a need for such excessive handling of possible error cases doesn't seem to be a good idea in a `static` function.

src/config.inc
```diff
@@ -76,7 +76,7 @@ endif
 OSX_IDENTITY="Developer ID Application: Daniel Kroening"
 
 # Detailed version information
-CBMC_VERSION = 5.26.0
+CBMC_VERSION = 5.26.1
```
thomasspriggs: :no_entry_sign: Please drop this commit. The current release is now 5.27.0 and the change to `config.inc` for that release has already been made and merged to `develop`.

src/config.inc
```diff
@@ -14,9 +14,12 @@ ifeq ($(CPROVER_WITH_PROFILING),1)
 endif
 
 # Select optimisation or debug info
-#CXXFLAGS += -O2 -DNDEBUG
+#CXXFLAGS += -O2 -DNDEBUG -flto -DLTO
```
martin-cs: `-O3`?

src/config.inc
```diff
@@ -76,7 +79,7 @@ endif
 OSX_IDENTITY="Developer ID Application: Daniel Kroening"
 
 # Detailed version information
-CBMC_VERSION = 5.26.0
+CBMC_VERSION = 5.26.1
```
martin-cs: I think you may have rebased on to a branch other than `develop`?

martin-cs:
@rurban Thank you for the timing results.  Normally our focus tends to be on run-times in the 10s--1200s run-times.  If we get a 10-20% performance benefit on those then that would be a very compelling argument.

Am I reading your comments correctly that you are using SatAbs?  If so, which version?

src/config.inc
rurban: oops, sure

src/config.inc
rurban: On top of the version PR #6014

rurban:
I'm using satabs 3.2 with boom 1.8. 
I haven't checked for updated versions there. I still run into WIP array stuff there. 
```
Refining set of predicates according to counterexample (WP)
*** CEGAR Loop Iteration 2
Computing Predicate Abstraction for Program
sorry, wp(array_copy...) not implemented
```

src/util/CMakeLists.txt
tautschnig: I didn't really get rid of the error handling, but still did the refactoring in #6017 to make this warning go away.

src/util/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ if(GIT_FOUND)
       OUTPUT_STRIP_TRAILING_WHITESPACE
       WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     )
-    configure_file(\${CUR}/version.cpp.in version.cpp)
+    configure_file(\${CUR}/version.cpp.in \${CBMC_SOURCE_DIR}/util/version.cpp)
```
thomasspriggs: :no_entry_sign: Please drop this commit as this "fix" is not needed.

tautschnig:
Would you mind including the following patch (or something to that effect):
````patch
--- a/src/solvers/CMakeLists.txt
+++ b/src/solvers/CMakeLists.txt
@@ -109,7 +109,13 @@ elseif("${sat_impl}" STREQUAL "cadical")
     )

     message(STATUS "Building CaDiCaL")
-    execute_process(COMMAND make WORKING_DIRECTORY ${cadical_SOURCE_DIR})
+    if("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
+      execute_process(COMMAND make CXXFLAGS=${CMAKE_CXX_FLAGS_RELEASE} WORKING_DIRECTORY ${cadical_SOURCE_DIR})
+    elseif("${CMAKE_BUILD_TYPE}" STREQUAL "RelWithDebInfo")
+      execute_process(COMMAND make CXXFLAGS=${CMAKE_CXX_FLAGS_RELWITHDEBINFO} WORKING_DIRECTORY ${cadical_SOURCE_DIR})
+    else()
+      execute_process(COMMAND make WORKING_DIRECTORY ${cadical_SOURCE_DIR})
+    endif()

     target_compile_definitions(solvers PUBLIC
         SATCHECK_CADICAL HAVE_CADICAL
````
to ensure the CaDiCaL solver is also built with the new options?

File: src/util/CMakeLists.txt (hunk 5 above)
Comment (rurban) [reply src/util/CMakeLists.txt]: without it version.cpp gets generated in the build dir. and the new version is ignored.

File: src/util/CMakeLists.txt (hunk 1 above)
Comment (rurban) [reply src/util/CMakeLists.txt]: Let's see of the smoke comes out green with your version and O3 on Ubuntu 18

File: CMakeLists.txt [hunk 6]
```diff
@@ -191,6 +191,36 @@ endif()
...
+   AND (CMAKE_BUILD_TYPE STREQUAL "Release"))
+  cmake_policy(SET CMP0069 NEW)
+  include(CheckIPOSupported)
+  check_ipo_supported(RESULT ipo_supported OUTPUT error)
+endif()
+
+if(ipo_supported)
+  message(STATUS "IPO / LTO enabled")
+  set_property(GLOBAL PROPERTY INTERPROCEDURAL_OPTIMIZATION True)
+  add_definitions(-DLTO)
```
tautschnig, line 208: CMake noob here: could we use `set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -DLTO")` instead?

CMakeLists.txt
```diff
@@ -191,6 +191,36 @@ endif()
...
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "i686")
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64") # freebsd
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64") # windows
+       OR (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86"))
+     set(PROCESSOR_FAMILY "Intel")
+     if(CMAKE_COMPILER_IS_GNUCC
+         OR (CMAKE_C_COMPILER_ID STREQUAL AppleClang)
+         OR (CMAKE_C_COMPILER_ID STREQUAL Clang)
+         OR (CMAKE_C_COMPILER_ID STREQUAL Intel))
+       add_definitions(-march=native)
```
tautschnig:
```suggestion
     if(CMAKE_COMPILER_IS_GNUCC
         OR (CMAKE_CXX_COMPILER_ID STREQUAL AppleClang)
         OR (CMAKE_CXX_COMPILER_ID STREQUAL Clang)
         OR (CMAKE_CXX_COMPILER_ID STREQUAL Intel))
       set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -march=native")
```

tautschnig (commented):
Is there a reliable way in CMake to obtain the actual compile flags? As a CMake noob the best I could find is not using `add_definitions` and instead going for `set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} ...")` so that `CMAKE_CXX_FLAGS_RELEASE` can be used. I'm asking to make sure the CaDiCaL build actually benefits.

martin-cs:
On Tue, 2021-04-06 at 12:30 -0700, Reini Urban wrote:
> I'm using satabs 3.2 with boom 1.8. 
> I haven't checked for updated versions there. I still run into WIP
> array stuff there. 
> ```
> Refining set of predicates according to counterexample (WP)
> *** CEGAR Loop Iteration 2
> Computing Predicate Abstraction for Program
> sorry, wp(array_copy...) not implemented
> ```


At the risk of side-tracking this PR, I am surprised to run find
someone using SatAbs.  I had thought it long dead.

The last version I have was built on CBMC 4.7 which was released in Jan
2014. As far as I know that was a case of light maintainance and the
last time it had active development was probably pre-2012.

I am not aware of any currently available copies of the source code.  I
have a check out of the subversion from late 2015 and I can send you
that / post it on github if you would like.  I believe at least one of
@tautschnig or @schrammel has a more up-to-date one.  I don't know
which version I have and it will almost certainly take a non-trivial
amount of work to get it to build / work with more modern versions of
CPROVER.

Given the improvements we have made to CPROVER, I do wonder whether it
might be better to just rewrite it completely if there is interest.

CMakeLists.txt
rurban: Yes, but this is easier. It only adds it to the Release flags anyway, because only there ipo_supported will be true.

CMakeLists.txt
rurban: CXX compiler! good catch

CMakeLists.txt
tautschnig: Understood, but then I wasn't able to find a way to surface this to the CaDiCaL build configuration. You might know how to make this work?

src/util/CMakeLists.txt
thomasspriggs: The `version.cpp` in the build dir is compiled and linked into the build. Ignoring this file would result in a failed build as it would not exist elsewhere. The latest Ubuntu build on our release page shows the correct version number when I download and install it and it was built without this proposed change. I also end up with CBMC builds showing the correct version number for builds I produce locally, with altered version numbers in `config.inc`. Under what conditions have you seen incorrect version numbers?

If you have also built cbmc using `make` and the makefiles, then you can end up with a `version.cpp` in the `cbmc/src/util` directory. that would be a stale build artefact, from the build in place nature of the make build. It should not be committed to git. To my knowledge it is correctly ignored and neither overwritten or linked by the cmake build, if it exists.

rurban:
Cannot repro the ubuntu20 cmake-gcc crash, but looks bad. Was introduced by `--march=native`, so it looks like a bad alignment bug. Either a function, but more like an argument, which wanted to be vectorized but was not 16-byte aligned.

I also notized that the GLOBAL target did not enable all targets to be compiled with `-flto`. Got that working, but am now in different, more expected fields of bugs. Will need some time.

CMakeLists.txt
chrisr-diffblue: I think you might actually need _both_ the C and CXX variants?

Schrammel:
Hello @martin-cs!

I never had the source code of this project, sorry,I believe that you have marked the wrong person...

CMakeLists.txt
chrisr-diffblue, line 221: I'm a bit nervous of the `-march=native` - if I'm reading the CMake correctly (a big if :-) ) then this is enabled by default on release builds, but thats quite likely to cause problems building CBMC release distributions (or at least, be a trap waiting to cause problems). `native` means "use any instructions available on the build machine" I believe - but we've got no real guarantee that our build machines are not higher spec/newer CPUs/whatever than any of the downstream users of our binary packages, meaning they run the risk of getting illegal instruction crashes/traps from our released binaries. While we _could_ override default options for our binary packaging, I'd feel much more comfortable if we didn't have to...

chrisr-diffblue:
While I'm definitely supportive of "free speed" form -O3/LTO - I must admit I'm a bit weary of enabling both of these by default. In all honesty, I may just be being paranoid for the decade I spent working as a compiler engineer, :-) but I've seen an awful lot of codegen bug reports where '-O3' was a common factor - and codegen bugs are insidious to discover. Given what CBMC is, and that this PR enables `-O3` by default, I think it's worth asking an explicit question: How much do we want to trade potential speed for potential increase risk of silent incorrect results? Not against this going in, but just think it is worth an explicit decision on the risk aspect.

CMakeLists.txt
rurban: True. Packagers need to get rid of that somehow. Maybe only add it to CFLAGS and CXXFLAGS via env. The idea was to find sloppy visibility and alignment, and test it in develop. Also have locally fast binaries

src/util/CMakeLists.txt
rurban: Sorry. You are right. My mistake with the wrong prefix.

In the general case build vs src mixup, you need to build out of src to get wrong version.cpp files

rurban:
The `-O3` idea is to catch compiler or code bugs earlier in the develop branch. Inlining and vectorization catches much more than without, E.g. wrong visibility, wrong alignment, bad compiler versions, ...

Same for `-flto` and `-march-native`. I'm already blocked by some bugs caught by improved lto builds.
`-march-native` of course needs to be turned off by packagers.

tautschnig:
> The `-O3` idea is to catch compiler or code bugs earlier in the develop branch. Inlining and vectorization catches much more than without, E.g. wrong visibility, wrong alignment, bad compiler versions, ...
> 
> Same for `-flto` and `-march-native`. I'm already blocked by some bugs caught by improved lto builds.
> `-march-native` of course needs to be turned off by packagers.

That suggests 1) adding code to permit building with that configuration and 2) adding a further GitHub action using that configuration rather than making it the default?

Also, can you provide some hints on how to best debug this? I wasn't able to reproduce the failure on a 20.04 system myself.

rurban:
I'll redo in a new branch, because of the blockers

thomasspriggs:
It might be worth considering `-O3` and the IPO / LTO changes as separate PRs. If we switch on one before the other, then we might be in a better place to work out which of the changes caused any issues which we run into later. Do you think IPO / LTO is less risky than `-O3` optimisation?

rurban:
LTO is harder to achieve.

O3 is riskier to fall into compiler bugs. I found several.

PR #6012, opened 2021-04-05
Allow quantifiers within loop invariants

ArenBabikian:
This PR adds support for quantified expressions in loop contracts. ~~This PR depends on #5942.~~

We have included 3 regression tests in this PR. However, two of them are marked as `KNOWNBUG` due to a current limitation of quantifiers that does not allow symbolic ranges for the quantified variable when using the SAT back-end.

### DIFF
--- a/regression/contracts/quantifiers-loop-01/main.c
+++ b/regression/contracts/quantifiers-loop-01/main.c
@@ -1,0 +1,29 @@
+#include <assert.h>
+
+#define N 16
+
+void main()
+{
+  int a[N];
+  a[10] = 0;
+
+  for(int i = 0; i < N; ++i)
+    // clang-format off
+    __CPROVER_loop_invariant(
+      (0 <= i) && (i <= N) &&
+      __CPROVER_forall {
+        int k;
+        // constant bounds for explicit unrolling with SAT backend
+        (0 <= k && k <= N) ==> (
+          // the actual symbolic bound for `k`
+          k < i ==> a[k] == 1
+        )
+      }
+    )
+    // clang-format on
+    {
+      a[i] = 1;
+    }
+
+  assert(a[10] == 1);
+}
--- a/regression/contracts/quantifiers-loop-01/test.desc
+++ b/regression/contracts/quantifiers-loop-01/test.desc
@@ -1,0 +1,15 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^\[main.1\] line .* Check loop invariant before entry: SUCCESS
+^\[main.2\] line .* Check that loop invariant is preserved: SUCCESS
+^\[main.assertion.1\] line .* assertion a\[10\] == 1: SUCCESS
+^VERIFICATION SUCCESSFUL$
+--
+--
+This test case checks the handling of a `forall` quantifier within a loop invariant.
+
+This test case uses explicit constant bounds on the quantified variable,
+so that it can be unrolled (to conjunctions) with the SAT backend.
--- a/regression/contracts/quantifiers-loop-02/main.c
+++ b/regression/contracts/quantifiers-loop-02/main.c
@@ -1,0 +1,32 @@
+#include <assert.h>
+
+void main()
+{
+  int N, a[64];
+  __CPROVER_assume(0 <= N && N < 64);
+
+  for(int i = 0; i < N; ++i)
+    // clang-format off
+    __CPROVER_loop_invariant(
+      (0 <= i) && (i <= N) &&
+      __CPROVER_forall {
+        int k;
+        (0 <= k && k < i) ==> a[k] == 1
+      }
+    )
+    // clang-format on
+    {
+      a[i] = 1;
+    }
+
+  // clang-format off
+  assert(__CPROVER_forall {
+    int k;
+    (0 <= k && k < N) ==> a[k] == 1
+  });
+  // clang-format on
+
+  int k;
+  __CPROVER_assume(0 <= k && k < N);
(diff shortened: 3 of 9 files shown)


SaswatPadhi:
Hi @ArenBabikian,

For the `KNOWNBUG` test cases, could you please try the SMT backend? CPROVER SMT2 is not expected to succeed if the SAT backend does not, but we might have some luck with Z3.

- Try running CBMC with `--z3` locally
- For CI, replace `KNOWNBUG` with `CORE smt-backend broken-cprover-smt-backend`

SaswatPadhi:
Marking this as draft for now.

The failure traces look like the one from #6106, but they don't use `char*`. Need to take a closer look, and may be #6106 is due to some issue with contracts -> assertion/assumption translation and not `char *` or `__CPROVER_w_ok`.

regression/contracts/quantifiers-loop-01/main.c
```diff
@@ -0,0 +1,20 @@
...
+{
+  int N = 64, a[64];
+
+  for(int i = 0; i < N; ++i)
+    // clang-format off
+    __CPROVER_loop_invariant(
+      (0 <= i) && (i <= N) &&
+      __CPROVER_forall {
+        int k;
+        (0 <= k && k < N) ==> 1 == 1
```
martin-cs: Is the `1 == 1` so that it could simplify away completely intentional?

regression/contracts/quantifiers-loop-01/test.desc
```diff
@@ -0,0 +1,11 @@
...
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^\[main.1\] line .* Check loop invariant before entry: SUCCESS
+^\[main.2\] line .* Check that loop invariant is preserved: SUCCESS
+^VERIFICATION SUCCESSFUL$
+--
+--
+This test case checks the handling of a `forall` quantifier within a loop invariant.
```
martin-cs, line 12: If it is, then please can you expand the description here?

martin-cs (approved):
This doesn't look too controversial...

regression/contracts/quantifiers-loop-01/main.c
feliperodri: Shouldn't the invariant be the following?
```c
      __CPROVER_forall {
         int k;
         (0 <= k && k < N) ==> a[k] = 1
      }
```

regression/contracts/quantifiers-loop-02/main.c
```diff
@@ -0,0 +1,32 @@
...
+  // clang-format off
+  assert(__CPROVER_forall {
+    int k;
+    (0 <= k && k < N) ==> a[k] == 1
+  });
+  // clang-format on
+
+  int k;
+  __CPROVER_assume(0 <= k && k < N);
+  assert(a[k] == 1);
```
feliperodri, line 31: Why do we need this block here? It seems redundant.

src/goto-instrument/code_contracts.cpp
```diff
@@ -213,12 +225,10 @@ void code_contractst::add_quantified_variable(
   else if(expression.id() == ID_exists || expression.id() == ID_forall)
   {
     // When a quantifier expression is found,
-    // 1. get quantified variables
+    // for each quantified variable ...
```
feliperodri, line 234: If you decide to remove `1.`, update the following comments too; otherwise, leave it here.

src/goto-instrument/code_contracts.cpp
```diff
@@ -127,9 +137,11 @@ void code_contractst::check_apply_invariants(
...
 
   // assume the invariant
-  code_assumet assumption{invariant};
-  assumption.add_source_location() = loop_head->source_location;
-  converter.goto_convert(assumption, havoc_code, mode);
+  {
+    code_assumet assumption{invariant_expr()};
+    assumption.add_source_location() = loop_head->source_location;
+    converter.goto_convert(assumption, havoc_code, mode);
+  }
```
feliperodri, line 148: Why are you adding `{...}`?

feliperodri (changes requested):

regression/contracts/quantifiers-loop-01/main.c
SaswatPadhi: `(0 <= k && k < N) ==> a[k] = 1` wouldn't hold, but I am adding `(0 <= k && k < i) ==> a[k] = 1` to address this concern.

src/goto-instrument/code_contracts.cpp
SaswatPadhi: For scoping variables.

The invariant gets expanded to multiple assertions and assumptions. I scope these, for example the `assumption` object in this case, so they don't pollute the scope of the overall function. They are immediately compiled to goto instructions and added to `havoc_code`.

regression/contracts/quantifiers-loop-01/test.desc
SaswatPadhi: Added some description.

src/goto-instrument/code_contracts.cpp
SaswatPadhi: Done; I updated the comments.

regression/contracts/quantifiers-loop-02/main.c
SaswatPadhi: I kept it around just to nondeterministically check the array without invoking quantifiers at all, and ensure consistency between both the quantified interpretation and this non-det interpretation.

feliperodri (approved):

tautschnig (approved):

PR #6014, opened 2021-04-06
Fix generate_version_cpp target

rurban:
Create version.cpp in the src dir, where it is picked up.
And not in the build dir, where it is ignored.
    
Repro: check `bin/cbmc --version` for the correct version.
Or just  `strings src/util/CMakeFiles/util.dir/version.cpp.o`

### DIFF
--- a/src/config.inc
+++ b/src/config.inc
@@ -76,7 +76,7 @@ endif
 OSX_IDENTITY="Developer ID Application: Daniel Kroening"
 
 # Detailed version information
-CBMC_VERSION = 5.26.0
+CBMC_VERSION = 5.26.1
 
 # Use the CUDD library for BDDs, can be installed using `make -C src cudd-download`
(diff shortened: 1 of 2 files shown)


TGWDB:
@rurban The generated version.cpp file should be in a build directory not a source directory (as this PR seems to specify). Can you provide a little more clarity on what the problem this fix is meant to solve is? Note that it appears the current build process picks up the version number correctly.

P.S. This will conflict with #6015 , so if there does need to be a fix to the version.cpp location please separate this from a version number change.

thomasspriggs:
The current build system is correct. A version mismatch was shown in the file names / release page for 5.26.1 due to a test of the github action which performs the release packaging, without us doing the corresponding update to `config.inc`. The current version is intended to be 5.26.0, not 5.26.1 and we have removed the invalid 5.26.1 release from the release page.

PR #6023, opened 2021-04-12
Vsd meet and assume

jezhiggins:
Two related strands of work in this PR. 

* Extending `abstract_environmentt::assume` to cover all value representations. Up until now, `assume` was only correct for constants. It _may_ have evaluated a given expression correctly for intervals and value-sets, but not reliably. These changes all assume expressions to be properly evaluated for all value representation, and in the face of different representations within the same expression.

* Filling out the implementation of `abstract_objectt::meet` for value representations. Where `merge` is the union of two values, `meet` is the intersection. `meet` is used when evaluating an assumed equality expression. 

+2,331/-384 looks alarming, but the major bulk of that is test cases exercising various combinations of constants, intervals, value-sets, and operators in assume and in meet.


jezhiggins:
MacOS build failures look to be a meta-problem getting the tool dependencies 

```
==> Downloading https://homebrew.bintray.com/bottles/ninja-1.10.2.catalina.bottle.tar.gz
curl: (22) The requested URL returned error: 403 Forbidden
Error: Failed to download resource "ninja"
Download failed: https://homebrew.bintray.com/bottles/ninja-1.10.2.catalina.bottle.tar.gz
Error: Process completed with exit code 1.
```

martin-cs:
Paging @NlightNFotis and @thomasspriggs for the Mac OS CI issues.

NlightNFotis:
I took a quick look at it - in the beginning I thought that they upgraded the version so it's hitting the wrong URL because of outdated database issues (kind of like needing an apt-get update, if you want).

But it looks like it's still the same version of the formula (https://formulae.brew.sh/formula/ninja), and it's 403 if you hit the URL, which seems to be sitting on their servers, so I'm assuming this must be a temporary issue they are having.

Could you try it again (just restart the jobs from the github interface) later, and let me know if it doesn't work? I'll take a deeper look at it then.

tautschnig:
> MacOS build failures look to be a meta-problem getting the tool dependencies
> 
> ```
> ==> Downloading https://homebrew.bintray.com/bottles/ninja-1.10.2.catalina.bottle.tar.gz
> curl: (22) The requested URL returned error: 403 Forbidden
> Error: Failed to download resource "ninja"
> Download failed: https://homebrew.bintray.com/bottles/ninja-1.10.2.catalina.bottle.tar.gz
> Error: Process completed with exit code 1.
> ```

For the record: https://github.com/actions/virtual-environments/issues/3165 - presumably the current run will have updated runners already.

regression/goto-analyzer/branching-ge/test-always-intervals.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main-always.c
+--variable-sensitivity --vsd-values intervals --show
+^EXIT=0$
+^SIGNAL=0$
+^main::1::i .* \[5, 5\] @ \[17\]
+^main::1::p .* \[0, A\] @ \[3, 8, 14\]
```
martin-cs, line 7: One day, I will have to fix the printing in hex...

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -196,21 +225,18 @@ bool abstract_environmentt::assume(const exprt &expr, const namespacet &ns)
...
     // Should be of the right type
     INVARIANT(
-      possibly_constant.type().id() == ID_bool, "simplication preserves type");
+      assumption.type().id() == ID_bool, "simplification preserves type");
 
-    if(possibly_constant.is_false())
+    if(assumption.is_false())
     {
       bool currently_bottom = is_bottom();
       make_bottom();
```
martin-cs, line 242: The comment block a few lines down can be removed.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -239,6 +265,29 @@ bool abstract_environmentt::assume(const exprt &expr, const namespacet &ns)
...
+                                      {ID_and, assume_and},
+                                      {ID_or, assume_or},
+                                      {ID_equal, assume_eq},
+                                      {ID_notequal, assume_noteq},
+                                      {ID_le, assume_less_than},
+                                      {ID_lt, assume_less_than},
+                                      {ID_ge, assume_greater_than},
+                                      {ID_gt, assume_greater_than}};
+
+exprt abstract_environmentt::do_assume(const exprt &expr, const namespacet &ns)
```
martin-cs, line 266: Is this intended to only be used on expressions of boolean type?  If so, please can you add a PRECONDITION to document this.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -468,3 +517,210 @@ std::vector<abstract_object_pointert> eval_operands(
...
+           : std::dynamic_pointer_cast<const abstract_value_objectt>(obj);
+}
+
+bool is_value(const abstract_object_pointert &obj)
+{
+  return as_value(obj) != nullptr;
+}
+
+static auto inverse_operations =
+  std::map<dstringt, dstringt>{{ID_equal, ID_notequal},
```
martin-cs: `irep_idt` rather than raw `dstring`.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -468,3 +517,210 @@ std::vector<abstract_object_pointert> eval_operands(
...
+{
+  auto not_expr = to_not_expr(expr);
+
+  auto inverse_expression = invert_expr(not_expr.op());
+  if(inverse_expression.is_not_nil())
+    return env.do_assume(inverse_expression, ns);
+
+  auto result = env.do_assume(not_expr.op(), ns);
+  if(result.is_boolean())
+    result = result.is_true() ? exprt(false_exprt()) : true_exprt();
```
martin-cs: Why the `exprt` constructor on one side?

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -468,3 +517,210 @@ std::vector<abstract_object_pointert> eval_operands(
...
+    return nil_exprt();
+  if(!is_value(left) || !is_value(right))
+    return nil_exprt();
+
+  auto meet = left->meet(right);
+
+  if(meet->is_bottom())
+    return true_exprt();
+
+  return false_exprt();
```
martin-cs: Is this correct?  If left is `{0, 1, 2}` and right is `{1, 2, 3}`, the meet is `{1,2}` which is not enough to conclude anything.  We can only return `false_exprt()` if both left and right are the same, single possible value.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -468,3 +517,210 @@ std::vector<abstract_object_pointert> eval_operands(
...
+  auto left = env.eval(relationship_expr.lhs(), ns);
+  auto right = env.eval(relationship_expr.rhs(), ns);
+
+  if(left->is_top() || right->is_top())
+    return {};
+
+  return {as_value(left), as_value(right)};
+}
+
+exprt assume_less_than(
```
martin-cs, line 767: Is it worth a comment saying this can be ever-so slightly tightened by considering the cases of less-than and less-than-or-equal-to separately.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -468,3 +517,210 @@ std::vector<abstract_object_pointert> eval_operands(
...
+  const exprt &expr,
+  const namespacet &ns)
+{
+  auto operands = eval_operands_as_values(env, expr, ns);
+  if(!operands.are_good())
+    return nil_exprt();
+
+  auto left_upper = operands.left_interval().get_upper();
+  auto right_lower = operands.right_interval().get_lower();
+
```
martin-cs, line 823: Am I right in thinking that this will not resolve the following:

```
int x = nondet();

if (x < 10) {
  assert(x < 20);
}
```

because it does not prune the domain; it just identifies when a branch is infeasible (returning true / false)?

martin-cs (approved):
In principle happy but there are a few details to check.

regression/goto-analyzer/branching-ge/test-always-intervals.desc
jezhiggins: This is a pretty tiny change, but the main reason I haven't is the donkey work of updating a bunch of regression tests :)

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: It's a cast rather than a constructor. 
The ternary operator wants both operands to have the same type, but `result.is_true() ? false_exprt() : true_exprt()` fails because `false_exprt` and `true_exprt` are different types (even though both can be converted to the `exprt`type `result` wants. By forcing the first operand to `exprt`, the second operand decays to `exprt` automatically and it all compiles and runs. 
I've replaced with a `static_cast`.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: No, it will operate on any expression. It's not a boolean expression, it falls back to evaluating it in exactly the same way as before.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: We do consider the less-than and less-then-or-equal separately (also greater-than and greater-than-or-equal) - when we rewrite the expression in simplified form we use the operator ID from the original expression. We don't change the comparison, we just simplify the operands.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: That's right, the assert will be UNKNOWN. We only prune the domain in the case of ==

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: I believe it is, but that doesn't mean it is. :slightly_smiling_face: 

For the equals case, if objects meet we consider that there exists some equality and so we return true. If the meet is empty, they're not equal and we return false. 
From our conversations, I understood the != case to be essentially equivalent. If the object do not meet, they are not equal and we return true. If there's some meet they are not not equal and we return false. 

Reworking this to return `false_exprt()` only when left and right are the same and `nil_exprt` otherwise breaks (or at least changes) existing behaviour in, for example, some of the simplification regression tests. It also causes segfaults in some of the regression tests I wrote around this work. 

The code, as written here, behaves better than the previous code in many cases and no worse than the previous code in every other case. It may not be absolutely correct, but I feel we can still merge this change, gain the benefits, and then revisit as work continues.

regression/goto-analyzer/branching-ge/test-always-intervals.desc
martin-cs: Yep!  That was what was discouraging me too.

src/analyses/variable-sensitivity/abstract_environment.cpp
martin-cs: Ahhhhhh.... yeahhhhh... ugh; the C++ type system and when subtyping relations are taken.  That is an annoying wrinkle. Thank you.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -196,49 +225,51 @@ bool abstract_environmentt::assume(const exprt &expr, const namespacet &ns)
...
-   * 4. eval the current value of the variable
-   * 5. compute the meet (not merge!) of the two abstract_objectt's
-   * 6. assign the new value back to the environment.
-   */
-
   return false;
 }
 
+static auto assume_functions =
+  std::map<dstringt, assume_function>{{ID_not, assume_not},
```
martin-cs: s/dstring/irept/

`dstring` is the implementation but `irep_idt` is the "what irept.id stores" type.  At points there were other options apart from `dstring`.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -196,49 +225,51 @@ bool abstract_environmentt::assume(const exprt &expr, const namespacet &ns)
...
   // This should be enforced by the well-structured-ness of the
   // goto-program and the way assume is used.
-
   PRECONDITION(expr.type().id() == ID_bool);
 
-  // Evaluate the expression
-  abstract_object_pointert res = eval(expr, ns);
-
-  exprt possibly_constant = res->to_constant();
+  auto simplified = simplify_expr(expr, ns);
```
martin-cs, line 230: I wonder if it makes sense to `ai_simplify` this instead?  That way anything that is directly constant should be substituted out and also used in the rewriting based simplification.  For example if we have `x + y + z == 0` and in the domain we know that `y ==0` and `z == 0`, `ai_simplify` should get us `x == 0`, while just `simplify_expr` on it's own won't.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -196,49 +225,51 @@ bool abstract_environmentt::assume(const exprt &expr, const namespacet &ns)
...
+exprt abstract_environmentt::do_assume(const exprt &expr, const namespacet &ns)
+{
+  auto expr_id = expr.id();
+
+  auto fn = assume_functions[expr_id];
+
+  if(fn)
+    return fn(*this, expr, ns);
+
+  return eval(expr, ns)->to_constant();
```
martin-cs, line 275: This is a good fall-back option.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -196,49 +225,51 @@ bool abstract_environmentt::assume(const exprt &expr, const namespacet &ns)
...
+                                      {ID_and, assume_and},
+                                      {ID_or, assume_or},
+                                      {ID_equal, assume_eq},
+                                      {ID_notequal, assume_noteq},
+                                      {ID_le, assume_less_than},
+                                      {ID_lt, assume_less_than},
+                                      {ID_ge, assume_greater_than},
+                                      {ID_gt, assume_greater_than}};
+
+exprt abstract_environmentt::do_assume(const exprt &expr, const namespacet &ns)
```
martin-cs, line 266: Can we document somewhere that returning an `exprt` with `ID_nil` is "give up on this assumption" and that `ID_false` sets the domain to bottom.  It is obvious when reading `assume` but I think it helps in understanding this code.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -468,3 +499,211 @@ std::vector<abstract_object_pointert> eval_operands(
...
+                               {ID_lt, ID_ge},
+                               {ID_ge, ID_lt},
+                               {ID_gt, ID_le}};
+
+exprt invert_expr(const exprt &expr)
+{
+  auto expr_id = expr.id();
+  auto inverse_operation = inverse_operations.find(expr_id);
+  if(inverse_operation == inverse_operations.end())
+    return nil_exprt();
```
martin-cs, line 548: There is a subtle precision loss here; if we have `if (A || B)`  then when we performing the transform for the `GOTO` instructions we will wind up assuming `not_exprt(or_exprt(A, B))` which will fail here rather than assuming `and_exprt(not_exprt(A), not_exprt(B))`.

You can either extend `invert_exprt` and get it to handle logical operations... or (which might be simpler) have an extra boolean flag argument for "we are assuming this to be ..." which starts as true.  `assume_not` will flip this (as will `assume_or` see below) and in the handlers for any atomic condition, you check the flag and if it is false, invert_exprt and recurse.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -468,3 +499,211 @@ std::vector<abstract_object_pointert> eval_operands(
...
+exprt assume_and(
+  abstract_environmentt &env,
+  const exprt &expr,
+  const namespacet &ns)
+{
+  auto and_expr = to_and_expr(expr);
+  for(auto const &operand : and_expr.operands())
+  {
+    auto result = env.do_assume(operand, ns);
+    if(result.is_nil() || result.is_false())
```
martin-cs: If `is_nil()` you can continue with the other operands.  It's only worth returning early if it is `is_false()`.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -468,3 +499,211 @@ std::vector<abstract_object_pointert> eval_operands(
...
+exprt assume_or(
+  abstract_environmentt &env,
+  const exprt &expr,
+  const namespacet &ns)
+{
+  exprt or_result = false_exprt();
+  auto or_expr = to_or_expr(expr);
+  for(auto const &operand : or_expr.operands())
+  {
+    auto result = env.do_assume(operand, ns);
```
martin-cs: It is upsetting that this is not the dual of `assume_and` (and I fear I may not have communicated that very well) but this is not correct in the cases when `do_assume` can prune things (see comments below about pruning on ordering).  Consider the case of assuming `x == 1 || x == 2` as written this will (I think) result in the domain being bottom.

Instead I would have `assume_or(env, or_expr, ns, assuming_true)` call `assume_and(env, and_exprt(or_expr,operand()), ns, ! assume_true)`, i.e. you do the same thing as `assume_and` but inverting the each of the operands.

So rather than assuming `A || B`, you assume `!A && !B`, which means that if you prune the domain while handling `!A`, that is a good thing rather than conflicting with any of the pruning you do in `!B`.

src/analyses/variable-sensitivity/abstract_environment.cpp
martin-cs: There is an asymmetry which happens because the domains are representing /possible/ values for each variable rather than values they definitely can take.

So in the equal cases, we can take the meet of the two domains and get all of the values at which they could both have and be equal.  As you say, if it is empty then there are no ways they can be equal, so we return false and set the domain to bottom.  If there are things in the meet then it is possible they are equal and /we know that they must have a value from the meet/, so we can prune as you do.


The inequality case is a little bit harder and less useful.  If we take the meet of the values and it is empty then there was never any chance they were equal, so we return true because that constraint really is always true.  That much is symmetric.  However if the meet is non-empty (not bottom) then all is says is that it is /possible/ that they could be equal.  It doesn't mean they will be.  In the example I give, if the meet is `{1,2}` then it is possible that `left` is `1` and `right` is `2` or vica versa.  The only cases we need to remove are when they /have/ to be equal.  However that is really hard to represent if you have a set of alternatives for `left` and a separate one for `right` (this is where the non-relational part of the abstraction is an impediment, we don't have a way of representing the relationship between them).  To prune out possibilities we need to check when things are forced to be equal.  This only happens when one side only has one element (i.e. `to_constant` returns something other than `ID_nil`).  If the other side is also one element and the same one (meet is not empty) then they are necessarily equal, so we can return false as assuming they are not equal rules out all possibilities.  ( There is also a pruning case, where one side has one possibility and the other has multiple ones, in which case you can remove that value.  Easy-ish for value-set, possible for intervals if it is the boundary, otherwise not really possible. )

Whether this gets fixed as part of this PR or as a follow up ... either is fine.  I have a mild preference for fixing it now, esp. given the other changes I am requesting to this area.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -468,3 +499,211 @@ std::vector<abstract_object_pointert> eval_operands(
...
+
+  auto left_lower = operands.left_interval().get_lower();
+  auto right_upper = operands.right_interval().get_upper();
+
+  auto reduced_le_expr =
+    binary_relation_exprt(left_lower, expr.id(), right_upper);
+  return env.eval(reduced_le_expr, ns)->to_constant();
+}
+
+exprt assume_greater_than(
```
martin-cs, line 813: One of the symmetries that does work is `(a > b) == (b < a)` so if you want you can implement one of greater and less and simply recurse, flipping the arguments, to the other.

src/analyses/variable-sensitivity/abstract_environment.cpp
martin-cs: It would be really good if we could prune the domain on this case, it really makes a difference for precision and termination.  If there is no overlap then we return true or false (depending on if it is always greater than or never greater than).  If the two overlap (i.e. their meet is non-empty) then you can set the new bounds from the meet.

For example, if `x > y` then you can assign `x` to `[x.upper, y.upper]` and `y` to `[x.lower, y.lower]` (with the off-by-one for `>` vs. `>=`).  It might require me a little thought about how to do this in an abstraction agnostic way.

martin-cs (approved):
Sorry; I've done that awful reviewer thing of responding to your comments and then finding a bunch more things based on that.

If it is more efficient to talk these through, I am happy to do so.

jezhiggins:
> Sorry; I've done that awful reviewer thing of responding to your comments and then finding a bunch more things based on that.

That's ok. I'll work through these, and come back to you if there are any tricky bits :slightly_smiling_face:

src/analyses/variable-sensitivity/abstract_environment.cpp
thomasspriggs: Using `static_cast` isn't a particularly good solution either as it is an unchecked cast. `expr_checked_cast` would not currently work either. It would need its type signature tweaking slightly and `can_cast_expr<exprt>` to be defined. I suggest just using an `if` statement instead of the ternary operator here.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: I should have picked this up with the other dstring change :(

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: :+1:

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: `ai_simplify` isn't available here because that's a member function on `variable_sensitivity_domain`. Luckily, we call `assume` from `variable_sensitivity_domain`, so we can just call `assume` first :slightly_smiling_face: So I've done that.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: I've rewritten `assume_or` in terms of `assume_and`.

src/analyses/variable-sensitivity/abstract_environment.cpp
jezhiggins: Having rewritten `assume_or` in terms of `assume_and`, I think this is taken care of. If we're assuming `not_exprt(or_exprt(A, B))` then that will be transformed to (more or less) `not_expr(not_expr(and_exprt(not_exprt(A), not_exprt(B))))`. It's a little long winded, but I think it will evaluate correctly.

src/analyses/variable-sensitivity/abstract_environment.cpp
martin-cs: Great!

src/analyses/variable-sensitivity/abstract_environment.cpp
martin-cs: Sounds good; thanks.

jezhiggins:
@NlightNFotis @TGWDB I've got four of my regression tests that fail only on the VS with Make build. 

The failing tests all involve pruning a constant value TOP down to an interval where one bound is +/-MAX_INT. Curiously, when using intervals or value sets, the tests pass even though they will be pruned to the same interval. I have a suspicion, therefore, this may be related the typet used to construct the constant. Unfortunately, I don't have access to a Windows machine at the moment so am not able to investigate, but wondered if you had seen anything similar or had any ideas why it might affect this one platform?

Thanks.

regression/goto-analyzer/pruning-lt-nondet/test-constants.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+--variable-sensitivity --vsd-values constants --verify
+^EXIT=0$
+^SIGNAL=0$
+\[main\.assertion\.1\] line .* assertion x < 20: SUCCESS
+
```
NlightNFotis: Hello @jezhiggins, both this and `pruning-lt-nondet-2/test-constants.desc` contain an empty line.

This looks like at least in the windows system, it fails to much as a regex. Can you delete these empty lines and see if it behaves?

I'm surprised how this is not a problem on other platforms though.

jezhiggins: You're a superstar! That's got it sorted, thank you.

regression/goto-analyzer/value-set-function-pointers-simple/main.c
```diff
@@ -1,7 +1,6 @@
 #include <assert.h>
 
 typedef int (*fptr_t)(int);
-fptr_t fun_global, fun_global_show;
```
martin-cs: Would it be possible to add the reduced version of this as another test (can be another `.c` and another `.desc` in the same directory ) rather than just reducing this one?

regression/goto-analyzer/value-set-function-pointers-simple/test.desc
```diff
@@ -1,23 +1,8 @@
...
-^file main.c line 28 function main: replacing function pointer by 2 possible targets$
-^file main.c line 33 function main: replacing function pointer by 2 possible targets$
-^file main.c line 41 function main: replacing function pointer by 2 possible targets$
-^file main.c line 49 function main: replacing function pointer by 2 possible targets$
-^main::1::fun0 \(\) -> TOP
-^main::1::fun1 \(\) -> value-set-begin: ptr ->\(f\) :value-set-end
-^main::1::fun2_show \(\) -> value-set-begin: ptr ->\(f\), ptr ->\(g\) :value-set-end
-^main::1::fun3_show \(\) -> value-set-begin: ptr ->\(f\), ptr ->\(g\) :value-set-end
-^fun_global_show \(\) -> value-set-begin: ptr ->\(f\), ptr ->\(g\) :value-set-end
+--variable-sensitivity --vsd-pointers value-set --show
```
martin-cs: Likewise here.

regression/goto-analyzer/value-set-function-pointers/test.desc
```diff
@@ -0,0 +1,23 @@
+CORE
+main.c
+--variable-sensitivity --vsd-structs every-field --vsd-arrays every-element --vsd-pointers value-set --vsd-values set-of-constants --show --pointer-check --three-way-merge
```
martin-cs: Right - I see what you have done.  For the sake of a cleaner PR it might be good if the reduced one is the new one but ... it's not that big a deal.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -196,49 +225,56 @@ bool abstract_environmentt::assume(const exprt &expr, const namespacet &ns)
...
-    if(possibly_constant.is_false())
+    if(assumption.is_false())
     {
       bool currently_bottom = is_bottom();
       make_bottom();
       return !currently_bottom;
     }
   }
 
-  /* TODO : full implementation here
```
martin-cs, line 221: :tada: !!!

src/analyses/variable-sensitivity/abstract_environment.h
```diff
@@ -96,6 +96,7 @@ class abstract_environmentt
   /// The better the implementation the more precise the results
   /// will be.
   virtual bool assume(const exprt &expr, const namespacet &ns);
+  exprt do_assume(const exprt &e, const namespacet &ns);
```
martin-cs, line 99: Is there any harm in this being public?  I don't think so but I thought I should check.

src/analyses/variable-sensitivity/data_dependency_context.cpp
```diff
@@ -246,38 +246,64 @@ data_dependency_contextt::merge(abstract_object_pointert other) const
...
+    data_dominators.begin(),
+    data_dominators.end(),
+    other->data_dominators.begin(),
+    other->data_dominators.end(),
+    std::inserter(result->data_dominators, result->data_dominators.end()),
+    location_ordert());
+
+  // It is critically important that we only return a newly constructed result
+  // abstract object *iff* the data has actually changed, otherwise AI may
+  // never reach a fixpoint
```
martin-cs, line 299: :heart:

martin-cs (approved):

src/analyses/variable-sensitivity/abstract_environment.h
jezhiggins: It's public so I can exercise it directly in unit tests. (Rather than via `assume` which does some other things too.) However, there's no harm can come of it been public.

regression/goto-analyzer/value-set-function-pointers/test.desc
jezhiggins: Done!

src/analyses/variable-sensitivity/abstract_environment.h
martin-cs: That is fine.

PR #6025, opened 2021-04-12
Supports history variables in function contracts

ArenBabikian:
This PR adds support for history variables in `__CPROVER_ensures` contracts. Specifically, users may now access the pre-function call state of a variable within the post-condition. 

Let us consider the following example. Semantically, this example shows a case where the post-condition of a function call ensures that ( the post-call value of `x` ) is greater than ( the pre-call value of `x` + 2 ).
```
void foo(int *x) __CPROVER_assigns(*x)
  __CPROVER_requires(*x > 0)
  __CPROVER_ensures(*x > __CPROVER_old(*x) + 2)
{
  *x = *x + 2;
}
```
In this case, the variable that we are interested in is `x`, which is modified by the call to `foo(int *x)`. The pre-condition contract (`__CPROVER_requires`) may only access the pre-state of the variable `x` (this is done through the inclusion of `*x`). However, the post-condition (`__CPROVER_ensures`) accesses both the post-state of `x` (through `*x`) and the pre-state of `x` (through `__CPROVER_old(*x)`)

**Note:**
We currently only support pointers within the `__CPROVER_old()` construct. In the near future, we plan to add support for symbols and struct members.

---


regression/contracts/history-pointer-enforce-03/test.desc
```diff
@@ -0,0 +1,13 @@
+CORE
+main.c
+--enforce-all-contracts _ --trace
```
SaswatPadhi: What's the `_` for? Is it a typo?

SaswatPadhi: For consistency, could you please remove `--trace`? It's not used in other `.desc` files in this PR.

regression/contracts/history-pointer-enforce-03/main.c
```diff
@@ -0,0 +1,14 @@
+void foo(int *x) __CPROVER_assigns(*x)
+  __CPROVER_requires(*x > 0 && *x < __INT_MAX__ - 5) __CPROVER_ensures(
```
SaswatPadhi: Nitpick: Could you please move `__CPROVER_ensures` to next line?

regression/contracts/history-pointer-replace-01/test.desc
```diff
@@ -0,0 +1,13 @@
...
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test checks that history variables are supported with the use of the
+--replace-all-calls-with-contracts flag. In this case, the post-condition
+(which contains the history variable) becomes an assumption. We then manually
+assert this assumption. For this test, the assertion should succeed.
```
SaswatPadhi: "assert this assumption" is confusing. We just assume the post condition, right?

May be just remove that sentence?

SaswatPadhi: Also, missing newline at the end of the file.

regression/contracts/history-pointer-replace-02/test.desc
```diff
@@ -0,0 +1,13 @@
...
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Verification:
+This test checks that history variables are supported with the use of the
+--replace-all-calls-with-contracts flag. In this case, the post-condition
+(which contains the history variable) becomes an assumption. We then manually
+assert this assumption. For this test, the assertion should succeed.
```
SaswatPadhi: "assert this assumption" is confusing. We just assume the post condition, right?

May be just remove that sentence?

SaswatPadhi: Also, missing newline at the end of the file.

regression/contracts/history-pointer-replace-02/main.c
```diff
@@ -0,0 +1,17 @@
+#include <assert.h>
+
+void foo(int *x) __CPROVER_assigns(*x) __CPROVER_requires(*x == 0)
+  __CPROVER_ensures(*x > __CPROVER_old(*x) + 2)
```
SaswatPadhi: The contract is incorrect here -- should be `__CPROVER_ensures(*x >= __CPROVER_old(*x) + 2)`.

We don't enforce and only replace here, but better to have the correct contracts in test cases, or add comments regarding contracts that are intentionally incorrect.

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,109 @@ void code_contractst::add_quantified_variable(
...
+    if(op_result.has_value())
+    {
+      op = *op_result;
+      modified = true;
+    }
+  }
+
+  if(expr.id() == ID_old)
+  {
+    DATA_INVARIANT(expr.operands().size() == 1, "old must have one operands");
```
SaswatPadhi: old -> __CPROVER_old

SaswatPadhi: operands -> operand

regression/contracts/history-pointer-replace-03/test.desc
```diff
@@ -0,0 +1,11 @@
...
+main.c
+--replace-all-calls-with-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+--
+Verification:
+This test checks that history variables cannot be used as part of the
+pre-condition contract. In this case, verification should fail.
```
SaswatPadhi: Ideally, this should be a compilation error -- `__CPROVER_old` should not be allowed outside of `__CPROVER_ensures`. This could be done during type checking of `__CPROVER_requires`.

If not, could we generate a helpful error message on verification failure? Also, please add a regex for the expected error message above.

SaswatPadhi: Missing newline at the end of the file.

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,109 @@ void code_contractst::add_quantified_variable(
...
+  exprt expr,
+  std::map<exprt, exprt> &parameter2history,
+  source_locationt location,
+  const irep_idt &function,
+  const irep_idt &mode,
+  std::list<goto_programt::instructiont> &instructions)
+{
+  bool modified = false;
+
+  for(auto &op : expr.operands())
```
SaswatPadhi, line 255: `auto &` -> `const auto &`

src/goto-instrument/code_contracts.cpp
```diff
@@ -868,6 +984,30 @@ void code_contractst::add_contract_check(
...
+      ensures.source_location(),
+      wrapper_fun,
+      function_symbol.mode);
+
+    // add all the history variable initializations
+    std::list<goto_programt::instructiont>::iterator it;
+    for(it = ensures_pair.second.begin(); it != ensures_pair.second.end(); ++it)
+    {
+      check.add(std::move(*it));
+    }
```
SaswatPadhi: See my comment above on `std::pair<goto_programt, goto_programt> ensures_pair;`

Then I would replace this entire loop with `check.destructive_append(ensures_pair.second);`

regression/contracts/history-pointer-enforce-03/main.c
SaswatPadhi: Also, `#include <limits.h>` and use `INT_MAX` instead. I think this might fix the Windows build issues?

regression/contracts/history-pointer-replace-01/main.c
```diff
@@ -0,0 +1,18 @@
...
+void foo(int *x) __CPROVER_assigns(*x) __CPROVER_requires(*x > 0)
+  __CPROVER_ensures(*x == __CPROVER_old(*x) + 2)
+{
+  *x = *x + 2;
+}
+
+int main()
+{
+  int n;
+  __CPROVER_assume(n > 0 && n < __INT_MAX__ - 2);
```
SaswatPadhi: `#include <limits.h>` and use `INT_MAX` instead. I think this might fix the Windows build issues?

regression/contracts/history-pointer-enforce-07/test.desc
```diff
@@ -0,0 +1,13 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=1$
```
SaswatPadhi: Looks like exit code is 64 on Windows. So use `^EXIT=(1|64)$`

src/goto-instrument/code_contracts.cpp
```diff
@@ -366,13 +470,24 @@ bool code_contractst::apply_function_contract(
...
+        ensures.source_location(),
+        function,
+        symbol_table.lookup_ref(function).mode);
+
+    // add all the history variable initializations
+    std::list<goto_programt::instructiont>::iterator it;
+    for(it = ensures_pair.second.begin(); it != ensures_pair.second.end(); ++it)
+    {
+      goto_program.insert_before(target, std::move(*it));
+    }
```
SaswatPadhi: See my comment above on `std::pair<goto_programt, goto_programt> ensures_pair`

Then I would replace this entire loop and the next 2 lines with:
```c
    auto lines_to_iterate = ensures_pair.first.instructions.size() + ensures_pair.second.instructions.size();
    goto_program.insert_before_swap(target, ensures_pair.second);
    goto_program.insert_before_swap(target, ensures_pair.first);
```

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,109 @@ void code_contractst::add_quantified_variable(
...
+code_contractst::create_ensures_instruction(
+  codet expression,
+  source_locationt location,
+  const irep_idt &function,
+  const irep_idt &mode)
+{
+  std::map<exprt, exprt> parameter2history;
+  std::list<goto_programt::instructiont> instructions;
+
+  // Find and replace "old" expression in the above instruction
```
SaswatPadhi: "... in the ~~above instruction~~ expression"?

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,109 @@ void code_contractst::add_quantified_variable(
...
+  }
+  else if(modified)
+    return std::move(expr);
+  else
+    return {};
+}
+
+std::pair<goto_programt, std::list<goto_programt::instructiont>>
+code_contractst::create_ensures_instruction(
+  codet expression,
```
SaswatPadhi: Can this be `const codet &expression`?

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,109 @@ void code_contractst::add_quantified_variable(
...
+    // TODO: generalize below
+    PRECONDITION(parameter.id() == ID_dereference);
+
+    exprt ret;
+    if(parameter.id() == ID_dereference)
+    {
+      const auto &dereference_expr = to_dereference_expr(parameter);
+      const auto &cur_pointer = dereference_expr.pointer();
+
+      std::map<exprt, exprt>::iterator it = parameter2history.find(cur_pointer);
```
SaswatPadhi: `auto it`

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,109 @@ void code_contractst::add_quantified_variable(
...
+        // 3.2 Add an assignment such that the value pointed to by the new
+        // temporary variable is equal to the value of the corresponding
+        // parameter
+        instructions.push_back(goto_programt::make_assignment(
+          dereference_exprt(tmp_symbol),
+          dereference_exprt(cur_pointer),
+          location));
+      }
+
+      ret = dereference_exprt(parameter2history[cur_pointer]);
```
SaswatPadhi: Why not store the `dereference_exprt` in the map (after declaring it) in the first place, so we don't have to construct the `dereference_exprt` every time?

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,109 @@ void code_contractst::add_quantified_variable(
   }
 }
 
+optionalt<exprt> code_contractst::replace_old_parameter(
+  exprt expr,
+  std::map<exprt, exprt> &parameter2history,
+  source_locationt location,
+  const irep_idt &function,
+  const irep_idt &mode,
+  std::list<goto_programt::instructiont> &instructions)
```
SaswatPadhi: The last argument could be `goto_programt &history`.

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,109 @@ void code_contractst::add_quantified_variable(
...
+          new_tmp_symbol(cur_pointer.type(), location, function, mode)
+            .symbol_expr();
+
+        // 2. Associate the above temporary variable to it's corresponding
+        // expression
+        parameter2history[cur_pointer] = tmp_symbol;
+
+        // 3. Add the required instructions to the instructions list
+        // 3.1 Declare the newly created temporary variable
+        instructions.push_back(goto_programt::make_decl(tmp_symbol, location));
```
SaswatPadhi: This can be `history.add(goto_programt::make_decl(...))`.

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,109 @@ void code_contractst::add_quantified_variable(
...
+
+        // 3.2 Add an assignment such that the value pointed to by the new
+        // temporary variable is equal to the value of the corresponding
+        // parameter
+        instructions.push_back(goto_programt::make_assignment(
```
SaswatPadhi: Similarly `history.add(goto_programt::make_assignment(...))`.

SaswatPadhi (changes requested):
The PR overall looks good. I list several comments regarding simplifying the code and clarifying the regression tests.

One meta comment about regression tests: you only check for exit code and verification success and failure. Could you please also add some regexes in the `test.desc` files to check if the contracts are actually being processed? They must be generating some additional assertions, for which we should have regexes.

regression/contracts/history-pointer-enforce-03/test.desc
ArenBabikian: Oh, this is a mistake. I will fix it.

regression/contracts/history-pointer-enforce-07/test.desc
ArenBabikian: Thanks!

src/goto-instrument/code_contracts.cpp
ArenBabikian: I cannot do this because there is a path where I am modifying `op` within the loop.
I might be incorrect.

SaswatPadhi: Ah yes, you're doing `op = *op_result;` in the loop.

src/goto-instrument/code_contracts.cpp
ArenBabikian: I was able to change it to `codet &expression`. However I did make some slight modifications inside the function and at the call sites. But overally, I believeI was able to provide a certain improvement.

SaswatPadhi: Nice.

I can review the change when after you push it. I'm curious why `const codet &` wouldn't work:
- if you're modifying `expression` within this function, then `codet expression` would have been incorrect in the first place, because that was a copy
- if you're not modifying `expression` within this function, then `const codet &` should work fine

regression/contracts/history-pointer-enforce-03/main.c
ArenBabikian: I have fixed the `INT_MAX` issue. However, regarding the placement of `__CPROVER_ensures`, the current location is what clang is proposing. I will try what you suggested, but it might not work.

SaswatPadhi: Oh I see. Yeah `clang-format` goes crazy on contract annotations sometimes.

I usually surround contracts with `/* clang-format off */` and `/* clang-format on */`.

regression/contracts/history-pointer-replace-03/test.desc
ArenBabikian: In my upcoming commit, there will be a compilation error if there is a `__CPROVER_old` expressions in a `__CPROVER_requires` clause.

regression/contracts/history-pointer-enforce-03/main.c
ArenBabikian: This positioning of `_CPROVER_ensures` is what clang is proposing. I will try to move it ot the next line, but I might be unable to do this.

src/ansi-c/c_typecheck_base.cpp
```diff
@@ -729,6 +729,7 @@ void c_typecheck_baset::typecheck_declaration(
         if(as_const(code_type).requires().is_not_nil())
         {
           auto &requires = code_type.requires();
+          typecheck_expr_requires(requires);
```
SaswatPadhi: The function call right below this line is `typecheck_expr(requires);`, so this is confusing.

Could you please rename the new function to `disallow_history_variables(requires)`? That would resolve the confusion and the name of your function would also match exactly what it's doing -- which isn't type checking.

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,106 @@ void code_contractst::add_quantified_variable(
   }
 }
 
+const optionalt<exprt> code_contractst::replace_old_parameter(
+  exprt expr,
```
SaswatPadhi: Could you make this `exprt &expr` to avoid a copy. I see that it can't be `const exprt&` since you're modifying it.

Also, this pattern looks bad:
- you pass `expression` to this function, which could be modified within the function
- then you return an `optional<exprt>` from this function
- you replace `expression` if the `optional<exprt>` isn't nil

Why not just pass `exprt &expr` as first argument here and return `void`?
This call that you do later:

```c
  // Find and replace "old" expression in the "expression" variable
  const auto expr_w_replacement = replace_old_parameter(
    expression, parameter2history, location, function, mode, history);
  if(expr_w_replacement.has_value())
  {
    expression = to_code(*expr_w_replacement);
  }
```

could just be one line: `replace_old_parameter(expression, parameter2history, location, function, mode, history);` -- `expression` would be passed by reference, so would be modified within `replace_old_parameter` as needed.

SaswatPadhi (changes requested):
The regexes you added to `test.desc` files look good and now we're checking that contracts are actually working as expected.

I have only 2 comments:

src/goto-instrument/code_contracts.cpp
```diff
@@ -244,6 +245,89 @@ void code_contractst::add_quantified_variable(
...
+  {
+    DATA_INVARIANT(
+      expr.operands().size() == 1, CPROVER_PREFIX "old must have one operand");
+
+    const auto &parameter = to_old_expr(expr).variable();
+
+    // TODO: generalize below
+    PRECONDITION(parameter.id() == ID_dereference);
+
+    exprt ret;
```
SaswatPadhi: Unused, I think

src/util/pointer_expr.h
```diff
@@ -462,6 +462,30 @@ class null_pointer_exprt : public constant_exprt
   }
 };
 
+/// \brief A base class for a predicate that indicates the pre-function-call
```
SaswatPadhi: a predicate -> an expression

`old_exprt` is not a predicate.

src/util/pointer_expr.h
```diff
@@ -462,6 +462,30 @@ class null_pointer_exprt : public constant_exprt
...
+/// value of a variable passed as a parameter to a function
```
SaswatPadhi: a variable -> an expression

The argument can be a deref, for example.

src/util/pointer_expr.h
```diff
@@ -462,6 +462,30 @@ class null_pointer_exprt : public constant_exprt
...
+class old_exprt : public unary_exprt
+{
+public:
+  explicit old_exprt(irep_idt id, exprt variable)
+    : unary_exprt(id, std::move(variable))
+  {
+  }
+
+  const exprt &variable() const
```
SaswatPadhi: May be rename this to `parameter()` since it need not be just a variable? Can be a deref, for example.

src/util/pointer_expr.h
SaswatPadhi: I think this class shouldn't be in `src/util/pointer_expr.h`. Perhaps we should make a new file in this directory: `ghost_variables.h` that would contain all of the `exprt`s for history variables and other ghost variables.

What do you think @feliperodri?

feliperodri: Agreed.

SaswatPadhi (approved):
LGTM! Thanks for all the changes.

src/goto-instrument/code_contracts.cpp
```diff
@@ -22,11 +22,13 @@ Date: February 2016
 
 #include <linking/static_lifetime_init.h>
 
+#include <map>
```
feliperodri: This should go below `<algorithm>`.

feliperodri (approved):
LGTM.

feliperodri:
Hi @kroening @tautschnig could you review this PR? It's missing code owner approval.

kroening:
While it's conceivable that other languages may eventually offer a similar mechanism, may I suggest to add `old_exprt` to `ansi-c/c_expr.h` for now.

ArenBabikian:
@kroening Thanks for the comment, I have made the required changes.

src/ansi-c/c_expr.h
```diff
@@ -200,4 +200,28 @@ inline side_effect_expr_overflowt &to_side_effect_expr_overflow(exprt &expr)
   return static_cast<side_effect_expr_overflowt &>(side_effect_expr);
 }
 
+/// \brief A base class for an expression that indicates the pre-function-call
+/// value of an expression passed as a parameter to a function
+class old_exprt : public unary_exprt
+{
+public:
+  explicit old_exprt(irep_idt id, exprt variable)
+    : unary_exprt(id, std::move(variable))
```
kroening: I would fix the ID of `old_exprt` to `ID_old` -- I don't see an extension coming whereby you need several different kinds of "old expressions".

kroening: Likewise, this isn't a "base class".

src/ansi-c/c_expr.h
```diff
@@ -200,4 +200,28 @@ inline side_effect_expr_overflowt &to_side_effect_expr_overflow(exprt &expr)
...
+  {
+  }
+
+  const exprt &parameter() const
+  {
```
kroening, line 213: "Parameter" is wrong terminology here. Given that you don't support symbols yet, the expression that's returned is actually never a parameter. Call it "expression".

src/ansi-c/c_typecheck_base.h
```diff
@@ -205,6 +205,7 @@ class c_typecheck_baset:
     const symbol_exprt &function_symbol);
   virtual exprt
   typecheck_shuffle_vector(const side_effect_expr_function_callt &expr);
+  virtual void disallow_history_variables(exprt &expr);
```
kroening: This needs to be
`void disallow_history_variables(const exprt &) const;`
No need for this to be virtual, the parameter must be a const-ref, and the method itself is const.

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -2576,6 +2576,20 @@ exprt c_typecheck_baset::do_special_functions(
 
     return std::move(ok_expr);
   }
+  else if(identifier == CPROVER_PREFIX "old")
+  {
+    if(expr.arguments().size() != 1)
+    {
+      error().source_location = f_op.source_location();
+      error() << identifier << " expects one operands" << eom;
+      throw 0;
```
kroening, line 2630: operands -> operand

src/ansi-c/c_typecheck_base.cpp
```diff
@@ -727,6 +727,7 @@ void c_typecheck_baset::typecheck_declaration(
         if(as_const(code_type).requires().is_not_nil())
         {
           auto &requires = code_type.requires();
+          disallow_history_variables(requires);
           typecheck_expr(requires);
           implicit_typecast_bool(requires);
```
kroening, line 731: How about doing that call to `disallow_history_variables` after the typechecking, which would simplify it significantly.

src/ansi-c/library/cprover.h
```diff
@@ -35,6 +35,7 @@ __CPROVER_size_t __CPROVER_zero_string_length(const void *);
 __CPROVER_size_t __CPROVER_buffer_size(const void *);
 __CPROVER_bool __CPROVER_r_ok(const void *, __CPROVER_size_t);
 __CPROVER_bool __CPROVER_w_ok(const void *, __CPROVER_size_t);
+void __CPROVER_old(const void *);
```
kroening, line 40: I would make that
`void __CPROVER_old();`
since you want to support arbitrary expressions eventually.

src/goto-instrument/code_contracts.cpp
```diff
@@ -241,6 +244,88 @@ void code_contractst::add_quantified_variable(
...
+  if(expr.id() == ID_old)
+  {
+    DATA_INVARIANT(
+      expr.operands().size() == 1, CPROVER_PREFIX "old must have one operand");
+
+    const auto &parameter = to_old_expr(expr).parameter();
+
+    // TODO: generalize below
+    PRECONDITION(parameter.id() == ID_dereference);
+
```
kroening: `PRECONDITION` isn't meant for restrictions on user-input -- the idea is that these are provable.
Throw an exception whenever user input doesn't satisfy the restrictions.

kroening (approved):

chrisr-diffblue (approved):

PR #6026, opened 2021-04-13
Fix havocing of arrays when enforcing invariants

### LINKED ISSUE #6020: Soundness bug with array havocing in loop contracts
**CBMC version:** 99f1a2e

**Operating system:** 10.15.7

**Test case:**

```c
#include <assert.h>
#include <stdlib.h>

#define SIZE 8

void main()
{
  char *data = malloc(SIZE);
  data[5] = 0;

  for (unsigned i = 0; i < SIZE; i++)
    __CPROVER_loop_invariant(i <= SIZE)
  {
    data[i] = 1;
  }

  assert(data[5] == 0);
}
```

```console
$ goto-cc test.c -o test.1.gb
$ goto-instrument --enforce-all-contracts test.1.gb test.2.gb
$ cbmc test.2.gb
```

Clearly the verification should fail because we are initializing all elements of `data` to `1`.

**What happened instead:**

CBMC reports that verification was successful!

Running `cbmc test.c` unwinds the loop and ignored the loop invariant contract. This results in a verification failure, as expected.

**Additional Information:**

I think, the bug is in the `build_havoc_code` routine:
https://github.com/diffblue/cbmc/blob/99f1a2e0d3fb3171a21e5e2c811ea8c7ebc26182/src/goto-instrument/loop_utils.cpp#L38-L56

It simply havocs a pointer without considering its size. So for arrays, it only havocs the first element.

SaswatPadhi:
Fixes #6020.

As reported in #6020, only the first element of an array was being havoced earlier. In this change, we fix this behavior using `havoc_object`.

### DIFF
--- a/regression/contracts/invar_havoc_dynamic_multi-dim_array_partial_const_idx/test.desc
+++ b/regression/contracts/invar_havoc_dynamic_multi-dim_array_partial_const_idx/test.desc
@@ -1,0 +1,22 @@
+KNOWNBUG
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
+^\[main.1\] .* Check loop invariant before entry: SUCCESS$
+^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
+^\[main.assertion.1\] .* assertion data\[1\]\[2\]\[3\] == 0: FAILURE$
+^\[main.assertion.2\] .* assertion data\[4\]\[5\]\[6\] == 0: FAILURE$
+^VERIFICATION FAILED$
+--
+--
+Test case related to issue #6020: it checks that dynamically allocated multi
+dimensional arrays are correctly havoced when enforcing invariant contracts.
+
+Here, both assertions are expected to fail -- the entire `data` array should
+be havoced since we are assigning to a non-constant index.
+We _could_ do a more precise analysis in future where we only havoc
+`data[4][5][6]` but not `data[1][2][3]` since the latter clearly cannot be
+modified in the given loop.
+
+Blocked on issue #6021 -- alias analysis is imprecise and returns `unknown`.
--- a/src/goto-instrument/loop_utils.cpp
+++ b/src/goto-instrument/loop_utils.cpp
@@ -11,11 +11,34 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "loop_utils.h"
 
+#include <util/expr_util.h>
 #include <util/pointer_expr.h>
 #include <util/std_expr.h>
 
+#include <goto-programs/pointer_arithmetic.h>
+
 #include <analyses/natural_loops.h>
 #include <analyses/local_may_alias.h>
+
+class loop_utils_is_constantt : public is_constantt
+{
+public:
+  explicit loop_utils_is_constantt(const modifiest &mod) : modifies(mod)
+  {
+  }
+  bool is_constant(const exprt &expr) const override
+  {
+    // Besides the "usual" constants (checked in is_constantt::is_constant),
+    // we also treat unmodified symbols as constants
+    if(expr.id() == ID_symbol && modifies.find(expr) == modifies.end())
+      return true;
+
+    return is_constantt::is_constant(expr);
+  }
+
+protected:
+  const modifiest &modifies;
+};
 
 goto_programt::targett get_loop_exit(const loopt &loop)
 {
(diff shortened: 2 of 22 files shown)


tautschnig:
@SaswatPadhi Please update `scripts/expected_doxygen_warnings.txt` to silence the `doxygen` warning.

regression/contracts/invar_havoc_dynamic_multi-dim_array_partial_const_idx/test.desc
```diff
@@ -0,0 +1,21 @@
+KNOWNBUG
```
tautschnig, line 1: I'd prefer to have all the `KNOWNBUG` tests in a separate commit, especially as they (also) relate to a different issue.

src/goto-instrument/loop_utils.cpp
```diff
@@ -35,23 +58,51 @@ goto_programt::targett get_loop_exit(const loopt &loop)
...
 
+void build_havoc_code_for_scalar(
+  const goto_programt::targett loop_head,
+  const exprt &lhs,
+  goto_programt &dest)
+{
+  side_effect_expr_nondett rhs(lhs.type(), loop_head->source_location);
+
+  goto_programt::targett t = dest.add(goto_programt::make_assignment(
+    code_assignt(std::move(lhs), std::move(rhs)), loop_head->source_location));
```
tautschnig: `lhs` is `const &`, the `std::move` will not add value. Also, you can nowadays use `make_assignment(lhs, std::move(rhs), loop_head->source_location)` (avoiding the `code_assignt`).

src/goto-instrument/loop_utils.cpp
```diff
@@ -35,23 +58,51 @@ goto_programt::targett get_loop_exit(const loopt &loop)
...
+  const exprt &lhs,
+  goto_programt &dest)
+{
+  codet havoc(ID_havoc_object);
+  havoc.add_source_location() = loop_head->source_location;
+  havoc.add_to_operands(std::move(lhs));
+
+  goto_programt::targett t =
+    dest.add(goto_programt::make_other(havoc, loop_head->source_location));
+  t->code_nonconst().add_source_location() = loop_head->source_location;
```
tautschnig: This is redundant, line 79 already sets that source location.

src/goto-instrument/loop_utils.cpp
```diff
@@ -35,23 +58,51 @@ goto_programt::targett get_loop_exit(const loopt &loop)
...
-      m_it!=modifies.end();
-      m_it++)
+  loop_utils_is_constantt is_constant(modifies);
+  for(const auto &lhs : modifies)
   {
-    exprt lhs=*m_it;
-    side_effect_expr_nondett rhs(lhs.type(), loop_head->source_location);
+    if(lhs.id() == ID_index || lhs.id() == ID_dereference)
+    {
+      address_of_exprt address_of_lhs(std::move(lhs));
```
tautschnig: `lhs` is `const &`, no need for `std::move`.

src/goto-instrument/loop_utils.cpp
```diff
@@ -63,17 +114,23 @@ void get_modifies_lhs(
 {
   if(lhs.id()==ID_symbol)
     modifies.insert(lhs);
-  else if(lhs.id()==ID_dereference)
+  else if(lhs.id() == ID_dereference)
```
tautschnig: Nit pick: I'd leave this line unchanged.

src/goto-instrument/loop_utils.cpp
```diff
@@ -63,17 +114,23 @@ void get_modifies_lhs(
...
+        modifies.insert(dereference_exprt{mod});
+      else
+        modifies.insert(dereference_exprt{plus_exprt{mod, ptr.offset}});
+    }
   }
   else if(lhs.id()==ID_member)
   {
   }
-  else if(lhs.id()==ID_index)
+  else if(lhs.id() == ID_index)
```
tautschnig: Nit pick: I'd leave this line unchanged.

tautschnig (approved):

src/goto-instrument/loop_utils.cpp
SaswatPadhi: Thanks! Fixed.

regression/contracts/invar_havoc_dynamic_multi-dim_array_partial_const_idx/test.desc
SaswatPadhi: Good point! Done.

TGWDB (approved):

src/goto-instrument/loop_utils.cpp
```diff
@@ -11,12 +11,35 @@ Author: Daniel Kroening, kroening@kroening.com
...
 #include <analyses/natural_loops.h>
 #include <analyses/local_may_alias.h>
 
+class loop_utils_is_constantt : public is_constantt
+{
+public:
+  explicit loop_utils_is_constantt(const modifiest &mod) : modifies(mod)
+  {
+  }
+  bool is_constant(const exprt &expr) const override
```
chrisr-diffblue, line 29: Minor query - I wonder if this would be better with a more explicit name, such as `is_constant_or_unmodified` ?

chrisr-diffblue (approved):
Looks good, and happy for it to go in - just a minor query about naming, which you can feel free to ignore if it gets too complicated.

src/goto-instrument/loop_utils.cpp
SaswatPadhi: I agree, that would be a better name. But the `operator()` defined on the base class (`is_constantt`) calls `is_constant` and so I chose to override the `is_constant` function.

chrisr-diffblue: Yeah, I assumed it was a "more hassle than its worth" choice :-)

Issue #6028, opened 2021-04-13
Still SIMD1 failures with ia32_vec_ext_v4si


rurban:
When I compile on AMD64 with `-march=native`
the SIMD1 test fails still after #5919 with
```
no body for function __builtin_ia32_vec_ext_v4si
main.c function main
[main.assertion.1] line 12 assertion val1 == 0x1234: FAILURE
```

CBMC version: cbmc-5.27.0-16-gd79e5f0b6
Operating system: Fedora Core 32
Exact command line resulting in the issue: make test
What behaviour did you expect:

[LastTest-marchnative.log](https://github.com/diffblue/cbmc/files/6302924/LastTest-marchnative.log)

See also my march-native branch with the `-Denable_marchnative=true` option to make it easier. https://github.com/rurban/cbmc/commits/march-native

tautschnig:
Won't be today, but I'll take a look at this.

tautschnig:
@rurban In the meantime, could you perhaps post the upload of running CBMC on this test with `--preprocess`? Also, could you post the full output, not just those three lines? I'm really surprised that `__builtin_ia32_vec_ext_v4si` is not found, it should be provided by the internal library. Thank you!

rurban:
See the https://github.com/diffblue/cbmc/files/6302924/LastTest-marchnative.log link, already there. Or more?
Could be Fedora specific

tautschnig:
> See the https://github.com/diffblue/cbmc/files/6302924/LastTest-marchnative.log link, already there. Or more?
> Could be Fedora specific

Ah, thank you, I thought this was a log of a successful run. (The `Malloc13` failure is also surprising.) The output of running just that test with `--preprocess` would still be useful.

tautschnig:
I got hold of a Fedora 33 system, but I'm unable to reproduce. Here is what I did:

```
$ $ cat /proc/cpuinfo
processor	: 0
vendor_id	: AuthenticAMD
cpu family	: 23
model		: 49
model name	: AMD EPYC 7R32
[...]
$ g++ --version
g++ (GCC) 10.2.1 20201125 (Red Hat 10.2.1-9)
[...]
$ cmake -Denable_marchnative=true -Bbuild
[...]
$ cmake --build build -- -j cbmc
[...]
```

The only test I'm getting to fail is `integer-assignments1`.

tautschnig:
> The only test I'm getting to fail is `integer-assignments1`.

Fixed in #6030.

[...]

PR #6029, opened 2021-04-13
Add -Denable_marchnative=true option

rurban:
Valid for gcc-compat on Intel. Not valid on arm or other archs.
    
And pass CFLAGS and CXXFLAGS to cudd configure.
Don't use that on releases, just privately or on known CPU HW.
Added a doc section Maximum Performance in compilation-and-development
    
10% faster than -O3.
make test cbmc-CORE went from 84.02.sec with -O3 to 75.92 sec with -O2 -march=native

    test regression: SIMD1
    no body for function __builtin_ia32_vec_ext_v4si
    main.c function main
    [main.assertion.1] line 12 assertion val1 == 0x1234: FAILURE

See #6028 


TGWDB (changes requested):
This PR is a mixture of at least four different things.

1. Adding the titular option related to native architecture optimisations. This mostly looks ok although I'd like to see some more details on the data/improvement on more than one run on one system.
2. There are changes to flag passing (for CaDiCaL) which seems like a bugfix? Can this also be checked to see if this is only `cmake`, or also a `make` issue?
3. Improvements to output during cmake build.
4. A modification related to location and changes to `version.cpp` that is not clear needs to be fixed. Can this be explained/justified?

There may also be some changes to formatting that are cosmetic and unrelated?

I would like these to at least be clearly explained and separated, if not split into different PRs.

rurban:
> Adding the titular option related to native architecture optimisations. This mostly looks ok although I'd like to see some more details on the data/improvement on more than one run on one system.

I did several runs, compared to -O3. Before I only benchmarked private code, now I've benchmarked against cbmc-CORE.

> There are changes to flag passing (for CaDiCaL) which seems like a bugfix? Can this also be checked to see if this is only cmake, or also a make issue?

We needed to pass extra compiler options to all extra compiled backends, such as cudd or CaDiCaL. cudd already has -O3, but for -march=native we'd need to add these lines. More so for -flto later. You won't get any SIMD vectorization without.

> Improvements to output during cmake build.

Only for this feature.

>  A modification related to location and changes to version.cpp that is not clear needs to be fixed. Can this be explained/justified?

Sorry, I've split the version fix into a seperate branch. I forgot if I already submitted that as extra PR already, if not will do. There was some denial, that it is not broken.

doc/architectural/compilation-and-development.md
```diff
@@ -302,3 +302,22 @@ each function; and the "call graph", which includes context, and tells you how
...
 https://ftp.gnu.org/old-gnu/Manuals/gprof-2.9.1/html_chapter/gprof_5.html
+
+\section compilation-and-development-section-max-performance Maximum Performance
+
+On Intel with gcc-compatible compilers (i.e. gcc, clang, icc) you may
+use the `-Denable_marchnative=1` option, to pass `-march=native` to cc
+and c++.
+
+    CXXFLAGS="-O3" CFLAGS="-O3" cmake -S . -Bbuild-native -Denable_marchnative=1
+    make -j4 -C build-native
```
TGWDB: I think this would be clearer if it only showed the new and important flags, we don't need to show the specification of the `-O3` flag or `-j4`. So something elike:
```
cmake -S . -Bbuild-native -Denable_marchnative=1
make -C build-native
```

src/solvers/CMakeLists.txt
```diff
@@ -109,7 +112,7 @@ elseif("${sat_impl}" STREQUAL "cadical")
     download_project(PROJ cadical
         URL https://github.com/arminbiere/cadical/archive/rel-1.4.1.tar.gz
         PATCH_COMMAND true
-        COMMAND CXX=${CMAKE_CXX_COMPILER} ./configure -O3 -s CXXFLAGS=-std=c++14
+        COMMAND CXX=${CMAKE_CXX_COMPILER} ./configure -O3 -s CXXFLAGS=-std=c++14 -j
```
tautschnig: As much as a parallel build would be nice: this doesn't seem to work.

src/memory-analyzer/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ target_link_libraries(memory-analyzer-lib
 
 # Executable
 add_executable(memory-analyzer memory_analyzer_main.cpp)
+
```
tautschnig: Should there be an actual change here?

src/ansi-c/CMakeLists.txt
```diff
@@ -10,6 +10,10 @@ add_custom_command(OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/cprover_library.inc"
     DEPENDS converter ${ansi_c_library_sources})
 
 add_executable(file_converter file_converter.cpp)
+if(ipo_supported)
+    set_property(TARGET converter PROPERTY INTERPROCEDURAL_OPTIMIZATION True)
+    set_property(TARGET file_converter PROPERTY INTERPROCEDURAL_OPTIMIZATION True)
+endif()
```
tautschnig, line 16: Fair enough for consistency, but performance won't matter that much for these, they are build-time only tools.

scripts/glucose_CMakeLists.txt
```diff
@@ -14,6 +14,10 @@ set_target_properties(
     XCODE_ATTRIBUTE_CODE_SIGN_IDENTITY "Developer ID Application: Daniel Kroening"
 )
 
+if(ipo_supported)
+    set_property(TARGET minisat2-condensed PROPERTY INTERPROCEDURAL_OPTIMIZATION True)
```
tautschnig: Copy&paste error: Should be "glucose-condensed"

src/solvers/CMakeLists.txt
rurban: yes, fixed

src/memory-analyzer/CMakeLists.txt
rurban: sorry, removed

scripts/glucose_CMakeLists.txt
rurban: oops, fixed

CMakeLists.txt
```diff
@@ -204,6 +208,36 @@ endif()
 option(WITH_MEMORY_ANALYZER
   "build the memory analyzer" ${WITH_MEMORY_ANALYZER_DEFAULT})
 
+if((CMAKE_MAJOR_VERSION EQUAL 3 AND CMAKE_MINOR_VERSION GREATER_EQUAL 9)
```
tautschnig: Shouldn't this be `CMAKE_MAJOR_VERSION GREATER_EQUAL 4 OR CMAKE_MINOR_VERSION GREATER_EQUAL 9` for our minimum requirement (by now) is 3.8 anyway?

src/solvers/CMakeLists.txt
```diff
@@ -122,6 +125,9 @@ elseif("${sat_impl}" STREQUAL "cadical")
 
     add_library(cadical STATIC IMPORTED)
 
+    if(ipo_supported)
+      set_property(TARGET cadical PROPERTY INTERPROCEDURAL_OPTIMIZATION True)
+    endif()
```
tautschnig: How effective is this going to be? I am asking for `-march=native` will (AFAIK) not be forwarded to the cadical build, but perhaps we need to find a way of doing so?

CMakeLists.txt
rurban: true, fixed

src/solvers/CMakeLists.txt
rurban: I thought that subprojects inherit the cache, so it would inherit -march=native and ipo_supported. I'll check, this was years ago

tautschnig: The cadical build is done at configure time (https://github.com/diffblue/cbmc/blob/develop/src/solvers/CMakeLists.txt#L117), I don't think this is a subproject? (Unlike the minisat/glucose builds.)

tautschnig:
It seems we need a compilation rule that adds `-Dyyalloc=<parser-specific-prefix>_alloc` to all build rules turning `*.tab.cpp` files into objects. (It seems that `bison` source hard-codes the name `yyalloc` for the `union yyalloc` type: https://sources.debian.org/src/bison/2%3A3.8.2%2Bdfsg-1/data/skeletons/yacc.c/#L590)

rurban:
union yyalloc: Yes, that's where I got stuck with -flto

tautschnig:
> union yyalloc: Yes, that's where I got stuck with -flto

Your cmake foo is probably way better than mine, but I'd expect something like

`set_source_files_properties(ansi_c_y.tab.cpp PROPERTIES COMPILE_FLAGS -Dyyalloc=ansi_c_yyalloc)`

Perhaps this can even be done within the `generic_bison` macro?

tautschnig:
@rurban The following commit fixes the linking problem with LTO:
````patch
commit e8818876128e86caa84caf5f108629a6d5b6e640
Author: Michael Tautschnig <tautschn@amazon.com>
Date:   Wed May 31 11:50:49 2023 +0000

    Rename yyalloc to avoid duplicate definitions with LTO

    This is for CMake only as we haven't currently got an LTO set-up for
    Makefile builds.

diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 51bfee0e30..1a38a61705 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -25,6 +25,7 @@ macro(generic_bison name)
     endif()

     set(bison_source "${name}_y.tab.cpp")
+    set_source_files_properties(${bison_source} PROPERTIES COMPILE_FLAGS -Dyyalloc=${name}_yyalloc)
     if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${bison_source}")
         message(FATAL_ERROR "Generated file ${bison_source} found in source tree. If you previously built with `make`, run `make clean` and try again")
     endif()
````

**Comment by rurban** [comment]
Good, but then we still get those warnings:
```
src/json/json_y.tab.cpp:186: warning: type ‘yysymbol_kind_t’ violates the C++ One Definition Rule [-Wodr]
  186 | enum yysymbol_kind_t
      | 
src/ansi-c/ansi_c_y.tab.cpp:157: note: an enum with different value name is defined in another translation unit
  157 | enum yysymbol_kind_t
      | 
src/json/json_y.tab.cpp:192: note: name ‘YYSYMBOL_TOK_STRING’ differs from name ‘YYSYMBOL_TOK_AUTO’ defined in another translation unit
  192 |   YYSYMBOL_TOK_STRING = 3,                 /* TOK_STRING  */
      | 
src/ansi-c/ansi_c_y.tab.cpp:163: note: mismatching definition
  163 |   YYSYMBOL_TOK_AUTO = 3,                   /* "auto"  */
```

**Comment by tautschnig** [comment]
> Good, but then we still get those warnings:
> 
> ```
> src/json/json_y.tab.cpp:186: warning: type ‘yysymbol_kind_t’ violates the C++ One Definition Rule [-Wodr]
>   186 | enum yysymbol_kind_t
>       | 
> src/ansi-c/ansi_c_y.tab.cpp:157: note: an enum with different value name is defined in another translation unit
>   157 | enum yysymbol_kind_t
>       | 
> src/json/json_y.tab.cpp:192: note: name ‘YYSYMBOL_TOK_STRING’ differs from name ‘YYSYMBOL_TOK_AUTO’ defined in another translation unit
>   192 |   YYSYMBOL_TOK_STRING = 3,                 /* TOK_STRING  */
>       | 
> src/ansi-c/ansi_c_y.tab.cpp:163: note: mismatching definition
>   163 |   YYSYMBOL_TOK_AUTO = 3,                   /* "auto"  */
> ```

Hmm, I hadn’t seen that in my local testing. Maybe my proposed patch just needs to be extended to have more defines?

File: src/CMakeLists.txt [hunk 8]
```diff
@@ -25,6 +25,7 @@ macro(generic_bison name)
     endif()
 
     set(bison_source "${name}_y.tab.cpp")
+    set_source_files_properties(${bison_source} PROPERTIES COMPILE_FLAGS -Dyyalloc=${name}_yyalloc)
```
tautschnig:
```suggestion
    set_source_files_properties(${bison_source} PROPERTIES COMPILE_FLAGS -Dyyalloc=${name}_yyalloc -Dyysymbol_kind_t=${name}_yysymbol_kind_t)
```

scripts/glucose_CMakeLists.txt
tautschnig: Looks like this needs to be fixed again.

src/CMakeLists.txt
tautschnig: Looks like you need to do the same for yytokentype.

lzaoral:
Hey! Fedora packager of `cbmc` here.

I've come across this PR by an accident (and I'm glad I did).  I've got a few comments:

1. Making `cbmc` compilable with LTO is a great idea and it was on my TODO list for quite a while since LTO is the default for Fedora packages since 2020.  ~~Why do you want to restrict it to `x86_64` only, though?~~ (my bad, it is allowed everywhere) And what is the reason to make it mandatory? 

2. Using `-march=native` by default for everyone is a horrible idea.  Especially for downstream packagers like me because binary built on a newer hardware won't run on older almost surely because of the usage of newer CPU instructions. 

3. Using `-O3` does not necessarily guarantee faster software and it may also cause some subtle bugs.  There is a reason why some Linux distributions forbid it's usage.  Therefore, it is not a good idea to force it on everyone.

If you decide to merge this PR in its current form, please inform your downstream packagers so that they are not surprised that things would break due to issues mentioned above.

/cc: @vmihalko 

Edit: fix wrong statement about arch- limited LTO

lzaoral:
@rurban Could you please address my concerns raised in https://github.com/diffblue/cbmc/pull/6029#issuecomment-1694333365?

thomasspriggs:
@lzaoral I agree that it is the change to use LTO in this PR which is most worth merging. However if I remember correctly, last time I looked at this there was further work required to split the LTO changes from the other work and get them working in isolation. I have done some work towards this myself, but not enough to raise a PR. In the mean time, I have kicked off CI on this PR to check for incidental breakages.

rurban:
Will change lto fixes only

PR #6034, opened 2021-04-13
Fix havocing of struct members when enforcing invariants

SaswatPadhi:
Members expressions were ignored when computing the modified set of loops, and were therefore not being havoc'ed correctly.

In this change we fix this behavior by adding these expressions to the `modified` set.

### DIFF
--- a/regression/contracts/invar_dynamic_struct_member/main.c
+++ b/regression/contracts/invar_dynamic_struct_member/main.c
@@ -1,0 +1,30 @@
+typedef struct test
+{
+  int x;
+} test;
+
+void main()
+{
+  struct test *t = malloc(sizeof(test));
+  t->x = 0;
+
+  unsigned n;
+  for(unsigned i = 0; i < n; ++i)
+    __CPROVER_loop_invariant(i <= n)
+    {
+      switch(i % 4)
+      {
+      case 0:
+        break;
+
+      case 1:
+      case 2:
+        t->x += 1;
+
+      default:
+        t->x += 2;
+      }
+    }
+
+  assert(t->x == 0 || t->x == 1 || t->x == 2);
+}
--- a/regression/contracts/invar_struct_member/main.c
+++ b/regression/contracts/invar_struct_member/main.c
@@ -1,0 +1,19 @@
+struct test
+{
+  int x;
+};
+
+void main()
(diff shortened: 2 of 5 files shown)


SaswatPadhi:
cc: @feliperodri

regression/contracts/invar_dynamic_struct_member/main.c
```diff
@@ -0,0 +1,30 @@
...
+    __CPROVER_loop_invariant(i <= n)
+    {
+      switch(i % 4)
+      {
+      case 0:
+        break;
+
+      case 1:
+      case 2:
+        t->x += 1;
```
danielsn, line 22: Is this meant to fall through

regression/contracts/invar_struct_member/main.c
```diff
@@ -0,0 +1,19 @@
...
+  t.x = 0;
+
+  unsigned n;
+  for(unsigned i = 0; i < n; ++i)
+    __CPROVER_loop_invariant(i <= n)
+    {
+      t.x += 2;
+    }
+
+  assert(t.x == 0 || t.x == 2);
```
danielsn, line 18: Why this assertion?

SaswatPadhi: This assertion would pass if `t.x` is not havoced (so stays at value `0`), and the loop is replaced by a single nondet iteration. But should fail if the modifies set of the loop is havoced correctly.

I ~will add~ have added a comment explaining this in the `.desc` file.

regression/contracts/invar_dynamic_struct_member/main.c
SaswatPadhi: Yes, I wanted to keep one fall-through case in this `switch`.

tautschnig: May be worth adding a comment `// fall through`

tautschnig (approved):

TGWDB (approved):
It would be nice to have a test that passes as well as failures in the regression tests (or maybe there is one somewhere?).

NlightNFotis (approved):
LGTM

chrisr-diffblue (approved):

PR #6038, opened 2021-04-15
doc/man: Do not install the jbmc man pages when jbmc is not installed

vmihalko:
The commit message is explaining the change.


tautschnig:
I wonder whether it would be worth including a test of the `make install` target, just listing all the files that we expect to be installed (and doing a `diff` against a known-good list).

PR #6046, opened 2021-04-22
Vsd - widening during abstract object merge

jezhiggins:
Implements widening during abstract object merges. 

Widening during a merge is determined by the `trace_ptrt` locations passed into `variable_sensitivity_domaint::merge`. When widening should occur hasn't been changed, what this PR does is actually try and widen when requested. 

If a merge result is unmodified, no widening will take place. 

Constants don't need to be widened - rather, they already do because they go TOP.

For intervals, we consider which bound is being extended by the merge then extend that further out by the width of the interval. Eg.
* normal merge of [1, 3] with 4 results in [1, 4]
* widening merge of [1,3] with 4 results in [1, 8]. 
  * The merge result [1, 4] has been extended on the upper bound (4 > 3). 
  * The interval width is also 4, so the new upper bound is 4 + 4, hence [1, 8]
* widening merge of [4, 5] with -1 results in [-4, 5]
  * The merge result [1, 5] has been extended on the lower bound (1 < 4).
  * The interval width is 5, so the new lower bound is 1 - 5, hence [-4, 5].
If the merge extends both upper and lower bounds, we extend at both ends. 

For value-sets, a similar mechanism is applied. If the upper or lower value in the set has changed in the merge, we extend that further, either by replacing a constant value with an interval or by extending the interval. Eg.
* merging { 1, 3 } with 4 gives { 1, 3, 4 }
* widening merge of { 1, 3 } with 4 results in { 1, 3, [4, 8] }

This PR also includes value-set compacting. When a value-set becomes large, that is has more than value_set_abstract_objectt::max_value_set_size elements, it is compacted. This has two stages 
* non-destructive compacting - if the set contains any intervals, combine overlapping intervals together and then discard any values contained in those intervals
  * compacting { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, [5, 12] } would give { 1, 2, 3, 4, [5, 12] }
  * compacting { 1, 2, 3, 4, 5, 7, 8, [5, 10], [8, 12] } would give { 1, 2, 3, 4, [5, 12] }
  * compacting { [1, 4], 1, 2, 4, 5, 6, 7, [8, 10], [6, 12] } would give { [1, 4], 5, [6, 12] }
If the value-set is still too large, we move to the second stage
* destructive compacting - creating new intervals in value-set, then reapplying the non-destructive compact.
  * the intervals span the lower and upper third of the set's range 
  * so { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 } would become { [1, 4], 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, [9, 12] }, which would compact to { [1, 4], 5, 6, 7, 8, [9, 12] }.
  * if value set is still too large (because of outlier values), split the value-set into two intervals spanning the lower and upper half of the range.


unit/analyses/variable-sensitivity/value_set_abstract_object/widening_merge.cpp
```diff
@@ -0,0 +1,380 @@
...
+      auto merged = widening_merge(op1, op2);
+
+      THEN("widen lower bound - {[-3, 3], [2, 4]}")
+      {
+        auto interval_3minus_3 =
+          constant_interval_exprt(from_integer(-3, type), val3);
+        EXPECT_MODIFIED(merged, {interval_3minus_3, interval_2_4});
+      }
+    }
+    /*
```
martin-cs: Why is this commented out?

martin-cs (approved):
Although the PR is large, it seems the changes are mostly pretty straight-forward and a lot of it is tracing the extra parameter to merge through the various interfaces.

unit/analyses/variable-sensitivity/value_set_abstract_object/widening_merge.cpp
jezhiggins: I have no idea!

jezhiggins: Ah - I was using the tests I'd written around widening intervals as a template for widening value-sets. I'd commented these out because I hadn't got to them, and I think they must have been left dangling when we moved back to assume and meet. 
I'll see if there's anything useful in there, and remove the rest.

PR #6047, opened 2021-04-22
Build cbmc with cadical using make and IPASIR

TGWDB:
This adds documentation for CaDiCaL with IPASIR linking to `cbmc`. This is tested for Ubuntu and MacOS.

Note that while CaDiCaL can be compiled for Windows (Cygwin), there are problems with building `cbmc` that need to be fixed, and also problems with Cygwin handling of some libraries. For this reason Windows build support is not included in this PR, although this may come later (or via `cmake`).

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -380,6 +380,47 @@ For the packaged builds of CBMC on our release page we currently build CBMC
 with the MiniSat2 SAT solver statically linked at compile time. However it is
 also possible to build CBMC using alternative SAT solvers.
 
+### Compiling with CaDiCaL
+
+The [CaDiCaL](https://github.com/arminbiere/cadical) solver supports the
+[IPASIR](https://github.com/biotomas/ipasir) C interface to incremental SAT
+solvers, which is also supported by CBMC. So the process for producing a CBMC
+with CaDiCaL build is to build CaDiCaL as a static library then compile CBMC
+with the IPASIR build options and link to the CaDiCaL static library.
+
+Note that at the time of writing this has been tested to work with the CaDiCaL
+1.4.0 on Ubuntu 18.04 & 20.04.
+
+1. Download CaDiCaL:
+   ```
+   git clone --branch rel-1.4.0 https://github.com/arminbiere/cadical.git
+   ```
+   This will clone the CaDiCaL repository into a `cadical` subdirectory and
+   checkout release 1.4.0, which has been checked for compatibility with CBMC at
+   the time these instructions were written.
+
+2. Build CaDiCaL:
+   ```
+   cd cadical
+   ./configure
+   make cadical
+   cd ..
+   ```
+   This will create a build directory called `build` inside the clone of the
+   CaDiCaL repository. The `cadical` make target is specified in this example in
+   order to avoid building targets which are not required by CBMC. The built
+   static library will be placed in `cadical/build/libcadical.a`.
+
+3. Build CBMC:
+   ```
+   make -C src LIBS="$PWD/cadical/build/libcadical.a" IPASIR=$PWD/cadical/src
+   ```
+   This links the CaDiCaL library as part of the build. Passing the IPASIR
+   parameter tells the build system to build for the IPASIR interface. The
+   argument for the IPASIR parameter gives the build system the location for
+   the IPASIR headers, which is needed for the cbmc includes of `ipasir.h`. The
+   compiled binary will be placed in `cbmc/src/cbmc/cbmc`.
+
 ### Compiling with Riss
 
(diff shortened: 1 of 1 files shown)


COMPILING.md
```diff
@@ -380,6 +380,47 @@ For the packaged builds of CBMC on our release page we currently build CBMC
 with the MiniSat2 SAT solver statically linked at compile time. However it is
 also possible to build CBMC using alternative SAT solvers.
 
+### Compiling with CaDiCaL
+
+The [CaDiCaL](https://github.com/arminbiere/cadical) solver supports the
+[IPASIR](https://github.com/biotomas/ipasir) C interface to incremental SAT
+solvers, which is also supported by CBMC. So the process for producing a CBMC
+with CaDiCaL build is to build CaDiCaL as a static library then compile CBMC
+with the IPASIR build options and link to the CaDiCaL static library.
```
tautschnig, line 430: While that is one way of doing it, it's also possible to use the simpler approach using `-Dsat_impl=cadical` (with `cmake`) or `make -C src cadical-download` followed by `make -C src CADICAL=../../cadical`.

TGWDB: Yes. This was to document how to do it with the IPASIR interface.

tautschnig: I figured as much from the commit message, but could you perhaps also include a few words on those simpler approaches? I don't think the remainder of the section makes clear that this is only needed when choosing the IPASIR approach. (That said, it may be worth pondering long term whether we should stop supporting interfaces other than IPASIR to simplify maintenance.)

TGWDB: Pushed a change now that I believe clarifies how to build and link different SAT solvers and makes clear the differences.

tautschnig (approved):

COMPILING.md
```diff
@@ -380,7 +380,85 @@ For the packaged builds of CBMC on our release page we currently build CBMC
 with the MiniSat2 SAT solver statically linked at compile time. However it is
 also possible to build CBMC using alternative SAT solvers.
 
-### Compiling with Riss
+### Compiling with CBMC Wrapper
```
thomasspriggs: :pick: Not sure I would use the phrase "CBMC Wrapper". It could suggest a third party piece of software which wraps around cbmc, rather than our intended meaning of classes which live in CBMC and wrap around the native interfaces of the solver. Maybe "Compiling CBMC using solver native interfaces." would be better.

COMPILING.md
```diff
@@ -380,7 +380,85 @@ For the packaged builds of CBMC on our release page we currently build CBMC
...
+`SATPATH` is the path to the SAT interface.
+
+The rest of this section provides detailed instructions for some example
+SAT solvers.
+
+#### Compiling with CaDiCaL
+
+The [CaDiCaL](https://github.com/arminbiere/cadical) solver supports the
+[IPASIR](https://github.com/biotomas/ipasir) C interface to incremental SAT
+solvers, which is also supported by CBMC. So the process for producing a CBMC
```
thomasspriggs, line 428: Might be worth noting here that CaDiCaL can alternatively be used via its own interface.

COMPILING.md
```diff
@@ -380,7 +380,85 @@ For the packaged builds of CBMC on our release page we currently build CBMC
...
+make -C src LIBS="$PWD/SATOBJ SATLINKFLAGS" IPASIR=$PWD/SATPATH
+```
+Where `SATOBJ` is the pre-compiled IPASIR compatible SAT binary,
+`SATLINKFLAGS` are any flags required by the SAT object file, and
+`SATPATH` is the path to the SAT interface.
+
+The rest of this section provides detailed instructions for some example
+SAT solvers.
+
+#### Compiling with CaDiCaL
```
thomasspriggs: I suggest that we make it clear in this title that we are referring to use of CaDiCaL via the IPASIR interface. I know that we are already within an IPASIR subsection, but that could be missed when skimming through.

COMPILING.md
TGWDB: Updated.

COMPILING.md
TGWDB: Fixed.

COMPILING.md
TGWDB: Added paragraph on this (as well as other clarity changes).

COMPILING.md
thomasspriggs: :+1:

COMPILING.md
thomasspriggs: :+1:

COMPILING.md
thomasspriggs: :+1:

thomasspriggs (approved):
LGMT

PR #6048, opened 2021-04-22
Fix Cywin make build

TGWDB:
This fixes various details to ensure that make under Cywin can build
cbmc.

This is just some small fixes that appear to stop `cbmc` from being built using `make` and Cygwin.

### DIFF
--- a/src/util/run.cpp
+++ b/src/util/run.cpp
@@ -448,43 +448,10 @@ int run(
 #endif
 }
 
+#ifndef _WIN32
 /// quote a string for bash and CMD
 static std::string shell_quote(const std::string &src)
 {
-  #ifdef _WIN32
-  // first check if quoting is needed at all
-
-  if(src.find(' ')==std::string::npos &&
-     src.find('"')==std::string::npos &&
-     src.find('&')==std::string::npos &&
-     src.find('|')==std::string::npos &&
-     src.find('(')==std::string::npos &&
-     src.find(')')==std::string::npos &&
-     src.find('<')==std::string::npos &&
-     src.find('>')==std::string::npos &&
-     src.find('^')==std::string::npos)
-  {
-    // seems fine -- return as is
-    return src;
-  }
-
-  std::string result;
-
-  result+='"';
-
-  for(const char ch : src)
-  {
-    if(ch=='"')
-      result+='"'; // quotes are doubled
-    result+=ch;
-  }
-
-  result+='"';
-
-  return result;
-
-  #else
-
   // first check if quoting is needed at all
 
   if(src.find(' ')==std::string::npos &&
@@ -519,8 +486,8 @@ static std::string shell_quote(const std::string &src)
   result+='\'';
 
   return result;
-  #endif
 }
+#endif
 
 int run(
   const std::string &what,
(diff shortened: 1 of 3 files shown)


src/util/run.cpp
```diff
@@ -448,43 +448,10 @@ int run(
 #endif
 }
 
+#ifndef _WIN32
```
tautschnig, line 451: This change doesn't look like "just a small change" at all?

tautschnig (changes requested):
I'd prefer some more clarity in the commit message, and the failing Visual Studio builds suggest that these "small changes" aren't that small at all (the changes to `run.cpp` seem quite wrong, actually).

TGWDB:
> I'd prefer some more clarity in the commit message, and the failing Visual Studio builds suggest that these "small changes" aren't that small at all (the changes to `run.cpp` seem quite wrong, actually).

I'll comment directly here, then update the commit message(s) at the end if this makes it through review.

The Windows builds are all currently failing due to some CI problem, not related to this PR (that said, I created this PR to get the results from CI, so there is a chance there could be a problem).

There are three changes here:

1. Remove the `static` linking in the Cygwin build since this causes the build to fail (this was also suggested in the `Makefile` for Cygwin anyway).
2. Update and link the missing dependencies and fix the use of `SYMBOL_INFO` that appears to have been wrong for some time and not detected/tested.
3. Change the `ifdef` statements related to Cygwin (`_WIN32`) in `run.cpp`. Note that this appears to be a large change, but the root of the change is that Cygwin build does not call the `shell_quote` function. This causes the build to fail since the `-Werror` flag makes this an error. Rather than create a dummy call to this function, I chose to simply change the `ifdef` code to not include the function. This also meant deleting code inside the function that was inside an `ifdef` for building in Cygwin.

I hope this clarifies and that CI can be fixed and a proper CI test done.

tautschnig:
> > I'd prefer some more clarity in the commit message, and the failing Visual Studio builds suggest that these "small changes" aren't that small at all (the changes to `run.cpp` seem quite wrong, actually).
> 
> I'll comment directly here, then update the commit message(s) at the end if this makes it through review.
> 

Thank you!

> The Windows builds are all currently failing due to some CI problem, not related to this PR (that said, I created this PR to get the results from CI, so there is a chance there could be a problem).
> 

My apologies for not digging deeper and wrongly attributing these failures to your changes!

> There are three changes here:
> 
>     1. Remove the `static` linking in the Cygwin build since this causes the build to fail (this was also suggested in the `Makefile` for Cygwin anyway).
> 

Yes, this one seemed less surprising,

>     2. Update and link the missing dependencies and fix the use of `SYMBOL_INFO` that appears to have been wrong for some time and not detected/tested.
> 

and that change looked good.

>     3. Change the `ifdef` statements related to Cygwin (`_WIN32`) in `run.cpp`. Note that this appears to be a large change, but the root of the change is that Cygwin build does not call the `shell_quote` function. This causes the build to fail since the `-Werror` flag makes this an error. Rather than create a dummy call to this function, I chose to simply change the `ifdef` code to not include the function. This also meant deleting code inside the function that was inside an `ifdef` for building in Cygwin.
> 

This one I'm still not very clear about. `_WIN32` is more than Cygwin, this is all Windows builds. I am surprised to hear that `shell_quote` wasn't being invoked at all, it seemed rather important to do the work it is doing (proper escaping). Looking at the code base you are of course right, but it seems we have another problem to solve: `shell_quote` exists twice, once in `util/` and once in `ansi-c`. These are byte-by-byte the same, so one of them ought to be removed. This might fix the problem as the function would then actually be used?

tautschnig:
@TGWDB See #6054 for the fix of Windows builds in CI.

TGWDB:
Closing this PR as it mixes different things. The fix needed for general WIN32 build is now #6091.

PR #6049, opened 2021-04-22
Properly converts conditions into assertions and assumptions in function contracts

feliperodri:

### DIFF
--- a/regression/contracts/assigns_enforce_arrays_02/test.desc
+++ b/regression/contracts/assigns_enforce_arrays_02/test.desc
@@ -10,3 +10,5 @@ Checks whether the instrumentation process does not duplicate function calls
 used as part of array-index operations, i.e., if a function call is used in
 the computation of the index of an array assignment, then instrumenting that
 statement with an assigns clause will not result in multiple function calls.
+This test also ensures that assigns clauses correctly check for global
+variables modified by subcalls.
--- a/regression/contracts/assigns_enforce_arrays_03/test.desc
+++ b/regression/contracts/assigns_enforce_arrays_03/test.desc
@@ -6,5 +6,5 @@ main.c
 ^VERIFICATION FAILED$
 --
 --
-Checks whether verification fails when an array is assigned at an index
-below its lower bound.
+Checks whether verification fails when an assigns clause contains pointer,
+but function only modifies its content.
--- a/regression/contracts/assigns_enforce_functions_in_contracts/header.h
+++ b/regression/contracts/assigns_enforce_functions_in_contracts/header.h
@@ -1,0 +1,11 @@
+#include "utility.h"
+#include <stdbool.h>
+#include <stdlib.h>
+
+int foo(int *x) __CPROVER_requires(s2n_result_is_ok(validity3(x)))
+  __CPROVER_assigns(*x) __CPROVER_ensures(
+    __CPROVER_return_value == *x + 5 && s2n_result_is_ok(validity3(x)))
+{
+  *x = *x + 0;
+  return *x + 5;
+}
--- a/regression/contracts/assigns_enforce_functions_in_contracts/main.c
+++ b/regression/contracts/assigns_enforce_functions_in_contracts/main.c
@@ -1,0 +1,9 @@
+#include "header.h"
+
+int main()
+{
+  int *n = malloc(sizeof(*n));
+  *n = foo(n);
+
+  return 0;
+}
--- a/regression/contracts/assigns_enforce_functions_in_contracts/utility.h
+++ b/regression/contracts/assigns_enforce_functions_in_contracts/utility.h
@@ -1,0 +1,40 @@
+#include <stdbool.h>
+#include <stdlib.h>
+
+/* Function return code  */
+#define S2N_SUCCESS 0
+#define S2N_FAILURE -1
+
+/* A value which indicates the outcome of a function */
+typedef struct
+{
+  int __error_signal;
+} s2n_result;
+
+#define S2N_RESULT s2n_result
+#define S2N_RESULT_OK ((s2n_result){S2N_SUCCESS})
+#define S2N_RESULT_ERROR ((s2n_result){S2N_FAILURE})
+
+bool s2n_result_is_ok(s2n_result result)
+{
+  return result.__error_signal == S2N_SUCCESS;
+}
+
+bool validity1(int *x)
+{
+  return (x > 0);
+}
+
+bool validity2(int *x)
+{
+  return (x == 0);
+}
+
+S2N_RESULT validity3(int *x)
+{
+  if(x == NULL)
+    return S2N_RESULT_ERROR;
+  if(!validity1(x))
+    return S2N_RESULT_ERROR;
+  return S2N_RESULT_OK;
+}
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -17,6 +17,7 @@ Date: February 2016
 
 #include <analyses/local_may_alias.h>
 
+#include <goto-programs/goto_convert_class.h>
 #include <goto-programs/remove_skip.h>
 
 #include <linking/static_lifetime_init.h>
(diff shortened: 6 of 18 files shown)


regression/contracts/assigns_enforce_functions_in_contracts/header.h
```diff
@@ -0,0 +1,11 @@
...
+#include <stdbool.h>
+#include <stdlib.h>
+
+int foo(int *x) __CPROVER_requires(s2n_result_is_ok(validity3(x)))
+  __CPROVER_assigns(*x) __CPROVER_ensures(
+    __CPROVER_return_value == *x + 5 && s2n_result_is_ok(validity3(x)))
+{
+  *x = *x + 0;
+  return *x + 5;
+}
```
SaswatPadhi: Missing newline.

regression/contracts/assigns_enforce_functions_in_contracts/utility.h
```diff
@@ -0,0 +1,40 @@
...
+}
+
+S2N_RESULT validity3(int *x)
+{
+  if(x == NULL)
+    return S2N_RESULT_ERROR;
+  if(!validity1(x))
+    return S2N_RESULT_ERROR;
+  return S2N_RESULT_OK;
+}
```
SaswatPadhi: Missing newline.

regression/contracts/assigns_enforce_functions_in_contracts/main.c
```diff
@@ -0,0 +1,9 @@
+#include "header.h"
+
+int main()
+{
+  int *n = malloc(sizeof(*n));
+  n = foo(n);
```
SaswatPadhi: Should this be `*n = foo(n)`, because `foo` returns an `int` and `n` is an `int *`?

regression/contracts/assigns_enforce_arrays_02/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
SaswatPadhi: Please explain why this is a `KNOWNBUG` now.

regression/contracts/assigns_enforce_arrays_02/main.c
```diff
@@ -8,7 +8,7 @@ int nextIdx() __CPROVER_assigns(idx)
   return idx;
 }
 
-void f1(int a[], int len)
+void f1(int a[], int len) __CPROVER_assigns(*a)
```
SaswatPadhi: May be if we add `idx` to the `__CPROVER_assigns` clause, this might pass.

src/goto-instrument/code_contracts.cpp
```diff
@@ -327,11 +337,14 @@ bool code_contractst::apply_function_contract(
...
-      goto_programt::make_assertion(requires, target->source_location);
-
-    goto_program.insert_before_swap(target, a);
-    ++target;
+    goto_programt assertion;
+    convert_to_goto(
+      code_assertt(requires),
+      symbol_table.lookup_ref(function).mode,
+      assertion);
+    int lines_to_iterate = assertion.instructions.size();
```
SaswatPadhi: Use `auto` or `std::size_t` instead?

src/goto-instrument/code_contracts.cpp
```diff
@@ -353,12 +366,16 @@ bool code_contractst::apply_function_contract(
...
-  }
-  else
-  {
-    *target = goto_programt::make_skip();
+    goto_programt assumption;
+    convert_to_goto(
+      code_assumet(ensures),
+      symbol_table.lookup_ref(function).mode,
+      assumption);
+    int lines_to_iterate = assumption.instructions.size();
```
SaswatPadhi: Use `auto` or `std::size_t` instead?

src/goto-instrument/code_contracts.cpp
```diff
@@ -860,15 +868,18 @@ void code_contractst::add_contract_check(
   code_contractst::add_quantified_variable(
     requires, replace, function_symbol.mode);
 
+  goto_convertt converter(symbol_table, log.get_message_handler());
```
SaswatPadhi: Is this variable used somewhere?

src/goto-instrument/code_contracts.h
```diff
@@ -104,6 +104,12 @@ class code_contractst
   /// \brief Enforce contract of a single function
   bool enforce_contract(const std::string &);
 
+  /// \brief Create code corresponding to assertion/assumption
```
SaswatPadhi: This comment could be reworded for clarity.
1. Is it supposed to work only for assertions / assumptions?
2. It creates goto instructions (not `codet`), so saying "create code" might be confusing.

Meta comment about the approach: the function seems to be creating a `goto_convert` and just making a single call to it. Can we make this `converter` a private member of the `code_contractst` class instead, and just call the `converter.goto_convert` directly?

SaswatPadhi (changes requested):

regression/contracts/assigns_enforce_arrays_02/main.c
SaswatPadhi: I would actually suggest keeping this test as is, and modify the `test.desc` back to `CORE` and explain why this test should fail. This is to ensure that the ASSIGNS clause correctly checks for globals modified by subcalls.

Then, we should add another test, with the fix I proposed above and update its `test.desc` to reflect that it passes.

regression/contracts/assigns_enforce_functions_in_contracts/header.h
feliperodri: Fixed it.

regression/contracts/assigns_enforce_functions_in_contracts/main.c
feliperodri: Fixed it.

regression/contracts/assigns_enforce_functions_in_contracts/utility.h
feliperodri: Fixed it.

regression/contracts/assigns_enforce_arrays_02/main.c
feliperodri: Implemented as suggested.

regression/contracts/assigns_enforce_arrays_02/test.desc
feliperodri: Fixed it.

src/goto-instrument/code_contracts.cpp
feliperodri: Implemented as suggested.

src/goto-instrument/code_contracts.cpp
feliperodri: Implemented as suggested.

src/goto-instrument/code_contracts.cpp
feliperodri: Nope, removed it.

src/goto-instrument/code_contracts.h
feliperodri: I update the comment as suggested, but I don't see how adding `converter` as a member of this class would be beneficial here.

feliperodri:
@SaswatPadhi thank you for the comments, could you take another look?

chrisr-diffblue (approved):
Looks good - and I appreciate the better naming too :-)

TGWDB (approved):

regression/contracts/assigns_enforce_arrays_03/test.desc
```diff
@@ -6,5 +6,5 @@ main.c
 ^VERIFICATION FAILED$
 --
 --
-Checks whether verification fails when an array is assigned at an index
-below its lower bound.
+Checks whether verification fails when an assigns clause contans pointer,
```
SaswatPadhi: typo: conta[i]ns

SaswatPadhi (approved):
LGTM. Thanks for the fixes :D

src/goto-instrument/code_contracts.h
SaswatPadhi: I meant that `converter` is being initialized with the same 2 arguments:

```
goto_convertt converter(symbol_table, log.get_message_handler());
```

so why not make it a `private`/`protected` member, initialize it within `code_contractst` constructor, and directly call `converter.goto_convert` in place of `convert_to_goto`?

But we only call it twice anyway, so not a big deal.

regression/contracts/assigns_enforce_arrays_03/test.desc
feliperodri: Fixed it.

PR #6060, opened 2021-04-27
Remove unnecessary includes

tautschnig:
Manual removal of includes based on include-what-you-use's output,
filtered for includes that should be removed.

A fully automated use of include-what-you-use doesn't seem feasible at this time for there are a number of false positives of includes that iwyu claims can be safely removed.


peterschrammel (approved):

kroening (approved):

src/goto-cc/gcc_message_handler.cpp
```diff
@@ -8,7 +8,9 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "gcc_message_handler.h"
 
-#include <util/unicode.h>
+#ifdef _MSC_VER
```
martin-cs, line 11: This seems to be an idiom.  Would it be better to have the `ifdef` guard in `util/unicode.h`?

martin-cs (approved):
I am really glad this PR exists and even more glad that wasn't the one to have to write it!  I think it is definitely a good thing and should reduce build times.  Do you have a rough number for how much time it saves?  A few concerns:

1. Not sure why the coverage metrics drop.  Is it counting forward declarations of `class`es as executable code?
2. This has a very "breaks all existing PRs" feel to it.  What, if anything, are we going to do about that?
3. Having gone to all of this trouble, is there anything we could add to the CI / review process to avoid accumulating unnecessary includes and forward declarations?  Could we have some kind of 'try to remove #include lines from the PR and see if it still builds' task.  If it build develop first and then just tried a loop of removing lines adding headers from the diff, applying and seeing if it still builds?

src/goto-cc/gcc_message_handler.cpp
tautschnig: Well, it's really just the use of `widen` (in most of those files) that's required with Visual Studio only. The header provides more functionality than that, so I don't think that's really an option.

tautschnig:
> I am really glad this PR exists and even more glad that wasn't the one to have to write it! I think it is definitely a good thing and should reduce build times. Do you have a rough number for how much time it saves? A few concerns:

I've now added data to the commit message. Builds seem approximately 7% faster.

> 
>     1. Not sure why the coverage metrics drop.  Is it counting forward declarations of `class`es as executable code?
> 

I have no idea, and the Codecov page doesn't seem to be working at the moment. It seems that quite often with code removal the coverage metrics appear to drop by a tiny fraction, which is weird.

>     2. This has a very "breaks all existing PRs" feel to it.  What, if anything, are we going to do about that?
> 

Yes, there may be an impact on several PRs, but I don't expect that many actually: most of the includes simply weren't necessary, and when PRs change implementations only and not header files then most likely they're not impacted at all as an `#include` may have been shifted from the header file to the cpp file.

>     3. Having gone to all of this trouble, is there anything we could add to the CI / review process to avoid accumulating unnecessary includes and forward declarations?  Could we have some kind of 'try to remove #include lines from the PR and see if it still builds' task.  If it build develop first and then just tried a loop of removing lines adding headers from the diff, applying and seeing if it still builds?

I do think that we could use more recent versions of include-what-you-use in CI, perhaps 8.0 (which is in Ubuntu 20.04) is quite reasonable already. I've worked with 5.0 (from Ubuntu 18.04), which wasn't too great. The issue mainly is around detecting use of types in templates. I'll create a separate PR to play with this (also need to figure out how long it takes to run).

chrisr-diffblue (approved):
Wow - thats a lot of work... 👍

PR #6063, opened 2021-04-29
introduce new address constructor expressions

kroening:
This PR introduces three new expressions, which fix the largest
asymmetry in the expression system: expressions that denote pointers to
objects.

The first expression is the base constructor: it returns the address of an
object with a given identifier.  It is semantically equivalent to the
existing address_of expression with a symbol subexpression.

The next two expressions both take a base address, and then add to it the
offset (in char units) of either a named struct field or an array element
(given by its index).  They are semantically equivalent to but much more
concise and readable than the equivalent combination of typecasts and
pointer arithmetic.

They key benefit is that algorithms that iterate over expressions no longer
need to special case the sub-expressions of address_of expressions.


src/util/pointer_expr.h
```diff
@@ -381,6 +381,235 @@ inline address_of_exprt &to_address_of_expr(exprt &expr)
   return ret;
 }
 
+/// \brief Operator to return the address of an object
+class object_address_exprt : public nullary_exprt
```
tautschnig, line 385: How about making this a `unary_exprt`, storing the `symbol_exprt` as an operand? This would avoid taking apart and then rebuilding the `symbol_exprt`.

src/util/pointer_expr.h
```diff
@@ -381,6 +381,235 @@ inline address_of_exprt &to_address_of_expr(exprt &expr)
...
+  return ret;
+}
+
+/// \brief Operator to return the address of a field relative
+/// to a base address
+class field_address_exprt : public unary_exprt
+{
+public:
+  /// constructor for field addresses.
+  /// The base address must be a pointer to a compound type.
```
tautschnig, line 455: So should the `base` actually be an `object_address_exprt`?

src/util/pointer_expr.h
```diff
@@ -381,6 +381,235 @@ inline address_of_exprt &to_address_of_expr(exprt &expr)
...
+  return ret;
+}
+
+/// \brief Operator to return the address of an array element
+/// relative to a base address
+class element_address_exprt : public binary_exprt
+{
+public:
+  /// constructor for element addresses. The base address must have array type.
+  element_address_exprt(exprt base, exprt index);
```
tautschnig: Hmm, `field_address_exprt` suggests that the `base` must be a pointer, but here this isn't the case?

tautschnig (approved):
Some bits need clarification (see comments), but otherwise looking forward to rolling this out in the codebase.

src/util/pointer_expr.h
kroening: That is tempting, but recreates some of the problems that `address_of_exprt` has. In particular, this means that substitution of equals-for-equals requires special-casing `object_address_exprt`.

src/util/pointer_expr.h
kroening: No, this can be any pointer, say "NULL", or a pointer-typed symbol.

src/util/pointer_expr.h
kroening: The `to_pointer_type(base.type())` does enforce this.

tautschnig: Ok, but then the comment is inaccurate?

kroening: Yes, fixed the comment.

PR #6074, opened 2021-05-01
Enforce purity on missing assigns clause

feliperodri:
CBMC currently doesn't enforce assigns clauses if there isn't one.
However, whenever enforcing a function contract, the absence of an
assigns clause should be interpreted as an empty write set
(i.e., no inputs will be modified by this function).
CBMC now properly checks for empty assigns clause behavior.

### DIFF
--- a/regression/contracts/assigns_enforce_05/test.desc
+++ b/regression/contracts/assigns_enforce_05/test.desc
@@ -1,9 +1,11 @@
 CORE
 main.c
 --enforce-all-contracts
-^EXIT=10$
+^EXIT=0$
 ^SIGNAL=0$
-^VERIFICATION FAILED$
+^\[f1.1\] line \d+ .*: SUCCESS$
+^VERIFICATION SUCCESSFUL$
 --
 --
-This test checks that verification fails when a function with an assigns clause calls a function without an assigns clause.
+This test checks that verification succeeds when enforcing a contract
+for functions, without assigns clauses, that don't modify anything.
--- a/regression/contracts/assigns_enforce_15/main.c
+++ b/regression/contracts/assigns_enforce_15/main.c
@@ -1,0 +1,29 @@
+int global = 0;
+
+int foo(int *x) __CPROVER_assigns(*x)
+  __CPROVER_ensures(__CPROVER_return_value == *x + 5)
+{
+  int z = 5;
+  int y = bar(&z);
+  return *x + 5;
+}
+
+int bar(int *a) __CPROVER_ensures(__CPROVER_return_value >= 5)
+{
+  *a = *a + 2;
+  return *a;
+}
+
+int baz() __CPROVER_ensures(__CPROVER_return_value == global)
+{
+  global = global + 1;
+  return global;
+}
+
+int main()
+{
+  int n;
+  n = foo(&n);
+  n = baz();
+  return 0;
+}
--- a/regression/contracts/assigns_enforce_15/test.desc
+++ b/regression/contracts/assigns_enforce_15/test.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
+^\[bar.1\] line \d+ .*: FAILURE$
+^\[baz.1\] line \d+ .*: FAILURE$
+^VERIFICATION FAILED$
+--
+--
+Checks whether verification fails when enforcing a contract
+for functions, without assigns clauses, that modify an input.
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -513,50 +513,38 @@ void code_contractst::instrument_call_statement(
 
   exprt called_assigns =
     to_code_with_contract_type(called_symbol.type).assigns();
-  if(called_assigns.is_nil()) // Called function has no assigns clause
+  if(!called_assigns.is_nil()) // Called function has assigns clause
   {
-    // Create a false assertion, so the analysis
-    // will fail if this function is called.
-    goto_programt failing_assertion;
-    failing_assertion.add(goto_programt::make_assertion(
-      false_exprt(), instruction_iterator->source_location));
-    program.insert_before_swap(instruction_iterator, failing_assertion);
-    ++instruction_iterator;
-
-    return;
-  }
-    else // Called function has assigns clause
+    replace_symbolt replace;
+    // Replace formal parameters
+    code_function_callt::argumentst::const_iterator a_it =
+      call.arguments().begin();
+    for(code_typet::parameterst::const_iterator p_it =
+          called_type.parameters().begin();
+        p_it != called_type.parameters().end() &&
+        a_it != call.arguments().end();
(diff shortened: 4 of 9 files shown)


regression/contracts/assigns_enforce_05/test.desc
```diff
@@ -1,9 +1,10 @@
 CORE
 main.c
 --enforce-all-contracts
-^EXIT=10$
+^EXIT=0$
 ^SIGNAL=0$
-^VERIFICATION FAILED$
+^VERIFICATION SUCCESSFUL$
```
SaswatPadhi, line 7: Is there any way we could add some additional regexes to make sure that contracts are processed? The enforce flag should be generating some additional assertions, so can we match some regex on them?

The idea here is to fail this test (instead of vacuously passing) if somehow `--enforce-all-contracts` flag is a no-op.

regression/contracts/assigns_enforce_15/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
```
SaswatPadhi, line 8: Can we add some regexes to match on the assertions that failed?

regression/contracts/assigns_enforce_15/main.c
```diff
@@ -0,0 +1,20 @@
...
+  return *x + 5;
+}
+
+int bar(int *a) __CPROVER_ensures(__CPROVER_return_value >= 5)
+{
+  *a = *a + 2;
+  return *a;
+}
+
+int main()
```
SaswatPadhi, line 23: Can we add another function that doesn't have assigns clause, but modifies a global variable?
We can check for the appropriate assertion in the `test.desc` file instead of creating a separate regression test just for this case.

SaswatPadhi:
I just realized that the PR doesn't actually add support for `__CPROVER_assigns()`. It only checks for purity of functions when assigns clause is missing. So may be we can rename the PR (and update the commit message) to: "Enforce purity on missing assigns clause"?

kroening:
> I just realized that the PR doesn't actually add support for `__CPROVER_assigns()`. It only checks for purity of functions when assigns clause is missing. So may be we can rename the PR (and update the commit message) to: "Enforce purity on missing assigns clause"?

Yes, renaming the PR would help!

regression/contracts/assigns_enforce_05/test.desc
feliperodri: Implemented as suggested.

regression/contracts/assigns_enforce_15/test.desc
feliperodri: Implemented as suggested.

regression/contracts/assigns_enforce_15/main.c
feliperodri: Implemented as suggested.

regression/contracts/assigns_enforce_15/test.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=10$
+^SIGNAL=0$
+^\[bar.1\] line 13 .*: FAILURE$
+^\[baz.1\] line 18 .*: FAILURE$
```
TGWDB: This looks like it will fail, isn't it line 19? I think other tests use regex like:
```
^\[main.enum-range-check.6\] line \d+ enum range check in ev3: SUCCESS$
^\[main.enum-range-check.7\] line \d+ enum range check in ev2: FAILURE$
```
taken from `regression/cbmc/enum_is_in_range/enum_test5.desc`

src/goto-instrument/code_contracts.cpp
```diff
@@ -513,20 +513,8 @@ void code_contractst::instrument_call_statement(
...
-      false_exprt(), instruction_iterator->source_location));
-    program.insert_before_swap(instruction_iterator, failing_assertion);
-    ++instruction_iterator;
-
-    return;
-  }
-    else // Called function has assigns clause
-    {
+  if(!called_assigns.is_nil()) // Called function has assigns clause
+  {
```
TGWDB, line 517: The alignment of the old code now inside this conditional is off by two spaces. I don't know if this is something clang-format dislikes, but it would be nice to fix.

TGWDB:
I think this looks ok. There are some small tidying bits I noted, and I think it will fail regression tests (but this should just be fixing the test regexs). The only bit that is unclear to me is whether removing the old code that used to enforce failure is safe and won't break expected functionality. (Why was it written if this is fairly clearly failing when there is no assignment if this should instead be pure and not evaluated as a failure?)

SaswatPadhi (approved):
Approving since I don't have any additional comments, but let's fix all of the comments from @TGWDB.

regression/contracts/assigns_enforce_15/test.desc
feliperodri: Thanks, @TGWDB! Implemented as suggested.

src/goto-instrument/code_contracts.cpp
feliperodri: Implemented as suggested.

feliperodri:
> The only bit that is unclear to me is whether removing the old code that used to enforce failure is safe and won't break expected functionality. (Why was it written if this is fairly clearly failing when there is no assignment if this should instead be pure and not evaluated as a failure?)

@TGWDB Before, when enforcing a contract for function foo, CBMC would fail if any of its inner functions don't have an assigns clause, even if the inner function preserve all inputs. This behavior is too restrict. If a function doesn't change any of the inputs, its assigns clause should be empty. On this PR, we do one step further and instead of adding a false assertion for functions without assigns clause, we check for purity of the inner functions (i.e., they don't change any of the inputs).

src/goto-instrument/code_contracts.cpp
```diff
@@ -634,9 +622,6 @@ bool code_contractst::add_pointer_checks(const std::string &function_name)
   const auto &type = to_code_with_contract_type(function_symbol.type);
 
   exprt assigns_expr = type.assigns();
-  // Return if there are no reference checks to perform.
-  if(assigns_expr.is_nil())
-    return false;
```
chrisr-diffblue, line 639: Just a check - is there a test case that will fail if this code ever gets inserted again? :-) It superficially looks like an "obvious" optimisation, so I can imaging someone in the future re-adding it with good intent.

chrisr-diffblue (approved):
Looks good to me, thanks - with just one small sanity check on my part :-)

peterschrammel (approved):

src/goto-instrument/code_contracts.cpp
feliperodri: If somebody add this back, `assigns_enforce_15` will fail. For this case, we have assignments, but because there is no assigns clause, we wouldn't even complain about them. we should report an error in this case, because we want to enforce contracts across all functions using `--enforce-all-contracts` flag.

PR #6075, opened 2021-05-04
Add CMake support for building and linking against CaDiCaL using the IPASIR interface

NlightNFotis:
Expand the CMake build to be able to link against IPASIR supporting solvers. Two
new solver configurations have been added:

* CaDiCaL, which should be easy to compile against (`-Dsat_impl=ipasir_cadical` will handle downloading, building and linking against the solver), and
* Any arbitrary solver supporting the IPASIR interface (for the more ~adventurous~ demanding users, who may want to link against a different variety of solvers). This is more involved as an option (requires a number of different flags to be passed to CMake).

This expands our configuration options significantly. Previously, we could do one of two things:
  
  * Build with `cmake` using the internal CaDiCaL interface (`satcheck_cadical`), or
  * Build with `make` using the IPASIR interface.

This now allows us to build using both of the build systems, and link with a number of different solvers, for both testing and production use.

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -467,9 +467,21 @@ Note that at the time of writing this has been tested to work with the CaDiCaL
 It's also possible to build CBMC using CaDiCaL through IPASIR via `cmake`,
 controlled with the flag `-Dsat_impl=ipasir-cadical`, like so:
 
-```
+```sh
 $ cmake -Bbuild_ipasir -S. -Dsat_impl=ipasir-cadical
 ```
+
+An advanced user may also take a more adventurous route, trying to link
+CBMC against any solver supporthing the IPASIR interface. To do that,
+an invocation like this is needed:
+
+```sh
+$ cmake -Bbuild_ipasir -S . -Dsat_impl=ipasir-custom -DIPASIR=<source_location> -DIPASIR_LIB=<lib_location>
+```
+
+with `<source_location>` being the absolute path to the folder containing
+the solver implementation and `<lib_location>` being the absolute path that
+contains a precompiled static library of the solver (`.a` file).
 
 #### Compiling with Riss via IPASIR
 
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -462,6 +462,15 @@ Note that at the time of writing this has been tested to work with the CaDiCaL
    the IPASIR headers, which is needed for the cbmc includes of `ipasir.h`. The
    compiled binary will be placed in `cbmc/src/cbmc/cbmc`.
 
+---
+
+It's also possible to build CBMC using CaDiCaL through IPASIR via `cmake`,
+controlled with the flag `-Dsat_impl=ipasir-cadical`, like so:
+
+```
+$ cmake -Bbuild_ipasir -S. -Dsat_impl=ipasir-cadical
+```
+
 #### Compiling with Riss via IPASIR
 
 The [Riss](https://github.com/conp-solutions/riss) solver supports the
--- a/src/solvers/CMakeLists.txt
+++ b/src/solvers/CMakeLists.txt
@@ -160,6 +160,38 @@ elseif("${sat_impl}" STREQUAL "ipasir-cadical")
         ${cadical_SOURCE_DIR}/src
     )
     target_link_libraries(solvers cadical_ipasir)
+elseif("${sat_impl}" STREQUAL "ipasir-custom")
+    message(STATUS "Building with IPASIR solver linking: custom solver provided")
+
+    if (NOT DEFINED IPASIR)
+        message(FATAL_ERROR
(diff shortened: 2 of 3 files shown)


src/solvers/CMakeLists.txt
```diff
@@ -51,6 +55,7 @@ list(REMOVE_ITEM sources
     ${lingeling_source}
     ${booleforce_source}
     ${minibdd_source}
+    # ${ipasir_source}
```
TGWDB, line 58: Commented out option?

src/solvers/CMakeLists.txt
```diff
@@ -128,6 +133,33 @@ elseif("${sat_impl}" STREQUAL "cadical")
     )
 
     target_link_libraries(solvers cadical)
+elseif("${sat_impl}" STREQUAL "ipasir-cadical")
+    message(STATUS "Building with IPASIR solver linking against: CaDiCaL")
+
+    download_project(PROJ cadical
+        URL https://github.com/arminbiere/cadical/archive/rel-1.4.0.tar.gz
```
TGWDB, line 140: This uses CaDiCaL 1.4.0 and the non-IPASIR path uses CaDiCaL 1.3.0, can these by made the same version?

TGWDB:
I'm curious how much of the IPASIR building configuration is specific to CaDiCaL and how much is generic? Is it easy to link against another IPASIR compatible solver if one already has the correct path information to pass in to `cmake`? If this is a small addition/extension, this might be nice to include here.

src/solvers/CMakeLists.txt
NlightNFotis: Yes, I have it on my todo list, though it needs to be submitted as a different PR - I prefer PRs to be isolated to single feature changes.

src/solvers/CMakeLists.txt
NlightNFotis: No - this is a weird mechanism in the cmake file, where the files specified there are _removed_ from the build, and then re-introduced later, when the particular solver to be linked against has been identified.

I haven't understood the mechanism exactly (how it reintroduces the files for, say, minisat or glucose) so I've left this there as a reminder for me to investigate further. If I remove the commenting out of the ipasir sources there, what happens is that we get a linking error when building `smt2_solver`, where it complains that it can't find references for `satcheck_ipasir` (i.e. failing to build `satcheck_ipasir.cpp`).

COMPILING.md
```diff
@@ -462,6 +462,27 @@ Note that at the time of writing this has been tested to work with the CaDiCaL
    the IPASIR headers, which is needed for the cbmc includes of `ipasir.h`. The
    compiled binary will be placed in `cbmc/src/cbmc/cbmc`.
 
+---
+
+It's also possible to build CBMC using CaDiCaL through IPASIR via `cmake`,
+controlled with the flag `-Dsat_impl=ipasir-cadical`, like so:
+
+```sh
+$ cmake -Bbuild_ipasir -S . -Dsat-impl=ipasir-cadical
```
TGWDB: `-Dsat_impl`

COMPILING.md
```diff
@@ -462,6 +462,27 @@ Note that at the time of writing this has been tested to work with the CaDiCaL
...
+```
+
+An advanced user may also take a more adventurous route, trying to link
+CBMC against any solver supporthing the IPASIR interface. To do that,
+an invocation like this is needed:
+
+```sh
+$ cmake -Bbuild_ipasir -S . -Dsat-impl=ipasir-custom -DIPASIR=<source_location> -DIPASIR_LIB=<lib_location>
```
TGWDB: `-Dsat_impl`

COMPILING.md
```diff
@@ -462,6 +462,27 @@ Note that at the time of writing this has been tested to work with the CaDiCaL
...
+```
+
+with `<source_location>` being the path to the folder containing the solver implementation
+and `<lib_location>` being the path that contains a precompiled static library of
+the solver (`.a` file).
```
TGWDB: Note that these paths must be absolute and not relative!

TGWDB (approved):

TGWDB:
I've checked this on Ubuntu 20.04 with the CaDiCaL and also linked to SATCOMP via the IPASIR path, all worked for me.

chrisr-diffblue (approved):
Looks good to me. Thanks for the clear instructions!

src/solvers/CMakeLists.txt
tautschnig: I'd have the commit accomplishing this at https://github.com/tautschnig/cbmc/commit/330542a37c768f00f71959b2b2229336f15c3079, but thus far held back on creating a PR as performance needs to be evaluated (I vaguely recall a regression).

tautschnig (approved):

Issue #6065, opened 2021-04-29
Invariant check failed in `goto-analyzer`


feliperodri:
CBMC version: `develop` branch
Operating system: 18.04.1-Ubuntu

First, clone the [s2n-tls](https://github.com/aws/s2n-tls) repository.
```console
$ cd s2n-tls
$ git submodule update --initi --recursive
$ cd tests/cbmc/proofs/s2n_hash_free
$ make goto
$ goto-analyzer gotos/s2n_hash_free_harness.c.goto
```

What behaviour did you expect: No errors.

What happened instead:
```console
Storage not specified, defaulting to --one-domain-per-history
GOTO-ANALYSER version 5.28.0 (cbmc-5.28.1-14-gafd8a6095) 64-bit x86_64 linux
Reading GOTO program from file
Reading: gotos/s2n_hash_free_harness.c.goto
Generating GOTO Program
Adding CPROVER library (x86_64)
Removal of function pointers and virtual functions
Partial Inlining
--- begin invariant violation report ---
Invariant check failed
File: goto_inline_class.cpp:459 function: goto_inline
Condition: check_inline_map(inline_map)
Reason: Precondition
Backtrace:
goto-analyzer(+0x4aec4e) [0x55cbaec14c4e]
goto-analyzer(+0x4afbda) [0x55cbaec15bda]
goto-analyzer(+0x3f9ff) [0x55cbae7a59ff]
goto-analyzer(+0x353514) [0x55cbaeab9514]
goto-analyzer(+0x2f2b3d) [0x55cbaea58b3d]
goto-analyzer(+0x2f3360) [0x55cbaea59360]
goto-analyzer(+0x288ca9) [0x55cbae9eeca9]
goto-analyzer(+0x40ec9) [0x55cbae7a6ec9]
goto-analyzer(+0x45207) [0x55cbae7ab207]
goto-analyzer(+0x3da48) [0x55cbae7a3a48]
goto-analyzer(+0x332e4) [0x55cbae7992e4]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7) [0x7feea193bbf7]
goto-analyzer(+0x3f63a) [0x55cbae7a563a]


--- end invariant violation report ---
Aborted (core dumped)
```

martin-cs:
[...]

PR #6076, opened 2021-05-04
Ensure locations are updated after inlining

TGWDB:
This commit ensures the locations in the goto program are
updated during the partial inlining process. This is due to
some paths (in goto-analyzer) checking for this part way through
various goto program transformations.

Fixes issue #6065

### DIFF
--- a/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/s2n_hash_inlined.c
+++ b/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/s2n_hash_inlined.c
@@ -1,7 +1,17 @@
+// This file is highly reduced from some open source projects.
+// The following four lines are adapted from the openssl library
+// Full repository here:
+// Exact file adapted from here: https://github.com/openssl/openssl/tree/master
+// https://github.com/openssl/openssl/blob/master/crypto/evp/evp_local.h
 struct evp_md_ctx_st
 {
   const void *digest;
 };
+// The rest of this file is adapted from, various files in the
+// AWS s2n library. Full repository and licence information is
+// available here: https://github.com/aws/s2n-tls
+// Note that this is a highly cut down synthesis of many files
+// with most of their content and structure reduced.
 struct s2n_evp_digest
 {
   const void *ctx;
--- a/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/s2n_hash_inlined.c
+++ b/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/s2n_hash_inlined.c
@@ -1,0 +1,44 @@
+struct evp_md_ctx_st
+{
+  const void *digest;
+};
+struct s2n_evp_digest
+{
+  const void *ctx;
+};
+union s2n_hash_low_level_digest {
+};
+struct s2n_hash_evp_digest
+{
+  struct s2n_evp_digest evp_md5_secondary;
+};
+struct s2n_hash_state
+{
+  const struct s2n_hash *hash_impl;
+  union {
+    union s2n_hash_low_level_digest low_level;
+    struct s2n_hash_evp_digest high_level;
+  } digest;
+};
+struct s2n_hash
+{
+  int (*free)(struct s2n_hash_state *state);
+};
+void EVP_MD_CTX_free(struct evp_md_ctx_st *ctx)
+{
+  free(ctx->digest);
+  free(ctx);
+}
+static int s2n_evp_hash_free(struct s2n_hash_state *state)
+{
+  (EVP_MD_CTX_free((state->digest.high_level.evp_md5_secondary.ctx)));
+  return 0;
+}
+static const struct s2n_hash s2n_evp_hash = {
+  .free = &s2n_evp_hash_free,
+};
+static int s2n_hash_set_impl(struct s2n_hash_state *state)
+{
+  state->hash_impl = &s2n_evp_hash;
+  return 0;
+}
--- a/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/test.desc
+++ b/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/test.desc
@@ -1,0 +1,16 @@
+CORE
+test.c
+--verify
+^EXIT=0$
+^SIGNAL=0$
+Checking assertions
+^\[EVP_MD_CTX_free.precondition_instance.1\] line \d+ free argument must be NULL or valid pointer: SUCCESS
+--
+Invariant check failed
+--
+This test checks that after building the goto binary (see test.sh)
+that there is no errors that lead to invariant violations.
+This was created after a bug was found due to the
+instrument_preconditions code not correctly fixing locations and
+the invariant check of partial inlining detecting this location
+inconsistency.
--- a/src/goto-programs/instrument_preconditions.cpp
+++ b/src/goto-programs/instrument_preconditions.cpp
@@ -139,6 +139,12 @@ void instrument_preconditions(goto_modelt &goto_model)
   // now remove the preconditions
   for(auto &f_it : goto_model.goto_functions.function_map)
     remove_preconditions(f_it.second.body);
+  // The above may leave some locations uninitialized, this update is a
+  // sanity to check to ensure the goto model and functions are correct
+  // for later passes.
+  // Note that only the first loop is the one known to leave locations
+  // uninitialized.
+  goto_model.goto_functions.update();
 }
 
 void remove_preconditions(goto_functiont &goto_function)
--- a/regression/goto-cc-goto-analyzer/README.md
+++ b/regression/goto-cc-goto-analyzer/README.md
@@ -1,0 +1,9 @@
+This subdirectory is for tests where goto-cc is used to produce
+a goto binary that is then passed to goto-analyzer. There is also
+the option to have a custom script to produce the goto binary
+should this require special handling or multiple steps. This
+script is run if the root name of the test file has a script with
+the same filename. For example, if the desc file specifies 
+test.c as the test file then the chain.sh script will check for
+the existence of test.sh and if test.sh exists then it will be used
+to build the goto binary (assumed to be test.gb).
--- a/regression/goto-cc-goto-analyzer/chain.sh
+++ b/regression/goto-cc-goto-analyzer/chain.sh
@@ -9,12 +9,17 @@ is_windows=$3
 options=${*:4:$#-4}
 name=${*:$#}
 name=${name%.c}
+buildgoto=${name}.sh
 
-if [[ "${is_windows}" == "true" ]]; then
-  "${goto_cc}" "${name}.c"
-  mv "${name}.exe" "${name}.gb"
+if test -f ${buildgoto}; then
+  ./${buildgoto} ${goto_cc} ${is_windows}
 else
-  "${goto_cc}" "${name}.c" -o "${name}.gb"
+  if [[ "${is_windows}" == "true" ]]; then
+    "${goto_cc}" "${name}.c"
+    mv "${name}.exe" "${name}.gb"
+  else
+    "${goto_cc}" "${name}.c" -o "${name}.gb"
+  fi
 fi
 
 "${goto_analyzer}" "${name}.gb" ${options}
--- a/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/simple.c
+++ b/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/simple.c
@@ -1,3 +1,5 @@
+// This is highly adapted from the AWS s2n library.
+// Full repository and information here: https://github.com/aws/s2n-tls
 void s2n_hash_free_harness()
 {
 }
--- a/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/simple.c
+++ b/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/simple.c
@@ -1,0 +1,3 @@
+void s2n_hash_free_harness()
+{
+}
--- a/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/test.sh
+++ b/regression/goto-cc-goto-analyzer/instrument_preconditions_locations/test.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+set -e
+
+goto_cc=$1
+is_windows=$2
+
+if [[ "${is_windows}" == "true" ]]; then
+  ${goto_cc}  --export-file-local-symbols  simple.c 
+  mv simple.exe simple.gb
+  ${goto_cc}  --export-file-local-symbols  s2n_hash_inlined.c
+  mv s2n_hash_inlined.exe s2n_hash_inlined.gb
+  ${goto_cc}  --function s2n_hash_free_harness simple.gb s2n_hash_inlined.gb
+  mv simple.exe test.gb
+else
+  ${goto_cc}  --export-file-local-symbols  simple.c -o simple.gb
+  ${goto_cc}  --export-file-local-symbols  s2n_hash_inlined.c -o s2n_hash_inlined.gb
+  ${goto_cc}  --function s2n_hash_free_harness simple.gb s2n_hash_inlined.gb -o test.gb
+fi


TGWDB:
Note that regressions tests are NOT included (yet) as this is urgent/blocking for AWS.

thomasspriggs (changes requested):
This needs a regression test adding before we merge it.

NlightNFotis:
Yes, I also favour regression tests being in place before any merge.

As it stands, this looks like a random code change, with no way for me (as a reviewer) to verify that this fixes anything.

martin-cs (changes requested):
I really wouldn't be surprised if this line is the correct fix, failing to update location numbers after modifying goto-programs is a classic bug that many of us have made.  Particularly when duplicating code.

However I am less convinced that this is currently in the right place.  It seems that it is before the inlining changes things.  If this is the right place, please explain why in the commit message.

thomasspriggs:
> I really wouldn't be surprised if this line is the correct fix, failing to update location numbers after modifying goto-programs is a classic bug that many of us have made. Particularly when duplicating code.
> 
> However I am less convinced that this is currently in the right place. It seems that it is before the inlining changes things. If this is the right place, please explain why in the commit message.

I worked with TGWDB to help find the cause of the issue. There were definitely location numbers which were left uninitialised / zero valued before the inlining. I'd need to check to confirm whether or not additional bad locations were introduced during the inling process as well.

TGWDB:
> > I really wouldn't be surprised if this line is the correct fix, failing to update location numbers after modifying goto-programs is a classic bug that many of us have made. Particularly when duplicating code.
> > However I am less convinced that this is currently in the right place. It seems that it is before the inlining changes things. If this is the right place, please explain why in the commit message.
> 
> I worked with TGWDB to help find the cause of the issue. There were definitely location numbers which were left uninitialised / zero valued before the inlining. I'd need to check to confirm whether or not additional bad locations were introduced during the inling process as well.

I've been looking this morning (not updating each little bit). The problem with the locations in the example is occurring inside the `instrument preconditions` code (`instrument_preconditions.cpp`). Shifting the fix in the first commit (later commits to update this PR no doubt) to before `instrument preconditions` has the error, after the `instrument preconditions` prevents the error. I have not yet identified the exact conditions to minimise the example and also identify which part of the code is the culprit.

Note that this strongly implies the error is inside the `instrument preconditions` code and there is no indication of an error in the inlining code.

TGWDB:
Further update. During the unifying of the parsing options and goto code paths between `cbmc`, `goto-diff`, and `goto-analyzer` the `instrument_preconditions` function was added to the `goto-analyzer` behaviour. Previously `instrument_preconditions` was never called from `goto-analyzer` and so this is a new additional behaviour. The manipulations inside `instrument_preconditions` can add in locations that are not in correct order (and these can appear in `cbmc` and `goto-diff`) as well as in `goto-analyzer`. However, as a safety check (precondition) after partial function inlining `goto-analyzer` checks the locations are in correct order. Since the `instrument_preconditions` can violate this, `goto-analyser` now has an error due to new behaviour combined with an old check.

To summarise:

- `goto-analyser` has had `instrument_preconditions` added, which can create instructions with incorrect location order
- `goto-analyser` has a check for location order in the partial inlining code.
- Together these caused a new invariant violation.

With the above in mind, I recommend doing at least:
```
goto_model.goto_functions.compute_location_numbers();
```
in `process_goto_program.cpp` after `instrument_preconditions(goto_model);`. This will fix/update the locations (and put them in order), thus resolving this problem while leaving the checking in after the partial inlining.

TODO: also a small regression test.

martin-cs:
On Wed, 2021-05-05 at 02:28 -0700, thomasspriggs wrote:
> 
> I worked with TGWDB to help find the cause of the issue. There were
> definitely location numbers which were left uninitialised / zero
> valued before the inlining. I'd need to check to confirm whether or
> not additional bad locations were introduced during the inling
> process as well. 

Thank you; this is absolutely the correct thing to do.  There should be
no uninitialised locations at the end of any instrumentation pass so we
need to track down where they are added.  `goto_programt::update()`
should be called after modifications to a `goto-programt`.

https://github.com/diffblue/cbmc/blob/c3729fb105cc0b53c6d8ed0f5fa728b1f5109cc5/src/goto-programs/goto_program.cpp#L540

Sorry to be such a pain about this but it's the kind of thing that is
easy to 'paper over' but doing so causes long-term confusion and
inconsistency.

src/goto-programs/process_goto_program.cpp
```diff
@@ -55,6 +55,12 @@ bool process_goto_program(
   if(options.get_bool_option("partial-inline"))
   {
     log.status() << "Partial Inlining" << messaget::eom;
+    // Instrument preconditions can introduce locations that are not
+    // in correct order, while a postprocessing checking in the
+    // inlining can fail if locations are not in correct order.
+    // This refreshes the locations before inlining so any failures
+    // later are correctly attributed to errors in the inlining.
+    goto_model.goto_functions.compute_location_numbers();
```
martin-cs: I think this needs to be in `instrument_preconditions` because "`update` was called after modification" should be an invariant of `goto_programt`'s and things that work on them.

In an ideal world we could add this to the `validate` checks and call them as post-condition of all of these transforms.

martin-cs (changes requested):
Sorry to keep on about this but fixing this at the root cause is important.

TGWDB:
This is an update that goes with a commit that may come before or after this message. The commit moves the updating of the `goto_model.gotofunctions` inside the `instrument_preconditions` function. This should address the concerns about putting the validation/fixing of the code in the right place.

The main point for this message is to discuss the regression test (or lack of) and how to finish this PR. I've been refining the example that caused this originally but have not narrowed this down to anything suitable small and reasonably to turn into a regression test. Below are the (apparent) requirements to cause the invariant violation.

1. There need to be two goto files that are combined into a single goto file. Attempts to recreate the problem with only a single goto file or source file have been unsuccessful.
2. The `--export-file-local-symbols` flag must be used to create both goto files.

Unfortunately the example relies on openssl includes, and there is a LOT of complexity in sorting out the details of what is necessary and what isn't. So far shrinking the example has gone down to two files, one is 3 lines, the other is about 85 lines, but still includes openssl.

At this point there are five (all undesirable) options, which I'd like to have feedback on. (Presented in no particular order.)

1. Have PR with no regression test(s). This is clearly undesirable, but would be easy to approve and merge immediately (which is important to AWS).
2. Have a regression test that uses prebuilt goto files. This is undesirable for a couple of reasons. A: These goto binaries will inevitably go out of date in the future and the test will fail, hence losing the value of the test. B: The creation of these gotos may have the root cause of the error, and so the test will not catch the fix.
3. Have a regression test that relies upon having openssl libraries available. Again this is undesirable since this forces dependencies and other installs for regression tests and all the complexities that follow.
4. Continue to explore and try to find the exact interplay of options, files, dependencies, `cbmc` code, etc. that leads to the bug to either find the precise bug fix or be able to build a good regression test (which ever comes first). This is undesirable since it is complex and can be a (further) time sink of non-trivial proportions.
5. Revert the behaviour to not have `goto-analyser` call the `instrument_preconditions` function. This is undesirable since it again does not really fix or address the bug, but simply reverts to the behaviour before the error. Also this continues the separate code paths for different programs that #5807 aimed to fix (although this kind of holds true for the partial inlining option since only `goto-analyzer` can invoke it).

Feedback welcome!

src/goto-programs/instrument_preconditions.cpp
```diff
@@ -135,10 +135,15 @@ void instrument_preconditions(goto_modelt &goto_model)
     instrument_preconditions(
       goto_model,
       f_it.second.body);
-
```
martin-cs: What happens if you move the addition and call to `update` to here?

martin-cs (commented):
I am still not 100% convinced this is the right patch but I think it is close to fixing /a/ problem.  I would not block this patch.

src/goto-programs/goto_inline.cpp
```diff
@@ -203,7 +203,6 @@ void goto_partial_inline(
       }
     }
   }
-
```
thomasspriggs: :pick: This empty line removal looks unrelated to the changes this commit is supposed to be making. Presumably this is due to adding and removing different changes in this PR, but this formatting change should be cleaned up. Same applied to the line marked with a :dash:

src/goto-programs/instrument_preconditions.cpp
thomasspriggs: :dash: See previous comment.

thomasspriggs (commented):
Looks like this is still missing a regression test. Message me and we can work together on reducing the example down, if it is still causing problems.

martin-cs:
@TGWDB thank you so much for chasing this one.  I think the changes you have made are already a lot better but I am not quite sure we are at the end of the journey yet.

Thoughts on the options.

1. Undesirable but possible.  My concern is less that there isn't a regression test and more that we don't know for sure what causes this.
2. We have contemplated this before but it is a big ask and "haven't understood the bug enough" is a bad reason to add one.
3. Agreed; the a goal is to keep CPROVER dependency light.  This is not a good enough reason to add a dependency like that.
4. This is technically the correct and desirable option.  I guess the question is how simple / fast / cheap we can make this.
5. I'd really rather not.

How does the following sound for a (falsifiable) conjecture...?
1. The problem does not exist before `instrument_precondition` is called.  Check by commenting it out and seeing if that crashes.
2. The problem doesn't show up in other tools (at the moment...) is that they call another instrumentation pass which does call `update` before the first pass that actually uses the location numbers (i.e. `goto_partial_inline` in the case of `goto_analyzer`).  The bug exists but is masked (which is why common preprocessing is a good goal!).  This should be possible to test by enabling `goto_partial_inline` in CBMC and disabling any other instrumentation passes that call `update` inbetween the two.
3. https://github.com/diffblue/cbmc/blob/84ad28764e0b3da5383c20975f5a340dae321e84/src/goto-programs/instrument_preconditions.cpp#L117 adds an instruction to a `goto_programt` without calling `update`, which is definitely a bug.  This can be tested by modifying the code to output the location number or to look at the crash and check it is using one of the added instructions.
4. Two files are needed because if the precondition and the call site are in the same compilation unit this processing is done earlier?  This can be tested by ... asking @tautschnig ?  Looking at the code of `instrument_precondition` should give some idea of what triggers the addition.
5. OpenSSL is needed because we don't have any tests with preconditions that have preconditions and their usage in different files.  Testing this should be possible by isolating a function with a precondition and it's call into two separate files (or whatever other conditions are needed from 4).

If this lot is correct or correct enough then I think that gives you a route to a regression test and to option 4.

martin-cs:
Also @TGWDB have you come across https://github.com/csmith-project/creduce ?

NlightNFotis:
@thomasspriggs The example is very hard to be reduced further - I checked it out with Thomas G. yesterday.

The reason for that is that the problematic behaviour is manifesting on a reduced file that includes an `openssl` header file that is kind of an *all encompassing* `include` - meaning it transitively includes a huge number of other headers. If we remove that header file, the problem disappears. It's unclear which one of the transitively included headers is making the error manifest, and detecting/isolating that looks like is a significant amount of effort.

NlightNFotis:
@martin-cs This is a great idea, I didn't think of `c-reduce`. I don't know if Thomas tried it before, but this looks like it might be an option worth trying to see if we can come up with a regression test.

TGWDB:
Quick response to update, I'll add more later today/tomorrow.

> @TGWDB thank you so much for chasing this one. I think the changes you have made are already a lot better but I am not quite sure we are at the end of the journey yet.
> 

snip

> How does the following sound for a (falsifiable) conjecture...?
> 
> 1. The problem does not exist before `instrument_precondition` is called.  Check by commenting it out and seeing if that crashes.

This is true and tested before.

> 2. The problem doesn't show up in other tools (at the moment...) is that they call another instrumentation pass which does call `update` before the first pass that actually uses the location numbers (i.e. `goto_partial_inline` in the case of `goto_analyzer`).  The bug exists but is masked (which is why common preprocessing is a good goal!).  This should be possible to test by enabling `goto_partial_inline` in CBMC and disabling any other instrumentation passes that call `update` inbetween the two.

This is also true and tested before.

> 3. https://github.com/diffblue/cbmc/blob/84ad28764e0b3da5383c20975f5a340dae321e84/src/goto-programs/instrument_preconditions.cpp#L117
>     adds an instruction to a `goto_programt` without calling `update`, which is definitely a bug.  This can be tested by modifying the code to output the location number or to look at the crash and check it is using one of the added instructions.

I'll look into this further (it's part of what I was doing when I wrote above).

> 4. Two files are needed because if the precondition and the call site are in the same compilation unit this processing is done earlier?  This can be tested by ... asking @tautschnig ?  Looking at the code of `instrument_precondition` should give some idea of what triggers the addition.
> 5. OpenSSL is needed because we don't have any tests with preconditions that have preconditions and their usage in different files.  Testing this should be possible by isolating a function with a precondition and it's call into two separate files (or whatever other conditions are needed from 4).

I strongly suspect we can do it without openssl, eventually.

> If this lot is correct or correct enough then I think that gives you a route to a regression test and to option 4.

Agree, it's the path I'm on.

TGWDB:
> Also @TGWDB have you come across https://github.com/csmith-project/creduce ?

I tried `cpp-merge` but not this, I'll have a look. Thanks!

thomasspriggs (approved):
:+1:

TGWDB:
@martin-cs Can you please re-review this?

NlightNFotis (approved):
LGTM.

regression/goto-cc-goto-analyzer/instrument_preconditions_locations/s2n_hash_inlined.c
```diff
@@ -0,0 +1,44 @@
+struct evp_md_ctx_st
+{
+  const void *digest;
+};
+struct s2n_evp_digest
+{
+  struct evp_md_ctx_st *ctx;
```
martin-cs: It feels like this might be simplified a bit more, can this be replaced with the `const void *` field directly?

Equally well; if you are done with reducing this, that is also fine.

regression/goto-cc-goto-analyzer/instrument_preconditions_locations/test.desc
```diff
@@ -0,0 +1,15 @@
+CORE
+test.c
+--verify
+Checking assertions
+^EXIT=0$
```
martin-cs, line 4: Could you check for at least one of the assertions?

regression/goto-cc-goto-analyzer/instrument_preconditions_locations/test.desc
```diff
@@ -0,0 +1,15 @@
...
+^SIGNAL=0$
+--
+Invariant check failed
+--
+This test checks that after building the goto binary (see test.sh)
```
martin-cs, line 11: Brilliant!

src/goto-programs/instrument_preconditions.cpp
```diff
@@ -139,6 +139,12 @@ void instrument_preconditions(goto_modelt &goto_model)
   // now remove the preconditions
   for(auto &f_it : goto_model.goto_functions.function_map)
     remove_preconditions(f_it.second.body);
+  // The above may leave some locations uninitialized, this update is a
+  // sanity to check to ensure the goto model and functions are correct
+  // for later passes.
+  // Note that only the first loop is the one known to leave locations
+  // uninitialized.
+  goto_model.goto_functions.update();
```
martin-cs, line 147: :+1:

martin-cs (approved):
Thank you for chasing this down and sorry it has turned out to be more of an ordeal than it first looked.

Comments are mostly thoughts / suggestions but one last question -- what triggers the `instrument_preconditions` if there are no preconditions in the source code?

regression/goto-cc-goto-analyzer/instrument_preconditions_locations/s2n_hash_inlined.c
TGWDB: Done

regression/goto-cc-goto-analyzer/instrument_preconditions_locations/test.desc
TGWDB: Tested for the first verification  result.

TGWDB:
> Thank you for chasing this down and sorry it has turned out to be more of an ordeal than it first looked.
> 
> Comments are mostly thoughts / suggestions but one last question -- what triggers the `instrument_preconditions` if there are no preconditions in the source code?

Line 52 of `process_goto_program.cpp` always invokes `instrument_preconditions` regardless of anything else. I cannot comment on why this decision was made originally (or why this was added to `goto-analyzer`.

martin-cs:
Apologies; I wasn't clear. `instrument_preconditions` is always run in an attempt to make the preprocessing of goto programs common between all tools (this is a work-in-progress).

My confusion was that https://github.com/diffblue/cbmc/blob/0ea7f13c4f14f52d803b5b7004b43e101d48952c/src/goto-programs/instrument_preconditions.cpp#L120 should only modify the program ( and thus create the uninitialised locations ) if there are preconditions.  And the test case doesn't have any.  I feel like there is something I am not getting and I don't know if it is significant.

TGWDB:
@peterschrammel @tautschnig Any chance of a codeowner review for this?

regression/goto-cc-goto-analyzer/instrument_preconditions_locations/s2n_hash_inlined.c
```diff
@@ -0,0 +1,44 @@
+struct evp_md_ctx_st
```
peterschrammel, line 6: ❓ What's the license of this file?

peterschrammel (approved):

regression/goto-cc-goto-analyzer/instrument_preconditions_locations/s2n_hash_inlined.c
TGWDB: For `s2n_hash_inlined.c` lines 1-4 are adapted from openssl which has an Apache like licence and lines 5-44 are modified from the AWS s2n library that is also under Apache 2.0 licence. All of `simple.c` is modified from the AWS s2n library.

peterschrammel: Please add a header and comments to explain this so that we know where this is coming from.

martin-cs:
To answer my own question : it might well be `free` that is triggering this.  In which case it should be possible to make the example a *lot* simpler.

regression/goto-cc-goto-analyzer/instrument_preconditions_locations/s2n_hash_inlined.c
TGWDB: Done

martin-cs:
I still felt like I didn't understand why this bug was occurring so I started having a dig.

`s2n_hash_inlined.c` can be reduced to:

```
void *p;
void *q;

void my_free(void)
{
  // Need both frees
  free(p);
  free(q);
}

typedef void (*free_function)(void);

free_function global = &my_free;

void point(free_function *f)
{
  *f = global;
}
```

Doing this removes the need for the `--export-file-local-symbols` flags.

This reduces the possible cause of the issue but I am still pretty hazy on the exact sequence of events that cause it.  I /speculate/ that there is something to do with a pointer to `my_free` possibly escaping and so needing to be processed later?  Is it that the preconditions of `free` are being inlined into `my_free` and then need inlining again?

TGWDB:
> I still felt like I didn't understand why this bug was occurring so I started having a dig.
> 
> `s2n_hash_inlined.c` can be reduced to:
> 
> ```
> void *p;
> void *q;
> 
> void my_free(void)
> {
>   // Need both frees
>   free(p);
>   free(q);
> }
> 
> typedef void (*free_function)(void);
> 
> free_function global = &my_free;
> 
> void point(free_function *f)
> {
>   *f = global;
> }
> ```
> 
> Doing this removes the need for the `--export-file-local-symbols` flags.
> 
> This reduces the possible cause of the issue but I am still pretty hazy on the exact sequence of events that cause it. I /speculate/ that there is something to do with a pointer to `my_free` possibly escaping and so needing to be processed later? Is it that the preconditions of `free` are being inlined into `my_free` and then need inlining again?

Thanks for digging deeper into this. I merged since the invariant violation was a blocking issue for AWS.

I'm not sure if the "inlining" is a distraction here since the problems appear before the `cbmc` inlining code itself (unless you meant inlining somewhere else?).

martin-cs:
Inlining the precondition.

I think there is another bug which is a difference in how the preprocessing is done and possibly when / how the C library is linked.

Issue #6045, opened 2021-04-20
Memory primitives should handle invalid pointers


feliperodri:
CBMC version: 5.27.0 (cbmc-5.26.0-148-g7ced011bd-dirty)
Operating system: macOS Mojave 10.14.6
Exact command line resulting in the issue: `cbmc main.c --malloc-may-fail --malloc-fail-null --trace`
What behaviour did you expect: `line 7 assertion __CPROVER_r_ok(ptr, 0): FAILURE`
What happened instead: `VERIFICATION SUCCESSFUL`

To explain the behavior, let's consider the following example.

```c
// main.c
#include <stdlib.h>
#include <assert.h>

int main() {
  void *ptr = malloc(0);
  __CPROVER_assume(ptr != NULL);
  assert(__CPROVER_r_ok(ptr, 0));
}
```

I expect the assertion to fail, because `ptr` could be an invalid pointer; however, it succeeds.

```shell
CBMC version 5.27.0 (cbmc-5.26.0-148-g7ced011bd-dirty) 64-bit x86_64 macos
Parsing main.c
Converting
Type-checking main
Generating GOTO Program
Adding CPROVER library (x86_64)
Removal of function pointers and virtual functions
Generic Property Instrumentation
Running with 8 object bits, 56 offset bits (default)
Starting Bounded Model Checking
Runtime Symex: 0.0034956s
size of program expression: 94 steps
simple slicing removed 5 assignments
Generated 1 VCC(s), 1 remaining after simplification
Runtime Postprocess Equation: 1.1753e-05s
Passing problem to propositional reduction
converting SSA
Runtime Convert SSA: 0.000522958s
Running propositional reduction
Post-processing
Runtime Post-process: 8.1631e-05s
Solving with MiniSAT 2.2.1 with simplifier
437 variables, 313 clauses
SAT checker inconsistent: instance is UNSATISFIABLE
Runtime Solver: 9.8288e-05s
Runtime decision procedure: 0.000650914s

** Results:
<builtin-library-malloc> function malloc
[malloc.assertion.1] line 26 max allocation size exceeded: SUCCESS
[malloc.assertion.2] line 31 max allocation may fail: SUCCESS

main.c function main
[main.assertion.1] line 8 assertion __CPROVER_r_ok(ptr, 0): SUCCESS

** 0 of 3 failed (1 iterations)
VERIFICATION SUCCESSFUL
```
[...]

PR #6077, opened 2021-05-04
Check for validity of pointers in assigns clause

ArenBabikian:
Consider the following example:

```
void bar(int *x, int *y) __CPROVER_assigns(*x, *y)
{
  *x = 3;
  if(y != NULL)
    *y = 5;
}

void foo(int *x) __CPROVER_assigns(*x)
{
  bar(x, NULL);
  *x = 3;
}
```

The potential problem in this case is that `foo` calls `bar` by passing a null pointer. Of course, considering the contents of `bar`, this will not be problematic. However, the assigns clause for `bar` is not aware of what is inside `bar`, so the assigns clause previously could not handle the null pointer.

This PR adds support for such cases, by checking for passed null pointers at the goto program level.

**Note:** currently, this PR only supports pointers, which is required for function contracts for the [coreJSON](https://github.com/FreeRTOS/coreJSON) project.

### DIFF
--- a/regression/contracts/assigns_validity_pointer_01/test.desc
+++ b/regression/contracts/assigns_validity_pointer_01/test.desc
@@ -1,0 +1,29 @@
+CORE
+main.c
+--enforce-contract foo --replace-call-with-contract bar --replace-call-with-contract baz
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+SUCCESS
+// bar
+ASSERT \*x > 0
+IF !\(\*x == 3\) THEN GOTO \d
+IF !\(\(signed int \*\)\(void \*\)0 == \(\(signed int \*\)NULL\)\) THEN GOTO \d
+tmp_if_expr = \*\(\(signed int \*\)\(void \*\)0\) == 5 \? TRUE \: FALSE;
+tmp_if_expr\$\d = tmp_if_expr \? TRUE : FALSE;
+ASSUME tmp_if_expr\$\d
+// baz
+IF !\(z == \(\(signed int \*\)NULL\)\) THEN GOTO \d
+tmp_if_expr\$\d = \*z == 7 \? TRUE : FALSE;
+ASSUME tmp_if_expr\$\d
+// foo
+ASSUME \*tmp_cc\$\d > 0
+ASSERT \*tmp_cc\$\d == 3
+--
+\[3\] file main\.c line 6 assertion: FAILURE
+--
+Verification:
+This test checks support for a NULL pointer that is assigned to by
+a function (bar and baz). Both functions bar and baz are being replaced by
+their function contracts, while the calling function foo is being checked
+(by enforcing it's function contracts).
--- a/regression/contracts/assigns_validity_pointer_03/test.desc
+++ b/regression/contracts/assigns_validity_pointer_03/test.desc
@@ -1,0 +1,28 @@
+KNOWNBUG
+main.c
+--enforce-contract foo --replace-call-with-contract bar --replace-call-with-contract baz
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+// bar
+ASSERT \*x > 0
+IF !\(\*x == 3\) THEN GOTO \d
+tmp_if_expr = \*y == 5 \? TRUE \: FALSE;
+ASSUME tmp_if_expr
+// baz
+ASSUME \*z == 7
+// foo
+ASSUME \*tmp_cc\$\d > 0
+ASSERT \*tmp_cc\$\d == 3
+--
+--
+Verification:
+This test checks support for an uninitialized pointer that is assigned to by
+a function (bar and baz). Both functions bar and baz are being replaced by
+their function contracts, while the calling function foo is being checked
+(by enforcing it's function contracts).
+
+Known Bug:
+Currently, there is a known issue with __CPROVER_w_ok(ptr, 0) such that it
+returns true if ptr is uninitialized. This is not the expected behavior,
+therefore, the outcome of this test case is currently incorrect.
--- a/regression/contracts/assigns_validity_pointer_04/main.c
+++ b/regression/contracts/assigns_validity_pointer_04/main.c
@@ -1,0 +1,34 @@
+#include <assert.h>
+#include <stddef.h>
+
+int *z;
+
+void bar(int *x, int *y) __CPROVER_assigns(*x, *y) __CPROVER_requires(*x > 0)
+  __CPROVER_ensures(*x == 3 && *y == 5)
+{
+}
+
+void baz() __CPROVER_assigns(*z) __CPROVER_ensures(*z == 7)
+{
+}
+
+void foo(int *x) __CPROVER_assigns(*x) __CPROVER_requires(*x > 0)
+  __CPROVER_ensures(*x == 3)
+{
+  int *y = malloc(sizeof(int));
+  bar(x, y);
+  assert(*y == 5);
+
+  z = malloc(sizeof(int));
+  baz();
+  assert(*z == 7);
+}
+
+int main()
+{
+  int n;
+  foo(&n);
+
+  assert(n == 3);
+  return 0;
+}
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -1524,10 +1524,39 @@ goto_programt assigns_clauset::havoc_code(
   goto_programt havoc_statements;
   for(assigns_clause_targett *target : targets)
   {
+    // (1) If the assigned target is not a dereference,
+    // only include the havoc_statement
+
+    // (2) If the assigned target is a dereference, do the following:
+
+    // if(!__CPROVER_w_ok(target, 0)) goto z;
+    //      havoc_statements
+    // z: skip
+
+    // create the z label
+    goto_programt tmp_z;
+    goto_programt::targett z = tmp_z.add(goto_programt::make_skip(location));
+
+    const auto &target_ptr = target->get_direct_pointer();
+    if(to_address_of_expr(target_ptr).object().id() == ID_dereference)
+    {
+      // create the condition
+      exprt condition =
+        not_exprt(w_ok_exprt(target_ptr, from_integer(0, integer_typet())));
+      havoc_statements.add(goto_programt::make_goto(z, condition, location));
+    }
+
+    // create havoc_statements
     for(goto_programt::instructiont instruction :
         target->havoc_code(location).instructions)
     {
       havoc_statements.add(std::move(instruction));
+    }
+
+    if(to_address_of_expr(target_ptr).object().id() == ID_dereference)
+    {
+      // add the z label instruction
+      havoc_statements.destructive_append(tmp_z);
     }
   }
   return havoc_statements;
@@ -1577,8 +1606,28 @@ exprt assigns_clauset::compatible_expression(
     {
       if(first_iter)
       {
-        current_target_compatible =
-          target->compatible_expression(*called_target);
+        // TODO: Optimize the validation below and remove code duplication
+        // See GitHub issue #6105 for further details
+
+        // Validating the called target through __CPROVER_w_ok() is
+        // only useful when the called target is a dereference
+        const auto &called_target_ptr = called_target->get_direct_pointer();
+        if(
+          to_address_of_expr(called_target_ptr).object().id() == ID_dereference)
+        {
+          // or_exprt is short-circuited, therefore
+          // target->compatible_expression(*called_target) would not be
+          // checked on invalid called_targets.
+          current_target_compatible = or_exprt(
+            not_exprt(
+              w_ok_exprt(called_target_ptr, from_integer(0, integer_typet()))),
+            target->compatible_expression(*called_target));
+        }
+        else
+        {
+          current_target_compatible =
+            target->compatible_expression(*called_target);
+        }
         first_iter = false;
       }
       else
--- a/regression/contracts/assigns_replace_03/test.desc
+++ b/regression/contracts/assigns_replace_03/test.desc
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
 main.c
 --replace-all-calls-with-contracts
 ^EXIT=0$
@@ -7,3 +7,7 @@ main.c
 --
 --
 This test checks that a havocked variable can be constrained by a function post-condition.
+
+Known Bug:
+Currently, there is a bug when char variables are being passed to
+__CPROVER_w_ok(). See GitHub issue #6106 for further details.
--- a/regression/contracts/assigns_validity_pointer_01/main.c
+++ b/regression/contracts/assigns_validity_pointer_01/main.c
@@ -1,0 +1,36 @@
+#include <assert.h>
+#include <stddef.h>
+
+int *z;
+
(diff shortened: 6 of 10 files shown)


src/goto-instrument/code_contracts.cpp
```diff
@@ -1595,8 +1615,15 @@ exprt assigns_clauset::compatible_expression(
     {
       if(first_iter)
       {
-        current_target_compatible =
-          target->compatible_expression(*called_target);
+        // TODO: generalize this
+        // currently only supports pointers
```
feliperodri: Could you turn this comment into a INVARIANT and open a GitHub issue to track this case?

src/goto-instrument/code_contracts.cpp
```diff
@@ -1542,11 +1542,31 @@ goto_programt assigns_clauset::havoc_code(
...
+    //    if(target != NULL) goto x;
+    //      havoc_statements
+    // z: skip
+
+    // create the z label
+    goto_programt tmp_z;
+    goto_programt::targett z = tmp_z.add(goto_programt::make_skip(location));
+
+    // TODO: generalize this
+    // currently only supports pointers
```
feliperodri: Could you turn this comment into a INVARIANT and open a GitHub issue to track this case?

regression/contracts/assigns_null_pointer_02/test.desc
```diff
@@ -0,0 +1,13 @@
+CORE
+main.c
+--enforce-contract foo
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
```
feliperodri: Could you also add a test where we have an allocation that non-deterministically fails and pass the pointer to a function with an assigns clause including this pointer? I want to make sure we cover cases where the pointer comes from mallocs.

feliperodri (changes requested):
I few minor comments.

src/goto-instrument/code_contracts.cpp
ArenBabikian: I was obliged to do an if statement instead. Otherwise, the regression tests for structs and arrays would fail.

src/goto-instrument/code_contracts.cpp
ArenBabikian: I was obliged to do an if statement instead. Otherwise, the regression tests for structs and arrays would fail.

src/goto-instrument/code_contracts.cpp
```diff
@@ -1542,11 +1542,31 @@ goto_programt assigns_clauset::havoc_code(
   goto_programt havoc_statements;
   for(assigns_clause_targett *target : targets)
   {
+    //    if(target != NULL) goto x;
```
SaswatPadhi: May be you meant the following?
```suggestion
    //    if(target == NULL) goto z;
```

regression/contracts/assigns_null_pointer_02/test.desc
ArenBabikian: I have added a test case and marked it as a `KNOWNBUG`.

src/goto-instrument/code_contracts.cpp
```diff
@@ -1542,11 +1542,38 @@ goto_programt assigns_clauset::havoc_code(
...
+    //      havoc_statements
+    // z: skip
+
+    // create the z label
+    goto_programt tmp_z;
+    goto_programt::targett z = tmp_z.add(goto_programt::make_skip(location));
+
+    // TODO: generalize this. Currently only supports pointers
```
SaswatPadhi: Could you add a bit more explanation to this comment?

I am not sure what generalization is expected here. What needs to be checked for other `target_type`s: `Array` and `Struct`?

cc: @feliperodri

src/goto-instrument/code_contracts.cpp
```diff
@@ -1542,11 +1542,38 @@ goto_programt assigns_clauset::havoc_code(
...
+    // See GitHub issue #6087 for further details
+    if(target->target_type == assigns_clause_targett::target_type::Scalar)
+    {
+      exprt condition = equal_exprt(
+        target->get_direct_pointer(),
+        null_pointer_exprt(
+          to_pointer_type(target->get_direct_pointer().type())));
```
SaswatPadhi: This condition would not work for invalid pointers though. I would suggest checking validity of pointers, not just `NULL`-ness.

One way to check validity in this context would be to check `__CPROVER_w_ok(target, 0)`. So, something like:
```c
      exprt condition = not_exprt(w_ok_exprt(target->get_direct_pointer(), from_integer(0, int_type)));
```
would check if `target` is invalid. (not super sure about the syntax, but I hope you get the idea)

This is inspired by a similar validity check being done within the implementation of `free`:
https://github.com/diffblue/cbmc/blob/48893287099cb5780302fe9dc415eb6888354fd6/src/ansi-c/library/stdlib.c#L250-L252

src/goto-instrument/code_contracts.cpp
```diff
@@ -1595,9 +1622,23 @@ exprt assigns_clauset::compatible_expression(
...
-        first_iter = false;
+        // TODO: generalize this. Currently only supports pointers
+        // See GitHub issue #6087 for further details
+        if(target->target_type == assigns_clause_targett::target_type::Scalar)
+        {
+          current_target_compatible = or_exprt(
+            equal_exprt(
+              called_target->get_direct_pointer(),
+              null_pointer_exprt(
+                to_pointer_type(called_target->get_direct_pointer().type()))),
```
SaswatPadhi: Same comment as before about checking validity, not just `NULL`-ness.

But, why do we have this code duplication?
cc: @feliperodri

SaswatPadhi (changes requested):

src/goto-instrument/code_contracts.cpp
feliperodri: If we use memory primitives, this will fail (see https://github.com/diffblue/cbmc/issues/6045). @tautschnig ?

SaswatPadhi: We tested 4 things, and w_ok doesn't work as expected on one of them:
- `!__CPROVER_w_ok(NULL, 0)` returns `true`
- `!__CPROVER_w_ok(ptr, 0)` returns `true` on an invalid memory location, like `ptr = 0x12345678`
- `!__CPROVER_w_ok(ptr, 0)` returns `false` on a valid (non-`NULL` result from `malloc`) memory location
- `!__CPROVER_w_ok(ptr, 0)` returns `false` on an uninitialized `ptr`, which is unexpected

src/goto-instrument/code_contracts.cpp
feliperodri: We should
1. Remove the `TODO` comment. We don't need to generalize this, it should work only for dereferences;
2. Close the GitHub issue because of (1);
3. Check for the `expr.id()` in the if statement.

ArenBabikian: @feliperodri I think point 3 may not be necessary. The variables in the assigns clause are represented as `address_of` expressions which point to the corresponding symbol, dereference, etc... As a result, I think that any kind of if statement or invariant would not be necessary.

src/goto-instrument/code_contracts.cpp
ArenBabikian: This is for the case where we are using the `--enforce-contract` flag. So for any assignment, asside fromchecking if the assigned variable is part of the assigns clause, I am also asking CBMC to consider the case that the assigned variable may be NULL/invalid.

src/goto-instrument/code_contracts.cpp
```diff
@@ -1524,11 +1524,43 @@ goto_programt assigns_clauset::havoc_code(
   goto_programt havoc_statements;
   for(assigns_clause_targett *target : targets)
   {
+    // (1) If the assigned target is not a dereference,
+    // only include the havoc_statement
+
+    // (2) If the assigned target is a dereference, do the following:
+
+    //    if( !__CPROVER_w_ok(target, 0) goto z;
```
feliperodri:
```suggestion
    // if(!__CPROVER_w_ok(target, 0)) goto z;
```

feliperodri (approved):
LGTM

regression/contracts/assigns_validity_pointer_01/test.desc
```diff
@@ -0,0 +1,13 @@
+CORE
+main.c
+--enforce-contract foo --replace-call-with-contract bar
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
```
SaswatPadhi, line 6: In all test.desc files, could you please add some regexes for expected assertions? The point is to not let the tests vacuously pass (for example if we change the contracts flag names or flags meanings)

regression/contracts/assigns_validity_pointer_01/test.desc
```diff
@@ -0,0 +1,13 @@
...
+--
+--
```
SaswatPadhi, line 24: What was the earlier behavior on invalid pointers? @feliperodri 

Did CBMC crash, or was it generating spurious assertion failures?
If it was crashing, then the `EXIT=0` regex should now make sure that we don't, but if we were generating spurious assertion failures, could you please also add some negative regexes (that are _not_ supposed to match in the output)?

regression/contracts/assigns_validity_pointer_03/test.desc
```diff
@@ -0,0 +1,18 @@
...
+Verification:
+This test checks support for an uninitialized pointer passed as a parameter
+to a function that may assign to it. The uninitialized pointer is passed to
+function bar which has been replaced by it's function contracts, while the
+calling function foo is being checked (by enforcing it's function contracts).
+
+Known Bug:
+Currently, there is a known issue with __CPROVER_w_ok(ptr, 0) such that it
+returns true if ptr is uninitialized. This is not the expected behavior,
+therefore, the outcome of this test case is currently incorrect.
```
SaswatPadhi: Missing newline at the end of file

src/goto-instrument/code_contracts.cpp
```diff
@@ -1577,8 +1609,27 @@ exprt assigns_clauset::compatible_expression(
...
       {
-        current_target_compatible =
-          target->compatible_expression(*called_target);
+        // TODO: Optimize the validation below and remove code duplication
+        // See GitHub issue #6105 for further details
+
+        // Validating the called target through __CPROVER_w_ok() is
+        // only useful when the called target is a dereference
+        if(
+          to_address_of_expr(called_target->get_direct_pointer())
```
SaswatPadhi: I would recommend creating a temporary above the `if` condition:

```c
const auto &called_target_ptr = called_target->get_direct_pointer();
```

That would make the `if` condition and the `w_ok_exprt` both slightly cleaner.

src/goto-instrument/code_contracts.cpp
```diff
@@ -1524,11 +1524,43 @@ goto_programt assigns_clauset::havoc_code(
...
+
+    //    if( !__CPROVER_w_ok(target, 0) goto z;
+    //      havoc_statements
+    // z: skip
+
+    // create the z label
+    goto_programt tmp_z;
+    goto_programt::targett z = tmp_z.add(goto_programt::make_skip(location));
+
+    if(
```
SaswatPadhi: I would recommend creating a temporary above the `if` condition:

```c
const auto &target_ptr = target->get_direct_pointer();
```

That would make many of the following expressions slightly cleaner.

regression/contracts/assigns_validity_pointer_04/main.c
```diff
@@ -0,0 +1,24 @@
+#include <assert.h>
```
SaswatPadhi, line 1: Could we also test that global pointers are correctly checked for validity?

You don't have to make a separate regression test directory, just add another function to one or more of these `main.c` files and then check for the corresponding assertion regex.

SaswatPadhi (changes requested):
Mostly LGTM, but I left a few minor comments.

Also, please check why the builds are failing.

regression/contracts/assigns_validity_pointer_01/test.desc
feliperodri: it generating spurious assertion failures, since we'd use this pointers with memory primitives.

SaswatPadhi: I see. Thanks, that makes sense.

Then we should add some negative regexes to make sure that we don't generate them any more, i.e., showing that the proposed changes actually work.

ArenBabikian: Regarding the spurious assertion failures, would the `EXIT=0` not be enough to check that we no longer generate `FAILURE`s?

Secondly, it seems that in the current implementation, negative regexes are not supported (or I might not know how to do it). But from my investigation, making sure that a regex is not present in the `test.out` file will require changing the `perl` script.

SaswatPadhi: > Would the EXIT=0 not be enough to check that we no longer generate FAILUREs?

It would be an approximation. We would catch failures, but we can't be sure we have fixed the root cause.

> it seems that in the current implementation, negative regexes are not supported

They are. Negative regexes go between the two `---` `---` you currently have. You can add as many regexes as you want between those two lines and the `perl` script would make sure that none of them are matched in the output.

src/goto-instrument/code_contracts.cpp
```diff
@@ -1524,11 +1524,42 @@ goto_programt assigns_clauset::havoc_code(
...
+    // create the z label
+    goto_programt tmp_z;
+    goto_programt::targett z = tmp_z.add(goto_programt::make_skip(location));
+
+    const auto &target_ptr = target->get_direct_pointer();
+    if(to_address_of_expr(target_ptr).object().id() == ID_dereference)
+    {
+      // create the condition
+      exprt condition = not_exprt(w_ok_exprt(
+        target->get_direct_pointer(), from_integer(0, integer_typet())));
```
SaswatPadhi: Use `target_ptr`

src/goto-instrument/code_contracts.cpp
```diff
@@ -1524,11 +1524,42 @@ goto_programt assigns_clauset::havoc_code(
...
+
+    // create havoc_statements
     for(goto_programt::instructiont instruction :
         target->havoc_code(location).instructions)
     {
       havoc_statements.add(std::move(instruction));
     }
+
+    if(
+      to_address_of_expr(target->get_direct_pointer()).object().id() ==
```
SaswatPadhi: Use `target_ptr`

src/goto-instrument/code_contracts.cpp
```diff
@@ -1577,8 +1608,26 @@ exprt assigns_clauset::compatible_expression(
...
+        // TODO: Optimize the validation below and remove code duplication
+        // See GitHub issue #6105 for further details
+
+        // Validating the called target through __CPROVER_w_ok() is
+        // only useful when the called target is a dereference
+        const auto &called_target_ptr = called_target->get_direct_pointer();
+        if(
+          to_address_of_expr(called_target_ptr).object().id() == ID_dereference)
+        {
+          current_target_compatible = or_exprt(
```
SaswatPadhi, line 1621: Please add a comment above this line:
> `or_exprt` is short-circuited, therefore `target->compatible_expression(*called_target)` would not be checked on invalid `called_target`s.

SaswatPadhi (commented):
Almost there. A couple of previous comments that are unaddressed:

SaswatPadhi (approved):
LGTM. Thanks for all the changes! 😄

chrisr-diffblue (approved):
LGTM

PR #6070, opened 2021-04-30
Avoid Doxygen warnings for MiniSat and Glucose

tautschnig:
Move the destructor to the non-template class.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #6078, opened 2021-05-05
Manage lifetime of MiniSat and Glucose solver pointers using unique_ptr

thomasspriggs:
Manage lifetime of MiniSat solver pointer using `unique_ptr`. This has the benefit of allowing the use of a default destructor rather than a hand written destructor. Note that `satcheck_minisat2_baset` was already non-copyable because its `solver_hardness` field was non-copyable. I have initially raised a PR following this approach for the MiniSat solver only, so I can check whether other maintainers see this as a reasonable approach. If this meets with approval then similar changes could be made for other solvers as well.

### DIFF
--- a/scripts/expected_doxygen_warnings.txt
+++ b/scripts/expected_doxygen_warnings.txt
@@ -1,11 +1,3 @@
-/home/runner/work/cbmc/cbmc/src/solvers/sat/satcheck_glucose.cpp:246: warning: no matching class member found for 
-  template
-  satcheck_glucose_baset< Glucose::Solver >::~satcheck_glucose_baset()
-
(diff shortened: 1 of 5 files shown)


NlightNFotis (approved):
LGTM.

tautschnig:
Thank you @thomasspriggs, this is so much better than #6070! I had made a brief attempt using `unique_ptr`, but it didn't occur to me that I (still) had to place the destructor in the .cpp file. Could you please do the same for Glucose?

thomasspriggs:
> Thank you @thomasspriggs, this is so much better than #6070! I had made a brief attempt using `unique_ptr`, but it didn't occur to me that I (still) had to place the destructor in the .cpp file. Could you please do the same for Glucose?

@tautschnig  - Done.

tautschnig (approved):

PR #6080, opened 2021-05-06
Remove no-op handling of non-existent dynamic_size expression

tautschnig:
There is no code that creates an expression with id "dynamic_size" and
the three places handling this id just did nothing.


martin-cs (approved):
Does anyone remember what this was for?  Is this handled in another way?

tautschnig:
> Does anyone remember what this was for? Is this handled in another way?

I can only imagine that it was replaced by the `dynamic_size` predicate, which of course in turn now also has become redundant.

PR #6088, opened 2021-05-06
Rfc sat backend mergesat

conp-solutions:
**This is a RFC patch to investigate whether moving the SAT backend to a more recent solver would be beneficial. This patch set is scappy. In case the integration of the new solver results in performance benefits, required changes could be worked into the used MergeSat solver, and this patch series could be cleaned up.**

I am not aware how the performance of the SAT backend can be tested. Please advice. Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
     set(CMAKE_CXX_FLAGS_RELEASE "-O2")
     set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -g")
     #   Enable lots of warnings
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wpedantic -Werror -Wno-deprecated-declarations -Wswitch-enum")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wpedantic -Wno-deprecated-declarations -Wswitch-enum")
 elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     # This would be the place to enable warnings for Windows builds, although
     # config.inc doesn't seem to do that currently
--- a/src/Makefile
+++ b/src/Makefile
@@ -119,19 +119,18 @@ DOWNLOADER = curl -L --remote-name
 TAR = tar
 
 minisat2-download:
-	@echo "Downloading Minisat 2.2.1"
+	@echo "Downloading MergeSat (replacing Minisat 2.2.1)"
 	@for i in $$(seq 1 3) ; do \
 	  $(DOWNLOADER) \
-	    http://ftp.debian.org/debian/pool/main/m/minisat2/minisat2_2.2.1.orig.tar.gz && \
+	    https://github.com/conp-solutions/mergesat/archive/refs/tags/v3.0.tar.gz && \
 	    exit 0 ; \
-		$(RM) minisat2_2.2.1.orig.tar.gz ; \
+		$(RM) v3.0.tar.gz ; \
 	  if [ $$i -lt 3 ] ; then echo "Re-trying in 10 seconds" 1>&2 ; sleep 10 ; fi ; \
 	done ; exit 1
-	@$(TAR) xfz minisat2_2.2.1.orig.tar.gz
+	@$(TAR) xfz v3.0.tar.gz # will create mergesat-3.0
 	@rm -Rf ../minisat-2.2.1
-	@mv minisat2-2.2.1 ../minisat-2.2.1
-	@(cd ../minisat-2.2.1; patch -p1 < ../scripts/minisat-2.2.1-patch)
-	@rm minisat2_2.2.1.orig.tar.gz
+	@mv mergesat-3.0 ../minisat-2.2.1
+	@rm v3.0.tar.gz
 
 cudd-download:
 	@echo "Downloading Cudd 3.0.0"
--- a/src/config.inc
+++ b/src/config.inc
@@ -5,7 +5,7 @@ BUILD_ENV = AUTO
 ifeq ($(BUILD_ENV),MSVC)
   #CXXFLAGS += /Wall /WX
 else
-  CXXFLAGS += -Wall -pedantic -Werror -Wno-deprecated-declarations -Wswitch-enum
+  CXXFLAGS += -Wall -pedantic -Wno-deprecated-declarations -Wswitch-enum
 endif
 
 ifeq ($(CPROVER_WITH_PROFILING),1)
--- a/src/solvers/Makefile
+++ b/src/solvers/Makefile
@@ -17,7 +17,7 @@ endif
 ifneq ($(MINISAT2),)
   MINISAT2_SRC=sat/satcheck_minisat2.cpp
   MINISAT2_INCLUDE=-I $(MINISAT2)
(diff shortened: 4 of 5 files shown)


CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
     set(CMAKE_CXX_FLAGS_RELEASE "-O2")
     set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O2 -g")
     #   Enable lots of warnings
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wpedantic -Werror -Wno-deprecated-declarations -Wswitch-enum")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wpedantic -Wno-deprecated-declarations -Wswitch-enum")
```
TGWDB, line 65: The policy of the current build is to error on warnings, this should probably be discussed independently of other PRs. If this is required for this PR to work, I would recommend explaining why the warnings cannot be avoided.

src/Makefile
```diff
@@ -119,19 +119,18 @@ DOWNLOADER = curl -L --remote-name
...
 	  if [ $$i -lt 3 ] ; then echo "Re-trying in 10 seconds" 1>&2 ; sleep 10 ; fi ; \
 	done ; exit 1
-	@$(TAR) xfz minisat2_2.2.1.orig.tar.gz
+	@$(TAR) xfz v3.0.tar.gz # will create mergesat-3.0
 	@rm -Rf ../minisat-2.2.1
-	@mv minisat2-2.2.1 ../minisat-2.2.1
-	@(cd ../minisat-2.2.1; patch -p1 < ../scripts/minisat-2.2.1-patch)
-	@rm minisat2_2.2.1.orig.tar.gz
+	@mv mergesat-3.0 ../minisat-2.2.1
+	@rm v3.0.tar.gz
```
TGWDB, line 133: This appears to be editing and replacing the current MiniSAT build process. I would much prefer to see a new build path for MergeSat (like the ones for CaDiCaL or glucose) and then if this has much improved performance we can simply change the default. Also see recent PRs for documenting and updating alternative build systems (e.g. #6075 #6047).

src/config.inc
```diff
@@ -5,7 +5,7 @@ BUILD_ENV = AUTO
 ifeq ($(BUILD_ENV),MSVC)
   #CXXFLAGS += /Wall /WX
 else
-  CXXFLAGS += -Wall -pedantic -Werror -Wno-deprecated-declarations -Wswitch-enum
+  CXXFLAGS += -Wall -pedantic -Wno-deprecated-declarations -Wswitch-enum
```
TGWDB, line 8: Again, this should be separate.

src/solvers/Makefile
```diff
@@ -17,7 +17,7 @@ endif
 ifneq ($(MINISAT2),)
   MINISAT2_SRC=sat/satcheck_minisat2.cpp
   MINISAT2_INCLUDE=-I $(MINISAT2)
-  MINISAT2_LIB=$(MINISAT2)/minisat/simp/SimpSolver$(OBJEXT) $(MINISAT2)/minisat/core/Solver$(OBJEXT)
+  MINISAT2_LIB=$(MINISAT2)/minisat/simp/SimpSolver$(OBJEXT) $(MINISAT2)/minisat/core/Solver$(OBJEXT) $(MINISAT2)/minisat/utils/ccnr$(OBJEXT) $(MINISAT2)/minisat/utils/Options$(OBJEXT) $(MINISAT2)/minisat/utils/System$(OBJEXT)
```
TGWDB, line 20: Again, editing the MiniSAT and not providing a new option.

src/solvers/Makefile
```diff
@@ -201,6 +201,15 @@ ifeq ($(BUILD_ENV_),MSVC)
...
 
+$(MINISAT2)/minisat/utils/ccnr$(OBJEXT): $(MINISAT2)/minisat/utils/ccnr.cc
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+
+$(MINISAT2)/minisat/utils/Options$(OBJEXT): $(MINISAT2)/minisat/utils/Options.cc
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+
+$(MINISAT2)/minisat/utils/System$(OBJEXT): $(MINISAT2)/minisat/utils/System.cc
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+
```
TGWDB, line 212: Again, editing.

TGWDB (changes requested):

conp-solutions:
Thanks for the comments. I fully understand that I should create a separate back-end. I can happily to so, in case the performance of the proposed SAT back-end exceeds the performance of the current one.

This PR is an RFC. In case we decide to move forward, I will address all the comments. I could adapt MergeSat so that dropping "-Werror" is not required any more (or provide a CBMC local patch as is done for other solvers).

Until then, I would love to understand whether there is an agreed way to measure the performance of a new back-end. Thanks!

TGWDB:
> Thanks for the comments. I fully understand that I should create a separate back-end. I can happily to so, in case the performance of the proposed SAT back-end exceeds the performance of the current one.
> 
> This PR is an RFC. In case we decide to move forward, I will address all the comments. I could adapt MergeSat so that dropping "-Werror" is not required any more (or provide a CBMC local patch as is done for other solvers).

Great.

> Until then, I would love to understand whether there is an agreed way to measure the performance of a new back-end. Thanks!

Recent performance work has explored some different benchmarks including: full runs on regression tests, runs over open source repositories (with `cbmc` validation checks), and clear test examples. An example of a PR with some performance data can be found here: https://github.com/diffblue/cbmc/pull/5964

src/Makefile
martin-cs: Yes; absolutely.

martin-cs:
@tautschnig has probably the best scripting for doing large scale performance evaluation.  But I think this might be a little bit of a red herring.

There are already a number of back-ends that often have better performance than the default one.  However MiniSAT remains the default because:

1. It's simple and requires little external dependency.
2. Its performance is remarkably consistent and scales predictably.

So, patches adding new solvers (SAT, SMT, others) are most welcome and should be merged eagerly.  Changing the default is going to require a lot of benchmarking, including against the SMT solvers and a lot of time to see if there are performance regression in actual applications.

PS Is the IPASIR format still a going concern?  That might be an easier route to integrate new solvers.

tautschnig:
Here are results for two of the categories of SV-COMP, with Mergesat on the x axis of the scatter plots, and Minisat on the y axis. As can be seen, the results vary in both directions. On other categories the results are more consistent in that Mergesat is about 10% slower, but that slowdown does not affect overall scores. Hence, no clear-cut "solver X is uniformly better," which probably is the expected result anyway.

Screenshot 2021-05-14 at 11 19 08
Screenshot 2021-05-14 at 11 19 23

What is much better about Mergesat, of course, is that it is maintained :-)

TGWDB:
@conp-solutions Just checking in to see where you are with this PR/RFC.

conp-solutions:
Feel free to close this PR. I will likely follow up with a proper new SAT backend in the future, but cannot give a strict date (i should manage to iterate within the next month.

TGWDB:
Closing as recommended by opening author. Look forward to the new SAT back end in the future.

PR #6091, opened 2021-05-07
Fix WIN32 -Werror build error

TGWDB:
This removes an unused function from the WIN32 build so that this
can still build with the -Werror compiler option.

Note that this is split from #6048.


tautschnig (changes requested):
Could the cleanup please be done slightly differently so that the function (that is a literal copy!) in src/ansi-c/ is removed and that one is used instead? I think that would also make it used, and therefore avoid the build error.

tautschnig (approved):
Thank you for cleaning this up. (And please squash the commits before merging.)

TGWDB:
> Thank you for cleaning this up. (And please squash the commits before merging.)

Happy to, and thanks for pointing out the duplication.

PR #6101, opened 2021-05-10
[RFC] Two new failing tests demonstrating updates outside member bounds

tautschnig:
These examples write to a member different from the access path being
used, yet within the object bounds.

### DIFF
--- a/regression/cbmc/member_bounds1/main.c
+++ b/regression/cbmc/member_bounds1/main.c
@@ -1,0 +1,24 @@
+#include <assert.h>
+
+#pragma pack(push, 1)
+struct S
+{
+  int a[2];
+  int x;
+};
+#pragma pack(pop)
+
+#ifdef _MSC_VER
+#  define _Static_assert(x, m) static_assert(x, m)
+#endif
+
+int main()
+{
+  int A[3];
+  _Static_assert(sizeof(A) == sizeof(struct S), "");
+  struct S *s = A;
+  s->a[2] = 42;
+  assert(*((int *)s + 2) == 42);
+  assert(s->x == 42);
+  assert(A[2] == 42);
+}
--- a/regression/cbmc/member_bounds1/test.desc
+++ b/regression/cbmc/member_bounds1/test.desc
@@ -1,0 +1,11 @@
+KNOWNBUG
+main.c
+--no-simplify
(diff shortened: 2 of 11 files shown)


tautschnig:
I'm marking this RFC as I'd appreciate some thoughts about those tests. While they work perfectly fine with both gcc and clang, I'm not sure they are covered by the C standard.

regression/cbmc/member_bounds1/main.c
```diff
@@ -0,0 +1,24 @@
+#include <assert.h>
+
+#pragma pack(push, 1)
```
martin-cs, line 3: :-\ Do we actually support this?

regression/cbmc/member_bounds1/main.c
```diff
@@ -0,0 +1,24 @@
...
+#ifdef _MSC_VER
+#  define _Static_assert(x, m) static_assert(x, m)
+#endif
+
+int main()
+{
+  int A[3];
+  _Static_assert(sizeof(A) == sizeof(struct S), "");
+  struct S *s = A;
+  s->a[2] = 42;
```
martin-cs, line 20: Pretty sure this is undefined behaviour...

martin-cs (approved):
I think this falls in the general area of "defining undefined behaviour".  The out-of-bound write is UB so, as much as ISO 9899 has anything to say... whatever we do after is correct.  I think the most important thing is that we can detect this happening.  Apart from that ... emulating popular compilers is probably not a bad idea.

regression/cbmc/member_bounds1/main.c
tautschnig: We do :-) I thought it was the most portable way of requesting packing...

martin-cs: :-|

Well... you're not wrong...

Wow.  Glad I didn't have to implement that one.

kroening:
I am afraid that "access beyond member bounds" needs to be enabled by default. It's not an uncommon idiom at all.

tautschnig:
> I am afraid that "access beyond member bounds" needs to be enabled by default. It's not an uncommon idiom at all.

I'll try to work out a fix. The proposed patch only partly addresses the problem, we'll need to disable field sensitivity when detecting such an access beyond member bounds.

PR #6108, opened 2021-05-11
Add source location and comments to assertions generated from function contracts

### LINKED ISSUE #5526: Improve error message for assigns clause
CBMC version: `>5.0`
Operating system: Ubuntu 18.04 and macOS Mojave 10.14.6
Exact command line resulting in the issue: `goto-cc main.c -o main.gb && goto-instrument main.gb main-ins.gb --enforce-all-contracts && cbmc main-ins.gb`
What behaviour did you expect: A more verbose error message.
What happened instead: `[f1.2] line 8 assertion: FAILURE`

feliperodri:
Resolves #5526.

### DIFF
--- a/regression/contracts/assigns_enforce_scoping_01/test.desc
+++ b/regression/contracts/assigns_enforce_scoping_01/test.desc
@@ -3,8 +3,9 @@ main.c
 --enforce-all-contracts
 ^EXIT=10$
 ^SIGNAL=0$
-assertion: SUCCESS$
-assertion: FAILURE$
+^\[f1.\d+\] line \d+ Check that b is assignable: SUCCESS$
+^\[f1.\d+\] line \d+ Check that \*b is assignable: SUCCESS$
+^\[f1.\d+\] line \d+ Check that \*b is assignable: FAILURE$
 ^VERIFICATION FAILED$
 --
 --
--- a/regression/contracts/quantifiers-forall-ensures-enforce/test.desc
+++ b/regression/contracts/quantifiers-forall-ensures-enforce/test.desc
@@ -3,7 +3,10 @@ main.c
 --enforce-all-contracts
 ^EXIT=0$
 ^SIGNAL=0$
-^\[1\] assertion: SUCCESS$
+^\[postcondition.\d+\] file main.c line \d+ Check ensures clause: SUCCESS$
+^\[f1.\d+\] line \d+ Check that i is assignable: SUCCESS$
+^\[f1.\d+\] line \d+ Check that i is assignable: SUCCESS$
+^\[f1.\d+\] line \d+ Check that arr\[\(.*\)i\] is assignable: SUCCESS$
 ^VERIFICATION SUCCESSFUL$
 --
 ^warning: ignoring
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -441,6 +441,11 @@ bool code_contractst::apply_function_contract(
       code_assertt(requires),
       assertion,
       symbol_table.lookup_ref(function).mode);
+    assertion.instructions.front().source_location = requires.source_location();
+    assertion.instructions.back().source_location.set_comment(
+      "Check requires clause");
+    assertion.instructions.front().source_location.set_property_class(
+      ID_precondition);
     auto lines_to_iterate = assertion.instructions.size();
     goto_program.insert_before_swap(target, assertion);
     std::advance(target, lines_to_iterate);
@@ -563,7 +568,8 @@ void code_contractst::instrument_assign_statement(
   alias_assertion.add(goto_programt::make_assertion(
     assigns_clause.alias_expression(lhs),
     instruction_iterator->source_location));
-
+  alias_assertion.instructions.back().source_location.set_comment(
+    "Check that " + from_expr(ns, lhs.id(), lhs) + " is assignable");
   int lines_to_iterate = alias_assertion.instructions.size();
   program.insert_before_swap(instruction_iterator, alias_assertion);
   std::advance(instruction_iterator, lines_to_iterate);
@@ -626,6 +632,9 @@ void code_contractst::instrument_call_statement(
     goto_programt alias_assertion;
     alias_assertion.add(goto_programt::make_assertion(
       alias_expr, instruction_iterator->source_location));
+    alias_assertion.instructions.back().source_location.set_comment(
+      "Check that " + from_expr(ns, alias_expr.id(), alias_expr) +
+      " is assignable");
     program.insert_before_swap(instruction_iterator, alias_assertion);
     ++instruction_iterator;
   }
@@ -661,6 +670,8 @@ void code_contractst::instrument_call_statement(
     goto_programt alias_assertion;
     alias_assertion.add(goto_programt::make_assertion(
       compatible, instruction_iterator->source_location));
+    alias_assertion.instructions.back().source_location.set_comment(
+      "Check compatibility of assigns clause with the called function");
     program.insert_before_swap(instruction_iterator, alias_assertion);
     ++instruction_iterator;
   }
@@ -1003,8 +1014,13 @@ void code_contractst::add_contract_check(
 
     // get all the relevant instructions related to history variables
     auto assertion = code_assertt(ensures);
+    assertion.add_source_location() = ensures.source_location();
     ensures_pair = create_ensures_instruction(
       assertion, ensures.source_location(), wrapper_fun, function_symbol.mode);
+    ensures_pair.first.instructions.back().source_location.set_comment(
+      "Check ensures clause");
+    ensures_pair.first.instructions.front().source_location.set_property_class(
+      ID_postcondition);
 
     // add all the history variable initializations
     check.destructive_append(ensures_pair.second);
--- a/regression/contracts/assigns_enforce_scoping_02/test.desc
+++ b/regression/contracts/assigns_enforce_scoping_02/test.desc
@@ -3,8 +3,8 @@ main.c
 --enforce-all-contracts
 ^EXIT=10$
 ^SIGNAL=0$
-assertion: SUCCESS$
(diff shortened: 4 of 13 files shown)


src/goto-instrument/code_contracts.cpp
```diff
@@ -876,6 +876,9 @@ void code_contractst::add_contract_check(
     goto_programt assertion;
     convert_to_goto(
       code_assertt(ensures_cond), function_symbol.mode, assertion);
+    assertion.instructions.front().source_location.set_line(ensures.source_location().get_line());
+    assertion.instructions.back().source_location.set_comment("ensures clause");
```
kroening: What about the file name?

martin-cs (approved):
Is this the only place where something like this happens?  It feels like there are likely to be other cases of this in the code contract implementation.

src/goto-instrument/code_contracts.cpp
feliperodri: Yep, I added it.

src/goto-instrument/code_contracts.cpp
```diff
@@ -436,11 +436,17 @@ bool code_contractst::apply_function_contract(
     code_contractst::add_quantified_variable(requires, replace, mode);
     replace(requires);
 
+    //goto_programt assertion;
```
SaswatPadhi: Should this be removed?

SaswatPadhi (commented):
As part of this PR, could we also add source location and comments to assigns clause? (#5526)

src/goto-instrument/code_contracts.cpp
feliperodri: Yes.

src/goto-instrument/code_contracts.cpp
```diff
@@ -563,7 +568,8 @@ void code_contractst::instrument_assign_statement(
   alias_assertion.add(goto_programt::make_assertion(
     assigns_clause.alias_expression(lhs),
     instruction_iterator->source_location));
-
+  alias_assertion.instructions.back().source_location.set_comment(
+    "Check assigns clause");
```
SaswatPadhi: I think the `Check ensures clause` and `Check requires clause` messages are fine, because each function currently can have at most one ENSURES clause and at most one REQUIRES clause.

But multiple `Check assigns clause` assertions seem less useful because it's not clear which variable each assertion is looking at.

There might be a better way to do this, but this could be one way:

```cpp
alias_assertion.instructions.back().source_location.set_comment(
    "Check that " + lhs.pretty(0) + " is assignable");
```

@tautschnig Do you have any suggestions on what we should use here to `string`-ify `lhs` here? Is `pretty` the right thing to use?

SaswatPadhi (commented):
One comment below.

And please check the assertions introduced at lines 634 and 669. Those don't have any comments, but look related to assigns clause?

src/goto-instrument/code_contracts.cpp
SaswatPadhi: As we discussed offline, `pretty` seems to generate a verbose string (with lots of debug info).

Based on a quick search through the code base, I think `from_expr` (from `#include <langapi/language_util.h>`) could be the used here, e.g. https://github.com/diffblue/cbmc/blob/1ab5de1ac0893c7ca94ffe5463e2c7c4ee781266/src/goto-programs/builtin_functions.cpp#L752-L753

feliperodri: @SaswatPadhi I also tried `from_expr`, but at the end it also uses pretty and generates the same verbose strings.
https://github.com/diffblue/cbmc/blob/1ab5de1ac0893c7ca94ffe5463e2c7c4ee781266/src/langapi/language.cpp#L32-L39

I'm now converting `lhs` to a symbol and extracting the identifier, which results to the following message
```console
main.c function bar
[bar.1] line 4 Check that bar::y is assignable: SUCCESS
```

SaswatPadhi: The above `from_expr` definition looks suspicious, I don't think that's the one being called from the `set_comment` example I showed.

The `from_expr` call should return a `std::string` but the definition you posted above returns a `bool`. Could you please double check?

It's being used within `set_comment`, so if it prints a verbose output then that should break the CBMC output ...

SaswatPadhi: I believe this is the right definition:

https://github.com/diffblue/cbmc/blob/48893287099cb5780302fe9dc415eb6888354fd6/src/langapi/language_util.cpp#L20-L31

Also try calling:
https://github.com/diffblue/cbmc/blob/48893287099cb5780302fe9dc415eb6888354fd6/src/langapi/language_util.cpp#L59-L63

The problem with your "symbol extraction" approach would be that not all assignment targets would be symbols. We also have derefs and struct members, for example.

SaswatPadhi: Oh may be you're right, `p->from_expr` probably resolves to the definition you mentioned ... 

If that's the case, then my last suggestion would be to try `expr2c` function from `#include <ansi-c/expr2c.h>`. 

If that doesn't work either, then let's wait to hear from @tautschnig. But I wonder why `from_expr` is used within `set_comments` in the codebase, if it prints a debug output, not a small inline expression ...

SaswatPadhi: Okay, this should be called, not the abstract definition that you posted above.

https://github.com/diffblue/cbmc/blob/1ab5de1ac0893c7ca94ffe5463e2c7c4ee781266/src/ansi-c/ansi_c_language.cpp#L146-L153

This does call `expr2c` and not `pretty`.

Basically in this case the `get_language_from_identifier` should resolve to C language (i.e. `ansi_c_languaget`), not the abstract `languaget`. If that doesn't happen, let's look into the root cause.

feliperodri: Thanks for the tip! It worked perfectly.

regression/contracts/assigns_enforce_scoping_01/test.desc
```diff
@@ -3,8 +3,9 @@ main.c
 --enforce-all-contracts
 ^EXIT=10$
 ^SIGNAL=0$
-assertion: SUCCESS$
-assertion: FAILURE$
+^\[f1.\d+\] line \d+ Check that b is assignable: SUCCESS$
+^\[f1.\d+\] line \d+ Check that \*b is assignable: SUCCESS$
+^\[f1.\d+\] line \d+ Check that \*b is assignable: FAILURE$
```
SaswatPadhi, line 8: Were we generating 3 assertions before this change as well?

The function never assigns to `b`, only to `*b`, so why are we checking:
```
^\[f1.\d+\] line \d+ Check that b is assignable: SUCCESS$
```

Is it because of this line:

```c
int *b = a;
```

If so, is it necessary to check assigns clause at local variable creation? I mean they would never violate the assigns clause right? And probably shouldn't even be checked, just for efficiency.

If we agree and I'm not missing something here, then let's create a ticket for this and fix at some point?

src/goto-instrument/code_contracts.cpp
```diff
@@ -563,7 +568,11 @@ void code_contractst::instrument_assign_statement(
   alias_assertion.add(goto_programt::make_assertion(
     assigns_clause.alias_expression(lhs),
     instruction_iterator->source_location));
-
+  std::string source_expr_string;
+  (get_language_from_identifier(ns, lhs.id()))
+    ->from_expr(lhs, source_expr_string, ns);
```
SaswatPadhi: These 3 lines look like we are inlining the definition of https://github.com/diffblue/cbmc/blob/48893287099cb5780302fe9dc415eb6888354fd6/src/langapi/language_util.cpp#L20-L31

Could you please explain what is the issue with doing the following (after `#include`ing the appropriate headers):
```c
    "Check that " + from_expr(ns, lhs.id(), lhs) + " is assignable"
```

regression/contracts/assigns_enforce_scoping_01/test.desc
feliperodri: We currently check ALL assignments in the function body, even if they are over local variables. I agree with you that this is unnecessary and we could prune these cases for efficiency. See https://github.com/diffblue/cbmc/issues/6154

regression/contracts/quantifiers-forall-ensures-enforce/test.desc
```diff
@@ -3,7 +3,10 @@ main.c
 --enforce-all-contracts
 ^EXIT=0$
 ^SIGNAL=0$
-^\[1\] assertion: SUCCESS$
+^\[postcondition.\d+\] file main.c line \d+ Check ensures clause: SUCCESS$
+^\[f1.\d+\] line \d+ Check that i is assignable: SUCCESS$
+^\[f1.\d+\] line \d+ Check that i is assignable: SUCCESS$
+^\[f1.\d+\] line \d+ Check that arr\[\(signed long int\)i\] is assignable: SUCCESS$
```
SaswatPadhi: Ah on Windows this is `signed long long int`:

```
[f1.2] line 18 Check that arr[(signed long long int)i] is assignable: SUCCESS
```

I would just make this:

```
^\[f1.\d+\] line \d+ Check that arr\[\(.*\)i\] is assignable: SUCCESS$
```

regression/contracts/assigns_enforce_scoping_01/test.desc
SaswatPadhi: Great, thanks for creating the issue. Resolving this comment.

src/goto-instrument/code_contracts.cpp
feliperodri: Perfect! Implemented as suggested.

regression/contracts/quantifiers-forall-ensures-enforce/test.desc
feliperodri: Fixed it!

SaswatPadhi (approved):
Nice! These comments look great :)

PR #6109, opened 2021-05-12
Restructure compilation documentation

thomasspriggs:
This PR restructures the compilation documentation. We have received some feedback that the compilation instructions are difficult to follow for new users. This PR makes things more straight forward for new users to follow.

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -42,7 +42,7 @@ files.
      ```
      xcode-select --install
      ```
-     You shoud also install [Homebrew](https://brew.sh), after which you can
+     You should also install [Homebrew](https://brew.sh), after which you can
      run `brew install cmake` to install CMake.
    - On platforms where installing the Java Development Kit and Maven is
      difficult, you can avoid needing these dependencies by not building
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -50,7 +50,7 @@ files.
      run `brew install cmake` to install CMake.
    - On platforms where installing the Java Development Kit and Maven is
      difficult, you can avoid needing these dependencies by not building
-     JBMC. Just pass `-DWITH_JBMC=OFF` to cmake in step (4) below.
+     JBMC. Just pass `-DWITH_JBMC=OFF` to `cmake` in step (4) below.
    - On Windows, ensure you have Visual Studio 2019 or later installed. The
      developer command line that comes with Visual Studio 2019 has CMake
      already available. You will also need to ensure that you have winflexbison
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -5,7 +5,7 @@ following environments:
 
 - Linux
 - MacOS X
-- Microsoft Visual Studio 2019
+- Windows
 
 The above environments are currently tested as part of our continuous
 integration system. It separately tests both the CMake build system and the
@@ -17,7 +17,6 @@ past, but are not actively tested:
 
 - Solaris 11
 - FreeBSD 11
-- Cygwin
 
 # Building using CMake
 
@@ -202,66 +201,6 @@ Follow these instructions:
    make -C jbmc/src setup-submodules
    make -C jbmc/src
    ```
-
-## COMPILATION ON WINDOWS WITH CYGWIN
-
-There are two options: the Visual Studio compiler with version 14 (2015) or
-later, or the MinGW cross compiler with version 5.4 or later.
-We recommend Visual Studio.
-
-Follow these instructions:
-
-1. First install Cygwin, then from the Cygwin setup facility install the
-   following packages: `flex, bison, tar, gzip, git, make, wget, patch,
-   curl`.
-2. Get the CBMC source via
-   ```
-   git clone https://github.com/diffblue/cbmc cbmc-git
-   cd cbmc-git
-   ```
-3. Depending on your choice of compiler:
-   1. To compile with Visual Studio, change the second line of `src/config.inc`
-      to
-      ```
-      BUILD_ENV = MSVC
-      ```
-      Open the Developer Command Prompt for Visual Studio, then start the
-      Cygwin shell with
-      ```
-      bash.exe -login
-      ```
-      Please note that this might open a different shell instead, especially if
-      you have installed other Linux subsystems previously. To verify that you
-      are in the correct shell, make sure that the Windows file system can be
-      accessed via the folder`/cygdrive`. If the command above does not open
-      the Cygwin shell, you can also access it by using its absolute path,
-      `C:\cygwin64\bin\bash.exe` by default. In the Developer Command Prompt, 
-      simply type
-      ```
-      C:\cygwin64\bin\bash.exe -login
-      ```
-   2. To compile with MinGW, use Cygwin setup to install a mingw g++ compiler
-      package, i.e. one of `mingw{32,64}-{x86_64,i686}-gcc-g++`. You may also
-      have to adjust the section in `src/common` that defines `CC` and `CXX`
-      for BUILD_ENV = Cygwin.
-      Then start the Cygwin shell.
-4. To compile CMBC, open the Cygwin shell and type
-   ```
-   make -C src DOWNLOADER=wget minisat2-download
-   make -C src
-   ```
-5. To compile JBMC, you additionally need the JDK and Maven 3, which have
-   to be installed manually. Then open the Cygwin shell and type
-   ```
-   make -C jbmc/src setup-submodules
-   make -C jbmc/src
-   ```
-
-(Optional) A Visual Studio project file can be generated with the script
-"generate_vcxproj" that is in the subdirectory "scripts".  The project file is
-helpful for GUI-based tasks, e.g., the class viewer, debugging, etc., and can
-be used for building with MSBuild.  Note that you still need to run flex/bison
-using "make generated_files" before opening the project.
 
 ## COMPILATION ON SOLARIS 11
 
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -5,7 +5,7 @@ following environments:
 
 - Linux
 - MacOS X
-- Microsoft Visual Studio
+- Microsoft Visual Studio 2019
 
 The above environments are currently tested as part of our continuous
 integration system. The environments below have been used successfully in the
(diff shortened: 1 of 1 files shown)


COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+past, but are not actively tested:
+
 - Solaris 11
 - FreeBSD 11
 - Cygwin
-- Microsoft Visual Studio
 
-The rest of this document is split up into three parts: compilation on Linux,
-MacOS, Windows.  Please read the section appropriate for your machine.
+# Building using CMAKE
```
NlightNFotis: Would it be possible we have `CMake` consistently capitalised?

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+     difficult, you can avoid needing these dependencies by not building
+     JBMC. Just pass `-DWITH_JBMC=OFF` to cmake in step (4) below.
+   - On Windows, ensure you have Visual Studio 2019 or later installed. The
+     developer command line that comes with Visual Studio 2019 has cmake
+     already available. You will also need to ensure that you have winflexbison
+     installed and available in the path. winflexbison is available from
+     [the github release page](https://github.com/lexxmark/winflexbison/releases/)
+     or through
+     [the chocolatey package manager.](https://community.chocolatey.org/packages/winflexbison3)
+     Installing strawberryperl is advised if you want to run tests on Windows.
```
NlightNFotis: Can `strawberryperl` be added into a codeblock to indicate that it's a package name as well? Or add instructions for it with chocolatey as an addendum?

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+   - Use of CMake has not been tested on Solaris or FreeBSD. However, it should
+     be possible to install CMake from the system package manager or the
+     [official download page](https://cmake.org/download) on those systems.
+     The dependencies (as listed in the relevant sections above) will also be
+     required, and should be installed using the suggested methods.
+2. Navigate to the *top level* folder of the project. This is different from
```
NlightNFotis, line 69: Maybe add the second sentence as a bullet point, or add the paths in parentheses (`cbmc` vs `cbmc/src/`) to make it a bit less confusing?

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+   `-DCMAKE_BUILD_TYPE=Debug` (or `RelWithDebInfo`) to the initial invocation.
+   The default is to perform an optimized build via the `Release` configuration.
+
+   If you *do* need to manually add flags, use `-DCMAKE_CXX_FLAGS=...` and
+   `-DCMAKE_EXE_LINKER_FLAGS=...`. This is useful for enabling clang's
+   sanitizers.
+
+   Finally, to enable building universal binaries on macOS, you can pass the
+   flag `-DCMAKE_OSX_ARCHITECTURES=i386;x86_64`. If you don't supply this flag,
+   the build will just be for the architecture of your machine.
```
NlightNFotis: Would it be better phrased as _The build target is going to be the local machine's architecture_?

NlightNFotis (approved):
LGTM, with some minor comments.

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+integration system. It separately tests both the CMake build system and the
+hand-written make files. The latest build steps being used in CI can be
+[referenced here](https://github.com/diffblue/cbmc/blob/develop/.github/workflows/pull-request-checks.yaml).
+
+The environments below have been used successfully in the
+past, but are not actively tested:
+
 - Solaris 11
 - FreeBSD 11
 - Cygwin
```
tautschnig: Hmm, this is (already was!) a somewhat odd mix of operating systems and build environments. Perhaps say "Windows" instead of "Visual Studio" and drop "Cygwin"? There surely are tool chains for Linux that we are _not_ testing.

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+following environments:
 
 - Linux
 - MacOS X
+- Microsoft Visual Studio 2019
+
+The above environments are currently tested as part of our continuous
+integration system. It separately tests both the CMake build system and the
+hand-written make files. The latest build steps being used in CI can be
+[referenced here](https://github.com/diffblue/cbmc/blob/develop/.github/workflows/pull-request-checks.yaml).
```
tautschnig, line 13: Should probably say "found here" instead of "referenced here?"

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
 - Cygwin
-- Microsoft Visual Studio
 
-The rest of this document is split up into three parts: compilation on Linux,
-MacOS, Windows.  Please read the section appropriate for your machine.
+# Building using CMAKE
+
+Building with CMake is supported across Linux, MacOS X and Windows with Visual
+Studio 2019. There are also hand-written make files which can be used to build
+separate binaries independently. Usage of the hand-written make files is
```
tautschnig, line 25: What does "separate binaries independently" mean?

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+explained in a separate section. The CMake build can build the complete
```
tautschnig, line 26: Could that "separate section" be named/referenced here?

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+repository in fewer steps and supports better integration with various IDEs and
```
tautschnig, line 27: How did you find out that it was "fewer steps?"

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+     ```
+   - On macOS, do
+     ```
+     xcode-select --install
+     ```
+     You should also install [Homebrew](https://brew.sh), after which you can
+     run `brew install cmake` to install CMake.
+   - On platforms where installing the Java Development Kit and Maven is
+     difficult, you can avoid needing these dependencies by not building
+     JBMC. Just pass `-DWITH_JBMC=OFF` to cmake in step (4) below.
```
tautschnig: The above commands didn't install maven or a JDK! Also, having this item right here is a bit strange for both the ones before and after are about installing operating-system specific dependencies.

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+   flag `-DCMAKE_OSX_ARCHITECTURES=i386;x86_64`. If you don't supply this flag,
+   the build will just be for the architecture of your machine.
+5. Run the build:
+   ```
+   cmake --build build
+   ```
+   This command tells CMake to invoke the correct tool to run the build in the
+   `build` directory. You can also use the build back-end directly by invoking
+   `make`, `ninja`, or opening the generated IDE project as appropriate. The
+   complete set of built binaries can be found in `./build/bin` once the build
```
tautschnig, line 118: Nit picking: I'd probably use `build/bin/` instead of `./build/bin`.

COMPILING.md
```diff
@@ -1,19 +1,125 @@
...
+   is complete.
```
tautschnig, line 119: Perhaps add a comment about doing parallel builds in here? Most people will have multiple cores available these days, and the build process is a lot faster when using more than one core...

tautschnig (approved):
Thank you for improving the documentation!

COMPILING.md
```diff
@@ -15,10 +15,106 @@ past, but are not actively tested:
...
+   - On Red Hat/Fedora or derivates, do
+     ```
+     dnf install gcc gcc-c++ flex bison curl patch cmake
+     ```
+   - On macOS, do
+     ```
+     xcode-select --install
+     ```
+     You shoud also install [Homebrew](https://brew.sh), after which you can
+     run `brew install cmake` to install CMake.
```
TGWDB: Also similar to tautshnig's comment below, why is MacOS different in style and arguments to the others. I think it would be better to have single (displayed) comment for each system and let that do it all. If users want to play around, this can (probably) be covered elsewhere. (Note, I'm reviewing by commit, so they may be a little haphazard depending on the commits.)

COMPILING.md
```diff
@@ -15,10 +15,106 @@ past, but are not actively tested:
...
+   cmake --build build
+   ```
+   This command tells CMake to invoke the correct tool to run the build in the
+   `build` directory. You can also use the build back-end directly by invoking
+   `make`, `ninja`, or opening the generated IDE project as appropriate.
+
+# Building using hand-written make files.
+
+The rest of this section is split up based on the platform being built on. 
+Please read the section appropriate for your machine.
```
TGWDB, line 124: "machine" -> "platform"

COMPILING.md
```diff
@@ -1,14 +1,19 @@
 # WHAT ARCHITECTURE?
 
-CPROVER now needs a C++11 compliant compiler and works in the following
-environments:
+CPROVER now needs a C++11 compliant compiler and is known to work in the
+following environments:
 
 - Linux
 - MacOS X
+- Microsoft Visual Studio
```
TGWDB: This should be >=2019, also below should probably indicate MSVS 2015+-?

COMPILING.md
TGWDB: I think this might be clearer as something like "can be used to build (sub) components of the cbmc package" or something like this? I agree the current version is not clear (I don't know if mi version is any better). But the spirit I think is we could build `goto-diff` or `jbmc` without building all the others?

COMPILING.md
thomasspriggs: Done.

COMPILING.md
thomasspriggs: Done.

COMPILING.md
thomasspriggs: I have rephrased this, but slightly differently.

COMPILING.md
thomasspriggs: I have dropped cygwin.

COMPILING.md
tautschnig: But ... that's no different with cmake, is it? You can easily build individual components by just specifying them as targets (e.g., I sometimes do `ninja -C build cbmc` when I don't care about building all the rest).

TGWDB: I agree. I'm guessing at what the wording meant to convey.

PR #6111, opened 2021-05-12
Visual Studio compilation documentation

NlightNFotis:
This PR serves two purposes:

1. Add documentation on building with Visual Studio on Windows to the `COMPILING.md` file.
2. Do some further minor cleanup on top of #6109 and implement some suggestions from the
    review comments there.

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -253,7 +253,27 @@ Maven 3 manually.
    gmake -C jbmc/src
    ```
 
-# WORKING WITH ECLIPSE
+# Working with IDEs
+
+## Working with Visual Studio on Windows
+
+Follow these instructions to work on Windows with Visual Studio:
+
+1. Open the `cbmc` folder in Visual Studio. Visual Studio 2017 and
+   later have automated support for CMake projects, so you need to
+   give sometime to Visual Studio to index the project and load its
+   own plugins, and then it's going to be ready to build `cbmc`.
+2. Once indexing and plugin loading has finished, a menu `Build`
+   should have appeared on the top bar. From there, select `Build All`.
+3. After the build has finished, there should be a folder `out` present.
+   Navigating `out/build/<build_parameters>/bin` should get you to the
+   binaries and other artifacts built, with `<build_parameters>` corresponding
+   to something like `x64-Debug (Default)` or whatever the equivalent is
+   for your system.
+
+The above instructions have been tested against Visual Studio 2019.
+
+## Working with Eclipse
 
 To work with Eclipse, do the following:
 
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -10,7 +10,7 @@ following environments:
 The above environments are currently tested as part of our continuous
 integration system. It separately tests both the CMake build system and the
 hand-written make files. The latest build steps being used in CI can be
-[referenced here](https://github.com/diffblue/cbmc/blob/develop/.github/workflows/pull-request-checks.yaml).
+[found here](https://github.com/diffblue/cbmc/blob/develop/.github/workflows/pull-request-checks.yaml).
 
 The environments below have been used successfully in the
 past, but are not actively tested:
@@ -22,7 +22,7 @@ past, but are not actively tested:
 
 Building with CMake is supported across Linux, MacOS X and Windows with Visual
 Studio 2019. There are also hand-written make files which can be used to build
-separate binaries independently. Usage of the hand-written make files is
+separate targets independently. Usage of the hand-written make files is
 explained in a separate section. The CMake build can build the complete
 repository in fewer steps and supports better integration with various IDEs and
 static-analysis tools. On Windows, the CMake build has the advantage of not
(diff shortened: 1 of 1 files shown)


COMPILING.md
```diff
@@ -114,6 +114,25 @@ files.
...
+   own plugins, and then it's going to be ready to build `cbmc`.
+3. Once indexing and plugin loading has finished, a menu `Build`
+   should have appeared on the top bar. From there, select `Build All`.
+4. After the build has finished, there should be a folder `out` present.
+   Navigating `out/build/<build_parameters>/bin` should get you to the
+   binaries and other artifacts built, with `<build_parameters>` corresponding
+   to something like `x64-Debug (Default)` or whatever the equivalent is
+   for your system.
+
+The above instructions hae been tested against Visual Studio 2019.
```
TGWDB: Above says VS 2017, which version is it?

TGWDB (approved):

COMPILING.md
NlightNFotis: I have tested it with Visual Studio 2019, but 2017 is the first version that added support for this.

Do I convey this accurately through the text?

COMPILING.md
```diff
@@ -114,6 +114,25 @@ files.
    complete set of built binaries can be found in `./build/bin` once the build
    is complete.
 
+## Compilation on Windows using Visual Studio
+
+Follow these instructions to work on Windows with Visual Studio:
```
tautschnig: It would be useful to make clear whether that's on top of the above instructions, or a replacement of (some of) the steps.

COMPILING.md
```diff
@@ -114,6 +114,25 @@ files.
...
+
+1. First, clone the `cbmc` repository locally.
```
tautschnig: Fair call, but other build instructions don't suggest that. We should be consistent about that.

COMPILING.md
```diff
@@ -114,6 +114,25 @@ files.
...
+2. Open the `cbmc` folder into Visual Studio. Visual Studio 2017 and
```
tautschnig: English isn't my first language, so take this with a grain of salt: I think it should be "in," not "into." Perhaps someone whose first language _is_ English can do a proof read?

COMPILING.md
tautschnig: Nit pick: s/hae/have/

src/goto-checker/symex_coverage.cpp
```diff
@@ -14,6 +14,7 @@ Date: March 2016
 #include "symex_coverage.h"
 
 #include <chrono>
+#include <ctime>
```
tautschnig: Thank you for fixing this! Would you mind moving it to a PR of its own, I'll happily add my approval as soon as I get to see the PR. And: my apologies for messing this up.

tautschnig (approved):

src/goto-checker/symex_coverage.cpp
NlightNFotis: Yeah, no problem - I will spin this out to its own PR now.

Also, yeah it's hard to notice subtle breakages like this - I only discovered it yesterday when I played around with VS on windows to write these docs.

COMPILING.md
TGWDB: Yes, "into" -> "in"

COMPILING.md
thomasspriggs: I am OK with the two different version numbers above, as the versions match the facts. Maybe add 2019 to the section title though.

COMPILING.md
```diff
@@ -114,6 +114,25 @@ files.
    complete set of built binaries can be found in `./build/bin` once the build
    is complete.
 
+## Compilation on Windows using Visual Studio
```
thomasspriggs: :pick: I don't think this section should really be just about *compilation* with Visual Studio as much as its use as an IDE. As such I suggest that this new section be moved be moved just above the existing section on "WORKING WITH ECLIPSE" and that both sections be made subsections of a new "Working on CBMC using IDEs" section.

:bulb: We could also add a new "Working with CLion" section as a later enhancement (as a separate piece of work in a potential follow-up PR).

COMPILING.md
thomasspriggs: :pick: This isn't a Visual Studio specific instruction. Visual Studio 2019 features git integration. So it would be great to include instructions on cloning CBMC using that integration - [Visual Studio git documentation](https://docs.microsoft.com/en-us/visualstudio/version-control/git-with-visual-studio?view=vs-2019#clone-an-existing-git-repository)

thomasspriggs (commented):
Needs a rebase to get rid of the commits which have already been merged as separate PRs.

NlightNFotis:
Hi @tautschnig, can you please re-review the second commit in this PR?

We had to merge #6109 before we could incorporate your suggested fixes, so this is now including them.

TGWDB (approved):

thomasspriggs (approved):

PR #6118, opened 2021-05-14
Use simplify_exprtt::resultt in pre-order simplification steps

tautschnig:
The use of resultt increases type safety as the expression to be
simplified is no longer modified in place. All post-order simplification
steps already use resultt, but pre-order steps had been left to be
done.

### DIFF
--- a/src/util/simplify_expr_if.cpp
+++ b/src/util/simplify_expr_if.cpp
@@ -211,47 +211,66 @@ bool simplify_exprt::simplify_if_cond(exprt &expr)
   return no_change;
 }
 
-bool simplify_exprt::simplify_if_preorder(if_exprt &expr)
+static simplify_exprt::resultt<> build_if_expr(
+  const if_exprt &expr,
+  simplify_exprt::resultt<> cond,
+  simplify_exprt::resultt<> truevalue,
+  simplify_exprt::resultt<> falsevalue)
 {
-  exprt &cond = expr.cond();
-  exprt &truevalue = expr.true_case();
-  exprt &falsevalue = expr.false_case();
+  if(
+    !cond.has_changed() && !truevalue.has_changed() &&
+    !falsevalue.has_changed())
+  {
+    return simplify_exprt::resultt<>(
+      simplify_exprt::resultt<>::UNCHANGED, expr);
+  }
+  else
+  {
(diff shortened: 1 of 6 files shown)


martin-cs (approved):

src/util/simplify_expr_if.cpp
```diff
@@ -260,21 +279,18 @@ bool simplify_exprt::simplify_if_preorder(if_exprt &expr)
...
 
-    auto r_truevalue = simplify_rec(truevalue);
-    if(r_truevalue.has_changed())
-    {
-      truevalue = r_truevalue.expr;
-      no_change = false;
-    }
+    auto r_truevalue = simplify_rec(swap_branches ? falsevalue : truevalue);
+    if(swap_branches)
+      r_truevalue.expr_changed = CHANGED;
```
kroening, line 285: Why is this needed? The if-expression will be marked as changed anyway.

src/util/simplify_expr_if.cpp
```diff
@@ -283,48 +299,36 @@ bool simplify_exprt::simplify_if_preorder(if_exprt &expr)
...
 
-    auto r_falsevalue = simplify_rec(falsevalue);
-    if(r_falsevalue.has_changed())
-    {
-      falsevalue = r_falsevalue.expr;
-      no_change = false;
-    }
+    auto falsevalue = simplify_rec(swap_branches ? falsevalue : truevalue);
+    if(swap_branches)
+      r_falsevalue.expr_changed = CHANGED;
```
kroening, line 303: Same here

src/util/simplify_expr_if.cpp
tautschnig: Yes, but `build_if_expr` would not actually replace the `truevalue`.

src/util/simplify_expr_if.cpp
tautschnig: `build_if_expr` would not actually replace the `falsevalue`.

src/util/simplify_expr_if.cpp
tautschnig: I have refactored the code and added a comment.

kroening (approved):

PR #6121, opened 2021-05-14
Concurrency: do not remove shared or dirty pointers from the value set

tautschnig:
Just like we always add (and never replace) shared pointers in
goto_symex_statet::assignment, symex_dead should not remove shared or
dirty pointers.

This fixes a first problem, with further problems remaining as
indicated by the KNOWNBUG regression test.


martin-cs (approved):
Approve but if you have time to look at the `KNOWNBUG` before merging, that would be great.

tautschnig:
> Approve but if you have time to look at the `KNOWNBUG` before merging, that would be great.

Oh, my apologies for failing to make that clear: that's fixed in #6122.

kroening (approved):

PR #6123, opened 2021-05-14
Concurrency: treat updates to an unknown field as atomic

### LINKED ISSUE #123: bug in concurrency encoding for arrays and structs
regression tests coming...

tautschnig:
Writing to an object that is composed of multiple fields is generally
treated in a field-sensitive (and thus: single-field) manner, which had
fixed several issues in handling shared arrays or structs. When the
specific field being updated cannot be determined during symbolic
execution (as is the case with non-const indices into arrays), we update
all fields that make up the full object. Such a full-object update must
then be treated as an atomic operation to avoid losing intermittent
writes performed by other threads.

Fixes: #123

### DIFF
--- a/src/goto-symex/goto_symex.cpp
+++ b/src/goto-symex/goto_symex.cpp
@@ -117,10 +117,28 @@ void goto_symext::symex_assign(
         return;
     }
 
+    // We may end up reading (and writing) all components of an object composed
+    // of multiple fields. In such cases, we must do so atomically to avoid
+    // overwriting components modified by another thread. Note that this also
+    // implies multiple shared reads on the rhs being treated as atomic.
+    const bool maybe_divisible =
+      lhs.id() == ID_index ||
+      (is_ssa_expr(lhs) &&
+       state.field_sensitivity.is_divisible(to_ssa_expr(lhs), false));
+    const bool need_atomic_section = maybe_divisible &&
+                                     state.threads.size() > 1 &&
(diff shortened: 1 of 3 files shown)


TGWDB (approved):

src/goto-symex/goto_symex.cpp
```diff
@@ -114,9 +114,27 @@ void goto_symext::symex_assign(
         return;
     }
 
+    // We may end up reading (and writing) all components of an object composed
+    // of multiple fields. In such cases, we must do so atomically to avoid
+    // overwriting components modified by another thread. Note that this also
+    // implies multiple shared reads on the rhs being treated as atomic.
```
martin-cs, line 123: Thanks for the comment that makes this clear but ... I am a little concerned about this because now
`struct s = t;` and

```
struct s;
s.x = t.x;
s.y = t.y;
```

can have different behaviour.  I guess there is a very valid question about what the `atomic` actions in the CPROVER concurrency model are.

tautschnig: Although this case shouldn't arise for structs and mainly affects arrays: you are absolutely right that this is questionable either way. Without this patch, we'll have spurious results (as documented in the regression test). With this patch, we may miss some behaviour, depending on what exactly we otherwise consider to be atomic.

peterschrammel (approved):

PR #6126, opened 2021-05-18
Solver factory cleanup

thomasspriggs:
A couple of small tidy ups I wanted to do after reading the solver factory to do other work. This clean up is raised as a separate PR as it doesn't depend on the other work or vice versa

### DIFF
--- a/src/goto-checker/solver_factory.cpp
+++ b/src/goto-checker/solver_factory.cpp
@@ -411,7 +411,7 @@ void solver_factoryt::no_incremental_check()
 {
   const bool all_properties = options.get_bool_option("all-properties");
   const bool cover = options.is_set("cover");
-  const bool incremental_check = options.is_set("incremental-check");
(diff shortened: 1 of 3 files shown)


src/goto-checker/solver_factory.cpp
```diff
@@ -411,7 +411,6 @@ void solver_factoryt::no_incremental_check()
 {
   const bool all_properties = options.get_bool_option("all-properties");
   const bool cover = options.is_set("cover");
-  const bool incremental_check = options.is_set("incremental-check");
```
peterschrammel, line 414: This should be `incremental-loop`.

thomasspriggs: Ok. I have removed this deletion from this PR and pushed a fix with a regression test instead.

peterschrammel (approved):

PR #6128, opened 2021-05-18
Upgrade CaDiCaL to 1.4.1

tautschnig:
Armin Biere has released the version of CaDiCaL competing in the 2021
SAT competition. ~More precisely, there's also a tag sc2021 in the CaDiCaL repository, but the changes do not seem to substantially impact performance when evaluated over SV-COMP.~ Comparing releases 1.3.0 and sc2021 doesn't show any major performance differences, so really this is just to make consistent what the cadical and ipasir-cadical CMake configurations will use.

### DIFF
--- a/src/solvers/CMakeLists.txt
+++ b/src/solvers/CMakeLists.txt
@@ -107,14 +107,14 @@ elseif("${sat_impl}" STREQUAL "cadical")
     message(STATUS "Building solvers with cadical")
 
     download_project(PROJ cadical
-        URL https://github.com/arminbiere/cadical/archive/rel-1.3.0.tar.gz
+        URL https://github.com/arminbiere/cadical/archive/rel-1.4.1.tar.gz
         PATCH_COMMAND true
-        COMMAND CXX=${CMAKE_CXX_COMPILER} ./configure -O3 -s -j
-        URL_MD5 5bd15d1e198d2e904a8af8b7873dd341
+        COMMAND CXX=${CMAKE_CXX_COMPILER} ./configure -O3 -s CXXFLAGS=-std=c++14
+        URL_MD5 b44874501a175106424f4bd5de29aa59
     )
 
     message(STATUS "Building CaDiCaL")
-    execute_process(COMMAND make WORKING_DIRECTORY ${cadical_SOURCE_DIR})
+    execute_process(COMMAND make -j WORKING_DIRECTORY ${cadical_SOURCE_DIR})
 
     target_compile_definitions(solvers PUBLIC
         SATCHECK_CADICAL HAVE_CADICAL
@@ -137,10 +137,10 @@ elseif("${sat_impl}" STREQUAL "ipasir-cadical")
     message(STATUS "Building with IPASIR solver linking against: CaDiCaL")
 
     download_project(PROJ cadical
-        URL https://github.com/arminbiere/cadical/archive/rel-1.4.0.tar.gz
+        URL https://github.com/arminbiere/cadical/archive/rel-1.4.1.tar.gz
         PATCH_COMMAND true
         COMMAND ./configure CXX=g++
-        URL_MD5 9bad586a82995a1d95d1197d445a353a
+        URL_MD5 b44874501a175106424f4bd5de29aa59
     )
 
     message(STATUS "Building CaDiCaL")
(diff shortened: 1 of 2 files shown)


src/solvers/CMakeLists.txt
```diff
@@ -139,7 +139,7 @@ elseif("${sat_impl}" STREQUAL "ipasir-cadical")
     download_project(PROJ cadical
         URL https://github.com/arminbiere/cadical/archive/rel-1.4.0.tar.gz
         PATCH_COMMAND true
-        COMMAND ./configure CXX=g++
+        COMMAND CXX=${CMAKE_CXX_COMPILER} ./configure -O3 -s
```
NlightNFotis: Sorry, I'm not sure if this is correct.

I had tried this back when I was working with this, but this has a problem: On MacOS systems, we specify the compiler to be `/usr/bin/clang` for the C one, and `/usr/bin/clang++` for the C++ one, otherwise we get an error while building CBMC.

The problem is though, that CaDiCaL fails to build with `clang++` on any (mac) machine that I tried it - I cannot remember the reason exactly (I think it was dependent on some gcc intrinsics, though don't quote me on that).

This is the reason I had hardcoded it to be `g++`.

tautschnig: Ah, great call out! Perhaps we should actually upgrade to tag sc2021 then, where this has been fixed: https://github.com/arminbiere/cadical/blob/master/configure#L246. That said, https://github.com/arminbiere/cadical/blob/master/configure#L255 could probably still do with a fix.

tautschnig: I've updated the branch to use sc2021.

NlightNFotis: Wait, I looked into it a bit further, and my previous call is wrong - it just seems that when we build natively with `/usr/bin/clang++` the standard isn't set to C++11 by default.

I managed to get it to build by changing that line to:

```
COMMAND CXX=${CMAKE_CXX_COMPILER} ./configure CXXFLAGS=-std=c++11
```

but just touching CXXFLAGS is causing the build to break (e.g., if you change it to `CXXFLAGS="${CMAKE_CXX_FLAGS} -std=c++11"`).

I am not sure why it behaves this way though; I thought we setup the standard globally, so I'm surprised it doesn't propagate.

tautschnig: `COMMAND CXX=${CMAKE_CXX_COMPILER} ./configure CXXFLAGS="${CMAKE_CXX_FLAGS} -std=c++11"` won't work, because CMake has an interesting approach to quoting. I don't think there's a way to pass anything that uses spaces in this place. So `CXXFLAGS=-std=c++11` might actually work.

tautschnig: Flags updated in this PR.

NlightNFotis: @tautschnig Okay, I spent the last hour researching this and I think I now understand what's happening:

* When you pass a `CXX=` flag to CaDiCaL's `configure` script, for a strange kind of fashion it messes up its `CXXFLAGS` as well, losing the configuration for the standard (among other things).
* By losing the configuration for the standard, `clang++` becomes allergic to some things, for example template syntax (it complains that `>>` needs to have a space - this is a build error).
* As soon as you fix the standard before the configure script is run, then everything is working again.

---

So just as an example, `CXX=/usr/bin/clang++ ./configure` is giving us these:

* `configure: compiling with '/usr/bin/clang++ -W -O -DNDEBUG'`,
* `/usr/bin/clang++ -W -O -DNDEBUG -I../build -c ../src/analyze.cpp` (sample compilation)

But by default it looks like this:

* `configure: compiling with 'g++ -Wall -Wextra -O3 -DNDEBUG -std=c++0x'`
* `g++ -Wall -Wextra -O3 -DNDEBUG -std=c++0x -I../build -c ../src/analyze.cpp`.

So the conclusion is that if you manage to patch `CXXFLAGS` as well before the call to configure, it should work, even with CaDiCaL 1.4.0. Something like `CXX=/usr/bin/clang++ CXXFLAGS="-Wall -Wextra -O3 -DNDEBUG -std=c++11" ./configure` should work, translated to CMake syntax.

tautschnig: It seems we might have to wait for the next release as the build issues have been fixed: https://github.com/arminbiere/cadical/commit/c87965febd9a208d2422ac10bf76c7b90eb460ac.

martin-cs (approved):

peterschrammel (approved):

PR #6131, opened 2021-05-19
Add start of foundational work for incremental SMT2 solving support

thomasspriggs:
Add start of foundational work for incremental SMT2 solving support. This is the first piece of work for issue https://github.com/diffblue/cbmc/issues/6134

The work in this PR should serve as a basis for actual implementation which is work currently being under-taken within Diffblue. It is not useful to external users on its own, but it is raised now as its own PR in the interests of -

- Keeping PRs small and reviewable.
- Starting the review and feedback process before reaching the end the piece of work.
- Avoiding using a feature branch that becomes time consuming to resolve conflicts with the `develop` branch.

- [x] ~~My commit message includes data points confirming performance improvements (if claimed).~~ Non claimed.


src/ansi-c/ansi_c_entry_point.cpp
```diff
@@ -201,7 +218,7 @@ bool generate_ansi_c_start_function(
   init_code.add(code_labelt(CPROVER_PREFIX "HIDE", code_skipt()));
 
   // build call to initialization function
-
+  if(enable_static_lifetime_initialisation)
```
tautschnig: I'm aware this is a draft, but just an early warning: there could be all sorts of places that (implicitly) expect that this call be present. The exact instructions in the function won't matter as much.

tautschnig:
Incidentally, @kroening and I talked about incremental SMT earlier today. Just sharing some thoughts: one possible obstacle to consider is how to interface with the solver to facilitate incremental queries. The current, file-based interface will have poor performance. @kroening considered a sockets-based interface for Unix-like systems, with Windows being somewhat left out.

src/cbmc/cbmc_parse_options.cpp
```diff
@@ -1042,6 +1049,9 @@ void cbmc_parse_optionst::help()
     " --yices                      use Yices\n"
     " --z3                         use Z3\n"
     " --refine                     use refinement procedure (experimental)\n"
+    " --incremental-smt2-solver cmd\n"
+    "                              Command to invoke external SMT solver for\n"
```
TGWDB: "Command" -> c "command" for consistency

regression/cbmc-incr-smt2/nondeterministic-int-assert/incremental_solver_called.desc
```diff
@@ -0,0 +1,12 @@
+CORE
+empty.c
+--incremental-smt2-solver z3
+WARNING: Leaving variables of static lifetime uninitialised.
```
TGWDB: I don't think this warning is part of this test.

regression/cbmc-incr-smt2/nondeterministic-int-assert/incremental_solver_called.desc
```diff
@@ -0,0 +1,12 @@
...
+Passing problem to incremental SMT2 solving via "z3"
```
TGWDB, line 4: Should we check the passing/printing/etc of this string (how does it handle spaces or special characters in this string)?

src/ansi-c/ansi_c_entry_point.cpp
thomasspriggs: Thanks. I have every expectation that there will be things which are broken and broken in surprising ways. What we are trying to avoid is a partial implementation which ignores (with a warning or not) kinds of expression which are not implemented. The idea is that we will halt at the first expression which is not understood, in order to make these easy to find rather than like looking for a needle in a haystack. The problem is that we are expecting unimplemented functionality in the initialise function. Therefore by skipping it we can still test rudimentary functionality in trivial user code examples, in regression tests. So once we reach the point where everything in the initialise function is supported we can revert skipping it and proceed with more real-world examples. We are using the fairly coarse approach of skipping the initialise function entirely, rather than filtering out unsupported features because this is a temporary solution only.

regression/cbmc-incr-smt2/nondeterministic-int-assert/incremental_solver_called.desc
thomasspriggs: There is no special handling here because this is a rough initial version. Once we have more of the sub-process communication stuff in place we can can test this with examples that have spaces or other interesting characters. The reason to avoid over-thinking this now is in case we change the internal representation with things like splitting into multiple strings, which could entail rework of the printing if we are overly thorough about this upfront. Unless you know of a handy string sanitisation / escaping function in the code base already, which we could just slot in here...

regression/cbmc-incr-smt2/nondeterministic-int-assert/incremental_solver_called.desc
thomasspriggs: Ok. I'll remove this one, as it is probably throwing the net a little too wide.

regression/cbmc-incr-smt2/nondeterministic-int-assert/incremental_solver_called.desc
TGWDB: `util/run.cpp` has some shell quoting stuff we could reuse?

That said I agree we don't need to commit too seriously here. I was mostly thinking the test could check that if we pass in something like `"z3 -T:600"` (to have a 600 second timeout) that this is handled fine for now and shows up in the output.

regression/cbmc-incr-smt2/nondeterministic-int-assert/incremental_solver_called.desc
TGWDB: Great. I am hoping this test will survive a lot of the PRs without changes as a very basic sanity test. If we only test the string of the command then we can keep it when we reinstate the CPROVER initialization.

thomasspriggs:
> Incidentally, @kroening and I talked about incremental SMT earlier today. Just sharing some thoughts: one possible obstacle to consider is how to interface with the solver to facilitate incremental queries. The current, file-based interface will have poor performance. @kroening considered a sockets-based interface for Unix-like systems, with Windows being somewhat left out.

Dealing with the poor performance which a file-based interface would have when doing incremental solving is already one of our goals. We are currently thinking along the lines of starting the solver sub-process with the appropriate command line arguments to instruct it to take input from the standard command line input rather than a file and then redirecting its standard input/output to CBMC using pipes. I think that could be made to work across all operating systems including Windows, but we haven't investigated the Windows APIs for this too closely yet.

You mentioned using a sockets-based interface. Is support for sockets-based communication something which is commonly built into SMT2 solvers already or something which would have to be added on to them somehow?

kroening:
We have made a number of attempts to get pipe-based communication with subprocesses working. E.g., look at 38642cd07df2b8bee87d085ed1da830b87ed51df
Other attempts are the interface to gdb, and Georg Weissenbacher has experimented with incremental SMT in the past.

The key problem was always that we couldn't get it to work reliably on Windows.

src/ansi-c/ansi_c_entry_point.cpp
```diff
@@ -173,10 +174,25 @@ bool ansi_c_entry_point(
     return false; // give up
   }
 
-  static_lifetime_init(symbol_table, symbol.location);
+  if(enable_static_lifetime_initialisation)
```
martin-cs: What does this gains us over the `--nondet-static` ?

src/ansi-c/ansi_c_entry_point.h
```diff
@@ -15,15 +15,26 @@ class symbol_tablet;
...
 
+/// \brief Generate start/entry point function in the symbol table.
+/// \param symbol_table: The symbol table into which to add the new function.
+/// \param message_handlert: For reporting error conditions and warnings.
+/// \param c_object_factory_parameterst: For applying the object factory to all
+///   variables which need to be non deterministically initialised.
+/// \param enable_static_lifetime_initialisation: When set to false a start
+///   function will be generated which does not initialise variables with static
+///   lifetime. This was added in order to allow the partial incremental smt2
+///   support to be tested using a subset of variable types.
```
martin-cs: Fine for a draft but I'm not sure this should stay long-term.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.h
```diff
@@ -0,0 +1,36 @@
+
+#ifndef CBMC_SMT2_INCREMENTAL_DECISION_PROCEDURE_H
+#define CBMC_SMT2_INCREMENTAL_DECISION_PROCEDURE_H
+
+#include <solvers/stack_decision_procedure.h>
+
+class smt2_incremental_decision_proceduret final
+  : public stack_decision_proceduret
```
martin-cs, line 12: Is this not making use of the existing SMT-LIB back-end?

martin-cs (commented):
I am really enthusiastic about this as a piece of functionality and would love to see it as part of CPROVER.  In the spirit of this enthusiasm, let me contribute some complication!

When I have looked at this there are basically two big architectural questions:
1. How you connect to the solver?
2. Which of the incremental solving mechanisms you use?

For 1. you have either have A. a loose coupling where you interact with the solver via text / SMT-LIB or B. a tight integration where you link against the solver's API.  A. is slower (but unless you are absolutely hammering out queries, it is unlikely to be a major cost) but simpler... if you are on *NIX.  There you can start a child process and pipe `stdin` / `stdout`.  I am not a windows programmer so I hope I am wrong here but my understanding was the windows doesn't have reliable pipes.  Which means that A is either *NIX only or you have to write some kind of solver wrapper.

B. is about as fast as you can get but adds build complexity and dependency (more than our existing SAT integrations but maybe not that much more).  Also, it used to tie you to just one server.  However there has been progress on this! https://github.com/makaimann/smt-switch is a generic C++ API that allows you to create instances of Bitwuzla, CVC4, Z3, MathSAT and Yices.  Obviously it is another dependency but if you are going to the trouble of using a solver API, you might as well use that one.

2. also has two options, you can C. use push/pop or D. use solve-under-assumption.  C is a lot more general and allows you to get rid of potentially expensive or complex terms.  However it is a massive pain to implement and I think only CVC4 has a complete and mostly hysteresis-free implementation.  Z3 famously handles this badly.  D. is a lot more limited in that you can only really add formula (although adding an "if enabled..." guard allows you to simulate removal) but it is simple and fast.  If you can adapt the kind of incremental solving you are doing to D (and unless we are doing incremental symex then ... we probably can) then it really is the way to go.  I don't see there being much ongoing work on improving push/pop but solve-under-assumption seems a lot more popular.

I had implementing B/D on my "wouldn't it be great if" list as I think it is really the best way of approaching this and probably less work than refactoring the current SMT-LIB 2 back-end.

If there is anything I can do to actually help then do say!

src/ansi-c/ansi_c_entry_point.h
thomasspriggs: Yes. This should definitely be removed/reverted long-term.

src/ansi-c/ansi_c_entry_point.cpp
thomasspriggs: Complete removal of `__cprover_initialise` instead of nondet initialisation of some of the variables in there. This avoids ever adding pointer typed assignments such as `__CPROVER_memory_leak = NULL;` and it avoids array expressions such as `__CPROVER_threads_exited = ARRAY_OF(FALSE);`.

The idea is that this will facilitate implementation and integration testing of an initial subset of functionality with integer support only from source through to solver, without support for pointers and structs. Support for the missing functionality can then be added afterwards in piece-wise fashion, until we can put `__cprover_initialise` back in place. This will allow us to thoroughly test and debug the areas which are complete rather than attempting to implement everything at once and then hunting the bugs across a broad swathe of functionality.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.h
thomasspriggs: No. We are planning to use the existing implementation in `solvers/smt2/smt2_conv.cpp` for reference, but not to reuse it verbatim. This will enable structuring the implementation differently. The existing implementation has: 

-  5000 lines of code within a single file.
-  Result parsing and formula generation in the same file/class.
-  Functions which convert directly from `exprt` to smt2 formulas as strings.
-  No unit tests.

We were concerned that adding a major new feature into the code for the existing implementation would be highly error prone and make it tricky to track down bugs when they do occur. In particular we would like to add separated:
 * Abstract representation of smt2 formulas.
 * Conversion from `exprt` to abstract smt2 formulas.
 * Conversion from abstract smt2 formulas to string.
 * Unit tests for the separate conversion steps.
The idea is that the above outlined structure would make the testing and debugging comparatively straight forward. We will learn as much as possible from the existing implementation in order make good use of the work which has been done previously.

martin-cs: Seriously; look at SMT-Switch.  That already has most of what you need.

src/ansi-c/ansi_c_entry_point.cpp
martin-cs: Why not simply ignore certain types in the conversion?  That would be less intrusive and give you better diagnostics.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.h
thomasspriggs: Using `SMT-Switch` is an interesting idea and it warrants further discussion within Diffblue. So please do not think that I am discounting that option. However I can see at least 2 potential issues with that approach, namely:

1. The `SMT-switch` repository you linked to has no Windows CI. CBMC does have Visual Studio / Windows CI, testing and packages. So it would be good to choose a solution which allows us to keep parity between Windows support and support for Unix-like operating systems.
2. There is a desire for it to be possible to swap in new / different solvers at run time. Is that something which can be supported through `SMT-switch`, or would we be stuck with supporting only fixed versions of selected solvers linked to CBMC at build / compile time?

thomasspriggs:
> We have made a number of attempts to get pipe-based communication with subprocesses working. E.g., look at [38642cd](https://github.com/diffblue/cbmc/commit/38642cd07df2b8bee87d085ed1da830b87ed51df)
> Other attempts are the interface to gdb, and Georg Weissenbacher has experimented with incremental SMT in the past.
> 
> The key problem was always that we couldn't get it to work reliably on Windows.

Thank you for sharing this useful information about what has been tried previously. @NlightNFotis has been looking at the existing implementation in `src/memory-analyzer/gdb_api.cpp`, with a view to factoring out this existing sub process communication code so that the same code can be used by both the memory-analyzer and the SMT2 interface.

src/ansi-c/ansi_c_entry_point.cpp
thomasspriggs: Because we would like to error rather than ignore on encountering unimplemented functionality in the new `exprt` to smt2 conversion in the incremental support. That way we can be sure that all _user_ code up to the current point of execution is supported, without having to read back through the output checking for warnings. `__cprover_initialise` conflicts with this approach, because it introduces potentially unimplemented expressions at the beginning of the analysis before we even reach user code. Filtering out unsupported expressions encountered before reaching the user code and erroring on unsupported expressions reached after that seemed like over-engineering a temporary work around which we were already planning to revert.

martin-cs:
On Thu, 2021-05-20 at 07:25 -0700, thomasspriggs wrote:
> 
> Using `SMT-Switch` is an interesting idea and it warrants further
> discussion within Diffblue. So please do not think that I am
> discounting that option. However I can see at least 2 potential
> issues with that approach, namely:
> 
> 1. The `SMT-switch` repository you linked to has no Windows CI. CBMC
> does have Visual Studio / Windows CI, testing and packages. So it
> would be good to choose a solution which allows us to keep parity
> between Windows support and support for Unix-like operating systems.

I very much support feature parity on all platforms but it might be a
bit of a challenge with Windows unless you can find / build a reliable
pipe.  Even among the solvers that have windows binaries, the actual
level of testing and support is ... not great.

[ I could perhaps really troll and suggest that WSL is looking to have
a future and maybe that would be a route! ]


> 2. There is a desire for it to be possible to swap in new / different
> solvers at run time. Is that something which can be supported through
> `SMT-switch`, or would we be stuck with supporting only fixed
> versions of selected solvers linked to CBMC at build / compile time?

AFAIK you would be able to run-time switch against anything that you
have linked against.  This would be more flexible than, say, our long-
time handling of SAT solvers but not as free as a text SMT-LIB
interface.  I think smt-switch also has a generic SMT-LIB back-end as
well.


Forgive me if I am telling you things you already know; I don't know
how much experience you have with SMT solvers but a few other things to
consider:

* There is a long-standing issue with the representation of values in
models.  It seems simple, even trivial to start with but becomes pretty
horrific when you start having to deal with uninterpreted functions,
quantifiers, etc.  SMT-LIB 3.0 should help with this somewhat but it is
going to remain an annoying wrinkle for the next few years so you will
need some solver-specific code.

* One of the reasons that the existing SMT-LIB back-end is as large as
it is, is because it requires a two-pass design.  All variables have to
be declared or defined before they are used.  This is a disadvantage of
using SMT-LIB as your interface (it is rare you will hear me advocating
against SMT-LIB but, that is one point).

* Another reason is the variety of logics supported.  Unless you are
going to be strictly Z3 and CVC4/5 only (and there are good reasons not
to be) then you may well have to implement multiple encodings for
things like structs, strings, quantifiers, etc.  This intersects with
the previous issue.

TGWDB:
Added an RFC issue #6134 that provides much more context and information for this PR.

kroening:
The RFC suggests that the option to disable static lifetime initialisation will get merged. I am very strongly opposed to that, even if there's a plan to remove it later.

TGWDB:
> The RFC suggests that the option to disable static lifetime initialisation will get merged. I am very strongly opposed to that, even if there's a plan to remove it later.

The disabling of the static lifetime initialisation is planned to be disabled only inside the new SMT implementation and will always display a warning about this. I believe @thomasspriggs discussed above why this is the preferred option in this PR. My understanding is that we need to lifetime initialisation to be effectively disabled **only within the new SMT backend** to prevent issues while this new SMT backend is being developed. The choice to entirely disable is to ensure a light modification and minimal impact on the codebase.

With the above detail, can you please clarify that I understand you oppose all disabling of the lifetime initialisation even if strictly within a specific work in progress?

martin-cs:
I'm with @kroening on this one.  Temporary things have a nasty habit of becoming permanent.  Will `--no-library` or `--slice-formula` not omit or ignore the pointer and array things sufficiently?

thomasspriggs:
> I'm with @kroening on this one. Temporary things have a nasty habit of becoming permanent. Will `--no-library` or `--slice-formula` not omit or ignore the pointer and array things sufficiently?

I have done a little more investigation and it seems like `--slice-formula` might well do what we need. I am going to leave this PR alone for the moment and come back to it on Monday.

thomasspriggs:
> > I'm with @kroening on this one. Temporary things have a nasty habit of becoming permanent. Will `--no-library` or `--slice-formula` not omit or ignore the pointer and array things sufficiently?
> 
> I have done a little more investigation and it seems like `--slice-formula` might well do what we need. I am going to leave this PR alone for the moment and come back to it on Monday.

I have removed the code for skipping `cprover-initialise` and added regexes to the test in order to confirm that `--slice-formula` does what we need.

src/ansi-c/ansi_c_entry_point.cpp
thomasspriggs: This code is now gone. So that implicit expectation should no longer be an issue.

src/cbmc/cbmc_parse_options.cpp
thomasspriggs: done.

regression/cbmc-incr-smt2/nondeterministic-int-assert/incremental_solver_called.desc
thomasspriggs: This is now gone.

regression/cbmc-incr-smt2/nondeterministic-int-assert/incremental_solver_called.desc
thomasspriggs: This seems like a finer point that should be looked at as part of the follow-up sub-process communication work.

src/ansi-c/ansi_c_entry_point.cpp
thomasspriggs: This skipping of `__cprover_initialise` is now gone. Therefore I am going to mark this comment chain as resolved.

src/ansi-c/ansi_c_entry_point.h
thomasspriggs: Its gone.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.h
thomasspriggs: This seems like a conversation for the issue RFC - https://github.com/diffblue/cbmc/issues/6134
Ok, if I mark this as resolved for the minute as it doesn't really relate to the work which is currently in this PR?

martin-cs: Sure!

martin-cs:
( As an aside; I don't really know why `--slice-formula` is not on by default.  I guess it is a problem if you are trying to generate complete test cases, including irrelevant fields, etc. )

CODEOWNERS
```diff
@@ -54,6 +54,7 @@
 /jbmc/src/jdiff/ @smowton @peterschrammel
 /src/cpp/ @kroening @tautschnig @peterschrammel
 /src/solvers/smt2 @kroening @martin-cs @tautschnig @peterschrammel @allredj @romainbrenguier
+/src/solvers/smt2_incremental @peterschrammel @diffblue/diffblue-opensource
```
martin-cs: May I suggest some of the owners of `/src/solvers/smt2` as well.  @tautschnig does fantastic reviews and @romainbrenguier has the deepest understanding of the string solver which might well be one area you can really improve on.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+exprt smt2_incremental_decision_proceduret::handle(const exprt &expr)
+{
+  return expr;
+}
+
+exprt smt2_incremental_decision_proceduret::get(const exprt &expr) const
+{
+  INVARIANT(false, "Unimplemented get of:\n  " + expr.pretty(2, 0));
```
martin-cs: `UNIMPLEMENTED` rather than `INVARIANT` please?

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+void smt2_incremental_decision_proceduret::print_assignment(
+  std::ostream &out) const
+{
+  INVARIANT(false, "Unimplemented printing of assignments.");
```
martin-cs: Ditto.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+std::size_t
+smt2_incremental_decision_proceduret::get_number_of_solver_calls() const
+{
+  return number_of_solver_calls;
+}
+
+void smt2_incremental_decision_proceduret::set_to(const exprt &expr, bool value)
+{
+  INVARIANT(
```
martin-cs: Ditto.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -0,0 +1,75 @@
...
+    "Unimplemented set_to (" + std::string{value ? "true" : "false"} +
+      "):\n  " + expr.pretty(2, 0));
+}
+
+void smt2_incremental_decision_proceduret::push(
+  const std::vector<exprt> &assumptions)
+{
+  for(const auto &assumption : assumptions)
+  {
+    INVARIANT(
```
martin-cs: Ditto.

martin-cs (approved):
Approve modulo minor changes.

peterschrammel (approved):

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
thomasspriggs: The `UNIMPLEMENTED` macro is actually under the Legacy annotations section in `invariant.h` with a comment of "should not be used in new code" - https://github.com/diffblue/cbmc/blob/develop/src/util/invariant.h#L519

`UNIMPLEMENTED` also does not allow for a reason to be specified. This would prevent me from including the pretty printing of the expression in the printed output.

martin-cs: Yeah, I wrote that :-)  I put it in so that the number of `UNIMPLEMENTED` `TODO` and `UNHANDLED_CASE` asserts decreased with time once we had looked at and converted all of the rogue `asset(0)`s.  I think this is one of the legitimate cases where adding them would be OK.  It certainly would be better style than having direct `INVARIANT(false, "...")`.

Adding a variant of `UNIMPLEMENTED` which takes a message would be fine, as would defining your own custom macro:

```
#define SMT_LIB_WIP(X) INVARIANT(false,"SMT-LIB backend WIP : " X)
```

which would allow you to quickly locate how many of these there are in the code and filter out results that are caused by WIP issues, while running larger scale testing.

TGWDB (approved):
I like the suggestion of @martin-cs to have a specific failure case for the unimplemented/WIP SMT (but do not consider this blocking).

This looks good to me.

CODEOWNERS
thomasspriggs: Is this a hint that you would like to be a code owner as well?

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
NlightNFotis: I would also prefer this to be anything but an `INVARIANT`.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -12,8 +12,7 @@ smt2_incremental_decision_proceduret::smt2_incremental_decision_proceduret(
 
 exprt smt2_incremental_decision_proceduret::handle(const exprt &expr)
 {
-  UNIMPLEMENTED;
```
NlightNFotis: I do not understand this change and its commit message.

Why was this changed from `UNIMPLEMENTED` to just returning what was passed (it also loses the `const` modifier, unless I'm mistaken?)

NlightNFotis (commented):
Some questions from my end.

CODEOWNERS
thomasspriggs: I have checked with Romain and it is better to leave him off this list, as reviewing PRs on this piece of work would not currently be a priority for him.

martin-cs: Your code; it's up to you who is a owner!  I would at least add yourself so that you can authorise (or not) controls that others make.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
thomasspriggs: The `const` modifier is not "lost", as the type of this function is the same before and after this commit. The parameter is passed in by const reference and a copy is returned by value, therefore this is const correct and the return value cannot be used to mutate any other version of the expression.

Without this commit the console output from the test case is as follows, (both with and without `--slice-formula`)
```
Reason: Unimplemented handle of:
  constant
    * type: bool
    * value: true
```
This doesn't give us enough information in the regression test, to confirm that `--slice-formula` removes the expressions from `cprover_initialise` which we don't plan to support in the first slice of work. Implementing as the identity/copy function enables the test case to run a little further before hitting the unimplemented `set_to` member function which gives us more useful diagnostics.

There is likely a better way to implement this function. Unfortunately it is not documented in the base class. You can see the other current implementations of it here - https://github.com/diffblue/cbmc/blob/develop/src/solvers/smt2/smt2_conv.cpp#L756 and here - https://github.com/diffblue/cbmc/blob/develop/src/solvers/prop/prop_conv_solver.cpp#L40

CODEOWNERS
thomasspriggs: I am already in the `@diffblue/diffblue-opensource` group. I think I'll expand it to the group's members so that this is less opaque.

thomasspriggs:
This PR is now based on top of https://github.com/diffblue/cbmc/pull/6152/files so that the new macro can be reviewed separately, rather than adding ever more functionality to an already approved PR.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
thomasspriggs: This is now using the new `UNIMPLEMENTED_FEATURE` marco.

CODEOWNERS
thomasspriggs: I have pushed the expansion of the group.

thomasspriggs:
I am planning to merge this once CI completes. :crossed_fingers:

PR #6132, opened 2021-05-20
Use FloatingPoint theory when using --cprover-smt2

kroening:

### DIFF
--- a/src/solvers/smt2/smt2_parser.cpp
+++ b/src/solvers/smt2/smt2_parser.cpp
@@ -500,7 +500,10 @@ exprt smt2_parsert::function_application()
       if(next_token() != smt2_tokenizert::SYMBOL)
         throw error("expected symbol after '_'");
 
-      if(has_prefix(smt2_tokenizer.get_buffer(), "bv"))
+      // copy, the reference won't be stable
+      const auto id = smt2_tokenizer.get_buffer();
+
+      if(has_prefix(id, "bv"))
       {
         mp_integer i = string2integer(
           std::string(smt2_tokenizer.get_buffer(), 2, std::string::npos));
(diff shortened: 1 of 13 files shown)


martin-cs (approved):
Looks good.  Thanks for the clearly broken out commits.

src/solvers/smt2/smt2_parser.cpp
```diff
@@ -699,44 +728,80 @@ exprt smt2_parsert::function_application()
...
+                id2string(to_constant_expr(source_op).get_value());
+              auto dot_pos = real_number.find('.');
+              if(dot_pos == std::string::npos)
+              {
+                exponent = 0;
+                significand = string2integer(real_number);
+              }
+              else
+              {
+                // remove the '.', if any
```
tautschnig: The ", if any" in the comment seems strange, because we know that there is a '.' in the `else` case.

src/solvers/smt2/smt2_parser.cpp
```diff
@@ -699,44 +728,80 @@ exprt smt2_parsert::function_application()
...
+                std::string significand_str;
+                significand_str.reserve(real_number.size());
+                for(auto ch : real_number)
+                  if(ch != '.')
+                    significand_str += ch;
```
tautschnig: 1) Is the intended behaviour really such that the '.' is the only character not being copied over?
2) Nit pick: Braces around the body of the for-loop would be nice.
3) Perhaps just use `erase` instead:
```
std::string significand_str = real_number;
significand_str.erase(significand_str.begin() + dot_pos);
```
4) As further alternative, create `significand_str` by concatenating two substrings, which perhaps is more efficient in some future version of C++ with string views.

src/solvers/smt2/smt2_parser.cpp
```diff
@@ -699,44 +728,80 @@ exprt smt2_parsert::function_application()
...
+
+                exponent =
+                  mp_integer(dot_pos) - mp_integer(real_number.size()) + 1;
+                significand = string2integer(significand_str);
+              }
+
+              ieee_floatt a(spec);
+              a.rounding_mode = static_cast<ieee_floatt::rounding_modet>(
+                numeric_cast_v<int>(to_constant_expr(rounding_mode)));
```
tautschnig: We really ought to have a constructor of `ieee_floatt` that takes a rounding mode...

tautschnig (approved):

TGWDB (approved):

PR #6138, opened 2021-05-24
bv_pointerst: else case is just for byte updates

tautschnig:
We should not treat all other expressions as byte updates. Expressions
other than those handled in the if/else if chain should result in
failing conversion.


TGWDB (commented):
This looks nice and a good way to catch errors. Is it possible to actually reach this code path? If so can you create a regression test that catches it?

martin-cs (approved):

martin-cs:
I think @TGWDB is correct to request a test but I think this is a rare code path and may only be taken by other things that also need to be fixed so I am willing to merge without one if @tautschnig says that adding one would be testing for incorrect or buggy behaviour.

peterschrammel (approved):

PR #6143, opened 2021-05-24
Binary comparison <, <=, >, >= over pointers requires same-object

tautschnig:
We used to compare them just like integers, even though our
object:offset encoding restricted the space of possible results.

### DIFF
--- a/src/solvers/flattening/bv_pointers.cpp
+++ b/src/solvers/flattening/bv_pointers.cpp
@@ -172,8 +172,30 @@ literalt bv_pointerst::convert_rest(const exprt &expr)
       const bvt &bv0=convert_bv(operands[0]);
       const bvt &bv1=convert_bv(operands[1]);
 
-      return bv_utils.rel(
-        bv0, expr.id(), bv1, bv_utilst::representationt::UNSIGNED);
+      const pointer_typet &type0 = to_pointer_type(operands[0].type());
+      bvt offset_bv0 = offset_literals(bv0, type0);
+
+      const pointer_typet &type1 = to_pointer_type(operands[1].type());
+      bvt offset_bv1 = offset_literals(bv1, type1);
+
+      // Comparison over pointers to distinct objects is undefined behavior in
+      // C; we choose to always produce "false" in such a case.  Alternatively,
+      // we could do a comparison over the integer representation of a pointer
+
+      // do the same-object-test via an expression as this may permit re-using
+      // already cached encodings of the equality test
+      const exprt same_object = ::same_object(operands[0], operands[1]);
+      const literalt same_object_lit = convert(same_object);
+      if(same_object_lit.is_false())
+        return same_object_lit;
+
+      return prop.land(
+        same_object_lit,
+        bv_utils.rel(
+          offset_bv0,
+          expr.id(),
+          offset_bv1,
+          bv_utilst::representationt::SIGNED));
     }
   }
 
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -27,6 +27,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/namespace.h>
 #include <util/pointer_expr.h>
 #include <util/pointer_offset_size.h>
+#include <util/pointer_predicates.h>
 #include <util/prefix.h>
 #include <util/range.h>
 #include <util/simplify_expr.h>
@@ -3165,7 +3166,6 @@ void smt2_convt::convert_relation(const binary_relation_exprt &expr)
   const typet &op_type=expr.op0().type();
 
   if(op_type.id()==ID_unsignedbv ||
-     op_type.id()==ID_pointer ||
      op_type.id()==ID_bv)
   {
     out << "(";
@@ -3237,6 +3237,31 @@ void smt2_convt::convert_relation(const binary_relation_exprt &expr)
     out << " ";
     convert_expr(expr.op1());
     out << ")";
+  }
+  else if(op_type.id() == ID_pointer)
+  {
+    const exprt same_object = ::same_object(expr.op0(), expr.op1());
+
+    out << "(and ";
+    convert_expr(same_object);
+
+    out << " (";
+    if(expr.id() == ID_le)
+      out << "bvsle";
+    else if(expr.id() == ID_lt)
+      out << "bvslt";
+    else if(expr.id() == ID_ge)
+      out << "bvsge";
+    else if(expr.id() == ID_gt)
+      out << "bvsgt";
+
+    out << ' ';
+    convert_expr(pointer_offset(expr.op0()));
+    out << ' ';
+    convert_expr(pointer_offset(expr.op1()));
+    out << ')';
+
+    out << ')';
   }
   else
     UNEXPECTEDCASE(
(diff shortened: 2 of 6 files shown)


src/solvers/flattening/bv_pointers.cpp
```diff
@@ -172,8 +172,38 @@ literalt bv_pointerst::convert_rest(const exprt &expr)
...
+      // comparison over pointers to distinct objects is undefined and yields a
+      // non-deterministic result; alternatively, we could do a comparison over
+      // the integer representation of a pointer
+
+      // do the same-object-test via an expression as this may permit re-using
+      // already cached encodings of the equality test
+      const exprt same_object = ::same_object(operands[0], operands[1]);
+      const bvt &same_object_bv = convert_bv(same_object);
+      CHECK_RETURN(same_object_bv.size() == 1);
+
```
kroening, line 191: Simply use `convert` instead of `convert_bv`?

tautschnig: Done. I'll push a new version once I've also fixed the SMT back-end.

src/solvers/flattening/bv_pointers.cpp
```diff
@@ -172,8 +172,38 @@ literalt bv_pointerst::convert_rest(const exprt &expr)
...
+      literalt l = prop.new_variable();
+
+      if(!same_object_bv[0].is_false())
```
martin-cs: I like this optimisation but if we are being super strict, shouldn't the new literal be inside the if too.

martin-cs (approved):
Looks like a good bug fix.  One thought is that it will change (mostly for the better, I think) the behaviour of:

`__CPROVER_assume( p < &array + LENGTH );`

which will now subtly constrain `p` to be in the same object.  From a standard point of view, this is correct, but it means that having a pointer which points to anywhere valid in a bunch of different arrays is now quite hard to specify.

src/solvers/flattening/bv_pointers.cpp
tautschnig: No, I don't think so: the encoding is such that the objects not matching leaves it to the solver to choose a value for that literal. This also means that `__CPROVER_assume( p < &array + LENGTH );` does not constrain `p` to be in the same object. I'm expanding the regression test to include this and demonstrate that it isn't being constrained.

tautschnig: Additional note: if you _want_ to constrain it to the same object, this can be done using `__CPROVER_pointer_object(p) == __CPROVER_pointer_object(&array)`.

martin-cs: Thinking about it; is this correct?  The case in which the if is not taken is when it can be (encoding time) determined that the pointers cannot point to the same object, right?  In which case ISO 9899 says that the comparison is undefined behaviour.  With `goto_check` we should be able to detect this case and users shouldn't rely on something that happens after undefined behaviour.  So we are back to the question of what we do after UB?  Your solution is to make it non-deterministic, but it would be no less correct for it to be false or true.

I think non-deterministic probably is the best trade off but A. I want to make sure that I understand what is going on and B. can you add a comment to saying that it is non-deterministic, the case and why.

tautschnig: We do correctly cover this case in `goto_check`, reporting "same object violation in ..." (tried on the regression test I'm including here). So I think we're safe on that side. The question is what the back-end should do in such cases. As you say, and behaviour is permitted.

martin-cs: I wonder if there is a doxygen-based way of commenting in the source and building a page / list of ... definitions of undefined behaviour, i.e. what we do.

tautschnig: We could introduce a new group `\defgroup undef Undefined Behaviour` and then do `\addtogroup undef`.

tautschnig:
I have added an implementation for SMT, but my knowledge of that code is limited. @kroening and/or @martin-cs, could you please take a look? ~It seems to work with Z3, but the built-in SMT2 solver just says `ERROR` for each of the comparisons. I'm not sure why that's the case.~ (edit: I had just messed up my $PATH)

src/solvers/flattening/bv_pointers.cpp
kroening: I would use different terminology here. "Undefined behavior" has specific ramifications when dealing with C code (despite the name), and in a tool for checking C this is what a reader will think of.

There are pros and cons to using nondeterminism for such cases. As starting point, take a look at the discussion around the meaning of division by zero in SMT-LIB.

The choice 'nondeterminism' is safe when doing satisfiability checking. It will start to bite once quantifiers are used.

tautschnig: So if we make a choice: should it be always true or always false? I'd intuitively pick the latter, but I'm not sure I have arguments supporting this.

kroening: I'd also go for 'false'; this is what IEEE did with NaNs.

kroening: BTW, doing that requires a review of the simplifier. In particular, the rule !(a>b) = a<=b no longer holds.

kroening: If you wanted to retain that rule, and not introduce an implicit existential quantification, you could go back to the total ordering that was there before.

martin-cs:
@tautschnig I don't think you need to create the literal in the SMT back-end.  Generating:

```
"(=> " + convert_expr(same_object(left, right))
 + " " + convert_expr(binary_relation_exprt(get_offset(left), relation, get_offset(right))) + ")"
```

should do the trick.

But this would be the semantics of "comparison is true if they point to different objects".  Using 'and' instead would give the semantics of "comparison is false if they point to different objects".  Hmmm... actually maybe you do need the literal if you want the non-deterministic semantics.

src/solvers/flattening/bv_pointers.cpp
tautschnig: I have now implemented "false" as the choice for comparison of pointers to distinct objects, and have also updated the simplifier. Thank you for calling this out, @kroening!

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -3122,6 +3122,31 @@ void smt2_convt::convert_relation(const binary_relation_exprt &expr)
...
+    out << " (";
+    if(expr.id() == ID_le)
+      out << "bvule";
+    else if(expr.id() == ID_lt)
+      out << "bvult";
+    else if(expr.id() == ID_ge)
+      out << "bvuge";
+    else if(expr.id() == ID_gt)
+      out << "bvugt";
+
```
kroening, line 3257: You are doing a signed comparison in bv_pointers but an unsigned one here?

tautschnig: Thank you, this is now fixed (copy&paste mistake as this was previously handled in the same case as `ID_unsigned`).

NlightNFotis (approved):
LGTM

NlightNFotis:
@tautschnig Can you rebase on top of `develop` and push once more please? 

It seems that the GitHub action job for VS2019 bugged out. Hopefully a push again would restart it.

kroening:
BTW, I would be very reluctant to "guarantee" any behavior for the undefined case. I am ok that we make that "false", but the tests currently generate the impression that this is a feature. There is a risk that users decide to rely on that.

Perhaps add a comment that this is what we currently do, but this may change.

kroening (approved):

PR #6145, opened 2021-05-24
Cleanup regression tests for quantified contracts

SaswatPadhi:
Cleaned up some the quantified contracts, changed some KNOWNBUG cases to CORE, added regex patterns in descriptions to make sure we catch expected failures (and don't ignore unexpected ones) etc.

Please review individual commits. The second commit is just a renaming: `*-01` -> `*-enforce` and `*-02` -> `*-replace`. Some tests didn't adhere to this convention (and were double testing the same context, enforcement / replacement), so I changed the context appropriately and now I think we are reasonably exhaustive on corner cases.

- n/a ~~My PR is restricted to a single feature or bugfix.~~


SaswatPadhi:
@feliperodri Please review the first commit. The second one is just for renaming.

regression/contracts/quantifiers-exists-ensures-02/main.c
```diff
@@ -1,20 +1,41 @@
...
+  int *arr = malloc(len * sizeof(int));
+
+  f1(arr, len);
+
+  bool found_zero = false;
+  for(int i = 0; i <= MAX_LEN; i++)
+  {
+    if(i < len)
+      found_zero |= (arr[i] == 0);
+  }
```
feliperodri: We are not proving these programs are correct, because we're not unrolling these loops. Thus, we have two options here:
1. Add invariants to these loops;
2. Update our regression-test script to also call CBMC with the correct bounds and `--unwinding-assertions`;

I'd prefer option (1). In any case, we can create a GitHub issue for this now and tackle it on a separate PR (for all test cases).

regression/contracts/quantifiers-exists-ensures-02/test.desc
```diff
@@ -1,19 +1,14 @@
...
-assertion and the verification remains sound when using __CPROVER_exists.
+within negative contexts (replaced ENSURES clauses).
 
-Known Bug:
-We expect verification to fail because arr[i] is always equal to 0 for
-[0 <= i < 10]. In fact, we expect the (0 <= i && i < 10) statement to act as a
-range for i. However, in the current implementation of quantifier handling,
-the (0 <= i && i < 10) statement is not interpreted as an explicit range, but
-instead, as part of a logic formula, which causes verification to succeed.
+This is fully supported (without requiring full unrolling) with the SAT backend.
```
feliperodri: What do you mean by `without requiring full unrolling`?

regression/contracts/quantifiers-exists-requires-01/main.c
```diff
@@ -1,15 +1,40 @@
+#include <assert.h>
```
feliperodri: We don't need this assert here.

regression/contracts/quantifiers-exists-requires-01/test.desc
```diff
@@ -1,12 +1,14 @@
...
 --
+^warning: ignoring
 --
 The purpose of this test is to ensure that we can safely use __CPROVER_exists
-in __CPROVER_requires clauses. By using the --replace-all-calls-with-contracts
-flag, goto-instrument will transform the __CPROVER_requires clauses into an
-assertion and the verification remains sound when using __CPROVER_exists.
+within both negative contexts (enforced REQUIRES clauses).
+
+This is fully supported (without requiring full unrolling) with the SAT backend.
```
feliperodri: Fix empty new line.

feliperodri: Again, "without requiring full unrolling"?

regression/contracts/quantifiers-exists-requires-02/test.desc
```diff
@@ -1,19 +1,17 @@
...
 
-Known Bug:
-We expect verification to fail because arr[i] is never equal to 1 for
-[0 <= i < 10]. In fact, we expect the (0 <= i && i < 10) statement to act as a
-range for i. However, in the current implementation of quantifier handling,
-the (0 <= i && i < 10) statement is not interpreted as an explicit range, but
-instead, as part of a logic formula, which causes verification to succeed.
+With the SAT backend existential quantifiers in a positive context,
+e.g., the REQUIRES clause being replaced in this case,
+are supported only if the quantifier is bound to a constant range.
```
feliperodri: We should also create a section in the CPROVER manual about quantifiers support. Let's create a GitHub issue to track this.

regression/contracts/quantifiers-nested-05/test.desc
```diff
@@ -5,7 +5,7 @@ main.c
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
+^warning: ignoring
```
feliperodri, line 8: Why is this expected?

regression/contracts/quantifiers-nested-05/main.c
```diff
@@ -1,8 +1,14 @@
...
-int f1(int *arr) __CPROVER_requires(!__CPROVER_forall {
-    int i;
-    (0 <= i && i < 10) ==> arr[i] == 0
-  }) __CPROVER_ensures(__CPROVER_return_value == 0)
+int f1(int *arr)
+  __CPROVER_requires(
+    !__CPROVER_forall {
+      int i;
+      (0 <= i && i < 10) ==> arr[i] == 0
+    }
```
feliperodri, line 7: It'd be nice to prove equivalence of
```
(!__CPROVER_forall(P)) == (__CPROVER_exists(!P))
```

feliperodri: Perhaps out of scope for this PR.

regression/contracts/quantifiers-loop-02/test.desc
```diff
@@ -1,21 +1,22 @@
...
 ^VERIFICATION SUCCESSFUL$
 --
+^warning: ignoring
 --
 This test case checks the handling of a universal quantifier, with a symbolic
 upper bound, within a loop invariant.
 
-The test is tagged:
-- `smt-backend`:
+TODO: This test should:
```
feliperodri, line 16: Is there a GitHub issue to track this?

regression/contracts/quantifiers-forall-requires-02/test.desc
```diff
@@ -1,12 +1,14 @@
...
 --
+^warning: ignoring
 --
 The purpose of this test is to ensure that we can safely use __CPROVER_forall
-in __CPROVER_requires clauses. By using the --replace-all-calls-with-contracts
-flag, goto-instrument will transform the __CPROVER_requires clauses into an
-assertion and the verification remains sound when using __CPROVER_forall.
+within positive contexts (replaced REQUIRES clauses).
+
+This is fully supported (without requiring full unrolling) with the SAT backend.
```
feliperodri: Fix empty newline.

regression/contracts/quantifiers-forall-requires-02/main.c
```diff
@@ -1,15 +1,34 @@
+#include <assert.h>
```
feliperodri: We don't need this assert here.

regression/contracts/quantifiers-forall-requires-01/main.c
```diff
@@ -1,15 +1,25 @@
+#include <assert.h>
```
feliperodri: We don't need this here.

regression/contracts/quantifiers-forall-ensures-02/main.c
```diff
@@ -1,23 +1,29 @@
+#include <stdbool.h>
```
feliperodri: Let's add the assert include for consistency.

regression/contracts/quantifiers-forall-ensures-01/test.desc
```diff
@@ -3,10 +3,12 @@ main.c
...
 --
+^warning: ignoring
 --
 The purpose of this test is to ensure that we can safely use __CPROVER_forall
-in __CPROVER_ensures clauses. By using the --enforce-all-contracts
-flag, goto-instrument will transform the __CPROVER_ensures clauses into an
-assertion and the verification remains sound when using __CPROVER_forall.
+within positive contexts (enforced ENSURES clauses).
+
+This is fully supported (without requiring full unrolling) with the SAT backend.
```
feliperodri: newline.

feliperodri (changes requested):

regression/contracts/quantifiers-loop-02/test.desc
SaswatPadhi: Yes I left a comment here, that we should be handling SMT tags in all regression test directories: https://github.com/diffblue/cbmc/issues/5986#issuecomment-845529645

regression/contracts/quantifiers-nested-05/test.desc
SaswatPadhi: When CBMC (with SAT backend) does not support a quantifier in a particular context, say forall in a negative context, it prints:

```
warning: ignoring forall
```

So I added these negative regexes to make sure we never see that line in the output, otherwise the results would be unsound (since quantifier was completely ignored in the formula).

regression/contracts/quantifiers-exists-ensures-02/main.c
SaswatPadhi: But all loops in CBMC are always unrolled. This one is too, you can see in the CBMC output.

With symbolic bounds the unrolling just never stops, so we give `--unwind` to limit unrolling and then `--unwinding-assertions` to make sure our limited unrolling is sound.

Another way to stop the unrolling is to provide an invariant, and then we just reason about a single arbitrary iteration.

Here, since the loop bounds are constants (0 and 128), the unrolling stops automatically, and is always sound since there is nothing more to unroll (i.e. we didn't limit anything).

regression/contracts/quantifiers-exists-ensures-02/test.desc
SaswatPadhi: May be there is a better way to phrase this. I was being lazy.

I meant to say that `exists x . (L <= x <= U) && P(x)` is not unrolled to `P(L) \/ P(L+1) \/ ... \/ P(U-1) \/ P(U)` in a negative context.

In a negative context symbolic or even no explicit bounds at all would work for `exists`, e.g. `assume(exists x . P(x))` works fine because CBMC just introduces a temporary variable `x` and then `assume(P(x))`.

regression/contracts/quantifiers-nested-05/main.c
SaswatPadhi: You're right, we haven't tested these two quantifiers together. Each regression test currently is checking either universal or existential quantification, but not both.

I would add it in another iteration. I also plan to debug the quick-sort regression test that's marked as a KNOWNBUG.

feliperodri (approved):
LGTM.

regression/contracts/quantifiers-exists-requires-02/test.desc
SaswatPadhi: Done: #6149

feliperodri: https://github.com/diffblue/cbmc/issues/6149

regression/contracts/quantifiers-exists-ensures-02/test.desc
feliperodri: https://github.com/diffblue/cbmc/issues/6149

Issue #6134, opened 2021-05-21
[RFC] SMT Plan


TGWDB:
### Executive Summary

This RFC outlines an implementation plan for adding a standard and easily switchable SMT backend to CBMC. The goal of this work is to address two main areas. The first is to improve performance of CBMC by being able to both: use (runtime) switchable SMT solvers, and be extendable to exploit different theories. The second is for CBMC to have a high quality (clean, clear, robust, extensible, and maintainable) interaction with SMT solvers. This document includes the main details and measures for delivery of the project; the project plan; and key milestones and deliverables to be performed in a mostly iterative manner over vertical slices.

## Project Goals

The motivation for this project is to improve CBMC performance via SMT solvers. This has two main underlying aspects that both need to be addressed for this project to achieve long term success. The first is to allow CBMC to support (runtime) switchable SMT solvers and support multiple different theories. The second is to ensure that the SMT interaction with CBMC is high quality (clean, clear, robust, extensible and maintainable). These motivations drive the rest of the project detailed below.

The goal of this project is thus to allow CBMC to support multiple SMT solvers by using a common interface (SMT2LIB) and communication via a configurable interface (e.g. ports or pipes). The goals relating to the quality of the interaction should be taken into account and is somewhat reflected in the deliverables and overall project plan (incrementally developing a whole new SMT2 interaction rather than simply making small changes to the current SMT translation that is NOT clear, clean, robust, or maintainable).

The measures for delivery of the project are the following:

1. CBMC will have an interface based on SMT2LIB that allows communication with any SMT2LIB compliant SMT solver.
2. CBMC will be able to take the SMT solver as a configuration option and thus allow different SMT2LIB compliant solvers to be chosen at CBMC invocation time (not compile time).
3. CBMC is able to exploit different theories in the SMT solver to improve results. (Note that the current SMT interaction already supports more SMT theories than simply SAT, but this goal is to ensure improvement and parity.)

Note that these are not able to capture the detail of the quality, but quality should be kept in mind as it is one of the motivations of this project.

## Project Deliverables & Milestones

[...]

PR #6151, opened 2021-05-25
Piped process handler for CBMC/SMT solver IPC

NlightNFotis:
This PR is related to #6134 .

This is adding a piped process handler class, with the bulk of it extracted/re-arranged
from `src/memory-analyzer/gdb_api.cpp`.

The idea is that we want to abstract away that interface (since it is code that has been
validated once, and it's known to be working - or have been in the past) and use it to
allow IPC with SMT solvers in the new incremental SMT backend (described in #6134).
Additionally, the `gdb_api.cpp` is going to be adapted at a later time to use this interface.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -108,6 +108,7 @@ jobs:
         run: |
           make -C unit      test
           make -C jbmc/unit test
+          make TAGS="[z3]" -C unit test
       - name: Run expected failure unit tests
         run: |
           make TAGS="[!shouldfail]" -C unit test
@@ -288,7 +289,9 @@ jobs:
       - name: Print ccache stats
         run: ccache -s
       - name: Run unit tests
-        run: cd unit; ./unit_tests
+        run: |
+          cd unit; ./unit_tests
+          ./unit_tests "[z3]"
       - name: Run JBMC unit tests
         run: cd jbmc/unit; ./unit_tests
       - name: Run regression tests
--- a/doc/API/util/piped_process.md
+++ b/doc/API/util/piped_process.md
@@ -1,0 +1,24 @@
+# `src/util/piped_process.{cpp, h}`
+
+To utilise the `piped_process` API for interprocess communication with any binary:
+
+* You need to initialise it by calling the construct `piped_processt("binary with args")`.
+  * If IPC fails before child process creation, you will get a `system_exceptiont`.
+  * If the `binary command` does not correspond to a binary in the `$PATH` or is
+    not a path to a binary itself, you'll read a string `Launching <xyz> failed with error: <error>`
+    when you attempt to `receive()` output from the child process.
+* The child process is automatically killed with SIGTERM when the destructor for
+  the `piped_processt` object is called.
+  * This will occur automatically when the `piped_processt` goes out of scope if
+    it's locally scoped.
+* Use `send()` to send a string message to the child process' input.
+  * This returns a `send_responset`, an enum that shows whether the
+    sending of the message through the pipe succeeded or failed.
+* Use `receive()` to read a string message from the child process' output.
+  * It's always a good idea to guard a call to `receive` with `can_receive()`,
+    so that receiving is blocked until there's something to receive.
+    * `can_receive` with no arguments will default to infinite wait time for piped
+      process readiness.
+  * Alternatively, you can guard the call to `receive` with `wait_receivable`.
+   `wait_receivable` takes an integer value representing `microseconds` of waiting
+   time between checks for pipe readiness.
--- a/src/util/piped_process.cpp
+++ b/src/util/piped_process.cpp
@@ -1,0 +1,262 @@
+/// \file piped_process.cpp
+/// Subprocess communication with pipes.
+/// \author Diffblue Ltd.
+
+#ifdef _WIN32
+// Windows includes go here
+#else
+#  include <fcntl.h>  // library for fcntl function
+#  include <poll.h>   // library for poll function
+#  include <signal.h> // library for kill function
+#  include <unistd.h> // library for read/write/sleep/etc. functions
+#endif
+
+#ifdef _WIN32
+// Unimplemented on windows for now...
+#else
+
+#  include <cstring> // library for strerror function (on linux)
+#  include <iostream>
+#  include <vector>
+
+#  include "exception_utils.h"
+#  include "invariant.h"
+#  include "narrow.h"
+#  include "optional.h"
+#  include "piped_process.h"
+#  include "string_utils.h"
+
+#  define BUFSIZE 2048
+
+piped_processt::piped_processt(const std::vector<std::string> commandvec)
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE("Pipe IPC on windows.")
+#  else
+
+  if(pipe(pipe_input) == -1)
+  {
+    throw system_exceptiont("Input pipe creation failed");
+  }
+
+  if(pipe(pipe_output) == -1)
+  {
+    throw system_exceptiont("Output pipe creation failed");
+  }
+
+  // Default state
+  process_state = statet::NOT_CREATED;
+
+  if(fcntl(pipe_output[0], F_SETFL, O_NONBLOCK) < 0)
+  {
+    throw system_exceptiont("Setting pipe non-blocking failed");
+  }
+
+  // Create a new process for the child that will execute the
+  // command and receive information via pipes.
+  child_process_id = fork();
+  if(child_process_id == 0)
+  {
+    // child process here
+
+    // Close pipes that will be used by the parent so we do
+    // not have our own copies and conflicts.
+    close(pipe_input[1]);
+    close(pipe_output[0]);
+
+    // Duplicate pipes so we have the ones we need.
+    dup2(pipe_input[0], STDIN_FILENO);
+    dup2(pipe_output[1], STDOUT_FILENO);
+    dup2(pipe_output[1], STDERR_FILENO);
+
+    // Create a char** for the arguments (all the contents of commandvec
+    // except the first element, i.e. the command itself).
+    char **args =
+      reinterpret_cast<char **>(malloc((commandvec.size()) * sizeof(char *)));
+    // Add all the arguments to the args array of char *.
+    unsigned long i = 0;
+    while(i < commandvec.size())
+    {
+      args[i] = strdup(commandvec[i].c_str());
+      i++;
+    }
+    args[i] = NULL;
+    execvp(commandvec[0].c_str(), args);
+    // The args variable will be handled by the OS if execvp succeeds, but
+    // if execvp fails then we should free it here (just in case the runtime
+    // error below continues execution.)
+    while(i > 0)
+    {
+      i--;
+      free(args[i]);
+    }
+    free(args);
+    // Only reachable if execvp failed
+    // Note that here we send to std::cerr since we are in the child process
+    // here and this is received by the parent process.
+    std::cerr << "Launching " << commandvec[0]
+              << " failed with error: " << std::strerror(errno) << std::endl;
+    abort();
+  }
+  else
+  {
+    // parent process here
+    // Close pipes to be used by the child process
+    close(pipe_input[0]);
+    close(pipe_output[1]);
+
+    // Get stream for sending to the child process
+    command_stream = fdopen(pipe_input[1], "w");
+    process_state = statet::CREATED;
+  }
+#  endif
+}
+
+piped_processt::~piped_processt()
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE("Pipe IPC on windows: piped_processt constructor")
+#  else
+  // Close the parent side of the remaining pipes
+  fclose(command_stream);
+  // Note that the above will call close(pipe_input[1]);
+  close(pipe_output[0]);
+  // Send signal to the child process to terminate
+  kill(child_process_id, SIGTERM);
+#  endif
+}
+
+piped_processt::send_responset piped_processt::send(const std::string &message)
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE("Pipe IPC on windows: send()")
+#  else
+
+  if(process_state != statet::CREATED)
+  {
+    return send_responset::ERROR;
+  }
+
+  // send message to solver process
+  int send_status = fputs(message.c_str(), command_stream);
+  fflush(command_stream);
+
+  if(send_status == EOF)
+  {
+    return send_responset::FAILED;
+  }
+
+  return send_responset::SUCCEEDED;
+#  endif
+}
+
+std::string piped_processt::receive()
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE("Pipe IPC on windows: receive()")
+#  else
+
+  INVARIANT(
+    process_state == statet::CREATED,
+    "Can only receive() from a fully initialised process");
+
+  std::string response = std::string("");
+  int nbytes;
+  char buff[BUFSIZE];
+
+  while(true)
+  {
+    nbytes = read(pipe_output[0], buff, BUFSIZE);
+    INVARIANT(
+      nbytes < BUFSIZE,
+      "More bytes cannot be read at a time, than the size of the buffer");
+    switch(nbytes)
+    {
+    case -1:
+      // Nothing more to read in the pipe
+      return response;
+    case 0:
+      // Pipe is closed.
+      process_state = statet::STOPPED;
+      return response;
+    default:
+      // Read some bytes, append them to the response and continue
+      response.append(buff, nbytes);
+    }
+  }
+
+  UNREACHABLE;
+#  endif
+}
+
+std::string piped_processt::wait_receive()
+{
+  // can_receive(PIPED_PROCESS_INFINITE_TIMEOUT) waits an ubounded time until
+  // there is some data
+  can_receive(PIPED_PROCESS_INFINITE_TIMEOUT);
+  return receive();
+}
+
+piped_processt::statet piped_processt::get_status()
+{
+  return process_state;
+}
+
+bool piped_processt::can_receive(optionalt<std::size_t> wait_time)
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE(
+    "Pipe IPC on windows: can_receive(optionalt<std::size_t> wait_time)")
+#  else
+  // unwrap the optional argument here
+  const int timeout = wait_time ? narrow<int>(*wait_time) : -1;
+  struct pollfd fds // NOLINT
+  {
+    pipe_output[0], POLLIN, 0
+  };
+  nfds_t nfds = POLLIN;
+  const int ready = poll(&fds, nfds, timeout);
+
+  switch(ready)
+  {
+  case -1:
+    // Error case
+    // Further error handling could go here
+    process_state = statet::ERROR;
+    // fallthrough intended
+  case 0:
+    // Timeout case
+    // Do nothing for timeout and error fallthrough, default function behaviour
+    // is to return false.
+    break;
+  default:
+    // Found some events, check for POLLIN
+    if(fds.revents & POLLIN)
+    {
+      // we can read from the pipe here
+      return true;
+    }
+    // Some revent we did not ask for or check for, can't read though.
+  }
+  return false;
+#  endif
+}
+
+bool piped_processt::can_receive()
+{
+  return can_receive(0);
+}
+
+void piped_processt::wait_receivable(int wait_time)
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE("Pipe IPC on windows: wait_stopped(int wait_time)")
+#  else
+  while(process_state == statet::CREATED && !can_receive(0))
+  {
+    usleep(wait_time);
+  }
+#  endif
+}
+
+#endif
--- a/src/util/piped_process.h
+++ b/src/util/piped_process.h
@@ -1,0 +1,99 @@
+/// \file piped_process.h
+/// Subprocess communication with pipes
+/// \author Diffblue Ltd.
+
+#ifndef CPROVER_UTIL_PIPED_PROCESS_H
+#define CPROVER_UTIL_PIPED_PROCESS_H
+
+#ifdef _WIN32
+// TODO: Windows definitions go here
+#else
+
+#  include <vector>
+
+#  include "optional.h"
+
+#  define PIPED_PROCESS_INFINITE_TIMEOUT                                       \
+    optionalt<std::size_t>                                                     \
+    {                                                                          \
+    }
+
+class piped_processt
+{
+public:
+  /// Enumeration to keep track of child process state.
+  enum class statet
+  {
+    NOT_CREATED,
+    CREATED,
+    STOPPED,
+    ERROR
+  };
+
+  /// Enumeration for send response.
+  enum class send_responset
+  {
+    SUCCEEDED,
+    FAILED,
+    ERROR
+  };
+
+  /// Send a string message (command) to the child process.
+  /// \param message The string message to be sent.
+  /// \return
+  send_responset send(const std::string &message);
+  /// Read a string from the child process' output.
+  /// \return a string containing data from the process, empty string if no data
+  std::string receive();
+  /// Wait until a string is available and read a string from the child
+  /// process' output.
+  /// \return a string containing data from the process, empty string if no data
+  std::string wait_receive();
+
+  /// Get child process status.
+  /// \return a statet representing the status of the child process
+  statet get_status();
+
+  /// See if this process can receive data from the other process.
+  /// \param wait_time Amount of time to wait before timing out, with:
+  ///        * positive integer being wait time in milli-seconds,
+  ///        * 0 signifying non-blocking immediate return, and
+  ///        * an empty optional representing infinite wait time.
+  /// \return true is there is data to read from process, false otherwise
+  bool can_receive(optionalt<std::size_t> wait_time);
+
+  /// See if this process can receive data from the other process.
+  /// Note this calls can_receive(0);
+  /// \return true if there is data to read from process, false otherwise.
+  bool can_receive();
+
+  /// Wait for the pipe to be ready, waiting specified time between
+  /// checks. Will return when the pipe is ready or the other process
+  /// is not in a statet::CREATED state (i.e. error has occured).
+  /// \param wait_time Time spent in usleep() (microseconds) between checks
+  //         of can_receive(0)
+  void wait_receivable(int wait_time);
+
+  /// Initiate a new subprocess with pipes supporting communication
+  /// between the parent (this process) and the child.
+  /// \param commandvec The command and arguments to create the process
+  explicit piped_processt(const std::vector<std::string> commandvec);
+
+  ~piped_processt();
+
+protected:
+  // Child process ID.
+  pid_t child_process_id;
+  FILE *command_stream;
+  // The member fields below are so named from the perspective of the
+  // parent -> child process. So `pipe_input` is where we are feeding
+  // commands to the child process, and `pipe_output` is where we read
+  // the results of execution from.
+  int pipe_input[2];
+  int pipe_output[2];
+  statet process_state;
+};
+
+#endif // endif _WIN32
+
+#endif // endifndef CPROVER_UTIL_PIPED_PROCESS_H
--- a/unit/util/piped_process.cpp
+++ b/unit/util/piped_process.cpp
@@ -1,0 +1,270 @@
+/// \file
+/// \author Diffblue Ltd.
+/// Unit tests for checking the piped process communication mechanism.
+
+#ifdef _WIN32
+// No unit tests yet!
+#else
+
+#  include <testing-utils/use_catch.h>
+#  include <util/optional.h>
+#  include <util/piped_process.h>
+#  include <util/string_utils.h>
+
+TEST_CASE(
+  "Creating a sub process and reading its output.",
+  "[core][util][piped_process]")
+{
+  const std::string to_be_echoed = "The Jabberwocky";
+  // Need to give path to avoid shell built-in invocation
+  std::vector<std::string> commands;
+  commands.push_back("/bin/echo");
+  commands.push_back(to_be_echoed);
+  piped_processt process = piped_processt(commands);
+
+  // This is an indirect way to detect when the pipe has something. This
+  // could (in theory) also return when there is an error, but this unit
+  // test is not doing error handling.
+  process.can_receive(PIPED_PROCESS_INFINITE_TIMEOUT);
+  std::string response = strip_string(process.receive());
+
+  REQUIRE(response == to_be_echoed);
+}
+
+TEST_CASE(
+  "Creating a sub process with a binary that doesn't exist.",
+  "[core][util][piped_process]")
+{
+  const std::string expected_error("Launching abcde failed");
+  std::vector<std::string> commands;
+  commands.push_back("abcde");
+  piped_processt process = piped_processt(commands);
+
+  // This is an indirect way to detect when the pipe has something. This
+  // could (in theory) also return when there is an error, but this unit
+  // test is not doing error handling.
+  process.can_receive(PIPED_PROCESS_INFINITE_TIMEOUT);
+  std::string response = process.receive();
+  // This tracks how many times we tried, if for some reason we are stuck
+  // give up eventually for this test.
+  int too_many_tries = 0;
+  // The expected length of the output string is 6 characters, keep
+  // trying up to the retry limit of 5 or the string is long enough
+  while(too_many_tries < 5 && response.length() < 64)
+  {
+    // Wait a short amount of time to try and receive
+    process.can_receive(10);
+    response += process.receive();
+    too_many_tries++;
+  }
+
+  REQUIRE(response.find(expected_error) < response.length() - 5);
+}
+
+TEST_CASE(
+  "Creating a sub process of z3 and read a response from an echo command.",
+  "[core][util][piped_process][.z3]")
+{
+  std::vector<std::string> commands;
+  commands.push_back("z3");
+  commands.push_back("-in");
+  piped_processt process = piped_processt(commands);
+
+  REQUIRE(
+    process.send("(echo \"hi\")\n") ==
+    piped_processt::send_responset::SUCCEEDED);
+
+  process.can_receive(PIPED_PROCESS_INFINITE_TIMEOUT);
+  std::string response = strip_string(process.receive());
+  REQUIRE(response == "hi");
+
+  REQUIRE(
+    process.send("(exit)\n") == piped_processt::send_responset::SUCCEEDED);
+}
+
+TEST_CASE(
+  "Creating a sub process and interacting with it.",
+  "[core][util][piped_process][.z3]")
+{
+  std::vector<std::string> commands;
+  commands.push_back("z3");
+  commands.push_back("-in");
+  const std::string termination_statement = "(exit)\n";
+  piped_processt process = piped_processt(commands);
+
+  REQUIRE(
+    process.send("(echo \"hi\")\n") ==
+    piped_processt::send_responset::SUCCEEDED);
+
+  process.can_receive(PIPED_PROCESS_INFINITE_TIMEOUT);
+  std::string response = strip_string(process.receive());
+  REQUIRE(response == "hi");
+
+  std::string statement = std::string("(echo \"Second string\")\n");
+  REQUIRE(process.send(statement) == piped_processt::send_responset::SUCCEEDED);
+
+  process.can_receive(PIPED_PROCESS_INFINITE_TIMEOUT);
+  response = strip_string(process.receive());
+  REQUIRE(response == "Second string");
+
+  REQUIRE(
+    process.send(termination_statement) ==
+    piped_processt::send_responset::SUCCEEDED);
+}
+
+TEST_CASE(
+  "Use a created piped process instance of z3 to solve a simple SMT problem",
+  "[core][util][piped_process][.z3]")
+{
+  std::vector<std::string> commands;
+  commands.push_back("z3");
+  commands.push_back("-in");
+  commands.push_back("-smt2");
+  piped_processt process = piped_processt(commands);
+
+  std::string message =
+    "(set-logic QF_LIA) (declare-const x Int) (declare-const y Int) (assert (> "
+    "(+ (mod x 4) (* 3 (div y 2))) (- x y)))  (check-sat)\n";
+  REQUIRE(process.send(message) == piped_processt::send_responset::SUCCEEDED);
+
+  process.can_receive(PIPED_PROCESS_INFINITE_TIMEOUT);
+  std::string response = strip_string(process.receive());
+  REQUIRE(response == "sat");
+
+  REQUIRE(
+    process.send("(exit)\n") == piped_processt::send_responset::SUCCEEDED);
+}
+
+TEST_CASE(
+  "Use a created piped process instance of z3 to solve a simple SMT problem "
+  "with wait_receive",
+  "[core][util][piped_process][.z3]")
+{
+  std::vector<std::string> commands;
+  commands.push_back("z3");
+  commands.push_back("-in");
+  commands.push_back("-smt2");
+  piped_processt process = piped_processt(commands);
+
+  std::string statement =
+    "(set-logic QF_LIA) (declare-const x Int) (declare-const y Int) (assert (> "
+    "(+ (mod x 4) (* 3 (div y 2))) (- x y)))  (check-sat)\n";
+
+  REQUIRE(process.send(statement) == piped_processt::send_responset::SUCCEEDED);
+
+  std::string response = strip_string(process.wait_receive());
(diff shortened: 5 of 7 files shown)


src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
+/// \file Subprocess communication with pipes
+///
+/// \author Diffblue Ltd.
+
+#ifdef _WIN32
+// Windows includes go here
```
martin-cs, line 6: FWIW I think this PR either needs the windows code OR a deliberate and acknowledged decision that some functionality is simply not going to be supported on Windows.  I'm fine with either but I think this PR is the decision point.

( Or, maybe, a "Linux" build on WSL as the "windows version")

src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
...
+#endif
+
+#include "invariant.h"
+#include "piped_process.h"
+
+piped_processt::piped_processt(const std::string &command)
+{
+#ifdef _WIN32
+  // This should use the new error state from PR #6131 once that is done
+  INVARIANT(false, "New SMT2 backend WIP: Windows piped_process constructor.");
```
martin-cs: I have given @thomasspriggs a hard time about `INVARIANT(false, "...")`, please can we use something else?  My objection is that it is missing the point of what an invariant is and that it is not supposed to be able to be triggered by user input.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+char **piped_processt::split_command_args(const std::string &command)
+{
+  char **res = NULL;
+  int n_spaces = 0;
+  char *p = strtok(strdup(command.c_str()), " ");
+
+  while(p)
+  {
+    res = reinterpret_cast<char **>(realloc(res, sizeof(char *) * ++n_spaces));
```
martin-cs: Couldn't you have a `std::vector<char *>` to do the segmentation and then create the `char *` array at the end?

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
+#ifndef CPROVER_UTIL_PIPED_PROCESS_H
+#define CPROVER_UTIL_PIPED_PROCESS_H
+
+// TODO: Revisit buffer size and storage location.
+// Note that this may impact number of characters correctly communicated
+// by the pipe, to check/fix.
```
martin-cs: It shouldn't affect correctness, if it does, yeah, please reconsider!

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
...
+
+protected:
+  FILE *response_stream;
+  FILE *command_stream;
+  int pipe_input[2];
+  int pipe_output[2];
+  process_statet process_state;
+
+public:
+  bool send(const std::string &message);
```
martin-cs: In some implementations of the STL it used to be possible to create an `ostream` or an `istream` based on a file descriptor.  If that were possible it would give a much cleaner and more consistent interface to the pipe.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,98 @@
...
+  "We create a pipe and we can read from it",
+  "[core][util][piped_process]")
+{
+  std::string to_be_echoed = "The Jabberwocky";
+  // Need to give up to avoid shell built-in invocation
+  std::string binary = "/bin/echo";
+  std::string command = binary + " " + to_be_echoed;
+  piped_processt process = piped_processt(command);
+
+  sleep(1);
```
martin-cs: ???

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+TEST_CASE(
+  "We create a pipe, send and receive from it",
+  "[core][util][piped_process]")
+{
+  std::string binary = "z3 -in";
+  std::string statement = "(echo \"hi\")";
+  std::string termination_statement = "(exit)";
+  piped_processt process = piped_processt(binary);
+  // Wait a moment for z3 to start
```
martin-cs: Eh?  Why?  Pipes should be reliable and asynchronous.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,98 @@
...
+  REQUIRE(response.length() == 2);
+
+  // Tell z3 to terminate
+  res = process.send(termination_statement);
+  REQUIRE(res == true);
+}
+
+TEST_CASE("We create a pipe, interact", "[core][util][piped_process]")
+{
+  std::string binary = "z3 -in";
```
martin-cs: Could we use CPROVER's SMT solver instead of Z3 so that the unit tests don't add another dependency?

src/util/piped_process.cpp
thomasspriggs: Yep, this is a draft PR. I think the plan is to use the macro in - https://github.com/diffblue/cbmc/pull/6152
Presumably this branch doesn't have it as it wasn't ready at the time of writing.

src/util/piped_process.cpp
thomasspriggs: We are planning to do implementation for both Windows and non-windows OS. We are just doing the non-windows support *first*. I guess splitting up the PRs vs a single larger PR is a separate discussion.

src/util/piped_process.h
thomasspriggs: :pick: Can we have some doxygen please. Especially to document what true/false means for the return value. Add doxygen comment also applies to everything with a :mechanical_leg: emoji

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
...
+  std::string receive();
+
+  process_statet get_status()
```
thomasspriggs: :mechanical_leg:

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
...
+  {
+    return process_state;
+  }
+
+  piped_processt(const std::string &command);
```
thomasspriggs: :mechanical_leg:

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
...
+class piped_processt
+{
+  enum class process_statet
+  {
+    NOT_CREATED,
+    CREATED,
+    STOPPED
+  };
+
+  char **split_command_args(const std::string &command);
```
thomasspriggs: :hammer_and_pick: This member function doesn't appear to need the state of the class. Can this be refactored into a free function? That would make the class cleaner / smaller and make the function unit-testable in isolation.

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
...
+
+protected:
+  FILE *response_stream;
+  FILE *command_stream;
+  int pipe_input[2];
+  int pipe_output[2];
```
thomasspriggs, line 93: :bulb: These streams and pipes are operating system specific implementation details. It might be cleaner to use the "PImpl" coding pattern to keep all the `#ifdef _WIN32` stuff in the `.cpp` file. This should: 

- Make the header file easier to read for programmers of other parts of the code base who just want to use this API.
- Mean that we can keep the windows specific includes in the `.cpp` file, if we need fields with types based on `windows.h` when we come to do the windows support. This should also reduce compile times for the rest of the code base by avoiding putting `windows.h` in our header file.

This seems like a reasonable introduction to the code pattern if you aren't familiar with it - https://en.cppreference.com/w/cpp/language/pimpl We should be able to just leave off the `std::experimental::propagate_const` stuff it mentions.

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
+#ifndef CPROVER_UTIL_PIPED_PROCESS_H
+#define CPROVER_UTIL_PIPED_PROCESS_H
+
+// TODO: Revisit buffer size and storage location.
+// Note that this may impact number of characters correctly communicated
+// by the pipe, to check/fix.
+#define BUFSIZE 2048
```
thomasspriggs: :grey_question: Is this just an implementation detail that should be in the `.cpp` file? I am thinking about the increased compilation time for anyone tuning this parameter later.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if(res == NULL)
+      exit(-1); /* memory allocation failed */
```
thomasspriggs: :pick: `INVARIANT`

src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
...
+#endif
+}
+
+char **piped_processt::split_command_args(const std::string &command)
+{
+  char **res = NULL;
+  int n_spaces = 0;
+  char *p = strtok(strdup(command.c_str()), " ");
+
+  while(p)
```
thomasspriggs: :bulb: It might be cleaner to reuse `split_string` from `src/util/string_utils.h`

src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
...
+    close(pipe_input[1]);
+    close(pipe_output[0]);
+
+    // Duplicate pipes so we have the ones we need.
+    dup2(pipe_input[0], STDIN_FILENO);
+    dup2(pipe_output[1], STDOUT_FILENO);
+    dup2(pipe_output[1], STDERR_FILENO);
+
+    // Create the arguments to execvp from the construction string
+    char **args = split_command_args(command);
```
thomasspriggs: :question: Where does this raw pointer get freed up?

thomasspriggs (commented):
Some initial comments.

unit/util/piped_process.cpp
TGWDB: Pipes are asynchronous, but the pipes here are non-blocking. This means if you try to read immediately the other process may have not written anything to the pipe yet and you can read nothing (and then fail the following test).

Having the ability to read from a pipe that is not ended/closed is significant to being able to interact (e.g. with a solver) without requiring it to be terminated/closed after each request for a response. That said, without an explicit synchronisation system (yet?) it will not always be clear when the pipe is ready to be read. An alternate is to read lines, but this requires knowing how many lines to read, which may not be known in advance (e.g. how many lines make up a trace?) or doing parsing/checking after each line to know when to continue/stop. (This PR is clearly WIP, mostly I believe created to test the CI and various builds, it is not ready for review or merging.)

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
...
+
+class piped_processt
+{
+  enum class process_statet
```
thomasspriggs: :grey_question: Given that the public `get_status` member function returns a `process_statet`, shouldn't the definition of a `process_statet` also be public?

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
...
+  int pipe_input[2];
+  int pipe_output[2];
+  process_statet process_state;
+
+public:
+  bool send(const std::string &message);
+  std::string receive();
+
+  process_statet get_status()
+  {
```
thomasspriggs: :pick: Can we keep the implementation in the `.cpp`, even if it is short?

src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
...
+    // Execute the command
+    execvp(args[0], args);
+    // Only reachable if execvp failed
+    throw std::runtime_error(
+      "Launching \"" + command + "\" failed with error: " + strerror(errno));
+  }
+  else
+  {
+    // parent process here
+    // Close pipes to be used by the child process
```
thomasspriggs, line 104: :grey_question: Is this closing whole pipes or just access to specific ends of specific pipes by specific processes?

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#  include <testing-utils/use_catch.h>
+#  include <unistd.h>
+#  include <util/piped_process.h>
+
+TEST_CASE(
+  "We create a pipe and we can read from it",
+  "[core][util][piped_process]")
+{
+  std::string to_be_echoed = "The Jabberwocky";
```
thomasspriggs: :pick: `const` for all variables which are not mutated throughout this test, please.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,98 @@
...
+#ifdef _WIN32
+// No unit tests yet!
+#else
+
+#  include <testing-utils/use_catch.h>
+#  include <unistd.h>
+#  include <util/piped_process.h>
+
+TEST_CASE(
+  "We create a pipe and we can read from it",
```
thomasspriggs:
```suggestion
  "Creating a sub process and reading its output.",
```

src/util/piped_process.h
thomasspriggs: :pick: `explicit`

src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  // send message to solver process
+  int send_status = fputs(message.c_str(), command_stream);
+  fflush(command_stream);
+
+  if(send_status == EOF)
+  {
+    // Some kind of error occured, maybe we should update the
+    // solver status here?
```
thomasspriggs: :pick: As this layer is not intended to be solver specific, we probably shouldn't explain how it works in terms of the solver.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
...
+    throw std::runtime_error("Input pipe creation failed");
+  }
+
+  if(pipe(pipe_output) == -1)
+  {
+    throw std::runtime_error("Output pipe creation failed");
+  }
+
+  // Default state
+  process_state = process_statet::NOT_CREATED;
```
thomasspriggs: :pick: Might be cleaner to put this field initialisation into the constructors initialiser list.

src/util/piped_process.h
```diff
@@ -0,0 +1,39 @@
...
+// TODO: Revisit buffer size and storage location.
+// Note that this may impact number of characters correctly communicated
+// by the pipe, to check/fix.
+#define BUFSIZE 2048
+
+class piped_processt
+{
+  enum class process_statet
+  {
+    NOT_CREATED,
```
thomasspriggs, line 27: :bulb: Can we get rid of this `NOT_CREATED` status, based on the `piped_processt` not being constructed if the construction fails? This status is unreachable on successful return of the constructor, based on all error conditions throwing an exception. The `CREATED` state could then be renamed to `RUNNING`, or we could simplify things even further by using an `is_stopped` bool instead of an enum.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,168 @@
...
+    nbytes = read(pipe_output[0], buff, BUFSIZE);
+    switch(nbytes)
+    {
+    case -1:
+      // Nothing more to read in the pipe
+      return response;
+    case 0:
+      // Pipe is closed.
+      process_state = process_statet::STOPPED;
+      if(response == std::string(""))
```
thomasspriggs: :no_entry_sign: This check currently does nothing and the function would work the same way with it removed. This is because returning NULL will construct an empty string, when the return type is `std::string`. I consider this to be a blocking comment as a no-op check is highly misleading to the reader.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,98 @@
+/// \file
+/// \author Diffblue Ltd.
+/// Unit tests for checking the piped process communication mechanism.
+
+#ifdef _WIN32
+// No unit tests yet!
+#else
+
+#  include <testing-utils/use_catch.h>
+#  include <unistd.h>
```
thomasspriggs: :no_entry_sign: The fact that we need an operating system specific include here in the unit tests says to me that our abstraction is leaky and that we should reconsider how to avoid it being needed.

src/util/string_utils.cpp
```diff
@@ -1,3 +1,5 @@
+
+
```
TGWDB: Should we be adding extra whitespace/newlines?

src/util/piped_process.cpp
thomasspriggs: I think I would have split `std::string` into `std::vector<string>` and then converted that into the required `char**` style array.

src/util/piped_process.cpp
TGWDB: Specific ends of specific pipes for the process involved.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  // send message to solver process
+  int send_status = fputs(message.c_str(), command_stream);
+  fflush(command_stream);
+
+  if(send_status == EOF)
+  {
+    // Some kind of error occured, maybe we should update the
+    // process status here?
```
TGWDB: Maybe this should be done now.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,233 @@
...
+  switch(ready)
+  {
+  case -1:
+    // Error case
+    // Further error handling could go here
+    process_state = process_statet::ERROR;
+    // fallthrough intended
+  case 0:
+    // Timeout case
+    // Do nothing for timeout and error fallthrough, default is return false.
```
TGWDB: "default" -> "function defaul behaviour"

src/util/piped_process.cpp
chrisr-diffblue: This just closes one end of the pipe.

src/util/string_utils.cpp
```diff
@@ -252,3 +255,23 @@ std::string wrap_line(
 
   return result;
 }
+
+char **string_to_cstr_array(const std::string &s, const char *delims)
```
chrisr-diffblue: I think this should be named better - the name doesn't really imply that this is actually splitting the string... Also, given what this is used for (splitting a command string intro arguments) it's probably not complete/safe either - it doesn't handle any kind of quoting of spaces which would/might be needed for safety... e.g. if the command string is something like "mysolver --config-file '/path/with spaces/config'" I think this would break....

src/util/piped_process.cpp
```diff
@@ -0,0 +1,233 @@
...
+  UNIMPLEMENTED_FEATURE("Pipe IPC on windows: receive()")
+#else
+
+  INVARIANT(
+    process_state == process_statet::CREATED,
+    "Can only receive() from a fully initialised process");
+
+  // This is necessary to ensure the buffer is synced with the
+  // latest data.
+  fsync(pipe_output[0]);
```
chrisr-diffblue: Rather than using explicit syncs/flushes, is it possible to put the pipe into non-buffered mode? There should be an ioctl operation on the pipe for that, I think....

src/util/piped_process.h
kroening: I'd get rid of `split_command_args` -- let the user pass in a vector of strings, i.e., using the interface of `run`.

chrisr-diffblue: That would be my preference too - though later on we'll still want this argument splitting for allowing the user to specify the solver command to use. Also worth noting that for windows support, we may need to be able to re-construct a single (correctly quoted) string representing the command to run.

unit/util/piped_process.cpp
NlightNFotis: This was temporary while we were iterating to stabilise the tests.

All of the instances of `sleep` have been removed from the current iteration of the test code.

src/util/piped_process.h
NlightNFotis: It doesn't, we have verified it.

unit/util/piped_process.cpp
NlightNFotis: The problem is that depending on the build type, the location of the SMT solver will be different (in an out-of-source cmake build vs. a make one) so this looks hairy to implement.

We will probably revisit this decision in a bit.

src/util/piped_process.cpp
NlightNFotis: There are two cases for this:

1) If `execvp` succeeds, it's not handled anywhere - and it doesn't need to, for the simple reason that the process image changes to the binary invoked in `exec` so the pointer is no longer present there.
2) In case `execvp` fails, it's freed later on, but there could be an argument made that it wouldn't need to be handled there as well, as the failure of execvp is leading to a terminal exception.

src/util/piped_process.h
NlightNFotis: Yes, done.

src/util/piped_process.h
NlightNFotis: Done.

src/util/piped_process.cpp
NlightNFotis: We deleted `fsync` as we found out we don't really need it.

src/util/piped_process.h
martin-cs: https://gist.github.com/rajatkhanduja/2012695 gives an example of this.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,256 @@
...
+  const std::string termination_statement = "(exit)\n";
+  piped_processt process = piped_processt(binary);
+
+  piped_processt::process_send_responset res = process.send(statement);
+  REQUIRE(res == piped_processt::process_send_responset::SUCCEEDED);
+
+  process.can_receive(-1);
+  std::string response = process.receive();
+  response.erase(
+    std::remove(response.begin(), response.end(), '\n'), response.end());
```
thomasspriggs: :pick: If you use the existing `strip_string` function from `src/util/string_utils.h` then the resulting code will be shorter and it will be easier to read, because it will be stating the intent rather than the algorithms.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,256 @@
...
+  REQUIRE(response == std::string("hi"));
```
thomasspriggs: :pick: `std::string("hi")` could be just `"hi"`, because the standard template library defines the `==` operator for and left hand side of `std::string` and a right hand side of `const char *`

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,256 @@
...
+  REQUIRE(response.length() == 2);
```
thomasspriggs: :pick: This requirement is redundant and should be removed because the requirement above will already check that `response` has length 2 because strings are only equal if they have the same length.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,256 @@
...
+TEST_CASE(
+  "We create a pipe, send and receive from it",
+  "[core][util][piped_process]")
+{
+  const std::string binary = "z3 -in";
+  const std::string statement = "(echo \"hi\")\n";
+  const std::string termination_statement = "(exit)\n";
+  piped_processt process = piped_processt(binary);
+
+  piped_processt::process_send_responset res = process.send(statement);
```
thomasspriggs: :pick: The coding standards say that we should avoid abbreviations. On this basis `res` should be `response`. However I'd go a step further and call this `statement_response` because is is the result of sending the statement. This is assuming that you keep this variable rather than acting on my other suggestion of inlining `process.send(statement)` into the requirement.
:pick: `const`.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,256 @@
...
+  REQUIRE(res == piped_processt::process_send_responset::SUCCEEDED);
+
+  process.can_receive(-1);
+  std::string response = process.receive();
+  response.erase(
+    std::remove(response.begin(), response.end(), '\n'), response.end());
+  REQUIRE(response == std::string("hi"));
+  REQUIRE(response.length() == 2);
+
+  res = process.send(termination_statement);
```
thomasspriggs: :pick: Can we have a different `const` variable with a different name, rather than reusing an existing variable?

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,256 @@
...
+  "We create a pipe, send and receive from it",
+  "[core][util][piped_process]")
+{
+  const std::string binary = "z3 -in";
+  const std::string statement = "(echo \"hi\")\n";
+  const std::string termination_statement = "(exit)\n";
+  piped_processt process = piped_processt(binary);
+
+  piped_processt::process_send_responset res = process.send(statement);
+  REQUIRE(res == piped_processt::process_send_responset::SUCCEEDED);
```
thomasspriggs: :bulb: If you inline `process.send(statement)` into this requirement, then catch will give more informative error messages in the case where this fails. This is because the error messaging will show `process.send(statement)` on that side of the equals instead of just `res`. `REQUIRE(process.send(statement) == piped_processt::process_send_responset::SUCCEEDED);`

src/util/piped_process.h
```diff
@@ -0,0 +1,70 @@
...
+  std::string receive();
+  /// Wait until a string is available and read a string from the child
+  /// process' output.
+  std::string wait_receive();
+
+  /// Get child process status.
+  process_statet get_status();
+
+  /// See if this process can receive data from the other process.
+  /// \param timeout Amount of time to wait before timing out and
```
thomasspriggs: :question: What unit of time is `timeout` in? This should be documented, if not part of the parameter name.

src/util/piped_process.h
```diff
@@ -0,0 +1,70 @@
...
+  ///     returning: -1 is infiniute, 0 is non-blocking immediate.
```
thomasspriggs: Using the magic value of `-1` is leaking the implementation detail, that `poll` uses `-1` to mean infinite waiting. I'd make this more expressive / API agnostic by using parameter name and type `optionalt<std::size_t> seconds_waiting_limit`. Note the use of an unsigned type to force users to use empty when they want unlimited waiting.

src/util/piped_process.h
thomasspriggs: This function is gone. Therefore I am going to consider this  comment to be resolved.

src/util/piped_process.h
```diff
@@ -0,0 +1,94 @@
...
+  bool can_receive(optionalt<std::size_t> wait_time);
+
+  /// See if this process can receive data from the other process.
+  /// Note this calls can_receive(0);
+  /// \return true if there is data to read from process, false otherwise.
+  bool can_receive();
+
+  /// Wait for the pipe to be ready, waiting specified time between
+  /// checks. Will return when the pipe is ready or the other process
+  /// is not in a process_statet::CREATED state (i.e. error has occured).
```
thomasspriggs: :pick: The name `process_statet` in this doxygen is now out of date. It should either just be `statet`, or `piped_processt::statet` if we want to be more specific.

src/util/piped_process.cpp
thomasspriggs: Looks like this has been fixed.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,272 @@
...
+/// Unit tests for checking the piped process communication mechanism.
+
+#ifdef _WIN32
+// No unit tests yet!
+#else
+
+#  include <testing-utils/use_catch.h>
+#  include <util/optional.h>
+#  include <util/piped_process.h>
+#  include <util/string_utils.h>
```
thomasspriggs, line 12: :grey_question: Won't these 4 includes be required regardless of whether the tests are being run on Windows or not?

src/util/piped_process.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+  UNREACHABLE;
+#  endif
+}
+
+std::string piped_processt::wait_receive()
+{
+  // can_receive(PIPED_PROCESS_INFINITE_TIMEOUT) waits an ubounded time until
+  // there is some data
+  can_receive(PIPED_PROCESS_INFINITE_TIMEOUT);
```
thomasspriggs, line 196: :pick: Seems like fairly gratuitous macro usage to me. Why not just use an ordinary constant, defined in either the function or the class. Such as - `const auto infinite_timeout = optionalt<std::size_t>{};`

src/util/piped_process.h
```diff
@@ -0,0 +1,94 @@
...
+  /// \param wait_time Time spent in usleep() between checks of can_receive(0)
```
thomasspriggs: :pick: Please can you document the unit of time here? I think I can safely assume this isn't [Planck Time](https://simple.wikipedia.org/wiki/Planck_time), but is it seconds, microseconds, milliseconds, etc?

src/util/piped_process.h
thomasspriggs: Thanks for implementing this change.

scripts/expected_doxygen_warnings.txt
```diff
@@ -3,7 +3,7 @@ warning: Included by graph for 'goto_model.h' not generated, too many nodes (95)
 warning: Included by graph for 'arith_tools.h' not generated, too many nodes (167), threshold is 60. Consider increasing DOT_GRAPH_MAX_NODES.
 warning: Included by graph for 'c_types.h' not generated, too many nodes (128), threshold is 60. Consider increasing DOT_GRAPH_MAX_NODES.
 warning: Included by graph for 'config.h' not generated, too many nodes (79), threshold is 60. Consider increasing DOT_GRAPH_MAX_NODES.
-warning: Included by graph for 'invariant.h' not generated, too many nodes (172), threshold is 60. Consider increasing DOT_GRAPH_MAX_NODES.
+warning: Included by graph for 'invariant.h' not generated, too many nodes (173), threshold is 60. Consider increasing DOT_GRAPH_MAX_NODES.
```
thomasspriggs: :exclamation: If you rebase on the latest version of the `develop` branch, then this change will no longer be required.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,264 @@
...
+/// Subprocess communication with pipes.
+/// \author Diffblue Ltd.
+
+#ifdef _WIN32
+// Windows includes go here
+#else
+#  include <fcntl.h>  // library for fcntl function
+#  include <poll.h>   // library for poll function
+#  include <signal.h> // library for kill function
+#  include <string.h> // library for strerror function (on linux)
```
thomasspriggs: :pick: As we are working in C++ rather than C, we should be including `cstring`, not `string.h`. The code should then refer to `std::strerror`, rather than just `strerror` on its own.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,264 @@
...
+    process_state == statet::CREATED,
+    "Can only receive() from a fully initialised process");
+
+  std::string response = std::string("");
+  int nbytes;
+  char buff[BUFSIZE];
+
+  while(true)
+  {
+    nbytes = read(pipe_output[0], buff, BUFSIZE);
```
thomasspriggs, line 169: :bulb: `INVARIANT(nbytes < BUFSIZE), "More bytes cannot be read at a time, than the size of the buffer");`

src/util/piped_process.cpp
```diff
@@ -0,0 +1,264 @@
...
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE(
+    "Pipe IPC on windows: can_receive(optionalt<std::size_t> wait_time)")
+#  else
+  // unwrap the optional argument here
+  int timeout;
+  if(wait_time.has_value())
+  {
+    timeout = wait_time.value();
```
thomasspriggs: :pick: This implicit cast should be checked for data loss - `narrow<int>(wait_time.value())`.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+bool piped_processt::can_receive(optionalt<std::size_t> wait_time)
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE(
+    "Pipe IPC on windows: can_receive(optionalt<std::size_t> wait_time)")
+#  else
+  // unwrap the optional argument here
+  int timeout;
+  if(wait_time.has_value())
```
thomasspriggs: :bulb: This entire if statement could be expressed more succinctly and with added `const`, by using the ternary operator. `const int timeout = wait_time ? narrow<int>(*wait_time) : -1;`

src/util/piped_process.cpp
```diff
@@ -0,0 +1,264 @@
...
+    timeout = -1;
+  }
+
+  int ready;
+  struct pollfd fds // NOLINT
+  {
+    pipe_output[0], POLLIN, 0
+  };
+  nfds_t nfds = POLLIN;
+  ready = poll(&fds, nfds, timeout);
```
thomasspriggs: :pick: I would prefer that `ready` was defined where it is first assigned - `const int ready = poll(&fds, nfds, timeout);`

src/util/piped_process.h
```diff
@@ -0,0 +1,94 @@
...
+
+  /// Initiate a new subprocess with pipes supporting communication
+  /// between the parent (this process) and the child.
+  /// \param commandvec The command and arguments to create the process
+  explicit piped_processt(const std::vector<std::string> commandvec);
+
+  ~piped_processt();
+
+protected:
+  // Child process ID.
```
thomasspriggs, line 85: :bulb: Make the comment the variable name, instead of using a separate comment! This is one of the easiest ways to make code easier to read. In this case that would mean renaming `pid` to `child_process_id`.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,272 @@
...
+TEST_CASE(
+  "Creating a sub process and reading its output.",
+  "[core][util][piped_process]")
+{
+  const std::string to_be_echoed = "The Jabberwocky";
+  // Need to give path to avoid shell built-in invocation
+  std::vector<std::string> commands;
+  commands.push_back("/bin/echo");
+  commands.push_back(to_be_echoed);
+  piped_processt process = piped_processt(commands);
```
thomasspriggs, line 23: :bulb: The command vector could be initialised inline in order to avoid mutating a vector with `push_back`. For example - `piped_processt process{{"/bin/echo", to_be_echoed}};`

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+  // This is an indirect way to detect when the pipe has something since
+  // -1 is an infinite timeout wait. This could (in theory) also return
```
thomasspriggs: :pick: This comment does not match the current API being used. As we are using empty rather than `-1`.

thomasspriggs (approved):
My remaining comments amount to nit picks and ideas for improvement. I think this looks pretty good overall. My preference would be to get this PR merged sooner, rather than later and to perform any further improvements as follow up PRs.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+piped_processt::piped_processt(const std::vector<std::string> commandvec)
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE("Pipe IPC on windows.")
+#  else
+
+  if(pipe(pipe_input) == -1)
+  {
+    throw std::runtime_error("Input pipe creation failed");
```
peterschrammel: 💡 use `system_exceptiont` from `util/exception_utils.h`.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  if(pipe(pipe_output) == -1)
+  {
+    throw std::runtime_error("Output pipe creation failed");
```
peterschrammel: 💡

src/util/piped_process.cpp
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  // Default state
+  process_state = statet::NOT_CREATED;
+
+  if(fcntl(pipe_output[0], F_SETFL, O_NONBLOCK) < 0)
+  {
+    throw std::runtime_error("Setting pipe non-blocking failed");
```
peterschrammel: 💡

src/util/piped_process.h
```diff
@@ -0,0 +1,94 @@
...
+  pid_t pid;
+  FILE *command_stream;
+  int pipe_input[2];
```
peterschrammel, line 92: Document from which perspective this is "input" resp. "output".

src/util/piped_process.cpp
```diff
@@ -0,0 +1,259 @@
...
+    // if execvp fails then we should free it here (just in case the runtime
+    // error below continues execution.)
+    while(i > 0)
+    {
+      i--;
+      free(args[i]);
+    }
+    free(args);
+    // Only reachable if execvp failed
+    std::cerr << "Launching " << commandvec[0]
```
peterschrammel, line 97: 🚫 Use a message handler

src/util/piped_process.cpp
```diff
@@ -0,0 +1,259 @@
...
+  }
+#  endif
+}
+
+piped_processt::~piped_processt()
+{
+#  ifdef _WIN32
+  UNIMPLEMENTED_FEATURE("Pipe IPC on windows: piped_processt constructor")
+#  else
+  // Close the parent side of the remaning pipes
```
peterschrammel:
```suggestion
  // Close the parent side of the remaining pipes
```

src/util/piped_process.h
```diff
@@ -0,0 +1,94 @@
...
+#  include <vector>
+
+#  include "optional.h"
+
+#  define PIPED_PROCESS_INFINITE_TIMEOUT                                       \
+    optionalt<std::size_t>                                                     \
+    {                                                                          \
+    }
+
+class piped_processt
```
peterschrammel, line 21: 📖 Give an overview (with example) of how this API should be used.

unit/util/piped_process.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+  REQUIRE(response.substr(0, 64) == expected_error);
+}
+
+TEST_CASE(
+  "Creating a sub process of z3 and read a response from an echo command.",
+  "[core][util][piped_process]")
+{
+  std::vector<std::string> commands;
+  commands.push_back("z3");
```
peterschrammel, line 69: How do we tell CBMC developers who want to run the unit tests that they need Z3?

src/util/piped_process.cpp
TGWDB: `std::cerr` is correct here since this is in the child process and this will have been redirected to the parent. It is the responsibility of the parent to read (or ignore) this ouptut.

unit/util/piped_process.cpp
NlightNFotis: This (and all other tests referencing `z3`) is now hidden behind a *hide tag* `.` - `[.z3]`.

An ordinary invocation of `bin/unit` without any tags will skip all of these tests. I will modify the CI scripts to invoke the unit test binary with the z3 flag passed as well - this we felt is the best compromise between not requiring z3 for the unit tests for an ordinary user, and testing the functionality we need to test in CI as well as locally.

src/util/piped_process.h
NlightNFotis: It's now under `doc/API/util/piped_process.md`.

.github/workflows/pull-request-checks.yaml
```diff
@@ -107,6 +107,7 @@ jobs:
       - name: Run unit tests
         run: |
           make -C unit      test
+          make -C unit      test "[z3]"
```
TGWDB: I'm curious why this is in some of the builds but not others. On the one hand I understand that testing this is largely unrelated to the various flavours of make/ubuntu and so testing once per OS is sufficient. On the other hand, this makes the tests between OS (versions) less consistent and maybe has potential for lack of clarity.

doc/API/util/piped_process.md
```diff
@@ -0,0 +1,25 @@
+# `src/util/piped_process.{cpp, h}`
+
+To utilise the `piped_process` API for interprocess communication with any binary:
+
+* You need to initialise it by calling the construct `piped_processt("binary with args", message_handler)`.
+  * A `message_handlert` should usually be present down from the call stack at 
+    the place you want to call `piped_process` from.
```
TGWDB: Does not match API, `message_handlert` no longer required?

doc/API/util/piped_process.md
```diff
@@ -0,0 +1,25 @@
...
+  * If IPC fails during child process creation, you will get a `system_exceptiont`.
```
TGWDB: during -> before

doc/API/util/piped_process.md
```diff
@@ -0,0 +1,25 @@
...
+  * If the `binary command` does not correspond to a binary in the `$PATH` or is
+    not a path to a binary itself, you'll read a string `Launching <xyz> failed with error: <error>`
+    when you attempt to `receive()` output from it.
```
TGWDB: it -> the child process.

doc/API/util/piped_process.md
```diff
@@ -0,0 +1,25 @@
...
+* To close the underlying child process, you need a call to the destructor - `~piped_processt()`.
```
TGWDB: I think I would rephrase this to be something like:
"The child process is automatically killed with SIGTERM when the destructor for the piped_processt is called."

src/util/piped_process.h
TGWDB: There may be some potential in having an enum with values like `RUNNING`, `STOPPED`, and `ERROR`, but agree that `NOT_CREATED` is probably not useful.

unit/util/piped_process.cpp
TGWDB: Probably. Although without a Windows implementation this does not matter either way.

TGWDB (approved):

chrisr-diffblue (approved):
While it would be nice to have the windows implementation ready as well, this is here for a dedicated purpose that still an ongoing task, so I'm OK with it going in as is.

NlightNFotis:
Hi @tautschnig, would it be possible to review this PR at your earliest convenience?

We are blocked on a code owner's review from you to get this moving forward.

Thanks for your time 🙏🏻

peterschrammel (approved):

PR #6165, opened 2021-06-10
Initial data structures for incremental smt2 solving

thomasspriggs:
Initial data structures for implementation of incremental smt2 solving.

Standard PR checklist: 

### DIFF
--- a/src/solvers/smt2_incremental/internal_ast.txt
+++ b/src/solvers/smt2_incremental/internal_ast.txt
@@ -1,0 +1,255 @@
+@startuml
+' This diagram is written using PlantUML. It can be visualised locally or using
+' the PlantUML website - http://www.plantuml.com/plantuml/uml/.
+' On Linux the visualisation tool can be installed using -
+'   `sudo apt-get install plantuml`
+' Running the `plantuml` binary from the directory where this file exists will
+' give the option to view the file.
+
+title SMT2 Internal Abstract Syntax Tree Data Structures
+legend bottom right
+  Specification references refer to "The SMT-LIB Standard" Version 2.6
+endlegend
+
+class irept {
+	--
+	Protected inheritance to be used to prevent direct data access from outside
+	of the classes inheriting from it.
+	__
+}
+
+class smt_commandt {
+	..
+	Note - Protected constructor, so as to only allow construction as part of a
+	sub-class. Public copy/upcast/assignment to this class still available.
+	__
+}
+irept <|-- smt_commandt
+
+class smt_set_option_commandt {
+	""(set-option :produce-models true)""
+	--
+	Section 4.1.7 page 55
+	__
+	option : smt_optiont
+}
+smt_commandt <|-- smt_set_option_commandt
+smt_set_option_commandt *-- smt_optiont
+
+class smt_set_logic_commandt {
+	(set-logic QF_AUFBV)
+	--
+	__
+	logic : smt_logict
+}
+smt_commandt <|-- smt_set_logic_commandt
+smt_set_logic_commandt *-- smt_logict
+
+class smt_declare_function_commandt {
+	""(declare-fun |main::1::x!0@1#1| () (_ BitVec 32))""
+	--
+	Declares the sort (rank?) of a function, introducing the new identifier
+	""|main::1::x!0@1#1|"" in the above example.
+	__
+	identifier
+	parameter_sorts : std::vector<smt_sortt>
+	return_sort : smt_sortt
+}
+smt_commandt <|-- smt_declare_function_commandt
+smt_declare_function_commandt *-- smt_sortt
+
+class smt_define_function_commandt {
+	""(define-fun |B0| () Bool (= |main::1::x!0@1#1| |main::1::x!0@1#1|))""
+	--
+	Defines the implementation of a function. Note that the return_sort could be
+	stored in the "definition" term because all terms have a sort for our
+	definition of term.
+	__
+	identifier
+	parameters : std::vector<smt_identifier_termt>
+	return_sort : smt_sortt
+	definition : smt_termt
+}
+smt_commandt <|-- smt_define_function_commandt
+smt_define_function_commandt *-- smt_identifier_termt
+smt_define_function_commandt *-- smt_sortt
+smt_define_function_commandt *-- smt_termt
+
+class smt_assert_commandt {
+	""(assert |B1|)""
+	--
+	Note condition should be of sort ""smt_bool_sortt""
+	__
+	condition : smt_termt
+}
+smt_commandt <|-- smt_assert_commandt
+smt_assert_commandt *-- smt_termt
+
+class smt_check_sat_commandt {
+	""(check-sat)""
+	..
+	Section 4.2.5 page 64
+	__
+}
+smt_commandt <|-- smt_check_sat_commandt
+
+class smt_get_value_commandt {
+	""(get-value (|foo|))""
+	..
+	Section 4.2.6 page 65
+	Note: Identifier would usually be a smt_identifier_termt for our use case, but
+	the syntax allows more flexability than that.
+	__
+	identifier : smt_termt
+}
+smt_commandt <|-- smt_get_value_commandt
+smt_get_value_commandt *-- smt_termt
+
+class smt_exit_commandt {
+	""(exit)""
+	..
+	Section 4.2.1 page 59
+	Instructs the solver process to exit.
+	__
+}
+smt_commandt <|-- smt_exit_commandt
+
+class smt_pop_commandt {
+	""(pop 1)""
+	..
+	__
+	levels : std::size_t
+}
+smt_commandt <|-- smt_pop_commandt
+
+class smt_push_commandt {
+	""(push 1)""
+	..
+	__
+	levels : std::size_t
+}
+smt_commandt <|-- smt_push_commandt
+
+class smt_termt {
+	..
+	Section 3.6 page 26
+	Analogous to `exprt`.
+	Note - Protected constructor, so as to only allow construction as part of a
+	sub-class. Public copy/upcast/assignment to this class still available.
+	__
+	sort : smt_sortt
+}
+irept <|-- smt_termt
+smt_termt *-- smt_sortt
+
+class smt_optiont {
+	..
+	Note - Protected constructor, so as to only allow construction as part of a
+	sub-class. Public copy/upcast/assignment to this class still available.
+	__
+}
+irept <|-- smt_optiont
+
+class smt_option_produce_modelst {
+	:produce-models true
+	..
+	Section 4.1.7 page 56
+	__
+	value : bool
+}
+smt_optiont <|-- smt_option_produce_modelst
+
+class smt_logict {
+	..
+	Note - Protected constructor, so as to only allow construction as part of a
+	sub-class. Public copy/upcast/assignment to this class still available.
+	__
+}
+irept <|-- smt_logict
+
+class smt_logic_qf_aufbvt {
+	""QF_AUFBV""
+	..
+	https://smtlib.cs.uiowa.edu/logics-all.shtml
+	__
+}
+smt_logict <|-- smt_logic_qf_aufbvt
+
+class smt_bool_literal_termt {
+	""true""
+	""false""
+	..
+	Sort is set to smt_bool_sortt in constructor. Idea - factory member functions
+	for true and false.
+	__
+	value : bool
+}
+smt_termt <-- smt_bool_literal_termt
+
+class smt_bit_vector_constant_termt {
+	""(_ bv0 32)""
+	..
+	http://smtlib.cs.uiowa.edu/logics-all.shtml#QF_BV - See bitvector constants
+	__
+	value : mp_integert
+	width : std::size_t
+}
+smt_termt <|-- smt_bit_vector_constant_termt
+
+class smt_identifier_termt {
+	""|main::1::x!0@1#1|""
+	..
+	__
+	value : irep_idt
+}
+smt_termt <|-- smt_identifier_termt
+
+class smt_not_termt {
+	(not false)
+	..
+	__
+	operand : smt_termt
+}
+smt_termt <|-- smt_not_termt
+smt_not_termt *-- smt_termt
+
+class smt_function_application_termt {
+	""(= x y)""
+	..
+	Section 3.6 page 27.
+	Sort checking should be carried out somewhere, to confirm that the sorts of
+	the term arguments match the sorts of the function definition.
+	__
+	function_identifier : irep_idt
+	arguments : std::vector<smt_termt>
+}
+smt_termt <|-- smt_function_application_termt
+
+class smt_sortt {
+	..
+	Section 3.5 page 26
+	Analogous to `typet`.
+	Note - Protected constructor, so as to only allow construction as part of a
+	sub-class. Public copy/upcast/assignment to this class still available.
+	__
+}
+irept <|-- smt_sortt
+
+class smt_bool_sortt {
+	""Bool""
+	..
+	Section 3.5 page 26
+	__
+}
+smt_sortt <|-- smt_bool_sortt
+
+class smt_bit_vector_sortt {
+	""(_ BitVec 32)""
+	..
+	http://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml
+	__
+	bit_width : std::size_t
+}
+smt_sortt <|-- smt_bit_vector_sortt
+
+@enduml
--- a/src/solvers/smt2_incremental/smt_commands.def
+++ b/src/solvers/smt2_incremental/smt_commands.def
@@ -4,6 +4,8 @@
 /// set of commands which are implemented and it is used to automate repetitive
 /// parts of the implementation. These include -
 ///  * The constant `irep_idt`s used to identify each of the command classes.
+///  * The member functions of the `smt_command_const_downcast_visitort` class.
+///  * The type of command checks required to implement `smt_commandt::accept`.
 COMMAND_ID(assert)
 COMMAND_ID(check_sat)
 COMMAND_ID(declare_function)
--- a/src/solvers/smt2_incremental/smt_commands.def
+++ b/src/solvers/smt2_incremental/smt_commands.def
@@ -1,0 +1,16 @@
+/// \file
+/// This set of definitions is used as part of the
+/// [X-macro](https://en.wikipedia.org/wiki/X_Macro) pattern. These define the
+/// set of commands which are implemented and it is used to automate repetitive
+/// parts of the implementation. These include -
+///  * The constant `irep_idt`s used to identify each of the command classes.
+COMMAND_ID(assert)
+COMMAND_ID(check_sat)
+COMMAND_ID(declare_function)
+COMMAND_ID(define_function)
+COMMAND_ID(exit)
+COMMAND_ID(get_value)
+COMMAND_ID(pop)
+COMMAND_ID(push)
+COMMAND_ID(set_logic)
+COMMAND_ID(set_option)
--- a/src/solvers/smt2_incremental/smt_commands.h
+++ b/src/solvers/smt2_incremental/smt_commands.h
@@ -1,0 +1,121 @@
+// Author: Diffblue Ltd.
+
+#ifndef CPROVER_SOLVERS_SMT2_INCREMENTAL_SMT_COMMANDS_H
+#define CPROVER_SOLVERS_SMT2_INCREMENTAL_SMT_COMMANDS_H
+
+#include <solvers/smt2_incremental/smt_logics.h>
+#include <solvers/smt2_incremental/smt_options.h>
+#include <solvers/smt2_incremental/smt_terms.h>
+#include <util/irep.h>
+
+class smt_commandt : protected irept
+{
+public:
+  // smt_commandt does not support the notion of an empty / null state. Use
+  // optionalt<smt_commandt> instead if an empty command is required.
+  smt_commandt() = delete;
+
+  using irept::pretty;
+
+  bool operator==(const smt_commandt &) const;
+  bool operator!=(const smt_commandt &) const;
+
+protected:
+  using irept::irept;
+};
+
+class smt_assert_commandt : public smt_commandt,
+                            private smt_termt::storert<smt_assert_commandt>
+{
+public:
+  explicit smt_assert_commandt(smt_termt condition);
+  const smt_termt &condition() const;
+};
+
+class smt_check_sat_commandt : public smt_commandt
+{
+public:
+  smt_check_sat_commandt();
+};
+
+class smt_declare_function_commandt
+  : public smt_commandt,
+    private smt_sortt::storert<smt_declare_function_commandt>,
+    private smt_termt::storert<smt_declare_function_commandt>
+{
+public:
+  smt_declare_function_commandt(
+    smt_identifier_termt identifier,
+    std::vector<smt_sortt> parameter_sorts);
+  const smt_identifier_termt &identifier() const;
+  const smt_sortt &return_sort() const;
+  std::vector<std::reference_wrapper<const smt_sortt>> parameter_sorts() const;
+
+private:
+  using sort_storert = smt_sortt::storert<smt_declare_function_commandt>;
+  using term_storert = smt_termt::storert<smt_declare_function_commandt>;
+};
+
+class smt_define_function_commandt
+  : public smt_commandt,
+    private smt_termt::storert<smt_define_function_commandt>
+{
+public:
+  smt_define_function_commandt(
+    irep_idt identifier,
+    std::vector<smt_identifier_termt> parameters,
+    smt_termt definition);
+  const irep_idt &identifier() const;
+  std::vector<std::reference_wrapper<const smt_identifier_termt>>
+  parameters() const;
+  const smt_sortt &return_sort() const;
+  const smt_termt &definition() const;
+};
+
+class smt_exit_commandt : public smt_commandt
+{
+public:
+  smt_exit_commandt();
+};
+
+class smt_get_value_commandt : public smt_commandt,
+                               protected smt_termt::storert<smt_assert_commandt>
+{
+public:
+  explicit smt_get_value_commandt(smt_termt descriptor);
+  const smt_termt &descriptor() const;
+};
+
+class smt_push_commandt : public smt_commandt
+{
+public:
+  explicit smt_push_commandt(std::size_t levels);
+  size_t levels() const;
+};
+
+class smt_pop_commandt : public smt_commandt
+{
+public:
+  explicit smt_pop_commandt(std::size_t levels);
+  size_t levels() const;
+};
+
+class smt_set_logic_commandt
+  : public smt_commandt,
+    private smt_logict::storert<smt_set_logic_commandt>
+{
+public:
+  explicit smt_set_logic_commandt(smt_logict logic);
+  const smt_logict &logic() const;
+};
+
+class smt_set_option_commandt
+  : public smt_commandt,
+    private smt_optiont::storert<smt_set_option_commandt>
+{
+public:
+  explicit smt_set_option_commandt(smt_optiont option);
+  const smt_optiont &option() const;
+};
+
+#endif // CPROVER_SOLVERS_SMT2_INCREMENTAL_SMT_COMMANDS_H
--- a/src/solvers/smt2_incremental/smt_commands.h
+++ b/src/solvers/smt2_incremental/smt_commands.h
@@ -8,6 +8,8 @@
 #include <solvers/smt2_incremental/smt_terms.h>
 #include <util/irep.h>
 
+class smt_command_const_downcast_visitort;
+
 class smt_commandt : protected irept
 {
 public:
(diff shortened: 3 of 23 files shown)


src/solvers/smt2_incremental/smt_terms.cpp
```diff
@@ -0,0 +1,166 @@
...
+  static const std::regex valid{R"(^[^\|\\]*$)"};
+  return std::regex_match(id2string(identifier), valid);
+}
+
+smt_identifier_termt::smt_identifier_termt(irep_idt identifier, smt_sortt sort)
+  : smt_termt(ID_smt_identifier_term, std::move(sort))
+{
+  INVARIANT(
+    is_valid_smt_identifier(identifier),
+    R"(Identifiers may not contain | or \ characters.)");
```
TGWDB, line 76: I'm not sure that this exactly matches the SMT2 specification (I'm looking at pages 95-96), is this meant to match the specification or is the cbmc version slightly different here?

unit/solvers/smt2_incremental/smt_terms.cpp
```diff
@@ -0,0 +1,89 @@
...
+  "[core][smt2_incremental]")
+{
+  REQUIRE(smt_bool_literal_termt{true}.value());
+  REQUIRE_FALSE(smt_bool_literal_termt{false}.value());
+}
+
+TEST_CASE("smt_not_termt upcasts.", "[core][smt2_incremental]")
+{
+  smt_termt not_term{smt_not_termt{smt_bool_literal_termt{false}}};
+  not_term = smt_not_termt{smt_bool_literal_termt{false}};
```
TGWDB: false -> true? for one of the cases

unit/solvers/smt2_incremental/smt_terms.cpp
```diff
@@ -0,0 +1,89 @@
...
+  const smt_bool_literal_termt bool_term{true};
+  const smt_not_termt not_term{bool_term};
+  REQUIRE(not_term.operand() == bool_term);
+}
+
+TEST_CASE("smt_identifier_termt upcasts.", "[core][smt2_incremental]")
+{
+  smt_termt term{smt_identifier_termt{"foo", smt_bool_sortt{}}};
+  term = smt_identifier_termt{"bar", smt_bool_sortt{}};
+}
```
TGWDB: Test case for failing the regex?

unit/solvers/smt2_incremental/smt_terms.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+  smt_termt false_term = smt_bool_literal_termt{false};
+  CHECK(false_term == false_term);
+  CHECK(false_term == smt_bool_literal_termt{false});
+  smt_termt true_term = smt_bool_literal_termt{true};
+  CHECK(false_term != true_term);
+  CHECK(
+    smt_bit_vector_constant_termt{42, 8} !=
+    smt_bit_vector_constant_termt{12, 8});
+}
```
TGWDB: Equality test using `not_term`?

TGWDB (commented):
Also no unit tests for the smt_commands?

thomasspriggs:
> Also no unit tests for the smt_commands?

Its on the TODO list in the PR description.

src/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: My intention here is to check that our internal representation is not escaped and then to perform escaping as part of the printing. I was concerned we could accidentally end up with doubly escaped identifiers. We should probably discus outside of github what the precise requirements should be here.

unit/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: Done.

unit/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: New / separate constructor tests added.

unit/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: Done.

TGWDB:
For bookkeeping, note that this is part of the SMT backend work with broad explanation and RFC in #6134

unit/solvers/smt2_incremental/smt_terms.cpp
TGWDB: Seems to only test for `|`?

src/solvers/smt2_incremental/smt_sorts.h
```diff
@@ -0,0 +1,91 @@
...
+#ifndef CPROVER_SOLVERS_SMT2_INCREMENTAL_SMT_SORTS_H
+#define CPROVER_SOLVERS_SMT2_INCREMENTAL_SMT_SORTS_H
+
+#include <util/irep.h>
+
+#include <type_traits>
+
+class smt_sort_const_downcast_visitort;
+
+class smt_sortt : protected irept
```
martin-cs, line 16: Why define these separately when you could use `typet`?  The type system is essentially identical.  I can see a sort of value in having a `class smt_sortt : public typet` and a `smt_sortt to_smt_sort(typet &t)` function so you can use the C++ type safety to enforce conversion but this really seems like overkill.

src/solvers/smt2_incremental/smt_commands.h
```diff
@@ -0,0 +1,92 @@
+// Author: Diffblue Ltd.
+
+#ifndef CPROVER_SOLVERS_SMT2_INCREMENTAL_SMT_COMMANDS_H
+#define CPROVER_SOLVERS_SMT2_INCREMENTAL_SMT_COMMANDS_H
+
+#include <solvers/smt2_incremental/smt_terms.h>
+#include <util/irep.h>
+
+class smt_commandt : protected irept
```
martin-cs, line 13: As a sanity test for this, would it be worth integrating these into `smt2_solver` / `smt2_parser`? They should make the separation between parsing and solving much clearer.

src/solvers/smt2_incremental/smt_sorts.h
```diff
@@ -0,0 +1,91 @@
...
+
+  bool operator==(const smt_sortt &) const;
+  bool operator!=(const smt_sortt &) const;
+
+  void accept(smt_sort_const_downcast_visitort &) const;
+  void accept(smt_sort_const_downcast_visitort &&) const;
+
+  /// \brief Class for adding the ability to up and down cast smt_sortt to and
+  ///   from irept. These casts are required by other irept derived classes in
+  ///   order to store instances of smt_sortt inside them.
```
martin-cs, line 33: Why not just publicly inherit from `irept`?

src/solvers/smt2_incremental/smt_sorts.def
```diff
@@ -0,0 +1,2 @@
+SORT_ID(bool)
```
martin-cs, line 9: I honestly don't see the value of using this over `ID_bool` used in `bool_typet`.

src/solvers/smt2_incremental/smt_terms.h
```diff
@@ -0,0 +1,126 @@
...
+
+#include <functional>
+
+class BigInt;
+using mp_integer = BigInt;
+
+class smt_sortt;
+class smt_term_const_downcast_visitort;
+
+class smt_termt : protected irept, private smt_sortt::storert<smt_termt>
```
martin-cs, line 15: Why not just user `exprt` or if you want to do some kind of conversion, then have a simple wrapper around `exprt` that marks the converted values?

src/solvers/smt2_incremental/smt_terms.h
```diff
@@ -0,0 +1,126 @@
...
+  smt_termt() = delete;
+
+  using irept::pretty;
+
+  bool operator==(const smt_termt &) const;
+  bool operator!=(const smt_termt &) const;
+
+  const smt_sortt &get_sort() const;
+
+  void accept(smt_term_const_downcast_visitort &) const;
```
martin-cs, line 29: It is good to see visitors!

martin-cs (commented):
I feel like I must be missing / not understanding a requirement here.  The commands; yes, they are good and would be useful.  But the rest -- why not just use `exprt` and `typet`?  The underlying representation is the same and there are already a load of utility functions.  Also, in the vast majority of cases, the semantics are the same as well.  I can understand that it might be nice to use the C++ type system to mark which are values used in this subsystem or having had some conversions / checks applied but that can be done in a much more lightweight fashion.

src/solvers/smt2_incremental/smt_terms.h
thomasspriggs: Because this gives us more compile-time checking of the soundness of the conversion and better guarantees about the separation of concerns. This applies both in terms of the conversion from `exprt` to `smt_termt` and the conversion from `smt_termt` to string or other representation understood by the solver. We are only currently planning a conversion to SMT2 strings, but by keeping the conversion to strings separate it would theoretically allow for more easily adding an additional conversion algorithm such as to `smt-switch` data structures to be added on top as an additional run time option. Without the new data structures, the conversion from `exprt` to `smt_termt` could "leak" unconverted `exprt` classes which can only be caught at run time. The way the visitor is defined will also mean that missing conversions from `smt_termt` to string would also cause a compile error. There has to be at least a placeholder method in place with the `UNIMPLEMENTED` macro. 

Defining the data structure will give us a clean separation between the conversion from `exprt` to `smt_termt` and `smt_termt` to string. It will allow for separately unit testing the two conversions. A new data structure only for use with `smt2` has the same advantages that having separate data structures from the language front ends such as for `Java` and `C` vs the internal `goto_program` representation would have.

src/solvers/smt2_incremental/smt_sorts.def
thomasspriggs: Please bear in mind that this file is not only being used to define the IDs of the `smt_sortt` classes. It is also being used to automate writing some of the more repetitive parts of the visitor pattern. :thinking: I should probably add some doxygen to this file explaining how it is used. What would be the value in rationing the number of different IDs used rather than keeping them separate?

src/solvers/smt2_incremental/smt_sorts.h
thomasspriggs: Because I want to restrict the ability to cast to and from `irept`, to the inside of the storage classes. I see the fact that each of these inherit from `irept` to be an implementation detail. Ideally, there should be no need to upcast to `irept` as part of the algorithms which live outside the storage classes. Keeping variables declared as `smt_termt` or `smt_sortt` rather than `irept` where possible will give us more specific compile time type information.

unit/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: I have added a test for `\`.

src/solvers/smt2_incremental/smt_commands.def
```diff
@@ -0,0 +1,8 @@
+COMMAND_ID(assert)
+COMMAND_ID(check_sat)
+COMMAND_ID(declare_function)
+COMMAND_ID(define_function)
+COMMAND_ID(exit)
+COMMAND_ID(get_value)
+COMMAND_ID(set_logic)
+COMMAND_ID(set_option)
```
SaswatPadhi, line 18: Are `(push)` and `(pop)` out of scope for this PR?

thomasspriggs: I had not considered them yet. However I think it will make sense to add them early on. So I will likely add them before I move this PR out of draft status.

src/solvers/smt2_incremental/smt_sorts.h
tautschnig: Wouldn't your rationale then suggest using an `irept` member instead of inheriting?

src/solvers/smt2_incremental/smt_commands.def
martin-cs: Would it be possible to add `check-sat-assuming` as well please?

src/solvers/smt2_incremental/smt_terms.h
martin-cs: I think you can achieve much of the compile time checking at a much lower cost of having something like...

```
class smt_sortt : public typet {};
class smt_termt : public exprtt {};

class smt_translator {
public :
  virtual smt_sortt type_to_sort(const typet &t) = 0;
  virtual smt_termt expr_to_term(const exprt &e) = 0;
}
```
If this is the only way of producing these two types and the rest of the code is `const`-clean and the critical functions take `smt_sortt` and `smt_termt` then it will be very rare to have failed-to-convert bug.

I would suggest an object because these transformations tend to wind up being stateful no matter how good everyone's intent is.  They will likely also need to be `stackable` in some sense.  As soon as you have a couple of options on how to translate each theory / type (IIRC this was on your list of goals?) then you will have a combinatorial explosion of normal forms, so it is not so easy to use types* to enforce normal forms.

Using the visitor to do the actual conversion or caching converted things is smart idea and it is good to see it early integrated into the design here.

*.  I could be wrong, if I am I would love to know because the handling of goto-programs has exactly this issue.

src/solvers/smt2_incremental/smt_commands.def
thomasspriggs: Incremental solving should be possible without `check-sat-assuming` right? I'd prefer not to add too many different things up front, so that we can get a MVP working initially and then add incremental improvements to the incremental solving support. There is no reason that more data structures cannot be added as they are needed.

src/solvers/smt2_incremental/smt_terms.h
thomasspriggs: I am not in favour of publicly inheriting from `typet`, for the following reasons -

- It would support down casting from `typet` to `smt_sortt` and require checks to be written for these casts.
- It would add public methods to `smt_sortt` such as `subtype()` and `source_location()` which don't necessarily make sense in the context of a sort.
- It would add public access to member functions from `irept`, which would allow direct manipulation of the `irept` data structures through weakly typed member functions such as `move_to_sub`.
- It would add public access to `irept`'s `is_nil` member function, whereas I would prefer to make the classes in this hierarchy non-nullable/nillable. That way `optionalt` can be added on top of these classes only in the places where nullability is a desirable property.

On the point of the translator - Yes, we will need functionality to convert from `exprt` to `termt`. The choice of whether to use member functions or free functions with configuration/state struct parameters to do the conversions is mostly a stylistic choice in my opinion, rather than a crucial difference. This is unless we really need virtual inheritance and multiple different translators. I have found free functions to be a little easier to unit test in some circumstances, which is a point in their favour. I do think it would improve readability to store configuration and state separately assuming we end up needing both though. That way the configuration can be kept const and the state mutable.

src/solvers/smt2_incremental/smt_sorts.h
thomasspriggs: > Wouldn't your rationale then suggest using an `irept` member instead of inheriting?

Yes, I could have done that. I could also have implemented this without using `irept` at all and handled lifetime using `std::unique_ptr` or some such. By using `irept` with inheritance this code retains a greater semblance to other parts of the code base. This should aid cbmc developers familiar with other parts of the code base to transition to maintaining or improving this part and it should aid with reusing the existing `pretty`, `operator==` and hashing algorithms defined for `irept`. Otherwise we would need to write more new code to define these pieces of functionality for the new data structures.

src/solvers/smt2_incremental/smt_commands.def
thomasspriggs: I have now added `(push n)` and `(pop n)`.

unit/solvers/smt2_incremental/smt_sorts.cpp
```diff
@@ -0,0 +1,32 @@
...
+  const smt_bool_sortt bool_sort;
+  smt_sortt sort{bool_sort};
+  sort = bool_sort;
+}
+
+TEST_CASE("Test upcasts of smt_bit_vec_sortt.", "[core][smt2_incremental]")
+{
+  const smt_bit_vector_sortt bit_vec_sort{32};
+  smt_sortt sort{bit_vec_sort};
+  sort = bit_vec_sort;
```
NlightNFotis: What are these two test cases testing for exactly?

unit/solvers/smt2_incremental/smt_sorts.cpp
```diff
@@ -30,3 +30,50 @@ TEST_CASE(
 {
   REQUIRE(smt_bit_vector_sortt{32}.bit_width() == 32);
 }
+
+TEST_CASE("Visiting smt_bool_sortt.", "[core][smt2_incremental]")
+{
+  class : public smt_sort_const_downcast_visitort
```
NlightNFotis, line 22: ⛏️ Given that the visitor is nearly the same in both of the following test cases, does it make sense for it to be pulled out somewhere?

unit/solvers/smt2_incremental/smt_to_string.cpp
```diff
@@ -0,0 +1,12 @@
+// Author: Diffblue Ltd.
+
+#include <testing-utils/use_catch.h>
+
+#include <solvers/smt2_incremental/smt_sorts.h>
+#include <solvers/smt2_incremental/smt_to_string.h>
+
+TEST_CASE("Test smt_sortt to string conversion", "[core][smt2_incremental]")
+{
+  CHECK(smt_to_string(smt_bool_sortt{}) == "Bool");
```
NlightNFotis: 👍🏻 I really like how clean these are because of the visitor - I feel like it was a good design choice.

src/solvers/smt2_incremental/smt_sorts.cpp
```diff
@@ -10,6 +10,16 @@
 #include <solvers/smt2_incremental/smt_sorts.def>
 #undef SORT_ID
 
+bool smt_sortt::operator==(const smt_sortt &other) const
+{
+  return irept::operator==(other);
```
NlightNFotis, line 15: Question: since this just calls the base class' method, with no other change to it, does it make sense to kill this method here and just `using irept::operator==;` in the header file?

NlightNFotis (commented):
Some questions from me.

src/solvers/smt2_incremental/smt_sorts.cpp
thomasspriggs: The `irept::operator==` member function takes an `irept` as its parameter / right hand side. So yes we could add `using irept::operator==;`, but the existing tests which use the `smt_sortt::operator==` would not compile because the implicit upcast to `irept` is banned from outside of the data structure's member functions due to the use of protected inheritance. Also I don't think it would make sense to allow comparisons such as `smt_bool_sortt{} == irept{}`. 1) because sorts should not be stored as `irept` outside of the irep based data structures and 2) because if we know that one side is not a sort at compile time and that the other side is a sort, then we know that the result would always be false.

unit/solvers/smt2_incremental/smt_sorts.cpp
thomasspriggs: It tests that the copy constructor of the `smt_sortt` class can be called with a smt_bit_vector_sortt typed parameter and the same for the assignment operator. Effectively testing the upcast which is carried out in these two operations is accessible. I added these two tests because the same test with the `irept` constructor should fail due to the access restrictions. Writing a test that it does not work with `irept`, could be done using some template meta programming to write a "does not compile" function, but probably would not be straight forward code to for us to read or write.

unit/solvers/smt2_incremental/smt_sorts.cpp
thomasspriggs: Yes. I usually err on the side of more duplication in unit tests, in order to make it more obvious why the properties the test is testing should hold. The duplication isn't too bad for the sorts at the moment as there are only a couple of cases. But it will be much worse for the terms and commands and it will be worse if we expand this with more sorts. Therefore I'll come up with a way to factor it out for the terms and commands and then back port the solution to the sorts.

thomasspriggs:
Note: Some of the make build jobs are currently failing CI because the makefile miscounts the tests which are using TEMPLATE_TEST_CASE

src/solvers/smt2_incremental/internal_ast.txt
```diff
@@ -0,0 +1,255 @@
+@startuml
+' This diagram is written using PlantUML. It can be visualised locally or using
+' the PlantUML website - http://www.plantuml.com/plantuml/uml/.
+' On Linux the visualisation tool can be installed using -
+'   `sudo apt-get install plantuml`
+' Running the `plantuml` binary from the directory where this file exists will
+' give the option to view the file.
+
```
TGWDB, line 8: This file is great to have (and also to be able to generate). I just have the general concern that this is not generated or linked from the source code? How easy will it be for the two to lose synchronisation?

src/solvers/smt2_incremental/smt_logics.cpp
```diff
@@ -0,0 +1,47 @@
...
+  ::accept(*this, id(), visitor);
+}
+
+void smt_logict::accept(smt_logic_const_downcast_visitort &&visitor) const
+{
+  ::accept(*this, id(), std::move(visitor));
+}
+
+smt_logic_quantifier_free_bit_vectorst::smt_logic_quantifier_free_bit_vectorst()
+  : smt_logict{ID_smt_logic_quantifier_free_bit_vectors}
```
TGWDB, line 45: There is some tension here since the code here (and later also marked) uses `quantifier_free_bit_vectors` but I think the actual logic is likely to be `QF_AUFBV` (based on what is used now). Note that this is not `QF_BV` which corresponds to quantifier free bit vector theory. (`QF_AUFBV` also allows arrays and free sort and function symbols, more details on logics [here](http://smtlib.cs.uiowa.edu/logics.shtml)).

src/solvers/smt2_incremental/smt_logics.def
```diff
@@ -0,0 +1,9 @@
+/// \file
+/// This set of definitions is used as part of the
+/// [X-macro](https://en.wikipedia.org/wiki/X_Macro) pattern. These define the
+/// set of options which are implemented and it is used to automate repetitive
+/// parts of the implementation. These include -
+///  * The constant `irep_idt`s used to identify each of the logic classes.
+///  * The member functions of the `smt_logic_const_downcast_visitort` class.
+///  * The type of option checks required to implement `smt_logict::accept`.
+LOGIC_ID(quantifier_free_bit_vectors)
```
TGWDB, line 9: logic tension

src/solvers/smt2_incremental/smt_logics.h
```diff
@@ -0,0 +1,78 @@
...
+const smt_logict &smt_logict::storert<derivedt>::downcast(const irept &irep)
+{
+  return static_cast<const smt_logict &>(irep);
+}
+
+class smt_logic_quantifier_free_bit_vectorst : public smt_logict
+{
+public:
+  smt_logic_quantifier_free_bit_vectorst();
+};
```
TGWDB, line 68: logic tension

src/solvers/smt2_incremental/smt_terms.cpp
```diff
@@ -0,0 +1,166 @@
...
+  smt_bit_vector_sortt sort)
+  : smt_termt{ID_smt_bit_vector_constant_term, std::move(sort)}
+{
+  INVARIANT(
+    value < power(mp_integer{2}, get_sort().bit_width()),
+    "value must fit in number of bits available.");
+  INVARIANT(
+    !value.is_negative(),
+    "Negative numbers are not supported by bit vector constants.");
+  set(ID_value, integer2bvrep(value, get_sort().bit_width()));
```
TGWDB, line 96: Some of the current code in `smt2_conv.cpp` handles wrapping and negative values for conversion between bitvectors and integers (see `integer2bvrep` and `bvrep2integer` for examples), are we sure of these invariants here? Alternatively, do we want to make clear that integers must have their conversion to unsigned representations done first before conversion to bitvector?

src/solvers/smt2_incremental/smt_terms.cpp
```diff
@@ -0,0 +1,166 @@
...
+  const mp_integer &value,
+  std::size_t bit_width)
+  : smt_bit_vector_constant_termt{value, smt_bit_vector_sortt{bit_width}}
+{
+}
+
+mp_integer smt_bit_vector_constant_termt::value() const
+{
+  return bvrep2integer(get(ID_value), get_sort().bit_width(), false);
+}
```
TGWDB, line 109: Like mentioned above, the existing code converts based on whether or not the bitvector is signed, we might need this information and conversion?

unit/solvers/smt2_incremental/smt_terms.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+#include <type_traits>
+
+TEST_CASE("smt_bool_literal_termt upcasts.", "[core][smt2_incremental]")
+{
+  smt_termt term{smt_bool_literal_termt{true}};
+  term = smt_bool_literal_termt{false};
+}
+
+TEST_CASE("Test smt_termt.pretty is accessible.", "[core][smt2_incremental]")
```
TGWDB, line 11: Technically this is testing both that the `pretty` function is accessible and the output. I guess a change to the base `pretty` function could invalidate this test without any changes to the `smt_termt` code.

unit/solvers/smt2_incremental/smt_terms.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+TEST_CASE("smt_identifier_termt construction", "[core][smt2_incremental]")
+{
+  cbmc_invariants_should_throwt invariants_throw;
+  CHECK_NOTHROW(smt_identifier_termt{"foo bar", smt_bool_sortt{}});
+  CHECK_THROWS(smt_identifier_termt{"|foo bar|", smt_bool_sortt{}});
+  CHECK_THROWS(smt_identifier_termt{"foo\\ bar", smt_bool_sortt{}});
+}
+
+TEST_CASE("smt_identifier_termt upcasts.", "[core][smt2_incremental]")
```
TGWDB: No actual `CHECK` inside this test case, I guess it's a style choice?

src/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: The `INVARIANT` is correct whilst we haven't implemented/tested the signed -> unsigned conversion. If we decide to implement the conversion in this constructor, then the invariant can be removed. 

Of note is that bitvector terms are not inherently signed or unsigned. Therefore if we were to do the conversion in the constructor then we could end up with potentially surprising behaviour where the term is constructed with a negative value, but the value is positive when we get it back out. Therefore I am tempted to say that the term should only handle positive values and that the conversions should be done as part of the other conversion code pieces.

unit/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: Yeah, I wrote these upcast tests when I was initially working on the accessibility, at that time I just wanted to be sure that the upcasts compiled. However now that the rest of the tests are in place and full of upcasts, I am not sure that this has value. Therefore I might just remove all the upcast tests.

src/solvers/smt2_incremental/smt_logics.cpp
thomasspriggs: Yes. `quantifier_free_bit_vectors` would be just `QF_BV`. `QF_AUFBV` is what we were using in the old back end up until recently. I should fix the full names to match the abbreviations which get written out to the solver.

src/solvers/smt2_incremental/smt_terms.cpp
martin-cs: Yeah, it is one of the differences between the CPROVER type system and the SMT-LIB sorts.  CPROVER has three kinds of bit-vector, signed, unsigned and ... general bit-vector, while SMT-LIB only has signed and unsigned operations.  Working with / interpreting everything as hex / binary and just presenting everything as unsigned would be a good solution.

src/solvers/smt2_incremental/internal_ast.txt
thomasspriggs: I think there is a doxygen plugin which can generate a graph. However I remember these being less than ideal. I suggest we either keep the plantUML and live with the reality that it will likely get out of date, or I remove it entirely and put any useful information from the graph nodes into the doxygen above the corresponding classes.

unit/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: I have now removed these tests, as I was unconvinced of the value they added.

TGWDB (approved):

src/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: Ok. I think this question is sufficiently answered. So I am going to mark it as resolved.

src/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: Yes. We will need to keep track of signedness. But signedness is not a property held by the SMT flavour of bit vectors. Therefore the tracking and conversion should be handled separately.

src/solvers/smt2_incremental/smt_commands.h
thomasspriggs: Yes. Using the new structures for the solving will help sanity test them. I have already found several minor issues by working on solving/converting to string locally. However I feel it is better to get a first version merged and then refine the structures as we work on the solving, rather than perfect the first version before merging the first PR. Integrating the new structures into the parsing would be a nice-to-have. But we are currently thinking of working on exprt -> new structures and new structures -> string before parsing.

chrisr-diffblue (approved):
Looks sensible to me.

PR #6167, opened 2021-06-10
Handling function pointers in contracts instrumentation

feliperodri:
During instrumentation for enforcing/replacing function contracts,
all calls to function pointers were not handled. We now consider
that a function call might be a dereference and properly handle it
during instrumentation.

### DIFF
--- a/regression/contracts/function-calls-05/main.c
+++ b/regression/contracts/function-calls-05/main.c
@@ -1,0 +1,25 @@
+#include <assert.h>
+#include <stdlib.h>
+
+static int add_operator(const int *a, const int *b)
+{
+  return (*a + *b);
+}
+
+// clang-format off
+int foo(int *x, int *y)
+  __CPROVER_requires(x != NULL &&  y != NULL)
+  __CPROVER_ensures(__CPROVER_return_value == 10)
+// clang-format on
+{
+  int (*sum)(const int *, const int *) = add_operator;
+  return sum(x, y);
+}
+
+int main()
+{
+  int x = 5;
+  int y = 5;
+  assert(foo(&x, &y) == 10);
+  return 0;
+}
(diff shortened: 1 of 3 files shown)


SaswatPadhi (approved):
Great!! LGTM :D

Minor nitpick: The `comparator` function doesn't actually compare, so may be consider renaming it to `adder`?

regression/contracts/function-calls-05/main.c
```diff
@@ -0,0 +1,27 @@
+#include <assert.h>
+#include <stdlib.h>
+
+static int add_operator(const int *a, const int *b)
+  __CPROVER_requires(a != NULL && b != NULL)
+    __CPROVER_ensures(__CPROVER_return_value > 0)
```
SaswatPadhi: Also, this wouldn't hold right? For `*a == *b == 0`?

feliperodri:
> Minor nitpick: The `comparator` function doesn't actually compare, so may be consider renaming it to `adder`?

Good catch! 😅 Thanks!

regression/contracts/function-calls-05/main.c
feliperodri: Yes, but it might be part of the specification, right? :)

feliperodri: You're right! I removed the contracts for the this "add_operator". We should implement now the nondet assignments when enforcing contracts.

martin-cs:
If I understand this correctly then you are intending / assuming that code contracts will be run before function pointer removal.  Does it work if it is after function pointer removal as well?  If not, please can you document the dependency somewhere?  There are loads of dependencies between passes and the vast majority are undocumented so it would be nice if the problem didn't get worse.

PR #6172, opened 2021-06-11
Retry with different servers if codesigning fails on `.msi`.

NlightNFotis:
For the Windows package installer build script, we have had
issues in the past where the package build script would fail
completely if the timestamp server was not available.

This now changes it to retry with different timestamp servers
if one of them fails.

### DIFF
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -204,7 +204,16 @@ jobs:
       - name: Sign the installer
         id: code_sign
         run: |
-          & signtool.exe sign /f $env:pfxcert /p "${{ secrets.CODESIGNCERTPASSWORD }}" /tr http://ts.ssl.com ${{ steps.create_packages.outputs.msi_installer }}
+          $servers = @('http://ts.ssl.com', 'http://timestamp.digicert.com', 'http://timestamp.sectigo.com')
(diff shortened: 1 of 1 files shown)


.github/workflows/release-packages.yaml
```diff
@@ -204,7 +204,16 @@ jobs:
...
+          $servers = @('http://ts.ssl.com', 'http://timestamp.digicert.com', 'http://timestamp.sectigo.com')
+          foreach($ts_server in $servers)
+          {
+            & signtool.exe sign /f $env:pfxcert /p "${{ secrets.CODESIGNCERTPASSWORD }}" /tr $ts_server ${{ steps.create_packages.outputs.msi_installer }}
+            if ($LastExitCode -eq "0")
+            {
+              # Stop if code-signing the binary using this server was successful.
+              break
+            }
+          }
```
chrisr-diffblue, line 216: If _none_ of the time servers work, does this still fail the build? Or does the exit code of signtool.exe get swallowed by the foreach loop, making it _look_ like the loop was a success?

NlightNFotis: No, at least in my experiments locally, the fail code *propagates*.

thomasspriggs (approved):

TGWDB (approved):

PR #6173, opened 2021-06-11
Fix doxygen "too many nodes" warning filter

thomasspriggs:
The "too many nodes" warnings were previously filtered out by the python
file. However the regex in the python before this commit does not match
the exact message from the current version of doxygen. This commit
removes the expected warnings from the list file and makes the python
work as it was previously intended, by fixing the regex to match either
the current or previous warning texts.


tautschnig (approved):
Is it worth having a follow-up PR that does away with the `expected_doxygen_warnings.txt` file and any code supporting it?

TGWDB (approved):

NlightNFotis (approved):
Yes please, thank you 🎉

thomasspriggs:
> Is it worth having a follow-up PR that does away with the `expected_doxygen_warnings.txt` file and any code supporting it?

The `expected_doxygen_warnings.txt` file and code supporting it could be useful next time we want to move to a newer doxygen version. However on balance it makes sense to remove it, as less code means less maintenance.

PR #6176, opened 2021-06-13
Support multiple requires and ensures clauses

kroening:
The C grammar supports multiple ensures and requires clauses.  This commit
adds support for this in the typechecker and in goto-instrument.  The
semantics of multiple clauses is the same as that of the conjunction of the
clauses.

### DIFF
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -402,13 +402,13 @@ bool code_contractst::apply_function_contract(
   const auto &type = to_code_with_contract_type(function_symbol.type);
 
   // Isolate each component of the contract.
-  exprt assigns = type.assigns();
-  exprt requires = type.requires();
-  exprt ensures = type.ensures();
+  auto assigns = type.assigns();
+  auto requires = conjunction(type.requires());
+  auto ensures = conjunction(type.ensures());
 
-  // Check to see if the function  contract actually constrains its effect on
+  // Check to see if the function contract actually constrains its effect on
   // the program state; if not, return.
-  if(ensures.is_nil() && assigns.is_nil())
+  if(ensures.is_true() && assigns.is_nil())
     return false;
 
   // Create a replace_symbolt object, for replacing expressions in the callee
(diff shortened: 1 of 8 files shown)


feliperodri (approved):
LGTM. Thank you for implementing this!! It'll really useful when writing contracts.

src/goto-instrument/code_contracts.cpp
```diff
@@ -984,14 +984,14 @@ void code_contractst::add_contract_check(
...
   const symbolt &function_symbol = ns.lookup(mangled_fun);
-  auto code_type = to_code_with_contract_type(function_symbol.type);
+  const auto &code_type = to_code_with_contract_type(function_symbol.type);
 
-  exprt &assigns = code_type.assigns();
-  exprt &requires = code_type.requires();
-  exprt &ensures = code_type.ensures();
+  exprt assigns = code_type.assigns();
+  exprt requires = conjunction(code_type.requires());
+  exprt ensures = conjunction(code_type.ensures());
```
SaswatPadhi, line 991: I wonder if it would be better to create separate assertions instead of just creating a single conjunctive assertion.

I'm thinking about failure reporting. Would we be able to report the right source location if say the 3rd ensures clause failed? Since we would generate a single assertion, we would only generate one source location for the failure, and I think that would be the location of the first ensures clause?

SaswatPadhi (commented):
Thanks for implementing this. The changes look good, I only have one high-level question:

src/goto-instrument/code_contracts.cpp
kroening: Yes, please do PR!

SaswatPadhi: Okay let's merge this since it fixes the issue of contracts being silently ignored. I / Felipe would then make a separate PR to improve error reporting.

SaswatPadhi (approved):

chrisr-diffblue (approved):

peterschrammel (approved):
lgtm

PR #6177, opened 2021-06-15
allow an empty assigns clause

kroening:
This adds support for an explicit empty assigns clause, denoting a pure
function.

### DIFF
--- a/regression/contracts/assigns_enforce_15/main.c
+++ b/regression/contracts/assigns_enforce_15/main.c
@@ -20,10 +20,16 @@ int baz() __CPROVER_ensures(__CPROVER_return_value == global)
   return global;
 }
 
+void qux(void) __CPROVER_assigns()
+{
+  global = global + 1;
+}
+
 int main()
 {
   int n;
   n = foo(&n);
   n = baz();
+  qux();
   return 0;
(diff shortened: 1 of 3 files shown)


regression/contracts/assigns_enforce_15/main.c
```diff
@@ -20,10 +20,16 @@ int baz() __CPROVER_ensures(__CPROVER_return_value == global)
   return global;
 }
 
+void qux(void) __CPROVER_assigns()
```
feliperodri, line 23: The semantics of an empty assigns clause should be that the function doesn't change any (global) arguments, right?

kroening: Yes!

martin-cs (approved):

martin-cs:
One thing you could do with this is to allow calls to pure functions like these to appear in function contracts.  SPARK uses these (expression functions IIRC) to remove the need to duplicate common annotations.  Obviously we can use macros but once they need non-trivial control-flow, pure functions are much more preferable.

feliperodri (approved):

chrisr-diffblue (approved):

PR #6180, opened 2021-06-15
Piped communication on Windows

TGWDB:
Piped inter-process communication for Windows.

**Old text below.**

This is a PR to test CI and see how the piped communication progresses on Windows. This started based on #6151 and so will need to either wait for that to be merged and rebased, or supersede that PR.

Note that this is part of the larger SMT plan available at #6134 .

### DIFF
--- a/src/util/piped_process.cpp
+++ b/src/util/piped_process.cpp
@@ -79,6 +79,8 @@
 #  include "run.h"     // for Windows arg quoting
 #  include "unicode.h" // for widen function
 #  include <tchar.h>   // library for _tcscpy function
+#  include <util/make_unique.h>
+#  include <windows.h>
 #else
 #  include <fcntl.h>  // library for fcntl function
 #  include <poll.h>   // library for poll function
@@ -185,7 +187,8 @@ piped_processt::piped_processt(const std::vector<std::string> commandvec)
   }
   // Create the child process
   STARTUPINFOW start_info;
-  ZeroMemory(&proc_info, sizeof(PROCESS_INFORMATION));
+  proc_info = util_make_unique<PROCESS_INFORMATION>();
+  ZeroMemory(proc_info.get(), sizeof(PROCESS_INFORMATION));
   ZeroMemory(&start_info, sizeof(STARTUPINFOW));
   start_info.cb = sizeof(STARTUPINFOW);
   start_info.hStdError = child_std_OUT_Wr;
@@ -211,7 +214,7 @@ piped_processt::piped_processt(const std::vector<std::string> commandvec)
     NULL,                     // use parent's environment
     NULL,                     // use parent's current directory
     &start_info,              // STARTUPINFO pointer
-    &proc_info);              // receives PROCESS_INFORMATION
+    proc_info.get());         // receives PROCESS_INFORMATION
   // Close handles to the stdin and stdout pipes no longer needed by the
   // child process. If they are not explicitly closed, there is no way to
   // recognize that the child process has ended (but maybe we don't care).
@@ -298,7 +301,7 @@ piped_processt::piped_processt(const std::vector<std::string> commandvec)
 piped_processt::~piped_processt()
 {
 #  ifdef _WIN32
-  TerminateProcess(proc_info.hProcess, 0);
+  TerminateProcess(proc_info->hProcess, 0);
   // Disconnecting the pipes also kicks the client off, it should be killed
   // by now, but this will also force the client off.
   // Note that pipes are cleaned up by Windows when all handles to the pipe
@@ -307,8 +310,8 @@ piped_processt::~piped_processt()
   DisconnectNamedPipe(child_std_IN_Wr);
   CloseHandle(child_std_OUT_Rd);
   CloseHandle(child_std_IN_Wr);
-  CloseHandle(proc_info.hProcess);
-  CloseHandle(proc_info.hThread);
+  CloseHandle(proc_info->hProcess);
+  CloseHandle(proc_info->hThread);
 #  else
   // Close the parent side of the remaining pipes
   fclose(command_stream);
@@ -363,6 +366,8 @@ std::string piped_processt::receive()
     success = ReadFile(child_std_OUT_Rd, buff, BUFSIZE, &nbytes, NULL);
 #else
     nbytes = read(pipe_output[0], buff, BUFSIZE);
+    // Added the status back in here to keep parity with old implementation
+    // TODO: check which statuses are really used/needed.
     if(nbytes == 0) // Update if the pipe is stopped
       process_state = statet::STOPPED;
     success = nbytes > 0;
--- a/src/util/piped_process.cpp
+++ b/src/util/piped_process.cpp
@@ -102,7 +102,13 @@
 #  define BUFSIZE 2048
 
 #ifdef _WIN32
-std::wstring process_windows_args(const std::vector<std::string> commandvec)
+/// This function prepares a single wide string for the windows command
+/// line.
+/// \param commandvec: A vector of strings that contain the command and
+///                    arguments to the command.
+/// \returns A single wide string of the command appropriate for windows
+static std::wstring
+prepare_windows_command_line(const std::vector<std::string> &commandvec)
 {
   std::wstring result = widen(commandvec[0]);
   for(int i = 1; i < commandvec.size(); i++)
@@ -114,7 +120,7 @@ std::wstring process_windows_args(const std::vector<std::string> commandvec)
 }
 #endif
 
-piped_processt::piped_processt(const std::vector<std::string> commandvec)
+piped_processt::piped_processt(const std::vector<std::string> &commandvec)
 {
 #  ifdef _WIN32
   // Security attributes for pipe creation
@@ -204,7 +210,7 @@ piped_processt::piped_processt(const std::vector<std::string> commandvec)
   start_info.hStdOutput = child_std_OUT_Wr;
   start_info.hStdInput = child_std_IN_Rd;
   start_info.dwFlags |= STARTF_USESTDHANDLES;
-  const std::wstring cmdline = process_windows_args(commandvec);
+  const std::wstring cmdline = prepare_windows_command_line(commandvec);
   // Note that we do NOT free this since it becomes part of the child
   // and causes heap corruption in Windows if we free!
   const BOOL success = CreateProcessW(
--- a/src/util/piped_process.cpp
+++ b/src/util/piped_process.cpp
@@ -144,8 +144,12 @@ piped_processt::piped_processt(const std::vector<std::string> &commandvec)
     PIPE_TYPE_BYTE | PIPE_NOWAIT, // Bytes and non-blocking
     PIPE_UNLIMITED_INSTANCES,     // Probably doesn't matter
     BUFSIZE,
-    BUFSIZE,    // Output and input bufffer sizes
-    0,          // Timeout in ms, 0 = use system default
+    BUFSIZE, // Output and input bufffer sizes
+    0,       // Timeout in ms, 0 = use system default
+    // This is the timeout that WaitNamedPipe functions will wait to try
+    // and connect before aborting if no instance of the pipe is available.
+    // In practice this is not used since we connect immediately and only
+    // use one instance (no waiting for a free instance).
     &sec_attr); // For inheritance by child
   if(child_std_IN_Rd == INVALID_HANDLE_VALUE)
   {
--- a/src/util/piped_process.cpp
+++ b/src/util/piped_process.cpp
@@ -2,18 +2,89 @@
 /// Subprocess communication with pipes.
 /// \author Diffblue Ltd.
 
+// NOTES ON WINDOWS PIPES IMPLEMENTATION
+//
+// This is a note to explain the choices related to the Windows pipes
+// implementation and to serve as information for future work on the
+// Windows parts of this class.
+//
+// Windows supports two kinds of pipes: anonymous and named.
+//
+// Anonymous pipes can only operate in blocking mode. This is a problem for
+// this class because blocking mode pipes (on Windows) will not allow the
+// other end to read until the process providing the data has terminated.
+// (You might think that this is not necessary, but in practice this is
+// the case.) For example, if we ran
+//    echo The Jabberwocky; ping 127.0.0.1 -n 6 >nul
+// on the command line in Windows we would see the string "The Jabberwocky"
+// immediately, and then the command would end about 6 seconds later after the
+// pings complete. However, a blocking pipe will see nothing until the ping
+// command has finished, even if the echo has completed and (supposedly)
+// written to the pipe.
+//
+// For the above reason, we NEED to be able to use non-blocking pipes. Since
+// anonymous pipes cannot be non-blocking (in theory they have a named pipe
+// underneath, but it's not clear you could hack this to be non-blocking
+// safely), we have to use named pipes.
+//
+// Named pipes can be non-blocking and this is how we create them.
+//
+// Aside on security:
+// Named pipes can be connected to by other processes and here we have NOT
+// gone deep into the security handling. The default used here is to allow
+// access from the same session token/permissions. This SHOULD be sufficient
+// for what we need.
+//
+// Non-blocking pipes allow immediate reading of any data on the pipe which
+// matches the Linux/MacOS pipe behaviour and also allows reading of the
+// string "The Jabberwocky" from the example above before waiting for the ping
+// command to terminate. This reading can be done with any of the usual pipe
+// read/peek functions, so we use those.
+//
+// There is one problem with the approach used here, that there is no Windows
+// function that can wait on a non-blocking pipe. There are a few options that
+// appear like they would work (or claim they work). Details on these and why
+// they don't work are over-viewed here:
+// -  WaitCommEvent claims it can wait for events on a handle (e.g. char
+//    written) which would be perfect. Unfortunately on a non-blocking pipe
+//    this returns immediately. Using this on a blocking pipe fails to detect
+//    that a character is written until the other process terminates in the
+//    example above, making this ineffective for what we want.
+// -  Setting the pipe timeout or changing blocking after creation. This is
+//    theoretically possible, but in practice either has no effect, or can
+//    cause a segmentation fault. This was attempted with the SetCommTimeouts
+//    function and cause segfault.
+// -  Using a wait for event function (e.g. WaitForMultipleObjects, also single
+//    object, event, etc.). These can in theory wait until an event, but have
+//    the problem that with non-blocking pipes, the wait will not happen since
+//    they return immediately. One might think they can work with a blocking
+//    pipe and a timeout (i.e. have a blocking read and a timeout thread and
+//    wait for one of them to happen to see if there is something to read or
+//    whether we could timeout). However, while this can create the right
+//    wait and timeout behaviour, since the underlying pipe is blocking this
+//    means the example above cannot read "The Jabberwocky" until the ping has
+//    finished, again undoing the interactive behaviour desired.
+// Since none of the above work effectivley, the chosen approach is to use a
+// non-blocking peek to see if there is anthing to read, and use a sleep and
+// poll behaviour that might be much busier than we want. At the time of
+// writing this has not been made smart, just a first choice option for how
+// frequently to poll.
+//
+// Conclusion
+// The implementation is written this way to mitigate the problems with what
+// can and cannot be done with Windows pipes. It's not always pretty, but it
+// does work and handles what we want.
+
 #ifdef _WIN32
-// Windows includes go here
+#  include "run.h"     // for Windows arg quoting
+#  include "unicode.h" // for widen function
+#  include <tchar.h>   // library for _tcscpy function
 #else
 #  include <fcntl.h>  // library for fcntl function
 #  include <poll.h>   // library for poll function
 #  include <signal.h> // library for kill function
 #  include <unistd.h> // library for read/write/sleep/etc. functions
 #endif
-
-#ifdef _WIN32
-// Unimplemented on windows for now...
-#else
 
 #  include <cstring> // library for strerror function (on linux)
 #  include <iostream>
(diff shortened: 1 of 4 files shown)


src/util/piped_process.cpp
```diff
@@ -0,0 +1,361 @@
...
+  // Default state
+  process_state = statet::NOT_CREATED;
+#ifdef _WIN32
+  // Security attributes for pipe creation
+  SECURITY_ATTRIBUTES sec_attr;
+  // Ensure pipes are inherited
+  sec_attr.nLength = sizeof(SECURITY_ATTRIBUTES);
+  sec_attr.bInheritHandle = TRUE;
+  sec_attr.lpSecurityDescriptor = NULL;
+  // Use named pipes to allow non-blocking read
```
kroening, line 135: What access rights does this security descriptor grant? The default one for named pipes gives read access to the EVERYONE group.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,361 @@
...
+  child_std_IN_Rd = CreateNamedPipe(
+      TEXT("\\\\.\\pipe\\cbmc\\SMT2\\child\\IN"),
+      PIPE_ACCESS_INBOUND,          // Reading for us
```
kroening: What happens when there are multiple instances of CBMC running simultaneously? Do they all try to talk over the same named pipe?

src/util/piped_process.cpp
```diff
@@ -0,0 +1,361 @@
...
+  start_info.hStdError = child_std_OUT_Wr;
+  start_info.hStdOutput = child_std_OUT_Wr;
+  start_info.hStdInput = child_std_IN_Rd;
+  start_info.dwFlags |= STARTF_USESTDHANDLES;
+  // Unpack the command into a single string for Windows API
+  std::string cmdline = "";
+  for(int i = 0; i < commandvec.size(); i++)
+  {
+    cmdline.append(commandvec[i].c_str());
+    cmdline.append(" ");
```
kroening: This will require quoting. Use `quote_windows_arg` from util/run.cpp

src/util/piped_process.cpp
```diff
@@ -0,0 +1,361 @@
...
+  while(timeout < 0 || waited_time >= timeout)
+  {
+    PeekNamedPipe(child_std_OUT_Rd, &buffer, 1, nbytes, rbytes, rmbytes);
+    if(buffer != 0)
+    {
+      return true;
+    }
+// TODO, make this define and choice better
+#  define WIN_POLL_WAIT 10
+    Sleep(WIN_POLL_WAIT);
```
kroening, line 432: There is no better option than polling? This will heat a lot of CPUs!

src/util/piped_process.cpp
TGWDB: If I read the documentation [here](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/aa379560(v=vs.85)) correctly then by defining the security attributes for `null` for `lpSecurityDescriptor` (`sec_attr` approx line 30 of `src/util/piped_process.cpp`) this sets the security to be the default for the current session access token. In particular this is NOT available to all users.

That said, I'll add comment about this to the code.

src/util/piped_process.cpp
TGWDB: Fixed with use of random number.

src/util/piped_process.cpp
TGWDB: Added comment to code as well.

src/util/piped_process.cpp
TGWDB: I'd prefer to leave this to the calling code to properly ensure the handling of quotations (also this breaks unit tests and makes for messy dependencies for now).

src/util/piped_process.cpp
TGWDB: I haven't found evidence of one that works nicely with our asynchronous reading from named pipes. The options available for waiting appear to work with blocking pipes, but this does not work with our interactive approach (at least at the moment).

If someone who knows Windows better can suggest an alternative I'll look into it.

src/util/piped_process.cpp
kroening: `rand` isn't random, but produces a deterministic sequence unless you change the seed.
Consider using something you know to be unique, say a process ID.

src/util/piped_process.cpp
kroening: This is a no-go, since it has security implications.

src/util/piped_process.cpp
kroening: I'd say this is a blocker for using this.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,383 @@
...
+  {
+    cmdline.append(commandvec[i].c_str());
+    cmdline.append(" ");
+  }
+  // Note that we do NOT free this since it becomes part of the child
+  // and causes heap corruption in Windows if we free!
+  TCHAR *szCmdline =
+    reinterpret_cast<TCHAR *>(malloc(sizeof(TCHAR) * cmdline.size()));
+  _tcscpy(szCmdline, cmdline.c_str());
+  success = CreateProcess(
```
kroening: This should use `CreateProcessW`, and `widen(cmdline)`

src/util/piped_process.cpp
```diff
@@ -0,0 +1,383 @@
...
+{
+  // Close the parent side of the remaining pipes
+  // and kill child process
+#ifdef _WIN32
+  CloseHandle(child_std_OUT_Rd);
+  CloseHandle(child_std_IN_Wr);
+  TerminateProcess(proc_info.hProcess, 0);
+  CloseHandle(proc_info.hProcess);
+  CloseHandle(proc_info.hThread);
+#else
```
kroening: Does this clean up the named pipe?

TGWDB: According to the docs when all handles are closed/released then Windows cleans up the pipe. I've pushed some more aggressive killing and disconnecting of the child process, but it probably makes no difference in practice.

src/util/piped_process.cpp
kroening: `MsgWaitForMultipleObjectsEx` might do the right thing.

TGWDB: `WaitForMultipleObjects` (and similar ones) rely upon the events they wait on being synchronous. This (so far in testing and from what I've read) means we have to wait for the child process to finish before we can read from the pipe. I saw some example code that used this approach with asynchronous pipes, but then used a busy loop with `peek` in the same way I've implemented currently.

I'll keep looking, and play around more now I've sorted out most of the other quirks.

src/util/piped_process.cpp
TGWDB: Added quoting here.

src/util/piped_process.cpp
TGWDB: It was seeded, but process ID is better - done.

src/util/piped_process.cpp
TGWDB: Done, along with quoting.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,388 @@
...
+  for(int i = 1; i < commandvec.size(); i++)
+  {
+    cmdline.append(L" ");
+    cmdline.append(quote_windows_arg(widen(commandvec[i])));
+  }
+  // Note that we do NOT free this since it becomes part of the child
+  // and causes heap corruption in Windows if we free!
+  WCHAR *szCmdline =
+    reinterpret_cast<WCHAR *>(malloc(sizeof(WCHAR) * cmdline.size()));
+  wcscpy(szCmdline, cmdline.c_str());
```
kroening: You need to add one `WCHAR` to the size of the allocated buffer for the terminating zero.
To avoid this problem altogether, use `_wcsdup`.

src/util/piped_process.cpp
```diff
@@ -0,0 +1,388 @@
...
+    // Create a char** for the arguments (all the contents of commandvec
+    // except the first element, i.e. the command itself).
+    char **args =
+      reinterpret_cast<char **>(malloc((commandvec.size()) * sizeof(char *)));
+    // Add all the arguments to the args array of char *.
+    unsigned long i = 0;
+    while(i < commandvec.size())
+    {
+      args[i] = strdup(commandvec[i].c_str());
+      i++;
```
kroening: No need for the `strdup`, and the subsequent `free`.

src/util/piped_process.cpp
TGWDB: Done

src/util/piped_process.cpp
TGWDB: The child process wants to be able to modify the arguments (i.e. non `const`) while the `commandvec`  with `c_str()`can only provide `const` instances.

kroening: This memory is not shared with the child; the child gets a copy. `const` is fine.

TGWDB: Fixed. :)

kroening: Please add the `const` to the type of `args`. There is then no need for the `const_cast`.

TGWDB:
Now rebased on develop after merging of #6151 .

src/util/piped_process.cpp
TGWDB: See comment/discussion at the top of the file for more on why this is done this way.

src/util/piped_process.cpp
```diff
@@ -30,8 +101,127 @@
...
+  // Unpack the command into a single string for Windows API
+  std::wstring cmdline = widen(commandvec[0]);
+  for(int i = 1; i < commandvec.size(); i++)
+  {
+    cmdline.append(L" ");
+    cmdline.append(quote_windows_arg(widen(commandvec[i])));
+  }
+  // Note that we do NOT free this since it becomes part of the child
+  // and causes heap corruption in Windows if we free!
+  success = CreateProcessW(
```
thomasspriggs: :pick: I suggest declaring success where it is used - `const BOOL success = CreateProcessW(`

src/util/piped_process.cpp
```diff
@@ -30,8 +101,127 @@
...
+  BOOL success = FALSE;
+  ZeroMemory(&proc_info, sizeof(PROCESS_INFORMATION));
+  ZeroMemory(&start_info, sizeof(STARTUPINFOW));
+  start_info.cb = sizeof(STARTUPINFOW);
+  start_info.hStdError = child_std_OUT_Wr;
+  start_info.hStdOutput = child_std_OUT_Wr;
+  start_info.hStdInput = child_std_IN_Rd;
+  start_info.dwFlags |= STARTF_USESTDHANDLES;
+  // Unpack the command into a single string for Windows API
+  std::wstring cmdline = widen(commandvec[0]);
```
thomasspriggs: :pick: I suggest splitting the lines of code which build `cmdline` out into a separate function and then declaring the variable holding the string returned as `const`.

src/util/piped_process.cpp
```diff
@@ -128,15 +326,17 @@ piped_processt::~piped_processt()
...
 piped_processt::send_responset piped_processt::send(const std::string &message)
 {
-#  ifdef _WIN32
-  UNIMPLEMENTED_FEATURE("Pipe IPC on windows: send()")
-#  else
-
   if(process_state != statet::CREATED)
   {
-    return send_responset::ERROR;
+    return send_responset::ERRORED;
```
thomasspriggs, line 342: :pick: I would have kept this rename in a separate commit, in order to keep the commits smaller and restricted to a single change. I consider this something to consider for next time, rather than something worth editing the commit history to fix.

src/util/piped_process.cpp
```diff
@@ -107,15 +295,25 @@ piped_processt::piped_processt(const std::vector<std::string> commandvec)
...
   }
 #  endif
+  process_state = statet::CREATED;
 }
 
 piped_processt::~piped_processt()
 {
 #  ifdef _WIN32
-  UNIMPLEMENTED_FEATURE("Pipe IPC on windows: piped_processt constructor")
+  TerminateProcess(proc_info.hProcess, 0);
```
thomasspriggs: :no_entry_sign: :question: Are we violating the C++ rule of three which says that if we define a custom destructor then we also need to define the copy constructor and copy assignment operator? I wasn't sure if these other operations were implicitly deleted. If we violate this rule then we are at risk of copying this class and then freeing the resources it handles multiple times. The simplest fix for this issue could be to explicitly delete the other 2, to prevent copying. Alternatively the code in the destructor could be moved into custom deleters for the process info and handles using `std::unique_ptr` / `std::shared_ptr` pointers pointing to the managed resources.

src/util/piped_process.cpp
```diff
@@ -30,8 +101,127 @@
...
+  // See following link for details
+  // https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/aa379560(v=vs.85) //NOLINT
+  sec_attr.lpSecurityDescriptor = NULL;
+  // Use named pipes to allow non-blocking read
+  // Build the base name for the pipes
+  std::string base_name = "\\\\.\\pipe\\cbmc\\SMT2\\child\\";
+  // Use process ID as a unique ID for this process at this time.
+  base_name.append(std::to_string(GetCurrentProcessId()));
+  std::string tmp_name = base_name;
+  tmp_name.append("\\IN");
```
thomasspriggs: :bulb: I suggest using `const std::string in_name = base_name + "\\IN";` rather than using `.append`. This allows us to both make the variable const and use a slightly more succinct syntax.

src/util/piped_process.cpp
```diff
@@ -145,48 +345,40 @@ piped_processt::send_responset piped_processt::send(const std::string &message)
...
       "More bytes cannot be read at a time, than the size of the buffer");
-    switch(nbytes)
+    if(nbytes > 0)
     {
-    case -1:
-      // Nothing more to read in the pipe
-      return response;
-    case 0:
-      // Pipe is closed.
-      process_state = statet::STOPPED;
```
thomasspriggs, line 180: :no_entry_sign: :question: Why is the usage of the STOPPED state in the non-Windows version being removed as part of the implementation of the Windows version?

src/util/piped_process.cpp
```diff
@@ -204,12 +396,28 @@ piped_processt::statet piped_processt::get_status()
...
   // unwrap the optional argument here
   const int timeout = wait_time ? narrow<int>(*wait_time) : -1;
+#ifdef _WIN32
+  int waited_time = 0;
+  // The next four need to be initialised for compiler warnings
+  DWORD buffer = 0;
+  LPDWORD nbytes = 0;
+  LPDWORD rbytes = 0;
+  LPDWORD rmbytes = 0;
+  while(timeout < 0 || waited_time >= timeout)
```
thomasspriggs, line 423: :pick: The magic value of `-1` for `timeout` is set due to the unix API being used. I suggest the "unwrap" is moved into the non-Windows code and this line be written as `while(!wait_time || waited_time >= *wait_time)`

src/util/piped_process.h
```diff
@@ -6,11 +6,10 @@
 #define CPROVER_UTIL_PIPED_PROCESS_H
 
 #ifdef _WIN32
-// TODO: Windows definitions go here
-#else
+#  include <windows.h>
```
thomasspriggs: :no_entry_sign: Including `windows.h` in this header file exposes the entire windows API to any other file which includes this one when building on Windows. This has the potential to make it more difficult for contributors working on Windows to write platform independent code and could also increase build times. This issue could be worked around by forward declaring `PROCESS_INFORMATION` and `HANDLE` and making the member fields pointers to these types, or by using the "Pointer to implementation" / "pImpl" technique to only declare fields of these types in the `.cpp` file.

unit/util/piped_process.cpp
```diff
@@ -61,6 +66,39 @@ TEST_CASE(
...
+// if run at midnight, but it's sufficient for a basic check for now.
+TEST_CASE(
+  "Creating a sub process and terminate it.",
+  "[core][util][piped_process]")
+{
+  std::vector<std::string> commands;
+#ifdef _WIN32
+  commands.push_back("cmd /c ping 127.0.0.1 -n 6 > nul");
+  SYSTEMTIME st;
+  GetSystemTime(&st);
```
thomasspriggs: :pick: This should probably be done using C++11's `std::chrono::steady_clock` from the `<chrono>` header, so as to avoid unnecessarily making the test operating system specific.

thomasspriggs (changes requested):
Great to see some more progress on this, but there are some :no_entry_sign: issues which we should address before merging this.

src/util/piped_process.cpp
```diff
@@ -2,19 +2,90 @@
...
+//
+// Named pipes can be non-blocking and this is how we create them.
+//
+// Aside on security:
+// Named pipes can be connected to by other processes and here we have NOT
+// gone deep into the security handling. The default used here is to allow
+// access from the same session token/permissions. This SHOULD be sufficient
+// for what we need.
+//
+// Non-blocking pipes allow immediate reading of any data on the pipre which
```
peterschrammel: pipe

src/util/piped_process.cpp
```diff
@@ -2,19 +2,90 @@
...
+// matches the Linux/MasOC pipe behvariour and also allows reading of the
```
peterschrammel: behaviour

src/util/piped_process.cpp
```diff
@@ -2,19 +2,90 @@
...
+//    they return immediately. One might think they can work with a blocking
+//    pipe and a timeout (i.e. have a blocking read and a timeout thread and
+//    wait for one of them to happen to see if there is something to read or
+//    whether we could timeout). However, while this can create the right
+//    wait and timeout behaviour, since the underlying pipe is blocking this
+//    means the example above cannot read "The Jabberwocky" until the ping has
+//    finished, again undoing the interactive behaviour desired.
+// Since none of the above work effectivley, the chosen approach is to use a
+// non-blocking peek to see if there is anthing to read, and use a sleep and
+// poll behaviour that might be puch busier than we want. At the time of
```
peterschrammel: much?

src/util/piped_process.cpp
```diff
@@ -2,19 +2,90 @@
...
+//    the problem that with non-blocking pipes, the wait will not happen since
+//    they return immediately. One might think they can work with a blocking
+//    pipe and a timeout (i.e. have a blocking read and a timeout thread and
+//    wait for one of them to happen to see if there is something to read or
+//    whether we could timeout). However, while this can create the right
+//    wait and timeout behaviour, since the underlying pipe is blocking this
+//    means the example above cannot read "The Jabberwocky" until the ping has
+//    finished, again undoing the interactive behaviour desired.
+// Since none of the above work effectivley, the chosen approach is to use a
+// non-blocking peek to see if there is anthing to read, and use a sleep and
```
peterschrammel, line 68: Would something like this work instead of sleep or is this no better but just with more overhead?
```
do {
  // check whether there's something new
} while (WaitForSingleObject(hEvent, 100) == WAIT_TIMEOUT)
```

src/util/piped_process.cpp
```diff
@@ -30,8 +101,127 @@
 
 piped_processt::piped_processt(const std::vector<std::string> commandvec)
 {
+  // Default state
+  process_state = statet::NOT_CREATED;
 #  ifdef _WIN32
-  UNIMPLEMENTED_FEATURE("Pipe IPC on windows.")
+  // Security attributes for pipe creation
+  SECURITY_ATTRIBUTES sec_attr;
+  // Ensure pipes are inherited
```
peterschrammel: comment should be one line further down

src/util/piped_process.cpp
```diff
@@ -30,8 +101,127 @@
...
+  tmp_name.append("\\IN");
+  LPCSTR tmp_str = tmp_name.c_str();
+  child_std_IN_Rd = CreateNamedPipe(
+    tmp_str,
+    PIPE_ACCESS_INBOUND,          // Reading for us
+    PIPE_TYPE_BYTE | PIPE_NOWAIT, // Bytes and non-blocking
+    PIPE_UNLIMITED_INSTANCES,     // Probably doesn't matter
+    BUFSIZE,
+    BUFSIZE,    // Output and input bufffer sizes
+    0,          // Timeout in ms, 0 = use system default
```
peterschrammel: This is the timeout to detect that the other side hasn't connected, I guess?

src/util/piped_process.cpp
```diff
@@ -30,8 +101,127 @@
...
+    &sec_attr); // For inheritance by child
+  if(child_std_IN_Rd == INVALID_HANDLE_VALUE)
+  {
+    throw std::runtime_error("Input pipe creation failed for child_std_IN_Rd");
```
peterschrammel: I think we have exception types in `util/exception_util.h` that are more suitable.

src/util/piped_process.cpp
```diff
@@ -30,8 +101,127 @@
 
 piped_processt::piped_processt(const std::vector<std::string> commandvec)
 {
+  // Default state
```
peterschrammel: Would it be possible to split up this mega-function a bit?

src/util/piped_process.cpp
tautschnig: Also: MacOS

src/util/piped_process.cpp
```diff
@@ -30,8 +101,127 @@
 
 piped_processt::piped_processt(const std::vector<std::string> commandvec)
```
tautschnig: I recognise this was here before, but any insight as to why this isn't being passed as a reference? The `const` doesn't seem terribly useful at present.

src/util/piped_process.cpp
```diff
@@ -30,8 +101,127 @@
...
+  // Ensure pipes are inherited
+  sec_attr.nLength = sizeof(SECURITY_ATTRIBUTES);
+  sec_attr.bInheritHandle = TRUE;
+  // This sets the security to the default for the current session access token
+  // See following link for details
+  // https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/aa379560(v=vs.85) //NOLINT
+  sec_attr.lpSecurityDescriptor = NULL;
+  // Use named pipes to allow non-blocking read
+  // Build the base name for the pipes
+  std::string base_name = "\\\\.\\pipe\\cbmc\\SMT2\\child\\";
```
tautschnig: This is generic code in `util`, I don't think it should hard-code `SMT2`.

src/util/piped_process.cpp
TGWDB: Those don't work on non-blocking pipes, they immediately return and so would not actually sleep.

src/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
TGWDB: Old behaviour preserved, added comment about to check if we really need this state (TODO)

src/util/piped_process.h
TGWDB: Fixed.

unit/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
TGWDB: Both fixed

src/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
TGWDB: No insight, but fixed now.

src/util/piped_process.cpp
TGWDB: Split out a small helper function, but the function is pretty straightforward, just many lines due to many arguments to function calls to create Windows processes/pipes (also lots of commends/explanation).

src/util/piped_process.cpp
TGWDB: Removed excess states (in recent separate commit).

src/util/piped_process.cpp
TGWDB: Fixed

src/util/piped_process.cpp
```diff
@@ -101,6 +101,19 @@
 
 #  define BUFSIZE 2048
 
+#ifdef _WIN32
+std::wstring process_windows_args(const std::vector<std::string> commandvec)
```
thomasspriggs: :bulb: I would declare this as `static std::wstring process_windows_args(const std::vector<std::string> commandvec)`. Adding the `static` tells the compiler that this function has internal linkage only. Meaning that it can only be called from within this `.cpp` file. This informs the reader that this function is an internal implementation detail, not part of the external interface.

src/util/piped_process.cpp
```diff
@@ -193,13 +206,7 @@ piped_processt::piped_processt(const std::vector<std::string> commandvec)
   start_info.hStdOutput = child_std_OUT_Wr;
   start_info.hStdInput = child_std_IN_Rd;
   start_info.dwFlags |= STARTF_USESTDHANDLES;
-  // Unpack the command into a single string for Windows API
```
thomasspriggs: :pick: I suggest reusing this comment as doxygen just above the definition of the new function.

src/util/piped_process.cpp
```diff
@@ -193,13 +206,7 @@ piped_processt::piped_processt(const std::vector<std::string> commandvec)
...
-  std::wstring cmdline = widen(commandvec[0]);
-  for(int i = 1; i < commandvec.size(); i++)
-  {
-    cmdline.append(L" ");
-    cmdline.append(quote_windows_arg(widen(commandvec[i])));
-  }
+  const std::wstring cmdline = process_windows_args(commandvec);
```
thomasspriggs: :pick: `process` is a little non-descriptive. Maybe `prepare_windows_command_line` would be better.

src/util/piped_process.cpp
```diff
@@ -102,19 +102,19 @@
 #  define BUFSIZE 2048
 
 #ifdef _WIN32
-std::wstring process_windows_args(const std::vector<std::string> commandvec)
+std::wstring process_windows_args(const std::vector<std::string> *commandvec)
```
thomasspriggs: :no_entry_sign: Adding that `*` passes by pointer, not by reference. To pass by reference you should use `&`. Passing by pointer is an inferior solution as it allows `nullptr` to be passed as an argument, whereas a reference is non-nullable (as is passing by value). This issue should be fixed before this PR is merged, because it affects the way the constructor is used. So if we leave resolving it later then we may need to fix additional usages of the constructor built on top of this PR.

thomasspriggs: Looks to be fixed.

thomasspriggs (approved):
:+1:

src/util/piped_process.cpp
peterschrammel: Any insights on this? Would be good to explain to save follow-ups the time for investigating how the API works.

unit/util/piped_process.cpp
```diff
@@ -35,10 +37,15 @@ TEST_CASE(
   "Creating a sub process with a binary that doesn't exist.",
   "[core][util][piped_process]")
 {
-  const std::string expected_error("Launching abcde failed");
   std::vector<std::string> commands;
+#ifdef _WIN32
+  const std::string expected_error("'abcde' is not recogni");
```
peterschrammel:
```suggestion
  const std::string expected_error("'abcde' is not recognized");
```

unit/util/piped_process.cpp
```diff
@@ -61,14 +68,51 @@ TEST_CASE(
...
+    piped_processt process(commands);
+  }
+  std::chrono::steady_clock::time_point end_time =
+    std::chrono::steady_clock::now();
+  std::chrono::duration<double> time_span =
+    std::chrono::duration_cast<std::chrono::duration<double>>(
+      end_time - start_time);
+  size_t calc = time_span.count();
+#else
+  // Currently not working under Linxu/MacOS?!
```
peterschrammel:
```suggestion
  // Currently not working under Linux/MacOS?!
```

unit/util/piped_process.cpp
```diff
@@ -61,14 +68,51 @@ TEST_CASE(
...
+  // commands.push_back("sleep 6");
```
peterschrammel: Put in `#if 0` rather than commenting.

peterschrammel (approved):

PR #6183, opened 2021-06-16
Parsing and type-checking loop variants (aka ranking functions)

LongPham7:
In this pull request, I have implemented the parsing and type-checking of loop variants (also known as ranking functions). A loop variant is an "integer expression" that (i) is bounded below and (ii) strictly decreases after each iteration of a loop. We use loop variants to prove termination of loops. This is the feature I am currently working on during my research internship at AWS. 


src/util/std_types.h
```diff
@@ -40,6 +40,17 @@ class bool_typet:public typet
   }
 };
 
+/// The int type. 
+/// This is necessary for the type checking of loop variants, which should have
+/// some kind of integer type.
+class int_typet:public typet
```
martin-cs: See
https://github.com/diffblue/cbmc/blob/0ae6010a4e3977e591d9f4dd6029f6cfc6dc4508/src/util/mathematical_types.h#L20

src/ansi-c/scanner.l
```diff
@@ -1281,6 +1281,7 @@ __decltype          { if(PARSER.cpp98 &&
 {CPROVER_PREFIX}"finally"      { loc(); return TOK_CPROVER_FINALLY; }
 {CPROVER_PREFIX}"ID"           { loc(); return TOK_CPROVER_ID; }
 {CPROVER_PREFIX}"loop_invariant" { loc(); return TOK_CPROVER_LOOP_INVARIANT; }
+{CPROVER_PREFIX}"loop_variant" { loc(); return TOK_CPROVER_LOOP_VARIANT; }
```
martin-cs: Are you committed to the name?  My concern is that if we have `loop_invariant` and `loop_variant` it is natural to think that they are duals / inverses which is not really the intuition we want.  Might it be better to call them ranking functions or similar?

martin-cs (changes requested):
Looks good apart from a few small things.  Please can we have some tests and some documentation!

src/util/std_types.h
LongPham7: Thank you so much for your suggestion! Ideally, I want to use a "bounded" int type (as opposed to an unbounded int type) because every loop variant/ranking function must be bounded below. Although `ID_signed_int` is used in CBMC's code base, it doesn't seem to be a valid type of GOTO programs. That is, the function `implicit_typecast_followed` in https://github.com/diffblue/cbmc/blob/0ae6010a4e3977e591d9f4dd6029f6cfc6dc4508/src/ansi-c/c_typecast.cpp#L489 doesn't accept `ID_signed_int` as the target of type-casting. Please let me know if you have a suggestion!

src/ansi-c/scanner.l
SaswatPadhi: Thanks, this is a good point.
There are several options (in addition to `loop_variant` and `ranking_function`) here. For example, Dafny calls it `decreases` (and I think ACSL too). Coq calls it `measures`.
I personally think `__CPROVER_decreases` would be the clearest.

What do you think?

cc: @kroening @tautschnig

src/goto-instrument/loop_utils.cpp
```diff
@@ -116,6 +116,7 @@ void get_modifies_lhs(
     const pointer_arithmetict ptr(to_dereference_expr(lhs).pointer());
     for(const auto &mod : local_may_alias.get(t, ptr.pointer))
     {
+      if(mod.id() == ID_unknown) continue;
```
SaswatPadhi: Please remove this. Unrelated to this PR.

src/ansi-c/parser.y
```diff
@@ -28,6 +28,8 @@ extern char *yyansi_ctext;
 
 #include <util/mathematical_expr.h>
 
+#include <iostream>
```
SaswatPadhi: Please remove this.

src/ansi-c/parser.y
```diff
@@ -497,6 +500,13 @@ loop_invariant_opt:
         { $$=$3; }
         ;
 
+loop_variant_opt:
+        /* nothing */
+        { init($$); parser_stack($$).make_nil(); }
+        | TOK_CPROVER_LOOP_VARIANT '(' ACSL_binding_expression ')'
+        { $$=$3; std::cout<<"Loop variant has been parsed.\n"; }
```
SaswatPadhi: Please remove debug output.

SaswatPadhi (commented):
Some initial comments.

As Martin suggested, let's add some regression tests. I'll take a closer look again later.

src/util/std_types.h
martin-cs: In CPROVER, Integer generally means mathematically bounded integer and C `int` is a `ID_signedbv` or a `ID_unsignedbv`.  The choice between the two has significant impact in terms of decidability and performance.

Do you need to fix the type of this annotation?  I am /guessing/ what you are going to do is take the expression `dec(x,y,z,...)` and generate two assertions : `dec(x,y,z,...) < dec(__CPROVER_old(x), __CPROVER_old(y), ... )` and `0 <= dec(x,y,z,...)`?  If so then you can do this with any ordered type that has a 0.

src/goto-instrument/loop_utils.cpp
LongPham7: Yes, this's been removed in the latest version.

src/ansi-c/scanner.l
LongPham7: For the syntax of loop variants, I've swithced from `__CPROVER_loop_variant(...)` to `__CPROVER_decreases(...)`. If we change our mind later, we can easily change the syntax.

src/util/std_types.h
LongPham7: Thanks a lot for your swift response! As you remarked, one idea is to bound loop variants from below; i.e. `0 <= dec(x,y,z,...)`. Another idea is to use an integer type that is already bounded below (e.g. `ID_signedbv` and `ID_unsignedbv`). I was initially considering the first idea, and then switched to the second idea. However, maybe the first idea is better.

LongPham7: By the way, I finished implementing a prototype of loop variants. If I use the type `integer_typet` from https://github.com/diffblue/cbmc/blob/0ae6010a4e3977e591d9f4dd6029f6cfc6dc4508/src/util/mathematical_types.h#L20, I can successfully run `goto-cc toy.c -o toy.goto` and `goto-instrument --enforce-all-contracts toy.goto toy.2.goto`. However, `cbmc toy.2.goto` fails - it returns an error message of UNIMPLEMENTED. Instead, if I use `ID_signedbv`, `cbmc toy.2.goto` successfully runs. So I probably should switch from `integer_typet` to a different integer type.

martin-cs: From a decision procedures point of view, bit-vectors (`ID_signedbv`, `ID_unsignedbv`, `ID_bv`) and mathematical integers (`ID_integer`) are very very different things.  For a fixed bit-width, bit-vector decision problems are in NP.  There are even some sub-theories that are in P.  Mathematical integers are in undecidable in the general case and NP in the restricted linear case.  The decision procedures you use for them are very different and converting between them in an expression is generally very expensive because it is hard for the decision procedures to exchange information in a useful way (because they work so very differently).  The native back-end does not support more than trivial reasoning over mathematical integers.

My suggestion would be to avoid fixing which you are using unless you really have to.  What you are doing should work just fine for both.  If you just have it as an expression then the user can give a ranking function in `ID_signedbv`, `ID_integer`, `ID_float`, whatever.  You only have to block the case of `ID_real` or add something like `2*(dec(x'',y'',z'') - dec(x',y',z')) < (dec(x',y',z') - dec(x,y,z))` where each `'` denotes an previous iteration.

peterschrammel: It is feasible to specify/construct ranking functions using bitvector types. 2LS does that, for instance. However, you need to be careful to use sufficiently wide types so as to avoid any overflows in the arithmetic occurring in the expressions involved.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -805,16 +807,28 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
   break_is_allowed=old_break_is_allowed;
   continue_is_allowed=old_continue_is_allowed;
 
-  typecheck_spec_expr(code, ID_C_spec_loop_invariant);
+  typecheck_loop_invariant(code);
+  typecheck_spec_decreases(code);
 }
 
-void c_typecheck_baset::typecheck_spec_expr(codet &code, const irep_idt &spec)
+void c_typecheck_baset::typecheck_loop_invariant(codet &code)
```
SaswatPadhi: [nit]

For consistency, can we name this `typecheck_spec_loop_invariant`?

So we would have `ID_C_spec_loop_invariant` checked by `typecheck_spec_loop_invariant` and `ID_C_spec_decreases` checked by `typecheck_spec_decreases`.

src/ansi-c/scanner.l
```diff
@@ -1281,6 +1281,7 @@ __decltype          { if(PARSER.cpp98 &&
 {CPROVER_PREFIX}"finally"      { loc(); return TOK_CPROVER_FINALLY; }
 {CPROVER_PREFIX}"ID"           { loc(); return TOK_CPROVER_ID; }
 {CPROVER_PREFIX}"loop_invariant" { loc(); return TOK_CPROVER_LOOP_INVARIANT; }
+{CPROVER_PREFIX}"decreases" { loc(); return TOK_CPROVER_DECREASES; }
```
SaswatPadhi: [nit]

```suggestion
{CPROVER_PREFIX}"decreases"    { loc(); return TOK_CPROVER_DECREASES; }
```

src/util/std_types.h
SaswatPadhi: Thanks for the pointers, Martin, Peter. For now we implemented `implicit_typecast_arithmetic` check on the ranking function. This check is quite liberal and avoids fixing the expected type for the ranking function.

In the next PR, when we instrument the code to generate the monotonic decrement assertions from ranking functions, we would carefully explore the various corner cases that you have pointed out.

SaswatPadhi (approved):
Approving with a couple of nitpicks.

src/ansi-c/c_typecheck_code.cpp
LongPham7: Thanks for the suggestion!

SaswatPadhi:
Hi @martin-cs,

Could you please take another look at this PR?

regression/contracts/variant_parsing_statement_fail/main.c
```diff
@@ -0,0 +1,14 @@
+
+int main()
+{
+  int i = 0; 
+  int N = 10;
+  while (i != N)
+    __CPROVER_loop_invariant(0 <= i && i <= N)
+    __CPROVER_decreases(int x = 0)
```
martin-cs, line 9: Is it possible to have a `__CPROVER_decreases` without a loop invariant?

regression/contracts/variant_parsing_tuple_fail/test.desc
```diff
@@ -0,0 +1,14 @@
+CORE
+main.c
+--enforce-all-contracts
+^main.c: In function 'main':$
+^main.c:7:1: error: syntax error before ','$
+^     __CPROVER_decreases\(N - i, 42\)$
+^PARSING ERROR$
```
martin-cs, line 6: Nice to see some negative tests!

martin-cs (approved):

regression/contracts/variant_parsing_statement_fail/main.c
LongPham7: Yes, it is possible. The default loop invariant is `true`. The code for that is already in my local repository, and I will file a second PR soon.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -805,16 +807,28 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
 }
 
-void c_typecheck_baset::typecheck_spec_expr(codet &code, const irep_idt &spec)
+void c_typecheck_baset::typecheck_spec_loop_invariant(codet &code)
 {
-  if(code.find(spec).is_not_nil())
+  if(code.find(ID_C_spec_loop_invariant).is_not_nil())
   {
-    exprt &constraint = static_cast<exprt &>(code.add(spec));
+    exprt &invariant = static_cast<exprt &>(code.add(ID_C_spec_loop_invariant));
```
feliperodri, line 818: Is there anything else we could check at this point? For instance, the number of operands? @SaswatPadhi 
```cpp 
 // this comes with 1 operand, which is a Boolean statement
  if(code.operands().size()!=1)
  {
    error().source_location = code.source_location();
    error() << "decreases expected to have 1 operand" << eom;
    throw 0;
  }
```

feliperodri (approved):
LGTM, great work!

src/ansi-c/c_typecheck_code.cpp
SaswatPadhi: This is a good idea, but that's already taken care of by the parsing rule, so we don't need to add it to the type checker.

For instance, take a look at `regression/contracts/variant_parsing_tuple_fail` regression test added in this PR.

Also note that in future this restriction would be relaxed to allow multi-dimensional ranking functions. I'm not sure if we'll make `__CPROVER_decreases` variadic, or allow some other syntax for tuples.

kroening (approved):

regression/contracts/variant_parsing_statement_fail/test.desc
```diff
@@ -0,0 +1,14 @@
+CORE
+main.c
+--enforce-all-contracts
```
SaswatPadhi, line 3: Add this line after `--enforce-all-contracts`
```
activate-multi-line-match
```

regression/contracts/variant_parsing_statement_fail/test.desc
```diff
@@ -0,0 +1,14 @@
...
+^main.c: In function 'main':$
```
SaswatPadhi: Please remove this line.

regression/contracts/variant_parsing_statement_fail/test.desc
```diff
@@ -0,0 +1,14 @@
...
+^main.c:8:1: error: syntax error before 'int'$
+^     __CPROVER_decreases\(int x = 0\)$
```
SaswatPadhi:
```
^main.c.*: error: syntax error before 'int'\n\s*__CPROVER_decreases\(int x = 0\)$
```

regression/contracts/variant_parsing_statement_fail/test.desc
```diff
@@ -0,0 +1,14 @@
...
+^PARSING ERROR$
+^EXIT=1$
```
SaswatPadhi:
```suggestion
^EXIT=(1|64)$
```

SaswatPadhi (changes requested):
1. Please fix the `test.desc` files (which are failing on Windows)
2. Fix the formatting of `main.c` files.

SaswatPadhi (approved):

PR #6184, opened 2021-06-16
New flag for only processing loop contracts

SaswatPadhi:
Created a new flag, `--apply-loop-contracts`, to be able to separately process loop and function contracts.

The `--enforce-contract foo` flag would only enforce function contracts on `foo` now. The previous behavior of enforcing both function and loop contracts can be achieved by `--apply-loop-contracts --enforce-contract foo`.

### DIFF
--- a/regression/contracts/quantifiers-loop-02/test.desc
+++ b/regression/contracts/quantifiers-loop-02/test.desc
@@ -1,6 +1,6 @@
-CORE
+CORE smt-backend
 main.c
---enforce-all-contracts _ --z3
+--apply-loop-contracts _ --z3
 ^EXIT=0$
 ^SIGNAL=0$
 ^\[main.\d+\] line .* Check loop invariant before entry: SUCCESS
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -1186,6 +1186,12 @@ bool code_contractst::replace_calls(
   return false;
 }
 
+void code_contractst::apply_loop_contracts()
+{
+  for(auto &goto_function : goto_functions.function_map)
+    apply_loop_contract(goto_function.first, goto_function.second);
+}
+
 bool code_contractst::replace_calls()
 {
   std::set<std::string> funs_to_replace;
@@ -1204,8 +1210,6 @@ bool code_contractst::enforce_contracts()
   {
     if(has_contract(goto_function.first))
       funs_to_enforce.insert(id2string(goto_function.first));
-    else
-      apply_loop_contract(goto_function.first, goto_function.second);
   }
   return enforce_contracts(funs_to_enforce);
 }
@@ -1225,7 +1229,6 @@ bool code_contractst::enforce_contracts(
                   << messaget::eom;
       continue;
     }
-    apply_loop_contract(goto_function->first, goto_function->second);
 
     if(!has_contract(fun))
     {
--- a/src/goto-instrument/code_contracts.h
+++ b/src/goto-instrument/code_contracts.h
@@ -90,6 +90,8 @@ public:
   /// \return `true` on failure, `false` otherwise
   bool replace_calls();
 
+  void apply_loop_contracts();
+
   const symbolt &new_tmp_symbol(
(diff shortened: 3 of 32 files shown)


regression/contracts/quantifiers-loop-02/test.desc
```diff
@@ -1,6 +1,6 @@
-CORE
+CORE smt-backend
```
feliperodri, line 1: Can it work with both back-ends or it was spuriously passing?

src/goto-instrument/code_contracts.h
```diff
@@ -216,6 +214,11 @@ class code_contractst
     const irep_idt &mode);
 };
 
+#define FLAG_LOOP_CONTRACTS "use-loop-contracts"
+#define HELP_LOOP_CONTRACTS                                                    \
+  " --use-loop-contracts\n"                                                    \
+  "                              simplify loops using loop contracts\n"
```
feliperodri: Is simplify the correct wording here? How about "apply loop invariants on annotated loops"?

feliperodri (approved):
Thanks!

regression/contracts/quantifiers-loop-02/test.desc
SaswatPadhi: This test works correctly either way, we explicitly provide `--z3` in the next line.

I added the tag because on systems that don't have `z3` installed and that only run SAT-specific tests, this test should be ignored. Also see #6181.

Currently test tags are not respected within `regression/contract`. I have reported this already @ https://github.com/diffblue/cbmc/issues/5986#issuecomment-845529645. When that's fixed, this test would be automatically ignored in SAT-only mode.

src/goto-instrument/code_contracts.h
SaswatPadhi: Sure, either works. I'll update as per your suggestion.

I used "simplify" because we replace a loop with a conditional block, which is "simpler" control-flow structure.

src/goto-instrument/code_contracts.cpp
```diff
@@ -549,22 +549,23 @@ bool code_contractst::apply_function_contract(
   return false;
 }
 
-void code_contractst::apply_loop_contract(
-  const irep_idt &function_name,
-  goto_functionst::goto_functiont &goto_function)
+void code_contractst::apply_loop_contracts()
```
martin-cs: I realise this is a little pedantic but I think the pattern of having a `goto_programt` and a `goto_functionst` version of functions, even if `apply_loop_contracts()` is simply a loop that calls `code_contractst::apply_loop_contract(const irep_idt &function_name,  goto_functionst::goto_functiont &amp;goto_function)`.

src/goto-instrument/code_contracts.h
martin-cs: Again, this is possibly pedantic but I can think of at least two ways of using a loop contract: A. converting to an assertion and checking it holds on all iterations, B. using as an over-approximation of what the loop does by removing the loop, havocing modified variables and then assuming it.  Often these are done together but they really don't have to be.  I would only regard B. as simplifying the loop and then ... summarising or (over-)approximating the loop feel more accurate.  For example, if it is a short-ish loop that modifies arrays, unrolling might actually be "simpler" than using a quantifier.

martin-cs (approved):
Approve although it might be good if you could think about the comments.

src/goto-instrument/code_contracts.h
SaswatPadhi: Good catch! :)
I totally agree. All of the instrumentation code currently only does (B), so that's why I called it a "simplification," but you're right that (A) can be "simpler" for unrolled short-ish loops.

Also going to name this `--apply-loop-contracts` instead, just to be consistent.

src/goto-instrument/code_contracts.cpp
SaswatPadhi: I kept this original function and added `apply_loop_contracts()` that just loops over all goto functions and calls `add_loop_contract` on each one.

feliperodri (approved):

PR #6186, opened 2021-06-17
Fix distance check in overflow checking of left shift

tautschnig:
The code wrongly compared the shift operand when checking against
negative shifts.

### DIFF
--- a/regression/cbmc/Overflow_Leftshift1/main.c
+++ b/regression/cbmc/Overflow_Leftshift1/main.c
@@ -1,3 +1,5 @@
+#include <inttypes.h>
+
 int main()
 {
   unsigned char x;
@@ -22,6 +24,10 @@ int main()
 
   // not an overflow in ANSI-C, but undefined in C99
   s = 1 << (sizeof(int) * 8 - 1);
+
+  // overflow in an expression where operand and distance types are different
+  uint32_t u32;
+  int64_t i64 = ((int64_t)u32) << 32;
 
   return 0;
 }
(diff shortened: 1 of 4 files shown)


kroening (approved):

regression/cbmc/Overflow_Leftshift1/main.c
```diff
@@ -23,5 +25,9 @@ int main()
   // not an overflow in ANSI-C, but undefined in C99
   s = 1 << (sizeof(int) * 8 - 1);
 
+  // overflow in an expression where operand and distance types are different
+  uint32_t u32;
+  int64_t i64 = (int64_t)u32 << 32;
```
martin-cs: This is clever.  For those of us who struggle to remember the relative precedence of cast and shift and the C type promotion rules, would it be possible to be write this as:

```
((int64_t)u32) << 32
```

martin-cs: Ah, hold on, is that the point, this is a C89 / C99 difference?

src/analyses/goto_check.cpp
```diff
@@ -777,8 +777,10 @@ void goto_checkt::integer_overflow_check(
       if(distance_type.id() == ID_unsignedbv)
         neg_dist_shift = false_exprt();
       else
-        neg_dist_shift =
-          binary_relation_exprt(op, ID_lt, from_integer(0, distance_type));
```
martin-cs, line 781: :scream: oops.

martin-cs (approved):
Yes; we need this.

regression/cbmc/Overflow_Leftshift1/main.c
tautschnig: Done - no, this C89/C99 difference was not mean to be important, it's just relevant that the types of the operand to be shifted and the type of the shift distance need to differ.

martin-cs: Thanks!

PR #6187, opened 2021-06-17
Tag another test that fails if you don't have z3 on your path

martin-cs:

### DIFF
--- a/regression/contracts/quantifiers-loop-02/test.desc
+++ b/regression/contracts/quantifiers-loop-02/test.desc
@@ -1,4 +1,4 @@
-CORE
+CORE smt-backend broken-cprover-smt-backend
 main.c
 --enforce-all-contracts _ --z3
 ^EXIT=0$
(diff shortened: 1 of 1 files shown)


martin-cs:
@db-ci-cprover please could one of the CI images not have z3 on the path so that we can catch this kind of issue before it is merged?

SaswatPadhi (commented):
The Makefiles in `regression/contracts` do not respect backend tags (see https://github.com/diffblue/cbmc/issues/5986#issuecomment-845529645), so I left this as TODO.

I don't think just adding the tags would leads to these tests being ignored; we also need to update the Make and CMake files.

martin-cs:
At the moment you can't cleanly run the regression tests without z3.  I don't quite know why the CI doesn't flag this up.  Presumably there is some common set-up which includes adding z3?  I think we should either fix this or add z3 as a dependency.  My preference is fix.

SaswatPadhi:
>  I think we should either fix this or add z3 as a dependency. My preference is fix.

+1. I think we should fix this.
1. The Make and CMake scripts should correctly enforce the tags
2. May be for one platform, say linux?, we could have two separate images: one with Z3 and one without Z3?

martin-cs:
Sounds great.  I can try to do 1. but we will need @db-ci-cprover to do 2.

TGWDB:
Note that #6213 has merged a CI job that specifically does some regression tests with `z3`.

martin-cs:
Fixed by a different route.

PR #6190, opened 2021-06-19
VSD - correct widening merge to handle extreme values

jezhiggins:
During a widening merge, it is possible for the upper or lower bound to be extended to integer limits - `max_exprt(type)` or `min_exprt(type)`. This was not handled gracefully and could have resulted in an (inappropriate) invariant violation constructing new intervals. Extreme values, along with overflow and underflow calculating new bounds, are now handled appropriately.

### DIFF
--- a/src/analyses/variable-sensitivity/widened_range.cpp
+++ b/src/analyses/variable-sensitivity/widened_range.cpp
@@ -14,10 +14,9 @@ static bool has_underflowed(const exprt &value)
   return constant_interval_exprt::greater_than(
     value, from_integer(0, value.type()));
 }
-static bool has_overflowed(const exprt &value)
+static bool has_overflowed(const exprt &value, const exprt &initial_value)
 {
-  return constant_interval_exprt::less_than(
-    value, from_integer(0, value.type()));
+  return constant_interval_exprt::less_than(value, initial_value);
 }
 
 exprt widened_ranget::widen_lower_bound() const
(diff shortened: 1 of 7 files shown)


src/analyses/variable-sensitivity/widened_range.cpp
```diff
@@ -0,0 +1,51 @@
...
+ Author: Jez Higgins
+
+\*******************************************************************/
+
+#include "widened_range.h"
+#include <util/simplify_expr.h>
+
+static bool has_underflowed(const exprt &value)
+{
+  return constant_interval_exprt::greater_than(
```
martin-cs, line 14: Do these work for unsigned types?

martin-cs (approved):

src/analyses/variable-sensitivity/widened_range.cpp
jezhiggins: Annoyingly no.

jezhiggins: underflow yes, overflow no. Now fixed.

martin-cs: Thanks!

martin-cs:
Only failure is from the code coverage and TBH I am suspicious of the numbers it reports.  The patch is listed as having 100% coverage so I am going to merge it.

PR #6192, opened 2021-06-19
Introduce binding_exprt::instantiate

kroening:


src/util/std_expr.cpp
```diff
@@ -134,3 +136,136 @@ void let_exprt::validate(const exprt &expr, const validation_modet vm)
...
+  std::map<symbol_exprt, exprt> value_map;
+
+  for(std::size_t i = 0; i < variables.size(); i++)
+  {
+    // types must match
+    PRECONDITION(variables[i].type() == values[i].type());
+    value_map[variables[i]] = values[i];
+  }
+
+  // build a subsitution map
```
peterschrammel:
```suggestion
  // build a substitution map
```

src/util/std_expr.cpp
```diff
@@ -134,3 +136,136 @@ void let_exprt::validate(const exprt &expr, const validation_modet vm)
       "let bindings must be type consistent");
   }
 }
+
+static optionalt<exprt> substitute_symbols_rec(
+  const std::map<irep_idt, exprt> &substitutions,
+  exprt src)
```
peterschrammel, line 142: Is it necessary to pass by value here (a copy is taken below)?

peterschrammel (approved):

src/util/std_expr.cpp
kroening: The copy happens in most cases -- so this is the pattern "use pass by value when copy is inevitable".

src/util/std_expr.cpp
kroening: Fixed

TGWDB (approved):

Issue #6130, opened 2021-05-19
Spurious malloc behavior for large allocations


feliperodri:
**CBMC version:** `develop`
**Operating system:** macOs Majave
**Exact command line resulting in the issue:** `cbmc main.c --malloc-may-fail --malloc-fail-null --pointer-overflow-check --pointer-check`
**What behaviour did you expect:** Consistent behaviour using malloc for large allocations.
**What happened instead:**

Consider the following example.
```c
     1  #include <stdlib.h>
     2  #include <stdint.h>
     3  #include <assert.h>
     4
     5  int main () {
     6      size_t size;
     7      //size = SIZE_MAX;
     8      uint8_t *ptr = malloc(size);
     9      __CPROVER_assume(ptr != NULL);
    10      uint8_t *ptr_end = ptr + size;
    11      assert(ptr <= ptr_end);
    12  }
```

Running this example with `cbmc main.c --malloc-may-fail --malloc-fail-null --pointer-overflow-check --pointer-check` leads to verification failure.
```console
...
[main.pointer_arithmetic.16] line 11 pointer relation: dead object in ptr_end: SUCCESS
[main.pointer_arithmetic.17] line 11 pointer relation: pointer outside object bounds in ptr_end: FAILURE
[main.pointer_arithmetic.18] line 11 pointer relation: invalid integer address in ptr_end: SUCCESS

** 1 of 22 failed (2 iterations)
VERIFICATION FAILED
```

Running this example with `cbmc main.c --pointer-overflow-check --pointer-check` leads to another kind of verification failure.
```console
...
[main.pointer_arithmetic.4] line 10 pointer arithmetic: dead object in ptr + (signed long int)size: SUCCESS
[main.pointer_arithmetic.5] line 10 pointer arithmetic: pointer outside object bounds in ptr + (signed long int)size: FAILURE
[main.pointer_arithmetic.6] line 10 pointer arithmetic: invalid integer address in ptr + (signed long int)size: SUCCESS
...
[main.pointer_arithmetic.16] line 11 pointer relation: dead object in ptr_end: SUCCESS
[main.pointer_arithmetic.17] line 11 pointer relation: pointer outside object bounds in ptr_end: FAILURE
[main.pointer_arithmetic.18] line 11 pointer relation: invalid integer address in ptr_end: SUCCESS

** 2 of 22 failed (3 iterations)
VERIFICATION FAILED
```

If we include line 7, we get different results too.
[...]

PR #2108, opened 2018-04-24
Enable automatic limit for array-as-uninterpreted-function [depends-on: #1874]

tautschnig:
Previously, the command line permitted setting uninterpreted functions
to "never" or "always", where "never" actually was the default. The
"automatic" mode could not be enabled in any way.

This reverts #6232, but now includes additional changes to the array
theory to handle nested struct members no different from members of a
top-level struct.

Fixes: #2018

PR #6194, opened 2021-06-21
Enable and use the automatic limit for array-as-uninterpreted-function (duplicated, fixes 6130)

TGWDB:
This is a recreation of PR #2108 that should fix Issue #6130 . This has been created as a PR to run CI after rebasing on develop (as of this morning).


TGWDB:
@martin-cs Any chance of approval, this is the same as #2108 that you approved (just rebased and passing newer CI).

src/solvers/flattening/boolbv_width.cpp
```diff
@@ -136,13 +135,7 @@ const boolbv_widtht::entryt &boolbv_widtht::get_entry(const typet &type) const
     else
     {
       mp_integer total = *array_size * sub_width;
-      if(total>(1<<30)) // realistic limit
-        throw analysis_exceptiont("array too large for flattening");
-
-      if(total < 0)
```
martin-cs, line 142: This case does not seem to be handled any more.  Maybe it could become an invariant?

martin-cs (approved):
As I said in #2108 , I am OK with this but there is some kind of bug in equality of unbounded arrays which I have yet to reliably trigger via C or locate. I figure that this might expose it so it would be good to have a few tests of what happens when an array is over `MAX_FLATTENED_ARRAY_SIZE`.

src/solvers/flattening/boolbv_width.cpp
TGWDB: I did some checking and this error handling should be left in since it can trigger an exception if not handled. I've pushed a fix here.

PR #6197, opened 2021-06-24
goto instrumentation for decreases clauses on loops

LongPham7:
This pull request contains the full-fledged implementation of one-dimensional loop variants (also known as ranking functions). A loop variant is an integer expression (i) that is bounded below and (ii) strictly decreases after each iteration of a loop. We use loop variants to prove termination of loops. 

A loop variant in C is converted to the following statements in the GOTO language: 
1. Declarations of two temporary variables
2. Assign a loop variant to the first temporary variable at the "beginning" of a loop body
3. Assign a loop variant to the second temporary variable at the "end" of a loop body
4. Assert that the second temporary variable < first temporary variable
5. Discard the two temporary variables by the "dead" statements in the GOTO language. 

### DIFF
--- a/regression/contracts/invar_check_multiple_loops/test.desc
+++ b/regression/contracts/invar_check_multiple_loops/test.desc
@@ -5,8 +5,10 @@ main.c
 ^SIGNAL=0$
 ^\[main.1\] .* Check loop invariant before entry: SUCCESS$
 ^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
-^\[main.3\] .* Check loop invariant before entry: SUCCESS$
-^\[main.4\] .* Check that loop invariant is preserved: SUCCESS$
+^\[main.3\] .* Check decreases clause on loop iteration: SUCCESS$
+^\[main.4\] .* Check loop invariant before entry: SUCCESS$
+^\[main.5\] .* Check that loop invariant is preserved: SUCCESS$
+^\[main.6\] .* Check decreases clause on loop iteration: SUCCESS$
 ^\[main.assertion.1\] .* assertion x == y \+ 2 \* n: SUCCESS$
 ^VERIFICATION SUCCESSFUL$
 --
--- a/src/goto-instrument/code_contracts.cpp
+++ b/src/goto-instrument/code_contracts.cpp
@@ -112,8 +112,9 @@ exprt get_size(const typet &type, const namespacet &ns, messaget &log)
   return result;
 }
 
-void code_contractst::check_apply_invariants(
+void code_contractst::check_apply_loop_contracts(
   goto_functionst::goto_functiont &goto_function,
+  const irep_idt &function_name,
   const local_may_aliast &local_may_alias,
   const goto_programt::targett loop_head,
   const loopt &loop,
@@ -129,11 +130,25 @@ void code_contractst::check_apply_invariants(
       t->location_number > loop_end->location_number)
       loop_end = t;
 
-  // see whether we have an invariant
+  // see whether we have an invariant and a decreases clause
   exprt invariant = static_cast<const exprt &>(
     loop_end->get_condition().find(ID_C_spec_loop_invariant));
+  exprt decreases_clause = static_cast<const exprt &>(
+    loop_end->get_condition().find(ID_C_spec_decreases));
   if(invariant.is_nil())
-    return;
+  {
+    if(decreases_clause.is_nil())
+      return;
+    else
+    {
+      invariant = true_exprt();
+      log.warning()
+        << "The loop at " << loop_head->source_location.as_string()
+        << " does not have a loop invariant, but has a decreases clause. "
+        << "Hence, a default loop invariant ('true') is being used."
+        << messaget::eom;
+    }
+  }
 
   // change
   //   H: loop;
@@ -142,9 +157,12 @@ void code_contractst::check_apply_invariants(
   //   H: assert(invariant);
   //   havoc;
   //   assume(invariant);
+  //   old_decreases_value = decreases_clause(current_environment);
   //   if(guard) goto E:
   //   loop;
+  //   new_decreases_value = decreases_clause(current_environment);
   //   assert(invariant);
+  //   assert(new_decreases_value < old_decreases_value);
   //   assume(false);
   //   E: ...
 
@@ -188,6 +206,32 @@ void code_contractst::check_apply_invariants(
     converter.goto_convert(assumption, havoc_code, mode);
   }
 
+  // Temporary variables that store the decreases clause's values before and
+  // after the loop
+  symbol_exprt old_decreases_symbol =
+    new_tmp_symbol(
+      decreases_clause.type(), loop_head->source_location, function_name, mode)
+      .symbol_expr();
+  symbol_exprt new_decreases_symbol =
+    new_tmp_symbol(
+      decreases_clause.type(), loop_head->source_location, function_name, mode)
+      .symbol_expr();
+
+  if(!decreases_clause.is_nil())
+  {
+    // Generate: a declaration of the temporary variable that stores the
+    // decreases clause's value before the loop
+    havoc_code.add(goto_programt::make_decl(
+      old_decreases_symbol, loop_head->source_location));
+
+    // Generate: an assignment to store the decreases clause's value before the
+    // loop
+    code_assignt old_decreases_assignment{old_decreases_symbol,
+                                          decreases_clause};
+    old_decreases_assignment.add_source_location() = loop_head->source_location;
+    converter.goto_convert(old_decreases_assignment, havoc_code, mode);
+  }
+
   // non-deterministically skip the loop if it is a do-while loop
   if(!loop_head->is_goto())
   {
@@ -209,14 +253,46 @@ void code_contractst::check_apply_invariants(
     converter.goto_convert(assertion, havoc_code, mode);
     havoc_code.instructions.back().source_location.set_comment(
       "Check that loop invariant is preserved");
-    auto offset = havoc_code.instructions.size();
-    goto_function.body.insert_before_swap(loop_end, havoc_code);
-    std::advance(loop_end, offset);
   }
+
+  if(!decreases_clause.is_nil())
+  {
+    // Generate: a declaration of a temporary variable that stores the decreases
+    // clause after one arbitrary iteration of the loop
+    havoc_code.add(goto_programt::make_decl(
+      new_decreases_symbol, loop_head->source_location));
+
+    // Generate: an assignment to store the decreases clause's value after one
+    // iteration of the loop
+    code_assignt new_decreases_assignment{new_decreases_symbol,
+                                          decreases_clause};
+    new_decreases_assignment.add_source_location() = loop_head->source_location;
+    converter.goto_convert(new_decreases_assignment, havoc_code, mode);
+
+    // Generate: assertion that the decreases clause's value after the loop is
+    // smaller than the value before the loop
+    code_assertt monotonic_decreasing_assertion{
+      binary_relation_exprt(new_decreases_symbol, ID_lt, old_decreases_symbol)};
+    monotonic_decreasing_assertion.add_source_location() =
+      loop_head->source_location;
+    converter.goto_convert(monotonic_decreasing_assertion, havoc_code, mode);
+    havoc_code.instructions.back().source_location.set_comment(
+      "Check decreases clause on loop iteration");
+
+    // Discard the temporary variables that store decreases clause's values
+    havoc_code.add(goto_programt::make_dead(
+      old_decreases_symbol, loop_head->source_location));
+    havoc_code.add(goto_programt::make_dead(
+      new_decreases_symbol, loop_head->source_location));
+  }
+
+  auto offset = havoc_code.instructions.size();
+  goto_function.body.insert_before_swap(loop_end, havoc_code);
+  std::advance(loop_end, offset);
 
   // change the back edge into assume(false) or assume(guard)
   loop_end->targets.clear();
-  loop_end->type=ASSUME;
+  loop_end->type = ASSUME;
   if(loop_head->is_goto())
     loop_end->set_condition(false_exprt());
   else
@@ -559,12 +635,15 @@ void code_contractst::apply_loop_contract(
   // Iterate over the (natural) loops in the function,
   // and apply any invariant annotations that we find.
   for(const auto &loop : natural_loops.loop_map)
-    check_apply_invariants(
+  {
+    check_apply_loop_contracts(
       goto_function,
+      function_name,
       local_may_alias,
       loop.first,
       loop.second,
       symbol_table.lookup_ref(function_name).mode);
+  }
 }
 
 const symbolt &code_contractst::new_tmp_symbol(
--- a/src/goto-instrument/code_contracts.h
+++ b/src/goto-instrument/code_contracts.h
@@ -98,12 +98,14 @@ public:
     const irep_idt &function_id,
     const irep_idt &mode);
 
-  void check_apply_invariants(
+  void check_apply_loop_contracts(
     goto_functionst::goto_functiont &goto_function,
+    const irep_idt &function_name,
     const local_may_aliast &local_may_alias,
     const goto_programt::targett loop_head,
     const loopt &loop,
     const irep_idt &mode);
+
   const namespacet &get_namespace() const;
 
   // for "helper" classes to update symbol table.
--- a/src/goto-programs/goto_convert.cpp
+++ b/src/goto-programs/goto_convert.cpp
@@ -857,25 +857,40 @@ void goto_convertt::convert_assume(
   dest.add(goto_programt::make_assumption(op, code.source_location()));
 }
 
-void goto_convertt::convert_loop_invariant(
+void goto_convertt::convert_loop_contracts(
   const codet &code,
   goto_programt::targett loop,
   const irep_idt &mode)
 {
-  exprt invariant=
-    static_cast<const exprt&>(code.find(ID_C_spec_loop_invariant));
+  exprt invariant =
+    static_cast<const exprt &>(code.find(ID_C_spec_loop_invariant));
+  exprt decreases_clause =
+    static_cast<const exprt &>(code.find(ID_C_spec_decreases));
 
-  if(invariant.is_nil())
-    return;
-
-  if(has_subexpr(invariant, ID_side_effect))
+  if(!invariant.is_nil())
   {
-    throw incorrect_goto_program_exceptiont(
-      "Loop invariant is not side-effect free.", code.find_source_location());
+    if(has_subexpr(invariant, ID_side_effect))
+    {
+      throw incorrect_goto_program_exceptiont(
+        "Loop invariant is not side-effect free.", code.find_source_location());
+    }
+
+    PRECONDITION(loop->is_goto());
+    loop->guard.add(ID_C_spec_loop_invariant).swap(invariant);
   }
 
-  PRECONDITION(loop->is_goto());
-  loop->guard.add(ID_C_spec_loop_invariant).swap(invariant);
+  if(!decreases_clause.is_nil())
+  {
+    if(has_subexpr(decreases_clause, ID_side_effect))
+    {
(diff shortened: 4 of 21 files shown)


src/goto-programs/goto_program.h
```diff
@@ -1078,6 +1079,17 @@ class goto_programt
       nil_exprt(),
       {});
   }
+
+  // Returns an assertion smaller < larger. 
+  // This is necessary for the implementation of loop variants. 
+  static instructiont make_inequality(
```
SaswatPadhi: This function is only ever called once, so why don't we just inline its definition there? It's a single line anyway ...

src/goto-programs/goto_convert.cpp
```diff
@@ -857,25 +857,40 @@ void goto_convertt::convert_assume(
   dest.add(goto_programt::make_assumption(op, code.source_location()));
 }
 
-void goto_convertt::convert_loop_invariant(
+void goto_convertt::convert_loop_invariant_and_decreases(
```
SaswatPadhi: I'd just call it `convert_loop_contracts` now. Assigns clause stuff would also be added in this function eventually

src/goto-programs/goto_convert.cpp
```diff
@@ -857,25 +857,40 @@ void goto_convertt::convert_assume(
...
+    PRECONDITION(loop->is_goto());
+    loop->guard.add(ID_C_spec_loop_invariant).swap(invariant);
   }
 
-  PRECONDITION(loop->is_goto());
-  loop->guard.add(ID_C_spec_loop_invariant).swap(invariant);
+  if(!variant.is_nil())
+  {
+    // Interestingly enough, a function call is considered a side effect even if
+    // the callee entails no side effects.
```
SaswatPadhi: I'd remove these two  lines -- this is a known issue. If you want to document this, let's document that before the first call to `has_subexpr` as well. May be add a `TODO` at line 868 and explain this, if you want to.

src/goto-instrument/code_contracts.h
```diff
@@ -98,12 +98,14 @@ class code_contractst
     const irep_idt &function_id,
     const irep_idt &mode);
 
-  void check_apply_invariants(
+  void check_apply_invariants_and_decreases(
```
SaswatPadhi: I'd just call it `check_apply_loop_contracts` now. Assigns clause stuff would also be added in this function eventually

src/goto-instrument/code_contracts.cpp
```diff
@@ -129,11 +130,18 @@ void code_contractst::check_apply_invariants(
...
-  if(invariant.is_nil())
-    return;
+  exprt variant = static_cast<const exprt &>(
+    loop_end->get_condition().find(ID_C_spec_decreases));
+  if(invariant.is_nil()) 
+  {
+    if (variant.is_nil())
+      return;
+    else 
+      invariant = true_exprt(); // The Boolean true expression
```
SaswatPadhi: [nit]

remove comment; self-explanatory

src/goto-instrument/code_contracts.cpp
```diff
@@ -129,11 +130,18 @@ void code_contractst::check_apply_invariants(
...
     loop_end->get_condition().find(ID_C_spec_loop_invariant));
-  if(invariant.is_nil())
-    return;
+  exprt variant = static_cast<const exprt &>(
+    loop_end->get_condition().find(ID_C_spec_decreases));
+  if(invariant.is_nil()) 
+  {
+    if (variant.is_nil())
+      return;
+    else
```
SaswatPadhi: I'd also log a warning here that we are using a default `true` invariant . It's not unsound -- if insufficient this would fail the checks; but just might be surprising to users.

src/goto-programs/goto_program.cpp
```diff
@@ -117,9 +117,17 @@ std::ostream &goto_programt::output_instruction(
...
     break;
+  case ASSIGN:
+    out << from_expr(ns, identifier, instruction.get_code());
+    {
+      const irep_idt &commentw = instruction.source_location.get_comment();
+      if (!commentw.empty())
+        out << " // " << commentw;
+    }
+    out << '\n'; 
+    break;
```
SaswatPadhi: Why was this change needed?

src/goto-instrument/code_contracts.cpp
```diff
@@ -209,11 +241,35 @@ void code_contractst::check_apply_invariants(
...
+    havoc_code.add(goto_programt::make_assignment(new_variant_symbol, variant, declaration_new_variant_source_location));
+    havoc_code.instructions.back().source_location.set_comment("Evaluate and store the loop variant after the loop");
+
+    // Generate: assertion that the loop variant after the loop is smaller than
+    // the loop variant before the loop
+    source_locationt assignment_source_location {loop_head->source_location}; 
+    havoc_code.add(goto_programt::make_inequality(new_variant_symbol, old_variant_symbol, assignment_source_location));
+    havoc_code.instructions.back().source_location.set_comment("Compare the old and new variant values");
+
+    // Discartd the temporary variables that store loop variants
```
SaswatPadhi:
```suggestion
    // Discard the temporary variables that store loop variant values
```

SaswatPadhi (changes requested):
Some initial comments below, and could you please:
- fix the formatting issues? (see the build failures for details)
- change the message `Compare the old and new variant values` to `Check that loop variant monotonically decreases`?

src/goto-programs/goto_program.cpp
LongPham7: When we display a GOTO program by the command `goto-instrument --show-goto-functions`, I want to show a comment `// Evaluate and store the loop variant before the loop` next to the assignment for the first temporary variable, and likewise a comment `// Evaluate and store the loop variant after the loop` next to the second temporary variable's assignment. This is why I modified the function `output_instruction` in `goto_program.cpp` such that it can print out not only comments of assertions but also comments of assignments. If you think this feature is redundant, please let me know!

src/goto-programs/goto_program.h
LongPham7: Good point. Yes, I will fix it.

src/goto-programs/goto_convert.cpp
LongPham7: OK. I will just remove these two lines of comments.

src/goto-instrument/code_contracts.cpp
LongPham7: Yes, I will do it.

src/goto-instrument/code_contracts.cpp
```diff
@@ -129,11 +130,24 @@ void code_contractst::check_apply_invariants(
...
+  {
+    if(variant.is_nil())
+      return;
+    else
+    {
+      invariant = true_exprt();
+      log.warning() << "A loop in function '" << function_name
+                    << "' does not have a loop invariant, but a loop variant. "
+                    << "Hence, a loop invariant is set to true by default. "
+                    << messaget::eom;
```
SaswatPadhi: How about:

```c
      log.warning() << "The loop at " << loop_head->source_location.as_string()
                    << "does not have a loop invariant, but has a loop variant. "
                    << "Hence, a default loop invariant ('true') is being used."
                    << messaget::eom;
```

src/goto-instrument/code_contracts.cpp
```diff
@@ -129,11 +130,24 @@ void code_contractst::check_apply_invariants(
...
+    loop_end->get_condition().find(ID_C_spec_decreases));
   if(invariant.is_nil())
-    return;
+  {
+    if(variant.is_nil())
+      return;
+    else
+    {
+      invariant = true_exprt();
+      log.warning() << "A loop in function '" << function_name
```
SaswatPadhi: Could you please also add a regression test to check if this warning is correctly triggered?

src/goto-programs/goto_program.cpp
SaswatPadhi: We usually don't add comments on assignments, but only on the assertions (which you're doing already in code_contracts.cpp). The comments on assignments are not guaranteed to be preserved across subsequent transformations (simplification passes etc.), so may be we can remove these comments and revert this printing logic.

src/goto-instrument/code_contracts.cpp
```diff
@@ -188,6 +205,37 @@ void code_contractst::check_apply_invariants(
     converter.goto_convert(assumption, havoc_code, mode);
   }
 
+  // Temproary variables that store the variant before and after the loop
```
SaswatPadhi:
```suggestion
  // Temporary variables that store the variant before and after the loop
```

src/goto-instrument/code_contracts.cpp
```diff
@@ -188,6 +205,37 @@ void code_contractst::check_apply_invariants(
...
+    // variant before the loop entry
+    havoc_code.add(
+      goto_programt::make_decl(old_variant_symbol, loop_head->source_location));
+
+    // Generate: an assignment to store the loop variant before the loop entry
+    // We would like to set a comment associated with a source location.
+    // However, we have already used loop_head->source_location to set a
+    // comment. Furthermore, the third argument of make_assignment is passed by
+    // reference rather than by value. Therefore, it is necessary to clone
+    // loop_head->source_location.
```
SaswatPadhi: I don't understand the issue here. What exactly is the problem with passing the `source_location` by value? Is it that we are adding a comment? If so, please remove the comment on the assignment (I explained the rationale in another comment below).

(same comment for the other assignment below)

src/goto-instrument/code_contracts.cpp
```diff
@@ -188,6 +205,37 @@ void code_contractst::check_apply_invariants(
...
+    source_locationt declaration_old_variant_source_location{
+      loop_head->source_location};
+    havoc_code.add(goto_programt::make_assignment(
+      old_variant_symbol, variant, declaration_old_variant_source_location));
```
SaswatPadhi: Instead of generating a goto instruction directly, we should be calling `goto_convert` to properly translate a high-level assignment to appropriate goto instructions. Use:

```c++
    code_assignt old_variant_assignment{old_variant_symbol, variant};
    old_variant_assignment.add_source_location() = loop_head->source_location;
    converter.goto_convert(old_variant_assignment, havoc_code, mode);
```

(same for the other assignment)

src/goto-instrument/code_contracts.cpp
```diff
@@ -209,14 +257,47 @@ void code_contractst::check_apply_invariants(
...
+      new_variant_symbol, variant, declaration_new_variant_source_location));
+    havoc_code.instructions.back().source_location.set_comment(
+      "Evaluate and store the loop variant after the loop");
+
+    // Generate: assertion that the loop variant after the loop is smaller than
+    // the loop variant before the loop
+    source_locationt assignment_source_location{loop_head->source_location};
+    havoc_code.add(goto_programt::make_assertion(
+      binary_relation_exprt(new_variant_symbol, ID_lt, old_variant_symbol),
+      assignment_source_location));
```
SaswatPadhi: Instead of generating a goto instruction directly, we should be calling `goto_convert` to properly translate a high-level assertion to appropriate goto instructions. Use:

```c
    code_assertt variant_assertion{binary_relation_exprt(new_variant_symbol, ID_lt, old_variant_symbol)};
    variant_assertion.add_source_location() = loop_end->source_location;
    converter.goto_convert(variant_assertion, havoc_code, mode);
```

SaswatPadhi (changes requested):

src/goto-programs/goto_program.cpp
LongPham7: OK. I will revert the change I had made.

src/goto-instrument/code_contracts.cpp
LongPham7: Yes, I will.

src/goto-instrument/code_contracts.cpp
LongPham7: Thanks a lot for the code.

src/goto-instrument/code_contracts.cpp
LongPham7: Now that we use `goto_convert` instead of `make_assignment`, we no longer need to clone a `source_location` object. So this issue is now gone.

src/goto-instrument/code_contracts.cpp
```diff
@@ -129,11 +130,25 @@ void code_contractst::check_apply_invariants(
...
-    return;
+  {
+    if(variant.is_nil())
+      return;
+    else
+    {
+      invariant = true_exprt();
+      log.warning()
+        << "The loop at " << loop_head->source_location.as_string()
+        << " does not have a loop invariant, but has a loop variant. "
```
martin-cs: Please can you be consistent with naming?  Personally I don't like "loop variant" but I care most about it being consistent.  SO if you are going to call it a loop variant in the help text and comments then maybe it should be `__CPROVER_loop_variant`.

src/goto-instrument/code_contracts.cpp
```diff
@@ -188,6 +206,29 @@ void code_contractst::check_apply_invariants(
...
+  {
+    // Generate: a declaration of the temporary variable that stores the loop
+    // variant before the loop entry
+    havoc_code.add(
+      goto_programt::make_decl(old_variant_symbol, loop_head->source_location));
+
+    // Generate: an assignment to store the loop variant before the loop entry
+    code_assignt old_variant_assignment{old_variant_symbol, variant};
+    old_variant_assignment.add_source_location() = loop_head->source_location;
+    converter.goto_convert(old_variant_assignment, havoc_code, mode);
```
martin-cs: @tautschnig is this use of `goto_convert` fitting with how and when / where it is supposed to be used?

martin-cs (approved):
Looks good apart from the things flagged.

src/goto-instrument/code_contracts.cpp
LongPham7: Thanks for your suggestion. As of now, the concrete syntax for loop variants is `__CPROVER_decreases(...)`. However, "decreases" is not a noun (or at least, it would be awkward to use it as a noun). This is why I currently use the phrase "loop variants" in the code/comments to refer to the loop variants/ranking functions. I will discuss this with Saswat.

LongPham7: @martin-cs I misunderstood your comment - I thought you were suggesting changing all function names and comments in the code base so that we never mention "loop variants", but now I believe you are just referring to the warning issued to a user. Yes, I will change the warning message to something like "The loop at ... does not have a loop invariant, but has a decreases-clause." In the code base, however, I will keep using the phrase "loop variants" in comments and function names.

martin-cs: So, I am very bad at using consistent terminology and it causes confusion.  My comments, here and earlier, are intended to try to help you avoid a similar fate.

I would /suggest/ that you use one terminology for *everything*; the annotation, the user error reporting, in the comments, in the code, everywhere.  Either call them ranking functions, or loop variants, or ... whatever name you want.  Personally I feel that loop variant can cause confusion with loop invariants but that is just my view.  I think it is far more important to be consistent than it is to pick one particular name.

LongPham7: OK. Thanks for the clarification! I will update my comments and function names in the code.

LongPham7: I've just updated all function names, comments, and warning messages in the code such that the phrase "DECRAESES clause" is used in place of "loop variants."

regression/contracts/invar_check_multiple_loops/test.desc
```diff
@@ -5,8 +5,10 @@ main.c
 ^SIGNAL=0$
 ^\[main.1\] .* Check loop invariant before entry: SUCCESS$
 ^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
-^\[main.3\] .* Check loop invariant before entry: SUCCESS$
-^\[main.4\] .* Check that loop invariant is preserved: SUCCESS$
+^\[main.3\] .* Check that DECREASES clause monotonically decreases: SUCCESS$
```
SaswatPadhi: We don't capitalize the other contracts to this looks a bit odd. How about:

```
Check monotonicity of decreases clause: ...
```

LongPham7: Yeah, initially, I didn't capitalize "decreases clauses". However, in that warning message, we use "decreases" twice, and I thought it might confuse the user. This is why I decided to capitalize one of the two occurrences of "decreases." Your suggested warning message is much better. I will modify the code accordingly.

src/goto-programs/goto_program.cpp
```diff
@@ -131,7 +131,7 @@ std::ostream &goto_programt::output_instruction(
     {
       out << from_expr(ns, identifier, instruction.get_condition());
 
-      const irep_idt &comment=instruction.source_location.get_comment();
+      const irep_idt &comment = instruction.source_location.get_comment();
```
SaswatPadhi: [nit]

Please undo this change.

src/goto-programs/goto_convert_class.h
```diff
@@ -236,7 +236,7 @@ class goto_convertt:public messaget
     goto_programt &dest,
     const irep_idt &mode);
   void convert_cpp_delete(const codet &code, goto_programt &dest);
-  void convert_loop_invariant(
+  void convert_loop_contract(
```
SaswatPadhi:
```suggestion
  void convert_loop_contracts(
```

(here and elsewhere)

src/goto-programs/goto_convert.cpp
```diff
@@ -857,25 +857,40 @@ void goto_convertt::convert_assume(
...
   }
 
-  PRECONDITION(loop->is_goto());
-  loop->guard.add(ID_C_spec_loop_invariant).swap(invariant);
+  if(!decreases_clause.is_nil())
+  {
+    if(has_subexpr(decreases_clause, ID_side_effect))
+    {
+      throw incorrect_goto_program_exceptiont(
+        "decreases clause is not side-effect free.",
```
SaswatPadhi:
```suggestion
        "Decreases clause is not side-effect free.",
```

SaswatPadhi (approved):
Couple of minor comments, but otherwise LGTM! Thanks for all the work!

feliperodri (approved):
Thank you for tackling all comments!

src/goto-instrument/code_contracts.cpp
```diff
@@ -209,14 +253,46 @@ void code_contractst::check_apply_invariants(
...
+
+    // Generate: assertion that the decreases clause's value after the loop is
+    // smaller than the value before the loop
+    code_assertt monotonic_decreasing_assertion{
+      binary_relation_exprt(new_decreases_symbol, ID_lt, old_decreases_symbol)};
+    monotonic_decreasing_assertion.add_source_location() =
+      loop_head->source_location;
+    converter.goto_convert(monotonic_decreasing_assertion, havoc_code, mode);
+    havoc_code.instructions.back().source_location.set_comment(
+      "Check monotonicity of decreases clause");
```
kroening: "monotonicity" is perhaps not the right word. It might not be an intuitive description for a broader audience, and being pedantic, you need strict monotonicity to get termination.
How about "ranking function decreases"?

kroening:
What should happen when there are multiple `__CPROVER_decreases` clauses for one loop?
You could make that a lexicographic ranking function.

kroening (approved):

SaswatPadhi:
> What should happen when there are multiple `__CPROVER_decreases` clauses for one loop?
You could make that a lexicographic ranking function.

Yes, this is a good point. Currently we don't allow multiple invariant and decreases annotations, but my proposal was also to treat multiple decreases clause annotations as tuples that would be compared lexicographically.

src/goto-instrument/code_contracts.cpp
SaswatPadhi: I think "ranking function" might also sound unfamiliar, plus we haven't used ranking function elsewhere (the annotation just reads `__CPROVER_decreases`).

How about simply, `"Check decreases clause on loop iteration"`?

LongPham7: Thanks a lot for both of your suggestions. I will change the comment to `"Check decreases clause on loop iteration"`.

PR #6199, opened 2021-06-25
floating-point mod and remainder

kroening:


martin-cs (commented):
The PR is fine apart from the semantics.  I am not sure we should be implementing wrong semantics, or, at least if we are going to, there are a number of less wrong and cheaper options.

For discussion, IEEE-754 2008 says:

"When y ≠ 0, the remainder r = remainder(x, y) is defined for finite x and y regardless of the
rounding-direction attribute by the mathematical relation r = x − y × n , where n is the integer nearest the exact number x/y ; whenever | n − x/y | = 1⁄2 , then n is even. Thus, the remainder is always exact.
If r = 0, its sign shall be that of x. remainder(x, ∞) is x for finite x."

ISO-9899 1999 says:

"The fmod functions return the value x − ny, for some integer n such that, if y is nonzero,
the result has the same sign as x and magnitude less than the magnitude of y. If y is zero,
whether a domain error occurs or the fmod functions return zero is implementation-
defined."

"The remainder functions compute the remainder x REM y required by IEC 60559." (i.e. IEEE-754)

"The remquo functions compute the same remainder as the remainder functions. In
the object pointed to by quo they store a value whose sign is the sign of x/y and whose
magnitude is congruent modulo 2^n to the magnitude of the integral quotient of x/y, where
n is an implementation-defined integer greater than or equal to 3."


Despite what the comments here https://github.com/diffblue/cbmc/blob/9a044360888d62e36aaa1b5335a9c3b4ff3e55c7/src/solvers/floatbv/float_utils.cpp#L543 I have implemented this:
https://github.com/martin-cs/symfpu/blob/master/core/remainder.h

The expensive bit is here https://github.com/martin-cs/symfpu/blob/c3acaf62b137c36aae5eb380f1d883bfa9095f60/core/remainder.h#L143 as exact iteration costs but you need them for fewer and fewer numbers.  I do wonder whether it is worth implementing as a C library so that we can use loop unrolling to limit the cost.

kroening:
The change of the semantics should be a different PR; this PR establishes a way to get the operation from the language frontend to the solver.

It's tempting to make this a library function; the key problem with that we don't have a way to constant-fold C code during typechecking.

martin-cs (approved):
Approve given @kroening 's comments on the scoping of the PR.

jbmc/src/java_bytecode/java_bytecode_convert_method.cpp
```diff
@@ -1558,8 +1558,10 @@ java_bytecode_convert_methodt::convert_instructions(const methodt &method)
     else if(bytecode == patternt("?rem"))
     {
       PRECONDITION(op.size() == 2 && results.size() == 1);
+      // This is _not_ the remainder operation defined by IEEE 754,
```
peterschrammel, line 1561: Good catch. Maybe add that this is explained in JLS 15.17.3.

peterschrammel (approved):

tautschnig (approved):

PR #6200, opened 2021-06-25
Introduce euclidean_mod expression

kroening:
This introduces a variant of the modulo operator, to be used on integers,
which uses Boute's Euclidean definition.  The result is never negative, and
thus differs from the C99 `%` operator.  The semantics match what SMT-LIB2
does in the integer theory.

The operator `%` is not defined by SMT-LIB2, and is removed from the parser.


martin-cs (approved):
A test case would be nice if possible.

TGWDB (approved):
Looks good, some test cases would be nice (particularly showing this difference on negative numbers).

peterschrammel (approved):

Issue #5977, opened 2021-03-25
Incorrect (hard-coded) logic for SMT backend


SaswatPadhi:
**CBMC version:** `develop`

**Operating system:** Mac OS 10.15.7

**Test case:**
```c
#include <assert.h>

void main()
{
  int i;
  __CPROVER_assume(i % 2 == 0);
  assert(__CPROVER_exists { int j ; i == j + j });
}
```

**Exact command line resulting in the issue:**

`cbmc --z3 test.c` or `cbmc --cvc4 test.c`

**What behaviour did you expect:**

CBMC would verify the assertion and report success.

**What happened instead:**

```console
$ cbmc --z3 test.c | tail
SMT2 solver returned error message:
        "line 88 column 17: model is not available"
Running SMT2 QF_AUFBV using Z3
Runtime Solver: 1.69421s
Runtime decision procedure: 1.69455s

** Results:
test.c function main
[main.assertion.1] line 7 assertion __CPROVER_exists { int j ; i == j + j }: ERROR

** 0 of 1 failed (1 iterations)
VERIFICATION ERROR

$ cbmc --cvc4 test.c | tail
SMT2 solver returned error message:
        "The logic was specified as QF_AUFBV, which doesn't include THEORY_QUANTIFIERS, but got a preprocessing-time fact for that theory.
The fact:
(forall ((|main::1::1::j!0#0| (_ BitVec 32))) (not (= (concat ((_ extract 30 0) |main::1::1::j!0#0|) (_ bv0 1)) |main::1::i!0@1#1|)) )"
Running SMT2 QF_AUFBV using CVC4
Runtime Solver: 0.0263352s
Runtime decision procedure: 0.0265899s

** Results:
main.c function main
[main.assertion.1] line 7 assertion __CPROVER_exists { int j ; i == j + j }: ERROR

** 0 of 1 failed (1 iterations)
VERIFICATION ERROR
```

**Additional context:**

CBMC also fails to verify this program with the SAT backend, but a similar issue has already been reported in #5395.

SaswatPadhi:
In my experience, just `(set-logic ALL)` works well with most solvers today. This is part of SMTLIB standard now (so is valid, syntactically) and might be the simplest fix.
But, what does everyone else think? cc: @kroening, @martin-cs, @tautschnig 

If performance hit is a concern, I can create a draft PR once we have Z3 running on the CI [#5949] and check how long the `smt-backend` tests take.

[...]

PR #6205, opened 2021-06-28
Fix bool versus bitvec 1 sort mismtach for CVC4

TGWDB:
This fixes a sort mismatch that can occur in CVC4 where an
array of booleans is translated into an array of (_ bitvec 1)
which is then a mismatch when CVC4 expects and can handle an
Array of Bool.

This solves some of the problems in #5977 .

Note that there are NO TESTS for this since we do not support cvc4 in CI.

### DIFF
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -96,6 +96,7 @@ smt2_convt::smt2_convt(
     break;
 
   case solvert::CVC4:
+    use_array_of_bool = true;
     use_as_const = true;
     break;
 


kroening:
All relevant solvers support arrays of bool now -- I'll do a PR that removes the `(_ bitvec 1)` encoding.

martin-cs (changes requested):
This seems like a work-around rather than a fix.  If you can find the code that is generating the sort mismatch then you should be able to fix this properly.

TGWDB:
@kroening @martin-cs It would be nice to approve and merge this PR to allow cvc4 to not error, regardless of future fixes to resolve the sort stuff generally as in #6206 .

peterschrammel (approved):
Let's merge this workaround, with a proper fix expected in #6206.

martin-cs (commented):
I don't think this is a good idea as it is a work around.  I am not sure #6206 will fix it either as I think that is needed for MathSAT support but ... @TGWDB if you really want to do it, I won't stand in your way.

martin-cs (approved):
I don't approve but this will remove my "request changes".

TGWDB:
@martin-cs Thanks. This is kind of a blocking concern for AWS since it prevents many things that should work with cvc4 from progressing. I agree a proper look at the SMT2 back end and fixing details there is preferable, but this is likely to be slow and/or caught up with other changes. Let's hope a proper solution is done in future.

PR #6206, opened 2021-06-28
remove encoding of arrays of bools into bitvector

kroening:
The encoding of arrays of booleans as arrays of `(_ bitvec 1)` was necessary
when solvers did not support arrays of booleans.

This support is now broadly available, removing the need for this code path.


martin-cs (commented):
I remember this being a MathSAT thing.  Have you checked if they now support it?

TGWDB:
> I remember this being a MathSAT thing. Have you checked if they now support it?

Sadly this looks true.
```
$ ./mathsat -version
MathSAT5 version 5.6.6 (218275631c24) (Apr 23 2021 07:55:22, gmp 6.1.2, gcc 7.5.0, 64-bit)
$ ./mathsat ~/github/testing/Issue5977/z3-latest.smt2
(error "Arrays with Bool as argument are not supported")
$ grep Array ~/github/testing/Issue5977/z3-latest.smt2 | grep Bool
(define-fun |__CPROVER_threads_exited#1| () (Array (_ BitVec 64) Bool) ((as const (Array (_ BitVec 64) Bool)) false))
```

martin-cs:
Yeah... IIRC it was because MathSAT has a strong separation between terms and formulae and so you can't have a formula-level proposition appearing in a term-level sort / expression.  Formally they are correct but it is a bit of a pain and one of the reasons why the foundations of SMT-LIB 3 are changing.

TGWDB:
Is there interest in updating this PR to include special handling for MathSAT? Even if @kroening is not interested in doing the change, if the end result is desirable I don't mind picking up from here and adding a special case for MathSAT to this code.

PR #6207, opened 2021-06-30
Fix bugs stemming from duplicate output streams in smt2_convt::write_footer() 

thomasspriggs:
This fixes bugs arising from some of the member functions of
`smt2::convt` writing to an `os` argument and some of the them writing
to `smt2_convt::out`. This was causing some of the generated text to be
written out when using `--outfile` argument with cbmc, but not without
it.

- [ ] n/a ~~My commit message includes data points confirming performance improvements (if claimed).~~

### DIFF
--- a/regression/cbmc/address_space_size_limit3/test.desc
+++ b/regression/cbmc/address_space_size_limit3/test.desc
@@ -1,4 +1,4 @@
-CORE
+CORE broken-smt-backend
 main.i
 --32 --little-endian --object-bits 25 --pointer-check
 ^EXIT=10$
--- a/src/solvers/smt2/smt2_dec.cpp
+++ b/src/solvers/smt2/smt2_dec.cpp
@@ -39,13 +39,15 @@ decision_proceduret::resultt smt2_dect::dec_solve()
     temp_file_stdout("smt2_dec_stdout_", ""),
     temp_file_stderr("smt2_dec_stderr_", "");
 
-  {
-    // we write the problem into a file
-    std::ofstream problem_out(
-      temp_file_problem(), std::ios_base::out | std::ios_base::trunc);
-    problem_out << stringstream.str();
-    write_footer(problem_out);
-  }
+  const auto write_problem_to_file = [&](std::ofstream problem_out) {
+    cached_output << stringstream.str();
+    stringstream.str(std::string{});
+    write_footer();
+    problem_out << cached_output.str() << stringstream.str();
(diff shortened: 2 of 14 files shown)


TGWDB (approved):

regression/cbmc/address_space_size_limit3/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+CORE broken-smt-backend
```
martin-cs, line 1: Why has this now broken?

src/solvers/smt2/smt2_dec.cpp
```diff
@@ -39,13 +39,15 @@ decision_proceduret::resultt smt2_dect::dec_solve()
...
     temp_file_stderr("smt2_dec_stderr_", "");
 
-  {
-    // we write the problem into a file
-    std::ofstream problem_out(
-      temp_file_problem(), std::ios_base::out | std::ios_base::trunc);
-    problem_out << stringstream.str();
-    write_footer(problem_out);
-  }
+  const auto write_problem_to_file = [&](std::ofstream problem_out) {
```
martin-cs, line 42: This seems longer, more complex and harder to debug than the code it replaces.  What am I missing?

martin-cs (approved):
Thanks for fixing all of these bugs; that is great to see.  I am not sure I really understand the changes to `smt2_dec.cpp` and why they were needed.

src/solvers/smt2/smt2_dec.cpp
TGWDB: The old code would do the following:

1. Write the internal stored string stream to file
2. Call the footer to explicitly write to the file

The problem was that the `write_footer` call would call another function (`define_object_size`) that would write to the internal string stream.

The desirable behaviour is to do the following:

1. Have stored string of the SMT that can be used for multiple calls to the solver
2. Write this stored string to the file
3. Add a footer to the file with the specific details of the current call/check (e.g. assumptions and `check-sat-assuming`).

The modified code stores the common SMT to the `cached_output` and then writes this at the beginning of each call. The `write_footer` function (and called functions such as `define_object_size`) now write to the same internal string stream as all the other SMT conversion code (reducing future misunderstandings like this fixes). The internal string stream is cleared before writing the footer and after, to ensure no accidental duplication of SMT instructions/data.

I hope this has helped clarify.

TGWDB:
I believe this also fixes #5952

PR #6209, opened 2021-07-02
VSD - bool intervals go top

jezhiggins:
An interval which spans its type's [min-max] goes TOP. By extension, a bool interval spanning [FALSE,TRUE] should also go TOP. 

This PR makes the changes to `interval_abstract_valuet` to add this behaviour. 
It also extends value-sets to set { FALSE, TRUE } to TOP.


regression/goto-analyzer/nondet-bool/test-constants.desc
```diff
@@ -0,0 +1,7 @@
+CORE
+main.c
+-vsd --one-domain-per-history --three-way-merge --loop-unwind-and-branching 16 --verify
```
martin-cs: Should that be `--vsd`?

martin-cs (approved):

PR #6210, opened 2021-07-05
SMT Z3 trace fix

TGWDB:
This PR fixes parsing of array values in traces from Z3.

### DIFF
--- a/regression/cbmc/enum-trace1/test_json.desc
+++ b/regression/cbmc/enum-trace1/test_json.desc
@@ -1,4 +1,4 @@
-CORE broken-smt-backend
+CORE
 main.c
 --json-ui --function test --trace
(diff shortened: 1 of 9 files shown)


peterschrammel (approved):
I don't see any arrays in these tests. Could you add any tests/checks for the added functionality?

martin-cs (approved):
Seems OK but 1. can we have a few tests for this and 2. this will probably need cross solver testing as it is not a standardised part of SMT-LIB.

TGWDB:
Merging this now that there are tests and everything passes. Also noting that this appears to do correct trace output for `cvc3` and `cvc4` as well, while also improving to some extent output for `cprover-smt2` albeit not as well.

PR #6211, opened 2021-07-05
Track conversion and use identifier instead of expression

TGWDB:
This tracks when an expression is converted by storing the
identifier in the SMT output in the defined_expressions
map. Then when asserting this expression, merely assert the
identifier and don't duplicate the expression in the output.

Note that this removed duplication of some SMT output where we used to output:
```
(define-fun |BXXX| () Bool (<some-complex-expression>)

(assert (<some-complex-expression>))
```
and instead this now outputs:
```
(define-fun |BXXX| () Bool (<some-complex-expression>)

(assert |BXXX|)
```

### DIFF
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -744,6 +744,14 @@ literalt smt2_convt::convert(const exprt &expr)
   no_boolean_variables++;
 
   out << "; convert\n";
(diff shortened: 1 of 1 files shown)


TGWDB:
@peterschrammel Any chance of codeowner review on this PR?

peterschrammel (approved):

PR #6213, opened 2021-07-06
Run regression tests with `z3` as the backend solver.

NlightNFotis:
As things currently stand, we install `z3` in a number of CI platforms,
but don't run the regression suite with `z3` as the backend solver.
This adds a job whose purpose is to run the regression tests under `regression/cbmc/`
with `z3`.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -119,6 +119,50 @@ jobs:
           env PATH=$PATH:`pwd`/src/solvers make -C regression/cbmc test-cprover-smt2
           make -C jbmc/regression test-parallel JOBS=2
 
+  # This job has been copied from the one above it, and modified to only build CBMC
+  # and run the `regression/cbmc/` regression tests against Z3. The rest of the tests
+  # (unit/regression) have been stripped based on the rationale that they are going
+  # to be run by the job above, which is basically the same, but more comprehensive.
+  # The reason we opted for a new job is that adding a `test-z3` step to the current
+  # jobs increases the job runtime to unacceptable levels (over 2hrs).
+  check-ubuntu-20_04-make-clang-smt-z3:
+    runs-on: ubuntu-20.04
+    env:
+      CC: "ccache /usr/bin/clang"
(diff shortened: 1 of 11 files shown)


.github/workflows/pull-request-checks.yaml
```diff
@@ -119,6 +119,50 @@ jobs:
           env PATH=$PATH:`pwd`/src/solvers make -C regression/cbmc test-cprover-smt2
           make -C jbmc/regression test-parallel JOBS=2
 
+  # This job has been copied from the one above it, and modified to only build CBMC
+  # and run the `regression/cbmc/` regression tests against Z3. The rest of the tests
+  # (unit/regression) have been stripped based on the rationale that they are going
+  # to be run by the job above, which is basically the same, but more comprehensive.
+  # The reason we opted for a new job is that adding a `test-z3` step to the current
+  # jobs increases the job runtime to unacceptable levels (over 2hrs).
+  check-ubuntu-20_04-make-clang-smt-z3:
```
thomasspriggs, line 128: :pick: I'd move `smt-z3` closer to the start of the name. That way the purpose of the job can be seen sooner and still read when the name has been truncated by github.

thomasspriggs (approved):
Looks good to me overall. However I note that the run time of the "Run regression/cbmc tests with z3 as the backend" step is down to 1m 10s with the long running test disabled. Given that the testing is that quick I am not sure how good an idea a separate job is. However I am approving as-is because running the tests with Z3 is a step forward whether they are done in a separate job or an existing job.

TGWDB:
> Looks good to me overall. However I note that the run time of the "Run regression/cbmc tests with z3 as the backend" step is down to 1m 10s with the long running test disabled. Given that the testing is that quick I am not sure how good an idea a separate job is. However I am approving as-is because running the tests with Z3 is a step forward whether they are done in a separate job or an existing job.

I would argue for a separate job since we don't want to enforce `z3` dependency any further than necessary... Also this way if there is some issue specifically with `z3`, it should only appear in one job.

TGWDB (approved):

PR #6215, opened 2021-07-07
SMT cache values and handle z3 errors

TGWDB:
This PR contains a fix for problems due to `cbmc` being able to send expressions with quantifiers to `z3`. This creates a problem where a `sat` result can be followed by an attempt to `(get-value |XXX|)` for various identifiers `XXX`. However, when `z3` has a quantifier in one of these identifiers this raises an error.

This PR addresses this problem by doing three things:

1. When we pass the solver input of the form `(assert |XXX|)` via the `set_to` function that is in turn referring to something in `defined_expressions`, then we also store what value we set this identifier to in `set_values`. (This covers various expressions, but most importantly here quantifier questions which `z3` will fail to handle with `(get-value |XXX|)`.)
2. When we parse output from the SMT solver if we hit an error relating to quantifiers after `z3` has determined the result is `sat`, then we ignore this error. This would lead to us missing results from the `(get-value ...)` expressions except for
3. When parsing results, if we already know the correct response for `sat` because we stored it in the `set_values` in 1 (above), then use this instead of parsing and checking the result from the solver. This also solves the problem with `z3` not being able to return the result and it being skipped in 2 (above).

### DIFF
--- a/regression/cbmc/union10/union_list2.desc
+++ b/regression/cbmc/union10/union_list2.desc
@@ -1,4 +1,4 @@
-CORE
+CORE broken-z3-smt-backend
 union_list2.c
 
 ^EXIT=0$
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -283,6 +283,7 @@ exprt smt2_convt::get(const exprt &expr) const
 
     if(it!=identifier_map.end())
       return it->second.value;
+    return expr;
   }
   else if(expr.id()==ID_nondet_symbol)
   {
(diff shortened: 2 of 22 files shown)


TGWDB:
@peterschrammel @tautschnig I would appreciate codeowner review (and ideally approved) here. I've tried to explain in the PR why this is done this way and what problem it solves. I'm also aware of critique regarding handling specific `z3` errors and special cases, but since we already give this input to `z3` (including in regression tests) I would argue that handling it in some form is better than simply reporting errors due to what `cbmc` creates internally. Also this is part of a series of fixes that AWS has found useful on a branch (https://github.com/NlightNFotis/cbmc/tree/smt_fast_response), so there is also argument from usage that this is valuable.

src/solvers/smt2/smt2_dec.cpp
```diff
@@ -175,17 +175,36 @@ decision_proceduret::resultt smt2_dect::read_result(std::istream &in)
...
+          messaget log{message_handler};
+          log.error() << "SMT2 solver returned error message:\n"
+                      << "\t\"" << message << "\"" << messaget::eom;
+          return decision_proceduret::resultt::D_ERROR;
+        }
       }
     }
   }
 
+  // If the result is satisfiable don't bother updating the assignments and
```
peterschrammel: Should be `is NOT satisfiable` ?

peterschrammel (approved):

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -283,6 +283,7 @@ exprt smt2_convt::get(const exprt &expr) const
 
     if(it!=identifier_map.end())
       return it->second.value;
+    return expr;
   }
```
kroening, line 287: It's wrong for `get` to return the symbol that it's asked to provide a model for.
The API is meant to return `nil` when no value is available.

kroening:
This is a horrible hack to work around a bug in Z3.
I would much prefer to raise this with the developers of Z3.

src/solvers/smt2/smt2_conv.cpp
TGWDB: This appears to have been caught from another PR/merging problem with git (and is already merged in #6210 ). That said, this fixes missing symbols in arrays when printing traces from SMT solvers.

regression/cbmc/union10/union_list2.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+CORE broken-z3-smt-backend
```
martin-cs, line 1: Why does this break?

martin-cs (commented):
I appreciate that "maybe do something else" is not a particularly helpful review comment but some general thoughts:

- In SMT-LIB 2.* there is little standarization of the representation of models and values.  For simple variables it is simple but for things like arrays and quantifiers it rapidly gets less obvious.
- This may change in SMT-LIB 3.0.  One of the reasons for adding things like lambdas to the language is to help create a way of outputting arrays and quantifiers.
- As such the output is quite solver-specific.
- `(get-model)` is a generally preferred interface than `(get-value)`.

So, maybe the correct way to handle this is to call `(get-model)` and then do a solver-specific parsing of what you get?

regression/cbmc/union10/union_list2.desc
TGWDB: These appear to have been broken before, previous version of `cbmc` have errors here.

TGWDB:
Broadly I agree on the points raised by both @kroening and @martin-cs , the ideal would be to do something better and not have this hack. However, the main concern here (and hence my raising this PR with the above comments/discussion) is that we are already sending "bad" input to `z3` (and other solvers). The goal here is to (better) handle where we send bad input to `z3` and in particular do this without major changes to how to invoke solvers. That said, feedback and commentary on how to do this right are welcome as the larger project (overview here #6134 ) can benefit from this.

PR #6218, opened 2021-07-08
VSD - pointers to heap allocations

jezhiggins:
Extends the variable sensitivity domain to be able to track heap allocated memory.

Within the goto-programs the heap allocation instruction is `ALLOCATE(malloc_size, 0 != 0)`. Within the vsd, this is an expression of type `ID_side_effect` with a statement of `ID_allocate`. When we eval this expression, we create a pointer object with a 'heap-object' placeholder expression. There's a little song and dance to propagate through the pointer type casts returns from malloc are subject to, but we can return 'heap-object' pointer back out.

When we assign pointer to a heap allocation to its symbol, we now know that actual (C) type of the object. We replace the 'heap object' placeholder with an array abstract object. The pointer abstract object can now act exactly as normal.


src/analyses/variable-sensitivity/abstract_pointer_object.cpp
```diff
@@ -87,6 +100,16 @@ abstract_object_pointert abstract_pointer_objectt::write_dereference(
   return std::make_shared<abstract_pointer_objectt>(type(), true, false);
 }
 
+abstract_object_pointert abstract_pointer_objectt::typecast(
+  const typet &new_type,
+  const abstract_environmentt &environment,
+  const namespacet &ns) const
+{
+  INVARIANT(is_void_pointer(type()), "Only allow pointer casting from void*");
```
martin-cs, line 108: I feel kind of conflicted about this.  In some ways it should not be an `INVARIANT` because it can be triggered by a user but also, at the moment we really aren't handling this case.  Could we return a `TOP` pointer of the correct type?  Reads from that should give `TOP` and writes to it could call `havoc()`?

src/analyses/variable-sensitivity/constant_pointer_abstract_object.cpp
```diff
@@ -201,14 +209,41 @@ abstract_object_pointert constant_pointer_abstract_objectt::write_dereference(
...
+    return shared_from_this();
   }
 }
 
+abstract_object_pointert constant_pointer_abstract_objectt::typecast(
+  const typet &new_type,
+  const abstract_environmentt &environment,
+  const namespacet &ns) const
+{
+  INVARIANT(is_void_pointer(type()), "Only allow pointer casting from void*");
```
martin-cs, line 224: As above.

src/analyses/variable-sensitivity/constant_pointer_abstract_object.cpp
```diff
@@ -201,14 +209,41 @@ abstract_object_pointert constant_pointer_abstract_objectt::write_dereference(
...
+  // Get an expression that we can assign to
+  exprt value = to_address_of_expr(value_stack.to_expression()).object();
+  if(value.id() == ID_dynamic_object)
+  {
+    auto &env = const_cast<abstract_environmentt &>(environment);
+
+    auto heap_array_type = array_typet(new_type.subtype(), nil_exprt());
+    auto array_object =
+      environment.abstract_object_factory(heap_array_type, ns, true, false);
+    auto heap_symbol = symbol_exprt("heap-object", heap_array_type);
```
martin-cs: Does this potentially clash heap objects of the same type?

src/analyses/variable-sensitivity/value_set_pointer_abstract_object.cpp
```diff
@@ -133,6 +142,15 @@ abstract_object_pointert value_set_pointer_abstract_objectt::write_dereference(
   return shared_from_this();
 }
 
+abstract_object_pointert value_set_pointer_abstract_objectt::typecast(
+  const typet &new_type,
+  const abstract_environmentt &environment,
+  const namespacet &ns) const
+{
+  INVARIANT(is_void_pointer(type()), "Only allow pointer casting from void*");
```
martin-cs, line 114: I see why but can this be left to the implementations of the objects in the set?

martin-cs (approved):

src/analyses/variable-sensitivity/constant_pointer_abstract_object.cpp
jezhiggins: Yes, and I'm working on that now.

src/analyses/variable-sensitivity/abstract_pointer_object.cpp
jezhiggins: I am similarly conflicted. Right now, the call to typecast() is itself guarded so that it will never be called if this invariant will fire

```
  if(expr.id() == ID_typecast)
  {
    const typecast_exprt &tce = to_typecast_expr(expr);
    if(tce.op().id() == ID_symbol && is_void_pointer(tce.op().type()))
``` 

Consequently existing typecasts, of which there are a handful, will evaluate as they do at the moment. I suppose this invariant is more of warning for future care than a guard against wacky corner cases.

src/analyses/variable-sensitivity/value_set_pointer_abstract_object.cpp
jezhiggins: This code is placeholder at the moment. More to follow.

jezhiggins: When we cast a value set of pointers, we need to also cast pointers it contains. We do this now.

src/analyses/variable-sensitivity/constant_pointer_abstract_object.cpp
jezhiggins: Each heap allocation gets its own symbol. Right now they just go 0, 1, 2, etc, but we could plug in something more sophisticated.

Issue #2295, opened 2018-06-06
goto-symex: "depth" option should be set to max-unsigned by default


tautschnig:
See https://github.com/diffblue/cbmc/pull/2262#discussion_r192809866.

PR #6222, opened 2021-07-12
Symex max depth default max

TGWDB:
This PR implements the recommended fix in [#2295](https://github.com/diffblue/cbmc/issues/2295).

### DIFF
--- a/jbmc/src/jbmc/jbmc_parse_options.cpp
+++ b/jbmc/src/jbmc/jbmc_parse_options.cpp
@@ -113,6 +113,7 @@ void jbmc_parse_optionst::set_default_options(optionst &options)
 
   // Other default
   options.set_option("arrays-uf", "auto");
+  options.set_option("depth", UINT32_MAX);
 }
 
 void jbmc_parse_optionst::get_command_line_options(optionst &options)
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -117,6 +117,7 @@ void cbmc_parse_optionst::set_default_options(optionst &options)
 
   // Other default
   options.set_option("arrays-uf", "auto");
+  options.set_option("depth", UINT32_MAX);
 }
 
 void cbmc_parse_optionst::get_command_line_options(optionst &options)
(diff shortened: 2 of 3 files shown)


martin-cs (approved):

kroening:
The default should be dealt with earlier, meaning when setting up `optionst`.

martin-cs (changes requested):
Daniel is correct; I should not have approved this.  The correct fix is an `else` here:

https://github.com/diffblue/cbmc/blob/06c563a76589ed00cc372d68e2819880537977b2/src/cbmc/cbmc_parse_options.cpp#L261

which sets the correct value.  It might be worth seeing if `unwind` has similar behaviour.

TGWDB:
I tried changing the creation of the `optiont` and this now breaks unit tests (these can be fixed). I'm unclear why the `"depth"` option is a special case that should be handled outside, when other options passed in to the `symex_config` that are missing are handled in the constructor. (The main difference appears to be that `"depth"` has a different default value, but this is also handled in the constructor for the `"max_field_sensitivity_array_size"` option.)

martin-cs:
@TGWDB It is a fair question.  In an ideal world the `*_parse_options.cpp` file should read the `cmdlinet` and use it to populate an `optionst` along with default values if they are not 0 (or whatever `optionst` returns for the "asking for an option that hasn't been set" case).  Unit tests can then simply create an appropriate `optionst`.  However it is made, it is the `optionst` which gets passed to the actual code and used to configure things.  Ideally this will be in a constructor, RAII-style.  There is a little bit of a grey area about whether and how the constructors should handle cases when the configuration is complex and has multiple, non-orthogonal options.

Historically, this bit of design intent has not been as widely followed as maybe it should have been, so in the older parts of the code you do find configuration parameters lurking much deeper.  These should be treated as a legacy issue and ideally modernised and moved into the "`parse_options.cpp` builds an `optionst` and that is it for configuration" world.

HTH

TGWDB:
Rather than an explicit `else` case I set the default value for the option in the creation of `optionst`. This seems like the cleanest solution.

martin-cs (approved):
Seems a reasonable solution.

TGWDB:
@kroening Any chance of a codeowner review now?

tautschnig (approved):

tautschnig:
@TGWDB Please cleanup the commit history on this one.

TGWDB:
> @TGWDB Please cleanup the commit history on this one.

Commit messages tidied up. This should just be waiting on codeowner review now.

martin-cs:
Paging @peterschrammel

peterschrammel (approved):

Issue #6230, opened 2021-07-15
Commit `cd1ec89b` breaks proofs


SaswatPadhi:
**CBMC version:** `develop` (currently 06c563a7)
**Operating system:** Mac OS / Linux

**Exact command line resulting in the issue:**
We don't have a "small" example yet, but here is a broken proof harness from s2n:
```console
$ git clone https://github.com/aws/s2n-tls.git && cd s2n-tls
$ git submodule update --init --checkout --recursive
$ cd tests/cbmc/proofs/s2n_socket_write_cork
$ make result
```

**What behaviour did you expect:**
This harnesses can still be verified after updating CBMC.

**What happened instead:**
CBMC crashes with an invariant violation:

```
--- begin invariant violation report ---
Invariant check failed
/Users/jmgruj/cbmc/src/solvers/flattening/arrays.cpp:200 function: collect_arrays
Condition: struct_op.id() == ID_symbol || struct_op.id() == ID_nondet_symbol
Reason: unexpected array expression: member with 'member'
Backtrace:
0   cbmc                                0x000000010ab25eea _Z15print_backtraceRNSt3__113basic_ostreamIcNS_11char_traitsIcEEEE + 74
1   cbmc                                0x000000010ab26482 _Z13get_backtracev + 210
2   cbmc                                0x000000010a56130c _Z29invariant_violated_structuredI17invariant_failedtJRKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEEEENS1_9enable_ifIXsr3std10is_base_ofIS0_T_EE5valueEvE4typeES9_S9_iS9_DpOT0_ + 44
3   cbmc                                0x000000010a5612d9 _Z25invariant_violated_stringRKNSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEES7_iS7_S7_ + 9
4   cbmc                                0x000000010a9081ab _ZN7arrayst14collect_arraysERK5exprt + 1451
5   cbmc                                0x000000010a9077e6 _ZN7arrayst21record_array_equalityERK11equal_exprt + 422
6   cbmc                                0x000000010a94de01 _ZN12bv_pointerst12convert_restERK5exprt + 1185
7   cbmc                                0x000000010a9a5358 _ZN17prop_conv_solvert12convert_boolERK5exprt + 2616
8   cbmc                                0x000000010a9a48a9 _ZN17prop_conv_solvert7convertERK5exprt + 905
[...]

PR #6232, opened 2021-07-16
Revert the change to the use of unbounded arrays

martin-cs:
This was introduced in #6194 but reported as breaking in #6230.

### DIFF
--- a/src/solvers/flattening/boolbv.h
+++ b/src/solvers/flattening/boolbv.h
@@ -47,7 +47,7 @@ public:
     message_handlert &message_handler,
     bool get_array_constraints = false)
     : arrayst(_ns, _prop, message_handler, get_array_constraints),
-      unbounded_array(unbounded_arrayt::U_AUTO),
+      unbounded_array(unbounded_arrayt::U_NONE),
       bv_width(_ns),
       bv_utils(_prop),
       functions(*this),


martin-cs:
@SaswatPadhi it would be great if you could let me know if this fixes your issue.

TGWDB (approved):
Happy to approve (specially for faster turn around). Any chance of a regression test?

martin-cs:
@TGWDB I would *love* a regression test and was hoping that @SaswatPadhi might be able to provide one.

SaswatPadhi:
Thanks for the fix, @martin-cs.

We noticed this on a large s2n proof harness, but let me try to reduce it to a small regression test. I would look into large / dynamically allocated arrays, as you suggested.

martin-cs:
Thanks @SaswatPadhi

SaswatPadhi:
This is the most I could shrink it to:

```c
#include <assert.h>
#include <stdint.h>
#include <stdlib.h>

static const uint16_t extensions[] = { 0, 1, 2, 3, 4, 5 };

#define COUNT (sizeof(extensions) / sizeof(extensions[0]))

typedef struct {
  uint8_t *data;
  uint32_t size;
  uint32_t allocated;
  unsigned growable :1;
} ext;

struct context {
  ext exts[COUNT];
};

struct conn {
  void *test;
  struct context ctx;
};

int main() {
  struct conn *s = malloc(sizeof(*s));
  if(s != NULL) s->test = NULL;
}
```

If I shrink the `ext` struct further, the invariant violation disappears, so may be there is some issue with alignment? Not sure ...

martin-cs:
I have tried this example with the latest develop (  06c563a76589ed00cc372d68e2819880537977b2 ) and it doesn't crash.  Can you double check and say something about the environment?

TGWDB:
I also tested the example on `cbmc` release `5.33.0`, `5.34.0`, and a branch with the fix and had no crashes for any of them (on Ubuntu 20.04 LTS). That said, when I was working on original PR that had this error, there were some behaviours that appeared to be system specific w.r.t. array sizes.

SaswatPadhi:
Sorry, I should have mentioned the full commandline:

```
cbmc --malloc-may-fail --malloc-fail-null --pointer-check test.c
```

martin-cs:
Ah, yes, that does it! Thanks @SaswatPadhi

feliperodri:
Any updates on that PR? @martin-cs @kroening

martin-cs:
@feliperodri I am working on the test case that @SaswatPadhi sent and will add it ASAP.

kroening (approved):

PR #6236, opened 2021-07-19
Multidimensional decreases clauses

LongPham7:
This pull request is about the implementation of "multidimensional" decreases clauses. "One-dimensional" decreases clauses are already supported by the existing version of CBMC. We extend one-dimensional decreases clauses to multidimensional ones by admitting tuples of integer-typed functions. The syntax of a multidimensional decrease clause is `__CPROVER_decreases(f1, ..., fn)`. We use the lexicographic order to compare two tuples. 


src/util/std_expr.h
```diff
@@ -2126,6 +2126,11 @@ inline not_exprt &to_not_expr(exprt &expr)
   return ret;
 }
 
+// Create a lexicographic less-than relation between two variables.
```
martin-cs: I see value in having this as a utility function but this is not the place for it.  `std_expr.h` is for the `irept` subclasses that give interfaces for expressions.  It's a bit of a weird case though that doesn't obviously fit into any of the other parts of `util/`.  Maybe it could have it's own file?

martin-cs (approved):
Looks good apart from the code placement issue flagged.

src/util/std_expr.h
LongPham7: Thanks a lot for your comment! I will discuss the issue with Saswat.

regression/contracts/variant_multidimensional_nested/test.desc
```diff
@@ -0,0 +1,13 @@
...
+activate-multi-line-match
+^\[main.1\].*Check loop invariant before entry: SUCCESS$
+^\[main.2\].*Check that loop invariant is preserved: SUCCESS$
+^\[main.3\].*Check decreases clause on loop iteration: SUCCESS$
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+This test succeeds. It tests whether a multidimensional decreases clause works
+properly.
```
SaswatPadhi: Why is this test case called "nested" though? ...

regression/contracts/variant_multidimensional_not_decreasing_fail/test.desc
```diff
@@ -0,0 +1,16 @@
...
+^\[main.3\] .* Check decreases clause on loop iteration: FAILURE$
+^.*1 of 3 failed \(2 iterations\)$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+This test fails because the given multidimensional decreases clause does not 
+monotonically decrease. A mistake is in the second component of the
+decreases clause: j. It should instead be N - j.
```
SaswatPadhi: Can we also add another loop in the function where the first component is not strictly monotonically decreasing? Just to be sure that we are indeed checking all components of tuples (well at least of size 2, in this case).

regression/contracts/variant_parsing_multiple_clauses_fail/test.desc
```diff
@@ -0,0 +1,16 @@
...
+^EXIT=(1|64)$
+^SIGNAL=0$
+--
+--
+This test fails because we have multiple decreases clauses. CBMC only admits
+one decreases clause for each loop. If one wants to write a multidimensional
+decreases clause, it should be placed inside a single __CPROVER_decreases(...),
+where each component of the multidimensional decreases clause is separated by a
+comma. Hence, in this test, the correct syntax is 
+__CPROVER_decreases(N - i, 42).
```
SaswatPadhi, line 16: Nice! :) These test.desc files are very well written.

src/ansi-c/parser.y
```diff
@@ -45,6 +45,8 @@ extern char *yyansi_ctext;
 // requires more stack space
 #define YYMAXDEPTH 25600
 
+#include <stdio.h>
```
SaswatPadhi: Why do we need this?

src/goto-instrument/code_contracts.cpp
```diff
@@ -155,6 +155,15 @@ void code_contractst::check_apply_loop_contracts(
     invariant = conjunction(invariant.operands());
   }
 
+  // Vector representing a (possibly multidimensional) decreases clause
+  exprt::operandst decreases_clause_vector = decreases_clause.operands();
```
SaswatPadhi:
```suggestion
  const auto decreases_clause_vector = decreases_clause.operands();
```

src/goto-instrument/code_contracts.cpp
```diff
@@ -155,6 +155,15 @@ void code_contractst::check_apply_loop_contracts(
...
+  int size_decreases_clasue = decreases_clause_vector.size();
```
SaswatPadhi: I think this should be `size_t` not `int`, but please double check what STL says. Not that we would have vectors that are longer than INT_MAX, but it's always good to use the correct type :)

```suggestion
  const size_t size_decreases_clasue = decreases_clause_vector.size();
```

Also use `size_t` in place of `int` for the loop counters below.

src/goto-instrument/code_contracts.cpp
```diff
@@ -211,30 +220,44 @@ void code_contractst::check_apply_loop_contracts(
...
 
   if(!decreases_clause.is_nil())
   {
-    // Generate: a declaration of the temporary variable that stores the
-    // decreases clause's value before the loop
-    havoc_code.add(goto_programt::make_decl(
-      old_decreases_symbol, loop_head->source_location));
+    // Generate: declarations of the temporary variables that stores the
+    // multidimensional decreases clause's value before the loop
+    for(auto &old_temp_var : old_temporary_variables)
```
SaswatPadhi:
```suggestion
    for(const auto &old_temp_var : old_temporary_variables)
```

src/goto-instrument/code_contracts.cpp
```diff
@@ -262,33 +285,45 @@ void code_contractst::check_apply_loop_contracts(
...
-    converter.goto_convert(new_decreases_assignment, havoc_code, mode);
-
-    // Generate: assertion that the decreases clause's value after the loop is
-    // smaller than the value before the loop
-    code_assertt monotonic_decreasing_assertion{
-      binary_relation_exprt(new_decreases_symbol, ID_lt, old_decreases_symbol)};
+    // Generate: declarations of temporary variables that stores the
+    // multidimensional decreases clause's value after one arbitrary iteration
+    // of the loop
+    for(auto &new_temp_var : new_temporary_variables)
```
SaswatPadhi:
```suggestion
    for(const auto &new_temp_var : new_temporary_variables)
```

src/util/std_expr.cpp
```diff
@@ -43,6 +43,58 @@ exprt conjunction(const exprt::operandst &op)
...
+// This is used in the implementation of multidimensional decreases clauses.
+exprt create_lexicographic_less_than(
+  const std::vector<symbol_exprt> &small,
+  const std::vector<symbol_exprt> &large)
+{
+  PRECONDITION(small.size() == large.size());
+
+  if(small.size() == 0)
+  {
+    return true_exprt();
```
SaswatPadhi: Shouldn't this be `false_exprt()`?

Empty tuple is not _strictly_ less than another empty tuple.

SaswatPadhi (changes requested):
Some general comments:
- Please remove `activate-multi-line-match` from .desc files that don't need multi-line regex matching
- Please only check regexes specific to the functionality you want to test (i.e. decreases clauses), because the other assertions (and especially their count, like `^.*0 of 5 failed \(1 iterations\)$`) may change in future if we introduce additional default checks.

regression/contracts/variant_multidimensional_nested/test.desc
LongPham7: Thanks for spotting this typo! Initially, I had a nested while-loop (i.e. an outer loop and an inner loop) in mind before writing this regression test. That's why I call this regression test "nested." However, I then decided to implement a single-while loop with two index variables, `i` and `j`. I will update the name to a more appropriate one.

regression/contracts/variant_multidimensional_not_decreasing_fail/test.desc
LongPham7: Got it.

src/ansi-c/parser.y
LongPham7: Thanks for spotting a redundant line.

src/util/std_expr.cpp
LongPham7: Thanks for spotting a bug! I agree that the output should be false.

src/util/std_expr.h
LongPham7: @kroening Can I have your input? As Martin pointed out, `std_expr.h` may not be a desirable place for the function `create_lexicographic_less_than`. However, creating a separate file just for the function `create_lexicographic_less_than` is not an ideal solution, either.

src/goto-instrument/code_contracts.cpp
SaswatPadhi: Interestingly, there is a recent Linux exploit that used a `size_t` to `int` overflow: https://www.qualys.com/2021/07/20/cve-2021-33909/sequoia-local-privilege-escalation-linux.txt
😃

src/util/std_expr.h
```diff
@@ -2126,6 +2126,11 @@ inline not_exprt &to_not_expr(exprt &expr)
   return ret;
 }
 
+// Create a lexicographic less-than relation between two tuples of variables.
+// This is used in the implementation of multidimensional decreases clauses.
+exprt create_lexicographic_less_than(
+  const std::vector<symbol_exprt> &small,
+  const std::vector<symbol_exprt> &large);
```
kroening: `std_expr.h` is the wrong place for this. Consider `expr_util.h`. Or leave it local with the code that uses it; `util/` is meant to be used when there is at least strong potential that there will be more than one user. This may happen one day, and then the function could move.

src/util/std_expr.cpp
```diff
@@ -43,6 +43,58 @@ exprt conjunction(const exprt::operandst &op)
   }
 }
 
+// Create a lexicographic less-than relation between two tuples of variables.
+// This is used in the implementation of multidimensional decreases clauses.
+exprt create_lexicographic_less_than(
+  const std::vector<symbol_exprt> &small,
+  const std::vector<symbol_exprt> &large)
+{
```
kroening: I'd call the parameters `lhs` and `rhs`, to be consistent with the other binary relations.

src/util/std_expr.cpp
```diff
@@ -43,6 +43,58 @@ exprt conjunction(const exprt::operandst &op)
...
+  PRECONDITION(small.size() == large.size());
+
+  if(small.size() == 0)
+  {
```
kroening: `.empty()` might be slightly more readable.

src/util/std_expr.cpp
```diff
@@ -43,6 +43,58 @@ exprt conjunction(const exprt::operandst &op)
...
+
+  consecutive_equalities[0] =
+    binary_relation_exprt(small[0], ID_equal, large[0]);
+  for(unsigned int i = 1; i < consecutive_equalities.size() - 1; i++)
+  {
+    exprt current_component_equality =
+      binary_relation_exprt(small[i], ID_equal, large[i]);
+    consecutive_equalities[i] =
+      and_exprt(consecutive_equalities[i - 1], current_component_equality);
+  }
```
kroening: I would separate the creation of the equalities and the forming of the conjunction. In particular, we have a `conjunction(...)` helper, which takes `exprt::operandst` as argument.

src/util/std_expr.cpp
```diff
@@ -43,6 +43,58 @@ exprt conjunction(const exprt::operandst &op)
...
+  lexicographic_possibilities[0] =
+    binary_relation_exprt(small[0], ID_lt, large[0]);
+  for(unsigned int i = 1; i < lexicographic_possibilities.size(); i++)
+  {
+    exprt current_component_less_than =
+      binary_relation_exprt(small[i], ID_lt, large[i]);
+    lexicographic_possibilities[i] =
+      and_exprt(consecutive_equalities[i - 1], current_component_less_than);
+  }
+
```
kroening: Same here.

src/util/std_expr.h
LongPham7: I see. Thanks for your suggestion! I will place the function locally in `contracts.cpp`.

src/util/std_expr.cpp
LongPham7: If we use `conjunction(...)` to create conjunctions of equalities of all necessary lengths (i.e. from 1 up to the size of the multidimensional decreases clause), I believe it will take quadratic time in total. Of course, the size of multidimensional decreases clauses is usually small. So the algorithmic performance should not be a concern. Nonetheless, I think it is better to simply traverse the vector and incrementally build conjunctions of equalities, which will take linear time.

src/util/std_expr.h
LongPham7: As @kroening suggested, I decided to put the function `create_lexicographic_less_than` in the file `contracts.cpp`.

regression/contracts/variant_missing_invariant_warning/test.desc
```diff
@@ -1,10 +1,7 @@
 CORE
 main.c
 --apply-loop-contracts
-activate-multi-line-match
 ^The loop at file main.c line 4 function main does not have a loop invariant, but has a decreases clause. Hence, a default loop invariant \('true'\) is being used.$
-^\[main.1\].*Check loop invariant before entry: SUCCESS$
-^\[main.2\].*Check that loop invariant is preserved: SUCCESS$
```
feliperodri: Why do we need to remove these from the test description?

regression/contracts/variant_multidimensional_ackermann/test.desc
```diff
@@ -0,0 +1,29 @@
+CORE
+main.c
+--apply-loop-contracts --replace-all-calls-with-contracts
+^\[ackermann.3\] .* Check decreases clause on loop iteration: SUCCESS$
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+This test succeeds. It tests whether we can prove (only partially) the
```
feliperodri: Instead of writing "This test succeeds", we could check if the final result is `VERIFICATION SUCCESSFUL`.

regression/contracts/variant_multidimensional_not_decreasing_fail1/main.c
```diff
@@ -0,0 +1,23 @@
+
```
feliperodri: remove empty space.

regression/contracts/variant_multidimensional_not_decreasing_fail1/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--apply-loop-contracts
+^\[main.3\] .* Check decreases clause on loop iteration: FAILURE$
+^EXIT=10$
+^SIGNAL=0$
```
feliperodri, line 9: VERIFICATION FAILED

regression/contracts/variant_multidimensional_not_decreasing_fail2/main.c
```diff
@@ -0,0 +1,23 @@
+
```
feliperodri: remove unnecessary empty line.

regression/contracts/variant_multidimensional_not_decreasing_fail2/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--apply-loop-contracts
+^\[main.3\] .* Check decreases clause on loop iteration: FAILURE$
+^EXIT=10$
+^SIGNAL=0$
```
feliperodri, line 9: VERIFICATION FAILED

regression/contracts/variant_multidimensional_two_index_variables/main.c
```diff
@@ -0,0 +1,23 @@
+
```
feliperodri: remove empty line.

regression/contracts/variant_multidimensional_two_index_variables/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--apply-loop-contracts
+^\[main.3\].*Check decreases clause on loop iteration: SUCCESS$
+^EXIT=0$
+^SIGNAL=0$
```
feliperodri, line 9: VERIFICATION SUCCESSFUL

regression/contracts/variant_multidimensional_two_index_variables/test.desc
```diff
@@ -0,0 +1,10 @@
...
+--
+--
+This test succeeds. It tests whether a multidimensional decreases clause works
```
feliperodri: Remove "This test succeeds" after adding `VERIFICATION SUCCESSFUL`.

regression/contracts/variant_not_decreasing_fail/test.desc
```diff
@@ -1,12 +1,7 @@
 CORE
 main.c
 --apply-loop-contracts
-^main.c function main$
-^\[main.1\] .* Check loop invariant before entry: SUCCESS$
-^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
```
feliperodri: Why removing these checks?

regression/contracts/variant_parsing_multiple_clauses_fail/test.desc
```diff
@@ -0,0 +1,16 @@
+CORE
+main.c
+--apply-loop-contracts
+activate-multi-line-match
```
feliperodri, line 4: Why do we need this here?

regression/contracts/variant_parsing_multiple_clauses_fail/test.desc
feliperodri: Shouldn't we support multiple decreases clauses in the future?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -828,9 +828,11 @@ void c_typecheck_baset::typecheck_spec_decreases(codet &code)
 {
   if(code.find(ID_C_spec_decreases).is_not_nil())
   {
-    exprt &variant = static_cast<exprt &>(code.add(ID_C_spec_decreases));
-
-    typecheck_expr(variant);
-    implicit_typecast_arithmetic(variant);
+    for(auto &decreases_clause_component :
+        (static_cast<exprt &>(code.add(ID_C_spec_decreases)).operands()))
```
feliperodri, line 832: Are there any guarantees that this cast is safe? Take a look at this comment https://github.com/diffblue/cbmc/pull/5403#discussion_r468825167.

src/util/std_expr.h
```diff
@@ -2126,7 +2126,6 @@ inline not_exprt &to_not_expr(exprt &expr)
   return ret;
 }
 
-
```
feliperodri: Clean ups should be on separate commit.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -34,6 +34,12 @@ Date: February 2016
 #include <util/pointer_offset_size.h>
 #include <util/replace_symbol.h>
 
+// Create a lexicographic less-than relation between two tuples of variables.
+// This is used in the implementation of multidimensional decreases clauses.
+exprt create_lexicographic_less_than(
+  const std::vector<symbol_exprt> &lhs,
+  const std::vector<symbol_exprt> &rhs);
```
feliperodri: Function declarations should go to header files.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -86,6 +92,15 @@ void code_contractst::check_apply_loop_contracts(
     invariant = conjunction(invariant.operands());
   }
 
+  // Vector representing a (possibly multidimensional) decreases clause
+  const auto decreases_clause_vector = decreases_clause.operands();
+  const size_t size_decreases_clasue = decreases_clause_vector.size();
```
feliperodri: You don't need to store the size on a separate variable, you can just call `size()`. Even better, vectors are traversed using iterators.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -142,30 +157,44 @@ void code_contractst::check_apply_loop_contracts(
...
+                                        function_name,
+                                        mode)
+                                        .symbol_expr());
+    new_temporary_variables.push_back(new_tmp_symbol(
+                                        decreases_clause_vector[i].type(),
+                                        loop_head->source_location,
+                                        function_name,
+                                        mode)
+                                        .symbol_expr());
+  }
```
```suggestion
  for(const auto &clause : decreases_clause.operands())
  {
    old_temporary_variables.push_back(new_tmp_symbol(
                                        clause.type(),
                                        loop_head->source_location,
                                        function_name,
                                        mode)
                                        .symbol_expr());
    new_temporary_variables.push_back(new_tmp_symbol(
                                        clause.type(),
                                        loop_head->source_location,
                                        function_name,
                                        mode)
                                        .symbol_expr());
  }
```
Same for all the other loops.

feliperodri (changes requested):

regression/contracts/variant_missing_invariant_warning/test.desc
LongPham7: Saswat recommended that I should only keep those regular expressions that are actually relevant to what I want to test (that is, decreases clauses). So I decided to remove those lines that are about loop "invariants" (but not loop variants). Having said that, loop invariants are still (usually) necessary for termination proofs, so it would also make sense to keep those lines. Any suggestion?

regression/contracts/variant_parsing_multiple_clauses_fail/test.desc
SaswatPadhi: It's necessary for the multi-line regex match for parsing error (see the next line).

regression/contracts/variant_multidimensional_ackermann/test.desc
LongPham7: OK. I will include `VERIFICATION SUCCESSFUL` in the regular expressions then. I used to have that line, but I removed it.

src/goto-instrument/contracts/contracts.cpp
LongPham7: Oh, yes, you're right. I should use an iterator.

regression/contracts/variant_parsing_multiple_clauses_fail/test.desc
SaswatPadhi: It's unclear as of now.

`__CPROVER_decreases` is unlike the other contracts. You can arbitrarily reorder multiple ENSURES, REQUIRES, ASSIGNS, LOOP_INVARIANT clauses, but the order of expressions matters for DECREASES clauses. They are not a conjunction (or a set), but an ordered tuple.

Moreover,

```
__CPROVER_decreases(X)
__CPROVER_decreases(Y)
```

reads as "both" X and Y monotonically decrease at each iteration, but for termination we simply need either one to monotonically decrease, and this sort of annotation would just be confusing.

src/util/std_expr.h
SaswatPadhi: I'd say best to simply remove this change, and not aim for this minor cleanup in this PR at all.

regression/contracts/variant_missing_invariant_warning/test.desc
SaswatPadhi: Yes, my point was mainly about the counts that you had `X of Y failed (Z iterations)`, and any other unrelated default checks.

I agree with Felipe that we should keep the assertions related to invariants, since they are going to be checked for variants as well.

LongPham7: OK. I just talked with Saswat, and apparently, I removed too many lines from regular expressions in regression tests. I will add them back!

src/util/std_expr.h
LongPham7: Thanks for spotting this! I removed this empty line by mistake.

src/goto-instrument/contracts/contracts.cpp
LongPham7: I don't want the function `create_lexicographic_less_than` to be automatically included whenever the header file `contracts.h` is included. As of now, this function is only used in the implementation of multidimensional decreases clauses. So I am basically the only client of this function. That's why I want to keep this function separate from the rest of the header file `contracts.h`. Any suggestion?

src/goto-instrument/contracts/contracts.cpp
LongPham7: Thanks a lot!

src/goto-instrument/contracts/contracts.cpp
LongPham7: Actually, I still need the index variable `i` because I traverse multiple vectors (of the same length) simultaneously. So as you suggested, I will access the size via `size()` instead of storing it in a designated variable.

src/goto-instrument/contracts/contracts.cpp
feliperodri: @LongPham7 take a look at [static functions](https://www.geeksforgeeks.org/what-are-static-functions-in-c/).

LongPham7: OK. I will make the function static. Thanks for the suggestion!

src/ansi-c/c_typecheck_code.cpp
SaswatPadhi: Good catch. As suggested on the other thread, we should create "safe" accessors at some point.

However, currently, this issue also applies to invariant contracts, so may be we can table this change for a separate future PR and fix both together?

LongPham7: Thanks for the link. For now, I will just leave it as it is. When someone cleans up the code for loop invariants, then we will also clean the code for decreases clauses.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -235,6 +271,54 @@ void code_contractst::check_apply_loop_contracts(
...
+  // l2, l3>.
+  // Then this vector stores <s1 == l1, s1 == l1 && s2 == l2,
+  // s1 == l1 && s2 == l2 && s3 == l3>.
+  // In fact, the last element is unnecessary, so we do not create it.
+  exprt::operandst equality_conjunctions(lhs.size());
+  equality_conjunctions[0] = binary_relation_exprt(lhs[0], ID_equal, rhs[0]);
+  for(unsigned int i = 1; i < equality_conjunctions.size() - 1; i++)
+  {
+    exprt current_component_equality =
+      binary_relation_exprt(lhs[i], ID_equal, rhs[i]);
```
SaswatPadhi: Consider removing redundant type annotation (`exprt`):
```suggestion
    binary_relation_exprt component_i_equality{lhs[i], ID_equal, rhs[i]};
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -235,6 +271,54 @@ void code_contractst::check_apply_loop_contracts(
...
+  // l2, l3>.
+  // Then this vector stores <s1 < l1, s1 == l1 && s2 < l2, s1 == l1 &&
+  // s2 == l2 && s3 < l3>.
+  exprt::operandst lexicographic_individual_comparisons(lhs.size());
+  lexicographic_individual_comparisons[0] =
+    binary_relation_exprt(lhs[0], ID_lt, rhs[0]);
+  for(unsigned int i = 1; i < lexicographic_individual_comparisons.size(); i++)
+  {
+    exprt current_component_less_than =
+      binary_relation_exprt(lhs[i], ID_lt, rhs[i]);
```
SaswatPadhi: Consider removing redundant type annotation (`exprt`):
```suggestion
    binary_relation_exprt component_i_less_than{lhs[i], ID_lt, rhs[i]};
```

SaswatPadhi (approved):
Thanks for all the changes!

LGTM with a couple of nitpicks:

regression/contracts/variant_multidimensional_ackermann/test.desc
```diff
@@ -0,0 +1,34 @@
...
+
+Furthermore, the Ackermann function has a function contract that the result
+is always non-negative. This post-condition is necessary for establishing
+the loop invariant. However, in this test, we do not enforce the function
+contract. Instead, we assume that the function contract is correct and use it
+(i.e. replace a recursive call of the Ackermann function with its contract). 
+
+We cannot verify/enforce the function contract of the Ackermann function, since
+CBMC does not support function contracts for recursively defined functions.
+As of now, CBMC only supports function contracts for non-recursive functions.
```
feliperodri, line 34: Kudos for your bar-raising test descriptions! @LongPham7

feliperodri (approved):
Thank you for implementing all comments!

chrisr-diffblue (approved):

PR #6237, opened 2021-07-20
Xml description addition

### LINKED ISSUE #3798: XML output of properties lacks description
See discussion at https://github.com/diffblue/cbmc/pull/3583#discussion_r247361722: JSON and plain-text output also print the description of a property, XML output doesn't.

TGWDB:
This unifies the XML output with the JSON output by including the `description` in XML.

Fixes #3798

### DIFF
--- a/doc/assets/xml_spec.xsd
+++ b/doc/assets/xml_spec.xsd
@@ -166,6 +166,7 @@
       <xs:sequence>
         <xs:element ref="goto_trace" minOccurs="0"/>
       </xs:sequence>
+      <xs:attribute name="description" type="xs:string"/>
       <xs:attribute name="property" type="xs:string"/>
       <xs:attribute name="status" type="xs:string"/>
     </xs:complexType>
--- a/regression/cbmc/ACSL/fail-operators.c
+++ b/regression/cbmc/ACSL/fail-operators.c
@@ -1,0 +1,36 @@
+void boolean()
+{
+  __CPROVER_bool a, b;
+  __CPROVER_assert(!(a ≡ b) == (a == b), "≥");
+  __CPROVER_assert(!(a ≢ b) == (a != b), "≢");
+  __CPROVER_assert(!(a ⇒ b) == (!a || b), "⇒");
+  __CPROVER_assert(!(a ⇔ b) == (a == b), "⇔");
+  __CPROVER_assert(!(a ∧ b) == (a && b), "∧");
+  __CPROVER_assert(!(a ∨ b) == (a || b), "∨");
+  __CPROVER_assert(!(a ⊻ b) == (a != b), "⊻");
+  __CPROVER_assert(!(¬ a) == (!a), "¬");
+}
+
+void relations()
+{
+  int a, b;
+  __CPROVER_assert(!(a ≥ b) == (a >= b), "≥");
+  __CPROVER_assert(!(a ≤ b) == (a <= b), "≤");
+  __CPROVER_assert(!(a ≡ b) == (a == b), "≥");
+  __CPROVER_assert(!(a ≢ b) == (a != b), "≢");
+  __CPROVER_assert(!(− a) == (-a), "−");
+}
+
+void quantifiers()
+{
+  __CPROVER_assert(!(∀ int i; 1) == 1, "∀");
+  __CPROVER_assert(!(∃ int i; 1) == 1, "∃");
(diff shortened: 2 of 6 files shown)


martin-cs (approved):
In principle, seems like a good idea to be consistent and that this is useful information.  However, we do need to be careful with output formats, esp. machine readable ones.  I think this should be backwards compatible though.

TGWDB:
I recommend reviewing by commit, each is quite straightforward but do different things.

regression/cbmc/ACSL/fail-operators.c
```diff
@@ -0,0 +1,36 @@
+void boolean()
+{
+  __CPROVER_bool a, b;
+  __CPROVER_assert(!(a ≡ b) == (a == b), "≥");
+  __CPROVER_assert(!(a ≢ b) == (a != b), "≢");
+  __CPROVER_assert(!(a ⇒ b) == (!a || b), "⇒");
```
NlightNFotis, line 6: These looked a bit suspicious to me, but it looks like the [C Compiler](https://gcc.gnu.org/onlinedocs/gcc-4.1.2/cpp/Character-sets.html) (or at least, GCC does) determines file encoding from reading the file, defaulting to either ASCII or UTF-8, and we are using `ifstream` to read files, which has its encoding information set through `std::locale`, so it looks like this should work.

NlightNFotis (approved):

chrisr-diffblue (approved):

regression/validate-trace-xml-schema/check.py
```diff
@@ -61,7 +61,10 @@
     # the SAT back-end only
     ['integer-assignments1', 'test.desc'],
     # this test is expected to abort, thus producing invalid XML
-    ['String_Abstraction17', 'test.desc']
+    ['String_Abstraction17', 'test.desc'],
+    # these tests use characters that cannot be displayed in XML
+    ['ACSL', 'operators.desc'],
```
thomasspriggs, line 66: :no_entry_sign: If the changes cause output to include more characters that are invalid XML then they should not be merged. The correct thing to do would be to escape these characters.

thomasspriggs (changes requested):

regression/validate-trace-xml-schema/check.py
thomasspriggs: For details of how escaping works in XML see - https://www.w3.org/TR/REC-xml/#syntax
I think the code for this escaping exists already. See `xmlt::escape_attribute` in `src/util/xml.cpp`.

martin-cs: :point_up: What he said.  Let's not generate invalid XML.

regression/cbmc/xml-interface1/test.desc
```diff
@@ -4,8 +4,8 @@ CORE
 ^EXIT=10$
 ^SIGNAL=0$
 Not unwinding loop foo\.0 iteration 3 file main\.c line 5 function foo thread 0
-<result property="foo\.assertion\.3" status="SUCCESS"/>
-<result property="foo\.assertion\.1" status="FAILURE">
+<result description="assertion 0" property="foo\.assertion\.3" status="SUCCESS"/>
+<result description="assertion 0" property="foo\.assertion\.1" status="FAILURE">
```
tautschnig, line 8: Nit pick: I'd prefer for this commit to be merged with prior commit that causes this test to break. Just makes any future bisecting easier.

doc/assets/xml_spec.xsd
```diff
@@ -166,6 +166,7 @@
       <xs:sequence>
         <xs:element ref="goto_trace" minOccurs="0"/>
       </xs:sequence>
+      <xs:attribute name="description" type="xs:string"/>
```
tautschnig, line 169: This one should also be merged with the code change.

regression/cbmc/ACSL/fail-operators.desc
```diff
@@ -0,0 +1,12 @@
...
+
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION FAILURE$
+--
+^VERIFICATION SUCCESSFUL$
+--
+This is to test printing of special operator characters in traces.
+The verification should fail. Note that this is a bug for normal
+and XML output, but not JSON.
```
tautschnig, line 12: It's not really clear to me what is failing here: is the actual verification outcome wrong just because of special characters?! Or is there just some issue with the output?

PR #6240, opened 2021-07-20
Add conversion of incremental smt2 data structures to string

thomasspriggs:
Add conversion of incremental smt2 data structures to string. So that they can be converted and sent to a solver. Visitors are used for this implementation, so that the conversion code can be kept separate from the data structures.

- [x] N/A None claimed - ~~My commit message includes data points confirming performance improvements (if claimed).~~

### DIFF
--- a/src/solvers/smt2_incremental/smt_to_string.cpp
+++ b/src/solvers/smt2_incremental/smt_to_string.cpp
@@ -2,10 +2,16 @@
 
 #include <solvers/smt2_incremental/smt_to_string.h>
 
+#include <solvers/smt2/smt2_conv.h>
 #include <solvers/smt2_incremental/smt_sorts.h>
+#include <solvers/smt2_incremental/smt_terms.h>
 
+#include <util/range.h>
+
+#include <functional>
 #include <iostream>
 #include <sstream>
+#include <stack>
 #include <string>
 
 class smt_sort_output_visitort : public smt_sort_const_downcast_visitort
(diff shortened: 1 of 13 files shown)


src/solvers/smt2_incremental/smt_to_string.cpp
```diff
@@ -229,3 +232,149 @@ std::string smt_to_string(const smt_termt &term)
...
+  std::ostream &os;
+
+public:
+  explicit smt_logic_to_string_convertert(std::ostream &os) : os(os)
+  {
+  }
+
+  void visit(const smt_logic_quantifier_free_bit_vectorst &) override
+  {
+    os << "QF_BV";
```
TGWDB: This is not the logic we currently use and this can cause errors later (we currently use `QF_AUFBV` explicitly, or for `z3` we don't specify and let `z3` assume `ALL`).

TGWDB (approved):

src/solvers/smt2_incremental/smt_to_string.cpp
```diff
@@ -41,3 +47,185 @@ std::string smt_to_string(const smt_sortt &sort)
...
+///   be printed.
+///   The `push_output(s)` functions exist as convenience functions to allow
+///   pushing the capturing functions to the stack in the reverse order required
+///   for printing, whilst keeping the visit functions easier to read by keeping
+///   the outputs in reading order and separating the capturing code.
+class smt_term_to_string_convertert : private smt_term_const_downcast_visitort
+{
+private:
+  using output_functiont = std::function<void(std::ostream &)>;
+  std::stack<output_functiont> output_stack;
```
NlightNFotis: I can't help but feel like the concern about call stack depth and the usage of `std::stack` is a case of a bit of premature optimisation... Do we know for sure we could *easily* exhaust the memory through the call stack? Keep in mind that a function call preamble is very lightweight (`push rbp; mov rbp, rsp;  mov DWORD PTR [rbp-4], edi` for an x64 linux system as an example).

If we could exhaust memory that way, what's stopping `std::stack` from exhausting memory? Apologies, but I couldn't find information about the underlying container it uses, but I suspect it could also be vulnerable to the same issue.

thomasspriggs: The size of the call stack has a fixed maximum. It will not grow arbitrarily large. The maximum stack size will also vary depending on which operating system the binary is compiled for. For example the size of the stack on Windows is as little as 1MB. Citation - `https://docs.microsoft.com/en-us/windows/win32/procthread/thread-stack-size` So using recursion to process data structures which may be deeply nested runs the risk of introducing out of stack space bugs which only show on certain platforms.

For a hypothetical example lets say each recursion adds 2 pointers, one for the output stream and one pointing to a term. On a 64 bit architecture those pointers would be 8 bytes each. If this is running with a stack size of 1MB = 2^20, then the maximum recursion depth available would be 2^20 / 8, which works out at 65536. Now this is a large enough number that you aren't likely to reach it with trivial examples but which could be reached with larger examples or if we happen to use an encoding which results in in deeply nested terms. The numbers chosen for this example are fairly arbitrary and are subject to change depending on issues such as -

1. Processor Architecture (for pointer size)
2. Platform (for stack size and calling conventions)
3. Implementation for functions in the algorithm.
4. The optimisations performed by the compiler.

For point 3 the number of arguments to the functions and the number of locally declared variables will affect the rate at which stack memory is used as the recursion gets deeper.

For point 4 it could be possible to write a particular algorithm in such a way that tail call optimisation can be applied and this leads to zero growth in the stack, as the recursion becomes deeper. This could then lead to the situation where the release build does not run out of stack space but the debug build does, due to not applying the optimisation! This could potentially lead to the debug builds being unusable for debugging other unrelated issues.

The unpredictability from above 4 points lead to the potential situation where you could have an example which fails, but does so in surprising ways and only in certain conditions.

The difference with `std::stack`, is that it stores the data contained therein on the heap, which can grow arbitrarily large. Technically the library implementation may have an optimisation to store small container instances on the stack and then expand into the heap as they expand. However the point is that they are only limited by the amount of available memory.

So yes, you could argue that this is a premature optimisation. I am going to argue that this is heading off a potentially difficult to resolve crash bug before it can become a problem.

NlightNFotis: Good stuff, my question was covered adequately - thanks.

src/solvers/smt2_incremental/smt_to_string.cpp
thomasspriggs: Is `QF_BV` enough for our first slice of functionality? I plan to add more different logics in a follow up PR, but I was deliberately limiting the scope of this and the preceding PR, in order to keep things more easily reviewable.

src/solvers/smt2_incremental/smt_to_string.cpp
```diff
@@ -0,0 +1,375 @@
...
+  }
+};
+
+std::ostream &operator<<(std::ostream &os, const smt_logict &logic)
+{
+  logic.accept(smt_logic_to_string_convertert{os});
+  return os;
+}
+
+std::string smt_to_string(const smt_logict &logic)
```
chrisr-diffblue: Should we make the naming here (and in the class/file name) a little more explicit? e.g. `smt_to_smt2_string` ? Otherwise its not explicitly clear what sort of 'string' is being returned.

chrisr-diffblue (approved):

src/solvers/smt2_incremental/smt_to_string.cpp
thomasspriggs: I am fine with making that change if you think it would make things clearer. I can't think of many other kinds of string which we could be converting to, but that is probably because I've been busy writing this one.

src/solvers/smt2_incremental/smt_to_string.cpp
TGWDB: We need at least `UF` for free function symbols:
```
tgw@DB0907:~$ z3 -smt2 -in
(set-logic QF_BV)
(declare-fun |f| ((_ BitVec 32) (_ BitVec 33)) (_ BitVec 31))
(error "line 2 column 60: logic does not support uninterpreted functions")
```
as above `QF_BV` alone will fail, but `QF_UFBV` below is ok:
```
tgw@DB0907:~$ z3 -smt2 -in
(set-logic QF_UFBV)
(declare-fun |f| ((_ BitVec 32) (_ BitVec 33)) (_ BitVec 31))
(define-fun |f| ((|g| (_ BitVec 32)) (|h| (_ BitVec 32))) Bool (|=| |g| |h|))
(check-sat)
sat
```

PR #6243, opened 2021-07-21
A simplified example that triggers issue 6230, from @SaswatPadhi

martin-cs:
https://github.com/diffblue/cbmc/pull/6232 got merged before I could add this test case.


SaswatPadhi (approved):
Thanks!

PR #6244, opened 2021-07-21
Mark a test that uses --smt2 as needing the smt-backend

martin-cs:
As I noted last time https://github.com/diffblue/cbmc/pull/6187#issuecomment-863213850 if one of the images that doesn't use the SMT back-end also doesn't have Z3 on the path then these bugs will manifest on PR and I won't need to keep patching them.

### DIFF
--- a/regression/contracts/Makefile
+++ b/regression/contracts/Makefile
@@ -1,4 +1,4 @@
-default: tests.log
+default: test
 
 include ../../src/config.inc
 include ../../src/common
@@ -6,23 +6,37 @@ include ../../src/common
 ifeq ($(BUILD_ENV_),MSVC)
 	exe=../../../src/goto-cc/goto-cl
 	is_windows=true
+	GCC_ONLY = -X gcc-only
 else
 	exe=../../../src/goto-cc/goto-cc
 	is_windows=false
+	GCC_ONLY =
 endif
 
 test:
-	@../test.pl -e -p -c '../chain.sh $(exe) ../../../src/goto-instrument/goto-instrument ../../../src/cbmc/cbmc $(is_windows)'
+	@../test.pl -e -p -c '../chain.sh $(exe) ../../../src/goto-instrument/goto-instrument ../../../src/cbmc/cbmc $(is_windows)' -X smt-backend $(GCC_ONLY)
 
-tests.log:
-	@../test.pl -e -p -c '../chain.sh $(exe) ../../../src/goto-instrument/goto-instrument ../../../src/cbmc/cbmc $(is_windows)'
+test-cprover-smt2:
+	@../test.pl -e -p -c '../chain.sh $(exe) ../../../src/goto-instrument/goto-instrument "../../../src/cbmc/cbmc --cprover-smt2" $(is_windows)' \
+					  -X broken-smt-backend -X thorough-smt-backend \
+					  -X broken-cprover-smt-backend -X thorough-cprover-smt-backend \
+					  -s cprover-smt2 $(GCC_ONLY)
+
+test-z3:
+	@../test.pl -e -p -c '../chain.sh $(exe) ../../../src/goto-instrument/goto-instrument "../../../src/cbmc/cbmc --z3" $(is_windows)' \
+					  -X broken-smt-backend -X thorough-smt-backend \
+					  -X broken-z3-smt-backend -X thorough-z3-smt-backend \
+					  -s z3 $(GCC_ONLY)
+
+tests.log: ../test.pl test
+
 
 clean:
 	@for dir in *; do \
 		$(RM) tests.log; \
 		if [ -d "$$dir" ]; then \
 			cd "$$dir"; \
-			$(RM) *.out *.gb; \
+			$(RM) *.out *.gb *.smt2; \
 			cd ..; \
 		fi \
 	done
--- a/regression/contracts/quantifiers-loop-02/test.desc
+++ b/regression/contracts/quantifiers-loop-02/test.desc
@@ -1,4 +1,4 @@
-CORE smt-backend
+CORE smt-backend broken-cprover-smt-backend
 main.c
 --apply-loop-contracts _ --z3
 ^EXIT=0$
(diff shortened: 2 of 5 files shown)


regression/contracts/quantifiers-loop-02/test.desc
```diff
@@ -1,6 +1,6 @@
-CORE smt-backend
+CORE smt-backend broken-cprover-smt-backend
 main.c
---apply-loop-contracts _ --z3
+--apply-loop-contracts _
```
TGWDB: I understand the tag to mark this is broken for the `cprover` backend, but removing the `--z3` tag hides that this uses functionality explicit to `z3`. I know the default at the moment is `z3`, but leaving this explicit has clarity that I'm not sure we want to remove.

martin-cs: Unless I have radically misunderstood, this should work with any SMT solver that supports quantifiers.  I will re-add --z3 but this means that the `smt-backend` tests will fail if another solver is used and `z3` is not present.

TGWDB: Not for this specific example, but a lot of the quantified output is a problem for other solvers. Typically `cvc3` and `cvc4` will error out due to quantifiers while specifying quantifier free logic, similar behaviour for `mathsat` and `boolector`. I think only `z3` can handle it out of the box from `cbmc`, while modifying the `set-logic` command allows `cvc4` to handle it (but can produce unknowns... which we were avoiding until it became clear this happens for `z3` and is now handled gracefully). Also in the current backend the output for `z3` is different to most other solvers, so turning of the `--z3` flag may reveal different `SMT` code.

martin-cs: > due to quantifiers while specifying quantifier free logic

This is the intended and correct behaviour of the solver.

> but can produce unknowns... which we were avoiding until it became clear this happens for z3 and is now handled gracefully

Again, if the logic you have requested / fragment of logic you have used is undecideable, this is intended behaviour.

regression/contracts/quantifiers-loop-02/test.desc
```diff
@@ -14,9 +14,5 @@ This test case checks the handling of a universal quantifier, with a symbolic
 upper bound, within a loop invariant.
 
 TODO: This test should:
-- be tagged `smt-backend`:
-  because the SAT backend does not support (simply ignores) `forall` in negative (e.g. assume) contexts.
-- be tagged `broken-cprover-smt-backend`:
-  because the CPROVER SMT2 solver cannot handle (errors out on) `forall` in negative (e.g. assume) contexts.
 - not use the `_ --z3` parameters:
```
SaswatPadhi, line 17: Do we still need the `_ --z3`, now that you have added Makefile rules to handle SMT-related tags?

regression/contracts/Makefile
```diff
@@ -1,28 +1,42 @@
...
+					  -X broken-cprover-smt-backend -X thorough-cprover-smt-backend \
+					  -s cprover-smt2 $(GCC_ONLY)
+
+test-z3:
+	@../test.pl -e -p -c '../chain.sh $(exe) ../../../src/goto-instrument/goto-instrument ../../../src/cbmc/cbmc --z3 $(is_windows)' \
+					  -X broken-smt-backend -X thorough-smt-backend \
+					  -X broken-z3-smt-backend -X thorough-z3-smt-backend \
+					  -s z3 $(GCC_ONLY)
+
+tests.log: ../test.pl test
```
SaswatPadhi, line 31: Do we also need to make similar changes in the CMakeLists.txt file?

regression/contracts/quantifiers-loop-02/test.desc
TGWDB: Apologies if my comment was unclear. I was not implying the behaviour was incorrect, only commenting on what we see now (and why I think being explicit on `z3` is helpful). No requested change or displeasure with the PR.

SaswatPadhi (commented):
Thanks for these changes. LGTM, but a couple of questions below:

regression/contracts/quantifiers-loop-02/test.desc
martin-cs: See @TGWDB 's request above.

regression/contracts/Makefile
martin-cs: Yes; good point.  Again @db-ci-cprover please can we have at least one build platform not have Z3 installed so these are caught.

martin-cs: Added.

martin-cs:
@SaswatPadhi @TGWDB please could you re-review?  @feliperodri please coul you have a look as a maintainer for code contracts?  I would like to get this merged ASAP.

regression/contracts/quantifiers-loop-02/test.desc
SaswatPadhi: Sorry about the late response on this, and for reopening this conversation.

I guess, until #6248 is merged, we could keep the `_ --z3` arguments, but it doesn't really matter for the CI because we don't run CVC4 anyway and so this test wouldn't really break anything (without the `_ --z3` part).
It would be nice to aim for solver-agnostic SMT encoding for as many cases as possible, and I think in this case we do get a solver-agnostic encoding that should work with both Z3 and CVC4 (modulo the `set-logic` issue). To me, `_ --z3` indicates that this test triggers some Z3-specific encoding that wouldn't work on other SMT solvers in general.

Moreover, if that's the case, we should use the `broken-<solver-X>-backend` tags to explicitly document that at least for Z3, CVC4 and CPROVER SMT2 solvers.

What do you think?

martin-cs: TBH for the purposes of this test all I want is to be able to run the regression tests on a machine that doesn't have Z3 installed.  For me anything else is a discussion for another ticket / issue.

SaswatPadhi: Approving this and unblocking you. We can revisit this issue once #6248 is merged may be.

SaswatPadhi (approved):

martin-cs:
Thanks @SaswatPadhi .  I have disabled the CMake SMT solver configurations for now as they were increasing the run time of the regressions by a lot.  As discussed above, there may be changes to how the regression tests are run for SMT solvers so I will leave enabling them to be part of that discussion.

SaswatPadhi:
Sounds good. Thanks a lot for adding the SMT tags support to regression/contracts!

PR #6196, opened 2021-06-23
Parser support for assigns contracts on loops

aalok-thakkar:
1. Created cprover_loop_contract that has cprover_contract_assigns_opt and loop_invariant_opt.

This adds support for parsing assigns clauses for loops.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [x] My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #6249, opened 2021-07-22
Add support for assigns clauses on loops

### LINKED ISSUE #6021: Crash in enforcing loop invariant contract due to imprecise alias analysis
**CBMC version:** 99f1a2e

**Operating system:** 10.15.7

**Test case:**

```c
#include <assert.h>
#include <stdlib.h>

#define SIZE 8

struct blob { char *data; };

void main()
{
  struct blob *b = malloc(sizeof(struct blob));
  b->data = malloc(SIZE);

  b->data[5] = 0;
  for (unsigned i = 0; i < SIZE; i++)
    __CPROVER_loop_invariant(i <= SIZE)
  {
    b->data[i] = 1;
  }

  assert(b->data[5] == 0);
}
```

```console
$ goto-cc test.c -o test.1.gb
$ goto-instrument --enforce-all-contracts test.1.gb test.2.gb
```

`goto-instrument` would decompose the invariant contract to appropriate assertions and assumptions.

**What happened instead:**

`goto-instrument` crashes with a precondition violation:

```console
$ goto-instrument --enforce-all-contracts test.1.gb test.2.gb
Reading GOTO program from 'test2.goto'
Invariant check failed
File: ../util/pointer_expr.h:387 function: dereference_exprt
Condition: op.type().id() == ID_pointer
Reason: Precondition
Backtrace:
...
```

**Additional Information:**

In this case `local_may_alias.get` is imprecise and returns an `unknown` expression for the pointer being modified (`b->data[i]`).

We should avoid the crash and show a warning / error regarding the imprecision of the alias analysis.

aalok-thakkar:
Resolves #6021.

In this PR, we add support for assigns clauses (`__CPROVER_assigns`) on loops.


src/goto-instrument/contracts/assigns.cpp
```diff
@@ -238,6 +238,11 @@ goto_programt assigns_clauset::havoc_code(
   return havoc_statements;
 }
 
+std::vector<assigns_clause_targett *> assigns_clauset::getTargets()
```
SaswatPadhi:
```suggestion
std::vector<assigns_clause_targett *> assigns_clauset::get_targets()
```

src/ansi-c/Search.code-search
```diff
@@ -0,0 +1,21 @@
+# Query: loop_invariant
```
martin-cs: What's This For?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -807,10 +809,29 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
 }
 
+void c_typecheck_baset::typecheck_spec_loop_assigns(codet &code)
+{
+  if(code.find(ID_C_spec_loop_assigns).is_not_nil())
+  {
+    for(const exprt &operand : static_cast<exprt &>(code.add(ID_C_spec_loop_assigns)).operands())
+    {
+      if(
+        operand.id() != ID_symbol && operand.id() != ID_ptrmember &&
```
martin-cs: These look like the restrictions on other assign clauses.  Is there any chance we could make them common.  Esp. as I recall someone saying that arrays were going to be added.

martin-cs (commented):
In principle seems fine but please look at the comments.

SaswatPadhi:
@martin-cs I have addressed some of the concerns from your first review. Please take another look if you have time and suggest any improvements / more interesting tests to try out. Thanks!!

regression/contracts/invar_assigns_alias_analysis/test.desc
```diff
@@ -1,14 +0,0 @@
...
-^SIGNAL=0$
-^\[main.1\] .* Check loop invariant before entry: SUCCESS$
-^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
-^\[main.assertion.1\] .* assertion b->data[5] == 0: FAILURE$
-^VERIFICATION FAILED$
---
---
-This test shows the need for assigns clauses. Currently we only
-parse the assigns clause for loops, but there is no implementation
-to actually enforce them. In the future, we will add this feature.
```
feliperodri, line 14: Was this an empty test case?

regression/contracts/invar_havoc_dynamic_multi-dim_array/test.desc
```diff
@@ -18,4 +18,8 @@ to pass -- we should not mistakenly havoc the allocations, just their values.
 However, the `data[1][2][3] == 0` assertion is expected to fail since `data`
 is havoced.
 
-Blocked on issue #6021 -- alias analysis is imprecise and returns `unknown`.
+The assigns clause in this case would have an entire 3D space,
+and some of it must be re-established, to the original objects
+but with different values at certain locations, by the loop invariant.
+However, currently we cannot restrict same_object for pointers
+within loop invariants.
```
feliperodri, line 26: Does this PR fixes https://github.com/diffblue/cbmc/issues/6021?

regression/contracts/invar_havoc_dynamic_multi-dim_array_all_const_idx/main.c
```diff
@@ -17,7 +17,7 @@ void main()
   data[4][5][6] = 0;
 
   for(unsigned i = 0; i < SIZE; i++)
-    __CPROVER_loop_invariant(i <= SIZE)
+    __CPROVER_assigns(i, data[4][5][6]) __CPROVER_loop_invariant(i <= SIZE)
```
feliperodri: Do we always have to specify in the assigns clause the iterators (e.g., `i`)? Wouldn't this be part of the "local variables set" of this loop? Not a blocker, but we'd deliver a better experience if users don't have to specify it.

regression/contracts/loop_assigns-03/test.desc
```diff
@@ -0,0 +1,17 @@
...
+^\[main.\d+\] .* Check that b->data\[(.*)i\] is assignable: SUCCESS$
+^VERIFICATION FAILED$
+--
+--
+This test (taken from #6021) shows the need for assigns clauses on loops.
+The alias analysis in this case returns `unknown`,
+and so we must manually annotate an assigns clause on the loop.
+
+Note that the annotated assigns clause in this case is an underapproximation,
+because `i` is also assigned in the loop and should be marked as assignable.
```
feliperodri, line 17: but again, isn't that part of the "local write set"?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -809,6 +818,41 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
 
   typecheck_spec_loop_invariant(code);
   typecheck_spec_decreases(code);
+
+  if(code.find(ID_C_spec_assigns).is_not_nil())
+    typecheck_spec_assigns(
+      static_cast<unary_exprt &>(code.add(ID_C_spec_assigns)).op().operands());
+}
```
feliperodri: Perhaps we need an example with `do { ... } while` to cover this case?

src/ansi-c/parser.y
```diff
@@ -2468,6 +2471,9 @@ iteration_statement:
           statement($$, ID_dowhile);
           parser_stack($$).add_to_operands(std::move(parser_stack($5)), std::move(parser_stack($2)));
 
+          if(!parser_stack($7).operands().empty())
+            static_cast<exprt &>(parser_stack($$).add(ID_C_spec_assigns)).operands().swap(parser_stack($7).operands());
+
```
feliperodri, line 2476: Why this is not being covered?

src/goto-instrument/contracts/utils.cpp
```diff
@@ -14,6 +14,21 @@ Date: September 2021
...
+    CONTRACT_PRAGMA_DISABLE_ASSIGNS_CHECK);
+  return instr;
+}
+
+goto_programt &add_pragma_disable_assigns_check(goto_programt &prog)
+{
+  Forall_goto_program_instructions(it, prog)
+    add_pragma_disable_assigns_check(*it);
+  return prog;
+}
```
feliperodri, line 30: Nice!

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -117,6 +117,8 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
     location));
 
   instructions.destructive_append(skip_program);
+
+  add_pragma_disable_assigns_check(instructions);
```
feliperodri: Why do we need to disable inclusion check here?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -133,34 +137,82 @@ void code_contractst::check_apply_loop_contracts(
...
+      get_modifies(local_may_alias, loop, modifies);
+    }
+    catch(const analysis_exceptiont &exc)
+    {
+      log.error() << "Failed to infer variables being modified by the loop at "
+                  << loop_head->source_location()
+                  << ".\nPlease specify an assigns clause.\nReason:"
+                  << messaget::eom;
+      throw exc;
+    }
```
feliperodri, line 178: Could the alias not throw an exception but still report the wrong write set for this loop?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -171,33 +223,54 @@ void code_contractst::check_apply_loop_contracts(
...
         .symbol_expr();
-    havoc_code.add(goto_programt::make_decl(
+    generated_code.add(goto_programt::make_decl(
       new_decreases_var, loop_head->source_location()));
     new_decreases_vars.push_back(new_decreases_var);
   }
 
-  // non-deterministically skip the loop if it is a do-while loop
+  // non-deterministically skip the loop if it is a do-while loop.
+  // TODO: Fix loop contract issues for do/while loops.
```
feliperodri: Let's create an Github issue to make sure we track this request and add the link here.
Shall we just stop applying invariants to do-while loops and report "not supported"?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -598,22 +672,54 @@ bool code_contractst::apply_function_contract(
...
+    typedef const typename natural_loops_mutablet::loopt &loopt;
+
+    targett target;
+    loopt loop;
+
+    loop_graph_nodet(targett t, loopt l) : target(t), loop(l)
+    {
+    }
+  };
+  grapht<loop_graph_nodet> loop_nesting_graph;
```
feliperodri, line 703: Why do we need to keep this struct here and not on a separate header file?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -636,7 +742,7 @@ void code_contractst::instrument_assign_statement(
     instruction_it->is_assign(),
     "The first instruction of instrument_assign_statement should always be"
     " an assignment");
-
+  add_pragma_disable_assigns_check(*instruction_it);
```
feliperodri: Could you add a comment here explaining why do we need to disable inclusion checks at this point? It might make debugging easier in the future.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -682,6 +788,7 @@ void code_contractst::instrument_call_statement(
       assigns,
       instruction_it,
       pointer_object(instruction_it->call_arguments().front()));
+    add_pragma_disable_assigns_check(*instruction_it);
```
feliperodri: Same as above.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -884,16 +999,22 @@ void code_contractst::check_frame_conditions(
...
   assigns_clauset &assigns)
 {
   goto_function_inline(goto_functions, function, ns, log.get_message_handler());
 
-  for(; instruction_it != body.instructions.end(); ++instruction_it)
+  for(; instruction_it != instruction_end; ++instruction_it)
   {
+    const auto &pragmas = instruction_it->source_location().get_pragmas();
+    if(pragmas.find(CONTRACT_PRAGMA_DISABLE_ASSIGNS_CHECK) != pragmas.end())
+      continue;
```
feliperodri, line 1019: Nice!

feliperodri (changes requested):
A few clarifications and small requests.

regression/contracts/invar_havoc_dynamic_multi-dim_array/test.desc
SaswatPadhi: Yep, it's supposed to.

regression/contracts/invar_havoc_dynamic_multi-dim_array_all_const_idx/main.c
SaswatPadhi: As of now, yes, users can either rely on automatic inference entirely or manual annotation entirely. If users manually provide an assigns clause, it has to capture (at least) everything that's assigned in the loop, just like function contracts.

In future, we could find a way to combine automatically inferred things with user annotations, both for function and loop contracts.

regression/contracts/loop_assigns-03/test.desc
SaswatPadhi: Note that the initialization of `i` is actually pulled "outside" of the loop. The loop body doesn't contain the definition of `i`, the enclosing scope does.

src/ansi-c/c_typecheck_code.cpp
SaswatPadhi: Yes, which I think requires other fixes to work correctly as well as we discussed this morning. I would leave it as future work for now and raise an exception on contract application on `do ... while` loops.

src/ansi-c/parser.y
SaswatPadhi: `do ... while` again :)

src/goto-instrument/contracts/assigns.cpp
SaswatPadhi: Because we don't need to check inclusion on definitions of CAR variables themselves,

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Nope, if the alias analysis is successful, it's supposed to compute a sound overapproximation of the loop's write set.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: That sounds like a good plan. I'll make this change. Thanks.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: It seemed too tiny and specific to this use case and not very useful generally, but I am open to moving it elsewhere.

@tautschnig, what do you suggest?

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: The idea is to not instrument the same instruction again. Do you think it's worth adding a comment on that?

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: The idea is to not instrument the same instruction again. Do you think it's worth adding a comment on that?

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: 😄

regression/contracts/invar_assigns_alias_analysis/test.desc
SaswatPadhi: GitHub's diff didn't detect this as a "rename" but shows it as a "delete" instead.

I moved it to `loop-assigns-01` (changed it from `KNOWNBUG` to `CORE`) and added variations of this test in `-02`, `-03` etc.

regression/contracts/loop_assigns-03/test.desc
remi-delmas-3000: moreover the information that `i` was declared `for( <here>; ... ; ...)` is destroyed by C-to-goto conversion

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -809,6 +818,41 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+  {
+    typecheck_expr(target);
+
+    if(target.type().id() == ID_empty)
+    {
+      error().source_location = target.source_location();
+      error() << "void-typed targets not permitted" << eom;
+      throw 0;
+    }
+    else if(target.id() == ID_pointer_object)
```
remi-delmas-3000, line 852: should we error-out on `void *` or any nested `(...((void *)*)* ...)` ?

SaswatPadhi: This case is hit when the user writes `__CPROVER_POINTER_OBJECT(ptr)`. Yes `ptr` in this case can be `void*` but should we not allow users to declare the underlying object as assignable? May be they are casting it correctly before actually assigning?

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I moved this line to the `for` loop that iterated over all instructions, instead of repeating it in all `instrument_*` functions, and added a clarifying comment.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I moved this line to the `for` loop that iterated over all instructions, instead of repeating it in all `instrument_*` functions, and added a clarifying comment.

remi-delmas-3000 (commented):
```
// 3.2. Skip storing the history if the expression is invalid
        history.add(goto_programt::make_goto(
          skip_target,
          not_exprt{all_dereferences_are_valid(parameter, ns)},
          location));
```
In this part of `contracts.cpp.L461` on snapshotting loop history variables, the history snapshot stays uninitialised, ie non-det if the ref was invalid. Is this the sound choice to ensure that any improper use of the history variable in a post condition or invariant would be detected ? Could we define it instead to call a function with the right type but would in fact `assert(false)` ?

SaswatPadhi:
> Is this the sound choice to ensure that any improper use of the history variable in a post condition or invariant would be detected ? Could we define it instead to call a function with the right type but would in fact assert(false) ?

This is to avoid spurious errors when the history variable _could_ be invalid, but appears in a properly guarded context within the invariant. Consider for example:

```c
__CPROVER_loop_invariant(
  (x != NULL) ==> (*x == __CPROVER_loop_entry(*x) + i)
)
```

In this case, we would snapshot `*x` before the loop, but it could be invalid if `x` is `NULL`. However, we don't want to `assert(false)` at the snapshot because the user has appropriately guarded it in the invariant, so would see the reported error as spurious.

Also, I think leaving the invalid history variable as nondet is sound? If the proof goes through when the history variable can be _anything_ then it'll also go through when it's _something_ specific, so in that sense I think it's sound?

regression/contracts/invar_havoc_dynamic_multi-dim_array_all_const_idx/main.c
feliperodri: For function contracts, users don't have to specify formal parameters and local variables in the assigns clauses. So, I was expecting that they wouldn't have to specify local variables for loop contracts either. Don't we automatically add all regions from `DECL` instructions on the write set? Again, this is not a blocker, but I don't think it's related to the automatic inference procedure we use here.

regression/contracts/loop_assigns-03/test.desc
feliperodri: > moreover the information that `i` was declared `for( <here>; ... ; ...)` is destroyed by C-to-goto conversion

Oh that clarifies this a bit more. So a simple loop like this
```c
  for(int i = 0; i < 2; i++)
    assert(i < 2);
```
is converted to something like this
```
        // 0 file main.c line 13 function main
        DECL main::1::1::i : signedbv[32]
        // 1 file main.c line 13 function main
        ASSIGN main::1::1::i := 0
        // 2 file main.c line 13 function main
     1: IF ¬(main::1::1::i < 2) THEN GOTO 2
        // 3 file main.c line 14 function main
        ASSERT main::1::1::i < 2 // assertion i < 2
        // 4 file main.c line 13 function main
        ASSIGN main::1::1::i := main::1::1::i + 1
        // 5 file main.c line 13 function main
        GOTO 1
        // 6 file main.c line 14 function main
     2: DEAD main::1::1::i
```

So the declaration of `i` in this context is outside of the loop body. If we have local variables in the loop body, do users must add them to the assign clause as well? It's a bit odd that we don't have access to these initial declarations since their scope is only the loop body.

src/ansi-c/c_typecheck_code.cpp
feliperodri: Perfect!

src/goto-instrument/contracts/assigns.cpp
feliperodri: I think a comment here would be helpful to smooth debugging in the future.

src/goto-instrument/contracts/contracts.cpp
feliperodri: > I moved this line to the `for` loop that iterated over all instructions, instead of repeating it in all `instrument_*` functions, and added a clarifying comment.

That's perfect! Thank you.

regression/contracts/invar_havoc_dynamic_multi-dim_array_all_const_idx/main.c
SaswatPadhi: Local variables (variables declared inside the loop / function _body_) are completely unrelated here. They are always automatically added to the write set and are not specified in the assigns clause (because they are not even visible in the enclosing scope). This is completely unrelated to loop counter declaration though. It's not declared inside the loop body.

Formal parameters are _somewhat_ related. One could see a `for` loop with `i_1`, ..., `1_n` initialization declarations as a lambda function that has `i_1`, ..., `1_n` as formal parameters and is invoked with the respective initialization values.
This could be implemented. I can grab all declarations from the initialization clause of a `for` loop and add them to the assigns clause automatically.

regression/contracts/loop_assigns-03/test.desc
SaswatPadhi: > If we have local variables in the loop body, do users must add them to the assign clause as well?

Local variables (variables declared inside the loop / function body) are completely unrelated here. Loop-local variables need not be specified in the assigns clause -- they aren't even visible at the enclosing scope.

I think the main confusion here is that `i` is "a local variable in the loop body". It's not; it's outside of the loop body.

Before C99 standard, declarations within `for` loop initialization were not allowed. So the declaration for `i` had to be necessarily made outside of the loop. Starting with C99, `i` can be declared in the initialization part of the `for` loop and has the same scope as the loop (not the loop body).

[cppreference states that](https://en.cppreference.com/w/cpp/language/for):
```c
{
  ...
  for ( init-statement ; condition ; iteration-expression ) statement
  ..
}
```
is equivalent to: (comments added by me) 
```c
{
  // This is the function-body scope
  ...
  {
    // This is NOT the loop-body scope
    init-statement
    while ( condition )
    // This is where the invariant goes
    {
      // This is the loop-body scope
      statement
      iteration-expression ;
    }
  }
  ...
} 
```

remi-delmas-3000:
> Also, I think leaving the invalid history variable as nondet is sound? If the proof goes through when the history variable can be _anything_ then it'll also go through when it's _something_ specific, so in that sense I think it's sound?

Then it would mean that the post condition is insensitive to the value so this seems OK. Resolving this

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -171,33 +223,58 @@ void code_contractst::check_apply_loop_contracts(
...
 
-  // non-deterministically skip the loop if it is a do-while loop
+  // TODO: Fix loop contract issues for do/while loops.
   if(!loop_head->is_goto())
   {
-    havoc_code.add(goto_programt::make_goto(
+    log.error() << "Loop contracts are currently unsupported on do/while loops."
+                << messaget::eom;
+    throw 0;
+
```
feliperodri: Shall we add a test case and label it as `KNOWNBUG`? It'd make codecov happy and it'd document this in our regression.

feliperodri (approved):

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -486,6 +487,10 @@ void c_typecheck_baset::typecheck_for(codet &code)
 
   typecheck_spec_loop_invariant(code);
   typecheck_spec_decreases(code);
+
+  if(code.find(ID_C_spec_assigns).is_not_nil())
+    typecheck_spec_assigns(
+      static_cast<unary_exprt &>(code.add(ID_C_spec_assigns)).op().operands());
```
tautschnig: Nit picking: braces around multi-line body.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -775,6 +780,10 @@ void c_typecheck_baset::typecheck_while(code_whilet &code)
...
```
tautschnig: Nit picking: braces around multi-line body.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -809,6 +818,41 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
```
tautschnig: Nit picking: braces around multi-line body.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -809,6 +818,41 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+void c_typecheck_baset::typecheck_spec_assigns(exprt::operandst &targets)
+{
+  for(auto &target : targets)
+  {
+    typecheck_expr(target);
+
+    if(target.type().id() == ID_empty)
+    {
+      error().source_location = target.source_location();
+      error() << "void-typed targets not permitted" << eom;
```
tautschnig: Perhaps add "in assigns clause"

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -809,6 +818,41 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+      throw 0;
+    }
+    else if(target.id() == ID_pointer_object)
+    {
+      // skip
+    }
+    else if(!target.get_bool(ID_C_lvalue))
+    {
+      error().source_location = target.source_location();
+      error() << "illegal target in assigns clause" << eom;
```
tautschnig: Perhaps say "non-lvalue target in assigns clause"

src/ansi-c/c_typecheck_base.cpp
```diff
@@ -739,33 +739,7 @@ void c_typecheck_baset::typecheck_declaration(
...
-            throw 0;
-          }
-          else if(has_subexpr(target, ID_side_effect))
-          {
-            error().source_location = target.source_location();
-            error() << "assigns clause is not side-effect free" << eom;
-            throw 0;
-          }
-        }
+        typecheck_spec_assigns(code_type.assigns());
```
tautschnig, line 742: Random place for comment: I would prefer, if at all possible, for tests to be part of this first commit. Those tests should make sure that the newly added parsing and type checking actually works. Also, this newly added support ought to be documented.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -942,8 +942,8 @@ void code_contractst::check_frame_conditions(
       instruction_it->is_other() &&
       instruction_it->get_other().get_statement() == ID_havoc_object)
     {
-      const exprt &havoc_argument = dereference_exprt(
-        to_typecast_expr(instruction_it->get_other().operands().front()).op());
+      const auto havoc_argument =
+        pointer_object(instruction_it->get_other().operands().front());
```
tautschnig, line 1078: Could the commit message please be a bit more verbose? You might know what the bug exactly is/was, but someone browsing the git history at a later time most likely won't.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -209,7 +209,10 @@ void code_contractst::check_apply_loop_contracts(
       converter.goto_convert(old_decreases_assignment, havoc_code, mode);
     }
 
-    goto_function.body.destructive_insert(std::next(loop_head), havoc_code);
+    const auto head_loc = loop_head->source_location();
+    while(loop_head->source_location() == head_loc)
+      loop_head = std::next(loop_head);
+    goto_function.body.destructive_insert(loop_head, havoc_code);
```
tautschnig: Several issues with both this change and the commit:
1) The commit message fails to answer the "why?" question.
2) There are no tests.
3) Why is the source location a sufficient criterion to select/deselect a particular instruction?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -133,34 +137,66 @@ void code_contractst::check_apply_loop_contracts(
...
   // and immediately convert it to goto instructions.
   {
     code_assertt assertion{invariant_expr()};
     assertion.add_source_location() = loop_head->source_location();
-    converter.goto_convert(assertion, havoc_code, mode);
-    havoc_code.instructions.back().source_location_nonconst().set_comment(
+    converter.goto_convert(assertion, generated_code, mode);
+    generated_code.instructions.back().source_location_nonconst().set_comment(
       "Check loop invariant before entry");
   }
```
tautschnig, line 153: Why do you need `goto_convert` here, wouldn't it be sufficient to just construct the goto program instruction in place?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -133,34 +137,66 @@ void code_contractst::check_apply_loop_contracts(
...
 
   // havoc the variables that may be modified
-  havoc_if_validt havoc_gen(modifies, ns);
-  havoc_gen.append_full_havoc_code(loop_head->source_location(), havoc_code);
+  modifiest modifies;
```
tautschnig, line 164: Is "modifiest" a good choice when we are talking about assigns?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -133,34 +137,66 @@ void code_contractst::check_apply_loop_contracts(
...
+  if(assigns.is_nil())
```
tautschnig, line 165: This method is way too big already, could the assigns-handling please be factored out into a method of its own?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -133,34 +137,66 @@ void code_contractst::check_apply_loop_contracts(
...
+                  << loop_head->source_location()
+                  << ".\nPlease specify an assigns clause.\nReason:"
+                  << messaget::eom;
+      throw exc;
+    }
+  }
+  else
+  {
+    for(const auto &target : assigns.operands())
+      modifies.insert(target);
```
tautschnig: `modifies.insert(assigns.operands().begin(), assigns.operands().end());` should work, I believe.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -133,34 +137,66 @@ void code_contractst::check_apply_loop_contracts(
...
 
   // Generate: assume(invariant) just after havocing
   // We use a block scope to create a temporary assumption,
   // and immediately convert it to goto instructions.
   {
     code_assumet assumption{invariant_expr()};
     assumption.add_source_location() = loop_head->source_location();
-    converter.goto_convert(assumption, havoc_code, mode);
+    converter.goto_convert(assumption, generated_code, mode);
   }
```
tautschnig, line 216: As above: likely doesn't actually require the use of `goto_convert`?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -171,33 +207,54 @@ void code_contractst::check_apply_loop_contracts(
...
   // Use insert_before_swap to preserve jumps to loop head.
-  insert_before_swap_and_advance(goto_function.body, loop_head, havoc_code);
+  insert_before_swap_and_advance(goto_function.body, loop_head, generated_code);
+
+  // Forward the loop_head iterator until the start of the body.
+  // This is necessary because complex C loop_head conditions are sometimes
+  // converted to multiple GOTO instructions (e.g. temporaries are introduced).
+  const auto head_loc = loop_head->source_location();
+  while(loop_head->source_location() == head_loc)
+    loop_head++;
```
tautschnig, line 263: So we had similar code in the preceding commit. Perhaps factor out this advancing-the-loop-head into a function. Also, I have no idea _why_ it is necessary to forward the `loop_head` iterator.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -171,33 +207,54 @@ void code_contractst::check_apply_loop_contracts(
...
+  // Copy the loop_head as we would increment the iterator while instrumenting.
+  if(assigns.is_not_nil())
+  {
+    auto copy_loop_head = loop_head;
+    check_frame_conditions(
+      function_name,
+      goto_function.body,
+      copy_loop_head,
+      loop_end,
+      loop_assigns);
```
tautschnig: Why are you taking a copy of an iterator? Pass the iterator by value and be happy?!

src/goto-instrument/contracts/contracts.cpp
tautschnig: It would be really helpful to have documentation that answers the "why" question of why this entire struct exists in the first place. That might help answer whether here or elsewhere is the right place.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -601,22 +656,54 @@ bool code_contractst::apply_function_contract(
...
+    for(size_t inner = 0; inner < loop_nesting_graph.size(); ++inner)
+    {
+      if(inner == outer)
+        continue;
+
+      if(loop_nesting_graph[outer].loop.contains(
+           loop_nesting_graph[inner].target))
+        loop_nesting_graph.add_edge(outer, inner);
+    }
+  }
```
tautschnig: This code would really benefit from some comments. Quadratic loops are always scary. Also, I'd at least rewrite this for the inner loop to start from `outer + 1` and test both inner-containers-outer and outer-contains-inner in the loop body. But that might be a matter of personal preference.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -887,11 +978,12 @@ void code_contractst::check_frame_conditions(
   const irep_idt &function,
   goto_programt &body,
   goto_programt::targett &instruction_it,
+  const goto_programt::targett &instruction_end,
```
tautschnig, line 1010: See earlier commit: pass this by value (there isn't much of a point of passing an iterator by reference).

src/goto-instrument/loop_utils.cpp
```diff
@@ -51,6 +51,10 @@ void get_modifies_lhs(
     for(const auto &mod : local_may_alias.get(t, ptr.pointer))
     {
       const typecast_exprt typed_mod{mod, ptr.pointer.type()};
+      if(mod.id() == ID_unknown)
+      {
+        throw analysis_exceptiont("Alias analysis returned UNKNOWN!");
+      }
```
tautschnig, line 57: Yes, so what?

regression/contracts/assigns_enforce_arrays_05/test.desc
```diff
@@ -1,6 +1,6 @@
 CORE
 main.c
---enforce-contract assigns_ptr --enforce-contract assigns_range
+--enforce-contract assigns_range
```
tautschnig, line 3: Doesn't quite match the commit message?

regression/contracts/invar_havoc_dynamic_multi-dim_array_all_const_idx/test.desc
```diff
@@ -1,12 +1,14 @@
-KNOWNBUG
+CORE
```
tautschnig, line 1: Should be merged with the commit that actually fixes the issue to enable this test.

src/goto-instrument/contracts/assigns.cpp
tautschnig: +100. I have no idea how this commit relates to the remainder of the PR. Does it perhaps touch code that was only introduced in this PR? If so, fold it into that commit. Where do I find tests? Why is this change being made at all?

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: This is a good point. Our assigns clauses don't have _modifies_ semantics. I'm going to make a separate GitHub issue to track this, because it requires changes outsides of contracts code (`loop_utils.{h,cpp}` and also `function_modifies.{h,cpp})`.

SaswatPadhi: Created #6467.

src/goto-instrument/contracts/assigns.cpp
SaswatPadhi: Added a small comment above it, but I am not sure what more I could say -- I mean this looks self-explanatory to me: we don't want to instrument our inclusion checks themselves. I am open to suggestions.

> I have no idea how this commit relates to the remainder of the PR.

The PR introduces write set checking in loop bodies. This commit ensures that each loop body is only instrumented once in presence of nested loops.

> Does it perhaps touch code that was only introduced in this PR?

Yes it improves the core checks introduced in `instrument loops per given assigns clauses` commit.

> If so, fold it into that commit.

I am not sure. This is an optimization commit. I intentionally kept it separate from the core commit that introduces checks. I thought it would make reviewing + debugging any potential issues in the future easier this way, but I could fold it into that commit if you think I should do that.

> Where do I find tests?

The test [loop-assigns-04] is included in `instrument loops per given assigns clauses` commit. This commit simply reduces the number of (repeated) checks generated in the output for that test.

I wasn't sure how to specify that in the test.desc, because we can only describe whether a regex pattern matches or not. Can we describe how many times it matches?

> Why is this change being made at all?

Inclusion checks are sloooow

src/goto-instrument/loop_utils.cpp
SaswatPadhi: So, to guarantee soundness, in this context one would have to add _every visible object_ to the `modifies` set in this case (unknown == \top).

I raise an exception because I was not sure how to even do that. This is caught by the caller and processed in whatever way appropriate in the calling context. In our case, loop contract instrumentation calls this function and shows a message to the user to add an assigns clause on the loop.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Done; thanks!

regression/contracts/invar_havoc_dynamic_multi-dim_array_all_const_idx/test.desc
SaswatPadhi: Done; thanks!

regression/contracts/assigns_enforce_arrays_05/test.desc
SaswatPadhi: This was not caught before (not sure why) but I had to fix it for the build to go through. Should I keep it in a separate commit just by itself?

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Done; thanks!

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: As discussed offline this morning, `goto_convert` is necessary here because the `assume(invariant)` statement could expand to multiple GOTO instructions.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Done; thanks!

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Done; thanks!

src/ansi-c/c_typecheck_code.cpp
SaswatPadhi: Done; thanks!

src/ansi-c/c_typecheck_code.cpp
SaswatPadhi: Done; thanks!

src/ansi-c/c_typecheck_code.cpp
SaswatPadhi: Done; thanks!

src/ansi-c/c_typecheck_code.cpp
SaswatPadhi: Done; thanks!

src/ansi-c/c_typecheck_code.cpp
SaswatPadhi: Done; thanks!

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: As discussed offline this morning, `goto_convert` is necessary here because the `assume(invariant)` statement could expand to multiple GOTO instructions.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I'm not sure ... I think it would lead to mode code duplication (of the containment-checking + edge-adding part):

```c
  for(size_t a = 0; a < loop_nesting_graph.size(); ++a)
  {
    for(size_t b = a + 1; b < loop_nesting_graph.size(); ++b)
    {
      if(loop_nesting_graph[a].loop.contains(
           loop_nesting_graph[b].target))
        loop_nesting_graph.add_edge(a, b);
      else if(loop_nesting_graph[b].loop.contains(
           loop_nesting_graph[a].target))
        loop_nesting_graph.add_edge(b, a);
    }
  }
```

In terms of number of operations it's basically the same: `n * n` vs `((n * (n-1) / 2) * 2)` == `n * (n-1)`

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I have added a regression test + more description there and also to the commit message.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I have added a regression test + more description there and also to the commit message.

src/ansi-c/c_typecheck_base.cpp
SaswatPadhi: I've added some description regarding this to the commit message now. Basically the tests (along with the syntax rules for parsing) were introduced in #6196. This PR only adds the parsing _actions_ etc.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Done; thanks!

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: This looks self-explanatory to me: we are creating a graph node that stores a loop information. Do you have any suggestions on what you'd like to see in a comment?

regression/contracts/invar_havoc_dynamic_multi-dim_array/test.desc
```diff
@@ -18,4 +18,9 @@ to pass -- we should not mistakenly havoc the allocations, just their values.
 However, the `data[1][2][3] == 0` assertion is expected to fail since `data`
 is havoced.
 
-Blocked on issue #6021 -- alias analysis is imprecise and returns `unknown`.
+Blocked on #6326:
```
feliperodri, line 21: Is this a PR or a GitHub issue? I'd just add the link here.

regression/contracts/invar_havoc_dynamic_multi-dim_array_partial_const_idx/test.desc
```diff
@@ -19,4 +19,9 @@ We _could_ do a more precise analysis in future where we only havoc
 `data[4][5][6]` but not `data[1][2][3]` since the latter clearly cannot be
 modified in the given loop.
 
-Blocked on issue #6021 -- alias analysis is imprecise and returns `unknown`.
+Blocked on #6326:
```
feliperodri, line 22: Is this a PR or a GitHub issue? I'd just add the link here.

regression/contracts/loop_contracts_do_while/test.desc
```diff
@@ -0,0 +1,9 @@
+KNOWNBUG
+main.c
+--apply-loop-contracts
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+--
+--
+This test checks that loop contracts work correctly on do/while loops.
```
feliperodri, line 9: Nice! Thanks.

regression/contracts/variant_multi_instruction_loop_head/test.desc
```diff
@@ -0,0 +1,19 @@
...
+^SIGNAL=0$
+--
+--
+This test fails even without the fix proposed in the commit, so it should be improved.
+It is expected to fail because the proposed invariant isn't strong enough to help prove
+termination using the specified variant.
+
+The test highlights a case where a C loop guard is compiled to multiple GOTO instructions.
+Therefore the loop_head pointer needs to be advanced multiple times to get to the loop body,
+where the initial value of the loop variant (/ ranking function) must be recorded.
```
feliperodri, line 15:
```suggestion
where the initial value of the loop variant (i.e., ranking function) must be recorded.
```

regression/contracts/variant_multi_instruction_loop_head/test.desc
```diff
@@ -0,0 +1,19 @@
...
+
+The proposed fix advances the pointer until the source_location differs from the original
+loop_head's source_location. However, this might still not work if the loop guard in the
+original C code was split across multiple lines in the first place.
```
feliperodri, line 19: Are we tracking this somewhere else? We should create a GitHub issue for this and add the link here.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -117,6 +117,10 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
     location));
 
   instructions.destructive_append(skip_program);
+
+  // The assignments above are only performed on locally defined temporaries
+  // and need not be checked for inclusion in the enclosing scope's write set.
+  add_pragma_disable_assigns_check(instructions);
```
feliperodri, line 123: Great!

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -598,22 +668,55 @@ bool code_contractst::apply_function_contract(
...
-  const irep_idt &function,
+  const irep_idt &function_name,
   goto_functionst::goto_functiont &goto_function)
 {
   local_may_aliast local_may_alias(goto_function);
   natural_loops_mutablet natural_loops(goto_function.body);
 
-  // Iterate over the (natural) loops in the function,
-  // and apply any invariant annotations that we find.
+  // a graph node type that stores information about a loop
```
```suggestion
  // A graph node type that stores information about a loop.
```

We could add a couple more comments here. For instance, which information are you referring to (e.g., loop body only, or an entire copy of the loop instructions, or only the set of local variables, or just loop head...)? Why do we need to use a graph to store this information and not another standard data structure instead? I see that every node contains "loop.first" and "loop.second", which I assume correspond to `loop_head` and `loop_body`, respectively - am I right? What about target?

I know that most of this information I could probably deduce from debugging the code a bit more. I also know that having these loops in topological order is the reason why we create this graph (unless I'm wrong). However, it might improve maintainability if we add a couple of more comments to make these bits of information even more obvious, specially answering questions like "_why topological order is important here?_" and set developers in the right direction next time they scan this code.

src/goto-instrument/contracts/contracts.cpp
feliperodri: I added a couple of suggestions on my previous comment.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -875,21 +981,34 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
   goto_function_inline(goto_functions, function, ns, log.get_message_handler());
 
-  for(; instruction_it != body.instructions.end(); ++instruction_it)
+  for(; instruction_it != instruction_end; ++instruction_it)
   {
+    const auto &pragmas = instruction_it->source_location().get_pragmas();
+    if(pragmas.find(CONTRACT_PRAGMA_DISABLE_ASSIGNS_CHECK) != pragmas.end())
+      continue;
+
+    // do not instrument this instruction again in the future,
```
```suggestion
    // Do not instrument this instruction again in the future,
```

feliperodri (approved):
Only minor comments about documentation. Thanks for implementing all the comments!

regression/contracts/variant_multi_instruction_loop_head/test.desc
SaswatPadhi: Good point. I will create a GitHub issue. I mentioned this to Michael just this morning.

regression/contracts/loop_assigns-03/test.desc
feliperodri: Thank you so much for adding a bit more context here! If there is a "loop scope" that encapsulates the entire loop and loop body, why are we focusing only in the loop body and not the most general scope? The only goal I'm trying to active with this discussion is: decrease the burden of annotation for users (this is not a blocker for this PR).

feliperodri: Your previous comment respond my previous question: https://github.com/diffblue/cbmc/pull/6249#discussion_r749646842. If you could grab all declarations from the initialization clause of a `for` loop and add them to the assigns clause automatically, this would decrease the annotation burden to developers and it's perfectly sound for our approach. Thanks!

src/ansi-c/c_typecheck_base.cpp
tautschnig: Is there anyway we can test that this was successful?

src/goto-instrument/contracts/contracts.cpp
tautschnig: Not quite?

tautschnig (approved):

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Do you still see it in the diff? I don't ...

regression/contracts/loop_assigns-03/test.desc
SaswatPadhi: Yes exactly. I'll try to push that change in this PR if it works (still playing with it), or would make a separate PR for it later.

src/goto-instrument/contracts/contracts.cpp
tautschnig: Ah, sorry, that was just because I reviewed commit-by-commit. The final commit amends this part.

regression/contracts/invar_havoc_dynamic_multi-dim_array/test.desc
SaswatPadhi: GitHub numbers both issues and PRs on the same number line. They also have a uniform URL for both: `<repo>/issues/<id>` and `<repo>/pull/<id>` are interchangeable.

So #6326 is a PR, but the URL [`https://github.com/diffblue/cbmc/issues/6326`](https://github.com/diffblue/cbmc/issues/6326) would work just fine.

regression/contracts/invar_havoc_dynamic_multi-dim_array_partial_const_idx/test.desc
SaswatPadhi: GitHub numbers both issues and PRs on the same number line. They also have a uniform URL for both: `<repo>/issues/<id>` and `<repo>/pull/<id>` are interchangeable.

So #6326 is a PR, but the URL [`https://github.com/diffblue/cbmc/issues/6326`](https://github.com/diffblue/cbmc/issues/6326) would work just fine.

src/ansi-c/c_typecheck_base.cpp
SaswatPadhi: Yes there are lots of function contracts regression tests that check each branch of the `if/else` structure. They still pass. I just refactored this into a common function that can be used for loop contracts too.

feliperodri:
@chrisr-diffblue @martin-cs @peterschrammel could you take a look at this PR? It's missing code owner approval.

regression/contracts/assigns_enforce_20/test.desc
```diff
@@ -3,7 +3,7 @@ main.c
 --enforce-contract foo
 ^EXIT=10$
 ^SIGNAL=0$
-^\[foo.\d+\] line \d+ Check that \*y is assignable: FAILURE$
+^\[foo.\d+\] line \d+ Check that POINTER_OBJECT\((\(.+\))?y\) is assignable: FAILURE$
```
martin-cs, line 6: Thanks for fixing this.  This is not a blocker for this PR but in general it is good to have a regression test that shows the difference that is created by the bug fix.

peterschrammel (approved):
lgtm

regression/contracts/variant_multi_instruction_loop_head/test.desc
```diff
@@ -0,0 +1,19 @@
+CORE
+main.c
+--apply-loop-contracts
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+This test fails even without the fix proposed in the commit, so it should be improved.
```
martin-cs, line 9: What is the observable difference in this test of the fix?  If none then ... can we have a test case that does have an observable difference when fixed?

Presumably it will need a short-cut operator with one operation that alters a value?  What about a fancy `for-else` loop like this:

```
int i = 0;
int i_valid_element = 1;

while (i < limit || i_valid_element = 0) {
  if (a[i] == target) {
    break;
  }
  ++i;
}

if (!i_valid_element) {
  realloc(a, ... );
  a[i] = target;
}

return i;
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -209,7 +209,13 @@ void code_contractst::check_apply_loop_contracts(
       converter.goto_convert(old_decreases_assignment, havoc_code, mode);
     }
 
-    goto_function.body.destructive_insert(std::next(loop_head), havoc_code);
+    // Forward the loop_head iterator until the start of the body.
+    // This is necessary because complex C loop_head conditions could be
+    // converted to multiple GOTO instructions (e.g. temporaries are introduced).
+    const auto head_loc = loop_head->source_location();
+    while(loop_head->source_location() == head_loc)
```
martin-cs: A. Frankly, this makes me nervous.  I don't think we have any (documented, checked, folklore, etc.) invariants about `source_location` and making it semantically meaningful.  There are a few places that do, things like the coverage code (I think @peterschrammel is the expert here) but ... I am concerned about this.

B. I can't think of a better way of distinguishing:

```
while (a && b) {
...
}
```

from

```
while (a) {
 if (! b) break;
}
```

I guess they produce the same `goto_programt`?

C. Do we need some common way of marking where the loop head / body is?  Should this code go into the `natural_loops` detection code?  Do we need a marker instruction?  When working on other analysis tools the question of "what is 'in a loop'" has come up and it is a lot more complex than it first appears.

regression/contracts/assigns_enforce_arrays_05/test.desc
martin-cs: I am less worried about where it is but some kind of comment about why you are removing `--enforce-contract assigns_ptr` would be good because I have been trying to figure it out.

regression/contracts/loop_assigns-03/test.desc
```diff
@@ -0,0 +1,17 @@
...
+^\[main.\d+\] .* Check that i is assignable: FAILURE$
+^\[main.\d+\] .* Check that b->data\[(.*)i\] is assignable: SUCCESS$
+^VERIFICATION FAILED$
+--
+--
+This test (taken from #6021) shows the need for assigns clauses on loops.
+The alias analysis in this case returns `unknown`,
+and so we must manually annotate an assigns clause on the loop.
+
+Note that the annotated assigns clause in this case is an underapproximation,
```
martin-cs, line 16: What is the consequence of this?  Does it produce unsound results or simply inefficient ones?

regression/contracts/loop_assigns-04/main.c
```diff
@@ -0,0 +1,41 @@
...
+    __CPROVER_assigns(i, __CPROVER_POINTER_OBJECT(b->data))
+    __CPROVER_loop_invariant(i <= SIZE)
+    // clang-format on
+    {
+      b->data[i] = 1;
+
+      int x;
+      for(unsigned j = 0; j < i; j++)
+        // clang-format off
+        // y is not assignable by outer loop, so this should be flagged
```
martin-cs, line 29: It is good you checked for this.

src/goto-instrument/contracts/contracts.cpp
martin-cs: +1 what he said.

regression/contracts/loop_contracts_do_while/test.desc
martin-cs, line 9: I am slightly confused about this commit.  You have a test to check that things work but the code that is added is error logging.  So... does this example work, but not exercise the code you added?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -236,10 +236,14 @@ void code_contractst::check_apply_loop_contracts(
     new_decreases_vars.push_back(new_decreases_var);
   }
 
-  // non-deterministically skip the loop if it is a do-while loop.
   // TODO: Fix loop contract issues for do/while loops.
   if(!loop_head->is_goto())
   {
+    log.error() << "Loop contracts are unsupported on do/while loops: "
+                << loop_head->source_location() << messaget::eom;
+    throw 0;
```
martin-cs, line 244: Please can you throw something, anything really, rather than a `0`?  I know other bits of code do it; they really shouldn't.  A string would be fine.

martin-cs (approved):
OK; I am really torn on this.  I don't want to hold up @SaswatPadhi or the good work done, so I have put approve BUT I think there are a number of things here that need some thought and discussion.  I have put these as comments.  I think the actual change is OK, it is mainly peripheral issues and bug fixes.

regression/contracts/assigns_enforce_20/test.desc
SaswatPadhi: Thanks!

src/goto-instrument/loop_utils.cpp
jimgrundy: I guess it's not strictly speaking necessary for soundness. You could chose the empty set or some random junk. The abstraction of the loop would be unsound, but the step-case check would fail because we wouldn't be able to prove that you only wrote what we said.

Having said that though, surely this would be easier for the user to debug then having them try to debug the fail of the proof that the loop didn't write to the empty set of targets.

regression/contracts/loop_assigns-03/test.desc
SaswatPadhi: Previously CBMC used to crash (see: #6021) because it couldn't resolve the `unknown` to an underlying object.

After merging this PR, the user would receive an error message suggesting them to add an assigns clause on the loop.

regression/contracts/loop_assigns-04/main.c
SaswatPadhi: Thanks!

regression/contracts/loop_contracts_do_while/test.desc
jimgrundy: do/while doesn't work yet

SaswatPadhi: So this is marked as `KNOWNBUG` and it's not even run on the CI yet.

May be I should change it to `CORE` and check for the expected error message on the output.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I agree that this is ugly, but (to avoid too much noise in this PR) I think should we make a global pass and remove it from all of the contracts code (and many other modules that `throw 0`) in its own PR. What do you think?

jimgrundy: If @martin-cs agrees, please file an issue to track this.

martin-cs: I will accept "that is the standard for this area of the code" and a good faith 'at some point' promise to do something about it.

I ... would be a little more reluctant to do it more widely because not all of them should be `throw`s; some should be `DATA_INVARIANTS`, some should be `CHECK_RETURN`, etc.

( For those who might be following who are new to this -- `throw` should only be used for things that are possible with incorrect user input, if it isn't possible (well, shouldn't be...) then it should be one of the `INVARIANTS` depending on why it shouldn't happen. )

regression/contracts/variant_multi_instruction_loop_head/test.desc
SaswatPadhi: > What is the observable difference in this test of the fix?

The difference is that the old value of the loop variant was recorded in the middle of loop head instructions. Simply doing `std::next(loop_head)` isn't sufficient to get to the body, where it should be recorded.

The `--show-goto-functions` on the test would show the difference, but there is no observation failure in the CBMC output. I could try to come up with a test where it would lead to a failure, though.

regression/contracts/loop_contracts_do_while/test.desc
martin-cs: Ah; forgive me, I missed that.  Yeah, `KNOWNBUG` is sufficient.  `CORE` with the error message is the extra classy option!

regression/contracts/loop_assigns-03/test.desc
martin-cs: If the error message is the "Check that i is assignable" VC failure then, OK, that is fine.  If there is something else then that should be checked for.

SaswatPadhi: Oops. I'm sorry, I was thinking of something else (not providing the assigns clause at all), which is unrelated to your question.

Yes, the expected error here is `Check that i is assignable: FAILURE`.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: In this commit I don't treat this an internal invariant failure (like violation of some precondition for a CBMC routine), but treat it as a user-side error -- the user trying to use an unsupported feature.

But I agree that this could be triggered either way and would be happy to change this to an invariant failure. What do you suggest?

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Much of the assigns clause checking is in a separate function (`check_frame_conditions`). In this function we call:

```c
  if(assigns.is_not_nil())
  {
    check_frame_conditions(
      function_name, goto_function.body, loop_head, loop_end, loop_assigns);
  }
```

and we have a few more lines to create snapshots of what's assignable:
```c
    for(const auto &car : loop_assigns.get_write_set())
    {
      auto snapshot_instructions = car.generate_snapshot_instructions();
      insert_before_swap_and_advance(
        goto_function.body, loop_head, snapshot_instructions);
    };
```

So these parts are not too big by themselves (and appear at different points within this function), but I agree that this function is growing to be too large. I would probably do this in a separate refactoring PR and pull out loop contracts to a separate `loop_contracts.cpp` file.

regression/contracts/variant_multi_instruction_loop_head/test.desc
martin-cs: If you have the time / energy it would be nice.  Adding test cases that don't have an observable difference is always a little iffy.  I appreciate that the test case you add would cover the source lines though.

It is pretty synthetic but...

```
uint64_t cookie = 0;
int connection_working = 0;
while (!connection_working && (cookie != 0 || cookie = get_new_cookie())) {
  connection_working = connect(cookie);
}
```

`get_new_cookie()` has a post condition that `cookie` is valid (i.e. non-zero) and `connect()` has a pre condition that `cookie` is valid as well.  It is reasonable for the user to think that cookie is guaranteed to be valid and invariant in the loop body.

src/goto-instrument/contracts/contracts.cpp
martin-cs: @jimgrundy An issue would even better than a good faith promise!

src/goto-instrument/loop_utils.cpp
SaswatPadhi: > I guess it's not strictly speaking necessary for soundness. You could chose the empty set or some random junk. 

I think that would be unsound. Because an empty set would be an underapproximation and would lead to proving incorrect properties (that should actually fail). Consider for example:

```c
a[5] = 0;
for(int i = 0; i < sz; ++i) a[i] = 1; // sz > 5
assert(a[5] == 0);
```

Suppose the alias analysis couldn't compute the write set of the loop, i.e, it returned `unknown` on the LHS `a[i]` for the statement `a[i] = 1` in the loop body.

If we consider the write set of this loop to be the empty set, we would be able to prove the postcondition, but we actually shouldn't.

regression/contracts/loop_assigns-03/test.desc
SaswatPadhi: Also, it wouldn't lead to unsound results, but raise a verification failure.

src/goto-instrument/loop_utils.cpp
jimgrundy: But, don't we then instrument the body with that (empty) write set and fail the step-case of the proof?

SaswatPadhi: Ah I see the confusion now.
We only instrument the body if an assigns clause was explicitly specified by the user. We don't resinstrument the body if the write set was automatically inferred from it.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Opened #6468 to track this.

src/goto-instrument/loop_utils.cpp
jimgrundy: Ah I see. So that write set inference is trusted code.

SaswatPadhi: Yes exactly, all internal CBMC analyses (in this case, the alias analysis) are trusted.

src/goto-instrument/contracts/contracts.cpp
tautschnig: Sharing some background discussion in here: I have the very same concern, and hence asked for some test to be added (the one you were wondering about) just so that at least we can investigate afterwards. We will likely have to rewrite some of this.

regression/contracts/loop_assigns-03/test.desc
martin-cs: Thanks for your thoughts; that is fine.

src/goto-instrument/contracts/contracts.cpp
martin-cs: Having one consistently used interface for identifying loops, the exit conditions, the exit locations, the "body", etc. would be great.  Note that the syntactic and semantic definitions of loops are different.

I am happy for this to be flagged as a future issue as I think the rest of the code base is not exactly consistent on this.

src/goto-instrument/contracts/contracts.cpp
martin-cs: I would be fine with a separate refactoring commit / PR / issue.

regression/contracts/assigns_enforce_arrays_05/test.desc
SaswatPadhi: This should be implementation as a check in the function contracts code -- basically CBMC should only enforce a single function at a time. We could enforce multiple functions, but we would have to havoc the global state and enforce each function under a "clean" (nondet) environment. 

We had fixed a bunch of these cases earlier (in other commits) but this one escaped somehow. We'll create an internal issue to check the number of functions being enforced at the same time.

PR #6251, opened 2021-07-24
goto_programt::output_instruction now uses format(...)

kroening:
Goto programs are intended as an internal representation that is independent
of the specific input programming language.  This commit switches the debug
output for the instructions of goto programs to `format(...)`, which is also
designed to be language independent.

### DIFF
--- a/src/goto-programs/goto_program.cpp
+++ b/src/goto-programs/goto_program.cpp
@@ -16,6 +16,8 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/base_type.h>
 #include <util/expr_iterator.h>
 #include <util/find_symbols.h>
+#include <util/format_expr.h>
+#include <util/format_type.h>
 #include <util/invariant.h>
 #include <util/pointer_expr.h>
 #include <util/std_expr.h>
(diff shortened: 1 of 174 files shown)


src/goto-programs/goto_program.cpp
```diff
@@ -103,22 +104,55 @@ std::ostream &goto_programt::output_instruction(
...
+    break;
+
   case DEAD:
+    out << "DEAD " << format(instruction.dead_symbol()) << '\n';
+    break;
+
   case FUNCTION_CALL:
+    out << "CALL ";
+    {
+      auto &call = instruction.get_function_call();
```
peterschrammel, line 134: Is this the only place where we print `codet`? It might be worth moving the formatting into a `util/format_code.h` (not in this PR) otherwise.

peterschrammel (approved):

src/goto-programs/goto_program.cpp
kroening: The idea is to get rid of the use of `codet` in goto functions altogether; function calls are in fact the only instruction that doesn't have an alternative API yet.

PR #6253, opened 2021-07-26
Add conversion for rol and ror operators to c output

### LINKED ISSUE #6241: `goto-instrument --dump-c` does not propery generate for `rol` or `ror` operators.
When working on a `--dump-c` for the Rust Model Checker, we are running into issues with `rol` and `ror` not being generated properly into C code.

CBMC version: 5.34.0 (cbmc-5.32.1-183-gc50985444)
Operating system: macOS Big Sur v11.4

Exact command line resulting in the issue: `symtab2gb test.json --out test.goto; goto-instrument --dump-c test.goto > test.c`
with [test.json](https://github.com/diffblue/cbmc/files/6858322/test.json.txt)
produces [test.c](https://github.com/diffblue/cbmc/files/6858390/test.c.txt)

What behaviour did you expect: The contents of `test.c` is valid C code.
What happened instead: The `rol` operator translates into an `irep` function call:
`unsigned char rol8=irep("(\"rol\" \"\" (\"constant\" \"type\" (\"unsignedbv\" \"width\" (\"8\")) \"value\" (\"38\")) \"\" (\"constant\" \"type\" (\"unsignedbv\" \"width\" (\"8\")) \"value\" (\"3\")) \"type\" (\"unsignedbv\" \"width\" (\"8\")))");`

TGWDB:
This adds conversion for the rol and ror operators from CBMC
internals into C code. This is done by bit-twiddling the values
to achieve the outcome.

Fixes #6241


src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,49 @@ std::string expr2ct::convert_with_precedence(
   else if(src.id()==ID_type)
     return convert(src.type());
 
+  else if(src.id() == ID_rol)
+  {
+    // AAAA rol n == (AAAA << n) | (AAAA >> (width(AAAA) - n))
```
martin-cs: What if `n` is greater than bit width?

martin-cs: Also note the cases when shift can give undefined behaviour.

TGWDB: This is done with modulo now.

src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,77 @@ std::string expr2ct::convert_with_precedence(
   else if(src.id()==ID_type)
     return convert(src.type());
 
+  else if(src.id() == ID_rol)
+  {
+    // AAAA rol n ==   (AAAA << n % width(AAAA))
```
thomasspriggs: :no_entry_sign: The majority of the other conversions in `convert_with_precedence` are either one liners, or separated out into other functions. This new code should be factored out in the same way. The new function will need to be a member function of the `expr2ct` class (see `src/ansi-c/expr2c_class.h`) so that it can include the recursive call to `convert_with_precedence`. The `convert_with_precedence` function is a bit long as it is, lets not add to the problem.

src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,77 @@ std::string expr2ct::convert_with_precedence(
...
+    auto rhs_expr = ashr_exprt(op0, complement_expr);
+    // AAAA << (n % width(AAAA))
+    auto lhs_exprt = shl_exprt(op0, op1);
+    // The whole thing
+    auto the_expr = bitor_exprt(lhs_exprt, rhs_expr);
+    return convert_with_precedence(the_expr, precedence);
+  }
+  else if(src.id() == ID_ror)
+  {
+    // AAAA rol n ==   (AAAA >> n % width(AAAA))
```
thomasspriggs: :no_entry_sign: This duplicated code will make maintenance more difficult. I suggest deduplicating it by parameterising the new function I suggested in my previous comment.
:bulb: I suggest creating a single use `left_or_rightt` enum class with left and right values to use as the type of the `left_or_right` parameter of the new function. This could alternatively be done without using a new enum with something like `bool is_right`, with true = right and false = left. But I think the enum would make for more self explanatory code.

src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,77 @@ std::string expr2ct::convert_with_precedence(
...
+    //               | (AAAA >> (width(AAAA) - n % width(AAAA)))
+    auto bin_expr = to_binary_expr(src);
+    // Get AAAA and if it's signed wrap it in a cast to remove
+    // the sign since this messes with C shifts
+    exprt &op0 = bin_expr.op0();
+    if(can_cast_type<signedbv_typet>(op0.type()))
+    {
+      // Shifts in C are arithmetic and care about sign, by forcing
+      // a cast to unsigned we force the shifts to perform rol/r
+      size_t ws = op0.type().get_size_t("width");
```
thomasspriggs: :pick: This and all other variables in this new code which are not mutated should be declared `const`.

src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,77 @@ std::string expr2ct::convert_with_precedence(
   else if(src.id()==ID_type)
     return convert(src.type());
 
+  else if(src.id() == ID_rol)
+  {
+    // AAAA rol n ==   (AAAA << n % width(AAAA))
+    //               | (AAAA >> (width(AAAA) - n % width(AAAA)))
+    auto bin_expr = to_binary_expr(src);
```
thomasspriggs: :pick: I suggest renaming `bin_expr` to `rotate_expr`, because at this point `bin_expr` can only be a rotate expression, even if the type only specifies that it is a `binary_exprt`.

src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,77 @@ std::string expr2ct::convert_with_precedence(
...
+    if(can_cast_type<signedbv_typet>(op0.type()))
+    {
+      // Shifts in C are arithmetic and care about sign, by forcing
+      // a cast to unsigned we force the shifts to perform rol/r
+      size_t ws = op0.type().get_size_t("width");
+      op0 = typecast_exprt(op0, unsignedbv_typet(ws));
+    }
+    // Get n because we need its type to construct the "width(AAAA)"
+    // value to use in operations below.
+    exprt &op1 = bin_expr.op1();
```
thomasspriggs: :pick: I suggest renaming `op1` to `distance`.

thomasspriggs: :pick: Use `.distance()` of `shift_exprt` instead of `.op1()`, because this is the specifically named getter for this data member.

src/ansi-c/expr2c.cpp
thomasspriggs: :pick: I suggest swapping out `to_binary_expr(src)` with `expr_checked_cast<shift_exprt>(src)`. That way we are using a more specific type of expression and we are using the checks which are defined alongside that expression.

src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,77 @@ std::string expr2ct::convert_with_precedence(
...
     return convert(src.type());
 
+  else if(src.id() == ID_rol)
+  {
+    // AAAA rol n ==   (AAAA << n % width(AAAA))
+    //               | (AAAA >> (width(AAAA) - n % width(AAAA)))
+    auto bin_expr = to_binary_expr(src);
+    // Get AAAA and if it's signed wrap it in a cast to remove
+    // the sign since this messes with C shifts
+    exprt &op0 = bin_expr.op0();
```
thomasspriggs: :pick: Use `.op()` of `shift_exprt` rather than `op0()`, because this is the specifically named getter for this data member.

src/ansi-c/expr2c.cpp
thomasspriggs: :no_entry_sign: Use `bitvector_typet::get_width()` instead of `get_size_t("width")`. By using `get_size_t` you are directly accessing the underlying data structure of the type. Fixing the accessibility to underlying `irept` would be a non-trivial refactor, but we should still use the well defined external interface where possible.

thomasspriggs: :pick: What is `ws` supposed to be short for? Reminder - the coding standards state that we should "Avoid abbreviations".

src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,77 @@ std::string expr2ct::convert_with_precedence(
...
+      // a cast to unsigned we force the shifts to perform rol/r
+      size_t ws = op0.type().get_size_t("width");
+      op0 = typecast_exprt(op0, unsignedbv_typet(ws));
+    }
+    // Get n because we need its type to construct the "width(AAAA)"
+    // value to use in operations below.
+    exprt &op1 = bin_expr.op1();
+    // Construct the "width(AAAA)" constant
+    irep_idt width_val = op0.type().get("width");
+    auto width = constant_exprt(width_val, op1.type());
```
thomasspriggs: :no_entry_sign: `from_integer` should be used to produce the `constant_exprt`, rather than using the constructor of `constant_exprt` directly. This is because the string based representation is not guaranteed to be stable. I believe we have made changes such as changing the base used for this representation before now. `from_integer` can take the `size_t` extracted via the `get_width` call, along with the same `typet`, which you are currently using.

src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,77 @@ std::string expr2ct::convert_with_precedence(
...
+      // Shifts in C are arithmetic and care about sign, by forcing
+      // a cast to unsigned we force the shifts to perform rol/r
+      size_t ws = op0.type().get_size_t("width");
+      op0 = typecast_exprt(op0, unsignedbv_typet(ws));
+    }
+    // Get n because we need its type to construct the "width(AAAA)"
+    // value to use in operations below.
+    exprt &op1 = bin_expr.op1();
+    // Construct the "width(AAAA)" constant
+    irep_idt width_val = op0.type().get("width");
```
thomasspriggs: :pick: Duplicated extraction of getting the width from `op0`. It should be possible to move the previous getting of the width out of the `if` statement and reuse it. We can expect the type to be a `bitvector_typet` whether or not it is signed. So extracting the width can be - `const size_t width = type_checked_cast<bitvector_typet>(op0.type()).get_width();`

src/ansi-c/expr2c.cpp
thomasspriggs: :pick: I suggest renaming this `width` to `width_expression`, in order to better differentiate it from the value of the expression.

src/ansi-c/expr2c.cpp
```diff
@@ -3794,6 +3794,77 @@ std::string expr2ct::convert_with_precedence(
...
+    auto width = constant_exprt(width_val, op1.type());
+    // Apply modulo to n since shifting will overflow
+    // That is: 0001 << 4 == 0, but 0001 rol 4 == 0001
+    op1 = mod_exprt(op1, width);
```
thomasspriggs: :pick: I suggest creating a new variable named `distance_modulo_width` instead, in order to better specify what this is and in order to avoid mutating.

thomasspriggs (changes requested):
It is good that we have a fix for the issue and that we have both regression and unit tests. However I think there are some issues which should be addressed for the sake of maintainability. The issues which I think really need to be addressed before we merge are marked with :no_entry_sign: Hopefully having the tests should make it easy to check the refactorings are sound as you address the issues.

src/ansi-c/expr2c.cpp
```diff
@@ -2290,6 +2290,71 @@ std::string expr2ct::convert_initializer_list(const exprt &src)
...
+  // Construct the "width(AAAA)" constant
+  const exprt width_expr =
+    from_integer(type_width, rotate_expr.distance().type());
+  // Apply modulo to n since shifting will overflow
+  // That is: 0001 << 4 == 0, but 0001 rol 4 == 0001
+  const exprt distance_modulo_width =
+    mod_exprt(rotate_expr.distance(), width_expr);
+  // Now put the pieces together
+  // width(AAAA) - (n % width(AAAA))
+  const auto complement_expr = minus_exprt(width_expr, distance_modulo_width);
```
chrisr-diffblue: I don't think this is actually the _complement expression_, is the complement _width_ expression ?

unit/ansi-c/expr2c.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <util/arith_tools.h>
+#include <util/bitvector_expr.h>
+#include <util/bitvector_types.h>
+#include <util/config.h>
+#include <util/namespace.h>
+#include <util/symbol_table.h>
+
+TEST_CASE("rol_2_c_conversion_unsigned", "[core][ansi-c][expr2c]")
+{
+  auto lhs = from_integer(31, unsignedbv_typet(32));
```
chrisr-diffblue, line 20: Personal preference here :-) But for tests used for bit-twiddling operations, I'd be rather tempted to use hex representations of the various immediate constants in tests like this. Might just be my background, but I find that easier to mentally model, rather than also adding a decimal->hex->binary mental translation :-)

unit/ansi-c/expr2c.cpp
```diff
@@ -0,0 +1,66 @@
...
+    expr2c(rol, namespacet{symbol_tablet{}}) ==
+    "31 << 3 % 32 | 31 >> 32 - 3 % 32");
+}
+
+TEST_CASE(
+  "rol_2_c_conversion_signed",
+  "[core][ansi-c][expr2c][convert_with_precedence]")
+{
+  config.ansi_c.mode = configt::ansi_ct::flavourt::GCC;
+  config.ansi_c.set_arch_spec_i386();
```
chrisr-diffblue, line 35: Might be worth a comment explaining why this is necessary?

chrisr-diffblue (approved):
Mostly looks OK, though I think @thomasspriggs has some blocking review comments that it would be good to resolve. Also, how cross-platform/cross-architecture/cross endian is this work? In particular, are we confident this works on, say ARM architecture where its switchable big/little endian? I _think_ cbmc has `--little-endian` and `--big-endian` options.

unit/ansi-c/expr2c.cpp
TGWDB: Done

src/ansi-c/expr2c.cpp
TGWDB: Fixed

src/ansi-c/expr2c.cpp
```diff
@@ -2290,6 +2290,72 @@ std::string expr2ct::convert_initializer_list(const exprt &src)
   return dest;
 }
 
+std::string expr2ct::convert_rox(const exprt &src, unsigned precedence)
+{
+  // AAAA rox n ==   (AAAA lhs_op n % width(AAAA))
+  //               | (AAAA rhs_op (width(AAAA) - n % width(AAAA)))
+  // Where lhs_op and rhs_op depend on whether it is rol or ror
+  // auto rotate_expr = to_binary_expr(src);
+  auto rotate_expr = expr_checked_cast<shift_exprt>(src);
```
tautschnig: In keeping with other functions I'd prefer for `convert_rox` to take a `shift_exprt` as first argument and `to_shift_expr` being used at the call site.

src/ansi-c/expr2c.cpp
```diff
@@ -2290,6 +2290,72 @@ std::string expr2ct::convert_initializer_list(const exprt &src)
...
+  else if(can_cast_type<unsignedbv_typet>(op0.type()))
+  {
+    // Set the type width
+    type_width = to_unsignedbv_type(op0.type()).get_width();
+  }
+  else
+  {
+    // Type that can't be represented as bitvector
+    // get some kind of width in a potentially unsafe way
+    type_width = op0.type().get_size_t("width");
```
tautschnig: I think you should just give up in that `else` case. You might well get back a `nil` irep, which will make the `from_integer` fail. You're trying to be very kind, but I don't think you're doing the user any favour.

tautschnig (approved):

src/ansi-c/expr2c.cpp
thomasspriggs: Agreed. I'd just put an `UNREACHABLE` here. I think the comments from codecov suggest that it is in fact unreachable with the way it is currently used.

src/ansi-c/expr2c.cpp
thomasspriggs: I now consider this to be sorted.

src/ansi-c/expr2c.cpp
thomasspriggs: I now consider this to be sorted.

src/ansi-c/expr2c.cpp
thomasspriggs: This looks to be sorted, with the exception of the unreachable code.

src/ansi-c/expr2c.cpp
thomasspriggs: This looks to be sorted.

thomasspriggs (approved):
My blocking comments have been addressed. Therefore I am going to approve this PR.

src/ansi-c/expr2c.cpp
```diff
@@ -2329,23 +2329,24 @@ std::string expr2ct::convert_rox(const exprt &src, unsigned precedence)
...
+    rhs_expr = ashr_exprt(op0, complement_width_expr);
   }
   else if(src.id() == ID_ror)
   {
     // AAAA >> (n % width(AAAA))
     lhs_expr = ashr_exprt(op0, distance_modulo_width);
-    // AAAA << complement_expr
-    rhs_expr = shl_exprt(op0, complement_expr);
+    // AAAA << complement_width_expr
+    rhs_expr = shl_exprt(op0, complement_width_expr);
```
thomasspriggs, line 2343: :pick: These changes appear to have been squashed into the wrong commit.

TGWDB: Fixed.

PR #6256, opened 2021-07-28
Add incremental SMT2 sort checked construction of bit vector predicate application and application of core theory functions

thomasspriggs:
Add incremental SMT2 sort checked construction of bit vector predicate application and application of core theory functions.

### DIFF
--- a/src/solvers/smt2_incremental/smt_commands.cpp
+++ b/src/solvers/smt2_incremental/smt_commands.cpp
@@ -198,3 +198,51 @@ void smt_commandt::accept(smt_command_const_downcast_visitort &&visitor) const
 {
   ::accept(*this, id(), std::move(visitor));
 }
+
+smt_command_functiont::smt_command_functiont(
+  const smt_declare_function_commandt &function_declaration)
+  : _identifier(function_declaration.identifier())
+{
+  const auto sort_references = function_declaration.parameter_sorts();
+  parameter_sorts =
+    make_range(sort_references).collect<decltype(parameter_sorts)>();
+}
+
+smt_command_functiont::smt_command_functiont(
+  const smt_define_function_commandt &function_definition)
+  : _identifier{function_definition.identifier()}
+{
+  const auto parameters = function_definition.parameters();
+  parameter_sorts =
+    make_range(parameters)
+      .map([](const smt_termt &term) { return term.get_sort(); })
+      .collect<decltype(parameter_sorts)>();
+}
+
+irep_idt smt_command_functiont::identifier() const
+{
+  return _identifier.identifier();
+}
+
+smt_sortt smt_command_functiont::return_sort(
+  const std::vector<smt_termt> &arguments) const
+{
+  return _identifier.get_sort();
+}
+
+void smt_command_functiont::validate(
+  const std::vector<smt_termt> &arguments) const
+{
+  INVARIANT(
+    parameter_sorts.size() == arguments.size(),
+    "Number of parameters and number of arguments must be the same.");
+  const auto parameter_sort_arguments =
+    make_range(parameter_sorts).zip(make_range(arguments));
+  for(const auto &parameter_sort_argument_pair : parameter_sort_arguments)
+  {
+    INVARIANT(
+      parameter_sort_argument_pair.first ==
+        parameter_sort_argument_pair.second.get_sort(),
+      "Sort of argument must have the same sort as the parameter.");
+  }
+}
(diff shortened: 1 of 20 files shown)


src/solvers/smt2_incremental/smt_commands.cpp
```diff
@@ -198,3 +198,51 @@ void smt_commandt::accept(smt_command_const_downcast_visitort &&visitor) const
...
+irep_idt smt_command_functiont::identifier() const
+{
+  return _identifier.identifier();
+}
+
+smt_sortt smt_command_functiont::return_sort(
+  const std::vector<smt_termt> &arguments) const
+{
+  return _identifier.get_sort();
+}
```
TGWDB, line 231: It's not clear here why the `return_sort` function needs the `arguments` parameter.

thomasspriggs:
This PR relates to https://github.com/diffblue/cbmc/issues/6134

src/solvers/smt2_incremental/smt_bitvector_theory.cpp
```diff
@@ -0,0 +1,42 @@
...
+  const smt_termt &right)
+{
+  const auto left_sort = left.get_sort().cast<smt_bit_vector_sortt>();
+  INVARIANT(left_sort, "Left operand must have bitvector sort.");
+  const auto right_sort = right.get_sort().cast<smt_bit_vector_sortt>();
+  INVARIANT(right_sort, "Right operand must have bitvector sort.");
+  INVARIANT(
+    left_sort->bit_width() == right_sort->bit_width(),
+    "Left and right operands must have the same bit width.");
+}
```
TGWDB: Do we need to check they are both signed/unsigned?

src/solvers/smt2_incremental/smt_bitvector_theory.cpp
```diff
@@ -0,0 +1,42 @@
...
+    const smt_termt &left, const smt_termt &right)                             \
+  {                                                                            \
+    validate_bitvector_predicate_arguments(left, right);                       \
+  }                                                                            \
+                                                                               \
+  smt_sortt smt_bit_vector_theoryt::the_name##t::return_sort(                  \
+    const smt_termt &, const smt_termt &)                                      \
+  {                                                                            \
+    return smt_bool_sortt{};                                                   \
+  }                                                                            \
```
TGWDB: Do we need the arguments here?

TGWDB (commented):
There are a lot of tests for different operators, but they all use the same factory code. Is there significant value in having them all?

src/solvers/smt2_incremental/smt_commands.cpp
thomasspriggs: It doesn't in this particular case. But it might need it for things like "bitvector bitwise or" or "ite" from the CORE theory for example.

src/solvers/smt2_incremental/smt_bitvector_theory.cpp
thomasspriggs: Bit vectors in SMT2 bitvector theory do not have signedness as a property. Therefore this information is not available to this factory enforcing these invariants. Good ways to check use of the correct kind of operator might be worth us discussing over slack.

src/solvers/smt2_incremental/smt_bitvector_theory.cpp
thomasspriggs: No. Which is why they are unnamed. It is more straightforward to leaved them unused than it is to make the template code pass them only in some circumstances.

TGWDB (approved):

src/solvers/smt2_incremental/smt_commands.cpp
```diff
@@ -198,3 +198,51 @@ void smt_commandt::accept(smt_command_const_downcast_visitort &&visitor) const
...
+
+void smt_command_functiont::validate(
+  const std::vector<smt_termt> &arguments) const
+{
+  INVARIANT(
+    parameter_sorts.size() == arguments.size(),
+    "Number of parameters and number of arguments must the the same.");
```
chrisr-diffblue: "must the the same" => "must be the same"

unit/solvers/smt2_incremental/smt_commands.cpp
```diff
@@ -94,3 +95,15 @@ TEST_CASE("smt_set_option_commandt getter", "[core][smt2_incremental]")
   CHECK(smt_set_option_commandt{models_on}.option() == models_on);
   CHECK(smt_set_option_commandt{models_off}.option() == models_off);
 }
+
+TEST_CASE("SMT2 function application factory tests", "[core][smt2_incremental]")
```
chrisr-diffblue, line 99: No `CHECK`s in this test? Whats its purpose?

src/solvers/smt2_incremental/smt_commands.cpp
thomasspriggs: Done.

unit/solvers/smt2_incremental/smt_commands.cpp
thomasspriggs: The checks were missing because I got distracted adding other functionality to this PR, before I finished writing this test. Thanks for spotting this oversight. I have now added some checks.

chrisr-diffblue (approved):

PR #6257, opened 2021-07-28
Dump c var overlap

### LINKED ISSUE #6242: `goto-instrument --dump-c` produces invalid code when parameter name overlaps with local variable name
When two local variables have the same base name, --dump-c automatically uses the unique name for all (except the first) variables in that scope to avoid name conflicts. However, this does not consider function parameter names at the moment.

CBMC version: 5.34.0 (cbmc-5.32.1-183-gc50985444)
Operating system: macOS Big Sur v11.4

Exact command line resulting in the issue: `symtab2gb test.json --out test.goto; goto-instrument --dump-c test.goto > test.c`
with [test.json](https://github.com/diffblue/cbmc/files/6858727/test.json.txt)
produces [test.c](https://github.com/diffblue/cbmc/files/6858728/test.c.txt)

What behaviour did you expect: The contents of `test.c` is valid C code.
What happened instead: The parameter to `helper` and one of the local variables have overlapping names.

Generated definition of `helper`:
```
void helper(unsigned int x) {
  int x;
  int helper$$1$$var_3$$x;
}
```

TGWDB:
This PR adds checking for variable scope collision between function local variables and function arguments. These were previously allowed to clash as shown in PR #6242 .  This PR fixes #6242 


src/ansi-c/expr2c.cpp
```diff
@@ -176,6 +176,34 @@ void expr2ct::get_shorthands(const exprt &expr)
       has_collision=!ns_collision[func].insert(sh).second;
     }
 
+    if(!has_collision)
+    {
+      // We could also conflict with a function argument, the code below
+      // finds the function we're in, and checks we don't conflict with
+      // any argument to the function
+      const std::string symbol_str = id2string(symbol_id);
```
chrisr-diffblue, line 184: Is there a reasons this symbol name/symboilt lookup code is not the same as the code in lines 168-173 above?  In fact, can the checks in this block not just be folded into that existing check block?

TGWDB: The check in lines 168-173 is seeing whether the symbol we're looking at now has a conflict with a function symbol name, e.g.
```
void f();

void main(int argc)
{
  int f;
  int g;
```
where here `f` is both a local variable in `main` and also the function `f`.

The new code is finding the current function we're in (e.g. inside `main` above) to check for collisions between local symbols and argument symbols (i.e. above `f` and `g` symbols in the `main` function do not collide with `argc`.

tautschnig (approved):
Would it be possible to add a `dump-c` regression test as well? It would seem rather simple to construct this case. Thank you!

TGWDB:
> Would it be possible to add a `dump-c` regression test as well? It would seem rather simple to construct this case. Thank you!

I can do this, but will wait and do it after #6253 since this includes infrastructure for regression tests of this form.

tautschnig:
> > Would it be possible to add a `dump-c` regression test as well? It would seem rather simple to construct this case. Thank you!
> 
> I can do this, but will wait and do it after #6253 since this includes infrastructure for regression tests of this form.

@TGWDB Did this actually happen? #6253 got merged, and now the diff in this PR looks almost as big, so I'm struggling to tell.

TGWDB:
@tautschnig Yes, two tests added for detecting different variable name collisions in this commit:
https://github.com/diffblue/cbmc/pull/6257/commits/aa0a5dca27403eba5efb614d21c84d45466bc057

PR #6265, opened 2021-08-03
Properly checks struct members in alias expression

feliperodri:

### DIFF
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -26,7 +26,8 @@ assigns_clause_targett::assigns_clause_targett(
     contract(contract),
     init_block(),
     log(log_parameter),
-    local_target(typet())
+    local_target(typet()),
+    target_id(object_ptr.id())
 {
   INVARIANT(
     pointer_object.type().id() == ID_pointer,
@@ -69,6 +70,11 @@ exprt assigns_clause_targett::alias_expression(const exprt &lhs)
   condition.push_back(same_object(lhs_ptr, target));
 
   // If assigns target was a dereference, comparing objects is enough
+  if(target_id == ID_dereference)
+  {
+    return conjunction(condition);
+  }
+
   if(target_id == ID_dereference)
   {
     return conjunction(condition);
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -18,16 +18,16 @@ Date: July 2021
 #include <util/pointer_predicates.h>
 
 assigns_clause_targett::assigns_clause_targett(
-  const exprt &object_ptr,
+  const exprt &object,
   code_contractst &contract,
   messaget &log_parameter,
   const irep_idt &function_id)
-  : pointer_object(pointer_for(object_ptr)),
+  : pointer_object(pointer_for(object)),
     contract(contract),
     init_block(),
     log(log_parameter),
-    local_target(typet()),
-    target_id(object_ptr.id())
+    target(typet()),
+    target_id(object.id())
 {
   INVARIANT(
     pointer_object.type().id() == ID_pointer,
@@ -40,13 +40,12 @@ assigns_clause_targett::assigns_clause_targett(
     function_symbol.location,
     function_symbol.mode);
 
-  local_target = standin_symbol.symbol_expr();
+  target = standin_symbol.symbol_expr();
 
   // Build standin variable initialization block
-  init_block.add(
-    goto_programt::make_decl(local_target, function_symbol.location));
+  init_block.add(goto_programt::make_decl(target, function_symbol.location));
   init_block.add(goto_programt::make_assignment(
-    code_assignt(local_target, pointer_object), function_symbol.location));
+    code_assignt(target, pointer_object), function_symbol.location));
 }
 
 assigns_clause_targett::~assigns_clause_targett()
@@ -56,7 +55,7 @@ assigns_clause_targett::~assigns_clause_targett()
 std::vector<symbol_exprt> assigns_clause_targett::temporary_declarations() const
 {
   std::vector<symbol_exprt> result;
-  result.push_back(local_target);
+  result.push_back(target);
   return result;
 }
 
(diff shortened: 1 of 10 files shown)


regression/contracts/assigns_enforce_arrays_02/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
martin-cs: A. If this was working before, has this PR broken it?
B. If it wasn't then why did the test pass?

regression/contracts/assigns_type_checking_valid_cases/test.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+KNOWNBUG
```
martin-cs: Same question again.

martin-cs (commented):
The code looks fairly good but I am worried about the tests and about disabling so many of them in a PR claiming to fix things.

regression/contracts/assigns_enforce_arrays_02/test.desc
feliperodri: Fixed bug.

regression/contracts/assigns_type_checking_valid_cases/test.desc
feliperodri: Fixed bug.

feliperodri:
> The code looks fairly good but I am worried about the tests and about disabling so many of them in a PR claiming to fix things.

@martin-cs could you take another look? I fixed the bug when dealing with arrays.

martin-cs (approved):
Thank you; that is a massive improvement on the previous version.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -56,25 +55,64 @@ assigns_clause_targett::~assigns_clause_targett()
...
+  // If assigns target was a dereference, comparing objects is enough
+  if(target_id == ID_dereference)
+  {
+    return conjunction(condition);
+  }
+
+  if(target_id == ID_dereference)
+  {
+    return conjunction(condition);
+  }
```
SaswatPadhi: Remove the 4 lines above? They look repeated

SaswatPadhi (changes requested):
I am still thinking about the new conditions we have added, but my main comment for now is to add enough test cases that cover this new change.

1. We are missing test cases for replacement context. We should check that during replacement, we don't havoc more than (the struct field) we are supposed to havoc
2. Both test cases assign to the first field of the struct. Since we look at the offsets etc, could we make one of the test assign to the second field? I'd suggest changing `regression/contracts/assigns_enforce_structs_04/main.c` to assign to `p->y` instead.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Rebase problem. Removed it.

feliperodri:
>     1. We are missing test cases for replacement context. We should check that during replacement, we don't havoc more than (the struct field) we are supposed to havoc

This PR is only for enforcement. We should add test cases for replacement once we update the havoc method.

>     2. Both test cases assign to the first field of the struct. Since we look at the offsets etc, could we make one of the test assign to the second field? I'd suggest changing `regression/contracts/assigns_enforce_structs_04/main.c` to assign to `p->y` instead.

I added two new test cases, could you take another look? @SaswatPadhi

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -705,7 +704,7 @@ void code_contractst::instrument_assign_statement(
     instruction_iterator->source_location));
   alias_assertion.instructions.back().source_location.set_comment(
     "Check that " + from_expr(ns, lhs.id(), lhs) + " is assignable");
-  int lines_to_iterate = alias_assertion.instructions.size();
+  size_t lines_to_iterate = alias_assertion.instructions.size();
   program.insert_before_swap(instruction_iterator, alias_assertion);
   std::advance(instruction_iterator, lines_to_iterate);
```
SaswatPadhi: If you rebase your branch on to the current `develop`, you wouldn't need to keep track of `lines_to_iterate` any more. I added a helper function for this, and you would be able to do:

```c
insert_before_swap_and_advance(program, instruction_iterator, alias_assertion);
```

(same for all other occurrences of `insert_before_swap` followed by `std::advance` pattern in this file)

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -753,10 +752,10 @@ void code_contractst::instrument_call_statement(
...
       // Make freshly allocated memory assignable, if we can determine its type.
       assigns_clause_targett *new_target =
-        assigns_clause.add_pointer_target(rhs);
+        assigns_clause.add_target(dereference_exprt(rhs));
       goto_programt &pointer_capture = new_target->get_init_block();
 
-      int lines_to_iterate = pointer_capture.instructions.size();
+      size_t lines_to_iterate = pointer_capture.instructions.size();
       program.insert_before_swap(local_instruction_iterator, pointer_capture);
       std::advance(instruction_iterator, lines_to_iterate + 1);
```
SaswatPadhi: If you rebase your branch on to the current `develop`, you wouldn't need to keep track of `lines_to_iterate` any more. I added a helper function for this, and you would be able to do:

```c
insert_before_swap_and_advance(program, instruction_iterator, pointer_capture);
```

I don't understand the `+ 1` on the `lines_to_iterate` though. Why is it needed?

SaswatPadhi (approved):
Thanks, LGTM. If you rebase on to `develop`, you could remove some repeated code. Commented below:

src/goto-instrument/contracts/contracts.cpp
feliperodri: Great! Implemented as suggested!

src/goto-instrument/contracts/contracts.cpp
feliperodri: `+1` wasn't necessary there, it came from this case:
```c
    // malloc statments return a void pointer, which is then cast and assigned
    // to a result variable. We iterate one line forward to grab the result of
    // the malloc once it is cast.
```
I re-wrote the code to use `insert_before_swap_and_advance`. Good catch!

SaswatPadhi: Awesome; thanks!

PR #6267, opened 2021-08-03
VSD - to predicate

jezhiggins:
This is a way that a user of the abstract interpreter to produce a logical expression that is true for any of the possible states represented by a single domain. 

For example, if we have three variables in scope and intervals for them:

* x is [0,20] 
* y is [-4,3000] 
* z is TOP

then `to_predicate` could generate

`(0 <= x <= 20) && (-4 <= y <= 3000)`


src/analyses/variable-sensitivity/abstract_object.cpp
```diff
@@ -172,6 +172,20 @@ exprt abstract_objectt::to_constant() const
...
+  if(is_top())
+    return true_exprt();
+  if(is_bottom())
+    return false_exprt();
+  return to_predicate_internal(name);
+}
+
+exprt abstract_objectt::to_predicate_internal(const exprt &name) const
+{
+  return nil_exprt();
```
martin-cs, line 187: As I understand it, this should never be directly called. right?  It is here because it is overloaded?  (I guess this is a point where you are right about the separation of interface and base implementation. )  If so, `UNREACHABLE` might be more appropriate.

martin-cs (approved):
Looks good.  The only thing missing is pointers, which should be relatively straight-forward.

martin-cs (commented):
To make this more directly accessible, would it be possible to add to this PR an implementation of this interface:

https://github.com/diffblue/cbmc/blob/d3dc5601b5680fb9b3e156595279ad8369c2b7a6/src/analyses/ai_domain.h#L157

( as `variable_sensitivity_domaint::to_predicate` ).  All it needs to do is to iterate over the environment map, `abstract_objectt::to_predicate()` each entry and give the conjunct.

Further, would it also be possible to add overloadings of `variable_sensitivity_domaint::to_predicate()` for:

- a single exprt : it should `eval` it and then `abstract_objectt::to_predicate()` the result.
- a vector of exprt : `eval` each one, `abstract_objectt::to_predicate()` and conjunct.

martin-cs (approved):
Thanks for this; looks good.

PR #6269, opened 2021-08-04
Fix evaluation of decreases clause

SaswatPadhi:
`__CPROVER_decreases` clauses on loops were evaluated outside of loop iterations, which was buggy. In this PR, we fix this behavior: decreases clause is evaluated inside the loop block (after the loop guard is satisfied).

### DIFF
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -38,15 +38,46 @@ Date: February 2016
 // This is used in the implementation of multidimensional decreases clauses.
 static exprt create_lexicographic_less_than(
   const std::vector<symbol_exprt> &lhs,
-  const std::vector<symbol_exprt> &rhs);
-
-exprt get_size(const typet &type, const namespacet &ns, messaget &log)
+  const std::vector<symbol_exprt> &rhs)
 {
-  auto size_of_opt = size_of_expr(type, ns);
-  CHECK_RETURN(size_of_opt.has_value());
-  exprt result = size_of_opt.value();
-  result.add(ID_C_c_sizeof_type) = type;
-  return result;
+  PRECONDITION(lhs.size() == rhs.size());
+
+  if(lhs.empty())
+  {
+    return false_exprt();
+  }
+
+  // Store conjunctions of equalities.
+  // For example, suppose that the two input vectors are <s1, s2, s3> and <l1,
+  // l2, l3>.
+  // Then this vector stores <s1 == l1, s1 == l1 && s2 == l2,
+  // s1 == l1 && s2 == l2 && s3 == l3>.
+  // In fact, the last element is unnecessary, so we do not create it.
+  exprt::operandst equality_conjunctions(lhs.size());
+  equality_conjunctions[0] = binary_relation_exprt(lhs[0], ID_equal, rhs[0]);
+  for(unsigned int i = 1; i < equality_conjunctions.size() - 1; i++)
+  {
+    binary_relation_exprt component_i_equality{lhs[i], ID_equal, rhs[i]};
+    equality_conjunctions[i] =
+      and_exprt(equality_conjunctions[i - 1], component_i_equality);
+  }
+
+  // Store inequalities between the i-th components of the input vectors
+  // (i.e. lhs and rhs).
+  // For example, suppose that the two input vectors are <s1, s2, s3> and <l1,
+  // l2, l3>.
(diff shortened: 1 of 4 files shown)


martin-cs (approved):
Looks like an appropriate fix.  Thanks for separating out the refactor into it's own commit; that does make reviewing easier.

SaswatPadhi:
Thanks for taking a look @martin-cs.
I realized that some of the refactoring (renaming) changes had spilled into the next commit, so I fixed that.

@feliperodri could you please take a look too?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -38,21 +38,62 @@ Date: February 2016
...
+  // s1 == l1 && s2 == l2 && s3 == l3>.
+  // In fact, the last element is unnecessary, so we do not create it.
+  exprt::operandst equality_conjunctions(lhs.size());
+  equality_conjunctions[0] = binary_relation_exprt(lhs[0], ID_equal, rhs[0]);
+  for(unsigned int i = 1; i < equality_conjunctions.size() - 1; i++)
+  {
+    binary_relation_exprt component_i_equality{lhs[i], ID_equal, rhs[i]};
+    equality_conjunctions[i] =
+      and_exprt(equality_conjunctions[i - 1], component_i_equality);
+  }
```
```suggestion
  exprt::operandst equality_conjunctions;
  for(unsigned int i = 0; i < lhs.size(); i++)
  {
    binary_relation_exprt component_i_equality{lhs[i], ID_equal, rhs[i]};
    equality_conjunctions.push_back(component_i_equality);
  }
  exprt equality_constraint = conjunction(equality_conjunctions);
```

feliperodri: You'd also need to update the next loop in the same style. Perhaps, even combine them together? Just a suggestion, note a hard requirement.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -199,7 +226,23 @@ void code_contractst::check_apply_loop_contracts(
...
   // Now havoc at the loop head.
   // Use insert_before_swap to preserve jumps to loop head.
-  goto_function.body.insert_before_swap(loop_head, havoc_code);
+  insert_before_swap_and_advance(goto_function.body, loop_head, havoc_code);
+
+  // Generate: assignments to store the multidimensional decreases clause's
+  // value before the loop
+  if(!decreases_clause.is_nil())
+  {
+    for(size_t i = 0; i < old_decreases_vars.size(); i++)
```
feliperodri, line 235: updated previous loops with `size_t` as well for consistency.

src/goto-instrument/contracts/contracts.h
```diff
@@ -123,7 +123,7 @@ class code_contractst
   void check_apply_loop_contracts(
     goto_functionst::goto_functiont &goto_function,
     const local_may_aliast &local_may_alias,
-    const goto_programt::targett loop_head,
+    goto_programt::targett loop_head,
```
feliperodri, line 126: I didn't see any modification in `loop_head`, is this change necessary?

feliperodri (changes requested):
Only minor changes.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: So the `equality_conjunctions` vector stores a list of conjunctions. In your suggestion, it only stores equality predicates and we compute a single conjunction at the end. That wouldn't work.

Basically we want to have `lhs[0] == rhs[0]`, `lhs[0] == rhs[0] && lhs[1] == rhs[1]`, and so on in the `equality_conjunctions` vector.

src/goto-instrument/contracts/contracts.h
SaswatPadhi: Yes, you can look at line 229 and the `insert_before_swap_and_advance` function (specifically, line 90 in that function).

We insert some code (havoc & assumptions) before the loop head, and this change should have been made earlier. We didn't care about it because we were not really using the `loop_head` after that insertion. Now that we do, we need to advance if correctly.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Fixed as suggested.

src/goto-instrument/contracts/contracts.h
feliperodri: Thanks for the clarification!

feliperodri (approved):

PR #6270, opened 2021-08-04
Fix quantifiers for arrays only works with Z3 and CVC4

### LINKED ISSUE #6263: MathSAT SMT2 not working
CBMC version: Tested with CBMC 5.27.0 and 5.35.0
Operating system: Fedora 34
Test Case:
```c
#include <assert.h>

void main(){
  assert(0);
}
```
```bash
$cbmc simpleProgram.c --mathsat | tail
SMT2 solver returned error message:
    "The CNF conversion does not support quantifiers"
Running SMT2 QF_AUFBV using MathSAT
Runtime Solver: 0.0204232s
Runtime decision procedure: 0.0205809s

** Results:
simpleProgram.c function main
[main.assertion.1] line 4 assertion 0: ERROR

** 0 of 1 failed (1 iterations)
VERIFICATION ERROR
```
What behaviour did you expect: Mathsat returns SAT
What happened instead: MathSAT exits with an error message

I believe this is due to commit "[Fixed SMT encoding of array_of_expr](https://github.com/diffblue/cbmc/commit/aaa7a3a5394aaf5c8a283cba55094c19e5daecc3)" first included in version 5.27.0, which enabled generation of using a quantifier-based initialization instead of lambda in smt2_conv.cpp line 4400, while MathSAT does not fully support quantifiers.
Version 5.26.0 correctly fails the verification.

Previously I created a [question on the CProver Support list](https://groups.google.com/u/1/g/cprover-support/c/hSkS_o9mddI).

TGWDB:
This commit fixes an error introduced in #5974 where quantifiers
were produced for all SMT2 solvers despite only being supported by
Z3 and CVC4. The fix here reverts to the older behaviour for other
solvers.

Fixes #6263

### DIFF
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -4742,11 +4742,15 @@ void smt2_convt::find_symbols(const exprt &expr)
           }
           out << ")))\n";
         }
-        // else
-        // {
-        //   // This is where an alternate for other solvers that does not use
-        //   // quantifiers should go.
-        // }
+        else
+        {
+          // This is where an alternate for other solvers that does not use
+          // quantifiers should go.
+          INVARIANT(
+            false,
+            "Cannot substitute quantified expression for lambda in current "
+            "solver.");
+        }
 
         defined_expressions[expr] = id;
       }
--- a/src/solvers/smt2/smt2_conv.cpp
+++ b/src/solvers/smt2/smt2_conv.cpp
@@ -4721,21 +4721,32 @@ void smt2_convt::find_symbols(const exprt &expr)
         convert_type(array_type);
         out << ")\n";
 
-        // use a quantifier-based initialization instead of lambda
-        out << "(assert (forall ((i ";
-        convert_type(array_type.size().type());
-        out << ")) (= (select " << id << " i) ";
-        if(array_type.element_type().id() == ID_bool && !use_array_of_bool)
+        // The code below only works in Z3 and CVC4, it was added (and
+        // approved/merged) despite breaking other solvers.
+        // This conditional removes this for other solvers
+        if(solver == solvert::CVC4 || solver == solvert::Z3)
         {
-          out << "(ite ";
-          convert_expr(array_of.what());
-          out << " #b1 #b0)";
+          // use a quantifier-based initialization instead of lambda
+          out << "(assert (forall ((i ";
+          convert_type(array_type.size().type());
+          out << ")) (= (select " << id << " i) ";
+          if(array_type.element_type().id() == ID_bool && !use_array_of_bool)
+          {
+            out << "(ite ";
+            convert_expr(array_of.what());
+            out << " #b1 #b0)";
+          }
+          else
+          {
+            convert_expr(array_of.what());
+          }
+          out << ")))\n";
         }
-        else
-        {
-          convert_expr(array_of.what());
-        }
-        out << ")))\n";
+        // else
+        // {
+        //   // This is where an alternate for other solvers that does not use
+        //   // quantifiers should go.
+        // }
 
         defined_expressions[expr] = id;
       }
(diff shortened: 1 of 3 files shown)


kroening:
MathSAT does not support `forall` either.

TGWDB:
> MathSAT does not support `forall` either.

Now we only produce the `forall` in the array code block for `Z3` and `CVC4`, all other solvers will have the old code.

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -4552,21 +4552,32 @@ void smt2_convt::find_symbols(const exprt &expr)
...
-        {
-          out << "(ite ";
-          convert_expr(array_of.what());
-          out << " #b1 #b0)";
-        }
-        else
+        // The code below only works in Z3 and CVC4, it was added (and
+        // approved/merged) despite breaking other solvers.
+        // This conditional removes this for other solvers
+        if(solver == solvert::CVC4 || solver == solvert::Z3)
```
SaswatPadhi, line 4727: Thanks for fixing this. In my original commit, I didn't realize that we also support solvers that do not have quantifier support.

Instead of checking for specific solvers, should we introduce `use_quantified_array_init`? This can be set to `true` for `Z3` and `CVC4`in the case split we have in the constructor, and we wouldn't need the solver-specific checks or the comments above and below. We use the same pattern for other solver-specific features, e.g. array of Bools support, as const support etc.

Actually, may be we should have a `use_quantifier` flag instead to track solvers that support quantifiers. The issue here is with the use of quantifiers, not really specific to array initialization. May be we should also go over other uses of quantifiers in this file and guard them with `use_quantifier`.

SaswatPadhi (changes requested):

src/solvers/smt2/smt2_conv.cpp
TGWDB: Agree that a flag for quantifiers is probably the better approach, the PR above is a bit limited and incomplete (also we should have a flag and not solver specific code in random places).

TGWDB: I've gone with a very lightweight (minimum) fix and rebase for now. Is this still blocking?

SaswatPadhi: Thank you! No this isn't a blocker currently.

SaswatPadhi (approved):

NlightNFotis (approved):

src/solvers/smt2/smt2_conv.cpp
```diff
@@ -4721,21 +4721,32 @@ void smt2_convt::find_symbols(const exprt &expr)
...
+            convert_expr(array_of.what());
+          }
+          out << ")))\n";
         }
-        out << ")))\n";
+        // else
+        // {
+        //   // This is where an alternate for other solvers that does not use
+        //   // quantifiers should go.
+        // }
```
tautschnig: We should not silently do this. It's ok to fail an invariant, but let's please not generate incomplete formulae without telling the user.

tautschnig (changes requested):

tautschnig (approved):

tautschnig:
@TGWDB Thank you for adding the invariant. This is now, not unexpectedly, failing in some tests. I'd suggest to reinstate the regex pattern in the test, and mark the test as known-broken.

kroening (changes requested):
This needs more work -- the `array_of` expression is an array constructor for an array that maps all indices to one value. Various solvers do support SMT2-LIB array constructor expressions, which should be used here.

kroening:
Perhaps a bit more controversial, we could consider dropping support for solvers that support neither `as const` nor the alternative that uses quantifiers. You won't get anything through CBMC without at least one of these.

TGWDB:
@kroening I agree that this PR does not correctly solve all the issues with various solvers. This was intended to be a sport fix for an issue introduced in a prior PR that allowed us to crash some solvers. Instead now this produces an error from CBMC (where we should be building something better). With that in mind, would you be willing to accept this spot fix since in practice this is a fix to the output/failure behaviour? (Also in the context where a new SMT backend is being developer, which may deprecate this code path or enable a better solution in future.)

PR #6271, opened 2021-08-04
VSD - liveness context

jezhiggins:
Adds a new abstract_objectt in the `context_abstract_objectt` hierarchy which stores a single location, indicating where this value became live.

- **Transform** : if the instruction is an assignment and a variable is written to (including merge 'maybe' writes) then the location is set to the current instruction.

- **Merge** : if the instruction is the same on both sides of the merge, then use that value. If it is different then use the location of the merge point.


src/analyses/variable-sensitivity/variable_sensitivity_domain.h
```diff
@@ -78,8 +78,8 @@
   "(vsd-pointers):"                                                            \
   "(vsd-unions):"                                                              \
   "(vsd-flow-insensitive)"                                                     \
-  "(vsd-data-dependencies)"
-
+  "(vsd-data-dependencies)"                                                    \
+  "(vsd-region-tracking)"
```
martin-cs: A backslash and a blank new line after so that future additions are just one line?

src/analyses/variable-sensitivity/region_context.cpp
```diff
@@ -0,0 +1,373 @@
...
+  return !all_matched;
+}
+
+/**
+ * Internal helper function to format and output a given set of locations
+ *
+ * \param out the stream on which to output the locations
+ * \param locations the set of locations to output
+ */
+void region_contextt::output_last_written_locations(
```
martin-cs: Is this needed?  If it is general to all of the domains that track locations then maybe it could be put somewhere more general.  If not then could it be simplified as only one location is stored.

regression/goto-analyzer/liveness-struct/test-write-location.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+--variable-sensitivity --vsd-structs every-field --vsd-values intervals --vsd-liveness --show
```
martin-cs: Should this have `--vsd-liveness` set?

regression/goto-analyzer/liveness/test-liveness.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+--variable-sensitivity --vsd-values set-of-constants --vsd-liveness --show
+^EXIT=0$
+^SIGNAL=0$
+^main::1::x .* TOP @ \[1\]
+^main::1::x .* value-set-begin: 0 :value-set-end @ \[2\]
+^main::1::x .* value-set-begin: 2, 3 :value-set-end @ \[7\]
```
martin-cs, line 8: Pedantry probably but could we have the other two regions for illustration and completeness.

src/analyses/variable-sensitivity/abstract_object.cpp
```diff
@@ -206,6 +206,19 @@ void abstract_objectt::output(
...
 }
 
+abstract_objectt::combine_result abstract_objectt::merge(
+  const abstract_object_pointert &op1,
+  const abstract_object_pointert &op2,
+  const locationt &merge_location,
+  const widen_modet &widen_mode)
+{
+  abstract_object_pointert result = merge(op1, op2, widen_mode).object;
+  result = result->merge_location_context(merge_location);
```
martin-cs, line 216: Is a separate call better than an extra argument here?

martin-cs: Ah; I think I see it.  It is the partner of `write_location_context` which can't easily be folded into eval.

src/analyses/variable-sensitivity/liveness_context.cpp
```diff
@@ -0,0 +1,220 @@
...
+  const namespacet &ns,
+  const std::stack<exprt> &stack,
+  const exprt &specifier,
+  const abstract_object_pointert &value,
+  bool merging_write) const
+{
+  abstract_object_pointert updated_child = child_abstract_object->write(
+    environment, ns, stack, specifier, value, merging_write);
+
+  // Only perform an update if the write to the child has in fact changed it...
```
martin-cs: I'm not exactly sure what (code level) condition this is checking?

src/analyses/variable-sensitivity/liveness_context.cpp
```diff
@@ -0,0 +1,220 @@
...
+  // Even if the pointers are different, it maybe that it has changed only
+  // because of a merge operation, rather than an actual write. Given that
+  // this class has knowledge of where writes have occured, use that
+  // information to determine if any writes have happened and use that as the
+  // proxy for whether the value has changed or not.
+  //
+  // For two sets of last written locations to match,
+  // each location in one set must be equal to precisely one location
+  // in the other, since a set can assume at most one match
+  return before_context->get_location() != get_location();
```
martin-cs: As `locationt`'s are iterators into (potentially different) lists, we should compare location numbers instead of directly comparing locations.

src/analyses/variable-sensitivity/variable_sensitivity_domain.h
```diff
@@ -78,8 +78,8 @@
   "(vsd-pointers):"                                                            \
   "(vsd-unions):"                                                              \
   "(vsd-flow-insensitive)"                                                     \
-  "(vsd-data-dependencies)"
-
+  "(vsd-data-dependencies)"                                                    \
+  "(vsd-liveness)"
```
martin-cs: Add a backslash and a new line so that future additions are single lines?

martin-cs (approved):
Thanks for the tidy up and refactoring as well as the new abstract object.  Please can you add a test case involving function calls.  It needs a function that is called at least twice and a version with and without `--three-way-merge`.

regression/goto-analyzer/liveness-struct/test-write-location.desc
jezhiggins: nope!

martin-cs (approved):
Thanks for chasing up the three-way-merge issues.

PR #6273, opened 2021-08-04
Predicate abstraction for monotonic change

LongPham7:
In this pull request, I have implemented predicate abstraction for monotonic change: it checks whether variables have monotonically increased or decreased since the beginning of given GOTO code. This predicate abstraction will be used in the automatic inference of decreases clauses (aka ranking functions or loop variants). For example, consider a while-loop whose loop guard is `x < y`. If we know that `x` monotonically increases in the loop body and that `y` stays constant, we can successfully infer that `y - x` is a correct decreases clause.

This predicate abstraction works as follows. For each variable, we keep track of its predicate, namely monotonic-change status. There are six possible statuses:
1. Declared but uninitialized
2. Constant
3. Monotonic increase: the variable has monotonically increased since the beginning of the code. In this context, "monotonic" increase/decrease means "strict" increase/decrease. 
4. Monotonic decrease
5. Top: inconclusive. That is, we have no information about the given variable.
6. Bottom: infeasible. That is, no variables can possess this status. 

Here is an example: 
```
DECL x            // x is declared but uninitialized
ASSIGN x := 42    // x has stayed constant
ASSIGN x := x + 1 // x has monotonically increased
ASSIGN x := x - 1 // x is now top
```

Note that this predicate abstraction does NOT keep track of how much a variable has changed. So if we have `x--` followed by `x++` as in the above example, the predicate abstraction cannot tell us the net effect on `x`. This is why `x`'s status after the last line of the above example is top. 

### DIFF
--- a/regression/goto-analyzer/monotonic_change_function_arguments/main.c
+++ b/regression/goto-analyzer/monotonic_change_function_arguments/main.c
@@ -1,0 +1,8 @@
+int foo(int x, int y)
+{
+  x++;
+  x += 6;
+
+  y--;
+  y -= 6;
+}
--- a/regression/goto-analyzer/monotonic_change_function_call/main.c
+++ b/regression/goto-analyzer/monotonic_change_function_call/main.c
@@ -1,0 +1,19 @@
+int nondet_bool(void);
+
+int increment(int x)
+{
+  if(nondet_bool())
+    x++;
+  else
+    x += 6;
+  return x;
+}
+
+int main()
+{
+  int x = 42;
+  x++;
+  int y = increment(x);
+  y--;
+  return 0;
+}
--- a/regression/goto-analyzer/monotonic_change_struct/main.c
+++ b/regression/goto-analyzer/monotonic_change_struct/main.c
@@ -1,0 +1,25 @@
+int nondet(void);
+int nondet_bool(void);
+
+struct cartesian_coorindate
+{
+  int x;
+  int y;
+};
+
+int main()
+{
+  struct cartesian_coorindate u = {nondet(), nondet()};
+  if(nondet_bool())
+  {
+    u.x++;
+    u.y--;
+  }
+  else
+  {
+    u.x += 6;
+    u.y -= 6;
+  }
+
+  return 0;
+}
--- a/regression/goto-analyzer/monotonic_change_values/main.c
+++ b/regression/goto-analyzer/monotonic_change_values/main.c
@@ -1,0 +1,22 @@
+int nondet(void);
+
+int main()
+{
+  int x = 42;
+  x++;
+  x = x + 2;
+
+  int y = 42;
+  y--;
+  y = y - 2;
+
+  int z = (y + 1729) * (x - 6);
+  z++;
+  z--;
+
+  int u = nondet();
+  u++;
+  u *= 2;
+
+  return 0;
+}
--- a/src/analyses/variable-sensitivity/monotonic_change.cpp
+++ b/src/analyses/variable-sensitivity/monotonic_change.cpp
@@ -1,0 +1,355 @@
+/*******************************************************************\
+
+ Module: Predicate abstraction for monotonic change
+
+ Authors: Long Pham, Saswat Padhi
+
+\*******************************************************************/
+
+// For the documentation of the class monotonic_changet, have a look at
+// the header file monotonic_change.h.
+
+#include <langapi/language_util.h>
+
+#include <util/interval.h>
+#include <util/std_expr.h>
+
+#include "abstract_object_statistics.h"
+#include "monotonic_change.h"
+
+// We will print out a message on the standard output when we encouter a
+// pathological case. An example is when we are try to merge the abstract value
+// "uninitialized" with other abstract values (e.g. "increase"). This is why we
+// need <iostream>.
+#include <iostream>
+
+monotonic_changet::monotonic_changet(const typet &t)
+  : abstract_value_objectt(t, false, false), monotonicity_value(uninitialized)
+{
+}
+
+monotonic_changet::monotonic_changet(const typet &t, bool tp, bool bttm)
+  : abstract_value_objectt(t, tp, bttm),
+    monotonicity_value((tp || bttm) ? top_or_bottom : uninitialized)
+{
+}
+
+monotonic_changet::monotonic_changet(
+  const typet &t,
+  bool tp,
+  bool bttm,
+  monotonicity_flags initial_value)
+  : abstract_value_objectt(t, tp, bttm),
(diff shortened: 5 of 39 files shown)


martin-cs:
Exciting stuff; looking forward to reviewing it!

LongPham7:
I've just finished adding comments to the code. It is now ready for an initial review. Thanks a lot!

regression/goto-analyzer/monotonic_change_function_arguments/main.c
```diff
@@ -0,0 +1,10 @@
+int foo(int x, int y);
+
```
SaswatPadhi: [nit]

Remove these 2 lines

```suggestion
```

regression/goto-analyzer/monotonic_change_function_call/main.c
```diff
@@ -0,0 +1,20 @@
+int nondet_bool(void);
+int increment(int x);
```
SaswatPadhi: [nit]

You could simply define `increment` here right?
Is there something specific that you want to test with the forward declaration and a definition later?

regression/goto-analyzer/monotonic_change_struct/main.c
```diff
@@ -0,0 +1,25 @@
...
+
+struct cartesian_coorindate
+{
+  int x;
+  int y;
+};
+
+int main()
+{
+  struct cartesian_coorindate u = {nondet(), nondet()};
```
SaswatPadhi, line 12: Would it matter if we just wrote:

```suggestion
  struct cartesian_coorindate u;
```

and perhaps remove the `nondet` declaration above?

regression/goto-analyzer/monotonic_change_values/main.c
```diff
@@ -0,0 +1,22 @@
...
+
+  int y = 42;
+  y--;
+  y = y - 2;
+
+  int z = (y + 1729) * (x - 6);
+  z++;
+  z--;
+
+  int u = nondet();
```
SaswatPadhi, line 17: Can we write:
```suggestion
  int u;
```
and remove the `nondet` declaration? Variables that are declared (but not defined) are nondet by default

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.cpp
```diff
@@ -6,9 +6,13 @@
 
 \*******************************************************************/
 #include "variable_sensitivity_object_factory.h"
+#include "abstract_value_object.h"
 #include "full_array_abstract_object.h"
+#include "monotonic_change.h"
 #include "value_set_pointer_abstract_object.h"
 
+#include <iostream>
```
SaswatPadhi: Remove this?

src/analyses/variable-sensitivity/monotonic_change.cpp
```diff
@@ -0,0 +1,331 @@
...
+        this->type(), false, false, increase);
+    case decrease:
+      return std::make_shared<monotonic_changet>(
+        this->type(), true, false); // top
+    case uninitialized:
+      std::cout << "We are trying to merge increase with uninitialized\n";
+      break;
+    case top_or_bottom:
+    default:
+      std::cout << "This should be unreachable\n";
```
SaswatPadhi: Please use:
```suggestion
      UNREACHABLE;
```

(at several other places below as well)

src/analyses/variable-sensitivity/monotonic_change.cpp
```diff
@@ -0,0 +1,331 @@
...
+      return std::make_shared<monotonic_changet>(
+        this->type(), false, false, increase);
+    case decrease:
+      return std::make_shared<monotonic_changet>(
+        this->type(), true, false); // top
+    case uninitialized:
+      std::cout << "We are trying to merge increase with uninitialized\n";
+      break;
+    case top_or_bottom:
+    default:
```
SaswatPadhi: It's be nice to explicitly document fall-through cases.

```suggestion
    case top_or_bottom:
      [[fallthrough]];
    default:
```

`[[fallthrough]];` is a C++17 feature though. If that doesn't compile, please use `// fall through` instead.

src/analyses/variable-sensitivity/monotonic_change.cpp
```diff
@@ -0,0 +1,331 @@
...
+#include <util/std_expr.h>
+
+#include "abstract_object_statistics.h"
+#include "monotonic_change.h"
+
+// We will print out a message on the standard output when we encouter a
+// pathological case. An example is when we are try to merge the abstract value
+// "uninitialized" with other abstract values (e.g. "increase"). This is why we
+// need <iostream>.
+#include <iostream>
```
SaswatPadhi, line 24: Should we use `log` functions instead of `cout`?

@martin-cs May be you could advise us?

SaswatPadhi (commented):
Some initial comments

regression/goto-analyzer/monotonic_change_function_call/main.c
LongPham7: Thanks for spotting a redundant function prototype - I wasn't carefully considering whether it was necessary.

regression/goto-analyzer/monotonic_change_struct/main.c
LongPham7: Good point. Your comment made me realize that there is a hole in my understanding of the semantics. Suppose we declare a variable `x` (without initializing it). At this point, the abstract value of `x` is "declared but uninitialized." Now suppose we run `x := x+1`. Should the abstract value now be "constant" or "monotonic increase"? In the current implementation, the abstract value will become "constant." However, I now wonder whether this is a reasonable behavior.

src/analyses/variable-sensitivity/monotonic_change.cpp
LongPham7: Thanks for the suggestion.

src/analyses/variable-sensitivity/monotonic_change.cpp
LongPham7: Thanks for the suggestion. `[[fallthrough]];` works.

src/analyses/variable-sensitivity/variable_sensitivity_object_factory.cpp
LongPham7: Oh, yes, I will remove it.

src/analyses/variable-sensitivity/monotonic_change.cpp
LongPham7: No, actually, the CI complains about `[[fallthrough]];`. I will use `// fall through` instead.

SaswatPadhi:
Closing as #6317 supersedes this PR.

PR #6277, opened 2021-08-06
Adds support for struct members in history variables

feliperodri:

### DIFF
--- a/doc/cprover-manual/contracts-history-variables.md
+++ b/doc/cprover-manual/contracts-history-variables.md
@@ -13,8 +13,9 @@ _ensures_ clause.
 
 ### Parameters
 
-`__CPROVER_old` takes a single argument, which is the identifier corresponding to
-a parameter of the function. For now, only scalar or pointer types are supported.
+`__CPROVER_old` takes a single argument, which is the identifier
+corresponding to a parameter of the function. For now, only scalars,
+pointers, and struct members are supported.
 
 ### Semantics
 
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -394,7 +394,6 @@ void code_contractst::replace_old_parameter(
 
     const auto &parameter = to_old_expr(expr).expression();
 
-    // TODO: generalize below
     if(
       parameter.id() == ID_dereference || parameter.id() == ID_member ||
       parameter.id() == ID_symbol)
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -395,22 +395,22 @@ void code_contractst::replace_old_parameter(
     const auto &parameter = to_old_expr(expr).expression();
 
     // TODO: generalize below
-    if(parameter.id() == ID_dereference)
+    if(
+      parameter.id() == ID_dereference || parameter.id() == ID_member ||
+      parameter.id() == ID_symbol)
     {
-      const auto &dereference_expr = to_dereference_expr(parameter);
-
-      auto it = parameter2history.find(dereference_expr);
+      auto it = parameter2history.find(parameter);
 
       if(it == parameter2history.end())
       {
         // 1. Create a temporary symbol expression that represents the
         // history variable
         symbol_exprt tmp_symbol =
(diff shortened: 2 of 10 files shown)


doc/cprover-manual/contracts-history-variables.md
```diff
@@ -13,8 +13,9 @@ _ensures_ clause.
 
 ### Parameters
 
-`__CPROVER_old` takes a single argument, which is the identifier corresponding to
-a parameter of the function. For now, only scalar or pointer types are supported.
+`__CPROVER_old` takes a single argument, which is the identifier
+corresponding to a parameter of the function. For now, only scalars,
+pointers, and struct members are supported.
```
SaswatPadhi, line 18: The `if` condition that was updated below only allows dereference and struct members. Do we need to have some other condition for simple scalars?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -395,22 +395,20 @@ void code_contractst::replace_old_parameter(
     const auto &parameter = to_old_expr(expr).expression();
 
     // TODO: generalize below
-    if(parameter.id() == ID_dereference)
+    if(parameter.id() == ID_dereference || parameter.id() == ID_member)
```
SaswatPadhi: I think in the assigns clause code I also saw `ID_ptrmember`. I thought that `a.b` has `ID_member` and `a->b` has `ID_ptrmember`, but the regression tests you added seem to work with just `ID_member` check, so I might be wrong.
But do we need an `ID_ptrmember` case here?

SaswatPadhi (commented):
Nice! Thanks 😃

The changes look good, but I have a couple of questions regarding the scope

doc/cprover-manual/contracts-history-variables.md
feliperodri: Good catch, we needed to add `ID_symbol`. I also added a new regression test to cover this case.

src/goto-instrument/contracts/contracts.cpp
feliperodri: I added more tests to cover these cases. `ID_member` seems enough.

feliperodri:
@SaswatPadhi Could you take another look?

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I did a quick search, and it looks like `ID_ptrmember` indeed corresponds to `->`:
https://github.com/diffblue/cbmc/blob/1ab5de1ac0893c7ca94ffe5463e2c7c4ee781266/src/cpp/cpp_typecheck_expr.cpp#L454

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -395,22 +395,22 @@ void code_contractst::replace_old_parameter(
     const auto &parameter = to_old_expr(expr).expression();
 
     // TODO: generalize below
```
SaswatPadhi: May be remove this comment now, since this matches the assigns clause constraints? Or actually, may be mention: `TODO: Look into ID_ptrmember`?

SaswatPadhi (approved):

src/goto-instrument/contracts/contracts.cpp
feliperodri: @SaswatPadhi This is for `cpp` not `ansi-c`.

src/goto-instrument/contracts/contracts.cpp
feliperodri: Remove it.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: @feliperodri Are you sure this doesn't apply to ANSI-C as well? From the ANSI-C parser:

https://github.com/diffblue/cbmc/blob/5592d7e414ee3f440631f7281073345516391d75/src/ansi-c/parser.y#L565-L569

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I don't think the `ID_ptrmember` issue was fully resolved. I think we should have resolved this fully before merging.

PR #6278, opened 2021-08-06
Added support for loop_entry history variable.

aalok-thakkar:
In this commit we introduce a new expression, `__CPROVER_loop_entry`, which can be used within loop invariants to track history of variables.
`__CPROVER_loop_entry(x)` returns the initial value of a variable just before the very first iteration of a loop.


src/goto-instrument/contracts/contracts.cpp
```diff
@@ -387,10 +398,11 @@ void code_contractst::replace_old_parameter(
     replace_old_parameter(op, parameter2history, location, mode, history);
   }
 
-  if(expr.id() == ID_old)
+  if(expr.id() == ID_old ||
+     expr.id() == ID_loop_entry)
   {
     DATA_INVARIANT(
-      expr.operands().size() == 1, CPROVER_PREFIX "old must have one operand");
+      expr.operands().size() == 1, "history variable must have one operand");
```
feliperodri: Why is this better than the previous message?

feliperodri (changes requested):
We need regression tests.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I think Aalok was trying to generalize the error message for both `__CPROVER_old` and `__CPROVER_loop_entry`.

I just have a minor suggestion:
```suggestion
      expr.operands().size() == 1, "A history variable must have one operand.");
```

Note that this is not a user-facing error message, but an internal invariant failure in case we tried to construct a bad history variable. But we can improve the invariant failure message, if this is unclear. Do you have any suggestions, @feliperodri?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -387,10 +398,11 @@ void code_contractst::replace_old_parameter(
     replace_old_parameter(op, parameter2history, location, mode, history);
   }
 
-  if(expr.id() == ID_old)
+  if(expr.id() == ID_old ||
+     expr.id() == ID_loop_entry)
```
SaswatPadhi: Do we need to split the line here?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -427,7 +439,7 @@ void code_contractst::replace_old_parameter(
     }
     else
     {
-      log.error() << CPROVER_PREFIX "old does not currently support "
+      log.error() << "history variables do not currently support "
```
SaswatPadhi:
```suggestion
      log.error() << "History variables do not currently support "
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -174,6 +174,17 @@ void code_contractst::check_apply_loop_contracts(
...
+  // Process "loop_entry" history variables
+
+  std::map<exprt, exprt> parameter2history;
+  goto_programt history;
+
+  // Find and replace "loop_entry" expression in the "invariant" variable
+  replace_old_parameter(invariant, parameter2history, loop_head->source_location, mode, history);
+
+  // Create 'loop_entry' history variables
+  insert_before_swap_and_advance(havoc_code, loop_head, history);
```
SaswatPadhi: May be you meant:
```suggestion
  insert_before_swap_and_advance(goto_function.body, loop_head, history);
```
because `loop_head` is an instruction iterator into that, not `havoc_code`?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -174,6 +174,17 @@ void code_contractst::check_apply_loop_contracts(
     return invariant_copy;
   };
 
+  // Process "loop_entry" history variables
+
+  std::map<exprt, exprt> parameter2history;
+  goto_programt history;
```
SaswatPadhi: As per our offline discussion, may be just reuse `havoc_code` instead of defining this?

SaswatPadhi (changes requested):

regression/contracts/loop-history-variables-01/main.c
```diff
@@ -0,0 +1,17 @@
...
+
+  while(y > 0)
+  __CPROVER_loop_invariant(*x == __CPROVER_loop_entry(*x))
+  {
+    --y;
+    *x = *x + 1;
+    *x = *x - 1;
+  }
+  assert(*x == &z);
+}
```
SaswatPadhi: Please add a newline at the end (this file and others).

regression/contracts/loop-history-variables-01/main.c
```diff
@@ -0,0 +1,17 @@
+#include <assert.h>
+
+void main ()
+{
+  int *x, y, z; 
+
+  *x = &z;
+
+  while(y > 0)
+  __CPROVER_loop_invariant(*x == __CPROVER_loop_entry(*x))
```
SaswatPadhi: Could you also add a couple more loops to this file and test for struct members and scalars as well?

SaswatPadhi (changes requested):

regression/contracts/loop-history-variables-01/test.desc
```diff
@@ -0,0 +1,11 @@
...
+main.c
+--apply-loop-contracts
+^EXIT=0$
+^SIGNAL=0$
+^\[main.1\] .* Check loop invariant before entry: SUCCESS$
+^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
+^VERIFICATION SUCCESSFUL$
+--
+--
+This test checks that the history variable for loop entry is supported.
```
feliperodri: Add new line.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -387,10 +398,19 @@ void code_contractst::replace_old_parameter(
...
+    if (expr.id() == ID_old)
+    {
+      DATA_INVARIANT(
       expr.operands().size() == 1, CPROVER_PREFIX "old must have one operand");
+    }
+    if (expr.id() == ID_loop_entry)
+    {
+      DATA_INVARIANT(
+      expr.operands().size() == 1, CPROVER_PREFIX "loop_entry must have one operand");
+    }
```
feliperodri: @SaswatPadhi if we add info of the source location, we could perhaps avoid code duplication and keep a single error message for both history variables. What do you think? Only a suggestion.

feliperodri (approved):
LGTM modulo Saswat's comments.

src/goto-instrument/contracts/contracts.cpp
feliperodri: See https://github.com/diffblue/cbmc/pull/6278#discussion_r684582662

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -427,9 +447,18 @@ void code_contractst::replace_old_parameter(
...
+        log.error() << CPROVER_PREFIX "old does not currently support "
                   << parameter.id() << " expressions." << messaget::eom;
       throw 0;
+      }
+      if (expr.id() == ID_loop_entry)
+      {
+        log.error() << CPROVER_PREFIX "loop_entry does not currently support "
+                  << parameter.id() << " expressions." << messaget::eom;
+      throw 0;
+      }
```
SaswatPadhi:
```suggestion
      log.error() << "History variables do not currently support "
                  << parameter.id() << " expressions." << messaget::eom;
      throw 0;
```

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: This is an internal invariant (`DATA_INVARIANT`) failure, not a user-facing error message; so I don't think source location would make much sense.

But looking at this code now, I don't think this entire thing is necessary at all. You can remove lines 404 - 414.

The `to_old_expr` call @ line 415 would call `validate_expr` for the base class `unary_expr`, which would anyway check the number of operands internally:

https://github.com/diffblue/cbmc/blob/5b8e8fa6fc8eb6b241aaca38e4e706568a7ac158/src/util/std_expr.h#L317-L320

If we want to show user-facing errors, that is already done within the parser.

SaswatPadhi (changes requested):

src/goto-instrument/contracts/contracts.cpp
aalok-thakkar: I've made separate functions and avoided generalization.

SaswatPadhi: Please don't do that. That looks like a lot of duplicate code.

As I have commented below, this operand check can simply be removed.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -375,24 +392,26 @@ void code_contractst::add_quantified_variable(
...
-  if(expr.id() == ID_old)
+  if(expr.id() == ID_old || expr.id() == ID_loop_entry)
   {
-    DATA_INVARIANT(
-      expr.operands().size() == 1, CPROVER_PREFIX "old must have one operand");
+    const auto &parameter = to_history_expr(expr, id).expression();
 
-    const auto &parameter = to_old_expr(expr).expression();
+    DATA_INVARIANT(
+      expr.operands().size() == 1, "A history variable must have one operand");
```
SaswatPadhi: The 2 lines above are redundant. Please remove those.

The number of operands check would be performed during `to_history_expr` call @ line 411, as I mentioned in https://github.com/diffblue/cbmc/pull/6278#discussion_r684686555.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Not addressed yet.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -836,3 +837,21 @@ void c_typecheck_baset::typecheck_spec_decreases(codet &code)
     }
   }
 }
+
+void c_typecheck_baset::disallow_function_history_variables(
```
SaswatPadhi: Why do we need this? There is `c_typecheck_baset::disallow_history_variables` defined in `c_typecheck_expr.cpp`

regression/contracts/loop-history-variables-04/test.desc
```diff
@@ -0,0 +1,12 @@
...
+main.c
+--apply-loop-contracts
+^EXIT=0$
+^SIGNAL=0$
+^\[main.1\] .* Check loop invariant before entry: SUCCESS$
+^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
+^VERIFICATION SUCCESSFUL$
+--
+--
+This test checks that struct objects are supported for
```
SaswatPadhi: 1. There are no `struct`s in the test case.
2. We are introducing too many tiny regression test files. Let's just have a single C file with multiple loops, each one checking a different kind of objects. The line number of the errors (whenever that happens) would indicate the root cause; we don't need separate files.

regression/contracts/loop-history-variables-03/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-all-contracts
+^PARSING ERROR$
+^EXIT=1$
+^SIGNAL=0$
+--
+--
+This test ensures that __CPROVER_loop_entry cannot be
+used for function contracts.
```
SaswatPadhi: But we are not checking that :) The C file is using `__CPROVER_old` in function contracts.

regression/contracts/loop-history-variables-03/main.c
```diff
@@ -0,0 +1,13 @@
+oid foo(int *x) __CPROVER_assigns(*x)
```
SaswatPadhi: `oid`?!

SaswatPadhi (changes requested):

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -3965,10 +3979,25 @@ void c_typecheck_baset::disallow_history_variables(const exprt &expr) const
   }
 
   if(expr.id() == ID_old)
+  {
+    error().source_location = expr.source_location();
+    error() << CPROVER_PREFIX "old expressions are not allowed" << eom;
```
feliperodri: Are not allowed where? Could you improve this error message?

regression/contracts/history-pointer-replace-03/test.desc
```diff
@@ -4,7 +4,7 @@ main.c
 ^EXIT=(1|64)$
 ^SIGNAL=0$
 ^CONVERSION ERROR$
-error: __CPROVER_old expressions are not allowed in __CPROVER_requires clauses
+error: __CPROVER_old expressions are not allowed
```
feliperodri: Not allowed where? Could you improve this error message?

regression/contracts/loop-history-variables-01/main.c
```diff
@@ -0,0 +1,48 @@
+#include <assert.h>
```
feliperodri: You need stdlib for malloc too.

regression/contracts/loop-history-variables-02/test.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main.c
+--apply-loop-contracts
+^CONVERSION ERROR$
```
feliperodri: Which error message shall we get in this case?

regression/contracts/loop-history-variables-03/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-all-contracts
+^CONVERSION ERROR$
```
feliperodri: Which error message shall we get in this case?

feliperodri (changes requested):
Almost there 😊

regression/contracts/history-pointer-replace-03/test.desc
SaswatPadhi: The line number is printed on the standard output, but we don't match it in the regex. Should we match the line number here?

src/ansi-c/c_typecheck_expr.cpp
SaswatPadhi: Since it's a common error message for disallowing `__CPROVER_old` within requires clauses and loop invariants, we shouldn't say "in __CPROVER_requires clauses" any more. May be we could say:

```
 error: __CPROVER_old expressions are not allowed in this context
```

Since we set the "source_location", the line number should also be printed and "in this context" would be clear from that.

SaswatPadhi (commented):
There is a bit of code duplication between `disallow_loop_history_variables` and `disallow_history_variables`.

Can we remove `disallow_loop_history_variables`, add an additional argument to `disallow_history_variables` which would be the `id` of the history variable expressions we want to disallow, and just reuse the code in both contexts?

SaswatPadhi (approved):
Thanks; LGTM modulo the build issues with CI.

feliperodri:
@aalok-thakkar CI failures are due to regex on test descriptions (not compatible with Windows). You also need to run clang-format.

regression/contracts/history-pointer-replace-03/test.desc
```diff
@@ -3,8 +3,8 @@ main.c
 --replace-all-calls-with-contracts
 ^EXIT=(1|64)$
 ^SIGNAL=0$
+^main.c:4:1: error: __CPROVER_old expressions are not allowed in this context$
```
feliperodri: Not compatible with Windows system.

regression/contracts/loop-history-variables-02/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--apply-loop-contracts
+^main.c:10:1: error: __CPROVER_old expressions are not allowed in this context$
```
feliperodri: Not compatible with Windows system.

regression/contracts/loop-history-variables-03/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--enforce-all-contracts
+^main.c:2:1: error: __CPROVER_loop_entry expressions are not allowed in this context$
```
feliperodri: Not compatible with Windows system.

feliperodri (approved):
LGTM modulo CI failures.

feliperodri (approved):

chrisr-diffblue (commented):
Code looks great, but I wonder if I might ask for a couple of extras:

1) Some more negative tests - at the moment the regression tests you've added all lead either to VERIFICATION SUCCESSFUL, or CONVERSION ERROR... Could we perhaps have some tests that also lead to verification failure? This might also help with my next point :-)
2) This is more optional... but it would be _super_ nice if we could also have a small bit of user documentation for this new predicate... indicating its purpose, how it should be used, etc.

I won't block the review just for missing user documentation, but I think we could do with the extra "negative test" before we merge.

regression/contracts/loop-history-variables-04/test.desc
```diff
@@ -0,0 +1,12 @@
...
+--apply-loop-contracts
+^EXIT=(10|6)$
+^SIGNAL=0$
+^\[main.1\] .* Check loop invariant before entry: SUCCESS$
+^\[main.2\] .* Check that loop invariant is preserved: SUCCESS$
+^\[main.assertion.1\] .* assertion x == z: FAILURE$
+^VERIFICATION FAILED$
+--
+--
+This test checks that __CPROVER_loop_entry is supported.
```
SaswatPadhi: How exactly are we checking that `__CPROVER_loop_entry` is supported? It doesn't appear in the the C program ...
The parsing test cases already check for support. The negative tests should be checking for assertion failures (see my comment above).

SaswatPadhi (changes requested):
Thanks @chrisr-diffblue!

@aalok-thakkar I don't understand the new test. I think by negative tests Chris was suggesting that we include a test where it would be violated, and we check that CBMC correctly catches it. An example would be:

```c
while(y > 0)
    __CPROVER_loop_invariant(x == __CPROVER_loop_entry(x))
    {
      --y;
      x += 1;
      x -= 2;
    }
}
```

SaswatPadhi (approved):
LGTM. Thanks for all the fixes, @aalok-thakkar :)

feliperodri:
@chrisr-diffblue could you take another look?

regression/contracts/loop-history-variables-01/main.c
```diff
@@ -0,0 +1,49 @@
...
+
+typedef struct
+{
+  int *n;
+} s;
+
+void main()
+{
+  int *x1, y1, z1;
+  *x1 = &z1;
```
SaswatPadhi:
```suggestion
  x1 = &z1;
```

regression/contracts/loop-history-variables-02/main.c
```diff
@@ -0,0 +1,17 @@
+#include <assert.h>
+
+void main()
+{
+  int *x, y, z;
+
+  *x = &z;
```
SaswatPadhi:
```suggestion
  x = &z;
```

SaswatPadhi (changes requested):
Some issues with pointers.

(They were not caught because the tests are not run with `--pointer-check`)

SaswatPadhi:
The CI failures are due to `.desc` files being out of sync. You should also update `*x` -> `x` in those files.

Minor comment on the commit description: please change `a new predicate __CPROVER_loop_entry` -> `a new unary expression __CPROVER_loop_entry`. `__CPROVER_loop_entry` is not a predicate (a `bool`-returning function).

chrisr-diffblue (approved):
Approved, pending fixing the issue @SaswatPadhi has mentioned with --pointer-check

SaswatPadhi (approved):

Issue #6238, opened 2021-07-20
Add flag to check pointer validity only for those pointer primitives that lack defined semantics for invalid pointers


jimgrundy:
CBMC version: 5.34.0
Operating system: N/A

The CBMC Manual section on “Memory Primitives / Detecting potential misuses of memory primitives” (http://cprover.diffblue.com/memory-primitives.html#autotoc_md108) ends with the following text:

"While the first three primitives [__CPROVER_POINTER_OBJECT, __CPROVER_POINTER_OFFSET, and __CPROVER_same_object]  have well-defined semantics even on invalid pointers, using them on invalid pointers is usually unintended in user programs. Thus, they have been included in the --pointer-primitive-check option."

We are encountering various cases where it is our intention to use these primitives on invalid pointers — specifically as assumptions about uninitialized pointers so as to constrain them. For example, we may malloc one pointer p and then seek to constrain a second pointer q to point into the same region as p.

To facilitate this without abandoning the pointer primitive checks entirely we need the addition of a new flag that causes CBMC to check for valid pointer use only on those primitives where behavior on invalid pointers is undefined. One possible name might be --lax-pointer-primitive-check. There are probably better ones.

It should be possible to use both this new flag (--lax-pointer-primitive-check) and the existing flag (--pointer-primitive-check) such that a #pragma can be used to disable just the stricter checks on a region of code.

feliperodri:
We need to be careful while using these primitives in assumptions (see https://github.com/diffblue/cbmc/issues/6057). Instead of creating a new flag, we might want to consider updating the behavior of these primitives to properly handle invalid pointers (see the discussion https://github.com/diffblue/cbmc/issues/6045#issuecomment-824334506). For instance, consider the following code snippet:
```c
int *a = malloc(sizeof(*a));
int *b;
__CPROVER_assume(__CPROVER_same_object(a, b));
assert(__CPROVER_r_ok(a, sizeof(*a)));
assert(__CPROVER_r_ok(b, sizeof(*b)));
```
I'd assume that without the pointer primitive check, both assertions should be valid.

feliperodri:
cc. @SaswatPadhi

jimgrundy:
[...]

PR #6279, opened 2021-08-08
Skip checking frame conditions for local variables

### LINKED ISSUE #6154: Assigns clause should only check assignments over variables from the write set
**CBMC version:** `develop` branch.
**Operating system:** N/A.
**Exact command line resulting in the issue:** N/A.
**What behaviour did you expect:** We only want to check assignments over variables from the write set. Local variables should be ignored.
What happened instead: When using assigns clause, CBMC checks for all assignments, even if they are over local variables.

feliperodri:
Resolves #6154 

It avoids unnecessary assertions (improving performance) and it prevents cases where memory primitives are invoked with dead objects.

### DIFF
--- a/regression/contracts/assigns_enforce_17/test.desc
+++ b/regression/contracts/assigns_enforce_17/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+--enforce-all-contracts
+^EXIT=0$
+^SIGNAL=0$
+^\[main.assertion.\d+\] line \d+ assertion x \=\= 0\: SUCCESS$
+^VERIFICATION SUCCESSFUL$
+--
+--
+Checks whether verification correctly distincts local variables
+and global variables with same name when checking frame conditions.
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -675,14 +675,19 @@ void code_contractst::instrument_assign_statement(
 
   const exprt &lhs = instruction_iterator->assign_lhs();
 
-  goto_programt alias_assertion;
-  alias_assertion.add(goto_programt::make_assertion(
-    assigns_clause.alias_expression(lhs),
-    instruction_iterator->source_location));
-  alias_assertion.instructions.back().source_location.set_comment(
-    "Check that " + from_expr(ns, lhs.id(), lhs) + " is assignable");
-  insert_before_swap_and_advance(
-    program, instruction_iterator, alias_assertion);
+  if(
+    freely_assignable_symbols.find(lhs.get(ID_identifier)) ==
+    freely_assignable_symbols.end())
+  {
+    goto_programt alias_assertion;
+    alias_assertion.add(goto_programt::make_assertion(
+      assigns_clause.alias_expression(lhs),
+      instruction_iterator->source_location));
+    alias_assertion.instructions.back().source_location.set_comment(
+      "Check that " + from_expr(ns, lhs.id(), lhs) + " is assignable");
+    insert_before_swap_and_advance(
+      program, instruction_iterator, alias_assertion);
+  }
 }
 
 void code_contractst::instrument_call_statement(
@@ -894,14 +899,6 @@ void code_contractst::check_frame_conditions(
 
   // Create a list of variables that are okay to assign.
   std::set<irep_idt> freely_assignable_symbols;
-  // Add all parameters that are not pointers to the freely assignable set
-  for(code_typet::parametert param : type.parameters())
-  {
-    if(param.type().id() != ID_pointer)
-    {
-      freely_assignable_symbols.insert(param.get_identifier());
-    }
-  }
 
   // Create temporary variables to hold the assigns
   // clause targets before they can be modified.
@@ -917,6 +914,7 @@ void code_contractst::check_frame_conditions(
   {
     if(instruction_it->is_decl())
     {
+      // Local variables are always freely assignable
       freely_assignable_symbols.insert(
         instruction_it->get_decl().symbol().get_identifier());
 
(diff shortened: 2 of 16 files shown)


src/goto-instrument/contracts/contracts.cpp
```diff
@@ -665,7 +665,7 @@ void code_contractst::instrument_assign_statement(
   goto_programt::instructionst::iterator &instruction_iterator,
   goto_programt &program,
   exprt &assigns,
-  std::set<irep_idt> &freely_assignable_symbols,
+  std::set<irep_idt> &local_assignable_symbols,
```
SaswatPadhi: I think the old name was more intuitive. Freely assignable would indicate that these things can be assigned to without any checks.

When we insert local variables into this set, we can leave a comment that local variables are always freely assignable.

SaswatPadhi (changes requested):
Just one comment below.

Other than that, can we add a test case where we assign to a local variable with the same name as the global variable and make sure that things work correctly? Something like:

```c
int x;
void pure()
__CPROVER_assigns()
{
  int x;
  x++;
}
```

regression/contracts/chain.sh
```diff
@@ -44,4 +44,4 @@ elif echo $args_inst | grep -q -- "--dump-c" ; then
   rm "${name}-mod.c"
 fi
 $goto_instrument --show-goto-functions "${name}-mod.gb"
-$cbmc "${name}-mod.gb" ${args_cbmc}
+$cbmc "${name}-mod.gb" ${args_cbmc} --pointer-primitive-check
```
SaswatPadhi: I think this would make it impossible to test certain behaviors on invalid pointers. Can we add this to the specific test that requires it? We can do so by adding `_ --pointer-primitive-check` to the list of flags (note the underscore). For example:

https://github.com/diffblue/cbmc/blob/61fd24de393b45c7f07f634ca162aa06c3974c42/regression/contracts/variant_init_inside_loop/test.desc#L3

src/goto-instrument/contracts/contracts.cpp
feliperodri: Implemented as suggested.

regression/contracts/chain.sh
feliperodri: I moved the flag to specific test cases. However, which behaviors would we want to test on invalid pointers? If CPROVER API says we can't use these primitives with invalid pointers, we should only use function contracts with `--pointer-primitive-check`, right?

feliperodri:
> Other than that, can we add a test case where we assign to a local variable with the same name as the global variable and make sure that things work correctly?

Done! Could you take another look? @SaswatPadhi

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -396,7 +396,7 @@ void code_contractst::replace_old_parameter(
 
     if(
       parameter.id() == ID_dereference || parameter.id() == ID_member ||
-      parameter.id() == ID_symbol)
+      parameter.id() == ID_symbol || parameter.id() == ID_ptrmember)
```
feliperodri, line 399: @SaswatPadhi to address your comment from https://github.com/diffblue/cbmc/pull/6277#discussion_r684679524

SaswatPadhi: Thanks! As you mentioned on the other PR, everything still works even with just `ID_member`, but I'm still not sure how. Let's leave `ID_ptrmember` in here as well. Do we allow these 4 cases in assigns clause as well?

regression/contracts/assigns_enforce_17/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--enforce-all-contracts _ --pointer-primitive-check
```
SaswatPadhi: For this particular test case, we don't really need pointer primitive check right? I mean there are no pointers in the program, so we can remove this check.

src/goto-instrument/contracts/contracts.cpp
feliperodri: > Do we allow these 4 cases in assigns clause as well?

Yes, we do.

https://github.com/diffblue/cbmc/blob/a7bfc101b9ad1b95c755a368d6eda84317e565f5/src/ansi-c/ansi_c_convert_type.cpp#L251-L266

SaswatPadhi (approved):
Thanks for adding the new test case.

LGTM modulo a minor comment below.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Perfect! Thanks

regression/contracts/chain.sh
SaswatPadhi: We may not specifically be checking any behavior for invalid pointers (although we might want to, in some form, in the future, see: #6238), but within the "contracts" regression suite, we shouldn't necessarily have to always create allocated pointers. Some property checks work just fine with uninitialized pointers and that might be sufficient for some cases.

One last question I have is regarding the need for this check. The memory primitives documentation (http://cprover.diffblue.com/memory-primitives.html) says that the check is necessary only for memory primitives. We don't seem to use them in the test cases directly. Do we generate them during instrumentation?

regression/contracts/assigns_enforce_17/test.desc
feliperodri: Good catch. We won't generate any assertions to check frame conditions to this program.

regression/contracts/chain.sh
feliperodri: > Do we generate them during instrumentation?

Yes, we do. We rely on them to check frame conditions. We use `__CPROVER_same_object`, `__CPROVER_POINTER_OFFSET`, and `__CPROVER_OBJECT_SIZE`. For instance, take a look at this

https://github.com/diffblue/cbmc/blob/b60259a9fc00a1c045d5585e4ead7639911afe31/src/goto-instrument/contracts/assigns.cpp#L68-L69

PR #6282, opened 2021-08-09
VSD - neater clone member function.

jezhiggins:
The VSD `abstract_objectt` hierarchy has a polymorphic `mutable_clone` member function, which abstract object's use when they need to return an updated copy of themselves. The member function declaration is 

  `virtual std::shared_ptr<abstract_objectt> mutable_clone() const`
 
Because `mutable_clone` returns a shared pointer to the base type, we almost always need to cast the return value, typically something like 

  `auto result = std::dynamic_pointer_cast<data_dependency_contextt>(mutable_clone());`

This is, frankly, a bit of a mouthful. 

This PR is pure refactoring which adds a non-virtual template member function to wrap up the pointer cast into a neater package. We can now write 

  `auto result = clone(this);`

and result will still have the correct shared pointer type. It's not absolutely ideal (if I could think of a way to eliminate the `this` parameter, believe me I'd take it), but it's much, much easier on the eye.

### DIFF
--- a/src/analyses/variable-sensitivity/abstract_object.cpp
+++ b/src/analyses/variable-sensitivity/abstract_object.cpp
@@ -62,7 +62,7 @@ abstract_object_pointert abstract_objectt::abstract_object_merge(
   if(is_top() || other->bottom)
     return this->abstract_object_merge_internal(other);
 
-  internal_abstract_object_pointert merged = mutable_clone();
+  auto merged = mutable_clone();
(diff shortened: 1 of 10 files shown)


martin-cs (approved):

jezhiggins:
The rebase is horrendous and there's genuine issue with the data-dependencies. I'll re-raise as a clean PR.

Issue #618, opened 2017-03-09
Idea : Connecting up producers and consumers of knowledge about termination


martin-cs:
1. There are various things we do that can be improved by having even crude information about loop and function termination and non-termination.

2. For example, I think that slicing can be improved with a small amount of information about termination so that irrelevant loops and functions can be removed.  This can also remove control dependencies on loop control variables.  In the past when I have used slicing this was one of the key limitations. @marek-trtik and @lucasccordeiro might care.

3. Another example, in goto-analyze, we can only show that an assertion is false "if reachable", termination and non-termination can give us a crude reachability analysis which may be enough to give a definitive false in some cases.  @danpoe  and @thk123 might care.

4. Various people have termination and non-termination analyses of varying weights in various branches.  I know @cristina-david  and @pkesseli  have some very cool things.  I know @peterschrammel  and @steffan711 are building something.  I know @danpoe has built a loop bound computation in goto-analyze.  I wouldn't be surprised if @tautschnig has something up his sleeve.

5. Could we provide an API for accessing / using this information:

loop_id -> TERMINATE & NON_TERMINATE & UNKNOWN
function_name -> TERMINATE & NON_TERMINATE & UNKNOWN

(i.e. a bit-wise combination of which cases are possible)

with a trivial default implementation (for loops, check if the guard is false, for functions check all loops) and then have a task to connect up more sophisticated analyses as they are merge-able / in branches that they exist?

What do people think?

mgudemann:
I think it's a good idea.

Not sure if that counts, but we have `--java-unwind-enum-static` in CBMC/Java which unrolls any loop it finds in the static initalization context of Java enumerations. The unwind parameter `k` is exactly the number of elements in the enumeration. The reasoning is that as there's no input parameter to `<clinit>`, the unwinding that makes most sense is `k`.

This is a heuristic, but could be worth a try to assume `<clinit>` loops to terminate in presence of the above option. In particular as enums with user-defined static init are often used in the Java benchmarks we have looked at.

martin-cs:
1. The unwind analysis in goto-analyze should be able to do this
soundly.

2. Maybe the interface should be extended to also return a bound on the
number of iterations if it returns 
TERMINATE | !NON_TERMINATE | !UNKNOWN

tautschnig:
Couldn't we use custom_bit_vector_domaint for this purpose?

martin-cs:
[...]

Issue #265, opened 2016-10-24
Feature request: --cover array


martin-cs:
Can we have an additional coverage instrumentation that adds coverage goals for each possible entry of a fixed-size array when an instruction reads from the array:

```
int lookup [] = {0, 1, 2, 3, 4, 5, 6, 7};

int interpolate (int x) {
    int target = find_address(x);

    __CPROVER_cover(target == 0);
    __CPROVER_cover(target == 1);
    __CPROVER_cover(target == 2);
    __CPROVER_cover(target == 3);
    __CPROVER_cover(target == 4);
    __CPROVER_cover(target == 5);
    __CPROVER_cover(target == 6);
    __CPROVER_cover(target == 7);

    return computeValue(x, lookup[target]);
}
```

Requested by an industrial partner.

TGWDB:
Closing due to detail in #6284

Issue #218, opened 2016-09-06
Function summaries


trtikm:
Main design goals of:
- context insensitivity (there is one summary applicable in any context the function is called from)
- transitive closure (effects of all callees are captured in summaries of the root caller function)
- generalised interface (suitable for many different concrete summary-computing algorithms)
- language independent (should work for Java, C, C++, ...)

Known links to other tasks/projects:
- test_gen (our summaries should fit to needs of this tests generation task)
- 2ls (their summaries should also be expressible in our summaries)
- path-symex (summaries must also be usable for path-based symbolic execution)

Method:
   I'll iteratively improve the structure of the summaries. I'll start with a simple implementation tested on few hand-written benchmarks and then I'll discuss the design of summaries with others (mainly from the related projects) to get a feedback. In each subsequent iteration I'll update the implementation according the the feedback from the previous iteration. I expect two or three iterations until the summaries will be ready to tackle real-world programs. An iteration may approximately take one week.

martin-cs:
On Tue, 2016-09-06 at 10:07 -0700, Marek Trtik wrote:
> Main design goals of:
>   - context insensitivity (there is one summary applicable in any context the function is called from)
>   - transitive closure (effects of all callees are captured in summaries of the root caller function)
>   - generalised interface (suitable for many different concrete summary-computing algorithms)
>   - language independent (should work for Java, C, C++, ...)
> 
> Known links to other tasks/projects:
>   - test_gen (our summaries should fit to needs of this tests generation task)
>   - 2ls (their summaries should also be expressible in our summaries)
>   - path-symex (summaries must also be usable for path-based symbolic execution)
>   
> Method:
>    I'll iteratively improve the structure of the summaries. I'll start with a simple implementation tested on few hand-written benchmarks and then I'll discuss the design of summaries with others (mainly from the related projects) to get a feedback. In each subsequent iteration I'll update the implementation according the the feedback from the previous iteration. I expect two or three iterations until the summaries will be ready to tackle real-world programs. An iteration may approximately take one week.

Are the summaries going to be over or under approximate?  2LS will
likely need over-approximate, while path-symex will need
under-approximate.

Cheers,
 - Martin

martin-cs:
Would this make use of our loop acceleration code?

TGWDB:
Closing due to detail in #6284

Issue #40, opened 2016-03-29
Advanced goto-diff


peterschrammel:
syntactic and semantic diffing of goto programs

peterschrammel:
basic syntactic diffing in pull request #65

peterschrammel:
integrate
- deltacheck's change impact analysis
- cemc equivalence checker

TGWDB:
Closing due to detail in #6284

PR #6284, opened 2021-08-10
Add FEATURE_IDEAS.md

### LINKED ISSUE #28: Refinement-based slicing
patches exist, but performance is not yet convincing

TGWDB:
This adds FEATURE_IDEAS.md to gather together new features and research ideas
that otherwise would fill the issues/PR backlog. Also provide a simple
reference point for finding feature ideas or projects.

This now also combines the MINI_PROJECTS.md file into this one so there is a single location for all these kinds of ideas and projects.

Note that this would replace/close #28 #29 #40 #218 #265 #618  (and others as they are transitioned here).


peterschrammel (approved):
I'd suggest to call the file `FEATURE_IDEAS.md` to make it clearer what its content is.

martin-cs (commented):
I would rather not do things this way.  There is already a mini-projects list:

https://github.com/diffblue/cbmc/pull/3930

which does get occasional updates:

https://github.com/diffblue/cbmc/pull/4119

It would be better to add to that than start a new file.  However despite being the original author of many of these I think that issues are a much better way of tracking this sort of thing.  They can be cross-referenced, have clearer updates from multiple people, integrate with the rest of how we work and have a much lower barrier to entry.

If this is a "number of open issues" issue then ... maybe a tag for "feature ideas" which can be excluded from the count?

martin-cs:
( The content is good though, as is making this sort of thing better recorded. )

chrisr-diffblue (approved):

PR #6286, opened 2021-08-11
Static Verifier: Status should be UNKNOWN when there are true and false reachable traces

jezhiggins:
Consider 
```
#include <assert.h>
int global;

int main(void) {
  int local;
  unsigned int nondet;

  local = 1;
  global = 1;

  if (nondet == 0)
    local = 2;
  if (nondet == 0)
    global = 2;

  assert(local == global);
}

```
Running `goto-analyzer wrong.c --verify --recursive-interprocedural --vsd  --loop-unwind-and-branching 10` gives 
```
[main.assertion.1] line 21 assertion local == global: FAILURE (if reachable)
```
which is, to put it kindly, an overapproximation.

However running ` `goto-analyzer wrong.c --verify --recursive-interprocedural --vsd` gives 
```
[main.assertion.1] line 21 assertion local == global: UNKNOWN
```
which is as correct as that configuration can give. 

There are four paths through the code. Two are safe (take both ifs, take neither if), but the two take-one-if paths (which are spurious) will definitely fail. In an ideal world they'd be blocked, but we are not yet in that world. 

The static verifier does not account for this "might be a red path but that doesn't mean it fails because that path could be an over-approximation" and returned a FAILURE when an UNKNOWN would be more appropriate. 

(In the test cases updated in this PR, there are existing comments articulating this same issue.)

### DIFF
--- a/regression/goto-analyzer/branching_history_01/main.c
+++ b/regression/goto-analyzer/branching_history_01/main.c
@@ -1,0 +1,23 @@
+#include <assert.h>
+
+int global;
+
+int main(void)
+{
+  int local;
+  unsigned int nondet;
+
+  local = 1;
+  global = 1;
+
+  if(nondet == 0)
+  {
+    local = 2;
+  }
+  if(nondet == 0)
+  {
+    global = 2;
+  }
+
+  assert(local == global);
+}
--- a/regression/goto-analyzer/branching_history_01/test-no-unwind.desc
+++ b/regression/goto-analyzer/branching_history_01/test-no-unwind.desc
@@ -1,0 +1,7 @@
+CORE
+main.c
+--verify --recursive-interprocedural --vsd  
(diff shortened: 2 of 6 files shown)


martin-cs (approved):
Looks good; thanks.

martin-cs:
I think the drop in the number of lines covered is because we have reformatted and so removed some lines.  I am going to ignore it for the purpose of merging.

jezhiggins:
> I think the drop in the number of lines covered is because we have reformatted and so removed some lines. I am going to ignore it for the purpose of merging.

Yes, it's entirely an artefact of having applied clang-format-7 to static_verifier.cpp https://github.com/diffblue/cbmc/pull/6286/files#diff-da2927e58d056b5906c1ed62c21527e7db528c499359547ddaa6d7b86f680c14

PR #6288, opened 2021-08-11
Introduce floatbv_rounding_mode(unsigned)

kroening:
This introduces a function to produce the constant expression that is
consumed by the floatbv_* expressions as the rounding mode.

Using this function in `ieee_floatt` removes the dependency on c_types.h.

### DIFF
--- a/src/util/floatbv_expr.cpp
+++ b/src/util/floatbv_expr.cpp
@@ -1,0 +1,20 @@
+/*******************************************************************\
+
+Module: API to expression classes for floating-point arithmetic
+
+Author: Daniel Kroening, kroening@kroening.com
+
+\*******************************************************************/
+
+#include "floatbv_expr.h"
+
+#include "arith_tools.h"
+#include "bitvector_types.h"
+
+constant_exprt floatbv_rounding_mode(unsigned rm)
+{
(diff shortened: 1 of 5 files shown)


src/util/floatbv_expr.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+#include "floatbv_expr.h"
+
+#include "arith_tools.h"
+#include "bitvector_types.h"
+
+constant_exprt floatbv_rounding_mode(unsigned rm)
+{
+  // the 32 bits are an arbitrary choice
+  return ::from_integer(rm, unsignedbv_typet(32));
```
tautschnig: ~~Shouldn't the comment also clarify that the type needs to be the same as that of the `__CPROVER_rounding_mode` symbol (which, actually, seems to be a signed int)?~~ Edit: no, it doesn't matter, `float_utilst::set_rounding_mode` will create constants of the appropriate size.

tautschnig (approved):
I'd prefer using an integer of minimal size of a fixed-32-bits one, but likely this won't matter much.

tautschnig:
@kroening I believe tests would pass after a rebase.

peterschrammel (approved):

PR #6290, opened 2021-08-11
VSD - constant expression evaluation speedup

jezhiggins:
This PR reduces the number of times the operands in a constant_value expression are reevaluated. Usually, this overhead is small but for nested expressions this can result in excessive runtimes. 

The motivating example has the form 
```
(
 (l >= mh.s ? 1 : 0)
&(
  (
   (mh.s >= s ? 1 : 0) 
  &(
    (
     (n >= mh.t ? 1 : 0) 
    &(
      (
       (mh.t >= o ? 1 : 0) 
      &(
        (
         ((int)mh.to == (int)nt ? 0 : 1) 
        &((int)mh.su == (int)ns ? 0 : 1)
        ) == 0 ? 0 : 1
       )
      ) == 0 ? 0 : 1
     )
    ) == 0 ? 0 : 1
   )
  ) == 0 ? 0 : 1
 )
) == 0
```
This doesn't look particularly difficult, but evaluation time for the goto-program in question was close to two hours. With these changes evaluation time is approximately two tenths of a second (usual caveats - as reported by `time` on my machine, etc).

### DIFF
--- a/regression/goto-analyzer-simplify/simplify-deeply-nested-expression/test.desc
+++ b/regression/goto-analyzer-simplify/simplify-deeply-nested-expression/test.desc
@@ -1,0 +1,10 @@
+CORE
+main.c
+--vsd
+^SIGNAL=0$
+^EXIT=0$
+--
+--
+This expression is derived from a real system under test. Due to inadvertent reevaluation of expression operands, all expression evaluations had geometric complexity (even more so in the presence of TOP values). This particular expression took nearly two hours to evaluate. Reworking to prevent repeated reevaluation brings this down to a few tenths.
+See https://github.com/diffblue/cbmc/pull/6290
+If this test appears to hang, the constant_evaluator is probably your first port of call.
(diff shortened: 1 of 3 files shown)


martin-cs (approved):
Looks good so far.  A test case would be good.

regression/goto-analyzer-simplify/simplify-deeply-nested-expression/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--vsd
+^SIGNAL=0$
+^EXIT=0$
+--
+--
+This expression is derived from a real system under test. Due to inadvertent reevaluation of expression operands, all expression evaluations had geometric complexity (even more so in the presence of TOP values). This particular expression took nearly two hours to evaluate. Reworking to prevent repeated reevaluation brings this down to a few tenths.
+See https://github.com/diffblue/cbmc/pull/6290
+If this test appears to hang, the constant_evaluator is probably your first port of call.
```
martin-cs, line 10: :heart:

martin-cs (approved):
Great; thank you!

PR #6292, opened 2021-08-12
Unify havocing codegen within code contracts

### LINKED ISSUE #6235: Replace with contracts - havoc entire array content
CBMC version: develop (current 06c563a76589ed00cc372d68e2819880537977b2)

Operating system: N/A.

`goto-cc *.c -o main.gb && goto-instrument main.gb main-mod.gb --replace-all-calls-with-contracts && cbmc main-mod.gb`

```c
#include <assert.h>

void foo(char c[]) __CPROVER_assigns(*c)
{
}

int main()
{
  char b[2];
  b[0] = 'a';
  b[1] = 'b';
  foo(b);
  assert(b[0] == 'a');
  assert(b[1] == 'b');
  return 0;
}
```

```shell
** Results:
main.c function main
[main.assertion.1] line 13 assertion b[0] == 'a': FAILURE
[main.assertion.2] line 14 assertion b[1] == 'b': FAILURE
```

What happened instead:
```shell
** Results:
main.c function main
[main.assertion.1] line 13 assertion b[0] == 'a': FAILURE
[main.assertion.2] line 14 assertion b[1] == 'b': SUCCESS
```

We need to havoc all contents of an array.

SaswatPadhi:
Resolves #6235.

In this PR, we unify the havocing code generation for both loop & function contracts.

This fixes multiple known issues:
1. (function contracts) entire arrays were not being havoced
2. (loop contracts) pointers not checked for nullness before havocing

- n/a ~The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/~
- n/a ~My commit message includes data points confirming performance improvements (if claimed).~

### DIFF
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -13,6 +13,8 @@ Date: July 2021
 
 #include "assigns.h"
 
+#include <goto-instrument/havoc_utils.h>
+
 #include <util/arith_tools.h>
 #include <util/c_types.h>
 #include <util/pointer_predicates.h>
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -106,22 +106,6 @@ exprt assigns_clause_targett::compatible_expression(
   return same_object(called_target.get_direct_pointer(), target);
 }
 
-goto_programt assigns_clause_targett::havoc_code() const
-{
-  goto_programt assigns_havoc;
-  source_locationt location = pointer_object.source_location();
-
-  exprt lhs = dereference_exprt(pointer_object);
-  side_effect_expr_nondett rhs(lhs.type(), location);
-
-  goto_programt::targett target =
-    assigns_havoc.add(goto_programt::make_assignment(
-      code_assignt(std::move(lhs), std::move(rhs)), location));
-  target->code_nonconst().add_source_location() = location;
-
-  return assigns_havoc;
-}
-
 const exprt &assigns_clause_targett::get_direct_pointer() const
 {
   return pointer_object;
@@ -199,47 +183,12 @@ goto_programt assigns_clauset::dead_stmts()
 
 goto_programt assigns_clauset::havoc_code()
 {
+  modifiest modifies;
+  for(const auto &t : targets)
+    modifies.insert(to_address_of_expr(t->get_direct_pointer()).object());
+
   goto_programt havoc_statements;
-  source_locationt location = assigns.source_location();
-
-  for(assigns_clause_targett *target : targets)
-  {
-    // (1) If the assigned target is not a dereference,
-    // only include the havoc_statement
-
-    // (2) If the assigned target is a dereference, do the following:
-
-    // if(!__CPROVER_w_ok(target, 0)) goto z;
-    //      havoc_statements
-    // z: skip
-
-    // create the z label
-    goto_programt tmp_z;
-    goto_programt::targett z = tmp_z.add(goto_programt::make_skip(location));
-
-    const auto &target_ptr = target->get_direct_pointer();
-
-    if(to_address_of_expr(target_ptr).object().id() == ID_dereference)
-    {
-      // create the condition
-      exprt condition =
-        not_exprt(w_ok_exprt(target_ptr, from_integer(0, unsigned_int_type())));
-      havoc_statements.add(goto_programt::make_goto(z, condition, location));
-    }
-
-    // create havoc_statements
-    for(goto_programt::instructiont instruction :
-        target->havoc_code().instructions)
-    {
-      havoc_statements.add(std::move(instruction));
-    }
-
-    if(to_address_of_expr(target_ptr).object().id() == ID_dereference)
-    {
-      // add the z label instruction
-      havoc_statements.destructive_append(tmp_z);
-    }
-  }
+  append_havoc_code(assigns.source_location(), modifies, havoc_statements);
   return havoc_statements;
 }
 
--- a/src/goto-instrument/havoc_utils.cpp
+++ b/src/goto-instrument/havoc_utils.cpp
@@ -1,0 +1,62 @@
+/*******************************************************************\
+
+Module: Utilities for building havoc code for expressions.
+
+Author: Saswat Padhi, saspadhi@amazon.com
+
+Date: July 2021
+
+\*******************************************************************/
+
+/// \file
+/// Utilities for building havoc code for expressions
+
+#include "havoc_utils.h"
+
+#include <util/expr_util.h>
+#include <util/pointer_expr.h>
+
+void append_object_havoc_code_for_expr(
+  const source_locationt source_location,
+  const exprt &expr,
+  goto_programt &dest)
+{
+  codet havoc(ID_havoc_object);
+  havoc.add_source_location() = source_location;
+  havoc.add_to_operands(expr);
+  dest.add(goto_programt::make_other(havoc, source_location));
+}
+
+void append_scalar_havoc_code_for_expr(
+  const source_locationt source_location,
+  const exprt &expr,
+  goto_programt &dest)
+{
+  side_effect_expr_nondett rhs(expr.type(), source_location);
+  goto_programt::targett t = dest.add(
+    goto_programt::make_assignment(expr, std::move(rhs), source_location));
+  t->code_nonconst().add_source_location() = source_location;
+}
+
+void append_havoc_code(
+  const source_locationt source_location,
+  const modifiest &modifies,
+  goto_programt &dest)
+{
+  havoc_utils_is_constantt is_constant(modifies);
+  for(const auto &expr : modifies)
+  {
+    if(expr.id() == ID_index || expr.id() == ID_dereference)
+    {
+      address_of_exprt address_of_expr(expr);
+      if(!is_constant(address_of_expr))
+      {
+        append_object_havoc_code_for_expr(
+          source_location, address_of_expr, dest);
+        continue;
+      }
+    }
+
+    append_scalar_havoc_code_for_expr(source_location, expr, dest);
+  }
+}
--- a/src/goto-instrument/havoc_utils.cpp
+++ b/src/goto-instrument/havoc_utils.cpp
@@ -13,18 +13,48 @@ Date: July 2021
 
 #include "havoc_utils.h"
 
-#include <util/expr_util.h>
+#include <util/arith_tools.h>
+#include <util/c_types.h>
 #include <util/pointer_expr.h>
+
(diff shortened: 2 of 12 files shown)


src/goto-instrument/havoc_utils.h
```diff
@@ -0,0 +1,38 @@
...
+
+void append_object_havoc_code_for_expr(
+  const source_locationt source_location,
+  const exprt &expr,
+  goto_programt &dest);
+
+void append_scalar_havoc_code_for_expr(
+  const source_locationt source_location,
+  const exprt &expr,
+  goto_programt &dest);
```
feliperodri, line 59: Why the suffix "append"? Perhaps we can come up with more intuitive names. What do you think about `havoc_code`, `havoc_object`, and `havoc_scalar`?

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -197,47 +183,16 @@ goto_programt assigns_clauset::dead_stmts()
...
-    // if(!__CPROVER_w_ok(target, 0)) goto z;
-    //      havoc_statements
-    // z: skip
+  modifiest modifies;
+  std::for_each(
+    targets.begin(),
+    targets.end(),
+    [&modifies](const assigns_clause_targett *t) {
+      modifies.insert(to_address_of_expr(t->get_direct_pointer()).object());
+    });
```
feliperodri: Very clean 👏 nice!

src/goto-instrument/havoc_utils.cpp
```diff
@@ -0,0 +1,115 @@
...
+    // we also treat unmodified symbols as constants
+    if(expr.id() == ID_symbol && modifies.find(expr) == modifies.end())
+      return true;
+
+    return is_constantt::is_constant(expr);
+  }
+
+protected:
+  const modifiest &modifies;
+};
```
feliperodri: Why keep this class here instead of the .h file?

src/goto-instrument/havoc_utils.cpp
```diff
@@ -0,0 +1,115 @@
...
+      address_of_exprt address_of_lhs(lhs);
+      if(!is_constant(address_of_lhs))
+      {
+        append_object_havoc_code_for_expr(
+          source_location, address_of_lhs, dest);
+        continue;
+      }
+    }
+
+    append_scalar_havoc_code_for_expr(source_location, lhs, dest);
```
feliperodri: Is `lhs.source_location() != source_location`? If not, we wouldn't need the extra parameter.

feliperodri: Wouldn't an else branch be better than using `continue`?

src/goto-instrument/havoc_utils.cpp
```diff
@@ -0,0 +1,115 @@
...
+void append_object_havoc_code_for_expr(
+  const source_locationt source_location,
+  const exprt &expr,
+  goto_programt &dest)
+{
+  append_safe_havoc_code_for_expr(source_location, expr, dest, [&]() {
+    codet havoc(ID_havoc_object);
+    havoc.add_source_location() = source_location;
+    havoc.add_to_operands(expr);
+
```
feliperodri: no need for empty line.

src/goto-instrument/havoc_utils.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+void append_scalar_havoc_code_for_expr(
+  const source_locationt source_location,
+  const exprt &expr,
+  goto_programt &dest)
+{
+  append_safe_havoc_code_for_expr(source_location, expr, dest, [&]() {
+    side_effect_expr_nondett rhs(expr.type(), source_location);
+
```
feliperodri: no need for empty line.

src/goto-instrument/havoc_utils.cpp
```diff
@@ -0,0 +1,115 @@
...
+  });
+}
+
+void append_scalar_havoc_code_for_expr(
+  const source_locationt source_location,
+  const exprt &expr,
+  goto_programt &dest)
+{
+  append_safe_havoc_code_for_expr(source_location, expr, dest, [&]() {
+    side_effect_expr_nondett rhs(expr.type(), source_location);
```
feliperodri, line 66: Again, if `expr.source_location() == source_location`, we wouldn't need the extra param.

src/goto-instrument/havoc_utils.cpp
```diff
@@ -0,0 +1,115 @@
...
+    const auto condition = not_exprt(w_ok_exprt(
+      to_dereference_expr(expr).pointer(),
+      from_integer(0, unsigned_int_type())));
+    dest.add(goto_programt::make_goto(skip_target, condition, source_location));
+  }
+
+  havoc_code_impl();
+
+  if(expr.id() == ID_dereference)
+    dest.destructive_append(skip_program);
```
feliperodri, line 44: Could you add a comment for this special case?

feliperodri (changes requested):
Thanks for implementing this fix!! Just a few comments.

src/goto-instrument/havoc_utils.cpp
SaswatPadhi: Notice that we are invoking `continue` from within a nested-`if` structure. It's not straightforward to write a single `else` branch here, so I just used `continue`.

src/goto-instrument/havoc_utils.cpp
SaswatPadhi: Good point! I should move it to the .h file, yes.

src/goto-instrument/contracts/assigns.cpp
SaswatPadhi: 😄 Thanks!

While writing this transformation, I noticed that `targets` is a `vector`, unlike `modifies`, which is a `set`. If we write the same expression twice in the `__CPROVER_assigns` clause, would we end up havocing / enforcing it twice? Or, does the parser / typechecker somehow guard against this?

src/goto-instrument/havoc_utils.h
SaswatPadhi: The original prefix was `build` and it was inserting goto statements at the end of the `dest` goto program. I was always confused about whether it adds the statements at the beginning or at the end, so I just changed `build` -> `append`. What part of the name is unintuitive?

The names `havoc_code`, `havoc_object`, and `havoc_scalar` seem okay but lead to the same confusion as `build`. Also notice that `append_havoc_code` takes an entire `modifies` `set`, but the `append_*_havoc_code_for_expr` functions take `exprt`.

The editor autocompletes anyway, so I just use longer function names that are more explicit about their effect.

src/goto-instrument/havoc_utils.cpp
SaswatPadhi: For `expr` that comes from an annotation (like assigns clause) this would be fine, but for the auto-computed modifies set, `expr` would have arbitrary source locations within the loop body. This extra source location is currently used to override those locations and just show the location of the loop head for the auto-computed modifies set.

feliperodri: Thanks for the clarification!

src/goto-instrument/contracts/assigns.cpp
feliperodri: > If we write the same expression twice in the `__CPROVER_assigns` clause, would we end up havocing / enforcing it twice? Or, does the parser / typechecker somehow guard against this?

We don't have any regression tests that cover this behavior. Could you write a regression test that cover this scenario for both loop contracts and function contracts?

> While writing this transformation, I noticed that `targets` is a `vector`, unlike `modifies`, which is a `set`.

Could you turn `targets` into a `set` too? Regardless of our type-checking procedure, `set` would be a more appropriate data structure here.

These modifications might be part of a separate PR. I'm happy to implement it too if that's the case.

src/goto-instrument/havoc_utils.cpp
feliperodri: You're right. I was thinking in something like this.

```cpp
  for(const auto &expr : modifies)
  {
    if((expr.id() == ID_index || expr.id() == ID_dereference) && !is_constant(address_of_expr(expr)))
    {
        append_object_havoc_code_for_expr(
            source_location, address_of_expr(expr), dest);
    }
    else
    {
        append_scalar_havoc_code_for_expr(source_location, expr, dest);
    }
  }
```

But your approach seems cleaner.

src/goto-instrument/havoc_utils.h
feliperodri: This is not a blocker, since it is probably just a style matter. For me, clear and concise method names yields clear code. 

> I was always confused about whether it adds the statements at the beginning or at the end, so I just changed `build` -> `append`.

I'd add this information to a header comment in each method. This will also help us to build document for this file using Doxygen. Some editors also pick up this information. For instance, take a look at this method:
https://github.com/diffblue/cbmc/blob/0bbbe195ff7e9fe55c2275dda37457de307e91a9/src/goto-instrument/contracts/contracts.h#L74-L87

> Also notice that `append_havoc_code` takes an entire `modifies` set, but the `append_*_havoc_code_for_expr` functions take `exprt`.

This is the type of information that could easily go to a comment header or inferred just by looking at the parameters.

feliperodri (approved):
I don't have any more blocking comments. 🚀

src/goto-instrument/havoc_utils.cpp
SaswatPadhi: Yeah I would keep the current structure, just in case we later add more stuff to the inner conditional block.

src/goto-instrument/havoc_utils.h
SaswatPadhi: Thanks, I agree I should add some comments here regardless, but for now I think I will leave `append_*_havoc_code_for_expr` and `append_havoc_code`, unless there is something specific that's incorrect / confusing about these names.

> What do you think about `havoc_code`, `havoc_object`, and `havoc_scalar`?

I thought a bit more about these names, but `havoc_code` seems incorrect -- it doesn't havoc a `codet` but a `set<exprt>`.

I guess we can revisit this later when we add comments & documentation for our functions.

src/goto-instrument/contracts/assigns.cpp
SaswatPadhi: Currently this is not a concern for loops because we don't allow manual `assigns` clause annotation yet.

I would separate this refactor (`vector` -> `set`) into another PR though. I'll open an internal ticket to track it.

src/goto-instrument/havoc_utils.h
chrisr-diffblue: Personally, I like the `append_*_` naming :-)

chrisr-diffblue (approved):

PR #6293, opened 2021-08-12
VSD - max array length

jezhiggins:
To prevent large arrays getting unwieldy, add a configuration parameter that limits the number of entries they have. 

Writes beyond the maximum index are all written to the same location, merging with any previous writes. 

* Adds `--vsd-arrays smash` option, which sets the maximum size to 1. All writes to the array are merged, and all reads return the same value.
* Add `--vsd-arrays up-to-n-elements` option. Default n is 10, `--vsd-max-array-elements` options gives finer control. (I would have liked n to be an additional parameter to --vsd-arrays, but that would have required significant rework in the arguments parsing and it's just not worth it.)


src/analyses/variable-sensitivity/variable_sensitivity_configuration.h
```diff
@@ -51,18 +52,15 @@ struct vsd_configt
 
   flow_sensitivityt flow_sensitivity;
 
+  size_t maximum_array_index = std::numeric_limits<size_t>::max();
```
martin-cs: I presume this is pending future changes to option handling?

martin-cs (approved):

src/analyses/variable-sensitivity/variable_sensitivity_configuration.h
jezhiggins: Yes, that's right.

src/analyses/variable-sensitivity/abstract_environment.cpp
```diff
@@ -337,10 +337,9 @@ abstract_object_pointert abstract_environmentt::abstract_object_factory(
     type, top, bttm, e, environment, ns);
 }
 
-abstract_object_pointert abstract_environmentt::add_object_context(
-  const abstract_object_pointert &abstract_object) const
+const vsd_configt &abstract_environmentt::configuration() const
 {
-  return object_factory->wrap_with_context(abstract_object);
+  return object_factory->config();
 }
```
jezhiggins, line 343: Bit of a diff mismatch. `add_object_context` is redundant and was removed. `configuration` is a new member function, which we need :)

PR #6296, opened 2021-08-13
Ensure result of to_predicates is deterministic

tautschnig:
The order of entries in an unordered map would previously yield varying
results. This caused spurious unit test failures as unit tests perform
exact string matching on the output.

See https://github.com/diffblue/cbmc/pull/5740/checks?check_run_id=3321470809 for an example of a spuriously failing run.


tautschnig:
@jezhiggins You might care to take a look.

martin-cs (approved):
Sorry I missed this and thanks for the fix.

jezhiggins (approved):
Thanks!

PR #6300, opened 2021-08-16
VSD - Regression tests to capture behaviour of assertions

jezhiggins:
A set of regressions tests capturing the behaviour of assertions in the VSD, and thus the behaviour of `variable_sensitivity_domaint::ai_simplify`.

There are no code changes in this PR, just the regression tests. Probably should have been included in https://github.com/diffblue/cbmc/pull/6023 but commit was accidentally left in a little branch of its own.

### DIFF
--- a/regression/goto-analyzer/variable-sensitivity-asserts-02/test-intervals.desc
+++ b/regression/goto-analyzer/variable-sensitivity-asserts-02/test-intervals.desc
@@ -1,0 +1,14 @@
+CORE
+main.c
+--variable-sensitivity --vsd-values intervals --verify
+^EXIT=0$
+^SIGNAL=0$
+\[main.assertion.1\] line 13 x == y: SUCCESS
+\[main.assertion.2\] line 14 y == x: SUCCESS
+\[main.assertion.3\] line 15 !\(x == y\): FAILURE
+\[main.assertion.4\] line 16 !\(y == x\): FAILURE
+\[main.assertion.5\] line 17 x != y: FAILURE
+\[main.assertion.6\] line 18 y != x: FAILURE
+\[main.assertion.7\] line 19 !\(x != y\): SUCCESS
+\[main.assertion.8\] line 20 !\(y != x\): SUCCESS
+--
--- a/regression/goto-analyzer/variable-sensitivity-asserts-02/test-value-sets.desc
+++ b/regression/goto-analyzer/variable-sensitivity-asserts-02/test-value-sets.desc
@@ -1,0 +1,14 @@
+CORE
+main.c
+--variable-sensitivity --vsd-values set-of-constants --verify
+^EXIT=0$
+^SIGNAL=0$
+\[main.assertion.1\] line 13 x == y: UNKNOWN
+\[main.assertion.2\] line 14 y == x: UNKNOWN
+\[main.assertion.3\] line 15 !\(x == y\): UNKNOWN
+\[main.assertion.4\] line 16 !\(y == x\): UNKNOWN
+\[main.assertion.5\] line 17 x != y: UNKNOWN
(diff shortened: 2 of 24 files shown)


regression/goto-analyzer/variable-sensitivity-asserts-02/test-intervals.desc
```diff
@@ -0,0 +1,14 @@
+CORE
+main.c
+--variable-sensitivity --vsd-values intervals --verify
+^EXIT=0$
+^SIGNAL=0$
+\[main.assertion.1\] line 13 x == y: SUCCESS
```
martin-cs, line 6: I can't see why this can be marked as a success.  If non-det is 0 then x = 1, y = 3, if non-det is non-zero then x = 3, y = 1, in neither case does the assertion pass....

regression/goto-analyzer/variable-sensitivity-asserts-02/test-intervals.desc
```diff
@@ -0,0 +1,14 @@
...
+\[main.assertion.2\] line 14 y == x: SUCCESS
+\[main.assertion.3\] line 15 !\(x == y\): FAILURE
```
martin-cs, line 8: This should be a success...

regression/goto-analyzer/variable-sensitivity-asserts-02/test-value-sets.desc
```diff
@@ -0,0 +1,14 @@
+CORE
+main.c
+--variable-sensitivity --vsd-values set-of-constants --verify
+^EXIT=0$
+^SIGNAL=0$
+\[main.assertion.1\] line 13 x == y: UNKNOWN
```
martin-cs, line 6: These I agree with because both sets will be `{1, 3}` and you can't determine the equality or inequality of these.

martin-cs (changes requested):
Thanks for the tests, these illustrate some subtle and interesting behaviour.  There is one issue, which I think might be a bug in the handling of `assume` / `ai_simplify`.  Maybe equality of intervals has been used somewhere it shouldn't have been?  I.E. [1,3] == [1,3] but if x \in [1,3] and y \in [1,3] that does not mean x = y.

regression/goto-analyzer/variable-sensitivity-asserts-02/test-intervals.desc
jezhiggins: Evaluating the assertion calls down into `variable_sensitivity_domain::ai_simplify`. The meat of `variable_sensitivity_domain::ai_simplify` is 
```
  abstract_object_pointert res = abstract_state.eval(condition, ns);
  exprt c = res->to_constant();
```

For the condition `x == y` where both `x` and `y` are intervals, the call to `abstract_state.eval` ends up in `interval_evaluator::transform`, and the ultimately in `constant_interval_exprt::equal`. 

`constant_interval_exprt::equal` returns true because the intervals have the same upper and lower bounds. 

I understand your objection, but this test is, I believe, capturing pre-existing behaviour.

regression/goto-analyzer/variable-sensitivity-asserts-02/test-value-sets.desc
jezhiggins: Again this is existing behaviour. The UNKNOWN arises because we try all the combinations of the values in the sets, in this case, `1 == 1` `1 == 3` `3 == 1` `3 == 3`, and because we have both true and false expressions, the result is unknown.

martin-cs (approved):
As you note, these are pre-existing issues and the tests just highlight them.  So I am going to merge the tests and then separately raise an issue to correct these.

PR #6302, opened 2021-08-16
Added support for loop_old history variable.

aalok-thakkar:
1. Created a new predicate __CPROVER_loop_old which can be used in the loop invariant
   to keep track of the value of a variable at the start of the loop.
2. Modified the code contract to support this history variable.
3. Added regression tests.


src/goto-instrument/contracts/contracts.cpp
```diff
@@ -246,6 +279,18 @@ void code_contractst::check_apply_loop_contracts(
     goto_function.body.destructive_insert(std::next(loop_head), havoc_code);
   }
 
+  // Reassign 'loop_old' history variables before the loop body
+  std::map<exprt, exprt> parameter2history_loop_old_2;
+  replace_history_parameter(
```
SaswatPadhi, line 276: This doesn't look right.

1. It looks like you are replacing `__CPROVER_loop_old` within invariant_copy again, which would actually not replace anything, since those have been replaced before.

2. You should generate assignments here, not replacement. Basically, for each expression in `parameter2history_loop_old` (not `parameter2history_loop_old_2`, which shouldn't be needed) you should generate an assignment statement to update the stored value with the current value of the expression.

regression/contracts/loop-history-variables-04/main.c
```diff
@@ -0,0 +1,49 @@
...
+
+typedef struct
+{
+  int *n;
+} s;
+
+void main()
+{
+  int *x1, y1, z1;
+  *x1 = &z1;
```
SaswatPadhi:
```suggestion
  x1 = &z1;
```

SaswatPadhi (changes requested):
I think there is a major issue (commented below).

Adding more tests, where `__CPROVER_loop_old` would differ from `__CPROVER_loop_entry` would have helped catch the issue, so we should add those. For instance:

```c
  while(y1 > 0)
    __CPROVER_loop_invariant(
        *x1 + y1 == __CPROVER_loop_old(*x1) + __CPROVER_loop_old(y1)
    )
    {
      --y1;
      *x1 = *x1 + 1;
    }
```
Here `*x1` changes after each iteration, and so does `__CPROVER_loop_old(*x1)`, but `__CPROVER_loop_entry(*x1)` remains unchanged (= `z1`).

PR #6311, opened 2021-08-22
Fix applicability condition of simplify_byte_update

peterschrammel:
The condition is supposed to guard against a crash in `basic_string::replace` in [line 1896](https://github.com/diffblue/cbmc/compare/develop...peterschrammel:fix-simplify-byte-update?expand=1#diff-748b330f9bcee071535ec8d20336452d8cb0664d4172dbd3aa6cdd0fc72aeec9R1896), which it fails to do if the offsets in the condition and the argument to `replace` don't match.


TGWDB (approved):
Looks ok, but it would be nice if we didn't have a magic number.

TGWDB (approved):

peterschrammel:
> Looks ok, but it would be nice if we didn't have a magic number.

This will be done in https://github.com/diffblue/cbmc/pull/917

PR #6313, opened 2021-08-23
Initial documentation for loop contracts

SaswatPadhi:
- n/a ~~Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).~~

### DIFF
--- a/doc/cprover-manual/contracts-functions.md
+++ b/doc/cprover-manual/contracts-functions.md
@@ -2,16 +2,9 @@
 
 CBMC offers support for function contracts, which includes three basic clauses:
 _requires_, _ensures_, and _assigns_.
-These clauses formally describe the specification of a function. CBMC
-also provides a series of built-in constructs to be used with functions
+These clauses formally describe the specification of a function.
+CBMC also provides a series of built-in constructs to be used with functions
 contracts (e.g., _history variables_, _quantifiers_, and _memory predicates_).
-
-To learn more about contracts, take a look at the chapter [Design by
-Contract](http://se.inf.ethz.ch/~meyer/publications/old/dbc_chapter.pdf) from
-the book Object-Oriented Software Construction by Bertrand Meyer or read the
-notes [Contract-based
-Design](https://www.georgefairbanks.com/york-university-contract-based-design-2021)
-by George Fairbanks.
 
 ## Overview
 
@@ -157,4 +150,3 @@ program using contracts.
 - [Memory Predicates](contracts-memory-predicates.md)
 - [History Variables](contracts-history-variables.md)
 - [Quantifiers](contracts-quantifiers.md)
-- [Loop Contracts]()
--- a/doc/cprover-manual/contracts-functions.md
+++ b/doc/cprover-manual/contracts-functions.md
@@ -75,7 +75,7 @@ example, developers can use the built-in construct `__CPROVER_return_value`,
 which represents the return value of a function. As postconditions, one may list
 possible return values (in this case, either `SUCCESS` or `FAILURE`) as well as
 describe the main property of this function: if the function returns `SUCCESS`,
-then `*out` stores the result of `*a + *b`.  We can also check that the value in
+then `*out` stores the result of `a + b`.  We can also check that the value in
 `*out` will be preserved in case of failure by using `__CPROVER_old`, which
 refers to the value of a given object in the pre-state of a function (see
 [History Variables](contracts-history-variables.md) Section for details). Thus, for the `sum` function, the
--- a/doc/cprover-manual/contracts-invariant.md
+++ b/doc/cprover-manual/contracts-invariant.md
@@ -1,0 +1,152 @@
+# Invariant Clause
+
+An _invariant_ clause specifies a property that must be preserved
+by every iteration of a loop.
+In general a loop has infinitely many invariants.
+For instance, `true` is a trivial invariant that holds before the loop,
+and after each iteration of the loop (thus, inductive).
+However, `true` is rarely useful --
+it is an extremely imprecise abstraction of a loop,
+which is generally not _sufficient_ to prove any subsequent assertions.
+
+### Syntax
+
+An _invariant_ clause accepts a valid Boolean expression
+over the variables visible at the same scope as the loop.
+Additionally, [history variables] are also allowed within invariant clauses.
+
+```c
+__CPROVER_loop_invariant(bool cond)
+```
+
+Invariant clauses may be specified just after the loop guard.
+Multiple invariant clauses on the same loop are allowed,
+and is equivalent to a single invariant clause that is a conjunction of those clauses.
+A few examples are shown below.
+
+```c
+while(i < n)
+__CPROVER_loop_invariant(0 <= i && i <= n)
+{ ... }
+```
+
+```c
+for(int i = 0; i < n; ++i)
+__CPROVER_loop_invariant(0 <= i)
+__CPROVER_loop_invariant(i <= n)
+{ ... }
+```
+
+```c
+do { ... }
+while (i < n)
+__CPROVER_loop_invariant(0 <= i)
+__CPROVER_loop_invariant(i <= n);
+```
+
+**Important.** Invariant clauses must be free of _side effects_,
+for example, mutation of local or global variables.
+
+
+### Semantics
+
+The loop invariant clause expands to several assumptions and assertions:
+1. The invariant is _asserted_ just before the first iteration.
+2. The invariant is _assumed_ on a non-deterministic state to model a non-deterministic iteration.
+3. The invariant is finally _asserted_ again to establish its inductiveness.
+
+Mathematical induction is the working principle here.
+(1) establishes the base case for induction, and
+(2) & (3) establish the inductive case.
+Therefore, the invariant must hold _after_ the loop execution for _any_ number of iterations.
+The invariant, together with the negation of the loop guard,
+must be sufficient to establish subsequent assertions.
+If it is not, the abstraction is too imprecise and the user must supply a stronger invariant.
+
+To illustrate the key idea,
+consider the following [binary search] loop with invariant clauses:
+
+```c
+int binary_search(int val, int *buf, int size)
+{
+  if(size <= 0 || buf == NULL) return NOT_FOUND;
+
+  long lb = 0, ub = size - 1;
+  long mid = (lb + ub) / 2;
+
+  while(lb <= ub)
+  __CPROVER_loop_invariant(0L <= lb && lb - 1L <= ub && ub < size)
+  __CPROVER_loop_invariant(mid == (lb + ub) / 2L)
+  __CPROVER_decreases(ub - lb)
+  {
+     if(buf[mid] == val) break;
+     if(buf[mid] < val)
+       lb = mid + 1;
+     else
+       ub = mid - 1;
+     mid = (lb + ub) / 2;
+  }
+  return lb > ub ? NOT_FOUND : mid;
+}
+```
+
+The instrumented GOTO program is conceptually similar to the following high-level C program:
+
+```c
+int binary_search(int val, int *buf, int size)
+{
+  if(size <= 0 || buf == NULL) return NOT_FOUND;
+
+  long lb = 0, ub = size - 1;
+  long mid = (lb + ub) / 2;
+
+  /* 1. assert invariant at loop entry */
+  assert(0L <= lb && lb - 1L <= ub && ub < size);
+  assert(mid == (lb + ub) / 2L);
+
+  /* 2. create a non-deterministic state for modified variables */
+  havoc(lb, ub, mid);
+
+  /* 3. establish invariant to model state at an arbitrary iteration */
+  __CPROVER_assume(0L <= lb && lb - 1L <= ub && ub < size)
+  __CPROVER_assume(mid == (lb + ub) / 2L)
+
+  /* 4. perform a single arbitrary iteration (or exit the loop) */
+  if(lb <= ub)
+  {
+    if(buf[mid] == val) break;
+    if(buf[mid] < val)
+      lb = mid + 1;
+    else
+      ub = mid - 1;
+    mid = (lb + ub) / 2;
+
+    /* 5. assert the invariant to establish inductiveness */
+    assert(0L <= lb && lb - 1L <= ub && ub < size);
+    assert(mid == (lb + ub) / 2L);
+
+    /* 6. terminate this symbolic execution path; similar to "exit" */
+    __CPROVER_assume(false);
+  }
+  return lb > ub ? NOT_FOUND : mid;
+}
+```
+
+A few things to note here:
+
+- At instrumented code point (2), when we `havoc` the modified variables,
+  this set of modified variables is automatically computed
+  using alias analysis of l-values appearing in the loop body.
+  However, the analysis is incomplete and may fail to characterize the set for complex loops.
+  In such cases, the user must manually annotate the set of modified variables
+  using an [_assigns clause_](contracts-assigns.md).
+
+- At instrumented code point (6), when we _assume_ `false`,
+  observe that this assumption only exists within the `lb <= ub` conditional.
+  Therefore, only the symbolic execution path _inside_ this conditional block terminates.
+  The code outside of the conditional block continues to be symbolically executed,
+  and subsequent assertions do not become vacuously `true`.
+
+[history variables]: contracts-history-variables.md
+
+[binary search]: https://en.wikipedia.org/wiki/Binary_search_algorithm
--- a/doc/cprover-manual/contracts-loops.md
+++ b/doc/cprover-manual/contracts-loops.md
@@ -1,0 +1,147 @@
+# Loop Contracts
+
+CBMC offers support for loop contracts, which includes three basic clauses:
+- _invariant_ clause for establishing safety properties
+- _decreases_ clause for establishing termination, and
+- _assigns_ clause for declaring the subset of variables that is modifiable in the loop.
+
+These clauses formally describe an abstraction of a loop for the purpose of a proof.
+CBMC also provides a series of built-in constructs
+to aid writing loop contracts (e.g., _history variables_ and _quantifiers_).
+
+## Overview
+
+Consider an implementation of the [binary search algorithm] below.
+
+```c
+#include <assert.h>
+#include <stdlib.h>
+#include <stdbool.h>
+
+#define NOT_FOUND (-1)
+
+int binary_search(int val, int *buf, int size)
+{
+  if(size <= 0 || buf == NULL) return NOT_FOUND;
+
+  long lb = 0, ub = size - 1;
+  long mid = (lb + ub) / 2;
+
+  while(lb <= ub)
+  {
+     if(buf[mid] == val) break;
+     if(buf[mid] < val)
+       lb = mid + 1;
+     else
+       ub = mid - 1;
+     mid = (lb + ub) / 2;
+  }
+  return lb > ub ? NOT_FOUND : mid;
+}
+
+int main() {
+  int val, size;
+  int *buf = size >= 0 ? malloc(size * sizeof(int)) : NULL;
+
+  int idx = binary_search(val, buf, size);
+  if(idx != NOT_FOUND)
+    assert(buf[idx] == val);
+}
+```
+
+The function stores a lower bound `lb` and an upper bound `ub`
+initialized to the bounds on the buffer `buf`, i.e., to `0` and `size-1` respectively.
+In each iteration, the midpoint `mid` is compared against the target value `val`
+and in case of a mismatch either the lower half or the upper half of the buffer is searched recursively.
+A developer might be interested in verifying two high-level properties on the loop on all possible buffers `buf` and values `val`:
+1. an out-of-bound access should never occur (at `buf[mid]` lookup)
+2. the loop must eventually always terminate
+
+To prove the first (memory-safety) property,
+we may declare [_loop invariants_](contracts-invariant.md)
+that must be preserved across all loop iterations.
+In this case, two invariant clauses would together imply that `buf[mid]` lookup is always safe.
+The first invariant clause would establish that the bounds (`lb` and `ub`) are always valid:
+```c
+__CPROVER_loop_invariant(0L <= lb && lb - 1L <= ub && ub < size)
+```
+Note that in the second conjunct,
+the `lb - 1 == ub` case is possible when the value `val` is not found in the buffer `buf`.
+The second invariant clause would establish that the midpoint `mid` is always a valid index.
+In this particular case we can in fact establish a stronger invariant,
+that `mid` is indeed always the midpoint of `lb` and `ub` in every iteration:
+```c
+__CPROVER_loop_invariant(mid == (lb + ub) / 2L)
+```
+
+To prove the second (termination) property,
+we may declare a [_decreases clause_](contracts-decreases.md)
+that indicates a bounded numeric measure
+which must monotonically decrease with each loop iteration.
+In this case, it is easy to see that `lb` and `ub` are approaching closer together with each iteration, since either `lb` must increase or `ub` must decrease in each iteration.
+```c
+__CPROVER_decreases(ub - lb)
+```
+
+The loop together with all its contracts is shown below.
+
+```c
+#include <assert.h>
+#include <stdlib.h>
+#include <stdbool.h>
+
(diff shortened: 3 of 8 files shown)


doc/cprover-manual/contracts-functions.md
```diff
@@ -0,0 +1,152 @@
...
+__CPROVER_requires(__CPROVER_is_fresh(out, sizeof(*out)))
+```
+
+We can use `__CPROVER_ensures` to specify postconditions (see [Requires \&
+Ensures Clauses](contracts-requires-and-ensures.md) Section for details).  In our
+example, developers can use the built-in construct `__CPROVER_return_value`,
+which represents the return value of a function. As postconditions, one may list
+possible return values (in this case, either `SUCCESS` or `FAILURE`) as well as
+describe the main property of this function: if the function returns `SUCCESS`,
+then `*out` stores the result of `*a + *b`.  We can also check that the value in
```
TGWDB: `*a + *b` --> `a + b`

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,123 @@
...
+int bin_search(int val, int buf[], int size)
+{
+  if(size < 0) return -1;
+
+  long lb = 0, ub = size - 1;
+  long mid = (lb + ub) / 2;
+
+  while(lb <= ub)
+  __CPROVER_loop_invariant(0 <= lb && lb - 1 <= ub && ub < size)
+  __CPROVER_loop_invariant(mid == (lb + ub) / 2)
```
TGWDB: This invariant seems counter-intuitive (this may be something much deeper than the actual cbmc checking though). There are points inside the loop where this invariant does not hold. For example, on completion of line 71 or 73  this invariant will typically not hold (although it will again after line 74 is completed).

I see this is shown below (how this is implemented), maybe a note that `__CPROVER_loop_invariant` is checked at each loop iteration and may not hold for the body of the loop?

doc/cprover-manual/contracts-loops.md
```diff
@@ -0,0 +1,118 @@
...
+initialized to the bounds on the buffer `buf`, i.e., to `0` and `size-1` respectively.
+In each iteration, the midpoint `mid` is compared against the target value `val`
+and in case of a mismatch either the lower half or the upper half of the buffer is searched recursively.
+A developer might be interested in verifying two high-level properties on the loop on all possible buffers `buf` and values `val`:
+1. an out-of-bound access should never occur (at `buf[mid]` lookup)
+2. the loop must eventually always terminate
+
+To prove the first (memory-safety) property,
+we may declare _loop invariants_ that must be preserved across all loop iterations.
+In this case, two invariant clauses would together imply that `buf[mid]` lookup is always safe.
```
TGWDB, line 63: Related to the above comment on the invariant only being at certain points in the loop. This is only here because `buf[mid]` only uses the value of `mid` that is safe at entry point, if we calculated `mid` inside the body and then tried to access `buf[mid]` things could go wrong, e.g.:
```
int badfun(int *buf, int size)
{
  int i = 0;
  int mid = i;
  while(i < size)
  __CPROVER_loop_invariant(0 <=i && i <= size)
  __CPROVER_loop_invariant(mid == i)
  {
    mid = rand();
    if(buf[mid] == 0xCAFE)
      return 1;
    i++;
    mid = i;
  }
}
```
Here even though we have an invariant that `mid` is equal to `i` and `i` is safe, inside the loop we can break this.

doc/cprover-manual/contracts-functions.md
SaswatPadhi: Thanks! Fixed.

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,123 @@
+# Invariant Clause
+
+An _invariant_ clause specifies a property that must be preserved
+by every iteration of a loop.
+
+### Syntax
+
+An _invariant_ clause accepts a valid Boolean expression
+over the variables visible at the same scope as the loop.
+Additionally, [history variables] are also allowed within invariant clauses.
```
jimgrundy, line 16: That bit about "history variables" looks like it should be a link to a history variables page, but it is being rendered just as "history variables" in square brackets, which reads oddly.

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,123 @@
...
+
+```c
+__CPROVER_loop_invariant(bool cond)
+```
+
+Invariant clauses may be specified just after the loop guard.
+Multiple invariant clauses on the same loop is allowed,
```
jimgrundy: "is allowed" -> "are allowed"

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: Ah yes! Good catch; thanks. Fixed it -- added a link to the history variables Markdown file.

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,128 @@
...
+    __CPROVER_assume(false);
+  }
+
+  assert(lb > ub || (0 <= mid && mid < size));
+  return lb > ub ? -1 : mid;
+}
+
+[history variables]: contracts-history-variables.md
+
+[binary search algorithm]: https://en.wikipedia.org/wiki/Binary_search_algorithm
```
jimgrundy: Should the links to history variables be outside the code block? Right now they are not rendering as links.

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: Thanks; I would add a note at the top explaining a bit more about how loop invariants work and why they help us obtain an unbounded correctness proof.

But, this is the standard interpretation of [loop invariants in Floyd-Hoare logic](https://en.wikipedia.org/wiki/Loop_invariant). Much like function contracts, loop contracts are enforced at boundary of the loop and may not hold at intermediate steps in the loop body. It must be preserved before the loop, and after each iteration of the loop, thereby providing an inductive argument that the invariant must _always_ hold after the loop, which is the key to proving any subsequent assertions.

doc/cprover-manual/contracts-loops.md
jimgrundy: True, but then the step case of the induction will fail here. So, you won't get a proof of correctness via this procedure of some code that is not correct.

This behavior that loop invariants are true only at the start and end of an iteration is my normal expectation. I usually see a loop as breaking an invariant (say by moving an index) and then doing the work needed to reestablish the invariant again before the end.

doc/cprover-manual/contracts-loops.md
```diff
@@ -0,0 +1,118 @@
...
+The first invariant clause would establish that the bounds (`lb` and `ub`) are always valid:
+```c
+__CPROVER_loop_invariant(0 <= lb && lb - 1 <= ub && ub < size)
+```
+Note that in the second conjunct, `lb - 1 == ub` case is possible when the value `val` is not found in the buffer `buf`.
+The second invariant clause would establish that the midpoint `mid` is always a valid index.
+In this particular case we can in fact establish a stronger invariant,
+that `mid` is indeed always the midpoint of `lb` and `ub` in every iteration:
+```c
+__CPROVER_loop_invariant(mid == (lb + ub) / 2)
```
jimgrundy: You could get away with a weaker invariant here:
(b <= mid && mid <= ub)
You only need it to be the mid-point if you want to show optimal performance.
Whether or not you would find that more intuitive depends on the reader I guess.
May or may not have a measurable effect on proof time.
Change or not as you wish.

doc/cprover-manual/contracts-quantifiers.md
```diff
@@ -9,7 +9,7 @@ __CPROVER_exists { *type* *identifier*; *boolean expression* }
 
 While quantified expressions with arbitrary Boolean expressions are supported with an SMT backend, the SAT backend only supports bounded quantification under _constant_ lower & upper bounds. This is because the SAT backend currently expands a universal quantifier (`__CPROVER_forall`) to a conjunction and an existential quantifier (`__CPROVER_exists`) to a disjunction on each value within the specified bound.
 
-Concretely, with the SAT backend, the following syntax must be used for quantifiers within function contracts:
+Concretely, with the SAT backend, the following syntax must be used for quantifiers:
```
jimgrundy, line 12: Is that really the syntax we have to used for SAT?
*lower bound* <= *id* && *id* <= *upper bound*
If so, that's a real pity because as illustrated by the code below often this syntax would be more natural:
*lower bound* <= *id* && *id* < *upper bound*

doc/cprover-manual/contracts-loops.md
SaswatPadhi: The bug here is not that the implementation (body) doesn't meet the abstraction (contracts), but that the implementation uses unsafe operations. Using a loop invariant wouldn't mask those. In particular, for this case, the pointer checking flags would expose the unsafe memory access. After instrumentation, CBMC would still check an arbitrary iteration and this particular unsafe access would be caught.

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: Added a note (lines 58 -- 64).

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: You are right, but I have pushed the change already. Perhaps you were looking at a stale copy before I pushed ...

doc/cprover-manual/contracts-quantifiers.md
SaswatPadhi: The `<=` vs `<` doesn't matter. The SAT backend works with both. The important point is that that both `*upper bound*` and `*lower bound*` are constants. CBMC normalizes the `<=` and `<` and works well with both.

But for consistency, I guess I should update either the example below, or the definition of `*range*` expression below. Do you have any suggestions?

doc/cprover-manual/contracts-loops.md
SaswatPadhi: Thanks; you are right about the weaker invariant. The performance is quite good already though -- I ran the same commands I specified in this file and CBMC generates an unbounded proof in fraction of a second. 
I'll keep the current abstraction since it's stronger and seems easier to connect it to the implementation.

doc/cprover-manual/contracts-quantifiers.md
jimgrundy: I'm the kinda guy who will read the documentation and take it at it's word, so if the documentation literally says that the range has to have the form "lower bound <= id && id <= upper bound" then I'm going to use that form.

If you could give your self some wiggle room and say that "the range must specify concrete lower and upper bounds for the quantification, for example by using an expression of the form ..." then I would suggest doing that, be explicit about being vague :-) 

Also, yes, I think in the example I would prefer "lower bound <= id && id < upper bound" but it's not important I guess.

SaswatPadhi: Fixed.

doc/cprover-manual/contracts-loops.md
```diff
@@ -0,0 +1,120 @@
...
+A developer might be interested in verifying two high-level properties on the loop on all possible buffers `buf` and values `val`:
+1. an out-of-bound access should never occur (at `buf[mid]` lookup)
+2. the loop must eventually always terminate
+
+To prove the first (memory-safety) property,
+we may declare _loop invariants_ that must be preserved across all loop iterations.
+In this case, two invariant clauses would together imply that `buf[mid]` lookup is always safe.
+The first invariant clause would establish that the bounds (`lb` and `ub`) are always valid:
+```c
+__CPROVER_loop_invariant(0 <= lb && lb - 1 <= ub && ub < size)
```
feliperodri: Wouldn't be cleaner to break this down into three invariants?
```suggestion
__CPROVER_loop_invariant(0 <= lb)
__CPROVER_loop_invariant(lb - 1 <= ub)
__CPROVER_loop_invariant(ub < size)
```

doc/cprover-manual/contracts-loops.md
```diff
@@ -0,0 +1,120 @@
...
+The function stores a lower bound `lb` and an upper bound `ub`
+initialized to the bounds on the buffer `buf`, i.e., to `0` and `size-1` respectively.
+In each iteration, the midpoint `mid` is compared against the target value `val`
+and in case of a mismatch either the lower half or the upper half of the buffer is searched recursively.
+A developer might be interested in verifying two high-level properties on the loop on all possible buffers `buf` and values `val`:
+1. an out-of-bound access should never occur (at `buf[mid]` lookup)
+2. the loop must eventually always terminate
+
+To prove the first (memory-safety) property,
+we may declare _loop invariants_ that must be preserved across all loop iterations.
```
feliperodri: Could you add a link to the invariant clause documentation in _loop invariants_?

doc/cprover-manual/contracts-loops.md
```diff
@@ -0,0 +1,120 @@
...
+Note that in the second conjunct, `lb - 1 == ub` case is possible when the value `val` is not found in the buffer `buf`.
+The second invariant clause would establish that the midpoint `mid` is always a valid index.
+In this particular case we can in fact establish a stronger invariant,
+that `mid` is indeed always the midpoint of `lb` and `ub` in every iteration:
+```c
+__CPROVER_loop_invariant(mid == (lb + ub) / 2)
+```
+
+To prove the second (termination) property,
+we may declare a _decreases clause_ that indicates a bounded numeric measure which must monotonically decrease with each loop iteration.
```
feliperodri: Could you add a link to the decreases clause documentation in _decreases clause_?

doc/cprover-manual/contracts-loops.md
```diff
@@ -0,0 +1,120 @@
...
+
+```sh
+# compile the C program to a GOTO binary
+goto-cc -o binary_search.goto binary_search.c
+
+# instrument loops using specified loop contracts
+goto-instrument --apply-loop-contracts binary_search.goto binary_search.2.goto
+
+# run CBMC to check the instrumented program
+cbmc binary_search.2.goto --signed-overflow-check --function binary_search
```
feliperodri: Don't you wanna check `--bounds-check` in this example?

doc/cprover-manual/contracts-loops.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+With CBMC we can now generate an unbounded proof using these contracts:
+
+```sh
+# compile the C program to a GOTO binary
+goto-cc -o binary_search.goto binary_search.c
+
+# instrument loops using specified loop contracts
+goto-instrument --apply-loop-contracts binary_search.goto binary_search.2.goto
```
feliperodri: Nitpic: Replace `binary_search.2.goto` with `binary_search_inst.goto`.

doc/cprover-manual/contracts-loops.md
```diff
@@ -0,0 +1,120 @@
...
+     mid = (lb + ub) / 2;
+  }
+  assert(lb > ub || (0 <= mid && mid < size));
+  return lb > ub ? -1 : mid;
+}
+```
+
+With CBMC we can now generate an unbounded proof using these contracts:
+
+```sh
```
feliperodri, line 129: In the function contract documentation we group all commands together and then explain them later. Let's keep both in the same style: either update function contract or this one. Both styles are fine.
```
The first command just compiles the GOTO program as usual,
the second command instruments the code to check the function
satisfies the contract, and the third one runs CBMC to do the checking.
```

doc/cprover-manual/contracts-loops.md
feliperodri: I'd include at the end the expected result, i.e., just the final checks generated from the contract. This is also missing from function contracts. This way, we can show users what to expect from verification output.

doc/cprover-manual/contracts-quantifiers.md
```diff
@@ -19,10 +19,13 @@ __CPROVER_exists { *id* *type*; *range* && *boolean expression* }
...
 
 ```
-*lower bound* <= *id* && *id* <= *upper bound*
+*lower bound* <= *id* && *id* < *upper bound*
 ```
 
 where `*lower bound*`and `*upper bound*` are constants.
+The bound predicates could be strict (e.g., `*lower bound* < *id*`),
+or non-strict (e.g., `*upper bound* <= *id*`),
+but both the bounds must be constants.
```
feliperodri: We should add this in bold: **both the bounds must be constants**.

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,143 @@
+# Invariant Clause
+
+An _invariant_ clause specifies a property that must be preserved
+by every iteration of a loop.
+In general a loop has infinitely many invariants.
+For instance, `true` is a trivial invariant that holds before the loop,
+and after each iteration of the loop (thus, inductive).
+However, it `true` is rarely useful --
```
feliperodri: typo: "it `true`" -> "`true`"

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+```c
+do { ... }
+while (i < n)
+__CPROVER_loop_invariant(0 <= i)
+__CPROVER_loop_invariant(i <= n);
+```
+
+**Important.** Invariant clauses must be free of _side effects_,
```
feliperodri, line 47: We should emphasize here that CBMC doesn't enforce this yet.

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,143 @@
...
+for example, mutation of local or global variables.
+
+
+### Semantics
+
+The loop invariant clause expands to several assumptions and assertions:
+1. The invariant is asserted just before the first iteration
+2. The invariant is assumed on a non-deterministic state to model a non-determistic iteration
+3. The invariant is finally asserted again to establish its inductiveness
```
feliperodri: Nitpick: I usually add `;` at the end of each line.

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,143 @@
...
+Mathematical induction is the working principle here.
+(1) establishes the base case for induction, and
+(2) & (3) establish the inductive case.
+Therefore, the invariant must hold _after_ the loop execution for _any_ number of iterations.
+The invariant, together with the negation of the loop guard,
+must be sufficient to establish subsequent assertions.
+If it is not, the abstraction is too imprecise and the user must supply a stronger invariant.
+
+To illustrate the key idea,
+consider the following [binary search] loop with invariant clauses:
```
feliperodri, line 67: [binary search algorithm]

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,143 @@
...
+
+int binary_search(int val, int buf[], int size)
+{
+  if(size < 0) return -1;
+
+  long lb = 0, ub = size - 1;
+  long mid = (lb + ub) / 2;
+
+  while(lb <= ub)
+  __CPROVER_loop_invariant(0 <= lb && lb - 1 <= ub && ub < size)
```
feliperodri: I'd break down this into three invariants.

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,143 @@
...
+
+  long lb = 0, ub = size - 1;
+  long mid = (lb + ub) / 2;
+
+  /* 1. assert invariant at loop entry */
+  assert(0 <= lb && lb - 1 <= ub && ub < size);
+  assert(mid == (lb + ub) / 2);
+
+  /* 2. create a non-deterministic state for modified variables */
+  havoc(lb, ub, mid);
```
feliperodri, line 108: Could you add more explanation in the text on how we establish the modified variables in this example and add a link to assigns clause?

doc/cprover-manual/contracts-invariant.md
```diff
@@ -0,0 +1,143 @@
...
+     else
+       ub = mid - 1;
+     mid = (lb + ub) / 2;
+
+    /* 5. assert the invariant to establish inductiveness */
+    assert(0 <= lb && lb - 1 <= ub && ub < size);
+    assert(mid == (lb + ub) / 2);
+
+    /* 6. terminate this symbolic execution path; similar to "exit" */
+    __CPROVER_assume(false);
```
feliperodri, line 129: Could you add more context in the text about why we must add this false assumption and how we can guarantee it won't lead to vacuous assertions.

feliperodri (changes requested):
Thanks so much for this great docs! A few more changes and we'll be good to go.

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: I think it's cleaner to keep the "bound inavriants" together and the invariant on `mid` separate from these. I am not sure how breaking it to 3 clauses makes it necessarily cleaner. It looks way too verbose.

doc/cprover-manual/contracts-loops.md
SaswatPadhi: I think it's cleaner to keep the "bound inavriants" together and the invariant on `mid` separate from these. I am not sure how breaking it to 3 clauses makes it necessarily cleaner. It looks way too verbose.

doc/cprover-manual/contracts-loops.md
SaswatPadhi: Good catch! Will do.

doc/cprover-manual/contracts-loops.md
SaswatPadhi: Good catch! Will do.

doc/cprover-manual/contracts-loops.md
SaswatPadhi: > Don't you wanna check `--bounds-check` in this example?

Not really, we don't have a proper harness that allocates `buf`, so `--bounds-check` wouldn't work. I could include a full harness, and that works (I've tested), but I don't want to complicate the example and distract readers from the main point.

I use `--signed-overflow-check` and we check that the accessed index (`mid`) stays between `0` and `size-1`, which is a proxy for the actual allocation bounds.

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: Thanks!

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: Actually for loop contracts, we do. For function contracts, we don't, yet.

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: Good point! Will do; thanks!

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: Good point! Will do; thanks!

doc/cprover-manual/contracts-invariant.md
SaswatPadhi: Hmm, "binary search algorithm loop" is kind of awkward to read. I think "binary search loop" is fine?

feliperodri: I was referring to add the reference link or kill the brackets.

SaswatPadhi: Oops, you are right! The link at the bottom has wrong label. I will fix it; thanks!

doc/cprover-manual/contracts-loops.md
feliperodri: My question here is: can we guarantee to catch out-of-bound access at `buf[mid]` by following these instructions? If yes, leave it as it is. If no, at least state in the text how to catch it.

SaswatPadhi: Since `buf` isn't even allocated, we can't. What we do guarantee is that `mid` stays between `0` and `size-1` (via overflow checks), rather than `mid` stays between `0` and `object_size(buf)-1` (via bounds checks).

In any case, I see your confusion and other readers might have the same confusion about this sort of "proxy" for bounds check, so I would expand the example a bit more. Thanks!

feliperodri (approved):

jimgrundy (approved):
I'm happy with, was by last night, sorry for being slow.

SaswatPadhi:
Thanks @jimgrundy, @feliperodri. I'll address the last remaining comments and merge tomorrow.

PR #6317, opened 2021-08-25
Abstract interpretation for strict and monotonic change

LongPham7:
In this pull request, I have implemented an abstract interpretation for strict and monotonic change: it checks whether variables have strictly and monotonically increased or decreased since the beginning of given GOTO code. This abstract interpretation will be used in the automatic inference of decreases clauses (aka ranking functions or loop variants). For example, consider a while-loop whose loop guard is `x < y`. If we know that `x` strictly (and monotonically) increases in the loop body and that `y` stays constant, we can successfully infer that `y - x` is a correct decreases clause.

This predicate abstraction works as follows. For each variable, we keep track of its abstract value. There are five abstract values in the domain:
1. Unchanged
2. Strict and monotonic increase: the variable has strictly and monotonically increased since the beginning of the code. 
3. Strict and monotonic decrease
4. Top: inconclusive.
5. Bottom: infeasible.

Here is an example: 
```
DECL x            // x has stayed unchanged so far
ASSIGN x := x + 1 // x has strictly increased so far
ASSIGN x := x - 1 // x is now top
```

Note that this abstract interpretation does NOT keep track of how much a variable has changed. So if we have `x--` followed by `x++` as in the above example, the abstract interpretation cannot tell us the net effect on `x`. This is why `x`'s abstract value after the last line of the above example is top. 

### DIFF
--- a/regression/goto-analyzer/monotonic_change_conditionals/main.c
+++ b/regression/goto-analyzer/monotonic_change_conditionals/main.c
@@ -1,0 +1,18 @@
+int nondet_bool(void);
+
+int main()
+{
+  int x;
+  if(nondet_bool())
+    x++;
+  else
+    x += 6;
+
+  int y;
+  if(nondet_bool())
+    y++;
+  else
+    y -= 6;
+
+  return 0;
+}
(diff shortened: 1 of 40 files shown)


feliperodri:
Closing this PR for now, as we will invest on a solution using an enumerative synthesizer (cc. @qinheping).

PR #6318, opened 2021-08-26
Add initial subset of `exprt` to internal `smt_termt` conversions

thomasspriggs:
Add initial subset of `exprt` to internal `smt_termt` conversions

### DIFF
--- a/src/solvers/smt2_incremental/convert_expr_to_smt.cpp
+++ b/src/solvers/smt2_incremental/convert_expr_to_smt.cpp
@@ -3,6 +3,7 @@
 #include <solvers/smt2_incremental/convert_expr_to_smt.h>
 
 #include <solvers/prop/literal_expr.h>
+#include <solvers/smt2_incremental/smt_bit_vector_theory.h>
 #include <solvers/smt2_incremental/smt_core_theory.h>
 #include <util/arith_tools.h>
 #include <util/bitvector_expr.h>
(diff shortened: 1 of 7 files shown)


TGWDB (commented):
Looks good so far (not going to approve as it is still draft).

src/solvers/smt2_incremental/convert_expr_to_smt.cpp
```diff
@@ -56,9 +56,15 @@ static smt_termt convert_expr_to_smt(const union_exprt &union_construction)
...
+    const auto bool_type =
+      type_try_dynamic_cast<bool_typet>(constant_literal.type()))
+  {
+    return smt_bool_literal_termt{constant_literal.is_true()};
+  }
   UNIMPLEMENTED_FEATURE(
-    "Generation of SMT formula for constant literal expression: " +
-    constant_literal.pretty());
+    "Generation of SMT formula for constant literal expression of type: " +
+    constant_literal.type().pretty());
```
TGWDB: I thought the literal would include the type information? Would it be more helpful to know both type and literal value when debugging?

TGWDB (approved):

peterschrammel (approved):

src/solvers/smt2_incremental/convert_expr_to_smt.cpp
thomasspriggs: The type is enough to determine the proximate cause of the invariant violation and is more specific to the condition this relates to in this case. This is because conversion of constant literal expressions are now (partially) implemented. This message is actually moved inside of `convert_type_to_smt_sort` instead in a subsequent refactor in this PR, where the full expression is unavailable. I would argue that if we are seriously engaged in debugging then we can add whatever additional printing is useful as part of that debugging process.

PR #6320, opened 2021-08-26
Automatic inference of decreases clauses [depends-on: #6317]

LongPham7:
This pull request consists of two commits:
1. Abstract interpretation for strict and monotonic change. It should be identical to the one from PR #6317. 
2. Automatic inference of decreases clauses. It depends on the first commit. 

To automatically infer decrease clauses, we first extract a loop body and wrap it inside a dummy function. We then pass it to the abstract interpretation of strict and monotonic change. 


feliperodri:
Closing this PR for now, as we will invest on a solution using an enumerative synthesizer (cc. @qinheping).

PR #6326, opened 2021-09-01
enable nondeterministic pointers

kroening:
This commit enables the use of nondeterministic pointers, to allow
declarative modeling of states that include pointers.

### DIFF
--- a/regression/cbmc/nondet-pointer/nondet-pointer2.c
+++ b/regression/cbmc/nondet-pointer/nondet-pointer2.c
@@ -1,0 +1,14 @@
+int *nondet_pointer();
+
+int main()
+{
+  int x = 123, y = 456;
+  int *px = &x;
+  int *py = &y;
+  int *q = nondet_pointer();
+
+  if(q == px || q == py)
+    __CPROVER_assert(*q == 123 || *q == 456, "value of *q");
+
+  return 0;
+}
--- a/regression/cbmc/nondet-pointer/nondet-pointer5.c
+++ b/regression/cbmc/nondet-pointer/nondet-pointer5.c
@@ -1,0 +1,19 @@
+int *nondet_pointer();
+
+int main()
+{
+  int some_array[10];
+  int *p = some_array;
+  int *q = nondet_pointer();
+  int index;
+
+  __CPROVER_assume(index >= 0 && index < 10);
+  __CPROVER_assume(__CPROVER_same_object(p, q));
+  __CPROVER_assume(__CPROVER_POINTER_OFFSET(q) == sizeof(int) * index);
+
+  *q = 123;
+
+  __CPROVER_assert(some_array[index] == 123, "value of array[index]");
+
+  return 0;
+}
--- a/src/pointer-analysis/value_set.cpp
+++ b/src/pointer-analysis/value_set.cpp
@@ -505,6 +505,23 @@ void value_sett::get_value_set_rec(
     else
       insert(dest, exprt(ID_unknown, original_type));
   }
+  else if(expr.id() == ID_nondet_symbol)
+  {
+    if(expr.type().id() == ID_pointer)
+    {
+      // we'll take the union of all objects we see, with unspecified offsets
+      values.iterate([this, &dest](const irep_idt &key, const entryt &value) {
+        for(const auto &object : value.object_map.read())
+          insert(dest, object.first, offsett());
+      });
+
+      // we'll add null, in case it's not there yet
+      insert(
+        dest,
+        exprt(ID_null_object, to_pointer_type(expr_type).subtype()),
+        offsett());
+    }
+  }
   else if(expr.id()==ID_if)
   {
     get_value_set_rec(
@@ -984,7 +1001,6 @@ void value_sett::get_value_set_rec(
   }
   else
   {
-    // for example: expr.id() == ID_nondet_symbol
     insert(dest, exprt(ID_unknown, original_type));
   }
 
--- a/regression/cbmc/Pointer14/main.c
+++ b/regression/cbmc/Pointer14/main.c
@@ -1,6 +1,4 @@
-typedef unsigned int size_t;
-
-void *malloc(size_t size);
+void *malloc(__CPROVER_size_t);
 
 enum blockstate
 {
--- a/regression/cbmc/String_Abstraction14/pass-in-implicit.c
+++ b/regression/cbmc/String_Abstraction14/pass-in-implicit.c
@@ -1,4 +1,4 @@
-void *malloc(unsigned);
+void *malloc(__CPROVER_size_t);
 
 void use_str(char *s)
 {
--- a/regression/cbmc/String_Abstraction20/structs2.c
+++ b/regression/cbmc/String_Abstraction20/structs2.c
@@ -1,4 +1,4 @@
-void *malloc(unsigned);
+void *malloc(__CPROVER_size_t);
 
 typedef struct str_struct
 {
--- a/regression/cbmc/double_deref/double_deref_with_pointer_arithmetic.desc
+++ b/regression/cbmc/double_deref/double_deref_with_pointer_arithmetic.desc
@@ -2,7 +2,7 @@ CORE
 double_deref_with_pointer_arithmetic.c
 --show-vcc
 ^\{-[0-9]+\} derefd_pointer::derefd_pointer!0#1 = \{ symex_dynamic::dynamic_object1#3\[\[0\]\], symex_dynamic::dynamic_object1#3\[\[1\]\] \}\[cast\(mod #source_location=""\(main::argc!0@1#1, 2\), signedbv\[64\]\)\]
-^\{1\} \(derefd_pointer::derefd_pointer!0#1 = address_of\(symex_dynamic::dynamic_object[23]\) \? main::argc!0@1#1 = 2 : \(derefd_pointer::derefd_pointer!0#1 = address_of\(symex_dynamic::dynamic_object[23]\) \? main::argc!0@1#1 = 1 : symex::invalid_object!0#0 = main::argc!0@1#1\)\)
+^\{1\} \(derefd_pointer::derefd_pointer!0#1 = address_of\(symex_dynamic::dynamic_object[23]\) \? main::argc!0@1#1 = 2 : main::argc!0@1#1 = 1
 ^EXIT=0$
 ^SIGNAL=0$
 --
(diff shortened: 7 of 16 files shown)


danielsn:
What would be the semantics for this case (slight modification of the previous test case)

```C
int *nondet_pointer();

int main()
{
  int *q = nondet_pointer(); // Reordered to be 1st
  int x = 123;
  int *p = &x;

  if(p == q)
    __CPROVER_assert(*q == 123, "value of *q");

  return 0;
}
```

danielsn:
Also

```C
int *nondet_pointer();

int main()
{
  int x = 123;
  int *p = &x;
  int *q = nondet_pointer();

  __CPROVER_assume(p == q);
  __CPROVER_assert(*q == 123, "value of *q");

  return 0;
}
```

jimgrundy:
> What would be the semantics for this case (slight modification of the previous test case)
> 
> ```c
> int *nondet_pointer();
> 
> int main()
> {
>   int *q = nondet_pointer(); // Reordered to be 1st
>   int x = 123;
>   int *p = &x;
> 
>   if(p == q)
>     __CPROVER_assert(*q == 123, "value of *q");
> 
>   return 0;
> }
> ```

My expectation would be that q must be NULL because there are no other pointers to choose from when q is assigned, therefore the then clause of the conditional is unreachable and the assertion vacuously true.

regression/cbmc/nondet-pointer/nondet-pointer2.c
```diff
@@ -0,0 +1,14 @@
+int *nondet_pointer();
+
+int main()
+{
+  int x = 123, y = 456;
+  int *px = &x;
+  int *py = &y;
+  int *q = nondet_pointer();
+
+  if(q == px || q == py)
```
jimgrundy, line 10: Why is it sufficient to test if q == px or q == py? Should we not also have to test same_object? 
Why do I say that? Daniel has pointed out to me that we can construct two pointers that are equal but have different provenance and accessing one may cause an error.
Suppose y comes after x in memory. Suppose we choose q to be derived from px (&x) but with an offset large enough to point it at y. Now q == py ... but is it legal to dereference it and get 456, or should I expect failure?

jimgrundy:
> Also
> 
> ```c
> int *nondet_pointer();
> 
> int main()
> {
>   int x = 123;
>   int *p = &x;
>   int *q = nondet_pointer();
> 
>   __CPROVER_assume(p == q);
>   __CPROVER_assert(*q == 123, "value of *q");
> 
>   return 0;
> }
> ```

Here I am expecting q must be either NULL or a pointer derived from p since that is the only pointer around.  The assumption makes q equal to p and since it must have been derived from p it is safe to dereference and so I expect *q to be 123 and the assertion to pass.

regression/cbmc/nondet-pointer/nondet-pointer2.c
kroening: Nondeterministic pointers are outside of the semantics of C; we can make that go either way.

remi-delmas-3000:
Would that last assertion pass or fail ?

```c
int *nondet_pointer();

int main()
{
  int x;
  int *p = &x;
  int *q = nondet_pointer();

  if(p == q) {
      __CPROVER_assume(*q == 123);
      __CPROVER_assert(x == 123, "would that pass or fail ?");
  }
  return 0;
}
```

jimgrundy:
We will have to put this one on hold for a bit. We discovered an unsoundness Friday when this is combined with the loop contracts feature. We will have to add some more checks to loop contracts to ensure we don't break soundness when this is added. I'll follow up with an explanation.

jimgrundy:
> We will have to put this one on hold for a bit. We discovered an unsoundness Friday when this is combined with the loop contracts feature. We will have to add some more checks to loop contracts to ensure we don't break soundness when this is added. I'll follow up with an explanation.

Maybe no issue. Will investigate further within a couple of days.

jimgrundy:
Looks like there is not an issue with nondet_pointer ... only perhaps with when we should use it. Let's proceed with documenting this so we can merge it.

SaswatPadhi:
> Looks like there is not an issue with nondet_pointer ... only perhaps with when we should use it. Let's proceed with documenting this so we can merge it.

According to our offline discussions, the invalid pointers introduced by `nondet_pointer` don't work with existing `IS_INVALID` check. It would be nice if we could update that check as part of this PR to keep things consistent.

jimgrundy:
@kroening Can you also make the necessary updates to the is_invalid check for consistency with this new feature?

jimgrundy:
Did I see the IS_INVALID which we need to pair with this make it in via another PR?

SaswatPadhi:
> Did I see the IS_INVALID which we need to pair with this make it in via another PR?

Perhaps #6366

regression/cbmc/nondet-pointer/nondet-pointer5.c
```diff
@@ -0,0 +1,19 @@
...
+
+int main()
+{
+  int some_array[10];
+  int *p = some_array;
+  int *q = nondet_pointer();
+  int index;
+
+  __CPROVER_assume(index >= 0 && index < 10);
+  __CPROVER_assume(__CPROVER_same_object(p, q));
```
qinheping, line 11: When I run `cbmc --pointer-primitive-check nondet-pointer5.c`, I got the following result
```
** Results:
nondet-pointer5.c function main
[main.pointer_primitives.1] line 11 dead object in POINTER_OBJECT((const void *)p): SUCCESS
[main.pointer_primitives.2] line 11 pointer outside object bounds in POINTER_OBJECT((const void *)p): SUCCESS
[main.pointer_primitives.3] line 11 pointer invalid in POINTER_OBJECT((const void *)q): FAILURE
[main.pointer_primitives.4] line 11 deallocated dynamic object in POINTER_OBJECT((const void *)q): SUCCESS
[main.pointer_primitives.5] line 11 dead object in POINTER_OBJECT((const void *)q): SUCCESS
[main.pointer_primitives.6] line 11 pointer outside object bounds in POINTER_OBJECT((const void *)q): FAILURE
[main.pointer_primitives.7] line 12 pointer invalid in POINTER_OFFSET((const void *)q): SUCCESS
[main.pointer_primitives.8] line 12 deallocated dynamic object in POINTER_OFFSET((const void *)q): SUCCESS
[main.pointer_primitives.9] line 12 dead object in POINTER_OFFSET((const void *)q): SUCCESS
[main.pointer_primitives.10] line 12 pointer outside object bounds in POINTER_OFFSET((const void *)q): FAILURE
[main.assertion.1] line 16 value of array[index]: SUCCESS

** 3 of 11 failed (3 iterations)
```
The trace showed that the offset of the pointer `q` exceeds the object size, which violates the pointer-primitive check
`POINTER_OBJECT(NULL) == POINTER_OBJECT(q) || POINTER_OFFSET(q) >= 0l && OBJECT_SIZE(q) >= (unsigned long int)POINTER_OFFSET(q) + 1ul`.

SaswatPadhi: Yeah `--pointer-primitive-check` shouldn't look into offsets when we are just comparing `same_object`

jimgrundy: Hey Daniel (@kroening). We were wondering about this regression that fails with this PR. We don't really need the offsets to be in range for the pointers for same_object to have a well defined meaning do we? Is there another assumption we could/should add to make this go away? Should same_object be relaxed to not worry about the legality of the indexes? Other thoughts? Thanks.

kroening: Not an easy question. You realise that this could result in a situation whereby you've just asserted same_object, and then you see a concrete trace of the program where that very pointer points at the beginning of some other object.

I can live with that, given that the C standard has a notion of a pointer that points just beyond the end of a sequence, but this is a prime candidate for confusing users.

jimgrundy: Yes, we do realize that. We're needing to constrain a pointer in a loop invariant such that as the pointer is iterated through the array it remains pointing to the same object. We'll then further constrain it to be inbounds for the array (or one beyond). To do this without warnings we need to shut off the assertion about the index being out of range in the same_object check. We'd like to keep the other sanity checks though. We _could_ define our own version were we shut off all the pointer-primitive checks, but just locally for that one assertion ... if we had the ability to do that with a pragma, which is a request on another thread. If we had that we could more easily make progress ourselves. Or, we could implement a version with no primitives checks as a new primitive - but that seems heavyweight.

tautschnig:
Shouldn't we also

1) include the invalid object in the list of candidate objects (just like `NULL` is included) and
2) amend https://github.com/diffblue/cbmc/blob/ce2680b724d650c404468eb0070c8b402735ec66/src/goto-symex/goto_symex_state.cpp#L823..L836 to use a `nondet` symbol (or perhaps provide a way to directly update the points-to set)?

regression/cbmc/nondet-pointer/nondet-pointer5.c
tautschnig: I have put up https://github.com/diffblue/cbmc/pull/6491 to adjust the behaviour of pointer-primitive-check.

src/pointer-analysis/value_set.cpp
```diff
@@ -505,6 +505,23 @@ void value_sett::get_value_set_rec(
     else
       insert(dest, exprt(ID_unknown, original_type));
   }
+  else if(expr.id() == ID_nondet_symbol)
+  {
+    if(expr.type().id() == ID_pointer)
+    {
+      // we'll take the union of all objects we see, with unspecified offsets
```
martin-cs, line 512: I worry a little that the semantics of "nondet but only on objects that exist at the point it was nondet'd" might introduce some unexpected effects BUT I am not sure I have a better idea of a clearer semantics.

martin-cs (approved):
Yes.  I wonder if it should include an / the invalid object or some other kind of "this pointer points to something that is not accessible via other means" / "writes to this are effectively ignored" partially because that would be useful but also because hooks on this could be used to drive a refinement loop of "what input data structure is assumed".

src/pointer-analysis/value_set.cpp
jimgrundy: I think @tautschnig has a 2nd PR in the works that addresses this concern, but it builds on this PR.

kroening:
@tautschnig will progress #6442, but we'll merge this for now.

PR #6327, opened 2021-09-02
No need to track write set with temporary variables

feliperodri:

### DIFF
--- a/regression/contracts/assigns_enforce_structs_07/test.desc
+++ b/regression/contracts/assigns_enforce_structs_07/test.desc
@@ -3,10 +3,10 @@ main.c
 --enforce-all-contracts _ --pointer-check
 ^EXIT=10$
 ^SIGNAL=0$
-^\[f1.\d+\] line \d+ Check that p\-\>buf\[\(.*\)0\] is assignable\: FAILURE$
-^\[f1.pointer\_dereference.\d+\] line \d+ dereference failure\: pointer NULL in p\-\>buf\: FAILURE$
-^\[f2.\d+\] line \d+ Check that pp\-\>p\-\>buf\[\(.*\)0\] is assignable\: FAILURE$
-^\[f2.pointer\_dereference.\d+\] line \d+ dereference failure\: pointer NULL in pp\-\>p\-\>buf\: FAILURE$
+^\[f1.\d+\] line \d+ Check that p->buf\[\(.*\)0\] is assignable: FAILURE$
+^\[f1.pointer\_dereference.\d+\] line \d+ dereference failure: pointer NULL in p->buf: FAILURE$
+^\[f2.\d+\] line \d+ Check that pp->p->buf\[\(.*\)0\] is assignable: FAILURE$
+^\[f2.pointer\_dereference.\d+\] line \d+ dereference failure: pointer NULL in pp->p->buf: FAILURE$
 ^VERIFICATION FAILED$
 --
 --
--- a/regression/contracts/assigns_enforce_structs_07/test.desc
+++ b/regression/contracts/assigns_enforce_structs_07/test.desc
@@ -1,0 +1,14 @@
+CORE
+main.c
+--enforce-all-contracts _ --pointer-check
+^EXIT=10$
+^SIGNAL=0$
+^\[f1.\d+\] line \d+ Check that p\-\>buf\[\(.*\)0\] is assignable\: FAILURE$
+^\[f1.pointer\_dereference.\d+\] line \d+ dereference failure\: pointer NULL in p\-\>buf\: FAILURE$
+^\[f2.\d+\] line \d+ Check that pp\-\>p\-\>buf\[\(.*\)0\] is assignable\: FAILURE$
+^\[f2.pointer\_dereference.\d+\] line \d+ dereference failure\: pointer NULL in pp\-\>p\-\>buf\: FAILURE$
+^VERIFICATION FAILED$
+--
+--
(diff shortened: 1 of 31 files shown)


src/goto-instrument/contracts/contracts.cpp
```diff
@@ -895,7 +891,12 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     return true;
   }
   goto_programt &program = old_function->second.body;
-  if(program.instructions.empty()) // empty function body
+
+  if(
+    program.empty() ||
+    (program.instructions.size() <= 2 &&
+     program.instructions.front().is_skip() &&
+     program.instructions.back().is_end_function())) // empty function body
```
SaswatPadhi: This check seems a bit fragile. We're checking if either `program` is empty, or has a single `skip` statement. What if it has 2 skip statements?
Is there a way (some goto analysis we can call) to normalize a `programt` and then check for emptiness? May be @tautschnig could advise us here,

Is this check really necessary though?
We `return false` at the end anyway. We should make sure that the `check_for_looped_mallocs` and `check_frame_conditions` calls in between don't misbehave on these empty functions (or functions that are semantically equivalent to empty).

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -24,49 +24,30 @@ assigns_clause_targett::assigns_clause_targett(
...
-}
-
 exprt assigns_clause_targett::alias_expression(const exprt &lhs)
 {
   exprt::operandst condition;
   exprt lhs_ptr = (lhs.id() == ID_address_of) ? to_address_of_expr(lhs).object()
                                               : pointer_for(lhs);
 
+  // NULL targets aren't writable
+  condition.push_back(not_exprt(null_pointer(target)));
```
SaswatPadhi: Should this check be more generally about validity, not just non-`NULL`-ness? Like the w_ok check we use in `compatible_expression`?

regression/contracts/assigns_enforce_structs_07/test.desc
```diff
@@ -0,0 +1,14 @@
+CORE
+main.c
+--enforce-all-contracts _ --pointer-check
+^EXIT=10$
+^SIGNAL=0$
+^\[f1.\d+\] line \d+ Check that p\-\>buf\[\(.*\)0\] is assignable\: FAILURE$
+^\[f1.pointer\_dereference.\d+\] line \d+ dereference failure\: pointer NULL in p\-\>buf\: FAILURE$
+^\[f2.\d+\] line \d+ Check that pp\-\>p\-\>buf\[\(.*\)0\] is assignable\: FAILURE$
+^\[f2.pointer\_dereference.\d+\] line \d+ dereference failure\: pointer NULL in pp\-\>p\-\>buf\: FAILURE$
```
SaswatPadhi: [nit]

We could reduce some of the escaping here. In particular, `-`, `>` and `:` need not be escaped here (and elsewhere, in other `.desc` files). Not a big issues, but just to makes these a little more readable

src/goto-instrument/contracts/contracts.cpp
feliperodri: This was indeed debugging left-overs. For any "empty" function in my examples, I observed at least 2 instructions: SKIP and END_FUNCTION; thus, program was never empty. What we need here is to improve the `program.empty()` method to effectively check for emptiness, which only relying in the number of instructions.

In any case, this is not necessary at all and we can remove it from our contract implementation.

regression/contracts/assigns_enforce_structs_07/test.desc
feliperodri: I removed all instances of these escaping partners from our contracts regression tests.

src/goto-instrument/contracts/assigns.cpp
feliperodri: We could add a check similar to
```cpp
condition.push_back(w_ok_exprt(target, from_integer(object_size(target), unsigned_int_type())));
```
How is that better than this nullness check? For me, it looks a bit redundant, because I get the size of the object, then I check if the object is writable up to its given size.

regression/contracts/assigns_enforce_structs_07/test.desc
SaswatPadhi: Thanks.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Thanks.

src/goto-instrument/contracts/assigns.cpp
SaswatPadhi: > because I get the size of the object, then I check if the object is writable up to its given size.

I might have missed this. If you do this, then yes it's redundant. Could you please point me to the specific lines that implement this?

> How is that better than this nullness check?

The idea is to use `w_ok` to make sure it's writable.

feliperodri: > Could you please point me to the specific lines that implement this?

I meant, I could replace the line 49 with this:
```c
condition.push_back(w_ok_exprt(target, from_integer(object_size(target), unsigned_int_type())));
```

It's not part of our alias check yet.

feliperodri: Implemented as suggested.

SaswatPadhi (approved):
LGTM :+1: 

Thanks!

PR #6331, opened 2021-09-03
Adds the notion of a local write set to contracts

feliperodri:

### DIFF
--- a/regression/contracts/assigns_enforce_18/test.desc
+++ b/regression/contracts/assigns_enforce_18/test.desc
@@ -12,8 +12,9 @@ main.c
 ^.* 1 of \d+ failed \(\d+ iteration.*\)$
 ^VERIFICATION FAILED$
 --
+^\[foo.\d+\] line \d+ Check that a is assignable: FAILURE$
 ^\[foo.pointer\_primitives.\d+\] line \d+ deallocated dynamic object in POINTER_OBJECT\(tmp\_cc\$\d+\): FAILURE$
-^.* 8 of \d+ failed (\d+ iterations)$
+^.* 3 of \d+ failed (\d+ iterations)$
 --
 Checks whether contract enforcement works with functions that deallocate memory.
 We had problems before when freeing a variable, but still keeping it on
--- a/regression/contracts/assigns_enforce_18/test.desc
+++ b/regression/contracts/assigns_enforce_18/test.desc
@@ -3,13 +3,16 @@ main.c
 --enforce-all-contracts _ --pointer-primitive-check
 ^EXIT=10$
 ^SIGNAL=0$
+^\[bar.\d+\] line \d+ Check that \*a is assignable: SUCCESS$
 ^\[bar.\d+\] line \d+ Check that \*b is assignable: SUCCESS$
+^\[baz.\d+\] line \d+ Check that \*c is assignable: FAILURE$
+^\[baz.\d+\] line \d+ Check that \*a is assignable: SUCCESS$
 ^\[foo.\d+\] line \d+ Check that a is assignable: SUCCESS$
 ^\[foo.\d+\] line \d+ Check that yp is assignable: SUCCESS$
 ^\[foo.\d+\] line \d+ Check that z is assignable: SUCCESS$
 ^\[foo.\d+\] line \d+ Check that \*xp is assignable: SUCCESS$
 ^\[foo.\d+\] line \d+ Check that y is assignable: FAILURE$
-^.* 1 of \d+ failed \(\d+ iteration.*\)$
+^.* 2 of \d+ failed \(\d+ iteration.*\)$
 ^VERIFICATION FAILED$
 --
 ^\[foo.\d+\] line \d+ Check that a is assignable: FAILURE$
--- a/regression/contracts/assigns_enforce_18/test.desc
+++ b/regression/contracts/assigns_enforce_18/test.desc
@@ -1,0 +1,21 @@
+CORE
+main.c
+--enforce-all-contracts _ --pointer-primitive-check
+^EXIT=10$
+^SIGNAL=0$
+^\[bar.\d+\] line \d+ Check that \*b is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that a is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that yp is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that z is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that \*xp is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that y is assignable: FAILURE$
+^.* 1 of \d+ failed \(\d+ iteration.*\)$
+^VERIFICATION FAILED$
+--
+^\[foo.pointer\_primitives.\d+\] line \d+ deallocated dynamic object in POINTER_OBJECT\(tmp\_cc\$\d+\): FAILURE$
+^.* 8 of \d+ failed (\d+ iterations)$
+--
+Checks whether contract enforcement works with functions that deallocate memory.
+We had problems before when freeing a variable, but still keeping it on
+the writable set, which lead to deallocated variables issues.
+Now, if a memory is deallocated, we remove it from the our freely assignable set.
--- a/regression/contracts/assigns_enforce_19/main.c
+++ b/regression/contracts/assigns_enforce_19/main.c
@@ -1,0 +1,33 @@
+#include <assert.h>
+
+static int b = 0;
+static int c = 0;
+
+int f() __CPROVER_assigns()
+{
+  static int a = 0;
+  a++;
+  return a;
+}
+
+int g(int *points_to_b, int *points_to_c)
+  __CPROVER_assigns(b) // Error: it should also mention c
+{
+  b = 1;
+  c = 2;
+  *points_to_b = 1;
+  *points_to_c = 2;
+}
+
+int main()
+{
+  assert(f() == 1);
+  assert(f() == 2);
+  assert(b == 0);
+  assert(c == 0);
+  g(&b, &c);
+  assert(b == 1);
+  assert(c == 2);
+
+  return 0;
+}
--- a/regression/contracts/assigns_validity_pointer_04/test.desc
+++ b/regression/contracts/assigns_validity_pointer_04/test.desc
@@ -1,9 +1,11 @@
-KNOWNBUG
+CORE
 main.c
 --enforce-contract foo --replace-call-with-contract bar --replace-call-with-contract baz _ --pointer-primitive-check
-^EXIT=0$
+^EXIT=10$
 ^SIGNAL=0$
-^VERIFICATION SUCCESSFUL$
+^\[foo.\d+\] line \d+ Check that z is assignable: FAILURE$
+^.* 1 of \d+ failed \(\d+ iteration.*\)$
+^VERIFICATION FAILED$
 // bar
 ASSERT \*foo::x > 0
 IF ¬\(\*foo::x = 3\) THEN GOTO \d
@@ -17,9 +19,3 @@ This test checks support for a malloced pointer that is assigned to by
 a function (bar and baz). Both functions bar and baz are being replaced by
 their function contracts, while the calling function foo is being checked
 (by enforcing it's function contracts).
-
-BUG: `z` is being assigned to in `foo`, but is not in `foo`s assigns clause!
-This test is expected to pass but it should not.
-It somehow used to (and still does on *nix), but that seems buggy.
-Most likely the bug is related to `freely_assignable_symbols`,
-which would be properly fixed in a subsequent PR.
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -138,6 +138,9 @@ goto_programt assigns_clauset::generate_havoc_code() const
   for(const auto &target : global_write_set)
     modifies.insert(target.address.object());
 
+  for(const auto &target : local_write_set)
+    modifies.insert(target.address.object());
+
   goto_programt havoc_statements;
   append_havoc_code(location, modifies, havoc_statements);
   return havoc_statements;
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -88,17 +88,18 @@ assigns_clauset::assigns_clauset(
   const exprt &expr,
   const messaget &log,
   const namespacet &ns)
-  : expr(expr), log(log), ns(ns)
+  : location(expr.source_location()), log(log), ns(ns)
 {
   for(const auto &target_expr : expr.operands())
   {
-    add_target(target_expr);
+    add_global_write_set(target_expr);
   }
 }
 
-void assigns_clauset::add_target(const exprt &target_expr)
+void assigns_clauset::add_global_write_set(const exprt &target_expr)
 {
-  auto insertion_succeeded = write_set.emplace(*this, target_expr).second;
+  auto insertion_succeeded =
+    global_write_set.emplace(*this, target_expr).second;
 
   if(!insertion_succeeded)
   {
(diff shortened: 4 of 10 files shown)


regression/contracts/assigns_enforce_18/test.desc
```diff
@@ -0,0 +1,21 @@
...
+--
+We had problems before when freeing a variable, but still kepting it on
+the writable set, which lead to dead variables issues. Anything that is local,
+including local allocations are now tracked by our freely assignable set;
+therefore, CBMC doesn't include these local variables in its alias check
+anymore (which avoids the issue previously described). Assignments to
+anything in our freely assignable set are bypassed. It’s important to
+notice that we track the address of all objects, thus, if a local pointer
+is created and points to something that is not local (e.g., inputs or global
+variables), CBMC will check every assignment to it.
```
TGWDB: "still kepting it on the writable set, which lead to dead variables issues." -> "still keeping it in the writable set, which lead to dead variable issues."

The information in this test appears to be useful to know, but in the wrong place. This describes how the implementation works, not what the test is actually checking for. I expect that someone trying to understand the code for contracts would look at the source code as a place to find information on how it works.

TGWDB (changes requested):
Good to have a regression test (and the test looks good), please put the information from the test description comment close to where it would be found/used by a developer (to explain why the code works the way it does).

TGWDB (approved):

feliperodri:
@SaswatPadhi could you take another look at it?

regression/contracts/assigns_enforce_18/test.desc
```diff
@@ -0,0 +1,21 @@
...
+^\[foo.\d+\] line \d+ Check that a is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that yp is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that z is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that \*xp is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that y is assignable: FAILURE$
+^.* 1 of \d+ failed \(\d+ iteration.*\)$
+^VERIFICATION FAILED$
+--
+^\[foo.pointer\_primitives.\d+\] line \d+ deallocated dynamic object in POINTER_OBJECT\(tmp\_cc\$\d+\): FAILURE$
+^.* 8 of \d+ failed (\d+ iterations)$
```
SaswatPadhi: I don't understand this negative regex. Why are we checking `8` out of "anything"?
We should list out the ones of interest that are not supposed to pass / fail explicitly, just like the assertion above.

src/goto-instrument/contracts/assigns.h
```diff
@@ -34,7 +34,7 @@ class assigns_clauset
 
     bool operator==(const targett &other) const
     {
-      return expr == other.expr;
+      return address == other.address;
```
SaswatPadhi, line 37: This is inconsistent with the hash function, because that hashes the `expr` not `address`.
We don't want to have two targets such that `t1 == t2` but `hash(t1) != hash(t2)`.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -99,7 +99,8 @@ assigns_clauset::assigns_clauset(
 
 void assigns_clauset::add_target(const exprt &target_expr)
 {
-  auto insertion_succeeded = targets.emplace(*this, target_expr).second;
+  auto insertion_succeeded =
+    write_set.emplace(*this, targett::normalize(target_expr)).second;
```
SaswatPadhi: I am not sure why this change is needed.
The constructor already calls `normalize` before storing the `address`.

I think this call should remain as before and we should be able to remove `expr` from `targett` entirely -- only `address` seems to be enough to track everything.

regression/contracts/assigns_validity_pointer_04/test.desc
```diff
@@ -1,9 +1,11 @@
-KNOWNBUG
+CORE
 main.c
 --enforce-contract foo --replace-call-with-contract bar --replace-call-with-contract baz _ --pointer-primitive-check
-^EXIT=0$
+^EXIT=10$
 ^SIGNAL=0$
-^VERIFICATION SUCCESSFUL$
+^\[foo.\d+\] line \d+ Check that z is assignable: FAILURE$
+^.* 1 of \d+ failed \(\d+ iteration.*\)$
```
SaswatPadhi, line 7: Same comment. Why are we checking `1` out of "anything"?

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -112,14 +113,32 @@ void assigns_clauset::add_target(const exprt &target_expr)
 
 void assigns_clauset::remove_target(const exprt &target_expr)
 {
-  targets.erase(targett(*this, targett::normalize(target_expr)));
+  write_set.erase(targett(*this, targett::normalize(target_expr)));
```
SaswatPadhi: Same comment -- constructor calls `normalize` already. This should remain unchanged.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -112,14 +113,32 @@ void assigns_clauset::add_target(const exprt &target_expr)
...
+  freely_assignable_set.erase(targett(*this, targett::normalize(expr)));
 }
 
 goto_programt assigns_clauset::generate_havoc_code() const
 {
   modifiest modifies;
-  for(const auto &target : targets)
-    modifies.insert(target.address.object());
+  for(const auto &target : write_set)
+    modifies.insert(to_address_of_expr(target.address).object());
```
SaswatPadhi: But `address` is already an `address_of_exprt`

SaswatPadhi (changes requested):
I am leaving some initial comments below, but I think we should discuss these two sets that are being maintained within assigns clause now. I think an assigns clause should just refer to the assigns clause on a function / loop -- it shouldn't be modified with other stuff.

We could create a separate class, may be called `assignable_set`, which would represent a union of the assigns clause + freely assignable local symbols. Let's discuss more offline.

src/goto-instrument/contracts/assigns.h
feliperodri: You're correct. I fixed the hash function as well.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Correct, we don't need to normalize this here again and we can completely remove `expr`. Done.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Correct, remove it.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Correct. This is unnecessary. Removed it.

regression/contracts/assigns_validity_pointer_04/test.desc
SaswatPadhi: Thanks for explaining offline. Resolving this.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -112,13 +111,30 @@ void assigns_clauset::add_target(const exprt &target_expr)
...
+  {
+    log.warning() << "Ignored duplicate expression '"
+                  << from_expr(ns, expr.id(), expr) << "' in assigns clause at "
+                  << expr.source_location().as_string() << messaget::eom;
+  }
+}
+
+void assigns_clauset::remove_freely_assignable_expr(const exprt &expr)
+{
+  freely_assignable_set.erase(targett(*this, targett::normalize(expr)));
```
SaswatPadhi:
```suggestion
  freely_assignable_set.erase(targett(*this, expr));
```
since `normalize` is called by the constructor.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -145,11 +166,11 @@ exprt assigns_clauset::generate_containment_check(const exprt &lhs) const
 exprt assigns_clauset::generate_subset_check(
   const assigns_clauset &subassigns) const
```
SaswatPadhi, line 164: This function needs to be updated for our "two sets" scheme now. We are only checking that:
`subassigns.write_set ⊆ write_set`.

I think we should be checking that:
`(subassigns.freely_assignable_set = {}) ∧ (subassigns.write_set ⊆ write_set ∪ freely_assignable_set)`

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -112,13 +111,30 @@ void assigns_clauset::add_target(const exprt &target_expr)
...
+{
+  freely_assignable_set.erase(targett(*this, targett::normalize(expr)));
 }
 
 goto_programt assigns_clauset::generate_havoc_code() const
 {
   modifiest modifies;
-  for(const auto &target : targets)
+  for(const auto &target : write_set)
     modifies.insert(target.address.object());
```
SaswatPadhi, line 135: For completeness, we should also havoc the `freely_assignable_set`. So we need a similar loop for that.

SaswatPadhi (changes requested):
Some more comments.

regression/contracts/assigns_validity_pointer_04/test.desc
feliperodri: For this case, it's OK to track `1` of "anything". Pointer primitive check can generate many properties to check and I want to make sure that there is only 1 failure: `Check that z is assignable`.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Done.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Implemented as suggested.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Implemented as suggested.

regression/contracts/assigns_enforce_18/test.desc
feliperodri: I ran this regression test with our `develop` branch and included the failed checks in the negative regex.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -89,17 +88,18 @@ assigns_clauset::assigns_clauset(
...
 {
   for(const auto &target_expr : expr.operands())
   {
-    add_target(target_expr);
+    add_global_write_set(target_expr);
   }
 }
 
-void assigns_clauset::add_target(const exprt &target_expr)
+void assigns_clauset::add_global_write_set(const exprt &target_expr)
```
SaswatPadhi: [nitpick]

I liked the names we discussed, `add_to_global_write_set` and `remove_from_global_write_set` better. The current names are a bit confusing because it sounds like we are adding a whole set or removing a whole set.

Before it was `add_target` and that made sense, because we were adding a new target.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -145,11 +170,15 @@ exprt assigns_clauset::generate_containment_check(const exprt &lhs) const
 exprt assigns_clauset::generate_subset_check(
   const assigns_clauset &subassigns) const
 {
-  if(subassigns.targets.empty())
+  if(subassigns.global_write_set.empty())
     return true_exprt();
```
SaswatPadhi, line 168: We should add an `INVARIANT` at this point to make sure we don't internally call this function on a `subassigns` that has a non-empty `local_write_set`, because `subassigns.local_write_set` is entirely ignored in the rest of the function -- we actually don't expect it for the callee's assigns clause.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -145,11 +170,15 @@ exprt assigns_clauset::generate_containment_check(const exprt &lhs) const
...
 
+  std::unordered_set<targett, targett::hasht> caller_write_set(
+    global_write_set);
+  caller_write_set.insert(local_write_set.begin(), local_write_set.end());
+
```
SaswatPadhi, line 173: I think it might be cleaner (and more efficient) to just have another loop. The loop only has a single statement in it anyway. What do you think?

regression/contracts/assigns_enforce_19/main.c
```diff
@@ -0,0 +1,33 @@
...
+  a++;
+  return a;
+}
+
+int g(int *points_to_b, int *points_to_c)
+  __CPROVER_assigns(b) // Error: it should also mention c
+{
+  b = 1;
+  c = 2;
+  *points_to_b = 1;
```
SaswatPadhi, line 18: Nice test case! We are now correctly looking at assignable (and freely assignable) "addresses" instead of symbols :)

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -110,32 +110,57 @@ void assigns_clauset::add_target(const exprt &target_expr)
...
+}
+
+void assigns_clauset::add_local_write_set(const exprt &expr)
+{
+  auto insertion_succeeded = local_write_set.emplace(*this, expr).second;
+
+  if(!insertion_succeeded)
+  {
+    log.warning() << "Ignored duplicate expression '"
+                  << from_expr(ns, expr.id(), expr) << "' in assigns clause at "
```
SaswatPadhi: This warning message can be a little confusing, since the user-specified targets only appear within `global_write_set`.

Should we just remove this warning and silently drop any local targets that we might internally add accidentally? Or, we could fatally fail on an `INVARIANT` failure here if we want to be absolutely sure that we never double-add local targets, but I think that might be too aggressive.

SaswatPadhi (commented):
Thanks! :) A few last comments

src/goto-instrument/contracts/assigns.cpp
feliperodri: Implemented as suggested.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Implemented as suggested.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Implemented as suggested.

src/goto-instrument/contracts/assigns.cpp
feliperodri: Implemented as suggested.

regression/contracts/assigns_enforce_18/test.desc
```diff
@@ -0,0 +1,22 @@
+CORE
+main.c
+--enforce-all-contracts _ --pointer-primitive-check
+^EXIT=10$
+^SIGNAL=0$
+^\[bar.\d+\] line \d+ Check that \*b is assignable: SUCCESS$
```
SaswatPadhi, line 7: We don't check that `a` is assignable at `free(a)`, or may be we do but don't have a regex for it?

regression/contracts/assigns_validity_pointer_04/test.desc
SaswatPadhi: I am a little confused as to why only 1 assertion fails here though.

The assigns clause for `foo` doesn't have `z`, nor does it have `*z`. So shouldn't we fail 2 assertions:
1. the containment check @ assignment `z = malloc...`
2. the subset check @ call `baz()` in the next line (which has `*z` in the assign clause for `baz` but not `foo`)?

SaswatPadhi (approved):
Awesome 🎉

2 comments on the regression tests, but otherwise LGTM.

regression/contracts/assigns_validity_pointer_04/test.desc
feliperodri: We discussed this offline. Two important points:
1. There is no containment check here, `malloc` is not treated as a normal function call.
2. There is a bug in checking the subset relation which is not stressed by this test case. We will address it on a separate PR.

regression/contracts/assigns_enforce_18/test.desc
feliperodri: Good catch!! I just sent a new commit to check frame conditions on free function calls.

regression/contracts/assigns_validity_pointer_04/test.desc
SaswatPadhi: > There is no containment check here, `malloc` is not treated as a normal function call.

Hmm, but `malloc` just returns a fresh allocation, the left-hand side of the assignment in `z = malloc...`, i.e. the variable `z`, must be checked for containment in the assigns clause.

> There is a bug in checking the subset relation which is not stressed by this test case. We will address it on a separate PR.

On digging a little bit deeper, I realized that the we don't actually need to separately check the assigns clause during replacement. We can go ahead and havoc directly, and those havocs would fail if something is not assignable in the calling context.

Issue #6149, opened 2021-05-25
Document quantifiers & code contracts support


SaswatPadhi:
Just a TODO ticket to properly document the code contracts support (Felipe -> function contracts, I -> loop contracts) at some point.

As Felipe suggested @ https://github.com/diffblue/cbmc/pull/6145#discussion_r638326431, we should also document the restricted quantifier support with the SAT backend. If this is already documented somewhere please point us to it.

martin-cs:
Yes; a page in the manual about code contracts and how to do deductive verification would be *amazing* and would be make accessible all of the great work you have been doing.

I don't think the limitations of the SAT backend are documented anywhere.

PR #6332, opened 2021-09-04
Documentation for decreases clauses

SaswatPadhi:
Related to #6149.

Thanks to @LongPham7 for an initial draft!

- n/a ~~Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).~~

### DIFF
--- a/doc/cprover-manual/contracts-decreases.md
+++ b/doc/cprover-manual/contracts-decreases.md
@@ -1,3 +1,172 @@
-# Decreases Clause
+# Decreases Clauses
 
-TODO: Document `__CPROVER_decreases`
+A _decreases_ clause specifies a measure that must strictly decrease at every iteration of a loop.
+By demonstrating that the measure
+
+1. is bounded from below, and
+2. strictly decreases at each iteration
+
+we can prove termination of loops.
+This is because the measure must eventually hit the lower bound
+at which point the loop must terminate,
+since the measure cannot strictly decrease further.
+This technique for proving termination was proposed by Robert Floyd,
+and interested readers may refer to his seminal paper
+"[_Assigning Meaning to Programs_](https://people.eecs.berkeley.edu/~necula/Papers/FloydMeaning.pdf)".
+
+### Syntax
+
+A one-dimensional (1D) decreases clause for a loop is an arithmetic expression `e`
+over the variables visible at the same scope as the loop,
+specified as `__CPROVER_decreases(e)`.
+
+Like invariant clauses, decreases clauses may be specified just after the loop guard.
+An example of a 1D decreases clause is shown below.
+
+```c
+for(int i = 0; i < n; i += 2)
+__CPROVER_loop_invariant(0 <= i && i <= n)
+__CPROVER_decreases(n - i)
+{ ... }
+```
+
+Please see the [invariant clauses](contracts-invariants.md) page
+for more examples on `for` and `do...while` loops.
+
+To help prove termination of more complex loops,
+CBMC also supports multi-dimensional decreases clauses.
+A multi-dimensional decreases clause is an [ordered tuple](https://en.wikipedia.org/wiki/Tuple)
+of arithmetic expressions, specified as `__CPROVER_decreases(e_1, e_2, ..., e_n)`.
+An example of a multi-dimensional decreases clause is given below.
+
+```c
+while(i < n)
+__CPROVER_loop_invariant(0 <= i && i <= n)
+__CPROVER_loop_invariant(0 <= j && j <= n)
+__CPROVER_decreases(n - i, n - j)
+{
+  if (j < n)
+    j++;
+  else
+  {
+    i++;
+    j = 0;
+  }
+}
+```
+
+We extend the strict arithmetic comparison for 1D decreases clauses
+to a strict [lexicographic comparison](https://en.wikipedia.org/wiki/Lexicographic_order)
+for multi-dimensional decreases clauses.
+
+**Important.**
+Like invariant clauses, decreases clauses must be free of side effects,
+for example, mutation of local or global variables.
+Otherwise, CBMC raises an error message during compilation:
+```
+Decreases clause is not side-effect free. (at: file main.c line 4 function main) 
+```
+
+### Semantics
+
+A decreases clause extends the loop abstraction introduced in the [invariants clause](contracts-invariants.md) manual.
+In addition to the inductiveness check asserted at the end of a single arbitrary iteration,
+CBMC would also assert the strict decrement of the measure specified in the decreases clause.
+At a high level, in addition to the assumptions and assertions introduced by the invariant clause,
+a decreases clause expands to three key steps:
+1. At the beginning of the loop body, record the initial value of the measure specified in the decreases clause.
+2. At the end of the loop body, record the final value of the measure specified in the decreases clause.
+3. After the loop iteration, assert that the final value is strictly smaller than the initial one.
+
+For a 1D decreases clause, we use the strict arithmetic comparison (i.e., `<`).
+For a multi-dimensional decreases clause, say `(e_1, ..., e_n)`,
+we extend the strict arithmetic comparison to a strict lexicographic comparison.
+
+As an example, consider our binary search implementation again,
+this time with a decreases clause annotation to prove its termination:
+
+```c
+int binary_search(int val, int *buf, int size)
+{
+  if(size <= 0 || buf == NULL) return NOT_FOUND;
+
+  long long lb = 0, ub = size - 1;
+  long long mid = ((unsigned int)lb + (unsigned int)ub) >> 1;
+
+  while(lb <= ub)
+  __CPROVER_loop_invariant(0L <= lb && lb - 1L <= ub && ub < size)
+  __CPROVER_loop_invariant(mid == ((unsigned int)lb + (unsigned int)ub) >> 1)
+  __CPROVER_decreases(ub - lb)
+  {
+     if(buf[mid] == val) break;
+     if(buf[mid] < val)
+       lb = mid + 1;
+     else
+       ub = mid - 1;
+     mid = ((unsigned int)lb + (unsigned int)ub) >> 1;
+  }
+  return lb > ub ? NOT_FOUND : mid;
+}
+```
+
+The instrumented GOTO program is conceptually similar to the following high-level C program:
+
+```c
+int binary_search(int val, int *buf, int size)
+{
+  if(size <= 0 || buf == NULL) return NOT_FOUND;
+
+  long long lb = 0, ub = size - 1;
+  long long mid = ((unsigned int)lb + (unsigned int)ub) >> 1;
+
+  /* 1. assert invariant at loop entry */
+  assert(0L <= lb && lb - 1L <= ub && ub < size);
+  assert(mid == ((unsigned int)lb + (unsigned int)ub) >> 1);
+
+  /* 2. create a non-deterministic state for modified variables */
+  havoc(lb, ub, mid);
+
+  /* 3. establish invariant to model state at an arbitrary iteration */
+  __CPROVER_assume(0L <= lb && lb - 1L <= ub && ub < size);
+  __CPROVER_assume(mid == ((unsigned int)lb + (unsigned int)ub) >> 1);
+
+  /* 4. perform a single arbitrary iteration (or exit the loop) */
+  if(lb <= ub)
+  {
+    /* 5. declare variables for tracking the loop variant */
+    int old_measure, new_measure;
+
+    /* 6. evaluate the variant at the start of the loop body */
+    old_measure = ub - lb;
+
+    if(buf[mid] == val) break;
+    if(buf[mid] < val)
+      lb = mid + 1;
+    else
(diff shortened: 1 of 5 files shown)


doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,183 @@
-# Decreases Clause
+# Decreases Clauses
 
-TODO: Document `__CPROVER_decreases`
+A _decreases_ clause specifies a measure that must strictly decrease
+monotonically at every iteration of a loop.
```
jimgrundy: I would simply say "must strictly decrease at every iteration" rather than "strictly decrease monotonically at every iteration". Monotonically decreasing means decreases or remains the same, but never increasing. Strictly decreasing is stronger so although adding "monotonically" doesn't alter the meaning it's presence only serves to add a bit of doubt as to what you meant.

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,183 @@
...
+at which point the loop must terminate,
+since the measure cannot strictly decrease further.
+
+In the scientific literature, this measure is also called a _loop variant_ or _ranking function_.
+Although they can be arbitrary (strict) well-founded relations on arbitrary domains,
+we restrict decreases clauses to (strict) arithmetic inequalities.
+Unlike mathematical integers, arithmetic types in C are all bounded from below.
+Furthermore, unlike mathematical real numbers, floating-point arithmetic types in C are not dense.
+Hence, any strictly decreasing arithmetic expression in C can be used as a decreases clause.
+
```
jimgrundy, line 16: Echoing Daniel's comments to Felipe on earlier documents, I wouldn't try to explain the science here. Maybe we write a background document for folks, but people come to a man-page type document like this for a short and sweet description of what a feature does and how should they use it. An explanation of how/why it works is done better in another format.

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,183 @@
...
+}
+```
+
+We extend the strict arithmetic comparison for 1D decreases clauses
+to a strict [lexicographic comparison](https://en.wikipedia.org/wiki/Lexicographic_order)
+for multi-dimensional decreases clauses.
+
+**Important.**
+Like invariant clauses, decreases clauses must be free of side effects,
+for example, mutation of local or global variables.
```
jimgrundy, line 64: Do we check that? Are we telling people to take care here lest they produce an unsound result, or are we telling them that if they make this mistake the tool will complain with an error. If so, it might be nice to describe the error they will get.

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,183 @@
...
+1. At the beginning of the loop body, record the initial value of the measure specified in the decreases clause.
+2. At the end of the loop body, record the final value of the measure specified in the decreases clause.
+3. After the loop iteration, assert that the final value is strictly smaller than the initial one.
+
+For a 1D decreases clause, we use the strict arithmetic comparison (i.e., `<`).
+For a multi-dimensional decreases clause, say `(e_1, ..., e_n)`,
+we extend the strict arithmetic comparison to a strict lexicographic comparison.
+Formally, given two evaluations `(a_1, ..., a_n)` and `(b_1, ..., b_n)`,
+a strict lexicographic comparison, denoted by `(a_1, ..., a_n) ≺ (b_1, ..., b_n)`,
+is defined as the predicate:
```
jimgrundy: Kind of repeating myself, but I think I would be better served by a separate document that described the unbounded verification feature and how it works (both mathematically and how implemented in CBMC) and keep this document as a light weight description of what it does and how to use it.

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: Yes, "monotonically decreases" / "decreases monotonically" means what you mentioned, but "strictly decreases monotonically" means a measure decreases consistently (doesn't remain the same / increases) and [is a widely used](https://www.google.com/search?q=%22strictly+decreases+monotonically%22) phrase. I don't think it's incorrect, but I agree that it may be confusing, so I'll change this as you suggested.

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: Thanks; makes sense. This paragraph could be removed.

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: Yes, we do check this and generate an error as shown in this regression test:
https://github.com/diffblue/cbmc/blob/3e0007f9ed818969185676eb96f8811dc75e8705/regression/contracts/variant_side_effects_fail/test.desc#L4

I would add this message to this doc & the invariants doc.

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: I agree on the previous comment, but this one seems relevant to end users. It's not describing the science, or "why" comparison of tuples is defined this way -- there are many different ways one may define a comparison over tuples), but is only describing "how" tuples are compared (i.e., lexicographically) which I think end users must know in order to be able to write multi-dimensional decreases clauses.

TGWDB (approved):

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,179 @@
-# Decreases Clause
+# Decreases Clauses
 
-TODO: Document `__CPROVER_decreases`
+A _decreases_ clause specifies a measure that must strictly decrease at every iteration of a loop.
+By demonstrating that the measure
+1. is bounded from below, and
+2. strictly decreases at each iteration,
+we can prove termination of loops.
+This is because the measure must eventually hit the lower bound
```
feliperodri: You're appending the last paragraph in `2.`. Add an empty line between them (or write them as a single paragraph).

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,179 @@
-# Decreases Clause
+# Decreases Clauses
 
-TODO: Document `__CPROVER_decreases`
+A _decreases_ clause specifies a measure that must strictly decrease at every iteration of a loop.
```
feliperodri, line 3: Do you have a link to any external resources with more details in the theory behind decreases clauses? It could be useful to add it here.

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,179 @@
...
+we can prove termination of loops.
+This is because the measure must eventually hit the lower bound
+at which point the loop must terminate,
+since the measure cannot strictly decrease further.
+
+### Syntax
+
+A one-dimensional (1D) decreases clause for a loop is an arithmetic expression `e`
+over the variables visible at the same scope as the loop,
+specified as `__CPROVER_decreases(e)`.
```
feliperodri, line 21: It doesn't have to be only arithmetic expressions, right? What about `__CPROVER_decreases(i)`?

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,179 @@
...
+Like invariant clauses, decreases clauses must be free of side effects,
+for example, mutation of local or global variables.
+Otherwise, CBMC raises an error message during compilation:
+```
+Decreases clause is not side-effect free. (at: file main.c line 4 function main) 
+```
+
+### Semantics
+
+A decreases clause extends the loop abstraction introduced in the [invariants clause](contracts-invariants.md) document.
```
feliperodri:
```
s/document/manual/g
```

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,179 @@
...
+}
+```
+
+The instrumented code points (5), (6), (8), and (9) are specific to the decreases clause.
+
+**Important.** 
+Decreases clauses work in conjunction with [loop invariants](contract-invariants.md),
+which model an arbitrary loop iteration at which the decreases clause is checked.
+If a decreases clause is annotated on a loop without an invariant clause,
+then the weakest possible invariant (i.e, `true`) is used to model an arbitrary iteration.
```
feliperodri, line 172: In this case, a successful verification will always means the loops terminates, but can't guarantee memory safety of the implementation?

feliperodri (changes requested):

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: Yes, exactly

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: Numeric variables by themselves are valid arithmetic expressions too -- [any expression whose return type is numeric is an arithmetic expression](https://www.cis.upenn.edu/~matuszek/General/JavaSyntax/arithmetic-expressions.html). `5`, `x` (for numeric `x`), `x+5`, `x <= 0 ? -x : x` are all valid arithmetic expressions.

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: Good catch! Thanks

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: Good point; I'll add some resources.

feliperodri:
@SaswatPadhi can we close #6149 after merging this PR? The only initial documentation missing will be assigns clause for loops, correct?

SaswatPadhi:
> @SaswatPadhi can we close #6149 after merging this PR? The only initial documentation missing will be assigns clause for loops, correct?

I think we can close it after this + assigns clause documentation is merged. I would push the assigns clause documentation to #6249

feliperodri (approved):

SaswatPadhi:
I have addressed all prior comments and fixed some build issues (with the new binary_search regression test).

Keeping this PR open for one final review from @jimgrundy.

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,184 @@
...
+Formally, given two evaluations `(a_1, ..., a_n)` and `(b_1, ..., b_n)`,
+a strict lexicographic comparison, denoted by `(a_1, ..., a_n) ≺ (b_1, ..., b_n)`,
+is defined as the predicate:
+
+```c
+   (a_1 < b_1)
+|| (a_1 == b_1 && a_2 < b_2)
+|| ...
+|| (a_1 == b_1 && ... && a_n < b_n)
+```
```
jimgrundy: I would suggest cutting the material after "strict lexicographic comparison". Maybe hyperlink off "lexicographic comparison" to the article you linked earlier, and you can skip the explanation of how you implement a lexicographic compare. If people know what one is already they don't need to see this code. If they don't, they might be better off reading the code. But, if you like it, then I'm OK with it.

doc/cprover-manual/contracts-decreases.md
```diff
@@ -1,3 +1,184 @@
...
+
+The instrumented GOTO program is conceptually similar to the following high-level C program:
+
+```c
+int binary_search(int val, int *buf, int size)
+{
+  if(size <= 0 || buf == NULL) return NOT_FOUND;
+
+  long long lb = 0, ub = size - 1;
+  long long mid = (lb + ub) / 2;
```
jimgrundy: This works because we use "long long" values to avoid overflow. But, we can do the computation another way and avoid overflow while using fewer bits, which might give better solver performance. Take a look at this article for a nice clear way to express the algorithm without the extra bits: [Extra, Extra - Read All About It: Nearly All Binary Searches and Mergesorts are Broken](https://ai.googleblog.com/2006/06/extra-extra-read-all-about-it-nearly.html). Now, since the use of int for the arguments and long long for the computation should mean this is correct (are there machines for which int is the same size as long long?) then I'm fine with leaving this as is, but I think you might like the version in this article better anyway

jimgrundy (approved):

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: Thanks! :D Great point!

> Now, since the use of int for the arguments and long long for the computation should mean this is correct (are there machines for which int is the same size as long long?)

It's correct (verified) only on 64-bit Windows and Linux machines -- the CBMC CI machines & my machine.
I originally had `long`, and that failed because on Windows `int` == `long`, apparently! Then I changed to `long long` and both Windows + Linux were happy. But that's still fragile -- even `long long` could be the same as `int` on some platform.

I implemented the change. The verification time with MiniSAT remains the same (~2.7s), but the number of variables in the SAT formula goes from ~12900 down to ~7700.

doc/cprover-manual/contracts-decreases.md
SaswatPadhi: Okay, I could remove it. I wasn't sure if end users would be familiar with exactly what lexicographic comparison on ordered tuples would mean. I didn't want to point them to research papers just for that, but I guess they could Google it if they are unsure :)

PR #6334, opened 2021-09-08
Refactor assigns clause implementation

SaswatPadhi:
I was planning to merge these changes as part of #6249, but given the volume of changes, I thought it would be best to create a separate PR.

Main changes are:
+ bug fixes in containment (previously called alias) and subset (previously called subset) checks
+ usused fields (e.g., `function_id`, `init_block`) are removed,
+ redundant functions (e.g. `pointer_for`) are removed,
+ const references are used throughout consistently
+ targets are tracked in an `unordered_set` as opposed to `vector`
+ introduced a "normalize" method for extracting base object of arrays

### DIFF
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -110,6 +110,11 @@ void assigns_clauset::add_target(const exprt &target_expr)
   }
 }
 
+void assigns_clauset::remove_target(const exprt &target_expr)
+{
+  targets.erase(targett(*this, targett::normalize(target_expr)));
+}
+
 goto_programt assigns_clauset::generate_havoc_code() const
 {
   modifiest modifies;
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -22,7 +22,10 @@ Date: July 2021
 assigns_clauset::targett::targett(
   const assigns_clauset &parent,
   const exprt &expr)
-  : expr(address_of_exprt(expr)), id(expr.id()), parent(parent)
+  : address(address_of_exprt(normalize(expr))),
+    expr(expr),
+    id(expr.id()),
+    parent(parent)
 {
 }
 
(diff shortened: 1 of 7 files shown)


SaswatPadhi:
Marked `draft` for now.

`assigns_enforce_16` is failing but I cannot reproduce the failure locally!

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -19,129 +19,107 @@ Date: July 2021
...
+  : expr(address_of_exprt(expr)), id(expr.id()), parent(parent)
 {
-  INVARIANT(
-    target.type().id() == ID_pointer,
-    "Assigns clause targets should be stored as pointer expressions.");
 }
 
-assigns_clause_targett::~assigns_clause_targett()
-{
-}
```
feliperodri: Why remove it? Don't we need to clean any memory for `assigns_clause_targett` objects?

feliperodri (approved):

src/goto-instrument/contracts/assigns.cpp
SaswatPadhi: Now we store `targett` (not `targett *`) in our `unordered_set`. We use `emplace` when inserting and don't seem to be making any copies elsewhere as well, so I thought this was cleaner than keeping raw pointers around.

Once the raw pointers and calls to `new` were removed, there's no leftover memory to clean.

SaswatPadhi:
> Marked `draft` for now.
> 
> `assigns_enforce_16` is failing but I cannot reproduce the failure locally!

Fixed this by removing dead variables from freely assignable set all together (although the implementation is still incorrect and, as discussed offline, @feliperodri would make a separate PR to fix that properly).

But, what's concerning here is that `goto-cc` / `goto-instrument` was generating different goto binaries on my machine vs CI.

EDIT: The remaining CI issues with Windows tests seem completely unrelated.

jimgrundy:
I realize I'm late here and that the coding practices for this code base are established ... but I would want to see more comments in the code. While particularly valuable for a new comer like myself, a header comment per method stating what it does, and maybe also per method argument/result, would help a great deal. I imagine one day being able to run doxygen on the code and learning something useful from it.

SaswatPadhi:
Will push a few more things before final review & merge, so marked as `draft`.

feliperodri (approved):
Thank you for all the subsequent clean up and fixes!

Issue #6168, opened 2021-06-10
`natural_loop`s not computed correctly in presence of function pointers [blocks #5530]


SaswatPadhi:
**CBMC version:** 556b4325
**Operating system:** Mac OS 10.15.7

**Test case:**
```c
#include <assert.h>
#include <stdlib.h>

static int adder(const int *a, const int *b)
{
  return (*a + *b);
}

int main()
{
  int x = 1024;

  int (*local_adder)(const int *, const int *) = adder;

  while(x > 0)
  __CPROVER_loop_invariant(1 == 1)
  {
    x += local_adder(&x, &x); // loop detection fails
    //x += adder(&x, &x);       // works fine
  }
}
```

**Exact command line resulting in the issue:**

```shell
$ goto-cc test.c -o test.gb
$ goto-instrument --enforce-all-contracts test.gb test.2.gb
$ cbmc test.2.gb
```

**What behaviour did you expect:**

The loop invariant would be used and unwinding wouldn't be necessary.

**What happened instead:**

The loop invariant is not used because the loop in `main` is not recognized.

Note that if we use `adder` in place of `local_adder` within the loop, then the loop is detected and the loop invariant is processed as expected.

**Additional context:**

If we use `adder` in place of `local_adder` above, i.e. call the function directly as opposed to using a function pointer, then loop is detected and invariant is processed as expected.

The following instrumentation code:

```c
for(auto &goto_function : goto_model.goto_functions.function_map) {
  natural_loops_mutablet natural_loops(goto_function.second.body);
  log.warning() << goto_function.first.c_str() << " has "
                << natural_loops.loop_map.size() << " loops."
                << messaget::eom;
}
```
produces the following output when using `local_adder` function pointer

```
Reading GOTO program from 'test.gb'
main has 0 loops.
adder has 0 loops.
__CPROVER__start has 0 loops.
__CPROVER_initialize has 0 loops.
Writing GOTO program to 'test.2.gb'
```

and produces the following output when using `adder` directly

```
Reading GOTO program from 'test.gb'
main has 1 loops.
adder has 0 loops.
__CPROVER__start has 0 loops.
__CPROVER_initialize has 0 loops.
Writing GOTO program to 'test.2.gb'
```

SaswatPadhi:
cc: @feliperodri @tautschnig

feliperodri:
cc. @chrisr-diffblue

martin-cs:
That's horrible :-\

[...]

PR #6340, opened 2021-09-14
Add workaround for issue described in #6168.

NlightNFotis:
This fixes the infinite unwinding by adding calls to `do_indirect_call_and_rtti_removal()`
before the contracts code invocation in `goto-instrument/`.

This is not yet finished, as it will need a test-case attached, but this is pushed so that
it can garner some initial comments regarding the approach.

### DIFF
--- a/regression/contracts/github_6168_infinite_unwinding_bug/main.c
+++ b/regression/contracts/github_6168_infinite_unwinding_bug/main.c
@@ -1,0 +1,21 @@
+#include <assert.h>
+#include <stdlib.h>
+
+static int adder(const int *a, const int *b)
+{
+  return (*a + *b);
+}
+
+int main()
+{
+  int x = 1024;
+
+  int (*local_adder)(const int *, const int *) = adder;
+
+  while(x > 0)
+    __CPROVER_loop_invariant(1 == 1)
+    {
+      x += local_adder(&x, &x); // loop detection fails
+      //x += adder(&x, &x);       // works fine
+    }
+}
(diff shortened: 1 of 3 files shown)


NlightNFotis:
Hi @SaswatPadhi, would it be possible to have a look at this PR? It works-around the issue described in #6168 using Martin's suggestion.

However, I couldn't get it to work without introducing a call to `apply_loop_contracts()` in `enforce_contracts()`. I'm not sure about this step, but this is what was preventing the change in latest `develop` from working (it worked cleanly against the version described in the issue, `cbmc-5.32.0`).

Besides, I'm not sure if that addition is changing the semantics of your contract code. It seems that in https://github.com/diffblue/cbmc/commit/9cdca480ed915041118734380b418c69d5878390#diff-1ff6abca57de19a68456666ed5c25ce09fcaae20fdc9c2b109cb60cb311c183aL1208 the call to `apply_loop_contracts()` is removed from a branch that could be exercised, without an equivalent call being reachable elsewhere, but that's just my understanding. Please let me know if there are issues with the current approach.

SaswatPadhi:
Thanks for implementing the workaround, @NlightNFotis!

We separated the function contract flags and loop contracts flags recently, so `--enforce-all-contracts` only enforces function contracts (for all functions). Loop contracts don't have separate enforcement and replacement steps, just a single apply step using the flag `--apply-loop-contracts`.

So you could replace the `--enforce-all-contracts` flag with `--apply-loop-contracts` and undo the new call to `apply_loop_contracts()`. Thanks!

NlightNFotis:
Thanks @SaswatPadhi, the latest suggestions have now been implemented, and I moved the test to the `regression/contracts` folder.

This patch should now fix the issue you are facing in #6168 if you call `goto-instrument --apply-loop-contracts`.

SaswatPadhi (approved):
Awesome! Thank you so much, @NlightNFotis.

Looks like that resolved all of the CI errors too.

peterschrammel (approved):

PR #6342, opened 2021-09-14
Annotate library function strlen with code contracts 

qinheping:
This PR 
1. added function contracts to the library function `strlen`;
2. added loop invariant to the loop in `strlen`.

However, we can't run unbounded proof with these code contracts. The blocker is that these contracts contain quantified expressions. With the current SAT engine, quantified expressions require constant bound (e.g., `MAX_BUFFER_SIZE`) on variables.

The hope is that, with SMT backend, we can remove the constant bounds and run unbounded proof on benchmarks calling `strlen`.


src/ansi-c/library/string.c
```diff
@@ -554,7 +554,20 @@ __CPROVER_HIDE:;
 
 #undef strlen
 
-inline size_t strlen(const char *s)
+inline size_t strlen(const char *s)  
+  __CPROVER_requires(s != NULL && 
+    __CPROVER_exists {
+      int i;
+      (0 <= i && i < MAX_BUFFER_SIZE) ==> (i < __CPROVER_OBJECT_SIZE(s)) ==> s[i] == '\0'
```
feliperodri: `MAX_STRING_SIZE`would be more appropriate.

src/ansi-c/library/string.c
```diff
@@ -563,7 +576,16 @@ inline size_t strlen(const char *s)
   return __CPROVER_zero_string_length(s);
   #else
   __CPROVER_size_t len=0;
-  while(s[len]!=0) len++;
+  while(s[len]!=0) 
+  __CPROVER_loop_invariant (
+        (0 <= len) && (len < __CPROVER_OBJECT_SIZE(s)) &&
+        ((len == 0) ||
```
SaswatPadhi: [nit]

Is `len == 0` clause needed?
Wouldn't `0 <= k && k < len ==> ...` vacuously pass anyway without it (`false ==> <anything>`)?

src/ansi-c/library/string.c
SaswatPadhi: I think neither of those actually belong within CBMC; both come from our harnesses. We have two options and I am not sure which is better:

1. Get rid of these constant bounds and write quantified expressions for the SMT backend, so simply:
```c
      (0 <= i && i < __CPROVER_OBJECT_SIZE(s)) ==> s[i] == '\0'
```
2. Define some constant bounds like `MAX_STRING_SIZE` for CBMC library functions, but then those would have to be set in addition to, and be consistent with, the harness-specific constants. This also raises more questions: How many such constant bounds to introduce into CBMC, like `MAX_BUFFER_SIZE` for `malloc` as well? What if the user wants to use two different bounds for different allocations? We have for example two bounds for some of the array functions: a MAX_ARRAY_SIZE and a MAX_ARRAY_ELEMENT_SIZE.

src/ansi-c/library/string.c
```diff
@@ -554,7 +554,20 @@ __CPROVER_HIDE:;
...
-inline size_t strlen(const char *s)
+inline size_t strlen(const char *s)  
+  __CPROVER_requires(s != NULL && 
+    __CPROVER_exists {
+      int i;
+      (0 <= i && i < MAX_BUFFER_SIZE) ==> (i < __CPROVER_OBJECT_SIZE(s)) ==> s[i] == '\0'
+  })
+  __CPROVER_ensures(
+    __CPROVER_return_value < __CPROVER_OBJECT_SIZE(s) &&
+    ((s[0] == '\0' && __CPROVER_return_value == 0) ||
```
SaswatPadhi: Same question. Do we need the special case of `__CPROVER_return_value == 0`?

src/ansi-c/library/string.c
qinheping: > Is `len == 0` clause needed?

You are right, we don't need the `len == 0` clause. The verification still successes after removing it.
> Wouldn't `0 <= k && k < len ==> ...` vacuously pass anyway without it (`false ==> <anything>`)?

The expression will vacuously pass when `len` is always `0`. But it is not the case here: `len` can be greater than `0`.

src/ansi-c/library/string.c
qinheping: The clause is not needed. I will remove it.

src/ansi-c/library/string.c
SaswatPadhi: Yes, I only meant vacuity in case of `len == 0`.

Issue #6343, opened 2021-09-15
Does jbmc build on MacOS?


markrtuttle:
CBMC version: 5.38
Operating system: MacOS 15.5.2 (Big Sur) on Apple M1

Does jbmc build on MacOS?

In the [cbmc formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/cbmc.rb) for brew, [Lines 35-36](https://github.com/Homebrew/homebrew-core/blob/b3387012ce55c4d6d41a7821050b4138ce2dc0d2/Formula/cbmc.rb#L35-L36) say that the java front end won't build on ARM.

I think the issue is that in [OpenJDK 17.13](https://github.com/openjdk/jdk) installed by brew under MacOS 15 on Apple M1, the March [commit a0c3f242](https://github.com/openjdk/jdk/commit/a0c3f2421862790cb64fb6d4a0d00c666345c090) deleted the methods isOtherUppercase and isOtherLowercase called on [Line 5463](https://github.com/diffblue/java-models-library/blob/198d541d3d33383f4c25f6cbd639bbe0d7b726ba/src/main/java/java/lang/Character.java#L5463) and [Line 5529](https://github.com/diffblue/java-models-library/blob/198d541d3d33383f4c25f6cbd639bbe0d7b726ba/src/main/java/java/lang/Character.java#L5529) of Character.java in DiffBlue's [java-models-library](https://github.com/diffblue/java-models-library.git) submodule.

In fact, if I delete these lines, then jbmc compiles on MacOS 15 on Apple M1.  Assuming brew installs OpenJDK 7.13 independent of ARM, isn't this a problem elsewhere?  Is it possible to rewrite these two lines without invoking these methods?

Edit: Changed OpenJDK version from 7.13 to 17.13.

thomasspriggs:
Isn't OpenJDK 7.13 rather out of date? If I am reading the POM file correctly, the models library is supposed to be built using Java version 8. It might be worth trying a newer version of OpenJDK to see if that fixes the build.

chrisr-diffblue:
I think 'OpenJDK 7.13' may be a typo for 'OpenJDK 17.3' ? Note also some version confusion on Apple M1 https://stackoverflow.com/questions/67782746/openjdk-16-shows-as-17-on-macos-apple-silicon-when-installed-with-homebrew

Possibly the two 'correct fixes' are:

[...]

PR #6344, opened 2021-09-15
Modify cmake build to build on MacOS

markrtuttle:
This pull request modifies the cmake build on MacOS to follow the instructions used in the [homebrew cbmc formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/cbmc.rb) for building cbmc on MacOS.  This should make it easy to build cbmc on MacOS everywhere with just `cmake -S . -B build -GNinja`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,7 +240,7 @@ cprover_default_properties(
     util
     xml)
 
-# Get the major version of the java installation used by maven.
+# Get the version of the java installation used by maven.
 #
 # The java in your PATH may not be the java used by maven.
 # Installing java and maven with `brew install openjdk maven` will install
@@ -249,9 +249,9 @@ cprover_default_properties(
 # and the cask-only installation directory /opt/homebrew/opt/openjdk/bin
 # will not be in your PATH by default.  See `brew info java`.
 #
-function(MAVEN_JAVA_MAJOR_VERSION VERSION)
+function(MAVEN_JAVA_VERSION VERSION)
     # Output of mvn --version includes a string like "Java version: 16.1.4"
-    # Return the major version number found in that string (like 16 in 16.1.4).
+    # Return the version number found in that string (like 16.1.4).
     # Return UNKNOWN if the major version number can't be found.
     execute_process(COMMAND mvn --version
                     OUTPUT_VARIABLE STDOUT
@@ -266,10 +266,11 @@ function(MAVEN_JAVA_MAJOR_VERSION VERSION)
         set(OUTPUT "Java version: ") # ensure VERSION=="UNKNOWN" below
     endif()
 
-    string(REGEX REPLACE ".*Java version: ([0-9]*).*" "\\1" FOUND ${OUTPUT})
+    string(REGEX REPLACE ".*Java version: ([0-9.]*).*" "\\1" FOUND ${OUTPUT})
     if(FOUND STREQUAL OUTPUT)
         message(STATUS
-                "Warning: Failed to find java version in mvn --version output: ${OUTPUT}")
+                "Warning: Failed to find java version in mvn --version output: "
+                "${OUTPUT}")
         set(FOUND "") # guarantee VERSION=="UNKNOWN" below
     endif()
 
@@ -279,15 +280,18 @@ function(MAVEN_JAVA_MAJOR_VERSION VERSION)
     set(${VERSION} ${FOUND} PARENT_SCOPE)
 endfunction()
 
-MAVEN_JAVA_MAJOR_VERSION(JAVA_VERSION)
+MAVEN_JAVA_VERSION(JAVA_VERSION)
 message(STATUS "Found Maven uses Java version ${JAVA_VERSION} by default")
 
 option(WITH_JBMC "Build the JBMC Java front-end" ON)
-if(JAVA_VERSION GREATER 16)
+if(NOT JAVA_VERSION VERSION_LESS 17)
   # JBMC won't build with OpenJDK version 17.
   # See https://github.com/diffblue/cbmc/issues/6343
-  message(STATUS "Setting WITH_JBMC to OFF: Java version ${JAVA_VERSION} > 16")
+  message(STATUS "Setting WITH_JBMC to OFF: Java version ${JAVA_VERSION} >= 17")
   set(WITH_JBMC OFF)
+endif()
+if(JAVA_VERSION STREQUAL "UNKNOWN")
+  message(STATUS "Use cmake option -DWITH_JBMC=OFF if jbmc fails to build.")
 endif()
 if(WITH_JBMC)
     add_subdirectory(jbmc)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,15 @@
 cmake_minimum_required(VERSION 3.2)
 
+# Compile with /usr/bin/clang on MacOS
+# See https://github.com/diffblue/cbmc/issues/4956
+#
+if (CMAKE_HOST_SYSTEM_NAME STREQUAL Darwin)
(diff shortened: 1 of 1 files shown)


thomasspriggs (commented):
I am not sure that forcing these settings on certain platforms, rather than fixing the build environment is the right approach. @NlightNFotis what do you think?

CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
...
 
+# Compile with /usr/bin/clang on MacOS
+# See https://github.com/diffblue/cbmc/issues/4956
+if (CMAKE_HOST_SYSTEM_NAME STREQUAL Darwin)
+    if(NOT DEFINED CMAKE_C_COMPILER)
+        message(STATUS "Setting CMAKE_C_COMPILER to /usr/bin/clang for MacOS")
+        set(CMAKE_C_COMPILER "/usr/bin/clang")
+    endif()
+endif()
+
```
chrisr-diffblue, line 12: This seems reasonable, if rather ugly - but probably the best we can do without external changes.

CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
 cmake_minimum_required(VERSION 3.2)
 
+# Compile with /usr/bin/clang on MacOS
+# See https://github.com/diffblue/cbmc/issues/4956
+if (CMAKE_HOST_SYSTEM_NAME STREQUAL Darwin)
+    if(NOT DEFINED CMAKE_C_COMPILER)
+        message(STATUS "Setting CMAKE_C_COMPILER to /usr/bin/clang for MacOS")
+        set(CMAKE_C_COMPILER "/usr/bin/clang")
```
NlightNFotis, line 9: This is a nice change - it's already documented in our [`COMPILING.md`](https://github.com/diffblue/cbmc/blob/develop/COMPILING.md) file that we need to do this by hand on MacOS, so I'm all for the automation here 👍🏻

CMakeLists.txt
```diff
@@ -231,6 +240,18 @@ cprover_default_properties(
     xml)
 
 option(WITH_JBMC "Build the JBMC Java front-end" ON)
+
+# Don't build jbmc on MacOS running on ARM
+# See https://github.com/Homebrew/homebrew-core/blob/master/Formula/cbmc.rb
+# See https://github.com/diffblue/cbmc/issues/6343
+execute_process(COMMAND uname -p
```
NlightNFotis: I don't really think that's needed at all - I'm able to build on my ARM Mac without this flag on, cleanly.

I believe this may be an artefact on the home-brew formula since the early days of `M1`, where various environments (including `java`) had issues with their support.

My local machine (Apple Silicon ARM) is reporting the following if it helps:

```sh
$ fotiskoutoulakis@Fotiss-Mac-mini build_develop % ./bin/jbmc
JBMC version 5.38.0 (cbmc-5.38.0-7-gf0ab57e29a) 64-bit arm64 macos
Please give exactly one class name, and/or use -jar jarfile or --gb goto-binary
$ fotiskoutoulakis@Fotiss-Mac-mini build_develop % java --version
java 16.0.1 2021-04-20
Java(TM) SE Runtime Environment (build 16.0.1+9-24)
Java HotSpot(TM) 64-Bit Server VM (build 16.0.1+9-24, mixed mode, sharing)
```

chrisr-diffblue: This I'm less happy about. Fundamentally, it looks like the issue is nothing to do with macos and/or M1 architecture. Its _entirely_ down to Java version. It just happens that _at this moment in time_, brew will install a JVM 17 preview release on M1 macs (but still installs a JVM16 on x86 macs). I strongly suspect that as soon as more JVM17's are installed on other platforms (and OpenJDK 17 has just been officially released: https://mail.openjdk.java.net/pipermail/jdk-dev/2021-September/006037.html) this will bite other platforms.

chrisr-diffblue (changes requested):

NlightNFotis (commented):
Hi Mark, good stuff 👍🏻 

I think the second change is not really needed. If that's removed it's good to go from me 👍🏻

CMakeLists.txt
chrisr-diffblue: @NlightNFotis Might be worth checking if that Java 16 you have installed is a x86 or arm64 build... but either way, I think this reinforces that its not specifically a macOS M1 issue - its a Java 16/17 issue.

NlightNFotis: @chrisr-diffblue Yeah, I think you're right - on my system, it looks like java is a fat binary of two architectures:

```sh
$ fotiskoutoulakis@Fotiss-Mac-mini build_develop % cat `which java` | file -
/dev/stdin: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64e:Mach-O 64-bit executable arm64e]
```

markrtuttle:
Instead of trying to force -DWITH_JBMC=OFF on Apple M1, I now suggest (for the inexperienced user) to try building with -DWITH_JBMC=OFF if jbmc fails to build on MacOS.

Can we finish off the [Java 16/17](https://github.com/diffblue/cbmc/issues/6343) issue quickly so we can delete this completely?

CMakeLists.txt
```diff
@@ -231,6 +240,10 @@ cprover_default_properties(
     xml)
 
 option(WITH_JBMC "Build the JBMC Java front-end" ON)
+if(CMAKE_HOST_SYSTEM_NAME STREQUAL Darwin)
+    # See https://github.com/diffblue/cbmc/issues/6343
+    message(STATUS "Use -DWITH_JBMC=OFF if jbmc fails to build on MacOS")
```
chrisr-diffblue: I think (until we can fix the models library) this would be better as a check against the Java version, and then print a warning message and automatically set WITH_JBMC=OFF. Ultimately this is not about Darwin or any particular OS/architecture - its purely a Java version/API change.

markrtuttle: @chrisr-diffblue Done.

CMakeLists.txt
markrtuttle: @chrisr-diffblue Done.

chrisr-diffblue (approved):
Looks great, thanks!

NlightNFotis:
Please do not merge this yet!

We have fixed the issues in the Models Library, and we are now unblocked in terms of the Homebrew PR. I am looking into making the homebrew PR based on Java 17 (I can build with Java 17 locally no problem after the Models Library fix), and after this, then the only change required here is the default setup of the compiler to the `/usr/bin/clang`.

CMakeLists.txt
```diff
@@ -230,7 +240,59 @@ cprover_default_properties(
...
+        set(FOUND "UNKNOWN")
+    endif()
+    set(${VERSION} ${FOUND} PARENT_SCOPE)
+endfunction()
+
+MAVEN_JAVA_VERSION(JAVA_VERSION)
+message(STATUS "Found Maven uses Java version ${JAVA_VERSION} by default")
+
 option(WITH_JBMC "Build the JBMC Java front-end" ON)
+if(NOT JAVA_VERSION VERSION_LESS 17)
```
NlightNFotis, line 287: When https://github.com/Homebrew/homebrew-core/pull/85720/files gets merged to homebrew, `cbmc` (and `jbmc` by extension) will build cleanly with `openJDK` (version 17).

If you could drop the java version checks here (or at least, not deactivate `jbmc` building if `java --version` >= 17), then that would be great.

NlightNFotis (changes requested):

markrtuttle:
Closing in favor of https://github.com/diffblue/cbmc/pull/6359.  The JDK 17 issues dealt with here are being fixed in another pull request.

PR #6357, opened 2021-09-22
Implement sending commands to solver in incremental SMT2 backend

thomasspriggs:
This PR includes implementation of sending commands to solver in incremental SMT2 backend, based on the exprts recieved. This is sufficient to get a sat/unsat response from the solver in sufficiently simple cases. However parsing the result from the solver is not included in this PR and will be in a follow up PR.

- [ ] N/A - Non claimed. ~~My commit message includes data points confirming performance improvements (if claimed).~~

### DIFF
--- a/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
+++ b/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
@@ -2,11 +2,15 @@
 
 #include "smt2_incremental_decision_procedure.h"
 
+#include <solvers/smt2_incremental/smt_to_smt2_string.h>
 #include <util/expr.h>
+#include <util/string_utils.h>
 
 smt2_incremental_decision_proceduret::smt2_incremental_decision_proceduret(
-  std::string solver_command)
-  : solver_command{std::move(solver_command)}, number_of_solver_calls{0}
+  std::string _solver_command)
+  : solver_command{std::move(_solver_command)},
+    number_of_solver_calls{0},
+    solver_process{split_string(solver_command, ' ', false, true)}
 {
 }
 
@@ -70,4 +74,10 @@ decision_proceduret::resultt smt2_incremental_decision_proceduret::dec_solve()
 {
   ++number_of_solver_calls;
   UNIMPLEMENTED_FEATURE("solving.");
+}
+
+void smt2_incremental_decision_proceduret::send_to_solver(
+  const smt_commandt &command)
+{
+  solver_process.send(smt_to_smt2_string(command) + "\n");
 }
--- a/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
+++ b/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
@@ -196,7 +196,10 @@ void smt2_incremental_decision_proceduret::pop()
 decision_proceduret::resultt smt2_incremental_decision_proceduret::dec_solve()
 {
   ++number_of_solver_calls;
-  UNIMPLEMENTED_FEATURE("solving.");
+  send_to_solver(smt_check_sat_commandt{});
+  const auto result = solver_process.wait_receive();
+  log.debug() << "Solver response - " << result << messaget::eom;
+  UNIMPLEMENTED_FEATURE("parsing of solver response.");
 }
 
 void smt2_incremental_decision_proceduret::send_to_solver(
--- a/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
+++ b/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
@@ -67,7 +67,7 @@ void smt2_incremental_decision_proceduret::define_dependent_functions(
     {
       const irep_idt &identifier = symbol_expr->get_identifier();
       const symbolt *symbol = nullptr;
-      if(!ns.lookup(identifier, symbol))
+      if(!ns.lookup(identifier, symbol) && !symbol->value.is_nil())
       {
         if(dependencies_needed(symbol->value))
           continue;
--- a/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
+++ b/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
@@ -2,6 +2,7 @@
 
 #include "smt2_incremental_decision_procedure.h"
 
+#include <solvers/smt2_incremental/smt_commands.h>
 #include <solvers/smt2_incremental/smt_to_smt2_string.h>
 #include <util/expr.h>
 #include <util/string_utils.h>
@@ -12,6 +13,9 @@ smt2_incremental_decision_proceduret::smt2_incremental_decision_proceduret(
     number_of_solver_calls{0},
     solver_process{split_string(solver_command, ' ', false, true)}
 {
+  send_to_solver(smt_set_option_commandt{smt_option_produce_modelst{true}});
+  send_to_solver(smt_set_logic_commandt{
+    smt_logic_quantifier_free_uninterpreted_functions_bit_vectorst{}});
 }
 
 exprt smt2_incremental_decision_proceduret::handle(const exprt &expr)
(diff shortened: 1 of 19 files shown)


src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -12,6 +13,9 @@ smt2_incremental_decision_proceduret::smt2_incremental_decision_proceduret(
     number_of_solver_calls{0},
     solver_process{split_string(solver_command, ' ', false, true)}
 {
+  send_to_solver(smt_set_option_commandt{smt_option_produce_modelst{true}});
+  send_to_solver(smt_set_logic_commandt{
+    smt_logic_quantifier_free_uninterpreted_functions_bit_vectorst{}});
```
TGWDB: Do we want to send these at solver construction time, or do we want to send/update these per query? I don't have a good answer now, but there was comment on another issue/PR that the logic could vary for different calls/inputs. Also do we want to be able to change these later (and is this allowed/easy to do)?

TGWDB (approved):

NlightNFotis (approved):
LGTM

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -8,10 +8,12 @@
...
 
 smt2_incremental_decision_proceduret::smt2_incremental_decision_proceduret(
-  std::string _solver_command)
-  : solver_command{std::move(_solver_command)},
+  std::string _solver_command,
+  message_handlert &message_handler)
+  : solver_command(std::move(_solver_command)),
     number_of_solver_calls{0},
-    solver_process{split_string(solver_command, ' ', false, true)}
+    solver_process{split_string(solver_command, ' ', false, true)},
```
peterschrammel: The split_string looks brittle.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
thomasspriggs: According to section 4.1 of the SMTLIB standard version 2.6, the `set-logic` command can only be issued whilst the solver is in "Start mode" and the only way to return to this mode is to issue the `reset` command. The actions of the `reset` command include undefining all functions and assertions we have sent to the solver. Therefore if we were reseting and setting different logics for different rounds of solving, then we would no longer be carrying out incremental solving.

TLDR; We should be sending these commands at construction time.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -2,11 +2,93 @@
...
+    {
+      dependent_expressions.insert(expr_node);
+    }
+  });
+  return dependent_expressions;
+}
+
+/// \brief Defines any functions which \p expr depends on, which have not yet
+///   been defined, along with their dependencies in turn.
+void smt2_incremental_decision_proceduret::define_dependent_functions(
```
peterschrammel, line 43: Would be great to unit-test this somehow.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
thomasspriggs: The arguments are provided to `cbmc` in space separated form. Therefore they currently need to be split in order separate them. Any suggestions for making this less brittle?

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
thomasspriggs: I have added unit tests of `smt2_incremental_decision_proceduret` these tests indirectly test the functionality of this `define_dependent_functions` function, via the `handle` and `set_to` functions of the public interface.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -67,21 +67,21 @@ void smt2_incremental_decision_proceduret::define_dependent_functions(
     {
       const irep_idt &identifier = symbol_expr->get_identifier();
       const symbolt *symbol = nullptr;
-      if(!ns.lookup(identifier, symbol) && !symbol->value.is_nil())
+      if(ns.lookup(identifier, symbol) || symbol->value.is_nil())
```
NlightNFotis, line 73: Are we sure this change is valid? It doesn't look like that to me.

[De Morgan's law in this particular instance would be a bi-implication](https://en.wikipedia.org/wiki/De_Morgan%27s_laws):

```
¬(A ∨ B) ⇔ ¬A ∧ ¬B
```

Assuming in your case `A : ns.lookup()` and `B : symbol->value.is_nil()` then I believe this should have been translated to `(!(ns.lookup() || symbol->value.is_nil()))`.

NlightNFotis: Yeah, I just verified this is with a quick run of the truth table:

`A : ns.lookup()`
`B : symbol->value.is_nil()`
`Original : !ns.lookup() && !symbol->value.is_nil()`
`Refactored : ns.lookup() || symbol->value.is_nil()`

| A  | B | Original | Refactored |
|---|---|--------|-------------|
| F  | F  |      T      |         F           |
| F  | T  |       F     |         T           |
| T  | F  |       F     |         T           |
| T  | T  |       F     |        T            |

NlightNFotis (changes requested):
The rest of the stuff looks good, but the refactoring of the condition expression changed its semantics.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
thomasspriggs: Yes. The outcome of the logic you present is the opposite way round. That final not was handled by flipping the contents of the code blocks rather than adding the `not` operation. Therefore the outcome is still the same, which is why the tests still pass.

NlightNFotis (approved):

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -2,16 +2,133 @@
...
+void smt2_incremental_decision_proceduret::define_dependent_functions(
+  const exprt &expr)
+{
+  std::unordered_set<exprt, irep_hash> seen_expressions =
+    make_range(expression_identifiers)
+      .map([](const std::pair<exprt, smt_identifier_termt> &expr_identifier) {
+        return expr_identifier.first;
+      });
+  std::stack<exprt> to_be_defined;
+  const auto dependencies_needed = [&](const exprt &expr) {
```
chrisr-diffblue, line 52: Any particular reason why this is a lambda, rather than a helper function, such as a hypothetical `push_dependencies_needed(exprt, std::stack<exprt>)` which might be more descriptive? Either way, I'd be tempted to change the name to make its behaviour more explicit... At the moment, it superficially looks like a call to `dependencies_needed(e)` would be state-free - but actually its changing the state of the method via the stack of exprts... I'd make that behaviour explicit in the name for ease of code reading...

src/solvers/smt2_incremental/smt_terms.cpp
```diff
@@ -50,7 +50,7 @@ bool smt_bool_literal_termt::value() const
 
 static bool is_valid_smt_identifier(irep_idt identifier)
 {
-  static const std::regex valid{R"(^[^\|\\]*$)"};
+  static const std::regex valid{R"(^[^\|]*$)"};
```
chrisr-diffblue, line 53: Minor quibble - but for the more hieroglyphic regex's like this its nice to include a comment showing an/some examples of what the regex is attempting to match.

chrisr-diffblue (approved):
Nothing blocking - though I've left a couple of small comments that might be worth considering.

peterschrammel (approved):

src/solvers/smt2_incremental/smt_terms.cpp
thomasspriggs: I agree this would be better with a comment. I didn't add one when I first wrote this, as it is relatively short and I am relatively comfortable with regex. I'll try and remember to comment this in a follow-up PR.

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
thomasspriggs: Because it was the quickest way for me to de-duplicate this code in the 2 places it is needed. A separate function would also have needed an additional parameter. I agree that the name I assigned to this lambda isn't necessarily the best. I think doing some pair programming together could be helpful in trying to come up with a more readable implementation.

PR #6363, opened 2021-09-23
Handle functions with mangled names in contract replacement

feliperodri:

In some cases, we use `source_location` to retrieve information (e.g., function name). However, if we are consuming a `goto_programt`, but also care about the name of that function, then it must be passed separately. We should not try to infer it via some means, such as the source location. In this case, `apply_function_contract` must also consume a `function_id`, so we can properly retrieve the symbol for the caller function.

### DIFF
--- a/regression/contracts/assigns_replace_09/test.desc
+++ b/regression/contracts/assigns_replace_09/test.desc
@@ -1,0 +1,12 @@
+CORE
+main.c
+--replace-call-with-contract bar
+^EXIT=0$
+^SIGNAL=0$
+\[\d+\] file main.c line \d+ Check that bar\'s assigns clause is a subset of foo\'s assigns clause: SUCCESS$
+^VERIFICATION SUCCESSFUL$
+--
+^Condition: \!not\_found$
+--
+Checks whether CBMC properly evaluates subset relationship on assigns
+during replacement with static functions.
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -473,18 +473,21 @@ code_contractst::create_ensures_instruction(
 }
 
 bool code_contractst::apply_function_contract(
-  goto_programt &goto_program,
-  goto_programt::targett target)
+  const irep_idt &function,
+  goto_programt &function_body,
+  goto_programt::targett &target)
 {
+  const auto &const_target =
+    static_cast<const goto_programt::targett &>(target);
   // Return if the function is not named in the call; currently we don't handle
   // function pointers.
-  PRECONDITION(as_const(*target).call_function().id() == ID_symbol);
+  PRECONDITION(const_target->call_function().id() == ID_symbol);
 
   // Retrieve the function type, which is needed to extract the contract
   // components.
-  const irep_idt &function =
-    to_symbol_expr(as_const(*target).call_function()).get_identifier();
-  const symbolt &function_symbol = ns.lookup(function);
+  const irep_idt &target_function =
+    to_symbol_expr(const_target->call_function()).get_identifier();
+  const symbolt &function_symbol = ns.lookup(target_function);
   const auto &type = to_code_with_contract_type(function_symbol.type);
 
   // Isolate each component of the contract.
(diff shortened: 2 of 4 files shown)


src/goto-instrument/contracts/contracts.cpp
```diff
@@ -473,18 +473,19 @@ code_contractst::create_ensures_instruction(
...
-  PRECONDITION(as_const(*target).call_function().id() == ID_symbol);
+  PRECONDITION((*target).call_function().id() == ID_symbol);
 
   // Retrieve the function type, which is needed to extract the contract
   // components.
-  const irep_idt &function =
-    to_symbol_expr(as_const(*target).call_function()).get_identifier();
-  const symbolt &function_symbol = ns.lookup(function);
+  const irep_idt &target_function =
+    to_symbol_expr((*target).call_function()).get_identifier();
```
SaswatPadhi:
```suggestion
    to_symbol_expr(target->call_function()).get_identifier();
```
(and similar changes below)

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -473,18 +473,19 @@ code_contractst::create_ensures_instruction(
...
-  goto_programt &goto_program,
-  goto_programt::targett target)
+  const irep_idt &function,
+  goto_programt &function_body,
+  goto_programt::targett &target)
 {
   // Return if the function is not named in the call; currently we don't handle
   // function pointers.
-  PRECONDITION(as_const(*target).call_function().id() == ID_symbol);
+  PRECONDITION((*target).call_function().id() == ID_symbol);
```
SaswatPadhi: Why are we removing `as_const`?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -473,18 +473,19 @@ code_contractst::create_ensures_instruction(
 }
 
 bool code_contractst::apply_function_contract(
-  goto_programt &goto_program,
-  goto_programt::targett target)
+  const irep_idt &function,
+  goto_programt &function_body,
+  goto_programt::targett &target)
```
SaswatPadhi, line 478: Could this be made `const`:

```suggestion
  const goto_programt::targett &target)
```

I can see that we might be modifying the `function_body`, but is the `target` modified too?

regression/contracts/assigns_replace_09/test.desc
```diff
@@ -0,0 +1,12 @@
...
+main.c
+--replace-call-with-contract bar
+^EXIT=0$
+^SIGNAL=0$
+\[\d+\] file main.c line \d+ Check that bar\'s assigns clause is a subset of foo\'s assigns clause: SUCCESS$
+^VERIFICATION SUCCESSFUL$
+--
+^Condition: \!not\_found$
+--
+Checks whether CBMC properly evaluates subset relationship on assigns
```
SaswatPadhi, line 11: We already have a regression test for subset checking. May be clarify here that this is specifically for `static` functions whose names are mangled?

SaswatPadhi (changes requested):

src/goto-instrument/contracts/contracts.cpp
feliperodri: Implemented as suggested.

src/goto-instrument/contracts/contracts.cpp
feliperodri: It's unnecessary.

src/goto-instrument/contracts/contracts.cpp
feliperodri: Target is used as an iterator by sub-calls and might be modified using operators `++` or `--`.

regression/contracts/assigns_replace_09/test.desc
feliperodri: Implemented as suggested.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Why?
It returns a `const` reference, so we wouldn't be able to unintentionally modify the underlying object.

For example, in this case, we were grabbing a `const` ref to `*target` and then calling `call_function()` on it, which previously wouldn't have been able to change `*target`, but now it could.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Thanks, resolving this then.

src/goto-instrument/contracts/contracts.cpp
feliperodri: @SaswatPadhi but now the code is
```c
PRECONDITION(target->call_function().id() == ID_symbol);
```

SaswatPadhi: Yes, but the `target->call_function()` is no longer guaranteed not to change `*target`.

SaswatPadhi (approved):
LGTM. Thanks! :D

PR #2000, opened 2018-04-03
[SV-COMP'18 11/19] invalid_object(pointer) is true for all non-existent objects

tautschnig:
Check whether the object part of a pointer is greater or equal the total
number of objects, or equals a dynamic object that has not actually been
allocated on a given trace.

PR #6366, opened 2021-09-25
is_invalid_pointer is now defined as a bi-implication

kroening:
The previous encoding of is_invalid_pointer is an implication,
which means that constraints that a pointer is not an invalid
pointer have no meaning.

This commit changes the encoding to be a bi-implication, using
the numerical range of the valid pointers.

### DIFF
--- a/regression/cbmc-primitives/r_w_ok_bug/test.desc
+++ b/regression/cbmc-primitives/r_w_ok_bug/test.desc
@@ -1,10 +1,7 @@
 CORE
 main.c
 --pointer-check --no-simplify --no-propagation
-^\[main.pointer_dereference.\d+\] line 8 dereference failure: pointer outside object bounds in \*p1: FAILURE$
-^\*\* 1 of \d+ failed
-^VERIFICATION FAILED$
-^EXIT=10$
+^VERIFICATION SUCCESSFUL$
+^EXIT=0$
 ^SIGNAL=0$
 --
-^warning: ignoring
--- a/regression/cbmc-primitives/r_w_ok_inconsistent_integer/test-no-cp.desc
+++ b/regression/cbmc-primitives/r_w_ok_inconsistent_integer/test-no-cp.desc
@@ -1,11 +1,10 @@
-FUTURE
+CORE
 main.c
 --pointer-check --no-simplify --no-propagation
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
-^warning: ignoring
 --
 Tests that an inconsistent assumption involving a pointer initialized to an
 integer address not pointing to memory and __CPROVER_r_ok() can be detected via
--- a/regression/cbmc-primitives/r_w_ok_inconsistent_integer/test.desc
+++ b/regression/cbmc-primitives/r_w_ok_inconsistent_integer/test.desc
@@ -1,11 +1,10 @@
-FUTURE
+CORE
 main.c
 --pointer-check
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
-^warning: ignoring
 --
 Tests that an inconsistent assumption involving a pointer initialized to an
 integer address not pointing to memory and __CPROVER_r_ok() can be detected via
--- a/regression/cbmc-primitives/r_w_ok_inconsistent_nondet/test-no-cp.desc
+++ b/regression/cbmc-primitives/r_w_ok_inconsistent_nondet/test-no-cp.desc
@@ -1,11 +1,10 @@
-FUTURE
+CORE
 main.c
 --pointer-check --no-simplify --no-propagation
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
-^warning: ignoring
 --
 Tests that an inconsistent assumption involving a nondet pointer and
 __CPROVER_r_ok() can be detected via assert(0). We use --no-simplify and
--- a/regression/cbmc-primitives/r_w_ok_inconsistent_nondet/test.desc
+++ b/regression/cbmc-primitives/r_w_ok_inconsistent_nondet/test.desc
@@ -1,11 +1,10 @@
-FUTURE
+CORE
 main.c
 --pointer-check
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
-^warning: ignoring
 --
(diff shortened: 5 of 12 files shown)


feliperodri:
@SaswatPadhi @jimgrundy @remi-delmas-3000

TGWDB (approved):

regression/cbmc-primitives/r_w_ok_inconsistent_integer/test.desc
```diff
@@ -1,11 +1,10 @@
-FUTURE
+CORE
 main.c
 --pointer-check
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
-^warning: ignoring
```
tautschnig, line 8: Why is this pattern being removed?

regression/cbmc-primitives/r_w_ok_inconsistent_nondet/test-no-cp.desc
```diff
@@ -1,11 +1,10 @@
-FUTURE
+CORE
 main.c
 --pointer-check --no-simplify --no-propagation
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
-^warning: ignoring
```
tautschnig, line 8: Why is this pattern being removed?

regression/cbmc-primitives/r_w_ok_inconsistent_nondet/test.desc
```diff
@@ -1,11 +1,10 @@
-FUTURE
+CORE
 main.c
 --pointer-check
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
-^warning: ignoring
```
tautschnig, line 8: Why is this pattern being removed?

regression/cbmc-primitives/r_w_ok_inconsistent_integer/test-no-cp.desc
```diff
@@ -1,11 +1,10 @@
-FUTURE
+CORE
 main.c
 --pointer-check --no-simplify --no-propagation
 ^EXIT=0$
 ^SIGNAL=0$
 ^VERIFICATION SUCCESSFUL$
 --
-^warning: ignoring
```
tautschnig, line 8: Why is this pattern being removed?

regression/cbmc-primitives/r_w_ok_bug/test.desc
```diff
@@ -1,10 +1,7 @@
...
 --pointer-check --no-simplify --no-propagation
-^\[main.pointer_dereference.\d+\] line 8 dereference failure: pointer outside object bounds in \*p1: FAILURE$
-^\*\* 1 of \d+ failed
-^VERIFICATION FAILED$
-^EXIT=10$
+^VERIFICATION SUCCESSFUL$
+^EXIT=0$
 ^SIGNAL=0$
 --
-^warning: ignoring
```
tautschnig, line 10: Why is this pattern being removed?

tautschnig (dismissed):
Is there any data on how much slower things get in practice?

tautschnig:
Returning to this after some time. A bunch of thoughts:
1) #2000 previously proposed a similar solution, but was rejected in out-of-band communication with @kroening for it would be incompatible with incrementally adding objects. Doesn't this problem exist here as well?
2) An aspect that #2000 tried to tackle in addition to what is covered here are objects that symex created, yet the path condition of which isn't true. #2000 specifically did this for dynamic objects, but the same would actually apply to stack-allocated data. I would therefore like to ask: should "is invalid object" also be true for objects the life span of which hasn't yet begun?
3) As a possible alternative to address concern 1) above: make "is invalid object" evaluate to false for any known object, but don't specify a value for all other cases. This permits adding constraints in subsequent solver calls.

jimgrundy:
@tautschnig suggests above 

> As a possible alternative to address concern 1) above: make "is invalid object" evaluate to false for any known object, but don't specify a value for all other cases. This permits adding constraints in subsequent solver calls.

Does that take us back to being an implication rather than an equivalence, and were we trying to get away from that?

Also, if this is an equivalence, can we please invert the sense and change the name to "is_valid_pointer". It's confusing to have to say that a pointer is valid by saying that it is not invalid.

tautschnig:
> @tautschnig suggests above
> 
> > As a possible alternative to address concern 1) above: make "is invalid object" evaluate to false for any known object, but don't specify a value for all other cases. This permits adding constraints in subsequent solver calls.
> 
> Does that take us back to being an implication rather than an equivalence, and were we trying to get away from that?
> 

It should be noted that the existing implementation (prior to the changes proposed in this PR) was a bi-implication on the object that is singled out as the "invalid" object. Therefore, "is invalid object" (at present) has to be read as "is _the_ invalid object," and not "is _not a valid_ object." It may well be that we should ...

> Also, if this is an equivalence, can we please invert the sense and change the name to "is_valid_pointer". It's confusing to have to say that a pointer is valid by saying that it is not invalid.

... keep the current `is_invalid_object` as-is and instead make the implementation proposed in this PR back a new primitive "is_valid_object" (we shouldn't use "is_valid_pointer," because then we'd also have to check pointer offsets).

jimgrundy:
> ... keep the current is_invalid_object as-is and instead make the implementation proposed in this PR back a new primitive "is_valid_object" (we shouldn't use "is_valid_pointer," because then we'd also have to check pointer offsets).

Agree about calling the proposed thing "is_valid_object" (not "is_valid_pointer") for sure.

What would keeping the current "is_invalid_object" do? Would it still track a single distinguished invalid object? Or would it be the negation of "is_valid_object". Is the motivation to keep it to avoid breaking existing code or is there another reason to have "is_invalid_object" kept around?

kroening:
I think we need a proper 'object factory' facility!

tautschnig:
Some notes as @kroening and myself discussed this out of band: the proposed approach may have a substantial performance penalty. This may be fixed by making improvements to post-processing, but even then `is_invalid_pointer` will not necessarily be what a user even wants: 1) use within assumptions requires a proper object factory as noted above, assumptions over `!is_invalid_pointer` are not the right approach; 2) `is_invalid_pointer` does not match the informal understanding of an "invalid" pointer, but really just says that the pointed-to object is CBMC's internal "invalid object."

jimgrundy:
This one doesn't seem like the way to go. Should we close it to declutter?

feliperodri:
> This one doesn't seem like the way to go. Should we close it to declutter?

@kroening ?

PR #6371, opened 2021-09-30
Propagate write-set checks to sub-functions

feliperodri:


regression/contracts/assigns_replace_08/test.desc
```diff
@@ -3,8 +3,8 @@ main.c
 --enforce-contract foo --replace-call-with-contract bar _ --pointer-primitive-check
 ^EXIT=10$
 ^SIGNAL=0$
-\[\d+\] file main.c line \d+ Check that bar\'s assigns clause is a subset of foo\'s assigns clause: FAILURE$
-\[\d+\] file main.c line \d+ Check that \*z is assignable: FAILURE$
+\[foo.\d+\] line \d+ Check that bar\'s assigns clause is a subset of foo\'s assigns clause: FAILURE$
+\[foo.\d+\] line \d+ Check that \*z is assignable: FAILURE$
```
SaswatPadhi, line 6: Do we not emit file names in source location any more?
What happens when we have multiple C files and we generate a single goto binary out of them?

regression/contracts/function-calls-03-1/test-enf.desc
```diff
@@ -1,16 +1,17 @@
...
+^file main.c line \d+ function f2: recursion is ignored on call to \'f2\'$
+^\[postcondition.\d+\] file main.c line \d+ Check ensures clause: FAILURE$
 ^VERIFICATION FAILED$
 --
 --
 Verification:
   function | pre-cond | post-cond
   ---------|----------|----------
   f1       | assumed  | asserted
-  f2       | assumed  | asserted
```
SaswatPadhi, line 13: Should we enforce `f2` and verify its contract, since `f2` also has some contract and it was being verified earlier?

regression/contracts/function-calls-04-1/test-enf.desc
```diff
@@ -1,17 +1,17 @@
...
+^\[postcondition.\d+\] file main.c line \d+ Check ensures clause: FAILURE$
 ^VERIFICATION FAILED$
 --
 --
 Verification:
   function | pre-cond | post-cond
   ---------|----------|----------
   f1       | assumed  | asserted
-  f2_out   | assumed  | asserted
-  f2_in    | assumed  | asserted
```
SaswatPadhi, line 14: Should we enforce `f2_in` and `f2_out` and verify their contracts, since they also has some contracts and those were being verified earlier?

src/ansi-c/ansi_c_convert_type.cpp
```diff
@@ -253,26 +254,22 @@ void ansi_c_convert_typet::read_rec(const typet &type)
     const exprt &as_expr =
       static_cast<const exprt &>(static_cast<const irept &>(type));
 
-    for(const exprt &operand : to_unary_expr(as_expr).op().operands())
+    for(const exprt &target : to_unary_expr(as_expr).op().operands())
     {
-      if(
-        operand.id() != ID_symbol && operand.id() != ID_ptrmember &&
-        operand.id() != ID_member && operand.id() != ID_dereference)
+      if(!is_lvalue(target))
```
SaswatPadhi: Nice! :tada:

src/ansi-c/ansi_c_convert_type.cpp
```diff
@@ -253,26 +254,22 @@ void ansi_c_convert_typet::read_rec(const typet &type)
...
       {
-        error().source_location = source_location;
+        error().source_location = target.source_location();
         error() << "illegal target in assigns clause" << eom;
         throw 0;
       }
+      else if(has_subexpr(target, ID_side_effect))
+      {
+        error().source_location = target.source_location();
+        error() << "Assigns clause is not side-effect free." << eom;
```
SaswatPadhi, line 262: We should add a regression test to make sure that this works as expected.

src/ansi-c/c_typecheck_base.cpp
```diff
@@ -741,10 +741,10 @@ void c_typecheck_baset::typecheck_declaration(
           }
         }
 
-        if(as_const(code_type).assigns().is_not_nil())
+        if(!as_const(code_type).assigns().empty())
```
SaswatPadhi: I don't think we need this condition any more. If it's empty, the next loop would just iterate over an empty range.

src/ansi-c/expr2c.cpp
```diff
@@ -583,10 +583,9 @@ std::string expr2ct::convert_rec(
       dest += " [[requires " + convert(requires) + "]]";
     }
 
-    if(!to_code_with_contract_type(src).assigns().operands().empty())
+    for(auto &assigns : to_code_with_contract_type(src).assigns())
```
SaswatPadhi:
```suggestion
    for(const auto &assigns : to_code_with_contract_type(src).assigns())
```

src/util/expr_util.cpp
```diff
@@ -24,6 +24,8 @@ bool is_lvalue(const exprt &expr)
     return is_lvalue(to_index_expr(expr).array());
   else if(expr.id() == ID_member)
     return is_lvalue(to_member_expr(expr).compound());
+  else if(expr.id() == ID_ptrmember)
+    return true;
```
SaswatPadhi: @tautschnig @kroening We are calling this function to check that the targets specified in assigns clause for functions are valid l-values. The `ptrmember` targets there aren't decomposed to deref + member before type checking. Should we be calling some utility function for the decomposition, or should we add this case here?

If we add this case, it should do something similar to the `member` case and not return `true` probably, because `a->b` == `(*a).b`.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -564,24 +555,31 @@ bool code_contractst::apply_function_contract(
...
   {
-    assigns_clauset assigns_clause(assigns, log, ns);
+    assigns_clauset assigns_clause(targets.operands(), log, ns);
 
     // Retrieve assigns clause of the caller function if exists.
     auto caller_assigns =
       to_code_with_contract_type(ns.lookup(function).type).assigns();
 
-    if(caller_assigns.is_not_nil())
+    if(enforced_functions.find(function.c_str()) != enforced_functions.end())
```
SaswatPadhi: This seems to check whether the contracts for caller (parent in the call chain) are being being checked.

However, as we discussed (documented in the specs doc), we need to check subset when contracts are _any_ ancestor is being checked.

SaswatPadhi (changes requested):
Thanks! That's quite a lot of cleanup and fixes.

I have a few minor comments and only major comments about subset check part.

src/ansi-c/ansi_c_convert_type.cpp
jimgrundy: Agree. Nice.

src/ansi-c/ansi_c_convert_type.cpp
jimgrundy: Definitely add a test.

src/ansi-c/c_typecheck_base.cpp
jimgrundy: Same would be true for the requires clause if/loop up on line 727.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -132,7 +132,8 @@ void assigns_clauset::remove_from_local_write_set(const exprt &expr)
   local_write_set.erase(targett(*this, expr));
 }
 
-goto_programt assigns_clauset::generate_havoc_code() const
+goto_programt
+assigns_clauset::generate_havoc_code(const source_locationt &location) const
 {
   modifiest modifies;
   for(const auto &target : global_write_set)
```
jimgrundy: I thought we didn't need separate local and global write sets now. Is that a future PR. Also, I see the addresses in the code, don't we need more than just the address to know what to havoc (location vs object containing location)?

src/goto-instrument/contracts/contracts.cpp
jimgrundy: +1

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -686,106 +685,61 @@ void code_contractst::instrument_assign_statement(
...
     "The first argument of instrument_call_statement should always be "
     "a function call");
 
-  irep_idt called_name;
-  if(instruction_iterator->call_function().id() == ID_dereference)
+  irep_idt called_function;
+  if(instruction_it->call_function().id() == ID_dereference)
   {
-    called_name =
+    called_function =
```
jimgrundy: Surely we want a test for this.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -686,106 +685,61 @@ void code_contractst::instrument_assign_statement(
...
+    assigns.remove_from_local_write_set(lhs_dereference);
+    assigns.remove_from_global_write_set(lhs_dereference);
     return;
   }
 
   if(
-    instruction_iterator->call_lhs().is_not_nil() &&
-    instruction_iterator->call_lhs().id() == ID_symbol)
+    instruction_it->call_lhs().is_not_nil() &&
+    instruction_it->call_lhs().source_location().is_not_nil() &&
```
jimgrundy: You seem to be going away from "foo.is_not_nill()" and towards "!foo.is_empty()" (a move I strongly endorse) so why not here?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -686,106 +685,61 @@ void code_contractst::instrument_assign_statement(
...
+    is_lvalue(instruction_it->call_lhs()))
```
jimgrundy: Test should be added to cover this.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -686,106 +685,61 @@ void code_contractst::instrument_assign_statement(
...
-    alias_assertion.add(goto_programt::make_assertion(
-      subset_check, instruction_iterator->source_location));
-    alias_assertion.instructions.back().source_location.set_comment(
-      "Check that " + id2string(called_name) +
-      "'s assigns clause is a subset of " +
-      id2string(instruction_iterator->source_location.get_function()) +
-      "'s assigns clause");
-    insert_before_swap_and_advance(
-      program, instruction_iterator, alias_assertion);
+      body, assigns, instruction_it, instruction_it->call_lhs());
```
jimgrundy: We need a test to cover.

src/ansi-c/ansi_c_convert_type.cpp
feliperodri: This is actually not necessary, because we check for `is_lvalue`prior, which would filter anything with side effects. Take a look at this example.
```c
#include <assert.h>
#include <stdlib.h>

int x = 0;

void bar(int *y)
{
  y = x;
}

void foo(int *y, int z) __CPROVER_assigns((x == 0) ? bar(y) : x)
{
  *y = z;
  z = -2;
}

int main()
{
  int *y = malloc(sizeof(*y));
  *y = 0;
  int z = 1;
  foo(y, z);
  assert(z == 1);
  return 0;
}
```
Verification  yields to this result.
```
main.c:11:1: error: illegal target in assigns clause
 void foo(int *y, int z) __CPROVER_assigns((x == 0) ? bar(y) : x)
CONVERSION ERROR
```

Are there any cases we might miss if we just remove this? @SaswatPadhi @jimgrundy

src/ansi-c/c_typecheck_base.cpp
feliperodri: Good catch! Implemented as suggested.

src/ansi-c/expr2c.cpp
feliperodri: Implemented as suggested. I also included for `requires` and `ensures`.

src/util/expr_util.cpp
feliperodri: @peterschrammel @martin-cs @chrisr-diffblue any thoughts on this too?

src/ansi-c/ansi_c_convert_type.cpp
jimgrundy: have we tried "__CPROVER_assigns(++x)" or similar?

jimgrundy: Also, coverage report says this isn't covered. If it is hittable then we should have a test, just need to think of one. If it isn't hittable then this code is dead and should be removed.

src/goto-instrument/contracts/assigns.cpp
feliperodri: > I thought we didn't need separate local and global write sets now. Is that a future PR.

We now only track one `write_set` now. I was implementing this change on a separate PR, but I pulled the commit to this one.

> Also, I see the addresses in the code, don't we need more than just the address to know what to havoc (location vs object containing location)?

The correct answer here is that a proper implementation of CARs will give us all necessary info for enforcement and replacement. Now, we keep track of the address in a `exprt` which will also give us the source location and the object associated with this address (we use this info to havoc).

src/goto-instrument/contracts/contracts.cpp
feliperodri: We can't use `empty` here, because the return type of this methods doesn't provide us with that functionality.
`instruction_it->call_lhs()` -> `exprt &`
`instruction_it->call_lhs().source_location()` -> `const source_locationt &`
So, the correct way to check if they are "empty" is using `is_not_nil()`.

src/goto-instrument/contracts/contracts.cpp
feliperodri: Our test case `assigns_enforce_22` cover this case. This is to make sure we properly check cases where the return of a function is assigned to a variable. For example,
```c
int bar() { ... }
int x = bar();
```
Perhaps what is missing here is a comment. I'll add it.

src/goto-instrument/contracts/contracts.cpp
feliperodri: Same as https://github.com/diffblue/cbmc/pull/6371#discussion_r720413355

src/goto-instrument/contracts/contracts.cpp
feliperodri: This is to handle function pointers. We should perhaps abort and report error message in this case. Correct?

regression/contracts/function-calls-03-1/test-enf.desc
feliperodri: I added an extra test description to enforce `f2` and verify its contract.

src/goto-instrument/contracts/contracts.cpp
jimgrundy: If we have a test that covers this... why is it being reported as not covered?

regression/contracts/assigns_replace_08/test.desc
feliperodri: We still provide source location, but now it used to organize the checks. Take a look at this example:
```
** Results:
header.h function f1
[f1.1] line 9 Check that x2 is assignable: SUCCESS
[f1.2] line 9 Check that y2 is assignable: SUCCESS
[f1.3] line 9 Check that z2 is assignable: SUCCESS

header.h function f2
[f2.1] line 14 Check that x3 is assignable: SUCCESS
[f2.2] line 14 Check that y3 is assignable: SUCCESS
[f2.3] line 14 Check that z3 is assignable: SUCCESS

header.h function f3
[f3.1] line 19 Check that *x3 is assignable: SUCCESS
[f3.2] line 20 Check that *y3 is assignable: SUCCESS
[f3.3] line 21 Check that *z3 is assignable: SUCCESS

** 0 of 9 failed (1 iterations)
VERIFICATION SUCCESSFUL
```

src/goto-instrument/contracts/contracts.cpp
jimgrundy: This is showing up as uncovered, I don't think we should have uncovered code in a system that must be trusted. But, if this is for function pointers and we aren't yet correctly instrumenting function pointers then let's abort here with an error message.

regression/contracts/assigns_replace_08/test.desc
SaswatPadhi: Nice!

Thanks.

src/ansi-c/ansi_c_convert_type.cpp
SaswatPadhi: > This is actually not necessary, because we check for `is_lvalue` prior, which would filter anything with side effects. 

I don't think so. In the example you tried, `bar(y)` would return `void` and so may be that's why it's not treated as a valid l-value. But, in general, function calls are valid within l-values.

Additionally, I don't think expressions like `*(++x)` would be flagged by `is_lvalue`. We should add a regression test with such a target in the assigns clause and make sure that we hit this branch in code.

regression/contracts/function-calls-04-1/test-enf.desc
feliperodri: Implemented as suggested.

src/goto-instrument/contracts/contracts.cpp
feliperodri: Based on an offline discussions, we decided to remove all subset relationship checks. The order in which we instrument contract allow us to rely on contract replacement and inlining of sub-function calls to catch any possible illegal writes.

src/goto-instrument/contracts/contracts.cpp
feliperodri: We don't need this case anymore. Inlining takes care of that problem, since all function calls (including the ones with function pointers) are inlined during enforcement.

regression/contracts/assigns_enforce_elvis_5/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-contract foo
+^EXIT=1$
+^SIGNAL=0$
+^main.c:\d+:\d+: error: illegal target in assigns clause$
+^CONVERSION ERROR$
+--
+--
+Check that Elvis operator expressions '(cond ? *if_true : *if_false)' are rejected from assigns clauses.
```
SaswatPadhi: Can we add a comment here that this is consistent with major compilers? (Perhaps also check the C standard?)
Because it's not super intuitive why this should be rejected and not `*(cond ? if_true : if_false)` ...

regression/contracts/assigns_enforce_function_calls/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-contract foo
+^EXIT=1$
+^SIGNAL=0$
+^main.c:\d+:\d+: error: illegal target in assigns clause$
+^CONVERSION ERROR$
+--
+--
+Check that function call expressions are rejected in assigns clauses.
```
SaswatPadhi: Missing newline here and a bunch of other ".desc" files.

regression/contracts/assigns_enforce_literal/test.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+main.c
+--enforce-contract foo
+^EXIT=1$
+^SIGNAL=0$
+^main.c:\d+:\d+: error: illegal target in assigns clause$
+^CONVERSION ERROR$
+--
+--
+Check that literal constants are rejected from assigns clauses.
```
SaswatPadhi: Missing newline.

Check other ".desc" files too.

src/goto-programs/instrument_preconditions.h
```diff
@@ -12,9 +12,17 @@ Date:   September 2017
...
+#include <util/replace_symbol.h>
+
 class goto_functiont;
 class goto_modelt;
 
+replace_symbolt actuals_replace_map(
+  const exprt &,
+  const exprt &,
+  const exprt::operandst &,
+  const namespacet &);
```
SaswatPadhi: Where is this function called in this PR? I did a quick search but couldn't find ...

src/goto-programs/instrument_preconditions.cpp
```diff
@@ -11,8 +11,6 @@ Date:   September 2017
 
 #include "instrument_preconditions.h"
 
-#include <util/replace_symbol.h>
```
SaswatPadhi: It's okay to move this to the `.h`, but if we don't need the new function introduced in the `.h` (see other comment), I would undo all changes to this file and the `.h` just to minimize this PR.

src/goto-instrument/contracts/contracts.cpp
feliperodri: Can you point me to this coverage info?

regression/contracts/assigns_enforce_function_calls/test.desc
feliperodri: Fixed it.

regression/contracts/assigns_enforce_literal/test.desc
feliperodri: Fixed it.

src/goto-instrument/contracts/contracts.cpp
jimgrundy: Just take a look at this file in the side-by-side diff view of the PR and you will see coverage warnings

image

src/ansi-c/expr2c.cpp
```diff
@@ -583,10 +583,9 @@ std::string expr2ct::convert_rec(
...
     {
-      dest += " [[assigns " +
-              convert(to_code_with_contract_type(src).assigns()) + "]]";
+      dest += " [[assigns " + convert(assigns) + "]]";
```
jimgrundy: Test coverage gap reported here. Can we have a regression test for this:
image

src/goto-programs/instrument_preconditions.h
feliperodri: This was debugging leftovers. Sorry. Removed it.

src/goto-programs/instrument_preconditions.cpp
feliperodri: This was debugging leftovers. Sorry. Removed it.

src/goto-instrument/contracts/contracts.cpp
jimgrundy: If we don't need this case why does it still exist?

regression/contracts/assigns_enforce_elvis_5/test.desc
jimgrundy: I guess @feliperodri has. found the explanation, so he could add a comment now.

jimgrundy (approved):

SaswatPadhi (approved):
LGTM. Thanks!

SaswatPadhi:
One minor comment: We have many tiny commits in this PR and could merge some of these together (especially if they are related) for a cleaner history. For example 77b2d946a7e363d2fe0f65963648585f73fe80de and d5374d74541a5ffed208210b764cf8314df9946a could definitely be merged together.

src/util/expr_util.cpp
kroening: There is something wrong here. ID_ptrmember is an expression that is internal to the C front-end, and should not be visible outside of it. It's possible that `is_value` is internal to the C front-end as well, and should possibly move there.

SaswatPadhi: Thanks.
It looks like `is_lvalue` is supposed to be used only on goto expressions, not C expressions. So calling it from out instrumentation code is okay, but calling it from our compilation code is not. 
Is there a similar function to check if an expression is a valid C l-value?

jimgrundy: Over in another slack thread Daniel mentioned "Within the C frontend, you can tell whether or not an expression is an lvalue by querying expr.get_bool(ID_C_lvalue)"

SaswatPadhi: For `__CPROVER_object(arg)` calls in the assigns clause, we want to make a special case and want to type check that the `deref(arg)` is an lvalue.

feliperodri: After a few offline discussions, CBMC now evaluates for `lvalue` expressions in assigns clauses after type checking using `get_bool(ID_C_lvalue)`.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -679,114 +656,48 @@ void code_contractst::instrument_assign_statement(
...
-  if(called_name == "malloc")
+  if(called_function == "malloc")
   {
     // malloc statments return a void pointer, which is then cast and assigned
     // to a result variable. We iterate one line forward to grab the result of
     // the malloc once it is cast.
-    instruction_iterator++;
-    if(instruction_iterator->is_assign())
+    instruction_it++;
+    if(instruction_it->is_assign())
```
jimgrundy, line 685: @SaswatPadhi @remi-delmas-3000 This code pattern looks like it is trying to recognize the assignment in code like "x = malloc(...);", but I think we just figured out this isn't the right way to do this. Can you check?

src/ansi-c/ansi_c_convert_type.cpp
```diff
@@ -13,6 +13,7 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <util/c_types.h>
 #include <util/config.h>
+#include <util/expr_util.h>
```
kroening, line 16: Is that include still needed?

kroening (approved):

src/ansi-c/ansi_c_convert_type.cpp
feliperodri: Yep, we use `has_subexpr` in this module now.

src/goto-instrument/contracts/contracts.h
```diff
@@ -39,19 +40,10 @@ Date: February 2016
...
-#define FLAG_REPLACE_ALL_CALLS "replace-all-calls-with-contracts"
-#define HELP_REPLACE_ALL_CALLS                                                 \
-  " --replace-all-calls-with-contracts\n"                                      \
-  "                              as above for all functions with a contract\n"
-
 #define FLAG_ENFORCE_CONTRACT "enforce-contract"
 #define HELP_ENFORCE_CONTRACT                                                  \
   " --enforce-contract <fun>     wrap fun with an assertion of its contract\n"
 
-#define FLAG_ENFORCE_ALL_CONTRACTS "enforce-all-contracts"
```
martin-cs, line 51: I am a little concerned by the change in functionality.  I realise that it would be good to have the extra flexibility but having to give a command-line argument for every function is quite a lot.  Elsewhere I have seen regexes used for this kind of thing.

martin-cs (approved):
I am approving this because 1. I do not see any blocking problems here, 2. as far as I know you are the main users of this code so changes in the functionality & interface are not likely to affect others too much and 3. I know it is urgent for you.  But, I found this quite hard to review.  In the future, if you have the time please could you consider:

1. Smaller PRs -- this does a number of related, but different things and would be easier to review in chunks.
2. A description of what you are doing and why in each commit message.  They are one line at the moment which is better than nothing but it gives more limited context.
3. More of a over-view description in the PR.

feliperodri:
> I am approving this because 1. I do not see any blocking problems here, 2. as far as I know you are the main users of this code so changes in the functionality & interface are not likely to affect others too much and 3. I know it is urgent for you. But, I found this quite hard to review. In the future, if you have the time please could you consider:
> 
>     1. Smaller PRs -- this does a number of related, but different things and would be easier to review in chunks.
> 
>     2. A description of what you are doing and why in each commit message.  They are one line at the moment which is better than nothing but it gives more limited context.
> 
>     3. More of a over-view description in the PR.

Thanks for the review @martin-cs. I'll at least add a better description for this PR now (3) and keep in mind (1) and (2) for future PRs.

src/goto-instrument/contracts/contracts.h
feliperodri: We we're using contracts now in a context where an user enforces one contract at a time, makes sure it is correct and then uses it during replacement. In the future, we still want to have an option where we can simply apply contracts whenever they are available, but for now we want to force users to specify which contracts must be enforced or replaced.

PR #6376, opened 2021-09-30
Use remove-function-pointers code for restricted function pointers

### LINKED ISSUE #6368: Input validation problems with `--function-pointer-restrictions-file`
CBMC version: 5.30.1 (cbmc-5.30.1)
Operating system: Ubuntu 20.04.2

```goto-instrument --function-pointer-restrictions-file restrictions``` (within the context of the [RMC](https://github.com/model-checking/rmc) tool). 

CBMC to restrict function pointer branch analysis to the cases specified, as described in the documentation for function pointer descriptions [here](http://cprover.diffblue.com/md__home_travis_build_diffblue_cbmc_doc_architectural_restrict-function-pointer.html). 

What happened instead:

I am running into two related problems with the `--function-pointer-restrictions-file` flag. 

**(1) Casting**

The functions I would like to restrict are a different type than the call site due to a cast in the source code to a function that takes a void pointer. The source code looks roughly like this:
```
// function pointer creation
.field=(int (*)(void *))_Animal_noise // struct field 

// function pointer use
var_5.vtable->field(var_1);
```

However, if I try to pass `_Animal_noise` as a restriction, I get this error:
```
Reason: type mismatch: `main.function_pointer_call.1' points to `int (void *)', but restriction `_Animal_noise' has type `int (struct _10007873221718841048 *self)'
```

However, if I try and include the cast in the restrictions file, it fails to parse:
```
Invalid restriction
Reason: symbol not found: (int (*)(void *))_Animal_noise
```

The pointer itself does not have a symbol name to my understanding, because it is a struct field.

~**(2) . in function name**~ wrong error, see below

~In a different example, some of my restricted function have `..` in the name, and fail to parse:~

~Example:~
```
  "_ZN80_$LT$devices..virtio..mmio..MmioTransport$u20$as$u20$devices..bus..BusDevice$GT$5write17h533d97e6e2ca6db0E.function_pointer_call.1": [
    "_ZN7devices6virtio6device12VirtioDevice20ack_features_by_page17hbff58082681c1a52E_wrapper",
    "_ZN7devices6virtio6device12VirtioDevice20ack_features_by_page17h2c20fca1a3ee805eE_wrapper",
    "_ZN7devices6virtio6device12VirtioDevice20ack_features_by_page17h1668e8cac23d16f1E_wrapper"
  ],
```

~Error:~
```
Reading GOTO program from 'parse_harness1/a.out'
file restrictions: syntax error
failed to read function pointer restrictions from restrictions
```

~This is despite CBMC analyzing the file with these names fine without restrictions.~

### PR DESCRIPTION
**Pull request by tautschnig**
We had two instances of code generating an if-else sequence of function
calls for function pointers. The code used with
--restrict-function-pointer, however, was not able to cope with type
mismatches. Make the code from remove_function_pointers re-usable in
both contexts, enabling support for additional type casts.

Fixes: #6368

### DIFF
--- a/src/goto-programs/remove_function_pointers.cpp
+++ b/src/goto-programs/remove_function_pointers.cpp
@@ -20,6 +20,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/source_location.h>
 #include <util/std_code.h>
 #include <util/std_expr.h>
+#include <util/string_utils.h>
 
 #include <analyses/does_remove_const.h>
 
@@ -44,24 +45,8 @@ public:
     goto_programt &goto_program,
     const irep_idt &function_id);
 
-  // a set of function symbols
-  using functionst = remove_const_function_pointerst::functionst;
-
-  /// Replace a call to a dynamic function at location
-  /// target in the given goto-program by a case-split
-  /// over a given set of functions
-  /// \param goto_program: The goto program that contains target
-  /// \param function_id: Name of function containing the target
-  /// \param target: location with function call with function pointer
-  /// \param functions: The set of functions to consider
-  void remove_function_pointer(
-    goto_programt &goto_program,
-    const irep_idt &function_id,
-    goto_programt::targett target,
-    const functionst &functions);
-
 protected:
-  messaget log;
+  message_handlert &message_handler;
   const namespacet ns;
   symbol_tablet &symbol_table;
   bool add_safety_assertion;
@@ -89,21 +74,6 @@ protected:
 
   typedef std::map<irep_idt, code_typet> type_mapt;
   type_mapt type_map;
-
-  bool is_type_compatible(
-    bool return_value_used,
-    const code_typet &call_type,
-    const code_typet &function_type);
-
-  bool arg_is_type_compatible(
-    const typet &call_type,
-    const typet &function_type);
-
-  void fix_argument_types(code_function_callt &function_call);
-  void fix_return_type(
-    const irep_idt &in_function_id,
-    code_function_callt &function_call,
-    goto_programt &dest);
 };
 
 remove_function_pointerst::remove_function_pointerst(
@@ -112,7 +82,7 @@ remove_function_pointerst::remove_function_pointerst(
   bool _add_safety_assertion,
   bool only_resolve_const_fps,
   const goto_functionst &goto_functions)
-  : log(_message_handler),
+  : message_handler(_message_handler),
     ns(_symbol_table),
     symbol_table(_symbol_table),
     add_safety_assertion(_add_safety_assertion),
@@ -131,9 +101,10 @@ remove_function_pointerst::remove_function_pointerst(
   }
 }
 
-bool remove_function_pointerst::arg_is_type_compatible(
+static bool arg_is_type_compatible(
   const typet &call_type,
-  const typet &function_type)
+  const typet &function_type,
+  const namespacet &ns)
 {
   if(call_type == function_type)
     return true;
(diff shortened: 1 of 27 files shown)

### CODE REVIEW

File: src/goto-programs/remove_function_pointers.cpp [hunk 1]
```diff
@@ -355,14 +330,56 @@ void remove_function_pointerst::remove_function_pointer(
...
+    add_safety_assertion);
 }
 
-void remove_function_pointerst::remove_function_pointer(
+static std::string function_pointer_assertion_comment(
+  const std::unordered_set<symbol_exprt, irep_hash> &candidates)
+{
+  std::stringstream comment;
+
+  comment << "dereferenced function pointer at must be ";
```
avanhatt: Needs location after `at`?

src/goto-programs/remove_function_pointers.cpp
```diff
@@ -355,14 +330,56 @@ void remove_function_pointerst::remove_function_pointer(
...
+
+  if(candidates.size() == 1)
+  {
+    comment << candidates.begin()->get_identifier();
+  }
+  else if(candidates.empty())
+  {
+    comment.str("no candidates for dereferenced function pointer");
```
avanhatt, line 357: Location here too would be great.

src/goto-programs/remove_function_pointers.cpp
tautschnig: Done, thanks! (Well, I actually removed the location.)

src/goto-programs/remove_function_pointers.cpp
tautschnig: My rationale for omitting this is that the location will be provided anyway, I don't think we need to repeat it.

src/goto-programs/remove_function_pointers.cpp
```diff
@@ -131,9 +101,10 @@ remove_function_pointerst::remove_function_pointerst(
   }
 }
 
-bool remove_function_pointerst::arg_is_type_compatible(
+static bool arg_is_type_compatible(
```
martin-cs, line 104: Is there a reason for preferring static functions over private member functions?

src/goto-programs/remove_function_pointers.h
```diff
@@ -46,4 +47,32 @@ bool remove_function_pointers(
...
+/// \param functions: The set of functions to consider
+/// \param add_safety_assertion: Iff true, include an assertion that the
+//         pointer matches one of the candidate functions
+void remove_function_pointer(
+  message_handlert &message_handler,
+  symbol_tablet &symbol_table,
+  goto_programt &goto_program,
+  const irep_idt &function_id,
+  goto_programt::targett target,
+  const std::unordered_set<symbol_exprt, irep_hash> &functions,
```
martin-cs, line 67: This is the key line right?  It gives a way of passing a map to function pointer removal and this splitting the process into "generate function maps" and "convert code using the maps"?

martin-cs (approved):
Yes.  There seem to be a number of different but overlapping changes in this PR.  There is a little formatting, threading the logging infrastructure through the functions, converting the generation code, catching up on the changed output, etc.  I realise that in this case it is not really easy to separate these but in future, if possible, separate commits would make this easier to review.

src/goto-programs/remove_function_pointers.cpp
tautschnig: As I made `remove_function_pointer` a global function, I had to have a way to use these functions.

tautschnig:
> Yes. There seem to be a number of different but overlapping changes in this PR. There is a little formatting, threading the logging infrastructure through the functions, converting the generation code, catching up on the changed output, etc. I realise that in this case it is not really easy to separate these but in future, if possible, separate commits would make this easier to review.

You're absolutely right that I should have tried to factor this into multiple commits. I had kept this in mind even when making the changes, but struggled to make it happen. And, upon a second look, I still struggle. So unless I'm really forced to, I'd rather keep it as is.

kroening (approved):

PR #6378, opened 2021-10-02
rename is_lvalue -> is_assignable

kroening:
The term 'lvalue' is nomenclature that is specific to C/C++.  The function
is_lvalue is meant to be used in the context of goto-programs, and thus, a
name that suggests a relationship to C programs may be confusing.

### DIFF
--- a/src/util/expr_util.cpp
+++ b/src/util/expr_util.cpp
@@ -18,12 +18,12 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "pointer_expr.h"
 #include "std_expr.h"
 
-bool is_lvalue(const exprt &expr)
+bool is_assignable(const exprt &expr)
 {
   if(expr.id() == ID_index)
-    return is_lvalue(to_index_expr(expr).array());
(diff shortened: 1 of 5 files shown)


chrisr-diffblue (approved):

TGWDB (approved):

src/util/expr_util.cpp
```diff
@@ -18,19 +18,20 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "pointer_expr.h"
 #include "std_expr.h"
 
-bool is_lvalue(const exprt &expr)
+bool is_assignable(const exprt &expr)
```
tautschnig, line 21: Should this perhaps be moved to goto-programs? The use in `replace_symbol` obviously conflicts with that, but some follow-up cleanup may be necessary.

tautschnig (approved):

src/util/expr_util.cpp
kroening: Yes; there may well be a need to make copies of this function for particular uses. I'd see this as an orthogonal question to the renaming done here.

PR #6382, opened 2021-10-05
change signature of interpretert::evaluate

kroening:
Instead of appending to a given vector, the method now returns a vector,
which is less prone to error.

### DIFF
--- a/src/goto-programs/interpreter_evaluate.cpp
+++ b/src/goto-programs/interpreter_evaluate.cpp
@@ -131,8 +131,7 @@ bool interpretert::count_type_leaves(const typet &ty, mp_integer &result)
   else if(ty.id()==ID_array)
   {
     const auto &at=to_array_type(ty);
-    mp_vectort array_size_vec;
-    evaluate(at.size(), array_size_vec);
+    mp_vectort array_size_vec = evaluate(at.size());
     if(array_size_vec.size()!=1)
       return true;
     mp_integer subtype_count;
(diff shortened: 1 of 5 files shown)


src/goto-programs/interpreter_evaluate.cpp
```diff
@@ -923,102 +865,101 @@ void interpretert::evaluate(
...
   {
     const auto &wexpr=to_with_expr(expr);
-    evaluate(wexpr.old(), dest);
-    std::vector<mp_integer> where;
-    std::vector<mp_integer> new_value;
-    evaluate(wexpr.where(), where);
-    evaluate(wexpr.new_value(), new_value);
+    mp_vectort dest = evaluate(wexpr.old());
+    std::vector<mp_integer> where = evaluate(wexpr.where());
+    std::vector<mp_integer> new_value = evaluate(wexpr.new_value());
```
tautschnig: For consistency: shouldn't these also use `mp_vectort` instead of `std::vector<mp_integer>`?

unit/interpreter/interpreter.cpp
```diff
@@ -32,9 +32,7 @@ class interpreter_testt
 
   mp_vectort evaluate(const exprt &expression)
   {
-    mp_vectort result;
-    interpreter.evaluate(expression, result);
-    return result;
+    return interpreter.evaluate(expression);
   }
```
tautschnig, line 36: Do we really need this function anymore?

tautschnig (approved):

unit/interpreter/interpreter.cpp
kroening: The alternative would require to make `interpreter` public, which is less nice.

src/goto-programs/interpreter_evaluate.cpp
kroening: done

feliperodri (approved):
Changes in the assigns clause module are fine. My only concern is the amount of Codecov alarms on the interpreter module. It seems we don't have any regression tests for this part of the codebase.

PR #6385, opened 2021-10-06
Fix invalidation of CARs on free and DEAD instructions

SaswatPadhi:
Previously, corresponding CARs were completely removed on `DEAD` instructions. While this is "okay" for unconditionally dead symbols, it resulted in spurious non-assignable errors when symbols were conditionally `DEAD`, e.g. `RETURN` within a conditional block.
Previously, invalid CARs from `free` calls were simply left behind, because (in theory) the CAR validation condition should automatically guard against conditions.

However, in both these cases, we need to check dead/deallocated status of individual objects to avoid spurious errors from CBMC's internal pointer checks. This PR fixes this.

### DIFF
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -637,7 +637,7 @@ void code_contractst::instrument_assign_statement(
     "The first instruction of instrument_assign_statement should always be"
     " an assignment");
 
-  add_containment_check(
+  add_inclusion_check(
     program, assigns_clause, instruction_it, instruction_it->assign_lhs());
 }
 
@@ -678,7 +678,7 @@ void code_contractst::instrument_call_statement(
   }
   else if(callee_name == "free")
   {
-    add_containment_check(
+    add_inclusion_check(
       body,
       assigns,
       instruction_it,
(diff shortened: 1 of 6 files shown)


remi-delmas-3000 (approved):
With this modification the the conditional  DEAD statement are correctly tracked using a condition embedded in the code rather than statically at compile time

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -838,20 +914,42 @@ void code_contractst::check_frame_conditions(
...
+          instruction_it->source_location);
+        // note that instruction_it is not advanced by this call,
+        // so no need to move it backwards
+        body.insert_before_swap(instruction_it, invalidation_assignment);
+      }
+      else
+      {
+        throw incorrect_goto_program_exceptiont(
+          "Found a `DEAD` variable without corresponding `DECL`!",
+          instruction_it->source_location);
```
feliperodri, line 937: Nice! We should use this instead of `throw 0`.

feliperodri (approved):

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -669,21 +668,98 @@ void code_contractst::instrument_call_statement(
...
+      const auto object_validity_var_addr =
+        new_tmp_symbol(
+          pointer_type(bool_typet{}),
+          instruction_it->source_location,
+          symbol_table.lookup_ref(function).mode,
+          symbol_table)
+          .symbol_expr();
+      write_set_validity_addrs.insert(object_validity_var_addr);
+
+      alias_checking_instructions.add(goto_programt::make_decl(
```
jimgrundy, line 710: Should we be concerned that there is no test covering this line?

SaswatPadhi: Most likely a false positive from Codecov. Other statements in the same loop are covered, and there is no `break` or `continue` in between.

PR #6387, opened 2021-10-08
introduce code_frontend_declt

kroening:
This introduces code_fontend_declt, which separates code_declt (for use in goto instructions) and the statement in C programs.

The two differ (the variant for goto programs has no initializer), and separating them enables changes.

### DIFF
--- a/src/ansi-c/c_typecheck_code.cpp
+++ b/src/ansi-c/c_typecheck_code.cpp
@@ -316,7 +316,7 @@ void c_typecheck_baset::typecheck_decl(codet &code)
     }
     else
     {
-      code_declt decl(symbol.symbol_expr());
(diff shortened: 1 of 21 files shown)


peterschrammel (commented):
What is the next step after this? Is it to remove deprecated uses of `code_Xt` in goto programs?

kroening:
yes, the goal is to remove the use of codet in goto-programs.

peterschrammel (approved):
Would it make sense to keep the name `code_declt` for the frontend occurrences that will be kept in the end and rename the deprecated uses in goto-program?

kroening:
This is to be consistent with `frontend_pointer`.

tautschnig (approved):

PR #6388, opened 2021-10-11
Address incremental SMT2 related code review comments from PR6357

thomasspriggs:
This PR addresses incremental SMT2 related code review comments from 
https://github.com/diffblue/cbmc/pull/6357. Some of the outstanding comments on this PR were not blocking and the PR was therefore merged without addressing them. This is the follow up PR which cleans this up.

### DIFF
--- a/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
+++ b/src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
@@ -49,7 +49,7 @@ void smt2_incremental_decision_proceduret::define_dependent_functions(
         return expr_identifier.first;
       });
   std::stack<exprt> to_be_defined;
-  const auto dependencies_needed = [&](const exprt &expr) {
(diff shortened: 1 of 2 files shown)


TGWDB (approved):

src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -49,7 +49,7 @@ void smt2_incremental_decision_proceduret::define_dependent_functions(
         return expr_identifier.first;
       });
   std::stack<exprt> to_be_defined;
-  const auto dependencies_needed = [&](const exprt &expr) {
+  const auto push_dependencies_needed = [&](const exprt &expr) {
```
NlightNFotis, line 52: Can the type of the return value of the lambda also be defined there? `-> bool` - because it made it surprising to see a lambda named `push-*` in a context that demands a predicate without it having an explicit book return value.

PR #6395, opened 2021-10-13
enabling `#pragma disable "pointer-primitive"`

### LINKED ISSUE #6239: Add support for the "check disable" pragma to disable the pointer-primitive check
CBMC version: 5.34.0
Operating system: N/A
Exact command line resulting in the issue: cbmc main.c --pointer-primitive-check

Running this program, I expect the analysis to pass, exactly as it does if I don't include the "--pointer-primitive-check" on the command line.

```
int main()
{
  void *p = malloc(1);
  void *q;

#pragma CPROVER check push
#pragma CPROVER check disable "pointer-primitive"
  if (__CPROVER_same_object(p,q)) {
  } else {
  }
#pragma CPROVER check pop
}
```

What happened instead:

CBMC version 5.34.0 (cbmc-5.34.0-dirty) 64-bit x86_64 macos
Parsing main.c
file main.c line 4 function main: Unsupported #pragma CPROVER before ' '
file main.c line 4 function main: syntax error before ' '
PARSING ERROR

Note: this is the same output you get if you use a complete random string in the pragma, say "dog" instead of "pointer-primtive". However, _some_ other check names, like "pointer" do work.

This request has 2 parts:
1/ Support pointer-primitive in this pragma so that it can be disabled for code regions.
2/ Give a better error message that includes the string that wasn't recognized instead of ''

remi-delmas-3000:
Fixes #6239.

This PR enables `#pragma disable "pointer-primitive"` by adding "pointer-primitive" to the list of named checks accepted by the ansi-c lexer, to allow users to disable pointer primitive checks using pragmas embedded in C code. This addresses issue https://github.com/diffblue/cbmc/issues/6239. The `goto_check` function already handles "disable:pointer-primitive-check" so only a lexer modification was needed. 

### DIFF
--- a/regression/cbmc/pragma_cprover3/test.desc
+++ b/regression/cbmc/pragma_cprover3/test.desc
@@ -1,0 +1,16 @@
+CORE
+main.c
+--pointer-primitive-check
+^main.c function main$
+^\[main.pointer_primitives.\d+\] line 17 pointer invalid in POINTER_OBJECT\(\(const void \*\)p\): SUCCESS$
+^\[main.pointer_primitives.\d+\] line 17 deallocated dynamic object in POINTER_OBJECT\(\(const void \*\)p\): SUCCESS$
+^\[main.pointer_primitives.\d+\] line 17 dead object in POINTER_OBJECT\(\(const void \*\)p\): SUCCESS$
+^\[main.pointer_primitives.\d+\] line 17 pointer outside object bounds in POINTER_OBJECT\(\(const void \*\)p\): SUCCESS$
+^\[main.pointer_primitives.\d+\] line 17 pointer invalid in POINTER_OBJECT\(\(const void \*\)q\): FAILURE$
+^\[main.pointer_primitives.\d+\] line 17 deallocated dynamic object in POINTER_OBJECT\(\(const void \*\)q\): SUCCESS$
+^\[main.pointer_primitives.\d+\] line 17 dead object in POINTER_OBJECT\(\(const void \*\)q\): SUCCESS$
+^\[main.pointer_primitives.\d+\] line 17 pointer outside object bounds in POINTER_OBJECT\(\(const void \*\)q\): FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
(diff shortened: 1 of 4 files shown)


regression/cbmc/pragma_cprover3/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+--pointer-primitive-check
+^\*\* 0 of 2 failed
```
SaswatPadhi: This pattern seems a bit fragile -- the `2` checks come from `malloc` but if the library implementation changes, that number could be different in the future.

I think just checking `VERIFICATION SUCCESSFUL`, as you are already doing below, should suffice?

SaswatPadhi (commented):
A minor comment below

jimgrundy:
We need this to unblock some of the loop-contracts work. Can someone take a look?

martin-cs (approved):
Looks good.  If you really want to gold plate it you could have something that fails the `--pointer-primitive-check` outside of the area covered by the `#pragma`.

martin-cs:
@jimgrundy HTH but note that there are currently regression failures of this PR and it cannot be merged until those are resolved.

kroening:
I had hoped to get some clarification on the use-case.
The __CPROVER_-prefixed pointer primitives are not really meant to be used in user-code, so why would one want to have checks for them?

remi-delmas-3000:
> Looks good. If you really want to gold plate it you could have something that fails the `--pointer-primitive-check` outside of the area covered by the `#pragma`.

will do

jimgrundy:
Should have been more specific. I was hoping one of "us" would be prompted to take a look at it, at least to get those build fails fixed so we can talk about merging.

jimgrundy:
@kroening "The _CPROVER-prefixed pointer primitives are not really meant to be used in user-code, so why would one want to have checks for them?". Well, yes, sure. But in a way we're agreeing with you here by asking for a way to shut them off for selected code.

Why do we turn them on globally in the first place though I guess would be a reasonable question. If we never turned them on globally with a command-line flag we wouldn't need a pragma to turn them off locally. One might ask more generally why these checks were implemented in the first place.

I think the answer is that we find that these checks can point out misunderstandings in the instrumentation code that we generate and so we've grown to like them. However, there are times when in local sections we feel like we know what we are doing and would be better served by turning them off.

More generally, there are a number of checks that can be enabled globally via a command-line flag and disabled locally with a pragma. This one has been the odd one out in that you can enable it globally but not disable it locally. It just seems that for consistency they should all work the same.

SaswatPadhi (approved):
LGTM; thanks!

peterschrammel (approved):

TGWDB:
@kroening Has your question been answered? Also are you the one that approval is now waiting on?

chrisr-diffblue (approved):

Issue #6231, opened 2021-07-15
Check flags are not respected within quantified expressions


SaswatPadhi:
**CBMC version:** 5.34.0
**Operating system:** Linux

**Test case:**

```c
#include <assert.h>
#include <stdlib.h>

int main() {
  char *a;

  // lots of errors with `--pointer-check` enabled
  assert(*a == *a);

  // no errors even with `--pointer-check` enabled
  assert(
    __CPROVER_forall {
      int i ; (0 <= i && i < 1) ==> *(a+i) == *(a+i)
    }
  );
}
```

**Exact command line resulting in the issue:**

```console
$ cbmc --pointer-check test.c
```

**What behaviour did you expect:**

`NULL` and invalid pointer issues would be checked within both assertions.

**What happened instead:**

`NULL` and invalid pointer issues are checked within first assertion but not the second. Commenting out the first assertion results in `VERIFICATION SUCCESSFUL` output even with `--pointer-check` flag.

The bounds for the quantified variable are constants and CBMC does not ignore the quantifier (with the SAT backend).

Even with the SMT backend, with which quantifiers are not unrolled, `--pointer-check` is not really checked within the quantified expression. Running the following gives `VERIFICATION SUCCESSFUL` on commenting the first assertion:

```console
$ cbmc --z3 --pointer-check test.c
```

SaswatPadhi:
cc: @feliperodri

martin-cs:
https://github.com/diffblue/cbmc/blob/06c563a76589ed00cc372d68e2819880537977b2/src/analyses/goto_check.cpp#L1565

is the problem but the fix is non-trivial because you may need the assert in side the range of the quantifiers which doesn't work because you can't have an `instructiont` inside an `exprt`.  Of the top of my head I think you might have to create a copy of the quantified expression and replace some terms with the conditions that check the assertion.  This feels kind of nasty though.

```
  assert( __CPROVER_forall { int i ; (0 <= i && i < 1) ==> *(a+i) == *(a+i)  } );
```

would become:

```
  assert( __CPROVER_forall { int i ; (0 <= i && i < 1) ==> POINTER_CHECK(*(a+i))  } );
  assert( __CPROVER_forall { int i ; (0 <= i && i < 1) ==> *(a+i) == *(a+i)  } );
```

martin-cs:
Of course, this is just the tip of the iceberg as what happens if the checks are from the side that is bounding the range, for example `0 <= i && i < *p`.

[...]

PR #6399, opened 2021-10-18
Allow checks to be performed inside quantifier statements.

NlightNFotis:
This allows checks from CLI flags like `--pointer-check` to be performed inside
expressions in quantifier statements.

This should be addressing issue #6231 .

### DIFF
--- a/regression/cbmc-primitives/exists_memory_checks/invalid_index_range.desc
+++ b/regression/cbmc-primitives/exists_memory_checks/invalid_index_range.desc
@@ -1,0 +1,12 @@
+CORE
+invalid_index_range.c
+--pointer-check
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+\[main\.assertion\.1\] line 9 assertion __CPROVER_exists \{ int i; \(0 <= i && i < 20\) && a\[i\] == i \*i \}: SUCCESS
+line 9 dereference failure: pointer outside object bounds in a\[\(signed (long|long long) int\)i\]: FAILURE
+--
+--
+Check that memory checks fail for pointer dereferences inside an existential
+qualifier, for out of bounds memory access.
--- a/regression/cbmc-primitives/exists_memory_checks/negated_exists.desc
+++ b/regression/cbmc-primitives/exists_memory_checks/negated_exists.desc
@@ -1,0 +1,17 @@
+CORE
+negated_exists.c
+--pointer-check
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+\[main\.assertion\.1\] line 9 assertion !__CPROVER_exists \{ int i; \(0 <= i && i < 10\) && a\[i\] == 42 \}: SUCCESS
+\[main\.pointer_dereference\.7\] line 9 dereference failure: pointer NULL in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.8\] line 9 dereference failure: pointer invalid in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.9\] line 9 dereference failure: deallocated dynamic object in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.10\] line 9 dereference failure: dead object in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.11\] line 9 dereference failure: pointer outside object bounds in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.12\] line 9 dereference failure: invalid integer address in a\[\(signed (long|long long) int\)i\]: SUCCESS
+--
+--
+Check that memory checks pass for valid pointer dereferences inside a negated
+existential qualifier.
--- a/regression/cbmc-primitives/exists_memory_checks/smt_missing_range_check.desc
+++ b/regression/cbmc-primitives/exists_memory_checks/smt_missing_range_check.desc
@@ -1,0 +1,15 @@
+CORE smt-backend
+smt_missing_range_check.c
+--pointer-check -z3
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+\[main\.assertion\.1\] line \d assertion __CPROVER_exists \{ int i; a\[i\] == i \*i \}: SUCCESS
+\[main\.pointer_dereference\.11\] line \d dereference failure: pointer outside object bounds in a\[\(signed (long|long long) int\)i\]: FAILURE
+--
+--
+Check that memory checks fail for pointer dereferences inside an existential
+qualifier, for out of bounds memory access, when using the smt backend and
+the range of the index is unbound. Note that this test is not expected to work
+with the SAT backend at the time of writing, as the SAT backend does not support
+qualifiers in this form.
--- a/regression/cbmc-primitives/exists_memory_checks/valid_index_range.desc
+++ b/regression/cbmc-primitives/exists_memory_checks/valid_index_range.desc
@@ -1,0 +1,17 @@
+CORE
+valid_index_range.c
+--pointer-check
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+\[main\.assertion\.1\] line 9 assertion __CPROVER_exists \{ int i; \(0 <= i && i < 10\) && a\[i\] == i \*i \}: SUCCESS
+\[main\.pointer_dereference\.7\] line 9 dereference failure: pointer NULL in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.8\] line 9 dereference failure: pointer invalid in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.9\] line 9 dereference failure: deallocated dynamic object in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.10\] line 9 dereference failure: dead object in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.11\] line 9 dereference failure: pointer outside object bounds in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.12\] line 9 dereference failure: invalid integer address in a\[\(signed (long|long long) int\)i\]: SUCCESS
+--
+--
+Check that memory checks pass for valid pointer dereferences inside an
+existential qualifier.
--- a/regression/cbmc-primitives/forall_6231_1/test.c
+++ b/regression/cbmc-primitives/forall_6231_1/test.c
@@ -1,0 +1,20 @@
+#include <assert.h>
+#include <stdlib.h>
+
+// clang-format off
+int main() {
+  char *a = malloc(1);
+
+  assert(*a == *a);
+
+  // BUG: In https://github.com/diffblue/cbmc/issues/6231, it was reported that
+  // no checks would be performed on the derefence inside the quantified statement,
+  // even when explicitly requested via for instance `--pointer-check`, because
+  // we would simply skip over these quantified statements in goto-check.
+  assert(
+    __CPROVER_forall {
+      int i ; (0 <= i && i < 1) ==> *(a+i) == *(a+i)
+    }
+  );
+}
+// clang-format on
--- a/regression/cbmc-primitives/forall_6231_2/test.c
+++ b/regression/cbmc-primitives/forall_6231_2/test.c
@@ -1,0 +1,23 @@
+#include <assert.h>
+#include <stdlib.h>
+
+// clang-format off
+int main() {
+  char *a = malloc(128);
+
+  // BUG: In https://github.com/diffblue/cbmc/issues/6231, it was reported that
+  // no checks would be performed on the derefence inside the quantified statement,
+  // even when explicitly requested via for instance `--pointer-check`, because
+  // we would simply skip over these quantified statements in goto-check.
+  assert(
+    __CPROVER_forall {
+      int i ; (0 <= i && i < 1) ==> *(a+i) == *(a+i)
+    }
+  );
+
+  assert(
+  __CPROVER_forall {
+    int j; !(0 <= j && j < 1) || (j == 0 && *(a+j) == *(a+j))
+  });
+}
+// clang-format on
--- a/regression/cbmc-primitives/forall_6231_2/test.desc
+++ b/regression/cbmc-primitives/forall_6231_2/test.desc
@@ -1,0 +1,23 @@
+CORE
+test.c
+--pointer-check
+^EXIT=0$
+^SIGNAL=0$
+\[main\.assertion\.1\] line \d+ assertion __CPROVER_forall \{ int i ; \(0 <= i && i < 1\) ==> \*\(a\+i\) == \*\(a\+i\) \}: SUCCESS
+\[main\.pointer_dereference\.1\] line \d+ dereference failure: pointer NULL in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.2\] line \d+ dereference failure: pointer invalid in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.3\] line \d+ dereference failure: deallocated dynamic object in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.4\] line \d+ dereference failure: dead object in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.5\] line \d+ dereference failure: pointer outside object bounds in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.pointer_dereference\.6\] line \d+ dereference failure: invalid integer address in a\[\(signed (long|long long) int\)i\]: SUCCESS
+\[main\.assertion.2] line \d+ assertion __CPROVER_forall \{ int j; \!\(0 <= j && j < 1\) || \(j == 0 && \*\(a\+j\) == \*\(a+j\)\) \}: SUCCESS
+\[main\.pointer_dereference\.7] line \d+ dereference failure: pointer NULL in a\[\(signed (long|long long) int\)j\]: SUCCESS
+\[main\.pointer_dereference\.8] line \d+ dereference failure: pointer invalid in a\[\(signed (long|long long) int\)j\]: SUCCESS
+\[main\.pointer_dereference\.9] line \d+ dereference failure: deallocated dynamic object in a\[\(signed (long|long long) int\)j\]: SUCCESS
+\[main\.pointer_dereference\.10] line \d+ dereference failure: dead object in a\[\(signed (long|long long) int\)j\]: SUCCESS
+\[main\.pointer_dereference\.11] line \d+ dereference failure: pointer outside object bounds in a\[\(signed (long|long long) int\)j\]: SUCCESS
+\[main\.pointer_dereference\.12] line \d+ dereference failure: invalid integer address in a\[\(signed (long|long long) int\)j\]: SUCCESS
+^VERIFICATION SUCCESSFUL$
+--
+--
+This is fixing an issue first reported in https://github.com/diffblue/cbmc/issues/6231
--- a/regression/cbmc-primitives/forall_6231_3/test.c
+++ b/regression/cbmc-primitives/forall_6231_3/test.c
@@ -1,0 +1,24 @@
+#include <assert.h>
+#include <stdlib.h>
+
+// This is essentially the same file as in test `forall_6231_1`, with the difference
+// being that the forall statement contains a bigger bound, so that we are to have
+// more concrete instantiations of the bound variable.
+
+// clang-format off
+int main() {
+  char *a = malloc(10);
+
+  assert(*a == *a);
+
+  // BUG: In https://github.com/diffblue/cbmc/issues/6231, it was reported that
+  // no checks would be performed on the derefence inside the quantified statement,
+  // even when explicitly requested via for instance `--pointer-check`, because
+  // we would simply skip over these quantified statements in goto-check.
+  assert(
+    __CPROVER_forall {
+      int i ; (0 <= i && i < 10) ==> *(a+i) == *(a+i)
+    }
+  );
+}
+// clang-format on
--- a/regression/cbmc-primitives/forall_6231_3/test_malloc_less_than_bound.c
+++ b/regression/cbmc-primitives/forall_6231_3/test_malloc_less_than_bound.c
@@ -1,0 +1,21 @@
+#include <assert.h>
+#include <stdlib.h>
+
+// Similar to our test in `test.c` of this folder, with the difference being
+// that the malloc size is less than the bound checked, which implies that the
+// check for the pointer being outside the object bounds is expected to fail.
+
+// clang-format off
+int main() {
+  char *a = malloc(4);
+
+  assert(*a == *a);
+
+  // BUG: no errors even with `--pointer-check` enabled -- now fixed.
+  assert(
+    __CPROVER_forall {
+      int i ; (0 <= i && i < 10) ==> *(a+i) == *(a+i)
+    }
+  );
+}
+// clang-format on
--- a/regression/cbmc-primitives/CMakeLists.txt
+++ b/regression/cbmc-primitives/CMakeLists.txt
@@ -1,3 +1,11 @@
-add_test_pl_tests(
-    "$<TARGET_FILE:cbmc>"
-)
+find_program(Z3_EXISTS "z3")
+message(${Z3_EXISTS})
+if(Z3_EXISTS)
+    add_test_pl_tests(
+        "$<TARGET_FILE:cbmc>"
+    )
+else()
+    add_test_pl_tests(
+        "$<TARGET_FILE:cbmc>" -X smt-backend
+    )
+endif()
--- a/regression/cbmc-primitives/alternating_quantifiers_6231/exists_in_forall.c
+++ b/regression/cbmc-primitives/alternating_quantifiers_6231/exists_in_forall.c
@@ -1,0 +1,14 @@
+#include <stdlib.h>
+
+// clang-format off
+int main(int argc, char **argv)
+{
+    int *i = malloc(sizeof(int));
+    *i = 1;
+
+    __CPROVER_assert(
+        __CPROVER_forall { int z; (0 < z && z < 10) ==>
+            __CPROVER_exists { int y; ( 10 < y && y < 20) && y == z + 10 && y > *i } },
+        "for all z, there exists a y so that y = z + 10 and y > 1");
+}
+// clang-format on
--- a/regression/cbmc-primitives/alternating_quantifiers_6231/exists_in_forall.desc
+++ b/regression/cbmc-primitives/alternating_quantifiers_6231/exists_in_forall.desc
@@ -1,0 +1,16 @@
+CORE
+exists_in_forall.c
+--pointer-check
+^EXIT=0$
+^SIGNAL=0$
+\[main\.assertion\.1\] line \d* for all z, there exists a y so that y = z \+ 10 and y > 1: SUCCESS
+\[main\.pointer_dereference\.7\] line \d+ dereference failure: pointer NULL in \*i: SUCCESS
+\[main\.pointer_dereference\.8\] line \d+ dereference failure: pointer invalid in \*i: SUCCESS
+\[main\.pointer_dereference\.9\] line \d+ dereference failure: deallocated dynamic object in \*i: SUCCESS
+\[main\.pointer_dereference\.10\] line \d+ dereference failure: dead object in \*i: SUCCESS
+\[main\.pointer_dereference\.11\] line \d+ dereference failure: pointer outside object bounds in \*i: SUCCESS
+\[main\.pointer_dereference\.12\] line \d+ dereference failure: invalid integer address in \*i: SUCCESS
+--
+--
+The assertion reference number here is present so that we make sure
+we always match the properties of the dereference inside the exists.
--- a/regression/cbmc-primitives/alternating_quantifiers_6231/forall_in_exists.c
+++ b/regression/cbmc-primitives/alternating_quantifiers_6231/forall_in_exists.c
@@ -1,0 +1,14 @@
+#include <stdlib.h>
+
+// clang-format off
+int main(int argc, char **argv)
+{
+    int *i = malloc(sizeof(int));
+    *i = 1;
+
+    __CPROVER_assert(
+        __CPROVER_exists { int z; (0 < z && z < 2) &&
+            __CPROVER_forall { int o; (10 < o && o < 20) ==> o > z && z == * i }},
+        "there exists a z between 0 and 2 so that for all o between 10 and 20, o > z and z = 1");
+}
+// clang-format on
--- a/regression/cbmc-primitives/alternating_quantifiers_6231/forall_in_exists.desc
+++ b/regression/cbmc-primitives/alternating_quantifiers_6231/forall_in_exists.desc
@@ -1,0 +1,14 @@
+CORE
+forall_in_exists.c
+--pointer-check
+^EXIT=0$
+^SIGNAL=0$
+\[main\.assertion\.1\] line \d* there exists a z between 0 and 2 so that for all o between 10 and 20, o > z and z = 1: SUCCESS
+\[main\.pointer_dereference\.7\] line \d+ dereference failure: pointer NULL in \*i: SUCCESS
+\[main\.pointer_dereference\.8\] line \d+ dereference failure: pointer invalid in \*i: SUCCESS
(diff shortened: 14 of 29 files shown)


martin-cs (commented):
I can't immediately see anything wrong with this BUT the whole thing makes me uneasy.  The comments have been there since 2013 which makes me think that they were probably there for a reason.  My immediate thought is about scoping and whether the quantified variable is actually live when the variable is checked.  I am also unsure about how nested quantifiers are going to work and to what degree the guards are applied to the checks as well.

Can you please say a little more about how you are expecting things to work?

regression/cbmc-primitives/forall_6231_2/test.c
```diff
@@ -0,0 +1,17 @@
+#include <assert.h>
+#include <stdlib.h>
+
+int main() {
+  char *a = malloc(128);
+
+  assert(
+    __CPROVER_forall {
+      int i ; (0 <= i && i < 1) ==> *(a+i) == *(a+i)
```
martin-cs, line 14: Likewise; not just one instantiation please.

regression/cbmc-primitives/forall_6231_1/test.c
```diff
@@ -0,0 +1,16 @@
...
+int main() {
+  char *a = malloc(1);
+
+  // lots of errors with `--pointer-check` enabled
+  assert(*a == *a);
+
+  // no errors even with `--pointer-check` enabled
+  assert(
+    __CPROVER_forall {
+      int i ; (0 <= i && i < 1) ==> *(a+i) == *(a+i)
```
martin-cs, line 16: Please can you change this so that it doesn't just have one true instantiation?

regression/cbmc-primitives/forall_6231_2/test.c
```diff
@@ -0,0 +1,17 @@
...
+    }
+  );
+
+  assert(
+  __CPROVER_forall {
+    int j; !(0 <= j && j < 1) || (j == 0 && *(a+j) == *(a+j))
```
martin-cs, line 20: Again, multiple instantiations.

regression/cbmc-primitives/forall_6231_2/test.desc
```diff
@@ -0,0 +1,23 @@
...
+\[main\.pointer_dereference\.5\] line \d dereference failure: pointer outside object bounds in a\[\(signed long int\)i\]: SUCCESS
+\[main\.pointer_dereference\.6\] line \d dereference failure: invalid integer address in a\[\(signed long int\)i\]: SUCCESS
+\[main\.assertion.2] line \d+ assertion __CPROVER_forall \{ int j; \!\(0 <= j && j < 1\) || \(j == 0 && \*\(a\+j\) == \*\(a+j\)\) \}: SUCCESS
+\[main\.pointer_dereference\.7] line \d+ dereference failure: pointer NULL in a\[\(signed long int\)j\]: SUCCESS
+\[main\.pointer_dereference\.8] line \d+ dereference failure: pointer invalid in a\[\(signed long int\)j\]: SUCCESS
+\[main\.pointer_dereference\.9] line \d+ dereference failure: deallocated dynamic object in a\[\(signed long int\)j\]: SUCCESS
+\[main\.pointer_dereference\.10] line \d+ dereference failure: dead object in a\[\(signed long int\)j\]: SUCCESS
+\[main\.pointer_dereference\.11] line \d+ dereference failure: pointer outside object bounds in a\[\(signed long int\)j\]: SUCCESS
+\[main\.pointer_dereference\.12] line \d+ dereference failure: invalid integer address in a\[\(signed long int\)j\]: SUCCESS
+^VERIFICATION SUCCESSFUL$
```
martin-cs, line 20: Please can we have one test that fails?

martin-cs (changes requested):
A few more concrete requests for other functionality.

kroening (changes requested):
This happens to work for the specific case of positive forall, since this is doing Skolemisation implicitly.

1. This will fail (generating VCCs that are too strong) for exists.
2. I would much prefer that the Skolemisation is done explicitly, i.e., by introducing a Skolem constant and instantiating the quantifier with it.

regression/cbmc-primitives/forall_6231_2/test.c
NlightNFotis: Hi @martin-cs, I've left these as they stand, and added a separate test that does multiple instantiations (by essence of it having a bigger upper bound).

Does that sound like an adequate solution to you?

regression/cbmc-primitives/forall_6231_2/test.desc
NlightNFotis: By a similar token, I've added a test that has the `pointer being outside of object bounds` condition fail by having the `malloc_size < upper_bound`.

Is that adequate, or should I spend some more time thinking about it?

martin-cs: Can I see the extra tests before I comment on them?

NlightNFotis: Yes, of course - I've added them as a separate commit (the third one).

NlightNFotis:
Hi Daniel @kroening.

With regards to your first query, we have pushed some changes that show that existentially quantified statements appear to be working as they should (or at least, how we expect them to). Could you review these and let us know if these much your expectations, or if there's a counterexample case you could think of that would help us test our change in a way that exposes a weakness of our current implementation?

regression/cbmc-primitives/exists_6231_2/test.c
```diff
@@ -0,0 +1,18 @@
...
+#include <stdlib.h>
+
+// clang-format off
+int main() {
+  char *a = malloc(1);
+
+  // lots of errors with `--pointer-check` enabled
+  assert(*a == *a);
+
+  // no errors even with `--pointer-check` enabled
```
TGWDB: This comment does not look correct.

regression/cbmc-primitives/exists_6231_2/test.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+test.c
+--pointer-check
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
```
TGWDB: Does this check the quantified expression, or will it "pass" when the assert on line 9 fails?

regression/cbmc-primitives/forall_6231_3/test.c
```diff
@@ -0,0 +1,22 @@
...
+
+// This is essentially the same file as in test `forall_6231_1`, with the difference
+// being that the forall statement contains a bigger bound, so that we are to have
+// more concrete instantiations of the bound variable.
+
+// clang-format off
+int main() {
+  char *a = malloc(10);
+
+  // lots of errors with `--pointer-check` enabled
```
TGWDB: This comment is false?

regression/cbmc-primitives/forall_6231_3/test_malloc_less_than_bound.c
```diff
@@ -0,0 +1,22 @@
...
+// check for the pointer being outside the object bounds is expected to fail.
+
+// clang-format off
+int main() {
+  char *a = malloc(4);
+
+  // lots of errors with `--pointer-check` enabled
+  assert(*a == *a);
+
+  // no errors even with `--pointer-check` enabled
```
TGWDB: False comment?

TGWDB (changes requested):

regression/cbmc-primitives/exists_6231_2/test.c
```diff
@@ -0,0 +1,18 @@
+#include <assert.h>
+#include <stdlib.h>
+
+// clang-format off
+int main() {
+  char *a = malloc(1);
+
+  // lots of errors with `--pointer-check` enabled
```
martin-cs: Why lots of errors; the only issue I can see is that `*a` is uninitialised.

regression/cbmc-primitives/exists_6231_2/test.desc
martin-cs: Yeah; this is too general.  Is there a reason why you are no longer checking each of the VCs gives the right answer?

regression/cbmc-primitives/exists_memory_checks/invalid_index_range.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+invalid_index_range.c
+--pointer-check
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+line 9 dereference failure: pointer outside object bounds in a\[\(signed long int\)i\]: FAILURE
```
martin-cs: OK; this is encouraging.

regression/cbmc-primitives/exists_memory_checks/smt_missing_range_check.desc
```diff
@@ -0,0 +1,15 @@
+CORE
+invalid_index_range.c
+--pointer-check -z3
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+line 9 assertion __CPROVER_exists \{ int i; \(0 <= i && i < 20\) && a\[i\] == i \*i \}: SUCCESS
```
martin-cs: Where does the `0 <= i && i < 20` come from?

regression/cbmc-primitives/forall_6231_1/test.c
```diff
@@ -0,0 +1,18 @@
+#include <assert.h>
+#include <stdlib.h>
+
+// clang-format off
+int main() {
+  char *a = malloc(1);
+
+  // lots of errors with `--pointer-check` enabled
```
martin-cs: Again, why?

regression/cbmc-primitives/forall_6231_2/test.c
martin-cs: It is OK but I think it might make these tests redundant.

regression/cbmc-primitives/forall_6231_2/test.desc
martin-cs: Thanks!

martin-cs (changes requested):
1. Do you have examples of quantifiers in assumptions?
2. Can you briefly outline how this works?  What logical expression is generated?  How does this interact with scoping?

regression/cbmc-primitives/exists_6231_2/test.c
SaswatPadhi: In [my original example](#6231), `a` was uninitialized so `*a` was resulting in lots of issues (with `--pointer-check`).

I think we could remove this comment now that we `malloc` above.

regression/cbmc-primitives/forall_6231_3/test.c
SaswatPadhi: In [my original example](#6231), `a` was uninitialized so `*a` was resulting in lots of issues (with `--pointer-check`).

I think we could remove this comment now that we `malloc` above.

regression/cbmc-primitives/forall_6231_1/test.c
SaswatPadhi: In [my original example](#6231), `a` was uninitialized so `*a` was resulting in lots of issues (with `--pointer-check`).

I think we could remove this comment now that we `malloc` above.

SaswatPadhi (commented):
Thanks for fixing this! This would help us catch issues in code contracts. Earlier, invariants and function contracts that used implications but performed incorrect/unsafe operations were not caught.

All test cases in this PR seem to check that the consequent is safe. May be the fix already checks the antecedent as well, but can we also add a regression test for that? Something like:
```c
__CPROVER_forall {
  int i;
  (0 <= i && i <= (n / 0))    /* (n/0) should be caught by `--div-by-zero` */
  ==> ...
}
```

martin-cs:
@SaswatPadhi makes a good point.  It is not clear to me whether this treats quantified expressions as guard/statement pairs and if so, whether it checks in both, does it use the guard for the checks in the statement?

regression/cbmc-primitives/exists_6231_2/test.c
thomasspriggs: This test was from a work in progress commit, which should have now been dropped.

regression/cbmc-primitives/exists_6231_2/test.desc
thomasspriggs: This test was from a work in progress commit, which should have now been dropped.

regression/cbmc-primitives/exists_memory_checks/smt_missing_range_check.desc
NlightNFotis: This was a mistake - the test was checking against the wrong test file.

NlightNFotis:
> @SaswatPadhi makes a good point. It is not clear to me whether this treats quantified expressions as guard/statement pairs and if so, whether it checks in both, does it use the guard for the checks in the statement?

Hi Martin, yes, it does treat quantified expressions as guard/statement pairs and yes it uses the guard for the checks in the statement. I have factored the change that does that in [its own commit](https://github.com/diffblue/cbmc/pull/6399/commits/1cdc6e5fed1ba60635d10f5b175612a406f32d0e), with an explanation of how it works there.

As a shorter explanation here, `goto_checkt::check` in the same file calls `check_rec` on the expression with `true_exprt` as a guard, which is then refined based on the exact nature of the expression. We already have support for refining the guard for logical operators, but we were dispatching to that only if the expression had `expr.id() == ID_and` or `expr.id() == ID_or`. This appears to be a latent bug that was not manifest because we didn't have any expressions of the particular form that the tests submitted for this PR, so we didn't observe that. As soon as we started performing checks for quantifier expressions (basically, not immediately bailing upon seeing one), we hit this bug.

Does that answer your question?

NlightNFotis:
Hi @kroening ,

I'm trying to better understand the requirement of the second part of your comment, which I quote:

> I would much prefer that the Skolemisation is done explicitly, i.e., by introducing a Skolem constant and instantiating the quantifier with it.

My understanding is that Skolemization is being done the same way it was being done before this patch, i.e. in `goto_symext::symex_assert()`  of `src/goto-symex/symex_main.cpp:157`, which appears to convert existential quantifiers into universals (the first part of the Skolemization process) and then performs a renaming of the bound variable (which performs the substitution by a Skolem function/Skolem constant).

Is that not adequate in some capacity? We have performed extended testing and it seems like this bug fix is working as intended - but if you believe that there's a chance some input might fail we're happy to investigate further and tighten the coverage of the tests for this.

Or is your request that we lift this all into a separate module? If yes, where do you think this would be better plugged in - the simplification modules, a completely different pass over the goto-program that hooks into the transformation pipeline, or somewhere else entirely?

TGWDB (approved):

martin-cs:
@NlightNFotis Thanks for your response, the explanation does indeed help and yes, I think you are correct about this being a latent bug.  Would it be possible to write a test case that doesn't use quantifiers that triggers this?  Would something like `y != 0 ==> 10/y <= 10` and `--div-by-zero-check` do?  That way the fixes could be split from the new feature and be approved even if there is more discussion over the checks in quantifiers.

Please could you have a look at, check and if necessary update:
https://github.com/diffblue/cbmc/blob/ce2680b724d650c404468eb0070c8b402735ec66/src/analyses/goto_check.cpp#L111
https://github.com/diffblue/cbmc/blob/ce2680b724d650c404468eb0070c8b402735ec66/src/analyses/goto_check.cpp#L1623

( The real cherry on top would be adding a `PRECONDITION` that checks for the expression types that are intended to be supported. )

kroening:
The problem is that the identifiers used in quantifiers are no longer guaranteed to be unique, i.e., you could have something like
`i[2]== 2 && exists i:int. x/i`
which will yield a check that reads
`(i[2] != 2) => i!=0`
In this formula, `i` is used twice, and isn't even type consistent -- the first instance is an array, and the second is an int. Formulas like that must not be generated.

You can't test that in the C front-end, since the C front-end's treatment of quantifiers pre-dates the introduction of scoping in the solver backends. Please don't even think about reintroducing this restriction, not even as a temporary hack.

The way to do this properly is to introduce a Skolem symbol for the bound identifier. You'll find that the classes for exists and forall have convenient methods for making that easy!

martin-cs:
@NlightNFotis while reviewing the changes you had kindly described I ran into some concerns.  I have documented them here: https://github.com/diffblue/cbmc/issues/6423  I don't consider them a precondition for this PR, but, obviously, they are linked.

NlightNFotis:
@kroening Hi Daniel, the [last commit](https://github.com/diffblue/cbmc/pull/6399/commits/63f683069c1e2096923e732329e74edf33aba048) is performing scolemisation of the formula in `goto_check`. Can you please review it now and let us know if this is adequate for what you had in mind?

@martin-cs Given that the bug-fix has now been merged on its own PR, could you please re-review this and let me know if there's anything more that needs be done?

src/analyses/goto_check.cpp
```diff
@@ -1747,6 +1749,26 @@ void goto_checkt::check_rec(const exprt &expr, guardt &guard)
     if(check_rec_member(to_member_expr(expr), guard))
       return;
   }
+  else if(expr.id() == ID_forall || expr.id() == ID_exists)
+  {
+    // In the case of a quantified statement, skolemise the expression
+    // and recurse.
+    const auto quantified_expr = to_quantifier_expr(expr);
+    const auto quantifier_sym = quantified_expr.symbol();
```
thomasspriggs: Clang-tidy suggests this should be `const auto &` to avoid unnecessary copy construction.

src/analyses/goto_check.cpp
```diff
@@ -1747,6 +1749,26 @@ void goto_checkt::check_rec(const exprt &expr, guardt &guard)
...
+    const auto new_symbol = get_fresh_aux_symbol(
+      quantifier_sym.type(),
+      CPROVER_PREFIX,
+      id2string(ns.lookup(quantifier_sym.get_identifier()).base_name),
+      quantified_expr.source_location(),
+      ID_C,
```
thomasspriggs: Given that this code is not part of the C front end, hard coding the language to C seems like the wrong choice. I suggest using the language mode of `quantifier_sym`.

src/analyses/goto_check.cpp
```diff
@@ -1747,6 +1749,26 @@ void goto_checkt::check_rec(const exprt &expr, guardt &guard)
...
+      const_cast<symbol_table_baset &>(ns.get_symbol_table()));
```
thomasspriggs: Don't cast away const! `get_fresh_aux_symbol` mutates the symbol table by adding symbols. Therefore casting away const violates the promise `goto_checkt` makes to not mutate the symbol table / by declaring them as const.

thomasspriggs (changes requested):

NlightNFotis:
Hi @kroening, can you take a look at the PR now and let us know if it is more or less what you envisioned for skolemization as a pass?

Specifically, it's the last commit that implements this.

src/goto-programs/skolemization.cpp
```diff
@@ -0,0 +1,69 @@
...
+      if(instruction.has_condition())
+      {
+        auto &cond = instruction.condition_nonconst();
+        recurse_on_condition(cond, goto_model);
+      }
+    }
+  }
+}
+
+void recurse_on_condition(exprt &cond, goto_modelt &model)
```
thomasspriggs: :pick: `recurse_on_` just tells the reader how this function is implemented, not what it does with/to conditions.

src/analyses/goto_check.cpp
thomasspriggs: No longer relevant.

src/goto-programs/skolemization.cpp
```diff
@@ -0,0 +1,69 @@
...
+  // recurse on the condition's subexpressions first - that allows us
+  // to correctly identify the nodes we care about in a compound
+  // expression, for instance something like `1 == 1 && exists(i.i>=0 && i>1)`
+  for(exprt &subexpr : cond.operands())
+  {
+    recurse_on_condition(subexpr, model);
+  }
+
+  // if the expression we're in doesn't have any more operands,
+  // it means it's a leaf node, so we go ahead and process it.
```
thomasspriggs: :no_entry: I find this comment really misleading. The following statement will be reached whether or not this is a leaf node. In fact the nodes of interest are not leaf nodes, as they are exists quantifiers and these quantifier expressions have child nodes. I think the phrase you are after is "depth first search".

src/analyses/goto_check.cpp
thomasspriggs: This is still hard coded in your new `rename_and_substitute` function and still needs fixing.

src/goto-programs/skolemization.h
```diff
@@ -0,0 +1,30 @@
...
+/// Rename bound variables in existential quantifiers. This is the entry
+/// point to the module.
+/// \param goto_model The goto_model on which we operate. Needed
+///    for making changes to the symbol table at variable renaming time.
+void rename_variables_in_existential_quantifiers(goto_modelt &goto_model);
+
+/// Helper function that recurses over the condition expression of instructions.
+/// \param cond The condition on which we recurse over the operands (subexpressions).
+/// \param model The goto_model we are inserting the new symbol to.
+void recurse_on_condition(exprt &cond, goto_modelt &model);
```
thomasspriggs: :pick: As this `recurse_on_condition` function and the `rename_and_substitute` function below are implementation details, I would not expose them as part of the interface in the header file. Note that functions can be forward declared in the `.cpp` file or re-ordered to avoid needing to forward declare at all.

src/goto-programs/skolemization.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+void recurse_on_condition(exprt &cond, goto_modelt &model)
+{
+  // recurse on the condition's subexpressions first - that allows us
+  // to correctly identify the nodes we care about in a compound
+  // expression, for instance something like `1 == 1 && exists(i.i>=0 && i>1)`
+  for(exprt &subexpr : cond.operands())
+  {
+    recurse_on_condition(subexpr, model);
```
thomasspriggs: Using recursion like this to visit all the nodes in the condition is re-implementing already standardised visitation. This should really be done with `exprt::visit_post`. I have tried out this refactor here - https://github.com/NlightNFotis/cbmc/compare/fix_6231...thomasspriggs:tas/fix_6231

src/goto-programs/skolemization.cpp
```diff
@@ -0,0 +1,69 @@
...
+    const auto new_symbol = get_fresh_aux_symbol(
+      quantifier_sym.type(),
+      CPROVER_PREFIX,
+      id2string(
+        model.symbol_table.lookup(quantifier_sym.get_identifier())->base_name),
+      quantifier_expr.source_location(),
+      ID_C,
+      model.symbol_table);
+    const auto symbol_expr =
+      symbol_exprt(new_symbol.name, quantifier_sym.type());
```
thomasspriggs: :bulb: Just use `symbolt::symbol_expr()`. Its a lot shorter and easier to read for me.

src/goto-programs/skolemization.cpp
```diff
@@ -0,0 +1,69 @@
...
+/// Rename variables in existentially quantified statements.
+/// Author: Diffblue Ltd.
+
+#include "skolemization.h"
+
+#include "goto_model.h"
+
+#include <ansi-c/expr2c.h>
+
+#include <util/expr_util.h>
```
thomasspriggs: :pick: Includes of `<ansi-c/expr2c.h>` and `<util/expr_util.h>` appear to be unused.

src/analyses/goto_check.cpp
thomasspriggs: This looks fixed.

src/goto-programs/skolemization.cpp
thomasspriggs: This comment is now gone.

src/goto-programs/skolemization.h
thomasspriggs: Ok.

src/goto-programs/skolemization.cpp
thomasspriggs: Ok.

src/goto-programs/skolemization.cpp
thomasspriggs: Ok, this is now resolved.

src/goto-programs/skolemization.cpp
thomasspriggs: Ok, this is now resolved.

src/goto-programs/skolemization.cpp
thomasspriggs: Ok, this is now resolved.

thomasspriggs (approved):

NlightNFotis:
Hello @martin-cs and @kroening , would it be possible to have a look at this again?

It's past our internal reviews and blocked on your reviews right now. I want to take this forward, either in terms of performing improvements and revisions if deemed necessary, or getting this merged.

martin-cs:
@NlightNFotis My apologies for the delay in reviewing.  I will look ASAP.

regression/cbmc-primitives/exists_memory_checks/invalid_index_range.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+invalid_index_range.c
+--pointer-check
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+line 9 dereference failure: pointer outside object bounds in a\[\(signed (long|long long) int\)i\]: FAILURE
```
martin-cs, line 8: Would it be possible to check that the assertion passes?

regression/cbmc-primitives/forall_6231_2/test.c
martin-cs: Ditto.

regression/cbmc-primitives/forall_6231_1/test.c
martin-cs: I see you have added this as a separate test.  OK.

regression/cbmc-primitives/exists_memory_checks/negated_exists.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+negated_exists.c
+--pointer-check
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+line 9 dereference failure:.*SUCCESS
```
martin-cs: A more complete check line as you have in other tests would be good.

regression/cbmc-primitives/exists_memory_checks/smt_missing_range_check.desc
```diff
@@ -0,0 +1,15 @@
+CORE
+smt_missing_range_check.c
+--pointer-check -z3
```
martin-cs, line 3: As we still don't have a CI run that doesn't have Z3 set up, please can this be tagged so that the regression tests still work on machines that don't have Z3 installed?

regression/cbmc-primitives/exists_memory_checks/valid_index_range.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+valid_index_range.c
+--pointer-check
+^EXIT=0$
+^SIGNAL=0$
+^VERIFICATION SUCCESSFUL$
+line 9 dereference failure:.*SUCCESS
```
martin-cs: More complete matching against results would be good.

regression/cbmc-primitives/forall_6231_1/test.c
```diff
@@ -0,0 +1,17 @@
+#include <assert.h>
+#include <stdlib.h>
+
+// clang-format off
+int main() {
+  char *a = malloc(1);
+
+  assert(*a == *a);
+
+  // BUG: no errors even with `--pointer-check` enabled -- now fixed.
```
martin-cs: I am not quite sure what this comment is trying to tell me.  Whatever it is, it feels like there might be a better place and way or saying it.

regression/cbmc-primitives/forall_6231_3/test_malloc_less_than_bound.c
```diff
@@ -0,0 +1,21 @@
...
+// that the malloc size is less than the bound checked, which implies that the
+// check for the pointer being outside the object bounds is expected to fail.
+
+// clang-format off
+int main() {
+  char *a = malloc(4);
+
+  assert(*a == *a);
+
+  // BUG: no errors even with `--pointer-check` enabled -- now fixed.
```
martin-cs, line 14: As above.

src/goto-programs/skolemization.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+/// Perform the first part of skolemisation, by identifying an existentially
+/// quantified statement and substituting the variables with skolem constants.
+/// \param condition The condition on which we search for quantifiers.
+/// \param symbol_table The symbol_table new symbols are inserted into.
+static void rename_variables_in_existential_quantifiers(
+  exprt &condition,
+  symbol_table_baset &symbol_table)
+{
+  condition.visit_post([&symbol_table](exprt &sub_expression) {
```
martin-cs: Thank you!

src/goto-programs/skolemization.cpp
```diff
@@ -0,0 +1,57 @@
+/// \file skolemization.h
+/// Rename variables in existentially quantified statements.
```
martin-cs: This is not what I would think of as Skolemization; this is renaming.  For me, Skolemization has to remove the existential so:

```
assert(__CPROVER_exists {  int i;   (0 <= i && i < 10) && a[i] == i *i   });
```

becomes:

```
int skolem_23_i;  // Note that if it is nested in a forall then it will need to be an uninterpreted function.
assert( (0 <= skolem_23_i && skolem_23_i < 10) && a[skolem_23_i] == skolem_23_i * skolem_23_i );
```

The general use-case for this is so that provers only have to deal with one kind of quantifier.

Note that the symex engine performs on-the-fly Skolemization where it can; that's how the SAT engine can handle quantifiers at all.  However it dynamically handles polarity, so depending on whether the quantifiers are in assume or assert, negated, etc. it will do the right thing.

martin-cs (commented):
I am removing my block on this but I still have some concerns:

* There are no tests using `assume`; that matters from a user point of view because of code contracts and also from a technical point of view as these are "flipped" during symex so what works with the SAT engine is different.
* A lot of the issues with quantifiers are simplified when there is only one level of quantifier alternation.  So I feel that alternating tests would be useful.
* The renaming code is not, IMHO, Skolemization and to me, doesn't match what @kroening requested.  However I believe that it fixes the immediate issue he raised and I will leave it to him to pursue that one.
* I'm still struggling a bit to see what is generated; I /think/ it is something like

`A && __CPROVER_forall { int i; B ==> C} `

will create a VC:

`assert(A && __CPROVER_forall { int i; B ==> CHECK_FROM_C } ); `

it would be good to get that confirmed.

martin-cs (approved):
Annoying -- github won't let me just remove my "Request changes" and return to a neutral state.  So, I am "approving" this but please don't treat this approved and merge.

kroening:
Please look at https://github.com/diffblue/cbmc/tree/goto_check_guard for an alternative approach, which relegates the problem of introducing the Skolem functions to the solver.

A key benefit of the alternative approach is that this means that there won't be two different kinds of goto-programs, one where bound symbols are unique and one where they aren't.

src/goto-programs/skolemization.cpp
NlightNFotis: Hi Martin,

Apologies for that - I'm aware of how skolemization works (in contrast to the code here), but when implemented in its proper form, it resulted in elimination of the existential quantifier far too early in the pipeline. There's code in `goto-check` or `goto-symex` (can't remember which of the two) - which depended on the quantifiers being there during VCC generation, and we were left with vccs that didn't semantically match the intended meaning, getting results that were all over the place.

The intended effect of the code above was to do a partial-skolemization, if we can admit that as a thing: perform the renaming, and let the existential quantifier stay, with the result being that the correct VCCs where generated, and it was being eliminated later in the pipeline in a way that felt safer (given that the variables were renamed before hand).

Definitely wasn't ideal, but there was a lot of uncertainty around the implementation of that feature in our pipeline.

My aim now is to drop this and substitute for Daniel's implementation (assuming it passes all tests).

regression/cbmc-primitives/exists_memory_checks/smt_missing_range_check.desc
NlightNFotis: Yes, in the last commit I have added a check for the `z3` binary in `CMakeLists.txt`

regression/cbmc-primitives/forall_6231_1/test.c
NlightNFotis: Apologies, I added this comment in a haste, and now realise that my phrasing was not the best 😞 

I have rephrased that in commit 604d12c. Can you let me know if that's clearer?

NlightNFotis:
Hi @martin-cs and @kroening , 

as requested, I've made several enhancements to the tests (added the extra checks Martin requested, tests based on the behaviour of the quantifiers in assumes, and alternating quantifier tests - each of which are in their own commits) and changed the implementation to be using Daniel's code (I lost the commit information because I couldn't cleanly cherry-pick after some changes to `goto-check`, but I hope my commit message preserves authorship information in an acceptable way).

Could you please have a look again and let me know how it looks to you now?

kroening:
May I suggest squashing the 9 commits that add/change the tests? Otherwise looks fine.

NlightNFotis:
Hi @kroening, if all is fine are you happy to lift your blocking review?

I will of course squash all the test commits before the final merge, I just want to give a chance to @martin-cs to review them on a commit by commit basis so that he can more easily check if his concerns have been addressed accordingly, and assuming he's also happy I will squash and merge.

kroening (approved):

PR #6144, opened 2021-05-24
Set the correct C++ standard version so that regression test pass rel…

martin-cs:
…iably

This should address issue:

https://github.com/diffblue/cbmc/issues/6090

- [ ] Each commit message has a non-empty body, explaining why the change was made.
- [ ] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [ ] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [ ] My PR is restricted to a single feature or bugfix.
- [ ] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #4053, opened 2019-02-03
OpenBSD compatibility

nanolith:
Patches to support CBMC on OpenBSD.

The ansi-c library has been patched to provide a compatible pthread_barrier_init function, and to bypass OpenBSD's stdio.h, which can't currently be parsed by the checker script.

The c_preprocess.cpp source file in ansi-c has been patched to pass the appropriate -std to Clang on OpenBSD.  OpenBSD does not use GNU extensions with the system Clang.

The config.cpp source file in util has been patched so that OpenBSD defaults to c99, just like FreeBSD.

Patches have been protected with a `defined(__OpenBSD__)` check to ensure that they do not impact other builds.

PR #6402, opened 2021-10-20
Set standard to C++17 when -std=c++17 or -std=gnu++17 are used

### LINKED ISSUE #6090: tests fail on parsing `noexcept` and `decltype` in c++17 mode
CBMC version: 5.29
Operating system: Fedora 34
Exact command line resulting in the issue: `ctest -L CORE`
What behaviour did you expect: all tests pass
What happened instead: 
`Failed test:  regression/cpp/virtual1/`
```
/usr/include/assert.h:69:1: error: parse error before '( true ) __attribute__'
      __THROW __attribute__ ((__noreturn__));
/usr/include/stdio.h:146:1: error: parse error before '( true ) ;'
 extern int remove (const char *__filename) __THROW;
PARSING ERROR
```
`Failed test: regression/cpp/Method_qualifier1`
```
/usr/include/c++/11/x86_64-redhat-linux/bits/c++config.h:2434:1: error: parse error before '; } namespace std'
   typedef decltype(nullptr)        nullptr_t;
/usr/include/assert.h:69:1: error: parse error before '( true ) __attribute__'
      __THROW __attribute__ ((__noreturn__));
PARSING ERROR
```

The problem is with parsing the `noexcept` and `decltype` expressions.

The reason of this problem is GCC 11 uses c++17 by default. If I specify `goto-cc -std=c++14/11/03`, then the mentioned tests pass.

lzaoral:

Based on suggestions by @martin-cs and @tautschnig.
Fixes: #6090 
Obsoletes: #6144

### DIFF
--- a/src/ansi-c/c_preprocess.cpp
+++ b/src/ansi-c/c_preprocess.cpp
@@ -551,6 +551,15 @@ bool c_preprocess_gcc_clang(
 #endif
         argv.push_back("-std=gnu++14");
       break;
+
+    case configt::cppt::cpp_standardt::CPP17:
+#if defined(__OpenBSD__)
+      if(preprocessor == configt::ansi_ct::preprocessort::CLANG)
+        argv.push_back("-std=c++17");
+      else
+#endif
+        argv.push_back("-std=gnu++17");
+      break;
     }
   }
   else
--- a/src/ansi-c/gcc_version.cpp
+++ b/src/ansi-c/gcc_version.cpp
@@ -120,6 +120,8 @@ void gcc_versiont::get(const std::string &executable)
               default_cxx_standard = configt::cppt::cpp_standardt::CPP11;
             else if(split[1] == "201402L")
               default_cxx_standard = configt::cppt::cpp_standardt::CPP14;
+            else if(split[1] == "201703L")
+              default_cxx_standard = configt::cppt::cpp_standardt::CPP17;
           }
         }
       }
--- a/src/analyses/goto_check.cpp
+++ b/src/analyses/goto_check.cpp
@@ -834,7 +834,8 @@ void goto_checkt::integer_overflow_check(
       {
         if(
           config.cpp.cpp_standard == configt::cppt::cpp_standardt::CPP11 ||
-          config.cpp.cpp_standard == configt::cppt::cpp_standardt::CPP14)
+          config.cpp.cpp_standard == configt::cppt::cpp_standardt::CPP14 ||
+          config.cpp.cpp_standard == configt::cppt::cpp_standardt::CPP17)
         {
           allow_shift_into_sign_bit = false;
         }
--- a/src/cpp/cpp_parser.cpp
+++ b/src/cpp/cpp_parser.cpp
@@ -21,9 +21,10 @@ bool cpp_parsert::parse()
 {
   // We use the ANSI-C scanner
   ansi_c_parser.cpp98=true;
-  ansi_c_parser.cpp11=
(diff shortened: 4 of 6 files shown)


peterschrammel (approved):

src/ansi-c/c_preprocess.cpp
```diff
@@ -551,6 +551,15 @@ bool c_preprocess_gcc_clang(
 #endif
         argv.push_back("-std=gnu++14");
       break;
+
+    case configt::cppt::cpp_standardt::CPP17:
+#if defined(__OpenBSD__)
```
kroening, line 556: This one is a bit puzzling? Why is OpenBSD different?

src/ansi-c/gcc_version.cpp
```diff
@@ -160,4 +162,7 @@ void configure_gcc(const gcc_versiont &gcc_version)
   config.ansi_c.gcc__float128_type =
     gcc_version.flavor == gcc_versiont::flavort::GCC &&
     gcc_version.is_at_least(4u, gcc_float128_minor_version);
+
+  config.ansi_c.c_standard = gcc_version.default_c_standard;
+  config.cpp.cpp_standard = gcc_version.default_cxx_standard;
 }
```
kroening: This is a bigger change in behavior; it may make sense to do so, but I'd take it out for now.

src/ansi-c/c_preprocess.cpp
chrisr-diffblue: Looks to me like this is a copy-n-paste of all the corresponding c++14, etc, condition blocks above this... And I'm suspecting that "OpenBSD" really means "is the platform using LLVM/clang by default" - which would also cover FreeBSD and similar...

chrisr-diffblue:
I'd suggest that the PR title should probably change - this PR doesn't, as far as I can tell, actually implement C++17 support - it adds some support for passing appropriate C++17 flags to the preprocessor, and detection of compiler C++17 modes. The PR title should probably make that clearer.

src/ansi-c/c_preprocess.cpp
kroening: Can we confirm that these platforms indeed default to -std=cpp++XX instead of gnu++XX?

lzaoral: Yes, that's just copy-pasted from the C++11/14 cases. The change @kroening refers to was made in b8116802b04a85813a38ff69f40df88a9d671ece as part of https://github.com/diffblue/cbmc/pull/4053 and the PR description says that:

> The c_preprocess.cpp source file in ansi-c has been patched to pass the appropriate -std to Clang on OpenBSD. OpenBSD does not use GNU extensions with the system Clang.

src/ansi-c/gcc_version.cpp
lzaoral: Thanks a lot for the remark, @kroening! This change was proposed by @tautschnig in https://github.com/diffblue/cbmc/pull/6144#issuecomment-847394738 so he will definitely know better as I'm not exactly sure what the real consequences of this change are.

However, I can confirm that even without this patch, #6090 seems to be fixed (at least on my Arch machine)! So I guess it's safe to drop this commit.

EDIT: typo

lzaoral:
Thanks @chrisr-diffblue! The PR title should be more appropriate now.

martin-cs (approved):
Looks good to me; are there any other blockers to this being merged?

tautschnig (approved):

PR #6412, opened 2021-10-22
introduce goto_programt::instructiont::type() and _nonconst()

kroening:
This introduces two methods to read and write the type of a GOTO program
instruction, and replaces various direct accesses.

### DIFF
--- a/jbmc/src/java_bytecode/remove_exceptions.cpp
+++ b/jbmc/src/java_bytecode/remove_exceptions.cpp
@@ -228,7 +228,7 @@ void remove_exceptionst::instrument_exception_handler(
   const goto_programt::targett &instr_it,
   bool may_catch)
 {
-  PRECONDITION(instr_it->type==CATCH);
+  PRECONDITION(instr_it->type() == CATCH);
 
   if(may_catch)
   {
@@ -404,7 +404,7 @@ bool remove_exceptionst::instrument_throw(
   const remove_exceptionst::stack_catcht &stack_catch,
   const std::vector<symbol_exprt> &locals)
 {
-  PRECONDITION(instr_it->type==THROW);
+  PRECONDITION(instr_it->type() == THROW);
 
   const exprt &exc_expr =
     uncaught_exceptions_domaint::get_exception_symbol(instr_it->get_code());
@@ -416,13 +416,11 @@ bool remove_exceptionst::instrument_throw(
   symbol_exprt exc_thrown =
     get_inflight_exception_global();
 
-  // add the assignment with the appropriate cast
-  code_assignt assignment(
+  // now turn the `throw' into an assignment with the appropriate cast
+  *instr_it = goto_programt::make_assignment(
     exc_thrown,
-    typecast_exprt(exc_expr, exc_thrown.type()));
-  // now turn the `throw' into `assignment'
-  instr_it->type=ASSIGN;
-  instr_it->code_nonconst() = assignment;
+    typecast_exprt(exc_expr, exc_thrown.type()),
+    instr_it->source_location());
 
   return true;
 }
@@ -437,7 +435,7 @@ remove_exceptionst::instrument_function_call(
   const stack_catcht &stack_catch,
   const std::vector<symbol_exprt> &locals)
 {
-  PRECONDITION(instr_it->type==FUNCTION_CALL);
+  PRECONDITION(instr_it->type() == FUNCTION_CALL);
 
   // save the address of the next instruction
   goto_programt::targett next_it=instr_it;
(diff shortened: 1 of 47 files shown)


src/goto-instrument/uninitialized.cpp
```diff
@@ -120,11 +120,11 @@ void uninitializedt::add_assertions(
           id2string(identifier)+"#initialized";
 
         symbol_exprt symbol_expr(new_identifier, bool_typet());
-        i1->type=DECL;
+        i1->type_nonconst() = DECL;
         i1->source_location_nonconst() = instruction.source_location();
         i1->code_nonconst() = code_declt(symbol_expr);
 
-        i2->type=ASSIGN;
+        i2->type_nonconst() = ASSIGN;
```
peterschrammel: use the 2-param variant of `insert_after`?

src/goto-instrument/uninitialized.cpp
```diff
@@ -120,11 +120,11 @@ void uninitializedt::add_assertions(
           id2string(identifier)+"#initialized";
 
         symbol_exprt symbol_expr(new_identifier, bool_typet());
-        i1->type=DECL;
+        i1->type_nonconst() = DECL;
```
peterschrammel: use the 2-param variant of `insert_after`?

jbmc/src/java_bytecode/remove_exceptions.cpp
```diff
@@ -421,7 +421,7 @@ bool remove_exceptionst::instrument_throw(
     exc_thrown,
     typecast_exprt(exc_expr, exc_thrown.type()));
   // now turn the `throw' into `assignment'
-  instr_it->type=ASSIGN;
+  instr_it->type_nonconst() = ASSIGN;
```
peterschrammel: swap?

src/goto-instrument/accelerate/disjunctive_polynomial_acceleration.cpp
```diff
@@ -846,7 +846,7 @@ void disjunctive_polynomial_accelerationt::build_fixed()
   {
     if(instruction.is_assert())
     {
-      instruction.type = ASSUME;
+      instruction.type_nonconst() = ASSUME;
```
peterschrammel: turn_into_assume?

peterschrammel (commented):
💡 I think by introducing `turn_into_assume` (we've already got `turn_into_skip`), we can do without `type_nonconst()`.

src/goto-programs/goto_program.h
```diff
@@ -340,8 +340,24 @@ class goto_programt
...
-    goto_program_instruction_typet type;
+    goto_program_instruction_typet type() const
+    {
+      return _type;
+    }
+
+    /// Set the kind of the instruction.
+    /// This method is best avoided to prevent mal-formed instructions.
+    /// Consider using the goto_programt::make_X methods instead.
+    goto_program_instruction_typet &type_nonconst()
```
martin-cs, line 351: As this only has 7 users, would it be better to just refactor those and not have this functionality at all?  Alternatively, could this be marked as deprecated?

martin-cs (approved):
Seems reasonable.

kroening:
I've added `turn_into_assume`, which removes some usage of `type_nonconst()`.

unit/goto-programs/structured_trace_util.cpp
```diff
@@ -56,7 +56,7 @@ TEST_CASE("structured_trace_util", "[core][util][trace]")
   add_instruction(loop_head_location, instructions);
   // 2: goto 1 # back_edge
   const auto back_edge = add_instruction(back_edge_location, instructions);
-  back_edge->type = GOTO;
+  back_edge->type_nonconst() = GOTO;
```
tautschnig, line 59: Can't we use `make_goto` for this one?

src/goto-programs/goto_program.h
tautschnig: Aren't we even down to 2 users, one of which can easily be fixed and the other being `read_bin_goto_object`? I think the latter could be handled by some low-level constructor that likely wants to be `private` with access only permitted via a `friend` declaration.

tautschnig (commented):
It would be great to figure out a story for the `nonconst` case right away.

unit/goto-programs/structured_trace_util.cpp
kroening: add_instruction() in that unit test should be replaced altogether, but this goes beyond this PR.

kroening:
Changing `read_bin_goto_object` is a major rejfactoring of that piece of code; much more involved than the initial scope of the PR. I'd do this separately.

tautschnig (approved):

peterschrammel (approved):

PR #6415, opened 2021-10-25
Separate file for Makefile based building of smt2_incremental directory

thomasspriggs:
So that the `.cpp` files for incremental SMT2 solving can be
added/removed/renamed without triggering codeowner review of the
`src/solvers` directory.

### DIFF
--- a/src/solvers/Makefile
+++ b/src/solvers/Makefile
@@ -68,6 +68,8 @@ ifneq ($(CADICAL),)
   CP_CXXFLAGS += -DHAVE_CADICAL
 endif
 
+include smt2_incremental/Makefile
+
 SRC = $(BOOLEFORCE_SRC) \
       $(CHAFF_SRC) \
       $(CUDD_SRC) \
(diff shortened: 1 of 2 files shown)


TGWDB (approved):

tautschnig:
If `CODEOWNERS` is what we're worried about, can't we just amend `CODEOWNERS` instead and list `src/solvers/Makefile` in there?

thomasspriggs:
> If `CODEOWNERS` is what we're worried about, can't we just amend `CODEOWNERS` instead and list `src/solvers/Makefile` in there?

That would also work. However it would grant code ownership to those listed against it for making build system changes for the other solvers. It would also mean keeping an additional list of people up to date in the `CODEOWNERS` file, rather than tying the list of people who can change what files are built for the `smt2_incremental` directory precisely to those who have code ownership of that directory.

tautschnig:
> That would also work. However it would grant code ownership to those listed against it for making build system changes for the other solvers. It would also mean keeping an additional list of people up to date in the CODEOWNERS file, rather than tying the list of people who can change what files are built for the smt2_incremental directory precisely to those who have code ownership of that directory.

I accept that there's some bureaucratic overhead in this, but then at least we're doing bureaucracy where bureaucracy belongs. Working around bureaucracy via a one-off Makefile hack is, well, a workaround. (And I'm not worried about granting code ownership to too many people on this occasion.)

An alternative fix is to introduce Makefiles for each subdirectory in `solvers` instead of this single big one. But then we should do this properly for all subdirectories.

thomasspriggs:
This has been superseded by the codeowners change PR which has been merged.

PR #6416, opened 2021-10-26
__CPROVER_r/w_ok does not require null or a valid pointer

### LINKED ISSUE #6472: Preconditions of cbmc library function memcpy are too strong
I got failure "memcpy source region readable" when I call ```memcpy(p,q,0)``` where ```q``` is an overflowing pointer (this happened in a c-common benchmark https://github.com/awslabs/aws-c-common/blob/main/source/array_list.c#L187). However, as the number of bytes we want to copy is 0, this call will not cause any memory safety problem. 

When I looked the cbmc library function ```memcpy```, I found that, even though the function body will be executed only when ```n``` (the number of bytes we want to copy) is greater than 0, the preconditions do not exclude the case of ```n==0```.
https://github.com/diffblue/cbmc/blob/d8296bdde10805b5f78ac3c6e61ccd86d388643e/src/ansi-c/library/string.c#L609
https://github.com/diffblue/cbmc/blob/d8296bdde10805b5f78ac3c6e61ccd86d388643e/src/ansi-c/library/string.c#L655

Modifying the preconditions to implications with premise ```n != 0``` can avoid such false-negative error. For example, the preconditions of ```memcpy``` can be following.
```c
  __CPROVER_precondition(
    n == 0 || (__CPROVER_POINTER_OBJECT(dst) != __CPROVER_POINTER_OBJECT(src) ||
      ((const char *)src >= (const char *)dst + n) ||
      ((const char *)dst >= (const char *)src + n)),
    "memcpy src/dst overlap");
  __CPROVER_precondition(
    n == 0 || (__CPROVER_r_ok(src, n)), "memcpy source region readable");
  __CPROVER_precondition(
    n == 0 || (__CPROVER_w_ok(dst, n)), "memcpy destination region writeable");
```

kroening:
The definition of `__CPROVER_r/w_ok` (in goto_check.cpp) does include the case
of an invalid pointer, and thus, there is no need to check that the pointer
given to these predicates is valid.


doc/architectural/memory-primitives.md
```diff
@@ -169,25 +169,28 @@ to pointers that point to within a memory object.
...
-The following two primitives can be used to check whether there is a memory
-segment starting at the given pointer and extending for at least the given
-number of bytes:
+The following three primitives can be used to check whether there is a
+memory segment starting at the given pointer and extending for at least the
+given number of bytes:
 
 - `_Bool __CPROVER_r_ok(const void *p, size_t size)`
 - `_Bool __CPROVER_w_ok(const void *p, size_t size)`
+- `_Bool __CPROVER_rw_ok(const void *p, size_t size)`
```
tautschnig, line 178: Why is this being removed from the documentation? (There's probably a point in completely removing it, but I don't understand why it's only being removed from the documentation.)

tautschnig (commented):
Looks good to me, except I'd like to understand the `rw_ok` removal from documentation.

doc/architectural/memory-primitives.md
```diff
@@ -169,25 +169,28 @@ to pointers that point to within a memory object.
...
 
-At present, both primitives are equivalent as all memory in CBMC is considered
+At present, all three primitives are equivalent as all memory in CBMC is considered
 both readable and writeable. If `p` is the null pointer, the primitives return
 false. If `p` is valid, the primitives return true if the memory segment
```
feliperodri: If `size` is zero, shouldn't this primitive return `true` even if the pointer is `NULL` or `invalid`? cc. @SaswatPadhi

feliperodri:
cc. @jimgrundy @markrtuttle @remi-delmas-3000

doc/architectural/memory-primitives.md
SaswatPadhi: +1. Intuitively, reading / writing "nothing" to any memory location could always (vacuously) succeed.

It's a design decision, however, and so we could choose to not do it this way and always for validity.
But currently there seems to be an issue with checking validity of a `malloc(0)` result. The standard says that it's not supposed to be deref'ed, so I am not exactly sure what "valid" would mean for the returned pointer.

```c
char *p = malloc(0);
assert(__CPROVER_r_ok(p, 0)); // fails with --pointer-primitive-check; passes otherwise
```

@feliperodri, correct me if I am wrong here, but this is what we observed, right?

doc/architectural/memory-primitives.md
kroening: The commit doesn't remove it? Note that the "-" is not indicative of a removed line here.

doc/architectural/memory-primitives.md
kroening: I don't think we should attempt to give meaning to "zero width memory accesses".

jimgrundy: I am perfectly happy for "__CPROVER_r_ok(p,0))" to fail if p is NULL or invalid.

But, I would want to dig a little into this example by looking at two cases (recall that an implementation of malloc is allowed to return a non-null value for malloc(0), though it may also return NULL):
```c
char *p = malloc(0);
if (p == NULL) {
    assert(__CPROVER_r_ok(p,0));
} else {
    assert(__CPROVER_r_ok(p,0));
}
```
Consider following cases:

1. First assert & pointer-primitive-checks enabled:
> I'm OK with any outcome here: fail with pointer primitive check, fails because __CPROVER_r_ok returned false, or passes. I think Daniel would like this to fail with a pointer primitive check, and I can get on board with that.
2. First assert & pointer-primtive-checks disabled:
> I'm OK with any outcome here: fails because __CPROVER_r_ok returned false, or passes
3. Second assert & pointer-primtive-checks enabled:
> I'm of the opinion that this must pass
4. Second assert & pointer-primtive-checks disabled:
> I'm of the opinion that this must pass

If cases 3 and 4 were to fail we're going to have to write a lot more special cases in our assertions than we need to handle degenerate cases where functions are called with size 0. Those special cases shouldn't be needed.

For case 1 - I think Daniel's desire to not give meaning to 0 size regions argues for keeping the expectation that this fails the pointer-primitive-check if that is enabled.

For case 2 - I'm not sure I care, but no matter whether you return true or false you are picking a meaning for a zero sized access to a NULL pointer.

jimgrundy: I suppose I'm happy with undefined behavior when size is 0... because I can always defined my own version that is __jim_r_ok(p,size) = (size == 0) || __CPROVER_r_ok(p, size);
Mind you if I wanted something like __jim_r_ok that would fail if p was invalid even if size was 0 and pointer-checks were in abled I guess I would be out of luck. I can probably live with that too.

kroening: You can get the version that fails with size 0 simply like this:
`(size == 0) ? 0 : __CPROVER_r_ok(p, size)`

kroening:
Ok, let's do the following: `__CPROVER_r/w_ok(p, s)` is true when `s` is zero and `p` points either into the object, or at the byte that's just one beyond the object.

This implies that it's true when `s` is zero and `p` has offset zero and the object has size zero.

doc/architectural/memory-primitives.md
jimgrundy: I meant when called with the -pointer-checks enabled would fail (not return false) with a pointer check error if called with __jim_r_ok(invalid-pointer,0).

jimgrundy:
I like very much where we ended up on this. Thanks Daniel.

doc/architectural/memory-primitives.md
kroening: Ah, apologies; I misunderstood.

feliperodri (approved):

doc/architectural/memory-primitives.md
tautschnig: My apologies for misreading this! Could the commit please be split up and the commit message(s) be clear in what is actually being changed?

tautschnig (dismissed):

tautschnig (changes requested):
The following example might serve to demonstrate why, in my opinion, this check should not be removed:
````c
#include <stdlib.h>

int main()
{
  int *p = malloc(sizeof(int));
  free(p);
  __CPROVER_assume(__CPROVER_r_ok(p, sizeof(int)));
  __CPROVER_assert(0, "should be unreachable");
}
````
CBMC wrongly reports that this assertion is reachable (but `--pointer-primitive-check` calls out the problem in the assumption).

jimgrundy:
> 

Wouldn't I want __CPROVER_r_ok(p, sizeof(int)) in this case (p has just been free-ed) to return false rather than error out?

martin-cs:
@tautschnig happy to review this when / if you are happy with the requested changes.

martin-cs:
@tautschnig as https://github.com/diffblue/cbmc/pull/7395 is now merged, does this need to be revisited?

PR #6419, opened 2021-10-28
Remove goto_programt::instructiont::type_nonconst

tautschnig:
Use of this method was never advised, and the remaining two uses can be
rewritten to use more appropriate (and modern) APIs.

### DIFF
--- a/src/goto-programs/read_bin_goto_object.cpp
+++ b/src/goto-programs/read_bin_goto_object.cpp
@@ -97,17 +97,20 @@ static bool read_bin_goto_object(
     for(std::size_t ins_index = 0; ins_index < ins_count; ++ins_index)
     {
       goto_programt::targett itarget = f.body.add_instruction();
-      goto_programt::instructiont &instruction=*itarget;
 
-      instruction.code_nonconst() =
+      // take copies as references into irepconverter are not stable
+      codet code =
         static_cast<const codet &>(irepconverter.reference_convert(in));
-      instruction.source_location_nonconst() =
-        static_cast<const source_locationt &>(
-          irepconverter.reference_convert(in));
-      instruction.type_nonconst() =
+      source_locationt source_location = static_cast<const source_locationt &>(
+        irepconverter.reference_convert(in));
+      goto_program_instruction_typet instruction_type =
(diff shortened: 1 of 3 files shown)


src/goto-programs/read_bin_goto_object.cpp
```diff
@@ -97,17 +97,15 @@ static bool read_bin_goto_object(
...
-          irepconverter.reference_convert(in));
-      instruction.type_nonconst() =
-        (goto_program_instruction_typet)irepconverter.read_gb_word(in);
-      instruction.guard =
-        static_cast<const exprt &>(irepconverter.reference_convert(in));
+          irepconverter.reference_convert(in)),
+        (goto_program_instruction_typet)irepconverter.read_gb_word(in),
+        static_cast<const exprt &>(irepconverter.reference_convert(in)),
+        {}};
+
```
kroening, line 113: The order of evaluation of function arguments is unspecified in C++ -- i.e., this may read the entries of the binary in the wrong order.

unit/goto-programs/structured_trace_util.cpp
```diff
@@ -53,20 +47,18 @@ TEST_CASE("structured_trace_util", "[core][util][trace]")
...
   add_instruction(basic_location, instructions);
   // 1 # loop_head
-  add_instruction(loop_head_location, instructions);
+  auto loop_head = add_instruction(loop_head_location, instructions);
   // 2: goto 1 # back_edge
-  const auto back_edge = add_instruction(back_edge_location, instructions);
-  back_edge->type_nonconst() = GOTO;
+  instructions.emplace_back(
+    goto_programt::make_goto(loop_head, back_edge_location));
+  loop_head->incoming_edges.insert(--instructions.end());
```
kroening: I'd use `std::prev` here, given that the side effect of the predecrement operator is not used.

tautschnig: Done.

src/goto-programs/read_bin_goto_object.cpp
tautschnig: Fixed!

kroening (approved):

PR #6420, opened 2021-10-29
Prepare DECL/DEAD construction with location

tautschnig:
Add new constructors to code_{decl,dead}t that also take a location.
This will help avoid some *_nonconst() calls. To facilitate this,
an additional make_dead instruction builder is also needed (which just
mirrors what we already have for other instruction types).

### DIFF
--- a/src/goto-programs/goto_instruction_code.h
+++ b/src/goto-programs/goto_instruction_code.h
@@ -129,6 +129,11 @@ public:
   {
   }
 
+  code_deadt(symbol_exprt symbol, source_locationt loc)
+    : codet(ID_dead, {std::move(symbol)}, std::move(loc))
+  {
+  }
+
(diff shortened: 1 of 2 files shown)


kroening:
This has a subtle problem: the location isn't supposed to go into the `_code` member any more, but into the `_source_location` member.

tautschnig:
> This has a subtle problem: the location isn't supposed to go into the `_code` member any more, but into the `_source_location` member.

I suppose we need to document this somewhere? Perhaps, however, this PR makes for a simpler transition as all those places can be spotted more easily?

kroening:
Sure, simply fix those two new constructors, and it's good to go!

tautschnig:
Closing as the new constructors shouldn't be necessary as indicated in the discussion.

PR #6424, opened 2021-11-01
Fix printing of `smt_get_value_commandt`

thomasspriggs:
This PR fixes the printing of `smt_get_value_commandt`. The get value command is supposed to be parameterised with a collection of one or more terms for which the solver should return the values. Our internal representation of the get-value command only supports a single term, which should be placed in brackets to make it into a collection containing a single term. The brackets were previously missing, which would have caused a solver error.


TGWDB (approved):
I recognise in cbmc we only ever ask for one value at a time, maybe we should note that the standard allows multiple in the code in case someone wants to change this one day?

PR #6426, opened 2021-11-01
Include function and parameter in `convert_function` error messages

adpaco-aws:
Include the name of the function and parameter causing `DATA_INVARIANT` to fail in the error messages. This saves time spent in debugging which function and parameter are problematic.

Fixes a couple of mistakes I spotted in the section titles in `COMPILING.md`.

### DIFF
--- a/src/goto-programs/goto_convert_functions.cpp
+++ b/src/goto-programs/goto_convert_functions.cpp
@@ -166,10 +166,19 @@ void goto_convert_functionst::convert_function(
   // we have a body, make sure all parameter names are valid
   for(const auto &p : f.parameter_identifiers)
   {
-    DATA_INVARIANT(!p.empty(), "parameter identifier should not be empty");
-    DATA_INVARIANT(
+    DATA_INVARIANT_WITH_DIAGNOSTICS(
+      !p.empty(),
+      "parameter identifier should not be empty",
+      "function:",
+      identifier);
+
+    DATA_INVARIANT_WITH_DIAGNOSTICS(
       symbol_table.has_symbol(p),
-      "parameter identifier must be a known symbol");
+      "parameter identifier must be a known symbol",
+      "function:",
+      identifier,
+      "parameter:",
+      p);
   }
 
   lifetimet parent_lifetime = lifetime;
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -1,4 +1,4 @@
-#WHAT ARCHITECTURE ?
+# What architecture?
 
 CPROVER now needs a C++11 compliant compiler and is known to work in the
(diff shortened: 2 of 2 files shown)


src/goto-programs/goto_convert_functions.cpp
```diff
@@ -163,13 +163,36 @@ void goto_convert_functionst::convert_function(
...
+               "parameter identifier must be a known symbol\n"
+               "function: %s\n"
+               "parameter: %s",
+               identifier.c_str(),
+               p.c_str());
+    }
     DATA_INVARIANT(
       symbol_table.has_symbol(p),
-      "parameter identifier must be a known symbol");
+      parameter_error_message);
```
TGWDB: I'm not sure that defining and constructing all of this outside the invariant is the right choice. There is also `DATA_INVARIANT_WITH_DIAGNOSTICS` that handles other arguments beyond the message.

adpaco-aws:
Hi @TGWDB , thanks for your suggestion!

I have changed the code to use `DATA_INVARIANT_WITH_DIAGNOSTICS` and print a similar message. Let me know if you see any other issues.

src/goto-programs/goto_convert_functions.cpp
martin-cs: Yes; what @TGWDB says.

martin-cs (approved):
Yes; looks good.

Please forgive the pedantry but in future it would be easier if this was two separate PRs.  That way each PR does / fixes one thing.  I realise this is hassle.

TGWDB (approved):

src/goto-programs/goto_convert_functions.cpp
```diff
@@ -166,10 +166,21 @@ void goto_convert_functionst::convert_function(
...
-    DATA_INVARIANT(!p.empty(), "parameter identifier should not be empty");
-    DATA_INVARIANT(
+    DATA_INVARIANT_WITH_DIAGNOSTICS(
+      !p.empty(),
+      "parameter identifier should not be empty",
+      "function:",
+      identifier,
+      "parameter:",
+      p);
+
```
kroening, line 174: That doesn't make sense: when the message is triggered, `p` is empty. Thus, why log it?

adpaco-aws: Right. Removed `p` from this case.

tautschnig (approved):
@adpaco-aws Could you please squash the three commits pertaining to the diagnostic message into a single one?

adpaco-aws:
@tautschnig just squashed them and rebased against develop, let me know if you see any issues.

PR #6430, opened 2021-11-02
Add new builtin __CPROVER_havoc_slice

### LINKED ISSUE #6351: Create new __CPROVER_havoc_slice & __CPROVER_havoc_from primitives
We need the following new havoc primitives in CBMC:

```
void __CPROVER_havoc_slice(void *p, size_t s);
void __CPROVER_havoc_from (void *p);
```
(Reference implementation for havoc_object: [symex_other.cpp](https://github.com/diffblue/cbmc/blob/56adbb21ea443f07d29b6056016cbc86da6c7dec/src/goto-symex/symex_other.cpp#L244))

The semantics of these primitives are described below.

***__CPROVER_havoc_slice(void \*p, size_t s)***

_Precondition:_

```
s == 0 || __CPROVER_w_ok(p, s)
```

- If s is zero then no action is performed.
- If s is non-zero and p is a valid pointer to an object with at least s writable bytes starting at address p, that is if __CPROVER_w_ok(p, s),  then the effect is to replace the s bytes starting at address p with fresh symbolic values.

***__CPROVER_havoc_from(void \*p)***

_Precondition:_

```
p == NULL ||
     __CPROVER_OBJECT_SIZE(p) == __CPROVER_POINTER_OFFSET(p) ||
     __CPROVER_w_ok(p, __CPROVER_OBJECT_SIZE(p) - __CPROVER_POINTER_OFFSET(p))
```

- If p is NULL then no action is performed.
- Otherwise, equivalent to:
    `__CPROVER_havoc_slice(p, __CPROVER_OBJECT_SIZE(p) - __CPROVER_POINTER_OFFSET(p))`

remi-delmas-3000:
Resolves #6351.

This PR adds a new builtin function :

```C
void __CPROVER_havoc_slice(void *p, __CPROVER_size_t size);
```
Which lets you update `size` consecutive bytes to nondet values in the object underlying a given pointer `p`. It requires that `__CPROVER_w_ok(p, size)` holds.
It is rewritten at goto level to an `array_replace(p, q)` operation where `q` is declared a nondet array of size `size`.

### DIFF
--- a/doc/cprover-manual/api.md
+++ b/doc/cprover-manual/api.md
@@ -211,6 +211,57 @@ returns true when it is safe to do both.  These predicates can be given an
 optional size; when the size argument is not given, the size of the subtype
 (which must not be **void**) of the pointer type is used.
 
+#### \_\_CPROVER\_havoc\_object
+
+
+This function requires a valid pointer and updates **all bytes** of the
+underlying object with nondeterministic values.
+
+```C
+void __CPROVER_havoc_object(void *p);
+```
+
+**Warning**
+
+This primitive havocs object bytes before
+the given `p` and after `p + sizeof(*p)`:
+
+```C
+struct foo {
+  int x;
+  int y;
+  int z;
+};
+
+struct foo thefoo = {.x = 1; .y = 2, .z = 3};
+
+int* p = &thefoo.y; // pointing to thefoo.y
+
+__CPROVER_havoc_object(p); // makes the whole struct nondet
+__CPROVER_assert(thefoo.x == 1, "fails because `thefoo.x` is now nondet");
+__CPROVER_assert(thefoo.y == 2, "fails because `thefoo.y` is now nondet");
+__CPROVER_assert(thefoo.z == 3, "fails because `thefoo.z` is now nondet");
+```
+
+#### \_\_CPROVER\_havoc\_slice
+
+This function requires requires that `__CPROVER_w_ok(p, size)` holds,
+and updates `size` consecutive bytes of the underlying object, starting at `p`,
+with nondeterministic values.
+
+```C
+void __CPROVER_havoc_slice(void *p, __CPROVER_size_t size);
+```
+
+**Caveat**
+
+- If the slice contains bytes that can be interpreted as pointers by the
+  program, this will cause these pointers to become invalid
+  (i.e. they will not point to anything meaningful).
+- If this slice only contains bytes that are not interpreted as pointers
+  by the program, then havocing the slice is equivalent to making the
+  interpretation of these bytes nondeterministic.
+
 ### Predefined Types and Symbols
 
 #### \_\_CPROVER\_bitvector
--- a/regression/cbmc/havoc_slice/declarations.h
+++ b/regression/cbmc/havoc_slice/declarations.h
@@ -1,0 +1,26 @@
+#include <stdint.h>
+#include <stdlib.h>
+
+// HAVOC STRUCT MEMBERS
+// TODO take into account plaform/pointer size in the tests
+typedef struct
+{
+  uint16_t a;    // 2 bytes
+  uint16_t b[5]; // 10 bytes
+  uint32_t c;    // 4 bytes
+  uint16_t *d;   // 4 or 8 bytes
+  union {
+    uint16_t a;    // 2 bytes
+    uint16_t b[5]; // 10 bytes
+    uint32_t c;    // 4 bytes
+    uint16_t *d;   // 4 or 8 bytes
+  } u;             // 10 bytes
+} st;              // 30 or 34 bytes total
+
+//                0  2          12   16       24       34
+//                |  |          |    |        |        |
+// struct layout: aa bbbbbbbbbb cccc dddddddd uuuuuuuuuu
+// union layout :                             aa--------
+//                                            bbbbbbbbbb
+//                                            cccc------
+//                                            dddddddd--
--- a/regression/cbmc/havoc_slice/test_whole_array_with_offset.desc
+++ b/regression/cbmc/havoc_slice/test_whole_array_with_offset.desc
@@ -1,0 +1,10 @@
+CORE
+test_whole_array_with_offset.c
+
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
--- a/regression/cbmc/havoc_slice_checks/main.c
+++ b/regression/cbmc/havoc_slice_checks/main.c
@@ -1,0 +1,31 @@
+// Here we test that __CPROVER_havoc_slice checks fail as expected
+#include <stdint.h>
+#include <stdlib.h>
+
+int main()
+{
+  // invalid pointer
+  int *invalid_ptr;
+  __CPROVER_havoc_slice(invalid_ptr, 1);
+
+  // null pointer
+  int *null_ptr = NULL;
+  __CPROVER_havoc_slice(null_ptr, 1);
+
+  int a[4];
+
+  // positive size but out of bounds
+  __CPROVER_havoc_slice(&a[2], 3 * sizeof(*a));
+
+  // positive size but out of bounds
+  __CPROVER_havoc_slice(a, 5 * sizeof(*a));
+
+  // pointer invalidated by overflow
+  __CPROVER_havoc_slice(&a[0] + 6, sizeof(*a));
+
+  // pointer invalidated by underflow
+  __CPROVER_havoc_slice(&a[0] - 1, sizeof(*a));
+
+  // negative size
+  __CPROVER_havoc_slice(&a[0], -2);
+}
--- a/src/goto-programs/builtin_functions.cpp
+++ b/src/goto-programs/builtin_functions.cpp
@@ -636,12 +636,92 @@ void goto_convertt::do_enum_is_in_range(
   copy(assignment, ASSIGN, dest);
 }
 
+void goto_convertt::do_havoc_slice(
+  const exprt &lhs,
+  const symbol_exprt &function,
+  const exprt::operandst &arguments,
+  goto_programt &dest,
+  const irep_idt &mode)
+{
+  irep_idt identifier = CPROVER_PREFIX "havoc_slice";
+
+  // We disable checks on the generated instructions
+  // because we add our own rw_ok assertion that takes size into account
+  auto source_location = function.find_source_location();
+  source_location.add_pragma("disable:pointer-check");
+  source_location.add_pragma("disable:pointer-overflow-check");
+  source_location.add_pragma("disable:pointer-primitive-check");
+
+  // check # arguments
+  if(arguments.size() != 2)
+  {
+    error().source_location = source_location;
+    error() << "'" << identifier << "' expected to have two arguments" << eom;
+    throw 0;
+  }
+
+  // check argument types
+  if(arguments[0].type().id() != ID_pointer)
+  {
+    error().source_location = source_location;
+    error() << "'" << identifier
+            << "' first argument expected to have `void *` type" << eom;
+    throw 0;
+  }
+
+  if(arguments[1].type().id() != ID_unsignedbv)
+  {
+    error().source_location = source_location;
+    error() << "'" << identifier
+            << "' second argument expected to have `size_t` type" << eom;
+    throw 0;
+  }
+
+  // check nil lhs
+  if(lhs.is_not_nil())
+  {
+    error().source_location = source_location;
+    error() << "'" << identifier << "' not expected to have a LHS" << eom;
+    throw 0;
+  }
+
+  // insert instructions
+  // assert(rw_ok(argument[0], argument[1]));
+  // char nondet_contents[argument[1]];
+  // __CPROVER_array_replace(p, nondet_contents);
+
+  r_or_w_ok_exprt ok_expr(ID_w_ok, arguments[0], arguments[1]);
+  ok_expr.add_source_location() = source_location;
+  goto_programt::targett t =
+    dest.add(goto_programt::make_assertion(ok_expr, source_location));
+  t->source_location_nonconst().set("user-provided", false);
+  t->source_location_nonconst().set_property_class(ID_assertion);
+  t->source_location_nonconst().set_comment(
+    "assertion havoc_slice " + from_expr(ns, identifier, ok_expr));
+
+  const array_typet array_type(char_type(), arguments[1]);
+
+  const symbolt &nondet_contents =
+    new_tmp_symbol(array_type, "nondet_contents", dest, source_location, mode);
+  const exprt &nondet_contents_expr = address_of_exprt{
+    index_exprt{nondet_contents.symbol_expr(), from_integer(0, index_type())}};
+
+  const exprt &arg0 =
+    typecast_exprt::conditional_cast(arguments[0], pointer_type(empty_typet{}));
+  const exprt &arg1 = typecast_exprt::conditional_cast(
+    nondet_contents_expr, pointer_type(empty_typet{}));
+
+  codet array_replace(ID_array_replace, {arg0, arg1}, source_location);
+  dest.add(goto_programt::make_other(array_replace, source_location));
+}
+
 /// add function calls to function queue for later processing
 void goto_convertt::do_function_call_symbol(
   const exprt &lhs,
   const symbol_exprt &function,
   const exprt::operandst &arguments,
-  goto_programt &dest)
+  goto_programt &dest,
+  const irep_idt &mode)
 {
   if(function.get_bool(ID_C_invalid_object))
     return; // ignore
@@ -681,8 +761,12 @@ void goto_convertt::do_function_call_symbol(
     return;
   }
 
-  if(identifier==CPROVER_PREFIX "assume" ||
-     identifier=="__VERIFIER_assume")
+  if(identifier == CPROVER_PREFIX "havoc_slice")
+  {
+    do_havoc_slice(lhs, function, arguments, dest, mode);
+  }
+  else if(
+    identifier == CPROVER_PREFIX "assume" || identifier == "__VERIFIER_assume")
   {
     if(arguments.size()!=1)
     {
--- a/src/goto-programs/goto_convert_class.h
+++ b/src/goto-programs/goto_convert_class.h
@@ -203,7 +203,8 @@ protected:
     const exprt &lhs,
     const symbol_exprt &function,
     const exprt::operandst &arguments,
-    goto_programt &dest);
+    goto_programt &dest,
+    const irep_idt &mode);
 
   virtual void do_function_call_symbol(const symbolt &)
   {
@@ -672,6 +673,12 @@ protected:
     const symbol_exprt &function,
     const exprt::operandst &arguments,
     goto_programt &dest);
+  void do_havoc_slice(
+    const exprt &lhs,
+    const symbol_exprt &function,
+    const exprt::operandst &arguments,
+    goto_programt &dest,
+    const irep_idt &mode);
 
   exprt get_array_argument(const exprt &src);
 };
--- a/src/goto-programs/goto_instruction_code.cpp
+++ b/src/goto-programs/goto_instruction_code.cpp
@@ -13,7 +13,10 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <util/arith_tools.h>
 #include <util/c_types.h>
+#include <util/namespace.h>
+#include <util/std_expr.h>
 #include <util/string_constant.h>
+#include <util/symbol_table_base.h>
 
 code_inputt::code_inputt(
   std::vector<exprt> arguments,
@@ -69,4 +72,14 @@ void code_outputt::check(const codet &code, const validation_modet vm)
 {
   DATA_CHECK(
     vm, code.operands().size() >= 2, "output must have at least two operands");
+}
+
+inline code_function_callt
+havoc_slice_call(const exprt &p, const exprt &size, const namespacet &ns)
+{
+  irep_idt identifier = CPROVER_PREFIX "havoc_slice";
+  symbol_exprt havoc_slice_function = ns.lookup(identifier).symbol_expr();
+  code_function_callt::argumentst arguments = {p, size};
+  return code_function_callt{std::move(havoc_slice_function),
+                             std::move(arguments)};
 }
--- a/src/goto-programs/goto_instruction_code.h
+++ b/src/goto-programs/goto_instruction_code.h
@@ -9,6 +9,7 @@ Author: Daniel Kroening, dkr@amazon.com
 #ifndef CPROVER_UTIL_GOTO_INSTRUCTION_CODE_H
 #define CPROVER_UTIL_GOTO_INSTRUCTION_CODE_H
 
+#include <util/cprover_prefix.h>
 #include <util/std_code_base.h>
 #include <util/std_expr.h>
 
@@ -536,5 +537,19 @@ inline code_returnt &to_code_return(codet &code)
   code_returnt::check(code);
   return static_cast<code_returnt &>(code);
 }
+
+/// \brief Builds a \ref code_function_callt
+/// to `__CPROVER_havoc_slice(p, size)`.
+///
+/// \param p The pointer argument.
+/// \param size The size argument.
+/// \param ns Namespace where the `__CPROVER_havoc_slice symbol` can be found.
+/// \remarks: It is a PRECONDITION that `__CPROVER_havoc_slice` exists
+///   in the namespace
+///
+/// \return A \ref code_function_callt expression
+///         `nil_exprt() := __CPROVER_havoc_slice(p, size)`.
+inline code_function_callt
+havoc_slice_call(const exprt &p, const exprt &size, const namespacet &ns);
 
 #endif // CPROVER_GOTO_PROGRAMS_GOTO_INSTRUCTION_CODE_H
--- a/regression/cbmc/havoc_slice/test_array_slice_1.c
+++ b/regression/cbmc/havoc_slice/test_array_slice_1.c
@@ -1,0 +1,17 @@
+void main(void)
+{
+  // INITIALIZE
+  int a[5] = {0, 1, 2, 3, 4};
+  int old_a[5] = {0, 1, 2, 3, 4};
+
+  // HAVOC ARRAY SLICE
+  __CPROVER_havoc_slice(a + 2, 2 * sizeof(*a));
+
+  // POSTCONDITIONS
+  __CPROVER_assert(a[0] == old_a[0], "expecting SUCCESS");
+  __CPROVER_assert(a[1] == old_a[1], "expecting SUCCESS");
+  __CPROVER_assert(a[2] == old_a[2], "expecting FAILURE");
+  __CPROVER_assert(a[3] == old_a[3], "expecting FAILURE");
+  __CPROVER_assert(a[4] == old_a[4], "expecting SUCCESS");
+  return;
+}
--- a/regression/cbmc/havoc_slice/test_array_slice_1.desc
+++ b/regression/cbmc/havoc_slice/test_array_slice_1.desc
@@ -1,0 +1,10 @@
+CORE
+test_array_slice_1.c
+
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
--- a/regression/cbmc/havoc_slice/test_array_slice_2.c
+++ b/regression/cbmc/havoc_slice/test_array_slice_2.c
@@ -1,0 +1,17 @@
+void main(void)
+{
+  // INITIALIZE
+  int a[5] = {0, 1, 2, 3, 4};
+  int old_a[5] = {0, 1, 2, 3, 4};
+
+  // HAVOC ARRAY SLICE
+  __CPROVER_havoc_slice(&a[2] - 1, 2 * sizeof(*a));
+
+  // POSTCONDITIONS
+  __CPROVER_assert(a[0] == old_a[0], "expecting SUCCESS");
+  __CPROVER_assert(a[1] == old_a[1], "expecting FAILURE");
+  __CPROVER_assert(a[2] == old_a[2], "expecting FAILURE");
+  __CPROVER_assert(a[3] == old_a[3], "expecting SUCCESS");
+  __CPROVER_assert(a[4] == old_a[4], "expecting SUCCESS");
+  return;
+}
--- a/regression/cbmc/havoc_slice/test_array_slice_2.desc
+++ b/regression/cbmc/havoc_slice/test_array_slice_2.desc
@@ -1,0 +1,10 @@
+CORE
+test_array_slice_2.c
+
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
--- a/regression/cbmc/havoc_slice/test_array_symbolic_size.c
+++ b/regression/cbmc/havoc_slice/test_array_symbolic_size.c
@@ -1,0 +1,46 @@
+#include <stdint.h>
+#include <stdlib.h>
+
+// byte-level comparison
+#define CHECK_SLICE(A, OLD_A, I, IDX, HAVOC_SIZE)                              \
+  {                                                                            \
+    __CPROVER_assert(                                                          \
+      !(IDX <= I && I <= IDX + HAVOC_SIZE) | A[I] == OLD_A[I],                 \
+      "expecting FAILURE");                                                    \
+                                                                               \
+    __CPROVER_assert(                                                          \
+      (IDX <= I && I <= IDX + HAVOC_SIZE) | A[I] == OLD_A[I],                  \
+      "expecting SUCCESS");                                                    \
+  }
+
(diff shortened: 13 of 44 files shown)


src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,81 @@ void goto_convertt::do_enum_is_in_range(
...
+  // check argument types
+  if(arguments[0].type().id() != ID_pointer)
+  {
+    error().source_location = source_loc;
+    error() << "'" << CPROVER_PREFIX "havoc_object_slice"
+            << "' first argument expected to have `void *` type" << eom;
+    throw 0;
+  }
+
+  if(arguments[1].type().id() != ID_unsignedbv)
```
SaswatPadhi, line 672: To improve readability, could we rewrite this as:

```suggestion
  if(arguments[1].type().id() != size_type().id())
```

regression/cbmc/havoc_object_slice/test.desc
```diff
@@ -0,0 +1,11 @@
...
+main.c
+--pointer-check --pointer-primitive-check
+^EXIT=10$
+^SIGNAL=0$
+^\*\* 2960 of \d+ failed.*
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
```
SaswatPadhi: missing newline

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,80 @@ void goto_convertt::do_enum_is_in_range(
...
+  const exprt &lhs,
+  const symbol_exprt &function,
+  const exprt::operandst &arguments,
+  goto_programt &dest)
+{
+  std::string identifier = CPROVER_PREFIX "havoc_object_slice";
+
+  auto source_location = function.find_source_location();
+  source_location.add_pragma("disable:pointer-check");
+  source_location.add_pragma("disable:pointer-primitive-check");
```
SaswatPadhi, line 653: Do we also need:

```cpp
   source_loc.add_pragma("disable:pointer-overflow-check");
```

doc/cprover-manual/api.md
```diff
@@ -211,6 +211,20 @@ returns true when it is safe to do both.  These predicates can be given an
 optional size; when the size argument is not given, the size of the subtype
 (which must not be **void**) of the pointer type is used.
 
+#### \_\_CPROVER\_havoc\_object, \_\_CPROVER\_havoc\_object\_slice
+
+
+This function requires a valid pointer and updates all the bytes of the underlying object with nondeterministic values.
+```C
+void __CPROVER_havoc_object(void *p);
```
danielsn, line 221: When you say "object", what do you mean?
For example, should this assertion pass or fail?
```C
struct foo {
  int x;
  int y;
};

struct foo thefoo = {.x = 1; .y = 12};
int* p = &thefoo.y;
__CPROVER_havoc_object(p);
assert(thefoo.x == 1);
```

remi-delmas-3000: the object in that case is the blob of memory that holds the whole struct `thefoo` so this assertion should fail.
on the other it passes when using `__CPROVER_havoc_object_slice(p, sizeof(*p))`

jimgrundy (approved):

regression/cbmc/havoc_object_slice/main.c
```diff
@@ -0,0 +1,579 @@
...
+    else
+    {
+      __CPROVER_assert(i == old_i, "value unchanged expecting SUCCESS");
+      HV32(s.i);
+      __CPROVER_assert(s.j == old_s.j, "value unchanged expecting SUCCESS");
+    }
+  }
+
+  return 0;
+}
```
SaswatPadhi: missing newline

src/goto-programs/builtin_functions.cpp
SaswatPadhi: If not, can we document (in a comment above these pragmas) what checks we are disabling and why?

doc/cprover-manual/api.md
danielsn: This seems like it could lead to a foot gun:

```C
int foo_stub(int* p) {
   __CPROVER_havoc_object(p);
   return *p;
}

struct foo { int x; int y; };

int main() [
  struct foo thefoo = {.x = 1; .y = 12};
  int v = foo_stub(&thefoo.y);
  assert(thefoo.x == 1);
  assert(thefoo.y == v);
}
```

I think the user would expect that foo_stub wouldn't modify memory which it doesn't have a valid pointer to.  Maybe we need a different name: 

```
__CPROVER_havoc_entire_allocation()
```

remi-delmas-3000: Agreed, we could name `__CPROVER_havoc_object` differently but it was introduced some time ago so this could be a breaking change for existing users of the primitive.

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: The idea with this check was to allow users that use the primitive directly from GOTO to use any `unsignedbv` type and not just `size_t` to specify sizes. This makes it more general than its C interface.

regression/cbmc/havoc_object_slice/declarations.h
```diff
@@ -0,0 +1,26 @@
+#include <stdint.h>
+#include <stdlib.h>
+
+// HAVOC STRUCT MEMBERS
+// TODO take into account plaform/pointer size in the tests
```
feliperodri: Create a GitHub issue to track this request and add it to the comment, so it doesn't get lost.

regression/cbmc/havoc_object_slice/test_array_slice_1.c
```diff
@@ -0,0 +1,17 @@
...
+  __CPROVER_havoc_object_slice(a + 2, 2 * sizeof(*a));
+
+  // POSTCONDITIONS
+  __CPROVER_assert(a[0] == old_a[0], "expecting SUCCESS");
+  __CPROVER_assert(a[1] == old_a[1], "expecting SUCCESS");
+  __CPROVER_assert(a[2] == old_a[2], "expecting FAILURE");
+  __CPROVER_assert(a[3] == old_a[3], "expecting FAILURE");
+  __CPROVER_assert(a[4] == old_a[4], "expecting SUCCESS");
+  return;
+}
```
feliperodri: Add newline.

regression/cbmc/havoc_object_slice/declarations.h
feliperodri: If we're not using this file, perhaps add it to the GitHub issue and delete from here instead.

regression/cbmc/havoc_object_slice/test_array_slice_1.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test_array_slice_1.c
+--pointer-check --pointer-primitive-check
```
feliperodri: Why do we need `--pointer-primitive-check` here?

regression/cbmc/havoc_object_slice/test_array_slice_1.desc
```diff
@@ -0,0 +1,10 @@
...
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
```
feliperodri: Add newline.

regression/cbmc/havoc_object_slice/test_array_slice_2.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test_array_slice_2.c
+--pointer-check --pointer-primitive-check
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
```
feliperodri: Could you add which assertions are expected to fail?

feliperodri: Same goes to all other test cases.

regression/cbmc/havoc_object_slice/test_array_slice_2.desc
```diff
@@ -0,0 +1,10 @@
...
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
```
feliperodri: Add newline.

regression/cbmc/havoc_object_slice/test_array_symbolic_size.c
```diff
@@ -0,0 +1,46 @@
...
+#define CHECK_SLICE(A, OLD_A, I, IDX, HAVOC_SIZE)                              \
+  {                                                                            \
+    __CPROVER_assert(                                                          \
+      !(IDX <= I && I <= IDX + HAVOC_SIZE) | A[I] == OLD_A[I],                 \
+      "expecting FAILURE");                                                    \
+                                                                               \
+    __CPROVER_assert(                                                          \
+      (IDX <= I && I <= IDX + HAVOC_SIZE) | A[I] == OLD_A[I],                  \
+      "expecting SUCCESS");                                                    \
+  }
```
feliperodri: Nice! 💯

regression/cbmc/havoc_object_slice/test_array_symbolic_size.c
```diff
@@ -0,0 +1,46 @@
...
+  __CPROVER_havoc_object_slice(&a[idx], havoc_size);
+
+  // POSTCONDITIONS
+  CHECK_SLICE(a, old_a, 0, idx, havoc_size);
+  CHECK_SLICE(a, old_a, 1, idx, havoc_size);
+  CHECK_SLICE(a, old_a, 2, idx, havoc_size);
+  CHECK_SLICE(a, old_a, 3, idx, havoc_size);
+  CHECK_SLICE(a, old_a, 4, idx, havoc_size);
+  return;
+}
```
feliperodri: Add newline.

regression/cbmc/havoc_object_slice/test_array_symbolic_size.desc
```diff
@@ -0,0 +1,10 @@
+CORE thorough-paths thorough-smt-backend
+test_array_symbolic_size.c
+--pointer-check --pointer-primitive-check
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
```
feliperodri: Add newline.

regression/cbmc/havoc_object_slice/test_array_symbolic_size.desc
```diff
@@ -0,0 +1,10 @@
+CORE thorough-paths thorough-smt-backend
+test_array_symbolic_size.c
+--pointer-check --pointer-primitive-check
```
feliperodri: Why do we need `--pointer-primitive-check`?

regression/cbmc/havoc_object_slice/test_nondet_conditional.c
```diff
@@ -0,0 +1,34 @@
...
+  }
+  else
+  {
+    __CPROVER_assert(i == old_i, "expecting SUCCESS");
+    __CPROVER_assert(s.i == old_s.i, "expecting FAILURE");
+    __CPROVER_assert(s.j == old_s.j, "expecting SUCCESS");
+  }
+
+  return;
+}
```
feliperodri: Add newline.

regression/cbmc/havoc_object_slice/test_nondet_conditional.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test_nondet_conditional.c
+--pointer-check --pointer-primitive-check
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
```
feliperodri: Add newline.

regression/cbmc/havoc_object_slice/test_nondet_conditional.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test_nondet_conditional.c
+--pointer-check --pointer-primitive-check
```
feliperodri: `--pointer-primitive-check` ?

regression/cbmc/havoc_object_slice/test_struct_a.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test_struct_a.c
+--pointer-check --pointer-primitive-check
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
```
feliperodri: Add newline.

regression/cbmc/havoc_object_slice/test_struct_a.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test_struct_a.c
+--pointer-check --pointer-primitive-check
```
feliperodri: `--pointer-primitive-check` ?

regression/cbmc/havoc_object_slice/test_struct_b.c
```diff
@@ -0,0 +1,31 @@
...
+  __CPROVER_assert(s.d == old_s.d, "expecting SUCCESS");
+  __CPROVER_assert(s.u.a == old_s.u.a, "expecting SUCCESS");
+  __CPROVER_assert(s.u.b[0] == old_s.u.b[0], "expecting SUCCESS");
+  __CPROVER_assert(s.u.b[1] == old_s.u.b[1], "expecting SUCCESS");
+  __CPROVER_assert(s.u.b[2] == old_s.u.b[2], "expecting SUCCESS");
+  __CPROVER_assert(s.u.b[3] == old_s.u.b[3], "expecting SUCCESS");
+  __CPROVER_assert(s.u.b[4] == old_s.u.b[4], "expecting SUCCESS");
+  __CPROVER_assert(s.u.c == old_s.u.c, "expecting SUCCESS");
+  __CPROVER_assert(s.u.d == old_s.u.d, "expecting SUCCESS");
+}
```
feliperodri: How many test do we have under `havoc_object_slice`? We should split them into test cases with descriptive names.

regression/cbmc/havoc_object_slice/test_struct_b.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test_struct_b.c
+--pointer-check --pointer-primitive-check
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
```
feliperodri: Add newline to all files.

regression/cbmc/havoc_object_slice/test_struct_b.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test_struct_b.c
+--pointer-check --pointer-primitive-check
```
feliperodri: Why do we need `--pointer-primitive-check`? Same for all test cases in this PR.

src/goto-programs/builtin_functions.cpp
feliperodri: +1

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+  {
+    error().source_location = source_location;
+    error() << "'" << identifier << "' not expected to have a LHS" << eom;
+    throw 0;
+  }
+
+  // insert instructions
+  // assert(rw_ok(argument[0], argument[1]));
+  // char ndarr[argument[1]];
+  // __CPROVER_array_replace(p, ndarr);
```
feliperodri: Nice!

src/goto-programs/goto_convert_class.h
```diff
@@ -672,6 +672,11 @@ class goto_convertt:public messaget
     const symbol_exprt &function,
     const exprt::operandst &arguments,
     goto_programt &dest);
+  void do_havoc_object_slice(
+    const exprt &lhs,
+    const symbol_exprt &function,
+    const exprt::operandst &arguments,
+    goto_programt &dest);
```
feliperodri: I'd suggest to add docs, but no other function in this file has Doxygen comments, so maybe we should address this on a separate PR.

feliperodri (changes requested):
Only minor changes. Almost ready to go 😄

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: added

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
   copy(assignment, ASSIGN, dest);
 }
 
+void goto_convertt::do_havoc_slice(
+  const exprt &lhs,
+  const symbol_exprt &function,
+  const exprt::operandst &arguments,
+  goto_programt &dest)
+{
+  std::string identifier = CPROVER_PREFIX "havoc_slice";
```
tautschnig: Use `irep_idt` instead of `std::string`.

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+  auto source_location = function.find_source_location();
+  source_location.add_pragma("disable:pointer-check");
+  source_location.add_pragma("disable:pointer-overflow-check");
+  source_location.add_pragma("disable:pointer-primitive-check");
+
+  // check # arguments
+  if(arguments.size() != 2)
+  {
+    error().source_location = source_location;
+    error() << "'" << CPROVER_PREFIX "havoc_slice"
```
tautschnig: Use `identifier` here.

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+            << "' expected to have two arguments" << eom;
+    throw 0;
+  }
+
+  // check argument types
+  if(arguments[0].type().id() != ID_pointer)
+  {
+    error().source_location = source_location;
+    error() << "'" << CPROVER_PREFIX "havoc_slice"
```
tautschnig: Use `identifier` here.

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+            << "' first argument expected to have `void *` type" << eom;
+    throw 0;
+  }
+
+  if(arguments[1].type().id() != ID_unsignedbv)
+  {
+    error().source_location = source_location;
+    error() << "'" << CPROVER_PREFIX "havoc_slice"
```
tautschnig: Use `identifier` here.

doc/cprover-manual/api.md
jimgrundy: Can we just add a more explicit warning to the document about havoc_object that it will havoc enclosing  structs.

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+  // __CPROVER_array_replace(p, ndarr);
+
+  r_or_w_ok_exprt ok_expr(ID_w_ok, arguments[0], arguments[1]);
+  ok_expr.add_source_location() = source_location;
+  goto_programt::targett t =
+    dest.add(goto_programt::make_assertion(ok_expr, source_location));
+  t->source_location_nonconst().set("user-provided", false);
+  t->source_location_nonconst().set_property_class(ID_assertion);
+  t->source_location_nonconst().set_comment(
+    "assertion havoc_slice " + from_expr(ns, identifier, ok_expr));
```
tautschnig, line 700: Rather than fixing up the location after the fact, make these changes to a source location beforehand, like:
````c++
source_locationt annotated_location = source_location;
annotated_location.set("user-provided", ...); ...
````
and then use `annotated_location` when constructing the goto-program instruction.

File: src/goto-programs/builtin_functions.cpp [hunk 31]
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+
+  const array_typet array_type(char_type(), arguments[1]);
+
+  const symbolt &free_array =
+    new_tmp_symbol(array_type, "ndarr", dest, source_location, ID_C);
```
tautschnig: Add a parameter to the method (`const irep_idt &mode`) as some other methods do and use `mode` instead of hard-wiring `ID_C`.

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+
+  const exprt &arg0 =
+    typecast_exprt::conditional_cast(arguments[0], pointer_type(empty_typet{}));
+  const exprt &arg1 = typecast_exprt::conditional_cast(
+    free_array_expr, pointer_type(empty_typet{}));
+
+  codet array_replace(ID_array_replace);
+  array_replace.add_source_location() = source_location;
+  array_replace.copy_to_operands(arg0);
+  array_replace.copy_to_operands(arg1);
```
tautschnig: Use the `codet` constructor that consumes a source location and operands (`codet(const irep_idt &statement, operandst op, source_locationt loc)`).

tautschnig (approved):
Approving, but please address the comments before merging.

src/goto-programs/builtin_functions.cpp
SaswatPadhi: Hmm, thinking more, I think it could actually be:

```suggestion
  if(arguments[1].type() != size_type())
```

We define the signature of `__CPROVER_havoc_slice` as:

```c
 void __CPROVER_havoc_slice(void *p, __CPROVER_size_t size);
```
for the frontend.

So, shouldn't the check here match the this signature?

Yes it should be an `unsignedbv` type, but I think it shouldn't be _any_ `unsignedbv` type. It should match the `size_t` semantics on the machine, which is computed by `size_type()`:

https://github.com/diffblue/cbmc/blob/48893287099cb5780302fe9dc415eb6888354fd6/src/util/c_types.cpp#L58-L72

Or am I missing something here?

src/goto-programs/goto_instruction_code.h
```diff
@@ -537,4 +538,18 @@ inline code_returnt &to_code_return(codet &code)
...
+/// \param p the pointer argument
+/// \param size the size argument
+/// \return a \ref code_function_callt expression `nil_exprt() := __CPROVER_havoc_slice(p, size)`
+inline code_function_callt havoc_slice_call(exprt &p, exprt &size)
+{
+  code_function_callt::argumentst arguments;
+  arguments.push_back(p);
+  arguments.push_back(size);
+  symbol_exprt havoc_slice_function =
+    symbol_exprt{CPROVER_PREFIX "havoc_slice", empty_typet()};
```
tautschnig: This needs to be of a `code_typet` type, and then makes me wonder whether you should rather fetch this symbol from the symbol table. And that, in turn, makes me wonder whether such a function should make sure it is present in the symbol table.

src/ansi-c/scanner.l
```diff
@@ -425,6 +425,20 @@ void ansi_c_scanner_init()
                   PARSER.set_pragma_cprover();
                 }
 
+<CPROVER_PRAGMA>{ws}"check"{ws}"enable"{ws}{named_check} {
```
tautschnig: I'm not sure this should be part of the same pull request?

doc/cprover-manual/api.md
```diff
@@ -211,6 +211,54 @@ returns true when it is safe to do both.  These predicates can be given an
...
+and updates `size` consecutive bytes of the underlying object, starting at `p`, with nondeterministic values.
+
+```C
+void __CPROVER_havoc_slice(void *p, __CPROVER_size_t size);
+```
+
+**Caveat**
+
+- If the slice contains bytes that can be interpreted as pointers by the program,
+  this will cause these pointers to become invalid (ie they will not point to anything meaningful).
```
feliperodri: i.e.

regression/cbmc/havoc_slice/declarations.h
```diff
@@ -0,0 +1,26 @@
+#include <stdint.h>
+#include <stdlib.h>
+
+// HAVOC STRUCT MEMBERS
+// TODO take into account plaform/pointer size in the tests
```
feliperodri, line 5: Do we have a GitHub issue to track this? If yes, we should add a link to it here.

regression/cbmc/havoc_slice_checks/main.c
```diff
@@ -0,0 +1,31 @@
+// Here we test that __CPROVER_havoc_slice checks fail as expected
```
feliperodri, line 1: We usually add all test descriptions at the end of the `test.desc` files. Same for all others.

regression/cbmc/havoc_slice/test_whole_array_with_offset.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+test_whole_array_with_offset.c
+
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^.*expecting FAILURE: SUCCESS$
+^.*expecting SUCCESS: FAILURE$
+^.*dereference .*: FAILURE$
```
feliperodri, line 10: I see that you have a different name for each test case (i.e., a `.c` file and a test description). The usual pattern across the regression tests is to create a folder with a descriptive name (like the ones you're using for each test case) and add a `main.c` file and a `test.desc` file inside that folder. At the end of each test description, we add a small explanation of what behavior the test is covering. Not a blocker, but I suggest you to update this test to follow the same pattern.

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+  source_location.add_pragma("disable:pointer-primitive-check");
+
+  // check # arguments
+  if(arguments.size() != 2)
+  {
+    error().source_location = source_location;
+    error() << "'" << CPROVER_PREFIX "havoc_slice"
+            << "' expected to have two arguments" << eom;
+    throw 0;
+  }
```
feliperodri, line 661: We probably don't have any cases where we try to pass more than 2 parameters, thus, we don't hit this case.

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+
+  // check argument types
+  if(arguments[0].type().id() != ID_pointer)
+  {
+    error().source_location = source_location;
+    error() << "'" << CPROVER_PREFIX "havoc_slice"
+            << "' first argument expected to have `void *` type" << eom;
+    throw 0;
+  }
```
feliperodri, line 670: Same as above for non-pointer parameters.

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+
+  if(arguments[1].type().id() != ID_unsignedbv)
+  {
+    error().source_location = source_location;
+    error() << "'" << CPROVER_PREFIX "havoc_slice"
+            << "' second argument expected to have `size_t` type" << eom;
+    throw 0;
+  }
```
feliperodri, line 678: Same as above. We can create one test to cover these error cases.

src/goto-programs/builtin_functions.cpp
```diff
@@ -636,6 +636,88 @@ void goto_convertt::do_enum_is_in_range(
...
+
+  // check nil lhs
+  if(lhs.is_not_nil())
+  {
+    error().source_location = source_location;
+    error() << "'" << identifier << "' not expected to have a LHS" << eom;
+    throw 0;
+  }
```
feliperodri, line 686: Same as above.

src/goto-programs/goto_instruction_code.h
```diff
@@ -537,4 +538,21 @@ inline code_returnt &to_code_return(codet &code)
   return static_cast<code_returnt &>(code);
 }
 
+/// \brief Builds a \ref code_function_callt
+/// to `__CPROVER_havoc_slice(p, size)`.
+/// \param p the pointer argument
+/// \param size the size argument
+/// \return a \ref code_function_callt expression
+///         `nil_exprt() := __CPROVER_havoc_slice(p, size)`
```
feliperodri:
```suggestion
/// \brief Builds a \ref code_function_callt
/// to `__CPROVER_havoc_slice(p, size)`.
///
/// \param p The pointer argument.
/// \param size The size argument.
///
/// \return A \ref code_function_callt expression
///         `nil_exprt() := __CPROVER_havoc_slice(p, size)`.
```

feliperodri (approved):
LGTM module a few more tests updates and Michael's comments.

src/goto-programs/goto_instruction_code.h
remi-delmas-3000: The symbol's signature is only declared in `src/ansi-c/cprover_builtin_headers.h`.
If someone were to build a goto model programatically without using the C front-end, is there a guarantee that this signature will be present in the symbol table ?

src/ansi-c/scanner.l
remi-delmas-3000: I'll clean this up

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: done.

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: done

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: done

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: done

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: done

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: done

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: done

src/goto-programs/builtin_functions.cpp
remi-delmas-3000: Since tests are expressed in C and uses of havoc_slice are checked by the front end these would be impossible to reach using a test expressed in C.
Covering them would require build and load a goto model directly or create them in C++ directly.

src/goto-programs/goto_instruction_code.h
tautschnig: No, there is no such guarantee. Hence, it would be necessary to check, and add the symbol if it isn't present just yet.

doc/cprover-manual/api.md
```diff
@@ -211,6 +211,57 @@ returns true when it is safe to do both.  These predicates can be given an
 optional size; when the size argument is not given, the size of the subtype
 (which must not be **void**) of the pointer type is used.
 
+#### \_\_CPROVER\_havoc\_object
```
tautschnig, line 214: Random place for comment: it would be great if the commit message also answered the "Why?" question. See https://chris.beams.io/posts/git-commit/ as linked from the pull-request template. On this occasion, it would be good to answer the "why did `__CPROVER_havoc_object` not suffice" question.

src/goto-programs/builtin_functions.cpp
tautschnig: Did you perhaps end up going back and forth on this? It doesn't seem to have been addressed.

src/goto-programs/goto_instruction_code.cpp
```diff
@@ -70,3 +73,13 @@ void code_outputt::check(const codet &code, const validation_modet vm)
   DATA_CHECK(
     vm, code.operands().size() >= 2, "output must have at least two operands");
 }
+
+inline code_function_callt
```
tautschnig, line 77: Remove `inline`

src/goto-programs/goto_instruction_code.h
```diff
@@ -537,4 +538,18 @@ inline code_returnt &to_code_return(codet &code)
...
+///
+/// \param p The pointer argument.
+/// \param size The size argument.
+/// \param ns Namespace where the `__CPROVER_havoc_slice symbol` can be found.
+/// \remarks: It is a PRECONDITION that `__CPROVER_havoc_slice` exists
+///   in the namespace
+///
+/// \return A \ref code_function_callt expression
+///         `nil_exprt() := __CPROVER_havoc_slice(p, size)`.
+inline code_function_callt
```
tautschnig, line 552: Remove `inline`

src/goto-programs/goto_instruction_code.h
```diff
@@ -9,6 +9,7 @@ Author: Daniel Kroening, dkr@amazon.com
 #ifndef CPROVER_UTIL_GOTO_INSTRUCTION_CODE_H
 #define CPROVER_UTIL_GOTO_INSTRUCTION_CODE_H
 
+#include <util/cprover_prefix.h>
```
tautschnig, line 12: You don't actually need this in the header file anymore.

tautschnig (approved):

Issue #5494, opened 2020-09-16
Possible performance optimisation: replacing multiple pointer dereferences with temporary variable storing dereferenced pointer


natasha-jeppu:
CBMC version: `5.12.6 (cbmc-5.12.6-47-ge06f678f2)`
Operating system: `macOS Catalina 10.15.6`
Exact command line resulting in the issue: `make result`
What behaviour did you expect: `Proof runtime below 10mins`
What happened instead: `Runtime without optimisation = ~ 5hrs, Runtime with optimisation = ~300s`

**To reproduce these results use the following:**
Code base: `awslabs/s2n` 
Proof: `s2n_stuffer_write_reservation`
Code without optimisation: https://github.com/natasha-jeppu/s2n/blob/without_opt/stuffer/s2n_stuffer.c#L38
Code with optimisation: https://github.com/natasha-jeppu/s2n/blob/with_opt/stuffer/s2n_stuffer.c#L38
Command: `make result`

**Optimisation**: Replace multiple dereferences of `struct s2n_stuffer_reservation* reservation` pointer with a temporary object `struct s2n_stuffer_reservation reservation_obj = *reservation` and use `reservation_obj` for subsequent struct member access. `Observed runtime decrease from 5hrs to 5mins`
It would be convenient if CBMC could do this for similar constructs.

This behaviour can be explained with two profiling metrics (PRs submitted):

- Solver query complexity (extension of --write-solver-stats-to) : https://github.com/diffblue/cbmc/pull/5480
- Points-to set metric (--show-points-to-sets) : https://github.com/diffblue/cbmc/pull/5475.

**Without the optimisation**, the `reservation` pointer is dereferenced 2656 times. Here, dereferences are counted as the number of calls to `value_set_dereferencet::dereference()`, for a given pointer, that generates the case split expression.

![Screenshot 2020-09-15 at 21 47 35](https://user-images.githubusercontent.com/13853048/93263036-1441bc80-f79d-11ea-8174-24c5378d3123.png)

The corresponding solver query complexity metric shows 10798 clauses for https://github.com/natasha-jeppu/s2n/blob/without_opt/stuffer/s2n_stuffer.c#L42 and SSA expr is evidence of the complex case split generated.

![Screenshot 2020-09-15 at 22 51 21](https://user-images.githubusercontent.com/13853048/93268981-07759680-f7a6-11ea-8263-bfb6b57d7807.png)

**With the optimisation**, the `reservation` pointer is dereferenced only once. 
[...]

PR #6431, opened 2021-11-02
Parallelize DIMACS formula printing

srogatch:

### DIFF
--- a/src/solvers/sat/dimacs_cnf.cpp
+++ b/src/solvers/sat/dimacs_cnf.cpp
@@ -13,6 +13,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/magic.h>
 
 #include <iostream>
+#include <thread>
 
 dimacs_cnft::dimacs_cnft(message_handlert &message_handler)
   : cnf_clause_listt(message_handler), break_lines(false)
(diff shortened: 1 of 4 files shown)


src/solvers/sat/dimacs_cnf.cpp
```diff
@@ -76,25 +77,62 @@ void dimacs_cnft::write_dimacs_clause(
   out << "0" << "\n";
 }
 
+void dimacs_cnft::wait_file_block(const size_t ordinal) {
+  std::unique_lock<std::mutex> lock(writing_sync);
+  for(;;) {
+    assert(next_file_block <= ordinal);
```
martin-cs, line 83: Please don't add `assert`s, use the various variants of `INVARIANT`.

martin-cs (commented):
Thanks for this PR.  Adding threads to CBMC would be quite a big architectural step.  It is not out of the question but I think it would need quite a significant use-case.  Can you talk a little more about why you need this functionality and how it works in a wider context?  Printing dimacs is not normally a particularly time-critical step, esp. compared with solving.  Does having multiple threads really give a significant speed up over, say, outputting to a file on a RAM disk?

srogatch:
> Thanks for this PR. Adding threads to CBMC would be quite a big architectural step. It is not out of the question but I think it would need quite a significant use-case. Can you talk a little more about why you need this functionality and how it works in a wider context? Printing dimacs is not normally a particularly time-critical step, esp. compared with solving. Does having multiple threads really give a significant speed up over, say, outputting to a file on a RAM disk?

Sure, thanks for taking a look! DIMACS printing indeed takes a considerable time: CBMC converts my program into a boolean formula in ~200 seconds, then something else is going on (don't remember for how much - I think it's converting to SSA taking ~70 seconds), then 12GB of the formula are written to disk at the speed of ~50MBytes per second on a super-fast SSD with a single thread. So that takes ~240 seconds. Then `kissat` solves the formula in ~600 seconds (because the formula is easy and has probable solutions with a random assignment of some variables). So DIMACS printing in a single thread takes as much as the rest of the processing (except the solver), and it is not bound by the SSD speed (SSD is PCIE-4.0 with up to 7GBytes per second transfers, 1 million IOPS). The printing is bound by C++ stream speed (conversion from binary values to text). With multithreading, on a 128 logical core CPU (Ryzen Threadripper 3990X) I get the speed of writing to the disk about 4GBytes/second, so the formula gets saved in a few seconds.

martin-cs:
> Sure, thanks for taking a look! DIMACS printing indeed takes a
> considerable time: CBMC converts my program into a boolean formula in
> ~200 seconds, then something else is going on (don't remember for how
> much - I think it's converting to SSA taking ~70 seconds),

The option `--timestamp monotonic` can help for this kind of
application level profiling.

>  then 12GB of the formula are written to disk at the speed of
> ~50MBytes per second on a super-fast SSD with a single thread. So
> that takes ~240 seconds. Then kissat solves the formula in ~600
> seconds (because the formula is easy and has probable solutions with
> a random assignment of some variables).

OK; thanks for the timings, that really helps with the context.

>  So DIMACS printing in a single thread takes as much as the rest of
> the processing (except the solver), and it is not bound by the SSD
> speed (SSD is PCIE-4.0 with up to 7GBytes per second transfers, 1
> million IOPS). The printing is bound by C++ stream speed (conversion
> from binary values to text). On a 128 logical thread CPU (Ryzen
> Threadripper 3990X) I get the speed of writing to the disk about
> 4GBytes/second, so the formula gets saved in a few seconds.

At the risk of sounding ungrateful, have you considered:

1. Profiling the printing part of the code and seeing if we can
optimise it at all?  If we could remove the bottleneck it might be
better than adding more compute to it.

2. Integrating Kissat as a SAT back-end so that the output to DIMACS is
not needed at all?  I think @natasha-jeppu was using kissat ( 
https://github.com/diffblue/cbmc/issues/5494 ) so I think it can be
done.

martin-cs:
https://github.com/diffblue/cbmc/pull/6439 is an example of adding support for a new SAT solver.

srogatch:
> At the risk of sounding ungrateful, have you considered: 1. Profiling the printing part of the code and seeing if we can optimise it at all? If we could remove the bottleneck it might be better than adding more compute to it. 2. Integrating Kissat as a SAT back-end so that the output to DIMACS is not needed at all? I think @natasha-jeppu was using kissat ( #5494 ) so I think it can be done.

1. I didn't profile it, but I know from experience that streams in C++ (and even C) are extremely slow. Only a custom printing routine could help, that would convert integers to strings in an efficient way. Still I think this routine will not achieve the speed of multiple cores.
2. In that PR the author used `kissat` via DIMACS formula printing. `kissat` is not yet ready for integration: not all the API is exposed.

PR #6436, opened 2021-11-03
goto_symex: implement SET_RETURN_VALUE

kroening:


src/goto-programs/mm_io.cpp
```diff
@@ -105,26 +106,37 @@ void mm_io(
...
+      to_code_type(mm_io_r.type()).return_type(),
+      id2string(id_r) + "$value",
+      id2string(id_r) + "$value",
+      maybe_symbol->location,
+      maybe_symbol->mode,
+      symbol_table);
+
+    mm_io_r_value = value_symbol.symbol_expr();
+  }
+
```
tautschnig, line 133: Why are the changes to `mm_io` necessary?

src/goto-programs/process_goto_program.cpp
```diff
@@ -58,9 +58,16 @@ bool process_goto_program(
   }
 
   // remove returns, gcc vectors, complex
-  remove_returns(goto_model);
+  if(
+    options.get_bool_option("remove-returns") ||
+    options.get_bool_option("string-abstraction"))
```
tautschnig, line 63: String abstraction was built well before return-value removal, so I wouldn't really expect a dependency here. In 425731b66727c40607457cef1a82acb3f6f316b6 I noticed a need for some fixes, but I don't think that there should be a general requirement for return-value removal to make string abstraction work.

src/goto-programs/validate_goto_model.cpp
```diff
@@ -48,7 +48,7 @@ class validate_goto_modelt
   /// - no return statements in any of the functions
   /// - lhs of every \ref code_function_callt instruction is nil
   /// - all return types are void (of both calls and functions themselves)
-  void check_returns_removed();
+  // void check_returns_removed();
```
tautschnig: I think we should either completely remove this check, or else introduce a bit map for goto programs that indicates which passes have run (and then make this check conditional).

src/goto-programs/mm_io.cpp
kroening: The current implementation assumes that the return value is in the function#return_value variable. It needs to be changed to use the lhs in the function call.

src/goto-programs/process_goto_program.cpp
kroening: It's unclear what the effort is. Might not be worth it, given that this feature has no known users.

src/goto-programs/validate_goto_model.cpp
kroening: Ok, will remove.

kroening: done

tautschnig (approved):

peterschrammel (approved):

PR #6439, opened 2021-11-04
Add support for MergeSat

tautschnig:
Adds support for MergeSat.

### DIFF
--- a/src/solvers/Makefile
+++ b/src/solvers/Makefile
@@ -22,6 +22,21 @@ ifneq ($(MINISAT2),)
   CLEANFILES += $(MINISAT2_LIB) $(patsubst %$(OBJEXT), %$(DEPEXT), $(MINISAT2_LIB))
 endif
 
+ifneq ($(MERGESAT),)
+  # MergeSat is based on MiniSat2 and is invoked (with suitable defines/ifdefs)
+  # via satcheck_minisat2.{h,cpp}
+  MERGESAT_SRC=sat/satcheck_minisat2.cpp
+  MERGESAT_INCLUDE=-I $(MERGESAT)
+  MERGESAT_LIB=$(MERGESAT)/minisat/core/Lookahead$(OBJEXT) \
+               $(MERGESAT)/minisat/core/Solver$(OBJEXT) \
+               $(MERGESAT)/minisat/simp/SimpSolver$(OBJEXT) \
+               $(MERGESAT)/minisat/utils/ccnr$(OBJEXT) \
+               $(MERGESAT)/minisat/utils/Options$(OBJEXT) \
+               $(MERGESAT)/minisat/utils/System$(OBJEXT)
+  CP_CXXFLAGS += -DHAVE_MERGESAT -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS
+  CLEANFILES += $(MERGESAT_LIB) $(patsubst %$(OBJEXT), %$(DEPEXT), $(MERGESAT_LIB))
+endif
+
 ifneq ($(IPASIR),)
   IPASIR_SRC=sat/satcheck_ipasir.cpp
   IPASIR_INCLUDE=-I $(IPASIR)
@@ -75,6 +90,7 @@ SRC = $(BOOLEFORCE_SRC) \
       $(GLUCOSE_SRC) \
       $(LINGELING_SRC) \
       $(MINISAT2_SRC) \
+      $(MERGESAT_SRC) \
       $(IPASIR_SRC) \
       $(MINISAT_SRC) \
       $(PICOSAT_SRC) \
@@ -234,6 +250,31 @@ $(MINISAT2)/minisat/core/Solver$(OBJEXT): $(MINISAT2)/minisat/core/Solver.cc
 endif
 endif
 
+ifneq ($(MERGESAT),)
+ifeq ($(BUILD_ENV_),MSVC)
+sat/satcheck_minisat2$(OBJEXT): sat/satcheck_minisat2.cpp
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+
+$(MERGESAT)/minisat/core/Lookahead$(OBJEXT): $(MERGESAT)/minisat/core/Lookahead.cc
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+
+$(MERGESAT)/minisat/core/Solver$(OBJEXT): $(MERGESAT)/minisat/core/Solver.cc
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+
+$(MERGESAT)/minisat/simp/SimpSolver$(OBJEXT): $(MERGESAT)/minisat/simp/SimpSolver.cc
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+
+$(MERGESAT)/minisat/utils/ccnr$(OBJEXT): $(MERGESAT)/minisat/utils/ccnr.cc
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+
+$(MERGESAT)/minisat/utils/Options$(OBJEXT): $(MERGESAT)/minisat/utils/Options.cc
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+
+$(MERGESAT)/minisat/utils/System$(OBJEXT): $(MERGESAT)/minisat/utils/System.cc
+	$(CXX) $(CP_CXXFLAGS) /w /nologo /c /EHsc $< /Fo$@
+endif
+endif
+
 ifneq ($(GLUCOSE),)
 ifeq ($(BUILD_ENV_),MSVC)
 sat/satcheck_glucose$(OBJEXT): sat/satcheck_glucose.cpp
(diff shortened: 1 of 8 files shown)


kroening:
I believe this is the only solver that uses the minisat identifiers (that isn't minisat). I may be worth adding some comments to make that apparent.

tautschnig:
> I believe this is the only solver that uses the minisat identifiers (that isn't minisat). I may be worth adding some comments to make that apparent.

Yes, good call! Where would you see the most natural place to find such comments (as a developer)?

src/solvers/sat/satcheck_minisat2.cpp
```diff
@@ -101,12 +105,20 @@ void satcheck_minisat2_baset<T>::clear_interrupt()
 
 const std::string satcheck_minisat_no_simplifiert::solver_text()
 {
+#ifdef HAVE_MERGESAT
+  return "Mergesat (latest) without simplifier";
```
martin-cs: Hard coded version number, even `"latest"` make me nervous.

scripts/mergesat.patch
```diff
@@ -0,0 +1,228 @@
...
+diff --git a/minisat/parallel/ParSolver.cc b/minisat/parallel/ParSolver.cc
+index ad4c235..b76566b 100644
+--- a/minisat/parallel/ParSolver.cc
++++ b/minisat/parallel/ParSolver.cc
+@@ -44,7 +44,7 @@ ParSolver::ParSolver()
+   : par_reparsed_options(updateOptions())
+   , parsing(false)
+   , verbosity(0)
+-  , use_simplification(true)
++  , use_simplification(false)
```
martin-cs: Where does this patch come from and what does it do?

martin-cs (approved):
In principle I don't see a problem.

src/solvers/sat/satcheck_minisat2.cpp
nmanthey: Is there a way to run `git describe` in the MergeSat directory during build time, and use the output of that?

scripts/mergesat.patch
nmanthey: This is disabling simplification by default. The alternative would be to support 2 solver types, with and without simplification, as you do for other solvers.

In case of using MergeSat's simplification, another tweak might be required: not allowing to run iterative simplifications with formula growth. That would be achieved by setting this parameter to 'false' by default: https://github.com/conp-solutions/mergesat/blob/3af84ff119ece9dd0ff6b9083022bb4264909ad1/minisat/simp/SimpSolver.cc#L48

Having an interface that allows to configure the SAT backend would be simpler going forward. Fortunately, MergeSat allows to do that by passing the parameters for the solver via an environment variable, e.g.

```
export MINISAT_RUNTIME_ARGS="-no-grow-iter"
```

would have the same effect.

nmanthey:
Thanks for picking this up. Let me know if there is anything on the SAT side to look at (except your PR)

src/solvers/Makefile
```diff
@@ -222,6 +231,28 @@ $(MINISAT2)/minisat/core/Solver$(OBJEXT): $(MINISAT2)/minisat/core/Solver.cc
 endif
 endif
 
+ifneq ($(MERGESAT),)
+ifeq ($(BUILD_ENV_),MSVC)
+sat/satcheck_minisat2$(OBJEXT): sat/satcheck_minisat2.cpp
```
kroening, line 255: I would add a comment here that Mergesat is invoked from satcheck_minisat2.cpp

src/solvers/sat/satcheck.h
```diff
@@ -80,9 +85,9 @@ typedef satcheck_booleforcet satcheck_no_simplifiert;
 typedef satcheck_minisat1t satcheckt;
 typedef satcheck_minisat1t satcheck_no_simplifiert;
 
-#elif defined SATCHECK_MINISAT2
+#elif defined SATCHECK_MINISAT2 || defined SATCHECK_MERGESAT
```
kroening, line 121: And another one here

src/solvers/sat/satcheck_minisat2.cpp
```diff
@@ -22,8 +22,8 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <minisat/core/Solver.h>
 #include <minisat/simp/SimpSolver.h>
 
-#ifndef HAVE_MINISAT2
```
kroening, line 29: And perhaps one here.

src/solvers/sat/satcheck_minisat2.cpp
martin-cs: Have a look at the set up for `src/util/version.cpp` here:

https://github.com/diffblue/cbmc/blob/74963e7f1d2bd584eafe715d5ed3ece538863760/src/util/Makefile#L122

scripts/mergesat.patch
martin-cs: My apologies; I wasn't clear.  My question is a bit higher-level.  Why are we patching MergeSat during the build?  With MiniSAT development is largely frozen.  Is this the case with MergeSat?  A number of these seem like pretty basic fixes so they shouldn't be too controversial?  Patching solvers after download is a hack and we should avoid it when possible.

conp-solutions: MergeSat is under development, so adding features that are useful for consumers of that solver is possible.

However, using the same default configuration of that solver for all use cases does not work well, due to environment restrictions - in MergeSats case that might mostly be the timeout. While parts of this could be addressed by making things more dynamic inside the solver, there is a lot of ground to cover to get there (and benchmarking to do, eventually adding ML models as we did for other solvers already in the past). The short-cut is to adapt the configuration for each use case accordingly. This patch is an attempt to do so for the/a CBMC use case.

I would love to have access to proper CBMC benchmarking, to get good data. Until then, it boils down to extrapolating from past experience.

src/solvers/Makefile
tautschnig: Done!

src/solvers/sat/satcheck.h
tautschnig: Done!

src/solvers/sat/satcheck_minisat2.cpp
tautschnig: Done.

tautschnig:
> IIUC, this variable needs to be set before the solver object is created. Have you checked whether this setting is actually passed through to the solver?
> 
> It might be simpler to always start from the SimpSolver object, and set the "use_simplification" parameter right after constructing the object, to indicate whether you want to have simplification or not. Then, you can also set the member behind the option you use in setenv, as this is always available as well. That would be member 'grow_iterations'

Done with minimal modifications, which, however, requires C++17. So I've added commits from #6749 to make this work (there's a CI job using MergeSat now).

tautschnig:
> nit: version "3.0" is not true

Is that still mentioned anywhere?

tautschnig:
> nit: Is this configuration actually tested on Windows? I did not test this myself, and know for earlier versions that this compilation does not work without modifications to MergeSat.

I've added MergeSat to the check-vs-2019-cmake-build-and-test CI job. Let's see!

tautschnig:
> > nit: Is this configuration actually tested on Windows? I did not test this myself, and know for earlier versions that this compilation does not work without modifications to MergeSat.
> 
> I've added MergeSat to the check-vs-2019-cmake-build-and-test CI job. Let's see!

Does not seem to work:

``` Lookahead.cc
D:\a\cbmc\cbmc\build\mergesat-src\minisat\core\Lookahead.cc(20,10): fatal error C1083: Cannot open include file: 'core/Lookahead.h': No such file or directory [D:\a\cbmc\cbmc\build\mergesat-build\mergesat-condensed.vcxproj]
cl : command line warning D9025: overriding '/W3' with '/w' [D:\a\cbmc\cbmc\build\mergesat-build\mergesat-condensed.vcxproj]
cl : Command line warning D9025: overriding '/W3' with '/w' [D:\a\cbmc\cbmc\build\mergesat-build\mergesat-condensed.vcxproj]
  Solver.cc
D:\a\cbmc\cbmc\build\mergesat-src\minisat\core\Solver.cc(50,10): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory 
```

I'll limit the experimenting to Linux-based platforms, but even then the uses from within refinement appear to be causing trouble. Needs looking at.

tautschnig:
Note to self: We'll likely want https://github.com/conp-solutions/mergesat/pull/124 to be merged and Mergesat 4.0 to be released before trying to merge this one.

src/solvers/sat/satcheck_minisat2.cpp
```diff
@@ -257,6 +272,14 @@ propt::resultt satcheck_minisat2_baset<T>::do_prop_solve()
 
 #endif
 
+#ifdef HAVE_MERGESAT
+    // We do not actually use MergeSat's "constrain" clauses at the moment, but
+    // MergeSat internally still uses them to track UNSAT. To make sure we
+    // aren't stuck with "UNSAT" in incremental calls the status needs to be
+    // reset.
+    ((Minisat::Solver *)solver.get())->reset_constrain_clause();
```
conp-solutions, line 299: This is not required anymore. There was an issue with how incremental solving was implemented. The latest update of MergeSat has that fixed. Hence, feel free to drop this hunk!

tautschnig: Thank you! I don't think we want to merge this PR before MergeSat 4.0 is released, so I suppose 4.0 will include this fix?

conp-solutions: Likely yes. On the other hand, I am not sure when I will release 4.0, i.. whether I wait for the SAT competition results or not. Hence, feel free to not wait. The major thing that will change is the default configuration of the solver - and the competition will help to identify which one to pick.

TGWDB:
The runtime for the new CI job looks unpleasantly long. Is this expected? It seems well outside the numbers on the job description.

tautschnig:
> The runtime for the new CI job looks unpleasantly long. Is this expected? It seems well outside the numbers on the job description.

We might take this as a first performance indicator (the tests taking 54 minutes as compared to 18 minutes that MiniSat takes). Perhaps the switch from 4.0-rc1 to 4.0-rc4 (of mergesat) wasn't a good idea?!

TGWDB:
> > The runtime for the new CI job looks unpleasantly long. Is this expected? It seems well outside the numbers on the job description.
> 
> We might take this as a first performance indicator (the tests taking 54 minutes as compared to 18 minutes that MiniSat takes). Perhaps the switch from 4.0-rc1 to 4.0-rc4 (of mergesat) wasn't a good idea?!

Yes. Without further knowledge/detail as to why (and whether it's an edge case we can accept), merging this seems like a bad idea.

PR #6447, opened 2021-11-08
Fix local static variables detection in assigns clauses

remi-delmas-3000:
Function contracts instrumentation:

1. Use the call graph of the function to detect local static variables declared in the functions it calls.

Other changes:
1. subfunction call inlining was moved from `code_contractst::check_frame_conditions` to `code_contractst::check_frame_conditions_function`.
2. Recursive functions are now detected during inlining and cause an INVARIANT violation
3. Function pointers are now detected during call graph construction and cause a PRECONDITION violation

### DIFF
--- a/regression/contracts/assigns_enforce_statics/main.c
+++ b/regression/contracts/assigns_enforce_statics/main.c
@@ -1,15 +1,15 @@
-#include <assert.h>
-#include <stdlib.h>
-
-static int x;
+static int x = 0;
 
 void foo() __CPROVER_assigns(x)
 {
   int *y = &x;
 
-  static int x;
+  static int x = 0;
+
+  // should pass (assigns local x)
   x++;
 
+  // should fail (assigns global x)
   (*y)++;
 }
 
--- a/regression/contracts/function-calls-recursive-01/main.c
+++ b/regression/contracts/function-calls-recursive-01/main.c
@@ -1,0 +1,23 @@
+#include <stdlib.h>
+
+int sum_rec(int i, int acc)
+{
+  if(i >= 0)
+    return sum_rec(i - 1, acc + i);
+  return acc;
+}
+
+int sum(int i) __CPROVER_requires(0 <= i && i <= 50)
+  __CPROVER_ensures(__CPROVER_return_value >= 0) __CPROVER_assigns()
+{
+  int j = i;
+  int res = sum_rec(j, 0);
+  return res;
+}
+
+int main()
+{
+  int result = sum(10);
+  __CPROVER_assert(result == 55, "result == 55");
+  return 0;
+}
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -13,6 +13,8 @@ Date: July 2021
 
 #include "assigns.h"
 
+#include <analyses/call_graph.h>
+
 #include <langapi/language_util.h>
 
 #include <util/arith_tools.h>
@@ -195,4 +197,32 @@ void havoc_assigns_targetst::append_havoc_code_for_expr(
   }
 
   havoc_utilst::append_havoc_code_for_expr(location, expr, dest);
+}
+
+void assigns_clauset::add_static_locals_to_write_set(
+  const goto_functionst &functions,
+  const irep_idt &root_function)
+{
+  auto call_graph =
+    call_grapht::create_from_root_function(functions, root_function, true)
+      .get_directed_graph();
+
+  for(const auto &sym_pair : symbol_table)
+  {
+    const auto &sym = sym_pair.second;
+    if(sym.is_static_lifetime)
+    {
+      auto fname = sym.location.get_function();
+      if(
+        !fname.empty() && (fname == root_function ||
+                           call_graph.get_node_index(fname).has_value()))
+      {
+        // We found a symbol with
+        // - a static lifetime
+        // - non empty location function
+        // - this function appears in the call graph of the function
+        add_to_write_set(sym.symbol_expr());
+      }
+    }
+  }
 }
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -29,6 +29,7 @@ Date: February 2016
 
 #include <util/c_types.h>
 #include <util/expr_util.h>
+#include <util/find_symbols.h>
 #include <util/fresh_symbol.h>
 #include <util/graph.h>
 #include <util/mathematical_expr.h>
@@ -37,9 +38,97 @@ Date: February 2016
 #include <util/pointer_offset_size.h>
 #include <util/pointer_predicates.h>
 #include <util/replace_symbol.h>
+#include <util/std_code.h>
 
 #include "memory_predicates.h"
 #include "utils.h"
+
+/// Decorator for \ref message_handlert that keeps track of warnings
+/// occuring when inlining a function.
+///
+/// It counts the number of :
+/// - recursive functions warnings
+/// - missing functions warnings
+/// - missing function body warnings
+/// - missing function arguments warnings
+class inlining_decoratort : public message_handlert
+{
+private:
+  message_handlert &wrapped;
+  unsigned int recursive_function_warnings_count = 0;
+
+  void parse_message(const std::string &message)
+  {
+    if(message.find("recursion is ignored on call") != std::string::npos)
+      recursive_function_warnings_count++;
+  }
+
+public:
+  explicit inlining_decoratort(message_handlert &_wrapped) : wrapped(_wrapped)
+  {
+  }
+
+  unsigned int get_recursive_function_warnings_count()
+  {
+    return recursive_function_warnings_count;
+  }
+
+  void print(unsigned level, const std::string &message) override
+  {
+    parse_message(message);
+    wrapped.print(level, message);
+  }
+
+  void print(unsigned level, const xmlt &xml) override
+  {
+    wrapped.print(level, xml);
+  }
+
+  void print(unsigned level, const jsont &json) override
+  {
+    wrapped.print(level, json);
+  }
+
+  void print(unsigned level, const structured_datat &data) override
+  {
+    wrapped.print(level, data);
+  }
+
+  void print(
+    unsigned level,
+    const std::string &message,
+    const source_locationt &location) override
+  {
+    parse_message(message);
+    wrapped.print(level, message, location);
+    return;
+  }
+
+  void flush(unsigned i) override
+  {
+    return wrapped.flush(i);
+  }
+
+  void set_verbosity(unsigned _verbosity)
+  {
+    wrapped.set_verbosity(_verbosity);
+  }
+
+  unsigned get_verbosity() const
+  {
+    return wrapped.get_verbosity();
+  }
+
+  std::size_t get_message_count(unsigned level) const
+  {
+    return wrapped.get_message_count(level);
+  }
+
+  std::string command(unsigned i) const override
+  {
+    return wrapped.command(i);
+  }
+};
 
 void code_contractst::check_apply_loop_contracts(
   const irep_idt &function_name,
@@ -69,6 +158,8 @@ void code_contractst::check_apply_loop_contracts(
 
   assigns_clauset loop_assigns(
     assigns.operands(), log, ns, function_name, symbol_table);
+
+  loop_assigns.add_static_locals_to_write_set(goto_functions, function_name);
 
   if(invariant.is_nil())
   {
@@ -688,8 +779,13 @@ void code_contractst::apply_loop_contract(
   if(!natural_loops.loop_map.size())
     return;
 
+  inlining_decoratort decorated(log.get_message_handler());
   goto_function_inline(
     goto_functions, function_name, ns, log.get_message_handler());
+
+  INVARIANT(
+    decorated.get_recursive_function_warnings_count() == 0,
+    "Recursive functions found during inlining");
 
   // A graph node type that stores information about a loop.
   // We create a DAG representing nesting of various loops in goto_function,
@@ -985,17 +1081,12 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     function,
     symbol_table);
 
-  // Adds formal parameters to write set
+  // detect and add static local variables
+  assigns.add_static_locals_to_write_set(goto_functions, function);
+
+  // Add formal parameters to write set
   for(const auto &param : to_code_type(target.type).parameters())
     assigns.add_to_write_set(ns.lookup(param.get_identifier()).symbol_expr());
-
-  // Adds local static declarations to write set
-  for(const auto &sym_pair : symbol_table)
-  {
-    const auto &sym = sym_pair.second;
-    if(sym.is_static_lifetime && sym.location.get_function() == function)
-      assigns.add_to_write_set(sym.symbol_expr());
-  }
 
   auto instruction_it = function_obj->second.body.instructions.begin();
   for(const auto &car : assigns.get_write_set())
@@ -1005,9 +1096,19 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
       function_obj->second.body, instruction_it, snapshot_instructions);
   };
 
-  // Inline all function calls.
-  goto_function_inline(
-    goto_functions, function_obj->first, ns, log.get_message_handler());
+  // Full inlining of the function body
+  // Inlining is performed so that function calls to a same function
+  // occurring under different write sets get instrumented specifically
+  // for each write set
+  inlining_decoratort decorated(log.get_message_handler());
+  goto_function_inline(goto_functions, function, ns, decorated);
+
+  INVARIANT(
+    decorated.get_recursive_function_warnings_count() == 0,
+    "Recursive functions found during inlining");
+
+  // restore internal invariants
+  goto_functions.update();
 
   // Insert write set inclusion checks.
   check_frame_conditions(
--- a/src/goto-instrument/contracts/contracts.h
+++ b/src/goto-instrument/contracts/contracts.h
@@ -58,7 +58,6 @@ public:
       goto_functions(goto_model.goto_functions),
       log(log),
       converter(symbol_table, log.get_message_handler())
-
   {
   }
 
--- a/regression/contracts/assigns_enforce_detect_local_statics/main.c
+++ b/regression/contracts/assigns_enforce_detect_local_statics/main.c
@@ -1,0 +1,31 @@
+#include <assert.h>
+#include <stdlib.h>
+
+static int x;
+static int xx;
+
+void foo()
+{
+  int *y = &x;
+  int *yy = &xx;
+
+  static int x;
+  // must pass (modifies local static)
+  x++;
+
+  // must pass (modifies assignable global static )
+  (*y)++;
+
+  // must fail (modifies non-assignable global static)
+  (*yy)++;
+}
+
+void bar() __CPROVER_assigns(x)
+{
+  foo();
+}
+
+int main()
+{
+  bar();
+}
--- a/regression/contracts/assigns_enforce_detect_local_statics/test.desc
+++ b/regression/contracts/assigns_enforce_detect_local_statics/test.desc
@@ -1,0 +1,14 @@
+CORE
+main.c
+--enforce-contract bar
+^EXIT=10$
+^SIGNAL=0$
+^\[foo.\d+\] line \d+ Check that y is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that foo\$\$1\$\$x is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that \*y is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that \*yy is assignable: FAILURE$
+^VERIFICATION FAILED$
+--
+--
+Checks whether static local and global variables are correctly tracked
+in assigns clause verification, accross subfunction calls.
--- a/regression/contracts/function-calls-03-1/test-enf-f1.desc
+++ b/regression/contracts/function-calls-03-1/test-enf-f1.desc
@@ -1,20 +1,12 @@
 CORE
 main.c
 --enforce-contract f1 _ --unwind 20 --unwinding-assertions
-^EXIT=10$
+^file main\.c line 13 function f2: recursion is ignored on call to 'f2'$
+^Invariant check failed$
+^Condition: decorated\.get_recursive_function_warnings_count\(\) == 0$
+^Reason: Recursive functions found during inlining$
+^EXIT=(127|134|137)$
 ^SIGNAL=0$
-^file main.c line \d+ function f2: recursion is ignored on call to \'f2\'$
-^\[postcondition.\d+\] file main.c line \d+ Check ensures clause: FAILURE$
-^VERIFICATION FAILED$
 --
 --
-Verification:
-  function | pre-cond | post-cond
-  ---------|----------|----------
-  f1       | assumed  | asserted
-
-Test should fail:
-The postcondition of f2 is incorrect, considering the recursion particularity. 
-
-Recursion:
-The base case for the recursive call to f2 provides different behavior than the common case (given the pre-conditions).
+Test should fail because a recursive function is found during inlining.
--- a/regression/contracts/function-calls-03-1/test-enf-f2.desc
+++ b/regression/contracts/function-calls-03-1/test-enf-f2.desc
@@ -1,21 +1,12 @@
 CORE
 main.c
 --enforce-contract f2 _ --unwind 20 --unwinding-assertions
-^EXIT=10$
+^file main\.c line 13 function f2: recursion is ignored on call to 'f2'$
+^Invariant check failed$
+^Condition: decorated\.get_recursive_function_warnings_count\(\) == 0$
+^Reason: Recursive functions found during inlining$
+^EXIT=(127|134|137)$
 ^SIGNAL=0$
-^file main.c line \d+ function f2: recursion is ignored on call to \'f2\'$
-^\[postcondition.\d+\] file main.c line \d+ Check ensures clause: FAILURE$
-^\[f2.\d+\] line \d+ Check that loc is assignable: SUCCESS$
-^VERIFICATION FAILED$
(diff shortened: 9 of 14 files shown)


src/goto-instrument/contracts/contracts.cpp
```diff
@@ -861,7 +861,7 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
   for(const auto &sym_pair : symbol_table)
   {
     const auto &sym = sym_pair.second;
-    if(sym.is_static_lifetime && sym.location.get_function() == function)
+    if(sym.is_static_lifetime && !sym.location.get_function().empty())
```
SaswatPadhi: Wouldn't this create CARs for ALL static local variables, like beyond the current function on which contracts are checked and its subfunction calls?

It might have a significant perf impact if lots of unrelated functions also introduce static local variables, because we'll add their CARs unnecessarily to the write set and check every assignment against them.

SaswatPadhi: It might be worth fixing the inlining routine instead. Inlined static locals should just behave as static locals within the caller. It seems weird and inconsistent to have the function body inlined into a caller but be mutating a static local that's tagged with a different function name.

remi-delmas-3000: Indeed. 
Since all sub-function calls are inlined for contract checking, a better solution would be to detect static locals used in the body of the function under analysis rather than at the symbol table level.

remi-delmas-3000:
Updated to a better solution is to detect static locals used in the body of the function under analysis rather than at the symbol table level.

regression/contracts/assigns_enforce_statics_subfunctions/main.c
```diff
@@ -0,0 +1,31 @@
+#include <assert.h>
```
feliperodri: You don't need to include this include here.

regression/contracts/assigns_enforce_statics_subfunctions/test.desc
```diff
@@ -0,0 +1,14 @@
...
+^SIGNAL=0$
+^\[foo.\d+\] line \d+ Check that y is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that foo\$\$1\$\$x is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that \*y is assignable: SUCCESS$
+^\[foo.\d+\] line \d+ Check that \*yy is assignable: FAILURE$
+^VERIFICATION FAILED$
+--
+--
+Checks whether static local and global variables are correctly tracked
+in assigns clause verification, accross subfunction calls.
```
feliperodri: nit:
```Makefile
in assigns clause verification, across sub-function calls.
```

regression/contracts/assigns_enforce_statics_subfunctions/test.desc
```diff
@@ -0,0 +1,14 @@
+CORE
+main.c
+--enforce-contract bar _ --pointer-primitive-check
```
feliperodri: Why do we need `--pointer-primitive-check` here?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1260,3 +1258,63 @@ bool code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
   }
   return fail;
 }
+
+void code_contractst::find_static_locals(
+  const goto_functiont &function,
+  assigns_clauset &assigns)
+{
+  find_symbols_sett tmp_set;
```
feliperodri: nit: we could call it `static_local_set` instead of `tmp_set`.

src/goto-instrument/contracts/contracts.h
```diff
@@ -210,6 +210,30 @@ class code_contractst
     codet &expression,
     source_locationt location,
     const irep_idt &mode);
+
+  
+  /// Finds static local symbols that occur in the body of the given function and adds them to the the write set of `assigns`.
```
feliperodri:
```suggestion
  /// \brief Finds static local symbols that occur in the body of the given function and adds them to the the write set of `assigns`.
```

src/goto-instrument/contracts/contracts.h
```diff
@@ -210,6 +210,30 @@ class code_contractst
...
+  ///
+  /// @param function the function to search local static symbols in
+  /// @param assigns assigns clause where search results are added
+  /// @return true if failure, false if success
```
feliperodri:
```suggestion
  /// @param[in]  function The function to search local static symbols in
  /// @param[out] assigns  Assigns clause where search results are added
```

src/goto-instrument/contracts/contracts.h
```diff
@@ -210,6 +210,30 @@ class code_contractst
...
+  ///
+  /// A symbol is considered a static local symbol iff:
+  /// - it has a static lifetime annotation
+  /// - its source location has a non-empty function attribute
```
feliperodri:
```suggestion
  /// A symbol is considered a static local symbol iff:
  /// - it has a static lifetime annotation;
  /// - its source location has a non-empty function attribute.
```

src/goto-instrument/contracts/contracts.h
```diff
@@ -210,6 +210,30 @@ class code_contractst
...
+  ///
+  /// Symbol occurrences are searched in:
+  /// - ASSERT/ASSUME conditions
+  /// - ASSIGN lhs and rhs
+  /// - FUNCTION_CALL function, lhs and operands
+  /// - SET_RETURN_VALUE instructions
+  /// - OTHER code.
```
feliperodri:
```suggestion
  /// Symbol occurrences are searched in:
  /// - ASSERT/ASSUME conditions;
  /// - ASSIGN lhs and rhs;
  /// - FUNCTION_CALL function, lhs and operands;
  /// - SET_RETURN_VALUE instructions;
  /// - OTHER code.
```

src/goto-instrument/contracts/contracts.h
```diff
@@ -210,6 +210,30 @@ class code_contractst
...
+  ///
+  /// The search does *not*:
+  /// - recurse into FUNCTION_CALL 
+  /// - test DECL or DEAD symbols (because local statics are never DECL'd or DEAD'd for being static).
+  /// - test attributes of GOTO, SET_RETURN_VALUE, THREAD*, ATOMIC*, THROW, CATCH  instructions
```
feliperodri:
```suggestion
  /// The search does *not*:
  /// - recurse into FUNCTION_CALL;
  /// - test DECL or DEAD symbols (because local statics are never DECL'd or DEAD'd for being static);
  /// - test attributes of GOTO, THREAD*, ATOMIC*, THROW, CATCH  instructions.
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1260,3 +1258,63 @@ bool code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
...
+  }
+
+  // filter symbols with static lifetime and non-empty function attribute
+  // and add them to the assigns clause
+  for(auto it = tmp_set.begin(); it != tmp_set.end(); it++)
+  {
+    symbolt sym = ns.lookup(*it);
+    bool is_local_static =
+      sym.is_static_lifetime && !sym.location.get_function().empty();
+    if(is_local_static)
```
```suggestion
    if(sym.is_static_lifetime && !sym.location.get_function().empty())
```

src/goto-instrument/contracts/contracts.h
feliperodri: This function doesn't have a return value.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1260,3 +1258,63 @@ bool code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
...
+
+      forall_expr(e_it, inst->call_arguments())
+      {
+        find_symbols(*e_it, tmp_set, true, false);
+      }
+    }
+
+    if(inst->is_other())
+      find_symbols(inst->get_other(), tmp_set, true, false);
+  }
```
feliperodri: Could we add more tests to cover all these cases?

feliperodri (changes requested):
Only minor requests.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1260,3 +1258,63 @@ bool code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
   }
   return fail;
 }
+
+void code_contractst::find_static_locals(
```
feliperodri: @tautschnig is there anything already available in CBMC that could help us to get all static local symbols of a given function?

SaswatPadhi: I was thinking of something simpler like:

```c++
// record all functions directly/indirectly called by foo
<some_data_structure> foo_calls = call_tree(foo);

// check if a global static variable comes from one of these functions
for(const auto &sym_pair : symbol_table)
  if(sym.is_static_lifetime && foo_calls.contains(sym.location.get_function()))
    local_statics.add(sym)
```

to find all local `static`s indirectly referred to by `foo`.

Do we have a CBMC library function that returns a call tree below a given function `foo`?

regression/contracts/assigns_enforce_statics_subfunctions/test.desc
remi-delmas-3000: I think it's safe to use subfunction
https://www.merriam-webster.com/dictionary/subroutine

feliperodri: It's fine to use subroutine, the core issue was the typo: `s/accross/across/g`.

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: the contents of this set only gets filtered for local statics in a second step

regression/contracts/assigns_enforce_statics/main.c
```diff
@@ -1,15 +1,18 @@
 #include <assert.h>
 #include <stdlib.h>
```
feliperodri: We don't need these includes here.

regression/contracts/assigns_enforce_statics_subfunctions/test1.c
```diff
@@ -0,0 +1,33 @@
+#include <assert.h>
+#include <stdlib.h>
```
feliperodri: We don't need these include here.

regression/contracts/assigns_enforce_statics_subfunctions/test2.c
```diff
@@ -0,0 +1,56 @@
+#include <stdbool.h>
```
feliperodri: We don't need this include here.

regression/contracts/assigns_enforce_statics_subfunctions/test2.desc
```diff
@@ -0,0 +1,25 @@
...
+^VERIFICATION FAILED$
+--
+^\[main\.assertion\.\d+] line 50 0 == x expecting FAILURE: SUCCESS$
+^\[main\.assertion\.\d+] line 54 old_x == x expecting FAILURE: SUCCESS$
+--
+Checks whether static local and global variables are correctly tracked
+in assigns clause verification, across subfunction calls.
+TODO add these lines once havocing of static pointers is fixed
+^\[bar\.assertion\.\d+\] line 21 i == tmp expecting FAILURE: SUCCESS$
+^\[bar\.assertion\.\d+\] line 30 \*i in range expecting FAILURE: SUCCESS$
```
feliperodri: Are we tracking this request on a GitHub issue? We should add the link here if yes.

feliperodri (changes requested):
Almost there.

regression/contracts/assigns_enforce_statics/main.c
remi-delmas-3000: fixed

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: now using the call_graph of the function under analysis to perform the detection

regression/contracts/assigns_enforce_statics_subfunctions/main.c
```diff
@@ -0,0 +1,31 @@
...
+#include <stdlib.h>
+
```
feliperodri: We still don't need these here.

regression/contracts/assigns_enforce_statics_subfunctions/test.desc
feliperodri: ?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1260,3 +1259,34 @@ bool code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
...
+  const goto_functionst &functions,
+  const irep_idt &root_function,
+  assigns_clauset &assigns)
+{
+  auto cg =
+    call_grapht::create_from_root_function(functions, root_function, true);
+  auto g = cg.get_directed_graph();
```
feliperodri: Why don't we hit this case?

src/goto-instrument/contracts/contracts.h
```diff
@@ -210,6 +209,23 @@ class code_contractst
...
+  /// and adds them to the given `assigns`.
+  ///
+  /// @param functions all functions of the goto_model
+  /// @param root_function the root function under which to search statics
+  /// @param assigns assigns clause where search results are added
+  ///
+  /// A symbol is considered a static local symbol iff:
+  /// - it has a static lifetime annotation
+  /// - its source location has a non-empty function attribute
+  /// - this function attribute is found in the call graph of the root_function
```
feliperodri:
```suggestion
  /// \brief Finds symbols declared with a static lifetime in the given
  /// `root_function` or one of the functions it calls,
  /// and adds them to the given `assigns`.
  ///
  /// @param functions     All functions of the `goto_model`.
  /// @param root_function The root function under which to search statics.
  /// @param assigns       Assigns clause where search results are added.
  ///
  /// A symbol is considered a static local symbol iff:
  /// - it has a static lifetime annotation;
  /// - its source location has a non-empty function attribute;
  /// - this function attribute is found in the call graph of the `root_function`.
```

feliperodri (approved):
LGTM modulo a few minor comments.

remi-delmas-3000:
We are now collecting symbols from the symbol table that have a static lifetime and are declare in functions occurring in the  call graph of the function under analysis

regression/contracts/assigns_enforce_statics_subfunctions/test.desc
tautschnig: Why is there an underscore here (after `bar`)?

regression/contracts/assigns_enforce_statics/main.c
```diff
@@ -1,15 +1,15 @@
-#include <assert.h>
-#include <stdlib.h>
-
-static int x;
+static int x = 0;
```
tautschnig, line 1: Random place for comment: the commit message ...
1) Could make clear that this is about code contracts;
2) Should likely have the "A symbol is ..." all moved into the body of the message;
3) Should likely talk about "called functions" instead of "subfunctions."

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -15,7 +15,10 @@ Date: February 2016
 
 #include <algorithm>
 #include <map>
+#include <util/find_symbols.h>
+#include <util/std_code.h>
```
tautschnig: Nit picking: please move the `util` includes a few lines down, where the other `util` includes are.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -853,18 +856,16 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     function,
     symbol_table);
 
-  // Adds formal parameters to write set
+  // fetch local statics of all subfunctions in the symbol_table
```
tautschnig: Much like the commit message: I don't think "subfunctions" is the right terminology.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -853,18 +856,16 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+  find_static_locals(goto_functions, function, assigns);
+
+  // Full inlining of the function body
+  goto_function_inline(goto_functions, function, ns, log.get_message_handler());
```
tautschnig: I think the comment could be more useful if it explained _why_ full inlining was necessary. (Also, will this work with recursive functions?)

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1260,3 +1259,34 @@ bool code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
...
+      auto fname = sym.location.get_function();
+      if(
+        !fname.empty() &&
+        (fname == root_function || g.get_node_index(fname).has_value()))
+      {
+        // the symbol has
+        // - a static lifetime
+        // - non empty location function
+        // - this function appears in the call graph of the function
+        assigns.add_to_write_set(sym.symbol_expr());
```
tautschnig: What about function pointers?

regression/contracts/assigns_enforce_statics_subfunctions/test.desc
remi-delmas-3000: this underscore separates two groups of options , the first group is passed to goto-instrument to perform the contract instrumentation, and the second set is passed to cbmc to analyse the instrumented contracts.

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: done

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: I'm curious too. Shouldn't we hit this on `regression/contracts/assigns_enforce_statics_subfunctions` test?

SaswatPadhi: [nit] do we need to keep `cg` around? also, may be use a slightly more descriptive name for `g` and make it `const`?

```suggestion
  const auto call_graph = call_grapht::create_from_root_function(functions, root_function, true).get_directed_graph();
```
(would need to reformat ^)

src/goto-instrument/contracts/contracts.h
```diff
@@ -210,6 +209,23 @@ class code_contractst
...
+  void find_static_locals(
```
SaswatPadhi: [nit] this function isn't just "finding" static locals, it's also adding them to `assigns`. It's worth thinking about:
1. should we make this function "pure": only find static locals and return a `set` of those?
2. if not, would it make sense to move this function into the `assigns_clauset` class as a member function?
3. if not, should we rename this to `add_static_locals_to_assigns_set`?

SaswatPadhi (commented):
Some minor comments:

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: the reason behind inlining is that in the function `f` we are instrumenting ,  a same function `g` can be called in several locations under different write sets. 
Without inlining, for each call site we would have to create a copy of g and instrument for the write set specific to this call site.
Inlining allows us to avoid having to duplicate g and instrument it specifically. 

As for inlining recursive functions, I'll implement a check to detect them and error out during inlining

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: we suppose function pointers have been removed/restricted before this instrumentation phase so that all possible target functions would appear explicitly in the call graph

tautschnig: Hmm, we should really include a `PRECONDITION` in `call_graph.cpp`'s `forall_callsites` to replace `if(function_expr.id()==ID_symbol)` by `PRECONDITION_WITH_DIAGNOSTICS(function_expr.id() == ID_symbol, "call graph computation requires function pointer removal");`.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -29,17 +30,126 @@ Date: February 2016
...
+  {
+    return missing_call_argument_warnings;
+  }
+
+  unsigned int get_missing_function_body_warnings()
+  {
+    return missing_function_body_warnings;
+  }
+
+  void print(unsigned level, const std::string &message)
```
tautschnig: Mark this one as `override`

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -29,17 +30,126 @@ Date: February 2016
...
+
+  void print(unsigned level, const jsont &json)
+  {
+    wrapped.print(level, json);
+  }
+
+  void print(unsigned level, const structured_datat &data)
+  {
+    wrapped.print(level, data);
+  }
```
tautschnig: All of these need to be marked `override`.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -29,17 +30,126 @@ Date: February 2016
...
+  {
+    parse_message(message);
+    wrapped.print(level, message, location);
+    return;
+  }
+
+  void flush(unsigned i)
+  {
+    return wrapped.flush(i);
+  }
```
tautschnig: Also to be marked `override`

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -29,17 +30,126 @@ Date: February 2016
...
+
+  ~inlining_decoratort()
+  {
+  }
```
tautschnig: Can be omitted.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -29,17 +30,126 @@ Date: February 2016
...
+    return wrapped.get_verbosity();
+  }
+
+  std::size_t get_message_count(unsigned level) const
+  {
+    return wrapped.get_message_count(level);
+  }
+
+  std::string command(unsigned i) const
+  {
```
tautschnig: These should be marked `virtual` in `util/message.h` and `override` in here.

src/goto-instrument/contracts/contracts.h
remi-delmas-3000: I moved the method to the `assigns_clauset` class.

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: renamed as required

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: `PRECONDITION` added in `call_graph.cpp`. 


Should we also modify `goto_function_inline` to warn on unresolved function pointers ?

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: This method is indeed covered when running  `regression/contracts/assigns_enforce_statics_subfunctions` otherwise the test would not pass.

SaswatPadhi: Thanks. Probably a false positive / stale result from Codecov then. Resolving this.

regression/contracts/function-calls-recursive-01/main.c
```diff
@@ -0,0 +1,23 @@
...
+  int res = sum_rec(j, 0);
+  return res;
+}
+
+int main()
+{
+  int result = sum(10);
+  __CPROVER_assert(result == 55, "result == 55");
+  return 0;
+}
```
SaswatPadhi: [nit] missing newline

SaswatPadhi (approved):
LGTM 🎉

src/goto-instrument/contracts/contracts.cpp
tautschnig: Yes, we should either warn or fail with a `PRECONDITION`.

feliperodri:
@chrisr-diffblue @martin-cs @peterschrammel could you review this PR? It's missing code owner approval.

regression/contracts/function-calls-recursive-01/main.c
feliperodri: +1

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -15,6 +15,8 @@ Date: July 2021
 
 #include <langapi/language_util.h>
 
+#include <analyses/call_graph.h>
```
SaswatPadhi, line 16: [nit] this should go above `langapi` include(s).

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -192,3 +194,32 @@ void havoc_assigns_targetst::append_havoc_code_for_expr(
...
+      {
+        // We found a symbol with
+        // - a static lifetime
+        // - non empty location function
+        // - this function appears in the call graph of the function
+        add_to_write_set(sym.symbol_expr());
+      }
+    }
+  }
+  return;
```
SaswatPadhi: Redundant return

SaswatPadhi (commented):
Couple of minor nitpicks but nothing that would block the merge.

src/goto-instrument/contracts/assigns.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/assigns.cpp

tautschnig (approved):

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -29,6 +29,7 @@ Date: February 2016
```
peterschrammel, line 29: This is file is getting way to large. Consider refactoring/splitting in future.

peterschrammel (approved):
Nothing blocking, just a few remarks for future work:
That's quite a lot in a single commit. It could have been split into a series of commits:
- introduce decorator
- add new tests
- add static local detection
- other changes 1
- other changes 2
- other changes 3
- activate/modify tests (separately or with the corresponding commit that makes the test pass)

each of which with a 'why' description in the commit body to make the intent of the PR clearer and easier and faster to review.

PR #6448, opened 2021-11-09
Byte updates using bit fields require extension to full bytes

tautschnig:
A byte update with an update value that has a size that's not a multiple
of bytes must not overwrite bits that are not in the update value. We
previously generated extractbits expressions that would go out of
bounds, which tripped up the SMT solvers. No longer generating these
out-of-bounds expressions makes the SMT solvers happy on the Promotion3
test, which exercised this feature.

### DIFF
--- a/src/goto-programs/builtin_functions.cpp
+++ b/src/goto-programs/builtin_functions.cpp
@@ -22,6 +22,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/prefix.h>
 #include <util/rational.h>
 #include <util/rational_tools.h>
+#include <util/simplify_expr.h>
 #include <util/symbol_table.h>
 
 #include <langapi/language_util.h>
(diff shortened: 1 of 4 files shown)


src/goto-programs/builtin_functions.cpp
```diff
@@ -699,7 +700,7 @@ void goto_convertt::do_havoc_slice(
   t->source_location_nonconst().set_comment(
     "assertion havoc_slice " + from_expr(ns, identifier, ok_expr));
 
-  const array_typet array_type(char_type(), arguments[1]);
+  const array_typet array_type(char_type(), simplify_expr(arguments[1], ns));
```
martin-cs, line 703: I get that this might be the right thing to do and that it fixes a bug but this doesn't immediately seem linked to the description of the PR.

martin-cs (approved):
Looks as expected.  Fewer tests marked as buggy is always good.

src/goto-programs/builtin_functions.cpp
tautschnig: Thanks, addressed in two ways: 1) I've moved this change to a commit of its own; 2) I've added a comment explaining when a `DATA_INVARIANT` newly introduced here could fail (which really was what made me detect the above).

peterschrammel (approved):

src/goto-programs/builtin_functions.cpp
martin-cs: Amazing!  Thank you.

PR #6453, opened 2021-11-11
Update documentation on quantifiers

### LINKED ISSUE #6449: Any way to specify the length to compare in `__CPROVER_array_equal`?
CBMC version: 5.43.0
Operating system: Debian

`_Bool __CPROVER_array_equal(const void array1[], const void array2[])` takes two arrays and returns true if the values stored in the arrays are equal. I'm wondering if it's possible to add a third argument that allows us to specify how many elements to compare for equality instead of comparing the whole array?

This would help to create models for functions that are similar to `memcmp`, but where we only care about whether the result is zero or non-zero (as opposed to whether it's positive or negative), without using iteration.

TGWDB:
Quantifiers are now broadly supported and work, so documentation
should reflect this.

Fixes #6449

### DIFF
--- a/doc/cprover-manual/cbmc-assertions.md
+++ b/doc/cprover-manual/cbmc-assertions.md
@@ -87,7 +87,7 @@ postconditions to create a function contract, which can be used for
 modular verification.
 
 
-Future CPROVER releases will support explicit quantifiers with a syntax
+CPROVER also supports explicit quantifiers with a syntax
 that resembles Spec\#:
 
 ```C
@@ -95,3 +95,11 @@ __CPROVER_forall { *type identifier* ; *expression* }
 __CPROVER_exists { *type identifier* ; *expression* }
 ```
 
+For example, the following can be used to check equality of
+two arrays `array` and `array2` to a bound `n`.
+```C
+assert(__CPROVER_forall { int i ;
+            (0 <= i && i < n) ==> arrary1[i] == array2[i]
+                     } );
+```
+Usage of `__CPROVER_exists` is similar.


thomasspriggs (approved):
It might be worth linking to the quantifiers document for details of the limitations of the quantifier support.

doc/cprover-manual/cbmc-assertions.md
```diff
@@ -87,11 +87,19 @@ postconditions to create a function contract, which can be used for
...
 ```C
 __CPROVER_forall { *type identifier* ; *expression* }
 __CPROVER_exists { *type identifier* ; *expression* }
 ```
 
+For example, the following can be used to check equality of
+two arrays `array` and `array2` to a bound `n`.
+```C
+assert(__CPROVER_forall { int i ;
+            (0 <= i && i < n) ==> arrary1[i] == array2[i]
```
feliperodri, line 102: `s/arrary1/array/g`

doc/cprover-manual/cbmc-assertions.md
```diff
@@ -87,11 +87,19 @@ postconditions to create a function contract, which can be used for
...
+                     } );
+```
+Usage of `__CPROVER_exists` is similar.
```
feliperodri, line 105: We already have documentation for quantifiers, take a look at this [section](https://github.com/diffblue/cbmc/blob/develop/doc/cprover-manual/contracts-quantifiers.md) of the contracts manual. could you add a point in here for this page too?

cc. @SaswatPadhi

feliperodri (approved):

doc/cprover-manual/cbmc-assertions.md
thomasspriggs: I think it would be better to use links rather than duplicating the same documentation in multiple places.

doc/cprover-manual/cbmc-assertions.md
```diff
@@ -87,11 +87,19 @@ postconditions to create a function contract, which can be used for
...
+CPROVER also supports explicit quantifiers with a syntax
 that resembles Spec\#:
 
 ```C
 __CPROVER_forall { *type identifier* ; *expression* }
 __CPROVER_exists { *type identifier* ; *expression* }
 ```
 
+For example, the following can be used to check equality of
+two arrays `array` and `array2` to a bound `n`.
```
danielsn, line 99: is there a limit on the bound `n`? I recall at one point it only worked up to 100?

feliperodri:
@TGWDB shall we close this PR or this update in the quantifier documentation is still necessary? cc. @SaswatPadhi

SaswatPadhi:
I agree with what @feliperodri and @thomasspriggs are suggesting -- we should add a link to `doc/cprover-manual/contracts-quantifiers.md`, since most of this material has been explained there.

I am also fine with renaming `doc/cprover-manual/contracts-quantifiers.md` to `doc/cprover-manual/quantifiers.md` since there's nothing in that doc that's specific to code contracts (except for the examples, which could be updated to also include some assertion and assumption examples). And after the renaming, we should just add pointers in the assertions docs and contracts docs to this quantifiers doc.

PR #6458, opened 2021-11-16
Add validation of smt parse trees to construct smt_responset instances.

thomasspriggs:
This PR adds validation of smt parse trees to construct smt_responset instances.


src/solvers/smt2_incremental/smt_response_validation.cpp
```diff
@@ -37,5 +37,14 @@ template class response_or_errort<smt_responset>;
 
 response_or_errort<smt_responset> validate_smt_response(const irept &parse_tree)
 {
+  if(parse_tree.id() == "sat")
```
NlightNFotis, line 309: ⛏️ Why not `switch (parse_tree.id()) { case "sat": }`. Might make it a bit more readable?

NlightNFotis (commented):
LGTM overall 👍🏻

src/solvers/smt2_incremental/smt_response_validation.cpp
thomasspriggs: Because you can't switch on a string in C++.

NlightNFotis: 😱

NlightNFotis (approved):

src/solvers/smt2_incremental/smt_response_validation.cpp
```diff
@@ -0,0 +1,327 @@
...
+  // unexpected in the parse tree is now considered to be an invalid response.
+  if(parse_tree.get_sub().size() == 1)
+  {
+    return {response_or_errort<smt_responset>{
+      "Error response is missing the error message."}};
+  }
+  if(parse_tree.get_sub().size() > 2)
+  {
+    return {response_or_errort<smt_responset>{
+      "Error response has multiple error messages."}};
```
TGWDB: Can we do something nicer here? Maybe gather all the error messages into a multiline string?

src/solvers/smt2_incremental/smt_response_validation.cpp
TGWDB: But you can switch on a/the first character... so if you want a big nested tree of case statements... ;)

TGWDB (approved):

src/solvers/smt2_incremental/smt_response_validation.cpp
thomasspriggs: I have done something nicer here. The original parse tree is now included in this error message, as instances where the messages are sub trees could also fall into this case.

PR #6459, opened 2021-11-16
Disable unnecessary pointer checks, enable necessary pointer checks i…

remi-delmas-3000:
This PR fixes some performance issues with assigns clause checking for function and loop contracts.

The problem was that the goto variables and instructions generated to instrument the checks were themselves re-instrumented with automatic VCCs in subsequent passes, causing an explosion in the number of VCCs (with unwanted VCCs representing up to 85% of the total VCCs in some problems).

Instructions performing CAR inclusion checks are read-only logical expressions whose evaluation cannot fail so we can safely disable automatic checks on them. 

CAR snapshotting instructions however perform some pointer arithmetic and need to be ultimately checked for pointer overflow.

In this PR:
- we disable all pointer checks on CAR inclusion checking instructions 
- we enable pointer overflow checks on CAR snapshot instructions

The disable pragmas guard against reinstrumentation in later passes, and the enable pragmas make sure necessary checks will eventually be instantiated before analysis.

### DIFF
--- a/regression/contracts/assigns_enforce_free_dead/test.desc
+++ b/regression/contracts/assigns_enforce_free_dead/test.desc
@@ -29,7 +29,9 @@ main.c
 ^\[foo.\d+\] line \d+ Check that y is assignable: FAILURE$
 ^\[foo.\d+\] line \d+ Check that z is assignable: FAILURE$
 ^\[foo.\d+\] line \d+ Check that POINTER_OBJECT\(\(void \*\)z\) is assignable: FAILURE$
-^.*tmp_cc\$\d+.*: FAILURE$
+^.*__car_valid.*: FAILURE$
+^.*__car_lb.*: FAILURE$
+^.*__car_ub.*: FAILURE$
 --
 Checks that invalidated CARs are correctly tracked on `free` and `DEAD`.
 
@@ -40,4 +42,4 @@ We also check (using positive regex matches above) that
 `**p` should be assignable in one case and should not be assignable in the other.
 
 Finally, we check that there should be no "internal" assertion failures
-on `tmp_cc` variables used to track CARs.
+on `__car_valid`, `__car_ub`, `__car_lb` variables used to track CARs.
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -52,6 +52,7 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
 }
 
 const symbolt assigns_clauset::conditional_address_ranget::generate_new_symbol(
+  const std::string &prefix,
   const typet &type,
   const source_locationt &location) const
 {
@@ -59,7 +60,8 @@ const symbolt assigns_clauset::conditional_address_ranget::generate_new_symbol(
     type,
     location,
     parent.symbol_table.lookup_ref(parent.function_name).mode,
-    parent.symbol_table);
+    parent.symbol_table,
+    prefix);
 }
 
 assigns_clauset::conditional_address_ranget::conditional_address_ranget(
@@ -70,11 +72,13 @@ assigns_clauset::conditional_address_ranget::conditional_address_ranget(
     parent(parent),
     slice(normalize_to_slice(expr, parent.ns)),
     validity_condition_var(
-      generate_new_symbol(bool_typet(), location).symbol_expr()),
+      generate_new_symbol("__car_valid", bool_typet(), location).symbol_expr()),
     lower_bound_address_var(
-      generate_new_symbol(slice.first.type(), location).symbol_expr()),
+      generate_new_symbol("__car_lb", slice.first.type(), location)
+        .symbol_expr()),
     upper_bound_address_var(
-      generate_new_symbol(slice.first.type(), location).symbol_expr())
+      generate_new_symbol("__car_ub", slice.first.type(), location)
+        .symbol_expr())
 {
 }
 
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -12,6 +12,7 @@ Date: July 2021
 /// Specify write set in code contracts
 
 #include "assigns.h"
+#include "utils.h"
 
 #include <analyses/call_graph.h>
 
@@ -87,41 +88,54 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
   const
 {
   goto_programt instructions;
+  // adding pragmas to the location to selectively disable checks
+  // where it is sound to do so
+  source_locationt location_no_checks = location;
+  disable_pointer_checks(location_no_checks);
 
-  instructions.add(goto_programt::make_decl(validity_condition_var, location));
-  instructions.add(goto_programt::make_decl(lower_bound_address_var, location));
-  instructions.add(goto_programt::make_decl(upper_bound_address_var, location));
+  instructions.add(
+    goto_programt::make_decl(validity_condition_var, location_no_checks));
+  instructions.add(
+    goto_programt::make_decl(lower_bound_address_var, location_no_checks));
+  instructions.add(
+    goto_programt::make_decl(upper_bound_address_var, location_no_checks));
 
   instructions.add(goto_programt::make_assignment(
     lower_bound_address_var,
     null_pointer_exprt{to_pointer_type(slice.first.type())},
-    location));
+    location_no_checks));
   instructions.add(goto_programt::make_assignment(
     upper_bound_address_var,
     null_pointer_exprt{to_pointer_type(slice.first.type())},
-    location));
+    location_no_checks));
 
   goto_programt skip_program;
-  const auto skip_target = skip_program.add(goto_programt::make_skip(location));
+  const auto skip_target =
+    skip_program.add(goto_programt::make_skip(location_no_checks));
 
   const auto validity_check_expr =
     and_exprt{all_dereferences_are_valid(source_expr, parent.ns),
               w_ok_exprt{slice.first, slice.second}};
   instructions.add(goto_programt::make_assignment(
-    validity_condition_var, validity_check_expr, location));
+    validity_condition_var, validity_check_expr, location_no_checks));
 
   instructions.add(goto_programt::make_goto(
-    skip_target, not_exprt{validity_condition_var}, location));
+    skip_target, not_exprt{validity_condition_var}, location_no_checks));
 
   instructions.add(goto_programt::make_assignment(
-    lower_bound_address_var, slice.first, location));
+    lower_bound_address_var, slice.first, location_no_checks));
+
+  // the computation of the CAR upper bound can overflow into the object ID bits
+  // of the pointer with very large allocation sizes.
+  // We enable pointer overflow checks to detect such cases.
+  source_locationt location_overflow_check = location;
+  location_overflow_check.add_pragma("enable:pointer-overflow-check");
 
   instructions.add(goto_programt::make_assignment(
     upper_bound_address_var,
     minus_exprt{plus_exprt{slice.first, slice.second},
                 from_integer(1, slice.second.type())},
-    location));
-
+    location_overflow_check));
   instructions.destructive_append(skip_program);
 
   // The assignments above are only performed on locally defined temporaries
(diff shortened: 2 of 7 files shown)


src/goto-instrument/contracts/utils.h
```diff
@@ -116,6 +116,8 @@ const symbolt &new_tmp_symbol(
   const typet &type,
   const source_locationt &location,
   const irep_idt &mode,
-  symbol_table_baset &symtab);
+  symbol_table_baset &symtab,
+  std::string suffix = "tmp_cc");
```
jimgrundy: This new parameter needs a doxygen comment above

jimgrundy (dismissed):

src/goto-instrument/contracts/utils.cpp
```diff
@@ -142,12 +142,13 @@ const symbolt &new_tmp_symbol(
...
 {
   return get_fresh_aux_symbol(
     type,
-    id2string(location.get_function()) + "::tmp_cc",
-    "tmp_cc",
+    id2string(location.get_function()) + "::",
+    suffix,
     location,
     mode,
     symtab);
```
tautschnig: While at it: should this code also be modified to have those symbols show up in counterexample traces? I found myself having a hard time understanding counterexamples for none of the assignments to these variables show up, but failing assertions then refer to them. The change required to facilitate this:
````c++
symbolt &new_symbol = get_fresh_aux_symbol(...);
new_symbol.is_auxiliary = false;
return new_symbol;
````

File: src/goto-instrument/contracts/assigns.cpp [hunk 3]
```diff
@@ -52,14 +52,16 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
 }
 
 const symbolt assigns_clauset::conditional_address_ranget::generate_new_symbol(
+  const std::string &prefix,
```
tautschnig, line 56: Random place for comment, and primarily for future reference: I would like to see commits like this be a pull request of their own. Changing the names of CAR instrumentation variables really has nothing to do with disabling selected checks.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -911,6 +911,9 @@ void code_contractst::instrument_call_statement(
   }
   else if(callee_name == "free")
   {
+    source_locationt &location_no_checks =
+      instruction_it->source_location_nonconst();
```
tautschnig: Do you really want or need to take a reference here? The reference will ensure (but is that desired?) that checks are disabled on the function call. Otherwise, I would have expected that you'd be taking a copy here, at which point you want to use `source_location()` instead of `source_location_nonconst()`.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1175,6 +1175,10 @@ void code_contractst::check_frame_conditions(
     else if(instruction_it->is_dead())
     {
       const auto &symbol = instruction_it->dead_symbol();
+      source_locationt &location_no_checks =
+        instruction_it->source_location_nonconst();
+      disable_pointer_checks(location_no_checks);
```
tautschnig, line 1178: As above: is the reference intentional?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1228,10 +1232,13 @@ code_contractst::add_inclusion_check(
     program, instruction_it, snapshot_instructions);
 
   goto_programt assertion;
-  assertion.add(goto_programt::make_assertion(
-    assigns.generate_inclusion_check(car), instruction_it->source_location()));
-  assertion.instructions.back().source_location_nonconst().set_comment(
+  source_locationt location_no_checks =
+    instruction_it->source_location_nonconst();
```
tautschnig, line 1234: Certainly `source_location()` instead of `source_location_nonconst()`.

regression/contracts/assigns_enforce_free_dead/test.desc
```diff
@@ -40,4 +42,4 @@ We also check (using positive regex matches above) that
 `**p` should be assignable in one case and should not be assignable in the other.
 
 Finally, we check that there should be no "internal" assertion failures
-on `tmp_cc` variables used to track CARs.
+on `__car_valid`, `__car_ub`, `__car_lb` variables used to track CARs.
```
tautschnig, line 45: Changes in this file should be folded into the commit introducing the new names.

regression/contracts/assigns_enforce_havoc_object/functions.txt
```diff
@@ -0,0 +1 @@
+Reading GOTO program from '__CPROVER_file_local_s2n_hash_c_s2n_evp_hash_reset_with_contracts_harness.c.goto'
```
tautschnig: What is the meaning of this file?

tautschnig (approved):
Generally ok with this, but the last commit introduces/modifies various tests and it is not obvious which tests are just additional tests, or which code changes these tests relate to.

regression/contracts/assigns_enforce_havoc_object/functions.txt
remi-delmas-3000: this is a debugging session leftover, removing it

src/goto-instrument/contracts/utils.cpp
remi-delmas-3000: Adding a new function argument 'is_auxiliary' to control that symbol per symbol.

src/goto-instrument/contracts/contracts.cpp

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/utils.h
remi-delmas-3000: fixed

remi-delmas-3000:
All review comments were addressed

regression/contracts/assigns_enforce_free_dead/test.desc
```diff
@@ -29,7 +29,9 @@ main.c
 ^\[foo.\d+\] line \d+ Check that y is assignable: FAILURE$
 ^\[foo.\d+\] line \d+ Check that z is assignable: FAILURE$
 ^\[foo.\d+\] line \d+ Check that POINTER_OBJECT\(\(void \*\)z\) is assignable: FAILURE$
-^.*tmp_cc\$\d+.*: FAILURE$
+^.*__car_valid.*: FAILURE$
+^.*__car_lb.*: FAILURE$
+^.*__car_ub.*: FAILURE$
```
feliperodri, line 34: nitpick: I'd prefer even more readable names for users: `__car_valid`, `__car_lower_bound`, and `__car_upper_bound`. However, the fact that these variables exist should be transparent to users. Why would this variables appear as failures? I guess you want to make sure here there are no memory safety errors in any operation using these variables, is that correct?

src/goto-instrument/contracts/assigns.h
```diff
@@ -60,8 +60,10 @@ class assigns_clauset
     const exprt
     generate_unsafe_inclusion_check(const conditional_address_ranget &) const;
 
-    const symbolt
-    generate_new_symbol(const typet &, const source_locationt &) const;
+    const symbolt generate_new_symbol(
+      const std::string &prefix,
+      const typet &,
+      const source_locationt &) const;
```
feliperodri, line 66: For consistency, add the name of the remaining parameters or remove `prefix`.

src/goto-instrument/contracts/utils.cpp
```diff
@@ -142,13 +142,18 @@ const symbolt &new_tmp_symbol(
   const typet &type,
   const source_locationt &location,
   const irep_idt &mode,
-  symbol_table_baset &symtab)
+  symbol_table_baset &symtab,
+  std::string suffix,
```
feliperodri, line 146:
```suggestion
  symbol_table_baset &symtab,
  std::string &suffix,
```

src/goto-instrument/contracts/utils.h
```diff
@@ -105,17 +105,21 @@ void insert_before_swap_and_advance(
...
 
-/// \brief Adds a new temporary symbol to the symbol table.
+/// \brief Adds a fresh and uniquely named symbol to the symbol table.
 ///
 /// \param type: The type of the new symbol.
 /// \param location: The source location for the new symbol.
 /// \param mode: The mode for the new symbol, e.g. ID_C, ID_java.
 /// \param symtab: The symbol table to which the new symbol is to be added.
+/// \param suffix: Suffix to use to generate the unique name
+/// \param is_auxiliary: Do not print symbol in traces if true (default = false)
```
feliperodri:
```suggestion
/// \param suffix: Suffix to use to generate the unique name.
/// \param is_auxiliary: Do not print symbol in traces if true (default = false).
```

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -87,41 +88,54 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
   const
 {
   goto_programt instructions;
-
-  instructions.add(goto_programt::make_decl(validity_condition_var, location));
-  instructions.add(goto_programt::make_decl(lower_bound_address_var, location));
-  instructions.add(goto_programt::make_decl(upper_bound_address_var, location));
+  // adding pragmas to the location to selectively disable checks
+  // where it is sound to do so
```
feliperodri, line 92: I think a brief explanation of "why" is sound to disable checks here would be more helpful, since one can easily infer this comment from the code.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -87,41 +88,54 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
...
 
   instructions.add(goto_programt::make_goto(
-    skip_target, not_exprt{validity_condition_var}, location));
+    skip_target, not_exprt{validity_condition_var}, location_no_checks));
 
   instructions.add(goto_programt::make_assignment(
-    lower_bound_address_var, slice.first, location));
+    lower_bound_address_var, slice.first, location_no_checks));
+
+  // the computation of the CAR upper bound can overflow into the object ID bits
```
feliperodri, line 128:
```suggestion
  // The computation of the CAR upper bound can overflow into the object ID bits
```

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -87,41 +88,54 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
...
+  // of the pointer with very large allocation sizes.
+  // We enable pointer overflow checks to detect such cases.
```
feliperodri, line 130:
```suggestion
  // We always enable pointer overflow checks to detect such cases.
```

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -134,14 +148,16 @@ const exprt
 assigns_clauset::conditional_address_ranget::generate_unsafe_inclusion_check(
   const conditional_address_ranget &lhs) const
 {
+  // remark: both lb and ub are derived from the same object so checking
+  // same_object(upper_bound_address_var, lhs.upper_bound_address_var)
+  // is redundant
```
feliperodri, line 153: Nice!
```suggestion
  // Both lb and ub are derived from the same object so checking
  // same_object(upper_bound_address_var, lhs.upper_bound_address_var)
  // is redundant
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -911,6 +911,8 @@ void code_contractst::instrument_call_statement(
   }
   else if(callee_name == "free")
   {
+    source_locationt location_no_checks = instruction_it->source_location();
+    disable_pointer_checks(location_no_checks);
```
feliperodri, line 915: We could add a comment here again explaining "why" is safe to always remove pointer checks in this location.

src/goto-instrument/contracts/contracts.cpp
feliperodri: We could add a comment here again explaining "why" is safe to always remove pointer checks in this location.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1228,10 +1230,13 @@ code_contractst::add_inclusion_check(
     program, instruction_it, snapshot_instructions);
 
   goto_programt assertion;
-  assertion.add(goto_programt::make_assertion(
-    assigns.generate_inclusion_check(car), instruction_it->source_location()));
-  assertion.instructions.back().source_location_nonconst().set_comment(
+  source_locationt location_no_checks =
+    instruction_it->source_location_nonconst();
+  disable_pointer_checks(location_no_checks);
```
feliperodri, line 1235: We could add a comment here again explaining "why" is safe to always remove pointer checks in this location.

feliperodri (commented):
I was a bit too late to the party, but I'll leave the comments here to be addressed on another PR.

Issue #5965, opened 2021-03-22
Feature request: specify input arguments of verified programs


vmihalko:
Hi,
We are (at RedHat) developing a tool for running dynamic analysis on RPM packages fully automatically. For this purpose, we want to use the open-source tools like Divine, Symbiotic, CBMC, Valgrind and others.
Because the "size" of the verified model is often too large, cbmc is running for too long (or forever). We would appreciate if CBMC had an option to enter specific input arguments for verified programs e.g. `cbmc print.c --argv 'Hello world'`. It would simplify the verification process a lot.
Do you plan to add such an option to CBMC in the near future?

martin-cs:
Glad to hear you are finding it useful.  I'm not sure I would class it as a dynamic analysis tool in the same way as Valgrind though.

You can pass arguments, sort of using two different approaches.  -DMACRO=VALUE allows you to set preprocessor macros.  The other option is to generate some C:

```
$ cat main_with_args.c
int main_with_args(void) {
  char *name="print";
  char *argument="Hello world";
  int argc = 2;
  char *argv[argc] = {name, argument};
  int return_value = main(argc, argv);
  assert(return_value == 0);
  return;
}
```

Then run `cbmc print.c main_with_args.c --function main_with_args`.   You might also want to try out `--depth` to limit the execution. HTH.

As far I know there is no plan to add the kind of option that you describe but it shouldn't be too hard to do.

vmihalko:
Thank you for your quick reply.

The things you have mentioned are possible if it were a one-time affair, but the whole process of automatic analysis of RPM packages is quite complex and uses several different programs and scripts. Therefore, I think that the implementation of some of the methods you mentioned is unnecessarily demanding (and opens up room for bugs).

If the implementation of such an option for cbmc isn't too hard to do, then let's do it. 
Would you mind implementing it, or should I try it myself? And in case I won't be successful, I kindly ask for your help or advice.

martin-cs:
I'll see if I can make the time to have a go.  Before I do though, could you clarify what you mean by dynamic analysis?  CBMC is really not the same kind of tool as valgrind and if you try to use it in the same way, your effort / result curve will likely not be very good.

vmihalko:
[...]

PR #55, opened 2016-04-04
goto-instrument --model-argc-argv: Explicitly initialise argc, argv

tautschnig:
Programs using argc and argv need to have these set up properly. This patch introduces an instrumentation achieving this.

PR #6462, opened 2021-11-17
goto-instrument --add-cmd-line-arg: Explicitly set argv (and argc) to user-specified values

vmihalko:
* Failed to find anything about [`--model-argc-argv`](https://github.com/diffblue/cbmc/pull/55) in `doc/cprover-manual/` so I don't know where to document `--add-cmd-line-arg`

This PR is related to https://github.com/diffblue/cbmc/issues/5965.


src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1842,7 +1870,7 @@ void goto_instrument_parse_optionst::help()
     HELP_REMOVE_CALLS_NO_BODY
     HELP_REMOVE_CONST_FUNCTION_POINTERS
     " --add-library                add models of C library functions\n"
-    " --model-argc-argv <n>        model up to <n> command line arguments\n"
+    HELP_ARGC_ARGV
```
martin-cs: Don't worry about fixing this for this particular PR but in general we try to make sure that the code compiles after every commit.  That way we can use `git bisect` more easily.

martin-cs (approved):
I am fine with this but I would like a second opinion from someone who wasn't involved in https://github.com/diffblue/cbmc/issues/5965

@peterschrammel ?

vmihalko:
I just discovered a problem. When arg string has prefix "--" or "-" (e.g. `--add-cmd-line-arg "--help"`) cbmc is interpreting it as another option, but it should be value for option...

regression/goto-instrument/argc-argv2/test2.desc
```diff
@@ -0,0 +1,8 @@
+CORE
+main.c
+--model-argc-argv 2
+^EXIT=10$
+^SIGNAL=0$
+^VERIFICATION FAILED$
+--
+^warning: ignoring
```
TGWDB: This test only tests that the `--model-argc-argv 2` is accepted, I don't think it tests the actual value `2` is used anywhere (this file fails without the flag).

src/config.inc
```diff
@@ -76,7 +76,7 @@ endif
 OSX_IDENTITY="Developer ID Application: Daniel Kroening"
 
 # Detailed version information
-CBMC_VERSION = 5.43.0
+CBMC_VERSION = 5.44.0
```
TGWDB: This should not be in this commit (but it's not important).

TGWDB (commented):
I'd like to see some more tests about different kinds of inputs/cases/failures as the current tests are very limited.

Also more explanation/documentation of how these can/cannot be combined, since it appears like `--model-argc-argv` and `--add-cmd-line-arg` are mutually exclusive. Which in turns makes me wonder if someone would want to specify unknown command line arguments for some positions (e.g. I want to test my flag, but with any value rather than specifying the value).

chrisr-diffblue:
There seems to be four commits that should not be part of this PR (commits authored by others, touching unrelated code such as the version number bump...) - I wonder if a rebase has gone wrong at some point? It would be worth a fresh rebase on top of the latest develop, and if those extra commits are still hanging around at that point, explicitly drop them so that this PR contains only your changes for this feature. That would also make it easier to review :-)

PR #6465, opened 2021-11-18
Replace __CPROVER_malloc_is_new_array by an uninterpreted function

tautschnig:
With this encoding we no longer non-deterministically track just one
such array. This enables the use in assumptions. Also, we will only
incur a cost in the encoding when actually using new/delete instead of
always having a global variable, even when unused.


tautschnig:
Closing as there are other PRs that more properly yield the desired improvements: #6506 introduces (unbounded) arrays, and #6590 ensures there only is a cost when the symbol is really used.

PR #6450, opened 2021-11-10
Adding `#pragma CPROVER enable "named-check"`

remi-delmas-3000:
This PR adds the possibility to selectively activate checks :
- on blocks of C code using `#pragma CPROVER enable "named-check"`
- to individual GOTO instructions and expressions using `source_location.add_pragma("enable:named-check")`

It is possible to enable/disable a check, to push a level on the stack and override the check status using another pragma enable/disable pragma.
Attempts to enable and disable a same check at the same stack level are detected and cause a PARSING_ERROR in the C front end and a INVARIANT violation in `goto_check`.

Modifications:
- the collection of allocations on the whole goto program using `goto_checkt::collect_allocation` is now always on
- a `messaget log` attribute is added to `goto_checkt` to be able to log warnings in case a same check is enabled/disabled more than once on a given GOTO instruction
- the `flag_resett` class is extended to detect multiple 'set_flag' operations on a same flag
- the ansi-c lexer/parser is modified to accept `#pragma CPROVER enable "named check"` and flag shadowing
- tests are added in `regression/cbmc/pragma_cprover_enable*` to test the feature

This feature is needed as part of the function and loop contracts verification, to be able to selectively activate checks on  instructions used for assigns-clause checking, in order to guarantee the soundness of the analysis, even if these checks are not globally activated in the current or ulterior `goto-instrument` or `cbmc` passes

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- [x] The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ ] My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- [x] White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #6471, opened 2021-11-23
Prevent multiple instrumentation of a same check and instrumentation of instrumentation assertions

remi-delmas-3000:
This PR addresses the following issues:

1) Running `goto-instrument` multiple times with a same named-check switch creates redundant assertions:

```
goto-cc test.c -o test.o
goto-instrument --pointer-check test.o test1.o
goto-instrument --pointer-check test1.o test2.o
goto-instrument --pointer-check test2.o test3.o
// 3x pointer-check assertions generated  in test3.o
```

2) Applying `--pointer-overflow-check` followed by `--unsigned-overflow-check` in different `goto-instrument` invocations  results in the `pointer-overflow-check` assertions to be instrumented for `unsigned-overflow-check`, and causes systematic errors.

```C
// overflow.c
#include <stdlib.h>
const int N = 100;
int main()
{
  int *buf = malloc(N * sizeof(*buf));
  char *lb = (((char *)buf) - __CPROVER_POINTER_OFFSET(buf));
  char *ub = (((char *)buf) - __CPROVER_POINTER_OFFSET(buf)) + __CPROVER_OBJECT_SIZE(buf) - 1;
}
```

Running these passes

```bash
goto-cc overflow.c o overflow.goto
goto-instrument --pointer-overflow-check overflow.goto check1.goto
goto-instrument --unsigned-overflow-check check1.goto check2.goto
cbmc check2.goto
```
results in this systematic error

```bash
[main.overflow.2] line 12 arithmetic overflow on unsigned + in (unsigned long int)(signed long int)OBJECT_SIZE(buf) + 18446744073709551615ul: FAILURE
```

The PR adds the following in `goto_check`:

- for 1): when an instruction is processed for a given `named-check`, either because the command line switch was used or because an "enable" pragma was added by the user in the source program, we add a pragma `"checked:named-check"` to the instruction. These pragmas get saved to goto binary files and persist across tool invocations. We check for the presence of `"checked:named-check"` pragmas to ensure check assertions are only generated once.
- for 2): we add `"disable:named-check"` pragmas for all named checks to all assertions generated by `goto_checkt::add_guarded_assertion`. This will prevent these check assertions to be instrumented by other checks in future invocations of `goto-instrument` or `cbmc`.

The "checked" pragmas are reserved for internal use and are not exposed to the user in the front-end.

- [x ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [x ] My PR is restricted to a single feature or bugfix.


regression/cbmc/pragma_cprover_enable_all/test.desc
```diff
@@ -0,0 +1,27 @@
+CORE broken-smt-backend
```
tautschnig: Just curious: any idea why this is failing with the SMT back-end?

regression/cbmc/pragma_cprover_enable_all/test.desc
```diff
@@ -0,0 +1,27 @@
...
+^\[main\.pointer_primitives\.\d+\] line 40 pointer outside object bounds in POINTER_OBJECT\(const void \*\)q\): FAILURE
+^\[main\.pointer_arithmetic\.\d+\] line 41 pointer arithmetic: pointer outside object bounds in p \+ 2000000000000(l|ll): FAILURE
+^\[main\.NaN\.\d+\] line 47 NaN on / in x / den: FAILURE
+^\[main\.division-by-zero\.\d+\] line 47 division by zero in x / den: FAILURE
+^\[main\.overflow\.\d+\] line 47 arithmetic overflow on floating-point division in x / den: FAILURE
+^\[main\.enum-range-check\.\d+\] line 48 enum range check in \(ABC\)10: FAILURE
+^\[main\.overflow\.\d+\] line 49 arithmetic overflow on signed type conversion in \(char\)\(signed int\)i \+ 1\): FAILURE
+^\[main\.overflow\.\d+\] line 50 arithmetic overflow on signed \+ in j \+ 1: FAILURE
+--
+This test uses all possible named-checks to maximize coverage.
```
tautschnig: Nit pick: newline at end of file.

src/analyses/goto_check.cpp
```diff
@@ -1851,30 +1858,64 @@ optionalt<exprt> goto_checkt::rw_ok_check(exprt expr)
...
   }
 
-  /// Restore the values of all flags that have been modified via `set_flag`.
+  /// \brief Sets the given flag to false, overriding any previous value.
+  ///
+  /// Contrary to \ref set_flag, does not trigger an INVARIANT
+  /// if the flag was already seen before.
+  void disable(bool &flag)
+  {
+    if(flag != false)
```
tautschnig: `if(!flag)` would seem more idiomatic.

src/analyses/goto_check.cpp
```diff
@@ -1897,39 +1938,96 @@ void goto_checkt::goto_check(
...
+        resetter.set_flag(enable_nan_check, false, d.first);
       else if(d.first == "disable:pointer-primitive-check")
-        flag_resetter.set_flag(enable_pointer_primitive_check, false);
+        resetter.set_flag(enable_pointer_primitive_check, false, d.first);
+      else if(d.first == "enable:bounds-check")
+        resetter.set_flag(enable_bounds_check, true, d.first);
+      else if(d.first == "enable:pointer-check")
+        resetter.set_flag(enable_pointer_check, true, d.first);
+      else if(d.first == "enable:memory_leak-check")
+        resetter.set_flag(enable_memory_leak_check, true, d.first);
```
tautschnig: Randomly placing my comment in this block of code: I think we should now look at refactoring this piece of code, perhaps as follows:
1) string-split `id2string(d.first)` on `:`
2) parse the first component into an enum { DISABLE, ENABLED, CHECKED }
3) lookup the second component in a map that is initialised in the constructor (the map will map strings or `irep_idt`to `bool*` or `bool&`)
4) perform a `set_flag` operation on the result of that lookup, directed by the enum value.

As is, this will become a maintenance nightmare.

tautschnig (commented):
Apart from comments as below: can this change please be turned into multiple commits as follows:
- Introduce a message handler to `goto_check`. [no change in behaviour expected, no tests required]
- Changes to the `ansi_c_parsert` API. [no change in behaviour expected, no tests required]
- Adding support for `pragma CPROVER enable` [new feature, includes documentation updates and tests]
- Changes to flag (re)setting [no change in behaviour expected, no tests required, could perhaps be done before the above commit.]
- Support for "checked:..." [tests expected as this avoids redundant instrumentation]
- Adding "disable:..." [tests expected]

regression/cbmc/pragma_cprover_enable_all/test.desc
remi-delmas-3000: I honestly have no idea. Seems related to some constructs being rejected by the SMT parser

src/analyses/goto_check.cpp
remi-delmas-3000: Yes, agreed.

martin-cs (commented):
I agree with @tautschnig that this would be better / easier to review as several commits.

Also, not wanting to be "that guy" but can we briefly discuss the motivation for this PR?  You give two interesting and relevant use cases.  In the first one, if you run the instrumentation multiple times, you get multiple copies of the assertion -- I am kind of OK with this?  Is avoiding this a major problem for your workflow(s)?  In the second case, should goto-check be looking at code in assertions by default, or really at all?  CPROVER assertions are "execution neutral" in that they are just checks, they can't affect the values of variables or alter execution[1], so why are the checks for undefined behaviour looking in them?  At least shouldn't the checks on expressions within assertions do things like convert `e` to `!undefined_cases_of_e => e`?

[1] But what about...
```
int g = 0;
int f00() { __CPROVER_assert(f00() != 0) ; return ++g; }
```

src/analyses/goto_check.cpp
```diff
@@ -277,6 +278,23 @@ class goto_checkt
   bool enable_assumptions;
   bool enable_pointer_primitive_check;
 
+  /// Maps a named-check name to the corresponding boolean flag.
+  std::map<irep_idt, bool *> name_to_flag{
+    {"no-enum-check", &no_enum_check},
```
tautschnig: Can you please insert a commit at the beginning of your commit series that entirely removes `no_enum_check` from goto_check.cpp? It is always `false`. Seems like an unnecessary piece of code introduced in 85b90f304ba90ca83d4f1f3c78e28293c51c057b, probably had a meaning at some point while developing that change back then.

src/analyses/goto_check.cpp
```diff
@@ -287,6 +305,35 @@ class goto_checkt
...
+  /// activation statuses for named checks
+  typedef enum
+  {
+    ENABLE,
+    DISABLE,
+    CHECKED
+  } check_statust;
+
+  /// optional (named-check, status) pair
+  typedef optionalt<std::pair<irep_idt, check_statust>> named_check_statust;
```
tautschnig: `CODING_STANDARD.md` says that we nowadays prefer `using` over `typedef`.

src/analyses/goto_check.cpp
```diff
@@ -287,6 +305,35 @@ class goto_checkt
...
+
+  /// Matches a named-check string of the form
+  ///
+  /// ```
+  /// ("enable"|"disable"|"checked") ":" <named-check>
+  /// ```
+  ///
+  /// \returns a pair (name, status) if the match succeeds
+  /// and the name is known, nothing otherwise.
+  named_check_statust match_named_check(irep_idt named_check);
```
tautschnig: `const irep_idt &named_check`; also, should this method be `const`?

src/analyses/goto_check.cpp
```diff
@@ -287,6 +305,35 @@ class goto_checkt
   allocationst allocations;
 
   irep_idt mode;
+
+  /// \brief Adds "checked" pragmas for all currently active named checks
+  /// on the given source location.
+  void add_active_named_check_pragmas(source_locationt &source_location);
```
tautschnig: Should be marked `const`

src/analyses/goto_check.cpp
```diff
@@ -287,6 +305,35 @@ class goto_checkt
...
+
+  /// \brief Adds "disable" pragmas for all named checks
+  /// on the given source location.
+  void add_all_disable_named_check_pragmas(source_locationt &source_location);
```
tautschnig: `const`

src/analyses/goto_check.cpp
```diff
@@ -1917,63 +1979,40 @@ void goto_checkt::goto_check(
...
+        case check_statust::CHECKED:
+          already_checked.push_back(flag);
+          break;
+        }
+      }
     }
 
+    // inhibit already checked checks
+    for(auto flag : already_checked)
+      resetter.disable(*flag);
```
tautschnig: Do you really need this `already_checked` list? Can't you invoke `resetter.disable(...)` directly earlier on?

src/analyses/goto_check.cpp
```diff
@@ -2453,3 +2492,48 @@ void goto_check(
   const namespacet ns(goto_model.symbol_table);
   goto_check(ns, options, goto_model.goto_functions, message_handler);
 }
+
+void goto_checkt::add_active_named_check_pragmas(
+  source_locationt &source_location)
+{
+  for(auto it : name_to_flag)
```
tautschnig: I'd avoid "it" as a name for this isn't an iterator. Also, I'm never quite sure what the exact type here is, so I tend to use `const auto &entry : name_to_flag` to make sure it's a const reference and not a copy.

src/analyses/goto_check.cpp
```diff
@@ -2453,3 +2492,48 @@ void goto_check(
...
+    if(*(it.second))
+      source_location.add_pragma("checked:" + id2string(it.first));
+}
+
+void goto_checkt::add_all_disable_named_check_pragmas(
+  source_locationt &source_location)
+{
+  for(auto it : name_to_flag)
```
tautschnig: As above: const reference.

src/analyses/goto_check.cpp
```diff
@@ -2453,3 +2492,48 @@ void goto_check(
...
+  auto status = check_statust::DISABLE;
+  if(s.find("enable:") == 0)
+  {
+    status = check_statust::ENABLE;
+  }
+  else if(s.find("disable:") == 0)
+  {
+    status = check_statust::DISABLE;
+  }
+  else if(s.find("checked:") == 0)
```
tautschnig: Use `has_prefix(s, "...")` instead of `s.find(...) == 0`.

src/analyses/goto_check.cpp
```diff
@@ -2453,3 +2492,48 @@ void goto_check(
...
+  {
+    status = check_statust::CHECKED;
+  }
+  else
+  {
+    return {};
+  }
+
+  auto sep = s.find(":");
```
tautschnig: Each of the above branches would actually have a constant known value for this. Perhaps avoid the duplicate `find`? Alternatively, do the `find` early on and then just compare the substrings up to that known position of the colon.

tautschnig (approved):

remi-delmas-3000:
> Also, not wanting to be "that guy" but can we briefly discuss the motivation for this PR? You give two interesting and relevant use cases. In the first one, if you run the instrumentation multiple times, you get multiple copies of the assertion -- 

Hi @martin-cs , the motivation for this PR is that the function contract and loop contract workflows involve several distinct goto-instrument passes that introduce new instrumentation variables, assertions and assumptions in a goto program which then get analysed using CBMC once all transformations are applied.

This ultimate analysis pass uses any automatic checks the user decides to activate (pointer-checks, overflow-checks, enum-checks, etc.). 

Some of this contract instrumentation is execution neutral and some of it requires extra checks:
- we add instrumentation that performs pointer validity checks (to guard snapshots of history variables, to guard write set inclusion checks, etc.) and are expected to manipulate invalid pointers, and we need to disable the built-in CBMC pointer checks
- we add instrumentation that performs pointer arithmetic for write set inclusion checking, which needs to get checked for overflow, regardless of what checks the user decides to use for the last analysis step.

> I am kind of OK with this? Is avoiding this a major problem for your workflow(s)? 

Our contract instrumentation happens early on and in later passes, with the current goto-instrument and cbmc behaviour, it gets picked up and instrumented if it were user code. This results in an explosion of the number of VCCs (x10, from 8k to 80k on our larger benchmarks) when all checks are activated in the last stage. Most of them being useless and hurting performance badly, some of them causing systematic errors. Most importantly we cannot guarantee that any of the checks we need for the soundness of our instrumentation will eventually be performed (eventhough our template Makefile for contracts proofs has all checks activated, the user can always decide to not activate pointer overflow checks and we have no control over this).

So this PR, combined with this other PR https://github.com/diffblue/cbmc/pull/6450 which introduces "enable" pragmas, gives us the level of control we need to ensure required checks will be instrumented exactly once, and that unnecessary checks or redundant checks will not be instrumented, while  letting the  user freely decide on what checks to instantiate on his own code. 

One thing to be noticed is that these "checked" pragmas are not exposed to the user so the possibilities for abuse are limited.

> In the second case, should goto-check be looking at code in assertions by default, or really at all? CPROVER assertions are "execution neutral" in that they are just checks, they can't affect the values of variables or alter execution[1], so why are the checks for undefined behaviour looking in them? 

This is precisely what this PR is trying to fix. My take is that checks should be instantiated on user-provided assertions that are part of the original program, but not necessarily on all instrumentation code and assertions (though in some cases checks might be required to ensure soundness, as is the case with write set inclusion checking).

A question to you :  is there any fundamental difference of intent in uses of `assert` vs uses of `__CPROVER_assert` ?

I was under the impression that `assert` would persist in the user program and be executed at runtime, and that `__CPROVER_assert` were a kind of "proof only" assertions that were not meant to persist in the program at runtime. But I don't know for sure.

> At least shouldn't the checks on expressions within assertions do things like convert `e` to `!undefined_cases_of_e => e`?

Assuming we're talking about `goto_check` generated assertions, this would not solve the problem that the whole `ASSERT !undefined_cases_of_e => e` would be instrumented again at the next invocation of goto-check or cbmc if CLI flags are set. 

> [1] But what about...
> 
> ```
> int g = 0;
> int f00() { __CPROVER_assert(f00() != 0) ; return ++g; }
> ```

Can `goto_check` generate this kind of assertions ? I hope not.
On the other hand, if this were a user-provided assertion, I would expect it to get instrumented with any check the user decides to activate on the CLI.

src/analyses/goto_check.cpp
remi-delmas-3000: This flag is an override to the user-controlled `enable_enum_check` that is set to true before checking LHS in assignments:

```
      // Reset the no_enum_check with the flag reset for exception
      // safety
      {
        flag_resett resetter(i);
        resetter.set_flag(no_enum_check, true, "no_enum_check");
        check(assign_lhs);
      }
```

If we remove this flag CBMC will start checking LHS expressions, I think the problem could be that these assertions will trigger when the LHS is uninitialised and can take any integer value.

src/analyses/goto_check.cpp
remi-delmas-3000: should I also change all other `typedef` in this file into `using` ?

src/analyses/goto_check.cpp
remi-delmas-3000: The check pragmas have priority over enable so they must be applied last.

I have moved the priority logic inside `flag_resett`

src/analyses/goto_check.cpp
remi-delmas-3000: fixed

src/analyses/goto_check.cpp
remi-delmas-3000: fixed

peterschrammel (approved):

martin-cs:
@remi-delmas-3000 thank you for the detailed and thoughtful exploration of the context.  I feel like I understand your workflow a bit better now.  I can see that it is a problem and none of the work-flows changes I proposed will easily fix it.  I am not 100% convinced there isn't an over-all better way but I think that would be a bit "global optimum" and I can see this is an improvement, even if to a local optimum.

> A question to you : is there any fundamental difference of intent in uses of assert vs uses of __CPROVER_assert ?

I think they should do different things.  I think `assert` should terminate the program if it fails (or `assume(0)`) while `__CPROVER_assert` is a pure check that should directly map to an `ASSERT` instruction.

martin-cs (approved):

tautschnig:
> > A question to you : is there any fundamental difference of intent in uses of assert vs uses of __CPROVER_assert ?
> 
> I think they should do different things. I think `assert` should terminate the program if it fails (or `assume(0)`) while `__CPROVER_assert` is a pure check that should directly map to an `ASSERT` instruction.

I suppose I should try to find some time for #5866 to enshrine this.

kroening:
> > > A question to you : is there any fundamental difference of intent in uses of assert vs uses of __CPROVER_assert ?
> > 
> > 
> > I think they should do different things. I think `assert` should terminate the program if it fails (or `assume(0)`) while `__CPROVER_assert` is a pure check that should directly map to an `ASSERT` instruction.
> 
> I suppose I should try to find some time for #5866 to enshrine this.

I would be very careful with this. Note that assert(0) does not abort when NDEBUG is set, which is often the case in production binaries. The failing assertion may then hide a genuine security vulnerability.

PR #6473, opened 2021-11-24
Fix handling of DEAD instructions and function call inlining

SaswatPadhi:
In this PR, we fix 3 minor issues related to code contracts:
1. an overflow error in handling `DEAD` instructions during write set inclusion checking.
2. refactor and optimize function call inlining during code contracts instrumentation
3. extend write set inclusion checks to loop head instructions (in addition to loop body) to check for possible side effects in the guard

### DIFF
--- a/regression/contracts/loop_assigns-05/main.c
+++ b/regression/contracts/loop_assigns-05/main.c
@@ -1,0 +1,59 @@
+#include <assert.h>
+
+int j;
+
+int lowerbound()
+{
+  return 0;
+}
+
+int upperbound()
+{
+  return 10;
+}
+
+void incr(int *i)
+{
+  (*i)++;
+}
+
+void body_1(int i)
+{
+  j = i;
+}
+
+void body_2(int *i)
+{
+  (*i)++;
+  (*i)--;
+}
+
+int body_3(int *i)
+{
+  (*i)++;
+  if(*i == 4)
+    return 1;
+
+  (*i)--;
+  return 0;
+}
+
+int main()
+{
+  for(int i = lowerbound(); i < upperbound(); incr(&i))
+    // clang-format off
+    __CPROVER_assigns(i, j)
+    __CPROVER_loop_invariant(0 <= i && i <= 10)
+    __CPROVER_loop_invariant(i != 0 ==> j + 1 == i)
+    // clang-format on
+    {
+      body_1(i);
+
+      if(body_3(&i))
+        return 1;
+
+      body_2(&i);
+    }
+
+  assert(j == 9);
+}
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -1054,20 +1054,28 @@ void code_contractst::check_frame_conditions(
         assigns_clauset::conditional_address_ranget{assigns, symbol});
       if(symbol_car != assigns.get_write_set().end())
       {
-        instruction_it++;
         auto invalidation_assignment = goto_programt::make_assignment(
           symbol_car->validity_condition_var,
           false_exprt{},
           instruction_it->source_location());
-        // note that instruction_it is not advanced by this call,
-        // so no need to move it backwards
-        body.insert_before_swap(instruction_it, invalidation_assignment);
+        // note that the CAR must be invalidated _after_ the DEAD instruction
+        body.insert_after(
+          instruction_it,
+          add_pragma_disable_assigns_check(invalidation_assignment));
       }
       else
       {
-        throw incorrect_goto_program_exceptiont(
-          "Found a `DEAD` variable without corresponding `DECL`!",
-          instruction_it->source_location());
+        // For loops, the loop_head appears after the DECL of counters,
+        // and any other temporaries introduced during "initialization".
+        // However, they go `DEAD` (possible conditionally) inside the loop,
+        // in presence of return statements.
+        // Notice that for them those variables be writable,
+        // they must appear as assigns targets anyway,
+        // but their DECL statements are outside of the loop.
+        log.warning() << "Found a `DEAD` variable "
+                      << name2string(symbol.get_identifier())
+                      << " without corresponding `DECL`, at: "
+                      << instruction_it->source_location() << messaget::eom;
       }
     }
     else if(
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -256,10 +256,13 @@ void code_contractst::check_apply_loop_contracts(
   // Forward the loop_head iterator until the start of the body.
   // This is necessary because complex C loop_head conditions could be
   // converted to multiple GOTO instructions (e.g. temporaries are introduced).
+  // If the loop_head location shifts to a different function,
+  // assume that it's an inlined function and keep skipping.
   // FIXME: This simple approach wouldn't work when
   // the loop guard in the source file is split across multiple lines.
   const auto head_loc = loop_head->source_location();
-  while(loop_head->source_location() == head_loc)
+  while(loop_head->source_location() == head_loc ||
+        loop_head->source_location().get_function() != head_loc.get_function())
     loop_head++;
 
   // At this point, we are just past the loop head,
@@ -678,6 +681,12 @@ void code_contractst::apply_loop_contract(
   local_may_aliast local_may_alias(goto_function);
   natural_loops_mutablet natural_loops(goto_function.body);
 
+  if(!natural_loops.loop_map.size())
+    return;
+
+  goto_function_inline(
+    goto_functions, function_name, ns, log.get_message_handler());
+
   // A graph node type that stores information about a loop.
   // We create a DAG representing nesting of various loops in goto_function,
   // sort them topologically, and instrument them in the top-sorted order.
@@ -992,7 +1001,11 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
       function_obj->second.body, instruction_it, snapshot_instructions);
   };
 
-  // Insert aliasing assertions
+  // Inline all function calls.
+  goto_function_inline(
+    goto_functions, function_obj->first, ns, log.get_message_handler());
+
+  // Insert write set inclusion checks.
   check_frame_conditions(
     function_obj->first,
     function_obj->second.body,
@@ -1010,8 +1023,6 @@ void code_contractst::check_frame_conditions(
   const goto_programt::targett &instruction_end,
   assigns_clauset &assigns)
 {
-  goto_function_inline(goto_functions, function, ns, log.get_message_handler());
-
   for(; instruction_it != instruction_end; ++instruction_it)
   {
     const auto &pragmas = instruction_it->source_location().get_pragmas();
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -189,6 +189,10 @@ void code_contractst::check_apply_loop_contracts(
       insert_before_swap_and_advance(
         goto_function.body, loop_head, snapshot_instructions);
     };
+
+    // Perform write set instrumentation on the entire loop.
+    check_frame_conditions(
+      function_name, goto_function.body, loop_head, loop_end, loop_assigns);
   }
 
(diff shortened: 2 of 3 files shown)


src/goto-instrument/contracts/contracts.cpp
```diff
@@ -251,29 +255,28 @@ void code_contractst::check_apply_loop_contracts(
...
   // Forward the loop_head iterator until the start of the body.
   // This is necessary because complex C loop_head conditions could be
   // converted to multiple GOTO instructions (e.g. temporaries are introduced).
+  // If the loop_head location shifts to a different function,
+  // assume that it's an inlined function and keep skipping.
   // FIXME: This simple approach wouldn't work when
   // the loop guard in the source file is split across multiple lines.
   const auto head_loc = loop_head->source_location();
-  while(loop_head->source_location() == head_loc)
+  while(loop_head->source_location() == head_loc ||
```
remi-delmas-3000, line 271: This seems fragile. An alternative solution could be to have the C front end inject LOCATION markers to precisely delimit loop entry, loop guard evaluation, loop condition test, loop body, loop step instructions, jump back to head instructions.

regression/contracts/loop_assigns-05/main.c
```diff
@@ -0,0 +1,38 @@
...
+}
+
+void body(int i)
+{
+  j = i;
+}
+
+void main()
+{
+  for(int i = lowerbound(); i < upperbound(); incr(&i))
```
remi-delmas-3000, line 43: could we also add another simple test such as 

```
int main()
{
   unsigned int max;
   assume(max > 0);
   unsigned int i=0;  
   while( i < max++)
   loop_decreases(max-i)
   {
     i++;
   }
}
```
This loop's decreases clause is incorrect, max and i can both overflow 
(but the loop still terminates when max overflows and becomes smaller than i or stays at max_int and i starts catching up on it)
We should be able to find all these errors.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -287,7 +290,8 @@ void code_contractst::check_apply_loop_contracts(
       converter.goto_convert(old_decreases_assignment, generated_code, mode);
     }
 
-    goto_function.body.destructive_insert(loop_body, generated_code);
+    goto_function.body.destructive_insert(
+      loop_body, add_pragma_disable_assigns_check(generated_code));
```
remi-delmas-3000, line 294: shouldn't the decreases clause's initial value be computed before the loop's guard expression is evaluated ?

on this program  

```
while (i < max++)
loop_assigns(i, max)
loop_decreases(max-i)
{
  i++;
}
```
instrumenting like this refutes the loop decreases clause
```
havoc(i, max);
decr_begin = max-i; 
max += 1;
if(i < max)
{
  i++;
  decr_end = max-i;
  assert(decr_end < decr_begin); // falsified
}
```

but instrumenting like this would tell you it holds 

```
havoc(i, max);
max += 1;
if(i < max)
{
  decr_begin = max-i; 
  i++;
  decr_end = max-i;
  assert(decr_end < decr_begin); // proved
}
```

Moreover the goto code for this program we can see that the backjump is before the guard's side effects computation:

```
int main()
{
  unsigned int max;
  unsigned int i = 0;
  while (i < max++)
  {
    i++;
  }
  return 0;
}

```

```
main /* main */
        // 0 file test.c line 6 function main
        DECL main::1::max : signedbv[32]
        // 1 file test.c line 7 function main
        DECL main::1::i : signedbv[32]
        // 2 file test.c line 7 function main
        ASSIGN main::1::i := 0
        // 3 file test.c line 8 function main
     1: DECL main::$tmp::tmp_post : signedbv[32]
        // 4 file test.c line 8 function main
        ASSIGN main::$tmp::tmp_post := main::1::max
        // 5 file test.c line 8 function main
        ASSIGN main::1::max := main::1::max + 1
        // 6 file test.c line 8 function main
        IF ¬(main::1::i < main::$tmp::tmp_post) THEN GOTO 2
        // 7 file test.c line 10 function main
        ASSIGN main::1::i := main::1::i + 1
        // 8 file test.c line 8 function main
        GOTO 1
        // 9 file test.c line 8 function main
     2: SKIP
        // 10 file test.c line 12 function main
        SET RETURN VALUE 0
        // 11 file test.c line 12 function main
        DEAD main::$tmp::tmp_post
        // 12 file test.c line 12 function main
        DEAD main::1::i
        // 13 file test.c line 12 function main
        DEAD main::1::max
        // 14 file test.c line 12 function main
        GOTO 3
        // 15 file test.c line 13 function main
     3: END_FUNCTION
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -678,6 +685,12 @@ void code_contractst::apply_loop_contract(
   local_may_aliast local_may_alias(goto_function);
   natural_loops_mutablet natural_loops(goto_function.body);
 
+  if(!natural_loops.loop_map.size())
+    return;
+
+  goto_function_inline(
+    goto_functions, function_name, ns, log.get_message_handler());
+
```
remi-delmas-3000, line 693: inlining after loop detection will miss loops that are hidden behind function calls that do not have contracts

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1054,20 +1069,28 @@ void code_contractst::check_frame_conditions(
...
+        // and any other temporaries introduced during "initialization".
+        // However, they go `DEAD` (possible conditionally) inside the loop,
+        // in presence of return statements.
+        // Notice that for them those variables be writable,
+        // they must appear as assigns targets anyway,
+        // but their DECL statements are outside of the loop.
+        log.warning() << "Found a `DEAD` variable "
+                      << name2string(symbol.get_identifier())
+                      << " without corresponding `DECL`, at: "
+                      << instruction_it->source_location() << messaget::eom;
```
remi-delmas-3000, line 1093: Do we really need to warn about constructs the user has no control over ? How could we distinguish such expected cases from really incorrect cases ?

remi-delmas-3000 (changes requested):
missing some tests with actual side effects in the guard evaluation, and unresolved issue with decreases clause evaluation

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: We could probably use `log.debug` or `log.info`, yes.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: which is okay? because they don't have contracts anyway?

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: We had discussed this yesterday and I think this should be a separate PR.

I also think we would find more bugs to squash if we want to handle loop guards that have side effects. That was out of scope until now.

src/goto-instrument/contracts/contracts.cpp
SaswatPadhi: Yes, @martin-cs suggested something similar (keeping some markers around), in my other PR.

I would suggest not to fix this this in the same PR especially because it would touch the C front end (parsing functions etc.) This PR is tiny and it only fixes the assigns clause issues (with `DEAD`) and makes it play well with function call inlining (modulo the existing issue with loop body finding).

SaswatPadhi:
Thanks for taking a look, @remi-delmas-3000 

> missing some tests with actual side effects in the guard evaluation, and unresolved issue with decreases clause evaluation

Guards with side effects were out of scope until now. This PR only changes assigns clause instrumentation to happen on the entire loop that CBMC identifies, rather than just the body. We haven't tested loops with guards that have side effects though, and there could be other issues with the invariant assumption / havoc statement / base case assertion placement, that we might want to fix before we can get decreases clauses working.

I would suggest keeping this PR only about assigns clause fixes (both function & loop contracts) and making separate PRs each for fixing issues with loops that with guards that have side effects, handling `do/while` loops etc.

remi-delmas-3000 (approved):

remi-delmas-3000:
approving to merge but we need to keep in mind the issues mentioned in my review

regression/contracts/loop_assigns-05/main.c
martin-cs: Agreed that you should be able to find these and agreed that this would be worth adding as a test.
Not entirely sure how this connects to the specific issue being fixed though.

src/goto-instrument/contracts/contracts.cpp
martin-cs: Yeah... I am not sure it is "logging at run-time to the user" issue, in part because, as @remi-delmas-3000 says, it is not clear what the user is supposed to do with this information.

This feels like a `DATA_INVARIANT` of a goto-programs; you can only `DEAD` a variable that you have previously `DECL`d.  This seems reasonable and feels like the kind of thing that should be in the `--validate-goto-model` checks.

src/goto-instrument/contracts/contracts.cpp
martin-cs: I think I agree with both.  I have already had Opinions at @SaswatPadhi about loop structure detection (@remi-delmas-3000 did I CC you?  If not I can resend. ); this is fragile and is asking for trouble and inconsistency with the rest of CPROVER.  I think this functionality should be factored out and used in all appropriate places.

But I also agree with @SaswatPadhi ; that is outside of the scope of this PR.

src/goto-instrument/contracts/contracts.cpp
martin-cs: I don't object but do note that this kind of inlining can get *expensive*.

martin-cs (approved):
I am not going to block this over the logging but please could you address it?

( Also thanks for the really well split-out changes and clear commit messages, makes it *so* much easier to read. )

regression/contracts/loop_assigns-05/main.c
feliperodri: I think this is a separate issue from the one addressed in this PR. @remi-delmas-3000 is there an issue where we track this?

feliperodri (approved):

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1054,20 +1054,28 @@ void code_contractst::check_frame_conditions(
...
-          instruction_it->source_location());
+        // For loops, the loop_head appears after the DECL of counters,
+        // and any other temporaries introduced during "initialization".
+        // However, they go `DEAD` (possible conditionally) inside the loop,
+        // in presence of return statements.
+        // Notice that for them those variables be writable,
+        // they must appear as assigns targets anyway,
+        // but their DECL statements are outside of the loop.
+        log.warning() << "Found a `DEAD` variable "
+                      << name2string(symbol.get_identifier())
```
tautschnig, line 1091: `name2string` seems to be rarely used (to the extent that I'm wondering whether we should get rid of it). I'd suggest you use `id2string` instead, though here you actually don't need either!

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -678,6 +681,12 @@ void code_contractst::apply_loop_contract(
   local_may_aliast local_may_alias(goto_function);
   natural_loops_mutablet natural_loops(goto_function.body);
 
+  if(!natural_loops.loop_map.size())
+    return;
+
+  goto_function_inline(
+    goto_functions, function_name, ns, log.get_message_handler());
```
tautschnig, line 692: As discussed offline: calls to `goto_function_inline` will require `goto_functions.update()` to be called afterwards.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -992,7 +1001,11 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
       function_obj->second.body, instruction_it, snapshot_instructions);
   };
 
-  // Insert aliasing assertions
+  // Inline all function calls.
+  goto_function_inline(
+    goto_functions, function_obj->first, ns, log.get_message_handler());
```
tautschnig, line 1010: As above: call `goto_functions.update()`

tautschnig (commented):
I was too late to the party: I'll create a follow-up PR to fix the issues I called out.

PR #6474, opened 2021-11-24
Add interpretation of SMT check-sat responses for incremental SMT solving support

thomasspriggs:
This PR adds interpretation of SMT check-sat responses for incremental SMT solving support. This is such that CBMC can report something useful rather than just reaching an UNIMPLEMENTED, when solving is complete.The contents of this PR is sufficient for CMBC to complete successfully when the response is UNSAT and the validation is successful.

### DIFF
--- a/regression/cbmc-incr-smt2/nondeterministic-int-assert/valid_unsat.desc
+++ b/regression/cbmc-incr-smt2/nondeterministic-int-assert/valid_unsat.desc
@@ -1,0 +1,14 @@
+CORE winbug
+valid_unsat.c
+--incremental-smt2-solver "z3 --smt2 -in" --verbosity 10
+Passing problem to incremental SMT2 solving via "z3 --smt2 -in"
+Sending command to SMT2 solver - \(check-sat\)
+Solver response - unsat
+VERIFICATION SUCCESSFUL
+^EXIT=0$
+^SIGNAL=0$
+--
+type: pointer
+--
+Test that given a `.c` program where all assertions hold, the solver responds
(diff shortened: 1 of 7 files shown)


src/solvers/smt2_incremental/smt2_incremental_decision_procedure.cpp
```diff
@@ -215,7 +229,8 @@ decision_proceduret::resultt smt2_incremental_decision_proceduret::dec_solve()
 {
   ++number_of_solver_calls;
   solver_process->send(smt_check_sat_commandt{});
-  const smt_responset result = solver_process->receive_response();
+  const smt_responset result =
```
NlightNFotis, line 246: You are sending the `smt_check_sat_commandt` command to the solver two times here, once in the line above and once in the new helper function.

Is that an error or by design?

thomasspriggs: Ah, that is a mistake I made in the addition of the success messages. I will fix this.

regression/cbmc-incr-smt2/nondeterministic-int-assert/valid_unsat.desc
```diff
@@ -0,0 +1,14 @@
+CORE winbug
+valid_unsat.c
+--incremental-smt2-solver "z3 --smt2 -in" --verbosity 10
```
TGWDB, line 3: Is the verbosity 10 needed here?

TGWDB (approved):

NlightNFotis (approved):

PR #6475, opened 2021-11-24
Tidying up the options for goto-instrument

martin-cs:
I'm working through the options of `goto-instrument` seeing:

1. Is it implemented?
2. If its implementation is basically a call to a single other header, are the flags described there with `OPT_` and `HELP_` macros?
3. Does it have at least one test case?

This is the first handful of these.  It is best to review commit by commit as each one is short and mostly separate.

Let me know if you think its worth me doing more of these!

### DIFF
--- a/src/goto-instrument/goto_instrument_parse_options.cpp
+++ b/src/goto-instrument/goto_instrument_parse_options.cpp
@@ -86,7 +86,6 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "call_sequences.h"
 #include "concurrency.h"
 #include "dot.h"
-#include "dump_c.h"
 #include "full_slicer.h"
 #include "function.h"
 #include "havoc_loops.h"
@@ -1722,11 +1721,11 @@ void goto_instrument_parse_optionst::help()
     "\n"
     "Main options:\n"
     HELP_DOCUMENT_PROPERTIES
-    " --dump-c                     generate C source\n"
-    " --dump-c-type-header m       generate a C header for types local in m\n"
-    " --dump-cpp                   generate C++ source\n"
     " --dot                        generate CFG graph in DOT format\n"
     " --interpreter                do concrete execution\n"
+    "\n"
+    "Dump Source:\n"
+    HELP_DUMP_C
     "\n"
     "Diagnosis:\n"
     " --show-loops                 show the loops in the program\n"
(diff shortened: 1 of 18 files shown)


martin-cs:
Please can someone who understands CTest and the build on Mac help me with this:

https://github.com/diffblue/cbmc/runs/4314110590?check_suite_focus=true#step:11:4571

It looks like the `uninitialized-check` test I added is failing but no reason is given because the `goto-instrument` tests where listed as passing:

```
20:   Running typedef4/test.desc  [OK] in 0 seconds
20:   Running uninitialized-check/test.desc  [FAILED]
15:   Running windows_h_VS_2013/test.desc  [OK] in 10 seconds
15: 
15: All tests were successful, 15 tests skipped
 7/68 Test  #15: ansi-c-CORE ..........................................   Passed  102.60 sec
test 28
      Start  28: cbmc-concurrency-CORE
```

martin-cs:
Is it possible to access the `*.log` files from the runs of the regression test?

martin-cs:
Weirdly, running it again got me an error message.  I guess there is something fragile about how CTest is set up?

NlightNFotis:
Hi @martin-cs , looks like everything is passing now.

Next time, if you need any help with a mac build feel free to ping me directly - I have access to an intel machine and an m1 machine if that helps.

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1851,6 +1851,7 @@ void goto_instrument_parse_optionst::help()
     "Other options:\n"
     " --version                    show version and exit\n"
     HELP_FLUSH
+    " --xml                        output files in XML where supported\n"
```
NlightNFotis, line 1842: Is it possible to add some documentation in the `docs/` folder for this? I wasn't aware of the difference between `--xml` and `--xml-ui`

NlightNFotis (approved):
Solid cleanup, minor comment.

src/goto-instrument/goto_instrument_parse_options.cpp
martin-cs: It is but...

A. The current state of the `goto-instrument` documentation ( https://github.com/diffblue/cbmc/blob/develop/doc/cprover-manual/goto-instrument.md ), apart from function body generation and contracts is pretty abysmal.  I do want to improve it  but it is after "tidying up, figuring out and testing" the options on my TODO list.

B. I am not sure that it should even be an option or called that.  At the moment I am going with preserving functionality but I am also building a set of "I think this would be sensible" changes.  As it stands it is a formerly undocumented option that works with only two of the flags.  I do wonder if it would make more sense if it followed the same convention as `goto-analyzer`.  Or maybe it would be better to move all of the functionality over to `goto-analyzer`?

tautschnig (approved):
❤️

peterschrammel (approved):

Issue #260, opened 2016-10-19
Test and repair --full-slice and --all-properties


peterschrammel:
add regression tests for C

@kroening, any observations what goes wrong with --full-slice at the moment?

lucasccordeiro:
@kroening: can you also provide some starting point for this particular task? For example, which test cases should I consider (or if I have to create a test suite)? What is actually wrong with --full-slice and --all-properties? Which additional options should I use to test --full-slice and --all-properties (e.g., cbmc file.c --full-slice, cbmc file.c --all-properties)?

lucasccordeiro:
@peterschrammel: which test suite can I add regression tests related to this issue? It seems that the --all-properties option is related to cbmc-incr. Is that correct? However, --full-slice option is not clear which test suite I can add test cases to.

lucasccordeiro:
Just to keep the history.

@tautschnig: This is the example with pointers that I mentioned in your visit at DiffBlue.

````
void *malloc(__CPROVER_size_t);
void free(void *);
int main()
{
  int *p=malloc(sizeof(int));
  int *q=p;
  int i, x;
  i=x;
  if(i==4711) free(q);
  // should fail if i==4711
  *p=1;
}
````
I was getting a different verification result when verifying it with --full-slice.

Additionally, the full-slice option gives incorrect results for the following C programs from cbmc regression suite:

Running Fixedbv3/test.desc  [FAILED]
Running Float3/test.desc  [FAILED]
Running Float3/test.desc  [FAILED]
Running Function5/test.desc  [FAILED]
Running Function6/test.desc  [FAILED]
Running Local_out_of_scope1/test.desc  [FAILED]
Running Memory_leak1/test.desc  [FAILED]
Running Overflow_Addition1/test.desc  [FAILED]
Running Pointer_Arithmetic5/test.desc  [FAILED]
Running Pointer_byte_extract5/test.desc  [FAILED]
Running Sideeffects5/test.desc  [FAILED]
Running Sideeffects6/test.desc  [FAILED]
Running String2/test.desc  [FAILED]
Running String6/test.desc  [FAILED]
Running Undefined_Function1/test.desc  [FAILED]
Running Undefined_Function2/test.desc  [FAILED]
Running array-tests/test.desc  [FAILED]
Running complex1/test.desc  [FAILED]
Running equality_through_union3/test.desc  [FAILED]
Running exit1/test.desc  [FAILED]
[...]

PR #6479, opened 2021-11-25
Warn user when unsound flags are enabled.

NlightNFotis:
Hello,

This is a draft PR that enables warning a user when `--reachability-slice` and `--full-slice`
are used in `cbmc` or `goto-instrument`.

It also adds a small doc for explaining the rationale for this, and how the warnings look.

For more discussion on this, please have a look at #6480

### DIFF
--- a/doc/cprover-manual/unsound_options.md
+++ b/doc/cprover-manual/unsound_options.md
@@ -1,0 +1,67 @@
+[CPROVER Manual TOC](../)
+
+## Unsound CLI options for various tools
+
+### Context
+
+In [#6480](https://github.com/diffblue/cbmc/issues/6480), there has been some extensive
+conversation about what it means for certain options to produce *unsound* behavior.
+
+We concluded that *unsound* in this context is a proxy for the following two behaviors
+that we want to avoid:
+
+* **A spurious counter example to an assertion**, which means that the tool may report
+  a coverage property (or a line of code) as reachable when in fact it is not, and
+* **Wrong proof to an assertion**, which means that it might indicate an assertion passing
+  when it's not.
+
+We expect that, by default, CBMC and JBMC display none of the behavior we
+described above (and if they do, it's an extremely serious bug that we aim to
+fix on an ASAP basis), but be aware that certain tools, like `goto-instrument`,
+may contain components that are experimental in nature and thus do
+transformations that eventually lead to behavior such as the ones described
+above. Furthermore, some options lead to unsound analysis results *by design*,
+and some transformations performed by `goto-instrument` will yield verification
+results that are (by design) unsound when taking verification of the
+non-transformed program as reference.
+
+### Examples of Options that may yield Unsound Results
+
+The following options will produce a warning when used with CBMC or JBMC:
+
+* Use of `--unwind` or `--unwindset` without `--unwinding-assertions`, or the
+  use of `--partial-loops`.
+* Depth or complexity-limited analysis (`--depth`, `--symex-complexity-limit`).
+
+See [Understanding Loop Unwinding](../cbmc/unwinding/) for an elaboration of
+these options.
+
+### Experimental Options
+
+Be advised that the following command line options to `cbmc` and `goto-instrument`
+have been reported to be unsound:
+
+* `--full-slice` has been reported to be unsound in issue [cbmc#260](https://github.com/diffblue/cbmc/issues/260)
+  In particular, `--full-slice` appears to lead to spurious counter examples,
+  because values that get assigned by a function whose body gets sliced out are
+  no longer present in the trace, but still result in flipped verification results.
+
+`cbmc` and `goto-instrument` have also been modified to warn that options used
+are unsound as part of their output. An example of how that output looks is shown
+below:
+
+```sh
+$ cbmc --full-slice ~/Devel/cbmc_bugs/6394/before-slice.out
+CBMC version 5.45.0 (cbmc-5.43.0-77-g99c5a92de1-dirty) 64-bit arm64 macos
+Reading GOTO program from file
+Reading: ~/Devel/cbmc_bugs/6394/before-slice.out
+Generating GOTO Program
+Adding CPROVER library (x86_64)
+Removal of function pointers and virtual functions
+Generic Property Instrumentation
+**** WARNING: Experimental option --full-slice, analysis results may be unsound. See https://github.com/diffblue/cbmc/issues/260
+Performing a full slice
+Running with 8 object bits, 56 offset bits (default)
+Starting Bounded Model Checking
+[...]
+```
--- a/src/cbmc/cbmc_parse_options.cpp
+++ b/src/cbmc/cbmc_parse_options.cpp
@@ -193,13 +193,26 @@ void cbmc_parse_optionst::get_command_line_options(optionst &options)
     options.set_option("mm", cmdline.get_value("mm"));
 
   if(cmdline.isset("symex-complexity-limit"))
+  {
     options.set_option(
       "symex-complexity-limit", cmdline.get_value("symex-complexity-limit"));
+    log.warning() << "**** WARNING: Complexity-limited analysis may yield "
+                     "unsound verification results"
+                  << messaget::eom;
+  }
 
   if(cmdline.isset("symex-complexity-failed-child-loops-limit"))
+  {
     options.set_option(
       "symex-complexity-failed-child-loops-limit",
       cmdline.get_value("symex-complexity-failed-child-loops-limit"));
+    if(!cmdline.isset("symex-complexity-limit"))
+    {
+      log.warning() << "**** WARNING: Complexity-limited analysis may yield "
+                       "unsound verification results"
+                    << messaget::eom;
+    }
+  }
 
   if(cmdline.isset("property"))
     options.set_option("property", cmdline.get_values("property"));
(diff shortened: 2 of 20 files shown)


src/cbmc/cbmc_parse_options.cpp
```diff
@@ -952,6 +952,8 @@ bool cbmc_parse_optionst::process_goto_program(
 
   if(options.get_bool_option("reachability-slice"))
   {
+    log.warning() << "WARNING: Unsound option --reachability-slice"
+                  << messaget::eom;
```
tautschnig: This option isn't fundamentally unsound, it's just that #6505 needs to be merged to fix a bug.

tautschnig (changes requested):
Thank you for kicking off work to annotate options and document what's going on. For "reachability-slice," however, I'd rather see us merge the fix.

TGWDB (changes requested):
As per the discussion we had in #6480 I thought the conclusion was to be explicit in how an option can be "unsound" as either *Spurious counter example to an assertion* or *Wrong proof to an assertion*. Can we update the messages and documentation here to say what can go wrong.

NlightNFotis:
Hi @tautschnig, some updates with regard to this:

1) I have removed the `--reachability-slice` mark as requested.
2) I have explored the behaviour here a bit, and can report the following:

If we enable any of the slice flags on a global level for a regressions test-suite run, for example `--full-slice` for all tests under `regression/cbmc/` we do get a number of failures that appear to be unexpected (to me at least). 

The failures appear to be flipping of verification results (from success to failure), and tend to refer to values not present in the trace. I have noticed that what happens in these cases is that we are asserting against values that appear to be assigned the result of a function call, but the slice options appear to eliminate the function body.

This is what happens in the case of a `--full-slice`, and which I've done some digging into. Test failures for similar reasons appear to be the result of other flags activated on a global basis, (for instance, `--reachability-slice` or `--slice-formula`) but I'm not sure if it's the exact same behaviour driving this.

I have done a number of quick fix-like experiments, as well as a few other structural changes to the full-slice but couldn't get anything to work properly (by this I mean passing all the tests), which signals a lack of understanding of the slicing internals from my end. But this appears to be genuine wrong behaviour (I expect any slicer to effectively preserve at least the direct path to an assignment).

NlightNFotis:
I have left a more detailed comment outlining the current issue in #260 .

TGWDB (approved):

NlightNFotis:
Hi @tautschnig, I have changed this to now also mark `--reachability-slice` as unsound, based on comments made during one of our regular catchup meetings.

Can you please confirm this is the case?

Could you also have another look at this at your earliest convenience, given that this is blocked by you?

NlightNFotis:
Hi @tautschnig, ping here - is this possible to review this so that we can get it merged?

tautschnig:
> Hi @tautschnig, I have changed this to now also mark `--reachability-slice` as unsound, based on comments made during one of our regular catchup meetings.
> 
> Can you please confirm this is the case?
> 
> Could you also have another look at this at your earliest convenience, given that this is blocked by you?

With apologies for the late response: is it still the case that `--reachability-slice` is knowingly unsound/producing wrong results? I am aware that `--full-slice` has issues and marking that one "unsound" appears fine.

src/cbmc/cbmc_parse_options.cpp
tautschnig: The above PR has been merged, so I don't think this applies any longer.

doc/cprover-manual/unsound_options.md
```diff
@@ -0,0 +1,37 @@
...
+have been reported to be unsound:
+
+* `--full-slice` has been reported to be unsound in issue [cbmc#260](https://github.com/diffblue/cbmc/issues/260)
+  In particular, `--full-slice` appears to lead to spurious counter examples,
+  because values that get assigned by a function whose body gets sliced out are
+  no longer present in the trace, but still result in flipped verification results.
+* `--reachability-slice` has been reported unsound at least once, in issue [cbmc#6394](https://github.com/diffblue/cbmc/issues/6394).
+  The state of this option these days is that this flag *should not* contain significant
+  soundness bugs, but it has a number of potential issues that pop up everyone now and then,
+  which renders it unsuitable for production use.
```
tautschnig: I think we might want to list other options in here, but I'm not sure `--reachability-slice` still warrants this warning. On the other hand, `--aggressive-slice` would fit in here.

doc/cprover-manual/unsound_options.md
```diff
@@ -0,0 +1,37 @@
+# Unsound CLI options for various tools
+
+Be advised that the following command line options to `cbmc` and `goto-instrument`
+have been reported to be unsound:
```
tautschnig: Can I ask that we start with a bit more context (reflecting the fact that https://github.com/diffblue/cbmc/issues/6480 had some extensive discussion)? I don't think there even is a universally agreed definition of "unsound," so please try to start with a summary of the conclusion of #6480.

doc/cprover-manual/unsound_options.md
```diff
@@ -0,0 +1,37 @@
...
+WARNING: Unsound option --full-slice
+Performing a full slice
+Running with 8 object bits, 56 offset bits (default)
+Starting Bounded Model Checking
+[...]
+```
+
+Be advised some of these options may be activated by auxiliary options, for instance,
+`--aggressive-slice` seems to lead to a path that performs the transformations that
+`--reachability-slice` does in `goto-instrument`.
```
tautschnig: We should not do this via documentation, but instead need to make sure that the code always issues a warning when one of the unsound code paths is entered.

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1806,6 +1814,7 @@ void goto_instrument_parse_optionst::instrument_goto_program()
     goto_model.goto_functions.update();
 
     log.status() << "Performing a reachability slice" << messaget::eom;
+
```
tautschnig: Whitespace change?

src/goto-instrument/goto_instrument_parse_options.cpp
```diff
@@ -1743,7 +1749,9 @@ void goto_instrument_parse_optionst::instrument_goto_program()
     do_indirect_call_and_rtti_removal();
     do_remove_returns();
 
+    log.warning() << "WARNING: Unsound option --full-slice" << messaget::eom;
     log.status() << "Performing a full slice" << messaget::eom;
+
```
tautschnig: Whitespace change?

NlightNFotis:
Hi @tautschnig,

I believe I have addressed all of the comments.

I have also removed the explanation for the `--reachability-slice` from the documentation, along with the `--aggressive-slice` option.

I'm not aware of any issues with the `--aggressive-slice` (the references before was due to a transitive relation to `--reachability-slice`) and I removed the references to it. If you are aware of any issues with it let me know and I can add it back in with a more elaborate explanation.

doc/cprover-manual/unsound_options.md
NlightNFotis: Added a paragraph with the conclusions from #6480

doc/cprover-manual/unsound_options.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Context
+
+In [#6480](https://github.com/diffblue/cbmc/issues/6480), there has been some extensive
+conversation about what it means for certain options to produce *unsound* behaviour.
+
+We concluded that *unsound* in this context is a proxy for the following two behaviours
+that we want to avoid:
+
+* **A spurious counter example to an assertion**, which gives a proof that the assertion
```
jimgrundy: I think a lot of people would fail to recognize this as a soundness problem. But if you cast it another way - that the tool may report a coverage property (or line of code) as reachable when, in fact, it is not - then I think more people will see that this is just as serious as a spurious proof. I would prefer we state it that way.

jimgrundy (changes requested):

NlightNFotis:
Hi,

@jimgrundy @tautschnig 

The comments you have left have now been addressed. Could you please have another look at your earliest convenience? Cheers.

jimgrundy (approved):

tautschnig (approved):
We can merge this as is, but I am wondering whether we might want to speak of "Unexpectedly" unsound options? Examples of transformations that incur unsoundness are `--partial-loops` (may yield spurious counterexamples) or `--aggressive-slicer` (may yield both spurious counterexamples as well as wrong proof). Both of these, AFAIK, work exactly as their documentation suggests, but do yield those unsound results.

jimgrundy (changes requested):
Hang on... there are options beyond the ones documented here that are known to cause unsoundness? Surely those must be documented here. It's almost worse than the present situation to go ahead as is, because if I now believe the tool will warn me when I ask for something unsound and I don't get a warning then I'll be even more convinced of the validity of the results I get.

tautschnig:
> Hang on... there are options beyond the ones documented here that are known to cause unsoundness? Surely those must be documented here. It's almost worse than the present situation to go ahead as is, because if I now believe the tool will warn me when I ask for something unsound and I don't get a warning then I'll be even more convinced of the validity of the results I get.

@jimgrundy I have now added a further commit as discussed out of band.

jimgrundy (changes requested):
I like these changes even more, but, since I'm asking
- It feels now like there should be a heading and _short_ section on options that are unsound by design (which probably isn't the most catchy name for them). If we could give an example or two of these + state that these options will give a warning when used.
Do we know how many such options there are? If we don't know or there are lots then we could defer adding the warning until we tackle soundness more completely, but if there if there are just a couple could we sneak the warnings in now?

tautschnig:
> I like these changes even more, but, since I'm asking
> 
>     * It feels now like there should be a heading and _short_ section on options that are unsound by design (which probably isn't the most catchy name for them). If we could give an example or two of these + state that these options will give a warning when used.

Done.

>       Do we know how many such options there are? If we don't know or there are lots then we could defer adding the warning until we tackle soundness more completely, but if there if there are just a couple could we sneak the warnings in now?

I believe I have covered all of them for CBMC. JBMC has a more elaborate environment generation facility, which might have its own share of risks. Perhaps @peterschrammel can comment on what I've covered so far?

jimgrundy (approved):

doc/cprover-manual/unsound_options.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Context
+
+In [#6480](https://github.com/diffblue/cbmc/issues/6480), there has been some extensive
+conversation about what it means for certain options to produce *unsound* behavior.
+
+We concluded that *unsound* in this context is a proxy for the following two behaviors
+that we want to avoid:
+
+* **A spurious counter example to an assertion**, which means that the tool may report
```
peterschrammel, line 13: counterexample

peterschrammel (approved):

Issue #6469, opened 2021-11-22
Calling all users of `goto-instrument`'s `--show-value-sets`, `--show-rw-sets`, `--remove-pointers`, `--race-check`, `--mm`, `--isr` and `--concurrency`


martin-cs:
These are the only options that use the older `static_analysist` code.  I am thinking of refactoring that and I am interested to know who that would affect.

The `git blame` for these suggest that they have been about for over a decade so although @kroening is listed as the author, the code may well predate the move to subversion and the original authorship has been lost.

Out of all of the options only `--remove-pointers` (in the acceleration tests) and `--mm` (in the weak memory) have tests in the regression suite.

Thoughts / opinions / preferences / hidden use cases / requests that things are not changed / requests that code is improved are most welcome.

martin-cs:
As no one objected, https://github.com/diffblue/cbmc/pull/6482 makes this change.

PR #6482, opened 2021-11-27
Port the last user of static_analysist

martin-cs:
This ports `value_set_analysis`, which was the last user of deprecated `static_analysist` to use `ait`.  Note that this is *not* the value set analysis code used by `cbmc`; this one is only used by `goto-instrument` (see https://github.com/diffblue/cbmc/issues/6469 ).  There are a few supporting changes to the `ait` infrastructure which are necessary but technically independent.  It is best to review commit by commit.  Note that the test case is added before the porting and works both before and after so most of the output is preserved (the afterwards, `ait` output has slightly more as it also outputs the instructions).


regression/goto-instrument/show-value-sets/main.c
```diff
@@ -0,0 +1,41 @@
...
+{
+  // Uninitialized pointer
+  int *u;
+
+  // Basic assignment
+  u = NULL;
+  int a;
+  u = &a;
+
+  // Test merging and the loss of corelation
```
peterschrammel: correlation

martin-cs:
At this stage I haven't removed `static_analysist` as I don't know if there are out-of-tree users but @tautschnig ported the other users 8 years ago ( https://github.com/diffblue/cbmc/blob/8a720e04f7706a16390abc8d70b627a74435f1e2/src/analyses/static_analysis.h#L16 ) so maybe it is time to get rid of it?

peterschrammel (approved):

peterschrammel:
> maybe it is time to get rid of it?

no objections

martin-cs:
I am seeing unit test failures in https://github.com/diffblue/cbmc/blame/develop/jbmc/unit/pointer-analysis/custom_value_set_analysis.cpp which I think are because `ait` lazily initializes domains while `static_analysist` would initialize everything up front.  This should be fixable but the existence and way the test is written makes me wonder if this is checking an interface that Diffblue code is using, any thoughts @smowton @chrisr-diffblue @peterschrammel @NlightNFotis ?

regression/goto-instrument/show-value-sets/test.desc
```diff
@@ -0,0 +1,27 @@
...
+It would be reasonable to ask about the checks for:
+
+main::1::q = \{ <main::1::b, 0, signedbv\[32\]> \}
+main::1::q = \{ <main::1::c, 0, signedbv\[32\]> \}
+
+but as domains are stored at the start of instructions the following merge will
+over-write it on one branch or the other.  To keep the test portable it is best
+avoided.
+
+
```
tautschnig, line 27: Nit pick: unnecessary blank lines?

src/pointer-analysis/value_set_domain.h
```diff
@@ -12,64 +12,125 @@ Author: Daniel Kroening, kroening@kroening.com
...
+    value_set.clear();
+    value_set.location_number = l->location_number;
+  }
 
-  bool merge(const value_set_domain_templatet<VST> &other, locationt)
+  void make_bottom() override
   {
-    return value_set.make_union(other.value_set);
+    reachable = false;
+    value_set.clear(); //???
```
tautschnig, line 44: ?

src/pointer-analysis/value_set_domain.h
```diff
@@ -12,64 +12,125 @@ Author: Daniel Kroening, kroening@kroening.com
...
-    value_setst::valuest &dest) override
+    value_setst::valuest &dest)
   {
     value_set.get_reference_set(expr, dest, ns);
   }
+
+  exprt get_return_lhs(locationt to) const
+  {
+    // get predecessor of "to"
+    to--;
```
tautschnig, line 101: I am wondering whether `get_return_lhs` should have a different interface: decrementing `to` is an unchecked operation, and would be undefined when `to` already is the beginning of a container.

src/pointer-analysis/value_set_domain.h
```diff
@@ -12,64 +12,125 @@ Author: Daniel Kroening, kroening@kroening.com
...
+
+    if(to->is_end_function())
+      return static_cast<const exprt &>(get_nil_irep());
```
tautschnig, line 104: Is that really a possible scenario, i.e., can decrementing `to` result in reaching an `END_FUNCTION` instruction? If so, is passing a one-past-the-end iterator to `get_return_lhs` a realistic scenario?

src/pointer-analysis/value_set_domain.h
```diff
@@ -116,4 +177,8 @@ void value_set_domain_templatet<VST>::transform(
   }
 }
 
+// To pass the correct location to the constructor we need a factory
+typedef ai_domain_factory_location_constructort<value_set_domaint>
+  value_set_domain_factoryt;
```
tautschnig, line 182: I think we prefer `using` over `typedef` nowadays.

tautschnig (approved):

smowton:
Memories of this far too foggy to be authoritative, but I think this was added so the security project (RIP) could do value-set-analysis-with-tweaks.

peterschrammel:
@martin-cs, please adapt the test for use with `ait` instead of `static_analysist`. Any custom value set analysis using this will need to upgrade to `ait` anyway.

src/pointer-analysis/value_set_domain.h
martin-cs: The underlying `value_sett` doesn't really have a clear idea of `top` and `bottom` because it's not really used / formalised like that.  The constructor of `value_set_domain_templatet` would always `value_set.clear()` which means I think it is the right thing to set when `make_bottom`.

src/pointer-analysis/value_set_domain.h
martin-cs: Yes; it is a risky bit of code and far from an ideal design.  It is moved over because I was trying to preserve functionality and structure where I could.  I should refactor this though.

src/pointer-analysis/value_set_domain.h
martin-cs: I really don't think it should be.  I think this bit of code is only used in `END_FUNCTION` where `to` is the instruction after the call.

martin-cs:
@chrisr-diffblue thanks, that does ring a bell...  I will assume that changes to this are not going to be high-risk for Diffblue.

@peterschrammel Will do, I think the changes should be very minimal.

src/pointer-analysis/value_set_domain.h
tautschnig: Perhaps replace the question marks by exactly that explanation?

regression/goto-instrument/show-value-sets/main.c
martin-cs: Fixed.

regression/goto-instrument/show-value-sets/test.desc
martin-cs: Fixed

NlightNFotis:
A couple of jobs failed because the reference to file has not been removed from the `makefiles.

src/pointer-analysis/value_set_domain.h
martin-cs: Done.

src/pointer-analysis/value_set_domain.h
martin-cs: Changed.

src/pointer-analysis/value_set_domain.h
martin-cs: I have had a look at this and when doing a "return" edge from `END_FUNCTION`, decrementing `to` really is the best way of getting the function call instruction.  That is probably a wider API failure.  However the only case in which this is actually needed is to handle the return value.  So if we are happy return removal has always been done first then it is fine to just remove all of this code.

martin-cs: I am mildly concerned that some of the more obscure users might not ensure this so I think I will just simplify but preserve the functionality.

martin-cs:
@peterschrammel @tautschnig : you both reviewed and approved this back in the distant past.  I am trying to get it moving again.  The PR is essentially identical but with the fix for the issue with the custom value set analysis and removal of static_analysist.  These were the outstanding issues from the reviews.  If you want another look then do shout, otherwise I will merge it when I can get CI to pass!

peterschrammel (approved):

martin-cs:
I'm going to ignore the patch coverage metric because of the increase of project coverage.

PR #6486, opened 2021-11-29
Byte extract lowering: refactor lowering of arrays/vectors [blocks: #6488]

tautschnig:
Both cases had some of the same code. Refactor those cases out of an
already-too-big function into one of its own.

This is in preparation of further changes to this code to support some
of the cases currently deemed unsupported.

- [x] Each commit message has a non-empty body, explaining why the change was made.
- [x] Methods or procedures I have added are documented, following the guidelines provided in CODING_STANDARD.md.
- n/a The feature or user visible behaviour I have added or modified has been documented in the User Guide in doc/cprover-manual/
- [ ] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- n/a My commit message includes data points confirming performance improvements (if claimed).
- [x] My PR is restricted to a single feature or bugfix.
- n/a White-space or formatting changes outside the feature-related changed lines are in commits of their own.

PR #6488, opened 2021-11-29
Byte extract lowering: support extracting non-byte-aligned arrays

tautschnig:
Alignment to bytes may be true for ANSI C, but isn't the case for our C front-end (which supports arrays of single bits), and not true for the overall framework in general. This required fixing alignment assumptions both in unpack_array_vector (which turns an arbitrary type into an array of bytes) as well as lower_byte_extract_array_vector (which picks an array/vector of arbitrary subtype from an array of bytes as prepared by unpack_array_vector).

~Depends on #6486, which is the first commit in here. Only the second commit is new.~

### DIFF
--- a/src/util/lower_byte_operators.cpp
+++ b/src/util/lower_byte_operators.cpp
@@ -563,29 +563,32 @@ static exprt unpack_array_vector(
   // refine the number of elements to extract in case the element width is known
   // and a multiple of bytes; otherwise we will expand the entire array/vector
   optionalt<mp_integer> num_elements = src_size;
-  if(element_bits > 0 && element_bits % bits_per_byte == 0)
+  if(element_bits > 0)
   {
     if(!num_elements.has_value())
     {
       // turn bytes into elements, rounding up
-      num_elements = (*max_bytes + el_bytes - 1) / el_bytes;
+      num_elements =
+        (*max_bytes * bits_per_byte + element_bits - 1) / element_bits;
     }
 
     if(offset_bytes.has_value())
     {
       // compute offset as number of elements
-      first_element = *offset_bytes / el_bytes;
-      // insert offset_bytes-many nil bytes into the output array
+      first_element = *offset_bytes * bits_per_byte / element_bits;
+      // insert offset_bytes-many zero bytes into the output array
       byte_operands.resize(
         numeric_cast_v<std::size_t>(std::min(
-          *offset_bytes - (*offset_bytes % el_bytes),
-          *num_elements * el_bytes)),
+          (*offset_bytes * bits_per_byte -
+           ((*offset_bytes * bits_per_byte) % element_bits)) /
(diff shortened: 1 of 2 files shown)


src/solvers/lowering/byte_operators.cpp
```diff
@@ -508,69 +508,117 @@ static exprt unpack_array_vector(
...
   // and a multiple of bytes; otherwise we will expand the entire array/vector
   optionalt<mp_integer> num_elements = src_size;
-  if(element_bits > 0 && element_bits % 8 == 0)
+  if(element_bits > 0)
   {
     if(!num_elements.has_value())
     {
       // turn bytes into elements, rounding up
-      num_elements = (*max_bytes + el_bytes - 1) / el_bytes;
+      num_elements = (*max_bytes * 8 + element_bits - 1) / element_bits;
```
martin-cs: I vaguely recall some hero going around and replacing all of the (appropriate) instances of `8` with `CHAR_BITS`.  Should we call on him again?

martin-cs (approved):
I see why this is needed and why it is useful and I cannot see anything wrong with it.  However, this kind of code is very prone to "off by one" type errors and is a nightmare to diagnose and fix.  So... please can we have a high degree of test coverage?  I realise you probably can't measure that until #6486 is merged but when you can, some test cases, maybe a `union` of an array of bits and a `struct` of some kind?

kroening:
I am reluctant to add more functionality to byte_extract/byte_update, given that these are already the most complicated expressions we have (maybe floating-point stuff can compete).

Would it make sense to separate the aspect of "extract some bytes from byte-aligned memory" from "map some non-byte-aligned encoding to a byte-aligned memory layout"? This might also help with the fat pointers.

tautschnig:
> I am reluctant to add more functionality to byte_extract/byte_update, given that these are already the most complicated expressions we have (maybe floating-point stuff can compete).

I hear you, but I would like to clarify that this PR really only adds functionality to lowering; the propositional back-end will already get this case right (though perhaps has an easier time doing so). It's been a while that I prepared this PR, but I think the reasons it came up were attempts to support more cases of non-det pointers, which just meant that cases that I previously was too lazy to implement were now triggered.

> Would it make sense to separate the aspect of "extract some bytes from byte-aligned memory" from "map some non-byte-aligned encoding to a byte-aligned memory layout"? This might also help with the fat pointers.

The architecture of the lowering implementation is of such nature, although it might not help that it's all baked into a single file. Any byte-extract or byte-update will first turn the underlying object into a byte array, and will then do all further operations on such a an array of bytes.

kroening:
I am still not convinced of the use-case: there shouldn't be pointers into bit-arrays.

peterschrammel (commented):
Isn't this in conflict with https://github.com/diffblue/cbmc/pull/7298?

PR #6489, opened 2021-11-29
Support byte update lowering of/from an array of bits [depends-on: #6488]

tautschnig:
Byte operator lowering made several assumptions about array elements
being byte aligned, which may be true for ANSI C, but isn't the case for
our C front-end (which supports arrays of single bits), and not true for
the overall framework in general.

A key part of this is factoring out code that was almost-the-same from
lower_byte_update_array_vector and lower_byte_update_struct.

Depends on #6488 for tests to pass.


martin-cs (approved):
As with #6488 PLEASE can we have some very extensive testing?

tautschnig:
Closing in favour of #7855.

PR #6490, opened 2021-11-30
C library header de-duplication

tautschnig:
Please review commit-by-commit: the goal of this PR is spelled out in the last commit in this series (removing duplicate-and-conflicting declarations), and the commits before this one enable this work while making sure `library_check.sh` remains happy.

### DIFF
--- a/src/ansi-c/library_check.sh
+++ b/src/ansi-c/library_check.sh
@@ -7,6 +7,7 @@ for f in "$@"; do
     echo "Checking ${f}"
     cp "${f}" __libcheck.c
     perl -p -i -e 's/(__builtin_[^v])/s$1/' __libcheck.c
+    perl -p -i -e 's/s(__builtin_unreachable)/$1/' __libcheck.c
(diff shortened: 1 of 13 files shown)


src/ansi-c/library_check.sh
```diff
@@ -7,6 +7,7 @@ for f in "$@"; do
     echo "Checking ${f}"
     cp "${f}" __libcheck.c
     perl -p -i -e 's/(__builtin_[^v])/s$1/' __libcheck.c
+    perl -p -i -e 's/s(__builtin_unreachable)/$1/' __libcheck.c
```
martin-cs, line 10: Couldn't you just do this for `__CPROVER_assume(0)` instead?

martin-cs (approved):

src/ansi-c/library_check.sh
tautschnig: That would be a valid hack, but then it's only `__CPROVER_assume(0)` that actually makes the code thereafter unreachable, so it's a bit strange perhaps?

kroening (approved):

PR #6491, opened 2021-11-30
Make pointer-primitive-check a no-op when behaviour is always defined

tautschnig:
pointer_object, pointer_offset have
well-defined behaviour even when the input is an unconstrained pointer:
the result is equally unconstrained.

Regression tests are updated to reflect the reduced number of checks
generated by --pointer-primitive-check. Note that
the patterns in pointer-primitive-check-03 never were effective as they
were placed in the patterns-not-to-seen section of test.desc while also
missing proper parenthesis escaping (making the patterns trivially
non-matching).

### DIFF
--- a/doc/architectural/memory-primitives.md
+++ b/doc/architectural/memory-primitives.md
@@ -199,17 +199,21 @@ CBMC has the option `--pointer-primitive-check` to detect potential misuses of
 the memory primitives. It checks that the pointers that appear in the following
 primitives are either null or valid:
 
-- `__CPROVER_POINTER_OBJECT`
-- `__CPROVER_POINTER_OFFSET`
-- `__CPROVER_same_object`
 - `__CPROVER_OBJECT_SIZE`
 - `__CPROVER_DYNAMIC_OBJECT`
 - `__CPROVER_r_ok`
 - `__CPROVER_w_ok`
 
-While the first three primitives have well-defined semantics even on invalid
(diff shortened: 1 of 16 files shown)


kroening (approved):

doc/architectural/memory-primitives.md
```diff
@@ -199,17 +199,21 @@ CBMC has the option `--pointer-primitive-check` to detect potential misuses of
...
+The following three primitives have well-defined semantics even on invalid
+pointers. Thus, they have been excluded from the `--pointer-primitive-check`
+option.
+
+- `__CPROVER_POINTER_OBJECT`
+- `__CPROVER_POINTER_OFFSET`
+- `__CPROVER_same_object`
+
+Using them on invalid pointers, however, may still be unintended in user
+programs.
```
martin-cs, line 216: This is really helpful; thanks for adding it.

martin-cs (approved):

martin-cs:
Is the patch coverage number correct?  If so it would be good to be over the current average.

tautschnig:
> Is the patch coverage number correct? If so it would be good to be over the current average.

Looking at GitHub's annotated diff, nothing in the diff is called out as being uncovered. Also, it would be very much surprising if any C/C++ changes were not covered. Perhaps Codecov measures the coverage against all lines in the diff, including changes to documentation? I wasn't able to figure out from Codecov's documentation how this is computed.

PR #6492, opened 2021-11-30
Fix malloc declaration in Linked_List1 regression test

tautschnig:
We had invalid pointers instead of dynamic objects in this test, which
didn't matter to much. It may, however, matter with upcoming support for
non-deterministic pointers.

### DIFF
--- a/regression/cbmc/Linked_List1/main.c
+++ b/regression/cbmc/Linked_List1/main.c
@@ -1,4 +1,4 @@
-void *malloc(unsigned int s);
+void *malloc(__CPROVER_size_t s);
 
 struct nodet
 {


martin-cs (approved):
Should that be `size_t`?

martin-cs:
https://github.com/diffblue/cbmc/runs/4369109439?check_suite_focus=true looks like an interesting failure.

tautschnig:
> Should that be `size_t`?

We're not including any headers in this file, so I resorted to the frontend-provided internal type instead.

tautschnig:
> https://github.com/diffblue/cbmc/runs/4369109439?check_suite_focus=true looks like an interesting failure.

I have added this to my todo list (seems dump-c is not failing to do some cleanup).

PR #6493, opened 2021-11-30
Invoke goto_functions.update() after inlining

tautschnig:
Inlining will alter location (instruction) numbers, which need to be
updated to once again be globally unique.

This is a follow-up to #6473.

### DIFF
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -773,10 +773,14 @@ void code_contractst::apply_loop_contract(
   const irep_idt &function_name,
   goto_functionst::goto_functiont &goto_function)
 {
-  local_may_aliast local_may_alias(goto_function);
-  natural_loops_mutablet natural_loops(goto_function.body);
+  const bool may_have_loops = std::any_of(
+    goto_function.body.instructions.begin(),
+    goto_function.body.instructions.end(),
+    [](const goto_programt::instructiont &instruction) {
+      return instruction.is_backwards_goto();
+    });
 
-  if(!natural_loops.loop_map.size())
+  if(!may_have_loops)
     return;
 
   inlining_decoratort decorated(log.get_message_handler());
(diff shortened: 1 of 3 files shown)


martin-cs:
https://github.com/diffblue/cbmc/runs/4369109439?check_suite_focus=true seems like an Interesting CI failure.

martin-cs (approved):
Backedges are a necessary but not sufficient test for loops so `may_have_loops` is good.

I have a vague recollection of a previous "Call `update()` after inline" issue / PR.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -682,14 +682,23 @@ void code_contractst::apply_loop_contract(
...
+    return instruction.is_backwards_goto();
+    });
 
-  if(!natural_loops.loop_map.size())
+  if(!may_have_loops)
     return;
 
   goto_function_inline(
     goto_functions, function_name, ns, log.get_message_handler());
+  goto_functions.update();
```
feliperodri: If one should always call them together, wouldn't be safer to add a proper API with both calls? It'd certainly help us to avoid these mistakes.

feliperodri (approved):

src/goto-instrument/contracts/contracts.cpp
tautschnig: Maybe. The problem is the cost of these calls. If you're making multiple modifications to goto programs and don't depend on unique instruction numbers, loop numbering in between, then you should not need to invoke it. There are, in fact, a lot more places in the contracts instrumentation code where such updates need to be considered.

remi-delmas-3000: Could we at least mention in the `goto_inline.h` doc that it breaks the internal consistency of the goto model / function  and that a call to `goto_functionst::update` is needed to restore it? Or maybe in the preconditions of `natural_loops` ? 

What type of modifications require calls to `update` ?

tautschnig: Hmm, I am wondering where to best document (or implement!) this. Any operations that add or remove instructions for a goto program require calls to `update`. I realise that `goto_programt::validate` does not actually check for these properties (as documented with the `location_number` and `loop_number` members of `goto_programt::instructiont`).

remi-delmas-3000 (approved):

PR #6500, opened 2021-12-03
Handle malloc(0) and malloc(__CPROVER_max_malloc_size) in assigns clause instrumentation

remi-delmas-3000:
With the previous upper bound expression`ub = lb + size - 1` pointer overflow were possible when the object size was zero.

The expression has been changed to `ub = lb + size` to allow for a zero size without overflows.

The CAR inclusion tests do not need to be changed.

Added tests to demonstrate that both `malloc(0)` and `malloc(__CPROVER_max_malloc_size)` are handled without issue by CAR instrumentation and contract replacement.

### DIFF
--- a/regression/contracts/assigns-enforce-malloc-zero/main.c
+++ b/regression/contracts/assigns-enforce-malloc-zero/main.c
@@ -1,0 +1,31 @@
+#include <stdlib.h>
+
+// returns the index at which the write was performed if any
+// -1 otherwise
+int foo(char *a, int size)
+  // clang-format off
+  __CPROVER_requires(0 <= size && size <= __CPROVER_max_malloc_size)
+  __CPROVER_requires(a == NULL || __CPROVER_is_fresh(a, size))
+  __CPROVER_assigns(__CPROVER_POINTER_OBJECT(a))
+  __CPROVER_ensures(
+    a && __CPROVER_return_value >= 0 ==> a[__CPROVER_return_value] == 0)
+// clang-format on
+{
+  if(!a)
+    return -1;
+  int i;
+  if(0 <= i && i < size)
+  {
+    a[i] = 0;
+    return i;
+  }
+  return -1;
+}
+
+int main()
+{
+  size_t size;
+  char *a;
+  foo(a, size);
+  return 0;
+}
--- a/regression/contracts/assigns-enforce-malloc-zero/test.desc
+++ b/regression/contracts/assigns-enforce-malloc-zero/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+--enforce-contract foo _ --malloc-may-fail --malloc-fail-null
+^EXIT=0$
+^SIGNAL=0$
+^\[foo\.\d+\] line 19 Check that a\[\(signed long (long )?int\)i\] is assignable: SUCCESS
+^VERIFICATION SUCCESSFUL$
+--
+This test checks that objects of size zero or __CPROVER_max_malloc_size
+do not cause spurious falsifications in assigns clause instrumentation
+in contract enforcement mode.
--- a/regression/contracts/assigns-replace-malloc-zero/main.c
+++ b/regression/contracts/assigns-replace-malloc-zero/main.c
@@ -1,0 +1,37 @@
+#include <stdlib.h>
+
+// returns the index at which the write was performed if any
+// -1 otherwise
+int foo(char *a, int size)
+  // clang-format off
+__CPROVER_requires(0 <= size && size <= __CPROVER_max_malloc_size)
+__CPROVER_requires(a == NULL || __CPROVER_rw_ok(a, size))
+__CPROVER_assigns(__CPROVER_POINTER_OBJECT(a))
+__CPROVER_ensures(
+    a && __CPROVER_return_value >= 0 ==> a[__CPROVER_return_value] == 0)
+// clang-format on
+{
+  if(!a)
+    return -1;
+  int i;
+  if(0 <= i && i < size)
+  {
+    a[i] = 0;
+    return i;
+  }
+  return -1;
+}
+
+int main()
+{
+  int size;
+  if(size < 0)
+    size = 0;
+  if(size > __CPROVER_max_malloc_size)
+    size = __CPROVER_max_malloc_size;
+  char *a = malloc(size * sizeof(*a));
+  int res = foo(a, size);
+  if(a && res >= 0)
+    __CPROVER_assert(a[res] == 0, "expecting SUCCESS");
+  return 0;
+}
--- a/regression/contracts/assigns-replace-malloc-zero/test.desc
+++ b/regression/contracts/assigns-replace-malloc-zero/test.desc
@@ -1,0 +1,11 @@
+CORE
+main.c
+--replace-call-with-contract foo _ --malloc-may-fail --malloc-fail-null
+^EXIT=0$
+^SIGNAL=0$
+\[main\.assertion\.1\] line 35 expecting SUCCESS: SUCCESS$
+^VERIFICATION SUCCESSFUL$
+--
+This test checks that objects of size zero or of __CPROVER_max_malloc_size
+do not cause spurious falsifications in assigns clause instrumentation
+in contract replacement mode.
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -133,8 +133,7 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
 
   instructions.add(goto_programt::make_assignment(
     upper_bound_address_var,
-    minus_exprt{plus_exprt{slice.first, slice.second},
-                from_integer(1, slice.second.type())},
+    plus_exprt{slice.first, slice.second},
     location_overflow_check));
   instructions.destructive_append(skip_program);
 


regression/contracts/assigns-enforce-malloc-zero/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--enforce-contract foo _ --malloc-may-fail
+^EXIT=0$
+^SIGNAL=0$
+^\[foo\.\d+\] line 22 Check that a\[\(signed long int\)i\] is assignable: SUCCESS
```
tautschnig: Use `signed (long (long )?)?int` to ensure this works across several platforms.

regression/contracts/assigns-enforce-malloc-zero/test.desc
```diff
@@ -0,0 +1,11 @@
...
+^VERIFICATION SUCCESSFUL$
+--
+This test checks that object of size zero or of __CPROVER_max_malloc_size
```
tautschnig: Nit: "objects" (plural)

regression/contracts/assigns-enforce-malloc-zero/main.c
```diff
@@ -0,0 +1,34 @@
+#include <stdlib.h>
+
+int MAX_SIZE = __CPROVER_max_malloc_size;
```
tautschnig: Nit pick: I'm not even sure why this is necessary, but if it's just about having a shorter name then a `#define` would seem more idiomatic?

regression/contracts/assigns-replace-malloc-zero/main.c
```diff
@@ -0,0 +1,40 @@
+#include <stdlib.h>
+
+int MAX_SIZE = __CPROVER_max_malloc_size;
```
tautschnig: See comment above.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -133,8 +133,7 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
 
   instructions.add(goto_programt::make_assignment(
     upper_bound_address_var,
-    minus_exprt{plus_exprt{slice.first, slice.second},
-                from_integer(1, slice.second.type())},
+    plus_exprt{slice.first, slice.second},
```
tautschnig, line 136: Why was the "-1" necessary before, or why is this the only change that's necessary?

tautschnig (commented):
I'm not yet clear why there wasn't any need to change a comparison operator in addition to the single arithmetic change. Could you please clarify?

regression/contracts/assigns-enforce-malloc-zero/main.c
jimgrundy: I'm allergic to macros, but if its just about having a shorter name then I'd go with "const int MAX_SIZE = ...;"

src/goto-instrument/contracts/assigns.cpp
jimgrundy: I guess we were expecting some "x <= y" checks to become "x < y" checks, so I too am curious to know why that didn't need to happen.

TGWDB: I'd also be curious to know why changing the actual value used does not appear to break anything. Was there a bug before, or is there an edge case now not being tested?

regression/contracts/assigns-enforce-malloc-zero/test.desc
remi-delmas-3000: fixed

remi-delmas-3000:
> I'm not yet clear why there wasn't any need to change a comparison operator in addition to the single arithmetic change. Could you please clarify?

these expressions define upper bounds for intervals. We just interpret them as being included/excluded of the interval, but the inclusion check for two intervals just compares the upper bounds using `<=`

src/goto-instrument/contracts/assigns.cpp
remi-delmas-3000: these expressions define upper bounds for intervals. We just interpret them as being included/excluded of the interval, but the inclusion check for two intervals just compares the upper bounds using `_ <= _`. Our test suite was missing a case with objects of size 0 and `__CPROVER_max_malloc_size`. My only fear removing the `-1` would be a pointer overflow into the object bits, but it seems like the fact that pointers pointing to one past an object's boundary are legal in C is already factored in `__CPROVER_max_malloc_size`, so everything works.

regression/contracts/assigns-enforce-malloc-zero/main.c
```diff
@@ -0,0 +1,34 @@
...
+int nondet_int();
```
feliperodri: Why do you need this `nondet_int()`?

regression/contracts/assigns-enforce-malloc-zero/main.c
```diff
@@ -0,0 +1,34 @@
...
+  __CPROVER_requires(0 <= size && size <= MAX_SIZE)
+  __CPROVER_requires(a == NULL || __CPROVER_is_fresh(a, size))
+  __CPROVER_assigns(__CPROVER_POINTER_OBJECT(a))
+  __CPROVER_ensures(
+    a && __CPROVER_return_value >= 0 ==> a[__CPROVER_return_value] == 0)
+// clang-format on
+{
+  if(!a)
+    return -1;
+  int i = nondet_int();
```
feliperodri: If you leave it uninitialized it'll have a nondeterministic value.

regression/contracts/assigns-enforce-malloc-zero/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--enforce-contract foo _ --malloc-may-fail
```
feliperodri: You should also include `--malloc-fail-null`.

regression/contracts/assigns-replace-malloc-zero/main.c
```diff
@@ -0,0 +1,40 @@
...
+int nondet_int();
```
feliperodri: Same as above.

regression/contracts/assigns-replace-malloc-zero/main.c
```diff
@@ -0,0 +1,40 @@
...
+__CPROVER_requires(0 <= size && size <= MAX_SIZE)
+__CPROVER_requires(a == NULL || __CPROVER_rw_ok(a, size))
+__CPROVER_assigns(__CPROVER_POINTER_OBJECT(a))
+__CPROVER_ensures(
+    a && __CPROVER_return_value >= 0 ==> a[__CPROVER_return_value] == 0)
+// clang-format on
+{
+  if(!a)
+    return -1;
+  int i = nondet_int();
```
feliperodri: Same as above.

regression/contracts/assigns-replace-malloc-zero/test.desc
```diff
@@ -0,0 +1,11 @@
+CORE
+main.c
+--replace-call-with-contract foo _ --malloc-may-fail
```
feliperodri: Same as above.

regression/contracts/assigns-replace-malloc-zero/test.desc
```diff
@@ -0,0 +1,11 @@
...
+^EXIT=0$
+^SIGNAL=0$
+\[main\.assertion\.1\] line 38 expecting SUCCESS: SUCCESS$
+^VERIFICATION SUCCESSFUL$
+--
+This test checks that object of size zero or of __CPROVER_max_malloc_size
```
feliperodri: objects

feliperodri (changes requested):

regression/contracts/assigns-enforce-malloc-zero/test.desc
remi-delmas-3000: fixed

regression/contracts/assigns-enforce-malloc-zero/main.c
remi-delmas-3000: fixed

regression/contracts/assigns-replace-malloc-zero/main.c
remi-delmas-3000: fixed

regression/contracts/assigns-enforce-malloc-zero/test.desc
remi-delmas-3000:
```
 --malloc-may-fail          allow malloc calls to return a null pointer
 --malloc-fail-null           set malloc failure mode to return null
 ```
 `--malloc-may-fail` seems more general than `--malloc-fail-null`, isn't it ?

File: regression/contracts/assigns-replace-malloc-zero/main.c (hunk 9 above)
Comment (remi-delmas-3000) [reply regression/contracts/assigns-replace-malloc-zero/main.c]: simplified

File: regression/contracts/assigns-enforce-malloc-zero/main.c (hunk 7 above)
Comment (remi-delmas-3000) [reply regression/contracts/assigns-enforce-malloc-zero/main.c]: simplified

File: regression/contracts/assigns-enforce-malloc-zero/test.desc (hunk 8 above)
Comment (feliperodri) [reply regression/contracts/assigns-enforce-malloc-zero/test.desc]: I'd expect the same thing, but you need the second flag to ensure that `malloc` will return NULL on failure. See the following example.
```c
// file main.c
#include <stdlib.h>

int main()
{
  int *i = malloc(0);
  assert(__CPROVER_DYNAMIC_OBJECT(i));
  //assert(i == NULL);
}
```

If you run it using `cbmc main.c --malloc-may-fail` the result is:
```shell
** Results:
<builtin-library-malloc> function malloc
[malloc.assertion.1] line 26 max allocation size exceeded: SUCCESS
[malloc.assertion.2] line 31 max allocation may fail: SUCCESS

main.c function main
[main.assertion.1] line 6 assertion IS_DYNAMIC_OBJECT((const void *)i): SUCCESS

VERIFICATION SUCCESSFUL
```

But if you run using `cbmc main.c --malloc-may-fail  --malloc-fail-null --trace`, the result is:
```shell
** Results:
<builtin-library-malloc> function malloc
[malloc.assertion.1] line 26 max allocation size exceeded: SUCCESS
[malloc.assertion.2] line 31 max allocation may fail: SUCCESS

main.c function main
[main.assertion.1] line 6 assertion IS_DYNAMIC_OBJECT((const void *)i): FAILURE

Trace for main.assertion.1:

State 26 file main.c function main line 5 thread 0
----------------------------------------------------
  i=((signed int *)NULL) (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 30 file main.c function main line 5 thread 0
----------------------------------------------------
  malloc_size=0ul (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

State 41 file main.c function main line 5 thread 0
----------------------------------------------------
  i=((signed int *)NULL) (00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000)

Violated property:
  file main.c function main line 6 thread 0
  assertion IS_DYNAMIC_OBJECT((const void *)i)
  IS_DYNAMIC_OBJECT((const void *)i)



** 1 of 3 failed (2 iterations)
VERIFICATION FAILED
```

remi-delmas-3000: Fixed. 

Remark: The CLI help and doc should be updated to reflect the actual behaviour.

regression/contracts/assigns-replace-malloc-zero/test.desc
remi-delmas-3000: fixed

feliperodri (approved):

Issue #6483, opened 2021-11-28
The function contract doesn't work when using loop


wcventure:
Hi there, 

I followed the [cprover-manual/contracts-functions.md](https://github.com/diffblue/cbmc/blob/develop/doc/cprover-manual/contracts-functions.md) and tried to use function contracts in my project. However, the function contracts don't seem to work correctly in some situations. Could you help me look at the following issue?

### The `foo` function contract doesn't work when using it in the loop body

I have written a function `foo` that Iteratively checks whether the two arrays have the same elements. If yes, it returns 0, otherwise, it returns -1. And I try to use its function contract in the `main` function. Unfortunately, I got an unexpected result that all wrong assertions are Judged as SUCCESS (When this function contract is not placed in the loop, the result is correct. But when it is put into a meaningless loop, the result is not correct).

```C
//main.c
#include <stdlib.h>
#include <string.h>
#include <stdint.h>
#include <assert.h>

#define SIZE 10

int foo(uint32_t *map, uint32_t *key, uint32_t len)
/* Precondition */
__CPROVER_requires(len <= SIZE)
__CPROVER_requires(__CPROVER_is_fresh(map, sizeof(uint32_t)*len))
__CPROVER_requires(__CPROVER_is_fresh(key, sizeof(uint32_t)*len))
/* Postconditions */
__CPROVER_ensures(__CPROVER_return_value == 0 || __CPROVER_return_value == -1)
__CPROVER_ensures(__CPROVER_return_value == -1 ==> __CPROVER_forall {int i; (0 <= i && i < len) ==> map[i] != key[i]})
__CPROVER_ensures(__CPROVER_return_value == 0 ==> __CPROVER_exists {int i; (0 <= i && i < len) && map[i] == key[i]})
{
	int ret = -1;
	for (int i = 0; i < len; i++)
	__CPROVER_assigns(i)
	__CPROVER_loop_invariant(0 <= i && i <= len)
	__CPROVER_loop_invariant(ret == -1 || ret == 0)
	__CPROVER_loop_invariant((ret == -1)==> __CPROVER_forall {int j; (0 <= j && j < len) ==> ( j < i ==> map[j] != key[j])})
	__CPROVER_loop_invariant((ret == 0) ==> map[i] == key[i])
	__CPROVER_decreases(len - i)
	{
		if (map[i] == key[i])
		{
			ret = 0;
			break;
		}
	}
	return ret;
}

int main()
{
[...]

PR #6502, opened 2021-12-03
havoc call return value when replacing a call by its contract

remi-delmas-3000:
This PR adds havoc instructions for the return value of function call when it gets replaced by the function's contract.

This havocking was missing previously and could result in vacuous instances when a same function was replaced more than once in a same scope (e.g. in loops), and the same return value variable was subjected to a superposition of post-conditions that can be contradicting each other.

- [ x] Each commit message has a non-empty body, explaining why the change was made.
- [ x] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).
- [ x] My PR is restricted to a single feature or bugfix.

### DIFF
--- a/regression/contracts/replace-nondet-return-value/main.c
+++ b/regression/contracts/replace-nondet-return-value/main.c
@@ -1,0 +1,23 @@
+int cmp(int i1, int i2)
+  // clang-format off
+  __CPROVER_ensures((i1 == i2) ==> (__CPROVER_return_value == 0))
+  __CPROVER_ensures((i1 != i2) ==> (__CPROVER_return_value == -1))
+// clang-format on
+{
+  if(i1 == i2)
+    return 0;
+  else
+    return -1;
+}
+
+int main()
+{
+  int ret = -1;
+  ret = cmp(0, 0);
+  __CPROVER_assert(ret == 0, "expecting SUCCESS");
+  ret = cmp(0, 1);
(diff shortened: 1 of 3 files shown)


regression/contracts/replace-nondet-return-value/main.c
```diff
@@ -0,0 +1,23 @@
+int cmp(int i1, int i2)
```
tautschnig, line 1: Random place for comment: can you please add "Fixes: #6483" to the commit message so that the issue is auto-resolved once this PR is merged?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -23,6 +23,7 @@ Date: February 2016
 #include <goto-instrument/havoc_utils.h>
 
 #include <goto-programs/goto_inline.h>
+#include <goto-programs/goto_program.h>
```
tautschnig, line 26: I don't think this is strictly necessary? (Even the `remove_skip.h` will already drag in this header.)

tautschnig (approved):

regression/contracts/replace-nondet-return-value/main.c
tautschnig: I'll go ahead and merge anyway, so this is just for future reference.

PR #6506, opened 2021-12-04
Track deallocated/dead in an array

tautschnig:
This enables tracking of all deallocations and all dead objects, and
thereby use of deallocated/dead_object predicates within assumptions.

This requires performance evaluation.

### DIFF
--- a/regression/cbmc/array_of_bool_as_bitvec/test-smt2-outfile.desc
+++ b/regression/cbmc/array_of_bool_as_bitvec/test-smt2-outfile.desc
@@ -1,13 +1,13 @@
 CORE broken-smt-backend
 main.c
 --smt2 --outfile -
-\(= \(select array_of\.2 i\) \(ite false #b1 #b0\)\)
+\(= \(select array_of\.\d+ i\) \(ite false #b1 #b0\)\)
 \(= \(select array\.3 \(\(_ zero_extend 32\) \|main::1::idx!0@1#1\|\)\) #b1\)
 \(= \(select array\.3 \(_ bv\d+ 64\)\) \(ite false #b1 #b0\)\)
 ^EXIT=0$
 ^SIGNAL=0$
 --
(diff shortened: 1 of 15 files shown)


tautschnig:
Screenshot 2022-03-29 at 20 05 15

The above scatter plot shows current develop (y axis) vs this proposed change (x axis) on SV-COMP's MemSafety-Heap category. This one is the benchmark set with the strongest negative signal I have found so far (several others show no significant difference, some even an improvement): on this benchmark set, CBMC is now only able to solve 140 instead of 145 tasks within the resource limits. Three of those that can no longer be solved now run out of memory, two trigger a bug (fixed in #6772). This is perhaps to be expected for we end up making use of the array theory, which may generate a lot of extra constraints.

Despite this data, I'd like to see this change being considered for inclusion as it would alleviate subtle usage limitations of `__CPROVER_r_ok` (it would now be safe to be used in assumptions), and might even make it possible to get rid of `--pointer-primitive-check` at some point, which is an option that is hard to explain to users.

peterschrammel (commented):
What are the special characteristics of the benchmark that is such an outlier?

tautschnig:
> What are the special characteristics of the benchmark that is such an outlier?

I'd say: a lot of objects to track where constant propagation might be unable to pick out one specific object each time there's a read to the array newly introduced in this PR. Two trouble makers are https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/list-ext-properties/test-0217_1.c and https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/memsafety/test-0235-3.c.

There are also two tests that appear to cause an invariant failure that may point to a bug elsewhere.

tautschnig:
> There are also two tests that appear to cause an invariant failure that may point to a bug elsewhere.

This bug is now addressed by #6772. The more frequent use of this code via this PR just made it visible.

PR #6521, opened 2021-12-08
Evaluate any non-trivial switch condition just once

tautschnig:
goto-program conversion should not turn switch(E) { case 1:... case 2:
...} into if(E == 1 || E == 2 || ...) for a non-trivial expression E.
Instead, the value of E must be copied into a temporary to avoid
repeated evaluation of E. We would already do this when E had side
effects, but even side-effect-free expression could result in expensive
assertions being generated when those expressions themselves would
trigger them with goto_check.

This change reduces verification time of a proof of
s2n_stuffer_skip_whitespace from several minutes down to seconds.

### DIFF
--- a/src/goto-programs/goto_convert.cpp
+++ b/src/goto-programs/goto_convert.cpp
@@ -1184,6 +1184,27 @@ void goto_convertt::convert_switch(
   goto_programt sideeffects;
   clean_expr(argument, sideeffects, mode);
 
+  // Avoid potential performance penalties caused by evaluating the value
+  // multiple times (as the below chain-of-ifs does). "needs_cleaning" isn't
+  // necessarily the right check here, and we may need to introduce a different
+  // way of identifying the class of non-trivial expressions that warrant
+  // introduction of a temporary.
+  if(needs_cleaning(argument))
+  {
+    symbolt &new_symbol = new_tmp_symbol(
+      argument.type(),
+      "switch_value",
+      sideeffects,
+      code.source_location(),
+      mode);
+
+    code_assignt copy_value{
+      new_symbol.symbol_expr(), argument, code.source_location()};
+    convert(copy_value, sideeffects, mode);
+
+    argument = new_symbol.symbol_expr();
+  }
+
   // save break/default/cases targets
   break_switch_targetst old_targets(targets);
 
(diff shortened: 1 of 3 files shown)


martin-cs:
Why does expression caching in the solver not catch this?

tautschnig:
> Why does expression caching in the solver not catch this?

If you're asking about the performance numbers I indicated: for the regression test included in this PR, `goto_check_ct` will find an expression along the lines of `IF (*p == 1 || *p == 2 || *p == 3) ...` and will build a chain of assertions along the lines of:
```
ASSERT p is not NULL
ASSERT *p != 1 ==> p is not NULL
ASSERT *p != 1 && *p != 2 ==> p is not NULL
```
and similarly for the various other pointer checks. Various sub-expressions will be cached, but those assertions overall have expressions never seen before.

src/goto-programs/goto_convert.cpp
```diff
@@ -1184,6 +1184,24 @@ void goto_convertt::convert_switch(
   goto_programt sideeffects;
   clean_expr(argument, sideeffects, mode);
 
+  // make sure evaluating the value multiple times (as the below chain-of-ifs
+  // does) doesn't have unintended effects
+  if(needs_cleaning(argument))
```
kroening, line 1192: I would change the comment: "unintended effects" sounds a bit like "side effect", which should be removed at this point by `clean_expr`.

I'd be clear that this is about reducing the cost of multiple copies of `argument`. This could well be independent of `needs_cleaning`. E.g., forall/exist "do not need cleaning", but may well be expensive.

tautschnig: Makes sense, but I did wonder about concurrency: in that situation, repeat evaluation might actually change the semantics of the program. Not that `needs_cleaning` currently covers this, though...

kroening: Concurrency is naughty. Repeated evaluation would certainly be something a compiler would be allowed to do. One-time evaluation would only be guaranteed if the variables in the expression are `volatile`.

tautschnig: Comment updated to reflect your input.

kroening: `does) does` -> `does)`

kroening (approved):

src/goto-programs/goto_convert.cpp
tautschnig: Thank you for careful proof reading!

kroening (approved):

PR #6525, opened 2021-12-09
local_bitvector_analysist: track whether an object may be alive

tautschnig:
A local variable is alive if it has been declared and not yet been marked dead.

### DIFF
--- a/regression/goto-instrument/show-local-bitvector-analysis1/main.c
+++ b/regression/goto-instrument/show-local-bitvector-analysis1/main.c
@@ -1,0 +1,18 @@
+int foo(int i)
+{
+  int x;
+  if(x > 0)
+  {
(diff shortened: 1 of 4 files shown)


kroening:
Does it make sense to move the "is live" analysis into a separate class? The bitvector analysis does seem to have different concerns.

tautschnig:
> Does it make sense to move the "is live" analysis into a separate class? The bitvector analysis does seem to have different concerns.

But then we would perhaps need a more abstract concept of a bitvector analysis and different specialisations? Isn't it that the existing bits all just happen to track the specific set of information that's convenient for `goto_check`?

kroening:
Yes, indeed, it makes sense to factor out a base class.

PR #6527, opened 2021-12-11
Skip function parameters and auxiliary variables when checking assigns clauses and check loop-freeness

remi-delmas-3000:
This PR skips some unnecessary instrumentation when checking assigns clauses.

# Skipping function parameters
Assignments to variables storing function parameters, as given by `symbolt::is_parameter` are always legal so we detect them and skip their instrumentation. We do however still add their CARs to the function's write set since they may be aliased and their addresses passed to sub-function calls. 

# Skipping auxiliary variables
Auxiliary variables, as given by `symbolt::is_auxiliary` are introduced by the goto compiler to decompose expressions in subexpressions, to store return values for function calls, etc. are only assigned once, cannot have aliases and are really just names for expressions. So we do not add them to the write set anymore, and we skip their assignments in the instrumentation.

# Checking loop freeness
We add a check for loop-freeness of the goto-program before attempting a function's assigns clause instrumentation. The assigns clause instrumentation is not sound in the presence of loops, and we did not check this condition before.
To detect loops we compute strongly connected components of the function's control flow graph and detect components that have a cardinality strictly greater than 1.

### DIFF
--- a/regression/contracts/function_check_02/main.c
+++ b/regression/contracts/function_check_02/main.c
@@ -15,10 +15,16 @@ int initialize(int *arr)
   )
 // clang-format on
 {
-  for(int i = 0; i < 10; i++)
-  {
-    arr[i] = i;
-  }
+  arr[0] = 0;
+  arr[1] = 1;
+  arr[2] = 2;
+  arr[3] = 3;
+  arr[4] = 4;
+  arr[5] = 5;
+  arr[6] = 6;
+  arr[7] = 7;
+  arr[8] = 8;
+  arr[9] = 9;
 
   return 0;
 }
--- a/regression/contracts/history-pointer-enforce-10/test.desc
+++ b/regression/contracts/history-pointer-enforce-10/test.desc
@@ -7,8 +7,6 @@ main.c
 ^\[postcondition.\d+\] file main.c line \d+ Check ensures clause: SUCCESS$
 ^\[postcondition.\d+\] file main.c line \d+ Check ensures clause: SUCCESS$
 ^\[bar.\d+\] line \d+ Check that p->y is assignable: SUCCESS$
-^\[baz.\d+\] line \d+ Check that p is assignable: SUCCESS$
-^\[baz.\d+\] line \d+ Check that p is assignable: SUCCESS$
 ^\[foo.\d+\] line \d+ Check that \*p->y is assignable: SUCCESS$
 ^\[foo.\d+\] line \d+ Check that z is assignable: SUCCESS$
 ^\[main.assertion.\d+\] line \d+ assertion \*\(p->y\) == 7: SUCCESS$
@@ -19,3 +17,5 @@ main.c
 This test checks that history variables are supported for structs, symbols, and
 struct members. By using the --enforce-contract flag, the postcondition
 (with history variable) is asserted. In this case, this assertion should pass.
+Note: A function is always authorized to assign the variables that store
+its arguments, there is no need to mention them in the assigns clause.
--- a/regression/contracts/loop-freeness-check/main.c
+++ b/regression/contracts/loop-freeness-check/main.c
@@ -1,0 +1,16 @@
+int foo(int *arr)
+  // clang-format off
+  __CPROVER_assigns(__CPROVER_POINTER_OBJECT(arr))
+// clang-format off
+{
+  for(int i = 0; i < 10; i++)
+    arr[i] = i;
+
+  return 0;
+}
+
+int main()
+{
+  int arr[10];
+  foo(arr);
+}
--- a/regression/contracts/quantifiers-forall-requires-enforce/main.c
+++ b/regression/contracts/quantifiers-forall-requires-enforce/main.c
@@ -12,8 +12,16 @@ bool f1(int *arr)
 // clang-format on
 {
   bool is_identity = true;
-  for(int i = 0; i < 10; ++i)
-    is_identity &= (arr[i] == i);
+  is_identity &= (arr[0] == 0);
+  is_identity &= (arr[1] == 1);
+  is_identity &= (arr[2] == 2);
+  is_identity &= (arr[3] == 3);
+  is_identity &= (arr[4] == 4);
+  is_identity &= (arr[5] == 5);
+  is_identity &= (arr[6] == 6);
+  is_identity &= (arr[7] == 7);
+  is_identity &= (arr[8] == 8);
+  is_identity &= (arr[9] == 9);
   return is_identity;
 }
 
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -309,7 +309,13 @@ void code_contractst::check_apply_loop_contracts(
 
   // Perform write set instrumentation on the entire loop.
   check_frame_conditions(
-    function_name, goto_function.body, loop_head, loop_end, loop_assigns);
+    function_name,
+    goto_function.body,
+    loop_head,
+    loop_end,
+    loop_assigns,
+    // do not skip checks on function parameter assignments
+    false);
 
   havoc_assigns_targetst havoc_gen(to_havoc, ns);
   havoc_gen.append_full_havoc_code(
@@ -1171,7 +1177,9 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     function_obj->second.body,
     instruction_it,
     function_obj->second.body.instructions.end(),
-    assigns);
+    assigns,
+    // skip checks on function parameter assignments
+    true);
 
   return false;
 }
@@ -1181,12 +1189,19 @@ void code_contractst::check_frame_conditions(
   goto_programt &body,
   goto_programt::targett instruction_it,
   const goto_programt::targett &instruction_end,
-  assigns_clauset &assigns)
+  assigns_clauset &assigns,
+  bool skip_parameter_assigns)
 {
   for(; instruction_it != instruction_end; ++instruction_it)
   {
     const auto &pragmas = instruction_it->source_location().get_pragmas();
     if(pragmas.find(CONTRACT_PRAGMA_DISABLE_ASSIGNS_CHECK) != pragmas.end())
+      continue;
+
+    if(skip_parameter_assigns && is_parameter_assign(instruction_it, ns))
+      continue;
+
+    if(is_auxiliary_decl_dead_or_assign(instruction_it, ns))
       continue;
 
     // Do not instrument this instruction again in the future,
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -1128,7 +1128,9 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     return true;
   }
 
-  if(check_for_looped_mallocs(function_obj->second.body))
+  auto &function_body = function_obj->second.body;
+
+  if(check_for_looped_mallocs(function_body))
   {
     return true;
   }
@@ -1142,20 +1144,23 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     function,
     symbol_table);
 
-  // detect and add static local variables
+  // Detect and add static local variables
   assigns.add_static_locals_to_write_set(goto_functions, function);
 
   // Add formal parameters to write set
   for(const auto &param : to_code_type(target.type).parameters())
     assigns.add_to_write_set(ns.lookup(param.get_identifier()).symbol_expr());
 
-  auto instruction_it = function_obj->second.body.instructions.begin();
+  auto instruction_it = function_body.instructions.begin();
   for(const auto &car : assigns.get_write_set())
   {
     auto snapshot_instructions = car.generate_snapshot_instructions();
     insert_before_swap_and_advance(
-      function_obj->second.body, instruction_it, snapshot_instructions);
+      function_body, instruction_it, snapshot_instructions);
   };
+
+  // Restore internal coherence in the programs
+  goto_functions.update();
 
   // Full inlining of the function body
   // Inlining is performed so that function calls to a same function
@@ -1168,15 +1173,23 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     decorated.get_recursive_function_warnings_count() == 0,
     "Recursive functions found during inlining");
 
-  // restore internal invariants
+  // Clean up possible fake loops that are due to `IF 0!=0 GOTO i` instructions
+  simplify_gotos(function_body, ns);
+
+  // Restore internal coherence in the programs
   goto_functions.update();
+
+  INVARIANT(
+    is_loop_free(function_body, ns, log),
+    "Loops remain in function '" + id2string(function) +
+      "', assigns clause checking instrumentation cannot be applied.");
 
   // Insert write set inclusion checks.
   check_frame_conditions(
-    function_obj->first,
-    function_obj->second.body,
+    function,
+    function_body,
     instruction_it,
-    function_obj->second.body.instructions.end(),
+    function_body.instructions.end(),
     assigns,
     // skip checks on function parameter assignments
     true);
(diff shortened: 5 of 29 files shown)


regression/contracts/function_check_02/main.c
```diff
@@ -15,10 +15,16 @@ int initialize(int *arr)
...
+  arr[0] = 0;
+  arr[1] = 1;
+  arr[2] = 2;
+  arr[3] = 3;
+  arr[4] = 4;
+  arr[5] = 5;
+  arr[6] = 6;
+  arr[7] = 7;
+  arr[8] = 8;
+  arr[9] = 9;
```
feliperodri, line 27: Why remove this loop?

regression/contracts/history-pointer-enforce-10/test.desc
```diff
@@ -19,3 +17,5 @@ main.c
 This test checks that history variables are supported for structs, symbols, and
 struct members. By using the --enforce-contract flag, the postcondition
 (with history variable) is asserted. In this case, this assertion should pass.
+Note: A function is always authorized to assign the variables that store
+its arguments, there is no need to mention them in the assigns clause.
```
feliperodri: Add newline.

regression/contracts/loop-freeness-check/main.c
```diff
@@ -0,0 +1,16 @@
+// clang-format off
+int foo(int *arr)
+  __CPROVER_assigns(*arr)
```
feliperodri: I guess it doesn't really matter for this case, but we probably want here `__CPROVER_assigns(__CPROVER_POINTER_OBJECT(arr))`.

regression/contracts/quantifiers-forall-requires-enforce/main.c
```diff
@@ -12,8 +12,16 @@ bool f1(int *arr)
...
+  is_identity &= (arr[0] == 0);
+  is_identity &= (arr[1] == 1);
+  is_identity &= (arr[2] == 2);
+  is_identity &= (arr[3] == 3);
+  is_identity &= (arr[4] == 4);
+  is_identity &= (arr[5] == 5);
+  is_identity &= (arr[6] == 6);
+  is_identity &= (arr[7] == 7);
+  is_identity &= (arr[8] == 8);
+  is_identity &= (arr[9] == 9);
```
feliperodri, line 24: Random place for a comment: we should at least create a GitHub issue to improve our `chain.sh` script to allow multiple instrumentation steps.

src/goto-instrument/contracts/contracts.h
```diff
@@ -211,6 +212,24 @@ class code_contractst
...
+  bool is_auxiliary_decl_dead_or_assign(
+    const goto_programt::targett &instruction_it);
+
+  /// Turns goto instructions `IF cond GOTO label` where the condition
+  /// statically simplifies to `false` into SKIP instructions.
+  void simplify_gotos(goto_programt &goto_program);
+
+  /// Returns true iff the given program is loop-free,
+  /// i.e. if each SCC of its CFG contains a single element.
+  bool is_loop_free(const goto_programt &goto_program);
```
feliperodri: All these functions should go to [contracts/utils.h](https://github.com/diffblue/cbmc/blob/develop/src/goto-instrument/contracts/utils.h)

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -285,7 +286,13 @@ void code_contractst::check_apply_loop_contracts(
 
     // Perform write set instrumentation on the entire loop.
     check_frame_conditions(
-      function_name, goto_function.body, loop_head, loop_end, loop_assigns);
+      function_name,
+      goto_function.body,
+      loop_head,
+      loop_end,
+      loop_assigns,
+      // do not skip checks on function parameter assignments
```
feliperodri: Why not?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1099,7 +1106,9 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     return true;
   }
 
-  if(check_for_looped_mallocs(function_obj->second.body))
+  auto &goto_program = function_obj->second.body;
```
feliperodri: nit: I'd name it `function_body`

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1120,14 +1129,17 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+  auto instruction_it = goto_program.instructions.begin();
   for(const auto &car : assigns.get_write_set())
   {
     auto snapshot_instructions = car.generate_snapshot_instructions();
     insert_before_swap_and_advance(
-      function_obj->second.body, instruction_it, snapshot_instructions);
+      goto_program, instruction_it, snapshot_instructions);
   };
 
+  // restore internal coherence in the programs
```
feliperodri: nit:
```suggestion
// Restore internal coherence in the programs
```

feliperodri: Same for all others.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1139,16 +1151,26 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+  // clean up possible fake loops that are due to `IF 0!=0 GOTO i` instructions
+  simplify_gotos(goto_program);
+
+  // restore internal coherence in the programs
   goto_functions.update();
 
+  INVARIANT(
+    is_loop_free(goto_program),
+    "Loops remain in function '" + id2string(function) +
+      "', assigns clause checking instrumentation cannot be applied.");
```
feliperodri: Nice!

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1553,3 +1582,107 @@ bool code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
   }
   return fail;
 }
+
+bool code_contractst::is_parameter_assign(
```
feliperodri: Move all implementation to [contracts/utils.cpp](https://github.com/diffblue/cbmc/blob/develop/src/goto-instrument/contracts/utils.cpp)

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1553,3 +1582,107 @@ bool code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
...
+    const auto &lhs = instruction_it->assign_lhs();
+    if(can_cast_expr<symbol_exprt>(lhs))
+      sym = to_symbol_expr(lhs);
+  }
+
+  // check condition
+  if(sym.has_value())
+    return ns.lookup(sym.value().get_identifier()).is_parameter;
+
+  return false;
```
```suggestion
  if(instruction_it->is_assign() && can_cast_expr<symbol_exprt>(instruction_it->assign_lhs()))
  {
    // Extract symbol from lhs
    optionalt<symbol_exprt> lhs_sym = to_symbol_expr(instruction_it->assign_lhs());
    // Check whether is parameter
    if(lhs_sym.has_value())
      return ns.lookup(lhs_sym.value().get_identifier()).is_parameter;
  }

  return false;
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1553,3 +1582,107 @@ bool code_contractst::enforce_contracts(const std::set<std::string> &to_enforce)
...
+
+  // check they are all of size 1
+  for(size_t scc_id = 0; scc_id < nof_sccs; scc_id++)
+  {
+    auto size = scc_size[scc_id];
+    if(size > 1)
+    {
+      log.error() << "Found CFG SCC with size " << size << messaget::eom;
+      for(const auto &node_id : node_to_scc)
+      {
```
feliperodri: I guess there is no way to exercise these failures from a C program, correct?

feliperodri (changes requested):

regression/contracts/function_check_02/main.c
remi-delmas-3000: This PR adds a loop freeness check before running the instrumentation, so the loop needs to be eliminated to satisfy this loop-freeness check.
This manual unrolling of loops is just a quick and dirty fix. A better solution would be to modify `chain.sh` add a pass of loop unrolling or loop contracts instrumentation.

regression/contracts/history-pointer-enforce-10/test.desc
remi-delmas-3000: fixed

regression/contracts/loop-freeness-check/main.c
remi-delmas-3000: good catch!

regression/contracts/quantifiers-forall-requires-enforce/main.c
remi-delmas-3000: yes.

src/goto-instrument/contracts/contracts.h
remi-delmas-3000: Do you see a potential for using these functions outside of `code_contractst` ?

feliperodri: Maybe, but the point is that we already have a module design to hold all "util" operations specific to contract instrumentation. We should aim to keep our core `code_contractst` class as simple as possible.

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: I expect this to always be true hence expressed it as an INVARIANT.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1122,7 +1128,9 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     return true;
   }
 
-  if(check_for_looped_mallocs(function_obj->second.body))
+  auto &function_body = function_obj->second.body;
+
+  if(check_for_looped_mallocs(function_body))
```
feliperodri, line 1133: This check is no longer necessary, you can remove it. It checks whether there is malloc within loops, but we assume contract instrumentation happens after all loops have been unrolled.

src/goto-instrument/contracts/utils.cpp
```diff
@@ -163,3 +167,118 @@ void disable_pointer_checks(source_locationt &source_location)
...
+    const auto &lhs = instruction_it->assign_lhs();
+    if(can_cast_expr<symbol_exprt>(lhs))
+      sym = to_symbol_expr(lhs);
+  }
+
+  // check condition
+  if(sym.has_value())
+    return ns.lookup(sym.value().get_identifier()).is_parameter;
+
+  return false;
```
feliperodri, line 189:
```suggestion
  if(instruction_it->is_assign() && can_cast_expr<symbol_exprt>(instruction_it->assign_lhs()))
  {
    // Extract symbol from lhs
    auto lhs_sym = to_symbol_expr(instruction_it->assign_lhs());
    // Check whether is parameter
    if(lhs_sym.has_value())
      return ns.lookup(lhs_sym.value().get_identifier()).is_parameter;
  }

  return false;
```

src/goto-instrument/contracts/utils.cpp
```diff
@@ -163,3 +167,118 @@ void disable_pointer_checks(source_locationt &source_location)
...
+    if(can_cast_expr<symbol_exprt>(lhs))
+      sym = to_symbol_expr(lhs);
+  }
+
+  // check condition
+  if(sym.has_value())
+    return ns.lookup(sym.value().get_identifier()).is_auxiliary;
+
+  return false;
+}
```
feliperodri, line 215:
```suggestion
bool is_auxiliary_decl_dead_or_assign(
  const goto_programt::targett &instruction_it,
  namespacet &ns)
{
  // Extract symbol
  optionalt<symbol_exprt> sym = {};
  if(instruction_it->is_assign() && can_cast_expr<symbol_exprt>(instruction_it->assign_lhs()))
      sym = to_symbol_expr(instruction_it->assign_lhs());
  else if(instruction_it->is_decl())
    sym = instruction_it->decl_symbol();
  else if(instruction_it->is_dead())
    sym = instruction_it->dead_symbol();
  else
    return false;

  // Check whether is auxiliary
  return ns.lookup(sym.value().get_identifier()).is_auxiliary;
}
```

src/goto-instrument/contracts/utils.cpp
```diff
@@ -163,3 +167,118 @@ void disable_pointer_checks(source_locationt &source_location)
...
+  {
+    if(instruction.is_goto())
+    {
+      const auto &condition = instruction.get_condition();
+      const auto &simplified = simplify_expr(condition, ns);
+      if(simplified.is_false())
+        instruction.turn_into_skip();
+    }
+  }
+}
```
feliperodri, line 229:
```suggestion
void simplify_gotos(goto_programt &goto_program, namespacet &ns)
{
  for(auto &it : goto_program.instructions)
    if(it.is_goto() && simplify_expr(it.get_condition(), ns).is_false())
        it.turn_into_skip();
}
```

src/goto-instrument/contracts/utils.cpp
```diff
@@ -163,3 +167,118 @@ void disable_pointer_checks(source_locationt &source_location)
...
+
+  // Check they are all of size 1
+  bool all_size_one = true;
+
+  for(size_t scc_id = 0; scc_id < nof_sccs; scc_id++)
+  {
+    auto size = scc_size[scc_id];
+    if(size > 1)
+    {
+      all_size_one = false;
```
feliperodri, line 267: We could add a regression test to explore these paths.

feliperodri (approved):
Only a couple of code suggestions.

src/goto-instrument/contracts/utils.cpp
tautschnig: On top of what Felipe suggested: there's no need for the `optionalt` anymore.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -309,7 +309,13 @@ void code_contractst::check_apply_loop_contracts(
...
   // Perform write set instrumentation on the entire loop.
   check_frame_conditions(
-    function_name, goto_function.body, loop_head, loop_end, loop_assigns);
+    function_name,
+    goto_function.body,
+    loop_head,
+    loop_end,
+    loop_assigns,
+    // do not skip checks on function parameter assignments
+    false);
```
tautschnig, line 318: If we see a need to comment on the Boolean value then it may be better to introduce an `enum` that has informative values. If we always pass constant values, then it might be even better to use a template parameter.

src/goto-instrument/contracts/utils.h
```diff
@@ -112,17 +112,29 @@ void insert_before_swap_and_advance(
...
+/// \param is_auxiliary: Do not print symbol in traces if true (default = true)
 /// \return The new symbolt object.
 const symbolt &new_tmp_symbol(
   const typet &type,
   const source_locationt &location,
   const irep_idt &mode,
   symbol_table_baset &symtab,
   std::string suffix = "tmp_cc",
-  bool is_auxiliary = false);
+  bool is_auxiliary = true);
```
tautschnig, line 125: Why is this being changed as part of this PR?

src/goto-instrument/contracts/utils.cpp
```diff
@@ -210,3 +214,71 @@ bool is_auxiliary_decl_dead_or_assign(
...
+
+bool is_loop_free(
+  const goto_programt &goto_program,
+  namespacet &ns,
+  messaget &log)
+{
+  // create cfg from instruction list
+  cfg_baset<empty_cfg_nodet> cfg;
+  cfg(goto_program);
+
```
tautschnig, line 239: You wouldn't have quite as elaborate error messages, but wouldn't a simple test of all instructions for `is_backwards_goto()` do the trick?

remi-delmas-3000:
superseded by https://github.com/diffblue/cbmc/pull/6533

PR #6528, opened 2021-12-15
Performance improvement and cleanup of local bitvector analysis

tautschnig:
Please review commit-by-commit. The commit messages include performance evaluation data.

### DIFF
--- a/src/analyses/local_bitvector_analysis.cpp
+++ b/src/analyses/local_bitvector_analysis.cpp
@@ -45,10 +45,7 @@ bool local_bitvector_analysist::merge(points_tot &a, points_tot &b)
     std::max(a.size(), b.size());
 
   for(std::size_t i=0; i<max_index; i++)
-  {
-    if(a[i].merge(b[i]))
-      result=true;
-  }
+    result |= a[i].merge(b[i]);
 
   return result;
 }
--- a/src/analyses/local_bitvector_analysis.cpp
+++ b/src/analyses/local_bitvector_analysis.cpp
@@ -70,7 +70,7 @@ void local_bitvector_analysist::assign_lhs(
 
     if(is_tracked(identifier))
     {
-      unsigned dest_pointer=pointers.number(identifier);
+      const auto dest_pointer = pointers.number(identifier);
       flagst rhs_flags=get_rec(rhs, loc_info_src);
       loc_info_dest[dest_pointer]=rhs_flags;
     }
@@ -125,7 +125,7 @@ local_bitvector_analysist::flagst local_bitvector_analysist::get_rec(
     const irep_idt &identifier=to_symbol_expr(rhs).get_identifier();
     if(is_tracked(identifier))
     {
-      unsigned src_pointer=pointers.number(identifier);
+      const auto src_pointer = pointers.number(identifier);
       return loc_info_src[src_pointer];
     }
     else
@@ -239,7 +239,7 @@ void local_bitvector_analysist::build()
   if(cfg.nodes.empty())
     return;
 
-  std::set<unsigned> work_queue;
+  std::set<local_cfgt::node_nrt> work_queue;
   work_queue.insert(0);
 
   loc_infos.resize(cfg.nodes.size());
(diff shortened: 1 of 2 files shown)


src/analyses/local_bitvector_analysis.cpp
```diff
@@ -45,10 +45,7 @@ bool local_bitvector_analysist::merge(points_tot &a, points_tot &b)
     std::max(a.size(), b.size());
 
   for(std::size_t i=0; i<max_index; i++)
-  {
-    if(a[i].merge(b[i]))
-      result=true;
-  }
+    result |= a[i].merge(b[i]);
```
kroening, line 48: I would add a comment that the bitwise or is used instead of the logical or to prevent the short-circuit semantics.

src/analyses/local_bitvector_analysis.cpp
```diff
@@ -242,8 +239,8 @@ void local_bitvector_analysist::build()
   if(cfg.nodes.empty())
     return;
 
-  work_queuet work_queue;
-  work_queue.push(0);
+  std::set<local_cfgt::node_nrt> work_queue;
+  work_queue.insert(0);
```
kroening, line 244: Queues are usually ordered (especially in the UK). Is the set really helping performance, and if so, I'd rename it to "work_set".

kroening:
Commit message: "enqueuing successor nodes" -> "dequeuing successor nodes"

tautschnig:
> Commit message: "enqueuing successor nodes" -> "dequeuing successor nodes"

Err, of course, thank you!

src/analyses/local_bitvector_analysis.cpp
```diff
@@ -242,8 +242,8 @@ void local_bitvector_analysist::build()
   if(cfg.nodes.empty())
     return;
 
-  work_queuet work_queue;
```
martin-cs, line 245: This one has a friend who would also like to become a set:

https://github.com/diffblue/cbmc/blob/eeddb3feb756be04e08709d9989d8ed2c36d793c/src/analyses/local_may_alias.h#L61

martin-cs (approved):
20x speed-up due to a change in data-structure!  I love it!  Classic computer science.  Would it be possible to have another PR that s/`stack`/`set`/ in `local_may_aliast` too?

src/analyses/local_bitvector_analysis.cpp
tautschnig: Good call, I'll do so in a follow-up PR.

remi-delmas-3000 (commented):
Can this result in a different fixed point being reached ? What are the implications for the precision of the analysis ? Are there any iteration strategies that can result in more precise results ? In any acceptable perf tradeoff that favors precision might more useful for users of this class.

martin-cs:
> Can this result in a different fixed point being reached ?

In the general case, yes.  In this specific analysis, I think not
because it is so simple and does not do any of the obvious "exploration
order matters" things like interprocedural analysis.


> What are the implications for the precision of the analysis ?

I don't think there are any.  I would be interested in being proven
wrong.  It should be possible to easily test for significant
differences by:

1. Patching the "remove from set" to remove a random element.
2. Ask @tautschnig to run the analysis a bunch of times with different
random seeds and see what happens.


> Are there any iteration strategies that can result in more precise 
> results ?

Given a program, there is a best order.  It is not clear (and I think
somewhere I have examples to show) that there is not one order which
works for all programs.


> In any
> acceptable perf tradeoff that favors precision might more useful for
> users of this class.

I think in this specific case it doesn't matter, plus, this is supposed
to be a very light-weight analysis.

If you want to experiment, `ai_historyt` has control over the order of
exploration, via:

https://github.com/diffblue/cbmc/blob/47c1c7e70cb17e3315ebacde99ef5b1f4a69eb91/src/analyses/ai_history.h#L123

Some of the existing implementation are ... to put it bluntly, kinda
crude:

https://github.com/diffblue/cbmc/blob/47c1c7e70cb17e3315ebacde99ef5b1f4a69eb91/src/analyses/ai_history.h#L199

So this could be improved.


If you want to do something more sophisticated then a new `ai_historyt`
or inheriting from `ai_baset` and changing how `fixpoint()` works:

https://github.com/diffblue/cbmc/blob/47c1c7e70cb17e3315ebacde99ef5b1f4a69eb91/src/analyses/ai.cpp#L224

HTH

Cheers,
 - Martin

PR #6532, opened 2021-12-16
Wrap NULL/0 is-a-null-pointer in a utility function

tautschnig:
A pointer-typed constant expression with value "NULL" is always a null
pointer expression. When NULL_is_zero is true in the platform
configuration, a value "0" is also a null pointer.

These possible options were repeatedly, but not consistently, evaluated
ad-hoc. This resulted in a least one missed simplification opportunity
as we noticed on an RMC-generated goto binary (which had 0 instead of
NULL in a pointer-typed constant).

### DIFF
--- a/src/goto-instrument/dump_c.cpp
+++ b/src/goto-instrument/dump_c.cpp
@@ -10,25 +10,23 @@ Author: Daniel Kroening, kroening@kroening.com
 /// Dump Goto-Program as C/C++ Source
 
 #include "dump_c.h"
+#include "dump_c_class.h"
 
 #include <util/byte_operators.h>
 #include <util/c_types.h>
 #include <util/config.h>
 #include <util/expr_initializer.h>
+#include <util/expr_util.h>
 #include <util/find_symbols.h>
 #include <util/get_base_name.h>
 #include <util/invariant.h>
-#include <util/namespace.h>
 #include <util/replace_symbol.h>
-#include <util/std_code.h>
 #include <util/string_utils.h>
 
 #include <ansi-c/expr2c.h>
 #include <cpp/expr2cpp.h>
-
 #include <linking/static_lifetime_init.h>
 
-#include "dump_c_class.h"
 #include "goto_program2code.h"
 
 dump_c_configurationt dump_c_configurationt::default_configuration =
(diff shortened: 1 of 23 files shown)


peterschrammel (approved):

src/util/simplify_expr_int.cpp
```diff
@@ -1572,9 +1572,12 @@ simplify_exprt::resultt<> simplify_exprt::simplify_inequality_rhs_is_constant(
     }
 
     // very special case for pointers
-    if(expr.id()==ID_equal &&
-       expr.op1().is_constant() &&
-       expr.op1().get(ID_value)==ID_NULL)
+    if(expr.id() != ID_equal)
+      return unchanged(expr);
+
+    const constant_exprt &op1_constant = to_constant_expr(expr.op1());
```
NlightNFotis, line 1578: Is it possible that this transfer function throws? Say, if `expr.op1()` is not a `constant_exprt`?

Before this change, a property was checked and there was no transfer function application, so I'm slightly worried that this could result in invariants at runtime in some cases.

src/goto-instrument/dump_c.cpp
```diff
@@ -1380,8 +1379,7 @@ void dump_ct::cleanup_expr(exprt &expr)
             // add a typecast for NULL or 0
             if(
               argument.id() == ID_constant &&
```
NlightNFotis, line 1383: Could this also be refactored to `argument.is_constant()`? Or is this not possible?

NlightNFotis (approved):
Solid, minor questions/comments from my end.

src/util/simplify_expr_int.cpp
tautschnig: This previous check actually was unnecessary: the function checks for this earlier on already.

src/goto-instrument/dump_c.cpp
tautschnig: Yes, which should likely be cleaned up across the code base. May I defer this to a PR of its own?

NlightNFotis: Yes, no problem.

tautschnig: Now done in https://github.com/diffblue/cbmc/pull/6542.

chris-ryder (approved):

PR #6533, opened 2021-12-16
Function contracts: check loop freeness before instrumentation, skip assignments to locals and prune write set using CFG info.

remi-delmas-3000:
The first two commits are from this draft PR https://github.com/diffblue/cbmc/pull/6525 which adds a `maybe_alive` bit to  `local_bitvector_analysist` and is needed by this PR.

This PR adds:

In `check_frame_conditions_function`:
- a pass that simplifies away GOTO instructions for which the guard is always false
- a check that the program is loop-free before attempting instrumentation
- manual loop unrolling in regression tests that still contained loops and are got rejected because of this new check

For assigns clause checking: 
- Assigning to local variables or function parameter variables is always legal so we skip instrumenting these assignments, 
- We avoid adding function parameters and local variables to the local write set, except when their address is taken at some point in the program and can later be assigned to indirectly via pointers.
- When generating inclusion checks for assignments, we remove from the local write set targets which are not possibly alive at the ASSIGN instruction that gets instrumented.
- Tests have been simplified by removing function parameters from assigns clauses, and removing function parameters and local variables from test specifications.

The net effect of these changes is a better proof performance because of a reduction in the number of generated assertions and of their size, but otherwise the contract checking functionality remains unchanged.

### DIFF
--- a/regression/contracts/assigns_enforce_04/test.desc
+++ b/regression/contracts/assigns_enforce_04/test.desc
@@ -3,12 +3,6 @@ main.c
 --enforce-contract f1
 ^EXIT=0$
 ^SIGNAL=0$
-^\[f1.\d+\] line \d+ Check that x2 is assignable: SUCCESS$
-^\[f1.\d+\] line \d+ Check that y2 is assignable: SUCCESS$
-^\[f1.\d+\] line \d+ Check that z2 is assignable: SUCCESS$
-^\[f2.\d+\] line \d+ Check that x3 is assignable: SUCCESS$
-^\[f2.\d+\] line \d+ Check that y3 is assignable: SUCCESS$
-^\[f2.\d+\] line \d+ Check that z3 is assignable: SUCCESS$
 ^\[f3.\d+\] line \d+ Check that \*x3 is assignable: SUCCESS$
 ^\[f3.\d+\] line \d+ Check that \*y3 is assignable: SUCCESS$
 ^\[f3.\d+\] line \d+ Check that \*z3 is assignable: SUCCESS$
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -159,6 +159,15 @@ assigns_clauset::conditional_address_ranget::generate_unsafe_inclusion_check(
                               pointer_offset(upper_bound_address_var)}});
 }
 
+bool assigns_clauset::conditional_address_ranget::maybe_alive(
+  cfg_infot &cfg_info) const
+{
+  if(can_cast_expr<symbol_exprt>(source_expr))
+    return cfg_info.is_maybe_alive(to_symbol_expr(source_expr));
+
+  return true;
+}
+
 assigns_clauset::assigns_clauset(
   const exprt::operandst &assigns,
   const messaget &log,
@@ -192,14 +201,26 @@ void assigns_clauset::remove_from_write_set(const exprt &target_expr)
 }
 
 exprt assigns_clauset::generate_inclusion_check(
-  const conditional_address_ranget &lhs) const
+  const conditional_address_ranget &lhs,
+  optionalt<cfg_infot> &cfg_info_opt) const
 {
   if(write_set.empty())
     return not_exprt{lhs.validity_condition_var};
 
   exprt::operandst conditions{not_exprt{lhs.validity_condition_var}};
-  for(const auto &target : write_set)
-    conditions.push_back(target.generate_unsafe_inclusion_check(lhs));
+
+  if(cfg_info_opt.has_value())
+  {
+    auto &cfg_info = cfg_info_opt.value();
+    for(const auto &target : write_set)
+      if(target.maybe_alive(cfg_info))
+        conditions.push_back(target.generate_unsafe_inclusion_check(lhs));
+  }
+  else
+  {
+    for(const auto &target : write_set)
+      conditions.push_back(target.generate_unsafe_inclusion_check(lhs));
+  }
 
   return disjunction(conditions);
 }
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -16,6 +16,7 @@ Date: February 2016
 #include <algorithm>
 #include <map>
 
+#include <analyses/local_bitvector_analysis.h>
 #include <analyses/local_may_alias.h>
 
 #include <ansi-c/c_expr.h>
@@ -307,9 +308,19 @@ void code_contractst::check_apply_loop_contracts(
       goto_function.body, loop_head, snapshot_instructions);
   };
 
+  optionalt<cfg_infot> cfg_empty_info;
+
   // Perform write set instrumentation on the entire loop.
   check_frame_conditions(
-    function_name, goto_function.body, loop_head, loop_end, loop_assigns);
+    function_name,
+    goto_function.body,
+    loop_head,
+    loop_end,
+    loop_assigns,
+    // do not skip checks on function parameter assignments
+    skipt::DontSkip,
+    // do not use CFG info for now
+    cfg_empty_info);
 
   havoc_assigns_targetst havoc_gen(to_havoc, ns);
   havoc_gen.append_full_havoc_code(
@@ -908,23 +919,25 @@ goto_functionst &code_contractst::get_goto_functions()
 }
 
 void code_contractst::instrument_assign_statement(
-  goto_programt::instructionst::iterator &instruction_it,
+  goto_programt::targett &instruction_it,
   goto_programt &program,
-  assigns_clauset &assigns_clause)
+  assigns_clauset &assigns_clause,
+  optionalt<cfg_infot> &cfg_info_opt)
 {
-  INVARIANT(
-    instruction_it->is_assign(),
-    "The first instruction of instrument_assign_statement should always be"
-    " an assignment");
   add_inclusion_check(
-    program, assigns_clause, instruction_it, instruction_it->assign_lhs());
+    program,
+    assigns_clause,
+    instruction_it,
+    instruction_it->assign_lhs(),
+    cfg_info_opt);
 }
 
 void code_contractst::instrument_call_statement(
-  goto_programt::instructionst::iterator &instruction_it,
+  goto_programt::targett &instruction_it,
   const irep_idt &function,
   goto_programt &body,
-  assigns_clauset &assigns)
+  assigns_clauset &assigns,
+  optionalt<cfg_infot> &cfg_info_opt)
 {
   INVARIANT(
     instruction_it->is_function_call(),
@@ -962,7 +975,8 @@ void code_contractst::instrument_call_statement(
       body,
       assigns,
       instruction_it,
-      pointer_object(instruction_it->call_arguments().front()));
+      pointer_object(instruction_it->call_arguments().front()),
+      cfg_info_opt);
 
     // skip all invalidation business if we're freeing invalid memory
     goto_programt alias_checking_instructions, skip_program;
@@ -1122,10 +1136,8 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     return true;
   }
 
-  if(check_for_looped_mallocs(function_obj->second.body))
-  {
-    return true;
-  }
+  const auto &goto_function = function_obj->second;
+  auto &function_body = function_obj->second.body;
 
   // Get assigns clause for function
   const symbolt &target = ns.lookup(function);
@@ -1136,20 +1148,8 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     function,
     symbol_table);
 
-  // detect and add static local variables
+  // Detect and add static local variables
   assigns.add_static_locals_to_write_set(goto_functions, function);
-
-  // Add formal parameters to write set
-  for(const auto &param : to_code_type(target.type).parameters())
-    assigns.add_to_write_set(ns.lookup(param.get_identifier()).symbol_expr());
-
-  auto instruction_it = function_obj->second.body.instructions.begin();
-  for(const auto &car : assigns.get_write_set())
-  {
-    auto snapshot_instructions = car.generate_snapshot_instructions();
-    insert_before_swap_and_advance(
-      function_obj->second.body, instruction_it, snapshot_instructions);
-  };
 
   // Full inlining of the function body
   // Inlining is performed so that function calls to a same function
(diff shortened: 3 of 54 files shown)


regression/contracts/assigns_enforce_04/test.desc
```diff
@@ -3,12 +3,6 @@ main.c
 --enforce-contract f1
 ^EXIT=0$
 ^SIGNAL=0$
-^\[f1.\d+\] line \d+ Check that x2 is assignable: SUCCESS$
```
tautschnig, line 6: Random place for comment: I'd love for the PR description ("This PR adds:

In check_frame_conditions_function: ...) to also be in the body of the commit message. This will ensure that using `git log` is truly meaningful.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -20,6 +20,7 @@ Date: July 2021
 
 #include <util/arith_tools.h>
 #include <util/c_types.h>
+#include <util/format_expr.h>
```
tautschnig: Debugging leftover?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1163,145 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+
+  // Restore internal coherence in the programs
+  goto_functions.update();
+
+  INVARIANT(
+    is_loop_free(function_body, ns, log),
+    "Loops remain in function '" + id2string(function) +
+      "', assigns clause checking instrumentation cannot be applied.");
+
+  // Create a deep copy of the iriginal goto_function object
```
tautschnig: s/iriginal/original/

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1163,145 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+  const goto_programt::targett &target,
+  const optionalt<cfg_infot> cfg_info_opt)
+{
+  if(cfg_info_opt.has_value())
+  {
+    const auto &lhs = target->assign_lhs();
+
+    if(can_cast_expr<symbol_exprt>(lhs))
+      return !cfg_info_opt.value().is_local(
+        to_symbol_expr(lhs).get_identifier());
```
tautschnig: With `can_cast_expr` you can use
```
if(auto lhs_symbol = can_cast_expr<symbol_exprt>(lhs))
  return !cfg_info_opt->is_local(lhs_symbol->get_identifier());
```

src/goto-instrument/contracts/utils.h
```diff
@@ -112,17 +118,102 @@ void insert_before_swap_and_advance(
...
+class cfg_infot
+{
+public:
+  cfg_infot(const namespacet &_ns, goto_functiont &_goto_function)
+    : ns(_ns),
+      goto_function(_goto_function),
+      target(goto_function.body.instructions.begin()),
+      bitvector_analysis(goto_function, ns),
+      dirty_analysis(goto_function),
+      dirty_idents(dirty_analysis.get_dirty_ids()),
```
tautschnig: Why do you copy these instead of using `dirty_analysis` when needed?

tautschnig (commented):
How does this relate to #6527?

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -192,14 +202,26 @@ void assigns_clauset::remove_from_write_set(const exprt &target_expr)
...
 
   exprt::operandst conditions{not_exprt{lhs.validity_condition_var}};
-  for(const auto &target : write_set)
-    conditions.push_back(target.generate_unsafe_inclusion_check(lhs));
+
+  if(cfg_info_opt.has_value())
+  {
+    auto &cfg_info = cfg_info_opt.value();
+    for(const auto &target : write_set)
+      if(target.maybe_alive(cfg_info))
```
feliperodri, line 216:
```suggestion
    for(const auto &target : write_set)
      if(target.maybe_alive(cfg_info_opt.value()))
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1122,10 +1136,9 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
     return true;
   }
 
-  if(check_for_looped_mallocs(function_obj->second.body))
-  {
-    return true;
-  }
+  // unpack
```
feliperodri: Unpack what? I think we can remove this comment.

src/goto-instrument/contracts/contracts.cpp
feliperodri: `s/iriginal/original/g`

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1163,145 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
-    function_obj->second.body,
+    function,
+    function_body,
     instruction_it,
-    function_obj->second.body.instructions.end(),
-    assigns);
+    function_body.instructions.end(),
+    assigns,
+    // skip checks on function parameter assignments
+    true,
```
feliperodri: This should really be an enum.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1163,145 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+
+/// Returns true iff a `DECL x` must be added to the local write set.
+///
+/// A variable is called 'dirty' if its address gets taken at some point in
+/// the program.
+///
+/// Assuming the goto program is obtained from a structured C program that
+/// passed C compiler checks, non-dirty varialbes can only be assigned to
+/// directly by name and by name cannot escape their lexical scope,
+/// and are always safe to assign to. Hence, we only track dirty variables
```
```suggestion
/// passed C compiler checks, non-dirty variables can only be assigned to
/// directly by name, they cannot escape their lexical scope, and they are
/// always safe to assign to. Hence, we only track dirty variables
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1163,145 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+  if(cfg_info_opt.has_value())
+  {
+    const auto &lhs = target->assign_lhs();
+
+    if(can_cast_expr<symbol_exprt>(lhs))
+      return !cfg_info_opt.value().is_local(
+        to_symbol_expr(lhs).get_identifier());
+  }
+
+  // unless proved local by the CFG analysis we assume it is not local.
```
```suggestion
  // Unless proved local by the CFG analysis we assume it is not local.
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1163,145 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+bool must_track_decl(
+  const goto_programt::targett &target,
+  const optionalt<cfg_infot> &cfg_info_opt)
+{
+  // Unless proved non-dirty by the CFG analysis we assume it is dirty.
+  if(!cfg_info_opt.has_value())
+    return true;
+
+  return cfg_info_opt.value().is_not_local_or_dirty_local(
+    target->decl_symbol().get_identifier());
```
```suggestion
  // Unless proved non-dirty by the CFG analysis we assume it is dirty.
  if(cfg_info_opt.has_value())
    return cfg_info_opt.value().is_not_local_or_dirty_local(
      target->decl_symbol().get_identifier());
  return true;
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1163,145 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+}
+
+/// Returns true iff a `DEAD x` must be processed to upate the local write set.
+/// The conditions are the same than for tracking a `DECL x` instruction.
```
feliperodri, line 1266:
```suggestion
/// Returns true iff a `DEAD x` must be processed to update the local write set.
/// The conditions are the same ones for tracking a `DECL x` instruction.
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1163,145 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+bool must_track_dead(
+  const goto_programt::targett &target,
+  const optionalt<cfg_infot> &cfg_info_opt)
+{
+  // Unless proved non-dirty by the CFG analysis we assume it is dirty.
+  if(!cfg_info_opt.has_value())
+    return true;
+
+  return cfg_info_opt.value().is_not_local_or_dirty_local(
+    target->dead_symbol().get_identifier());
```
feliperodri, line 1276:
```suggestion
  // Unless proved non-dirty by the CFG analysis we assume it is dirty.
  if(cfg_info_opt.has_value())
  return cfg_info_opt.value().is_not_local_or_dirty_local(
    target->dead_symbol().get_identifier());
  return true;
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1163,145 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+  assigns_clauset &assigns,
+  bool skip_parameter_assigns,
+  optionalt<cfg_infot> &cfg_info_opt)
 {
-  for(; instruction_it != instruction_end; ++instruction_it)
+  while(instruction_it != instruction_end)
   {
-    const auto &pragmas = instruction_it->source_location().get_pragmas();
-    if(pragmas.find(CONTRACT_PRAGMA_DISABLE_ASSIGNS_CHECK) != pragmas.end())
+    // skip instructions marked as disabled for assigns clause checking
```
```suggestion
    // Skip instructions marked as disabled for assigns clause checking.
```

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1257,19 +1370,26 @@ void code_contractst::check_frame_conditions(
     {
       const auto havoc_argument =
         pointer_object(instruction_it->get_other().operands().front());
-      add_inclusion_check(body, assigns, instruction_it, havoc_argument);
+      add_inclusion_check(
+        body, assigns, instruction_it, havoc_argument, cfg_info_opt);
     }
+
+    // move to the next instruction
```
```suggestion
    // Move to the next instruction.
```

src/goto-instrument/contracts/contracts.h
```diff
@@ -127,20 +128,35 @@ class code_contractst
...
+  /// \param instruction_it iterator to the instruction from which to start
+  /// instrumentation (inclusive)
+  /// \param instruction_end iterator to the instruction at which to stop
+  /// instrumentation (exclusive)
+  /// \param assigns assigns clause of the function (its write set attribute
+  /// gets extended during the instrumentation)
+  /// \param skip_parameter_assigns if true will cause assignments to symbol
+  /// marked as is_parameter to not be instrumented
+  /// \param cfg_info_opt cfg analysis that can will be used for write set
+  /// optimisation if available
```
feliperodri:
```suggestion
  /// \brief Insert assertion statements into the goto program to ensure that
  /// assigned memory is within the assignable memory frame.
  ///
  /// \param function Name of the function getting instrumented.
  /// \param body Body of the function getting instrumented.
  /// \param instruction_it Iterator to the instruction from which to start
  /// instrumentation (inclusive).
  /// \param instruction_end Iterator to the instruction at which to stop
  /// instrumentation (exclusive).
  /// \param assigns Assigns clause of the function (its write set attribute
  /// gets extended during the instrumentation).
  /// \param skip_parameter_assigns If true will cause assignments to symbol
  /// marked as is_parameter to not be instrumented.
  /// \param cfg_info_opt CFG analysis that can will be used for write set
  /// optimisation if available.
```

src/goto-instrument/contracts/utils.cpp
```diff
@@ -163,3 +167,67 @@ void disable_pointer_checks(source_locationt &source_location)
...
+  {
+    if(instruction.is_goto())
+    {
+      const auto &condition = instruction.get_condition();
+      const auto &simplified = simplify_expr(condition, ns);
+      if(simplified.is_false())
+        instruction.turn_into_skip();
+    }
+  }
+}
```
feliperodri: This PR also depends on https://github.com/diffblue/cbmc/pull/6527, perhaps it doesn't have to?

feliperodri (changes requested):

src/goto-instrument/contracts/assigns.cpp

src/goto-instrument/contracts/contracts.cpp

src/goto-instrument/contracts/utils.h
remi-delmas-3000: Removed

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: lhs_symbol would be a boolean in that case, not the result of the cast ?

src/goto-instrument/contracts/contracts.cpp

src/goto-instrument/contracts/contracts.cpp

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: An enum with TRUE and FALSE values ?

src/goto-instrument/contracts/assigns.cpp
remi-delmas-3000: this would reintroduce n function calls to `optional::value()`

src/goto-instrument/contracts/contracts.h
remi-delmas-3000: Fixed

src/goto-instrument/contracts/utils.cpp
remi-delmas-3000: This one supersedes [the](https://github.com/diffblue/cbmc/pull/6527), which has been closed.

remi-delmas-3000:
The dependency on  `local_bitvector_analysist` and https://github.com/diffblue/cbmc/pull/6533 was removed

src/goto-instrument/contracts/utils.cpp
feliperodri:
```suggestion
void simplify_gotos(goto_programt &goto_program, namespacet &ns)
{
  for(auto &instruction : goto_program.instructions)
    if(instruction.is_goto() && simplify_expr(instruction.get_condition(), ns).is_false())
        instruction.turn_into_skip();
}
```

feliperodri (approved):
Only minor comments left.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -159,6 +159,15 @@ assigns_clauset::conditional_address_ranget::generate_unsafe_inclusion_check(
                               pointer_offset(upper_bound_address_var)}});
 }
 
+bool assigns_clauset::conditional_address_ranget::maybe_alive(
+  cfg_infot &cfg_info) const
+{
+  if(can_cast_expr<symbol_exprt>(source_expr))
```
tautschnig, line 165: As discussed earlier today: could be implemented using `if(auto sym_expr = expr_try_dynamic_cast<symbol_exprt>(source_expr))`

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -192,14 +201,26 @@ void assigns_clauset::remove_from_write_set(const exprt &target_expr)
...
   exprt::operandst conditions{not_exprt{lhs.validity_condition_var}};
-  for(const auto &target : write_set)
-    conditions.push_back(target.generate_unsafe_inclusion_check(lhs));
+
+  if(cfg_info_opt.has_value())
+  {
+    auto &cfg_info = cfg_info_opt.value();
+    for(const auto &target : write_set)
+      if(target.maybe_alive(cfg_info))
+        conditions.push_back(target.generate_unsafe_inclusion_check(lhs));
```
tautschnig, line 217: Will this actually introduce the same check in multiple places? Wouldn't it be sufficient to check this once?

src/goto-instrument/contracts/contracts.cpp
tautschnig: A 2-value enum, where the "true" and "false" values carry meaningful names.

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1162,38 +1162,143 @@ bool code_contractst::check_frame_conditions_function(const irep_idt &function)
...
+    // skip checks on function parameter assignments
+    true,
+    cfg_info_opt);
 
   return false;
 }
 
+/// Returns true iff the target instruction is tagged with a
+/// 'CONTRACT_PRAGMA_DISABLE_ASSIGNS_CHECK' pragma.
+bool has_disable_assigns_check_pragma(const goto_programt::targett &target)
```
tautschnig: Do you really want an iterator that cannot change (`const goto_programt::targett&`) or an iterator so that whatever it points to cannot change (`goto_programt::const_targett`)?

src/goto-instrument/contracts/utils.cpp
```diff
@@ -163,3 +167,67 @@ void disable_pointer_checks(source_locationt &source_location)
...
+
+bool is_loop_free(
+  const goto_programt &goto_program,
+  namespacet &ns,
+  messaget &log)
+{
+  // create cfg from instruction list
+  cfg_baset<empty_cfg_nodet> cfg;
+  cfg(goto_program);
```
tautschnig, line 189: Comment from the earlier PR: do you really need such an elaborate analysis, or would looking at `is_backwards_goto()` for all instructions in `goto_program` also do the trick?

src/goto-instrument/contracts/contracts.cpp

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: Fixed. 
Following that logic we should change all `bool` return types to `SUCCESS/FAILURE` for 
- `replace_calls`, 
-`enforce_contracts`,  
- `enforce_contract`,
- `check_frame_conditions_function`,
- `apply_function_contract`,

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/contracts.cpp

src/goto-instrument/contracts/assigns.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/assigns.cpp
remi-delmas-3000: it generates an inclusion check for each possible target of the write set, and each check is specify to the lhs of the instrumented assignment, so no doublons are created here.

src/goto-instrument/contracts/utils.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: Both! Thanks for the tip

src/goto-instrument/contracts/utils.cpp
remi-delmas-3000: I ignored the existence of `is_backwards_goto`, if it is just jumping to an earlier instruction in the sequence it would not necessarily introduce a loop, would it ?

tautschnig: Hmm, that indeed does not imply the existence of a loop. But then, would `natural_loopst` do the trick here and relieve you of some of the heavy lifting?

PR #6535, opened 2021-12-17
Add missing DECL and DEAD instructions in function pointer call site labelling

remi-delmas-3000:
No change in functionality or performance is expected. 

Previously, function pointer call site labelling introduced ASSIGNS and CALL instructions to a fresh function pointer variable in the goto program without adding corresponding DECL and DEAD instructions, which made them look like global variables when they really are local to the function.

This caused  problems with function contracts checking, because a function would appear to be writing to a global that is not mentioned in its assigns clause.

We also add a small diagram to document what it means for `goto_programt::insert_before_swap` to preserve jumps to target.

- [] Regression or unit tests are included, or existing tests cover the modified code (in this case I have detailed which ones those are in the commit message).

### DIFF
--- a/src/goto-programs/goto_program.h
+++ b/src/goto-programs/goto_program.h
@@ -625,6 +625,22 @@ public:
 
   /// Insertion that preserves jumps to "target".
   /// The instruction is destroyed.
+  ///
+  /// Turns:
+  /// ```
+  /// ...->[a]->...
+  ///       ^
+  ///     target
+  /// ```
+  ///
+  /// Into:
+  /// ```
+  /// ...->[i]->[a]->...
+  ///       ^
+  ///     target
+  /// ```
+  ///
+  /// So that jumps to `a` now jump to the newly inserted `i`.
   void insert_before_swap(targett target, instructiont &instruction)
   {
     insert_before_swap(target);
--- a/src/goto-programs/label_function_pointer_call_sites.cpp
+++ b/src/goto-programs/label_function_pointer_call_sites.cpp
@@ -52,24 +52,32 @@ void label_function_pointer_call_sites(goto_modelt &goto_model)
           goto_model.symbol_table.lookup_ref(call_site_symbol_name)
             .symbol_expr();
 
-        // add assignment to the new function pointer variable, followed by a
-        // call of the new variable
+        // add a DECL instruction for the function pointer variable
+        auto decl_instruction =
+          goto_programt::make_decl(new_function_pointer, source_location);
+
+        goto_function.second.body.insert_before_swap(it, decl_instruction);
+        ++it;
+
+        // add assignment to the new variable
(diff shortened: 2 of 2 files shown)


src/goto-programs/goto_program.h
```diff
@@ -625,6 +625,22 @@ class goto_programt
...
+  /// ```
+  ///
+  /// Into:
+  /// ```
+  /// ...->[i]->[a]->...
+  ///       ^
+  ///     target
+  /// ```
+  ///
+  /// So that jumps to `a` now jump to the newly inserted `i`.
```
tautschnig, line 643: Thank you! But then: would you mind putting this in a commit of its own?

src/goto-programs/label_function_pointer_call_sites.cpp
```diff
@@ -52,17 +52,31 @@ void label_function_pointer_call_sites(goto_modelt &goto_model)
           goto_model.symbol_table.lookup_ref(call_site_symbol_name)
             .symbol_expr();
 
-        // add assignment to the new function pointer variable, followed by a
-        // call of the new variable
+        // add a DECL instruction for the function pointer variable
+        auto decl_instruction =
+          goto_programt::make_decl(new_function_pointer, source_location);
+        goto_function.second.body.insert_before_swap(it, decl_instruction);
+        it++;
```
tautschnig: Nit picking: `++it` (which the compiler will likely take care of for you, but why rely on compiler optimisations...).

src/goto-programs/label_function_pointer_call_sites.cpp
```diff
@@ -52,17 +52,31 @@ void label_function_pointer_call_sites(goto_modelt &goto_model)
...
                        function_pointer_dereference.pointer()},
           source_location);
 
         goto_function.second.body.insert_before_swap(it, assign_instruction);
-        const auto next = std::next(it);
-        to_code_function_call(next->code_nonconst()).function() =
+        it++;
+
+        // transform original call into a call to the new variable
+        to_code_function_call(it->code_nonconst()).function() =
```
tautschnig: `it->call_function() =` would be a shorter version thereof.

tautschnig (approved):
Since you touched this code, would you mind adding the following cleanup:
1) The lambda inserting the symbol into the symbol table should also set `is_thread_local` and `is_file_local` to `true`.
2) Instead of the repeated and long-winded `goto_model.symbol_table.lookup_ref` could you please introduce a `namespacet ns{goto_model.symbol_table}` early on and replace the `goto_model.symbol_table.lookup_ref` by `ns.lookup`?

Also, would it be possible to include a test, for example by adding patterns to `regression/goto-instrument/restrict-function-pointer-by-name-local/test.desc` that check for the existence of the new DECL/DEAD instructions?

feliperodri (approved):

PR #6539, opened 2021-12-20
Switch Glucose download source to GitHub

tautschnig:
www.labri.fr appears to be (temporarily?) offline, which already
happened several weeks ago, when it was offline for about one day. To
avoid spurious CI failures, switch to a GitHub source that hosts to
Glucose source. The use of Bruno Dutertre's fork also reduces the number
of patches that need to be applied, mostly leaving just Windows-specific
bits to be patched.

### DIFF
--- a/src/solvers/CMakeLists.txt
+++ b/src/solvers/CMakeLists.txt
@@ -88,10 +88,10 @@ elseif("${sat_impl}" STREQUAL "glucose")
     message(STATUS "Building solvers with glucose")
 
     download_project(PROJ glucose
-        URL http://www.labri.fr/perso/lsimon/downloads/softwares/glucose-syrup.tgz
+        URL https://github.com/BrunoDutertre/glucose-syrup/archive/0bb2afd3b9baace6981cbb8b4a1c7683c44968b7.tar.gz
         PATCH_COMMAND patch -p1 -i ${CBMC_SOURCE_DIR}/../scripts/glucose-syrup-patch
         COMMAND cmake -E copy ${CBMC_SOURCE_DIR}/../scripts/glucose_CMakeLists.txt CMakeLists.txt
-        URL_MD5 b6f040a6c28f011f3be994663338f548
+        URL_MD5 7c539c62c248b74210aef7414787323a
     )
 
     add_subdirectory(${glucose_SOURCE_DIR} ${glucose_BINARY_DIR})
--- a/scripts/glucose-syrup-patch
+++ b/scripts/glucose-syrup-patch
@@ -1,24 +1,3 @@
-diff -rupNw glucose-syrup/core/Solver.cc glucose-syrup-patched/core/Solver.cc
---- glucose-syrup/core/Solver.cc	2014-10-03 11:10:21.000000000 +0200
-+++ glucose-syrup-patched/core/Solver.cc	2018-04-21 16:58:22.950005391 +0200
-@@ -931,7 +931,6 @@ void Solver::uncheckedEnqueue(Lit p, CRe
- CRef Solver::propagate() {
-     CRef confl = CRef_Undef;
-     int num_props = 0;
--    int previousqhead = qhead;
-     watches.cleanAll();
-     watchesBin.cleanAll();
-     unaryWatches.cleanAll();
-@@ -1405,7 +1404,9 @@ lbool Solver::search(int nof_conflicts)
-                 decisions++;
-                 next = pickBranchLit();
-                 if (next == lit_Undef) {
-+#if 0
-                     printf("c last restart ## conflicts  :  %d %d \n", conflictC, decisionLevel());
-+#endif
-                     // Model found:
-                     return l_True;
-                 }
 diff -rupNw glucose-syrup/core/SolverTypes.h glucose-syrup-patched/core/SolverTypes.h
 --- glucose-syrup/core/SolverTypes.h	2014-10-03 11:10:22.000000000 +0200
 +++ glucose-syrup-patched/core/SolverTypes.h	2018-04-21 16:58:22.950005391 +0200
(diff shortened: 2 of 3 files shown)


src/solvers/CMakeLists.txt
```diff
@@ -88,10 +88,10 @@ elseif("${sat_impl}" STREQUAL "glucose")
     message(STATUS "Building solvers with glucose")
 
     download_project(PROJ glucose
-        URL http://www.labri.fr/perso/lsimon/downloads/softwares/glucose-syrup.tgz
+        URL https://github.com/BrunoDutertre/glucose-syrup/archive/refs/heads/main.tar.gz
         PATCH_COMMAND patch -p1 -i ${CBMC_SOURCE_DIR}/../scripts/glucose-syrup-patch
         COMMAND cmake -E copy ${CBMC_SOURCE_DIR}/../scripts/glucose_CMakeLists.txt CMakeLists.txt
-        URL_MD5 b6f040a6c28f011f3be994663338f548
+        URL_MD5 89a5183dd9b9eb5296fd9fab2c968a50
```
NlightNFotis: As far as I understand we are downloading the `main` branch, which seems to be something like a `master` branch of sorts.

How often is this being updated, and could random updates cause CI to fail by essence of this hash check failing?

NlightNFotis (commented):
Just one question.

src/solvers/CMakeLists.txt
tautschnig: The repo only has total of 12 commits, and the most recent commit was July 2021. I would expect the rate of updates to be very low, if any at all. So, yes, CI could fail, but when it does, it is right to do so: it would be a change that we should manually review to make sure it doesn't conflict with any of our expectations on this third-party component.

thomasspriggs: I am not entirely convinced that we should be following the main branch on someone else's fork. Depending on the development workflow they are using, we could be fetching untested or even non-compiling commits. Is there a way we can pin this to a release version?

kroening: An option would be to get
https://github.com/BrunoDutertre/glucose-syrup/archive/0bb2afd3b9baace6981cbb8b4a1c7683c44968b7.zip

chrisr-diffblue: Is it worth stepping back a pace or two and consider whether using a submodule or subtree is appropriate? Also other fleeting thoughts... should we contribute our local patches back to Bruno to avoid maintaining them ourselves (and maintaining our local patch management)? Should we fork Bruno's repo, apply our patches there, and download from our fork? No major preferences from my side, but just feels worth a thought and explicit decisions.

kroening: submodules are tedious to deal with when your build environment doesn't have network access.

tautschnig: Given that none of the glucose-syrup repositories have done a release, the only option seems to be what @kroening suggested. That would ensure stability, at the expense of not picking up future changes should they ever occur. Given that we are talking about a piece of software that is unlikely to see further development, maybe that's an ok price to pay?

thomasspriggs: - Fetching the archive of a specific commit seems like a better option than fetching the main branch to me.
- Downloading dependencies as part of the run of `make` / `cmake` vs checking out a submodule seems like moving the problem of needing network access around, rather than solving it to me. Therefore I am not going to argue in favour of one or the other.
- My understanding is that setting up a _subtree_ would mean that the dependency would be included in downloaded archives and initial checkouts of the repository. I can look into this possibility as an alternative to confirm if this is the case.

thomasspriggs: I can confirm that if we were to setup glucose as a subtree, then it would be checked out by default and it would be included in archive downloads from github - https://github.com/thomasspriggs/cbmc/tree/tas/glucose-subtree

peterschrammel: My preference would be to fork the repo and manage our patches there. I don't think there's anything to upstream to the original repo in our patches, though.
I think the discussion around not having network access in the build environment is not relevant - you already need network access to download minisat in the current setup. Unless this is a new requirement that we must satisfy I don't think subtree is a suitable option for that reason - it's an overkill. 
I'm fine with submodules, but I think they are an overkill for our use case as glucose is optional and the interface of glucose is very stable - i.e. no need to tightly track upgrades.
So, my feeling is that fetching the source zip from a fork is probably the must suitable and easiest solution.

tautschnig: Reviving this PR/discussion as we're once again blocked by network access issues to labri.fr. If anyone fancies doing (and maintaining) a fork then that's ok with me. I don't have the resources to do so, and will hence have to stick with what I've proposed in this PR at this point.

tautschnig: I have now added a comment that indicates the exact revision and  also states that our CI jobs will fail if "main" ever moves.

tautschnig: As we're in this situation once again (see https://github.com/diffblue/cbmc/pull/6986): pinging @NlightNFotis @thomasspriggs @peterschrammel for an update.

thomasspriggs: I am against merging this as-is, because I think that as a minimum we should be pinning the download to a fixed commit hash as suggested by @kroening rather than just taking the latest version of the main branch. @tautschnig Are you happy to make this update before we approve/merge this or would you like myself and @NlightNFotis to copy your branch and manage the change ourselves?

tautschnig: I've now changed download URL to use the specific commit hash rather than a branch.

TGWDB: Looks like the right solution for now. Thanks.

TGWDB (approved):

thomasspriggs (approved):

peterschrammel (approved):

PR #6540, opened 2021-12-21
Fix havocking of dependent assigns clause targets

remi-delmas-3000:
The first commit fixes a minor issue with history variables (allow history variables for constant and typecast expressions).

The second commit fixes missing declarations for return value variables introduced during replacement of function calls by contracts.

The third commit fixes a problem with assigns clause havocking when replacing calls by contracts.

When replacing a function call by its contract, we would directly havoc the targets of the assigns clause in an unspecified order. If the assigns clause contained dependent targets such as `assigns(s->ptr, *(s->ptr))`, havocking `s->ptr` before `*s->ptr` would lead to errors and incorrect results. 

We now introduce independent snapshot variables storing the address of each target and havoc them in a second time to avoid this dependency on the ordering of havoc instructions.

### DIFF
--- a/regression/contracts/assigns_replace_08/test.desc
+++ b/regression/contracts/assigns_replace_08/test.desc
@@ -3,7 +3,7 @@ main.c
 --enforce-contract foo --replace-call-with-contract bar _ --pointer-primitive-check
 ^EXIT=10$
 ^SIGNAL=0$
-\[foo.\d+\] line \d+ Check that \*z is assignable: FAILURE$
+\[foo.\d+\] line \d+ Check that \*z \(assigned by the contract of bar\) is assignable: FAILURE$
 ^.* 1 of \d+ failed \(\d+ iteration.*\)$
 ^VERIFICATION FAILED$
 --
--- a/regression/contracts/assigns_replace_havoc_dependent_targets_fail/enforce.desc
+++ b/regression/contracts/assigns_replace_havoc_dependent_targets_fail/enforce.desc
@@ -1,0 +1,9 @@
+CORE
+main_enforce.c
+--enforce-contract resize_vec _ --signed-overflow-check --unsigned-overflow-check --pointer-overflow-check
+^VERIFICATION SUCCESSFUL$
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+Verifies the contract being replaced in `replace.desc`.
--- a/regression/contracts/assigns_replace_havoc_dependent_targets_fail/vect.h
+++ b/regression/contracts/assigns_replace_havoc_dependent_targets_fail/vect.h
@@ -1,0 +1,48 @@
+#include <assert.h>
+#include <stdlib.h>
+
+typedef struct vect
+{
+  char *arr;
+  size_t size;
+} vect;
+
+void resize_vec(vect *v, size_t incr)
+  // clang-format off
+__CPROVER_requires(
+  __CPROVER_is_fresh(v, sizeof(vect)) &&
+  0 < v->size && v->size <= __CPROVER_max_malloc_size &&
+  0 < incr && incr < __CPROVER_max_malloc_size - v->size &&
+  __CPROVER_is_fresh(v->arr, v->size)
+)
+__CPROVER_assigns(v->size, v->arr, __CPROVER_POINTER_OBJECT(v->arr))
+__CPROVER_ensures(
+  v->size == __CPROVER_old(v->size) + __CPROVER_old(incr) &&
+  __CPROVER_is_fresh(v->arr, v->size)
+)
+// clang-format on
+{
+  free(v->arr);
+  v->size += incr;
+  v->arr = malloc(v->size);
+  return;
+}
+
+void resize_vec_incr10(vect *v)
+  // clang-format off
+__CPROVER_requires(
+  __CPROVER_is_fresh(v, sizeof(vect)) &&
+  0 < v->size && v->size <= __CPROVER_max_malloc_size &&
+  v->size + 10 < __CPROVER_max_malloc_size &&
+  __CPROVER_is_fresh(v->arr, v->size)
+)
+__CPROVER_assigns(v->size)
+__CPROVER_ensures(
+  v->size == __CPROVER_old(v->size) + 10 &&
+  __CPROVER_is_fresh(v->arr, v->size)
+)
+// clang-format on
+{
+  resize_vec(v, 10);
+  return;
+}
--- a/regression/contracts/history-constant/main.c
+++ b/regression/contracts/history-constant/main.c
@@ -1,0 +1,15 @@
+#include <stdio.h>
+
+int foo(int l) __CPROVER_requires(-10 <= l && l <= 10) __CPROVER_ensures(
+  __CPROVER_return_value == __CPROVER_old(l) + __CPROVER_old(10))
+{
+  return l + 10;
+}
+
+int main()
+{
+  int l;
+  __CPROVER_assume(-10 <= l && l <= 10);
+  foo(l);
+  return 0;
+}
--- a/regression/contracts/history-typecast/main.c
+++ b/regression/contracts/history-typecast/main.c
@@ -1,0 +1,23 @@
+#include <stdio.h>
+
+long bar(long l, long r) __CPROVER_requires(-10 <= l && l <= 10)
+  __CPROVER_requires(-10 <= r && r <= 10) __CPROVER_ensures(
+    __CPROVER_return_value == __CPROVER_old(l) + __CPROVER_old(r))
+{
+  return l + r;
+}
+
+int foo(int l, int r) __CPROVER_requires(-10 <= l && l <= 10)
+  __CPROVER_requires(-10 <= r && r <= 10) __CPROVER_ensures(
+    __CPROVER_return_value == __CPROVER_old(l) + __CPROVER_old(r))
+{
+  return bar((long)l, (long)r);
+}
+
+int main()
+{
+  int n;
+  __CPROVER_assume(-10 <= n && n <= 10);
+  foo(n, n);
+  return 0;
+}
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -43,6 +43,7 @@ Date: February 2016
 #include <util/replace_symbol.h>
 #include <util/std_code.h>
 
+#include "havoc_assigns_clause_targets.h"
 #include "memory_predicates.h"
 #include "utils.h"
 
@@ -790,15 +791,17 @@ bool code_contractst::apply_function_contract(
   // ...for assigns clause targets
   if(!assigns_clause.empty())
   {
-    assigns_clauset assigns_clause(
-      targets.operands(), log, ns, target_function, symbol_table);
-
-    // Havoc all targets in the write set
-    assignst assigns;
-    assigns.insert(targets.operands().cbegin(), targets.operands().cend());
-
-    havoc_assigns_targetst havoc_gen(assigns, ns);
-    havoc_gen.append_full_havoc_code(location, havoc_instructions);
+    // Havoc all targets in the assigns clause
+    // TODO: handle local statics possibly touched by this function
+    havoc_assigns_clause_targets(
+      target_function,
+      targets.operands(),
+      havoc_instructions,
+      // context parameters
+      location,
+      mode,
+      ns,
+      symbol_table);
   }
 
   // ...for the return value
@@ -840,6 +843,7 @@ bool code_contractst::apply_function_contract(
     ++target;
   }
 
+  // Erase original function call
   *target = goto_programt::make_skip();
 
   // Add this function to the set of replaced functions.
@@ -1422,8 +1426,20 @@ code_contractst::add_inclusion_check(
   source_locationt location_no_checks =
     instruction_it->source_location_nonconst();
   disable_pointer_checks(location_no_checks);
-  location_no_checks.set_comment(
-    "Check that " + from_expr(ns, lhs.id(), lhs) + " is assignable");
+
+  // does this assignment come from some contract replacement ?
+  const auto &comment = location_no_checks.get_comment();
+  if(is_assigns_clause_replacement_tracking_comment(comment))
+  {
+    location_no_checks.set_comment(
+      "Check that " + id2string(comment) + " is assignable");
+  }
+  else
+  {
+    location_no_checks.set_comment(
+      "Check that " + from_expr(ns, lhs.id(), lhs) + " is assignable");
+  }
+
   assertion.add(goto_programt::make_assertion(
     assigns.generate_inclusion_check(car, cfg_info_opt), location_no_checks));
   insert_before_swap_and_advance(program, instruction_it, assertion);
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -562,9 +562,10 @@ void code_contractst::replace_history_parameter(
   {
     const auto &parameter = to_history_expr(expr, id).expression();
 
+    const auto &id = parameter.id();
     if(
-      parameter.id() == ID_dereference || parameter.id() == ID_member ||
-      parameter.id() == ID_symbol || parameter.id() == ID_ptrmember)
+      id == ID_dereference || id == ID_member || id == ID_symbol ||
+      id == ID_ptrmember || id == ID_constant || id == ID_typecast)
     {
       auto it = parameter2history.find(parameter);
 
--- a/src/goto-instrument/contracts/contracts.cpp
+++ b/src/goto-instrument/contracts/contracts.cpp
@@ -672,6 +672,9 @@ bool code_contractst::apply_function_contract(
   // keep track of the call's return expression to make it nondet later
   optionalt<exprt> call_ret_opt = {};
 
+  // if true, the call return variable variable was created during replacement
+  bool call_ret_is_fresh_var = false;
+
   if(type.return_type() != empty_typet())
   {
     // Check whether the function's return value is not disregarded.
@@ -696,6 +699,7 @@ bool code_contractst::apply_function_contract(
       {
         // The postcondition does mention __CPROVER_return_value, so mint a
         // fresh variable to replace __CPROVER_return_value with.
+        call_ret_is_fresh_var = true;
         const symbolt &fresh = get_fresh_aux_symbol(
           type.return_type(),
           id2string(target_function),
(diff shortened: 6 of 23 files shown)


src/goto-instrument/contracts/contracts.cpp
```diff
@@ -562,9 +562,10 @@ void code_contractst::replace_history_parameter(
   {
     const auto &parameter = to_history_expr(expr, id).expression();
 
+    const auto &id = parameter.id();
     if(
-      parameter.id() == ID_dereference || parameter.id() == ID_member ||
-      parameter.id() == ID_symbol || parameter.id() == ID_ptrmember)
+      id == ID_dereference || id == ID_member || id == ID_symbol ||
+      id == ID_ptrmember || id == ID_constant || id == ID_typecast)
     {
```
tautschnig, line 570: Is there a test that demonstrates the need for this?

regression/contracts/assigns_replace_08/test.desc
```diff
@@ -3,7 +3,7 @@ main.c
 --enforce-contract foo --replace-call-with-contract bar _ --pointer-primitive-check
 ^EXIT=10$
 ^SIGNAL=0$
-\[foo.\d+\] line \d+ Check that \*z is assignable: FAILURE$
+\[foo.\d+\] line \d+ Check that \*z assigned by replaced bar is assignable: FAILURE$
```
tautschnig: I'm struggling to parse this - should it perhaps say "replaced *function* bar?"

src/goto-instrument/contracts/havoc_assigns_clause_targets.h
```diff
@@ -0,0 +1,62 @@
...
+/// Havoc function assigns clauses
+
+#ifndef CPROVER_GOTO_INSTRUMENT_CONTRACTS_HAVOC_ASSIGNS_CLAUSE_TARGETS_H
+#define CPROVER_GOTO_INSTRUMENT_CONTRACTS_HAVOC_ASSIGNS_CLAUSE_TARGETS_H
+
+#include <set>
+
+#include <goto-programs/goto_program.h>
+#include <util/namespace.h>
+#include <util/symbol_table.h>
```
tautschnig: Forward declarations (`class namespacet;`, `class symbol_tablet;`) will do the trick here. (This is to reduce compilation times where possible.)

src/goto-instrument/contracts/havoc_assigns_clause_targets.h
```diff
@@ -0,0 +1,62 @@
...
+
+\*******************************************************************/
+
+/// \file
+/// Havoc function assigns clauses
+
+#ifndef CPROVER_GOTO_INSTRUMENT_CONTRACTS_HAVOC_ASSIGNS_CLAUSE_TARGETS_H
+#define CPROVER_GOTO_INSTRUMENT_CONTRACTS_HAVOC_ASSIGNS_CLAUSE_TARGETS_H
+
+#include <set>
```
tautschnig: This isn't used here.

src/goto-instrument/contracts/havoc_assigns_clause_targets.h
```diff
@@ -0,0 +1,62 @@
...
+
+#include <goto-programs/goto_program.h>
```
tautschnig: I think you'll only need `#include <util/expr.h>`, everything else can be forward-declared.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -0,0 +1,239 @@
...
+/// - else trigger an error.
+///
+static exprt base_pointer(
+  const exprt &target,
+  // context parameters
+  const namespacet &ns)
+{
+  if(target.id() == ID_pointer_object)
+  {
+    const auto &ptr = target.operands().front();
```
tautschnig, line 61: Please use `to_unary_expr(target).op()` (for a minimal amount of sanity checking).

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -0,0 +1,239 @@
...
+    // bring the offset back to zero
+    auto res = minus_exprt{ptr, pointer_offset(ptr)};
+    res.type() = ptr.type();
```
tautschnig: Oh, we're really lacking a proper constructor for `minus_exprt` here. Adding this to my todo list.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -0,0 +1,239 @@
...
+    return std::move(res);
+  }
+  else if(is_assignable(target))
+  {
+    // check that the type of the target has a definite size
+    const auto &size = size_of_expr(target.type(), ns);
+
+    INVARIANT(
+      size.has_value(),
```
tautschnig: Please move this into the `INVARIANT` to avoid a situation where `size` ends up as an unused value (when completely disabling invariants). That is: `INVARIANT(size_of_expr(target.type(), ns).has_value(), ...)`

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -0,0 +1,239 @@
...
+      "`sizeof` must always be computable on assignable assigns clause "
+      "targets.");
+
+    return address_of_exprt{target, pointer_type(target.type())};
```
tautschnig: Wouldn't `is_assignable` be true for something like `a[5]`, but taking the address thereof would not actually be the "base address" (or at least not how I'd understand that term).

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -0,0 +1,239 @@
...
+  dest.add(goto_programt::make_decl(target_valid_var));
+
+  dest.add(goto_programt::make_decl(target_snapshot_var));
+
+  dest.add(goto_programt::make_assignment(
+    target_valid_var, all_dereferences_are_valid(target_pointer, ns)));
+
+  dest.add(goto_programt::make_assignment(
+    target_snapshot_var,
+    null_pointer_exprt{to_pointer_type(target_pointer.type())}));
```
tautschnig: For debugging purposes it would be helpful if all these had source locations that relate them in some way to the location a snapshot is taken at. Hmm, you're doing this after the fact. That's a bit confusing, in my opinion.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+  goto_programt skip_program;
+  const auto skip_target = skip_program.add(goto_programt::make_skip());
+
+  dest.add(goto_programt::make_goto(skip_target, not_exprt{target_valid_var}));
+
+  dest.add(goto_programt::make_assignment(target_snapshot_var, target_pointer));
+
+  dest.destructive_append(skip_program);
```
tautschnig: You could use `insert_before_swap` to avoid the `skip_program`. But then it probably doesn't make a whole lot of a difference.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -0,0 +1,239 @@
...
+    skip_program.add(goto_programt::make_skip(source_location));
+
+  dest.add(goto_programt::make_goto(
+    skip_target, not_exprt{target_valid_var}, source_location));
+
+  if(target.id() == ID_pointer_object)
+  {
+    // OTHER __CPROVER_havoc_object(target_snapshot_var)
+    codet code(ID_havoc_object);
+    code.add_to_operands(target_snapshot_var);
```
tautschnig: `codet code(ID_havoc_object, {target_snapshot_var});` would avoid the two-step construction.

tautschnig (approved):

regression/contracts/assigns_replace_08/test.desc
feliperodri: How about `Check that \*z is assignable (function bar replacement): `?

regression/contracts/assigns_replace_havoc_dependent_targets/vect.h
```diff
@@ -0,0 +1,48 @@
...
+{
+  char *arr;
+  size_t size;
+} vect;
+
+void resize_vec(vect *v, size_t incr)
+  // clang-format off
+__CPROVER_requires(
+  __CPROVER_is_fresh(v, sizeof(vect)) &&
+  0 < v->size && v->size <= __CPROVER_max_malloc_size &&
```
feliperodri: Why do you need to check whether `v->size` is smaller than `__CPROVER_max_malloc_size`? Haven't this problem been solved (see https://github.com/diffblue/cbmc/issues/6130)?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -786,15 +787,17 @@ bool code_contractst::apply_function_contract(
...
-      targets.operands(), log, ns, target_function, symbol_table);
-
-    // Havoc all targets in the write set
-    assignst assigns;
-    assigns.insert(targets.operands().cbegin(), targets.operands().cend());
-
-    havoc_assigns_targetst havoc_gen(assigns, ns);
-    havoc_gen.append_full_havoc_code(location, havoc_instructions);
+    // Havoc all targets in the assigns clause
+    // TODO handle local statics possibly touched by this function
```
feliperodri: Is there a issue to track this?

feliperodri: nit:
```suggestion
    // Havoc all targets in the assigns clause.
    // TODO: handle local statics possibly touched by this function.
```

regression/contracts/assigns_replace_havoc_dependent_targets_fail/enforce.desc
```diff
@@ -0,0 +1,9 @@
+CORE
+main_enforce.c
+--enforce-contract resize_vec _ --signed-overflow-check --unsigned-overflow-check --pointer-overflow-check
+^VERIFICATION SUCCESSFUL$
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+Verifies the contract being replaced in `replace_resize_vec.desc`.
```
feliperodri: Is that the correct file name? I didn't find it.

regression/contracts/assigns_replace_havoc_dependent_targets_fail/vect.h
```diff
@@ -0,0 +1,48 @@
...
+  v->size += incr;
+  v->arr = malloc(v->size);
+  return;
+}
+
+void resize_vec_incr10(vect *v)
+  // clang-format off
+__CPROVER_requires(
+  __CPROVER_is_fresh(v, sizeof(vect)) &&
+  0 < v->size && v->size <= __CPROVER_max_malloc_size &&
```
feliperodri, line 35: See https://github.com/diffblue/cbmc/issues/6130.

Hasn't this problem been fixed? Why do you need `v->size <= __CPROVER_max_malloc_size` here?

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: This computes the start address of the thing that gets assigned. If contract assigns `a[5]` the base address of the thing that gets assigned is `&(a[5])`. If the contract says it assigns `__CPROVER_pointer_object(a[5])` then the base address will be the address pointed to by `a-5`.

src/goto-instrument/contracts/havoc_assigns_clause_targets.h
```diff
@@ -0,0 +1,62 @@
...
+/// \param mode Language mode to use for newly generated symbols
+/// \param ns Namespace of the model
+/// \param st Symbol table of the model (new symbols will be added)
+///
+/// Assigns clause targets can be interdependent as shown in this example:
+///
+/// ```
+/// typedef struct vect { int *arr; int size; } vect;
+/// void resize(vect *v)
+/// __CPOVER_assigns(v->arr, v->size, __CPROVER_POINTER_OBJECT(v->arr))
```
feliperodri: typo: `__CPROVER_assigns`

feliperodri (approved):
Only minor comments.

regression/contracts/assigns_replace_havoc_dependent_targets/vect.h
remi-delmas-3000: Whatever the behaviour of CBMC is when this bound is exceeded, I want to avoid it in this example.

regression/contracts/assigns_replace_havoc_dependent_targets_fail/vect.h
remi-delmas-3000: Whatever the behaviour of CBMC is when this bound is exceeded, I want to avoid it in this example.

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: The tests added in this PR will fail without this.

remi-delmas-3000: added specific tests

regression/contracts/assigns_replace_08/test.desc
remi-delmas-3000: yes

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: it seems less natural, since after inserting the skip statement in dest, you would have to use a mix of insert_before_swap and skip_target++ or to insert instructions in reverse, or to create another program and insert_before_swap and then ++ the skip_target

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: yes

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
tautschnig: Ok!

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
tautschnig: I'd then find something along the lines of "build_address_of" a more natural function name, but this might be a matter of personal preference.

kroening: As a side, note that `address_of_exprt` has a convenience constructor that takes just the object, i.e., the above is equivalent to `address_of_exprt{target}`. The constructor will create the pointer type from the type of `target`.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -0,0 +1,252 @@
...
+  // context parameters
+  const namespacet &ns)
+{
+  if(target.id() == ID_pointer_object)
+  {
+    const auto &ptr = target.operands().front();
+    // bring the offset back to zero
+    auto res = minus_exprt{ptr, pointer_offset(ptr)};
+    res.type() = ptr.type();
+    return std::move(res);
```
tautschnig: Scratch my earlier comment, the following will actually do the right thing: `return minus_exprt{ptr, pointer_offset(ptr)};` (the type of the `minus_exprt` will be `ptr.type()` when constructed in this way).

src/goto-instrument/contracts/havoc_assigns_clause_targets.h
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.h
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.h
remi-delmas-3000: indeed this works

regression/contracts/assigns_replace_havoc_dependent_targets_fail/enforce.desc
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.h
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: thanks

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: renamed to `build_address_of`

remi-delmas-3000:
I took into account all pending comments, it should be good to go now.

tautschnig:
Requires approval from @peterschrammel @martin-cs @chrisr-diffblue as a new file was added to `goto-instrument/Makefile`.

regression/contracts/history-constant/main.c
```diff
@@ -0,0 +1,15 @@
+#include <stdio.h>
```
feliperodri, line 1: Why do we need this include?

regression/contracts/history-typecast/main.c
```diff
@@ -0,0 +1,23 @@
+#include <stdio.h>
```
feliperodri, line 1: Why do we need this include?

feliperodri (approved):
It needs code owner approval from @peterschrammel @chrisr-diffblue or @martin-cs.

chrisr-diffblue (approved):
Approving the Makefile changes. Nothing to add beyond the other detailed reviews.

PR #6541, opened 2021-12-22
Simplification of nested pointer arithmetic: do not assume same type

tautschnig:
Please review commit-by-commit: the first commit is a refactoring only.

### DIFF
--- a/regression/cbmc/Pointer_Arithmetic15/main.c
+++ b/regression/cbmc/Pointer_Arithmetic15/main.c
@@ -15,5 +15,9 @@ int main()
 
   a = p - q;
 
+  // mixing types: the C front-end will insert casts
+  unsigned long long u;
+  p = (p + a) + u;
+
   assert(0);
 }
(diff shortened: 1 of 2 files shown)


kroening:
Does it make sense to tighten the typing rules for pointer arithmetic, and then to change the front-end?
The standard is vague on these, we would need to invent our own.

tautschnig:
> Does it make sense to tighten the typing rules for pointer arithmetic, and then to change the front-end? The standard is vague on these, we would need to invent our own.

I have now extended an existing test: the C front-end already does ensure consistent typing, so I don't think we have an issue in the front-end to resolve. No other place, however, is strict about the types in pointer arithmetic. Perhaps we should be? I don't know.

kroening:
I'd be in favour of tightening the typing; this makes every bit of code that 'reads' a lot easier. There will be more of that than code that 'writes', and that would not get much more difficult.

tautschnig:
> I'd be in favour of tightening the typing; this makes every bit of code that 'reads' a lot easier. There will be more of that than code that 'writes', and that would not get much more difficult.

@kroening As now being discussed in #6731 making changes in Kani could well be the way to go. That said, however, this PR also fixes other issues in the offending simplifier code as indicated in the commit message, so perhaps merging these changes is a right first step?

kroening (approved):

PR #6542, opened 2021-12-22
Consistently use exprt::is_constant()

tautschnig:
We would sometimes use is_constant and at other times id() ==
ID_constant (which is exactly what is_constant() does). One option was
to get rid of is_constant(), but it seemed safer to use this function
than to stick with the lower-level id() == ID_constant: the function is
type safer as it can only be used on exprt, while comparisons on id()
would work on any irept.

To ensure there is no performance penalty, is_constant() has been moved
to the header file to permit inlining.

### DIFF
--- a/jbmc/src/java_bytecode/java_trace_validation.cpp
+++ b/jbmc/src/java_bytecode/java_trace_validation.cpp
@@ -104,19 +104,15 @@ bool check_struct_structure(const struct_exprt &expr)
   {
     return false;
   }
-  if(
-    expr.operands().size() > 1 && std::any_of(
-                                    ++expr.operands().begin(),
-                                    expr.operands().end(),
-                                    [&](const exprt &operand) {
-                                      return operand.id() != ID_constant &&
-                                             operand.id() !=
-                                               ID_annotated_pointer_constant;
-                                    }))
-  {
-    return false;
-  }
-  return true;
+
+  return expr.operands().size() == 1 ||
+         std::all_of(
+           std::next(expr.operands().begin()),
+           expr.operands().end(),
+           [&](const exprt &operand) {
+             return can_cast_expr<constant_exprt>(operand) ||
+                    can_cast_expr<annotated_pointer_constant_exprt>(operand);
+           });
 }
 
 bool check_address_structure(const address_of_exprt &address)
--- a/src/solvers/strings/string_refinement_util.cpp
+++ b/src/solvers/strings/string_refinement_util.cpp
@@ -51,8 +51,10 @@ bool has_char_array_subexpr(const exprt &expr, const namespacet &ns)
 std::string
 utf16_constant_array_to_java(const array_exprt &arr, std::size_t length)
 {
-  for(const auto &op : arr.operands())
-    PRECONDITION(op.id() == ID_constant);
+  PRECONDITION(std::all_of(
+    arr.operands().begin(),
+    arr.operands().end(),
+    can_cast_expr<constant_exprt>));
 
   std::wstring out(length, '?');
 
--- a/src/util/expr.cpp
+++ b/src/util/expr.cpp
@@ -22,13 +22,6 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <stack>
 
-/// Return whether the expression is a constant.
-/// \return True if is a constant, false otherwise
-bool exprt::is_constant() const
-{
-  return id()==ID_constant;
-}
-
 /// Return whether the expression is a constant representing `true`.
 /// \return True if is a Boolean constant representing `true`, false otherwise.
 bool exprt::is_true() const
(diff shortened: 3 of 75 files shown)


src/solvers/strings/string_refinement_util.cpp
```diff
@@ -51,8 +51,10 @@ bool has_char_array_subexpr(const exprt &expr, const namespacet &ns)
 std::string
 utf16_constant_array_to_java(const array_exprt &arr, std::size_t length)
 {
-  for(const auto &op : arr.operands())
-    PRECONDITION(op.id() == ID_constant);
+  PRECONDITION(std::all_of(
+    arr.operands().begin(), arr.operands().end(), [](const exprt &op) {
```
thomasspriggs: :bulb: Because `can_cast_expr` is a free function, rather than a member function, you can just write `can_cast_expr<constant_exprt>`, here rather than writing a lambda.

jbmc/src/java_bytecode/java_trace_validation.cpp
```diff
@@ -98,16 +98,12 @@ bool check_struct_structure(const struct_exprt &expr)
...
-      expr.operands().end(),
-      [&](const exprt &operand) { return operand.id() != ID_constant; }))
-  {
-    return false;
-  }
-  return true;
+
+  return expr.operands().size() == 1 ||
+         std::all_of(
+           ++expr.operands().begin(),
```
peterschrammel: ⛏️ `std::next(expr.operands().begin())`

peterschrammel (approved):

src/util/expr.cpp
```diff
@@ -21,13 +21,6 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <stack>
 
-/// Return whether the expression is a constant.
-/// \return True if is a constant, false otherwise
-bool exprt::is_constant() const
```
thomasspriggs, line 27: Do you actually have performance testing results to show that there are performance benefits to moving this into the header?

jbmc/src/java_bytecode/java_trace_validation.cpp
tautschnig: Done!

thomasspriggs (commented):
I'd prefer to see standardising on `can_cast_expr<constant_exprt>` instead, because I think it makes it clearer that the operation is a simple type check and it includes the type we are checking against. But the current changes in this PR are better than checking `.id()==ID_constant`.

src/util/expr.cpp
tautschnig: No, I haven't actually done measurements.

thomasspriggs: Do you think it would be more in-keeping with the PR review checklist you added to either leave the function where it is or to do the performance measurement and include it in the commit message? As a small aside I would have preferred to have seen the performance related change in a separate commit, in order to make it easier to evaluate in isolation.

tautschnig: There's a bit of a problem here: replacing `id() == ID_constant` by `.is_constant()` (without inlining it) will entail a function call. So there would inevitably be additional cost, although that need not translate into practically notable cost.

What I could do is not introduce new uses of `is_constant()` and just benchmark the performance impact of inlining based on the existing function calls.

tautschnig:
> I'd prefer to see standardising on `can_cast_expr<constant_exprt>` instead, because I think it makes it clearer that the operation is a simple type check and it includes the type we are checking against. But the current changes in this PR are better than checking `.id()==ID_constant`.

I'd propose to do this PR as it is as first step and then deprecate `exprt::is_constant` in favour of `can_cast_expr<constant_exprt>`. The reason I'd like to propose doing it in two steps is that a mechanic rewrite of all uses of `.is_constant()` is much easier than trying to rewrite the current mixed bag.

thomasspriggs (approved):
I am a little concerned about a creep of functions from being implemented in `.cpp` to header files over time. This is because it could may make a marginal difference to the build time, in the same way that adding function call overhead may make a difference to execution time. In my opinion if function calls like this are adding measurable overhead then that would be an argument to update the build systems to use link time optimisation instead of moving the definitions around.

I am approving this as is because using consistent interfaces/abstraction is more important than any argument about where functions should be defined.

tautschnig:
> I am a little concerned about a creep of functions from being implemented in `.cpp` to header files over time. This is because it could may make a marginal difference to the build time, in the same way that adding function call overhead may make a difference to execution time. In my opinion if function calls like this are adding measurable overhead then that would be an argument to update the build systems to use link time optimisation instead of moving the definitions around.
> 
> I am approving this as is because using consistent interfaces/abstraction is more important than any argument about where functions should be defined.

+100 to what you said. So I am planning to deprecate `is_constant()` and follow up by a PR that uses `can_cast_expr` instead as you suggested in an earlier comment. I do care about compile time, a lot.

kroening:
I feel that can_cast_expr is a downgrade in terms of readability compared to both .is_constant() and == ID_constant.

NlightNFotis (approved):

kroening (approved):

PR #6543, opened 2021-12-22
Consistently use exprt::is_boolean()

tautschnig:
We would sometimes use is_boolean and at other times type().id() ==
ID_bool (which is exactly what is_boolean() does). One option was to get
rid of is_boolean(), but it seemed safer to use this function than to
stick with the lower-level type().id() == ID_bool: the function is type
safer as it can only be used on exprt, while comparisons on id() would
work on any irept.

To ensure there is no performance penalty, is_boolean() has been moved
to the header file to permit inlining.

### DIFF
--- a/src/ansi-c/c_typecheck_expr.cpp
+++ b/src/ansi-c/c_typecheck_expr.cpp
@@ -431,7 +431,7 @@ void c_typecheck_baset::typecheck_expr_main(exprt &expr)
     // This is one of the few places where it's detectable
     // that we are using "bool" for boolean operators instead
     // of "int". We convert for this reason.
-    if(op.type().id() == ID_bool)
+    if(op.is_boolean())
       op = typecast_exprt(op, signed_int_type());
 
     irept::subt &generic_associations=
@@ -960,7 +960,7 @@ void c_typecheck_baset::typecheck_expr_sizeof(exprt &expr)
     // This is one of the few places where it's detectable
     // that we are using "bool" for boolean operators instead
     // of "int". We convert for this reason.
-    if(op.type().id() == ID_bool)
+    if(op.is_boolean())
       type = signed_int_type();
     else
       type = op.type();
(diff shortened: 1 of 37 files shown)


peterschrammel (approved):

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -3294,21 +3294,17 @@ void c_typecheck_baset::typecheck_function_call_arguments(
     }
     else if(i<parameter_types.size())
     {
-      const code_typet::parametert &parameter_type=
```
thomasspriggs: :pick: Could you separate this refactor out into a separate commit from the using of `exprt::is_boolean()`? This would have made it easier to review and it could make it easier to bisect later, if we find problems have been introduced.

src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -3294,21 +3294,17 @@ void c_typecheck_baset::typecheck_function_call_arguments(
...
+      const typet &parameter_type = parameter_types[i].type();
 
-      const typet &op_type=parameter_type.type();
-
-      if(op_type.id()==ID_bool &&
-         op.id()==ID_side_effect &&
-         op.get(ID_statement)==ID_assign &&
-         op.type().id()!=ID_bool)
+      if(
+        parameter_type.id() == ID_bool && op.id() == ID_side_effect &&
```
thomasspriggs: :pick: If you had left `parameter_type` as being a `code_typet::parametert`, then `parameter_type.id() == ID_bool` could be `parameter_type.is_boolean()`.
:bulb: Maybe `parameter_type` should just be `parameter` and `parameter_types` should be just `parameters`. These appear to be the parameter expressions, not their types.

thomasspriggs (approved):

NlightNFotis (approved):
LGTM

src/ansi-c/c_typecheck_expr.cpp
tautschnig: I have now dropped this refactor from this PR.

src/ansi-c/c_typecheck_expr.cpp
tautschnig: Now in #7226

kroening (approved):

PR #6546, opened 2021-12-23
Function contracts: preliminary support for conditional targets in assigns clause checking

remi-delmas-3000:
This PR adds support for conditional targets in assigns clauses.

A conditional target specifies a memory location that can be assigned by a function only when a condition holds at the call site.

When a function call is replaced by its contract, conditional targets are havocked only if the condition holds at the call site.

# Syntax
comma separated list of targets can now be preceded by `condition :`, and multiple target lists with or without conditions are separated by `;`:

```C
__CPROVER_assigns(
a, b, c; // unconditional targets
condition: x, y, z; // conditional targets
!condition && other_condition: q, r, t; // conditional targets
)
```

# Use cases

These conditional targets allow the user to specify frame conditions for functions that have distinct modes of operation and assign different locations depending on the mode.

They also allow the user to specify validity conditions for the assigned locations, in case they are accessed through chains of struct/union member selections, pointer dereferences, can are not always valid. For instance in the example below, `cond` would be a condition which guarantees that the target expression corresponds to a valid memory location:

```C
__CPROVER_assigns(cond: p->union.member->array[idx]->member) 
```

# Changes

In the ansi-c front-end
- In the parser, modify the target rule to accept the new syntax;
- Add an new expression class and `irep_id` for the new construct;
- Update the `expr2ct` expression formatter class to print the new expression type;
- Add typechecking rules:
  - targets must be lvalues, without ternary operators or side-effects;
  - conditions are still allowed to contain `side_effect_function_callt` but no other `side_effect_exprt` subtypes.

Remark: In a future PR we will check that any function used in a condition is side-effect-free, i.e. has a contract with an empty assigns clause. For now we trust the user not to use side-effecting functions in conditions.

In the `contracts/assigns.{h|c}` code:
1. Add an explicit `guard` expression to the internal representation of assigns clause targets;
2. Update the CAR validity condition to `guard && all_deref_valid(target) && rw_ok(target, size)`;
3. Propagate these changes to the instrumentation of assigns clause checking;
4. Propagate these changes to the havocking of targets for contract replacement.


regression/contracts/aaa-assigns_enforce_guarded_assigns_targets/main.c
```diff
@@ -0,0 +1,106 @@
+#include <stdbool.h>
```
tautschnig: Random place for comment: it would be great if the commit message explained

1) what a "guarded assigns clause target" is, and
2) _why_ this change is being made (the commit message answers the "what?" question only).

It might be that answering 1) already answers 2), though.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +70,98 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
 
+/// Result type for guarded assigns target pattern matching
+using if_match_resultt = optionalt<std::pair<exprt, exprt>>;
+
+/// Matches an expression with ID_NULL or  ID_constant and value "0",
+/// modulo some intermediate ID_typecasts
+bool match_NULL_or_zero(const exprt &expr)
+{
+  if(expr.id() == ID_typecast)
+    return match_NULL_or_zero(expr.operands().front());
```
tautschnig: You'll want to use `skip_typecast` from `expr_util.{h,cpp}`.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +70,98 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+  else
+    return expr.id() == ID_NULL ||
+           (expr.id() == ID_constant &&
+            id2string(to_constant_expr(expr).get_value()) == "0");
```
tautschnig: We should have a comment here that this is to be cleaned up once #6532 lands.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +70,98 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+/// Matches an expression of the form `guard ? target : (NULL|0)`
+/// modulo intermediary typecasts.
+/// \param expr the candidate expression
+/// \returns A pair `(guard,target)` if successful.
+if_match_resultt match_if(const exprt &expr)
+{
+  if(expr.id() == ID_typecast)
+  {
+    return match_if(expr.operands().front());
+  }
```
tautschnig: Use `skip_typecast`

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +70,98 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+  else if(expr.id() == ID_if)
+  {
+    if(!match_NULL_or_zero(expr.operands().at(2)))
+      return {};
+
+    return {{expr.operands().at(0), expr.operands().at(1)}};
```
tautschnig: Please use `to_if_expr` and then access the operands using `if_exprt`'s API instead of `.operands().at(...)`

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +70,98 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+normalize_to_guarded_slice(const exprt &expr, const namespacet &ns)
+{
+  if(expr.id() == ID_pointer_object)
+  {
+    if_match_resultt match_opt = match_if(expr.operands().at(0));
+    if(match_opt.has_value())
+    {
+      const auto &match = match_opt.value();
+      const auto &new_expr = pointer_object(match.second);
+      const auto &slice = normalize_to_slice(new_expr, ns);
```
tautschnig: Both of these return by value, so there isn't much of a point taking a reference here. It might even be undesirable as it might block copy elision.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +70,98 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+      return {match.first, new_expr, slice.first, slice.second};
+    }
+    else
+    {
+      const auto &slice = normalize_to_slice(expr, ns);
+      goto_programt empty_program;
```
tautschnig: Never used?

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +70,98 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+      return {true_exprt{}, expr, slice.first, slice.second};
+    }
+  }
+
+  if_match_resultt match_opt = match_if(expr);
+  if(match_opt.has_value())
+  {
+    const auto &match = match_opt.value();
+    const auto &slice = normalize_to_slice(match.second, ns);
```
tautschnig: Should not be a reference.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +70,98 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+    return {match.first, match.second, slice.first, slice.second};
+  }
+  else
+  {
+    const auto &slice = normalize_to_slice(expr, ns);
```
tautschnig: Should not be a reference.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +70,98 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+{
+  if(expr.id() == ID_pointer_object)
+  {
+    if_match_resultt match_opt = match_if(expr.operands().at(0));
+    if(match_opt.has_value())
+    {
+      const auto &match = match_opt.value();
+      const auto &new_expr = pointer_object(match.second);
+      const auto &slice = normalize_to_slice(new_expr, ns);
+      return {match.first, new_expr, slice.first, slice.second};
```
tautschnig: It might be worth adding a constructor that takes a `slicet` to avoid this `slice.first, slice.second` (which repeats below).

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -93,6 +207,17 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
...
+  null_message_handlert null_handler;
+
+  // clean up side effects from the guard expression
+  // we want checks on the guard evaluation instructions because it is user code
+  cleanert cleaner(parent.symbol_table, null_handler);
+  exprt clean_guard(guarded_slice.guard);
+  const auto &mode = parent.symbol_table.lookup_ref(parent.function_name).mode;
+  cleaner.clean(clean_guard, instructions, mode);
+  for(auto &inst : instructions.instructions)
+    inst.source_location_nonconst() = location;
```
tautschnig: Adding that `location` to `clean_guard` should suffice for source locations to be set up correctly, no need for this `for` loop once that's done.

tautschnig (approved):

regression/contracts/assigns_enforce_conditional_function_call_condition/main.c
```diff
@@ -0,0 +1,32 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
+
+bool cond()
+{
+  return true;
+}
+
+int foo(int a, int *x, int *y) __CPROVER_assigns(a && cond() ? *x)
```
tautschnig: I am not entirely convinced this is the best syntax to use: it seems that `?` now has a very difference precedence?

src/ansi-c/c_typecheck_base.h
```diff
@@ -146,7 +146,16 @@ class c_typecheck_baset:
...
+
+  /// \brief Typechecks a vector of assigns clause targets.
+  ///
+  /// \remark The vector will grow larger in case it contains conditional
+  /// targets of the form `c ? {t1, .. , tn}`, which get expanded to a
+  /// collection `{c ? t1, .. , c ? tn }` by distributing the condition `c`
+  /// on the list of targets.
   virtual void typecheck_spec_assigns(exprt::operandst &targets);
+  virtual void typecheck_spec_assigns_condition(exprt &condition);
+  virtual void typecheck_spec_assigns_target(exprt &target);
```
tautschnig: I think you should use Doxygen groups or `\copydoc` to make clear which methods this comment refers to.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +843,158 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
 
+void c_typecheck_baset::typecheck_spec_assigns_condition(exprt &condition)
+{
+  // compute type
+  typecheck_expr(condition);
+
+  // non-fatal warnings
+  if(has_subexpr(condition, [&](const exprt &subexpr) {
+       return can_cast_expr<side_effect_expr_function_callt>(subexpr);
+     }))
```
tautschnig: I'm not entirely sure, but the following might also work (and would be shorter):
`if(has_subexpr(condition, can_cast_expr<side_effect_expr_function_callt>))`

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +843,158 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+
+  if(has_subexpr(condition, [&](const exprt &subexpr) {
+       return can_cast_expr<side_effect_exprt>(subexpr) &&
+              !can_cast_expr<side_effect_expr_function_callt>(subexpr);
+     }))
+  {
+    must_throw = true;
+    error().source_location = condition.source_location();
+    error() << "side-effects not allowed in assigns clause conditions" << eom;
+  }
```
tautschnig, line 888: How about making this an `else if` of the earlier `side_effect_expr_function_callt` check?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +843,158 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
-    {
-      // skip
-    }
-    else if(!target.get_bool(ID_C_lvalue))
-    {
-      error().source_location = target.source_location();
-      error() << "non-lvalue target in assigns clause" << eom;
-      throw 0;
+      // typecheck condition
+      auto &condition = target.operands().at(0);
```
tautschnig: Please introduce a proper API in `ansi-c/c_expr.h` so you'd here end up with `target.condition()` (well, it likely wouldn't be "target" anymore as you'd have an intermediate `to_conditional_assigns_target` call).

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +843,158 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+      // make it boolean if needed
+      if(!can_cast_type<bool_typet>(condition.type()))
+      {
+        if(can_cast_type<pointer_typet>(condition.type()))
+          condition = notequal_exprt{
+            condition, null_pointer_exprt(to_pointer_type(condition.type()))};
+        else
+          condition =
+            notequal_exprt{condition, from_integer(0, condition.type())};
+      }
```
tautschnig: Use `implicit_typecast_bool(condition);` instead.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +843,158 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
-      error().source_location = target.source_location();
-      error() << "assigns clause is not side-effect free" << eom;
-      throw 0;
+      typecheck_spec_assigns_target(target);
+      tmp.push_back(std::move(target));
     }
   }
+  targets.clear();
+  for(auto &each : tmp)
+    targets.push_back(std::move(each));
```
tautschnig: `std::swap(targets, tmp);` might do the trick?

src/ansi-c/parser.y
```diff
@@ -3297,8 +3297,55 @@ cprover_function_contract:
...
+          $$=$1;
+        }
+        | logical_equivalence_expression '?' '{' logical_equivalence_expression_list '}'
+        { 
+          $$=$2;
+          parser_stack($$).id(ID_conditional_assigns_target);
+          mto($$, $1);
+          mto($$, $4);
+        }
+        | logical_equivalence_expression '?' logical_equivalence_expression
```
tautschnig: Is that really what you want/expect here? I would have thought a `unary_expression` would already be a superset of what the type checker will/should accepted on the right-hand side of `?`?

tautschnig (changes requested):
I believe the changes to the C front-end require some more work as detailed in my comments.

regression/contracts/assigns_enforce_conditional_function_call_condition/main.c
kroening: Frama-C has gone that way, with different precedence rules depending on the context of an expression. I would much like to avoid anything the like.

feliperodri:
@remi-delmas-3000 (nitpic) is there any particular reason to add "Function contracts: " as a suffix for every commit (and title) of this PR? I think this pollutes the message more than inform anything. I'd suggest removing it and keep the remaining text.

regression/contracts/assigns_enforce_conditional_function_call_condition/main.c
remi-delmas-3000: Do you have a suggestion for the operator syntax ? 

` cond ==> {target, .. }` was considered at some point but there would be similar issues in case the condition contains an operator like `<==>`.

The more verbose `if(cond) target` or `if(cond) {target, .., target }` ?

 Maybe a completely new digraph or trigraph operator  like`??` or `?=>` ?

src/ansi-c/c_typecheck_code.cpp
remi-delmas-3000: Since these cases are not exclusive (as you can  have both a `side_effect_expr_function_callt` and `side_effect_exprt` that is not a `side_effect_expr_function_callt` in the same expression) we need to keep these checks independent. What I'm also trying to do here is to report all warnings and errors before bailing out instead of stopping at the first one.

src/ansi-c/c_typecheck_code.cpp
remi-delmas-3000: OK

regression/contracts/assigns_enforce_conditional_function_call_condition/main.c
tautschnig: What about `:` (colon)?

src/ansi-c/c_typecheck_base.h
remi-delmas-3000: it only concerns `typecheck_spec_assigns`, I added a new line to split the block of declarations in two.

src/ansi-c/c_typecheck_code.cpp
remi-delmas-3000: Related question: is there already a function such as 

```
iterator<exprt &> iter_subexpr(
  const exprt &expr,
  const std::function<bool(const exprt &)> &pred),
  traversal_ordert order
)
```

That would let me iterate (in some chosen or unspecified order) on subexpressions that match a given predicate for further inspection?

tautschnig: I don't think we have that in full, but at least the iteration can be done using one of the iterators defined in `util/expr_iterator.h`. This is used in several places, you might take a look at `util/rename_symbol.cpp` for an example.

src/ansi-c/parser.y
remi-delmas-3000: Thanks, it seems to still allow what we need and more, but less than before.

src/ansi-c/c_typecheck_code.cpp
remi-delmas-3000: > @remi-delmas-3000 (nitpic) is there any particular reason to add "Function contracts: " as a suffix for every commit (and title) of this PR? I think this pollutes the message more than inform anything. I'd suggest removing it and keep the remaining text.

@feliperodri, it was a request from @tautschnig  to make it clear in the log history that a given commit is scoped to function contracts, and make it easier to filter the git history when there are regressions in CBMC-core.

remi-delmas-3000:
> @remi-delmas-3000 (nitpic) is there any particular reason to add "Function contracts: " as a suffix for every commit (and title) of this PR? I think this pollutes the message more than inform anything. I'd suggest removing it and keep the remaining text.

@feliperodri, it was a request from @tautschnig  to make it clear in the log history that a given commit is scoped to function contracts, and make it easier to filter the git history when there are regressions in CBMC-core.

regression/contracts/assigns_enforce_conditional_function_call_condition/main.c
remi-delmas-3000: Yes that could work !

How about this:

```C
assigns_clause: '__CPROVER_assigns' '(' [ cond_targets (';' cond_targets)* [ ';' ]  ] ')' 
cond_targets: [ logical_equiv_expr ':' ] unary_expr (',' unary_expr)*
```

This is retro compatible with `,`-separated lists of unconditional targets  
```C
__CPROVER_assigns(t1, t2, ... )
```
And we can have `;`-separated lists of lists of conditional targets:

```C
__CPROVER_assigns(t1, t2; cond: t3, t4; other_cond: t5, t6, t7)
``` 

with an optional ';' at the end you can indent like this:
```C
__CPROVER_assigns(
  t1, t2;
  cond: t3, t4;
  other_cond: t5, t6, t7;
)
```

tautschnig: I'm happy with that if you can come up with a grammar that doesn't require more lookahead!

src/ansi-c/c_expr.h
```diff
@@ -225,4 +225,69 @@ to_history_expr(const exprt &expr, const irep_idt &id)
...
+/// \brief A class for an expression that represents a
+/// conditional target in an assigns clause.
+class conditional_assigns_target_exprt : public binary_exprt
+{
+public:
+  explicit conditional_assigns_target_exprt(exprt _condition, exprt _target)
+    : binary_exprt(
+        std::move(_condition),
+        ID_conditional_assigns_target,
+        std::move(_target))
```
tautschnig: Nit pick: using this constructor of `binary_exprt`, the expression will implicitly have as `type()` whatever `_condition.type()` is. We shouldn't really use the type of this expression, so I'd suggest to explicitly set the type either as `typet{}` (and use the 4-parameter constructor of `binary_exprt`) to make sure no one ends up using that type.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -11,6 +11,8 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include "c_typecheck_base.h"
 
+#include <ansi-c/c_expr.h>
```
tautschnig: Move this further down, underneath `#include "ansi_c_declaration.h"`, and use `#include "c_expr.h"` for consistency.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +844,147 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+  // fatal errors
+  bool must_throw = false;
+
+  if(condition.type().id() == ID_empty)
+  {
+    must_throw = true;
+    error().source_location = condition.source_location();
+    error() << "void-typed expressions not allowed as assigns clause conditions"
+            << eom;
+  }
```
tautschnig, line 878: Isn't this caught by `implicit_typecast_bool` already?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +844,147 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
 void c_typecheck_baset::typecheck_spec_assigns(exprt::operandst &targets)
 {
+  exprt::operandst tmp;
   for(auto &target : targets)
   {
-    typecheck_expr(target);
-
-    if(target.type().id() == ID_empty)
+    // conditional target
+    if(can_cast_expr<conditional_assigns_target_exprt>(target))
```
tautschnig: You might contemplate the use of `if(auto conditional_target_ptr = expr_try_dynamic_cast<conditional_assigns_target_exprt>(target))`

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +844,147 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+      }
     }
-    else if(has_subexpr(target, ID_side_effect))
+    else // unconditional target
     {
-      error().source_location = target.source_location();
-      error() << "assigns clause is not side-effect free" << eom;
-      throw 0;
+      typecheck_spec_assigns_target(target);
+      tmp.push_back(std::move(target));
```
tautschnig: Would it be of any value to make up a conditional target with condition `true_exprt{}` here?

src/ansi-c/parser.y
```diff
@@ -3297,8 +3297,55 @@ cprover_function_contract:
...
+        }
+        | logical_equivalence_expression '?' '{' unary_expression_list '}'
+        { 
+          $$=$2;
+          parser_stack($$).id(ID_conditional_assigns_target);
+          mto($$, $1);
+          mto($$, $4);
+        }
+        | logical_equivalence_expression '?' unary_expression
+        { $$=$2;
```
tautschnig: Nit pick: newline after `{`

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -18,11 +18,30 @@ Date: July 2021
 
 #include <langapi/language_util.h>
 
+#include <goto-programs/goto_convert_class.h>
```
tautschnig: Nit pick: I'd prefer a blank line between includes from different directories.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +71,35 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+    const auto &new_expr = expr.operands().at(1);
+    const auto slice = normalize_to_slice(new_expr, ns);
+    return {expr.operands().at(0), new_expr, slice.first, slice.second};
+  }
+  else
+  {
+    const auto slice = normalize_to_slice(expr, ns);
+    return {true_exprt{}, expr, slice.first, slice.second};
+  }
+}
```
tautschnig: I suppose we could get rid of this if the the front-end always generated conditional assigns targets?

tautschnig (commented):
Looks much better! I'm holding off on approving until we've concluded the syntax discussion.

regression/contracts/assigns_enforce_conditional_unions/test.desc
```diff
@@ -0,0 +1,24 @@
...
+--enforce-contract update _ --pointer-check --pointer-overflow-check --signed-overflow-check --unsigned-overflow-check --conversion-check
+^\[is_high_level.\d+\] line 52 Check that latch is assignable: SUCCESS$
+^\[is_high_level.\d+\] line 53 Check that once is assignable: SUCCESS$
+^\[update.\d+\] line 85 Check that POINTER_OBJECT\(\(void \*\)state->digest.high_level.first.ctx->digest\) is assignable: SUCCESS$
+^\[update.\d+\] line 86 Check that state->digest.high_level.first.ctx->flags is assignable: SUCCESS$
+^\[update.\d+\] line 89 Check that POINTER_OBJECT\(\(void \*\)state->digest.high_level.second.ctx->digest\) is assignable: FAILURE$
+^\[update.\d+\] line 90 Check that state->digest.high_level.second.ctx->flags is assignable: FAILURE$
+^\[update.\d+\] line 96 Check that POINTER_OBJECT\(\(void \*\)state->digest.high_level.first.ctx->digest\) is assignable: FAILURE$
+^\[update.\d+\] line 97 Check that state->digest.high_level.first.ctx->flags is assignable: FAILURE$
+^\[update.\d+\] line 100 Check that POINTER_OBJECT\(\(void \*\)state->digest.high_level.second.ctx->digest\) is assignable: SUCCESS$
```
tautschnig: This one is failing in regression tests right now.

src/ansi-c/c_expr.h
```diff
@@ -225,4 +225,98 @@ to_history_expr(const exprt &expr, const irep_idt &id)
...
+inline const conditional_target_exprt &
+to_conditional_target_expr(const exprt &expr)
+{
+  PRECONDITION(expr.id() == ID_conditional_target);
+  auto &ret = static_cast<const conditional_target_exprt &>(expr);
+  validate_expr(ret);
+  return ret;
+}
+
+/// \copydoc to_conditional_target_expr(exprt &expr)
```
tautschnig: `const exprt &expr` to fix the Doxygen error

src/ansi-c/c_typecheck_base.h
```diff
@@ -146,7 +146,17 @@ class c_typecheck_baset:
   virtual void typecheck_start_thread(codet &code);
 
   // contracts
+
+  /// \brief Typechecks a vector of assigns clause targets.
+  ///
+  /// \remark The vector will grow larger in case it contains conditional
+  /// targets of the form `c ? {t1, .. , tn}`, which get expanded to a
+  /// collection `{c ? t1, .. , c ? tn }` by distributing the condition `c`
```
tautschnig: Syntax needs updating

src/ansi-c/c_expr.h
```diff
@@ -225,4 +225,98 @@ to_history_expr(const exprt &expr, const irep_idt &id)
...
+/// a list of targets sharing a common condition
+/// in an assigns clause.
+class conditional_target_exprt : public exprt
+{
+public:
+  explicit conditional_target_exprt(exprt _condition, exprt _target_list)
+    : exprt(ID_conditional_target, empty_typet{})
+  {
+    add_to_operands(std::move(_condition));
+    add_to_operands(std::move(_target_list));
```
tautschnig: Why did you choose to no longer make it a `binary_exprt`?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +844,167 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+      // place each single target in its own conditional target expr,
+      // distributing the condition
+      for(auto &actual_target : conditional_target.targets())
+      {
+        typecheck_spec_assigns_target(actual_target);
+        exprt ops{ID_expression_list};
+        ops.add_to_operands(actual_target);
+        conditional_target_exprt result{condition, ops};
+        result.add_source_location() = conditional_target.source_location();
+        tmp.push_back(std::move(result));
```
tautschnig: This appears to be the only place that constructs a `conditional_target_exprt`. I would, therefore, expect that `conditional_target_exprt` has a constructor that permit creating the desired expression with just a single statement, like this:
```
tmp.push_back(conditional_target_exprt{condition, exprt{ID_expression_list, typet{}, {actual_target}}, conditional_target.source_location()});
```

src/ansi-c/expr2c.cpp
```diff
@@ -3470,6 +3470,31 @@ std::string expr2ct::convert_sizeof(
   return dest;
 }
 
+std::string expr2ct::convert_conditional_target(const exprt &src)
```
tautschnig: Consume a `const conditional_target_exprt &src` to save yourself some trouble.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -18,8 +18,10 @@ Date: July 2021
 
 #include <langapi/language_util.h>
 
+#include <ansi-c/c_expr.h>
```
tautschnig, line 21: I'd appreciate a blank line after this one.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -15,6 +15,7 @@ Author: Remi Delmas, delmasrd@amazon.com
 
 #include <langapi/language_util.h>
 
+#include <ansi-c/c_expr.h>
```
tautschnig, line 18: Blank line after this one, please.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -23,6 +24,8 @@ Author: Remi Delmas, delmasrd@amazon.com
 #include <util/pointer_predicates.h>
 #include <util/std_code.h>
 
+#include <iostream>
+
```
tautschnig: Debugging left-over?

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -71,25 +74,29 @@ static exprt build_address_of(
 
     return address_of_exprt{target};
   }
+
+  std::cout << target.pretty() << std::endl;
```
tautschnig: Debugging left-over?

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -211,38 +265,45 @@ void havoc_assigns_clause_targets(
...
-      tracking_comment,
-      havoc_program,
-      source_location,
-      ns);
+    if(can_cast_expr<conditional_target_exprt>(target))
+    {
+      const auto &cond_target = to_conditional_target_expr(target);
+
+      // clean the condition if needed
+      null_message_handlert null_handler;
```
tautschnig: Here and in the other place that introduces a `null_message_handlert`: should you perhaps use `log.get_message_handler()` from code_contractst?

tautschnig (approved):
Several comments need addressing, but I'm generally ok with this.

src/ansi-c/c_expr.h
remi-delmas-3000: I was't really sure what a `binary_exprt` is supposed to represent, but from reading the code it seemed to be meant for binary operators such as `<`, `+`, where both left and right subtrees are homogenous.
This one being asymmetric and only meant to occur at a specify place in the AST seemed like a special case, it did not seem like reusing the `binary_exprt` interface would bring much.

regression/contracts/assigns_enforce_conditional_unions/test.desc
remi-delmas-3000: The tool is right and the test case is badly specified, fixing it in a new update.

feliperodri:
> @feliperodri, it was a request from @tautschnig to make it clear in the log history that a given commit is scoped to function contracts, and make it easier to filter the git history when there are regressions in CBMC-core.

@remi-delmas-3000 We should have "labels" to make it easier to filter contributions in the git history, similarly as we see in the [linux](https://github.com/torvalds/linux/commits/master) project. For example, take a look at this commit message:
```git
tracing: Tag trace_percpu_buffer as a percpu pointer

Tag trace_percpu_buffer as a percpu pointer to resolve warnings
reported by sparse:
...
```
We can adopt the tag `CONTRACT` to all work related to function and loop contracts (I don't think there is a need to distinct between them) and adopt [these guidelines](https://cbea.ms/git-commit/) for commit messages. Linux also offers some [guidelines about describing changes](https://github.com/torvalds/linux/blob/master/Documentation/process/submitting-patches.rst#describe-your-changes), but the  one from `cbea` it's clear IMHO. Following these guidelines, your current commit message could be
```git
CONTRACTS: Preliminary support for conditional assigns clauses

Conditional targets specify memory locations that can be assigned if and
only if a condition holds when a function is invoked. It does not yet
work for loops.

- Add a new `conditional_target_exprt` class and associated `irep_ID`;
- Modify ansi-c parser and typechecker to accept the new target syntax
  `cond: t1, ..., tn`;
- Update the CAR class `conditional_address_ranget` with explicit
  condition;
- Change the validity condition of a CAR to `condition &&
  all_deref_valid(target) && rw_ok(target, size)`, for both contract
checking and contract replacement logic;
- Update some tests, add new tests for the feature.
```
I can add a README file to our `src/goto-instrument/contracts` module with a reference for these guidelines and start a "contributors guide". (cc. @tautschnig)

regression/contracts/assigns_enforce_conditional_function_call_condition/main.c
```diff
@@ -0,0 +1,32 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>`.

regression/contracts/assigns_enforce_conditional_function_call_condition/test.desc
```diff
@@ -0,0 +1,17 @@
...
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+Checks that function calls are allowed in conditions.
+Remark: we allow function calls without further checks for now because they
+are mandatory for some applications.
+The next step must be to check that the called functions have a contract
+with an empty assigns clause and that they indeed satisfy their contract
+using a CI check.
```
feliperodri, line 17: Is there a issue to track this?

regression/contracts/assigns_enforce_conditional_lvalue/main.c
```diff
@@ -0,0 +1,31 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_lvalue/test.desc
```diff
@@ -0,0 +1,17 @@
...
+^\[foo\.\d+\] line 14 Check that \*x is assignable: FAILURE$
+^\[foo\.\d+\] line 15 Check that \*y is assignable: SUCCESS$
+^\[foo\.\d+\] line 18 Check that \*x is assignable: FAILURE$
+^\[foo\.\d+\] line 19 Check that \*y is assignable: FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+Check that conditional assigns clause  `c ? lvalue`
```
feliperodri: We don't use this pattern anymore.

regression/contracts/assigns_enforce_conditional_lvalue_list/main.c
```diff
@@ -2,12 +2,21 @@
 #include <stdbool.h>
 #include <stdlib.h>
```
feliperodri: We don't need `stdlib.h` here.

regression/contracts/assigns_enforce_conditional_lvalue_list/test.desc
```diff
@@ -0,0 +1,17 @@
...
+^\[foo\.\d+\] line 14 Check that \*x is assignable: FAILURE$
+^\[foo\.\d+\] line 15 Check that \*y is assignable: FAILURE$
+^\[foo\.\d+\] line 18 Check that \*x is assignable: FAILURE$
+^\[foo\.\d+\] line 19 Check that \*y is assignable: FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+Check that conditional assigns clause  `c ? {lvalue, ..}`
```
feliperodri, line 16: We don't use this pattern anymore.

regression/contracts/assigns_enforce_conditional_non_lvalue_target/main.c
```diff
@@ -0,0 +1,23 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>`.

regression/contracts/assigns_enforce_conditional_non_lvalue_target/main.c
```diff
@@ -0,0 +1,23 @@
...
+
+int *identity(int *ptr)
+{
+  return ptr;
+}
+
+int foo(bool a, int *x, int offset) __CPROVER_assigns(a : !x)
```
feliperodri, line 8: nit: it'd be nice to add a comment here like
```c
int foo(bool a, int *x, int offset) __CPROVER_assigns(a : !x) // !x Boolean not allowed as target, only lvalues.
```
or something similar.

regression/contracts/assigns_enforce_conditional_non_lvalue_target_list/main.c
```diff
@@ -0,0 +1,23 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_pointer_object/main.c
```diff
@@ -0,0 +1,37 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_pointer_object/test.desc
```diff
@@ -0,0 +1,17 @@
...
+^\[foo\.\d+\] line 20 Check that x\[\(signed (long )?long int\)0\] is assignable: FAILURE$
+^\[foo\.\d+\] line 21 Check that y\[\(signed (long )?long int\)0\] is assignable: SUCCESS$
+^\[foo\.\d+\] line 24 Check that x\[\(signed (long )?long int\)0\] is assignable: FAILURE$
+^\[foo\.\d+\] line 25 Check that y\[\(signed (long )?long int\)0\] is assignable: FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+Check that conditional assigns clause  `c ? __CPROVER_POINTER_OBJECT((p)`
```
feliperodri, line 16: We don't use this pattern anymore.

regression/contracts/assigns_enforce_conditional_pointer_object_list/main.c
```diff
@@ -0,0 +1,36 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_pointer_object_list/test.desc
```diff
@@ -0,0 +1,17 @@
...
+^\[foo\.\d+\] line 19 Check that x\[\(signed (long )?long int\)0\] is assignable: FAILURE$
+^\[foo\.\d+\] line 20 Check that y\[\(signed (long )?long int\)0\] is assignable: FAILURE$
+^\[foo\.\d+\] line 23 Check that x\[\(signed (long )?long int\)0\] is assignable: FAILURE$
+^\[foo\.\d+\] line 24 Check that y\[\(signed (long )?long int\)0\] is assignable: FAILURE$
+^VERIFICATION FAILED$
+^EXIT=10$
+^SIGNAL=0$
+--
+--
+Check that conditional assigns clause  `c ? {__CPROVER_POINTER_OBJECT((p), .. }`
```
feliperodri, line 16: We don't use this pattern anymore.

regression/contracts/assigns_enforce_conditional_side_effect_condition/main.c
```diff
@@ -2,16 +2,12 @@
 #include <stdbool.h>
 #include <stdlib.h>
```
feliperodri: We don't need these includes anymore.

regression/contracts/assigns_enforce_conditional_side_effect_target/main.c
```diff
@@ -0,0 +1,18 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_side_effect_target_list/main.c
```diff
@@ -0,0 +1,18 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_ternary_target/main.c
```diff
@@ -0,0 +1,18 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_ternary_target_list/main.c
```diff
@@ -0,0 +1,18 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_unions/main.c
```diff
@@ -0,0 +1,118 @@
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_void_target/main.c
```diff
@@ -0,0 +1,17 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_enforce_conditional_void_target_list/main.c
```diff
@@ -0,0 +1,18 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

regression/contracts/assigns_replace_conditional_targets/main.c
```diff
@@ -0,0 +1,76 @@
+#include <assert.h>
+#include <stdbool.h>
+#include <stdlib.h>
```
feliperodri: We only need `#include <stdbool.h>` here.

src/ansi-c/c_typecheck_base.h
feliperodri: +1

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +844,167 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+  {
+    must_throw = true;
+    error().source_location = condition.source_location();
+    error() << "ternary expressions not allowed in assigns "
+               "clause conditions"
+            << eom;
+  }
+
+  if(must_throw)
+    throw 0;
```
feliperodri, line 900: Why don't you `throw 0` on each if branch instead? You're now evaluating every if condition even if there is a previous error. Is that intentional?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +844,167 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+}
+
+void c_typecheck_baset::typecheck_spec_assigns_target(exprt &target)
+{
+  // compute type
+  typecheck_expr(target);
+
+  // fatal errors
+  bool must_throw = false;
+  if(target.type().id() == ID_empty)
```
feliperodri, line 910: **Same as above.**

Why don't you `throw 0` on each if branch instead? You're now evaluating every if condition even if there is a previous error. Is that intentional?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +844,167 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+  }
+
+  if(must_throw)
+    throw 0;
+}
+
 void c_typecheck_baset::typecheck_spec_assigns(exprt::operandst &targets)
 {
+  exprt::operandst tmp;
+  bool must_throw = false;
```
feliperodri, line 952: **Same as above.**

Why don't you `throw 0` on each if branch instead? You're now evaluating every if condition even if there is a previous error. Is that intentional?

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +844,167 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+            << eom;
+  }
+
+  if(must_throw)
+    throw 0;
+}
+
 void c_typecheck_baset::typecheck_spec_assigns(exprt::operandst &targets)
 {
+  exprt::operandst tmp;
```
feliperodri, line 951: Are those temporary targets? I'd give a more descriptive name for this temporary variable.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -842,35 +844,167 @@ void c_typecheck_baset::typecheck_dowhile(code_dowhilet &code)
...
+        conditional_target_exprt result{condition, ops};
+        result.add_source_location() = conditional_target.source_location();
+        tmp.push_back(std::move(result));
+      }
     }
   }
+
+  // at least one target was not as expected
+  if(must_throw)
+    throw 0;
```
feliperodri, line 1002: So why not `throw 0` as soon as we find this target?

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -52,6 +54,39 @@ static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
...
+/// ```
+static const guarded_slicet
+normalize_to_guarded_slice(const exprt &expr, const namespacet &ns)
+{
+  if(can_cast_expr<conditional_target_exprt>(expr))
+  {
+    const auto &conditional_target = to_conditional_target_expr(expr);
+    INVARIANT(
+      conditional_target.targets().size() == 1,
+      "expecting only a single target at this stage");
```
feliperodri: What stage? Could we be more descriptive?

src/goto-instrument/contracts/assigns.h
```diff
@@ -22,6 +22,27 @@ Date: July 2021
 
 typedef std::pair<const exprt, const exprt> slicet;
 
+/// A guarded slice of memory
```
feliperodri, line 25: `\brief`? Why struct and not a class here?

feliperodri (approved):
Only minor comments. Thanks for implementing this new feature and the improved doc/descriptions!

src/ansi-c/c_typecheck_code.cpp
remi-delmas-3000: fixed

src/ansi-c/c_typecheck_code.cpp
remi-delmas-3000: Since neiter `implicit_typecast_bool`  or `implicit_typecast` have documentation I wasn't sure what to expect.

src/ansi-c/c_expr.h
remi-delmas-3000: fixed

src/ansi-c/c_expr.h
tautschnig: Ok! (Adding/subtracting pointers is an existing example of `binary_exprt` with non-matching types.)

src/ansi-c/c_typecheck_code.cpp
tautschnig: Fair enough! I was actually hoping for your test cases to confirm whether it is/isn't caught...

src/goto-instrument/contracts/assigns.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: fixed

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: fixed

src/ansi-c/c_typecheck_code.cpp
remi-delmas-3000: very much intentional, in order to report all errors triggered by the expression instead of reporting them one by one.

src/ansi-c/c_typecheck_code.cpp
remi-delmas-3000: in order to report all errors found in the target list instead of stopping on the first one.

src/goto-instrument/contracts/assigns.h
remi-delmas-3000: there are no methods associated with this. Why would it need to be a class ?

PR #6547, opened 2021-12-26
Fix linking in case of function type conflict

tautschnig:
We already almost did the right thing, but were failing to update types at call sites. This PR also adds improved conflict reporting and an actual test of such type conflicts.


chris-ryder (approved):
Looks sensible, and thanks for the extra tests!

kroening (approved):

PR #6549, opened 2021-12-27
move solver_hardness.* from solvers/ to goto-symex/

kroening:
The `solver_hardnesst` class is inherently linked to goto-programs and goto-symex, and
introduces a surprising dependency on these corresponding includes into the
solver classes.

This commit moves the class into the goto-symex directory, removing the
dependency by introducing a base class for the callback.


src/goto-symex/solver_hardness.h
```diff
@@ -160,4 +161,26 @@ struct hash<solver_hardnesst::hardness_ssa_keyt>
...
+    if(auto hardness_collector = prop_conv_solver->get_hardness_collector())
+    {
+      if(hardness_collector->solver_hardness)
+      {
+        auto &solver_hardness = static_cast<solver_hardnesst &>(
+          *(hardness_collector->solver_hardness));
+        handler(solver_hardness);
+      }
+    }
+  }
```
tautschnig, line 183: This surely is better already, but don't we just need a Boolean function like `virtual bool set_hardness_collector(solver_hardnesst *) { return false; }` in `prop_conv_solvert` that will return `true` when the hardness collector was accepted?

tautschnig (approved):

src/goto-symex/solver_hardness.h
kroening: To make the above go away, you'd need a get_solver_hardness_collector, or the like, in `decision_proceduret`. I can't get myself to do that.

peterschrammel (approved):

PR #6551, opened 2021-12-27
rename post_process to finish_eager_conversion

kroening:
This renames a set of solver methods from post_process to
finish_eager_conversion.  The term 'post processing' may be used either in
lazy or eager solving.  The new name clarifies that this is about an eager
conversion.

### DIFF
--- a/src/solvers/prop/prop_conv_solver.h
+++ b/src/solvers/prop/prop_conv_solver.h
@@ -35,6 +35,9 @@ public:
 
   virtual ~prop_conv_solvert() = default;
 
+  // non-iterative eager conversion
+  virtual void finish_eager_conversion();
+
   // overloading from decision_proceduret
   decision_proceduret::resultt dec_solve() override;
   void print_assignment(std::ostream &out) const override;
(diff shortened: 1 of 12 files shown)


src/solvers/prop/prop_conv_solver.h
```diff
@@ -35,6 +35,9 @@ class prop_conv_solvert : public conflict_providert,
 
   virtual ~prop_conv_solvert() = default;
 
+  // non-iterative eager conversion
+  virtual void finish_eager_conversion();
```
tautschnig, line 39: Just curious: this used to be `protected`. Why is making it `public` the right thing to do?

tautschnig (approved):

src/solvers/prop/prop_conv_solver.h
kroening: Yes, I am thinking about the use of this for an eager conversion, say to Dimacs, or something else.
Right now, we trigger the "finishing" by calling `dec_solve`, which then doesn't actually do solving, but returns an error, which we then ignore. It would be clearer to feed the constraints to the decision procedure, and then call `finish_eager_conversion`, and then be done.

peterschrammel (approved):

PR #6552, opened 2021-12-27
introduce index_type and element_type for arrays and vectors

kroening:
This commit introduces convenience methods for retrieving the type of the
elements and of the indices for array and vector types.  This avoids using a
global, architecture-dependent function to generate the index type.

### DIFF
--- a/src/goto-instrument/goto_program2code.cpp
+++ b/src/goto-instrument/goto_program2code.cpp
@@ -405,9 +405,9 @@ void goto_program2codet::convert_assign_rec(
     forall_operands(it, assign.rhs())
     {
       index_exprt index(
-          assign.lhs(),
-          from_integer(i++, index_type()),
-          type.subtype());
+        assign.lhs(),
+        from_integer(i++, type.index_type()),
+        type.element_type());
       convert_assign_rec(code_assignt(index, *it), dest);
     }
   }
--- a/src/goto-symex/field_sensitivity.cpp
+++ b/src/goto-symex/field_sensitivity.cpp
@@ -9,7 +9,6 @@ Author: Michael Tautschnig
 #include "field_sensitivity.h"
 
 #include <util/arith_tools.h>
-#include <util/c_types.h>
 #include <util/simplify_expr.h>
 #include <util/std_expr.h>
 
@@ -204,7 +203,7 @@ exprt field_sensitivityt::get_fields(
 
     for(std::size_t i = 0; i < array_size; ++i)
     {
-      const index_exprt index(array, from_integer(i, index_type()));
+      const index_exprt index(array, from_integer(i, type.index_type()));
       ssa_exprt tmp = ssa_expr;
       bool was_l2 = !tmp.get_level_2().empty();
       tmp.remove_level_2();
@@ -318,7 +317,7 @@ void field_sensitivityt::field_assignments_rec(
     exprt::operandst::const_iterator fs_it = lhs_fs.operands().begin();
     for(std::size_t i = 0; i < array_size; ++i)
     {
-      const index_exprt index_rhs(lhs, from_integer(i, index_type()));
+      const index_exprt index_rhs(lhs, from_integer(i, type->index_type()));
       const exprt &index_lhs = *fs_it;
 
       field_assignments_rec(
--- a/src/goto-symex/symex_builtin_functions.cpp
+++ b/src/goto-symex/symex_builtin_functions.cpp
@@ -197,8 +197,8 @@ void goto_symext::symex_allocate(
   {
     const auto &array_type = to_array_type(*object_type);
     index_exprt index_expr(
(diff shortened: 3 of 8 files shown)


src/goto-instrument/goto_program2code.cpp
```diff
@@ -405,9 +405,7 @@ void goto_program2codet::convert_assign_rec(
     forall_operands(it, assign.rhs())
     {
       index_exprt index(
-          assign.lhs(),
-          from_integer(i++, index_type()),
-          type.subtype());
+        assign.lhs(), from_integer(i++, type.index_type()), type.subtype());
```
tautschnig: Use `.element_type()` instead of `.subtype()`

src/solvers/flattening/boolbv_get.cpp
```diff
@@ -156,18 +159,20 @@ exprt boolbvt::bv_get_rec(
     }
     else if(type.id()==ID_vector)
     {
-      const typet &subtype = type.subtype();
+      const auto &vector_type = to_vector_type(type);
+      const typet &subtype = vector_type.subtype();
```
tautschnig: Use `.element_type()`

src/util/std_types.h
```diff
@@ -768,6 +768,15 @@ class array_typet:public type_with_subtypet
     add(ID_size, std::move(_size));
   }
 
+  /// The type of the index expressions into any instance of this type.
+  typet index_type() const;
+
+  /// The type of the elements of the array.
+  const typet &element_type() const
+  {
+    return subtype();
```
tautschnig, line 779: Should we actually hide `subtype()` for `array_typet` and `vector_typet`?

src/util/std_types.h
```diff
@@ -768,6 +768,15 @@ class array_typet:public type_with_subtypet
     add(ID_size, std::move(_size));
   }
 
+  /// The type of the index expressions into any instance of this type.
+  typet index_type() const;
```
tautschnig, line 772: I am concerned that this may be confusing: we could end up in a situation where `array.index().type() != array.index_type()`!

tautschnig (commented):
Shouldn't we rather strive towards a situation where an array index can be of an arbitrary numeric type?

kroening:
> Shouldn't we rather strive towards a situation where an array index can be of an arbitrary numeric type?

I did consider it, but I don't think it's a good idea.
1. It would be problematic to use the SMT-LIB array theory, which is basically a map type.
2. Expressions that one would expect to be identical are not. Say a[1] and a[1l] would be different.

src/util/std_types.h
kroening: I'd like to enforce that, going forward, using a precondition.

src/util/std_types.h
kroening: Yes, that will improve readability.

tautschnig:
> > Shouldn't we rather strive towards a situation where an array index can be of an arbitrary numeric type?
> 
> I did consider it, but I don't think it's a good idea.
> 
>     1. It would be problematic to use the SMT-LIB array theory, which is basically a map type.
> 
>     2. Expressions that one would expect to be identical are not. Say a[1] and a[1l] would be different.

Will it always be possible to have the same index type, even when linking together objects from different source languages, for example C and Java?

kroening:
> Will it always be possible to have the same index type, even when linking together objects from different source languages, for example C and Java?

The plan is to make the index type part of the array type; you can then mix as you see fit.

src/goto-symex/field_sensitivity.cpp
```diff
@@ -204,7 +204,7 @@ exprt field_sensitivityt::get_fields(
 
     for(std::size_t i = 0; i < array_size; ++i)
     {
-      const index_exprt index(array, from_integer(i, index_type()));
+      const index_exprt index(array, from_integer(i, type.index_type()));
```
tautschnig, line 206: Please remove `#include <util/c_types.h>` from this file.

src/goto-symex/symex_builtin_functions.cpp
```diff
@@ -197,7 +197,7 @@ void goto_symext::symex_allocate(
   {
     const auto &array_type = to_array_type(*object_type);
     index_exprt index_expr(
-      value_symbol.symbol_expr(), from_integer(0, index_type()));
+      value_symbol.symbol_expr(), from_integer(0, array_type.index_type()));
```
tautschnig, line 200: There are further uses of `index_type()` in this file, which should also be removed.

tautschnig (approved):
My concerns/comments that I've left unresolved remain in place, but approving nonetheless.

kroening:
> My concerns/comments that I've left unresolved remain in place, but approving nonetheless.

Both of these won't be easy -- look at the branch typet_subtype

peterschrammel (approved):

PR #6560, opened 2022-01-06
Value sets should permit any order of operands in pointer arithmetic

tautschnig:
Previously, only \<integer-constant\> +/- pointer or
pointer-as-first-operand worked correctly. In all other cases, value
sets did not recurse on the correct operand.

### DIFF
--- a/src/pointer-analysis/value_set.cpp
+++ b/src/pointer-analysis/value_set.cpp
@@ -633,30 +633,45 @@ void value_sett::get_value_set_rec(
     object_mapt pointer_expr_set;
     optionalt<mp_integer> i;
 
-    // special case for pointer+integer
-
+    // special case for plus/minus and exactly one pointer
+    optionalt<exprt> ptr_operand;
     if(
-      expr.operands().size() == 2 && expr_type.id() == ID_pointer &&
+      expr_type.id() == ID_pointer &&
       (expr.id() == ID_plus || expr.id() == ID_minus))
     {
-      exprt ptr_operand;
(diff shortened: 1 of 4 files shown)


src/pointer-analysis/value_set.cpp
```diff
@@ -674,9 +689,12 @@ void value_sett::get_value_set_rec(
             i->negate();
```
TGWDB, line 693: This always negates the constant value `i`, but now we allow `i - ptr` and so the negation should respect the order of the operands. Also note that since the code has changed from explicitly checking for binary operands to n-ary, it is (in theory - I don't know that we construct n-ary pointer +/-) not sufficient to simply gather all the constant values, since we could have `i1 - ptr - i2`.

TGWDB (changes requested):

src/pointer-analysis/value_set.cpp
tautschnig: Thank you for calling out these scenarios! None of them should be possible (a `minus_exprt` is a binary expression, and subtraction of a pointer requires should only be possible when the left-hand side also is a pointer), so I've added a `DATA_INVARIANT` to make sure we catch all those cases, should they unexpectedly occur.

TGWDB (approved):
Thanks for addressing the potential issues, I agree they're probably extremely unlikely to occur, but since we removed a sanity check it's good to have something in it's place.

peterschrammel (approved):

PR #6564, opened 2022-01-11
Function contracts: check target validity, simplify snapshot instrumentation

remi-delmas-3000:
In assigns clause checking, for better usability and performance with conditional targets:

- Check that user-specified targets point to non-invalid locations when their condition holds
- Replace the `all_dereferences_valid` with a simpler `not_null` check in the validity condition, remove conditional jumps
- Rewrite inclusion checks to tolerate NULL targets only when necessary (calls to 'free')

Benefits:
- Detect and report errors user-specified target conditions
- Up to x1.5-x2 faster runs on large benchmarks (from ~1800s to ~900s on large `s2n-tls` benchmarks)

# Details

## Problem

We now support conditional assigns targets of the form `condition: target`.

The `condition` lets the user specify when the target expression is assignable by the function.

We also need to make sure the target expression denotes a valid memory location when the condition holds. 

Today this is achieved using this instrumentation code:

```
bool __car_valid := condition & all_dereferences_valid(target) & w_ok(target_start_address, target_size);
char *__car_lb := NULL;
char *__car_ub := NULL;
if(__car_valid)
{
  __car_lb := start_address_of(target);
  __car_ub := __car_lb + target_size(target); // this instruction is instrumented with pointer overflow checks
}
```

Inclusion checks are performed using:

```
ASSERT __car_valid ==> is_included(car, set-of-legit-locations)
```
 
With this encoding, if the target expression happens to be invalid when the user-provided condition is true, we silently compute an empty conditional address range and we will reject assignments to this invalid target location (as expected), 

The results can however be hard to interpret from the user's perspective since he does not know that the target is invalid when he specified it should be assignable. It is most likely the sign of an involuntary error in the target condition

Moreover, the use of `all_dereferences_valid` and `w_ok` predicates creates large formulas and causes long runtimes.

## Proposed Fix

We distinguish the following uses for an address range snaphsot  `(__car_valid, __car_lb, __car_ub)`:

1. Snapshotting a user-specified conditional target found in an assigns clause. In that case:
  a. when the condition holds, the `target_start_address` pointer shall be either `NULL` or writable with the expected size. We allow `NULL` because it can be passed as a valid argument to functions such as `free()`, and the user code might attempt to do this.
  b. we forbid invalid `target_start_address` pointers because they are not accepted in any meaningful operation.
2. Snapshotting the address range of a pointer passed to `free()`. The pointer must either be a `NULL` pointer or a pointer to the start address of a valid object (with offset 0, but we will not check that since `free` checks it itself).
3. Snapshotting the LHS of an assignment or an argument to a `havoc` function, in order to perform an inclusion check against a set of allowed target locations (themselves derived from cases 1 and 4). In that case:
  a. we want to check that the `target_start_address` is neither `NULL` nor `INVALID` when its condition holds.
4. Snapshotting a `DECL` or a pointer returned by `malloc`. Conditions 1.a. and 1.b. would apply here, but they are always satisfied by construction (a stack allocated var is always backed by an object of the right size for its type, and CBMC's `malloc` either returns a `NULL` pointer or a pointer backed by an object of the right size).

So, in order to check for target validity in cases 1. and 2. (user specified targets and parameters to free), we use:

```
ASSERT condition ==> (target_start_address == NULL || w_ok(target_start_address, target_size)) ;
```

To check for target validity in case 3. (assignment LHS), we use:

```
ASSERT condition ==> w_ok(target_start_address, target_size);
```

These assertions will be falsified if `target_start_address` is an invalid pointer.

As explained above, in case 4 the target is valid by construction, so we do not generate an assertion.

In cases 1, 2, 3 and 4 the snapshot is rid of conditional jumps and becomes:

```
bool __car_valid := condition & not_null(target_start_address);
__car_lb := target_start_address;
__car_ub := __car_lb + target_size(target);
```

Assuming the "target validity" assertion was proved, if `condition & not_null(target_start_address)` holds then `w_ok(target_start_address, target_size)` holds and the computed `__car_lb` and `__car_ub` are meaningful.

If the target happens to be `NULL`,  `__car_valid` will be false and `__car_lb` and `__car_ub` will contain garbage values but will never be used in inclusion checks, which are guarded by `__car_valid`.

If the target happens to be invalid, then all snapshot variables can take arbitrary and incoherent values which may or may not fail the inclusion checks.  In all cases, since the target validity assertions will surely fail, the assigns clause checking analysis as a whole will not succeed.

## Inclusion checks

The inclusion check for assignments and havoc parameters is written as:

```
ASSERT __car_valid & is_included(car, set-of-legit-locations);
```

An for parameters to `free` we allow NULL as a special case:

```
ASSERT target_start_address!=NULL ==> __car_valid & is_included(car, set-of-legit-locations);
```

## Summary

With this new encoding, 

- we can detect bad conditions which allow invalid targets in assigns clause specifications, 
- we remove conditional jumps and extra pointer assertions

For a full assigns clause check to be considered successful:
- the user specified targets must all pass the target validity checks
- all assignments and arguments to `havoc` or `free` must pass the inclusion check

-------

### DIFF
--- a/regression/contracts/assigns_enforce_offsets_4/test.desc
+++ b/regression/contracts/assigns_enforce_offsets_4/test.desc
@@ -1,9 +1,13 @@
-KNOWNBUG
+CORE
 main.c
---enforce-contract foo
+--enforce-contract foo _ --pointer-check
+^\[.*\d+\].* line 8 Check assigns target validity 'TRUE: x\[\(.*\)10\]': FAILURE$
 ^EXIT=10$
 ^SIGNAL=0$
 ^VERIFICATION FAILED$
 --
 --
-Check that a write at *x fails when the assigns clause specifies *(x + 1) and the actual underlying object is of size 1.
+Check that a write at *(x+10) fails when the assigns clause specifies *(x + 10)
+and the actual underlying object is too small.
+In that case the target inclusion succeeds because the LHS is in an invalid
+state, but the target validity check must fail.
--- a/regression/contracts/assigns_enforce_structs_07/test.desc
+++ b/regression/contracts/assigns_enforce_structs_07/test.desc
@@ -3,13 +3,14 @@ main.c
 --enforce-contract f1 --enforce-contract f2 _ --malloc-may-fail --malloc-fail-null --pointer-check
 ^EXIT=10$
 ^SIGNAL=0$
-^\[f1.\d+\] line \d+ Check that p->buf\[\(.*\)0\] is assignable: SUCCESS$
-^\[f1.pointer\_dereference.\d+\] line \d+ dereference failure: pointer invalid in p->buf\[\(.*\)0\]: FAILURE$
-^\[f1.pointer\_dereference.\d+\] line \d+ dereference failure: pointer NULL in p->buf: FAILURE$
-^\[f2.\d+\] line \d+ Check that pp->p->buf\[\(.*\)0\] is assignable: SUCCESS$
-^\[f2.pointer\_dereference.\d+\] line \d+ dereference failure: pointer invalid in pp->p->buf\[\(.*\)0\]: FAILURE$
+^\[f1.*\d+\].*line 18 Check assigns target validity 'TRUE: p->buf\[\(.*\)0\]': FAILURE$
+^\[f2.*\d+\].*line 23 Check assigns target validity 'TRUE: pp->p->buf\[\(.*\)0\]': FAILURE$
 ^VERIFICATION FAILED$
 --
 --
-Checks whether CBMC properly evaluates write set of members
-from invalid objects. In this case, they are not writable.
+In f1, the assigns clause specifies `*(p->buf)` as target (which can be invalid)
+and assigns `p->buf[0]` unconditionally. When both target and lhs are invalid,
+its inclusion check can be trivially satisfied (or not) but we expect the target
+validity check to fail.
+
+In f2 tests this behaviour with chained dereferences.
--- a/regression/contracts/assigns_replace_05/test.desc
+++ b/regression/contracts/assigns_replace_05/test.desc
@@ -1,12 +1,16 @@
-KNOWNBUG
+CORE
 main.c
 --replace-call-with-contract f2 --replace-call-with-contract f3
-^EXIT=10$
+main.c function main
+^\[.*\d+\] line 29 assertion p < 0: SUCCESS$
+^\[.*\d+\] line 30 assertion q == 32: SUCCESS$
+^\[.*\d+\] line 31 reachability test: SUCCESS$
+^VERIFICATION SUCCESSFUL$
+^EXIT=0$
 ^SIGNAL=0$
-^VERIFICATION FAILED$
 --
 --
-This test checks that verification fails when assigns clauses are combined with function contracts
-in a loop improperly, i.e., always assumes memory not mention in ensures clauses are unchanged.
-
-BUG: Currently, function call replacement using 'ensures' specifications encodes an implicit assumption that any memory not mentioned in the ensures clause remains unchanged throughout the function, even when an 'assigns' clause is not present.
+This test demonstrates that replacing a function call with a contract that has
+an empty assigns clause and a post condition involving its input parameters can 
+causes vacuous proofs. Checking the contract against the function would fail
+the assigns clause checks. *This is not a bug*.
--- a/src/goto-instrument/contracts/assigns.cpp
+++ b/src/goto-instrument/contracts/assigns.cpp
@@ -25,6 +25,7 @@ Date: July 2021
 #include <util/expr_util.h>
 #include <util/pointer_offset_size.h>
 #include <util/pointer_predicates.h>
+#include <util/simplify_expr.h>
 
 static const slicet normalize_to_slice(const exprt &expr, const namespacet &ns)
 {
@@ -126,14 +127,10 @@ assigns_clauset::conditional_address_ranget::conditional_address_ranget(
 }
 
 goto_programt
-assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
-  const
+assigns_clauset::conditional_address_ranget::generate_snapshot_instructions(
+  check_target_validityt check_target_validity) const
 {
   goto_programt instructions;
-  // adding pragmas to the location to selectively disable checks
-  // where it is sound to do so
-  source_locationt location_no_checks = location;
-  disable_pointer_checks(location_no_checks);
 
   // clean up side effects from the guard expression if needed
   cleanert cleaner(parent.symbol_table, parent.log.get_message_handler());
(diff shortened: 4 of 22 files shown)


regression/contracts/assigns_enforce_offsets_4/test.desc
```diff
@@ -1,9 +1,17 @@
...
 ^VERIFICATION FAILED$
 --
 --
-Check that a write at *x fails when the assigns clause specifies *(x + 1) and the actual underlying object is of size 1.
+Check that a write at *(x+10) fails when the assigns clause specifies *(x + 10)
+and the actual underlying object is too small.
+In that case the target inclusion succeeds because the LHS is in an invalid
+state, but the target validity check and the pointer checks for the assignment
+must fail.
+
```
tautschnig: Nit pick: lots of whitespace, but no newline at end-of-file.

regression/contracts/assigns_enforce_structs_07/test.desc
```diff
@@ -3,13 +3,18 @@ main.c
 --enforce-contract f1 --enforce-contract f2 _ --malloc-may-fail --malloc-fail-null --pointer-check
 ^EXIT=10$
 ^SIGNAL=0$
-^\[f1.\d+\] line \d+ Check that p->buf\[\(.*\)0\] is assignable: SUCCESS$
+^\[f1.\d+\] line \d+ Check that p->buf\[\(.*\)0\] is assignable: (SUCCESS|FAILURE)$
```
tautschnig: So we don't care what the verification result is? Would you actually want to see both of them? If so, two lines will be required.

regression/contracts/assigns_enforce_structs_07/test.desc
```diff
@@ -3,13 +3,18 @@ main.c
...
 ^\[f1.pointer\_dereference.\d+\] line \d+ dereference failure: pointer invalid in p->buf\[\(.*\)0\]: FAILURE$
 ^\[f1.pointer\_dereference.\d+\] line \d+ dereference failure: pointer NULL in p->buf: FAILURE$
-^\[f2.\d+\] line \d+ Check that pp->p->buf\[\(.*\)0\] is assignable: SUCCESS$
+^\[f2.\d+\] line \d+ Check that pp->p->buf\[\(.*\)0\] is assignable: (SUCCESS|FAILURE)$
```
tautschnig: As above: we don't care?

regression/contracts/assigns_replace_05/test.desc
```diff
@@ -1,12 +1,15 @@
...
-^VERIFICATION FAILED$
 --
 --
-This test checks that verification fails when assigns clauses are combined with function contracts
-in a loop improperly, i.e., always assumes memory not mention in ensures clauses are unchanged.
-
-BUG: Currently, function call replacement using 'ensures' specifications encodes an implicit assumption that any memory not mentioned in the ensures clause remains unchanged throughout the function, even when an 'assigns' clause is not present.
+This test checks that replacing a function call with a contract that has an 
+empty assigns clause and a post condition involving its input parameters can 
+causes vacuous proofs.
```
tautschnig: Missing newline at end-of-file.

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -168,10 +168,35 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
...
+    // We check that when guard condition is true,
+    // the target expression is either null or w_ok with the expected size.
+    // This assertion will be falsified whenever guarded_slice.start_adress
+    // is invalid.
+    auto target_validity_assertion =
+      instructions.add(goto_programt::make_assertion(
+        or_exprt{not_exprt{clean_guard},
+                 null_pointer(guarded_slice.start_adress),
+                 w_ok_exprt{guarded_slice.start_adress, guarded_slice.size}},
+        location_no_checks));
```
tautschnig: May I suggest that you wrap this `or_exprt` in a call to `simplify_expr(..., parent.ns)`?

tautschnig (approved):

regression/contracts/assigns_enforce_offsets_4/test.desc
remi-delmas-3000: fixed

regression/contracts/assigns_enforce_structs_07/test.desc
remi-delmas-3000: The inclusion check is meaningless if the target validity assertion does not hold, so it can have either status. What we really want to see is the target validity at line 8 fail, which was missing, thanks!

regression/contracts/assigns_enforce_structs_07/test.desc
remi-delmas-3000: same as above, the inclusion check is meaningless if the target validity assertion does not hold, so it can have either status, and we were missing a check on the target validity status. This is now fixed

src/goto-instrument/contracts/assigns.cpp

regression/contracts/assigns_replace_05/test.desc
remi-delmas-3000: fixed

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -985,7 +987,9 @@ void code_contractst::instrument_call_statement(
       // move past the call and then insert the CAR
       instruction_it++;
       const auto car = assigns.add_to_write_set(pointer_object(object));
-      auto snapshot_instructions = car->generate_snapshot_instructions();
+      auto snapshot_instructions = car->generate_snapshot_instructions(
+        // no need to check because malloc always returns a null or valid pointer
```
tautschnig: cppcheck says this line is too long

src/goto-instrument/contracts/assigns.cpp
```diff
@@ -148,54 +145,66 @@ assigns_clauset::conditional_address_ranget::generate_snapshot_instructions()
...
-  // of the pointer with very large allocation sizes.
-  // We enable pointer overflow checks to detect such cases.
-  source_locationt location_overflow_check = location;
-  location_overflow_check.add_pragma("enable:pointer-overflow-check");
+  instructions.add(
+    goto_programt::make_decl(upper_bound_address_var, location_no_checks));
 
   instructions.add(goto_programt::make_assignment(
     upper_bound_address_var,
     plus_exprt{guarded_slice.start_adress, guarded_slice.size},
```
tautschnig: Use `simplify_expr(plus_exprt{...}, parent.ns)` to get rid of some type casts that those expressions usually have.

remi-delmas-3000: done

src/goto-instrument/contracts/contracts.cpp

src/goto-instrument/contracts/assigns.cpp
remi-delmas-3000: done

regression/contracts/assigns_replace_05/test.desc
remi-delmas-3000: fixed

PR #6575, opened 2022-01-13
CONTRACTS: New class `instrument_spec_assignst` encapsulating assigns clause checking logic.

remi-delmas-3000:
The previous `assigns_clauset` class leaked its internal representations to the outside world and important control logic was spread between `assigns.cpp` and `contracts.cpp`.

New classes:
- New class `conditional_target_exprt` deriving `exprt`: represents a single conditional target from the source program
- New class`conditional_address_range_exprt` deriving `exprt`: represents a CAR derived from a conditional target 
- New class `instrument_spec_assignst` encapsulating assigns clause checking algos:
  - 3 methods to track targets by origin (from an assigns clause , from a DECL, from a malloc)
  - 1 method to detect local static variables used directly or indirectly by the function
  - 1 method to check assignments
  - 1 method to invalidate DEAD objects
  - 1 method to check freed objects and invalidate aliased CARs
- The generated assertions now have property class "assigns" for  better console output

Changes:
- The function `havoc_assigns_clause_targets` is now a class `havoc_assigns_clause_targetst` that inherits from `instrument_spec_assignst` and reuses all the snapshotting and target validity checking conditions
- Removed code from contracts.cpp that now lives in `instrument_spec_assignst`
- Function and loop contracts instrumentations now use the new classes

No performance changes expected.


src/goto-instrument/contracts/contracts.cpp
```diff
@@ -255,6 +251,18 @@ void code_contractst::check_apply_loop_contracts(
...
+  instrument_spec_assignst instrument_spec_assigns(
+    function_name, goto_functions, ns, symbol_table, log);
+
+  // assigns clause snapshots
+  goto_programt snapshot_instructions =
+    instrument_spec_assigns.track_static_locals();
+
+  // ^^^ FIXME ^^^ we should only allow assignments to static locals
+  // declared in functions that are called in the loop body, not from the whole
+  // function...
```
tautschnig, line 264: Are we tracking this task somewhere?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -274,7 +282,8 @@ void code_contractst::check_apply_loop_contracts(
           log.debug() << ", ";
         ran_once = true;
         log.debug() << format(target);
-        loop_assigns.add_to_write_set(target);
+        auto payload = instrument_spec_assigns.track_spec_target(target);
+        snapshot_instructions.destructive_append(payload);
```
tautschnig: Would it perhaps make for a better interface if `snapshot_instructions` were passed as an argument?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -293,22 +302,19 @@ void code_contractst::check_apply_loop_contracts(
...
+  // Insert instrumentation
   // This must be done before havocing the write set.
-  for(const auto &car : loop_assigns.get_write_set())
-  {
-    auto snapshot_instructions = car.generate_snapshot_instructions(
-      assigns_clauset::check_target_validityt::YES_ALLOW_NULL);
-    insert_before_swap_and_advance(
-      goto_function.body, loop_head, snapshot_instructions);
-  };
+  // ^^^ FIXME this is not true ^^^
```
tautschnig, line 317: Can we just fix this now?

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1425,10 +1340,11 @@ void code_contractst::check_frame_conditions(
   }
 }
 
+#if 0
```
tautschnig: At least a comment would be really helpful here.

src/goto-instrument/contracts/contracts.h
```diff
@@ -31,7 +31,7 @@ Date: February 2016
 #include <util/optional.h>
 #include <util/pointer_expr.h>
 
-#include "assigns.h"
+// #include "assigns.h"
```
tautschnig: Just remove it. It's very easy to re-add if needed...

src/goto-instrument/contracts/contracts.h
```diff
@@ -153,10 +154,11 @@ class code_contractst
     goto_programt &body,
     goto_programt::targett instruction_it,
     const goto_programt::targett &instruction_end,
-    assigns_clauset &assigns,
+    instrument_spec_assignst &instrument_spec_assigns,
     skipt skip_parameter_assigns,
     optionalt<cfg_infot> &cfg_info_opt);
 
+#if 0
```
tautschnig: Comment, please.

tautschnig (commented):
Just some preliminary comments below. Overall, this is really hard to review as it's hard to tell what is a change in behaviour (if any?) and what is just refactoring. It it truly were refactoring-only then surely no tests should change. Having this split up in several commits would have helped a lot.

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: this actually depends on wether
- a loop assigns clause is meant to represent the whole set of locations ever touched by the loop until completion, and hence 
  - no target expression should depend on other expressions also touched by the loop (i.e. we do not allow array slice expressions a[0..i]), 
  - it can be snapshotted before havoc+invariants

 or 
-  it should summarize an arbitrary execution prefix of the loop, in which case:
  - a target expressions could depend on each other like `(i, a[0..i])`
  - targets should be snapshotted after havocking and assuming the invariant.   
   - the whole clause should be proved to be included in itself post loop-step
   
  and we have not really decided on the semantics yet.

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: I refactored the API to use destination passing style, but since sometimes we have to append instructions, or to insert them before a target, or after a target, with or without swapping, we often end up having to declare a temporary `goto_programt` destination anyway.

src/goto-instrument/contracts/contracts.cpp
remi-delmas-3000: this was leftover from refactoring, now removed

src/goto-instrument/contracts/contracts.h
remi-delmas-3000: yes

src/goto-instrument/contracts/contracts.h
remi-delmas-3000: removed

src/goto-instrument/contracts/instrument_spec_assigns.h
```diff
@@ -0,0 +1,337 @@
...
+  // set of tracked conditional targets that come from assigns clauses specs
+  conditional_target_sett from_spec_assigns;
+
+  // set of tracked conditional targets that come from stack-allocated objects
+  conditional_target_sett from_stack_alloc;
+
+  // set of tracked conditional targets that come from heap-allocated objects
+  conditional_target_sett from_heap_alloc;
+
+  using car_mapt = std::map<const conditional_target_exprt, const car_exprt>;
```
SaswatPadhi: Why not:

```suggestion
  using car_mapt = std::unordered_map<const conditional_target_exprt, const car_exprt, irep_hash>;
```

src/goto-instrument/contracts/instrument_spec_assigns.h
```diff
@@ -0,0 +1,337 @@
...
+    optionalt<cfg_infot> &cfg_info_opt,
+    goto_programt &dest) const;
+
+  /// Generates instructions to invalidate all targets aliased with a
+  /// car that was passed to `free`, assuming the inclusion check passes,
+  /// ie. that the freed_car is fully included in one of the tracked targets.
+  void
+  invalidate_aliases(const car_exprt &freed_car, goto_programt &dest) const;
+
+  using conditional_target_sett = std::set<conditional_target_exprt>;
```
SaswatPadhi: Why not:

```suggestion
  using conditional_target_sett = std::unorederd_set<conditional_target_exprt, irep_hash>;
```

src/goto-instrument/contracts/utils.cpp
```diff
@@ -74,6 +59,35 @@ void havoc_if_validt::append_scalar_havoc_code_for_expr(
...
+  return instr;
+}
+
+goto_programt &add_pragma_disable_assigns_check(goto_programt &prog)
+{
+  Forall_goto_program_instructions(it, prog)
+    add_pragma_disable_assigns_check(*it);
+  return prog;
+}
+
```
SaswatPadhi, line 121: Am I missing something or have these two functions (this one + the one above) just been moved to a different location in the same file? If so, can we move those back to reduce the diff?

SaswatPadhi (commented):
Below are some minor initial comments based on a quick look at the changes.

I share the same concern as @tautschnig, that the change is huge (even ignoring the regression tests) and so is kind of hard to review. The new assigns clause code is MUCH larger than the previous assigns.cpp code and to me it seems quite complex (but I am also looking at it after a long time, and after the conditional assigns clause stuff was added).

My requests are to split the PR to multiple smaller commits if possible, and also add some description to the PR and/or the commit message as to why having these `exprt` subclasses is desirable.

src/goto-instrument/contracts/utils.cpp
remi-delmas-3000: This change was made so that the order of definitions in the cpp file follows the order of declarations in the header file, after moving `havoc_assigns_targett` from the now defunct `assigns` to `utils`.

src/ansi-c/c_typecheck_code.cpp
```diff
@@ -981,28 +981,21 @@ void c_typecheck_baset::typecheck_spec_assigns(exprt::operandst &targets)
...
   }
 
   // at least one target was not as expected
   if(must_throw)
     throw 0;
 
-  // now each target is either a simple side-effect-free lvalue expression
-  // or a conditional target expression with a single target
+  // now each target is either a simple side-effect-free unconditional lvalue
+  // expression conditional target group expression with a non-trivial condition
```
tautschnig: Missing "or?"

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+#include <langapi/language_util.h>
+
+#include "havoc_assigns_clause_targets.h"
+#include <util/arith_tools.h>
+#include <util/c_types.h>
+#include <util/expr_util.h>
+#include <util/pointer_offset_size.h>
+#include <util/pointer_predicates.h>
+#include <util/simplify_expr.h>
```
tautschnig: Nit pick for consistency with other places: use the following order
1) Include the header file corresponding to the implementation (as done here)
2) Includes from `util`
3) Includes from other directories
4) Includes from the current directory

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+///// PUBLIC METHODS /////
+
+void instrument_spec_assignst::track_spec_target(
+  const exprt &expr,
+  goto_programt &dest)
+{
+  if(can_cast_expr<conditional_target_group_exprt>(expr))
+    track_spec_target_group(to_conditional_target_group_expr(expr), dest);
```
tautschnig: How about
```
if(auto target = expr_try_dynamic_cast<conditional_target_group_exprt>(expr))
  track_spec_target_group(*target, dest);
```

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  create_snapshot(pair.second, dest);
+}
+
+bool instrument_spec_assignst::stack_allocated_is_tracked(
+  const symbol_exprt &symbol_expr) const
+{
+  conditional_target_exprt target{
+    true_exprt{}, symbol_expr, symbol_expr.source_location()};
+
+  return from_stack_alloc.find(target) != from_stack_alloc.end();
```
tautschnig: Is it really necessary to construct an expression this complicated just for the sake of a lookup? I'd expect the keys to be much more succinct (I'm particularly worried about the source location seemingly being required).

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+}
+
+void instrument_spec_assignst::invalidate_stack_allocated(
+  const symbol_exprt &symbol_expr,
+  goto_programt &dest)
+{
+  // ensure it is tracked
+  PRECONDITION_WITH_DIAGNOSTICS(
+    stack_allocated_is_tracked(symbol_expr),
+    "symbol is not tracked :" + from_expr(ns, "", symbol_expr));
```
tautschnig, line 62: I'd prefer the use of `format` (from `util/format_expr.h`) over `from_expr`. Ideally, we'd restrict `from_expr` to output that has to be language-specific.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+    true_exprt{}, symbol_expr, symbol_expr.source_location()};
+  const auto &car = car_map.find(target)->second;
+
+  goto_programt result;
+
+  auto t = result.add(goto_programt::make_assignment(
+    car.valid_var(), false_exprt{}, symbol_expr.source_location()));
+  disable_pointer_checks(t->source_location_nonconst());
+
+  add_pragma_disable_assigns_check(result);
```
tautschnig: It seems to different ways of adding `disable:...` pragma are being used here. Can we have a uniform style instead?

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  dest.destructive_append(result);
```
tautschnig: Shouldn't the above just use `dest.add(...)` instead of a temporary `result`?

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+    track_plain_spec_target(expr, dest);
+}
+
+void instrument_spec_assignst::track_stack_allocated(
+  const symbol_exprt &symbol_expr,
+  goto_programt &dest)
+{
+  // insert in tracking set
+  auto pair = get_or_insert(
+    true_exprt{}, symbol_expr, symbol_expr.source_location(), from_stack_alloc);
```
tautschnig: Should this rather be something like `const auto &bla = get_or_insert(...).second;`? I'd like to add that this isn't a great use of `auto`: it is not particularly obvious what the returned type is ("get_or_insert" makes me ask "what is it that I get?").

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  dest.destructive_append(result);
+}
+
+void instrument_spec_assignst::track_heap_allocated(
+  const exprt &expr,
+  goto_programt &dest)
+{
+  // insert in tracking set
+  auto pair =
+    get_or_insert(true_exprt{}, expr, expr.source_location(), from_heap_alloc);
```
tautschnig: As above: neither do you need/use the full pair, nor is "auto" very helpful.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+        // - a static lifetime
+        // - non empty location function
+        // - this function appears in the call graph of the function
+
+        // insert in tracking set
+        auto pair = get_or_insert(
+          true_exprt{}, sym.symbol_expr(), sym.location, from_stack_alloc);
+
+        // generate snapshot instructions for this target.
+        create_snapshot(pair.second, result);
```
tautschnig: Hmm, this entire code pattern appears to repeat, or should there be a variant of `create_snapshot` that takes more arguments?

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  if(has_subexpr(condition, ID_side_effect))
+    cleaner.clean(condition, result, st.lookup_ref(function_id).mode);
+
+  // create conditional address ranges by distributing the condition
+  for(const auto &target : group.targets())
+  {
+    // insert in tracking set
+    auto pair = get_or_insert(
+      condition, target, target.source_location(), from_spec_assigns);
+    const car_exprt &car = pair.second;
```
tautschnig: Is there any place in the code that uses the `.first` member of the result of `get_or_insert`? If not, perhaps `get_or_insert` should only return a const reference to the `car_exprt`?

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+    create_fresh_symbol("__car_lb", pointer_type(char_type()), source_location)
+      .symbol_expr();
+
+  const auto &upper_bound_var =
+    create_fresh_symbol("__car_ub", pointer_type(char_type()), source_location)
+      .symbol_expr();
+
+  if(target.id() == ID_pointer_object)
+  {
+    const auto &arg = target.operands().front();
```
tautschnig: I'd prefer: `const auto &arg = to_unary_expr(target).op();`

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+      upper_bound_var,
+      source_location};
+  }
+  else if(is_assignable(target))
+  {
+    const auto &size = size_of_expr(target.type(), ns);
+
+    INVARIANT(
+      size.has_value(),
+      "no definite size for lvalue target: \n" + target.pretty());
```
tautschnig: No need for a space before `\n`. (Those spaces at end-of-line are notorious for causing trouble when testing for the occurrence of some text in regression tests, where people might naively use something along the lines of `^no definite size for lvalue target:$`.)

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  result.add(
+    goto_programt::make_decl(car.upper_bound_var(), location_no_checks));
+
+  result.add(goto_programt::make_assignment(
+    car.upper_bound_var(),
+    simplify_expr(
+      plus_exprt{car.target_start_address(), car.target_size()}, ns),
+    location_no_checks));
+
+  add_pragma_disable_assigns_check(result);
```
tautschnig: Should that already be done for `location_no_checks` earlier to avoid fixing up the source location?

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  auto t = result.add(goto_programt::make_assertion(
+    target_validity_expr(car, allow_null_target), location_no_checks));
+
+  std::string comment = "Check that ";
+  comment += from_expr(ns, "", car.target());
+  comment += " is valid";
+  if(!car.condition().is_true())
+  {
+    comment += " when ";
+    comment += from_expr(ns, "", car.condition());
```
tautschnig, line 314: As above: `format` might be the better option than `from_expr`.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  }
+
+  t->source_location_nonconst().set_comment(comment);
+
+  add_pragma_disable_assigns_check(result);
+  t->source_location_nonconst().set_property_class("assigns");
+  dest.destructive_append(result);
```
tautschnig: It seems that just a single instruction is generated. It would, therefore, seem desirable to piece together the source location in full, and then construct the instruction directly in `dest` rather than the `source_location_nonconst()` dance plus the use of a temporary goto program.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+    comment += from_expr(ns, "", car.target());
+  }
+
+  comment += " is assignable";
+
+  t->source_location_nonconst().set_comment(comment);
+  t->source_location_nonconst().set_property_class("assigns");
+
+  add_pragma_disable_assigns_check(result);
+  dest.destructive_append(result);
```
tautschnig: As above: avoid this `source_location_nonconst()` dance and perhaps also the temporary goto program.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  // is redundant
+  return simplify_expr(
+    conjunction(
+      {candidate_car.valid_var(),
+       same_object(candidate_car.lower_bound_var(), car.lower_bound_var()),
+       less_than_or_equal_exprt{pointer_offset(candidate_car.lower_bound_var()),
+                                pointer_offset(car.lower_bound_var())},
+       less_than_or_equal_exprt{
+         pointer_offset(car.upper_bound_var()),
+         pointer_offset(candidate_car.upper_bound_var())}}),
```
tautschnig: You might use `and_exprt` directly, which happens to even have a four-operands constructor.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+}
+
+exprt instrument_spec_assignst::inclusion_check_full(
+  const car_exprt &car,
+  bool allow_null_lhs,
+  bool include_stack_allocated,
+  optionalt<cfg_infot> &cfg_info_opt) const
+{
+  bool no_targets = from_spec_assigns.empty() && from_heap_alloc.empty() &&
+                    (include_stack_allocated ? from_stack_alloc.empty() : true);
```
tautschnig: `!include_stack_allocated || from_stack_alloc.empty()` perhaps?

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+    bool is_stack_allocated =
+      from_stack_alloc.find(cond_target) != from_stack_alloc.end();
+
+    bool is_maybe_alive = true;
+    if(cfg_info_opt.has_value())
+    {
+      if(can_cast_expr<symbol_exprt>(cond_target.target()))
+        is_maybe_alive = cfg_info_opt.value().is_maybe_alive(
+          to_symbol_expr(cond_target.target()));
```
tautschnig: Braces around multi-line body, please. But then it might become a one-liner if you used `if(auto target = expr_try_dynamic_cast<symbol_exprt>(cond_target.target())`

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+    }
+
+    // skip dead targets and stack allocated if not they are not allowed
```
tautschnig: "if they are not" (spurious "not")

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  // disjunction over all allowed locations
+  exprt::operandst disjuncts;
+
+  // for all allowed locations
+  for(const auto &pair : car_map)
+  {
+    const conditional_target_exprt &cond_target = pair.first;
+
+    bool is_stack_allocated =
+      from_stack_alloc.find(cond_target) != from_stack_alloc.end();
```
tautschnig: How about doing an earlier `continue` here: `if(is_stack_allocated && !include_stack_allocated) continue;` to avoid an unnecessary `is_maybe_alive` check in that situation.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+    car_map.insert(
+      {*(insert.first),
+       create_car_expr(condition, target, target.source_location())});
+  }
+  else
+  {
+    log.warning() << "Ignored duplicate expression '"
+                  << from_expr(ns, target.id(), target)
+                  << "' in assigns clause spec at "
+                  << target.source_location().as_string() << messaget::eom;
```
tautschnig: To me, _this_ is a good use of `from_expr` for it is to reflect back to the user what they might have written.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
```diff
@@ -0,0 +1,559 @@
...
+  }
+
+  return {*(insert.first), car_map.find(*(insert.first))->second};
```
tautschnig: Hmm, can't the `find` be avoided in the `if` branch above for `car_map.insert` will return a useful iterator?

src/goto-instrument/contracts/instrument_spec_assigns.h
```diff
@@ -0,0 +1,354 @@
...
+  car_exprt(
+    const exprt &_condition,
+    const exprt &_target,
+    const exprt &_target_start_address,
+    const exprt &_target_size,
+    const symbol_exprt &_validity_var,
+    const symbol_exprt &_lower_bound_var,
+    const symbol_exprt &_upper_bound_var,
+    const source_locationt &_source_location)
+    : exprt()
```
tautschnig: The compiler will make this happen, no need for this explicit parent constructor call. What you might do, however, is:
`exprt(irep_idt{}, typet{}, {_condition, _target, _target_start_address, _target_size, _validity_var, _lower_bound_var, _upper_bound_var})`

src/goto-instrument/contracts/instrument_spec_assigns.h
```diff
@@ -0,0 +1,354 @@
...
+  const namespacet &ns;
+
+  /// Program symbol table
+  symbol_tablet &st;
+
+  /// Message handler
+  message_handlert &message_handler;
+
+  /// Logger
+  messaget log;
```
tautschnig: Why do you keep both of these?

src/goto-instrument/contracts/instrument_spec_assigns.h
```diff
@@ -0,0 +1,354 @@
...
+  ///  \param _function_id name of the instrumented function
+  ///  \param _functions other functions of the model
+  ///  \param _ns namespace of the goto_model
+  ///  \param _st symbol table of the goto_model
+  ///  \param _message_handler message handler used for warning/error messages
+  instrument_spec_assignst(
+    const irep_idt &_function_id,
+    const goto_functionst &_functions,
+    const namespacet &_ns,
+    symbol_tablet &_st,
```
tautschnig: Do you expect `_ns` to be the namespace wrapping `_st`? If so, I'd refrain from taking `_ns` here to avoid this implicit contract. Instead, construct your own `namespacet`.

src/goto-instrument/contracts/instrument_spec_assigns.h
```diff
@@ -0,0 +1,354 @@
...
+  conditional_target_sett from_spec_assigns;
+
+  /// Set of tracked conditional targets that come from stack-allocated objects
+  conditional_target_sett from_stack_alloc;
+
+  /// Set of tracked conditional targets that come from heap-allocated objects
+  conditional_target_sett from_heap_alloc;
+
+  using car_mapt = std::
+    unordered_map<const conditional_target_exprt, const car_exprt, irep_hash>;
```
tautschnig: I don't think the extra `const` on the map's key is necessary. (But then I've commented elsewhere that I'm not sure whether a `conditional_target_exprt` should be used as a key at all.)

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -177,137 +84,37 @@ static void havoc_if_valid(
...
-    side_effect_expr_nondett nondet(target.type(), source_location);
+    // ASSIGN *(target.type() *)__car_lb = nondet(car.target().type())
+    const auto &target_type = car.target().type();
+    side_effect_expr_nondett nondet(target_type, source_location);
     const auto &inst = dest.add(goto_programt::make_assignment(
-      dereference_exprt{target_snapshot_var}, nondet, source_location));
+      dereference_exprt{typecast_exprt::conditional_cast(
+        car.lower_bound_var(), pointer_type(target_type))},
+      nondet,
+      source_location));
```
tautschnig, line 94: I think we might be better off using `nondet_initializer` here for that would produce nondets for individual fields (based on the type). `goto-symex` could then take each of these in its field-sensitive encoding.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
```diff
@@ -177,137 +84,37 @@ static void havoc_if_valid(
...
-        mode);
-    }
-  }
-
-  dest.destructive_append(snapshot_program);
-  dest.destructive_append(havoc_program);
-}
+  dest.add(
+    goto_programt::make_dead(car.upper_bound_var(), source_location_no_checks));
+}
```
tautschnig: Github says there's no newline at end-of-file.

src/goto-instrument/contracts/havoc_assigns_clause_targets.h
```diff
@@ -49,17 +53,61 @@ class message_handlert;
...
+class havoc_assigns_clause_targetst : public instrument_spec_assignst
+{
+public:
+  havoc_assigns_clause_targetst(
+    const irep_idt &_function_id,
+    const std::vector<exprt> &_targets,
+    const goto_functionst &_functions,
+    const source_locationt &_source_location,
+    namespacet &_ns,
+    symbol_tablet &_st,
```
tautschnig: As said in another comment: don't take both a `namespacet` and a `symbol_tablet` (and essentially never take a non-const `namespacet`).

src/goto-instrument/contracts/contracts.cpp
```diff
@@ -1407,15 +1333,13 @@ void code_contractst::check_frame_conditions(
...
-      const auto havoc_argument =
-        pointer_object(instruction_it->get_other().operands().front());
-      add_inclusion_check(
-        body,
-        assigns,
-        instruction_it,
-        havoc_argument,
-        allow_null_lhst::NO,
-        cfg_info_opt);
+      auto havoc_argument = instruction_it->get_other().operands().front();
```
tautschnig, line 1335: You probably want `to_unary_expr(instruction_it->get_other()).op()`

src/goto-instrument/Makefile
```diff
@@ -16,9 +16,9 @@ SRC = accelerate/accelerate.cpp \
       alignment_checks.cpp \
       branch.cpp \
       call_sequences.cpp \
-      contracts/assigns.cpp \
       contracts/contracts.cpp \
       contracts/havoc_assigns_clause_targets.cpp \
+      contracts/instrument_spec_assigns.cpp \
```
tautschnig, line 21: This should be part of the commit that removed the old file so as not to break `git bisect` applications.

tautschnig (approved):
Looks ok in general, but a whole bunch of comments on individual bits of code.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: I specialized the lookup tables to have more specific key types depending on context, which caused some code inflation.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: All I see in `format_expr.h` is `std::ostream & format_rec(std::ostream &os, const expr &src)` which is barely used in the whole codebase (28 uses), however I see a lot more uses of `from_expr()` (139 occurences).

Also `format_rec` renders directly to ostreams so `from_expr` seems necessary when rendering expressions to `irep_ids` to generate comments, or precondition/invariant message strings. What should I use to render to a string in a language agnostic way ?

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: Fixed to use the correct location by construction

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: create snapshot is pretty much context agnostic so I don't think so.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: Which `format` is it then ? All I could find in `format_expr.h` was `format_rec` which works on `std::ostreams`

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: This went away when changing the cache representation

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: This went away when changing the cache representation

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: This went away when changing the cache representation

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: This went away when changing the cache representation

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: All comments or warning/error messages are eventually printed back to the user so following that logic `from_expr` would make sense everywhere, wouldn't it ?

src/goto-instrument/contracts/instrument_spec_assigns.h
remi-delmas-3000: Good point.

src/goto-instrument/contracts/instrument_spec_assigns.h
remi-delmas-3000: I did not realise the  message_handler was not used elsewhere.

src/goto-instrument/contracts/instrument_spec_assigns.h
remi-delmas-3000: I need to make sure that things stored in this map will never get modified.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: What is the relation between `side_effect_expr_nondett` (which also carries type information) and this `nondet_initializer` class ? It seemed to me that the latter was used to lower the former to basic GOTO before symex.

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
remi-delmas-3000: Fixed

remi-delmas-3000 (commented):
I addressed all comments except the one regarding `from_expr` and the use of `nondet_initializer`.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
remi-delmas-3000: Also, these comments are going to be shown to the user in the console output and logs, so as you mentioned below using the same syntax as the source code could be a plus.

src/goto-instrument/contracts/instrument_spec_assigns.cpp
tautschnig: Ah, my apologies for the lack of clarity. It's: `stream << format(expr);` or `string = format_to_string(expr);`

src/goto-instrument/contracts/havoc_assigns_clause_targets.cpp
tautschnig: The latter will use `side_effect_expr_nondett` by either returning exactly such (for all basic types), or by creating a compound object that in turn contains `side_effect_expr_nondett` (for example a struct expression with a nondet for each member).

tautschnig:
The change to the `Makefile` requires approval from one of { @peterschrammel, @martin-cs, @chrisr-diffblue }.

chrisr-diffblue (approved):
I've only reviewed the Makefile changes - so consider my approval as mainly for that.

PR #6578, opened 2022-01-13
Actually test linker script processing

tautschnig:
The goto-cc personality does not invoke linker script processing, hence
the tests located in the ansi-c folder were a no-op. Moving them to the
goto-gcc folder showed that they wouldn't work as is, which is also
fixed now:

1) Add the -N option so as not to generate ELF headers, which those
linker scripts leave no room for.
2) Make goto-cc accept (but ignore) the -N option. (Cosmetic change
only.)
3) Actually detect failure of linker script parsing by enabling warnings
and testing for them.
4) Make sure ls_parse.py is available on the PATH.


chrisr-diffblue (approved):
Good catch!

PR #6581, opened 2022-01-13
bitreverse_exprt: Expression to reverse the order of bits

tautschnig:
Clang has a __builtin_bitreverse (and at some point GCC might as well:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=50481).

Creating this as a draft: the actual implementation in the back-end is yet to be added, as are tests.

### DIFF
--- a/src/ansi-c/c_typecheck_expr.cpp
+++ b/src/ansi-c/c_typecheck_expr.cpp
@@ -3204,6 +3204,28 @@ exprt c_typecheck_baset::do_special_functions(
   {
     return typecheck_builtin_overflow(expr, ID_mult);
   }
+  else if(
+    identifier == "__builtin_bitreverse8" ||
+    identifier == "__builtin_bitreverse16" ||
+    identifier == "__builtin_bitreverse32" ||
+    identifier == "__builtin_bitreverse64")
+  {
+    // clang only
+    if(expr.arguments().size() != 1)
+    {
+      std::ostringstream error_message;
+      error_message << expr.source_location().as_string()
+                    << ": error: " << identifier << " expects one operand";
+      throw invalid_source_file_exceptiont{error_message.str()};
+    }
+
+    typecheck_function_call_arguments(expr);
+
+    bitreverse_exprt bitreverse{expr.arguments()[0]};
+    bitreverse.add_source_location() = source_location;
+
+    return std::move(bitreverse);
+  }
   else
     return nil_exprt();
   // NOLINTNEXTLINE(readability/fn_size)
--- a/src/ansi-c/expr2c.cpp
+++ b/src/ansi-c/expr2c.cpp
@@ -3496,6 +3496,22 @@ std::string expr2ct::convert_conditional_target_group(const exprt &src)
   return dest;
 }
 
+std::string expr2ct::convert_bitreverse(const bitreverse_exprt &src)
+{
+  if(auto type_ptr = type_try_dynamic_cast<unsignedbv_typet>(src.type()))
+  {
+    const std::size_t width = type_ptr->get_width();
+    if(width == 8 || width == 16 || width == 32 || width == 64)
+    {
+      return convert_function(
+        src, "__builtin_bitreverse" + std::to_string(width));
+    }
+  }
+
+  unsigned precedence;
+  return convert_norep(src, precedence);
+}
+
 std::string expr2ct::convert_with_precedence(
   const exprt &src,
   unsigned &precedence)
@@ -3903,6 +3919,9 @@ std::string expr2ct::convert_with_precedence(
   {
     return convert_conditional_target_group(src);
   }
+
+  else if(src.id() == ID_bitreverse)
+    return convert_bitreverse(to_bitreverse_expr(src));
 
   auto function_string_opt = convert_function(src);
   if(function_string_opt.has_value())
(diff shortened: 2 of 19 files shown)


src/ansi-c/c_typecheck_expr.cpp
```diff
@@ -3238,6 +3238,27 @@ exprt c_typecheck_baset::do_special_functions(
...
+    identifier == "__builtin_bitreverse16" ||
+    identifier == "__builtin_bitreverse32" ||
+    identifier == "__builtin_bitreverse64")
+  {
+    // clang only
+    if(expr.arguments().size() != 1)
+    {
+      error().source_location = f_op.source_location();
+      error() << identifier << " expects one operand" << eom;
+      throw 0;
```
chrisr-diffblue: Throw a structured exception?
```suggestion
       std::ostringstream error_message;
      error_message << expr.source_location().as_string() << ": " << identifier
                    << " expects expects one operand";
      throw invalid_source_file_exceptiont{error_message.str()};
```

src/ansi-c/expr2c.cpp
```diff
@@ -3496,6 +3496,22 @@ std::string expr2ct::convert_conditional_target_group(const exprt &src)
...
+std::string expr2ct::convert_bitreverse(const bitreverse_exprt &src)
+{
+  if(auto type_ptr = type_try_dynamic_cast<unsignedbv_typet>(src.type()))
+  {
+    const std::size_t width = type_ptr->get_width();
+    if(width == 8 || width == 16 || width == 32 || width == 64)
+    {
+      return convert_function(
+        src, "__builtin_bitreverse" + std::to_string(width));
+    }
```
chrisr-diffblue, line 3508: Belt-and-braces, add an `UNREACHABLE` or an assertion to be more descriptive, instead of using `convert_norep`? Or maybe just have a `PRECONDITION` ?

chrisr-diffblue (commented):
Generally looks great to me. Given the effort that went into removing unstructure exceptions in the error reporting previously, it seems a shame to be adding a `throw 0` back in here - so could I kindly ask if that could be replaced with a structured exception? It'd be an approval from me if that were done :-)

src/ansi-c/expr2c.cpp
tautschnig: Hmm, there isn't really anything fundamentally wrong in creating a `bitreverse_exprt` with a type that isn't 8/16/32/64 bits wide, we just don't have a good C representation available. So I'd stick with `convert_norep` as that's what the remainder of this code does, if that's ok?

src/ansi-c/c_typecheck_expr.cpp
tautschnig: Done, but I really wish we had #5837 (with its two dependencies) in place to make things even more structured!

tautschnig: And now also includes a test for this one :-)

src/ansi-c/expr2c.cpp
chrisr-diffblue: Fine by me - it was just a sanity check query from me, thanks.

chrisr-diffblue (approved):
Many thanks for the updates - looks great to me... And given that I poked a raw nerve on the structured exception point - feel free to assign any PRs that need reviews/comments on that front to me. 👍

src/solvers/flattening/boolbv.cpp
```diff
@@ -155,11 +155,14 @@ bvt boolbvt::convert_bitvector(const exprt &expr)
...
-  else if(expr.id()==ID_bitnot || expr.id()==ID_bitand ||
-          expr.id()==ID_bitor || expr.id()==ID_bitxor ||
-          expr.id()==ID_bitxnor || expr.id()==ID_bitnor ||
-          expr.id()==ID_bitnand)
+  else if(
+    expr.id() == ID_bitnot || expr.id() == ID_bitand || expr.id() == ID_bitor ||
+    expr.id() == ID_bitxor || expr.id() == ID_bitxnor ||
+    expr.id() == ID_bitnor || expr.id() == ID_bitnand ||
+    expr.id() == ID_bitreverse)
+  {
```
kroening: bit-reversing is not a bit-wise operation. Bit-wise operations are functions on individual bits, not bit-vectors. I'd stick this into a separate file.

src/util/bitvector_expr.cpp
```diff
@@ -142,3 +142,17 @@ exprt count_trailing_zeros_exprt::lower() const
...
+{
+  const std::size_t int_width = to_bitvector_type(type()).get_width();
+
+  exprt::operandst result_bits;
+  result_bits.reserve(int_width);
+
+  const exprt operand = op();
+  for(std::size_t i = 0; i < int_width; ++i)
+    result_bits.push_back(extractbit_exprt{operand, i});
+
```
kroening, line 156: This is potentially problematic when the lowered expression is passed to any piece of code that doesn't cope well with the sharing. This is worth at least a warning, and it may possibly make sense to fix this issue by introducing a `let_exprt` for `op()`.

src/solvers/flattening/boolbv.cpp
tautschnig: Done!

src/util/bitvector_expr.cpp
tautschnig: I've made a stab at it - is the second commit that I added what you were looking for?

kroening: yes!

kroening (approved):

NlightNFotis (approved):

PR #6584, opened 2022-01-17
Fix `irept` printing in failed unit tests

thomasspriggs:
The operator needed for the printing is defined in `unit_tests.cpp`, but it  also needs to be forward declared for the catch framework to find and  use it, instead of printing ireps as `{?}`.

A regression test of a failing unit test is included in this PR to ensure that this functionality for fault finding of failing unit tests works as intended.

### DIFF
--- a/regression/catch-framework/Makefile
+++ b/regression/catch-framework/Makefile
@@ -1,0 +1,9 @@
+default: test
+
+test:
+	@../test.pl -e -p -c "../../../unit/unit_tests"
+
+tests.log: ../test.pl test
+
+clean:
+	$(RM) tests*.log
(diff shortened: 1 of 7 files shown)


regression/catch-framework/Makefile
```diff
@@ -0,0 +1,12 @@
+default: test
+
+include ../../src/config.inc
+include ../../src/common
```
tautschnig: Nit picking: I don't think you actually need these here.

tautschnig (approved):

TGWDB (approved):

regression/catch-framework/Makefile
thomasspriggs: Removed.

PR #6590, opened 2022-01-18
Permit initializers in library functions

tautschnig:
The goal of this series of commits is to reduce the number of `__CPROVER_...` static-lifetime objects that appear in every program irrespective of their use by C (model) library functions. This will limit verification overhead to those cases where they are actually used. This will also enable more costly objects (such as unbounded arrays) to be used by future library instrumentation.


src/ansi-c/library/pthread_lib.c
```diff
@@ -290,12 +290,15 @@ inline int pthread_mutex_destroy(pthread_mutex_t *mutex)
...
 
-extern __CPROVER_bool __CPROVER_threads_exited[];
-extern __CPROVER_thread_local unsigned long __CPROVER_thread_id;
-
-extern __CPROVER_thread_local const void *__CPROVER_thread_keys[];
-extern __CPROVER_thread_local void (*__CPROVER_thread_key_dtors[])(void *);
-extern __CPROVER_thread_local unsigned long __CPROVER_next_thread_key;
+__CPROVER_bool __CPROVER_threads_exited[__CPROVER_constant_infinity_uint];
+__CPROVER_thread_local unsigned long __CPROVER_thread_id = 0;
+#if 0
```
chrisr-diffblue, line 296: Any particular reason for the `#if 0` rather than deletion? If theres a reason (other than "its a work in progress" :-) ) I'd suggest adding a comment to explain the reason.

src/ansi-c/library/pthread_lib.c
```diff
@@ -533,12 +540,17 @@ inline int pthread_rwlock_wrlock(pthread_rwlock_t *lock)
...
-extern __CPROVER_thread_local unsigned long __CPROVER_thread_id;
-
-extern __CPROVER_thread_local const void *__CPROVER_thread_keys[];
-extern __CPROVER_thread_local void (*__CPROVER_thread_key_dtors[])(void *);
-extern __CPROVER_thread_local unsigned long __CPROVER_next_thread_key;
+__CPROVER_bool __CPROVER_threads_exited[__CPROVER_constant_infinity_uint];
+#ifndef LIBRARY_CHECK
+__CPROVER_thread_local unsigned long __CPROVER_thread_id = 0;
+#endif
+#if 0
```
chrisr-diffblue, line 551: Ditto here and the other places in this file.

chrisr-diffblue (approved):
This seems reasonable and sensible to me.

src/ansi-c/library/pthread_lib.c
tautschnig: Comments added in all the places that have `#if 0`: thread key destructors are expensive. We'll have to figure out a viable solution at some point.

chrisr-diffblue: Thanks, much appreciated :-) I always hate encountering an `#if 0` with not context as to why its there, whether it works if re-enabled, etc.

chris-ryder (approved):

src/goto-programs/link_to_library.cpp
```diff
@@ -25,24 +79,27 @@ Author: Daniel Kroening, kroening@kroening.com
 void link_to_library(
   goto_modelt &goto_model,
   message_handlert &message_handler,
-  const std::function<
-    void(const std::set<irep_idt> &, symbol_tablet &, message_handlert &)>
-    &library)
+  const std::function<void(
+    const std::set<irep_idt> &,
+    const symbol_tablet &,
+    symbol_tablet &,
```
remi-delmas-3000, line 87: Can we document the two symbol table parameters and what they are supposed to be set to?

src/goto-programs/link_to_library.cpp
```diff
@@ -25,24 +79,27 @@ Author: Daniel Kroening, kroening@kroening.com
...
+    message_handlert &)> &library)
 {
   // this needs a fixedpoint, as library functions
   // may depend on other library functions
 
   std::set<irep_idt> added_functions;
+  replace_symbolt::expr_mapt object_type_updates;
```
remi-delmas-3000, line 97: Could you explain the role of this new map with a comment ?

src/goto-programs/link_to_library.cpp
```diff
@@ -59,30 +116,44 @@ void link_to_library(
         // already added
       }
       else
-        missing_functions.insert(id);
+      {
+        changed = true;
+        added_functions.insert(id);
+
```
remi-delmas-3000, line 127: How about reformulating the control flow structure as
```c
if(body_available || already_added)
 continue;

// add the symbol
changed = true;
added_functions.insert(id);
...
 
```

src/goto-programs/link_to_library.cpp
```diff
@@ -11,9 +11,63 @@ Author: Daniel Kroening, kroening@kroening.com
...
+  }
+
+  // check whether additional initialization may be required
+  if(
+    goto_model.goto_functions.function_map.find(INITIALIZE_FUNCTION) !=
+    goto_model.goto_functions.function_map.end())
+  {
+    for(const auto &entry : library_model.symbol_table)
+    {
+      if(
```
remi-delmas-3000, line 58: if a same static variable is declared more than once in the library code, do we check that the initial value for the symbol is the same at each declaration ?

Eventhough each declaration follows the pattern below, depending on which one gets pulled first results could be different:

```c
/* FUNCTION : foo */
#ifndef blah_is_defined
#define blah_is_defined
static int blah = 12;
#endif

int foo() 
{
  // uses blah
}

/* FUNCTION : bar */
#ifndef blah_is_defined
#define blah_is_defined
static int blah = 0; // a different initial value
#endif

int bar() {
  // uses blah
}
```

src/goto-programs/link_to_library.cpp
tautschnig: Could do, but then I'd have to touch more code in this PR.

src/goto-programs/link_to_library.cpp
tautschnig: Documentation added.

src/goto-programs/link_to_library.cpp
tautschnig: Done.

src/goto-programs/link_to_library.cpp
tautschnig: The call to `link_goto_model` will yield a warning in such a case (which I have just tested to make sure this really works as intended).

peterschrammel (approved):

PR #6593, opened 2022-01-18
Remove useless system_library_symbolst::is_type_internal

tautschnig:
Passing a symbolt that only has its "type" member set to
is_symbol_internal_symbol is useless as that function never even looks
at the type. Although fixing is_type_internal was possible, it turns out
to be completely unnecessary as the single user (dump-c) appears to work
just fine as demonstrated in the included test case.


chrisr-diffblue (approved):
Not entirely sure what the purpose of the original code was, and git blame doesn't appear to help me too much either... but I think its removal is likely safe.

chris-ryder (approved):

peterschrammel (approved):

